Issue #98, opened 2015-08-12
Refactor BibTeX file field


koppor:
JabRef should be convenient for new users to use. When using the `file` field, the format for each file entry is `description:link:type`. Multiple values can be put in separated by `;`. At each file, fields can be omitted. `file = {desc}` is a valid enry. Naively, I would have assumed that `desc` is the link. Furthermore `type` is a custom JabRef string, but [Media Types](http://www.iana.org/assignments/media-types/media-types.xhtml) are the accepted standard for providing types.

Therefore, I propose the new format `link:description:media type`.

Furthermore, I propose dropping support for the legacy fields pdf and ps. These fields were [legacy from the end of 2007 on](https://sourceforge.net/p/jabref/mailman/message/1369285/). Even though some users want to keep the old fields, I would vote for an intuitive handling of references to external files instead of clutter the UI with legacy things. (I currently can't find the email in the mailinglist where we discussed that :()

## Discussion points

- Is media type too over-engineered? Can we drop the type completely and derive the type by the  file extension?
- Are `:` and `;` good separators?
- Is `file` really more intuitive than separate `pdf` and `ps` links? Should we keep the `pdf` field?
- Should the new field be called `attachments` to enable easy migration from the old format to the new one?

## Next steps

I would propose a discussion at this issue and I'll update this text accordingly.

We should also remove "Tools -> Legacy Tools...".

## More to read

- Old style: http://jabref.sourceforge.net/help/ExternalFiles.php
- Current style: http://jabref.sourceforge.net/help/FileLinks.php

mlep:
About "Should we keep the pdf field?", I would say no IF it becomes possible to define several custom general fields to manage files (BTW: this is a registered feature request). Hence, users who want to have a PDF field and a PS field and a DVI field, etc. can get it.

koppor:
Related feature request: https://sourceforge.net/p/jabref/feature-requests/597/
(I didn't find the other @mlep mentioned, maybe someone else can?)

mlep:
Sorry for not taking the time to point to the specific feature requests. See:
http://sourceforge.net/p/jabref/feature-requests/282/
http://sourceforge.net/p/jabref/feature-requests/129/
http://sourceforge.net/p/jabref/feature-requests/163/
http://sourceforge.net/p/jabref/feature-requests/64/

koppor:
Also related: [DOS path in link](https://sourceforge.net/p/jabref/discussion/318825/thread/fa24606f/)

[...]

PR #251, opened 2015-10-22
Removes legacy options to sync files in the "pdf" or "ps" field

simonharrer:
Why?

- these legacy actions are replaced by using the "file" field and a much more sophisticated file sync action.


matthiasgeiger:
:+1:

koppor:
Relates to #98

PR #252, opened 2015-10-22
Update supported lookAndFeels

matthiasgeiger:
I propose to update the list of supported lookAndFeels:

- The FQCN for the `jgoodies` LaFs were wrong
- the "`javax.swing.plaf.mac.MacLookAndFeel`" is not available on Windows and not chosable on Macs either (as the whole input field is hidden on a Mac (see https://github.com/JabRef/jabref/blob/1a0d23b64efa72f5184ad7f4100a428b7cc06040/src/main/java/net/sf/jabref/gui/preftabs/AdvancedTab.java#L123-L148)
- `com.sun.java.swing.plaf.motif.MotifLookAndFeel` looks like this: 
jabref_metal


simonharrer:
:+1:

PR #253, opened 2015-10-22
Cleanup preferences

lenhard:
This PR remove preferences that are no longer used in JabRef.

I checked the usage each public field of `JabRefPreferences` and, if there was no such usage, searched for String occurrences of the preference value in the complete project. If such usage was found, I replaced it with a usage of the respective field. Otherwise, I deleted the preference.

All in all, only a surprisingly small amount of preferences is not used.


simonharrer:
:+1:

PR #255, opened 2015-10-23
Added JComboBox in MassSetFieldDialog

oscargus:
Added a JComboCox for "field" in MassSetFieldDialog (Set/clear/rename field in right-click menu).

Seemed to be quite convenient compared to typing the field name.

### DIFF
--- a/src/main/java/net/sf/jabref/util/MassSetFieldAction.java
+++ b/src/main/java/net/sf/jabref/util/MassSetFieldAction.java
@@ -1,4 +1,4 @@
-/*  Copyright (C) 2003-2011 JabRef contributors.
+/*  Copyright (C) 2003-2015 JabRef contributors.
     This program is free software; you can redistribute it and/or modify
     it under the terms of the GNU General Public License as published by
     the Free Software Foundation; either version 2 of the License, or
@@ -20,6 +20,7 @@ import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.TreeSet;
 
 import javax.swing.*;
 import javax.swing.event.ChangeEvent;
@@ -47,7 +48,6 @@ import net.sf.jabref.model.entry.BibtexEntry;
  */
 public class MassSetFieldAction extends MnemonicAwareAction {
 
-    private static final long serialVersionUID = -3335783568376999899L;
     private final JabRefFrame frame;
     private JDialog diag;
     private JRadioButton all;
@@ -55,11 +55,12 @@ public class MassSetFieldAction extends MnemonicAwareAction {
     private JRadioButton clear;
     private JRadioButton set;
     private JRadioButton rename;
-    private JTextField field;
+    private JComboBox<String> field;
     private JTextField text;
     private JTextField renameTo;
     private boolean cancelled = true;
     private JCheckBox overwrite;
+    private TreeSet<String> allFields;
 
 
     public MassSetFieldAction(JabRefFrame frame) {
@@ -70,7 +71,8 @@ public class MassSetFieldAction extends MnemonicAwareAction {
     private void createDialog() {
         diag = new JDialog(frame, Localization.lang("Set/clear/rename fields"), true);
 
-        field = new JTextField();
+        field = new JComboBox<>();
+        field.setEditable(true);
         text = new JTextField();
         text.setEnabled(false);
         renameTo = new JTextField();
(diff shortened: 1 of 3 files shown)


src/main/java/net/sf/jabref/util/MassSetFieldAction.java
```diff
@@ -85,6 +87,18 @@ private void createDialog() {
         set = new JRadioButton(Localization.lang("Set fields"));
         rename = new JRadioButton(Localization.lang("Rename field to") + ":");
         rename.setToolTipText(Localization.lang("Move contents of a field into a field with a different name"));
+        
+        Collection<BibtexEntry> be = frame.basePanel().database().getEntries();
+        allFields = new TreeSet<String>();
+        for (BibtexEntry e : be) {
+            allFields.addAll(e.getAllFields());
+        }
```
simonharrer: Can you extract the logic of getting all the fields of a database and move this method in the database class itself? Maybe use getAllVisibleFields as a method, then you can also ignore the ones that start with "__".

src/main/java/net/sf/jabref/util/MassSetFieldAction.java
```diff
@@ -150,7 +164,7 @@ public void stateChanged(ChangeEvent e) {
             public void actionPerformed(ActionEvent e) {
                 // Check if the user tries to rename multiple fields:
                 if (rename.isSelected()) {
-                    String[] fields = getFieldNames(field.getText());
+                    String[] fields = getFieldNames((String) field.getSelectedItem());
```
simonharrer: hm, strange that a generically typed combo box requires a cast to String. Is this really necessary?

koppor: Yeah, it is by design as arbitrary objects can be inserted if the ComboBox is editable: http://stackoverflow.com/a/7026724/873282

Solution: Replace the code by 

    field.getItemAt(field.getSelectedIndex())

(or add a comment pointing to the stackoverflow answer :innocent:)

simonharrer: Okay, so we can leave it at this.

oscargus:
As you may have noted, I have used the new Eclipse settings which apparently reformats everything when saving...

koppor:
When using git gui, you can choose the lines to commit. Just mark them and right click and   "stage lines to commit". Would that be possible? After committing: Ctrl r and revert the other changes!

Von: "Oscar Gustafsson" <notifications@github.com>
Gesendet: ‎24.‎10.‎2015 20:05
Betreff: Re: [jabref] Added JComboBox in MassSetFieldDialog (#255)

As you may have noted, I have used the new Eclipse settings which apparently reformats everything when saving...

simonharrer:
Hm, the new eclipse settings seem to be wrong. We should revert to the old ones as they cause a lot of havoc at the moment. See the strange indents in this commit.

koppor:
I tried with Eclipse Mars SR1 and it did NOT format it in a wrong way. I checked out this branch and did `CTRL+a` and `CTRL+SHIFT+f` at `BibtexDatabase` and it fixed everything.

@oscargus: which Eclipse version do you use? Mine is as follows

    Version: Mars.1 Release (4.5.1)
    Build id: 20150924-1200

oscargus:
I use the same version. By looking at the code it seemed like it was the `listener` variable that messed things up. I moved it to the end and now it looks much better. A bit strange since `listener` is also "correctly" formatted now.

koppor:
Could you switch the two commits `Move listener in the code ...` and `Added method`? Then we can properly read the diffs.

Side note: I tried to do CTRL+a and CTRL+f in Eclipse. Then it formatted correctly. Only the save actions seem to mess up the code. :cry:

oscargus:
I'll try that. rebase -i, right? Or should I simply just merge the commit
with the strange formatting and the better formatting?

koppor:
`rebase -i` should work. Think, you'll get many conflicts. Maybe you just keep the first commit only (meaning deleting the `pick` lines for the other two commits) and then do the other commits manually again.

Two options for that:
1. Copy the file to a temp folder and copy'n'paste the method during editing
2. Use a [second worktree](https://github.com/blog/2042-git-2-5-including-multiple-worktrees-and-triangular-workflows) and checkout the last commit there
3. Use `gitk --all&` and copy'n'paste the change from there. You have to remove the `-` though.

I personally do mostly option 3, sometimes option 2.

simonharrer:
you can also use `git reset HEAD~2` and then create a new commit.

PR #256, opened 2015-10-23
Bug fix in color preferences, replaced DefaultFormBuilder and clean ups

oscargus:
Fixed a bug in color preferences recently introduced.

Replaced the deprecated DefaultFormBuilder by FormBuilder in DatabasePropertiesDialog.

Removed some casts which weren't required.


koppor:
:+1:

PR #257, opened 2015-10-25
Improve About.html files

koppor:
* remove local license files in the help
  * replace License.html by online link
* replace copyright statements of used libraries by a link to [external-libraries.txt](https://github.com/JabRef/jabref/blob/master/external-libraries.txt)
* external links open in browser
* preview of link in status bar

Future work: Write a script converting external-libraries.txt to the About.html files. Thereby, the concrete license files of each library have to be put into the About.html files, too.


simonharrer:
:+1: Especially not having to translate the used libraries section is really helpful, I think. And only linking instead of shipping everything with it.

koppor:
After discussing, we should create a folder `licenses`, where each license file is stored with `id.txt`, where `id` is the id from `external-libraries.txt`. Then the build script should create the About.html txt content based on `external-libraries.txt`. Then, this conforms to the requirements of the BSD license.

koppor:
22c809c adds license information in `licenses/` and reproduces that information in About.html. Thus, we should comply with the license requirements.

Issue #244, opened 2015-10-16
2.11beta5 & xp: does not start...


mlep:
Hello, 

On my xp virtual machine, JabRef-2.11b5 does not start. No window is displayed.
Here are  the messages:

C:\test>java -jar JabRef-2.11b5.jar
Oct 16, 2015 9:04:31 AM org.java.plugin.registry.xml.ManifestParser <init>
INFO: got SAX parser factory - com.sun.org.apache.xerces.internal.jaxp.SAXParser
FactoryImpl@132ca2b
Oct 16, 2015 9:04:31 AM org.java.plugin.registry.xml.PluginRegistryImpl configur
e
INFO: configured, stopOnError=false, isValidating=true
Oct 16, 2015 9:04:31 AM org.java.plugin.registry.xml.PluginRegistryImpl register

INFO: plug-in and fragment descriptors registered - 2
Oct 16, 2015 9:04:31 AM net.sf.jabref.plugin.PluginCore initialize
INFO: Found 2 plugin(s):
  - net.sf.jabref.core (jar:file:/C:/Documents%20and%20Settings/fdarboux/Bureau/
JabRef-2.11b5.jar!/plugins/net.sf.jabref.core/plugin.xml)
  - net.sf.jabref.export.misq (jar:file:/C:/Documents%20and%20Settings/fdarboux/
Bureau/JabRef-2.11b5.jar!/plugins/net.sf.jabref.export.misq/plugin.xml)

Oct 16, 2015 9:04:31 AM org.java.plugin.standard.StandardPluginManager activateP
lugin
INFO: plug-in started - net.sf.jabref.core@2.0.0.4b1 (active/total: 1 of 2)
Oct 16, 2015 9:04:31 AM org.java.plugin.standard.StandardPluginManager activateP
lugin
INFO: plug-in started - net.sf.jabref.export.misq@1.0.0 (active/total: 2 of 2)
Exception in thread "main" java.lang.UnsatisfiedLinkError: Error looking up func
tion 'SetCurrentProcessExplicitAppUserModelID': La procÚdure spÚcifiÚe est intro
uvable.

        at com.sun.jna.Function.<init>(Function.java:208)
        at com.sun.jna.NativeLibrary.getFunction(NativeLibrary.java:536)
        at com.sun.jna.NativeLibrary.getFunction(NativeLibrary.java:513)
        at com.sun.jna.NativeLibrary.getFunction(NativeLibrary.java:499)
[...]

PR #224, opened 2015-10-09
Only show automated list of contributors

stefan-kolb:
Maintaining a list of active developers is tedious and not necessarily fair.
All authors are automatically included in the contributors list (if we regenerate it often enough).
I know this is not an easy decision especially for those that contributed more than others and or are/were project leads. Maybe we can add these people exclusively again (as maintainers or the like).

PR #258, opened 2015-10-25
Show developers

koppor:
As discussed in #244, especially at https://github.com/JabRef/jabref/pull/224#issuecomment-147650492, it would be good to show the currently active developers.

JabRef offers the team "developers", which are the currently developers of JabRef. There is also the organization, but these guys are more involved in the overall steering of JabRef and related projects (homepage, journal lists, ...) and not the concrete development and reviews of PRs in the JabRef code.

Similar to the AUTHORs list, I keep the alphabetical order. This is untypical, but more easy to create.

Side note: You have to have `GRADLE_OPTS` to be set to `-Dfile.encoding=UTF-8` on Windows to have the gradle build working correctly. See [README.md](https://github.com/JabRef/jabref/blob/master/README.md#building-jabref-from-source).

### DIFF
--- a/DEVELOPERS
+++ b/DEVELOPERS
@@ -1,0 +1,7 @@
+# This file lists all individuals being part of JabRef's developers team
+# For how it is generated, see `scripts/generate-developers.sh`.
+Jörg Lenhard
+Matthias Geiger
+Oliver Kopp
+Simon Harrer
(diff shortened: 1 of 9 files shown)


simonharrer:
:+1:

PR #259, opened 2015-10-25
Fixed some warnings, missing copyright statements, and some minor issues

oscargus:
PushToEmacs is causing formatting problems (works with manual formatting), but other than that, I believe that everything should be OK.

### DIFF
--- a/src/main/java/net/sf/jabref/bibtex/DuplicateCheck.java
+++ b/src/main/java/net/sf/jabref/bibtex/DuplicateCheck.java
@@ -1,4 +1,4 @@
-/*  Copyright (C) 2003-2011 JabRef contributors.
+/*  Copyright (C) 2003-2015 JabRef contributors.
     This program is free software; you can redistribute it and/or modify
     it under the terms of the GNU General Public License as published by
     the Free Software Foundation; either version 2 of the License, or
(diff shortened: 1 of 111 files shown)


simonharrer:
:+1:

koppor:
@simonharrer My comment on the else branches got lost. If following happens

    if (somecondition) {
        return a;
    } else {
        return b;
    }

Eclipse complains that it can be rewritten as


    if (somecondition) {
        return a;
    }
    return b;

Do we want that style? I like the first one more. If we decide for the first one, I'll turn off the Eclipse warning suggesting the second style.

simonharrer:
It depends. the second style is perfect when the if checks some error condition or some very special/minor case. The first style is perfect when both results are equally (likely/ok/...), are really two parallel cases being equal in their importance.

My proposal: disable the warning that this needs to be changed as this cannot be decided using the AST of Java. Only a developer can decide this.

koppor:
I disabled the warning in `master`. @oscargus Could you regenerate the Ecipse config

To save time, I would propose to merge this PR (instead of undoing the `else` branch changes), but not to do other else branch rewritings in the future. @simonharrer If OK, go ahead with merging.

PR #261, opened 2015-10-26
[WIP] Enable merge entries at context menu

koppor:
When selecting two entries, the context menu should offer merging of them, too.

- [ ] When only one entry is selected, the "Merge entries" entry should be disabled.

### DIFF
--- a/src/main/java/net/sf/jabref/gui/JabRefFrame.java
+++ b/src/main/java/net/sf/jabref/gui/JabRefFrame.java
@@ -394,7 +394,8 @@ public class JabRefFrame extends JFrame implements OutputPrinter {
             prefs.getKey(KeyBinds.CLEANUP),
             IconTheme.JabRefIcon.CLEANUP_ENTRIES.getIcon());
 
-    private final AbstractAction mergeEntries = new GeneralAction(Actions.MERGE_ENTRIES,
+    // used by RightClickMenu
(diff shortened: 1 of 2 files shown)


simonharrer:
Good idea.

The options in the right click menu are getting more and more. What happens when the right-click menu is larger (higher) and the current screen?

If you do a rebase you can get rid of the merge commit in that case.

oscargus:
I think that there are also some other entries in the right-click menu which does not disable themselves as they should (if I recall correctly from adding the Merge from DOI). Just saying that it may be worthwhile taking a step back and see how the right-click menu behaves. The mark/unmark functionality is a bit inconsistent. Sometimes I get "Unmark entry", even though the entry is unmarked. Some commands, like "Open URL or DOI" and "Get Bibtex data from DOI" doesn't work with multiple entries selected although there is a rather obvious solution to how it should work.

Regarding to large menu, the three special copy-commands could be moved into a submenu, "Copy special". Would at least help a bit...

simonharrer:
:+1: for having a close look at the disable/enable logic for the menu/taskbar/right-click-menu. We probably can delete a lot of duplicate code and make the code also more explicit. 

Yes, this could help. But it also introduces another step to click them. Always these trade-offs.... :)

simonharrer:
I am against merging in a hack. Either this is improved or we close this.

PR #264, opened 2015-10-26
Remove unused menus for import/export

matthiasgeiger:
Found some dead code in JabRefFrame which has been used to generate subMenus for custom imports/exports. 

Would be glad if someone checks whether this is okay.


simonharrer:
:+1:

PR #265, opened 2015-10-27
Enable coverity

simonharrer:
Why?

- coverity does static analysis of the source code and makes the found warnings available via reports. In addition to its own checks it also performs findbugs evaluations.

To fix still

- [x] coveralls gradle plugin and coverity gradle plugin do not get along. Coverity causes coveralls to fail by using/loading the wrong class. :(


koppor:
This is https://github.com/kt3k/coveralls-gradle-plugin/issues/47 (as we use `2.4.0`). (Outdated Issue found by google at first: https://github.com/kt3k/coveralls-gradle-plugin/issues/27? Should we update the coveralls plugin to `2.4.0x`?

The only thing, that we changed, is the inclusion of coverty `1.3.708`. Does that version depend on `httpmime-4.1` which seems to be broken?

PR #266, opened 2015-10-27
Removed a number of warnings, added copyright etc

oscargus:
Cleanups based on warnings, primarily:

 - Removed right-hand side arguments of diamond operators
 - Static functions
 - Commented empty blocks
 - Added try-with
 - Avoided hiding variables

Included copyright statements.

### DIFF
--- a/src/main/java/net/sf/jabref/bibtex/EntrySorter.java
+++ b/src/main/java/net/sf/jabref/bibtex/EntrySorter.java
@@ -116,7 +116,6 @@ public class EntrySorter implements DatabaseChangeListener {
                 changed = true;
                 break;
             case CHANGED_ENTRY:
-                // Entry changed. Resort list:
                 pos = Collections.binarySearch(set, e.getEntry(), comp);
                 int posOld = set.indexOf(e.getEntry());
                 if (pos < 0) {
--- a/src/main/java/net/sf/jabref/bibtex/EntrySorter.java
+++ b/src/main/java/net/sf/jabref/bibtex/EntrySorter.java
@@ -1,4 +1,4 @@
-/*  Copyright (C) 2003-2011 JabRef contributors.
+/*  Copyright (C) 2003-2015 JabRef contributors.
     This program is free software; you can redistribute it and/or modify
     it under the terms of the GNU General Public License as published by
     the Free Software Foundation; either version 2 of the License, or
@@ -23,7 +23,6 @@ import java.util.*;
 
 public class EntrySorter implements DatabaseChangeListener {
 
-    //TreeSet set;
     private final ArrayList<BibtexEntry> set;
     private final Comparator<BibtexEntry> comp;
     private String[] idArray;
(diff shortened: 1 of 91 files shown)


src/main/java/net/sf/jabref/bibtex/EntrySorter.java
```diff
@@ -127,18 +117,14 @@ public void databaseChanged(DatabaseChangeEvent e) {
                 break;
             case CHANGED_ENTRY:
                 // Entry changed. Resort list:
-                //Collections.sort(set, comp);
```
matthiasgeiger: Should the `resort list` comment be removed, too? Or is the implementation below some kind of resorting?!

koppor:
Other than my minor comments: :+1:

src/main/java/net/sf/jabref/bibtex/EntrySorter.java
oscargus: I interpreted the first line as a comment and the second as a comment-out. :-) But reading it, I see your point. Will fix.

oscargus:
When I saw System.*.print* I replaced it with LOGGER, but I haven't searched for it (yet). But, yes, I can changed that.

OK, will try and look for those else if:s.

oscargus:
XMPUtil is a command line tool, so I *think* that it makes sense to keep it as is there?

koppor:
I don't get the comment about `XMPUtil`. Isn't it just a plain class? May it also be used as executable? Even if, we should use logging, because we mainly use it in JabRef.

matthiasgeiger:
@koppor The only `System.out.`'s are in the `main()`-method of the class (and in a method which explains the `usage()` of the CLI mode. However, it seems to be that the possibility of the standalone usage is documented nowhere... 

Remove it? Leave it as is? Or use a Logger even for the CLI infos?

koppor:
Leave as is. Add a comment above the line (for our greppers). Add a note at our  ticket to extract as external lib. That will be a standalone tool then! (lib+cmd using that lib). 

Von: "Matthias Geiger" <notifications@github.com>
Gesendet: ‎28.‎10.‎2015 09:02
Betreff: Re: [jabref] Removed a number of warnings, added copyright etc (#266)

@koppor The only System.out.'s are in the main()-method of the class (and in a method which explains the usage() of the CLI mode. However, it seems to be that the possibility of the standalone usage is documented nowhere... 
Remove it? Leave it as is? Or use a Logger even for the CLI infos?

simonharrer:
The main method (CLI) cannot be called at the moment. I checked 2.10 and we only shipped it without building an additional jar with a different main method. Why not just remove this then?

koppor:
I believe, it might be a useful tool for some power users.

matthiasgeiger:
Maybe :wink: 
But wouldn't it be better to integrate it in the normal JabRef CLI?

simonharrer:
-> dev call :)

simonharrer:
Thank you @oscargus for this PR.

PR #267, opened 2015-10-27
Removes option to hide the BibTeX Code tab in the entry editor.

simonharrer:
Why?

- I do not see the point of hiding the BibTeX Code tab in the entry editor. Hence, I removed the option.


oscargus:
:+1:

matthiasgeiger:
LGTM, too.

Issue #269, opened 2015-10-27
String to Integer/int conversion


oscargus:
In the Util class, there is a method intValueOf which converts a String to an int using a method claimed at StackExchange to be the fastest, see http://stackoverflow.com/questions/1030479/most-efficient-way-of-converting-string-to-integer-in-java

Now, it makes either sense to consistently use this, because it is efficient, or remove it and use some "standard" function, to reduce the amount of code to maintain.

Any opinions on which way to go?

simonharrer:
I introduced this method because when saving (and generally working with) large .bib files (>1000) entries, this method could be called several 100.000 times per action. 

I replaced only the calls that had O(n²) complexity, causing this large number of calls to occur. 

From my point of view, it is OK to always use this method.

oscargus:
OK! Thanks for a quick response (although I do think that my actions will not be as quick). Didn't know the background of the method, but I'll use that then.

oscargus:
A related question: util.Util is quite large. Would it make sense to move intValueOf to logic.util.strings.StringUtil?

Should, in general, as many functions as possible move from util.Util to a more well defined home such as logic.util.*.*Util?

simonharrer:
Yes, splitting up large classes into smaller ones make sense.

matthiasgeiger:
Issue can be closed?

PR #270, opened 2015-10-27
String to int conversion

oscargus:
Replaced the Integer.parseInt() and Integer.valueOf() calls with Util.intValueOf(), which is a faster implementation.

Most of the time this won't matter, but since the method is there... (See #269)


stefan-kolb:
I'm not a fan of using custom methods for JDK defaults if it isn't necessary from a performance standpoint.

koppor:
I don't have a strong opinion on that. One the hand hand consistency, on the other hand least surprise. I lean towards the concept of least surprise.

matthiasgeiger:
Same here. If it is not **really** necessary due to performance issues, use the "standard" API methods.

simonharrer:
I think a consensus is reached here that this PR is to be closed without merging.

Issue #131, opened 2015-08-27
Auto cleanups/ Sensible defaults


stefan-kolb:
This issue should be a collection of things that should be auto monitored and corrected.
Some of the features can possibly be extracted from the existing `CleanupAction.java`.

The decisions are open for discussion :smiley:

__AUTO__
- [x] page ranges in the form num--num
- [x] convert 1st, 2nd to real superscrips
- [ ] convert umlauts etc. to BibTex format
- [ ] bibtex month format #month#

__MANUAL__

koppor:
Regarding option 3: I would NOT do that. The user should decide whether he wants UTF-8 encoded files (with all umlauts etc. encoded) or plain ASCII files. The former is possible without isses with [biblatex-biber](http://biblatex-biber.sourceforge.net/), while the latter is [required for bibtex users](http://wiki.lyx.org/BibTeX/Tips).

I would even go further and have the setting "convert BibTeX encoded characters to UTF-8 characters" as default (and offer it as option at the clean up dialog).

We're at the discussion of having a main bibtex file for collecting and editing and a bibtex file for using in papers, aren't we? (#113, UC1 and UC2 in [CloudRef](https://github.com/JabRef/cloudref/wiki))

koppor:
Regarding option 4: biblatex recommends the usage of numbers as they are most portable. Therefore, I would object that, too. Or do a conversion based on the type of the file (see #167).

Minor comments: Maybe referencing an issue at a commit which will probably be rebased is not a good idea :innocent:

stefan-kolb:
Haha true :dancers:

simonharrer:
But do the numbers start from zero or from one? :hamburger:  :)

stefan-kolb:
Should be adressed in separate PR and issues.

Issue #160, opened 2015-09-11
Support of different type of BibTeX files


koppor:
We need two types of BibTeX files: One for collecting (BC) and one for working in tex files (BW).

BC: This file is UTF-8 without encoded umlauts, without additional braces.

BW: This file should have all characters encoded using LaTeX directives and containing only the fields required by the publisher. E.g., no review field, series only in the short form, abbreviated journal name, no DOI, ... (to be defined for each publisher)

Background: The `.bst` file of the publisher allows more fields than typically used in publications. One always cleans up the file by hand. That effort should be reduced by JabRef.

ThomasA:
I think JabRef should keep working with the maximally informative format (BTS) to retain as much information as possible for storage. Ideally I think it is the responsibility of the style file to get rid of any unwanted fields. UTF is not necessarily a problem either, if using bibtex8 or Biblatex. I know one sometimes has to manually remove certain fields that should not be included in a paper, but for those cases I think it would be better if the BTW format you propose were a sort of export filter.

bluebirch:
👍

tobiasdiez:
I second the comment of @ThomasA! 

With Biblatex it is really easy to use the (BC) format and then delete/modify fields accordingly directly in latex. So in this case, no (BW) file is necessary at all. For the bibtex users, I would suggest the following workflow (which needs some additional features in JabRef):
1. Run latex on your document.
2. Create a new (BW) file based on (BC) and the aux-file. (Not implemented in JabRef I think)
3. Run a cleanup on the new (BW) file. (Here it would be convenient if there were different presets possible.)

So in my opinion, the cleanup facilities should be improved instead of supporting two types of files.

ThomasA:
Last time I checked, step 2 was possible in JabRef.

tobiasdiez:
I think this issue can be closed. The cleanup operations have been improved in the meanwhile so that it is relatively easy to convert a BC file to a BW one.

koppor:
I'd rather like to put it "on-hold", as sometimes one wants to have complete entries with volume and series, which one doesn't want in the final bib file. One can achieve that with hide/unhide (refs #759), but one then needs version control...

This feature is more for users who stick with the publisher's template (such as LNCS's `splncs03.bst` instead of https://github.com/neapel/biblatex-lncs).

[...]

PR #272, opened 2015-10-27
onSave actions discussion #131

stefan-kolb:
Will replace all ordinal numbers by superscripts when saving a database, e.g. `1st Conf. -> 1\textsuperscript{st} Conf.`
Will not alter the casing of the text at the moment, should we do this? E.g. `1ST Conf. -> 1st Conf.`

### DIFF
--- a/src/main/java/net/sf/jabref/exporter/SaveDatabaseAction.java
+++ b/src/main/java/net/sf/jabref/exporter/SaveDatabaseAction.java
@@ -25,9 +25,11 @@ import net.sf.jabref.gui.FileDialogs;
 import net.sf.jabref.collab.ChangeScanner;
 import net.sf.jabref.gui.worker.CallBack;
 import net.sf.jabref.gui.worker.Worker;
+import net.sf.jabref.logic.cleanup.AutoFormatter;
 import net.sf.jabref.logic.l10n.Encodings;
 import net.sf.jabref.logic.l10n.Localization;
 import net.sf.jabref.logic.util.io.FileBasedLock;
+import net.sf.jabref.model.entry.BibtexEntry;
 
 import javax.swing.*;
 import java.io.File;
@@ -52,7 +54,6 @@ public class SaveDatabaseAction extends AbstractWorker {
 
 
     public SaveDatabaseAction(BasePanel panel) {
-
         this.panel = panel;
         this.frame = panel.frame();
     }
@@ -183,6 +184,12 @@ public class SaveDatabaseAction extends AbstractWorker {
             // If the option is set, autogenerate keys for all entries that are
             // lacking keys, before saving:
             panel.autoGenerateKeysBeforeSaving();
+
+            // onSave cleanup actions
+            for(BibtexEntry entry: panel.database.getEntries()) {
+                AutoFormatter formatter = new AutoFormatter(entry);
+                formatter.runDefaultCleanups();
+            }
 
             if (!FileBasedLock.waitForFileLock(panel.getFile(), 10)) {
                 success = false;
--- a/src/main/java/net/sf/jabref/logic/cleanup/AutoFormatter.java
+++ b/src/main/java/net/sf/jabref/logic/cleanup/AutoFormatter.java
@@ -1,0 +1,58 @@
+package net.sf.jabref.logic.cleanup;
+
+import net.sf.jabref.model.entry.BibtexEntry;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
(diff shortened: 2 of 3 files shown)


oscargus:
"but the use of the superscript form seemed to have disappeared around the 1940s and 50s — until its corpse was reanimated by Microsoft." ;-)

stefan-kolb:
We thought it looks way better than without.

oscargus:
I do not really have a strong opinion on it, just found a quote from someone that does. I tend to remove ordinal numbers from conference names anyway (using strings to get consistent naming).

koppor:
I generally like the idea to be more pushy towards correct bibtex files.

If we really go towards that direction, could you remove the CleanUp functionality from the CleanUp dialog? It is not necessary there any more! (See screenshot below)

Does the RegEx obey enclosing in brackets (`{1st}`) or don't wee need that and force users to do the right thing?

The issue that I see, that we have to check **all exporters** (and the OpenOffice integration) whether they export `\textsuperscript` correctly. Furthermore, having my .bib file polluted with TeX commands is not appealing to me. We should change our cell renderers to render/remove the LaTeX commands.

Is this a discussion concerning: #160?

CleanUp dialog:
cleanup

stefan-kolb:
Thanks for input @koppor :+1: 
- [ ] I left the CleanupAction as is intentionally, but i can remove rewritten cleanups.
- [ ] Brackets will cause problems, need to recheck this, thanks!
- [ ] Exporters may have problems with that, but if that's true that issue already existed before

I'm not sure if we should hide this information from the user by removing LaTex commands from the cell renderer. I don't see a possibility to not "pollute" the bib file with necessary LaTex commands.

Open for further discussion :smile:

koppor:
If we have configurable save actions (as eclipse), we should keep it in the CleanUp diag. If don't, what's the benefit? The entry is always updated.

I'm not sure about the users acceptance though... Don't know, how many will complain and ask for configuration options...

Von: "Stefan Kolb" <notifications@github.com>
Gesendet: ‎27.‎10.‎2015 17:18
Betreff: Re: [jabref] Add onSave action for superscripts #131 (#272)

Thanks for input @koppor  
 I left the CleanupAction as is intentionally, but i can remove rewritten cleanups. 
 Brackets will cause problems, need to recheck this, thanks! 
 Exporters may have problems with that, but if that's true that issue already existed before 
I'm not sure if we should hide this information from the user by removing LaTex commands from the cell renderer. I don't see a possibility to not "pollute" the bib file with necessary LaTex commands.
Open for further discussion

stefan-kolb:
Yeah we need to discuss this, I think #131 should reach consensus, but this one might have different opinions.

simonharrer:
I think we should make save actions that alter the .bib file configurable via preferences. And I would not remove them from the cleanup actions either. I propose to only set them to active by default.

koppor:
:+1: for configuration options and keeping the CleanUp as is.

@stefan-kolb The code in CleanUp seems to differ from your new code: Could you modify the CleanUp dialog to use your code?

stefan-kolb:
I would remove the cleanup dialog and handle all that stuff in a preferences autocleanup section. If someone modifies this, it'll be run instantly and act like the old auto cleanup?! Anyway, it will always be run as onSave action.

koppor:
:-1: for removing the cleanup dialog. I'm used to having control on my entries and see what happens. We currently have the issue with the JabRef code changing (indent etc) due to the autosave without wanting to have these changes. 

For instance, I'm against performing following changes:

* auto bracketing is IMHO not working for `BPMN`, but I have to recheck. Also moving the DOIs around does not make sense if users are forced to use a certain `bst` showing the URL field and not the DOI field.
* PDF file renaming would cause many troubles. I do not want that JabRef renames my PDFs without asking.
* Making the path relative: I think, some use cases are not covered here
* Convert to biblatex format: Not sure if we want that really happening automatically
* I really like Unicode and would propose to store everything in UTF-8. However, not all users are Unicode ready and like the LaTeX encoding more.

I think, IntelliJ also doesn't have save actions, does it? Would you use it if they were there? Always changing your code?

matthiasgeiger:
I think the best way is to have a set of defined "clean up" actions which a) can be triggered manually, b) are performed automatically when saving, c) are performed automatically when a bib file is opend/entries are imported.

So the cleanup actions should only implemented once but can be triggered in different ways. And yes, the automatic transformations should be configurable in the preferences.

koppor:
:+1: for these options. Especially c) is really useful.

stefan-kolb:
Ok, I agree with your thoughts. We can close this PR, as the code changes will be made incremently in the direction that was pointed out here.

PR #273, opened 2015-10-27
Pagenumbers CleanupAction

stefan-kolb:
Replace pages field by num--num or keep existing field if Regex is not matching.
Pages field should only allow page range, i.e. `num--num` or page set `1,23,35`.

Algorithm:
```java
     * Format page numbers, separated either by commas or double-hyphens.
     * Converts the range number format of the <code>pages</code> field to page_number--page_number.
     * Removes all literals except [0-9,-].
     * Keeps the existing String if the resulting field does not match the expected Regex.
     *
     * <example>
     *     1-2 -> 1--2
     *     1,2,3 -> 1,2,3
     *     {1}-{2} -> 1--2
     *     Invalid -> Invalid
     * </example>
```


koppor:
See https://github.com/JabRef/jabref/pull/272#issuecomment-151554205:

 * is this #160?
 * remove from cleanup dialog

stefan-kolb:
- [ ] Brackets might also cause problems here, how should we handle this?

src/main/java/net/sf/jabref/logic/formatter/PageNumbersFormatter.java
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        // remove unwanted literals incl. whitespace
+        String cleanValue = value.replaceAll(rejectLiterals, "");
+        // try to find pages pattern
+        Matcher matcher = pagesPattern.matcher(cleanValue);
+        // replace
+        String newValue = matcher.replaceFirst(replace);
+        // replacement?
+        if(!newValue.equals(cleanValue)) {
```
simonharrer: just return the new value

stefan-kolb: We want to return the initial value and not the cleaned value if page numbers regex does not match?!

simonharrer: ah ok. then leave this untouched.

src/main/java/net/sf/jabref/gui/actions/CleanUpAction.java
```diff
@@ -445,14 +446,12 @@ private static void doCleanUpMonth(BibtexEntry entry, NamedCompound ce) {
...
-        if (!oldValue.equals(newValue)) {
-            entry.setField("pages", newValue);
-            ce.addEdit(new UndoableFieldChange(entry, "pages", oldValue, newValue));
-        }
+        // run formatter
+        new PageNumbersCleanup(entry).cleanup();
+        // new value
+        String newValue = entry.getField("pages");
+        // undo action
+        ce.addEdit(new UndoableFieldChange(entry, "pages", oldValue, newValue));
```
simonharrer: we should only create the undo action if anything was actually changed (oldValue != newValue).

PR #274, opened 2015-10-27
Save size of MergeEntryDialog and MergeEntryDOIDialog

oscargus:
As per @koppor s request (sort of), the `MergeEntry*` dialogs now save and restore their size and position (same for both).


koppor:
:+1: looks good.

matthiasgeiger:
A few days ago @lenhard ranted in a chat about a JabRef dialog which was not visible on his screen as it was placed outside the visible area. (Can happen quite often if a notebook is used with and without external screens)

I have not tested whether is also happens here. But it might be useful to ensure that a Dialog IS actually visible on the currently used screens...

But this might be another PR which implements this a Util class and introduce these checks for all Dialogs :wink:

lenhard:
@matthiasgeiger No, I did not. That was git gui. And it was no dialog, but the complete UI. Read the chat logs next time before ranting about other people ranting ;-)

matthiasgeiger:
:astonished: 

Nevertheless: We should check whether this might be a problem for JabRef...

oscargus:
Good points! I seem to recall some function "placeOnScreen" or similar
flashing by in some file, so worth checking into.

Nothing I'll do this week though.

simonharrer:
Would it be possible to extract the save and restore position into an additional class as this logic is being implemented for various classes at the moment? Maybe something which you can pass in four preferences  (for x,y, width, height) and a reference to the window. The save and load functionality could be moved into the new class.

matthiasgeiger:
:+1:

oscargus:
I've extracted the logic and put it into GUIGlobals. Not sure it was
actually the most elegant way to solve it (rather sure that it wasn't), but
at least there's less code cloning.

I'll push it to the PR once I have internet access for my laptop again
(travelling, the hotel WiFi code only allowed one device and I couldn't be
bothered to ask for another once I realized it).

When looking though the code, I noticed that there is a method
`Util.placeDialog` which basically place the dialog centered on the JabRef
frame and makes sure it is visible on screen (through a single API call).
Not sure if this is really the behavior that we want when we now save the
size and position. I can imagine that some dialogs would be better kept
outside the JabRef Frame, especially in a multi-screen setup (but then
there's the problem of possibly disappearing dialogs when the screen setup
is changed). Would probably require a bit more work and thinking.

oscargus:
Had some time while traveling today.

 - Created a class PositionWindow and moved to new package gui.util
 - Use the same class for storing and restoring the main window
 - Moved some utility classes from gui there
 - Moved the Util.placeDialog method there

oscargus:
Just realized that I should create a test class for the gui.util package and move at least one of the tests for util there.

oscargus:
Rebased on current master. Any outstanding issues here?

src/main/java/net/sf/jabref/groups/GroupSelector.java
```diff
@@ -74,6 +74,8 @@
 import net.sf.jabref.logic.search.rules.sets.SearchRuleSets;
 import net.sf.jabref.logic.search.rules.sets.SearchRuleSet;
 import net.sf.jabref.gui.undo.NamedCompound;
+import net.sf.jabref.gui.util.GUIGlobals;
+
```
matthiasgeiger: why is this import the only change in this file?

matthiasgeiger: Answer to my own question: GUIGlobals has been moved into `...util` package...

oscargus: Yes, I created that package to contain gui-related "utility" classes and methods (i.e., small useful stuff...).

matthiasgeiger:
I'll try to actually check the functionality tomorrow - code looks good to me.

matthiasgeiger:
Great! Works fine for me. Just tested it with different display configurations and the the dialog was always visible.

Maybe a minor improvement would be that `savePosition()` is also called if the "Cancel" button or the window close action ("X") is called. At least I was wondering why the changed size and position was not saved after I first closed the merge entries dialog with the "x".

oscargus:
I was thinking about that case (X and Cancel) and thought that if you hit, at least, X, it may be because you have messed up the size and positioning. But no strong opinion on it and easy to change (at least for Cancel it is just to add a single call as you say).

matthiasgeiger:
I also have no strong opinion on that - just wondered during testing why the changed size was not stored.

But let's see what other do: Just checked it with the programs currently open on my machine: All programs that store size and position do this even if the dialog is closed using X. 

Might it be possible to add a "WindowSizeAndPositionChangeListener" - no idea what this is actually in Swing :wink: - which automatically calls `savePosition()` regardless how the window is disposed?

oscargus:
Rebased on current master and added the Cancel button. No Listener though, just pure hard coding (but still a bit of code improvements).

matthiasgeiger:
You don't like the ComponentAdapter? :wink:

oscargus:
I went for quick (and dirty). Although I'm also not convinced of the idea to store the settings every time a dialog is moved (while I can see that there are not that many cycles required).

matthiasgeiger:
@JabRef/developers Opinions? Storing each time resizing is completed using a ComponentAdapter or only if the dialog is closed using a Button?

simonharrer:
Vote for using a ComponentAdapter. Is cleaner, and independent on adding new buttons. 

Another issue I have with this PR: there is no constant for the JabrefPreferences as they are computed by String concatenation. Previously, we tried hard to have constants for every JabrefPreference consolidated in the JabrefPreferences class. This has advantages that we can easily find all available preferences in the system in a single class. I think we should keep that. Because of this, I would suggest that instead of a single string, we pass the four preferences keys that were previously computed to the new positioning class.

oscargus:
I agree that the ComponentAdapter has some advantages. However, a (major) disadvantage is that it is a bit too much to grasp for me at the moment. :-)

Regarding the preferences, I also agree. It is indeed a bit of a hack, but with the current preference structure, this solution seemed simpler if nothing else. This one I should be able to fix though.

simonharrer:
Component adapter should work this way

```java
this.addComponentListener(new ComponentAdapter() {
            
            @Override
            public void componentResized(ComponentEvent e) {
                // store new infos
            }

            @Override
            public void componentMoved(ComponentEvent e) {
                // store new infos
            }

        });
```
Then, this is only stored when the window size is changed or the window is moved. Maybe we could also override the `isShown` method and only then set the size of the window. Then, every time the size/position is read or written would be within this adapter - preventing that the code is scattered around in the corresponding frame/window class. 

From an architectoral view, I would recommand making the PositionWindow class non-static with the four final fields for the four preferences. This way, we could create an instance of the PositionWindow class for each of the four frames which is then reused in the appropriate places.

simonharrer:
I would not move `GUIGlobals` to the `util` package because it basically hides this important class which is used by almost every GUI class.

oscargus:
OK! I'll move it back and try to implement that ComponentAdaptor thing.

oscargus:
Finally the ComponentAdaptor is in place. Had some strange problems where some of the dialogs saved a different size when closing the dialog, but it seems to work now.

PR #276, opened 2015-10-28
Refactor BibTex model (a bit)

stefan-kolb:
- Renames getAllFields() to getFieldNames() 
- add getFields() method

Often the workflow in our code is:
```java 
for(String name: entry.getFieldnames()) {
  entry.getField(name);
  // do something
}
```

We can directly use `entry.getFields()` for this now.
Are there any concurrency Problems behind the old rationale or was it just dumb?


simonharrer:
:+1: 

But we should not expose the internal map. Can you create a new one based on the internal values?

simonharrer:
Wanted to merge this, but tests seem to fail. Can you have a look @stefan-kolb, please?

PR #280, opened 2015-10-29
Removes option to disable renaming in FileChooser dialogs.

simonharrer:
Why?

- I think this is annoying as one would expect this to work. 

I can also live with the option that this is ALWAYS active, but giving the user the option to choose is not good in my opinion.


koppor:
:+1: for removing the configuration option.

I think, in typical programs, one can rename while the dialog is open. Therefore, I would always enable it.

PR #286, opened 2015-11-02
Fixed a number of warnings

oscargus:
Got rid of a number of warnings:

 - <>
 - static methods
 - commented empty blocks
 - general clean up

### DIFF
--- a/src/main/java/net/sf/jabref/JabRef.java
+++ b/src/main/java/net/sf/jabref/JabRef.java
@@ -160,7 +160,7 @@ public class JabRef {
         openWindow(loaded);
     }
 
-    private static void setupLogHandlerForErrorConsole() {
+    private void setupLogHandlerForErrorConsole() {
         Globals.handler = new CacheableHandler();
         ((Jdk14Logger)LOGGER).getLogger().addHandler(Globals.handler);
     }
@@ -486,7 +486,7 @@ public class JabRef {
      * @return A parser result containing the entries fetched or null if an
      *         error occurred.
      */
-    private static ParserResult fetch(String fetchCommand) {
+    private ParserResult fetch(String fetchCommand) {
 
         if ((fetchCommand == null) || !fetchCommand.contains(":") ||
                 (fetchCommand.split(":").length != 2)) {
@@ -529,7 +529,7 @@ public class JabRef {
         return new ParserResult(result);
     }
 
-    private static void setLookAndFeel() {
+    private void setLookAndFeel() {
         try {
             String lookFeel;
             String systemLnF = UIManager.getSystemLookAndFeelClassName();
--- a/src/main/java/net/sf/jabref/JabRef.java
+++ b/src/main/java/net/sf/jabref/JabRef.java
@@ -160,7 +160,7 @@ public class JabRef {
         openWindow(loaded);
     }
 
-    private void setupLogHandlerForErrorConsole() {
+    private static void setupLogHandlerForErrorConsole() {
         Globals.handler = new CacheableHandler();
         ((Jdk14Logger)LOGGER).getLogger().addHandler(Globals.handler);
     }
@@ -177,8 +177,6 @@ public class JabRef {
             cli.printUsage();
             return null; // TODO replace with optional one day
         }
-
-        boolean commandMode = cli.isDisableGui() || cli.isFetcherEngine();
 
         // Check if we should reset all preferences to default values:
         if (cli.isPreferencesReset()) {
@@ -217,8 +215,8 @@ public class JabRef {
         }
 
         // Vector to put imported/loaded database(s) in.
-        Vector<ParserResult> loaded = new Vector<ParserResult>();
-        Vector<String> toImport = new Vector<String>();
+        Vector<ParserResult> loaded = new Vector<>();
+        Vector<String> toImport = new Vector<>();
         if (!cli.isBlank() && (cli.getLeftOver().length > 0)) {
             for (String aLeftOver : cli.getLeftOver()) {
                 // Leftover arguments that have a "bib" extension are interpreted as
@@ -488,7 +486,7 @@ public class JabRef {
      * @return A parser result containing the entries fetched or null if an
      *         error occurred.
      */
-    private ParserResult fetch(String fetchCommand) {
+    private static ParserResult fetch(String fetchCommand) {
 
         if ((fetchCommand == null) || !fetchCommand.contains(":") ||
                 (fetchCommand.split(":").length != 2)) {
@@ -531,7 +529,7 @@ public class JabRef {
         return new ParserResult(result);
     }
 
-    private void setLookAndFeel() {
+    private static void setLookAndFeel() {
         try {
             String lookFeel;
             String systemLnF = UIManager.getSystemLookAndFeelClassName();
@@ -668,9 +666,9 @@ public class JabRef {
         // Add all loaded databases to the frame:
 
         boolean first = true;
-        List<File> postponed = new ArrayList<File>();
-        List<ParserResult> failed = new ArrayList<ParserResult>();
-        List<ParserResult> toOpenTab = new ArrayList<ParserResult>();
+        List<File> postponed = new ArrayList<>();
+        List<ParserResult> failed = new ArrayList<>();
+        List<ParserResult> toOpenTab = new ArrayList<>();
         if (!loaded.isEmpty()) {
             for (Iterator<ParserResult> i = loaded.iterator(); i.hasNext();) {
                 ParserResult pr = i.next();
--- a/src/main/java/net/sf/jabref/bst/VM.java
+++ b/src/main/java/net/sf/jabref/bst/VM.java
@@ -1,4 +1,4 @@
-/*  Copyright (C) 2003-2011 JabRef contributors.
+/*  Copyright (C) 2003-2015 JabRef contributors.
     This program is free software; you can redistribute it and/or modify
     it under the terms of the GNU General Public License as published by
     the Free Software Foundation; either version 2 of the License, or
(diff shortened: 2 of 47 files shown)


src/main/java/net/sf/jabref/JabRef.java
```diff
@@ -160,7 +160,7 @@ public void start(String[] args) {
         openWindow(loaded);
     }
 
-    private void setupLogHandlerForErrorConsole() {
+    private static void setupLogHandlerForErrorConsole() {
```
lenhard: Why do you turn this static? It is not called from a static context.

lenhard:
I find this PR very valuable (we need such improvements). The only thing I am unsure about is turning various methods static (see above). Can you explain the rationale behind this? Why is static needed here?

koppor:
The `static` conversion is because Eclipse warns if methods could be made static. On the one hand, we could add `@suppresswarning` if we are sure that Eclipse is wrong. On the other hand, we could just turn off the `static` warning. I don't have strong a opinion about that. I currently lean towards making methods static if possible to show that they don't access instance variables.

oscargus:
As @koppor correctly pointed out, this is based on Eclipse warnings and
Eclipse warnings only. I've trusted it so far, but recently I've noted that
some warnings disappear when I fix other, to me seemingly unrelated,
warnings, so maybe I should read up a bit...

lenhard:
I tend to disagree with the warning and would prefer non-static versions. The problem with the use of static is that it leads to a higher degree of coupling within the system. Every user of a static method uses the very same instance (that is, the class) and this transitively couples two users although there might be no need for it. 

Don't get me wrong, the code works with both versions. Especially if the class is instantiated only once, as is the case in the above line I commented at, there is no difference execution-wise. It is merely a "design smell". I have the opinion that a design with a lot of use of static is "worse" than a more object-oriented design. If we consistently favor static over non-static, we end up with a purely imperative program, which means we are propably using the wrong programming language.

@JabRef/developers This is about a coding guideline and concerns everyone. So what are the opinions of the remaining people? Should we consistently fix the Eclipse warning regarding the use of static, or should we turn it off?

oscargus:
OK, I've read up and see the point. I'd say that maybe one should go through all the static methods and remove the static keyword and add warning removal in case it does not make sense to access the method in a static way.

For reference: I've done this in the earlier cleanups as well, so it may be better to go over it once and for all (although I can of course start by changing these ones back).

matthiasgeiger:
I agree with @lenhard - we should try to reduce the usage of `static` methods and not introduce new ones. 
Quick rule of thumb: 
> Static is sensible in Utility classes which provide independent small methods, e.g, `getIntValue(...)`. Static should be avoided in complexer "logic classes" - even if only static or no members are accessed. 

Objections? @lenhard 

I don't like `@suppressWarnings` annotations - so would either turn off this warning completely - or just ignore those warnings "manually", i.e., think about it and decide whether the usage is okay or not.

lenhard:
@matthiasgeiger full ack :+1:

koppor:
Could you review eb441846817b11 and efce4ec2eae5bd79e30 (unrelated, but I think, it's important, too)

lenhard:
@koppor Looks got to me. I imported the project into Eclipse and am not getting the warning at the above position. I also agree with you on the usage of "this": it should not be mandatory.

However, I found another Eclipse warning, we should be aware of here. In the following line
``` java
new FocusRequester(((BasePanel) JabRef.jrf.tabbedPane.getComponentAt(0)).mainTable);
```
Eclipse warns that `The allocated object is never used`. The warning is a false positive. `FocusRequester` starts itself in a separate Thread in its constructor (which is unsafe publication and a potential bug, but that is a different issue) and it does not matter that it is not used at this position. I don't think we should ignore the `unused` warning, but *we should not blindly delete every variable for which it occurs*. Instead, we should add a `@SuppressWarnings("unused")`, if it is a false positive.

oscargus:
I have removed the statics introduced in this PR.

oscargus:
Rebased on current master so it should be possible to merge this (unless any other issues arise).

src/main/java/net/sf/jabref/bst/VM.java
```diff
@@ -1185,17 +1185,16 @@ private void iterate(Tree child) {
     }
 
     /**
-     * Sorts the entry list using the values of the string entry variable
-     * sort.key$. It has no arguments.
-     * 
-     * @param child
+     * Sorts the entry list using the values of the string entry variable sort.key$. It has no arguments.
+     *
+     * @param
```
matthiasgeiger: `@param` should be removed here completely

src/main/java/net/sf/jabref/exporter/layout/LayoutHelper.java
```diff
@@ -342,26 +283,22 @@ private Object parse() throws IOException, StringIndexOutOfBoundsException {
     }
 
     /**
-     * 
+     *
      */
```
matthiasgeiger: `/** */` block can be removed completely

src/main/java/net/sf/jabref/exporter/layout/LayoutHelper.java
```diff
@@ -371,83 +308,64 @@ private void parseField() throws IOException, StringIndexOutOfBoundsException
                     firstLetter = name.charAt(0);
                 } catch (StringIndexOutOfBoundsException ex) {
                     StringBuilder lastFive = new StringBuilder(10);
-                    for (StringInt entry : parsedEntries.subList(Math.max(0, parsedEntries.size() - 6), parsedEntries.size() - 1)) {
+                    for (StringInt entry : parsedEntries.subList(Math.max(0, parsedEntries.size() - 6),
+                            parsedEntries.size() - 1)) {
                         lastFive.append(entry.s);
                     }
-                    throw new StringIndexOutOfBoundsException(Localization.lang("Backslash parsing error near") + " \'" + lastFive.toString().replace("\n", " ") + '\'');
+                    throw new StringIndexOutOfBoundsException(Localization.lang("Backslash parsing error near") + " \'"
```
matthiasgeiger: To be checked: Is localization needed here? Should be only relevant if the exception message is shown somewhere in the UI.

src/main/java/net/sf/jabref/groups/AutoGroupDialog.java
```diff
@@ -96,7 +96,7 @@ public void actionPerformed(ActionEvent e) {
                         new ExplicitGroup(Localization.lang("Automatically created groups"),
                                 GroupHierarchyType.INCLUDING));
                 Set<String> hs = null;
-                String field = field();
+                String field1 = field();
```
matthiasgeiger: Maybe renaming the variable to "fieldText" would be more intuitive. 

Side note: Why the hell has the method "field()" been introduced, which does nothing more than returning `field.getText()`? 

@oscargus Would you remove those rather useless methods (`field(), remove()`), too? Thanks! :wink:

src/main/java/net/sf/jabref/gui/BasePanel.java
```diff
@@ -495,8 +482,7 @@ private void setupActions() {
         // The action for opening the string editor
         actions.put(Actions.EDIT_STRINGS, (BaseAction) () -> {
             if (stringDialog == null) {
-                StringDialog form = new StringDialog
-                        (frame, BasePanel.this, database, Globals.prefs);
```
matthiasgeiger: Those were really weired line breaks... Good to see that fixed :+1:

src/main/java/net/sf/jabref/importer/fileformat/InspecImporter.java
```diff
@@ -82,26 +80,25 @@ public boolean isRecognizedFormat(InputStream stream)
...
         String str;
-        while ((str = in.readLine()) != null) {
-            if (str.length() < 2) {
-                continue;
-            }
-            if (str.indexOf("Record") == 0) {
-                sb.append("__::__").append(str);
-            } else {
-                sb.append("__NEWFIELD__").append(str);
+        try (BufferedReader in = new BufferedReader(ImportFormatReader.getReaderDefaultEncoding(stream))) {
```
matthiasgeiger: :+1:

src/main/java/net/sf/jabref/wizard/auximport/AuxSubGenerator.java
```diff
@@ -224,8 +222,8 @@ private boolean parseAuxFile(String filename) {
                             }
                         }
                     }
-                } // line != null
-                else {
+                } else {
+                    // line != null
                     weiter = false;
```
matthiasgeiger: "weiter" should be renamed to "continue" for people who don't speak German :wink:

src/main/java/net/sf/jabref/wizard/auximport/AuxSubGenerator.java
```diff
@@ -300,12 +299,11 @@ private void resolveTags() {
      * @param auxDB The database to insert into.
```
matthiasgeiger: param must be renamed to

matthiasgeiger:
Huh.... requires quite some time to scroll through all your changes :wink:

As you can see above I had a few remarks. 

Apart from those: Good job! Should be ready to merge (which should be easy after your rebasing - bonus kudos for this :trophy:)

src/main/java/net/sf/jabref/exporter/layout/LayoutHelper.java
oscargus: No. However, there are some localized exception and log messages in the current code and these will be removed in one go and the translation files updated at the same time.

src/main/java/net/sf/jabref/wizard/auximport/AuxSubGenerator.java
oscargus: I do not speak enough German to be fully sure, but that was my interpretation of it as well. :-) (Except that continue is a quite bad choice... ;-))

oscargus:
Thanks for the review! Should be good to go now.

Issue #208, opened 2015-10-04
DOAJ and bibjson


oscargus:
Just found [Directory of Open Access Journals](http://doaj.org/). They provide an [API](https://doaj.org/api/v1/docs) for searching and returns the result as [bibjson](http://okfnlabs.org/bibjson/).

* It would be nice to support searching DOAJ (direct link to PDFs etc)
* bibjson import and export might be considered (import/conversion from required for the DOAJ search)

Any preference on json package? Any already included in JabRef? (With that said, I only have a vague idea of json parsers etc.)

koppor:
In [Winery](https://github.com/winery/winery), I use the 2.x branch of [Jackson](https://github.com/FasterXML/jackson). I rely, however, on the automatic mapping of JAX-RS and seldomly use directly a parsing component of Jackson.

I would propose to create a **separate library** with a Java data model of BibJSON.

I would use the [JAX-RS 2.0 client API](https://jax-rs-spec.java.net/nonav/2.0/apidocs/). This is the more modern way instead of using the Jersey-Client API. Instead of going with Apache CXF or Jersey, the [Jackson JAX-RS Providers](https://github.com/FasterXML/jackson-jaxrs-providers) seem the way to go.

Use [jackson-annotations](https://github.com/FasterXML/jackson-annotations) to specify the mapping to/from JSON. That would help reusing the functionality in other projects such as [CloudRef](https://github.com/JabRef/cloudref/).

[jackson-databind](https://github.com/FasterXML/jackson-databind) is the module you need to use to go from JSON to POJOs. Currently not sure whether you have to explicitly include it if you include the Jackson JAX-RS JSON Provider.

Hope, that is enough information for googling :)

koppor:
I just realized that @stefan-kolb added (I think in #101) [unirest](https://github.com/Mashape/unirest-java). In the [serialization section](https://github.com/Mashape/unirest-java#serialization) in their README.md, they also point to Jackson.

oscargus:
[...]

PR #287, opened 2015-11-03
DOAJ search

oscargus:
Search for Directory of Open Access Journals. Quite functional, but the remaining things should (probably) be done

 - [x] ~~Help page~~
 - [x] Decide on where the BibJSONConverter class should be positioned (input?)
 - [x] More fields? ~~ISSN?~~
 - [x] ~~Better/more error checking~~
 - [x] ~~Handle search results with more than 100 entries properly~~
 - [x] ~~Fill empty methods in DOAJFetcher~~
 - [x] ~~Update status bar~~
 - [x] Add test for BibJSONConverter
 - [x] ~~Enable stopping search~~

Implements what is discussed in #208

### DIFF
--- a/src/main/java/net/sf/jabref/importer/fileformat/JSONEntryParser.java
+++ b/src/main/java/net/sf/jabref/importer/fileformat/JSONEntryParser.java
@@ -24,12 +24,12 @@ import org.json.JSONObject;
 import net.sf.jabref.model.entry.BibtexEntry;
 import net.sf.jabref.model.entry.BibtexEntryTypes;
 
-public class JSONConverter {
+public class JSONEntryParser {
 
-    private static final Log LOGGER = LogFactory.getLog(JSONConverter.class);
+    private static final Log LOGGER = LogFactory.getLog(JSONEntryParser.class);
 
 
-    public JSONConverter() {
+    public JSONEntryParser() {
 
     }
 
--- a/src/main/java/net/sf/jabref/importer/fileformat/JSONEntryParser.java
+++ b/src/main/java/net/sf/jabref/importer/fileformat/JSONEntryParser.java
@@ -35,10 +35,10 @@ public class JSONEntryParser {
 
     public BibtexEntry BibJSONtoBibtex(JSONObject bibJsonEntry) {
         // Fields that are directly accessible at the top level BibJson object
-        String[] singleFieldStrings = new String[] {"year", "title", "abstract", "month"};
+        String[] singleFieldStrings = {"year", "title", "abstract", "month"};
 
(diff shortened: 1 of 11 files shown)


src/main/resources/help/en/DOAJHelp.html
```diff
@@ -0,0 +1,27 @@
+<html xmlns="http://www.w3.org/1999/xhtml">
+<head>
+<link rel="stylesheet" type="text/css" href="jabref_help.css"/>
+</head>
+
+<body>
+    <h1>Fetching entries from DOAJ</h1>
+            
+    <p><a href="http://www.diva-portal.org/">DOAJ (Directory of Open Access Journals)</a> is a database
```
matthiasgeiger: Wrong link? :hushed:

oscargus: Bah! Copy-paste-programming should be more automated. ;-)

oscargus:
I think I'm done there. :-)

src/main/java/net/sf/jabref/importer/fileformat/JSONEntryParser.java
```diff
@@ -0,0 +1,150 @@
...
+    public JSONEntryParser() {
+
+    }
+
+    public BibtexEntry BibJSONtoBibtex(JSONObject bibJsonEntry) {
+        // Fields that are directly accessible at the top level BibJson object
+        String[] singleFieldStrings = new String[] {"year", "title", "abstract", "month"};
+
+        // Fields that are accessible in the journal part of the BibJson object
+        String[] journalSingleFieldStrings = new String[] {"publisher", "number", "volume"};
```
simonharrer: You can just write 

`String[] journalSingleFieldStrings = {"publisher", "number", "volume"};`

instead of 

`String[] journalSingleFieldStrings = new String[] {"publisher", "number", "volume"};`

simonharrer:
Looks good to me.

matthiasgeiger:
One minor thing: The current title is "DOAJ (Directory of Open Source Journals)" which is much longer than the rest of the fetcher titles. This introduces horizontal scrollbars in the fetcher sidebar when using the default width:

doaj-screenshot


Apart from this, "ready to go" from my side, too.

koppor:
There is no such property like [text-overflow](http://www.w3schools.com/cssref/css3_pr_text-overflow.asp) in SWING, is it? :star2:

matthiasgeiger:
I think Java actually does snip too long texts - if the component has a fixed size. So this does not help here.

I'll merge the current state, but a potential improvement would be to define tooltips for the combobox to show some hints what the various abbreviations actually mean...

PR #291, opened 2015-11-04
Springer fetcher

oscargus:
Playing around with Json was quite fun, so I did a Springer fetcher (search) as well... 

Remains:
 - [x] Help file
 - [x] Replace spaces with %20 in queries
 - [x] Where to put SpringerJsonConverter?
 - [x] Test file


src/main/java/net/sf/jabref/importer/fileformat/JSONEntryParser.java
```diff
@@ -147,4 +154,103 @@ public BibtexEntry BibJSONtoBibtex(JSONObject bibJsonEntry) {
...
+            entry.setField("author", sb.toString());
+        } else {
+            LOGGER.info("No author found.");
+        }
+
+        // Direct accessible fields
+        for (String field : singleFieldStrings) {
+            if (springerJsonEntry.has(field)) {
+                String text = springerJsonEntry.getString(field);
+                if (text.length() > 0) {
```
simonharrer: use !isEmpty

src/main/java/net/sf/jabref/importer/fileformat/JSONEntryParser.java
```diff
@@ -147,4 +154,103 @@ public BibtexEntry BibJSONtoBibtex(JSONObject bibJsonEntry) {
...
+
+        // Authors
+        if (springerJsonEntry.has("creators")) {
+            JSONArray authors = springerJsonEntry.getJSONArray("creators");
+            StringBuffer sb = new StringBuffer();
+            for (int i = 0; i < authors.length(); i++) {
+                if (authors.getJSONObject(i).has("creator")) {
+                    sb.append(authors.getJSONObject(i).getString("creator"));
+                    if (i < (authors.length() - 1)) {
+                        sb.append(" and ");
```
simonharrer: this is the String.join logic.

src/main/java/net/sf/jabref/importer/fileformat/JSONEntryParser.java
```diff
@@ -147,4 +154,103 @@ public BibtexEntry BibJSONtoBibtex(JSONObject bibJsonEntry) {
...
+            nametype = "journal";
+        } else {
+            // Probably book chapter or from proceeding, go for book chapter
+            entry.setType(EntryTypes.getType("incollection"));
+            nametype = "booktitle";
+            entry.setField("isbn", isbn);
+        }
+
+        // Authors
+        if (springerJsonEntry.has("creators")) {
```
simonharrer: A lot of the parser code is untested.

src/main/java/net/sf/jabref/importer/fileformat/JSONEntryParser.java
```diff
@@ -147,4 +154,103 @@ public BibtexEntry BibJSONtoBibtex(JSONObject bibJsonEntry) {
...
+        // Fields that are directly accessible at the top level Json object
+        String[] singleFieldStrings = {"issn", "volume", "abstract", "doi", "title", "number",
+                "publisher"};
+
+        BibtexEntry entry = new BibtexEntry();
+        String nametype;
+
+        // Guess publication type
+        String isbn = springerJsonEntry.optString("isbn");
+        if ((isbn == null) || (isbn.length() == 0)) {
```
koppor: You can use `com.google.common.base.Strings.isNullOrEmpty(isbn)` for that.

src/test/java/net/sf/jabref/importer/fileformat/JSONEntryParserTest.java
```diff
@@ -28,8 +28,7 @@
     private final JSONEntryParser jc = new JSONEntryParser();
     @Test
     public void testBibJSONConverter() {
-        String jsonString = new String(
-                "{\n\"title\": \"Design of Finite Word Length Linear-Phase FIR Filters in the Logarithmic Number System Domain\",\n"
+        String jsonString = "{\n\"title\": \"Design of Finite Word Length Linear-Phase FIR Filters in the Logarithmic Number System Domain\",\n"
```
koppor: In https://github.com/JabRef/jabref/pull/378 we work out how to read files during test. I put the source files (`.xml`; `.json` in your case) and target files (`.bib`) into version control and the tests read the files and compare them. This makes maintenance IMHO easier. Furthermore, you can test all fields by one call and don't need to add `assertEquals` for each field.

simonharrer:
@oscargus if you have time, could you work on the comments? Then we can merge this in, and ship this feature alongside JabRef 3.0.

oscargus:
I've implemented the changes (and a few more). Testing is still lacking though.

koppor:
An idea regarding test files is implemented at [GVKParserTest.java](https://github.com/JabRef/jabref/blob/1332070cb62a8990567831a10f026d31acfc1738/src/test/java/net/sf/jabref/importer/fetcher/GVKParserTest.java).

oscargus:
Yes, I've noted that, but at the moment I do not see that there will be time to read up on it. But it is in fact the reason that I didn't hard code more tests using the current approach.

koppor:
OK, let's this go through and improve testing afterwards. Refs https://github.com/koppor/jabref/issues/62

matthiasgeiger:
I just tested the Fetcher which works well from a technical standpoint - however, springer is returning some shitty dates (article published in year 2102 and so...) is this a general problem or do they currently have some problems? @oscargus

oscargus:
I haven't noticed that problem earlier anyway (rather noted that quite a lot of papers had 2016 as year).

PR #294, opened 2015-11-05
[WIP] Partial translation in Farsi by Behrouz Javanmardi

mlep:


mlep:
Please, could this be "copied" to branch dev_2.11?

matthiasgeiger:
Will Behrouz provide some more translated strings in the near future?

Then I would wait with the integration until a more reasonable amount of Strings is translated.

mlep:
Behrouz plans to complete the translation in the next 2 days.
In such a case, why not releasing it into 2.11?

matthiasgeiger:
Great! 
I think I misformulated this a bit. I meant, that the current status should not be part of a release - as most of the strings are not translated (and as 2.11 was planned to be released yesterday it would not have been included). 

But as the release will be delayed we might include the more complete version of the Farsi translation.

mlep:
It makes sense. Thanks for clarifying.

koppor:
Is it possible to use `git commit --author="Behrouz Javanmardi <javanmardy@live.com>"` so that Behrouz gets credit also by automatic tooling such as being listed as current contributor [on openhub](https://www.openhub.net/p/jabref/contributors?sort=latest_commit&time_span=12+months)?

koppor:
I currently lean against inclusion of the translation in JabRef 2.11 as 3.0 is scheduled end of this month. I can, however, wait for a complete translation, but there is no need for Behrouz to rush through.

simonharrer:
Should we integrate this for JabRef 3.1?

matthiasgeiger:
Same reasoning as before: I don't think it is useful to integrate the current state, as it will result in a UI which is 95% English with only a few words in Farsi.

mlep:
I have just mailed Behrouz Javanmardi to know if he was able to send us a more complete localization.
BTW, a 5% Farsi localization could help in motivating new people in contributing to the localization.

simonharrer:
Any progress on this?

stefan-kolb:
How should we proceed with this PR?

simonharrer:
I think we can close it for now, as nothing happened for several months. It can still be reopened when there is traction again.

PR #297, opened 2015-11-06
Improved the push to application storing and restoring 

oscargus:
- Fixed so that the storing and restoring of last used application to push to does not rely on possible translation.
 - Ordered applications alphabetically
 - Set TeXstudio as default application

(I ran format source on PushToApplicationButton so it looks more modified than it actually is.)

### DIFF
--- a/src/main/java/net/sf/jabref/external/push/PushToApplicationButton.java
+++ b/src/main/java/net/sf/jabref/external/push/PushToApplicationButton.java
@@ -172,8 +172,8 @@ public class PushToApplicationButton implements ActionListener {
 
     static class BooleanHolder {
 
-        public BooleanHolder(boolean val) {
-            value = val;
+        public BooleanHolder(boolean value) {
+            this.value = value;
         }
 
 
@@ -249,8 +249,8 @@ public class PushToApplicationButton implements ActionListener {
         private final int index;
 
 
-        public PopupItemActionListener(int idx) {
-            index = idx;
+        public PopupItemActionListener(int index) {
+            this.index = index;
         }
 
         @Override
--- a/src/main/java/net/sf/jabref/external/push/PushToApplicationButton.java
+++ b/src/main/java/net/sf/jabref/external/push/PushToApplicationButton.java
@@ -1,4 +1,4 @@
-/*  Copyright (C) 2003-2011 JabRef contributors.
+/*  Copyright (C) 2003-2015 JabRef contributors.
     This program is free software; you can redistribute it and/or modify
     it under the terms of the GNU General Public License as published by
     the Free Software Foundation; either version 2 of the License, or
@@ -34,11 +34,9 @@ import java.util.HashMap;
 import java.util.List;
 
 /**
- * Customized UI component for pushing to external applications. Has a selection popup
- * menu to change the selected external application.
- * This class implements the ActionListener interface. When actionPerformed() is
- * invoked, the currently selected PushToApplication is activated. The actionPerformed()
- * method can be called with a null argument.
+ * Customized UI component for pushing to external applications. Has a selection popup menu to change the selected
+ * external application. This class implements the ActionListener interface. When actionPerformed() is invoked, the
+ * currently selected PushToApplication is activated. The actionPerformed() method can be called with a null argument.
  */
 public class PushToApplicationButton implements ActionListener {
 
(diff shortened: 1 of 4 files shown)


src/main/java/net/sf/jabref/external/push/PushToApplicationButton.java
```diff
@@ -56,9 +54,10 @@
     private final JPopupMenu optPopup = new JPopupMenu();
     private final JMenuItem settings = new JMenuItem(Localization.lang("Settings"));
 
-    public PushToApplicationButton(JabRefFrame frame, List<PushToApplication> pushActions) {
-        this.frame = frame;
-        this.pushActions = pushActions;
+
+    public PushToApplicationButton(JabRefFrame frameJR, List<PushToApplication> pushActionsList) {
+        frame = frameJR;
+        pushActions = pushActionsList;
```
simonharrer: Why the renaming?

oscargus: To remove the "this" part. Not sure what the standpoint is on this, but I have removed quite a bit of warnings earlier where some inputs are hiding member variables, so I thought I may just do that to avoid any later warnings in case someone changes something else and so on.

simonharrer: Hm, I would leave it for setting something in the constructor/setter. What do the other @JabRef/developers think?

matthiasgeiger: Yeah, have seen this in @oscargus' warning-removal-PRs. 
I don't have a strong opinion on this: I don't bother if a member name is reused in a method (especially not in a constructor/setter)...

oscargus: I see the difference between these cases (assigning in the constructor/init). There was also a case discussed where, translated to this case, `PushToApplicationButton.frame` was discouraged. I just interpolated to this, but no problem to remove it. Not sure what is best practice (there are quite a few places where things like this occur in the constructor, so probably better not touch it and restore this one).

simonharrer:
You write in your list above that you have set Texstudio as a default, but in the code there is `"LyX/Kile"` set as the default value of the setting. 

Can you revert the parameter changes in `BooleanHolder` and `PopupItemActionListener`?

oscargus:
Oh yeah, there were more of those... Sure!

Hmm, weird! Quite convinced I wrote "TeXstudio" at some stage, but I'll fix that as well.

src/main/java/net/sf/jabref/external/push/PushToApplicationButton.java
```diff
@@ -130,22 +126,26 @@ private void buildPopupMenu() {
...
-        this.selected = i;
+        selected = i;
         PushToApplication toApp = pushActions.get(i);
         pushButton.setIcon(toApp.getIcon());
         pushButton.setToolTipText(toApp.getTooltip());
         pushButton.setPreferredSize(buttonDim);
 
-        Globals.prefs.put("pushToApplication", toApp.getName());
+        // Store the last used application
+        Globals.prefs.put(Globals.prefs.PUSH_TO_APPLICATION, toApp.getApplicationName());
```
simonharrer: Please use JabRefPreferences.PUSH_TO_APPLICATION instead. As we do not want to access static variables on instances. Theoretically, Eclipse should issue a warning for these kinds of things.

src/main/java/net/sf/jabref/external/push/PushToApplicationButton.java
```diff
@@ -68,19 +67,18 @@ private void init() {
...
-            String appSelected = Globals.prefs.get("pushToApplication");
-            for (int i = 0; i < pushActions.size(); i++) {
-                PushToApplication toApp = pushActions.get(i);
-                if (toApp.getName().equals(appSelected)) {
-                    selected = i;
-                    break;
-                }
+
+        // Set the last used external application
+        String appSelected = Globals.prefs.get(Globals.prefs.PUSH_TO_APPLICATION);
```
simonharrer: Same problem as described in other comment about JabRefPreferences.CONSTANT.

src/main/java/net/sf/jabref/external/push/PushToApplicationButton.java
oscargus: That is exactly what I have done. :-)

src/main/java/net/sf/jabref/external/push/PushToApplicationButton.java
oscargus: But I found one place where it wasn't changed. Line 283.

oscargus: Or actually not. That was a key binding, which is fixed in another PR.

src/main/java/net/sf/jabref/external/push/PushToApplicationButton.java
simonharrer: Instead of 

`Globals.prefs.put(Globals.prefs.PUSH_TO_APPLICATION, toApp.getApplicationName());`

please write

`Globals.prefs.put(JabRefPreferences.PUSH_TO_APPLICATION, toApp.getApplicationName());`

because `PUSH_TO_APPLICATION` is a static variable of the `JabRefPreferences` class and should be accessed in a static way. And not through an instance as it would be when accessing it via `Globals.prefs`.

oscargus:
Ahh! Alright!

PR #298, opened 2015-11-06
Warningremoval5

oscargus:
Removed a bunch of warnings.

 - Redundant specifications in <>
 - Resource leakage
 - Commented empty blocks

 - Removed calls to Util.pr

### DIFF
--- a/src/main/java/net/sf/jabref/bst/BibtexCaseChanger.java
+++ b/src/main/java/net/sf/jabref/bst/BibtexCaseChanger.java
@@ -1,4 +1,4 @@
-/*  Copyright (C) 2003-2011 JabRef contributors.
+/*  Copyright (C) 2003-2015 JabRef contributors.
     This program is free software; you can redistribute it and/or modify
     it under the terms of the GNU General Public License as published by
     the Free Software Foundation; either version 2 of the License, or
@@ -15,7 +15,12 @@
 */
 package net.sf.jabref.bst;
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
 public class BibtexCaseChanger {
+
+    private static final Log LOGGER = LogFactory.getLog(BibtexCaseChanger.class);
 
     // stores whether the char before the current char was a colon
     private boolean prevColon = true;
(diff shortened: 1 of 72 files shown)


simonharrer:
:+1:

src/main/java/net/sf/jabref/bst/BibtexCaseChanger.java
```diff
@@ -129,17 +134,17 @@ private String doChangeCase(String s, FORMAT_MODE format) {
         return sb.toString();
     }
 
-    private int decrBraceLevel(String string, int braceLevel) {
-        if (braceLevel == 0) {
+    private int decrBraceLevel(String string, int bLevel) {
+        if (bLevel == 0) {
```
simonharrer: please do not abbreviate variable names.

oscargus: This is one of the actual cases where Eclipse warns. braceLevel is a variable in the BibTexCaseChanger class, so using the same name for a method argument seems like a bad idea. I can of course give it a longer name rather than abbreviate it. (Using the same name seems more incorrect here than in the earlier cases we've discussed: constructors and initfunctions.)

oscargus: I could of course define it static, but not sure if that is a better option.

oscargus: I read the code and decided to inline the methods as they were only called once. This also lead to warnings messages which made more sense, removal of an unused argument in BibtexWidth and change of the corresponding test.

simonharrer:
Any remaining issues? Otherwise we should merge this.

oscargus:
Yes, one: I realized that this PR kills the OpenOffice connection (which works with some other minor changes in the code). I will revert the "correction" here and eventually provide a PR which works with the most recent versions of OpenOffice and LibreOffice. Only minor things (paths, switches etc) so should be able to include in 3.0.

oscargus:
I've reverted this change now. It should be noted that the current version does not work with the latest versions of OpenOffice and LibreOffice, but at least it is not incorrect any more.

PR #299, opened 2015-11-06
Log message translation removals

oscargus:
Trying to remove translated log messages, unless the translation is needed elsewhere. In that case I add a comment stating where the string is found.

I'll do this every now and then, so while not complete yet, it will be split in separate PRs (unless there are any objections).


matthiasgeiger:
Quick note from the breakfast table: I would remove all translations in log statements regardless whether they are used elsewhere. This would improve our ability to unterstand bug reports with logs from japanese users ;-)

Regarding the clean up of the language files: There is a python script which is - I think - able to detect unused translations. So this step could be quite automated. Right? @koppor @mlep

Am 07.11.2015 12:08 vorm. schrieb Oscar Gustafsson <notifications@github.com>:

oscargus:
OK! I agree with the stand-point on translation in log messages.

I tried the translation script and there is a quite long list of possible obsolete keys (close to 100) and about 20 keys which are not in the translation files.

This basically means that I will replace the localization.lang for the log-messages, but not manually remove the keys at the moment.

At some stage, someone should do a major overhaul of the translation keys...

mlep:
Among the obsolete keys found by the script, I found false positive. It means the script may need to be modified before to be made automatic

oscargus:
OK! Yes, I was a bit surprised to see some of the "obsolete" keys. For example: "delete_entry", "unknown".

mlep:
This false positives by the way the string is coded in JabRef sources. I do not know if the script or the java sources should be altered.

simonharrer:
I vote for fixing the issues in the Java Source code as you already do in #301

PR #300, opened 2015-11-06
Fix #290 by removing the suggested url and text 

### LINKED ISSUE #290: SourceForge references
In PR #286 I have replaced most references to sourceforge.net, but the following would need a bit more work (from someone else):

In ManageJournalsPanel.java, there is a default link to a journal abbreviation list:
`http://jabref.sf.net/journals/journal_abbreviations_general.txt`

I guess that in the long run, this file is better moved to GitHub? I can of course change the source, but do not know enough about GitHub (nor have enough access rights) to be able to place the file somewhere sensible.

oscargus:
Removes the reference to a missing file earlier hosted on jabref.sf.net. Closes #290


simonharrer:
:+1:

PR #301, opened 2015-11-07
Fixed a number of false positive translation strings

oscargus:
Cleaned up the code (basically move Localization.lang to the "correct" places) for a number of files to reduce the amount of false positives when it comes to possibly obsolete keys. This also resulted in some other related clean-ups, such as removing translation requests for database names (try to translate "INSPIRE"...).

### DIFF
--- a/src/main/java/net/sf/jabref/collab/Change.java
+++ b/src/main/java/net/sf/jabref/collab/Change.java
@@ -32,17 +32,13 @@ abstract class Change extends DefaultMutableTreeNode {
         name = "";
     }
 
-    Change(String changeName) {
-        name = changeName;
-    }
-
-    private String getName() {
-        return name;
+    Change(String name) {
+        this.name = name;
     }
 
     @Override
     public String toString() {
-        return getName();
+        return name;
     }
 
     public boolean isAccepted() {
--- a/src/main/java/net/sf/jabref/collab/Change.java
+++ b/src/main/java/net/sf/jabref/collab/Change.java
@@ -1,4 +1,4 @@
-/*  Copyright (C) 2003-2011 JabRef contributors.
+/*  Copyright (C) 2003-2015 JabRef contributors.
     This program is free software; you can redistribute it and/or modify
     it under the terms of the GNU General Public License as published by
     the Free Software Foundation; either version 2 of the License, or
@@ -21,7 +21,6 @@ import net.sf.jabref.model.database.BibtexDatabase;
 import javax.swing.tree.DefaultMutableTreeNode;
 import javax.swing.JComponent;
 import net.sf.jabref.gui.undo.NamedCompound;
-import net.sf.jabref.logic.l10n.Localization;
 
 abstract class Change extends DefaultMutableTreeNode {
 
@@ -33,8 +32,8 @@ abstract class Change extends DefaultMutableTreeNode {
         name = "";
     }
 
-    Change(String name) {
-        this.name = Localization.lang(name);
+    Change(String changeName) {
+        name = changeName;
     }
 
     private String getName() {
@@ -60,7 +59,7 @@ abstract class Change extends DefaultMutableTreeNode {
      * @return boolean false if the parent overrides by not being accepted.
      */
     public boolean isAcceptable() {
-        if (getParent() != null && getParent() instanceof Change) {
+        if ((getParent() != null) && (getParent() instanceof Change)) {
             return ((Change) getParent()).isAccepted();
         } else {
             return true;
--- a/src/main/java/net/sf/jabref/exporter/SaveDatabaseAction.java
+++ b/src/main/java/net/sf/jabref/exporter/SaveDatabaseAction.java
@@ -279,7 +279,10 @@ public class SaveDatabaseAction extends AbstractWorker {
             String tryDiff = Localization.lang("Try different encoding");
             int answer = JOptionPane.showOptionDialog(frame, builder.getPanel(), Localization.lang("Save database"),
                     JOptionPane.YES_NO_CANCEL_OPTION, JOptionPane.WARNING_MESSAGE, null,
-                    new String[] {Localization.lang("Save"), tryDiff, Localization.lang("Cancel")}, tryDiff);
+                    // @formatter:off
+                    new String[] {Localization.lang("Save"), tryDiff,
+                            Localization.lang("Cancel")}, tryDiff);
+                    // @formatter:on
 
             if (answer == JOptionPane.NO_OPTION) {
                 // The user wants to use another encoding.
--- a/src/main/java/net/sf/jabref/exporter/SaveDatabaseAction.java
+++ b/src/main/java/net/sf/jabref/exporter/SaveDatabaseAction.java
@@ -370,10 +370,9 @@ public class SaveDatabaseAction extends AbstractWorker {
             }
             f = new File(chosenFile);
             // Check if the file already exists:
-            if (f.exists() && (JOptionPane.showConfirmDialog
-                    (frame, '\'' + f.getName() + "' " + Localization.lang("exists. Overwrite file?"),
-                            Localization.lang("Save database"), JOptionPane.OK_CANCEL_OPTION)
-                        != JOptionPane.OK_OPTION)) {
+            if (f.exists() && (JOptionPane.showConfirmDialog(frame,
+                    Localization.lang("'%0' exists. Overwrite file?", f.getName()),
+                    Localization.lang("Save database"), JOptionPane.OK_CANCEL_OPTION) != JOptionPane.OK_OPTION)) {
                 f = null;
             }
         }
--- a/src/main/java/net/sf/jabref/wizard/text/gui/TextInputDialog.java
+++ b/src/main/java/net/sf/jabref/wizard/text/gui/TextInputDialog.java
@@ -273,12 +273,12 @@ public class TextInputDialog extends JDialog implements ActionListener {
 
         // Radio buttons
         JRadioButton appRadio = new JRadioButton(Localization.lang("Append"));
-        appRadio.setToolTipText(Localization.lang("append_the_selected_text_to_bibtex_key"));
+        appRadio.setToolTipText(Localization.lang("Append_the_selected_text_to_bibtex_key"));
         appRadio.setMnemonic(KeyEvent.VK_A);
         appRadio.setSelected(true);
 
         overRadio = new JRadioButton(Localization.lang("Override"));
-        overRadio.setToolTipText(Localization.lang("override_the_bibtex_key_by_the_selected_text"));
+        overRadio.setToolTipText(Localization.lang("Override_the_bibtex_key_by_the_selected_text"));
         overRadio.setMnemonic(KeyEvent.VK_O);
         overRadio.setSelected(false);
 
@@ -676,11 +676,11 @@ public class TextInputDialog extends JDialog implements ActionListener {
                 this.setForeground(Color.gray);
                 this.setFont(usedFont);
                 this.setIcon(okIcon);
-                this.setToolTipText("filled");
+                this.setToolTipText(Localization.lang("Filled"));
             }
             else {
                 this.setIcon(needIcon);
-                this.setToolTipText("field is missing");
+                this.setToolTipText(Localization.lang("Field is missing"));
             }
             return this;
         }
@@ -726,22 +726,9 @@ abstract class BasicAction extends AbstractAction {
         putValue(Action.SHORT_DESCRIPTION, description);
     }
 
-    /* TODO: Not used. Will comment out for a while for possible later removal
-    public BasicAction(String text, String description, URL icon, KeyStroke key) {
-        super(Localization.lang(text), new ImageIcon(icon));
-        putValue(Action.ACCELERATOR_KEY, key);
-        putValue(Action.SHORT_DESCRIPTION, Localization.lang(description));
-    } */
-
     public BasicAction(String text) {
         super(text);
     }
-    /* TODO: Not used. Will comment out for a while for possible later removal
-    public BasicAction(String text, KeyStroke key) {
-        super(Localization.lang(text));
-        putValue(Action.ACCELERATOR_KEY, key);
-    }
-    */
 
     @Override
     public abstract void actionPerformed(ActionEvent e);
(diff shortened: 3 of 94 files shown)


simonharrer:
:+1:

mlep:
Great! Thanks.

oscargus:
So for, this is basically what I have checked for:

 - lines with two instances of `lang` (i.e. possibly two calls to `Localization.lang() `)
 - lines with two instances of `menuTitle` (i.e. possibly two calls to `Localization.menuTitle() `)
 - lines with a call to `menuTitle` where the first character of the argument is not a `"`
 - lines with a call to `lang` where the first character of the argument is not a `"`

For the final case, there are still some instances left. However, these require quite a bit of thinking/rewriting as far as I can tell.

From my perspective, I will not work further on this at the moment. However, if someone finds any false positives, please let me know, as I may have missed instances.

oscargus:
Oh, yeah. Some of the modifications are strictly not required. However, better safe than sorry, so now changing code somewhere is less likely to break a translation elsewhere.

src/main/java/net/sf/jabref/collab/Change.java
```diff
@@ -33,8 +32,8 @@
         name = "";
     }
 
-    Change(String name) {
-        this.name = Localization.lang(name);
+    Change(String changeName) {
```
simonharrer: can you reverse the name changes in this PR as well, please?

src/main/java/net/sf/jabref/exporter/SaveDatabaseAction.java
```diff
@@ -288,7 +288,10 @@ private boolean saveDatabase(File file, boolean selectedOnly, String encoding) t
             String tryDiff = Localization.lang("Try different encoding");
             int answer = JOptionPane.showOptionDialog(frame, builder.getPanel(), Localization.lang("Save database"),
                     JOptionPane.YES_NO_CANCEL_OPTION, JOptionPane.WARNING_MESSAGE, null,
-                    new String[] {Localization.lang("Save"), tryDiff, Localization.lang("Cancel")}, tryDiff);
+                    // @formatter:off
+                    new String[] {Localization.lang("Save"), tryDiff,
+                            Localization.lang("Cancel")}, tryDiff);
+                    // @formatter:on
```
simonharrer: Please try not no use these formatter tags as we want to have them as little as possible. A simple solution would be in this case to create the array in a line above using 

`String[] options = {Localization.lang("Save"), tryDiff, Localization.lang("Cancel")};`

instead.

src/main/java/net/sf/jabref/wizard/text/gui/TextInputDialog.java
```diff
@@ -710,24 +721,26 @@ private void maybeShowPopup(MouseEvent e) {
 
 abstract class BasicAction extends AbstractAction {
     public BasicAction(String text, String description, Icon icon) {
-        super(Localization.lang(text), icon);
-        putValue(Action.SHORT_DESCRIPTION, Localization.lang(description));
+        super(text, icon);
+        putValue(Action.SHORT_DESCRIPTION, description);
     }
 
+    /* TODO: Not used. Will comment out for a while for possible later removal
```
simonharrer: Just delete it if it is unused.

koppor:
I think, my [mail to jabref-devel on 2012-11-16](https://sourceforge.net/p/jabref/mailman/message/30107905/) provides some additional information that. Especially:

    Possible obsolete key: delete_entry

    JabRef-Source:
    BasePanel.java:
    (Globals.lang(bes.length > 1 ? "delete entries" : "delete entry"))

    Solution
    Move "?" to outside:
    (bes.length > 1 ? Globals.lang("delete entries") : Globals.lang("delete entry"))

Maybe, that helps even today :sunflower:

src/main/java/net/sf/jabref/exporter/SaveDatabaseAction.java
koppor: I think, the current `syncLang.py` can handle a single `Localization.lang` call per line. Therefore, the split among different lines is required.

We could use a [Java parser written in Python](https://github.com/musiKk/plyj/tree/java8) to improve `syncLang.py`, but I'm not sure, it's worth the effort.

oscargus: Yes, that is one of the main causes for false positive obsolete translation strings. In @simonharrer's example, "Cancel" will not be found (which I guess is OK in the specific case, but in general there were ten or so false positives removed).

matthiasgeiger:
Hmm... sorry @oscargus  but I think we have a problem here.

Removing translations from the exceptions which are thrown is good - especially if the exception is only logged.
However, if the Exception message is shown at some point to the user it should be localized. 
In the current state of your PR this is not always ensured. One example:
In BasePanel.java l. 1442 the content of a `SaveException.java` is shown to the user in a MessageDialog. As the exception message was localized during creation (see SaveException.java before your changes) the message was completely localized.
After your changes the `ex.getMessage()` part is always English.

@JabRef/developers I'm not sure what we should do here. Generally, I would propose: No localization at Exception instantiation - localization at the time a localized string is actually needed. 
However, this would result in calls like `Localization.lang(ex.getMessage())` which cannot be tracked by the Python script tracking all used/unused translation keys. 
Localizing only those exception message which are used somewhere else is also grap: Hard to find to out and even harder to ensure during the evolution of the code.
If we stay at the status quo: We have (potentially) bloated translation file which contains lots of keys which are shown never to the user (except he looks in the error console) and which may be harder for us the debug if a Japanese stack trace is pasted here.

simonharrer:
Good point. 

How often is the message of an exception displayed to the user in a translated way?

In principal, I prefer to translate at the latest point in time. Maybe instead of calling Localization.lang(ex.getMessage()) one could check the message and depending on its content return a specific translation. Something like

```java
if("asdf asdfas".equals(ex.getMessage())) {
  return Localization.lang("super user-friendly text");
}
```

matthiasgeiger:
> How often is the message of an exception displayed to the user in a translated way?
Can't tell you. Analyzing this requires some time (SaveExceptions are shown only at one position (see above) - but created at 11 different positions in the code).
Another problem:
`throw new SaveException(Localization.lang("Save failed while committing changes") + ": " + ex2.getMessage());` - where ex2 refers to an `IOException` - your proposal would not work with such cases.

If a custom exception is shown to the user we might introduce another field "localizedMessage" which *can* be created during instantiation and returned by a getter in the case a localized message is needed for the UI - but of course this does not help if an `Exception e` is caught. :scream_cat:

oscargus:
I was sort of aware of this, but inspired by a comment from @koppor on my local version of the branch, I decided to still give it a go (I can revert it to the previous version if required).

I note that there is a method getLocalizedMessage() defined, which require that we subclass the Exception. However, if it is primarily SaveException in combination with BasePanel (which I think it is), it might be a viable alternative.

What I think @simonharrer meant (or at least a feasible solution) would be to do:
`throw new SaveException("Sfwcc:" + ex2.getMessage());` and then split based on `':'` or whatever is a good delimiter.

koppor:
Please, no parsings with `:`. I would introduce new exceptions (inherit from SaveException) or use properties of SaveException for information of the filename. (Currently, I don't know whether Exceptions offer the possibility to add more information than using the message).

At the end, I would do `ex instanceof MyCoolException` or do an appropriate `catch` block (the more specific exceptions coming earlier) and inside the catch block outputting the user-friendly-text. Please, no `"asdf asdfas".equals(ex.getMessage())`. This is not typesafe and we are working hard to get rid from strings cluttered around the code. (See keybindings, preferences, ...)

oscargus:
Now, since `SaveException` is a JabRef exception, I plan to add a getLocalizedMessage() and set that with the translated string before throwing the exception. Hence, any dialog use getLocalizedMessage() and any logging getMessage(). OK?

I'll check carefully if there are other Exceptions that use this as well, but my impression is that with SaveException, we cover 90% at least (if not 100%).

koppor:
`getlocalizedMessage()` sounds good! :dizzy:

simonharrer:
For the future, please to try create small pull requests. It is hard to scan the changes of more than 300 lines. 

From my perspective, this PR looks good to be merged.

oscargus:
I understand that. However, for this type of "brute-force" conversion PR, it is hard to keep them small. Shouldn't be that many more though. I've seen some translations that can be improved/replaced (using "%0" etc instead of concatenating strings), but this should cover quite a bit of it and most warnings are also removed now.

Moving GUIUtils back removed 15 or so files from the PR anyway. :-)

matthiasgeiger:
I don't have much time to think it through - but I'm not convinced that the current solution regarding the localized messages in exception solves all problems... 

I hope I'll find some time this evening to explain my thougths.

oscargus:
@matthiasgeiger : You are correct. It doesn't solve all problems, but quite a bit of them. There are a handful or so exception messages which are not translated at the moment. However, I have not been able to fully figure out why they are translated (i.e., if there is a reason that they are translated to start with) and where they show up (some are quite obviously translated for a reason though). The way to solve this is to create subclasses of three(?) exception types which support getLocalizedMessage().

Somehow, at the moment, we get to choose between reports of missing translations or bug reports with stack traces in Japanese. ;-)

Also, looking at the complete translation/logging/error reporting situation, it can be improved. There is still quite a bit of exception handling which just prints the stack trace. I still find strings that should be translated. While most messages to SaveException is translated, it is also quite common that the exception message provided is "rt". Quite a few strings are built from concatenation, limiting the grammatical correctness of translations.

I think it would be brilliant if some guidelines for translations could be formed. Now, quite often, the strings doesn't include a full stop (.) nor are variables added with %0 etc. I also think it would be a good idea to make a major overhaul of this is one go, as a translator would benefit quite a bit of having the obsolete translations available when adding new, almost similar, strings.

matthiasgeiger:
Okay got some minutes time... 
You are right, your last commits are a big step in the right direction. The implementation of `getLocalizedMessage()` in `SaveException` is the right way to solve the issues and this should be adapted to more Exceptions.

But currently there still will be "mixed language" messages in the code as - for example in `FileActions.java` - you are creating localized messages which are relying on the implementation of `getLocalizedMessage()` method in standard API `IOException`s. And I don't think that, e.g., a `FileNotFoundException` does actually provide a localized message.

I have not checked whether the SaveExceptions in `FileActions.java` are only logged or actually displayed in the UI. But as you stated above:

> Hence, any dialog use getLocalizedMessage() and any logging getMessage().

We would use the getLocalizedMessage() only in dialogs/UI output. Maybe we should not append any technical messages from the cause at all? [An example is in `SaveSession`](https://github.com/oscargus/jabref/commit/1338137cff12a943b9e71888039d4e399bf46cee#diff-5cadfd8a296c63905e55cafd1321851fR130) where an IOException is caught:

```java
throw new SaveException("Save failed while committing changes: " + ex2.getMessage(),
                Localization.lang("Save failed while committing changes: %0", ex2.getMessage()));
```
here it might be sufficient to append the ex2.getMessage() part only to the "logMessage" but not to the "UI message":

```java
throw new SaveException("Save failed while committing changes: " + ex2.getMessage(),
                Localization.lang("Save failed while committing changes."));
```

However, this is only possible if the localized message is specific enough for the user to know what the problem might be... Applying this pattern consistently in whole project would require a huge effort, as the whole exception handling has to be improved: 
- clearer distinction between "API Exceptions" and JabRef exceptions we throw on our own
- replacing the catching and especially throwing of raw `Exception` (and even `Throwable` :scream_cat: )
- which requires: creating custom Exceptions with an implementation of `getLocalizedMessage()
- systematic usage of logging instead of System.out/err
- ...

This is enough stuff for some more huge PRs :wink:  - and I fear solving this in "batch-mode" will not be possible, but it requires quite some thinking time in various cases.

> I think it would be brilliant if some guidelines for translations could be formed. Now, quite often, the strings doesn't include a full stop (.) nor are variables added with %0 etc. I also think it would be a good idea to make a major overhaul of this is one go, as a translator would benefit quite a bit of having the obsolete translations available when adding new, almost similar, strings.

:+1: Full Ack! We should create a Wiki-Page for this - including the guidelines how to deal with localized exceptions :wink:

mlep:
Uhh... A few days ago, I just mentioned the existence of a few false positives, and now it seems a big trouble to a whole bunch of programmers... :-/

Anyway, I do not understand much at this very technical discussion.
Should the "guidelines how to deal with localized exceptions" be added to the wiki page regarding translation? (https://github.com/JabRef/jabref/wiki/Translating-JabRef) I guess not...

matthiasgeiger:
Don't worry @mlep you are not the cause of this trouble :wink:

The main trouble is the rather bad and inconsistent exception handling in the code...
Regarding the Wiki-documentation: No, this is not relavant for translators - but a guide for programmers how to enable localization in a "correct" way. (Still, a cross reference might be useful).


Regarding the PR: I think we can merge the current state as is and improve the Exception handling (and its localization) step-by-step.

oscargus:
As I understand it getLocalizedMessage() will just return getMessage() if it is not overridden, so it is safe from that perspective.

Regarding appending messages, I think it depends. Clearly, the current version is not correct though, but 
```
throw new SaveException("Save failed while committing changes: " + ex2.getMessage(),
                Localization.lang("Save failed while committing changes: %0", ex2.getLocalizedMessage()));
```
may be a viable alternative.

matthiasgeiger:
Yes, this would be the more appropriate alternative. However, currently (in master) a potential error dialog looks like this: 
import-failure

I do not like this for two reasons: First, the second part is not guaranteed to be localized (not in our hands if exceptions not thrown by us are overwriting the getLocalizedMessage() method or not). And second and more important: No normal user can understand what is actually the problem here (The selected file has a wrong format and cannot be loaded). 

So **we** should actually provide localized and understandable messages for all dialogs directly shown to the user.

But this is all out of the scope of this PR and maybe we should discuss this further in #320 - although I think we generally agree on the way to handle this stuff :wink:

So again: Imho, the state here can be merged - and will be improved eventually in the course of #320 (or related direct pushes to master).

oscargus:
That was indeed an ugly dialog from so many perspectives. :-)

matthiasgeiger:
I just finished a first version of the "Exception Handling" and "Localization" HowTos in the Wiki:

https://github.com/JabRef/jabref/wiki/Code-Howtos

Feel free to contribute/check ;-)

simonharrer:
:+1:

oscargus:
:+1:

PR #302, opened 2015-11-07
Replaced some of the GetKey arguments with constants from KeyBinds

oscargus:
Replaced some of the GetKey arguments with constants from KeyBinds

Change: the preamble editor dialog doesn't have its own key entry for closing it.


matthiasgeiger:
:+1:

PR #303, opened 2015-11-07
Replaces the final(?) KeyBinds.STRING and some Actions.STRING

oscargus:
Some more cleanups.

In GeneralFetcher, the key binding for "Fetch Medline" is defined, which seems a bit non-intuitive. Not sure how to proceed here.


simonharrer:
What do you mean with "In GeneralFetcher, the key binding for "Fetch Medline" is defined, which seems a bit non-intuitive. Not sure how to proceed here."? Why is this non-intuitive?

oscargus:
I replaced the key binding in general fetcher with "Web search", which seems more intuitive. :-) What I meant was that it doesn't make sense that the GeneralFetcher defines the Fetch Medline shortcut nor defines "Fetch Medline" as the name for switching the search panel.

src/main/java/net/sf/jabref/gui/fieldeditors/JTextAreaWithHighlighting.java
```diff
@@ -77,39 +77,37 @@ public void undoableEditHappened(UndoableEditEvent evt) {
...
-                    @Override
-                    public void actionPerformed(ActionEvent evt) {
-                        try {
-                            if (undo.canUndo()) {
-                                undo.undo();
-                            }
-                        } catch (CannotUndoException ignored) {
-                    // Ignored
-                        }
+        getActionMap().put("Undo", new AbstractAction(Actions.UNDO) {
```
simonharrer: Actions.UNDO == "undo" whereas you had "Undo" before. I am not sure which effects will this change have. We should be careful here. Does it still work for this special component?

simonharrer:
As the medline fetcher keybinding was a bug, can you report this in the CHANGELOG, please?

src/main/java/net/sf/jabref/gui/fieldeditors/JTextAreaWithHighlighting.java
oscargus: I *think* that it is just a arbitrary name, but it may be better to not change it... This naming vs Action.CONSTANT should be clarified somehow. Or maybe it is clear to other people than me...

I'll add the ChangeLog entry and also remove the other stuff discussed.

simonharrer:
Can you please rebase this? Then it is good to be merged. Thank you!

PR #305, opened 2015-11-08
Add ability to format/cleanup the date field

tobiasdiez:
The value in the date field has to be of the form yyyy-mm or yyyy-mm-dd according to the BibLatex manual. This PR adds the ability to format/cleanup the date field to the correct format.


CHANGELOG
```diff
@@ -1,4 +1,5 @@
 [master]
+	- Feature: Add ability to format/cleanup the date field
```
simonharrer: Please use spaces instead of a tab.

simonharrer:
Can you implement the `net.sf.jabref.logic.formatter.Formatter` interface in your `DatesCleanup` class, please? We try to implement all cleanup actions and formatters by using this interface. Have a look at the `PageNumbersFormatter` which is also part of a cleanup action as well.

tobiasdiez:
I will change the tab to spaces. But what do you mean with the Formatter interface? As far as I can see, the PageNumberFormatter does not implement any interface. I copied the PageNumberFormatter file and changed the logic accordingly.

simonharrer:
In the master branch, the PageNumberFormatter does implement the Formatter interface.

koppor:
[net.sf.jabref.logic.formatter.bibtexfields.PageNumbersFormatter](https://github.com/JabRef/jabref/blob/master/src/main/java/net/sf/jabref/logic/formatter/bibtexfields/PageNumbersFormatter.java)

tobiasdiez:
Ok I understand now. Thanks.
But what is then the purpose of the class PageNumbersCleanup? It just delegates the work to the formatter class. In my opinion this does not provides any real simplification since the formatter can be called directly in the CleanupAction class.
Should I implement a similar delegate mechanism for the date field or just remove PageNumbersCleanup?

simonharrer:
We try to build a cleanup mechanism based on formatters. The idea is that we can register these actions in arbitrary places, e.g., onSave, onImport, onExport, cleanup, etc. and apply them to different fields. The formatters themself are independent of any logic and therefore easy to test. 

Try to replicate the same structure, please.

tobiasdiez:
The Formatter interface is now used. However, this leads to the following problems with the code:
1. The DateCleanup class is now essentially a 1-1 copy of the PageNumberCleanup class.
2. The DateFormatter is not tested directly but only through the DateCleanup class. 

Essentially, the cleanup classes are now just passing through to the formatter and thereby create additional overhead. I would propose to remove one layer of abstraction using the following idea (would be a new PR):
Create a class `CleanupPreset` which collects all the possible formatters / cleanup operations and whatever the operation is activated or not. This class should also contain a method `cleanup(BibtexEntry)` which applies all the active formatters to the `BibtexEntry`. Maybe we also need a factory to create the (default) presets for save / import / export etc. Then on save (or cleanup or whatever) the appropriate preset will be created and then applied to the correct entries. I'm not sure whatever there should be an additional layer (i.e. move the cleanup method from the preset class to some commander) which keeps track of changes and allows them to be reverted.

src/main/java/net/sf/jabref/logic/formatter/bibtexfields/DateFormatter.java
```diff
@@ -0,0 +1,64 @@
...
+public class DateFormatter implements Formatter {
+    @Override
+    public String getName() {
+        return "Date";
+    }
+
+    /**
+     * Format date string to yyyy-mm-dd or yyyy-mm. Keeps the existing String if it does not match one of the following
+     * formats:
+     *  //@formatter:off
```
stefan-kolb: Do we need to use these tags even inside Javadoc? For me, then the formatting is fundamentally flawed...

simonharrer: We should discuss this in the dev call. Again....

src/main/java/net/sf/jabref/gui/actions/CleanUpAction.java
```diff
@@ -725,4 +740,22 @@ private static void doConvertLaTeX(BibtexEntry entry, NamedCompound ce) {
...
+     * Format dates correctly (yyyy-mm-dd or yyyy-mm)
+     */
+    private static void doCleanUpDate(BibtexEntry entry, NamedCompound ce) {
+        String oldValue = entry.getField("date");
+        // not set
+        if (oldValue == null) {
+            return;
+        }
+        // run formatter
+        new DatesCleanup(entry).cleanup();
```
simonharrer: The dates cleanup class is not required. Have a look at the doConvertUnits method. It should simply call the formatter which formats the string.

src/main/java/net/sf/jabref/logic/formatter/bibtexfields/DateFormatter.java
simonharrer: The formatter tags in comments can be removed as the new formatter rules do not format comments anymore.

src/test/java/net/sf/jabref/logic/cleanup/DatesCleanupTest.java
```diff
@@ -0,0 +1,205 @@
+package net.sf.jabref.logic.cleanup;
+
+import org.junit.Assert;
+import net.sf.jabref.model.entry.BibtexEntry;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+public class DatesCleanupTest {
+    private BibtexEntry entry;
```
simonharrer: Instead of testing DatesCleanup, it would be better to test the formatter with simple input/output string pairs.

simonharrer:
I tried to address your large comment with several minor comments inline in the code. I hope you get the idea of what I am aiming for.

tobiasdiez:
@simonharrer Thanks for your feedback. I refactored the code around formatters and cleanup actions, while keeping your suggestions in mind. In the end, I opted in for encapsulating a lot of logic in the new `doFieldFormatterCleanup` method and `FieldFormatterCleanup` class. What do you think about these changes?

Ps: Why is there the LayoutFormatter interface as well as the Formatter interface?

simonharrer:
Looks very nice now, your two now classes seems to a good approach to encapsulate this. Thank you. 

LayoutFormatter is currently used to define export formats. We did not want to touch this as we have not enough tests to catch all issues when chaging this. But the idea is to consolidate this at some point.

simonharrer:
Please rebase on master, so this can be merged in.

tobiasdiez:
Done!

PR #162, opened 2015-09-12
[WIP] Improvements to search

tobiasdiez:
This is a new version of my old pull request, now updated to the 2.8 version and incorporating all the suggestions from the last time. (Except the clean commit history )

The changes include:
- Search bar is now at the top
- Autocompletion suggestions are shown in a popup
- Search options are available via a drop-down list, this implements Feature Request #853
- "Clear search" button also clears search field, this implements Feature Request #601
- In incremental search mode: next result is selected by pressing enter (instead of Control + F as previously)
- Incremental search now uses the same search engine as the other search modes
- Added a new search mode (Live filter) which filters the table as soon as the search text is changed

I have a few questions:
- ~~There are quite a few changes from formatting the code in Eclipse. Is it ok if I include them?~~
- I have a problem with listening to changes in the textfield and programatically setting the text with setText(). The latter command triggers first an event for the deletion of the text and then another one for inserting the new one. Is there a way to be informed just once about such a change? (This occours in the following case: if the user accepts an autocompletion, then the text is changed via setText(). Thus, in incremental and live filter mode, two searches are performed.)
- ~~The SearchWorker implemented the AbstractWorker interface. As far as I understood, this enables one to do work asynchronously and then change the GUI later. However, the worker is simply run via 
        worker.getWorker().run();
        worker.getCallBack().update();
In my understanding, this defeats completely the purpose of using the AbstractWorker. What is the right way to run it?~~
- ~~A general observation about the `SearchRule` class. Every method of it takes the query as an argument. I would propose to save the query internally in  `SearchRule` and thus simplify the interface a little bit.~~

- ~~The setting "Autocomplete names" is not yet implemented. In my opinion this can be completely removed.~~
- ~~The setting "Select matches" is not yet implemented since I do not understand the purpose of this setting. Should it control whatever the first match is selected after the search is performed? Moreover, I think it was not implemented in the old search pane.~~

Things which should be done at some point (I will open a ticket after this pull request was accepted):
- All the other inline autocompletions (in fields for example) should also use the new dropdown style autocompleter
- Search result table is not styled the same as main table, related [SF Bug 847](http://sourceforge.net/p/jabref/bugs/847/)

PR #307, opened 2015-11-09
Improvements to search

tobiasdiez:
Cleaned-up version of #162.


simonharrer:
Can you please remove the `DONE` tags in your comments?

tobiasdiez:
DONE :sunglasses:

koppor:
Sorry for being picky at the formatting. When reading the diff, the indents don't seem to be right. Besides the content of the code, we also want to have the formatting consistent to ease newcomers to read and understand the code.

Do you use Eclipse or IntelliJ? With the recent master, `gradlew cleanEclipse eclipse` should generate an Eclipse config with the right indent settings.

tobiasdiez:
@koppor I totally agree about the formatting/indenting issues. However, the last time I run the formatter on the files resulted in many formatting changes. Is it possible that you run the formatter on the files after accepting the PR? Otherwise, my commit will be again screwed. 
Moreover, yes the searched text should be always highlighted. I don't understand the purpose of the preference. Maybe this should be done in a new PR.
The split button is necessary because both sides of the button are clickable whereas for a combobox only the right is clickable.

koppor:
I will discuss that with the other developers. Meanwhile, I'd like to ask whether you are aware that [git gui](https://git-scm.com/docs/git-gui) can do line-based committing: Mark the lines to commit, right click, choose "stage lines to commit". Then, you can collect only the "important" modified lines. I would propose to do the formatting and create a new commit. Amending does not work as your PR already consists of two commits. You could use `git rebase -i` and then "fixup" the last two commits (DONE and formatting) into the first one. Afterwards, you have to do a `git push -f` ("force overwrite existing branch" in git gui). BTW: As far as I remember, it is possible to mark some lines in the code and then push "CTRL+SHIFT+f" to format. Then, the line-based commit is not necessary; only the merging of the three commits into one ("fixup" at `git rebase -i`).

koppor:
I created a new branch `uisearchhistory` and merged the content there. We will continue working on the code there. Hope, that's OK!

Issue #155, opened 2015-09-10
[WIP] Consistent encoding strings


koppor:
JabRef's [current encoding list](https://github.com/JabRef/jabref/blob/be4b2aa2d7744b8b4e28c4662b2145a76470e4be/src/main/java/net/sf/jabref/logic/l10n/Encodings.java) lists possible encodings. 

The strings have flaws:

* `UTF8` and `UTF-16` are possible encodings. I really dislike the inconsistency between them: `UTF8` has no `-`, but `UTF-16` has: Both should have a dash resulting in `UTF-8`. (see https://de.wikipedia.org/wiki/UTF-8)
*  `Cp1251` should be `CP1251` (see http://man7.org/linux/man-pages/man7/cp1251.7.html for example). Similar for all other code pages
*  `ISO8859_15` should be `ISO8859-15` (see https://de.wikipedia.org/wiki/ISO_8859-15). Similar for all other encodings

On the one hand, I am very well aware that a change here results in a large compatibility code where the old strings have to be mapped to the new strings. On the other hand, I would like to see JabRef having strings not causing confusion.

Related: #118, #131

lenhard:
I fully agree with what you write and just had a look at this issue.

As far as I can see, we can just replace the faulty names with the correct ones in `ALL_ENCODINGS` and add the correct names to `ENCODING_NAMES_LOOKUP`. This means we have two keys (the correct name and the old name) for the same encoding. This should not break anything, since `ALL_ENCODINGS` is only used to display available encoding names in the UI and `ENCODING_NAMES_LOOKUP` is used for for actually applying an encoding. Hence, duplicate keys won't hurt as long as the values are ok (which, I assume, they are).

To sum up: This change looks pretty easy to me, but perhaps I just overlooked a difficult problem. Can you point me to code that breaks with the above changes?

simonharrer:
Can the user select the encoding in the UI? Do we want to automatically upgrade from an old encoding name to the new encoding name? Where are the used encodings stored?

lenhard:
1. *Can the user select the encoding in the UI?*

Yes, at various positions, for instance when a save action fails with an `UnsupportedCharsetException`. Example code is:
```
Object choice = JOptionPane.showInputDialog(frame, Localization.lang("Select encoding"), Localization.lang("Save database"),
                        JOptionPane.QUESTION_MESSAGE, null, Encodings.ENCODINGS, encoding);
```
[...]

PR #310, opened 2015-11-10
Copacimportertest

zellerdev:
New PR on a new branch.
Edited Tests. No assignments. Better formatting. Cover DT (documenttype).
New Test in case an empty Text is imported.

Edit: Added the suggestions into the testfile. Checking the equality of Copac and RIS is ongoing.


simonharrer:
Can you please have a look at https://ondemand.coverity.com/jobs/uj1ep5hu8118n66eqa61rs5ofc/results/f/4/8/9/620b152144b541f48db807c7eed9.html which indicates issues in the test class. Please close the resources with a try-with-resources block. 

Regarding the `testIsRecognizedFormat`, I think it would be better if there are two lists, one for the accepted and one for the rejected files which can then be iterated and check in a for loop each. This would allow to add all the other files from the other importers that will trickle in over the next month.

zellerdev:
Am I assuming right that you mean something like that?
```java
    @Test
    public void testIsRecognizedFormatAccept() throws IOException {
        CopacImporter importer = new CopacImporter();
            LinkedList<String> list = new LinkedList<>();
            list.add("CopacImporterTest1.txt");
            list.add("CopacImporterTest2.txt");

            for (String str : list) {
                try (InputStream is = CopacImporterTest.class.getResourceAsStream(str)) {
                    Assert.assertTrue(importer.isRecognizedFormat(is));
                    is.close();
                }
            }
    }
```

lenhard:
Almost :) As you are using try-with-resources, the line `is.close()` is no longer needed. The try block takes care of that automatically.

oscargus:
I read somewhere that it is still recommended (by many) to keep the close(), primarily for readability/understanding I guess. Now, I haven't followed it that carefully though...

simonharrer:
Do you have a link for that @oscargus ?

oscargus:
Maybe not. I tried to find it, but to no avail. I found something which might be what I referred to, but it didn't actually say exactly that. Rather, I found many examples without the close().

src/test/java/net/sf/jabref/importer/fileformat/CopacImporterTest.java
```diff
@@ -69,15 +122,55 @@ public void testImportEntries() throws IOException {
...
+    }
+
+    @Test
+    public void testImportEntries3() throws IOException {
+        CopacImporter importer = new CopacImporter();
+
+        //CopacImporterTest3.txt is an empty file.
+        List<BibtexEntry> entries = importer.importEntries(
+                CopacImporterTest.class.getResourceAsStream("CopacImporterTest3.txt"), new OutputPrinterToNull());
+        Assert.assertEquals(0, entries.size());
```
simonharrer: When comparing lists, it is better to do an assert that way:

`Assert.assertEquals(Collections.emptyList(), entries);`

Then, the message when the assertion fails is much better, as it contains the entries in the list using their toString method. 

Example:

```java
List<String> values = new LinkedList<>();
values.add("asdf");

Assert.assertEquals(Collections.emptyList(), values);
```

results in 

```
Exception in thread "main" java.lang.AssertionError: expected:<[]> but was:<[asdf]>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:118)
	at org.junit.Assert.assertEquals(Assert.java:144)
	at Test.main(Test.java:13)
```

zellerdev: Do you want me to only edit the ```testImportEntries3()``` ? 
Or should I create two Lists on every ```testImport()``` and only make one Assert at the end? For example 
```java
    @Test
    public void testImportEntries() throws IOException {
        Globals.prefs.put("defaultEncoding", "UTF8");

        CopacImporter importer = new CopacImporter();

        List<BibtexEntry> entries = importer.importEntries(
                CopacImporterTest.class.getResourceAsStream("CopacImporterTest1.txt"), new OutputPrinterToNull());
        Assert.assertEquals(1, entries.size());
        BibtexEntry entry = entries.get(0);

        List<String> idealValue = new LinkedList<>();
        idealValue.add("The SIS project : software reuse with a natural language approach");
        idealValue.add("Prechelt, Lutz and Universität Karlsruhe. Fakultät für Informatik");
        idealValue.add("Interner Bericht ; Nr.2/92");
        idealValue.add("1992");
        idealValue.add("Karlsruhe :  Universitat Karlsruhe, Fakultat fur Informatik");
        idealValue.add("Edinburgh");
        idealValue.add("TXT");

        List<String> isValue = new LinkedList<>();
        isValue.add(entry.getField("title"));
        isValue.add(entry.getField("author"));
        isValue.add(entry.getField("series"));
        isValue.add(entry.getField("year"));
        isValue.add(entry.getField("publisher"));
        isValue.add(entry.getField("HL"));
        isValue.add(entry.getField("documenttype"));

        Assert.assertEquals(idealValue, isValue);

        Assert.assertEquals(BibtexEntryTypes.BOOK, entry.getType());
    }
```

simonharrer: I meant only when checking that the size of the list should be 0. You do not get too much information that the list has size 1. It is much easier to debug when you know what values the element in the list has.

zellerdev: Okay. Now I know what you mean. Thank you.
After I have changed ```testImportEntries()``` to the "two list model" I got a failure. I couldn't resolve it because it was one big Message and I didn't saw my mistake. 
This is why I asked before I changed the other one and pushed it.

lenhard:
I just realized that a few fixes I implemented when going through coverty warnings this morning (see 5b66b9868e4c7ca97361ba44c2f2f89131828765) conflict with this PR. I fixed the resource leaks and the remainder of this PR is still valuable.

@zellerdev To get it into master, please merge master into your branch. Sorry for the inconvenience, I'll keep my hands off the importers from now.

koppor:
I noticed that line 53 reads

    Globals.prefs.put("defaultEncoding", "UTF8");

Could you please change that string to `UTF-8` and check if your test cases still work? - JabRef changed the strings to be more conforming to usual strings - see #155.

Furthermore, could you follow the pattern "If you modify preference, use following pattern:" stated at https://github.com/JabRef/jabref/wiki/Code-Howtos#test-cases?

lenhard:
I'd like to add that you should not refer to preferences via their string key, but rather via their static variable in `JabRefPreferences`, e.g.:
```
Globals.prefs.put(JabRefPreferences.DEFAULT_ENCODING, "UTF-8");
```

koppor:
I think, it's easier for reviewing if you squash the whole commits into one commit. `git reset master` is your friend.

src/test/java/net/sf/jabref/importer/fileformat/CopacImporterTest.java
```diff
@@ -6,46 +6,85 @@
...
+    public static void setUp() {
+        Globals.prefs = JabRefPreferences.getInstance();
     }
 
+
+    @Rule
+    public ExpectedException thrown = ExpectedException.none();
+
+
     @Test
```
tobiasdiez: I think it is slightly better to understand if you use `@Test(expect=IOException.class)` instead of the `thrown` construction.

koppor: I agree. In this case, no the whole method may throw an exception and not a single line (as IMHO discussed with @bartsch-dev in another PR)

koppor:
Coverage [100%](https://codecov.io/github/JabRef/jabref/src/main/java/net/sf/jabref/importer/fileformat/CopacImporter.java?ref=f3bfd6edcb39b55ce4ef9bd3198112fbe5777abe). :+1:

While wondering why the import source is so small, I checked the header comment http://copac.jisc.ac.uk/faq/#format to check whether there are more fields. There, I found "How do I import Copac records in reference management software? (e.g. Endnote/Zotero)" which contains "If you use Reference Manager you can use the RIS filter to import Copac records."

@obraliar @mairdl @zellerdev @ayanai1 Could you please check whether RIS is really equal to Copac? Can you read each other's code and test cases and report back here?

In case they are equal, I would propose to merge the importer code.

koppor:
@zellerdev I can't see what you changed why in your commit. In this case, it would have been better to add another commit. Then, I can easily see, what you changed after my approval. Now, I have to review the whole commit again and I don't know what I have been approved and what not.

zellerdev:
I apology for that. Because of the problems building the project I had  in the past this was a routine for me.

src/test/java/net/sf/jabref/importer/fileformat/CopacImporterTest.java
```diff
@@ -76,17 +97,56 @@ public void testImportEntries() throws IOException {
...
+            Assert.assertEquals("x, 102p, leaf : ill., form, port ; 22cm", one.getField("physicaldimensions"));
+            Assert.assertEquals("0407517006 (Pbk)", one.getField("isbn"));
+            Assert.assertEquals("Bibl.p.94-97. - Index", one.getField("note"));
+            Assert.assertEquals(
+                    "London Hospital and Medical College, Electronic data processing - Medicine, Computers - Hospital administration, Hospital planning, Operations research, Hospital equipment and supplies, Electronic data processing - Hospitals - Administration, Hospitals, London, London Hospital and Medical College, Records management, Applications of computer systems, to 1971",
+                    one.getField("keywords"));
+            Assert.assertEquals("TXT, PDF", one.getField("documenttype"));
+            Assert.assertEquals(
+                    "Aberdeen ; Birmingham ; Edinburgh ; Trinity College Dublin ; UCL (University College London)",
+                    one.getField("HL"));
```
tobiasdiez: What is `HL`? From the content, I would suspect it represents the location/city of the publisher. In this case it should  be written to the `address` field in bibtex.

obraliar:
After research we can say that the both file-formats are syntactically not the same. There are a lot of tags (like: SE, PU, PD, NT, HL, ...) in Copac which are note parsable with RISImporter. E.g. `NT` (Copac) <-> `N1` (RIS), which means 'notes'. On closer inspection at the line 163 (else condition) it becomes clear, that Copac accepts all kind of tags which differ from those listed above. That's probably the reason why the code looks short.

src/test/java/net/sf/jabref/importer/fileformat/CopacImporterTest.java
zellerdev: This implementation of CopacImporter knows the tags TI, AU, PY, PU, SE, IS, KW, NT, PD and DT. In this case Copac doesn't know what HL is. So to address this field you have to use the tag HL as field. This ressources already existed. In this case this is a test to watch whether the importer correctly handles with unknown tags.

koppor:
My feeling is that you inspected the code and not the format specifications and concrete results of the exporters linked somewhere from http://copac.jisc.ac.uk/faq/#format. Possibly, it is implicitly clear that the RIS format is completely covered? At #505, I don't see any testfiles of http://copac.jisc.ac.uk.

Is there any Copac format specification? Is there any RIS format specification? My feeling is that tags such as `SE` etc. should be treatable by RIS as well, but I'm not sure.

zellerdev:
Updated CopacImporterTest so all checks are succesful and removed conflicts with base branch.

src/test/java/net/sf/jabref/importer/fileformat/CopacImporterTest.java
```diff
@@ -75,17 +95,56 @@ public void testImportEntries() throws IOException {
...
+            Assert.assertEquals("0070844828", two.getField("isbn"));
+            Assert.assertEquals("Index", two.getField("note"));
+            Assert.assertEquals("Real-time data processing - Management, Real time computer systems, Design",
+                    two.getField("keywords"));
+            Assert.assertEquals("TXT, PDF", two.getField("documenttype"));
+            Assert.assertEquals(
+                    "Aberdeen ; Birmingham ; Edinburgh ; Imperial College ; Liverpool ; Manchester ; Oxford ; Trinity College Dublin",
+                    two.getField("HL"));
+
+            Assert.assertEquals("book", two.getType());
```
simonharrer: Please have a look at how the other successfully merged PRs compared bibentries after the import...

zellerdev:
This PR is open since several months now. At this time I used the old method.
I now updated it to use BibtexEntryAssert.

stefan-kolb:
@koppor WDYT?

src/test/java/net/sf/jabref/importer/fileformat/SilverPlatterImporterTestNotRecognized.java
```diff
@@ -25,7 +25,8 @@ public void setUp() throws Exception {
     @Test
     public final void testIsNotRecognizedFormat() throws Exception {
         List<String> notAccept = Arrays.asList("emptyFile.xml", "IsiImporterTest1.isi",
-                "InspecSilverPlatterImporterTest.txt", "oai2.xml", "RisImporterTest1.ris");
+ "CopacImporterTest1.txt",
```
koppor: Please fix the formatting

src/test/java/net/sf/jabref/importer/fileformat/CopacImporterTestFiles.java
```diff
@@ -0,0 +1,76 @@
...
+        try (InputStream copacStream = CopacImporterTest.class.getResourceAsStream(fileName)) {
+
+            List<BibEntry> copacEntries = copacImporter.importEntries(copacStream, new OutputPrinterToNull());
+            fileName = fileName.replace(".txt", "");
+
+            int size = copacEntries.size();
+
+            if (size != 1) {
+                for (int i = 1; i <= size; i++) {
+                    BibtexEntryAssert.assertEquals(CopacImporterTest.class, "Bib" + fileName + "-" + i + ".bib",
```
koppor: Would it be possible to name the copac files `CopacImporter....txt` and the bib files `CopacImporter....bib`. The prefix `Bib` is not necessary.

Please remove the `Bib` prefix and adapt the filter on line 46: Add `.filter(n -> n.endsWith(".txt)` before `.collect`.

Adapt line 61 to replace `.txt` by `.bib` and just pass `fileName` here.

In case there are multiple entries in your list, possibly BibTexEntryAssert has to be modified. We have to discuss then.

src/test/resources/net/sf/jabref/importer/fileformat/InspecSilverPlatterImporterTest.txt
```diff
@@ -1,30 +0,0 @@
-Record INSPEC
```
koppor: Isn't that file really not used at the InspecImporter or SilverPlatterImportTests?

zellerdev: As far as I know only @tschechlovdev  used it in SilverPlatterImportTest.

koppor: Maybe, it should be renamed then. I wonder, why the test case do not fail when removing that file.

zellerdev: It is used in testIsNotRecognizedFormat. So there is no specific test for the fields. Does this answer your question?

koppor: I thought that Inspec and SilverPlatter are two different formats. Therefore, I would name the file accordingly. - Maybe `SilverPaltterGeneralFile.txt`

koppor: Personal discussion: test case covered by other files. Therefore, OK to remove. @tschechlovdev

simonharrer:
Can you please rebase on current master?

koppor:
Coverage [98,48%](https://codecov.io/github/JabRef/jabref/src/main/java/net/sf/jabref/importer/fileformat/CopacImporter.java?ref=ede533d62b4ebcf69f2615605622e4dcf0e239c)

[LGTM](http://lgtm.in/i/KT0PIyUZp)

src/test/java/net/sf/jabref/importer/fileformat/CopacImporterTestFiles.java
```diff
@@ -0,0 +1,81 @@
...
+        try (InputStream copacStream = CopacImporterTest.class.getResourceAsStream(fileName)) {
+
+            List<BibEntry> copacEntries = copacImporter.importEntries(copacStream, new OutputPrinterToNull());
+            fileName = fileName.replace(".txt", ".bib");
+
+            Assert.assertFalse(copacEntries.isEmpty());
+
+            int size = copacEntries.size();
+
+            // workaround because BibtexEntryAssert can only test 1 entry
```
simonharrer: Then provide a method in the BibtexEntryAssert class which can do this, please.

src/test/java/net/sf/jabref/importer/fileformat/CopacImporterTest.java
```diff
@@ -1,91 +1,84 @@
...
+    public void testIsRecognizedFormatReject() throws IOException {
         CopacImporter importer = new CopacImporter();
 
-        try (InputStream stream = CopacImporterTest.class
-                .getResourceAsStream("CopacImporterTest2.txt")) {
-            List<BibEntry> entries = importer.importEntries(stream, new OutputPrinterToNull());
-            Assert.assertEquals(2, entries.size());
-            BibEntry one = entries.get(0);
+        List<String> list = testFiles.stream().filter(n -> !n.startsWith("CopacImporterTest"))
+                .filter(n -> !n.startsWith("InspecSilverPlatterImporterTest")).collect(Collectors.toList());
```
simonharrer: Why do you filter out InspecSilverPlatter files as well? This is quite unexpected.

src/test/java/net/sf/jabref/importer/fileformat/CopacImporterTest.java
```diff
@@ -1,91 +1,84 @@
...
-            Assert.assertTrue(importer.isRecognizedFormat(stream2));
-
-            Assert.assertFalse(importer.isRecognizedFormat(stream3));
+    /**
+     * Generates a List of all files in the package "/src/test/resources/net/sf/jabref/importer/fileformat"
+     * @return A list of Names
+     */
+    public List<String> getTestFiles() {
+        List<String> files = new ArrayList<>();
+        File d = new File(System.getProperty("user.dir") + "/src/test/resources/net/sf/jabref/importer/fileformat");
```
simonharrer: Please use Java NIO instead. Moreover, you do not need `System.getProperty("user.dir")` as this is the default.

zellerdev: In the latest commit I worked in your feedback except the usage of NIO. I am currently researching it.

src/test/java/net/sf/jabref/importer/fileformat/CopacImporterTest.java
zellerdev: After talking with @tschechlovdev we both agreed to delete the InspecSilverPlatterImporterTest.txt and replace it with another Inspec file which is not recognized by CopacImporter

zellerdev:
Using java.nio for getTestFiles(), removed InspecSilverPlatter and replaced it with another Inspec file which is not recognized by copac and added a new method for BibtexEntryAssert using InputStreams and an ImportFormat as parameter to generate to lists to compare. I hope I solved everything as expected.

src/test/java/net/sf/jabref/importer/fileformat/CopacImporterTest.java
```diff
@@ -1,91 +1,85 @@
...
 
-            Assert.assertFalse(importer.isRecognizedFormat(stream3));
+    /**
+     * Generates a List of all files in the package "/src/test/resources/net/sf/jabref/importer/fileformat"
+     * @return A list of Names
+     * @throws IOException
+     */
+    public List<String> getTestFiles() throws IOException {
+        List<String> files = new ArrayList<>();
+        Files.newDirectoryStream(Paths.get(FILEFORMAT_PATH)).forEach(n -> files.add(n.getFileName().toString()));
```
simonharrer: a directory stream is a resource which must be closed. use a try-with-resources construct here, please.

PR #311, opened 2015-11-10
[WIP] OpenOffice.org/LibreOffice improvements

oscargus:
Changed default file names suitable for OO.o 4 and LO 5 and added support for having both OpenOffice.org and LibreOffice installed (sort of)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ to [sourceforge feature requests](https://sourceforge.net/p/jabref/features/) by
 ## [Unreleased]
 
 ### Changed
+ - Updated to support OpenOffice 4 and LibreOffice 5
  - Better support for IEEEtranBSTCTL entries
  - Quick selection of month in entry editor
  - Unknown entry types will be converted to 'Misc' (was 'Other' before).
--- a/src/main/java/net/sf/jabref/openoffice/AutoDetectPaths.java
+++ b/src/main/java/net/sf/jabref/openoffice/AutoDetectPaths.java
@@ -27,13 +27,13 @@ import java.awt.event.ActionEvent;
 import java.util.*;
 import java.util.List;
 
-import com.jgoodies.forms.builder.DefaultFormBuilder;
+import com.jgoodies.forms.builder.FormBuilder;
 import com.jgoodies.forms.layout.FormLayout;
 import net.sf.jabref.logic.l10n.Localization;
 import net.sf.jabref.logic.util.OS;
 
 /**
- * Tools for automatically detecting jar and executable paths to OpenOffice.
+ * Tools for automatically detecting jar and executable paths to OpenOffice and/or LibreOffice.
  */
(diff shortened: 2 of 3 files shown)


simonharrer:
What do you mean with (sort of)? Can you add a CHANGELOG entry, please? Why is OpenOffice now renamed to OpenOffice.org?

We have changed a lot of defaults from JabRef 2 to JabRef 3. Users upgrading will not get these new defaults. But I do not have an idea how to solve this with minimal effort.

oscargus:
Yes, forgot about CHANGELOG.

Sort of: it should be able to find all versions installed. However, as JabRef and OO/LO must both run in the same JRE and that OO only (easily) supports 32-bit JREs, it will be that JabRef running in a 64-bit JRE will not connect to OO, while one may still select it. There are some things related to that which I have not figured out, such as how to figure out if JabRef is running in a 32-bit or 64-bit JRE or if the 32-bit or 64-bit version of LibreOffice is installed (on Windows one can make an educated guess based on the directory it is installed in).

I should erase my JabRef preferences for a bit more testing, so it may be a good idea to put this on hold for a while.

oscargus:
Oh yeah, I was wrong about OpenOffice.org, that was the previous name. I guess OpenOffice works better than Apache OpenOffice.

Related: should we bother distinguishing between OO and LO or just use OO for both? Considering the 32-bit issue, it is easier to connect to LO at the moment (if you run a 64-bit JRE only).

oscargus:
One more thing... I should look at it more carefully to be 100% sure, but using a manual connection will update using the new "default" paths (which are present both in the Preferences and in the connection code).

This will by the way probably break using earlier versions of OO/LO. Should one keep support for those as well? Seems like more people will upgrade to OO 4-series or LO 5-series anyway (and it works with some earlier versions, not just sure when the change was done).

koppor:
I would also lean towards loosing support for old versions of OO. We don't have the man power to maintain backwards compatibility. :fire:

stefan-kolb:
What is the status of this PR? Ready for 3.0 or to be postponed for 3.1?

oscargus:
Apart from changing the OpenOffice.org naming, I think it is at least better than the current master version. Primarily since it looks for the correct paths. I will probably not have time until Thursday or Friday to sort these things out though, but I believe it makes sense to include it in 3.0.

koppor:
@oscargus Could you add a CHANGELOG.md entry? Think, you can do it within seconds - hopefully.

Then, I'll let this through!

simonharrer:
Plus can you rebase or merge so that the tests may run successfully on circleci?

oscargus:
I've done the changes (including rebase), but lack of internet access right
now.

Will commit this evening.

oscargus:
I've updated but the test fails on what looks like a totally unrelated issue.

stefan-kolb:
@simonharrer Problem with the install4j task?!

simonharrer:
Should be fixed in master now.

PR #317, opened 2015-11-11
Initial author minimizer impl

stefan-kolb:
We wanted to have some minimizer formatters for a functionality to auto-minify bibtex files for conference papers.
This PR adds a first minimizer, the author minimizer from Simon's textools project.


oscargus:
Just for info: IEEEtran.bst has a control field to handle this. Now, not everyone writes papers to be published by IEEE...

While I agree that three authors should be the limit, it may make sense to make this configurable (except that there may be lots of configurations).

Issue #318, opened 2015-11-11
Normalize to Bibtex name format


stefan-kolb:
Rightclick formatter in authors field goes wrong.

Minimal example:
```
Stefan Kolb and J{\"o}rg Lenhard and Guido Wirtz
```
turns out to be
```
Kolb, Stefan and J{\"o}rg Lenhard and Wirtz, Guido
````

but should be

``` 
Kolb, Stefan and Lenhard, J{\"o}rg and Wirtz, Guido
```

- [x] `NameListNormalizer` should be transformed to a Formatter on that occasion.

On another note: is this really THE BibTex name format? This should help with multiple firstnames?!

**Comment by oscargus** [comment]
Yes, the Lastname, Firstname is THE BibTeX name format (if you should pick one). This is the only one that can handle Jr. and it also doesn't rely on "von" being spelled correctly (without capitals) for correct results.

A (possibly) related problem: The first name abbreviation for author column (and probably elsewhere) doesn't work as expected. Not sure if this happens to J{\"o}rg, but for sure H{\aa}kan is abbreviated to H{\aa}. instead of H.

**Comment by stefan-kolb** [comment]
What about Andreas Jr. {von Lastname}?

**Comment by oscargus** [comment]
Assuming his name is Andreas von Lastname, Jr. the correct way would be "von Lastname, Jr., Andreas". Your entry results in: von Lastname, A. J. or A. J. von Lastname.

"In BibTeX's world view, a name has four components:
1) First name (which includes any middle names provided)
2) von ("de la" or "van der" like components)
3) Last (Surname without the von part)
4) Jr (Things like "Jr.", "III", etc)"

http://nwalsh.com/tex/texhelp/bibtx-23.html says

 To summarize, you may type a name in one of three forms:
"First von Last" "von Last, First" "von Last, Jr, First"
You may almost always use the first form; you shouldn't if either there's a Jr part or the Last part has multiple tokens but there's no von part.

**Comment by oscargus** [comment]
There are more issues here:

```
Heng-Yu Jian and Xu, Z. and Chang, M.-C.F.
```
becomes
```
Heng-Yu Jian and Xu, Z. and Chang, M. -. C. F.
```
but I would expect
```
Jian, Heng-Yu and Xu, Z. and Chang, M.-C. F.
```

```
Oscar Gustafsson and Linda S. DeBrunner and Victor DeBrunner and H{\aa}kan Johansson
```
becomes
```
[...]

PR #319, opened 2015-11-11
Change NameListNormalizer interface to new Formatter interface

stefan-kolb:
See #318 
- [ ] What dividers are allowed between authors? (and)


koppor:
`and` is AFAIK the bibtex default. The idea of the `NameListNormalizer` is to convert pasted data from other sources into bibtex format. This allows for copying and pasting data from a literate section and improving it without much manual work. Alternative approach: Use the build-in "Prase with freecite" feature.

koppor:
In LabelPatternUtil (e.g., in `net.sf.jabref.logic.labelPattern.LabelPatternUtil.authEtal(String, String, String)`), methods from `net.sf.jabref.model.entry.AuthorList` are used. I think, you search for `net.sf.jabref.model.entry.AuthorList.getAuthorsFirstFirstAnds()` or `net.sf.jabref.model.entry.AuthorList.getAuthorsLastFirstAnds(false)`.

stefan-kolb:
existing problems will be tricky to solve. Postponing to a later stage.

PR #320, opened 2015-11-11
Concept: Improve exception handling

matthiasgeiger:
(Inspired by #301) 

Ultimate Goal: Improving the exception handling by...
- replacing generic Exceptions by custom exceptions
- support localization already on exception level
... in the complete project.

Current status: 
- showing concept using newly introduced `JabRefException` for the preferences import

### DIFF
--- a/src/main/java/net/sf/jabref/JabRef.java
+++ b/src/main/java/net/sf/jabref/JabRef.java
@@ -400,7 +400,7 @@ public class JabRef {
         if (cli.isPreferencesExport()) {
             try {
                 Globals.prefs.exportPreferences(cli.getPreferencesExport());
-            } catch (IOException ex) {
+            } catch (JabRefException ex) {
                 LOGGER.error("Cannot export preferences", ex);
             }
         }
--- a/src/main/java/net/sf/jabref/JabRef.java
+++ b/src/main/java/net/sf/jabref/JabRef.java
@@ -206,7 +206,7 @@ public class JabRef {
                 Globals.prefs.importPreferences(cli.getPreferencesImport());
                 CustomEntryTypesManager.loadCustomEntryTypes(Globals.prefs);
                 ExportFormats.initAllExports();
-            } catch (IOException ex) {
+            } catch (JabRefException ex) {
                 LOGGER.error("Cannot import preferences", ex);
             }
         }
--- a/src/main/java/net/sf/jabref/JabRefException.java
+++ b/src/main/java/net/sf/jabref/JabRefException.java
@@ -3,7 +3,6 @@ package net.sf.jabref;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
-
 public class JabRefException extends Exception {
 
     private String localizedMessage;
@@ -30,7 +29,7 @@ public class JabRefException extends Exception {
 
     @Override
     public String getLocalizedMessage() {
-        if(localizedMessage==null) {
+        if (localizedMessage == null) {
             LOGGER.warn("No localized message exception message defined. Falling back to getMessage().");
             return getMessage();
         } else {
--- a/src/main/java/net/sf/jabref/JabRefException.java
+++ b/src/main/java/net/sf/jabref/JabRefException.java
@@ -1,0 +1,41 @@
+package net.sf.jabref;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+
+public class JabRefException extends Exception {
+
+    private String localizedMessage;
+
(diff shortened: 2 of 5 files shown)


koppor:
Looks good. Why not naming `JabRefException` `LocalizedExcpetion` to be closer to the intended usage of the class?

tobiasdiez:
Since exceptions shouldn't be used to control the flow of control (eg [here](http://programmers.stackexchange.com/questions/189222/are-exceptions-as-control-flow-considered-a-serious-antipattern-if-so-why) or [here](http://stackoverflow.com/questions/729379/why-not-use-exceptions-as-regular-flow-of-control)) the problem of localizing exceptions shouldn't occur in the first place. As far as I understood, exceptions are there to be reported to us as bugs :). 

My rule of thumb is:
- Exceptions = where failure is not expected as an option
- (Boolean) return values = where failure might be an expected result

For example, the [SaveException](https://github.com/oscargus/jabref/blob/master/src/main/java/net/sf/jabref/exporter/SaveException.java) shouldn't derive from exception. And instead of throwing a SaveException it should be the return value of the save method (and then better renamed to SaveReturnValue). 

This is at least what I'm used from other programming languages / landscapes.

matthiasgeiger:
@tobiasdiez I don't agree. 
The bottom line of the discussions you linked above is that "`Exceptions` should not be used to control the **normal** flow" (e.g. using an `Exception` to leave a loop). 
**But**, `Exceptions` are explicitly allowed and should be used for *exceptional cases*. 

*(Sticking stricly to "do not use exception to control the flow" paradigm would render* ***all*** *checked exceptions obsolete as each handling of checked exceptions steers the control flow in a program...)*

Regarding your example: When a file is saved in JabRef this should work in almost each case - an error during saving is an *exception*, which most often occurs due to an underlying `IOException`. Therefore, it is reasonable to throw an `Exception` to indicate the *exceptional termination* of a save method.

The `SaveExceptions` are used at various places in the code: In `void` methods, in methods returning a `SaveSession`, ... - so it is not possible to define one `SaveResult` which is returned from the methods which may produce an error during saving. Even if it would be possible (for example by providing a generic saveMethod which is used throughout the code - you still have to do some "exception handling": What if the method returns `false` or a String `"Saving failed due to an IOException`? The result has to be checked by the method caller and must be handled somehow. If saving was not possible, all the things which rely on a saved file cannot be executed - so you depend on constructs like this:
```java 
if (saveResult.isSuccess()) { 
  // ... do the work and return something useful
} else {
  // return null? 
  // return another "wrapper object" containting not the intended result but a flag indicating failure?
}
```
And this goes all the way through the code back to the origininal call in the UI. 

I think using a `throws` declaration at the methods and catching the exception late in the UI is the more appropriate way (especially if you see the "application logic" as an isolated part of the system which is used by the UI only through API calls).

At the end [something to read what shows why I have problems with the current way of exception handling in the JabRef code](https://today.java.net/article/2006/04/04/exception-handling-antipatterns). I think we can find more than one instance of each "anti pattern" in our code base :wink:

@koppor This should be the "root JabRef exception" replacing the usage of raw `Exception`which must be caught by an `catch (Exception e)` :cry:  It also should be extended by other more specific exception such as the  `SaveException`.
So the ability to provide a localized message is only one aspect of the `JabRefException`.

tobiasdiez:
@matthiasgeiger 
Ok, you are right in many points. Nonetheless, I think it is not a good idea to directly output exceptions to the user (except for debugging). Exceptions is something for cases where Jabref shouldn't be able to recover from the failure. If the save operation goes wrong, then the code should catch the corresponding exception and write a message to the user. In pseudocode:
```
    handle(NoSpaceException)
       show(you have not enough space on your disk to save)
    handle(FileExistsException)
       show(there exists already a file with the same name. do you want to overwrite?)
    handle(OtherException)
       error(I don't know why the save operation didn't work. please report this bug to the developers)
```
Yes, it is more work then a simple `show(exceptionMessage)` but in my opinion the better way to go.

matthiasgeiger:
We are actually doing exactly this?!

The SaveException wraps the techncial reason and is thrown to the UI where the non-technical message and the causing `Exception` are logged - and the user is informed by a dialog only showing the non-technical, localized message.

matthiasgeiger:
If there are no objections regarding the concept I'll merge this PR - further improvements can either be made en bloc in another PR or on-the-fly when working on other issues.

koppor:
LGTM :+1:

matthiasgeiger:
Hmpf... there was a comment regarding translation in one of the commits which is no longer accessible due to the rebase - However, this is obsolete as Oscar has already changed this in `master`.

PR #321, opened 2015-11-11
Endnote importer test

bartsch-dev:


koppor:
Could you add some real EndNote entries? Currently, there seem only test entries to be tested, but no real-life entries.

### Minor comments

  * Is `txt` really the extension for EndNote entries?
  * You don't need to make separate commits for each method you test. **Before** issuing the pull request, you can do `git rebase -i`, merge the commits and push afterwards. Or a `git push -f` if you already pushed the commits into your repository.
  * Please use present tense in the commit messages. See http://chris.beams.io/posts/git-commit/, 
    https://git-scm.com/book/ch5-2.html, and 
    [the discussion on stackoverflow](http://stackoverflow.com/questions/3580013/should-i-use-past-or-present-tense-in-git-commit-messages)

simonharrer:
Please rebase on master so that circleci and codecov will show correct results.

koppor:
We discovered, that EndNote is also capable of importing `RIS` format and that both IEEExplore and Springer export `RIS` as "EndNote". (See for example http://link.springer.com/chapter/10.1007/978-3-662-44879-3_8 and http://ieeexplore.ieee.org/xpl/articleDetails.jsp?arnumber=4578545).

EndNote imports both: The RISFormat and the EndNote format (es exported by JabRef). I vote for removing the EndNote importer and focus on the RisImporter.

Furthermore, I vote for removing the EndNote exporter and focus on testing the RISExporter.

koppor:
I started a [discussion at jabref-users](https://sourceforge.net/p/jabref/mailman/message/34627267/). It seems that [users want to keep support for the old format](https://sourceforge.net/p/jabref/mailman/jabref-users/thread/c01c8e1e734a0bc12db55495a8ca0e80%40mail.swcp.com/#msg34627364).

simonharrer:
Just leave it as it is right now. :)

koppor:
OK for me even though some minor issues remain: https://codecov.io/github/JabRef/jabref/src/main/java/net/sf/jabref/importer/fileformat/EndnoteImporter.java?ref=6fb8635b36920b760e3e003d169e4a4fb78626ef - line 194, line 227.

src/test/java/net/sf/jabref/importer/fileformat/EndnoteImporterTest.java
```diff
@@ -0,0 +1,126 @@
...
+    @Rule
+    public ExpectedException thrown = ExpectedException.none();
+
+    EndnoteImporter importer;
+
+    @Before
+    public void setUp() throws Exception {
+        if (Globals.prefs == null) {
+            Globals.prefs = JabRefPreferences.getInstance();
+        }
```
simonharrer: You do not need the if-condition here. Just set the `Globals.prefs`

koppor: Reference in the docs: https://github.com/JabRef/jabref/wiki/Code-Howtos#preferences-in-tests

bartsch-dev:
Fixed the mentioned issues and did some more refactoring.

src/test/java/net/sf/jabref/importer/fileformat/EndnoteImporterTest.java
```diff
@@ -0,0 +1,135 @@
...
+            assertEquals("testR", be4.getField("doi"));
+            assertEquals("testK", be4.getField("keywords"));
+            assertEquals("testO1", be4.getField("note"));
+            assertEquals("testN", be4.getField("number"));
+            assertEquals("testP", be4.getField("pages"));
+            assertEquals("testI1", be4.getField("school"));
+            assertEquals("testU", be4.getField("url"));
+            assertEquals("testV", be4.getField("volume"));
+        }
+
```
tobiasdiez: Please split this in 3 different tests.

tobiasdiez:
The tests with small test files should be tested by inline strings as described here: https://github.com/JabRef/jabref/pull/354#issuecomment-166359106

koppor:
I agree with using direct equals tests than using bib files. However, tests with full bib files should not vanish.

We told the students to have a look at [GVKParserTest.java](https://github.com/JabRef/jabref/blob/master/src/test/java/net/sf/jabref/importer/fetcher/GVKParserTest.java) and do similar testing. :innocent:

koppor:
Coverage [97.90%](https://codecov.io/github/JabRef/jabref/src/main/java/net/sf/jabref/importer/fileformat/EndnoteImporter.java?ref=824f4672ccb57cd053a2d480d420d661eb4f270a). I don't see more test cases to turn the the yellow lines into green ones.

If the test cases and test files get speaking names, I'm fine with it. @tobiasdiez Do you agree?

tobiasdiez:
Yes, looks good to me :+1: 
Just a minor point: add assertions that the bibentry-type is correct.

bartsch-dev:
I used `be0` because it's the first item in the list, I changed it to `be`.

I exported the `.enw` from a Library, I removed the space.

PR #322, opened 2015-11-11
Moved YearUtilTest to tests

oscargus:
Found this test-file among the program files, so thought I'd better move it (the test works!).


matthiasgeiger:
:+1:

PR #329, opened 2015-11-12
Moved things from the Util file

oscargus:
I've moved some methods from the Util file to more suitable places.

The destination was determined by where the method was (most) used or where it seems to fit.

As a result of this I also moved the corresponding tests to a more suitable location.


src/main/java/net/sf/jabref/model/entry/BibtexEntry.java
```diff
@@ -25,6 +25,7 @@
 import java.text.ParsePosition;
 import java.text.SimpleDateFormat;
 import java.util.*;
+import net.sf.jabref.logic.util.date.YearUtil;
```
oscargus: I will move this to model...

oscargus:
Rebased, but will have to double-check more carefully to see that everything worked out as expected.

simonharrer:
Can you please rebase so that this can be merged in?

oscargus:
I'll do that after #335 is merged as that will probably create new conflicts...

oscargus:
As far as I can tell all tests are still here.

PR #330, opened 2015-11-12
Fix #328: Added more and better information with unsaved database

### LINKED ISSUE #328: Improve dialog for saving/discarding changes on exit ->Better usability
There is room for improvement of the "Save changes on exit".

First of all: When I have opened multiple bib files and then close JabRef, the dialog pops up, but it is not clear which database I save/don't save.
jabrefsavebeforeclosing

There should be the name of the current database visible.
And furthermore it's bad UI design to ask questions in dialogs.
http://www.informit.com/articles/article.aspx?p=1146301

I suggest the following improvement:
Database changed. Click "Save changes" to  save the changes, click "Discard changes" to discard any changes. Click "return to JabRef" to return back to the main window.

The dialog itself should then have the above mentioned buttons.

This would be a huge improvent in usabilty.

oscargus:
capturedialog

(In this case it was a new database, otherwise the full path is shown.)

### DIFF
--- a/src/main/java/net/sf/jabref/gui/JabRefFrame.java
+++ b/src/main/java/net/sf/jabref/gui/JabRefFrame.java
@@ -897,10 +897,7 @@ FindUnlinkedFilesDialog.ACTION_COMMAND,
                         filename = GUIGlobals.untitledTitle;
                     }
                     int answer = JOptionPane.showOptionDialog(JabRefFrame.this,
-                            "<html><br><b>" + Localization.lang("Database %0 has changed.", filename) + "</b>\n\n"
-                                    + Localization.lang(
-                                            "Click \"%0\" to save the changes,\nclick \"%1\" to discard any changes,\nclick \"%2\" to return to the main window.",
-                                            options[0].toString(), options[1].toString(), options[2].toString()),
+                            Localization.lang("Database %0 has changed.", filename),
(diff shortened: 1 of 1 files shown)


src/main/java/net/sf/jabref/gui/JabRefFrame.java
```diff
@@ -886,12 +886,23 @@ public boolean quit() {
...
+
+                    if (baseAt(i).getFile() != null) {
+                        filename = baseAt(i).getFile().getAbsolutePath();
+                    } else {
+                        filename = GUIGlobals.untitledTitle;
+                    }
+                    int answer = JOptionPane.showOptionDialog(JabRefFrame.this,
+                            "<html><br><b>" + Localization.lang("Database %0 has changed.", filename) + "</b>\n\n"
+                                    + Localization.lang(
+                                            "Click \"%0\" to save the changes,\nclick \"%1\" to discard any changes,\nclick \"%2\" to return to the main window.",
```
simonharrer: I think we can skip this long localization as the different options are self-explanatory. Just state that the Database %0 has changed.

koppor:
@oscargus Tooltip for screenshots: [GrabNDrop](http://www.scyv.de/gnd/).

src/main/java/net/sf/jabref/gui/JabRefFrame.java
matthiasgeiger: :+1:

PR #331, opened 2015-11-12
Make model independent

simonharrer:
net.sf.jabref.model has no dependencies anymore to whatsoever in jabref. This would allow to extract these classes in a separate library. Or to put them under heavy testing as this is the core of JabRef.


koppor:
:+1:

I assume you long thought about whether the prefix `net.sf.jabref.util` is necessary in some cases. :8ball:

matthiasgeiger:
:+1: - should be merged quickly as so much files are affected... 

Will definetly produce conflicts with #329 :stuck_out_tongue_winking_eye:

Issue #104, opened 2015-08-18
search options


biohegedus:
The naming and docs for search options are not clear and/or logical for me.

koppor:
Do you mean "Incremental" vs. "Float" vs. "Show results in dialog" vs. "Global search" or something else?

biohegedus:
I mean exactly these.

simonharrer:
Do you have an idea how they should be named?

lenhard:
For the record, this seems to come from the mailing list: http://sourceforge.net/p/jabref/mailman/message/34383062/

biohegedus:
I have not thougth about the naming yet, since I do not understand these or their logic. E.g. If these options are different types of searches why the "show results in dialog" is an option in this group. These should be two option groups. One group for the search types the other for the display type. Or if all of these options indicate some of the display option or mode of the search then checking any of them and search for a word I shuld get the same results. I ran into this, since I could not perform a search within only a given group. I was tald to perform "float" to have a search only in the given group.

mlep:
Some ("Incremental", "Float", "Filter" "Show results in dialog") are different ways to display the results.
But some are about the scope ("Global" = all of the opened databases).
Right?

simonharrer:
Search: *query*, *caseSensitive*, *isRegex*.
Search (ext): Autocomplete Words

Scope: current file or all files (global)

Display: float, filter, dialog, incremental
Display (ext): Highlight Word, Select matches (does not seem to work for me.)

Incremental is an option that is more a unix style of going through the results using a next button. I am not sure if we still need such an option.

mlep:
Thank you Simon for the clarification.

About Incremental:
- we can ask users which one they use the most
- where is the next button?!? (as you see this is not my prefer display style)

In general, I conclude that:
- Help should be expanded.
- The search GUI should be reorganized to avoid mixing things, no?

simonharrer:
I suggest to make it simpler by

- removing incremental search
- make float the default search and hide the other options
- bring caseSensitive and isRegex to the front as they have an impact on the search results

I do not have good suggestions on scope and the dialog option, as they are intertwined. A global search always opens up in a dialog.

mlep:
Currently, the various possibilities take 6 lines. Let's try to keep it as compact, or even smaller
Here is the design I suggest:
- Floating | Filter | New window (Display)
(A dropdown menu with the 3 possibilities, which displays the selected one)
- Local | Global
[...]

PR #332, opened 2015-11-13
Better search

koppor:
This PR is intended to work on the search capability.

It includes https://github.com/JabRef/jabref/pull/307 and should include https://github.com/JabRef/jabref/pull/162.

There is an ongoing discussion of features at https://github.com/JabRef/jabref/issues/104#issuecomment-155466078.

- [x] concurrency bug as data structures are shared (boolean)
- [x] documentation update (english)
- [x] documentation update (german)
- [x] disable/enable open in new window and global search only when a search has been executed previously
- [x] Why is "Regex" search enabled by default? I would have expected it to be turned off as default. (I reset my preferences and started a vanilla JabRef".
- [x] Remove debug output
- [x] ESC should leave the search and focus the currently selected entry
- [x] I pressed `CTRL+f` multiple times. The view changed (behavior due to old JabRef behavior, I would like to have that removed). 
- [x] Further, I cannot switch back from "Filter" to "Float". Have to restart JabRef. Steps: "Float" is activated. Search for a string where you can see the floated area. Switch to "Filter". OK. Switch back to "Float": Nothing happens. After a restart, float works as usual.
- [x] Preview is broken in combination with the new search. When switching from "Float" to "Fitler" and back: The preview panel constantly updates and the empty area above the entry grows and grows. However, this is a separate issue. --> fixed as this was a larger concurrency issue as a lot of code was NOT executed on the Swing EDT...
- [ ] Search is not aware of currently active groups. Hence, the displayed search results (found x matches) may be strange to the user when no entries are shown at all.
- [x] When the advanced search is active, the search icon on the left is marked red. 
- [x] CRITICAL Focus is lost from the search window sometimes as the entry editor can grab the focus in some cases. Fixed by removing the set focus upon changing the entry editor tab.


simonharrer:
Implements all improvements from #104 

Please try this out.

koppor:
I don't know whether the following exception has to do something with that:

```
Exception in thread "AWT-EventQueue-0" javax.swing.text.StateInvariantError: GlyphView: Stale view: javax.swing.text.BadLocationException: Length must be
positive
        at javax.swing.text.GlyphView.getText(GlyphView.java:137)
```

koppor:
The Exception is something else. Relates to #184.

grabbed_20151120-172717

simonharrer:
- Advanced search should work. I can try this out again.

simonharrer:
This PR is open for the final review. Then, it can be merged in.

matthiasgeiger:
Documentation has still to be improved? Or is there still an "incremental search"?

"Clear search" works not as intended: After removing the text from the search box "open results in new window" still shows the results. (This button is also never disabled - which might be sensible in case no search has been performed yet).

Already asked before: Toggle search is removed intentionally?

Apart from this: Still awesome :wink:

simonharrer:
Toggle search is removed by intent, as it is annoying and hidden from the user. 
Documentation should be updated. Where is it still stating something about incremental search?
Hm, handling the open results in new window and global search could be improved, yes. Makes sense to enable/disable them appropriately. Good idea!

stefan-kolb:
LGTM :+1:

koppor:
Looks good here. Minor issues are follows. If they are hard to fix, just let it go through and fix afterwards.

- [ ] Why is "Regex" search enabled by default? I would have expected it to be turned off as default. (I reset my preferences and started a vanilla JabRef".
- [ ] Remove debug output

Regarding the second point: When searching, I lines similar to the following line:

    INFORMATION: Searching "tes" (Groß-/Kleinschreibung wird nicht unterschieden, regular expression) in bibliography.bib

Why is there some German string and why is the line output at all?

koppor:
Looks better. Still can't switch between "Filter" and "Float".

Steps: "Float" is activated. Search for a string where you can see the floated area. Switch to "Filter". OK. Switch back to "Float": Nothing happens. After a restart, float works as usual.

Side effect: When switching: The preview panel constantly updates and the empty area above the entry grows and grows. However, this is a separate issue.

simonharrer:
Fixed all existing issues. Please test it again as this uncovers a lot of issues that are related to the search in other components....

matthiasgeiger:
:+1:

koppor:
Works good on my site. I don't use groups :innocent:.

The only thing I noticed is an issue with the key/value-search

I expected

    kopp leymann

and 

    author=kopp author=leymann

To reveal the same results.

However, I have to add an `and` inbetween:

    author=koppor and author=leymann

Maybe, that's a really different issue.

matthiasgeiger:
And this is not introduced by this PR ;-)

simonharrer:
This is why you have to read the HELP file ;)

koppor:
I like the indication of the advanced search. Possibly, we could use a really different icon? I think, we only use colors for changing the background, not the font itself. If we decide to really change the color, it should be more like black or something. Red is a signal color and is typically used to indicate that something is wrong. See also the highlighting in the box if nothing is found.

Quick rendering proposal: Mirror the glass so that the circle is shown at the upper right corner. Do not change the color at all. Or use one of following icons: `torch`,  `eye`

simonharrer:
Changing
- background looks awful
- color black: is hard to see the change
- mirror is too hard to see
- other icon: torch I could not find in the material design icons. Eye is already used elsewhere. 

Hence, I leave it at red for the moment.

koppor:
Permalinks did not work here. 

Am I at the right web site?

grabbed_20151126-164936

koppor:
The name I meant was "flashlight". Sorry.

grabbed_20151126-181814

simonharrer:
Your suggested color would also be ok. I do not like it when the icon changes. This is confusing.

simonharrer:
I think we can merge this in. The problem by combining group selection and search is a much bigger effort which at the moment there is not time for. The current improvements should be good enough. Search is on my machine (i7, 8gb ram, ssd) very very fast, even with a 6800-entry-bibtex-file it is still usable with only short wait times. The problem is not the search itself, but the GUI updates associated with it. This may be optimized later on if an issue arises.

koppor:
LGTM :+1:

PR #334, opened 2015-11-13
Add "open folder" icon

tobiasdiez:
Adds a icon to the "open folder" command (under the tools menu as well
as for the menu appearing on right-click). This fixes [SF bug #1149 Missing icon for "Open folder"](https://sourceforge.net/p/jabref/bugs/1149/).

(Before you ask: the line `prefs.getKey(KeyBinds.OPEN_FOLDER)` is formatted in this way by eclipse on save...)

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,5 @@
 [master]
+    - Fixed bug SF #1149: Missing icon for "Open folder"
     - Feature: Search DOAJ, Directory of Open Access Journals
     - The key bindings for searching specific databases are removed
     - Removes option to toggle native file dialog on mac by making JabRef always use native file dialogs on mac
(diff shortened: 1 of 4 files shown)


oscargus:
To get nicer formatting (I recognize the annoying behaviour): mark a suitable part of the code (like the whole function call in this case, i.e. rows 403-407) and press Shift+Ctrl+F.

simonharrer:
Hm, the issue with this is that we now use the same icon for opening an existing database and opening a folder. Because of this, we did not add the folder icon to the open folder action as we only want to use unique icons per action.

tobiasdiez:
What about instead using [library-plus](https://materialdesignicons.com/icon/library-plus) and [library-books](https://materialdesignicons.com/icon/library-books) for new and open database, respectively? They would also harmonize nicely with the [Save all icon](https://materialdesignicons.com/icon/content-save-all).

matthiasgeiger:
As the icon is no longer used in the IconBar (where there was a 1*1px "missing icon" before) I do not think we need an icon for "open folder" which would only be used in the right-click menu and menu bar.

PR #335, opened 2015-11-13
Clean-up of StringUtil and some added tests

oscargus:
Replaced/removed/unified some methods in StringUtil and added some tests in StringUtilTest.


simonharrer:
Can you please rebase on current master so that I can merge this in?

simonharrer:
I will merge this in. But you missed a few tests as the code coverage shows for the StringUtil class.

oscargus:
Yes, but they will show later (I moved the tests twice in different PRs...).

oscargus:
The missing test (makeBibtexExtension) is because I removed that function and made use of getCorrectFileName instead (where ever that change is...).

oscargus:
Oh, yeah. You mean that it is not fully tested? Correct! I only added tests for those methods I played with (and I don't know how to get that coverage information without pushing and creating a PR...).

koppor:
[EclEmma](http://eclemma.org/) seems to be the tool of choice for code coverage.

PR #336, opened 2015-11-13
Inlined Abbreviation button from JournalAbbreviationButton to EntryEd…

oscargus:
Moved the method generating the journal abbreviation button and inlined it in EntryEditor.

I changed (and corrected) the translations strings and while I agree that the append-approach is not preferred there was a reason: If any of the menu items, "Options" and "Manage journal abbreviations", are translated, but the tool tip is not, it will still give the correct information. The reason for not using %0 etc is that anything after the first string to be inserted is not translatable.

### DIFF
--- a/src/main/java/net/sf/jabref/gui/entryeditor/EntryEditor.java
+++ b/src/main/java/net/sf/jabref/gui/entryeditor/EntryEditor.java
@@ -481,7 +481,7 @@ public class EntryEditor extends JPanel implements VetoableChangeListener, Entry
                 public void actionPerformed(ActionEvent actionEvent) {
                     String text = editor.getText();
                     if (Abbreviations.journalAbbrev.isKnownName(text)) {
-                        String s = Abbreviations.journalAbbrev.getNextAbbreviation(text).orElse(text);
+                        String s = Abbreviations.toggleAbbreviation(text);
 
                         if (s != null) {
                             editor.setText(s);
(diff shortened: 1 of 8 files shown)


simonharrer:
I get the thing about localization. But I do not get why you inlined the static method.

oscargus:
To me it seemed easier from a maintenance perspective to inline it instead of having static functions in a different file that is only used once anyway. Maybe one should to the same with toggleAbbreviation in fact, since it is only used once a few lines up.

simonharrer:
I think the `toggleAbbreviation` method should be moved to the logic where it can be tested easily.

oscargus:
That's an even better idea! :+1: I'll do that.

oscargus:
I ended up inlining the function, but adding a test for the particular call used there.

src/main/java/net/sf/jabref/gui/entryeditor/EntryEditor.java
```diff
@@ -467,8 +474,29 @@ public void mouseClicked(MouseEvent e) {
...
+            // Button to toggle abbreviated/full journal names
+            JButton button = new JButton(Localization.lang("Toggle abbreviation"));
+            button.setToolTipText(ABBREVIATION_TOOLTIP_TEXT);
+            button.addActionListener(new ActionListener() {
+
+                @Override
+                public void actionPerformed(ActionEvent actionEvent) {
+                    String text = editor.getText();
+                    if (Abbreviations.journalAbbrev.isKnownName(text)) {
+                        String s = Abbreviations.journalAbbrev.getNextAbbreviation(text).orElse(text);
```
simonharrer: Can you move this line so that instead of `Abbreviations.journalAbbrev.getNextAbbreviation(text).orElse(text);` one can call `Abbreviations.journalAbbrev.toggleAbbreviation(text);`?

The aim is to have as much code as possible in logic and independent of the GUI classes.

oscargus:
I reintroduced the toggleAbbreviation method as requested, but at the same time found some other unused methods which I removed.

PR #337, opened 2015-11-13
Replaced EntryTypes.getType('string') with direct access to the corre…

oscargus:
No more EntryTypes.getType("article"), instead BibtexEntryTypes.ARTICLE etc is directly accessed.

### DIFF
--- a/src/main/java/net/sf/jabref/importer/fileformat/SixpackImporter.java
+++ b/src/main/java/net/sf/jabref/importer/fileformat/SixpackImporter.java
@@ -28,6 +28,8 @@ import net.sf.jabref.bibtex.EntryTypes;
 import net.sf.jabref.model.entry.IdGenerator;
 import net.sf.jabref.model.entry.BibtexEntry;
 import net.sf.jabref.model.entry.BibtexEntryType;
+import net.sf.jabref.model.entry.BibtexEntryTypes;
+
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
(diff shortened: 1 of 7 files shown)


src/main/java/net/sf/jabref/importer/fileformat/SixpackImporter.java
```diff
@@ -139,20 +141,18 @@ public boolean isRecognizedFormat(InputStream stream) throws IOException {
...
                     }
                     if (fields[1].equals("PhD-Thesis")) {
-                        type = "phdthesis";
+                        typ = BibtexEntryTypes.PHDTHESIS;
                     }
                     if (fields[1].equals("miscellaneous")) {
-                        type = "misc";
+                        typ = BibtexEntryTypes.MISC;
                     }
                     if (fields[1].equals("Conference")) {
```
matthiasgeiger: why not using `else if` in those checks?

matthiasgeiger:
:+1:

src/main/java/net/sf/jabref/importer/fileformat/SixpackImporter.java
oscargus: Yeah, I asked myself that as well in the back of my head, totally focused on replacing stuff. :-)

But I'll change it now. Good point!

simonharrer:
This might destroy biblatex support. As EntryTypes.getType may return a biblatex version if biblatex is activated.

oscargus:
OK! So then it should be the other way around? Never use BibtexEntryTypes.*?

simonharrer:
Hm, would make sense, yes. But there may be exceptions, probably. We should investigate this.

koppor:
Damn it, we should really investigate it. We really like constants and it is common practice to use constants instead of strings. In the meantime, a comment should be added to BibtexEntryTypes.* to never use them in the code (and reference this pull request). :innocent:

PR #338, opened 2015-11-13
Removed unused gui.components classes

oscargus:
These are not used so I thought one might just as well get rid of them...


matthiasgeiger:
:+1:

PR #339, opened 2015-11-14
Fix issue that prevents circleci to build successfully in PRs from br…

simonharrer:
…anches outside of JabRef repo when environment variables are not available


oscargus:
I have no idea if that code is correct, but the line break I blame on Gmail...

Issue #114, opened 2015-08-19
Fix line breaking


koppor:
Somehow, the abstract and keyword fields are incorrectly wrapped. See https://sourceforge.net/p/jabref/discussion/318825/thread/9609d582/ for details.

stefan-kolb:
The bug should be caused by `FieldContentParser`

```text
abstract = {The Quick Brown Fox Jumps Over The Lazy Dog. The Quick Brown Fox Jumps\r\n
	\tOver The Lazy Dog. The Quick Brown Fox Jumps Over The Lazy Dog. The
```

```java
 if (content.length() > i + 1 && content.charAt(i + 1) == '\t' && (content.length() == i + 2 || !Character.isWhitespace(content.charAt(i + 2)))) {
// We have either \n\t followed by non-whitespace, or \n\t at the
// end. Both cases indicate a wrap made by JabRef. Remove and insert space if necessary.
// @formatter:on

content.deleteCharAt(i); // \n
content.deleteCharAt(i); // \t
```

stefan-kolb:
The question here is, which fields should accept line breaks?

koppor:
All multiline fields :joy:

Arent't they enlisted in the code somewhere?

Multiline: Abstract, Review
Unsure: comment
Non-multiline: all others

Side note: No one has complained about the difference between UI and serialization. To have the `.bib` files editable, I would keep the autowrap-feature (refs #123).

sbitzer:
Hi, I'm very much for keeping some line breaks in multiline fields. I previously used the review field to save my own thoughts about a paper. For some entries this has been quite a bit of text which I had separated into paragraphs by putting an empty line between paragraphs in the entry editor. I recently noticed that linebreaks are removed when the database is saved. When it is saved the first time, my empty lines are gone. When it is saved the second time, the remaining linebreaks are gone, too. Needless to say that I found that all my old reviews are only a single paragraph mess now. I don't think that this was an issue earlier and I'm a bit upset about the change in behaviour.

My jabref version is 2.10b2.

koppor:
Related issue: https://sourceforge.net/p/jabref/feature-requests/854/

stefan-kolb:
The writer should also not write line breaks for non-multiline fields...

lenhard:
Fixed #187

PR #342, opened 2015-11-15
Added tests for most methods in StringUtils 

oscargus:
The ones that remain are:
- [ ] getPart - not really clear to me how it works
- [ ] findEncodingsForString - unused, but seems quite useful (e.g. we could possibly suggest an encoding when trying to save with a bad encoding)

Related change in this PR:
* Change in FileListTableModel, which basically reimplemented StringUtil.encodeStringArray

### DIFF
--- a/src/main/java/net/sf/jabref/gui/FileListEntry.java
+++ b/src/main/java/net/sf/jabref/gui/FileListEntry.java
@@ -57,6 +57,11 @@ public class FileListEntry {
         this.type = type;
     }
 
+    public String[] getStringArrayRepresentation() {
+        String type = getType() != null ? getType().getName() : "";
+        return new String[] {getDescription(), getLink(), type};
+    }
+
     @Override
     public String toString() {
         return description + " : " + link + " : " + type;
--- a/src/main/java/net/sf/jabref/logic/util/strings/StringUtil.java
+++ b/src/main/java/net/sf/jabref/logic/util/strings/StringUtil.java
@@ -173,7 +173,7 @@ public class StringUtil {
 
         String[] lines = in.split("\n");
         StringBuilder result = new StringBuilder();
-        addWrappedLine(result, lines[0], wrapAmount);
+        addWrappedLine(result, lines[0].replaceAll("\\s+$", ""), wrapAmount);
         for (int i = 1; i < lines.length; i++) {
 
             if (!lines[i].trim().equals("")) {
@@ -224,10 +224,6 @@ public class StringUtil {
         return result.toString();
     }
 
-    public static String quote(String toQuote, String specials, char quoteChar) {
-        return quote(toQuote, specials, quoteChar, 0);
-    }
-
     /**
      * Quote special characters.
      *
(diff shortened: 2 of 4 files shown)


src/main/java/net/sf/jabref/gui/FileListEntry.java
```diff
@@ -57,6 +57,11 @@ public void setType(ExternalFileType type) {
         this.type = type;
     }
 
+    public String[] getStringArrayRepresentation() {
+        String type = getType() != null ? getType().getName() : "";
+        return new String[] {getDescription(), getLink(), type};
```
simonharrer: Do we need String arrays or can we use an instance of a class instead?

src/main/java/net/sf/jabref/logic/util/strings/StringUtil.java
```diff
@@ -167,7 +167,7 @@ public static String wrap(String in, int wrapAmount) {
 
         String[] lines = in.split("\n");
         StringBuilder result = new StringBuilder();
-        addWrappedLine(result, lines[0], wrapAmount);
+        addWrappedLine(result, lines[0].replaceAll("\\s+$", ""), wrapAmount);
```
simonharrer: What does this regex mean? And why is it necessary?

oscargus: It is stated a few lines down. The idea is to remove \r if the line breaking is \r\n, but it turned out that it didn't work on the first line, only on the remaining ones.

However, now I realize that if the line-breaking is only \r (MAC, right?) the complete function will fail its purposed. Still, the current version is at least slightly better than the earlier...

src/main/java/net/sf/jabref/gui/FileListEntry.java
oscargus: I was thinking about that as well. ArrayList<ArrayList<String>>? But then I wasn't sure which was the most suitable one.

This especially holds for the encode and decode functions in StringUtil.

src/main/java/net/sf/jabref/logic/util/strings/StringUtil.java
oscargus: I guess the actual question here is if we should line wrap fields when writing to the .bib-file or not? It can also be noted that there's a comment stating that this function should be "symmetric" to the parse function above, which doesn't exist... Hence, I guess that maybe one should wait a bit with sorting out this. @lenhard ?

koppor: The RegEx is `CharMatcher.WHITESPACE.trimTrailingFrom` (see http://www.leveluplunch.com/java/examples/right-trim-string/).

koppor: Regarding the wrapping: I think, the entry point for the discussion is https://github.com/JabRef/jabref/issues/114. Think we agreed not not wrap single-line fields and that multiline fields should not be merged into one line.

oscargus: After reading the code a few times, I think that if the preference "Line wrap as in 2.9.2" is removed, so is this function. May be worth double-checking a few times though... :-)

lenhard:
The symmetry between parser and writer is realized through the class `FieldContentParser`, which is used during parsing (`BibtexParser`) and writing (`LatexFieldFormatter`). Both of these in turn use `StringUtil`. We agreed to leave field content alone from now on and that includes wrapping from my point of view.

oscargus:
OK! So even fixing a bug in the wrapping? :-)

lenhard:
It looks like that. good job ;-)

src/test/java/net/sf/jabref/logic/util/strings/StringUtilTest.java
```diff
@@ -57,53 +57,128 @@ public void testPutBracesAroundCapitals() {
...
+
+
+    String[][] stringArray1 = {{"a", "b"}, {"c", "d"}};
+    String encStringArray1 = "a:b;c:d";
+    String[][] stringArray2null = {{"a", null}, {"c", "d"}};
+    String encStringArray2 = "a:;c:d";
+    String[][] stringArray2 = {{"a", ""}, {"c", "d"}};
+    String encStringArray2null = "a:" + null + ";c:d";
+    String[][] stringArray3 = {{"a", ":b"}, {"c;", "d"}};
+    String encStringArray3 = "a:\\:b;c\\;:d";
```
simonharrer: make them static final

simonharrer:
findEncodingForStrings is removed in #396 

Hm, I would suggest to create test cases for getParts and try to get an idea what this method does by the test case results. From what I understand it does a kind of substring which is also affected by curly braces.

oscargus:
Done. Still tests for two methods are missing, but I think it makes sense to merge it as I do not understand what getParts does and findEncodingsForStrings is not used ( @tobiasdiez is this method of interest in relation to #396 ? Now that you have a fresh view of the encoding.).

oscargus:
OK! I assume that #396 will be merged eventually. Yes, I agree that it theoretically seems like a feasible way to go. :-) Well, apart from a few rebase commands, I guess that there's no obvious drawback in not merging this PR now.

oscargus:
I'll have a go at getParts eventually. Seems like we had thread-unsafe commenting at the end. :-)

simonharrer:
Ok, just open up a new PR for this.

PR #343, opened 2015-11-15
[WIP] Replaced deprecated Glazedlists-API with current

oscargus:
I replaced the deprecated calls to the glazedlists-API with the current versions. The major difference is that the new ones are not thread-safe and as such in most placed I've used the suggested factory method to obtained a wrapped list/model.

This seems to work, but may require a bit more testing/thinking. Especially to figure out if the wrapping may not be needed in all places and if there actually should be wrapping where there currently isn't.

A better option would be to explicitly lock and unlock the tables, but that require much more work.


matthiasgeiger:
Is still [WIP] or can this be merged?

oscargus:
Not sure. I think it works, but since it is a threading issue, it is hard to know for sure. However, using a naïve approach doesn't work (I've tried that earlier), so something is working better now. It is probably more likely that some tables are thread-safe where it isn't required.

simonharrer:
Hm, can you rebase this? I think then we should merge this as it improves the code still.

oscargus:
Rebased.

Issue #349, opened 2015-11-16
We need to add further info to the database tabs to distinguish open references


stefan-kolb:
Often reference files have the same name over multiple papers, e.g. references.bib.
When open in JabRef all tabs have the same name making it difficult to distinguish the different files.
We should add more context to make this easier for the users.
@simonharrer suggested to add the number of references in (100).
An alias retrieved from somewhere would solve this problem best, but is hard to determine.

simonharrer:
Other alternative could be to determine the minimal unique path between the files as a header. So when one uses different file names, everything stays the same. But for the entries that have the same file names, add parent folders until the name is unique. Should not be too difficult to determine.

tobiasdiez:
Is this a big problem? If this additional data in the bib file is inconvenient for the user, then he just don't have to specify a display name. Alternatively, we can save the display name in Jabref's config instead of the file.

stefan-kolb:
Another drawback is that the name must be set explicitely, this should be an (automatic) help feature to distinguish databases that share the same name, which is (at least for me) often the case when using the same base template.

koppor:
@wielanms and me also favour the idea of a display name (see also https://github.com/JabRef/jabref/issues/341#issuecomment-157182168). The initial display name could be derived automatically. Think, there are already solutions used in other software. Using the first two letters of each path element etc. The default can be overwritten by the user. The name could be written in the header of the BibTeX file (and not using `@comment{jabref-meta: ...`).

The naming scheme by Simon is also interesting. However, the display name will change depending on other files which may be misleading. Therefore, I propose to use `<parent folder name>/<bib file name>` as initial display name.

simonharrer:
Why make the label of the tab long when it is unnecessary? The problem I see is when `bib` files are stored in `bibliography` folders instead of in the root folder directly, which is common in larger tex projects.

matthiasgeiger:
I do not like the idea of a display name...

How about: 
`filename.bib` iff this label is unique.
`.../parent/filename.bib` iff this label is unique.
`.../parent/parent/filename.bib` iff this label is unique
...

Apart from this I think we are already applying the best practices from other tools: Showing the full path of the currently open bib DB in the window title and tool tips with the full path for each tab.

simonharrer:
:+1: for the sublime idea with `$filename [-- $unique-folder-path]`

stefan-kolb:
:+1:  for implementing the sublime behavior.

stefan-kolb:
[...]

PR #351, opened 2015-11-16
Show folder info on tab if file names are identical

stefan-kolb:
- Tab shows the minimal unique folder name substring if multiple database files share the same name

### DIFF
--- a/src/main/java/net/sf/jabref/gui/JabRefFrame.java
+++ b/src/main/java/net/sf/jabref/gui/JabRefFrame.java
@@ -126,10 +126,10 @@ public class JabRefFrame extends JFrame implements OutputPrinter {
         @Override
         public void actionPerformed(ActionEvent e) {
             IntegrityCheck check = new IntegrityCheck();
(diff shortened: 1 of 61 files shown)


src/main/java/net/sf/jabref/gui/JabRefFrame.java
```diff
@@ -1646,14 +1644,12 @@ public void addTab(BasePanel bp, File file, boolean raisePanel) {
                 title = title + '*';
             }
         } else {
-            title = file.getName();
+            title = file.getName() + " (" + bp.database.getEntryCount() + ")";
```
simonharrer: One minor suggestion: the code to determine the title is distributed in two places. One could create a method in BasePanel which computes the title and just reuse the method from there.

stefan-kolb: Of course this should be done. I just don't know if the majority wants this feature so just demoing it here.

simonharrer:
#349

PR #354, opened 2015-11-17
Medline plain importer test

boceckts:
Fixed IndexOutOfBounds for MedlinePlainImporter when Testfile has lines with too few characters. Removed unreachable code. Created 100% Testcoverage for MedlinePlainImporter.


simonharrer:
Please use rebase --interactive to squash the commits with messages like `""` or `see last commit`.

src/test/java/net/sf/jabref/importer/fileformat/MedlinePlainImporterTest.java
```diff
@@ -0,0 +1,120 @@
...
+
+        Assert.assertFalse(Testimporter
+                .isRecognizedFormat(MedlinePlainImporter.class.getResourceAsStream("IsiImporterTestMedline.isi")));
+
+        Assert.assertFalse(Testimporter.isRecognizedFormat(
+                MedlinePlainImporter.class.getResourceAsStream("MedlinePlainImporterTestEmpty.txt")));
+
+        //Test with txt File who is in recognized Format
+        Assert.assertTrue(Testimporter
+                .isRecognizedFormat(MedlinePlainImporter.class.getResourceAsStream("MedlinePlainImporterTest.txt")));
```
simonharrer: Please see #310 for how to handle tests for this part.

src/test/java/net/sf/jabref/importer/fileformat/MedlinePlainImporterTest.java
```diff
@@ -0,0 +1,120 @@
...
+    @Test
+    public void testGetCLIId() {
+        MedlinePlainImporter Testimporter = new MedlinePlainImporter();
+        String expected = Testimporter.getCLIId();
+        String actual = "";
+        Assert.assertFalse(expected.equals(actual));
+        actual = "MedlinePLain";
+        Assert.assertFalse(expected.equals(actual));
+        actual = "medlineplain";
+        Assert.assertTrue(expected.equals(actual));
```
simonharrer: Why not use `Assert.assertEquals`? Expected and actual must be switched, as expected is the fixed output String you know is correct whereas actual is the result of the method under test. 

`Assert.assertEquals("medlineplain", importer.getCLIId());`

What is more, variables must always start with a lower case letter and camel case (except when they are constants).

koppor:
@boceckts `java.lang.StringIndexOutOfBoundsException: String index out of range: -1` during test. Please fix :innocent:

koppor:
@boceckts Please squash the whole branch history into one commit. The magic command is `git rebase master`. See https://github.com/JabRef/jabref/pull/379#issuecomment-158661990 for a longer discussion.

boceckts:
Commits sind nun zu einem Commit zusammengefasst, index out of bounds exception gefixt und die Testmethoden sind nun leicht abgeändert zu AssertEquals.

koppor:
Commit looks good :+1:, but the test fail. Could you please investigate?

boceckts:
Welche Tests genau? Wenn ich die Klasse MedlinePlainImporterTest.java als JUnit Test ausführe, laufen alle Test bei mir ohne Fehler..

src/main/java/net/sf/jabref/importer/fileformat/MedlinePlainImporter.java
```diff
@@ -99,7 +99,7 @@ public boolean isRecognizedFormat(InputStream stream) throws IOException {
 
         for (String entry1 : entries) {
 
-            if (entry1.trim().isEmpty()) {
+            if (entry1.trim().isEmpty() || (entry1.indexOf('-') < 0)) {
```
simonharrer: !contains is better readable then indexOf, I think.

koppor:
Siehst du hier beim Pull-Request die Meldung "All checks have failed"? Screenshot:

grabbed_20151122-153227

Du kannst dann auf "Details" bei "ci/circleci" klicken. Dort steht dann "junit failures: testIsRecognizedFormat - in net.sf.jabref.importer.fileformat.MedlinePlainImporterTest". Hier kannst du auf "more info" klicken. Da steht dann

```
	at java.io.Reader.<init>(Reader.java:78)
	at java.io.InputStreamReader.<init>(InputStreamReader.java:97)
	at net.sf.jabref.importer.ImportFormatReader.getReaderDefaultEncoding(ImportFormatReader.java:352)
	at net.sf.jabref.importer.fileformat.MedlinePlainImporter.isRecognizedFormat(MedlinePlainImporter.java:69)
	at net.sf.jabref.importer.fileformat.MedlinePlainImporterTest.testIsRecognizedFormat(MedlinePlainImporterTest.java:50)
```

boceckts:
@simonharrer changed it to !contains.
@koppor Seems like one character in the name of the test file was a different case than what I wrote in the java test class.

koppor:
Your test file looks good. Seems to cover seven different entry types with different fields and different settings (doi, multine line titles). However, your test only treats two of them. Why don't you check all test entries?

koppor:
@boceckts We are ready to go from our side. Please have a look at [GVKParserTest.java](https://github.com/JabRef/jabref/blob/1332070cb62a8990567831a10f026d31acfc1738/src/test/java/net/sf/jabref/importer/fetcher/GVKParserTest.java).

boceckts:
The Unit Test failes because multiple comments are being imported with a newline separator. However in the bibtex source the new lines are being ommitted therefore the assertion fails when testing with a import file using multiple comments such as "MedlinePlainImporterTest2.txt vs MedlinePlainImporterTestBib2.bib".

src/test/java/net/sf/jabref/importer/fileformat/MedlinePlainImporterTest.java
```diff
@@ -0,0 +1,127 @@
...
+        Assert.assertEquals("06", test2.getField("month"));
+        Assert.assertNull(test2.getField("author"));
+        Assert.assertNull(test2.getField("title"));
+    }
+
+    @Test
+    public void testEmptyImport() throws Exception {
+        //Test entries with empty txt File
+        List<BibEntry> emptyEntries = importer.importEntries(
+                MedlinePlainImporter.class.getResourceAsStream("MedlinePlainImporterTestEmpty.txt"),
```
simonharrer: Always use try-with-resources for `getResourceAsStream` usages.

src/test/java/net/sf/jabref/importer/fileformat/MedlinePlainImporterTest.java
```diff
@@ -0,0 +1,127 @@
...
+        //Test with txt Files which are not in recognized Format
+        for (String str : list) {
+            try (InputStream is = MedlinePlainImporter.class.getResourceAsStream(str)) {
+                Assert.assertFalse(importer.isRecognizedFormat(is));
+            }
+        }
+
+        //Test with txt File which is in recognized Format
+        Assert.assertTrue(importer.isRecognizedFormat(MedlinePlainImporter.class
+                .getResourceAsStream("MedlinePlainImporterTest.txt")));
```
simonharrer: what about `MedlinePlainImporterTest1.txt` ?

src/test/java/net/sf/jabref/importer/fileformat/MedlinePlainImporterTest.java
```diff
@@ -0,0 +1,127 @@
...
+     * test multiple entries in one import file
+     *
+     * @throws Exception
+     */
+    @Test
+    public void testImportEntries() throws Exception {
+
+        //create entries list from the txt file
+        List<BibEntry> entries = importer.importEntries(
+                MedlinePlainImporter.class.getResourceAsStream("MedlinePlainImporterTest.txt"),
```
simonharrer: Always use try-with-resources for `getResourceAsStream` usages.

src/test/java/net/sf/jabref/importer/fileformat/MedlinePlainImporterTest.java
```diff
@@ -0,0 +1,127 @@
...
+        }
+
+        //Test with txt File which is in recognized Format
+        Assert.assertTrue(importer.isRecognizedFormat(MedlinePlainImporter.class
+                .getResourceAsStream("MedlinePlainImporterTest.txt")));
+
+    }
+
+    /**
+     * test multiple entries in one import file
```
simonharrer: use this for naming the test method, e.g., `public void testMultipleEntriesInSingleImportFile`, because then you do not need the comment anymore.

src/test/java/net/sf/jabref/importer/fileformat/MedlinePlainImporterTest.java
```diff
@@ -0,0 +1,127 @@
...
+import net.sf.jabref.JabRefPreferences;
+import net.sf.jabref.bibtex.BibtexEntryAssert;
+import net.sf.jabref.importer.OutputPrinterToNull;
+import net.sf.jabref.model.entry.BibEntry;
+
+/**
+ * Test class for the MedlinePlain importer.
+ *
+ */
+
```
simonharrer: This empty line can be removed

src/test/java/net/sf/jabref/importer/fileformat/MedlinePlainImporterTest.java
```diff
@@ -0,0 +1,127 @@
...
+public class MedlinePlainImporterTest {
+
+    private MedlinePlainImporter importer;
+
+
+    @Before
+    public void setUp() throws Exception {
+        if (Globals.prefs == null) {
+            Globals.prefs = JabRefPreferences.getInstance();
+        }
```
simonharrer: Just remove the if-condition and only set `Globals.prefs = JabRefPreferences.getInstance();` as this must not be optimized here.

src/test/java/net/sf/jabref/importer/fileformat/MedlinePlainImporterTest.java
```diff
@@ -0,0 +1,127 @@
...
+                    + ".txt");
+                    InputStream nis = MedlinePlainImporter.class.getResourceAsStream("MedlinePlainImporterTestBib" + n
+                            + ".bib")) {
+
+                List<BibEntry> entries = importer.importEntries(is, new OutputPrinterToNull());
+                Assert.assertNotNull(entries);
+                Assert.assertEquals(1, entries.size());
+                BibtexEntryAssert.assertEquals(nis, entries.get(0));
+            }
+        }
```
simonharrer: You can write it more clearly by separating the input pairs from the logic. Make it obvious what is getting tested. For instance:

```
for (int n = 1; n <= 5; n++) {
    String medlineFile = "MedlinePlainImporterTest" + n + ".txt";
    String bibtexFile = "MedlinePlainImporterTestBib" + n + ".bib";
    assertImportOfMedlineFileEqualsBibtexFile(medlineFile, bibtexFile);
}
```

And the assert method you can implement based on the other code.

boceckts:
Now using try-with-resource whenever I use the ResourceAsStream methode.
Renamed methodes and removed some comments. 
Added all test files to a list to check for recognized format.
Extracted test logic from testImportSingleEntriesInSingleFiles to separate it from the actual testing data.

src/test/java/net/sf/jabref/importer/fileformat/MedlinePlainImporterTest.java
boceckts: Which line exactly do you mean?

src/test/java/net/sf/jabref/importer/fileformat/MedlinePlainImporterTest.java
```diff
@@ -0,0 +1,122 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+import net.sf.jabref.Globals;
+import net.sf.jabref.JabRefPreferences;
+import net.sf.jabref.bibtex.BibtexEntryAssert;
+import net.sf.jabref.importer.OutputPrinterToNull;
+import net.sf.jabref.model.entry.BibEntry;
+
+/**
```
tobiasdiez: Comments can be removed since it contains exactly the same information as the class name.

src/test/java/net/sf/jabref/importer/fileformat/MedlinePlainImporterTest.java
```diff
@@ -0,0 +1,122 @@
...
+                "MedlinePlainImporterTestEmpty.txt");
+
+        //Test with txt Files which are not in recognized Format
+        for (String str : list) {
+            try (InputStream is = MedlinePlainImporter.class.getResourceAsStream(str)) {
+                Assert.assertFalse(importer.isRecognizedFormat(is));
+            }
+        }
+
+        list = Arrays.asList("MedlinePlainImporterTest.txt", "MedlinePlainImporterTest1.txt",
```
tobiasdiez: Split tests. Definitely a separate test for recognized and one for not-recognized files. I would even say that there should be a separate test for every file (maybe not for all files, just the ones which actually test a new behavior...say one for txt files and one for isi).

src/test/java/net/sf/jabref/importer/fileformat/MedlinePlainImporterTest.java
```diff
@@ -0,0 +1,122 @@
...
+            }
+        }
+    }
+
+    @Test
+    public void testImportMultipleEntriesInSingleFile() throws Exception {
+        try (InputStream is = MedlinePlainImporter.class.getResourceAsStream("MedlinePlainImporterTest.txt")) {
+            //create entries list from the txt file
+            List<BibEntry> entries = importer.importEntries(is, new OutputPrinterToNull());
+            Assert.assertEquals(7, entries.size());
```
tobiasdiez: If there are 7 entries then you should also test that all entries are parsed correctly and not just the first two. Probably it is more convenient to reduce the number of entries in the test file and thus simplify the test case.

src/test/java/net/sf/jabref/importer/fileformat/MedlinePlainImporterTest.java
```diff
@@ -0,0 +1,122 @@
...
+            BibEntry test2 = entries.get(1);
+            Assert.assertEquals("06", test2.getField("month"));
+            Assert.assertNull(test2.getField("author"));
+            Assert.assertNull(test2.getField("title"));
+        }
+    }
+
+    @Test
+    public void testEmptyFileImport() throws Exception {
+        try (InputStream is = MedlinePlainImporter.class.getResourceAsStream("MedlinePlainImporterTestEmpty.txt")) {
```
tobiasdiez: There is no need to create an empty file. Better create a stream from an empty string, see http://stackoverflow.com/questions/782178/how-do-i-convert-a-string-to-an-inputstream-in-java.

src/test/java/net/sf/jabref/importer/fileformat/MedlinePlainImporterTest.java
```diff
@@ -0,0 +1,122 @@
...
+            List<BibEntry> emptyEntries = importer.importEntries(is, new OutputPrinterToNull());
+            Assert.assertEquals(0, emptyEntries.size());
```
tobiasdiez: Test against an empty list as described here https://github.com/JabRef/jabref/wiki/Code-Howtos#test-cases.

src/test/java/net/sf/jabref/importer/fileformat/MedlinePlainImporterTest.java
```diff
@@ -0,0 +1,122 @@
...
+        }
+    }
+
+    @Test
+    public void testImportSingleEntriesInSingleFiles() throws IOException {
+        for (int n = 1; n <= 5; n++) {
+            String medlineFile = "MedlinePlainImporterTest" + n + ".txt";
+            String bibtexFile = "MedlinePlainImporterTestBib" + n + ".bib";
+            assertImportOfMedlineFileEqualsBibtexFile(medlineFile, bibtexFile);
```
tobiasdiez: Please directly test that the imported entries are correct (as it is done above). The test as it stands could also fail because the bib-file import has a bug.

src/test/java/net/sf/jabref/importer/fileformat/MedlinePlainImporterTest.java
boceckts: Split tests is a great idea. I think one test method for isRecognized and one for isNotRecognized should be enough since this methods should not test any data but only check if the format is recognized or not.

tobiasdiez: Yes, you are right, normally one should be enough. But have a look at the `isRecognized` method and see if you cover all cases. Also it does not harm to check edge cases, like empty files.

tobiasdiez:
Good work :+1: . 
Some tests are hard to read since they compare a medline import with a bib import. So it may be hard to see why they fail.

Better test specific features in small test cases (and maybe leave one big "integration" test). For example, (using pseudo-code)
```` java
public importParsesDoi() {
  Stream stream = StreamFromString("PMID-22664220" + "\n" + "AID - doi:10.1016/j.cpr.2005.02.002");
  List<Entries> entries = Import(stream);

  BibEntry expected = new BibEntry();
  expected.setField("doi", "10.1016/j.cpr.2005.02.002"); 
  Assert(entries = Collections.singletonList(expected));
}
````
could replace the test `4`. See also https://github.com/JabRef/jabref/wiki/Code-Howtos#test-cases and http://stackoverflow.com/questions/782178/how-do-i-convert-a-string-to-an-inputstream-in-java.

File: src/test/java/net/sf/jabref/importer/fileformat/MedlinePlainImporterTest.java (hunk 13 above)
Comment (boceckts) [inline src/test/java/net/sf/jabref/importer/fileformat/MedlinePlainImporterTest.java]: this test should only check if multiple entries in one file could be read and imported together. The specific test for an entry is being checked in the testImportSingleEntriesInSingleFiles methode using the methode suggested by @koppor with checking imported entry vs. bibtex file.

**Comment by koppor** [comment]
Your test fail, because `equals` is not working on the canonical representation. We are afraid that changing `equals` (and `hashCode`) leads to unwanted side effects.

Therefore, please add a method to `BibtexEntryAssert` for comparing Lists (please use the List interface, not the ArrayList implementation). There, you can use `assertEquals(BibEntry shouldBeEntry, BibEntry entry)` for comparing the entries themselves.

**Comment by koppor** [comment]
Please also try to get rid of the warnings by eclipse. I think try-with-resources is the solution, isn't it?

grabbed_20151223-182645

**Comment by boceckts** [comment]
#579

**Comment by boceckts** [comment]
Added more test methodes to cover more if conditions and included the suggestions of the comments.

File: src/test/java/net/sf/jabref/importer/fileformat/MedlinePlainImporterTest.java [hunk 17]
```diff
@@ -0,0 +1,171 @@
...
+            expectedEntry.setType(EntryTypes.getType("article"));
+            BibtexEntryAssert.assertEquals(Arrays.asList(expectedEntry), actualEntries);
+        }
+    }
+
+    @Test
+    public void testGetFormatName() {
+        Assert.assertNotEquals("", importer.getFormatName());
+        Assert.assertNotEquals("medlineplain", importer.getFormatName());
+        Assert.assertEquals("MedlinePlain", importer.getFormatName());
```
tobiasdiez: Only the last (positive) assert statement is necessary.

src/test/java/net/sf/jabref/importer/fileformat/MedlinePlainImporterTest.java
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    @Test
+    public void testGetCLIId() {
+        Assert.assertNotEquals("", importer.getCLIId());
+        Assert.assertNotEquals("MedlinePlain", importer.getCLIId());
+        Assert.assertEquals("medlineplain", importer.getCLIId());
```
tobiasdiez: Same as above.

src/test/java/net/sf/jabref/importer/fileformat/MedlinePlainImporterTest.java
```diff
@@ -0,0 +1,171 @@
...
+        Assert.assertFalse(importer.isRecognizedFormat(emptyFileStream));
+    }
+
+    @Test
+    public void testImportMultipleEntriesInSingleFile() throws Exception {
+        try (InputStream is = MedlinePlainImporter.class
+                .getResourceAsStream("MedlinePlainImporterTestMultipleEntries.txt")) {
+
+            List<BibEntry> entries = importer.importEntries(is, new OutputPrinterToNull());
+            Assert.assertEquals(7, entries.size());
```
tobiasdiez: You should verify all 7 entries or delete the last 5 in the test file.

tobiasdiez:
Looks very good to me! I just pointed out some smaller things. If they are fixed, this can be merged in (from my point of view).

boceckts:
Thank you for the comments, I removed the unnecessary NotEquals and added tests for the remaining entries in the test file.

koppor:
[Coverage 98.65%](https://codecov.io/github/JabRef/jabref/src/main/java/net/sf/jabref/importer/fileformat/MedlinePlainImporter.java?ref=6d09b72bd7cd97b6ab537905fe6142748f87ce8c). I assume that covering [line 126](https://codecov.io/github/JabRef/jabref/src/main/java/net/sf/jabref/importer/fileformat/MedlinePlainImporter.java?ref=6d09b72bd7cd97b6ab537905fe6142748f87ce8c#l-126) was difficult, because there is no entry where the fourth character of the second line is NOT `-`. Or similar :smiley:

PR #355, opened 2015-11-17
Add purple background in Windows 8.1+ start menu

tobiasdiez:
Add a gray background in JabRef's start menu tile (for Windows 8.1 and 10) by implementing the [VisualElementsManifest](https://msdn.microsoft.com/en-us/library/windows/apps/dn393983.aspx).
screen

Note: I couldn't test the install procedure since `gradlew releaseWindows` gives an error:
```
[ERROR] [org.gradle.BuildExceptionReporter] > A problem occurred starting process 'command 'launch4j''
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -333,6 +333,7 @@ task releaseWindows(dependsOn: "launch4j") {
     doLast {
         copy {
             from("$projectDir/LICENSE")
+            from("$projectDir/buildres/JabRef.VisualElementsManifest.xml")		
             from("$projectDir/README.md")
             from("$buildDir/launch4j/JabRef.exe")
             into("$buildDir/nsis")
--- a/buildres/JabRef.VisualElementsManifest.xml
+++ b/buildres/JabRef.VisualElementsManifest.xml
@@ -1,0 +1,4 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Application xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+  <VisualElements ForegroundText="light" BackgroundColor="#5C6FAD" ShowNameOnSquare150x150Logo="on"></VisualElements>
+</Application>
--- a/buildres/JabRef.VisualElementsManifest.xml
+++ b/buildres/JabRef.VisualElementsManifest.xml
@@ -1,4 +1,4 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Application xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
-  <VisualElements ForegroundText="light" BackgroundColor="#5C6FAD" ShowNameOnSquare150x150Logo="on"></VisualElements>
+  <VisualElements ForegroundText="light" BackgroundColor="#4F5F8F" ShowNameOnSquare150x150Logo="on"></VisualElements>
 </Application>


koppor:
Regarding the test with installer. Please consult https://github.com/JabRef/jabref#release-process

simonharrer:
:+1: we are currently in the process of moving to install4j installers. This should be integrated there as well.

stefan-kolb:
Why make it gray? The picture is purple, should it be purple?!

tobiasdiez:
Strange... on my system it is more gray than purple but on the screenshot it is more purple. Anyway, I took the main color in the jabref icon and made it a little bit lighter.

stefan-kolb:
What about taking the exact same color, to free the 'R' as Windows does with their icons?

tobiasdiez:
I oriented myself on MS Office.
untitled

stefan-kolb:
So what about this one? image

stefan-kolb:
You were faster :smile: 
Good work :+1:

simonharrer:
:+1:

stefan-kolb:
@JabRef/developers Does this work for any of you? Not working for me in the current setup.

koppor:
What does "not working" mean? Here, there is a difference with and without `JabRef.VisualElementsManifest.xml` in `C:\Program Files\jabref`. I do the following command to let Windows know that I played around with the manifest file:

    (ls "C:\ProgramData\Microsoft\Windows\Start Menu\Programs\JabRef\JabRef.lnk").lastwritetime = get-date

System: Windows 8.1 Pro, JabRef 64bit.

What I don't see is a background with one color, I see a background with some shading. However, the color changes depended of the presence or absence of `JabRef.VisualElementsManifest.xml`.

stefan-kolb:
I don't see a custom background at all. I'm on Windows 10.

matthiasgeiger:
So it is currently a manual step, which should be integrated in the new install4j process (if possible)?

koppor:
No, no manual Step, install4j works perfectly. Only for debugging issues there are manual steps required.

matthiasgeiger:
Okay... does not work for me. After installing and selecting "Pin to 'Start'" i only see the an uncolored background on Win10 (but I'm not sure whether if have changed some Windows prefs regarding this feature as I do not use the start tiles for other "Apps")

simonharrer:
Hm, install4j should put the file in the folder with the exe file.

stefan-kolb:
It is placed inside the jabref folder but has no effect for me. Maybe Matthias and I have overidden anything in Windows? Does it work for you Simon?

koppor:
Windows 10 Pro Version 1511 Build 10586.11 displays purple background.

grabbed_20151125-114228

Steps taken: Download 64 bit installer from http://builds.jabref.org/master/. Install. Open Windows Startmenu (additional step since I use [Classic Shell](http://www.classicshell.net/)). Right click on JabRef (should be there), "Pin to Start menu" (or similar). At the right side: Scroll down and see all apps. The last app is "JabRef" and displayed as shown above.

tobiasdiez:
I think the problem happens if you already had Jabref installed on the computer and/or installed the new version in the same folder. The problem is then that the shortcut pointing to the executable (usually `C:\ProgramData\Microsoft\Windows\Start Menu\Programs\JabRef\JabRef.lnk`)
is not updated. But Windows just reads the manifest file if this shortcut changes. Does it work if you invoke the command posted by @koppor in https://github.com/JabRef/jabref/pull/355#issuecomment-159449481?

stefan-kolb:
The command does not work for me.
```
(ls "C:\Users\Stefan\AppData\Roaming\Microsoft\Windows\Start Menu\Programs\JabRef\JabRef.lnk").lastwritetime = get-date
".lastwritetime" kann syntaktisch an dieser Stelle nicht verarbeitet werden.
```

I'll try to update Windows to the latest Patch release and see if it works, at least it does for Simon.

simonharrer:
Works on my machine (tm) ;-)

I uninstalled all JabRef versions before.

matthiasgeiger:
(not working after removing all jabref versions and a new clean install in another for win10 build 10240)

tobiasdiez:
@stefan-kolb did you tried it in a powershell or usual command shell? should work as powershell.
@matthiasgeiger uninstalling Jabref might not work because the lnk file is not removed during uninstall?

matthiasgeiger:
I've installed in a new start menu "folder" so this should not be the problem.

matthiasgeiger:
Works now after update to Win10 1511 Build 10586 :+1:

koppor:
@stefan-kolb Tooltip: [ConEmu](https://conemu.github.io/). Dropdown right of the "+": "Shells" -> "PowerShell"

stefan-kolb:
Ok, so it is probably the update that fixes the problem. The other solutions do not work for me.

simonharrer:
Maybe related: https://github.com/koppor/jabref/issues/57

PR #356, opened 2015-11-17
Test for InspecImporter

mairdl:
Completed code coverage and discovered a bug in BibtexEntryType (BibtetEntryType.getType() ALL_TYPES.get() returns null)


src/test/java/net/sf/jabref/importer/fileformat/InspecImportTest.java
```diff
@@ -0,0 +1,157 @@
...
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+
+
+    }
+    @Test
+    public void testImportEntries3() {
+
+        InputStream inStream = InspecImportTest.class.getResourceAsStream("InspecImportTest4.txt");
```
simonharrer: try-with-resources please to ensure that the input stream is closed.

src/test/java/net/sf/jabref/importer/fileformat/InspecImportTest.java
```diff
@@ -0,0 +1,157 @@
...
+        try {
+            List<BibtexEntry> entries = inspecImp.importEntries(InspecImportTest.class
+                    .getResourceAsStream("InspecImportTest4.txt"), new OutputPrinterToNull());
+            assertEquals(1, entries.size());
+            BibtexEntry entry = entries.get(0);
+            assertEquals(BibtexEntryTypes.OTHER, entry.getType());
+
+
+        } catch (IOException e) {
+            e.printStackTrace();
```
simonharrer: just annotate the testImportEntries3 with throws IOException. We want to avoid e.printStackTrace() at all cost as it is evil. ;)

src/test/java/net/sf/jabref/importer/fileformat/InspecImportTest.java
```diff
@@ -0,0 +1,157 @@
...
+    @Test
+    public void testIsRecognizedFormat() throws IOException {
+
+        InputStream inStream = InspecImportTest.class.getResourceAsStream("InspecImportTest.txt");
+        InputStream inStreamFalse = InspecImportTest.class.getResourceAsStream("InspecImportTestFalse.txt");
+
+        Assert.assertTrue("InspecImporter isRecognizedFormaTrue test failed",
+                inspecImp.isRecognizedFormat(inStream));
+        Assert.assertFalse("InspecImporter isRecognizedFormatFalse test failed",
+                inspecImp.isRecognizedFormat(inStream));
```
simonharrer: Have a look at #310 for handling the tests for format recognition.

src/test/java/net/sf/jabref/importer/fileformat/InspecImportTest.java
mairdl: shall i do the same for the other testImportEntries test as well?

simonharrer: :+1:

src/test/java/net/sf/jabref/importer/fileformat/InspecImportTest.java
mairdl: I editet the format recognition, so that it equals the one of #310

simonharrer: Hm, I do not see this in the current PR.

src/test/java/net/sf/jabref/importer/fileformat/InspecImportTest.java
```diff
@@ -0,0 +1,178 @@
...
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+import net.sf.jabref.Globals;
+import net.sf.jabref.JabRefPreferences;
+
+/**
+ *
+ * This is the test class of the InspecImporter. It tests all 4 methods of the class.
```
simonharrer: Can you remove this comment, please? The name of the class name says exactly that.

src/test/java/net/sf/jabref/importer/fileformat/InspecImportTest.java
```diff
@@ -0,0 +1,178 @@
...
+        }
+        this.inspecImp = new InspecImporter();
+
+    }
+
+    /**
+     * @throws java.lang.Exception
+     */
+    @After
+    public void tearDown() throws Exception {
```
simonharrer: An empty method should be removed.

src/test/java/net/sf/jabref/importer/fileformat/InspecImportTest.java
mairdl: Of course i can remove that, but isn't giving classes a JavaDoc comment in the sense of SE?

simonharrer: Yes, but only when it has more information as the class name. The class name already states that this class contains tests of the InspecImporter. Why should I duplicate this in a comment? Especially in a test class which always has the purpose to include tests for the methods of the class it is created to test. 

JavaDoc is especially useful for application code. But, then again, one should only add comments that have value, e.g., contain examples, metaphors, explanation on how to use this method and class, pre/post conditions, invariants, ...

mairdl: thanks for clearing that up :+1:

src/test/java/net/sf/jabref/importer/fileformat/InspecImportTest.java
```diff
@@ -0,0 +1,144 @@
...
+
+
+        try(InputStream inStream = InspecImportTest.class.getResourceAsStream("InspecImportTest2.txt")) {
+            List<BibtexEntry> entries = inspecImp.importEntries(inStream, new OutputPrinterToNull());
+            assertEquals(1, entries.size());
+            BibtexEntry entry = entries.get(0);
+
+            assertEquals("The SIS project : software reuse with a natural language approach", entry.getField("title"));
+            assertEquals("Prechelt, Lutz", entry.getField("author"));
+            assertEquals(" 1992", entry.getField("year"));
```
koppor: Why is the year imported with a space in front, this is strange.

src/test/java/net/sf/jabref/importer/fileformat/InspecImportTest.java
```diff
@@ -0,0 +1,144 @@
...
+            assertEquals("Abstrakt", entry.getField("abstract"));
+            assertEquals("key", entry.getField("keywords"));
+            assertEquals("10000", entry.getField("journal"));
+            assertEquals("20", entry.getField("pages"));
+            assertEquals(" 19", entry.getField("volume"));
```
koppor: Why is there a space in front in the volume? This is strange.

src/test/java/net/sf/jabref/importer/fileformat/InspecImportTest.java
mairdl: My bad I understood it the wrong way. I'll fix it in the next commit.

src/test/java/net/sf/jabref/importer/fileformat/InspecImportTest.java
mairdl: Have a look at the test files (line 8), there are spaces between the ". ; :". If there was no space it would return ".1992".

src/test/java/net/sf/jabref/importer/fileformat/InspecImportTest.java
mairdl: Have a look at the test files (line 8), there are spaces between the ". ; :". If there was no space it would return ";19"

koppor: I assume, the test files reflect real files. Then, please improve the importer by stripping white spaces from the fields "volumes" (and other fields where white spaces were added).

mairdl: In the beginning i though we weren't allowed to change the Import classes, that's why i created the tes tlike this, I'll fix that in the next commit, so that it strips the white spaces off.

mairdl:
It seems like my commits have some issues, but I don't get the problem. Can someone help me out?

oscargus:
I do not think it is your commit that causes problems. For some reason the SpringerLinkFetch test fails.

koppor:
@mairdl I think, it would make sense if you squash all commits into a single commit and to a `git push -f`. You could also do a `git rebase master` as described at https://github.com/JabRef/jabref/pull/379#issuecomment-158661990. After having done that, the rest by CircleCI should not fail any more.

src/test/java/net/sf/jabref/importer/fileformat/InspecImportTest.java
```diff
@@ -0,0 +1,108 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+public class InspecImportTest {
+
+    InspecImporter inspecImp;
+
+    @Before
+    public void setUp() throws Exception {
+        if (Globals.prefs == null) {
```
tobiasdiez: If statement not necessary.

src/test/java/net/sf/jabref/importer/fileformat/InspecImportTest.java
```diff
@@ -0,0 +1,108 @@
...
+        }
+    }
+
+    @Test
+    public void testImportEntries2() throws IOException{
+        try(InputStream inStream = InspecImportTest.class.getResourceAsStream("InspecImportTest3.txt")) {
+            List<BibEntry> entries = inspecImp.importEntries(inStream, new OutputPrinterToNull());
+            assertEquals(1, entries.size());
+            BibEntry entry = entries.get(0);
+            assertEquals(BibtexEntryTypes.INPROCEEDINGS, entry.getType());
```
tobiasdiez: If this is everything you want to test, then the test file can be simplified (i.e. strip everything except the RT line). Then it is also possible to inline the file content as a string and load the stream from the string. Also please adapt the test name to something more meaningful, like `importConferencePaperGivesInproceedings`.

src/test/java/net/sf/jabref/importer/fileformat/InspecImportTest.java
```diff
@@ -0,0 +1,108 @@
...
+        }
+    }
+
+    @Test
+    public void testImportEntries3()throws IOException {
+        try (InputStream inStream = InspecImportTest.class.getResourceAsStream("InspecImportTest4.txt")){
+            List<BibEntry> entries = inspecImp.importEntries(inStream, new OutputPrinterToNull());
+            assertEquals(1, entries.size());
+            BibEntry entry = entries.get(0);
+            assertEquals(BibtexEntryTypes.MISC, entry.getType());
```
tobiasdiez: Same as for the test above.

src/test/java/net/sf/jabref/importer/fileformat/InspecImportTest.java
```diff
@@ -0,0 +1,109 @@
...
+
+            assertEquals("The SIS project : software reuse with a natural language approach", entry.getField("title"));
+            assertEquals("Prechelt, Lutz", entry.getField("author"));
+            assertEquals("1992", entry.getField("year"));
+            assertEquals("Abstrakt", entry.getField("abstract"));
+            assertEquals("key", entry.getField("keywords"));
+            assertEquals("10000", entry.getField("journal"));
+            assertEquals("20", entry.getField("pages"));
+            assertEquals("19", entry.getField("volume"));
+            assertEquals(BibtexEntryTypes.ARTICLE.getName().toLowerCase(), entry.getType());
```
simonharrer: Pleas use compare the canonical bibentry format with one written in a file. With the current code, the expected and the actual bibentry is hard to see, and it does not detect if the actual bibentry has more key/value pairs than the expected one.

tobiasdiez: No, Simon, this is the better way to test for equality. You really test that the result is what you want and not just that the 'canonical serialization' is correct (I could cheat and just store a serialization in the bib entry without filling the actual fields. Then all the tests would pass but of course this is not what you want). Especially, the two methods below are way more readable then two separate input-output files. 

You are right, that this test doesn't fail if more fields are present, but this is #625.

src/test/java/net/sf/jabref/importer/fileformat/InspecImportTest.java
```diff
@@ -0,0 +1,109 @@
...
+            BibEntry entry = entries.get(0);
+
+            assertEquals("The SIS project : software reuse with a natural language approach", entry.getField("title"));
+            assertEquals("Prechelt, Lutz", entry.getField("author"));
+            assertEquals("1992", entry.getField("year"));
+            assertEquals("Abstrakt", entry.getField("abstract"));
+            assertEquals("key", entry.getField("keywords"));
+            assertEquals("10000", entry.getField("journal"));
+            assertEquals("20", entry.getField("pages"));
+            assertEquals("19", entry.getField("volume"));
```
tobiasdiez: write directly "article" here (same also below)

src/test/java/net/sf/jabref/importer/fileformat/InspecImportTest.java
```diff
@@ -0,0 +1,109 @@
...
+        try (InputStream inStream = new ByteArrayInputStream(testInput.getBytes())) {
+            List<BibEntry> entries = inspecImp.importEntries(inStream, new OutputPrinterToNull());
+            assertEquals(1, entries.size());
+            BibEntry entry = entries.get(0);
+            assertEquals(BibtexEntryTypes.MISC.getName().toLowerCase(), entry.getType());
+        }
+    }
+
+    @Test
+    public void testGetFormatName() {
```
tobiasdiez: "testGetFormatName failed" can be removed (it doesn't add any information...) same below

tobiasdiez:
LGTM :+1: 
(can be merged in if the small comments are fixed)

src/test/java/net/sf/jabref/importer/fileformat/InspecImportTest.java
simonharrer: You can have the input/output file pairs inline, if you like. That is not an issue. 

Well, you could cheat but why would you? For me, a test is much easier to understand if I can reduce it to a simple input/output system. If I have to rely on a canonical format, this is not a unit test anymore, I admit. But testing only the input/output pairs you have an acceptance test which could be shown to a user as well and the user would understand this. And the higher a test, the more stable it will be. A unit test in this regard may be modified if you change any of the methods of the bibentry. 

To me, the input/output is much more maintainable, as this could be easily applied to all importers. Hence, we could just add triples of (importer, input, output) and this could be used to test the importer. Even users could create test cases and a single test class could implement this as a data driven tests.

tobiasdiez: I don't understand what you are trying to say. Right now the test is of the format "input -> processing -> compare against expected output". So it perfectly fits the setting you described. 

So, I think the only question is how "compare against expected output" should be defined. In an ideal world, I would just write `parsedEntry.equals(expectedEntry)` and let the BibEntry class take care of what "equality" should mean. Since this is not possible right now, the next best thing for me is to compare that I get the data I want from the entry (which is what the test is doing).
In the end, it comes down to the type of the "expected output". For me it is an BibEntry and for you a string. I really like the typed version more! (And by the way, I have a really strong opinion about this one :smile: ...so maybe we should shift the discussion to the next dev call)

simonharrer: In production code, I completely agree: To compare one BibEntry with another one, a BibEntry must provide an equals method. We should implement one in the near future. Same holds for a BibDatabase which basically is a List of BibEntries. 

In test code, such an equals method does not help as it can only either give you true or false. Because of this, one uses Assert.assertEquals("a", "b") instead of Assert.assertTrue("a".equals("b")). Hence, we need a more sophisticated comparison method which can also show the difference, giving reasons why it failed. I think, we also agree on this one. 

I think we have to distinguish between the representation of the expected output and the comparison. 

We can represent the expected output using the canonical format in a file, in a String or by using the API to build up a BibEntry. I think using a file is more maintainable, as even normal users could provide us with test cases - but I do not have a strong opinion about this. All three alternatives are OK. 

So it boils down to how to achieve the comparison of two entries, or even two databases (list of entries) as the importers import a file which can contain more than one entry. I can also live with a comparison method which takes in two lists of BibEntries and compares them internally (while I personally prefer the canonical serialization as well as one would see all differences between two BibEntries quickly instead of only getting the single difference which caused the test to fail without knowing what other fields were also the problem). But nevertheless, we should require and use such a comparison method to ensure that no additional (and suprising) fields are found during the import. I thought we had such a comparison method already available, but I can only find `BibtexEntryAssert.assertEquals(List<BibEntry> expected, List<BibEntry> actuals)` which uses the canonical format internally for the comparison instead of calling the BibEntry methods directly.

tobiasdiez: Ok, I think we both want the same thing in the end!

As far as I know, the assertEquals method just compares the objects by calling the `equals` method and if they differ returns the values of `toString`. So to have a proper support we need to overwrite equals (compare the internal fields arrays) and toString (return the canonical serialization). So we need a solution until the corresponding #625 is implemented.

I would propose to create the bib entry via 
`BibEntry expectedEntry = new BibEntry()` and `expectedEntry.setField("field", "data")`. Then call `assertEquals(parsedEntry.getCanonicalSerialization(), expectedEntry.getCanonicalSerialization())`. 
As soon as we have a proper implementation of the equals method, the `getCanonicalSerialization` part can be removed.
What do you think about this solution?

koppor: The used method should be `net.sf.jabref.bibtex.BibtexEntryAssert.assertEquals(BibEntry, BibEntry)`.

koppor: @tobiasdiez One cannot cheat and store a canonical representation as `net.sf.jabref.model.entry.CanonicalBibtexEntry.getCanonicalRepresentation(BibEntry)` walks through the fields.

Or did I get something wrong?

Maybe we should write down the motivation for `CanonicalBibtexEntry` and `BibtexEntryAssert` somewhere to make it clear. This current discussion is a great input for that. See also https://github.com/JabRef/jabref/blob/master/CONTRIBUTING.md#when-making-an-architectural-decision :innocent:

tobiasdiez: Good point @koppor. Then use `BibtexEntryAssert.assertEquals(BibEntry, BibEntry)`.
And yes, in my opinion we should remove this class completely. If two entries are (or lists of them) are equal in some sense for the tests, then the production code should also regard them as equal.

simonharrer: I like the idea of having a toString as canonical representation and equals method. Good idea!

Hence, we can conclude that in this PR `BibtexEntryAssert.assertEquals(BibEntry, BibEntry)` should be used which is later changed accordingly.

simonharrer:
:+1: with the condition that the comments above are implemented (see assert equals of BibEntry)

Braunch:
Thats fine with me since I got lost in that discussion. I'll implement that in the evening or tomorrow.

src/test/java/net/sf/jabref/importer/fileformat/InspecImportTest.java
tobiasdiez: @mairdl can you change the code accordingly? Thanks!

Braunch: Am I getting it right that the BibtexEntryAssert makes the other assertions obsolete?

tobiasdiez: Yes instead of 
```
assertEqual("bla", entry.getField("field"));
assertEqual("foo", entry.getField("field2"));
```
you now write
```
BibEntry expected = new Entry();
expected.setField("field1", "bla")
expected.setField("field2", "foo")
BibtexEntryAssert.assertEqual(expected, entry);
```

Braunch: The assertEquals(BibEntry shouldBeEntry, BibEntry entry) Method is private. Is there any reason not to change this to public?

tobiasdiez: With https://github.com/JabRef/jabref/commit/2a80e9c98d5940e4ef2276902a9b8a244c303893 It is public since 4 hours :)

Braunch:
I reverted the branch but the parallel changes on the assert method did not get merged, since my last commit made it private again. Before investing more time in git stuff I just made it public again.

src/test/java/net/sf/jabref/importer/fileformat/InspecImportTest.java
```diff
@@ -0,0 +1,118 @@
...
+        shouldBeEntry.setField("pages", "20");
+        shouldBeEntry.setField("volume", "19");
+
+        try (InputStream inStream = InspecImportTest.class.getResourceAsStream("InspecImportTest2.txt")) {
+            List<BibEntry> entries = inspecImp.importEntries(inStream, new OutputPrinterToNull());
+
+            assertEquals(1, entries.size());
+            BibEntry entry = entries.get(0);
+
+            BibtexEntryAssert.assertEquals(shouldBeEntry, entry);
```
tobiasdiez: Replace this with BibtexEnrtyAssert.assertEquals(Arrays.ListOf(shouldBeEntry), entries).

koppor:
@Braunch: Do reduce the mess, just rebase on recent master and squash everything into one commit. A description is available at https://github.com/JabRef/jabref/wiki/Tools#rebase-everything-as-one-commit-on-master

Braunch:
Is it ready for a merge now?

tobiasdiez:
`testCompleteBibtexEntryOnJournalPaperImport` still has one not addressed comment. Please rename the commit to something more meaning full like `Test for InspecImporter`. Other than that :+1: for merge.

src/test/java/net/sf/jabref/importer/fileformat/InspecImportTest.java
```diff
@@ -0,0 +1,121 @@
...
+    public void importConferencePaperGivesInproceedings() throws IOException {
+        String testInput = "Record.*INSPEC.*\n" +
+                "\n" +
+                "RT ~ Conference-Paper";
+        try (InputStream inStream = new ByteArrayInputStream(testInput.getBytes())) {
+            List<BibEntry> entries = inspecImp.importEntries(inStream, new OutputPrinterToNull());
+            inStream.close();
+            assertEquals(1, entries.size());
+            BibEntry entry = entries.get(0);
+            assertEquals(BibtexEntryTypes.INPROCEEDINGS.getName().toLowerCase(), entry.getType());
```
simonharrer: Replace this with BibtexEntryAssert.assertEquals(Arrays.ListOf(shouldBeEntry), entries) as well.

src/test/java/net/sf/jabref/importer/fileformat/InspecImportTest.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    @Test
+    public void importConferencePaperGivesInproceedings() throws IOException {
+        String testInput = "Record.*INSPEC.*\n" +
+                "\n" +
+                "RT ~ Conference-Paper";
+        try (InputStream inStream = new ByteArrayInputStream(testInput.getBytes())) {
+            List<BibEntry> entries = inspecImp.importEntries(inStream, new OutputPrinterToNull());
+            inStream.close();
```
simonharrer: remove this, it is not necessary in try-with-resoruces

src/test/java/net/sf/jabref/importer/fileformat/InspecImportTest.java
```diff
@@ -0,0 +1,121 @@
...
+    @Test
+    public void importMiscGivesMisc() throws IOException {
+        String testInput = "Record.*INSPEC.*\n" +
+                "\n" +
+                "RT ~ Misc";
+        try (InputStream inStream = new ByteArrayInputStream(testInput.getBytes())) {
+            List<BibEntry> entries = inspecImp.importEntries(inStream, new OutputPrinterToNull());
+            assertEquals(1, entries.size());
+            BibEntry entry = entries.get(0);
+            assertEquals(BibtexEntryTypes.MISC.getName().toLowerCase(), entry.getType());
```
simonharrer: Replace this with BibtexEntryAssert.assertEquals(Arrays.ListOf(shouldBeEntry), entries) as well.

src/test/java/net/sf/jabref/importer/fileformat/InspecImportTest.java
```diff
@@ -0,0 +1,121 @@
...
+
+        assertEquals("INSPEC", inspecImp.getFormatName());
+
+    }
+
+    @Test
+    public void testGetCLIId() {
+
+        assertEquals("inspec", inspecImp.getCLIId());
+
```
simonharrer: remove empty lines

src/test/java/net/sf/jabref/importer/fileformat/InspecImportTest.java
```diff
@@ -0,0 +1,121 @@
...
+            BibEntry entry = entries.get(0);
+            assertEquals(BibtexEntryTypes.MISC.getName().toLowerCase(), entry.getType());
+        }
+    }
+
+    @Test
+    public void testGetFormatName() {
+
+        assertEquals("INSPEC", inspecImp.getFormatName());
+
```
simonharrer: remove empty lines

src/test/java/net/sf/jabref/importer/fileformat/InspecImportTest.java
mairdl: Are you sure you mean "Arrays.ListOf(shouldBeEntry)"? Because I can't find this methode.

tobiasdiez: Simon meant the `Arrays.asList`, but `Collections.singletonList` also works (I think IntelliJ even suggest replacing asList with singletonList)

mairdl: If i try to solve it with `Arrays.asList` or with `Collections.singletonList`, I get an error from the assertion because I try to compare a list of BibEntry with a BibEntry.

tobiasdiez: But `entries` is a list, isn't it?

mairdl: `entries` is a `BibEntry` as suggested in earlier comments

koppor:
@Braunch: I don't know, what you did, but all commits seem to be lost. Please use `gitk` to investigate and ask someone else (@mairdl?) if he still has a local copy. He should NOT delete his local code. If all of you deleted it, you have to use the hard man git tooling and look for: `git reflog` is your friend. See for example at http://gitready.com/advanced/2009/01/17/restoring-lost-commits.html.

src/test/java/net/sf/jabref/importer/fileformat/InspecImportTest.java
koppor: Please reference the comment. You can get the deep link to a comment using the time given at a commit. While scrolling through here, I don't see that comment directly. Please help the reader by clear references.

mairdl: @koppor Here you go [,](https://github.com/JabRef/jabref/pull/356#discussion_r55875576)

mairdl:
@koppor I think Chris(braunch) squashed all commits into 1

koppor:
@mairdl It looks like follows on my site:

grabbed_20160318-231122

10ffb1c is an empty commit and 3e4c73c shows a removal of empty lines. Where is the single commit you are talking about?

mairdl:
@koppor Now i understand what you mean, my bad. I'll contact @Braunch.

Braunch:
I hope my last commit fixed that. Otherwise I am doing something wrong with git.

stefan-kolb:
Tests are failing, please investigate.

mairdl:
Now the tests are working (Y)

stefan-kolb:
@tobiasdiez @koppor  @simonharrer Are you happy enough with it now? We can squash the commits via the merge button.

src/test/java/net/sf/jabref/importer/fileformat/InspecImportTest.java
tobiasdiez: please fix this

src/test/java/net/sf/jabref/importer/fileformat/InspecImportTest.java
```diff
@@ -0,0 +1,118 @@
...
+    public void importMiscGivesMisc() throws IOException {
+        String testInput = "Record.*INSPEC.*\n" +
+                "\n" +
+                "RT ~ Misc";
+        BibEntry shouldBeEntry = new BibEntry();
+        shouldBeEntry.setType("Misc");
+
+        try (InputStream inStream = new ByteArrayInputStream(testInput.getBytes())) {
+            List<BibEntry> entries = inspecImp.importEntries(inStream, new OutputPrinterToNull());
+            assertEquals(1, entries.size());
```
tobiasdiez: Use BibtexEntryAssert.assertEquals(Collections.singletonList(shouldbeEntry), entries);

simonharrer:
Interesting that the code coverage decreases according to codecov by 0.85% in a PR which aims to increase test coverage. Very strange.

oscargus:
I think it is because the code in this PR is based on an older code base, so fewer tests included and the comparison is against the (at the time of the push) current master. Push again and it would decrease > 2%...

PR #359, opened 2015-11-17
Add test for bibtex key generation and refactoring

tobiasdiez:
Verify that special latex characters in author names are handled correctly in bibtex key generation (from [SF bug 1186](http://sourceforge.net/p/jabref/bugs/1186/)).

Small code cleanup (`makeLabel` now returns voids since it updates the entry instead of returning a updated one)


stefan-kolb:
:+1:

PR #361, opened 2015-11-17
Fix bug: encoding was not read from comment

tobiasdiez:
The encoding was not correctly inferred from the first/second line of the file, i.e. from
````
% Encoding: encoding
```
This bug is covered in the tests by changing the default encoding to something different from the encoding in the test files. Fixed this bug by completely rewriting `checkForEncoding`.

### DIFF
--- a/src/main/java/net/sf/jabref/importer/OpenDatabaseAction.java
+++ b/src/main/java/net/sf/jabref/importer/OpenDatabaseAction.java
@@ -19,11 +19,13 @@ import java.awt.event.ActionEvent;
 import java.io.BufferedReader;
 import java.io.File;
 import java.io.IOException;
+import java.io.InputStreamReader;
 import java.io.Reader;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Optional;
 
 import javax.swing.Action;
 import javax.swing.JOptionPane;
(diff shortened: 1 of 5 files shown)


koppor:
The current code does not allow for putting the encoding manually in the header, maybe with empty lines at the beginning or some comment lines above. Not sure if that will be an issue.

koppor:
Maybe more test cases are required to test encoding headers different from `UTF-8`. Like your `CP1252`.

tobiasdiez:
@koppor I changed the "defaultEncoding" in the test file not for JabRef. Since all the test bib files have UTF-8 encoding in the header, this was the easiest way to check if the specified encoding was correctly determined. I fear I don't understand your second comment.

simonharrer:
What bothers me is that you changed the defaultEncoding variable for all existing tests instead of creating a new test where the defaultEncoding would be wrong, and the real encoding of the file must be selected.

src/main/java/net/sf/jabref/importer/OpenDatabaseAction.java
```diff
@@ -410,74 +411,33 @@ public static ParserResult loadDatabase(File fileToOpen, String encoding) throws
...
+                    // Line not recognized so stop parsing
+                    return null;
                 }
-            }
+             }
         } catch (IOException ignored) {
             // Ignored
         }
-        return suppliedEncoding != null ? suppliedEncoding.trim() : null;
+        return null;
```
simonharrer: We try to eliminate null. Would it be possible to use Optional<String> as a return type instead?

tobiasdiez:
I updated the code accordingly.

simonharrer:
Good idea. This should probably be done. Can you open an issue for this, please?

simonharrer:
Looks good to me. 

One minor thing: maybe make the files and strings static so that they are not recreated for each test.

stefan-kolb:
:+1:

PR #306, opened 2015-11-09
CopacImporterTest

zellerdev:
Added a few trivial Tests.
Added a Test for Exception.
Completed given Tests with given resources.
6 JUnit Tests without failure or error.

PR #362, opened 2015-11-18
BiblioscapeImporterTest

obraliar:


src/test/java/net/sf/jabref/importer/fileformat/BiblioscapeImporterTest.java
```diff
@@ -0,0 +1,96 @@
...
+                    }
+                }
+            } catch (Throwable var6_8) {
+                if (throwable == null) {
+                    throwable = var6_8;
+                } else if (throwable != var6_8) {
+                    throwable.addSuppressed(var6_8);
+                }
+                throw throwable;
+            }
```
simonharrer: 1) Please use Arrays.asList instead of adding all the entries with their list.add line. 
2) please have a look at how try-with-resources work. In #310 this is implemented more concisely.

koppor:
In the future, create a separate branch for each improvement. Now, you cannot use your `master` branch any more, because all commits to `master` go into this pull request. See https://github.com/JabRef/jabref/pull/306#issuecomment-155374056 for a similar case.

koppor:
I think, @bartsch-dev posted an http://web.archive.org link to a test file for the Biblioscape importer to @bruehldev.

bruehldev:
I wrote the link in the Biblioscapeimportertest.java file in the first
lines and pushed it.
Am 18.11.2015 15:47 schrieb "Oliver Kopp" <notifications@github.com>:

koppor:
@bruehldev i don't see your comment in the first lines. Please check where you pushed them. Please try to address our comments (if possible).

koppor:
Follow up is #504

PR #370, opened 2015-11-19
Add compatibility check for UTF8

### LINKED ISSUE #367: UTF-8 instead of UTF8
As discussed in #155, we switched from `UTF8` to `UTF-8`. In `net.sf.jabref.importer.OpenDatabaseActionTest.defaultEncoding`, however, `UTF8` is used. When changing that to `UTF-8`, tests fail. I think, that string should also be `UTF-8` and the test not failing.

lenhard:
This PR attempts to fix #367

We have a bit of a problem. `UTF8` came from these lines
```  java
        try (InputStreamReader reader = openFile(fileToOpen, suppliedEncoding, fallbackEncoding)) {
            BibtexParser bp = new BibtexParser(reader);

            ParserResult pr = bp.parse();
            pr.setEncoding(reader.getEncoding());
```
It is set in [this method call](https://docs.oracle.com/javase/8/docs/api/java/io/InputStreamReader.html#getEncoding--). That's the Java API. The `InputStreamReader` says that the proper name of the encoding is `UTF8`. 

My solution is an ugly compatibility hack in which we just insert the `-` manually. Ugly, but works... What do you think?

### DIFF
--- a/src/main/java/net/sf/jabref/importer/OpenDatabaseAction.java
+++ b/src/main/java/net/sf/jabref/importer/OpenDatabaseAction.java
@@ -107,20 +107,20 @@ public class OpenDatabaseAction extends MnemonicAwareAction {
 
     class OpenItSwingHelper implements Runnable {
 
-        final BasePanel bp;
+        final BasePanel basePanel;
         final boolean raisePanel;
         final File file;
 
 
-        OpenItSwingHelper(BasePanel bp, File file, boolean raisePanel) {
-            this.bp = bp;
+        OpenItSwingHelper(BasePanel basePanel, File file, boolean raisePanel) {
+            this.basePanel = basePanel;
             this.raisePanel = raisePanel;
             this.file = file;
         }
 
         @Override
         public void run() {
-            frame.addTab(bp, file, raisePanel);
+            frame.addTab(basePanel, file, raisePanel);
 
         }
     }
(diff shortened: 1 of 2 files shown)


tobiasdiez:
What about replacing all string representation of encodings by [`Charset`](https://docs.oracle.com/javase/8/docs/api/java/nio/charset/Charset.html)? Then on export `getEncoding.name()` returns the correct description "UTF-8".

lenhard:
Sounds like a good idea! We should give it a try.

lenhard:
@tobiasdiez great advice! `Charset` already implements the correct encoding name translation. All we have to do is through it into `Charset` and read it our again and we can skip the hack:
``` java
pr.setEncoding(Charset.forName(reader.getEncoding()).name());
```

koppor:
Please merge master: We did a mayor overhaul of the OpenDatabaseAction: Deleted code etc.

tobiasdiez:
@lenhard Glad this was helpful. I think the code would even further improve if the ParserResult (and maybe some other classes) doesn't hold the encoding as a string but as a `Charset`, i.e. ̀public Charset getEncoding()` and similarly for `setEncoding`.

lenhard:
@koppor Done!
@tobiasdiez Sure. But I think this should be part of a new PR, since this PR just attempts to fix one particular issue.

PR #371, opened 2015-11-19
Moved AssertUtil to tests

oscargus:
Found a file which better resides in testutils.


koppor:
Sounds good :+1:

oscargus:
I must admit that I thought I found AssertUtil in the non-test source tree, but now I realize it wasn't the case... Anyway, I guess it is slightly better situated in the new location.

PR #373, opened 2015-11-19
Extracted logic out of FileHistory

oscargus:
Separated FileHistory into FileHistoryMenu which contains the menu functions and FileHistory which contains the logic for the file history.

### DIFF
--- a/src/main/java/net/sf/jabref/gui/FileHistoryMenu.java
+++ b/src/main/java/net/sf/jabref/gui/FileHistoryMenu.java
@@ -17,27 +17,23 @@ package net.sf.jabref.gui;
 
 import net.sf.jabref.JabRefExecutorService;
 import net.sf.jabref.JabRefPreferences;
-import net.sf.jabref.gui.JabRefFrame;
 import net.sf.jabref.logic.l10n.Localization;
+import net.sf.jabref.logic.util.io.FileHistory;
 
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
 import java.io.File;
-import java.util.Iterator;
-import java.util.LinkedList;
-
 import javax.swing.JMenu;
 import javax.swing.JMenuItem;
 import javax.swing.JOptionPane;
 
-public class FileHistory extends JMenu implements ActionListener {
+public class FileHistoryMenu extends JMenu implements ActionListener {
 
-    private final JabRefPreferences prefs;
-    private final LinkedList<String> history = new LinkedList<>();
+    private final FileHistory history;
     private final JabRefFrame frame;
 
 
-    public FileHistory(JabRefPreferences prefs, JabRefFrame frame) {
+    public FileHistoryMenu(JabRefPreferences prefs, JabRefFrame frame) {
         String name = Localization.menuTitle("Recent files");
         int i = name.indexOf('&');
         if (i >= 0) {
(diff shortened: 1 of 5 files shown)


src/main/java/net/sf/jabref/gui/FileHistoryMenu.java
```diff
@@ -88,16 +69,13 @@ public void newFile(String filename) {
...
-            addItem(i.next(), count);
-            count++;
+        for (int count = 0; count < history.size(); count++) {
+            addItem(history.getFileName(count), count + 1);
         }
     }
 
     private void addItem(String filename, int num) {
-        String number = num + "";
+        String number = Integer.toString(num);
```
simonharrer: String.valueOf is the typical function for this.

src/main/java/net/sf/jabref/logic/util/io/FileHistory.java
```diff
@@ -0,0 +1,89 @@
...
+
+    private final JabRefPreferences prefs;
+    private final LinkedList<String> history = new LinkedList<>();
+
+
+    public FileHistory(JabRefPreferences prefs) {
+        this.prefs = prefs;
+        String[] old = prefs.getStringArray(JabRefPreferences.RECENT_FILES);
+        if ((old != null) && (old.length > 0)) {
+            for (int i = 0; i < old.length; i++) {
```
simonharrer: You can use for(String s : old) { list.add(s); } instead. The check for old.length > 0 is also not necessary as the loop would ignore the case in which the array is empty.

src/main/java/net/sf/jabref/logic/util/io/FileHistory.java
```diff
@@ -0,0 +1,89 @@
...
+                history.remove(i);
+                return;
+            }
+            i++;
+        }
+    }
+
+    public void storeHistory() {
+        if (!history.isEmpty()) {
+            String[] names = new String[history.size()];
```
simonharrer: String[] names = history.toArray(new String[history.size()]);

src/main/java/net/sf/jabref/logic/util/io/FileHistory.java
```diff
@@ -0,0 +1,89 @@
...
+
+    public void newFile(String filename) {
+        int i = 0;
+        while (i < history.size()) {
+            if (history.get(i).equals(filename)) {
+                history.remove(i);
+                i--;
+            }
+            i++;
+        }
```
simonharrer: a simple history.remove(filename) should suffice for this case. Then the entry would have been removed if it was in the list. This can replace the whole while loop.

simonharrer:
Nice. A test for the FileHistory class would be even nicer. :)

src/main/java/net/sf/jabref/gui/FileHistoryMenu.java
oscargus: String.valueOf = 60 hits, Integer.toString = 32 hits. So I kept it, but can make a general search-and-replace if that is deemed to be the correct solution.

simonharrer:
Looks good to me.

oscargus:
@koppor Probably you should check the latest PR. I think you are commenting on an old commit. Now most methods are one or two lines...

koppor:
Sorry for that! Following minor things remain. I merge and hack the first one.

* `List` instead of `LinkedList` (line 27)
* `TODO for next PR` :innocent:

oscargus:
No problems! I appreciate the feedback from you and @simonharrer ! (and the rest as well, of course.)

PR #375, opened 2015-11-19
Integrated key word handling in BibtexEntry

oscargus:
Moved the key word handling to BibtexEntry.

Added new methods addKeyword and addKeywords to simply add one or more keywords. Will be useful when merging key words later on.


src/main/java/net/sf/jabref/model/entry/BibtexEntry.java
```diff
@@ -489,4 +488,70 @@ public String getPublicationDate() {
...
+        String oldValue = this.getField("keywords");
+        String newValue;
+        if (!keywords.isEmpty()) {
+            StringBuilder sb = new StringBuilder();
+            for (String keyword : keywords) {
+                sb.append(keyword);
+                sb.append(", ");
+            }
+            sb.delete(sb.length() - 2, sb.length());
+            newValue = sb.toString();
```
simonharrer: String.join(", ", keywords) should do the same as these lines.

src/main/java/net/sf/jabref/model/entry/BibtexEntry.java
```diff
@@ -484,4 +485,64 @@ public String getPublicationDate() {
...
+            this.putKeywords(keywords);
+        }
+    }
+
+    /**
+     * Add multiple keywords to entry
+     *
+     * @param keywords Keywords to add
+     */
+    public void addKeywords(ArrayList<String> keywords) {
```
simonharrer: Use List (the abstraction) instead

src/main/java/net/sf/jabref/model/entry/BibtexEntry.java
```diff
@@ -484,4 +485,64 @@ public String getPublicationDate() {
...
+        if (keywords != null) {
+            for (String keyword : keywords) {
+                this.addKeyword(keyword);
+            }
+        }
+    }
+
+    public ArrayList<String> getSeparatedKeywords() {
```
simonharrer: Use List instead.

src/main/java/net/sf/jabref/model/entry/EntryUtil.java
```diff
@@ -57,4 +61,25 @@ public static String capitalizeFirst(String toCapitalize) {
         }
         return al;
     }
+
+    /**
+     * @param keywords a String of keywords
+     * @return an ArrayList containing the keywords. An emtpy list if keywords are null or empty
+     */
+    public static ArrayList<String> getSeparatedKeywords(String keywords) {
```
simonharrer: I would move the method + constants in separate class because they should be kept together and are independent of all the other methods. Maybe create a class `Keywords`.

src/main/java/net/sf/jabref/model/entry/BibtexEntry.java
```diff
@@ -484,4 +485,64 @@ public String getPublicationDate() {
...
+        String oldValue = this.getField("keywords");
+        String newValue;
+        if (!keywords.isEmpty()) {
+            newValue = String.join(", ", keywords);
+        } else {
+            newValue = null;
+        }
+        if ((oldValue == null) && (newValue == null)) {
+            return;
+        }
```
simonharrer: tests are missing for the putKeywords method.

src/main/java/net/sf/jabref/model/entry/BibtexEntry.java
```diff
@@ -484,4 +485,64 @@ public String getPublicationDate() {
         }
         return year;
     }
+
+    public void putKeywords(ArrayList<String> keywords) {
+        // Set Keyword Field
```
simonharrer: use Objects.requireNonNull(keywords) at the beginning to ensure that a null value is detected early on.

simonharrer:
I think this is valuable. If you have time and improve it a little bit plus rebase on master, I think we can merge this in.

oscargus:
I'll do that. Maybe tonight, but cannot promise anything. There are indeed only minor things. I'll wait with the KeyWords class, but I think it makes sense in the long term to have a separate class to deal with this. There may also eventually be something similar for linking entries, but I will need to think more carefully about that...

oscargus:
Might be good to go now.

PR #377, opened 2015-11-20
Removed some random unused stuff

oscargus:
Removed unused variables, methods, and classes that I happened to stumble upon.


koppor:
Looks OK. The only thing is `TextAnalyzer`. I think, this could need a more thorough check: It seems that this class can be used to parse plain text references. Would be an offline alternative to the FreeCite parsing. [PRRV](https://sourceforge.net/projects/jabrefprrvp/) also seems to do parse text citations. I don't know, what's better, but I would not delete `TextAnalyzer` without having investigated alternative code. I understand that the current code is used nowhere. However, I would more like to put it into production than deleting it. For instance, the code could help at "Paste entries from plain text" when no internet connection is available.

oscargus:
OK! I'll figure out how to recover that file. Seemed like a waste to delete it, but since it wasn't used... If you have any feedback on koppor/jabRef #25 this is another case where we could throw away fully usable, non-trivial code, which is currently unused (except for a constructor, which in turn is unused).

koppor:
I think, following steps should do the trick:

1. `git rebase -i "HEAD^"`
2. in the opened text file, replace `pick` by `edit`, save and close
3.  `git gui`
4. select the file in the lower left box 
5. choose "Commit" -> "Unstage from Commit"
6. select the file in the top box
7. "Commit" -> "Revert changes", accept
9. close git gui (no commit necessary)
10. `git rebase --continue`

oscargus:
Great, that worked! (Except that I my GUI was in Swedish...) Thanks!

koppor:
Change to English: set environment variable `LANG` to `en`. I personally don't like the German translation of `git gui`.

PR #379, opened 2015-11-21
Add tests for the Bibtex parser

tobiasdiez:
I added a few tests for the Bibtex parser. I appreciate your feedback about the following points.
- [x] Please have a look at the `TODO` comments. These concern string concatenation and `@String` fields. For example in `parseRecognizesDateFieldWithConcatenation` the field `date = {1-4~} # nov` is parsed as `1-4~#nov#`. I wouldn't have expect the last sharp symbol.
- [x] Tests about JabRef's metadata (groups etc.) are still missing. Mostly, because I don't know the specifications :). Maybe in a new PR.
- [x] Related to the previous point, the `GUIGlobals.META_FLAG` flags should be moved to another place (they are not GUI, but parser/writer logic). What would be an appropriate new home for them?
- [x] There are 4 tests at the end of the file starting with `test`. These tests fail (without the `@Ignore` tag) but I'm not sure what they test and what would be the expected behavior in these cases.
- [x] There are quite a few ignored tests. These are related to open bugs in SF or are new bugs which I found while writing the tests (nothing big). I will open a new issue for the new bugs after this PR is merged.

For the record: While going through the code, I also changed the following things:
- Remove `BibtexParser.isRecognizedFormat` since the code has quite a few shortcomings (whitespace before @ or closing brackets on new lines were a problem) and the execution was quite slow for big files
- Fixed a bug where loading a file with corrupted bibtex entries showed an exception, cf. `parseIgnoresAndWarnsAboutCorruptedEntry()`
- Made `EntryType` an abstract class which implements the `Comparable` interface and overwrites `equals(Object other)`. Thus new expressions like `entry.getType() == BibtexEntryTypes.ARTICLE` return the expected result.

### DIFF
--- a/src/main/java/net/sf/jabref/importer/fileformat/BibtexImporter.java
+++ b/src/main/java/net/sf/jabref/importer/fileformat/BibtexImporter.java
@@ -31,11 +31,12 @@ import net.sf.jabref.model.entry.BibEntry;
 public class BibtexImporter extends ImportFormat {
 
     /**
-     * @return false as that does not cause any harm in the current implementation of JabRef
+     * @return true as we have no effective way to decide whether a file is in bibtex format or not. See
+     *         https://github.com/JabRef/jabref/pull/379#issuecomment-158685726 for more details.
      */
     @Override
     public boolean isRecognizedFormat(InputStream in) throws IOException {
-        return BibtexParser.isRecognizedFormat(new InputStreamReader(in));
+        return true;
     }
 
     /**
--- a/src/main/java/net/sf/jabref/importer/fileformat/BibtexParser.java
+++ b/src/main/java/net/sf/jabref/importer/fileformat/BibtexParser.java
@@ -15,14 +15,11 @@
 */
 package net.sf.jabref.importer.fileformat;
 
-import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.PushbackReader;
 import java.io.Reader;
 import java.io.StringReader;
 import java.util.*;
-import java.util.regex.Pattern;
-
 import net.sf.jabref.*;
 import net.sf.jabref.bibtex.EntryTypes;
 import net.sf.jabref.logic.CustomEntryTypesManager;
@@ -91,7 +88,7 @@ public class BibtexParser {
      * Parses BibtexEntries from the given string and returns the collection of all entries found.
      *
      * @param bibtexString
-     * @return Returns null if an error occurred, returns an empty collection if no entries where found.
+     * @return Returns returns an empty collection if no entries where found or if an error occurred.
      */
     public static Collection<BibEntry> fromString(String bibtexString) {
         StringReader reader = new StringReader(bibtexString);
@@ -100,7 +97,7 @@ public class BibtexParser {
         try {
             return parser.parse().getDatabase().getEntries();
         } catch (Exception e) {
-            return null;
+            return new ArrayList<>();
         }
     }
 
@@ -121,26 +118,8 @@ public class BibtexParser {
     }
 
     /**
-     * Check whether the source is in the correct format for this importer.
-     */
-    public static boolean isRecognizedFormat(Reader reader) throws IOException {
-        // Our strategy is to look for the "@<type>    {" line.
-        BufferedReader in = new BufferedReader(reader);
-
-        Pattern formatPattern = Pattern.compile("@[a-zA-Z]*\\s*\\{");
-
-        String bibtexString;
-
-        while ((bibtexString = in.readLine()) != null) {
-            if (formatPattern.matcher(bibtexString).find()) {
-                return true;
-            }
-        }
-        return false;
-    }
-
-    /**
-     * Will parse the BibTex-Data found when reading from reader.
+     * Will parse the BibTex-Data found when reading from reader. Ignores any encoding supplied in the file by
+     * "Encoding: myEncoding".
      * <p>
      * The reader will be consumed.
      * <p>
@@ -492,7 +471,6 @@ public class BibtexParser {
             skipWhitespace();
         }
         String key = parseKey();
-
         result.setField(BibEntry.KEY_FIELD, key);
         skipWhitespace();
 
@@ -559,7 +537,7 @@ public class BibtexParser {
         while (((character = peek()) != ',') && (character != '}') && (character != ')')) {
 
             if (eof) {
-                throw new RuntimeException("Error in line " + line + ": EOF in mid-string");
+                throw new IOException("Error in line " + line + ": EOF in mid-string");
             }
             if (character == '"') {
                 StringBuffer text = parseQuotedFieldExactly();
@@ -571,7 +549,7 @@ public class BibtexParser {
                  *
                  * while (!((peek() == '"') && (j != '\\'))) { j = read(); if
                  * (_eof || (j == -1) || (j == 65535)) { throw new
-                 * RuntimeException("Error in line "+line+ ": EOF in
+                 * IOException("Error in line "+line+ ": EOF in
                  * mid-string"); }
                  *
                  * value.append((char) j); }
@@ -848,7 +826,7 @@ public class BibtexParser {
 
             int character = read();
             if (isEOFCharacter(character)) {
-                throw new RuntimeException("Error in line " + line + ": EOF in mid-string");
+                throw new IOException("Error in line " + line + ": EOF in mid-string");
             } else if (character == '{') {
                 brackets++;
             } else if (character == '}') {
@@ -891,7 +869,7 @@ public class BibtexParser {
 
             int character = read();
             if (isEOFCharacter(character)) {
-                throw new RuntimeException("Error in line " + line + ": EOF in mid-string");
+                throw new IOException("Error in line " + line + ": EOF in mid-string");
             } else if (character == '{') {
                 brackets++;
             } else if (character == '}') {
@@ -915,7 +893,7 @@ public class BibtexParser {
         while (!((peek() == '"') && (brackets == 0))) {
             int j = read();
             if (isEOFCharacter(j)) {
-                throw new RuntimeException("Error in line " + line + ": EOF in mid-string");
+                throw new IOException("Error in line " + line + ": EOF in mid-string");
             } else if (j == '{') {
                 brackets++;
             } else if (j == '}') {
@@ -933,7 +911,7 @@ public class BibtexParser {
         int character = read();
 
         if (character != expected) {
-            throw new RuntimeException("Error in line " + line + ": Expected " + expected
+            throw new IOException("Error in line " + line + ": Expected " + expected
                     + " but received " + (char) character);
         }
     }
@@ -958,12 +936,12 @@ public class BibtexParser {
         int character = read();
 
         if ((character != firstOption) && (character != secondOption)) {
-            throw new RuntimeException("Error in line " + line + ": Expected " + firstOption + " or "
+            throw new IOException("Error in line " + line + ": Expected " + firstOption + " or "
                     + secondOption + " but received " + character);
         }
     }
 
-    private void checkEntryTypes(ParserResult parserResult) {
+    private void checkEntryTypes(ParserResult result) {
         for (BibEntry bibEntry : database.getEntries()) {
             if (bibEntry.getType() instanceof UnknownEntryType) {
                 // Look up the unknown type name in our map of parsed types:
@@ -972,7 +950,7 @@ public class BibtexParser {
                 if (type != null) {
                     bibEntry.setType(type);
                 } else {
-                    parserResult.addWarning(
+                    result.addWarning(
                             Localization.lang("Unknown entry type")
                                     + ": " + name + "; key: " + bibEntry.getCiteKey()
                     );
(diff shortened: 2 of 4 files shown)


koppor:
Regarding `#`: It might be that JabRef diverges from BibTeX's format. [BaseFrameHelp](http://jabref.sourceforge.net/help/BaseFrameHelp.php) states that `#` is used to reference strings. Is that BibTeX standard? [StringEditorHelp](http://jabref.sourceforge.net/help/StringEditorHelp.php) also states that "this syntax is specific for JabRef, and differs slightly from the bibtex notation that is produced when you save your database". Do you have time to investigate further?

Regarding your commits: The first two (a3a4b57 and 408a1a0) seem to be merge commits but also contain modifications by you. This is very hard to review. Proposal: make a *single* clean commit based on master.

1. (already done, just for the record) Ensure that `Jabref/master` is up to date. GitHub doc: https://help.github.com/articles/syncing-a-fork/
2. ensure that the latest `master` is merged into `parser`
3. `git checkout parser`
4. `git reset master`
5. commit - preferably using `git gui` to really investigate the differences.
6. commit and `git push -f`

Result: **one** commit in branch `parser` showing the difference to master.

oscargus:
The sharp symbol is used differently in Bibtex and Jabref. In Jabref it is used to mark a string which shouldn't be enclosed in {} (or ""). In Bibtex it is used as a separator between enclosed text and strings. So I think it very much makes sense that `date = {1-4~} # nov` ends up as `1-4~#nov#` (which is what is shown in the JabRef entry editor, right?).

oscargus:
To clarify further (if needed) `1-4~#nov#` in JabRef means that `1-4~` will be enclosed and `nov`is a string alias/key. In the Bibtex file it means `{1-4~}` or `"1-4~"` and `nov` concatenated with `#` inbetween.

tobiasdiez:
@koppor I remember to have analyzed the regex used by `BibtexParser.isRecognizedFormat` and that I recognized some shortcomings (as stated above). While trying to fix them, I realized that there is no simple check to recognize a valid bibtex file without actually parsing it. And since the `isRecognizedFormat` actually tried to match a regex on every single line (in the worst case), I thought that it is an easier way to just parse every bib file and then state the errors. (Moreover, the method was used on import but not while opening a database). But if you prefer, I can reinsert it.

Yes, in bibtex the sharp symbol is used to concatenate strings as @oscargus explained. See also for example http://www.bibtex.org/Format/ . Apparently the current state works, so I just wanted to go safe and double-check if this is really the expected behavior. Thanks @oscargus for the clarifying words.

Ps: Yeah, I will clean up the git history...

koppor:
Regarding the `#` issue: I would propose that we remove bibtex incompatible behavior. Is it possible that we exchange JabRef's character "`#`" to backtick "`"?

koppor:
For me, it's OK to remove it. Just a comment in the code to this discussion. Maybe directly to your explanation above.

oscargus:
I do not think it is incompatible, just confusing. :-) Using `#...#` to directly define strings in JabRef is quite well established and most of the time it is used "stand-alone", so only the string in the field (journal and month abbreviations). Somehow, to me, it only becomes a problem once things are concatenated and one tries to compare the bibtex field with the JabRef field and realize that the same character is used.

Backtick is anyway a bad choice since it is a combining accent (on most(?) keyboard layouts) and therefore require two key presses. (GitHub is really annoying me in that respect.)

koppor:
The (German) keyboard layout [neo](http://www.neo-layout.org/) offers the backtick with a single key (and a modifier key).  :stuck_out_tongue_winking_eye:

I don't quite get the usage of `#...#`. Is it also written as such in the bib file or is it just a rendering thing in JabRef? I can live with the latter even though I find it confusing if the `bib` differs from the UI.

oscargus:
Just rendering in JabRef. Try writing `#a#` in a field, it will show up as `a` (not `{a}`) in the bibtex source. Write `a#b#` and it will be `{a} # b`.

tobiasdiez:
Could I please get feedback about the last two open points (`GUIGlobals.META_FLAG` and the tests about newlines). Thanks!

src/main/java/net/sf/jabref/importer/fileformat/BibtexImporter.java
```diff
@@ -31,11 +31,11 @@
 public class BibtexImporter extends ImportFormat {
 
     /**
-     * @return false as that does not cause any harm in the current implementation of JabRef
+     * @return true as we have no effective way to decide whether a file is in bibtex format or not
      */
     @Override
     public boolean isRecognizedFormat(InputStream in) throws IOException {
-        return BibtexParser.isRecognizedFormat(new InputStreamReader(in));
+        return true;
```
simonharrer: I do not like that we remove the logic to detect a bibtex file. On the contrary, we should extend this as we also want to detect whether a file is a bibtex or a biblatex file. This should not be removed.

src/main/java/net/sf/jabref/model/entry/EntryType.java
```diff
@@ -52,10 +50,35 @@
...
+
+    @Override
+    public int hashCode() {
+        return this.getName().hashCode();
+    }
+
+    @Override
+    public int compareTo(EntryType o) {
+        return getName().compareTo(o.getName());
+    }
```
simonharrer: I prefer interfaces over class extension. Would a Comparator<EntryType> also work for the use case here?

src/test/java/net/sf/jabref/importer/fileformat/BibtexParserTest.java
```diff
@@ -32,219 +104,682 @@ public void testParseReader() throws IOException {
...
+        Assert.assertTrue(result.hasWarnings());
+
+        Collection<BibtexEntry> c = result.getDatabase().getEntries();
+        Assert.assertEquals(1, c.size());
+
+        BibtexEntry e = c.iterator().next();
+        Assert.assertEquals(BibtexEntryTypes.ARTICLE, e.getType());
+        Assert.assertEquals(null, e.getCiteKey());
+        Assert.assertEquals(2, e.getFieldNames().size());
+        Assert.assertEquals("Ed von Test", e.getField("author"));
```
simonharrer: Comparison should be done using the canonical bibtex format. Then, we can a) detect more errors, and b) the tests itself are getting easier to read.

simonharrer:
> Made EntryType an abstract class which implements the Comparable interface and overwrites equals(Object other). Thus new expressions like entry.getType() == BibtexEntryTypes.ARTICLE return the expected result.

The second sentence is wrong. == never calls equals or hashCode or comparable, it simply checks if both reference the same instance in the heap. Because of this, I would not make the EntryType an abstract class. 

> Remove BibtexParser.isRecognizedFormat since the code has quite a few shortcomings (whitespace before @ or closing brackets on new lines were a problem) and the execution was quite slow for big files

To reiterate an inline comment: we aim to detect whether a file is a bibtex or biblatex file. This code can help with this. We should not delete it, but improve it. 

> Related to the previous point, the GUIGlobals.META_FLAG flags should be moved to another place (they are not GUI, but parser/writer logic). What would be an appropriate new home for them?

Agreed. Maybe a new class called JabRefDataFormat? It is currently used in MetaData and BibtexParser. I think we should assemble all read/write logic in a single place instead of distributing it over so many classes, as the MetaData class itself does not need to know how it can persist its structure. But I am open for other ideas regarding placing and naming. 

>  There are 4 tests at the end of the file starting with test. These tests fail (without the @Ignore tag) but I'm not sure what they test and what would be the expected behavior in these cases.

I think we should investigate. 2022983 refers to an issue number, somehow. Maybe to an old sourceforge number?

koppor:
SourceForge changed global numbering of bugs to local numbering of bugs years ago. But they have redirects. Just use https://sourceforge.net/p/jabref/bugs/2022983/ as URL and it will redirect to https://sourceforge.net/p/jabref/bugs/679/.

src/main/java/net/sf/jabref/importer/fileformat/BibtexImporter.java
tobiasdiez: The check equaled more or less to a search of the `@` symbol. So it was purely semantic. Moreover, there is difference between a Bibtex and a Biblatex file on this semantic level. As I have explained above, in my opinion there is no easy way to tell if a file is a valid bibtex file or not. This is why I have deleted this method (which, by the way was only called upon importing a bibtex file, but not on opening one).

src/main/java/net/sf/jabref/model/entry/EntryType.java
tobiasdiez: I think `Comparator` is not applicable here. What is actually required is a proper equals method, the comparison is just an extra.

src/test/java/net/sf/jabref/importer/fileformat/BibtexParserTest.java
tobiasdiez: What is the canonical bibtex format? Do you mean, that the `BibtexEntry` should be converted to a bibtex string? I strongly oppose this change as then the parser tests could fail for two different reasons: the parser logic is wrong or the export to bibtex has a flaw.

simonharrer: Have a look at `CanonicalBibtexEntry`. This format will never change and can be used to compare bibtex entries with predefined texts.

tobiasdiez:
Ok, I was wrong about the `==`. What I actually meant was that the line `Assert.assertEquals(BibtexEntryTypes.ARTICLE, e.getType());` works as expected. 

I liked the idea of a new `JabRefDataFormat` class and moved the meta flag there.

@koppor thanks for finding the old bug report! It says that the problem was multiple spaces in file names (which would explain the naming of the tests). However, I think this bug is not fixed: if you assign a file with multiple spaces to a entry. After saving and reopening the database, the file is no longer found. Somehow the spaces are lost during parsing, which is exactly why the tests fail.

src/main/java/net/sf/jabref/importer/fileformat/FieldContentParser.java
```diff
@@ -37,7 +34,7 @@ public FieldContentParser() {
         multiLineFields.add("abstract");
         multiLineFields.add("review");
         // the file field should not be formatted, therefore we treat it as a multi line field
-        multiLineFields.addAll(Arrays.asList(Globals.prefs.getStringArray(JabRefPreferences.NON_WRAPPABLE_FIELDS)));
+        //multiLineFields.addAll(Arrays.asList(Globals.prefs.getStringArray(JabRefPreferences.NON_WRAPPABLE_FIELDS)));
```
simonharrer: why did you comment out this line?

src/main/java/net/sf/jabref/importer/fileformat/BibtexParser.java
```diff
@@ -283,7 +262,7 @@ public ParserResult parse() throws IOException {
                                     + be.getAuthorTitleYear(40) + " ("
                                     + Localization.lang("grouping may not work for this entry") + ")");
                         }
-                    } catch (IOException ex) {
+                    } catch (Exception ex) {
```
simonharrer: why did you change to Exception here? The code above does not state that Exception is required.

tobiasdiez: If a bibtex field has for example no closing bracket, then another exception is thrown (not sure which concrete one, something along the lines of "I got something, but expected something else"). So I changed `IOException` to a general exception to handle also these cases. Without this change the test `parseIgnoresAndWarnsAboutCorruptedEntry` fails.

src/main/java/net/sf/jabref/importer/fileformat/FieldContentParser.java
tobiasdiez: Not sure why. I'll revert it.

src/main/java/net/sf/jabref/importer/fileformat/BibtexParser.java
lenhard: I think it would be better to find out the exact Exception (should be easy with the test) and then write a multi-catch block: `} catch (IOException | ThatOtherException ex) {`

If you catch `Exception`, you also catch Exceptions we do not want to catch, such as `NullPointerException`, `ArrayIndexOutOfBoundsException`, etc.

src/main/java/net/sf/jabref/importer/fileformat/FieldContentParser.java
tobiasdiez: I remember (or better: the tests reminded me): This line was commented out because there were some null point exceptions due to uninitialized preferences. Fixed this in the new commit.

src/main/java/net/sf/jabref/importer/fileformat/BibtexParser.java
tobiasdiez: Yes you are right. I changed it to only catch `RuntimeExceptions`.

tobiasdiez:
Ok, should I rebase (and thus probably delete some of your code comments)?

simonharrer:
Can you please rebase this on master?

tobiasdiez:
Rebased. 

It should be noted that I reversed a changed made in #469. Namely, Bibtexentry types are just compared by name (so a bibtex type with the same name as a biblatex type is considered the same). Otherwise all the tests would break in biblatex mode as then getType() returns the biblatex type.

I will propose a better solution tomorrow based on on the [state design pattern](https://en.wikipedia.org/wiki/State_pattern).

tobiasdiez:
I removed all the code overwriting the `equals()` method for entry types. Thus now the PR just contains the added tests.From my side, this is ready to be merged in.

simonharrer:
The issue with the isRecognizedFormat still persists. Will discuss in developer call how to proceed there. We need to check what effects will this have to set it to true. Maybe the current method is good enough?

simonharrer:
Please rebase. Will merge this right away then. Dev call said true is OK for import format detection.

tobiasdiez:
Rebased, but the findByDOI fails (probably a problem with the CI server not the code...)

PR #380, opened 2015-11-21
Added month selection to EntryEditor

oscargus:
Added a button for the month field such that one can select a month and get the correct bibtex string for that month.

Ideally there should probably be localized month names in the list, but I didn't manage to find a way to do that which still keeps the entry model independent, avoids hard coding the month names again and make sure that syncLang can find the strings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ to [sourceforge feature requests](https://sourceforge.net/p/jabref/features/) by
 
 ### Changed
  - Better support for IEEEtranBSTCTL entries
+ - Quick selection of month in entry editor
  - Unknown entry types will be converted to 'Misc' (was 'Other' before).
  - EntryTypes are now clustered per group on the 'new entry' GUI screen.
  - Tab shows the minimal unique folder name substring if multiple database files share the same name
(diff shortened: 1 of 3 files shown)


koppor:
When in biblatex mode, could you add `month = {<monthnumber>}` instead of `bibtexFormat`. See [biblatex](https://www.ctan.org/pkg/biblatex) manual on page 20. Text for "month" field.

oscargus:
Good point! Should be sorted out now.

simonharrer:
At first, I thought this is a really nice feature. But then I thought about it a little bit and found the following issues
- what if someone uses JabRef without bibtex or biblatex, e.g., by using the open office bridge? 
- what if someone used the month field wrongly, e.g., just putting the month in their local language?

But the really problematic thing is that a dropdown field is used. And when someone has put in a value that is not recognized by the dropdown field (old value is invalid, e.g., "März" instead of March), then this value is lost upon save and not shown in the UI. Because the dropdown while either show nothing or the first value of the dropdown. And with this information, the user is not able to fix this problem as he cannot easily see the previous value and he will not be notified that there even is a previous value. 

Because of this, I vote for an improvement or exclusion of this feature.

koppor:
The dropdown is displayed additionally to the text field. One can input plain text instead of using the drop down. I tried to input "test123" and it saved perfectly. I don't see why this should cause troubles.

grabbed_20151122-153012

simonharrer:
Ah ok. I just have looked at the code and misunderstood things - and I did not try the feature out... sorry.

PR #381, opened 2015-11-21
Change default preferences: Remote server is started

koppor:
I think, the new default makes sense as JabRef sometimes opens twice which is a bit annoying for me.

Before resetting the preference every now and then, I had the remote server activated for a few months and it worked fine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,7 @@ to [sourceforge feature requests](https://sourceforge.net/p/jabref/features/) by
  - Update supported LookAndFeels
  - Show replaced journal abbreviations on console
  - The three options to manage file references are moved to their own separated group in the Tools menu. 
+ - Default preferences: Remote server (port 6050) always started on first JabRef instance. This prevents JabRef loaded twice when opening a bib file.
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


oscargus:
Don't you think people will freak out when the firewall asks if the program should accept connections? (Or whatever happens?) Anyway, could be worth a try.

simonharrer:
Could not reproduce that the firewall prompts for accepting connections when used locally. I did this by removing all firewall exclusions and then started jabref twice with remote connections enabled on Windows 10. Can someone else please confirm this on their machine?

oscargus:
Sorry, I didn't test it. I just assumed that there might be some issues with this and that the "general public" gets confused and scared.

simonharrer:
:+1:

simonharrer:
Hm, this effectively disables running multiple JabRef versions simultaneously.

koppor:
That'S the intention: I really dislike that JabRef opens multiple Windows. We have tabs for multiple databases. I don't get the use case for multiple windows.

I know that we should have both somehow to mirror Firefox behavior. However, I don't assume that so many database files are opened in parallel. I know that Internet Explorer long refused to offer tabs, because the multiple window concept was more appealing.

PR #387, opened 2015-11-24
Use install4j to build the setup files. This enables bundling of JREs…

simonharrer:
… and Win7+ features like pin to taskbar


koppor:
LGTM :+1:

Regarding the resetting of the preferences. "reset" is the same as "delete"? Offering resetting is nice, but not mandatory for the PR. We can do that later. For now, I'd prefer https://github.com/koppor/jabref/issues/47.

Issue #116, opened 2015-08-19
Define optimal serialization format and variations (dropping or reviving the 2.9.2 format)


koppor:
In JabRef 2.10, we introduced following new feature:

```
    Changed serialization of BibTeX entries:
        First, the required, then the optional and then all other fields are written.
        Thereby, fields are now ordered by name. Except the title, which is written first.
        The second word in of the BibTeX type is capitalized. E.g., Inproceedings got InProceedings
        Configurable: Start field contents in same column. Enabled by default.
        Configurable: Use camel case for field names (e.g., "HowPublished" instead of "howpublished"). Enabled by default.
        If no field name is given, then "UNKNOWN" is used. For instance, " = {X}" gets " UNKNOWN = {X}".
```

When a user collaborates with a person using an older version of JabRef, they cannot use version control properly as the serialization always changes. In version 2.11 beta 2, we offer a quick hack to go back to the old 2.9.2 behavior, which is somehow incomplete. See https://github.com/JabRef/jabref/pull/10#issuecomment-43582135

We can just focus on other issues and let time solve the issue or invest time to implement the old serialization again.

Related issue: #115

Request at sourceforge: https://sourceforge.net/p/jabref/feature-requests/864/

==================================================================
To track the progress of implementation, the consensus described below is added here.

Rationale: People are very emotional about formatting -> Modify as little as possible in the bib file
- [x] Unmodified entries will not be formatted and written back exactly as read in (including all formatting issues, etc.
- [x] Modifying any part of an entry results in a reformatting of the complete entry
- [x] No more global configuration  forsorting of entries on load / save. (sorting information is now stored in the file)
- [x] New entries are added to the bottom of the file
- [x] Fixed  and non-configurable format for new or modified entries. Dropping all field saving options (Preferences -> File -> Field saving options)
- [x] Fields are sorted as stated in the Bibtex / Biblatex manuals.
- [x] The ` =` is appended directly to the field key. The `value` part is indented to one space past the longest field key name + ` = ` (so that all values are aligned).
- [x] Bibtex entry keys are written in camel case, starting upper case
- [x] Field keys are formatted in lower case only
- [x] No longer perform any changes of field content (no more space/tab/newline elimination, etc.)
- [x] Provide formatter (Example: Remove tabs / newlines / duplicate spaces) that can be enabled by a user for specific fields onLoad / onSave. These are not enabled by default.
- [x] If in doubt, do as biber does

stefan-kolb:
[...]

Issue #123, opened 2015-08-24
Replace BibTexParser & Writer


stefan-kolb:
The class `BibtexParser`should be replaced or rewritten if possible. Currently, it is quite a mess. Also, it only does the parsing of the files and not the writing. 

We need to investigate if we can replace it by some external library like [jBibTex](https://github.com/jbibtex/jbibtex) or need to rewrite it properly. The same probably holds for the writing process which should be aligned with the reading process.

Any information appreciated.

koppor:
Maybe we should check all Java-based [BibTeX-based managers](https://github.com/JabRef/jabref/wiki/Alternatives#bibtex-based) how they do it?

koppor:
jBibTex should be checked if arbitrary keys are supported. The commit https://github.com/S-Shimotori/jbibtex/commit/da1bb9eae13dc399fe775245fedb38aae67ffb33 indicates that the keys are are hardcoded. Another option is to drop support of customized entries completely and only support hard-coded ones.

koppor:
The [ANTLRv3 Grammar](http://stackoverflow.com/a/7597864/873282) could be a good start. If we use it, we have to ask @bkiers if he is willing to relicense it GPLv2+ or any other GPL-compatible license.

bluebirch:
Erathostenes (https://bitbucket.org/mkmatlock/eratosthenes/wiki/Home) use
jBibTeX and arbitrary keys are supported, as far as I am aware since I use
it.

mån 24 aug 2015 21:27 Oliver Kopp <notifications@github.com> skrev:

koppor:
We should generate test cases for it. I started to collect publicly available bibtex files at [the wiki](https://github.com/JabRef/jabref/wiki/Publicly-available-BibTeX-files). Should we include them using git submodules? I think, that is too much overhead.

lenhard:
[jBibTex uses some weird proprietary license](https://github.com/jbibtex/jbibtex/blob/master/LICENSE.txt). I don't think we can use it on that basis.

koppor:
Why? The README.md states: "JBibTeX is licensed under the BSD 3-Clause License." And the text seems to be the BSD 3 licensed. Or do you think, he mixed in a strange statement?

lenhard:
My Bad! I just did not recognize the text of BSD. So jBibTex is fine.

koppor:
Comments in the file before all entries and after all entries should be kept.

Example (from http://shelah.logic.at/eindex.html)

```
[...]

PR #391, opened 2015-11-25
Write unmodified entries to bib file in the same format as they were read

lenhard:
As discussed in #116, we want to write entries that are not modified during a session back in the exact same format as they were read. This PR is WIP in this direction and not complete yet. I will continue working on it.

To be able to write back an entry in the same fashion as it is read, we need to store it upon reading. I added a field to `BibtexEntry` for storing this and tried to modify `BibtexParser` to store the file content, but to no avail. The current code of that class is pure hell (uses a global `PushBackReader` to read the file in a very confusing way) and this PR is hopefully a step towards its replacement.

Instead of modifying the logic in `BibtexParser`, I extend it with additional methods that perform the new functionality on top. Current status:
 1. On initialization the complete file is read into a `List<String>` of its lines before handing it over to the old parser.
 2. After an entry has been parsed, the original represenation of the entry is looked up in the `List<String>` and stored in the entry. Blank lines following the entry belong to the entry.

~~That means we read the file twice, which is terrible for large files, but as this is WIP...~~

Next step will be to detect when an entry has been changed and the modification of the writing logic.

The handling of newlines currently is not consistent.


simonharrer:
Nice. Hm, what to do when two entries share a line? E.g. `@article{key1, title={a}}@article{key2, title={b}}`?

lenhard:
In that case, the current implementation fails.

But that gives me an idea: Isn't `@` a valid delimiter that is only used for setting bibtex entry types, but forbidden in other parts of the file? If so, I could use it for reading the file, instead of the newline character. This is very easy to do with a `Scanner` and would make things way easier.

oscargus:
There is the issue of comments, which potentially can contain `@`. But as far as I know, everything outside of `@xxx{..}` is a comment.

tobiasdiez:
The `@` symbol can also occur in an entry, for example in an email-address. The current implementation has also the problem that it fails if the entry has no bibtex key. I fear there is no easy way to find an entry in a bibtex file without actually parsing it completely.

koppor:
Isn't it possible to add a reference to the source line and column during parsing? This should be offered by ANTLR somehow, isn't it?

I assume that [JBibTeX](https://github.com/jbibtex/jbibtex). Doesn't support that. Refs #123.

lenhard:
@oscargus: that would not be a problem. It's more of a problem if an `@` is within an entry

@koppor: This is work in progress. I'll start commenting when I have something closer to finalization. Do we want to parse the file twice? If so, then jbibtex might be an option. Can you tell me if jbibtex writes back an entry in exactly the same format (up to every line feed character) as it was read?

@tobiasdiez Thanks, I am aware of that. This is work in progress. I'd first like to arrive at a solution that does the round-trip for a well-structured database. Then, I'll modify it to match incomplete entries.

lenhard:
Ok, now this works nicely with my large local database that doesn't include analomies.

Now comes the hard part: Properly reading and writing files with strange content. This is also the reason why some tests are failing at the moment. I might need to discuss some of these aspects and will ping you in this thread if I am unsure.

lenhard:
Dammit, I wanted to get some work done today and now I am only coding JabRef...

Anyways, this seems to work (and passes the tests)! I ditched my previous parsing logic and rewrote it in integration with the `BibtexParser`, which wasn't as hard as I claimed above (my bad). 

I am now able to store each entry exactly as it is in the file. Each entry serialization contains all text from its end up to the end of *previous* entry. The main problem with this is that with this logic, the first entry also stores the file header and then the header is serialized twice when rewriting the file. I was able to get around that with some hacking, but there might still be edge cases I am not yet aware of.

To sum up, this needs some more manual testing, but should be pretty close. Feedback is welcome :)

src/main/java/net/sf/jabref/model/entry/BibtexEntry.java
```diff
@@ -60,6 +61,14 @@
     private boolean searchHit;
     private boolean groupHit;
 
+    private String serialization;
+
+    /*
+    * marks if the complete serialization, which was read from file, should be used.
+    * Is set to false, if parts of the entry change
+     */
+    private boolean useCustomSerialization;
```
matthiasgeiger: I would rename this (and the corresponding methods, of course) as it is always used as "if not changed then..." to "useParsedSerialization" or even to "hasChanged".

src/main/java/net/sf/jabref/bibtex/BibtexEntryWriter.java
```diff
@@ -96,6 +96,12 @@ public BibtexEntryWriter(LatexFieldFormatter fieldFormatter, boolean write) {
     }
 
     public void write(BibtexEntry entry, Writer out) throws IOException {
+        // if the entry has not been modified, write it as it was
+        if(entry.shouldUseCustomSerialization()){
+            out.write(entry.getSerialization());
+            return;
+        }
```
matthiasgeiger: should be `if(!...)` ?

matthiasgeiger:
Tested it manually with some entries and it works smoothly! 

Good job! :+1:

src/main/java/net/sf/jabref/model/entry/BibtexEntry.java
lenhard: Good point, I'll do some renaming. On second thought, I will also replace the `Deque` with a `Stack`, because this is the way in which it is used.

lenhard:
Should we really introduce this in v3.0 already? I'm not entirely sure it works on all occaison. For instance, I haven't tested it with strings in the bib file so far.

simonharrer:
I think this is something for v3.1.

koppor:
:+1: For postponing. I think, we don't need a separate `release` branch, do we? We are just careful with mering the next days :8ball:

lenhard:
There are problems with newlines in this PR. If I open JabRef, save the current file without modification, and close JabRef multiple times over, then the save removes newlines between entries step by step. Currently, I am not at all sure why that is.

lenhard:
Working again. 

In the current version, all newlines / whitespaces at the end of the file, those that follow the last entry, are deleted. I don't really see a problem with that, or am I wrong?

simonharrer:
For version control, you always want to have an empty line at the bottom. http://stackoverflow.com/questions/5813311/no-newline-at-end-of-file

lenhard:
Why? The only argument in the stackoverflow discussion you link is "because C says, there should be one". That's not really convincing for a BibTeX file.

Anyways, this shouldn't be too hard to implement, so I'll just do it.

simonharrer:
I should have been clearer. Not the selected answer but one of the others is more relevant:

    If you add a new line at the end of the existing file which is not having a newline at the end already, the diff will show old last line also as modified, when conceptually its not.

    At least one good reason to add a newline at the end.

matthiasgeiger:
https://github.com/JabRef/jabref/commit/9b46dc28043dd41a62edd65ae33a37b9c5e76be1#diff-63c594fa122cbaa34c72e860d42a75efR2372 

Praxisbeispiel ;-)

lenhard:
Ok, this seems to work (I also tested with BibTeX strings and JabRef comments) and is ready to merge from my side. If somebody has further remarks and comments (or, god forbid, bugs), this is the time :)

Since this change is big and our test set is not so big, I would object to release this into the wild immediately. Since a bugfix release is imminent for 3.0, I guess we should still wait with merging?

koppor:
Just to be sure: Is it possible to keep lines outside entries? For instance at https://github.com/JabRef/jabref/blob/v3.0/src/test/resources/testbib/cleanup.bib, the lines above `@article{test,` should IMHO be kept when writing the file again.

lenhard:
@koppor: I don't think JabRef ever did that. I might be wrong of course. From what I know, the parser only parses stuff that starts with `@` and is a `preamble`, `string`, `comment`, or any entry type. I also can't find code in `FileActions` that saves arbitrary text.

If we want to store arbitrary text as well, then this is new functionality and should be done in a separate PR.

lenhard:
I just checked on master and can confirm that the text is removed.

tobiasdiez:
Even unrecognized comments by the user are removed, see https://github.com/JabRef/jabref/blob/master/src/main/java/net/sf/jabref/importer/fileformat/BibtexParser.java#L250

lenhard:
By the way, the `BibtexParser` still includes code for reading the version number (`This file was created with JabRef X.Y`). Since we stopped writing that, should I delete the parsing code?

koppor:
AFAIK, we use that information for file upgrades. Maybe we should move the file update functionality in clean up or something other more accessible. I think, the "upgrade warning" dialog can be disabled permanently, which is not helpful when opening an old file half a year later.

lenhard:
As far as @simonharrer knows :smile: the version number is not used anywhere. Can you provide evidence for such file upgrades?

matthiasgeiger:
e.g. https://github.com/JabRef/jabref/blob/master/src/main/java/net/sf/jabref/migrations/FileLinksUpgradeWarning.java#L65

lenhard:
@matthiasgeiger But that does not really depend on the version information. The version is checked, but also every file written by JabRef 3.0 is marked as problematic. The file links upgrade really depends on whether the file contains fields named `pdf` or `ps`. It wouldn't make a difference `isActionNecessary` always returned true, regardless of the version.

Anyways, if you really really want the version parsing, we can keep it.

matthiasgeiger:
okay... I have to confess that I only performed a "find usages" search without truly checking whether the usages are really needed.

So feel free to remove those lines ;-)

lenhard:
I removed the old meta flag and version parsing. Version headers are now simply removed on rewrite. File upgrade functionality ultimately depends on preferences (and on whether JabRef finds errors in the file to begin with).

lenhard:
So, from my point of view this PR is complete and ready to merge.

@koppor and I did some more testing with more advanced things like meta-data and it seems fine. I'm not guaranteeing the absence of bugs, but there should be no more obvious mistakes. 

Somebody else can take a final look and merge this with master.

simonharrer:
Looks good to me. It worked with my large .bib file without any issues.

koppor:
@simonharrer Could you check what happens if you change the sort ordering of the file? Change the ordering. If everything is OK, change the ordering and change an entry. :rabbit2: I it doesn't work, maybe we drop the ordering possibilities of entries within JabRef now or leave that as open issue.

simonharrer:
What do you think is the problem here? When I change the sort ordering, the order of the entries is changed in the file. But nothing else is. And if I also change an entry, this entry is changed as well. I do not get what you imply.

PR #392, opened 2015-11-25
Removed unused TXMLReader

oscargus:
As discussed in koppor/jabref#25.


koppor:
OK, let's get this through. If we need the code, we'll revert the commit.

PR #396, opened 2015-11-26
Change representation of encodings to Charset

tobiasdiez:
Instead of representing an encoding as a string we now use a Charset.
This idea was already discussed in PR #370.


simonharrer:
I do like this. Can there be any issues if a previous file has an encoding format that is not understood and is loaded with a newer version of jabref?

tobiasdiez:
As far as I can see there is no problem. The encoding from the file is given to `Charset.forName()`, which is the same behavior as previously. Maybe some encodings are now written to the bib-file in a slightly different way, but this shouldn't be a big problem.

simonharrer:
Ok. Please rebase, I will merge it in.

tobiasdiez:
Rebased.

PR #403, opened 2015-11-27
Force to have at least english localization

simonharrer:
What has been done:

- [x] Cleanup python script
- [x] Localization keys no longer contain trailing/leading spaces and newlines. Tests are in place to ensure that this will never happen again. 
- [x] Localization keys do not use %e and %c for = or : anymore
- [x] Unit Test find ALL usages of of Localization.lang and Localization.menuTitle. Hence, the workaround to disable the formatter is no longer required.
- [x] If the unit test fails, on the console the appropriate measures are printed how to add the keys. 
- [x] Fixed a lot of other issues as well, e.g., when the exception message was part of the localization key. 
- [x] That all non-english property files have the same keys as the english property file is checked by the unit test. However, there is also gradle task/python script for this.
- [x] Unit test that checks that every english translation is equal to the key. 
- POSTPONE `<br>` is similar to `\n`, but for this we do not have checks yet as well.
- POSTPONE Rename `Localization.lang` to `Localization.message`
- [x] The KeyBindings use `Localization.lang` directly.
- [x] Refactored code of KeyBinding
- [ ] update documentation
- [x] if any resource bundle is not available or if the given locale is not known or if the resource bundles have not bin initialized, always use "en" language. 

This is good for a review. The remaining points I would like to ignore as this is really annoying work, and this should be good enough with this PR already. 

New process:
1. Add new Localization.lang("KEY") to Java file.
2. Tests fail. In the test output a snippet is generated which must be added to the english translation file. There is also a snippet generated for the non-english files, but this is irrelevant.
3. Add snippet to english translation file
4. With `gradlew -b localization.gradle compareAndUpdateTranslationsWithEnglishTranslation` the "KEY" is added to the other translation files as well. 
5. Tests are green again.


stefan-kolb:
If you refactor this, we may also rename the methods in `Localization.java`. `#lang` may be named `msg` or `message` for example.

simonharrer:
@JabRef/developers please review, as I would like to merge it. Important changes are given in the first post of this PR.

matthiasgeiger:
:+1:  works fine and looks good to me

src/test/java/net/sf/jabref/logic/l10n/LocalizationParser.java
```diff
@@ -0,0 +1,206 @@
...
+        Properties properties = new Properties();
+        try (InputStream is = LocalizationTest.class.getResourceAsStream(path)) {
+            properties.load(is);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+        return properties;
+    }
+
+    @Test
```
tobiasdiez: Move these tests to the LocalizationTest class?

src/test/java/net/sf/jabref/logic/l10n/LocalizationTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+import java.util.StringJoiner;
+import java.util.stream.Collectors;
+
+import static org.junit.Assert.assertEquals;
+
+public class LocalizationTest {
```
tobiasdiez: Some of the tests should have more expressive names which describe what they test (probably also without the `test` prefix). 
I also find that some assertion are hard to read/understand and in my opinion they would profit if they are written over multiple lines (so that occurring objects get a name :) )

src/test/java/net/sf/jabref/logic/l10n/LocalizationParser.java
simonharrer: Hm, yes, should be better to move them out.

src/test/java/net/sf/jabref/logic/l10n/LocalizationTest.java
simonharrer: Agreed.

PR #407, opened 2015-11-28
Coverityfixes4

oscargus:
Fixed a number of things:

* Unnecessary imports
* Some resource leaks

I reformatted all files in gui/autocompleter as they had tabs in them and the diff here looked like a mess. The actual changes are that I inserted some missing `@override` and removed some imports, so nothing that affects any functionality.


oscargus:
Rebased and fixed spelling error.

Issue #388, opened 2015-11-25
proxy authentication support


tolot27:
There is still no support for proxy authentication. Unfortunately, username and password can not be specified by command line parameters or preferences (network tab) like it is possible for the proxy host and port (-Dhttp.proxyHost=).

The reason is that the Authenticator class has to be used.

A possible implementation is described here: http://memorynotfound.com/configure-http-proxy-settings-java.

koppor:
Would it be possible that you fork JabRef, create a patch and create a pull request?

![I can haz pull request](https://raw.githubusercontent.com/dscho/images/master/i-can-haz-pull-request.png)

Image source: https://github.com/dscho/images

tolot27:
I already tried this but gradle is a nightmare behind a proxy.

After getting it to work fetching https://services.gradle.org/distributions/gradle-2.9-all.zip by providing gradle.properties with proxy settings, it still fails to fetch the jars.
In the debug log I can see that it cannot fetch shadow-1.2.2.jar from http://maven.ej-technologies.com/repository/com/github/jengelman/gradle/plugins/shadow/1.2.2/shadow-1.2.2.jar.
It is quite obvious because there is no github directory in http://maven.ej-technologies.com/repository/com.

How to proceed?

tolot27:
Is github a virtual folder or plugin in maven which just does not work for the moment at http://maven.ej-technologies.com?

tolot27:
If I switch to another maven repository (i. e. http://repo.spring.io/libs-release-remote/), `shadow-1.2.2.jar` gets downloaded but then other plugins fail to download. There seem to be no cache because switching back to the ej-technologies repro again fails on shadow.jar.
Thats quite frustrating and a waste of time.

matthiasgeiger:
Sorry... but I cannot reproduce this... after clearing all my gradle caches to force gradle to download the libs again the following is executed:
```
$ gradlew build
Download http://maven.ej-technologies.com/repository/com/install4j/gradle-plugin/6.0/gradle-plugin-6.0.pom
Download https://plugins.gradle.org/m2/com/github/jengelman/gradle/plugins/shadow/1.2.2/shadow-1.2.2.pom
Download https://plugins.gradle.org/m2/org/kt3k/gradle/plugin/coveralls-gradle-plugin/2.4.0x/coveralls-gradle-plugin-2.4.0x.pom
[...]
[...]

PR #418, opened 2015-11-30
proxy authentication support (feature #388)

tolot27:
This pull request enables support for proxy authentication via command line arguments (`-D..`) and JabRef preferences (Network tab).


tolot27:
BTW: The tests in `net.sf.jabref.logic.fetcher` and the `URLDownloadTest` still fails because the Authenticator is not initialized in unit tests.

tolot27:
Why is the codecov/project check marked as failed but shows passed on details?

src/main/java/net/sf/jabref/JabRef.java
```diff
@@ -96,6 +99,26 @@ public void start(String[] args) {
...
+                    String password = System.getProperty(prot + ".proxyPassword", "");
+                    if (getRequestingHost().equalsIgnoreCase(host)) {
+                        if (Integer.parseInt(port) == getRequestingPort()) {
+                            return new PasswordAuthentication(user, password.toCharArray());
+                        }
+                    }
+                }
+                return null;
+            }
+        });
```
simonharrer: Is the `Authenticator` always needed? Can we add it only in the case we really know for sure that username/password credentials are required?

src/main/java/net/sf/jabref/JabRefPreferences.java
```diff
@@ -455,6 +458,9 @@ private JabRefPreferences() {
         defaults.put(USE_PROXY, Boolean.FALSE);
         defaults.put(PROXY_HOSTNAME, "my proxy host");
         defaults.put(PROXY_PORT, "my proxy port");
+        defaults.put(USE_PROXY_AUTHENTICATION, Boolean.FALSE);
+        //defaults.put(PROXY_USERNAME, "my proxy username");
+        //defaults.put(PROXY_PASSWORD, "my proxy password");
```
simonharrer: Why not use empty string as default? Otherwise, we would have to copy with null values, I think.

src/main/java/net/sf/jabref/gui/preftabs/NetworkTab.java
```diff
@@ -48,31 +53,55 @@ public NetworkTab(JabRefPreferences prefs) {
...
-        builder.add(defProxyHostname).xy(4, 5);
+        builder.add(defProxyHostname).xyw(4, 5, 3);
         builder.add(Localization.lang("Port") + ':').xy(2, 7);
-        builder.add(defProxyPort).xy(4, 7);
+        builder.add(defProxyPort).xyw(4, 7, 3);
+        builder.add(useProxyAuthentication).xyw(4, 9, 3);
+        builder.add(Localization.lang("Username:")).xy(4, 11);
+        builder.add(defProxyUsername).xy(6, 11);
+        builder.add(Localization.lang("Password:")).xy(4, 13);
+        builder.add(defProxyPassword).xy(6, 13);
```
simonharrer: Could you perhaps add a message to the user stating that passwords are stored in plaintext upon save?

simonharrer:
Is marked as fail as you added code which is not covered by a JUnit test, reducing code coverage. But we do not have a GUI strategy yet. And you added mostly GUI code, hence, no problem there.

src/main/java/net/sf/jabref/JabRefPreferences.java
tolot27: Because I followed the defaults for hostname and port I didn't get the idea behind populating it with "my proxy ...".

src/main/java/net/sf/jabref/gui/preftabs/NetworkTab.java
tolot27: Is a tooltip is enough? Maybe, if permantly shown?

tolot27:
I forgot to create a special branch for this PR... Hmmm.

src/main/java/net/sf/jabref/JabRefPreferences.java
tolot27: We don't have to cope with null values because NetworkTab.validateSettings ensures that the properties are neither null nor empty, as long as `useProxyAuthentication` is enabled.

src/main/java/net/sf/jabref/gui/preftabs/NetworkTab.java
matthiasgeiger: I'd prefer a JLabel underneath the password text field.

matthiasgeiger:
This is not a big problem - as long as you do not start working on another issue in your master branch :wink:

tolot27:
Hmm, if I change the proxy settings, a restart is required.

src/main/java/net/sf/jabref/JabRefPreferences.java
tolot27: Can we remove the defaults for `PROXY_HOSTNAME` and `PROXY_PORT`?

simonharrer: Yes, makes sense.

simonharrer:
We have other settings which upon change indicate to the user that JabRef has to be restarted for the settings to be applied. See the remote preferences under AdvancedTab. Maybe reuse the code from there.

tolot27:
Hmm, looking for config changes in six values looks quite ugly with `if` conditions. I can create a HashMap or StringBuilder out of these values and later on just compare the hash values. Or do you have any other suggestion?

simonharrer:
You just need to compute the final proxy string `username:password@hostname:port` which you can easily compare and then the two remaining boolean variables.

tolot27:
That's what I have implemented with a StringBuilder. Comparing the hashCode is much faster than comparing the string. ;-)

tolot27:
Okay, please check my PR.

src/main/java/net/sf/jabref/gui/preftabs/NetworkTab.java
```diff
@@ -48,31 +58,60 @@ public NetworkTab(JabRefPreferences prefs) {
...
 
+        useProxyAuthentication = new JCheckBox(Localization.lang("Proxy requires authentication"));
+        useProxyAuthentication.setEnabled(false);
+
+        defProxyUsername = new JTextField();
+        defProxyUsername.setEnabled(false);
+        defProxyPassword = new JPasswordField();
+        defProxyPassword.setEnabled(false);
+        lblPasswordWarning = new JLabel(Localization.lang("Attention: Password is stored in plain text!"));
+        lblPasswordWarning.setEnabled(false);
```
simonharrer: You can delete this line, as JLabels must not be disabled.

src/main/java/net/sf/jabref/gui/preftabs/NetworkTab.java
```diff
@@ -87,23 +126,57 @@ public void setValues() {
...
+        prefs.put(JabRefPreferences.PROXY_USERNAME, defProxyUsername.getText().trim());
+        prefs.put(JabRefPreferences.PROXY_PASSWORD, new String(defProxyPassword.getPassword()));
+        changedSettings = ((oldUseProxy != useProxy.isSelected()) || (oldUseProxyAuth != useProxyAuthentication
+                .isSelected()));
+        if (!changedSettings) {
+            StringBuilder sb = new StringBuilder();
+            sb.append(prefs.get(JabRefPreferences.PROXY_USERNAME)).append(':')
+                    .append(prefs.get(JabRefPreferences.PROXY_PASSWORD));
+            sb.append('@').append(prefs.get(JabRefPreferences.PROXY_HOSTNAME)).append(':')
+                    .append(prefs.get(JabRefPreferences.PROXY_PORT));
```
simonharrer: This code is in there two times. Can you extract that into a method?

src/main/java/net/sf/jabref/JabRef.java
```diff
@@ -97,6 +100,29 @@ public void start(String[] args) {
...
+                        if (getRequestingHost().equalsIgnoreCase(host)) {
+                            if (Integer.parseInt(port) == getRequestingPort()) {
+                                return new PasswordAuthentication(user, password.toCharArray());
+                            }
+                        }
+                    }
+                    return null;
+                }
+            });
+        }
```
simonharrer: It would be perfect if you could extract these lines related to the proxy settings into a separate class (named `ProxyConfigurator` or something like that), as the JabRef class is already too big to handle well.

src/main/java/net/sf/jabref/gui/preftabs/NetworkTab.java
tolot27: Indeed, but it is in general a good practice to disable irrelevant labels to not disturb/irritate the user. I can change it to be kept enabled as long as the password is not empty. That is IMHO the best compromise. Do you agree?

simonharrer: Ah, ok, you enable it with the other fields. I overlooked this fact. Leave it as it is.

simonharrer:
Thank you for the PR. Nice and clean implementation! We would gladly accept more PRs from you. ;)

PR #424, opened 2015-11-30
Removed file history size from preferences

oscargus:
Removed file history size as a preference variable, since it could not be set anyway and replaced it with a constant. See discussion in koppor/Jabref#43.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ to [sourceforge feature requests](https://sourceforge.net/p/jabref/features/) by
 - Biblatex preserves capital letters, checking whether letters may be converted to lowercase within the Integrity Check action is obsolete. Fixes #412 
 
 ### Removed
-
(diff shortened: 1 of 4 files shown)


simonharrer:
Please add a CHANGELOG entry. Otherwise, it is good to be merged.

oscargus:
Is it really required? I can of course do it, but since it is more of an internal change I thought it didn't make sense. (Or do people actually edit the preferences by hand?)

simonharrer:
Hm, well, in this case, it is debatable as the preference was never setable through the GUI. But if you would have loaded a preference.xml, you were able to change this - which is now not possible anymore. Hence, I would add a line. Otherwise, for internal changes, no CHANGELOG entry is required (normally, as no rule without an exception :)

simonharrer:
:+1:

oscargus:
Done!

PR #427, opened 2015-12-01
Cleanuptests

oscargus:
Refactored some of the cleanup actions and added tests. The tests could be more complete, but better than nothing.

### DIFF
--- a/src/main/java/net/sf/jabref/logic/cleanup/LatexCleanup.java
+++ b/src/main/java/net/sf/jabref/logic/cleanup/LatexCleanup.java
@@ -26,10 +26,24 @@ public class LatexCleanup {
         this.entry = entry;
     }
 
+    public LatexCleanup() {
+        this.entry = null;
+    }
+
     public void cleanup() {
+        if (this.entry == null) {
+            return;
+        }
+
         final String field = "title";
 
-        String newValue = entry.getField(field);
+        String newValue = format(entry.getField(field));
+
+        entry.setField(field, newValue);
+    }
+
+    public String format(String oldString) {
+        String newValue = oldString;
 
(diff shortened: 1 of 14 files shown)


src/main/java/net/sf/jabref/logic/cleanup/LatexCleanup.java
```diff
@@ -0,0 +1,64 @@
...
+
+    You should have received a copy of the GNU General Public License along
+    with this program; if not, write to the Free Software Foundation, Inc.,
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+package net.sf.jabref.logic.cleanup;
+
+import net.sf.jabref.model.entry.BibtexEntry;
+
+public class LatexCleanup {
```
simonharrer: Please try to create a formatter without any dependencies to a BibtexEntry. The setting of the new value should be done in the `Cleanup` class whereas the formatter should also implement the `Formatter` interface. Have a look at other formatters already created, e.g., PageNumberFormatter.

oscargus: Oh, quite a few of these things were already done... I'll get back with an updated PR. (I looked at PageNumberFormatter, but clearly not good enough...)

tobiasdiez: You might also have a look at #305

oscargus: Yes, that PR changed things a bit! :+1:

oscargus:
Replaced with #440

Issue #420, opened 2015-11-30
Cannot switch to BibLatex mode in preferences


stefan-kolb:
```
Exception in thread "AWT-EventQueue-0" java.lang.ClassCastException: sun.nio.cs.UTF_8 cannot be cast to java.lang.String
	at net.sf.jabref.gui.preftabs.GeneralTab.storeSettings(GeneralTab.java:245)
	at net.sf.jabref.gui.preftabs.PreferencesDialog$OkAction$1.run(PreferencesDialog.java:266)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:497)
	at spin.Invocation.evaluate(Invocation.java:175)
	at spin.off.SpinOffEvaluator$1.run(SpinOffEvaluator.java:108)
	at java.lang.Thread.run(Thread.java:745)
Exception in thread "AWT-EventQueue-0" java.lang.ClassCastException: sun.nio.cs.UTF_8 cannot be cast to java.lang.String
	at net.sf.jabref.gui.preftabs.GeneralTab.storeSettings(GeneralTab.java:245)
	at net.sf.jabref.gui.preftabs.PreferencesDialog$OkAction$1.run(PreferencesDialog.java:266)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:497)
	at spin.Invocation.evaluate(Invocation.java:175)
	at spin.off.SpinOffEvaluator$1.run(SpinOffEvaluator.java:108)
	at java.lang.Thread.run(Thread.java:745)
```

Related to charset PR #396

simonharrer:
@tobiasdiez can you have a look as this relates to your PR?

tolot27:
Change GeneralTab.java:245
```java
        prefs.put(JabRefPreferences.DEFAULT_ENCODING, (String) encodings.getSelectedItem());
```
into
```java
[...]

PR #429, opened 2015-12-01
Fix similar bug as #420 for the database properties dialog

tobiasdiez:
A similar bug as #420 happened when the encoding was changed in the database properties dialog. This is now fixed in this PR. 

Sorry for introducing these bugs in the first place.


simonharrer:
Thanks for the fixup PR.

PR #432, opened 2015-12-01
Refactors how the default encoding is stored in the preferences

tobiasdiez:


src/main/java/net/sf/jabref/JabRefPreferences.java
```diff
@@ -1668,4 +1669,12 @@ public boolean isNonWrappableField(String fieldName) {
...
         singleton = owPrefs;
     }
+
+    public Charset getDefaultEncoding() {
+        return Charset.forName(get(JabRefPreferences.DEFAULT_ENCODING));
+    }
+
+    public void setDefaultEncoding(Charset encoding) {
+        put(JabRefPreferences.DEFAULT_ENCODING, encoding.name());
+    }
```
simonharrer: Could you create a Unit test which tests this method pair?

tobiasdiez: Done and rebased.

PR #441, opened 2015-12-01
Cleanups, warning removals, move and test.

oscargus:
Some cleanups and warning removals. Should get rid of six high impact Coverity issues (I hope).

### DIFF
--- a/src/main/java/net/sf/jabref/bst/BibtexWidth.java
+++ b/src/main/java/net/sf/jabref/bst/BibtexWidth.java
@@ -46,29 +46,7 @@ public class BibtexWidth {
     private static int[] widths;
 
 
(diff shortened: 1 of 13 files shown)


simonharrer:
:+1: looks good to me. The stupro PRs may have to rebase after this is merged.

oscargus:
Maybe, but I do not think so. However, it may be that the IsiImporterTest is not longer needed once the StuPro tests are merged. Either way, we can wait with merging this if it is a potential problem (or I remove the final commit which is the one potentially causing problems, I guess.)

simonharrer:
It would be best if you would remove the final commit, then I can merge your changes immediately.

oscargus:
OK! I've done that and hopefully stored the removed commit in such a way that it can be recreated. :-)

PR #446, opened 2015-12-02
Update of French localization

mlep:
Update of French localization


matthiasgeiger:
Thanks for the update!

Issue #445, opened 2015-12-02
Moving ranking column: moved visually, but function still in old place


sbitzer:
In JabRef 3.0 I moved the ranking column to the end of my entry table using drag and drop. The file and URL columns took its place. When I then clicked on the file icon, the ranking dialogue popped up (star selection) and when I clicked the URL icon the file was opened. I could also not drag and drop the ranking column back in place - it stayed after the author column. When I switched off the ranking column in preferences, the file and URL icons worked as expected. I had also moved the entry type column towards the end of the table, but guess that this didn't cause the observed behaviour.

I'm on Linux Mint 17.1. My Java version is Oracle Java 1.8.0_66.

koppor:
Possibly #443 is related.

matthiasgeiger:
Fixed with #452 - changes are available now in current [development builds](http://builds.jabref.org/master/) and will be released with 3.1

Issue #443, opened 2015-12-02
GUI: Clicking on Ranking Col Header switches position of DOI column before/after Entrytype


Siedlerchr:
In 3.0 I just discovered a strange behavior:

When I click on the Ranking Column Header, the DOI Column switches position from before entrytype to after entrytype. It seems this only happens with the Ranking Column, clicking on any other header does not change the col position.

See screenshots below:

![jabrefguirankingdoicolumn](https://cloud.githubusercontent.com/assets/320228/11529475/97b07390-98ed-11e5-8294-11c17f1c43fe.png)
![jabrefguirankingdoicolumn2](https://cloud.githubusercontent.com/assets/320228/11529476/97b80b5a-98ed-11e5-90e5-5da52949038e.png)

matthiasgeiger:
Confirmed for 3.0 and `master`.

matthiasgeiger:
Fixed with #452 - changes are available now in current [development builds](http://builds.jabref.org/master/) and will be released with 3.1

Issue #384, opened 2015-11-24
Tooltip for title or long field


melTr:
Adding a tooltip so we know the meaning of each column headtitle (R, R, Q, F, P is not obvious) 

When a field doesn't fit in column width, show a tooltip with the complete content.

Thanks

oscargus:
Thanks for your input. This issue is actively discussed among the developers, so something will happen. :-)

matthiasgeiger:
Postponed to 3.1 (will be part of a general overhaul of the MainTableFormat).

stefan-kolb:
Also see #315

matthiasgeiger:
Implemented with #452 - changes are available now in current [development builds](http://builds.jabref.org/master/) and will be released with 3.1

Issue #315, opened 2015-11-11
Rating colum has no header description


stefan-kolb:
We should add a description here, especially since clicking inside the column always opens up the rating menu which is inconvenient when i just want to open the dit view of the entry which works for all other columns.

Two small columns next to rating also have no description and I don't even know what they do.

koppor:
The reason for directly changing the flags convenience for quick changing of flags and ratings. I fully agree that the columns should be labeled. As far as I remember, that seemed to be really difficult to achieve. Maybe postpone for #113?

matthiasgeiger:
The problem is not to place a header label, but the space required for this: E.g.  the column for flagging, "print status", etc. are only a few px wide - but a description would require far more space.
There is actually already the preference setting to "show one letter column headings" - which helps a bit ;-)

The best solution would be to show the "icon" of the column as header label (which should not be a problem with our new font based icon set) and provide a [tooltip when hovering over the column headers](http://docs.oracle.com/javase/tutorial/uiswing/components/table.html#headertooltip)

matthiasgeiger:
... and remove the mentioned prefs option :wink:

oscargus:
There is also a method setAlternativeDisplayName() in BibtexSingleField (in BibTexFields) which should do the trick (that is currently unused).

matthiasgeiger:
Implemented with #452 - changes are available now in current [development builds](http://builds.jabref.org/master/) and will be released with 3.1

PR #452, opened 2015-12-03
Revise maintable

### LINKED ISSUE #498: Open pdf file when clicking pdf icon

matthiasgeiger:
Complete check and rewrite of MainTable and related classes

This will fix and implement:
- [x] #315 
- [x] #384 - show tooltip for header columns
- [x] #384 - show tooltip for contents which are too long for table columns
- [x] #443 
- [x] #445 
- [x] Add ability to drag and drop specialfield columns
- ~~[ ] Allow storing position of specialfield columns~~ postponed to another PR
- [x] remove table column for "PDF/PS" which is obsolete (fixes #498)
- [x] remove option "Show one letter heading for icon columns" which is now obsolete
- [x] Changed default order in main table: # | all file based icons (file, URL/DOI, ...) | all bibtex field based icons (bibtexkey, entrytype, author, title, ...) | all special field icons
- ~~[ ] #490~~ postponed
- ~~[ ] Think about "must be implemented "vernünftig (matthias)" "Entry table -> fit table horizontally ...."" from the minutes~~ postponed
- [x] final cleanup + CHANGELOG


simonharrer:
Awesome.

matthiasgeiger:
Before I invest much time to fix it...

@JabRef/developers **Are we dropping support for the "PDF/PS" column in the main table?**

In this column potential contents of the "legacy fields" `pdf` and `ps` are shown... 
In #251 the legacy tools are already dropped...
And we already have the abilility to define special columns based on a filetype (e.g. show a column for Excel files) - so it would be possible to show specific columns for PDFs (stored in the `file` field).

simonharrer:
Yes, support has been dropped for pdf/ps fields all over the place, we should do this as well for the main table.

koppor:
:+1: for removing these legacy columns.

Could someone create a FAQ entry (in htdocs) or send me a plain FAQ on how to mimic that behavior with the "special columns"?

simonharrer:
Very nice PR. With this, we already have a 3.1 :)

matthiasgeiger:
@koppor: 
> Q: The 'PDF/PS' column in the main table is missing - how can I restore them?
A: As the BibTeX fields `pdf` and `ps` are migrated to `file` the table column was obsolete. However, it is possible to define columns specifically for each file type in the preferences. (See: Preferences -> Entry Table Columns -> Show extra columns)
show-extra

@simonharrer Mhmm... but this gui stuff is quite unnerving :wink:

lenhard:
And we already have issues with the pdf/ps columns, which no longer seem to work correctly: #498
This confirms the need for their removal.

src/main/java/net/sf/jabref/gui/ImportInspectionDialog.java
```diff
@@ -1391,9 +1384,9 @@ private void setupComparatorChooser() {
             comparators = comparatorChooser.getComparatorsForColumn(i);
             comparators.clear();
             if (i == FILE_COL) {
-                comparators.add(new IconComparator(new String[]{Globals.FILE_FIELD}));
+                comparators.add(new IconComparator(Arrays.asList(Globals.FILE_FIELD)));
             } else if (i == URL_COL) {
-                comparators.add(new IconComparator(new String[]{"url"}));
+                comparators.add(new IconComparator(Arrays.asList("url")));
```
simonharrer: You could use Collections.singletonList instead.

src/main/java/net/sf/jabref/gui/maintable/MainTableColumn.java
```diff
@@ -0,0 +1,133 @@
...
+
+    private final boolean isIconColumn;
+
+    private final JLabel iconLabel;
+
+    private final BibtexDatabase database;
+
+    public MainTableColumn(String columnName) {
+        this.columnName = columnName;
+        this.bibtexFields = new ArrayList<>();
```
simonharrer: Collections.emptyList

src/main/java/net/sf/jabref/gui/maintable/MainTableColumn.java
```diff
@@ -0,0 +1,133 @@
...
+     *
+     * @return name to be displayed
+     */
+    public String getDisplayName() {
+        if(!bibtexFields.isEmpty()) {
+            StringBuilder builder = new StringBuilder();
+            int i=0;
+            for (String field : bibtexFields) {
+                if (i > 0) {
+                    builder.append(MainTableFormat.COL_DEFINITION_FIELD_SEPARATOR);
```
simonharrer: Use StringJoiner instead, which should handle this more readable.

src/main/java/net/sf/jabref/gui/maintable/MainTableColumn.java
```diff
@@ -0,0 +1,133 @@
...
+    public boolean isIconColumn() {
+        return isIconColumn;
+    }
+
+    public boolean isFileFilter() {
+        return false; // Overridden in SpecialMainTableColumns for file filter columns
+    }
+
+    public Object getColumnValue(BibtexEntry entry) {
+        if (!bibtexFields.isEmpty()) {
```
simonharrer: fail fast:
```
if(...isEmpty()) {
  return null;
}

// other code
```

src/main/java/net/sf/jabref/gui/maintable/MainTableFormat.java
```diff
@@ -0,0 +1,171 @@
...
+    // Values to gather iconImages for those columns
+    // These values are also used to put a heading into the table; see getColumnName(int)
+    private static final String[] URL_FIRST = {"url", "doi"};
+    private static final String[] DOI_FIRST = {"doi", "url"};
+    private static final String[] ARXIV = {"eprint"};
+    public static final String[] FILE = {Globals.FILE_FIELD};
+
+    private final BibtexDatabase database;
+
+    private List<MainTableColumn> tableColumns = new ArrayList<>();
```
simonharrer: can we make this list final?

src/main/java/net/sf/jabref/gui/maintable/SpecialMainTableColumns.java
```diff
@@ -0,0 +1,188 @@
...
+
+    /**
+     * Creates a MainTableColumn which shows an icon instead textual content
+     *
+     * @param columnName the name of the column
+     * @param fields     the entry fields which should be shown
+     * @return the crated MainTableColumn
+     */
+    public static MainTableColumn createIconColumn(String columnName, String[] fields, JLabel iconLabel) {
+        return new MainTableColumn(columnName, fields, iconLabel) {
```
simonharrer: Is this method every called with more than one field? Could it be simplified to just passing one field? (I don't like arrays :)

src/main/java/net/sf/jabref/gui/util/IconComparator.java
```diff
@@ -18,16 +18,17 @@
...
  * Comparator that handles icon columns.
  */
 public class IconComparator implements Comparator<BibtexEntry> {
 
-    private final String[] fields;
+    private final List<String> fields;
 
 
-    public IconComparator(String[] fields) {
+    public IconComparator(List fields) {
```
simonharrer: List<String> you probably mean.

lenhard: You probably mean `List<String>` :smirk:

simonharrer: Yes. Interestingly, when you type in `List<String>` without the quotation marks, the `<String>` part is not shown, only when editing the post.

src/main/java/net/sf/jabref/gui/maintable/SpecialMainTableColumns.java
matthiasgeiger: Yes, for "DOI/URL" resp. "URL/DOI".

matthiasgeiger:
Just rebased - from my point of view the current state is good enough to be merged into master.

src/main/java/net/sf/jabref/gui/maintable/MainTableColumn.java
```diff
@@ -0,0 +1,129 @@
...
+    private final JLabel iconLabel;
+
+    private final BibtexDatabase database;
+
+    public MainTableColumn(String columnName) {
+        this.columnName = columnName;
+        this.bibtexFields = Collections.emptyList();
+        this.isIconColumn = false;
+        this.iconLabel = null;
+        this.database = null;
```
simonharrer: Idea: use optional instead to avoid null checks.

src/main/java/net/sf/jabref/gui/maintable/MainTableNameFormatter.java
```diff
@@ -0,0 +1,40 @@
...
+        // Read name format options:
+        boolean namesNatbib = Globals.prefs.getBoolean(JabRefPreferences.NAMES_NATBIB); //MK:
+        boolean namesLastOnly = Globals.prefs.getBoolean(JabRefPreferences.NAMES_LAST_ONLY);
+        boolean namesAsIs = Globals.prefs.getBoolean(JabRefPreferences.NAMES_AS_IS);
+        boolean abbr_names = Globals.prefs.getBoolean(JabRefPreferences.ABBR_AUTHOR_NAMES); //MK:
+        boolean namesFf = Globals.prefs.getBoolean(JabRefPreferences.NAMES_FIRST_LAST);
+        boolean namesLf = !(namesAsIs || namesFf || namesNatbib || namesLastOnly); // None of the above.
+
+        if (nameToFormat == null) {
+            return null;
```
simonharrer: I would put this right at the beginning of the method. So that the rest of the code does not have think about null at all.

simonharrer:
:+1:

PR #453, opened 2015-12-03
Some clean ups/issue fixes

oscargus:
Got rid of some more Coverity issues.


simonharrer:
Please do not use abbreviations in variable names, especially if the variable is used in a large method. If the method would be only three lines, the name does not matter much.

oscargus:
Had some problems moving the branch between computers, so hopefully the online edit worked...

PR #456, opened 2015-12-03
Default time stamps follows ISO-8601

koppor:
Default time stamps follows ISO-8601. Reason: https://xkcd.com/1179/

I am not aware of side effects at the CurrentDate formatter.


simonharrer:
:+1:

PR #460, opened 2015-12-03
Even more cleanups

oscargus:
Some more fixes based on Coverity and findbugs.

### DIFF
--- a/src/main/java/net/sf/jabref/importer/fetcher/GoogleScholarFetcher.java
+++ b/src/main/java/net/sf/jabref/importer/fetcher/GoogleScholarFetcher.java
@@ -37,6 +37,7 @@ import java.nio.charset.StandardCharsets;
 import java.util.*;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 public class GoogleScholarFetcher implements PreviewEntryFetcher {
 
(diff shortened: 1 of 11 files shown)


src/main/java/net/sf/jabref/importer/fetcher/GoogleScholarFetcher.java
```diff
@@ -173,9 +173,9 @@ private static void runConfig() throws IOException {
...
             formItems.put("num", String.valueOf(GoogleScholarFetcher.MAX_ENTRIES_TO_LOAD));
             StringBuilder ub = new StringBuilder(GoogleScholarFetcher.URL_SETPREFS + "?");
-            for (Iterator<String> i = formItems.keySet().iterator(); i.hasNext();) {
-                String name = i.next();
-                ub.append(name).append("=").append(formItems.get(name));
+            for (Iterator<Map.Entry<String, String>> i = formItems.entrySet().iterator(); i.hasNext();) {
+                Map.Entry<String, String> name = i.next();
+                ub.append(name.getKey()).append("=").append(name.getValue());
                 if (i.hasNext()) {
                     ub.append("&");
```
simonharrer: Maybe use StringJoiner instead of StringBuilder. This would then allow you to remove this iterator stuff.

src/main/java/net/sf/jabref/model/entry/UnknownEntryType.java
```diff
@@ -68,4 +68,20 @@ public String getName() {
...
+        } else {
+            return false;
+        }
+    }
+
+    @Override
+    public int hashCode() {
+        // TODO Auto-generated method stub
+        return super.hashCode();
+    }
```
simonharrer: Can you please remove the TODO tags?

src/main/java/net/sf/jabref/importer/fetcher/GoogleScholarFetcher.java
oscargus: I do not really understand how to use a StringJoiner with a Map.

oscargus: But I got an idea now, so let's see of ot works...

oscargus:
Actually not. It turns out that Entry<String,String>.toString() returns "key=value". Which is a bit unexpected, but makes sense and especially is efficient here.

simonharrer:
Ah ok. :)

PR #464, opened 2015-12-04
Refactor code from cleanup action to separate classes

tobiasdiez:
I removed all the functionality from `CleanupAction` to separate classes. The next step would be to add tests....next PR :)


oscargus:
:+1:

I'd add copyright headings to all new files.

PR #467, opened 2015-12-05
Add ability to restore focus to last focused tab on startup

tobiasdiez:


simonharrer:
Nice idea. 

I would prefer a class LastFocusedTabPreference with two methods setting the lastFocused Path and getting the lastFocusedP Path, putting the functionality of setting/comparing at a single place instead of two different places in the large JabRef class.

tobiasdiez:
You mean that the lines 
````
Globals.prefs.get(JabRefPreferences.LAST_FOCUSED);
---
String focusedDatabase = getCurrentBasePanel().getDatabaseFile().getAbsolutePath();
prefs.put(JabRefPreferences.LAST_FOCUSED, focusedDatabase);
````
should be moved to a separate class? I can do this but I doubt that this helps with the spaghetti code in `JabRef.java`  :smirk:

**Comment by simonharrer** [comment]
It only helps a little bit, I admit, but laboriously the squirrel lives. :)

I had something as the following in mind:

```
public class LastFocusedTabPreferences {

    public void setLastFocusedTab(File file) {
       // make absolute path, set, ...
    }

    public File getLastFocusedTab() {

    }

    public boolean hasHadLastFocus(File file) {
       // true if file.absolutePath.equals(getLastFocusedTab())
    }

}
```

With that, you can save a few lines. Have a look at the `RemotePreferences` class, it does something similar.

**Comment by tobiasdiez** [comment]
Ok, I understand. This is indeed cleaner and easier to use. One last question before I implement your suggestion: wouldn't it be better to move these two methods into JabRefPreferences instead of creating a new class? This would allow to make the field `LAST_FOCUSED` private so that nobody gets the idea to use the good old `get/put(String)` methods.

**Comment by simonharrer** [comment]
Hm, with this argument, the JabRefPreferences class would get extremely large if we did this for every preference field. Because of this, I would like to keep it separated.

**Comment by tobiasdiez** [comment]
Done.

File: src/main/java/net/sf/jabref/logic/preferences/LastFocusedTabPreferences.java [hunk 1]
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    public void setLastFocusedTab(File file) {
+        preferences.put(JabRefPreferences.LAST_FOCUSED, file.getAbsolutePath());
+    }
+
+    public boolean hadLastFocus(File file) {
+        String lastFocusedDatabase = preferences.get(JabRefPreferences.LAST_FOCUSED);
+        return file.getAbsolutePath().equals(lastFocusedDatabase);
+    }
```
simonharrer: Nice! Last comment: I think we should make this class also null-reference robust. I suggest to use Objects.requireNonNull for the preferences and for the file references just provide a special routine in case null is set. Why? At the moment, we have so many cases in which a reference can be null, I rather would avoid that this code is the cause of another bug report.

simonharrer:
Please harden the class and rebase, then this can be merged.

tobiasdiez:
Thanks for your feedback @simonharrer. I feel the code improved considerably through it. I added the suggested null tests now. Hence this should be good to go.

src/main/java/net/sf/jabref/logic/preferences/LastFocusedTabPreferences.java
```diff
@@ -0,0 +1,50 @@
...
+            filePath = file.getAbsolutePath();
+        }
+
+        preferences.put(JabRefPreferences.LAST_FOCUSED, filePath);
+    }
+
+    public boolean hadLastFocus(File file) {
+        String lastFocusedDatabase = preferences.get(JabRefPreferences.LAST_FOCUSED);
+        if (file == null) {
+            return false;
```
simonharrer: One last remark. ;)

I would use fail fast, that is, write the method as follows:

```
if(file == null) {
  return false;
}
String lastFocusedDatabase ...
// remaining code
``` 

This has the advantage that input checks do not have to be found somewhere in the code. Furthermore, you reduce the number of indention by 1 for the typical use case, you separate the typical use case from the failure case. And if/else should be used when both branches are equally likely - but typically, we have one case which is very likely and one which is very unlikely - and the code should reflect that as well.

tobiasdiez: Good point, changed it.

PR #468, opened 2015-12-05
Improved duplicate checker

oscargus:
I changed the core function of the duplicate checker from correlation to approximate string matching. This enables the test to work (which earlier failed as `Bob` and `Joyce` was deemed as close enough to indicate a duplicate...). Probably some of the constants used for comparison needs to be tuned. Also, as the approximate string matching is based on edit distance (the number of edits required to turn one string into the other), it might be that one may operate the complete comparison on the full string rather than the word by word style as currently is used.

I think this is good to go (in the sense that it probably performs better than the previous), but please check it with a few different databases to make sure that the false positives or missing hits are to frequent.

### DIFF
--- a/src/main/java/net/sf/jabref/bibtex/DuplicateCheck.java
+++ b/src/main/java/net/sf/jabref/bibtex/DuplicateCheck.java
@@ -30,6 +30,8 @@ import org.apache.commons.logging.LogFactory;
  */
 public class DuplicateCheck {
 
+    private static final Log LOGGER = LogFactory.getLog(DuplicateCheck.class);
+
     /*
      * Integer values for indicating result of duplicate check (for entries):
      *
@@ -49,8 +51,6 @@ public class DuplicateCheck {
 
     // Extra weighting of those fields that are most likely to provide correct duplicate detection:
     private static final HashMap<String, Double> fieldWeights = new HashMap<>();
-
-    private static final Log LOGGER = LogFactory.getLog(DuplicateCheck.class);
 
 
     static {
--- a/src/main/java/net/sf/jabref/bibtex/DuplicateCheck.java
+++ b/src/main/java/net/sf/jabref/bibtex/DuplicateCheck.java
@@ -22,6 +22,9 @@ import net.sf.jabref.model.entry.BibtexEntry;
 import java.util.HashMap;
 import java.util.HashSet;
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
 /**
  * This class contains utility method for duplicate checking of entries.
  */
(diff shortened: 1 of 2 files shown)


src/main/java/net/sf/jabref/bibtex/DuplicateCheck.java
```diff
@@ -234,88 +240,59 @@ static double correlateByWords(String s1, String s2, boolean truncate) {
...
+        String longer = s1, shorter = s2;
+        if (s1.length() < s2.length()) { // longer should always have greater length
+            longer = s2;
+            shorter = s1;
         }
-        return DuplicateCheck.corrCoef(n1, n2);
-    }
+        int longerLength = longer.length();
+        if (longerLength == 0) {
+            return 1.0;
```
simonharrer: This line is untested.

oscargus: Correct, I haven't added any tests, just enabled the old ones. There are probably quite a bit more things to test, like complete entries, to get complete code coverage.

I can do that, but somehow, the interesting thing would be to see how it behaves in practice.

(Those lines would just return that the strings are identical in case both are `""`.)

src/main/java/net/sf/jabref/bibtex/DuplicateCheck.java
```diff
@@ -47,6 +50,9 @@
     // Extra weighting of those fields that are most likely to provide correct duplicate detection:
     private static final HashMap<String, Double> fieldWeights = new HashMap<>();
 
+    private static final Log LOGGER = LogFactory.getLog(DuplicateCheck.class);
```
stefan-kolb: I would tend to add the logger always in the first line after class declaration..

oscargus:
I added some more test cases to get the coverage up and fixed @stefan-kolb's comment.

simonharrer:
If you want feedback for a PR, it is best if you publish your branch on the jabref repository. Because in that case, CircleCI can automatically build a version with your improvements. Otherwise, it is much more cumbersome for the other developers to try out your changes.

oscargus:
OK! I'll do that!

oscargus:
OK, now I pushed to upstream instead and the branch shows up in the list of branches. Will be patient for the tests to finish, but after that there may be a version to test.

simonharrer:
Ok, tried it. Seems good. But got 4 false positives in my library. But these four false positives refer to very very similar entries. In addition, it found an entry the previous algorithm did not find. So this is better from my perspective. 

I vote for merging.

simonharrer:
This uncovered another issue, namely, #477

koppor:
Should the merge entries dialog also open if the DOIs differ? http://link.springer.com/chapter/10.1007%2F978-3-642-14292-5_55 and http://link.springer.com/chapter/10.1007%2F978-3-642-14292-5_54 are considered as duplicates here.

koppor:
Standards such as BPMN 1.0 and BPMN 2.0 are considered as duplicate. Also different editions of a book (Weske: Business Process Management). But I think, all of this has nothing to do with the new implementation ^^.

PR #469, opened 2015-12-05
Added tests leading to small fixes in EntryTypes

oscargus:
"Major" change in behavior: if `EntryTypes.getBibtexEntryType` is called with a non-existing entry type name, now `BibLatexEntryTypes.MISC` is returned in BibLatex-mode instead of `BibtexEntryTypes.MISC`. If something breaks because of this, it is probably good that we found it...


simonharrer:
:+1:

oscargus:
How on earth do I make sure that a test on CircleCI starts in BibLatex mode? The test works on my computer, but it seems like the BibLatex-version of the test doesn't start in BibLatex-mode... Current approach:
``` java
@Before
        Globals.prefs = JabRefPreferences.getInstance();
        biblatex = Globals.prefs.getBoolean(JabRefPreferences.BIBLATEX_MODE);
        Globals.prefs.putBoolean(JabRefPreferences.BIBLATEX_MODE, true);
}
```

(and of course restoring in @After.)

simonharrer:
> java.lang.AssertionError: expected:<net.sf.jabref.model.entry.BibLatexEntryTypes$1@38e92bb1> but was:<net.sf.jabref.model.entry.BibtexEntryTypes$1@18b49f09>

Could you add a toString Method for the entry types classes?

> How on earth do I make sure that a test on CircleCI starts in BibLatex mode? The test works on my computer, but it seems like the BibLatex-version of the test doesn't start in BibLatex-mode... Current approach:

Hm, should work.

koppor:
I would replace the third statement by a more generic approach as shown at https://github.com/JabRef/jabref/wiki/Code-Howtos#test-cases. Otherwise, I think, that this the current way to go.

Then, all test classes changing the preferences look similar.

oscargus:
@simonharrer : Yes, I was thinking about that. However, it is not strictly the same since (at least in the application) a user may redefine the "article" entry type. From a test perspective I guess it can be OK for now though. However, for a complete test, one would actually like to redefine a standard field and make sure that the original is restored once the custom field is deleted and then I cannot see any other way than to check it like this.

@koppor: OK, could do that. Didn't read those line that carefully so I thought I was doing the same. :-)

oscargus:
@koppor: overwritePreferences in not public...

koppor:
@oscargus Adapt it accordingly. @lenhard Will have a look as this concept is IMHO invented by him.

oscargus:
Hmm, of course the length test will fail as well... Any idea why CircleCI refuses to switch to BibLatex mode? Maybe it makes sense to just remove the BibLatex-test version (as it is currently not testing the BibLatex-mode anyway) for now?

koppor:
Could you try to add a test case whether the current mode is BibLaTeX that we are really sure that the switch does not work? On the one hand, an idea would be to have the preferences stored externally. On the other hand, I would not understand why a change in a HashMap succeeds locally and not on CircleCI.

BTW: Fine with me to add an `@Ingore` tag to the whole biblatex type test class if we can't move forward here.

oscargus:
I restored the original tests comparing objects, but added `@Ignore` to the BibLatex-test. Added the test:

``` Java
    @Test
    public void textIsThisBibLatex() {
        assertTrue(Globals.prefs.getBoolean(JabRefPreferences.BIBLATEX_MODE));
    }
```

oscargus:
Appears as if that test passed, so somehow the initialization of EntryTypes is done before the preference is changed...

simonharrer:
Yes, you are right. The `EntryTypes` class uses a static initializer which is only called once when loading the class for the first time. During that time, it sets of the entry types, and ignores the value afterwards. We should refactor this, as we aim to have a file-based bibtex/latexmode switch instead of a global one.

oscargus:
Weird though that the test works on my computer, so there the
initialization of EntryTypes is apparently not done before the mode is
switched.

Anyway, with database-local EntryTypes this will be solved (eventually).
May require quite a bit of refactoring though, considering how things work
now (or maybe not). I guess the major thing will be that one cannot create
entries without having a database, which already is that way in the UI.

simonharrer:
Nevertheless, I vote for merging this PR, as it is still helpful, despite being unable to test everything to its fullest.

lenhard:
For the record: I wrote the `overwritePreferences` method to be able to write tests for `JabRefPreferences`. The method wasn't needed anywhere else, so I reduced visibility as far as possible. If you need more far-reaching visibility, this is fine.

tobiasdiez:
Could you please explain the reason for the test
```` java
CustomEntryType newArticle = new CustomEntryType("article", requiredFields,
                BibtexEntryTypes.ARTICLE.getOptionalFields());

bibtexentrytypes.addOrModifyCustomEntryType(newArticle);
// Should not be the same any more
assertNotEquals(BibtexEntryTypes.ARTICLE, bibtexentrytypes.getType("article"));
````

I would have expected that the test should fail (i.e. that the entry type is still an article).

**Comment by oscargus** [comment]
"article" is now a custom type with the same name as a standard type.

I saw this comment in the remove-method which inspired the test:
```
// In this case the user has removed a customized version
// of a standard type. We reinstate the standard type.
```

This is also the reason that I check against the actual type and not against the name.

tobiasdiez:
I would have expected that all entry types with the same name are considered equal. So even `assertEquals(BibtexEntryTypes.ARTICLE, BibLatexEntryTypes.ARTICLE)` should return true, shouldn't it?

lenhard:
A Bibtex article and a BibLatex article differ in the fields they require or have as optional. They should not be considered equal.

tobiasdiez:
But what is then the best way to say that an entry has a specific type without using explicit strings?
I thought a `assertEquals(BibtexEntryTypes.ARTICLE, entry.getType())` was a neat way to test weather the entry is an article. This test is already used in some tests.

But if bibtex types are considered different from biblatex types and from customized types, then the result of the above statement depends on the state of the program while parsing / creating the entry.

lenhard:
Sorry, I don't see the problem with that. The types are different, so why should we treat them as equal?

Issue #470, opened 2015-12-05
Define alternative fields for displaying in main table


tobiasdiez:
Currently, if an entry does not have an author then nothing is shown in author column of the main table. It would be nice if in such cases JabRef would fall-back and try another field (e.g. editor).
(If this is implemented, the column header should be adapted accordingly, i.e. show "Author / Editor")

The following combinations present themselves:
- Author / Editor
- Journal (Article) / Publisher (Book)
- Title could be supplemented by subtitle (in the case we implement the multi-line display of entries as [suggested by sci](http://sourceforge.net/p/jabref/mailman/message/34519834/) )

oscargus:
You can just write "author/editor" in the entry table editor and it will happen. (I learned a few weeks ago.)

oscargus:
Or you mean it should be automagic?

tobiasdiez:
Thanks for the tip! In this case, it would be nice if this feature is used by default. Accomplished in #472.

lenhard:
If the issue is fixed in #472 it can be closed, right?

tobiasdiez:
Yes can be closed.

PR #472, opened 2015-12-05
Show author/editor and journal/publisher in the main table by default

tobiasdiez:
Implements #470.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ to [sourceforge feature requests](https://sourceforge.net/p/jabref/features/) by
 ## [Unreleased]
 
 ### Changed
+- Implements #470: Show editor (as an alternative to author) and booktitle (as an alternative to journal) in the main table by default
 - Add ability to format/cleanup the date field
 - Add support for proxy authentication via VM args and GUI settings, this implements Feature Request #388
 - Move Bibtex and Biblatex mode switcher to File menu
--- a/src/main/java/net/sf/jabref/JabRefPreferences.java
+++ b/src/main/java/net/sf/jabref/JabRefPreferences.java
@@ -521,7 +521,7 @@ public class JabRefPreferences {
         defaults.put(SIDE_PANE_COMPONENT_NAMES, "");
         defaults.put(SIDE_PANE_COMPONENT_PREFERRED_POSITIONS, "");
 
-        defaults.put(COLUMN_NAMES, "entrytype;author;title;year;journal;bibtexkey");
+        defaults.put(COLUMN_NAMES, "entrytype;author/editor;title;year;journal/booktitle;bibtexkey");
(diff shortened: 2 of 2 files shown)


koppor:
Looks like follows:

grabbed_20151205-225839

"Author/Editor" is OK.

Unsure if "Journal/Publisher" is confusing. I think, "booktitle/journal" would be the right correspondence. We should replace the default "journal" column by "booktitle/journal".

How often is `InBook` used? Consequently, `title` should become `chapter/title`, but that heading is confusing.

oscargus:
I'd also vote for journal/booktitle (as that is what I use, dealing primarily with journal and conference papers). On the other hand, a good thing with this PR is that it indicates that these things are possible and how to do it.

tobiasdiez:
The journal says where an article is published. Similarly, the publisher is...well... the publisher of the book. So I would advocate the proposed `journal/publisher`. Booktitle would probably only make sense for articles in conference proceedings. I think articles and books are the most common types, so we should focus on them.

@koppor: `InBook`s also have an author and title, so there shouldn't be a problem.

oscargus:
Guess it depends on scientific area. I would possibly argue that in general, a publisher doesn't really add any important information, even for books, while in the case of conference papers, the booktitle is crucial. (I know that the publisher is important to create a valid reference, but it doesn't really add any information when browsing a database.)

koppor:
In my discipline (computer science), I'd see the correspondence as follows (bibtex fields)

inproceedings | journal
------- | -------
title | title
author | author
booktitle | journal
publisher | publisher
year | year

For instance, take an Elsevier article (http://www.sciencedirect.com/science/article/pii/S0268401215000808). The publisher is Elsevier, but the Journal is "International Journal of Information Management". Elsevier publishes more than one journal.

I read "where something is published" as the conference and the concrete journal. The conference is stored in "booktitle", whereas the concrete journal is stored in "journal". Both kinds of works are published by a publisher.

Is it different in your discipline @tobiasdiez?

oscargus:
I do not think @tobiasdiez disagrees with that analysis, but I got the
impression that he cites more books than conference papers. In that case I
can probably agree that publisher is the "next" field to add (mainly
because of lack of others).

Journal/booktitle/publisher? (Except for the very long heading...)

tobiasdiez:
Ok apparently the majority is for booktitle. Hence I changed it accordingly. 

My reasoning was indeed based on `book`s for which the publisher is the equivalent of a journal. Since books occur more often in my database, I thought this addition was more valuable. For `InProceedings` the booktitle indeed makes more sense.

koppor:
I suppose, `journal/booktitle/publisher` is too confusing? I really agree that the column `journal/booktitle` is empty too often if many books are in the database.

I'll merge it and let's see what others say. :sunflower:

Issue #473, opened 2015-12-06
Import/Export to external database fails for PostgreSQL in 3.0 of jabref


fishforbreakfast:
Hi,

Just tried 3.0.  The import and export features for PostgreSQL fail with the following messages:

For Export - "Could not export to SQL database for the following reason:
                     This connection has been closed"

For Import - "Could not import from SQL database for the following reason:
                      This ResultSet is closed."

Establishing a connection to the database returns successfully.

Thanks

Fred

koppor:
Which version of PostgreSQL do you use? I suppose that 2.11.1 worked fine? If not, which was the JabRef version working correctly?

fishforbreakfast:
Hi Oliver,

Everything works fine in 2.11.1.  I am using 9.4.5 of PostgreSQL. Just 
thought I'd try 3.0 since it is out.  I really like the new look by the 
way.  Thank you for the efforts to continually improve JabRef.

Cheers,

Fred

oscargus:
I made some changes that may solve this in #476. Not sure how you can get a complete build of it before being merged though...

simonharrer:
@oscargus if you create the branch on jabref instead of your local repository, your builds will show up on builds.jabref.org

oscargus:
Ahh, OK! Will have to read up in that!

@simonharrer Now you created a branch, right? (It looked like.)

simonharrer:
I think @koppor created one.

koppor:
@fishforbreakfast Please try the build at http://builds.jabref.org/databasefix/

fishforbreakfast:
Hi,

Just tried your build with exactly the same behavior.  ???  I'll keep checking for more updates to test.  If there is some debugging that I could do, let me know.

Cheers,

Fred

oscargus:
OK! I've reversed some more changes. A new build is available at: http://builds.jabref.org/databasefix/

Not sure what else could be done to get more information about exactly where the problem actually is. (The origin is that I introduced a bit too much automatic closing to get rid of code warnings, sorry about that.)

fishforbreakfast:
Hi,

Import is working.  Export will not work with 2.11 series of database schema.  It also will not create a new entry/entries in an empty database.  The following error occurs:

"Could not export to SQL database for the following reason:
Error: more than one row returned by a subquery used as an expression"

From postgresql log:
[...]

PR #476, opened 2015-12-06
Fixed(?) PostgreSQL problems

oscargus:
This may solve the problem in #473 

I was apparently a bit to enthusiastic in introducing try-with resources...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ to [sourceforge feature requests](https://sourceforge.net/p/jabref/features/) by
 - Fixed #438: Cut, Copy and Paste are now translated correctly in the menu
 - Fixed #443/#445: Fixed sorting and moving special field columns
 - Fixed #498: non-working legacy PDF/PS column removed
+- Fixed #473: Import/export to external database works again
 
 ### Removed
 - Removed file history size preference (never available from the UI)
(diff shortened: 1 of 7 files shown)


oscargus:
The important change is in PostgreSQLExporter where the connection is not automatically closed, which is probably what causes the issue.

The other changes are:
* Better code by not having two try directly following each other (now both statements are in one try)
* Avoided duplicate translations which is bound to give warnings using another locale

The reformatting may make it hard to see though.

koppor:
@oscgarcus If you created the branch in `upstream` (meaning `JabRef/jabref`) instead of your fork, the build would be available at http://builds.jabref.org/. Reason: We authenticate the upload with an SSH key which doesn't seem to be available when branches of forks are being built. - Refs https://github.com/JabRef/jabref/issues/473#issuecomment-162327856

I created a `databasefix` branch in the main repo and now the build is available at http://builds.jabref.org/databasefix/.

oscargus:
:+1:

oscargus:
For reference, this PR is now primarily hosted in JabRef/jabref databasefix. I'll try to keep this branch updated as well.

stefan-kolb:
Any updates on the progress of this PR?

oscargus:
Not really. It works again, but there are some issues with escaping which probably should be fixed (but is not introduced as far as I can tell between 2.11 and 3.0). So, the issues I introduced in 3.0 is removed, which is the part I was primarily looking at solving, but some other issues with the database management have been revealed.

simonharrer:
Please add CHANGELOG and rebase on master. We will merge this then. Other issues can be fixed in other PRs.

oscargus:
Rebased and added change log entry.

PR #484, opened 2015-12-07
[WIP] Error console and debugging

oscargus:
Changed the order of the error console tabs.

Add CLI switch `--debug` which enables logging of debug messages (shown as `FINE`).


oscargus:
Hmm... Seems like this only works for the JabRef.class logger. Not sure why or if there's an easy way to make it work for all classes...

koppor:
I would like to have that setting in the preferences or in the Error Console. Like a button "enable debug logging for this session". or "enable debug logging" with a confirmation dialog that debug logging is activated until restart of JabRef.

Regarding the change of the log level, I don't really know.

oscargus:
Yeah, me too. CLI was a quick way to check it though. :-)

http://stackoverflow.com/questions/13760095/java-dynamically-change-logging-level

simonharrer:
Hm, maybe the option should also increase the number of log statements saved in the log statement cache. At the moment, the cache is set to 500 statements, see the `net.sf.jabref.logic.logging.Cache` class.

PR #485, opened 2015-12-08
Removed CsaImporter from the import formats. The csa import format

boceckts:
refers to the cambridge science abstract format which could be obtained
on their website csa.com. However the website has been merged with
proquest.com and now the there is no export into the csa format
available. Therefore we came to the conclusion to remove the support for
the csa import format. For more information see also:
http://sourceforge.net/p/jabref/mailman/jabref-users/thread/2E36AF04-A83F-44FF-8160-B62EF48B6FAA%40gmail.com/#msg34662091


simonharrer:
:+1:

Issue #324, opened 2015-11-12
Text input in authors field has strange behavior


stefan-kolb:
It sometimes jumps to another position while entering authors or overwrites text.
Should have todo with any on input actions that may be running.
Experienced while overwriting something like 

``` 
Khajeh-Hosseini, Ali and others
```

with 

```
Khajeh-Hosseini, Ali and Sommerville, Ian and Bogaerts, Jurgen and Teregowda, Pradeep
```

Edit: Should be an auto-completion problem.

koppor:
ACK for auto-completion problem. No idea how to write test cases and how to fix :cry:

stefan-kolb:
See #486

stefan-kolb:
See #644

PR #486, opened 2015-12-08
Add autocomplete tests

tobiasdiez:
- Add mockito test framework: It felt really easy to write tests changing properties with mockito. See for example `NameFieldAutoCompleterTest`, where the preference changes are just a one line thing. This however only works if the dependencies can be easily injected into the class. 

- Add a few tests to the AutoCompleter framework. Fixed some NullPointExceptions which I found.
- Fixed the bug that short keys and names were not completed, tests: `BibtexKeyAutoCompleterTest.completeShortKeyReturnsKey` and `NameFieldAutoCompleterTest.completeShortNameReturnsName`


build.gradle
```diff
@@ -99,6 +99,7 @@ dependencies {
     compile 'commons-logging:commons-logging:1.2'
 
     testCompile 'junit:junit:4.12'
+    testCompile 'org.mockito:mockito-core:1.+'
```
simonharrer: Please always use the precise version.

src/test/java/net/sf/jabref/logic/autocompleter/BibtexKeyAutoCompleterTest.java
```diff
@@ -0,0 +1,153 @@
...
+        new BibtexKeyAutoCompleter(null);
+    }
+
+    @Test
+    public void completeWithoutAddingAnythingReturnsNothing() {
+        AutoCompletePreferences preferences = mock(AutoCompletePreferences.class);
+        BibtexKeyAutoCompleter autoCompleter = new BibtexKeyAutoCompleter(preferences);
+
+        String[] result = autoCompleter.complete("test");
+        Assert.assertEquals(0, result.length);
```
simonharrer: please compare the array with an empty array, because then the failure message will be much more helpful than 1 instead of 0.

tobiasdiez: Mhhh... do you really find the test failure message `AssertionError: array lengths differed, expected.length=0 actual.length=1` more expressive? I have more problems understanding the line `Assert.assertArrayEquals(new String[] {}, result);`. 

Maybe I should just replace the array by an `ArrayList`, then one could simply check `isEmpty()`. Best solution?

simonharrer: Hm, normally, I do not use arrays but lists instead. Then, I get a detailed message comparing two lists by Assert.assertEquals(Collections.emptyList(), actualList);

tobiasdiez: I now completely removed arrays and used lists instead, which overall resulted in cleaner code.

src/main/java/net/sf/jabref/gui/autocompleter/AutoCompleteListener.java
```diff
@@ -410,7 +411,8 @@ private void resetAutoCompletion() {
     }
 
     private String[] findCompletions(String beginning) {
-        return completer.complete(beginning);
+        List<String> results = completer.complete(beginning);
+        return results.toArray(new String[results.size()]);
```
simonharrer: Would it make sense to change the method signature to List<String> instead of converting a list into an array?

src/main/java/net/sf/jabref/gui/autocompleter/AutoCompleteSupport.java
```diff
@@ -233,9 +234,9 @@ private void postProcessTextChange() {
         }
 
         String text = textComp.getText();
-        E[] candidates = autoCompleter.complete(text);
+        List<E> candidates = autoCompleter.complete(text);
         renderer.update(candidates);
-        if (textComp.isEnabled() && (candidates != null) && (candidates.length > 0)) {
+        if (textComp.isEnabled() && (candidates != null) && (candidates.size() > 0)) {
```
simonharrer: What is the contract here, can candidates ever be null or is always != null but may be empty?

src/main/java/net/sf/jabref/logic/autocompleter/AbstractAutoCompleter.java
```diff
@@ -58,9 +58,12 @@ public AbstractAutoCompleter(AutoCompletePreferences preferences) {
      * Otherwise the completion is case insensitive.
      */
     @Override
-    public String[] complete(String toComplete) {
+    public List<String> complete(String toComplete) {
+        if(toComplete == null) {
+            return new ArrayList<>();
```
simonharrer: If you need an empty list which is never modified but only read, you can use Collections.emptyList() which should be faster.

src/main/java/net/sf/jabref/logic/autocompleter/NameFieldAutoCompleter.java
```diff
@@ -43,26 +47,27 @@
      * @see AutoCompleterFactory
      */
     NameFieldAutoCompleter(String fieldName, AutoCompletePreferences preferences) {
-        this(new String[] {fieldName}, false, preferences);
+        this(new String[] {Objects.requireNonNull(fieldName)}, false, preferences);
     }
 
     public NameFieldAutoCompleter(String[] fieldNames, boolean lastNameOnlyAndSeparationBySpace,
```
simonharrer: Would it make sense to change the fieldNames array to a List<String> as well? If you only pass a list with one element, you can use Collections.singletonList("element");

simonharrer:
Good idea. I noted a few other things as well, following your idea of using lists instead of arrays. We can merge this in soon.

src/main/java/net/sf/jabref/gui/autocompleter/AutoCompleteListener.java
tobiasdiez: Yes it would make sense. However I see `AutoCompleteListener` as deprecated and it should be replaced by `AutoCompleteSupport` in the future.

src/main/java/net/sf/jabref/gui/autocompleter/AutoCompleteSupport.java
tobiasdiez: Normally one shouldn't return null as a list, right? As far as I see, only empty lists are returned. Should I remove the null check or leave it to be on the safe side?

src/main/java/net/sf/jabref/logic/autocompleter/AbstractAutoCompleter.java
tobiasdiez: Yes, I wondered about that too. In the end I have chosen the new list since I saw no reason to prohibit that the caller adds further strings afterwards.

src/main/java/net/sf/jabref/logic/autocompleter/NameFieldAutoCompleter.java
tobiasdiez: Good idea. Will change it this evening.

src/main/java/net/sf/jabref/gui/autocompleter/AutoCompleteSupport.java
simonharrer: Yes, one should never return null - which is true for lists as well. ;)

Please remove the null check, as it is obsolete.

stefan-kolb:
Also see #324

tobiasdiez:
@simonharrer Your suggestions are now implemented. From my side, this PR is ready to be merged.

@stefan-kolb I didn't noticed anything strange while writing tests for the autocompleters. Probably the bug #324 is only a gui problem.

simonharrer:
Maybe we should address 324 in another pull request. Merging this in.

PR #487, opened 2015-12-08
PdfXmpImporterTest added

obraliar:


koppor:
The importer class is a very small one. 100% covered: https://codecov.io/github/JabRef/jabref/src/main/java/net/sf/jabref/importer/fileformat/PdfXmpImporter.java?ref=fb612f9bb0acdca18317dbc6fb3c657c9eb3907d

Very closely related is XMPUtil. Not covered completely: https://codecov.io/github/JabRef/jabref/src/main/java/net/sf/jabref/logic/xmp/XMPUtil.java?ref=fb612f9bb0acdca18317dbc6fb3c657c9eb3907d I would vote to let this through and do a full coverage of XMPUtil in a separate pull request.

src/test/java/net/sf/jabref/importer/fileformat/PdfXmpImporterTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @Rule
+    public ExpectedException thrown = ExpectedException.none();
+
+    private PdfXmpImporter importer;
+
+
+    @Before
+    public void setUp() {
+        if (Globals.prefs == null) {
```
tobiasdiez: The if test can be removed.

src/test/java/net/sf/jabref/importer/fileformat/PdfXmpImporterTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @Test
+    public void testGetFormatName() throws Exception {
+        Assert.assertEquals("XMP-annotated PDF", importer.getFormatName());
+    }
+
+    @Test
+    public void testImportEntriesEncrypted() throws Exception {
+        InputStream isEncrypt = PdfXmpImporterTest.class.getResourceAsStream("encrypted.pdf");
+        thrown.expect(IOException.class);
```
tobiasdiez: Replace the `thrown` statement by adding a expected tag to the test, i.e. 
`@Test(expected=IOException.class)`

src/test/java/net/sf/jabref/importer/fileformat/PdfXmpImporterTest.java
```diff
@@ -0,0 +1,70 @@
...
+        importer = new PdfXmpImporter();
+    }
+
+    @Test
+    public void testGetFormatName() throws Exception {
+        Assert.assertEquals("XMP-annotated PDF", importer.getFormatName());
+    }
+
+    @Test
+    public void testImportEntriesEncrypted() throws Exception {
```
tobiasdiez: Rename tests to reflect the expected behavior. For example, this here could be `ImportEncryptedFileThrowsException`.

src/test/java/net/sf/jabref/importer/fileformat/PdfXmpImporterTest.java
```diff
@@ -0,0 +1,70 @@
...
+        InputStream isEncrypt = PdfXmpImporterTest.class.getResourceAsStream("encrypted.pdf");
+        thrown.expect(IOException.class);
+        importer.importEntries(isEncrypt, new OutputPrinterToNull());
+    }
+
+    @Test
+    public void testImportEntries() throws Exception {
+        InputStream is0 = PdfXmpImporterTest.class.getResourceAsStream("annotated.pdf");
+        List<BibtexEntry> bibtexEntries = importer.importEntries(is0, new OutputPrinterToNull());
```
tobiasdiez: Add test that bibtexEntries has length 1.

src/test/java/net/sf/jabref/importer/fileformat/PdfXmpImporterTest.java
koppor: See also https://github.com/JabRef/jabref/wiki/Code-Howtos#preferences-in-tests

src/test/java/net/sf/jabref/importer/fileformat/PdfXmpImporterTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+        Assert.assertEquals("how to annotate a pdf", bibtexEntries.get(0).getField("abstract"));
```
tobiasdiez: Replace the four `bibtexEntries.get(0)` calls by creating an instance `BibtexEntry entry = bibtexEntries.get(0)` and use this.

src/test/java/net/sf/jabref/importer/fileformat/PdfXmpImporterTest.java
```diff
@@ -0,0 +1,70 @@
...
+        Assert.assertEquals("pdf, annotation", bibtexEntries.get(0).getField("keywords"));
+        Assert.assertEquals("The best Pdf ever", bibtexEntries.get(0).getField("title"));
+    }
+
+    @Test
+    public void testIsRecognizedFormat() throws Exception {
+        InputStream isT = PdfXmpImporterTest.class.getResourceAsStream("annotated.pdf");
+        Assert.assertTrue(importer.isRecognizedFormat(isT));
+
+        InputStream isF = PdfXmpImporterTest.class.getResourceAsStream("empty.pdf");
```
tobiasdiez: Split into two tests.

tobiasdiez:
Good work. :+1: 
Please just rename the tests and fix some minor issues.
I think you forgot to add the `encrypted` and `empty` pdf.

src/test/java/net/sf/jabref/importer/fileformat/PdfXmpImporterTest.java
bartsch-dev: Isnt it better to leave the thrown part so that jUnit only expects the exception at line 43?
Wehn i write `@Test(expected = IOException.class)` the test will also pass if the expection occurs at line 41.

bartsch-dev:
the `encrypted.pdf` already was at `src/test/resources/encrypted.pdf`. Since no other test uses that file anymore i will move it.

koppor:
Eclipse shows the warning that the InputStreams are not closed. Reading the code, this seems to be true. Could you please use try-with-resources?

grabbed_20151222-120709

src/test/java/net/sf/jabref/importer/fileformat/PdfXmpImporterTest.java
koppor: Convinced. Leave as is.

bartsch-dev:
Fixed all the issues mentioned above.

koppor:
I would propose to leave it as `src/test/resources/`, because the file is used at both your test case and `XMPUtilTest`. The new path `.../jabref/importer/fileformat` indicates that the file is only used at the respective class and not elsewhere.

bartsch-dev:
OK, I left it there and fixed my `PdfXmpImporterTest`. Everything should be fine now.

PR #488, opened 2015-12-08
IEEEXplore fetcher downloads the raw bibtex records from IEEEXplore

jeff-kerr:
Summary of changes
- moved away from the freesearchresult.jsp script for older browsers (notice on the IEEE suggests support for this will be dropped in Jan 2016), instead use the 'search' script and POST method
- results are returned in JSON format and the article numbers are obtained directly; HTML parsing has been removed
- the article numbers are used to construct a second query for the raw Bibtex records, which now captures all fields available from the IEEE
- the Bibtex entries are still passed to the existing cleanup() routine to improve the formatting consistency
- cleaned up unsued code as a result of the changes (removed HTML parsing code)
- added a method to Util for performing HTTP POST

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ to [sourceforge feature requests](https://sourceforge.net/p/jabref/features/) by
 - Move Bibtex and Biblatex mode switcher to File menu
 - Display active edit mode (BibTeX or Biblatex) at window title
 - Implements #444: The search is cleared by either clicking the clear-button or by pressing ESC with having focus in the search field. 
+- IEEEXplore search now downloads the full Bibtex record instead of parsing the fields from the HTML webpage result (fixes #1146, #1267)
 
 ### Fixed
 - Fixed: Cleanup process aborts if linked file does not exists
--- a/src/main/java/net/sf/jabref/importer/fetcher/IEEEXploreFetcher.java
+++ b/src/main/java/net/sf/jabref/importer/fetcher/IEEEXploreFetcher.java
@@ -24,9 +24,10 @@ import java.net.CookieHandler;
 import java.net.CookieManager;
 import java.net.MalformedURLException;
 import java.net.URL;
-
-import java.util.HashMap;
-import java.util.Set;
+import java.net.UnknownHostException;
+import java.net.HttpURLConnection;
+import java.util.Collection;
+import java.util.Iterator;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -34,14 +35,16 @@ import javax.swing.JCheckBox;
 import javax.swing.JOptionPane;
 import javax.swing.JPanel;
 
-import net.sf.jabref.bibtex.EntryTypes;
 import net.sf.jabref.model.entry.EntryType;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.json.JSONArray;
+import org.json.JSONException;
+import org.json.JSONObject;
 
 import net.sf.jabref.*;
 import net.sf.jabref.importer.*;
-import net.sf.jabref.model.entry.IdGenerator;
+import net.sf.jabref.importer.fileformat.BibtexParser;
 import net.sf.jabref.logic.formatter.bibtexfields.UnitFormatter;
 import net.sf.jabref.logic.formatter.casechanger.CaseKeeper;
 import net.sf.jabref.logic.formatter.casechanger.CaseKeeperList;
@@ -53,65 +56,32 @@ import net.sf.jabref.util.Util;
 public class IEEEXploreFetcher implements EntryFetcher {
 
     private static final Log LOGGER = LogFactory.getLog(IEEEXploreFetcher.class);
+    private static final String URL_SEARCH = "http://ieeexplore.ieee.org/rest/search?reload=true";
+    private static final String URL_BIBTEX_START = "http://ieeexplore.ieee.org/xpl/downloadCitations?reload=true&recordIds=";
+    private static final String URL_BIBTEX_END = "&download-format=download-bibtex&x=0&y=0";
+    private static final String DIALOG_TITLE = Localization.lang("Search %0", "IEEEXplore");
+    private static final int MAX_FETCH = 100;
 
     final CaseKeeperList caseKeeperList = new CaseKeeperList();
     private final CaseKeeper caseKeeper = new CaseKeeper();
     private final UnitFormatter unitFormatter = new UnitFormatter();
-
     private final HTMLConverter htmlConverter = new HTMLConverter();
-
     private final JCheckBox absCheckBox = new JCheckBox(Localization.lang("Include abstracts"), false);
 
-    private static final int MAX_FETCH = 100;
-    private final int perPage = IEEEXploreFetcher.MAX_FETCH;
     private int hits;
-    private int unparseable;
     private int parsed;
-    private int piv;
     private boolean shouldContinue;
     private boolean includeAbstract;
 
     private String terms;
-    private final String endUrl = "&rowsPerPage=" + Integer.toString(perPage) + "&pageNumber=";
-    private String searchUrl;
-
-    private final Pattern hitsPattern = Pattern.compile("([0-9,]+) Results");
-    private final Pattern typePattern = Pattern.compile("<span class=\"type\">\\s*(.+)");
-    private final HashMap<String, String> fieldPatterns = new HashMap<>();
-    private final Pattern absPattern = Pattern.compile("<p>\\s*(.+)");
-
-    Pattern stdEntryPattern = Pattern.compile(".*<strong>(.+)</strong><br>" + "\\s+(.+)");
+    private String postData;
 
     private final Pattern publicationPattern = Pattern.compile("(.*), \\d*\\.*\\s?(.*)");
     private final Pattern proceedingPattern = Pattern.compile("(.*?)\\.?\\s?Proceedings\\s?(.*)");
-    Pattern abstractLinkPattern = Pattern.compile("<a href=\'(.+)\'>\\s*<span class=\"more\">View full.*</span> </a>");
-
-    Pattern ieeeArticleNumberPattern = Pattern.compile("<a href=\".*arnumber=(\\d+).*\">");
-
-    private final Pattern authorPattern = Pattern.compile("<span id=\"preferredName\" class=\"(.*)\">");
-    private static final String START_URL = "http://ieeexplore.ieee.org/search/freesearchresult.jsp?queryText=";
-
-    private static final String DIALOG_TITLE = Localization.lang("Search %0", "IEEEXplore");
-
-
-    // Common words in IEEE Xplore that should always be
 
     public IEEEXploreFetcher() {
         super();
         CookieHandler.setDefault(new CookieManager());
-
-        fieldPatterns.put("title", "<a\\s*href=[^<]+>\\s*(.+)\\s*</a>");
-        //fieldPatterns.put("author", "</h3>\\s*(.+)");
-        //fieldPatterns.put("author", "(?s)</h3>\\s*(.+)</br>");
-        // fieldPatterns.put("author", "<span id=\"preferredName\" class=\"(.+)\">");
-        fieldPatterns.put("volume", "Volume:\\s*([A-Za-z-]*\\d+)");
-        fieldPatterns.put("number", "Issue:\\s*(\\d+)");
-        //fieldPatterns.put("part", "Part (\\d+),&nbsp;(.+)");
-        fieldPatterns.put("year", "(?:Copyright|Publication) Year:\\s*(\\d{4})");
-        fieldPatterns.put("pages", "Page\\(s\\):\\s*(\\d+)\\s*-\\s*(\\d*)");
-        //fieldPatterns.put("doi", "Digital Object Identifier:\\s*<a href=.*>(.+)</a>");
-        fieldPatterns.put("doi", "<a href=\"http://dx.doi.org/(.+)\" target");
-        fieldPatterns.put("url", "<a href=\"(/stamp/stamp[^\"]+)");
     }
 
     @Override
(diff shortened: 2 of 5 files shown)


jeff-kerr:
I've coded in Java before, but never contributed to an open source project, so here goes.... I welcome your feedback!

Basically the idea here was that I'm a heavy IEEEXplore user and I like having all the fields.  I noticed keywords were missing and also the month was not working in the latest version (V3.0).  Then I looked under the hood and started getting up to speed and realized that Jabref was parsing the HTML result instead of downloading the full Bibtex records (which are available).  And this functionality seemed to be breaking from time to time as IEEE would update the web interface.

Rather than just complain as I have done in the past, I decided I would try to fix this.

So this update partially rewrites the fetcher to now perform the search using a newer script in IEEEXplore (the freesearchresult.jsp is being deprecated next month) which directly gives data in JSON.  Then the article numbers are used to construct a second query for the Bibtex records.  As I was finishing this I saw Oscar's note about the new API for IEEE, which gives results in XML (search results only, not the bibtex record).  Probably it makes sense to eventually switch to this, but I had already written the JSON part at that point.

I also cleaned up the code, removing the old HTML parsing stuff.  I kept the extensive clean-up routine since it seems like this still improves the formatting consistency (the IEEE Bibtex records still aren't the greatest).  Let me know what you think!

src/main/java/net/sf/jabref/util/Util.java
```diff
@@ -665,6 +664,66 @@ public static String getResultsWithEncoding(URLConnection source, Charset encodi
...
+                    response.append(inputLine);
+                }
+                in.close();
+            }
+        } else {
+            try (BufferedReader in = new BufferedReader(new InputStreamReader(source.getInputStream()));) {
+                while ((inputLine = in.readLine()) != null) {
+                    response.append(inputLine);
+                }
+                in.close();
```
simonharrer: With the try-with-resources construct, you do not need to call close() anymore.

src/main/java/net/sf/jabref/importer/fetcher/IEEEXploreFetcher.java
```diff
@@ -269,23 +296,21 @@ private BibtexEntry cleanup(BibtexEntry entry) {
...
-            	author = author.replaceAll("  ", " ");
-            	author = author.replaceAll("[ ,;]+$", "");
-            	entry.setField("author", author);
-            }
-        }*/
+        String author = entry.getField("author");
+        if (author != null) {
+            author = author.replaceAll("\\s+", " ");
+            author = author.replaceAll("\\.", ". ");
+            //author = author.replaceAll("([^;]+),([^;]+),([^;]+)", "$1,$3,$2"); // Change order in case of Jr. etc
```
simonharrer: Why did you comment this out?

src/main/java/net/sf/jabref/importer/fetcher/IEEEXploreFetcher.java
```diff
@@ -125,71 +95,137 @@ public JPanel getOptionsPanel() {
...
 
-            if (page.contains("Error Page")) {
+            //if max hits were exceeded, display the warning
+            if (hits > IEEEXploreFetcher.MAX_FETCH) {
                 // @formatter:off
-                status.showMessage(
-                        Localization.lang("Intermittent errors on the IEEE Xplore server. Please try again in a while."),
-                       DIALOG_TITLE, JOptionPane.INFORMATION_MESSAGE);
+                status.showMessage(Localization.lang("%0 entries found. To reduce server load, only %1 will be downloaded.",
+                        new String[] {String.valueOf(hits), String.valueOf(IEEEXploreFetcher.MAX_FETCH)}),
```
simonharrer: You do not need to pass a String array, just present the two parameters as additional parameters to the lang method.

src/main/java/net/sf/jabref/importer/fetcher/IEEEXploreFetcher.java
```diff
@@ -125,71 +95,137 @@ public JPanel getOptionsPanel() {
...
-            if (page.contains("You have entered an invalid search")) {
-                status.showMessage(Localization.lang("You have entered an invalid search '%0'.", terms),
-                        DIALOG_TITLE, JOptionPane.INFORMATION_MESSAGE);
-                return false;
-            }
+            HttpURLConnection con = (HttpURLConnection) url.openConnection();
+
+            //add request header
+            con.setRequestMethod("POST");
+            con.setRequestProperty("User-Agent", "Mozilla/5.0 (Windows NT 5.1; rv:31.0) Gecko/20100101 Firefox/31.0");
```
simonharrer: I have seen this code in the Util, maybe create a util method for opening a url with the default user agent?

src/main/java/net/sf/jabref/importer/fetcher/IEEEXploreFetcher.java
```diff
@@ -53,65 +56,32 @@
...
-    private final String endUrl = "&rowsPerPage=" + Integer.toString(perPage) + "&pageNumber=";
-    private String searchUrl;
-
-    private final Pattern hitsPattern = Pattern.compile("([0-9,]+) Results");
-    private final Pattern typePattern = Pattern.compile("<span class=\"type\">\\s*(.+)");
-    private final HashMap<String, String> fieldPatterns = new HashMap<>();
-    private final Pattern absPattern = Pattern.compile("<p>\\s*(.+)");
-
-    Pattern stdEntryPattern = Pattern.compile(".*<strong>(.+)</strong><br>" + "\\s+(.+)");
+    private String postData;
```
simonharrer: Can you try to convert the field variables as local variables in the method? With a quick check, I realized that this works for at least two fields.

simonharrer:
Thank you for the PR. This is exactly the spirit we are looking for! :)

Would you mind creating tests as well? This would allow use to determine when this will break. You could start of by looking at the `GoogleScholarTest` which tests a special kind of fetcher, namely, the full text finder.

src/main/java/net/sf/jabref/util/Util.java
```diff
@@ -665,6 +664,66 @@ public static String getResultsWithEncoding(URLConnection source, Charset encodi
     }
 
     /**
+     * Get the results of HTTP post on a URL and return contents as a String.
+     *
+     * @param source postData encoding
+     * @return
+     * @throws IOException
+     */
+    public static String getPostResultsWithEncoding(URL source, String postData, Charset encoding) throws IOException {
```
simonharrer: Maybe the methods should be better put into `URLUtil`.

src/main/java/net/sf/jabref/util/Util.java
```diff
@@ -665,6 +664,66 @@ public static String getResultsWithEncoding(URLConnection source, Charset encodi
...
+        HttpURLConnection con = (HttpURLConnection) source.openConnection();
+
+        //add a default request header
+        con.setRequestMethod("POST");
+        con.setRequestProperty("User-Agent", "Mozilla/5.0 (Windows NT 5.1; rv:31.0) Gecko/20100101 Firefox/31.0");
+        con.setRequestProperty("Accept-Language", "en-US,en;q=0.5");
+
+        return net.sf.jabref.util.Util.getPostResultsWithEncoding(con, postData, encoding);
```
simonharrer: You can just call return getPostResult... as you call a method within this class.

src/main/java/net/sf/jabref/importer/fetcher/IEEEXploreFetcher.java
```diff
@@ -125,71 +95,137 @@ public JPanel getOptionsPanel() {
...
+                bibtexQueryURLStringBuf.append("&citations-format=citation-abstract");
+            } else {
+                bibtexQueryURLStringBuf.append("&citations-format=citation-only");
+            }
+
+            //append the remaining URL
+            bibtexQueryURLStringBuf.append(URL_BIBTEX_END);
+
+            //fetch the raw Bibtex results from IEEEXplore and parse
+            URL bibtexURL = new URL(bibtexQueryURLStringBuf.toString());
```
simonharrer: This method is really large. Can you split it up a little bit? May be creating a method to compute the bibtexQueryURL can help.

src/main/java/net/sf/jabref/importer/fetcher/IEEEXploreFetcher.java
jeff-kerr: It was giving strange results I thought it might not be required any longer. But today I dug deep for some records with Jr. Sr. in the name and confirmed it is still an issue.  I just added new code for proper parsing for this Jr. Sr. re-ordering and managed to find a few samples to confirm the new code is working.

src/main/java/net/sf/jabref/importer/fetcher/IEEEXploreFetcher.java
jeff-kerr: This was left over from previous code, but fixed!

src/main/java/net/sf/jabref/importer/fetcher/IEEEXploreFetcher.java
jeff-kerr: Moved the code to the Util method since I suspect nearly all requests will want it.

src/main/java/net/sf/jabref/importer/fetcher/IEEEXploreFetcher.java
jeff-kerr: Done!

src/main/java/net/sf/jabref/util/Util.java
jeff-kerr: I left it in Util for now, since this is where I saw all the other URL accesses.  But I suppose future work would be to move all these to URLUtil?

src/main/java/net/sf/jabref/util/Util.java
jeff-kerr: Great!

src/main/java/net/sf/jabref/util/Util.java
jeff-kerr: Thanks, this was copied from the existing methods as well. Fixed.

src/main/java/net/sf/jabref/importer/fetcher/IEEEXploreFetcher.java
jeff-kerr: Done!

jeff-kerr:
Thanks for the comments.  I'll try to add some tests when I have more time, and I'll have a look at `GoogleScholarTest`. I don't see an existing test for the basic `EntryFetcher`, or anything that implements it, unless I missed it...

src/main/java/net/sf/jabref/importer/fetcher/IEEEXploreFetcher.java
```diff
@@ -18,15 +18,16 @@
 import java.awt.BorderLayout;
 
 import java.io.IOException;
-
+import java.lang.reflect.Array;
```
simonharrer: I am pretty sure you do not need the reflection API in your code. Maybe you imported the wrong class?

src/main/java/net/sf/jabref/importer/fetcher/IEEEXploreFetcher.java
```diff
@@ -53,65 +56,23 @@
 public class IEEEXploreFetcher implements EntryFetcher {
 
     private static final Log LOGGER = LogFactory.getLog(IEEEXploreFetcher.class);
+    private static final String URL_SEARCH = "http://ieeexplore.ieee.org/rest/search?reload=true";
+    private static final String URL_BIBTEX_START = "http://ieeexplore.ieee.org/xpl/downloadCitations?reload=true&recordIds=";
+    private static final String URL_BIBTEX_END = "&download-format=download-bibtex&x=0&y=0";
+    private static final String DIALOG_TITLE = Localization.lang("Search %0", "IEEEXplore");
+    private static final int MAX_FETCH = 100;
 
     final CaseKeeperList caseKeeperList = new CaseKeeperList();
```
simonharrer: Hm, why is this list not private?

src/main/java/net/sf/jabref/importer/fetcher/IEEEXploreFetcher.java
```diff
@@ -377,6 +407,8 @@ private BibtexEntry cleanup(BibtexEntry entry) {
                 fullName = fullName.replaceAll(" on", " ").replace("  ", " ");
             }
 
+            Pattern publicationPattern = Pattern.compile("(.*), \\d*\\.*\\s?(.*)");
```
simonharrer: Pattern should be private final static fields because it is very expensive to compile them. And in this case, it would have to be compiled for each fetch.

src/main/java/net/sf/jabref/util/Util.java
```diff
@@ -665,6 +664,63 @@ public static String getResultsWithEncoding(URLConnection source, Charset encodi
...
+        source.setRequestMethod("POST");
+        source.setRequestProperty("User-Agent", "Mozilla/5.0 (Windows NT 5.1; rv:31.0) Gecko/20100101 Firefox/31.0");
+        source.setRequestProperty("Accept-Language", "en-US,en;q=0.5");
+
+        // Send post request
+        source.setDoOutput(true);
+        try (DataOutputStream wr = new DataOutputStream(source.getOutputStream());) {
+            wr.writeBytes(postData);
+            wr.flush();
+            wr.close();
```
simonharrer: you do not need to call flush and close, as try-with-resources will call close which in turn will also call flush.

src/main/java/net/sf/jabref/util/Util.java
```diff
@@ -665,6 +664,63 @@ public static String getResultsWithEncoding(URLConnection source, Charset encodi
...
+        }
+
+        int responseCode = source.getResponseCode();
+
+        String inputLine;
+        StringBuffer response = new StringBuffer();
+
+        if (encoding != null) {
```
simonharrer: Allowing that a parameter of a method can be null is not our preferred way, as we want to avoid null as much as possible. I would vote for removing that option - as an encoding should always be used.

simonharrer:
We currently have very little code coverage through tests, but we are trying to improve. For the GVKFetcher, there is a test, but it works in a different way as it basically is a parser only.

src/main/java/net/sf/jabref/importer/fetcher/IEEEXploreFetcher.java
jeff-kerr: Yes, was being dumb about array size.  Turns out there is a length member.  I have coded some Java a long time ago, but clearly still a newb :)

src/main/java/net/sf/jabref/importer/fetcher/IEEEXploreFetcher.java
jeff-kerr: Not sure, existing code. But I can't find it anywhere else in the project, so I removed it.

src/main/java/net/sf/jabref/importer/fetcher/IEEEXploreFetcher.java
jeff-kerr: Ok makes sense, reverted it.  But does that mean we should do the same for all the regex's here?  There are a lot of them...

src/main/java/net/sf/jabref/util/Util.java
jeff-kerr: Great! Very useful

src/main/java/net/sf/jabref/util/Util.java
jeff-kerr: Sure, I was keeping consistency with previous code.  It seems that if you don't specify then the "system default" is used.  Not an encoding expert, but I hard coded UTF-8 for now.  If folks have a better idea let me know!

jeff-kerr:
Thanks for the review, see my comments above in the outdated diff.  Is the preferred way to respond there, or in a new note here?  I am also new to git and github.  Learning all kinds of stuff. Thanks.

simonharrer:
Awesome, yes, that way it the preferred one! Can you still have a go at providing tests? We really aim to improving code coverage here one test at a time.

jeff-kerr:
Sure, yes I plan to try that. It might take a bit of time since I need to learn tests as well.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ to [sourceforge feature requests](https://sourceforge.net/p/jabref/features/) by
 - Move Bibtex and Biblatex mode switcher to File menu
 - Display active edit mode (BibTeX or Biblatex) at window title
 - Implements #444: The search is cleared by either clicking the clear-button or by pressing ESC with having focus in the search field. 
+- IEEEXplore search now downloads the full Bibtex record instead of parsing the fields from the HTML webpage result (fixes #1146, #1267)
```
oscargus: Not sure if this is a valid comment, but it looks like you wrote `#1146` etc, which I think should be `feature 1146` (see top of the file).

koppor:
I'll try to resolve the conflicts and hope that tests will be added in a separate pull request :innocent:

Issue #314, opened 2015-11-11
Add close all/others/... in Database Tab menu/File menu


stefan-kolb:
Similar to IDEs. So I can close all open databases at once and continue working on a new or only one database..

koppor:
:+1: :sunglasses:

stefan-kolb:
Continue discussion in PR #491

PR #491, opened 2015-12-09
Add close all/others/... in Database Tab menu

stefan-kolb:
See #314


src/main/java/net/sf/jabref/gui/JabRefFrame.java
```diff
@@ -2337,4 +2282,104 @@ private int showSaveDialog(String filename) {
...
+
+    private void removeTab(BasePanel panel) {
+        panel.cleanUp();
+        AutoSaveManager.deleteAutoSaveFile(panel);
+        tabbedPane.remove(panel);
+        if (tabbedPane.getTabCount() > 0) {
+            markActiveBasePanel();
+        }
+        setWindowTitle();
+        updateEnabledState(); // FIXME: Man, this is what I call a bug that this is not called.
```
simonharrer: Nice FIXME. Any idea what this could mean here?

stefan-kolb: Nope, I forgot what it meant :smile:

simonharrer:
Changelog is missing. Otherwise :+1:

src/main/java/net/sf/jabref/gui/JabRefFrame.java
stefan-kolb: Maybe it is even fixed by calling this method and I added the FIXME because we must make sure that this is called every time we remove something, i don't know...

src/main/java/net/sf/jabref/gui/JabRefFrame.java
```diff
@@ -2337,4 +2282,104 @@ private int showSaveDialog(String filename) {
                 Localization.lang("Save before closing"), JOptionPane.YES_NO_CANCEL_OPTION,
                 JOptionPane.WARNING_MESSAGE, null, options, options[2]);
     }
+
+    private void closeTab(BasePanel panel) {
+        // TODO: this menu should be tab based not on the DragDropPopupPane
```
simonharrer: what about this TODO?

stefan-kolb: Can be removed as this doesn't work with our current impl.

Issue #495, opened 2015-12-09
[Code] Some problems with entries and entry types


tobiasdiez:
I have some problems how entries and entry types are handled at the moment. See also #469 for some background.

1. There is currently (nearly) no way to access the data stored in the entry in a meaningful way. I would like to call `entry.getAuthors()` to get an `AuthorList` containing the data in a non-stringy version. 
2. There are four different modes for a type (BibTex, BibLatex and customized; or unknown). The user could switch between these modes at any time. In the ideal case, this switch would be localized in some class. The localization would make it obsolete to change the type of every entry everywhere by hand if the user switches modes.
3. There should be a way to say that an entry has a given type (e.g. is an article) irrespective of its mode (BibTex, BibLatex, ...).

I thought a little bit about these issues but couldn't come up with a solution I liked. Hopefully somebody else has a good idea to improve the system.

lenhard:
I aggree, there is potential for building a more beatiful API here.

Regarding 1.: Yes, maybe methods like this could be offered for all required fields that are shared by all entry types. I just fear that there is no intersection in required fields for all entry types. 

Regarding 2. and 3.: I think there is a misunderstanding here. The semantics of `equals` are very narrowly defined and a `BibtexEntryType` is just not equal to a `BibLatexEntryType`. This is not a "mode", but different instances of `EntryType`. Their contents are different. What you are talking about is the name of the entry type, which can very well be equal, regardless of whether there is a Bibtex entry or a Biblatex entry underneath. To sum up: you can say that an entry has a given *type name* regardless of its `EntryType`. I mean something like:
```
myBibtexEntry.getType().getName().equals(myBibLatexEntry.getType().getName()); // returns true
```
instead of 
```
myBibtexEntry.getType().equals(myBibLatexEntry.getType()); // returns false
```
From what I understand, this is all that you want.

tobiasdiez:
@lenhard 
1. Do you think it also makes sense to define such methods if the field is not defined for every entry type? For example, `Optional<AuthorList> getAuthors()`?

[...]

Issue #167, opened 2015-09-13
Biblatex mode should be a per-file setting


tobiasdiez:
Currently, a global setting controls whatever JabRef uses biblatex mode or not. It is however usual to use biblatex files parallel with older bibtex files. Thus I would suggest that the biblatex mode can be activated just for certain files (i.e. should be an option under the database settings).

Moreover, it would be nice to recognize the file type automatically when the file is opened. That is, instead of showing the "The following entry types are unknown"-dialog when biblatex entries are present, it would be better to ask the user if JabRef should activate biblatex mode for this file.

bluebirch:
biblatex format can also be discovered based on certain fields (e. g. subtitle, journaltitle, location, etc), not only entry types.

stefan-kolb:
An auto detection per file would also be good :+1:

PR #500, opened 2015-12-10
Database type detection

stefan-kolb:
Please see `BibDatabaseTypeDetection.java`for review.

Some refactoring and a class for auto-detection of database types.

### Algorithm
- should we introduce a hard-coded field like encodings inside the bib file?
- fast-enough?
- what about false detection?

### Refactoring
- Bibtexdatabase -> BibDatabase
- BibtexEntry -> BibEntry

also see #495, #167


lenhard:
The refactorings are fine, from my point of view.

The detection is based on in-memory operations on collections, so no need to worry about speed at the moment.

If I understand the algorithm correctly, any database that contains a non-Biblatex/non-Bibtex type will be categorized as Bibtex. So, a database with one BibLatex entry and a customn entry would be detected as Bibtex. Is that really the desired result? I would rather ignore all non-Bibtex / non-Biblatex types for the decision (CustomEntry, UnkownEntry, IEEETranEntry, ...) and do the decision solely on valid types with Bibtex as the default result.

stefan-kolb:
No it should always detect as Biblatex if it finds one exclusive entry.

lenhard:
The algorithm looks good to me. I would move the detection class into the logic package instead of model, but that is your decision.

Strangely, the CircleCi test says there are still compilation errors due to the renamings. As far as I can see locally, all is fine.

PR #501, opened 2015-12-10
[WIP] Prototype for new EntryType interface

tobiasdiez:
This is a prototype for an idea to resolve the issues discussed in #337 and #495.

Right now the entry type is converted from a string representation (in the bibtex file) to an EntryType by the parser. The problem with this approach is that the parser has to know about the bibtex vs biblatex mode. The idea was to add an intermediate step:
String (in Bibtex file) -> EntryClass (by parser) -> EntryType (by some factory using information from the preferences)
 
- There is an enum `StandardEntryClass` holding all the standard types regardless of bibtex or biblatex (so article, mvbook, etc, all are there). To support also unknown entry types we have the class `NonStandardEntryClass`. So for example, parsing `@article{}` results in `class = StandardEntryClass.ARTICLE` while `@gibberish{}` yields `class = NonStandardEntryClass` with name `gibberish`. In this way the type information is stored in a relative convenient form (better then a string).

The idea of an enum implementing an interface was taken from the way java handles standard file locations. See [the oracle blog ](https://blogs.oracle.com/darcy/entry/enums_and_mixins) (or [another blog](http://blog.pengyifan.com/how-to-extend-enum-in-java/)).
- The `EntryClass` does not know about required and optional fields. These information are provided by the `EntryTypeFactory`. For example, `entryTypeFactory(prefs).getTypeFor(StandardEntryClass.ARTICLE).getRequiredFields()` gives the required fields. 

Please ignore all the "my" prefixes in the class names.

### DIFF
--- a/src/main/java/net/sf/jabref/importer/fileformat/BibtexParser.java
+++ b/src/main/java/net/sf/jabref/importer/fileformat/BibtexParser.java
@@ -28,7 +28,6 @@ import net.sf.jabref.bibtex.EntryTypes;
 import net.sf.jabref.logic.CustomEntryTypesManager;
 import net.sf.jabref.model.database.KeyCollisionException;
 import net.sf.jabref.importer.ParserResult;
-import net.sf.jabref.model.entry.IdGenerator;
 import net.sf.jabref.logic.l10n.Localization;
 import net.sf.jabref.logic.util.strings.StringUtil;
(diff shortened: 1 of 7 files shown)


tobiasdiez:
So, what are your opinions on this idea? Should I start implementing it properly or is there a better way to solve this issue?

tobiasdiez:
New and simpler solution can be found in PR #605.

PR #504, opened 2015-12-13
BiblioscapeImporterTest

obraliar:


src/test/java/net/sf/jabref/importer/fileformat/BiblioscapeImporterTest.java
```diff
@@ -0,0 +1,76 @@
...
+        for (String str : list) {
+            Throwable throwable = null;
+            try (InputStream is = BiblioscapeImporterTest.class.getResourceAsStream(str);) {
+                    Assert.assertTrue(importer.isRecognizedFormat(is));
+            } catch (Throwable var6_8) {
+                if (throwable == null) {
+                    throwable = var6_8;
+                }
+                throw throwable;
+            }
```
bruehldev: Simon Harrer: Remove catch clause.

src/test/java/net/sf/jabref/importer/fileformat/BiblioscapeImporterTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+    @Test
+    public void testIsRecognizedFormat() throws Throwable {
+        BiblioscapeImporter importer = new BiblioscapeImporter();
+        List<String> list = Arrays.asList("BiblioscapeImporterTest.txt", "CopacImporterTest1.txt",
+                "CopacImporterTest2.txt", "CopacImporterTest3.txt", "IEEEImport1.txt", "IsiImporterTest1.isi",
+                "IsiImporterTestInspec.isi", "IsiImporterTestWOS.isi", "IsiImporterTestMedline.isi",
+                "RisImporterTest1.ris");
+        for (String str : list) {
+            Throwable throwable = null;
```
bruehldev: Remove variable

src/test/java/net/sf/jabref/importer/fileformat/BiblioscapeImporterTest.java
```diff
@@ -0,0 +1,76 @@
...
+        }
+    }
+
+    @Test
+    public void testImportEntries() throws Throwable {
+        BiblioscapeImporter importer = new BiblioscapeImporter();
+        Throwable throwable = null;
+        try (InputStream is = BiblioscapeImporter.class.getResourceAsStream("BiblioscapeImporterTest.txt")) {
+            List<BibtexEntry> entries = importer.importEntries(is, new OutputPrinterToNull());
+                Assert.assertEquals(7, entries.size());
```
bruehldev: How do you assert that the right seven entries with the right values are read correctly?

src/test/java/net/sf/jabref/importer/fileformat/BiblioscapeImporterTest.java
```diff
@@ -0,0 +1,76 @@
...
+import net.sf.jabref.model.entry.BibtexEntry;
+
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+/*
+ * Biblioscape Tag File with an example:
+ * https://web.archive.org/web/20080820010408/http://www.biblioscape.com/manual_bsp/Biblioscape_Tag_File.htm
+ */
```
bruehldev: Put into appropriate place - JavaDoc.

obraliar:
Ready to merge?

src/test/java/net/sf/jabref/importer/fileformat/BiblioscapeImporterTest.java
```diff
@@ -0,0 +1,121 @@
...
+
+public class BiblioscapeImporterTest {
+
+    @Before
+    public void setUp() throws Exception {
+        if (Globals.prefs == null) {
```
tobiasdiez: If check not necessary.

src/test/java/net/sf/jabref/importer/fileformat/BiblioscapeImporterTest.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    @Test
+    public void testImportEntries() throws Throwable {
+        BiblioscapeImporter importer = new BiblioscapeImporter();
+        BibtexImporter bibtexImporter = new BibtexImporter();
+
+        try (InputStream bScapeStream = BiblioscapeImporter.class.getResourceAsStream("BiblioscapeImporterTest.txt");
+                InputStream bibStream = BiblioscapeImporter.class
+                        .getResourceAsStream("BiblioscapeImporterTest_comparefile.txt")) {
```
tobiasdiez: change the ending of the comparefile to bib

src/test/java/net/sf/jabref/importer/fileformat/BiblioscapeImporterTest.java
```diff
@@ -0,0 +1,121 @@
...
+            List<BibEntry> bScapeEntries = importer.importEntries(bScapeStream, new OutputPrinterToNull());
+            List<BibEntry> bibEntries = bibtexImporter.importEntries(bibStream, new OutputPrinterToNull());
+
+            Assert.assertEquals(bScapeEntries.size(), bibEntries.size());
```
tobiasdiez: Use the appropriate method in BibtexAssert to compare the lists

src/test/java/net/sf/jabref/importer/fileformat/BiblioscapeImporterTest.java
```diff
@@ -0,0 +1,121 @@
...
+                    }
+                }
+            }
+
+            Assert.assertEquals(c, bScapeEntries.size());
+
+        }
+
+
+        try (InputStream is = BiblioscapeImporter.class.getResourceAsStream("BiblioscapeImporterTest1.txt")) {
```
tobiasdiez: Move to new test method

src/test/java/net/sf/jabref/importer/fileformat/BiblioscapeImporterTest.java
```diff
@@ -0,0 +1,121 @@
...
+                // Extract the bibtex entry type
+                BibEntry be = entries.get(i);
+                String type = be.getField("note").split(": ")[1];
+
+                Assert.assertEquals(type, be.getType());
+
+            }
+        }
+
+        try (InputStream is = BiblioscapeImporter.class.getResourceAsStream("BiblioscapeImporterTest2.txt")) {
```
tobiasdiez: Move to new test method

src/test/java/net/sf/jabref/importer/fileformat/BiblioscapeImporterTest.java
```diff
@@ -0,0 +1,121 @@
...
+
+
+        try (InputStream is = BiblioscapeImporter.class.getResourceAsStream("BiblioscapeImporterTest1.txt")) {
+            List<BibEntry> entries = importer.importEntries(is, new OutputPrinterToNull());
+            Assert.assertEquals(17, entries.size());
+
+            for (int i = 0; i < entries.size(); i++) {
+                // Extract the bibtex entry type
+                BibEntry be = entries.get(i);
+                String type = be.getField("note").split(": ")[1];
```
tobiasdiez: Why did you decided to write the tests in this way? Is the `bibtex-typ` a standard field in a Biblioscape file?

obraliar: No, but `--NT--` (note). Therefore we use this field to carry the information about the bib-type, cause there is no appropriate field in Biblioscape.

tobiasdiez: Ok, I see. I would prefer a parametrized test here with pairs of strings like
`+--RT-- proceedings`
as input and `inproceedings` as output.
Then the test creates a proper Biblioscape entry (for example by adding ` +--AU-- Baklouti, F.
+--YP-- 1999
+--KW-- Cells; Rna; Isoforms`) and calls the parser. Finally it creates a bib entry instance and compares it against the parsed one.

src/test/java/net/sf/jabref/importer/fileformat/BiblioscapeImporterTest.java
obraliar: Done.

src/test/java/net/sf/jabref/importer/fileformat/BiblioscapeImporterTest.java
obraliar: Done.

src/test/java/net/sf/jabref/importer/fileformat/BiblioscapeImporterTest.java
obraliar: The problem about this is that the entry sequence is not unique.
So it is not guaranteed that the right entries are going to be compared.
BibtexEntryAssert compares the list like: `list1.get(i) == list2.get(i)`.
The best method we discovered is to compare the `isbn` field for all combinations (see l. 73) to get proper objects to compare.

src/test/java/net/sf/jabref/importer/fileformat/BiblioscapeImporterTest.java
obraliar: Done.

src/test/java/net/sf/jabref/importer/fileformat/BiblioscapeImporterTest.java
obraliar: I tried to implement this but while doing I realized that it will become too complicated for a test, cause there are also entries which differ from the other.
Loop based reading (for example with `BufferedReader`) would be too dangerous, cause the sequence the importer uses to save entries is not clear.
Now I made the test a little bit safer by removing the keyword `bibtex-typ:` in field `--NT--`, so we only need to query the `note` field instead of 
splitting the string. I hope it looks a little bit better.

src/test/java/net/sf/jabref/importer/fileformat/BiblioscapeImporterTest.java
obraliar: Done.

obraliar:
If this is acceptable, I'm going to merge the commits.

src/test/java/net/sf/jabref/importer/fileformat/BiblioscapeImporterTest.java
simonharrer: Can you please handle this comment as well?

src/test/java/net/sf/jabref/importer/fileformat/BiblioscapeImporterTest.java
```diff
@@ -0,0 +1,125 @@
...
+        BiblioscapeImporter importer = new BiblioscapeImporter();
+        BibtexImporter bibtexImporter = new BibtexImporter();
+
+        try (InputStream bScapeStream = BiblioscapeImporter.class.getResourceAsStream("BiblioscapeImporterTest.txt");
+                InputStream bibStream = BiblioscapeImporter.class
+                        .getResourceAsStream("BiblioscapeImporterTest_comparefile.bib")) {
+            List<BibEntry> bScapeEntries = importer.importEntries(bScapeStream, new OutputPrinterToNull());
+            List<BibEntry> bibEntries = bibtexImporter.importEntries(bibStream, new OutputPrinterToNull());
+
+            Assert.assertEquals(bScapeEntries.size(), bibEntries.size());
```
simonharrer: Have a look at the other PRs how to compare two lists of bibentries for equality.

src/test/java/net/sf/jabref/importer/fileformat/BiblioscapeImporterTest.java
simonharrer: Then you can sort it and then compare it.

src/test/java/net/sf/jabref/importer/fileformat/BiblioscapeImporterTest.java
simonharrer: I still do not get it: is the `--NT--` field used for storing the bibtex type in there? Or is this just some guess from JabRef?

obraliar: Yes, it is used to store the bibtex entry type because there is no appropriate field in Biblioscape. Once the object is parsed as bib-entry we use this field value to compare the Biblioscape type with. The Biblioscape format stores the type in `--WT--` or `--RT--` tags.

simonharrer: Ok, but then the comparison scheme of the previous method can be reused as well, as one can also create a .bib file which matches the import, simplifying the test logic even further.

obraliar: If I "hard code" the entry type check (e.g.  `Assert.assertEquals("inbook", be.getType());`), then yes. Otherwise we have to get the type from the source, because there is a mapping between the types of Biblioscape an Bibtex (e. g. `conference <->inproceedings`, `report <-> techreport`,...).

simonharrer: create a bib file which contains what you expect. You can use it similarly to the previous test. This expected should assert also the mapping logic.

tobiasdiez: Regarding "realized that it will become too complicated for a test, cause there are also entries which differ from the other.": the whole point of this test should be to test the bib type. So just normalize all entries and only differ the type. All other items should be tested separately.

obraliar:
We are working on it...

src/test/java/net/sf/jabref/importer/fileformat/BiblioscapeImporterTest.java
```diff
@@ -98,19 +99,19 @@ public void testImportEntries() throws Throwable {
...
+        BibtexImporter bibtexImporter = new BibtexImporter();
+
+        try (InputStream biblioStream = BiblioscapeImporter.class.getResourceAsStream("BiblioscapeImporterTest1.txt");
+                InputStream bibStream = BiblioscapeImporter.class
+                        .getResourceAsStream("BiblioscapeImporterTest1_comparefile.bib")) {
 
-            Assert.assertEquals(17, entries.size());
+            List<BibEntry> isEntries = importer.importEntries(biblioStream, new OutputPrinterToNull());
+            List<BibEntry> shouldBeEntries = bibtexImporter.importEntries(bibStream, new OutputPrinterToNull());
+            Assert.assertEquals(shouldBeEntries.size(), isEntries.size());
```
simonharrer: not required anymore.

obraliar:
Something went wrong. I'm going to reset the branch to current master and to apply changes.

src/test/java/net/sf/jabref/importer/fileformat/BiblioscapeImporterTest.java
obraliar: Done.

src/test/java/net/sf/jabref/importer/fileformat/BiblioscapeImporterTest.java
obraliar: Done. See testImportEntries3().

src/test/java/net/sf/jabref/importer/fileformat/BiblioscapeImporterTest.java
```diff
@@ -0,0 +1,156 @@
...
+            for (int i = 0; i < isEntries.size(); i++) {
+                for (int j = 0; j < shouldBeEntries.size(); j++) {
+                    BibEntry isEntry = isEntries.get(i);
+                    BibEntry shouldBeEntry = shouldBeEntries.get(j);
+                    if (isEntry.getField("note").equals(shouldBeEntry.getField("note"))) {
+                        BibtexEntryAssert.assertEquals(shouldBeEntry, isEntry);
+                        c++;
+                    }
+                }
+            }
```
obraliar: @simonharrer I don't know why but the `BibtexImporter` does not return the Bibentries from the `List` in same order as they are saved in origin bib file. But `BiblioscapeImporter` does. Therefore I inserted a sort of ID for each entry, so that I can compare correctly (line 144). To compare  them manually will blow up the method, cause there are at least 10 entries.

tobiasdiez: #982 is now merged in so that a simple `BibtexEntryAssert.assertEquals` should be enough. 
Next time something didn't work as you would have expected, then just create an issue. The whole point of writing tests is to find and fix strange behavior.

obraliar:
Now this test should also be ready. Can you check it please?

tobiasdiez:
The `*testFiles` file looks good. Regarding the typ conversion test, I propose the following modifications also using parametrized tests (the code might not compile):
````java
@Parameter(value = 0)
public String biblioscapeType;

@Parameter(value = 1)
public String expectedBibType;

@Parameters
public static Collection<String> types() {
    return biblioscapeToBibTypeMap; // Essentially, just in the double array format as described in https://github.com/junit-team/junit/wiki/Parameterized-tests        
}

@Test
public void importConvertsToCorrectBibType() {
    String input = 
        '+--AU-- Baklouti, F.
        +--YP-- 1999
        +--KW-- Cells; Rna; Isoforms
        +--RT-- ' + biblioscapeType;

    List<Entries> entries = import above string using the procedure described in one of previous PRs (should be in your wiki)

    entry.setField("author", "Baklouti, F.");
    entry.setField("keywords", "Cells; Rna; Isoforms");
    entry.setField("year", "1999");
    entry.setType(expectedBibType);
    BibtexEntryAssert.assertEquals(Colletions.singletonOf(entry), entries);
}

````

**Comment by obraliar** [comment]
Done.

**Comment by tobiasdiez** [comment]
LGTM :+1:

**Comment by obraliar** [comment]
Commits were merged.

File: src/test/java/net/sf/jabref/importer/fileformat/BiblioscapeImporterTest.java [hunk 14]
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    @Test
+    public void testGetFormatName() {
+        BiblioscapeImporter importer = new BiblioscapeImporter();
+        Assert.assertEquals(importer.getFormatName(), "Biblioscape");
+    }
+
+    @Test
+    public void testGetCLIID() {
```
koppor: I think, this test does not depend on the parameters. Please separate parameterized tests from non-parameterized.


Maybe move the parameterized ones to `BiblioscapeImporterTestTypes.java`.

src/test/java/net/sf/jabref/importer/fileformat/BiblioscapeImporterTestFiles.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @Before
+    public void setUp() throws Exception {
+        Globals.prefs = JabRefPreferences.getInstance();
+        bsImporter = new BiblioscapeImporter();
+    }
+
+    @Parameters(name = "{0}")
+    public static Collection<String> fileNames() {
+        return Arrays.asList(new String[] {"BiblioscapeImporterTest0a", "BiblioscapeImporterTest0b",
```
koppor: Is it possible to give the file names more speaking names. Meaning: The names should be self-explanary. For instance, for ttest `1`, you could use `BakloutiEtAl1996`. Oh, I see that all articles are from him. Maybe, you can state the real difference of the test cases using the file name.

src/test/java/net/sf/jabref/importer/fileformat/BiblioscapeImporterTest.java
obraliar: Done.

src/test/java/net/sf/jabref/importer/fileformat/BiblioscapeImporterTestFiles.java
obraliar: Done.

src/test/java/net/sf/jabref/importer/fileformat/BiblioscapeImporterTest.java
```diff
@@ -34,7 +34,7 @@ public void testGetCLIID() {
 
     @Test
     public void testImportEntriesAbortion() throws Throwable {
-        try (InputStream is = BiblioscapeImporter.class.getResourceAsStream("BiblioscapeImporterTest2.txt")) {
+        try (InputStream is = BiblioscapeImporter.class.getResourceAsStream("BiblioscapeImporterTestAbortion.txt")) {
```
koppor: "Abortion" indicates an exception or a real worst case. (Double check: [dict.cc's translation of Programmabbruch](http://www.dict.cc/deutsch-englisch/Programmabbruch.html))

Maybe "InvalidFile" is the right suffix?

koppor:
I went through the [concrete coverage](https://codecov.io/github/JabRef/jabref/src%2Fmain%2Fjava%2Fnet%2Fsf%2Fjabref%2Fimporter%2Ffileformat/BiblioscapeImporter.java?ref=dba4eb1133317e77cb0fa02e6be7835a4073ef1e). Do you also test for types not covered at [the if-cascade](https://codecov.io/github/JabRef/jabref/src%2Fmain%2Fjava%2Fnet%2Fsf%2Fjabref%2Fimporter%2Ffileformat/BiblioscapeImporter.java?ref=dba4eb1133317e77cb0fa02e6be7835a4073ef1e#l-213)? If yes, I cannot explain, why the [last else if](https://codecov.io/github/JabRef/jabref/src%2Fmain%2Fjava%2Fnet%2Fsf%2Fjabref%2Fimporter%2Ffileformat/BiblioscapeImporter.java?ref=dba4eb1133317e77cb0fa02e6be7835a4073ef1e#l-230) is only partially covered.

Why are statements not covered in the [inbook case](https://codecov.io/github/JabRef/jabref/src%2Fmain%2Fjava%2Fnet%2Fsf%2Fjabref%2Fimporter%2Ffileformat/BiblioscapeImporter.java?ref=dba4eb1133317e77cb0fa02e6be7835a4073ef1e#l-244)?

obraliar:
I just changed some fields (types) in test files which were responsible for lower coverage. Now it should be OK.

src/test/resources/net/sf/jabref/importer/fileformat/BiblioscapeImporterTestCorrupt.txt
```diff
@@ -1,5 +1,5 @@
 
-teststring
```
koppor: Does it really have to be `te`? Couldn't you use `invalidStringMakingTheImporterFail`?

obraliar: the length should be less than 7 to reach the max. possible coverage.

src/test/resources/net/sf/jabref/importer/fileformat/BiblioscapeImporterTestUnknownFields.bib
```diff
@@ -1,18 +1,15 @@
...
+  comment = {Quaternary Title: Clinical medicine.},
   edition = {Second Edition},
   isbn = {0006-4971},
-  journal = {Blood},
   keywords = {Membrane skeletal protein-4.1; Insertion deletion mutations; Expression; Spectrin; Cells; Rna; Isoforms},
   language = {English},
-  note = {article},
   number = {9},
-  pages = {3934--3941},
+  pages = {3934},
```
koppor: The thing is: If you test with real data, you should not manipulate it to contain wrong values. Why not creating a really different test file with artificial, but readable test data. For instance: `booktitle = {The Booktitle}`, `author = {Firstname Lastname}`, `pages = {2}`, etc?

src/test/resources/net/sf/jabref/importer/fileformat/BiblioscapeImporterTestStandard.bib
```diff
@@ -1,14 +1,13 @@
...
-comment = {Quaternary Title: Chemistry.},
-isbn = {0022-3263},
-journal = {Journal of Organic Chemistry},
-language = {English},
-number = {11},
-pages = {3749--3752},
-title = {Electronically Mediated Selectivity IN Ring Opening OF 1-azirines - THE 3-x Mode - Convenient Route To 3-oxazolines},
-volume = {61},
-year = {1996}
+@misc{,
```
koppor: Maybe the filename suffix "Standard" is bad. When reading the test, I might think, that the importer always imports to `@misc` type. Why not calling it `UnkownType`, because the test is for an unknown type now.

koppor:
The goal is **not** to produce a minimal set of tests to get 100% coverage. What happens if one changes the importer code? Then, the tests might cover more if there are more tests.

Therefore, I don't understand, why you changed tests in 42f02862bb23bb0c2d8e instead of simply adding new tests aiming to cover the missed branches.

obraliar:
I didn't add new files, because the existing files are testing some code areas multiple times. Therefore I made sure that the changes would not affect the previous tests. Should I undo the changes and create new files anyway?

koppor:
That's what I meant with "minimal set of tests". - Yes, please undo the change of the tests and add new ones.

Furthermore, please try that your tests are not in contrast to existing data. For instance, stating the wrong page range for an existing paper.

obraliar:
Done.

obraliar:
Now ready? Should I squash the commits?

src/test/resources/net/sf/jabref/importer/fileformat/BiblioscapeImporterTestUnknownType.bib
```diff
@@ -0,0 +1,6 @@
+@misc{,
+  author = {Name},
+  booktitle = {Test Title},
+  pages = {--11},
```
matthiasgeiger: Why "--"?

obraliar: See `BiblioscapeImporterTestUnknownType.txt`. The tag `--PE--` tag means "page end". Cause there is no start page tag (`--PS--`) the pages are interpreted as `--11` in bibtex. This was the purpose of this test file.

matthiasgeiger: Thanks for the clarification!

I thought the test was created to "test unknown types" :wink:

But it's okay imho that you test this here, too.

matthiasgeiger:
Yes, Please squash the commits and than we'll merge this in.

obraliar:
Done.

stefan-kolb:
@koppor @matthiasgeiger Merge this in if your happy with it :smile:

Issue #399, opened 2015-11-27
Window content looks messed up


Bananach:
I attached four screenshots of one and the same search field window, after different resizing and moving around of the containing window. This phenomenon is not restricted to the search field, however. I use Windows 7.
![jabref](https://cloud.githubusercontent.com/assets/16045385/11439263/4f1ceca2-950b-11e5-9079-60ae9426dcca.png)

matthiasgeiger:
Hi,
which version of JabRef and Java are you using? 

Works fine for me with 2.11.1 and 3.0dev builds (on Win10 with recent java version installed).

Bananach:
I am using Jabref 2.11.1 and the problem occurs for both the exe and jar.
I also just updated Java.

matthiasgeiger:
Can you find anything special in the Error Console (Help -> Show Error Console) (esp. "Exceptions" tab)?

Bananach:
Nov 27, 2015 1:59:49 PM org.java.plugin.registry.xml.ManifestParser <init>
INFO: got SAX parser factory - com.sun.org.apache.xerces.internal.jaxp.SAXParserFactoryImpl@77afea7d
Nov 27, 2015 1:59:49 PM org.java.plugin.registry.xml.PluginRegistryImpl configure
INFO: configured, stopOnError=false, isValidating=true
Nov 27, 2015 1:59:49 PM org.java.plugin.registry.xml.PluginRegistryImpl register
INFO: plug-in and fragment descriptors registered - 2
Nov 27, 2015 1:59:49 PM net.sf.jabref.plugin.PluginCore initialize
INFO: Found 2 plugin(s):
  - net.sf.jabref.export.misq (jar:file:/C:/Users/wolfersf/Desktop/JabRef-2.11.1.jar!/plugins/net.sf.jabref.export.misq/plugin.xml)
  - net.sf.jabref.core (jar:file:/C:/Users/wolfersf/Desktop/JabRef-2.11.1.jar!/plugins/net.sf.jabref.core/plugin.xml)

Nov 27, 2015 1:59:49 PM org.java.plugin.standard.StandardPluginManager activatePlugin
INFO: plug-in started - net.sf.jabref.core@2.0.0.4b1 (active/total: 1 of 2)
Nov 27, 2015 1:59:49 PM org.java.plugin.standard.StandardPluginManager activatePlugin
INFO: plug-in started - net.sf.jabref.export.misq@1.0.0 (active/total: 2 of 2)
[...]

Issue #296, opened 2015-11-06
RIS import error : Booktitle / Series


g10996694:
It seemed that there would be a problem when importing RIS format data.
Most likely for chapter type, generally the Booktitle tag is T2, but when the Series tag (for RIS tag :  T3) exists, JabRef would take T3 as booktitle, instead of T2.
you could test the following RIS
<pre>
TY  - CHAP
PY  - 2015
Y1  - 2015/09/10
Y2  - 2015/11/06
M1  - 0
DO  - doi:10.1201/b19107-2
UR  - http://dx.doi.org/10.1201/b19107-2
N1  - doi:10.1201/b19107-2
M3  - doi:10.1201/b19107-2
AU  - Catarina Barata
AU  - MEmre Celebi
AU  - JorgeS Marques
SP  - 1-22
TI  - Toward a Robust Analysis of Dermoscopy Images Acquired under Different 					Conditions
T2  - Dermoscopy Image Analysis
T3  - Digital Imaging and Computer Vision
SN  - 978-1-4822-5326-9
PB  - CRC Press
ER  - 
</pre>
Here, "T2" is the Booktitle, "T3" is Series, 
I'm under 2.11b3.

stefan-kolb:
@JabRef/developers Can anyone confirm this? And mark as a bug if reproducible?

obraliar:
Confirmed. In such case T3 is taken as booktitle  instead of T2. See https://github.com/JabRef/jabref/pull/505

simonharrer:
Is this fixed in #817 @obraliar ?

obraliar:
Yes.

PR #505, opened 2015-12-13
[WIP] RISImporterTest

obraliar:


koppor:
[Not fully covered](https://codecov.io/github/JabRef/jabref/src/main/java/net/sf/jabref/importer/fileformat/RisImporter.java?ref=b89d758f48d1e461ded1dcea997b1d08f71b428b). Is it possible to get this for line 95, too?

RIS is a very widely used import format. The tests should be really good here.

@obraliar and @mairdl Could you please check https://github.com/JabRef/jabref/issues/296

Could you also check https://github.com/JabRef/jabref/issues/399#issuecomment-160138158 and check whether Mendeley exports the `file` field somehow in the RIS format? Please answer at #399. If Mendeley does export the field, at a test case for it!

koppor:
@obraliar This test fails, please fix. Please also work together with @zellerdev @mairdl and @ayanai1 to answer https://github.com/JabRef/jabref/pull/310#issuecomment-168193536

obraliar:
This test fails due to the bug (see https://github.com/JabRef/jabref/issues/296). So it shouldn't be up on the test method.

tobiasdiez:
@obraliar Could you try to fix the bug? You are probably the only person since ages which looked at the RIS importer code. So you are the perfect candidate to fix this bug :)

It is usually a good idea to write a test replicating a bug (which you have already done). Then fix the bug. Here the test verifies that the bug is indeed fixed. Finally, leave the test as a regression test to prevent that the same bug is reintroduced later.

koppor:
Could you please add a test for the RIS given at http://bibtexml.sourceforge.net/examples.html#ris, too? You can reuse the `bib` files @zellerdev will use from there for #511.

koppor:
@obraliar Please also address http://tex.stackexchange.com/questions/288511/failing-to-import-into-jabref-3-0 and resolve the conflicts.

lenhard:
@obraliar Please build and integrate a test case for the RIS file available in #742

koppor:
New pull request at #817

PR #508, opened 2015-12-14
MsBibImporterTest

tschechlovdev:
the part, that is commented 
```java
         //   		NodeList rootLst = docin.getElementsByTagName("b:Sources");
        //   		if(rootLst.getLength()==0)
        //   			rootLst = docin.getElementsByTagName("Sources");
        //   		if(rootLst.getLength()==0)
        //   			return false;
        // System.out.println(docin.getDocumentElement().getTagName());
```
is covered with the part 
```java
if ((docin != null) && !docin.getDocumentElement().getTagName().contains("Sources")) {
            return false;
        }
```
because in the commented part, you check, wether there is an element "b:Sources". If not, then you check wether there is an element with sources. When you check wether there is an element,which contains "Sources", you have both in one. Also if there stands something before or after "Sources", then you go in the if part despite of what stands before or after "Sources". 
 So we deleted the commented part.


src/main/java/net/sf/jabref/importer/fileformat/MedlinePlainImporter.java
```diff
@@ -1,289 +1,279 @@
...
+
+            // create one here
+            b.setField(hm);
+            bibitems.add(b);
+
+        }
+
+        return bibitems;
+
+    }
```
simonharrer: Please check the codecov reports, as you do not have 100% code coverage in this importer. 

And also check why the git diff tries to replace all lines in this file. Probably wrong setting with line endings.

koppor:
Coverage of MSBibDatabase.java: [38,07%](https://codecov.io/github/JabRef/jabref/src/main/java/net/sf/jabref/logic/msbib/MSBibDatabase.java?ref=3fa43b7e140914285653758b620e094d6ec3bcd8), but the relevant parts of the importer seem to be covered. Coverage of MSBibEntry.java: [26.23%](https://codecov.io/github/JabRef/jabref/src/main/java/net/sf/jabref/logic/msbib/MSBibEntry.java?ref=3fa43b7e140914285653758b620e094d6ec3bcd8). `getBibtexRepresentation` from there is not fully covered and neither related methods. For me, these should be tested, too.

src/test/java/net/sf/jabref/importer/fileformat/MsBibImporterTest.java
```diff
@@ -0,0 +1,102 @@
...
+/**
+ *
+ * @author tschechlovdev, mairdl
+ *
+ */
+public class MsBibImporterTest {
+
+    @Before
+    public void setUp() throws Exception {
+        if (Globals.prefs == null) {
```
tobiasdiez: If statement is not necessary.

src/test/java/net/sf/jabref/importer/fileformat/MsBibImporterTest.java
```diff
@@ -0,0 +1,102 @@
...
+        //test with files, which have non recognized format
+        List<String> notAccepted = Arrays.asList("CopacImporterTest1.txt", "IsiImporterTest1.isi",
+                "IsiImporterTestInspec.isi", "emptyFile.xml", "IsiImporterTestWOS.isi");
+        for (String s : notAccepted) {
+            try (InputStream stream = MsBibImporter.class.getResourceAsStream(s)) {
+                Assert.assertFalse(testImporter.isRecognizedFormat(stream));
+            }
+        }
+
+        //test with recognized format files
```
tobiasdiez: Split this test in one for recognized formats and one for rejected ones.

src/test/java/net/sf/jabref/importer/fileformat/MsBibImporterTest.java
```diff
@@ -0,0 +1,102 @@
...
+        for (String s : Accepted) {
+            try (InputStream stream = MsBibImporter.class.getResourceAsStream(s)) {
+                Assert.assertTrue(testImporter.isRecognizedFormat(stream));
+            }
+        }
+
+    }
+
+    @Test
+    public final void testImportEntries() throws java.io.IOException {
```
tobiasdiez: Split test into many tests which only check one file.

src/test/java/net/sf/jabref/importer/fileformat/MsBibImporterTest.java
```diff
@@ -0,0 +1,102 @@
...
+        MsBibImporter testImporter = new MsBibImporter();
+
+        //test with the normal xml File, that was exported by Jabref
+        List<BibEntry> entries = testImporter.importEntries(
+                MsBibImporterTest.class.getResourceAsStream("MsBibImporterTest1.xml"), new OutputPrinterToNull());
+        BibEntry entry = entries.get(0);
+        Assert.assertEquals(12, entry.getFieldNames().size());
```
tobiasdiez: You should check that the values of all these 12 fields are correct.

src/test/java/net/sf/jabref/importer/fileformat/MsBibImporterTest.java
```diff
@@ -0,0 +1,102 @@
...
+        Assert.assertEquals(1, entries.size());
+
+        //Test with xml file, that has only one child named source
+        entries = testImporter.importEntries(MsBibImporterTest.class.getResourceAsStream("MsBibImporterTest.xml"),
```
tobiasdiez: Better use a string here instead of a file. (Have a look at the other pull request how to do this)

src/test/java/net/sf/jabref/importer/fileformat/MsBibImporterTest.java
```diff
@@ -0,0 +1,102 @@
...
+                new OutputPrinterToNull());
+        Assert.assertEquals(Collections.emptyList(), entries);
+
+        //Test with non recognized format file
+        try {
+            entries = testImporter.importEntries(MsBibImporterTest.class.getResourceAsStream("CopacImporterTest1.txt"),
```
tobiasdiez: Have a look at the `expected` rule of JUnit.

src/test/java/net/sf/jabref/importer/fileformat/MsBibImporterTest.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    @Test
+    public final void testImportSingleEntry() throws Exception {
+        MsBibImporter testImporter = new MsBibImporter();
+        try (InputStream in = MsBibImporter.class.getResourceAsStream("MsBibImporterTest1.xml");
+                InputStream bibIn = MsBibImporter.class.getResourceAsStream("MsBibImporterTest1.bib")) {
+            List<BibEntry> entries = testImporter.importEntries(in, new OutputPrinterToNull());
+            Assert.assertNotNull(entries);
+            BibtexEntryAssert.assertEquals(bibIn, entries.get(0));
```
tobiasdiez: Check that only one entry is in the list of entries.

tobiasdiez:
Could you pls split `testImportEntries`? At least move the empty xml file to an separate test. 
Otherwise: :+1:

tobiasdiez:
Looks good. 
However, there a few information which apparently get lost upon import (for example, `issue` and `issn`). Could you add these mappings pls?

tschechlovdev:
It's not, like the mappings don't exists. issue is shown as number in the bibfile. I don't know why it is represented as number, but i can change the name to issue. 
The mapping is for the standardNumber also exists in 
```
    private void parseSingleStandardNumber(String type, String bibtype, String standardNum, HashMap<String, String> map) {
        // tested using http://www.javaregex.com/test.html
        Pattern pattern = Pattern.compile(':' + type + ":(.[^:]+)");
        Matcher matcher = pattern.matcher(standardNum);
        if (matcher.matches()) {
            map.put(bibtype, matcher.group(1));
        }
    }

    private void parseStandardNumber(String standardNum, HashMap<String, String> map) {
        if (standardNumber == null) {
            return;
        }
        parseSingleStandardNumber("ISBN", "isbn", standardNum, map); /* SM: 2010.10: lower case */
        parseSingleStandardNumber("ISSN", "issn", standardNum, map); /* SM: 2010.10: lower case */
        parseSingleStandardNumber("LCCN", "lccn", standardNum, map); /* SM: 2010.10: lower case */
        parseSingleStandardNumber("MRN", "mrnumber", standardNum, map);
        /* SM: 2010.10 begin DOI support */
        parseSingleStandardNumber("DOI", "doi", standardNum, map);
        /* SM: 2010.10 end DOI support */
    }
```

the problem is, that you have to write  :ISBN:1234-5678. So i didn't know, that I have to put a : before ISBN

tschechlovdev:
Is there some other mapping that I should add or something that I should change? Or ist the test so far ok ?

koppor:
The "regenerate eclipse settings" confuse me. Could you do a single commit as described at https://github.com/JabRef/jabref/wiki/Tools#rebase-everything-as-one-commit-on-master?

koppor:
Meanwhile, we learned about parameterized tests. Please split up your test class into two. One for iterating on testfiles and checking matching of the imported `xml` and the `bib `(see https://github.com/JabRef/jabref/blob/master/src/test/java/net/sf/jabref/importer/fileformat/PdfContentImporterTestFiles.java) and one for everything else (see https://github.com/JabRef/jabref/blob/master/src/test/java/net/sf/jabref/importer/fileformat/PdfContentImporterTest.java).

There, you also see the usage of `net.sf.jabref.bibtex.BibtexEntryAssert.assertEquals(Class<? extends Object>, String, List<BibEntry>)` (which is an improvement of the long-available `net.sf.jabref.bibtex.BibtexEntryAssert.assertEquals(Class<? extends Object>, String, BibEntry)`)

As a final step, the `...Files` test cases could be generalized: The only difference are 
 - the formatter
 - the filenames to test
 - the extension of the input test file

tschechlovdev:
I splitted the test class into two classes and tried to generalize the ...Files

src/test/resources/net/sf/jabref/importer/fileformat/MsBibImporterTest2.bib
```diff
@@ -0,0 +1,45 @@
+% This file was created with JabRef 2.10.
```
koppor: Where does this bib come from?

It is shown as new, but the header says JabRef 2.10. Where id you have it from? And where is the bibliography data from?

I thought, you were creating these test files for yourself.

src/test/java/net/sf/jabref/importer/fileformat/MsBibImporterTest.java
```diff
@@ -0,0 +1,64 @@
...
+import org.junit.Test;
+
+import net.sf.jabref.Globals;
+import net.sf.jabref.JabRefPreferences;
+import net.sf.jabref.importer.OutputPrinterToNull;
+import net.sf.jabref.model.entry.BibEntry;
+
+/**
+ *
+ * @author tschechlovdev, mairdl
```
koppor: The team decided not to put `@author` tags: `git blame` gives the authors even though [git does not directly support multiple authors](http://stackoverflow.com/q/7442112/873282).

src/main/java/net/sf/jabref/importer/fileformat/MsBibImporter.java
```diff
@@ -1,3 +1,4 @@
+
```
koppor: Why did you insert an empty line?

src/main/java/net/sf/jabref/importer/fileformat/MsBibImporter.java
```diff
@@ -51,9 +52,7 @@ public boolean isRecognizedFormat(InputStream in) throws IOException {
          */
         Document docin;
         try {
-            DocumentBuilder dbuild = DocumentBuilderFactory.
-                    newInstance().
-                    newDocumentBuilder();
+            DocumentBuilder dbuild = DocumentBuilderFactory.newInstance().newDocumentBuilder();
```
koppor: You should not reformat code if you do not change anything else in the class. That leads to `git blame` outputting something wrong.

koppor:
The refactoring of the tests is good and the commits are easy to review. Thank you for that.

However, I don't see a generalization of the tests. I expected an abstract class being the parent class. This class has a constructor getting passed: the formatter, the filenames to test, and the extension of the input test file.

The importer code is pretty simple, because the whole logic is in MSBibDatabase.java, covered [33,87%](https://codecov.io/github/JabRef/jabref/src/main/java/net/sf/jabref/logic/msbib/MSBibDatabase.java?ref=aacca4a894c069f4efd8109b1ad486fdf80ae7d0) and MSBibEntry.java, covered [42.09%](https://codecov.io/github/JabRef/jabref/src/main/java/net/sf/jabref/logic/msbib/MSBibEntry.java?ref=aacca4a894c069f4efd8109b1ad486fdf80ae7d0).

When I said "neither related methods. For me, these should be tested, too.", I meant for instance `mapMSBibToBibtexType`, which [is nearly not covered](https://codecov.io/github/JabRef/jabref/src/main/java/net/sf/jabref/logic/msbib/MSBibEntry.java?ref=aacca4a894c069f4efd8109b1ad486fdf80ae7d0#l-909).

Since your project is now in the "production" phase and not in the warmup any more, I expect the numbers of both MSBibDatabase.java and MSBibEntry.java to raise to nearly 100%.

tschechlovdev:
I tried to increase the test coverage, but the problem is, that most of the pars in the MsBibDatabse and the MsBibEntry only used in the class MSBibExportFormat. So they're not used when importing a msbib file. Is it ok to test them, when testing the MSBibExportFormat?

tobiasdiez:
Yes of course you can create tests for the export and thereby test MsBibDatabse and -Entry. But please create a new test class for the exporter and don't mix it up with the tests for the importer.

tschechlovdev:
It would fit into our plan, so I will work on the exporter test and open a new PR for the tests of the exporter.

stefan-kolb:
@tobiasdiez @koppor Are you fine with merging this in?

tobiasdiez:
Yes looks good to me.

PR #509, opened 2015-12-14
[WIP] Created MedlineImporterTest

bruehldev:
used an example xml file from http://www.nlm.nih.gov
test methods not implementet yet


simonharrer:
Please open WIP pull requests only if you need review during the creation of the PR to let it guide your implementation. The current state of this PR I cannot review because it basically contains no real value yet.

PR #511, opened 2015-12-14
BibTeXMLImporterTest

zellerdev:


koppor:
Why do you use `txt` as file extension? Shouldn't it be `xml`? Is BibTeXXML not an XML format?

koppor:
Coverage of the related BibTeXMLHandler: [31,70%](https://codecov.io/github/JabRef/jabref/src/main/java/net/sf/jabref/importer/fileformat/BibTeXMLHandler.java?ref=d78831caca83c170516b0b5b7aa52dd0d0f5fd9a)

src/test/resources/net/sf/jabref/importer/fileformat/BibteXMLImporter2.xml
```diff
@@ -0,0 +1 @@
+<bibt
```
tobiasdiez: Rename file to something more meaningful (indicating that this is an incomplete file)

tobiasdiez:
At the current stage no imports are actually checked (there is not a single test of a successful import). These should be added!

src/test/java/net/sf/jabref/importer/fileformat/BibteXMLImporterTest.java
```diff
@@ -0,0 +1,131 @@
...
+            Assert.assertEquals(0, entries.size());
+        }
+    }
+
+    @Test
+    public void testImportEntries2() throws IOException {
+
+        BibteXMLImporter importer = new BibteXMLImporter();
+
+        try (InputStream is = BibteXMLImporter.class.getResourceAsStream("BibteXMLImporter2.xml")) {
```
tobiasdiez: There is no need to test files which are rejected.

src/test/java/net/sf/jabref/importer/fileformat/BibteXMLImporterTest.java
```diff
@@ -0,0 +1,185 @@
...
+            entry = entries.get(3);
+            Assert.assertEquals("book", entry.getType());
+            Assert.assertEquals("musterman", entry.getCiteKey());
+            Assert.assertEquals("musterman", entry.getField("bibtexkey"));
+            Assert.assertEquals("Musterman, M.", entry.getField("author"));
+            Assert.assertEquals("Java is good!", entry.getField("title"));
+            Assert.assertEquals("Book publisher", entry.getField("publisher"));
+            Assert.assertEquals("2016", entry.getField("year"));
+            Assert.assertEquals("Stuttgart", entry.getField("address"));
+            Assert.assertEquals("Muster, M.", entry.getField("person"));
```
tobiasdiez: What is this "person" field?

src/test/java/net/sf/jabref/importer/fileformat/BibteXMLImporterTest.java
```diff
@@ -0,0 +1,185 @@
...
+
+        BibteXMLImporter importer = new BibteXMLImporter();
+
+        try (InputStream is = BibteXMLImporter.class.getResourceAsStream("BibteXMLImporter.xml")) {
+            List<BibEntry> entries = importer.importEntries(is, new OutputPrinterToNull());
+
+            BibEntry entry = entries.get(0);
+            Assert.assertEquals(entry.getCiteKey(), "");
+            Assert.assertEquals("article", entry.getType());
+            Assert.assertEquals("Test Article", entry.getField("article"));
```
tobiasdiez: Normally an article does not have a field "article" (in similarly for the other types). What is saved in this field?

src/test/java/net/sf/jabref/importer/fileformat/BibteXMLImporterTest.java
zellerdev: I searched for all available fields in the type Book. There was this field person. So I took it to test whether JabRef can read the value of this type.
I guess a correct use is:
```xml
    <bibtex:editor>
      <bibtex:person>Mustermann, M.</bibtex:person>
    </bibtex:editor>
```
In this and the following way JabRef wouldn´t have found the value of the field "editor".
```xml
    <bibtex:editor>Muster, M.</bibtex:editor>
    <bibtex:editor>
      <bibtex:person>Mustermann, M.</bibtex:person>
    </bibtex:editor>
```
Only if I would use this format JabRef would read both values correctly.
```xml
    <bibtex:editor>
      <bibtex:person>Mustermann, M.</bibtex:person>
    </bibtex:editor>
    <bibtex:editor>Muster, M.</bibtex:editor>
```
So my goal was only to get the values of the available fields. This way there is less code and a better overview over the field values.

To answer your question. If the first code example is correct it´s a way to add persons to the group authors or editors. But when I tested it, only the last entry is returned. And I can only reach the field person. There is no "editors" field to get a list of all person who are the editors.
If I am wrong please correct me.

src/test/java/net/sf/jabref/importer/fileformat/BibteXMLImporterTest.java
zellerdev: Just now I realise that there were a few changes since I merged with master. 
Because before that I could read with getFieldNames() and getOptionalFieldNames() all available fields for any kind of entry. In this case every entry had the same fields except the one field. The field named after the type. So I used it to see whether JabRef can read the fields. Still when I use getFieldNames() I get the fields "article" and "bibtexkey". getOptionalFieldNames() was removed. So in this case it´s just to test if the value is readable.

tobiasdiez:
Could you please convert a few existing small bib-files with [BibtexXML](https://sourceforge.net/projects/bibtexml/?source=navbar) to xml and then try to read the output back in JabRef. In this way we ensure that JabRef is able to properly read the 'real' BibtexXML-format.

zellerdev:
I will update the tests as soon as possible. Right now I have issues with the converter.

zellerdev:
testIsRecognizedFormatAccept() and testImportEntriesArticle2() are ignored because JabRef only accepts xml files with the syntax <bibtex:"fieldname">. The BibteXML converter creates xml files without bibtex:. So only <"fieldname">. 
Because of that the testIsRecognizedFormatAccept() fails and the testImportEntriesArticle2() gets an error.  #898

simonharrer:
Get the codacy warnings fixed as well, please.

zellerdev:
Fixed codacy warnings and updated the test so all checks are succesful and removed conflicts with base branch.

simonharrer:
why is the test testImportEntriesArticle2 ignored? I thought when not using a prefix, the importer can handle the import?

zellerdev:
No. The Importer can only import BibTeXML files with the prefix 'bibtex:'. That is one reason why the importer needs to be reimplement. For example if I convert a .bib file with an external tool to BibTeXML there is no 'bibtex:' prefix.
Now the question why I ignore testImportEntriesArticle. Well because the importer reads all tags/fields from the BibTeXML file when using BibtexEntryAssert. Even <article>. So an empty field is imported but the .bib file has no article field. So there is a fail comparing the files.

simonharrer:
So you plan a rewrite of the import logic which passes all tests in another PR, right? Normally, PRs only adding tests which are ignored should not be merged as they do not move the codebase forward - they are like a huge TODO. From my point of view, we could make an exception in that case as it is the stupro. Hence, LGTM.

tobiasdiez:
I don't like that there are quite a few tests which are identical except that the type of the entry is changed. This a perfect case for a parametrized test case. Using parametrized tests also gets rid of the nearly identical test files. All of this was actually discussed in one of the recent importer PRs...

stefan-kolb:
@tobiasdiez WDYT?

src/main/java/net/sf/jabref/importer/fileformat/BibTeXMLImporter.java
```diff
@@ -37,9 +37,9 @@
  * check here for details on the format
  * http://www.ecst.csuchico.edu/~jacobsd/bib/formats/endnote.html
  */
-public class BibteXMLImporter extends ImportFormat {
+public class BibTeXMLImporter extends ImportFormat {
```
tobiasdiez: Isn't there a X missing, i.e Bibtex_X_ML

src/test/java/net/sf/jabref/importer/fileformat/BibTeXMLImporterTest.java
```diff
@@ -0,0 +1,83 @@
...
+    public void testGetCLIId() {
+        BibTeXMLImporter importer = new BibTeXMLImporter();
+        Assert.assertEquals("bibtexml", importer.getCLIId());
+    }
+
+    @Test
+    public void testIsRecognizedFormatReject() throws IOException {
+        BibTeXMLImporter importer = new BibTeXMLImporter();
+
+        List<String> list = testFiles.stream().filter(n -> !n.startsWith("BibTeXMLImporterTest"))
```
tobiasdiez: I would prefer if you the test files are explicitly mentioned. Moreover, there is no need to specify 20 files if they are essentially the same.

tobiasdiez:
I still think that there are too many test files. The input can easily be created from a string and also the bibentry can be created directly. So the parametrized test should not contain a fileName as a parameter but a map XMLType -> BibTeXType.
But on second thought, such a map would just replicate the map present in the importer file. @koppor do you think it is worth to test each conversation to have a high test coverage? In my opinion, one conversation as a test would suffice.

src/main/java/net/sf/jabref/importer/fileformat/BibTeXMLImporter.java
koppor: The authors also name it that strange :see_no_evil: - see http://bibtexml.sourceforge.net/ and https://github.com/Zearin/BibTeXML

koppor:
I really like the test files. They could be used by other importers, too. Reason: For non-programmers this is more readable and one can really inspect whether the importing is reasonable. One does not have to read the code.

src/test/java/net/sf/jabref/importer/fileformat/BibTeXMLImporterTest.java
koppor: I like the idea to ensure that all newly added test files are automatically tested that they are NOT recognized.

tobiasdiez: Ok, now I understand the test :). Then it is a good idea!

tobiasdiez:
But do we need to test the conversation of each type? I would create a test that the conversation works in principle and then just have a Map<String, String> in the importer class. In my opinion there is no need to replicate this information as tests (probably you also get a high coverage with the map approach :) )

src/test/java/net/sf/jabref/importer/fileformat/BibTeXMLImporterTestFiles.java
```diff
@@ -0,0 +1,76 @@
...
+
+    @Ignore("Because the importer needs to be reimplemented")
+    @Test
+    public void testIsRecognizedFormat() throws IOException {
+        try (InputStream stream = BibTeXMLImporterTest.class.getResourceAsStream(fileName)) {
+            Assert.assertTrue(bibtexmlImporter.isRecognizedFormat(stream));
+        }
+    }
+
+    @Ignore("Because the importer needs to be reimplemented")
```
koppor: Is it really true that all test files are not working? Isn't there a single file working? All files carry `bibtex:` as prefix. Please elaborate more why the tests are ignored. Or, better, add working files only and keep the non-working files at your fork. We should really focus to keep the JabRef codebase clean. It might happen that the results of the discussion at #938 reveals, that we will drop BibTeXML support.

koppor:
I like that each type is tested. What happens if the implementation is changed? For instance, if XSLT is used for conversion. The aim of the plethora of test cases is to cover as much usecases as possible and not as much code as possible :sun_with_face:

koppor:
Coverage [96,87%](https://codecov.io/github/JabRef/jabref/src/main/java/net/sf/jabref/importer/fileformat/BibTeXMLHandler.java?ref=9c000dded0bc2406a9ba5afe5d833a323837ee63) and [75.86%](https://codecov.io/github/JabRef/jabref/src/main/java/net/sf/jabref/importer/fileformat/BibTeXMLImporter.java?ref=9c000dded0bc2406a9ba5afe5d833a323837ee63). Not more achievable.

LGTM.

src/main/java/net/sf/jabref/importer/fileformat/BibTeXMLHandler.java
```diff
@@ -32,17 +31,17 @@
...
     private BibEntry b; // the entry being read
 
     // XML parsing stuff
     private String currentChars;
 
+
     public List<BibEntry> getItems() {
-        if (bibitems == null) {
-            return Collections.emptyList();
-        }
```
simonharrer: Why is this deleted? We want to make sure that getItems can NEVER return null. Alternatively, you could set an initial value to bibitems directly when it is declared.

src/test/java/net/sf/jabref/importer/fileformat/BibTeXMLImporterTestFiles.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    @Test
+    public void testImportEntries() throws IOException {
+
+        try (InputStream bitexmlStream = BibTeXMLImporterTest.class.getResourceAsStream(fileName)) {
+
+            List<BibEntry> bibtexmlEntries = bibtexmlImporter.importEntries(bitexmlStream, new OutputPrinterToNull());
+
+            fileName = fileName.replace(".xml", ".bib");
```
simonharrer: Please do not reuse a variable name. Just create a new one.

src/test/java/net/sf/jabref/importer/fileformat/BibTeXMLImporterTestFiles.java
```diff
@@ -0,0 +1,79 @@
...
+
```
simonharrer: There are a lot of empty lines in this method. Try to use the empty lines to separate important things from each other. For instance, place one empty line between the action under test (the import in this case) and the assertions (the creation of the filename and the actual assertion in that case). I would remove the other empty lines as they make it harder to understand the code.

src/test/java/net/sf/jabref/importer/fileformat/BibTeXMLImporterTestFiles.java
```diff
@@ -0,0 +1,79 @@
...
+    @Before
+    public void setUp() {
+        Globals.prefs = JabRefPreferences.getInstance();
+        bibtexmlImporter = new BibTeXMLImporter();
+    }
+
+    @Parameters(name = "{0}")
+    public static Collection<String> fileNames() {
+        List<String> files = new ArrayList<>();
+        File d = new File(System.getProperty("user.dir") + "/src/test/resources/net/sf/jabref/importer/fileformat");
```
simonharrer: Use NIO instead. The property is not required here.

src/main/java/net/sf/jabref/importer/fileformat/BibTeXMLHandler.java
koppor: It ensures that a user will never not follow the protocol. The protocol of the usage states that first the entries have to be parsed, then the result retrieved. However, this is not explicitly documented. Should we create another exception "DetectedWrongUsageOfTheAPI"?

Initializing the variable before is a work around, then at EndDocument, this variable has to be reset to ensure that parsing of a second document does not lead to entries added to the result of the first run. - I know, that this is an edge case, but also not following the protocol. If we treat the first wrong usage (getResults before parsing), the we also have to tret the second wrong usage (using the parser twice), don't we?

simonharrer: During startDocument a new list will be created. Hence, it can be used twice without any issues.

koppor: Then, we add a ridiculous test case for covering these three lines of code? Currently, these lines are never used.

koppor: Personal discussion: quick fix: keep the if check; initialization at class variable; reset at EndDocument. - Test case is obsolete, but you can add it to achieve 100% coverage.

A good redesign is a private static class or an anonym class.

tobiasdiez:
I still prefer that the type conversation is implemented analogously to BiblioscapeImporterTestTypes

simonharrer:
One last remark: try not to use forEach, this is the worst method in using streams as it produces side effects.

PR #513, opened 2015-12-14
RepecNepImporterTest

tschechlovdev:


koppor:
It is impossible to review this PR. Yould you please do the usual `git fetch --all --prune`; `git merge upstream/master`; `git reset upstream/master`; `git gui`  + commit + force push chain?

koppor:
Coverage is [90%](https://codecov.io/github/JabRef/jabref/src/main/java/net/sf/jabref/importer/fileformat/RepecNepImporter.java?ref=2d294130b5b4773aa32358cc507698f28dbbceb2). I think, more is hard to achieve. Nevertheless, it should be tried to cover the [missed line 473](https://codecov.io/github/JabRef/jabref/src/main/java/net/sf/jabref/importer/fileformat/RepecNepImporter.java?ref=2d294130b5b4773aa32358cc507698f28dbbceb2#l-473).

Further, an IOException can be simulated using the Mockito framework. See https://github.com/zellerdev/jabref/commit/d78831caca83c170516b0b5b7aa52dd0d0f5fd9a

src/test/java/net/sf/jabref/importer/fileformat/RepecNepImporterTest.java
```diff
@@ -0,0 +1,122 @@
...
+import net.sf.jabref.JabRefPreferences;
+import net.sf.jabref.bibtex.BibtexEntryAssert;
+import net.sf.jabref.importer.OutputPrinterToNull;
+import net.sf.jabref.model.entry.BibEntry;
+
+public class RepecNepImporterTest {
+
+    @Before
+    public void setUp() throws Exception {
+        if (Globals.prefs == null) {
```
tobiasdiez: No need for the if check.

src/test/java/net/sf/jabref/importer/fileformat/RepecNepImporterTest.java
```diff
@@ -0,0 +1,122 @@
...
+        //test with non recognized format files
+        List<String> notAccepted = Arrays.asList("RepecNep1.xml", "CopacImporterTest1.txt", "RisImporterTest1.ris",
+                "CopacImporterTest2.txt", "IEEEImport1.txt");
+        for (String s : notAccepted) {
+            try (InputStream stream = RepecNepImporter.class.getResourceAsStream(s)) {
+                Assert.assertFalse(testImporter.isRecognizedFormat(stream));
+            }
+        }
+
+        //test with recognized format files
```
tobiasdiez: Split test in tests for recognize vs non-recognized format.

src/test/java/net/sf/jabref/importer/fileformat/RepecNepImporterTest.java
```diff
@@ -0,0 +1,122 @@
...
+        List<String> accepted = Arrays.asList("RepecNepImporterTest1.txt", "RepecNepImporterTest2.txt",
+                "RepecNepImporterTest3.txt");
+        for (String s : accepted) {
+            try (InputStream stream = RepecNepImporter.class.getResourceAsStream(s)) {
+                Assert.assertTrue(testImporter.isRecognizedFormat(stream));
+            }
+        }
+    }
+
+    @Test
```
tobiasdiez: Split test into multiple tests, each of which just covers on test file.
Moreover, test also that all parsed entries have the correct fields. For this you better don't import a file with 29 entries :)

src/test/java/net/sf/jabref/importer/fileformat/RepecNepImporterTest.java
```diff
@@ -0,0 +1,122 @@
...
+
+        entries = testImporter.importEntries(RepecNepImporter.class.getResourceAsStream("RepecNepImporterTest3.txt"),
+                new OutputPrinterToNull());
+        Assert.assertEquals(17, entries.size());
+
+        entries = testImporter.importEntries(RepecNepImporter.class.getResourceAsStream("RepecNepImporterTest4.txt"),
+                new OutputPrinterToNull());
+        Assert.assertEquals(29, entries.size());
+
+        try {
```
tobiasdiez: Have a look at the [expected exception rule](http://stackoverflow.com/questions/156503/how-do-you-assert-that-a-certain-exception-is-thrown-in-junit-4-tests)

src/test/java/net/sf/jabref/importer/fileformat/RepecNepImporterTest.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    @Test
+    public final void testImportSingleEntries() throws Exception {
+        RepecNepImporter testImporter = new RepecNepImporter();
+
+        //check all 4 test files and take the first two entries and compare with the bib file
+        for (int i = 1; i < 5; i++) {
+            try (InputStream in = RepecNepImporter.class.getResourceAsStream("RepecNepImporterTest" + i + ".txt");
+                    InputStream bibIn1 = RepecNepImporter.class
```
tobiasdiez: Better test the parsed entries directly instead of comparing them with a bib file.

koppor: We asked the students to use `bib` files to avoid endless `assertEquals` lines. I agree that for small tests (maybe <= 3 assertEquals), it makes sense to directly compare it. In contrast, for larger entries, I would vote to keep comparing with `bib` files. If someone starts to change the importer, the tests will ensure that nothing breaks. It won't help development, which is OK for me.

tobiasdiez: The problem with tests like this is that they can break because of three reasons:
1. The actual importer has a bug
2. The comparison logic has a bug
3. The bib-parser has a bug
Using direct assertEquals on the resulting bibentries (almost) eliminates possibility 2+3.
I think the main problem is that the entries are so big. In an idea world, you have many tests with small entries. Then if a tests fails you know exactly which feature has a bug.

On the other hand, I can see why it might be good idea to compare with the bibfile. And since the importer are relatively stable, I think it is ok to leave the tests as they are right now.

Ps: Something which worries me a little bit: while adding all these tests for the importer almost no bug where found. I find this strange.

src/test/java/net/sf/jabref/importer/fileformat/RepecNepImporterTest.java
```diff
@@ -0,0 +1,109 @@
...
+import net.sf.jabref.bibtex.BibtexEntryAssert;
+import net.sf.jabref.importer.OutputPrinterToNull;
+import net.sf.jabref.model.entry.BibEntry;
+
+public class RepecNepImporterTest {
+
+    RepecNepImporter testImporter;
+    @Before
+    public void setUp() throws Exception {
+            Globals.prefs = JabRefPreferences.getInstance();
```
tobiasdiez: There seems to be an issue with the indention.

src/test/resources/net/sf/jabref/importer/fileformat/RepecNepImporterTest1.bib
```diff
@@ -0,0 +1,12 @@
...
+  author = {Uwe Cantner and Andreas Nicklisch and Torsten Weiland},
+  institution = {University of Jena, Faculty of Economics and Max-Planck-Institute for Research into Collective Goods and Max-Planck-Institute for Economics, Strategic Interactions Group},
+  Jel                      = {D81 O31},
```
tobiasdiez: What is this `Jel`?

src/test/resources/net/sf/jabref/importer/fileformat/RepecNepImporterTest3.bib
```diff
@@ -0,0 +1,6 @@
+@TechReport{,
+  month = {7},
```
tobiasdiez: The month seems to be wrong. It should be 8, right?

src/test/resources/net/sf/jabref/importer/fileformat/RepecNepImporterTest1.bib
tschechlovdev: https://ideas.repec.org/j/
Here are top classification for JEL. The D for example means, that it is classified as Microeconomics. But I'm not sure what the meaning of the number behind is meaning.

src/test/resources/net/sf/jabref/importer/fileformat/RepecNepImporterTest3.bib
tschechlovdev: you're right. I don't know why it is imported wrong. In the other test the month was also wrong. In both it was one to low. So I've added  +1 for the month.
Now it should be correct. 
I've also added a field for the day.

tobiasdiez:
Good job :+1: 
@koppor if you are happy with [85% coverage](https://codecov.io/github/JabRef/jabref/src/main/java/net/sf/jabref/importer/fileformat/RepecNepImporter.java?ref=68d076e484cb821a3d3307bd5adaaab5cfa42aac), then this can be merged in from my point of view.

koppor:
This test is about RepecNepImporter. **No changes on other classes should be done**. Please remove the change on `MsBibImporter.java`.

koppor:
Otherwise, it's OK for me.

PR #514, opened 2015-12-15
add gradle test encoding 'utf-8'

bartsch-dev:
Added important missing Encoding information for testing with Gradle.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -240,6 +240,10 @@ compileTestJava {
     options.encoding = 'UTF-8'
 }
 
+test {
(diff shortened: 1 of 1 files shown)


simonharrer:
Tests should not depend on the default encoding of the system. I would vote for always reading only with UTF_8 set explicitly instead of relying on the default encoding.

koppor:
The proposed solution was also proposed at https://stackoverflow.com/a/71639886/873282. I don't understand how this relates to JDK18's default charset.

PR #515, opened 2015-12-15
BibtexImporter test

bartsch-dev:
Added tests for BibtexImporter, depends on PR #514 (missing Gradle test encoding, special char 'ß')

### DIFF
--- a/src/test/java/net/sf/jabref/importer/fileformat/BibtexImporterTest.java
+++ b/src/test/java/net/sf/jabref/importer/fileformat/BibtexImporterTest.java
@@ -1,0 +1,96 @@
+package net.sf.jabref.importer.fileformat;
+
+import net.sf.jabref.Globals;
+import net.sf.jabref.JabRefPreferences;
+import net.sf.jabref.importer.OutputPrinterToNull;
+import net.sf.jabref.model.entry.BibEntry;
+import org.junit.Before;
+import org.junit.Test;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.List;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+
+public class BibtexImporterTest {
+
+    private BibtexImporter importer;
+
+    @Before
+    public void setUp() throws Exception {
+        Globals.prefs = JabRefPreferences.getInstance();
+        importer = new BibtexImporter();
+    }
+
+    @Test
+    public void testIsRecognizedFormat() throws IOException {
+        try (InputStream stream = BibtexImporterTest.class.getResourceAsStream("BibtexImporter.examples.bib")) {
+            assertTrue(importer.isRecognizedFormat(stream));
+        }
+    }
+
+    @Test
+    public void testImportEntries() throws Exception {
+        try (InputStream stream = BibtexImporterTest.class.getResourceAsStream("BibtexImporter.examples.bib")) {
+            List<BibEntry> bibEntries = importer.importEntries(stream, new OutputPrinterToNull());
+
+            assertEquals(4, bibEntries.size());
+
+            BibEntry be0 = bibEntries.get(0);
+            assertEquals("Aks{\\i}n, {\\\"O}zge and T{\\\"u}rkmen, Hayati and Artok, Levent and {\\c{C}}etinkaya, " +
+                    "Bekir and Ni, Chaoying and B{\\\"u}y{\\\"u}kg{\\\"u}ng{\\\"o}r, Orhan and {\\\"O}zkal, Erhan",
+                     be0.getField("author"));
+            assertEquals("aksin", be0.getField("bibtexkey"));
+            assertEquals("2006", be0.getField("date"));
+            assertEquals("Effect of immobilization on catalytic characteristics", be0.getField("indextitle"));
+            assertEquals("#jomch#", be0.getField("journaltitle"));
+            assertEquals("13", be0.getField("number"));
+            assertEquals("3027-3036", be0.getField("pages"));
+            assertEquals("Effect of immobilization on catalytic characteristics of saturated {Pd-N}-heterocyclic " +
+                    "carbenes in {Mizoroki-Heck} reactions", be0.getField("title"));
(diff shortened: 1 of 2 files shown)


koppor:
I had a personal discussion with @bartsch-dev, that the current BibtexImporter is *not* aware of the encoding header. The reading of JabRef's encoding header is implemented in `OpenDatabaseAction` and not somewhere in the BibtexParser.

Therefore, there is *no* test of different encoding strings and it can IMHO only handle ASCII encoded files.

koppor:
I would let this go through and handle the usage of the open method of `OpenDatabaseAction` in a separate issue.

koppor:
Note that none of the files written at https://github.com/JabRef/jabref/wiki/Publicly-available-BibTeX-files are tested here.

src/test/java/net/sf/jabref/importer/fileformat/BibtexImporterTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @Rule
+    public ExpectedException thrown = ExpectedException.none();
+
+    private BibtexImporter importer;
+
+
+    @Before
+    public void setUp() throws Exception {
+        if (Globals.prefs == null) {
```
tobiasdiez: The if test can be removed.

src/test/java/net/sf/jabref/importer/fileformat/BibtexImporterTest.java
```diff
@@ -0,0 +1,70 @@
...
+        try (InputStream stream = BibtexImporter.class.getResourceAsStream("BibtexImporter.example.git.bib")) {
+            assertTrue(importer.isRecognizedFormat(stream));
+        }
+    }
+
+    @Test
+    public void testImportEntries() throws Exception {
+        try (InputStream stream = BibtexImporter.class.getResourceAsStream("BibtexImporter.example.git.bib")) {
+            List<BibEntry> entries = importer.importEntries(stream, new OutputPrinterToNull());
+
```
tobiasdiez: Add assertion that the length of `entries` is 1

src/test/java/net/sf/jabref/importer/fileformat/BibtexImporterTest.java
```diff
@@ -0,0 +1,70 @@
...
+    public void setUp() throws Exception {
+        if (Globals.prefs == null) {
+            Globals.prefs = JabRefPreferences.getInstance();
+        }
+        importer = new BibtexImporter();
+    }
+
+    @Test
+    public void testIsRecognizedFormat() throws Exception {
+        try (InputStream stream = BibtexImporter.class.getResourceAsStream("BibtexImporter.example.git.bib")) {
```
tobiasdiez: why does the filename contain a `.git.` part?

koppor: Because the content is about `git`.

src/test/java/net/sf/jabref/importer/fileformat/BibtexImporterTest.java
koppor: See also https://github.com/JabRef/jabref/wiki/Code-Howtos#preferences-in-tests

src/test/java/net/sf/jabref/importer/fileformat/BibtexImporterTest.java
```diff
@@ -0,0 +1,70 @@
...
+            assertTrue(importer.isRecognizedFormat(stream));
+        }
+    }
+
+    @Test
+    public void testImportEntries() throws Exception {
+        try (InputStream stream = BibtexImporter.class.getResourceAsStream("BibtexImporter.example.git.bib")) {
+            List<BibEntry> entries = importer.importEntries(stream, new OutputPrinterToNull());
+
+            assertEquals("Heidelberg", entries.get(0).getField("address"));
```
tobiasdiez: There are multiple calls of  `entries.get(0)`. Replace them by creating a new entry with `BibEntry entry = entries.get(0)`.

bartsch-dev:
Fixed all the mentioned issues.
The `.git` part in the file is, as already mentioned, the title of the book which i took as an example.

src/test/java/net/sf/jabref/importer/fileformat/BibtexImporterTest.java
```diff
@@ -0,0 +1,101 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+
+public class BibtexImporterTest {
+
+    private BibtexImporter importer;
+
+    @Rule
+    public ExpectedException thrown = ExpectedException.none();
```
tobiasdiez: This can be deleted.

tobiasdiez:
I think this can be merged in.

src/test/java/net/sf/jabref/importer/fileformat/BibtexImporterTest.java
koppor: @bartsch-dev The `ExpectedException` IMHO delares a variable which can be used for checking exceptions. See http://junit.org/apidocs/org/junit/rules/ExpectedException.html, http://stackoverflow.com/a/16725280/873282, and http://blog.codeleak.pl/2014/03/junit-expectedexception-rule-beyond.html.

bartsch-dev: As far as I understand the [javadoc](http://junit.org/javadoc/latest/org/junit/rules/ExpectedException.html) it just declares the standard expected Exception for all tests in that class (which usually is none). 
I will remove it if you think otherwise.

koppor: Please remove it. It only has an effect when the variable is used. See "This doesn't affect your existing tests" in the javadoc :innocent:

koppor:
As discussed in https://github.com/JabRef/jabref/issues/558#issuecomment-195430095, it is not necessary to refactor `OpenDatabaseAction` because of this importer.

PR #517, opened 2015-12-15
Fixed null issue in GoogleScholarFetcher

oscargus:
Fix so that the Google Scholar fetcher is working again after the new BibtexEntry class require nun-null values.


oscargus:
I was a bit too quick. There were more issues of this type, so I fixed them all and pushed without seeing that it was already merged.

oscargus:
I'll open a new PR (seems to be the best solution).

PR #519, opened 2015-12-15
Fixed all fields set to null

oscargus:
As entry.setField() does not accept a null argument anymore, the method entry.clearField() is used instead.


simonharrer:
are you done? ;)

oscargus:
Haha, yes, I think so. :-) (It seems like UndoableFieldChange handles clearing properly.)

Issue #479, opened 2015-12-07
Trouble importing after upgrading to Jabref 3.0


mazeratti:
Hi,
After the upgrade, I can only import .bib files. no other formats e.g. .ris, works. I am using jabref on OS X El Capitan (10.11.1). I got this error while installing: 
 an error occurred java.lang.nullpointerexception

matthiasgeiger:
Hi,
can you please paste the complete Stacktrace hier?

Either from the terminal or from the error console (Help -> Show error console).

Thanks!

mazeratti:
user.language=en
java.specification.vendor=Oracle Corporation
awt.toolkit=sun.lwawt.macosx.LWCToolkit
java.vm.info=mixed mode
java.version=1.8.0_25
java.ext.dirs=/Users/mazeratti/Library/Java/Extensions:/Library/Internet Plug-Ins/JavaAppletPlugin.plugin/Contents/Home/lib/ext:/Library/Java/Extensions:/Network/Library/Java/Extensions:/System/Library/Java/Extensions:/usr/lib/java
sun.boot.class.path=/Library/Internet Plug-Ins/JavaAppletPlugin.plugin/Contents/Home/lib/resources.jar:/Library/Internet Plug-Ins/JavaAppletPlugin.plugin/Contents/Home/lib/rt.jar:/Library/Internet Plug-Ins/JavaAppletPlugin.plugin/Contents/Home/lib/sunrsasign.jar:/Library/Internet Plug-Ins/JavaAppletPlugin.plugin/Contents/Home/lib/jsse.jar:/Library/Internet Plug-Ins/JavaAppletPlugin.plugin/Contents/Home/lib/jce.jar:/Library/Internet Plug-Ins/JavaAppletPlugin.plugin/Contents/Home/lib/charsets.jar:/Library/Internet Plug-Ins/JavaAppletPlugin.plugin/Contents/Home/lib/jfr.jar:/Library/Internet Plug-Ins/JavaAppletPlugin.plugin/Contents/Home/classes
install4j.appDir=/Volumes/jabref/JabRef Installer.app/Contents/Resources/app/
java.vendor=Oracle Corporation
file.separator=/
java.vendor.url.bug=http://bugreport.sun.com/bugreport/
sun.font.fontmanager=sun.font.CFontManager
sun.io.unicode.encoding=UnicodeBig
sun.cpu.endian=little
install4j.systemLanguage=en
[...]

PR #531, opened 2015-12-17
Import Format Reader Fix

boceckts:
The Problem is, that the stream can not be resetted without a BufferedInputStream. 
However this is necessary since we check if a import file is in the right format with
looking into the stream and searching for specific patterns.

Result of the above mentioned issue is that I always got an error message saying
"No entries found" when trying to import a file with a valid format and entries. 

This fixes the issue without changing any method interfaces.


simonharrer:
Please add a CHANGELOG entry as this fixes #479

oscargus:
Probably some line ending issue as all lines in ImportFormatReader are marked as changed and a new copy of the CHANGELOG is appended at the start of it.

PR #536, opened 2015-12-18
Changes to ParserResult warnings

oscargus:
Some restructuring of the ParserResult warnings code. Introduced a new class that is used in a number of other classes to do the same thing.

### DIFF
--- a/src/main/java/net/sf/jabref/gui/ParserResultWarningDialog.java
+++ b/src/main/java/net/sf/jabref/gui/ParserResultWarningDialog.java
@@ -15,7 +15,8 @@
 */
 package net.sf.jabref.gui;
 
-import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
 
 import javax.swing.JOptionPane;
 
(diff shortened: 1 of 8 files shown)


src/main/java/net/sf/jabref/gui/ParserResultWarningDialog.java
```diff
@@ -0,0 +1,85 @@
...
+    private final JabRefFrame jabRefFrame;
+    private final int dbCount;
+
+
+    public ParserResultWarningDialog(ParserResult pr, JabRefFrame jrf) {
+        parserResult = pr;
+        maxWarnings = Integer.MAX_VALUE;
+        jabRefFrame = jrf;
+        dbCount = -1;
+    }
```
simonharrer: use this(...) to call the other constructor

src/main/java/net/sf/jabref/gui/ParserResultWarningDialog.java
```diff
@@ -0,0 +1,85 @@
...
+
+    public ParserResultWarningDialog(ParserResult pr, JabRefFrame jrf, int maxWarns, int dataBaseCount) {
+        parserResult = pr;
+        maxWarnings = maxWarns;
+        jabRefFrame = jrf;
+        dbCount = dataBaseCount;
+    }
```
simonharrer: use Objects.requireNonNull for the two references

src/main/java/net/sf/jabref/importer/ImportMenuItem.java
```diff
@@ -154,23 +153,9 @@ public void run() {
...
-                            if (wrn.length() > 0) {
-                                wrn.deleteCharAt(wrn.length() - 1);
-                            }
-                            JOptionPane.showMessageDialog(frame, wrn.toString(),
-                                    Localization.lang("Warnings"),
-                                    JOptionPane.WARNING_MESSAGE);
-                        }
+                    if (Globals.prefs.getBoolean(JabRefPreferences.DISPLAY_KEY_WARNING_DIALOG_AT_STARTUP)) {
+                        ParserResultWarningDialog prwDialog = new ParserResultWarningDialog(pr, frame);
+                        prwDialog.show();
```
simonharrer: you can also put this in one line, as this is probably more readable. 

`new ParserResultWarningDialog(pr, frame).show()`

src/main/java/net/sf/jabref/importer/ParserResult.java
```diff
@@ -126,12 +126,8 @@ public boolean hasWarnings() {
...
     }
 
-    public String[] warnings() {
-        String[] s = new String[warnings.size()];
-        for (int i = 0; i < warnings.size(); i++) {
-            s[i] = warnings.get(i);
-        }
-        return s;
+    public ArrayList<String> warnings() {
+        return warnings;
```
simonharrer: Please use `List` instead and create a new list if possible to not let a caller change the warnings from the outside.

src/main/java/net/sf/jabref/gui/ParserResultWarningDialog.java
oscargus: Aha,  `this` was the magic that I didn't recall...

src/main/java/net/sf/jabref/gui/ParserResultWarningDialog.java
```diff
@@ -0,0 +1,85 @@
...
+
+import net.sf.jabref.importer.ParserResult;
+import net.sf.jabref.logic.l10n.Localization;
+
+public class ParserResultWarningDialog {
+
+    private final ParserResult parserResult;
+    private final int maxWarnings;
+    private final JabRefFrame jabRefFrame;
+    private final int dbCount;
```
simonharrer: what does `dbCount` mean?

oscargus: Database number (which tab to activate when showing the warning dialog).

I'll fix all things later on and comment where appropriate. Out of laptop battery now. :-)

Thanks for the feedback!

src/main/java/net/sf/jabref/importer/ImportMenuItem.java
oscargus: I changed it to a static method showing the dialog. Is this a better or worse method than the previous?

src/main/java/net/sf/jabref/gui/ParserResultWarningDialog.java
```diff
@@ -0,0 +1,91 @@
...
+     * @param jabRefFrame - the JabRefFrame
+     * @param maxWarnings - Maximum number of warnings to display
+     * @param dataBaseNumber - Database tab number to activate when showing the warning dialog
+     */
+    public static void showParserResultWarningDialog(ParserResult parserResult, JabRefFrame jabRefFrame,
+            int maxWarnings, int dataBaseNumber) {
+        Objects.requireNonNull(parserResult);
+        Objects.requireNonNull(jabRefFrame);
+        Objects.requireNonNull(maxWarnings);
+        Objects.requireNonNull(dataBaseNumber);
```
simonharrer: for the two int values, you do not need to add a requireNonNull check, as they are primitive types which can never be null.

src/main/java/net/sf/jabref/gui/ParserResultWarningDialog.java
```diff
@@ -0,0 +1,91 @@
...
+        if (parserResult.hasWarnings()) {
+            if (dataBaseNumber < 0) {
+                jabRefFrame.showBasePanelAt(dataBaseNumber);
+            }
+
+            // Generate string with warning texts
+            List<String> wrns = parserResult.warnings();
+            StringBuilder wrn = new StringBuilder();
```
simonharrer: I would use `warnings` instead of `wrns` as names should not be shortened. And instead of `wrn` I would use `dialogContent` which is in line with `dialogTitle` which is computed below.

src/main/java/net/sf/jabref/gui/ParserResultWarningDialog.java
oscargus: Makes sense. :-)

oscargus:
Fixed those things and a really stupid bug...

In general, what is the better approach for these types of warning dialogs?

* Like now, a static method showing the dialog and not returning anything? (`Dialog.showDialog()`)
* As the first approach, creating an object on which a method is directly called (still not returning anything)? (`new Dialog.show()`)

oscargus:
Will rebase...

src/main/java/net/sf/jabref/gui/ParserResultWarningDialog.java
```diff
@@ -0,0 +1,90 @@
...
+     * @param parserResult - ParserResult for the current import/open
+     * @param jabRefFrame - the JabRefFrame
+     * @param maxWarnings - Maximum number of warnings to display
+     * @param dataBaseNumber - Database tab number to activate when showing the warning dialog
+     */
+    public static void showParserResultWarningDialog(ParserResult parserResult, JabRefFrame jabRefFrame,
+            int maxWarnings, int dataBaseNumber) {
+        Objects.requireNonNull(parserResult);
+        Objects.requireNonNull(jabRefFrame);
+        if (parserResult.hasWarnings()) {
```
simonharrer: I would again use fail fast, namely, return quickly if there are no warnings in this method.

simonharrer:
I favor classes/instances over static methods. In that case, both is possible.

src/main/java/net/sf/jabref/gui/ParserResultWarningDialog.java
oscargus: Is there a difference between
``` Java
if (...) {
   return
}
Lots of code
```
and
``` Java
if (!...) {
    Lots of code
}
return
```
?

simonharrer: Semantically, there is no difference. 

But the normal control flow should be obvious to the reader and not hidden in some if branch. The issue with indention is that you always have to remember which condition has lead to this if branch, and you have to have it in mind to understand it. For the normal control flow or the most-used and typical control flow path, it should be obvious and we can put all failures on top, reducing the level of indention and putting the focus on the default control flow.

oscargus: OK, makes sense!

(The question was actually from a performance perspective, but I see you point.)

simonharrer: Performance wise, it probably makes no difference as well. But I cannot offer any facts. :)

oscargus:
May be good to go now.

PR #538, opened 2015-12-19
Extracted FieldExtra components to separate class

oscargus:


src/main/java/net/sf/jabref/gui/entryeditor/FieldExtraComponents.java
```diff
@@ -0,0 +1,258 @@
...
+        }
+
+        return new ExternalFilePanel(frame, panel.metaData(), entryEditor, fieldEditor.getFieldName(), off,
+                fieldEditor);
+    }
+
+    public static JComponent getExternalExtraComponent(FieldEditor fieldEditor, EntryEditor entryEditor) {
+        ((JComponent) fieldEditor).addMouseListener(entryEditor.new ExternalViewerListener());
+
+        return null;
```
simonharrer: Hm, I do not like that null is returned in some cases. Could this just be converted to returning an Optional? Would also apply to the getExtra method.

oscargus:
Optional should work. Will read up on it.

simonharrer:
:+1:

PR #540, opened 2015-12-19
Some minor Coverity warning fixes

oscargus:


src/main/java/net/sf/jabref/bibtex/CamelCaser.java
```diff
@@ -75,7 +75,7 @@
      * @return
      */
     public static String toCamelCase(String fieldName) {
-        if (fieldName == null || fieldName == "") {
+        if ((fieldName == null) || "".equals(fieldName)) {
```
tobiasdiez: or replace by `fieldName.isEmpty()`

simonharrer: Good point!

oscargus:
I replace the `"".equals(text)` with `text.isEmpty()` in "all" files where applicable, except for (most of) the Importers where it requires a bit of checking to see if the String is set to empty in the first place. (There are 60 occurrences in those files.)

PR #541, opened 2015-12-20
More use of Optional and some other fixes

oscargus:
* Used `Optional` instead of `return null`. Sometimes it makes much sense, sometimes less.
* Replaced checking zero or non-zero `String` lengths with `.isEmpty()`
* Changed the default case in some switch-statements to avoid `return null` (as that case should never happen anyway)
* Avoided an exception from happening when importing items without author and generating a BibTexKey which is based on the author name.

### DIFF
--- a/src/main/java/net/sf/jabref/JabRef.java
+++ b/src/main/java/net/sf/jabref/JabRef.java
@@ -247,8 +247,7 @@ public class JabRef {
                     if (initialStartup) {
                         toImport.add(aLeftOver);
                     } else {
-                        Optional<ParserResult> res = JabRef.importToOpenBase(aLeftOver);
-                        loaded.add(res.orElse(ParserResult.INVALID_FORMAT));
+                        loaded.add(JabRef.importToOpenBase(aLeftOver).orElse(ParserResult.INVALID_FORMAT));
                     }
                 } else if (pr != ParserResult.FILE_LOCKED) {
                     loaded.add(pr);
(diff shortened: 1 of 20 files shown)


src/main/java/net/sf/jabref/JabRef.java
```diff
@@ -898,11 +889,11 @@ private static ParserResult importFile(String argument) {
...
-    private static ParserResult importToOpenBase(String argument) {
-        ParserResult result = JabRef.importFile(argument);
+    private static Optional<ParserResult> importToOpenBase(String argument) {
+        Optional<ParserResult> result = JabRef.importFile(argument);
 
-        if (result != null) {
-            result.setToOpenTab(true);
+        if (result.isPresent()) {
+            result.get().setToOpenTab(true);
         }
```
simonharrer: in that case, use the following:

```
optional.ifPresent(value -> value.setToOpenTab(true));
```

The function is only executed if the optional is present.

src/main/java/net/sf/jabref/external/SynchronizeFileField.java
```diff
@@ -123,16 +123,16 @@ public void run() {
...
+            
                 Object old = sel[i].getField(fieldName);
                 FileListTableModel tableModel = new FileListTableModel();
                 if (old != null)
                     tableModel.setContent((String)old);
                 Thread t = FileListEditor.autoSetLinks(sel[i], tableModel, null, null);
-
+            
                 if (!tableModel.getStringRepresentation().equals(old)) {
                     String toSet = tableModel.getStringRepresentation();
```
simonharrer: Just delete commented out code.

src/main/java/net/sf/jabref/logic/labelPattern/LabelPatternUtil.java
```diff
@@ -175,8 +175,11 @@ private static String unifyDiacritics(String content) {
      * @return True if the author or editor is an institution.
      */
     private static boolean isInstitution(String author) {
-        return (author.charAt(0) == '{')
-                && (author.charAt(author.length() - 1) == '}');
+        if (!(author.isEmpty())) {
+            return (author.charAt(0) == '{') && (author.charAt(author.length() - 1) == '}');
+        } else {
+            return false; // In case of empty or null author
```
simonharrer: if author is null, a NPE will be thrown.

src/main/java/net/sf/jabref/JabRef.java
oscargus: Yes, there are a few more cases like that. I'll change them (now that I learned more about how to use Optional, didn't want to update during possible review...).

src/main/java/net/sf/jabref/logic/labelPattern/LabelPatternUtil.java
oscargus: Yes, bad comment. I had null check earlier, but then I realized that the method is never(?) called with a null argument anyway. Let me double check and come back with the correct solution. :-)

oscargus: As this method will at worst be called with an empty String, I just changed the comment. Also, the same holds for the methods `removeDiacritics()` and `generateInstitutionKey()` so I changed the null test to `isEmpty()` there.

PR #542, opened 2015-12-20
[WIP] Refactor cleanup code

tobiasdiez:
I moved the rest of the logic in `CleanupAction` to its new home `CleanupWorker`. This change should make it very easy to run cleanup operations at other points (for example after import).

The new code is not yet ready for review but I opened the PR to get some assistance: what is the expected behavior of the following cleanups:
- Unicode to Latex: `"Réflexions"` to `"R\'eflexion"`?
- Keep case in titles: "TEST" to "{TEST}" ? The dialog refers to "selected words"? which ones?
- Html to Latex: `<b>hallo</b>` to `"\\textbf{hallo}"`?

These are my expectations, but in fact all of them are wrong as failing tests show. What are the correct ones?


oscargus:
* `Réflexions` => `R{\'{e}}flexions`
* There is a hard coded list in CaseKeeperList mainly for Electrical Engineering. Ideally one should be able to provide more than one list, which easily can be selected based on subject area, and also keep a user defined list of commonly occurring words where the capitals should be kept. Why? There are names which are more or less common in certain areas. If I were into search engines I would add "Page", but probably not in other areas. Also, if you import "HOW TO TEST JAVA PROGRAMS" you do typically not want to keep it that way just because the database provided you that.
* No, that doesn't work, but it would be nice if it did. Currently, very few tags are honored and all are stripped.

tobiasdiez:
@oscargus Thanks! This was very helpful.

I close this PR as its originally purpose is fulfilled (and for some reason it was based on my master and not the correct branch...)

koppor:
@tobiasdiez I don't get why it's original purpose is fulfilled. Is there "better" code in the master branch already?

tobiasdiez:
It is now surpassed by #572.

PR #546, opened 2015-12-20
[WIP] First attempt at using TextAnalyzer

oscargus:
I thought I'd give the TextAnalyzer a go and it turns out that it can find year and pages numbers from my limited experiments. However, although volume and journal theoretically could be found I haven't figured out which format it succeeds in. 

To test it select "New entry from plain text", paste a reference in some format and press the wand.

The reverse marking is not fully working, I know. Not sure if I will progress on this one, it seems to be quite a bit of work to improve TextAnalyzer, but as a reference and starting point.

### DIFF
--- a/src/main/java/net/sf/jabref/wizard/text/TextAnalyzer.java
+++ b/src/main/java/net/sf/jabref/wizard/text/TextAnalyzer.java
@@ -13,7 +13,7 @@
     with this program; if not, write to the Free Software Foundation, Inc.,
     51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 */
-package net.sf.jabref.importer;
+package net.sf.jabref.wizard.text;
 
 import java.util.TreeSet;
 import java.util.Vector;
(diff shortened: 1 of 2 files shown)


oscargus:
Ready-built versions available at http://builds.jabref.org/usetextanalyzer/

koppor:
The idea is great. I'd love to have the result of "Parse with Freecite" integrated the same way.

However, could you provide us with an example entry? When I tried the following entry, only "12" is highlighted.

    Binz, Tobias; Breiter, Gerd; Leymann, Frank; Spatzier, Thomas: Portable Cloud Services Using TOSCA. In: IEEE Internet Computing. Vol. 16(03), Los Alamitos, CA, USA: IEEE Computer Society, 2012.

grabbed_20151221-114440

oscargus:
Yes, that is indeed one of the problems. The year is found and later on processed in several steps before the actual position is stored. This is still slightly better than the original version of TextAnalyzer. I couldn't really bother to work more on it yesterday as there were so many other issues. However, this particular one should be rather manageable. 

If you add page numbers those will be correctly extracted (but incorrectly highlighted) as well.

simonharrer:
What about this PR? Any idea how to progress there? Is it good enough already to be merged? Does it already create value or improves the code quality of JabRef?

oscargus:
The problem is that TextAnalyzer only can guess very few fields (year, pages, volume, journal) and not really consistently. The GUI issues can be fixed quite quickly, but I do not think it is worthwhile as long as TextAnalyzer is not more mature.

It is worthwhile merging from the perspective that if anyone gives it a go, TextAnalyzer is slightly better now. However, the GUI part should probably be removed for now, commenting out lines 232-233 in TextInputDialog for easy restoring.

But I strongly believe that no one will improve TextAnalyzer enough, so it may be better to remove it completely.

simonharrer:
We decided at JabCon that we remove this feature.

PR #547, opened 2015-12-20
[WIP] MedlineImporterTest

bruehldev:
New PR started because the old one was on a wrong branch.

MedlineImporTest + Testfile added

Couldn't cover two exceptions in importEntries.

### DIFF
--- a/src/main/java/net/sf/jabref/logic/util/strings/StringUtil.java
+++ b/src/main/java/net/sf/jabref/logic/util/strings/StringUtil.java
@@ -23,6 +23,7 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import net.sf.jabref.Globals;
+import net.sf.jabref.model.entry.Author;
 
 import com.google.common.base.CharMatcher;
 
@@ -623,6 +624,51 @@ public class StringUtil {
             result = result.replace(chrAndReplace.getKey(), chrAndReplace.getValue());
         }
         return result;
+    }
+
+    /**
+     * Expand initials, e.g. EH Wissler -> E. H. Wissler or Wissler, EH -> Wissler, E. H.
+     *
+     * @param name
+     * @return The name after expanding initials.
+     */
+    public static String expandAuthorInitials(String name) {
+        String[] authors = name.split(" and ");
+        StringBuilder sb = new StringBuilder();
+        for (int i = 0; i < authors.length; i++) {
+            if (authors[i].contains(", ")) {
+                String[] names = authors[i].split(", ");
+                if (names.length > 0) {
+                    sb.append(names[0]);
+                    if (names.length > 1) {
+                        sb.append(", ");
+                    }
+                }
+                for (int j = 1; j < names.length; j++) {
+                    if (j == 1) {
+                        sb.append(Author.addDotIfAbbreviation(names[j]));
+                    } else {
+                        sb.append(names[j]);
+                    }
+                    if (j < (names.length - 1)) {
+                        sb.append(", ");
+                    }
+                }
+            } else {
+                String[] names = authors[i].split(" ");
+                if (names.length > 0) {
+                    sb.append(Author.addDotIfAbbreviation(names[0]));
+                }
+                for (int j = 1; j < names.length; j++) {
+                    sb.append(' ');
+                    sb.append(names[j]);
+                }
+            }
+            if (i < (authors.length - 1)) {
+                sb.append(" and ");
+            }
+        }
+        return sb.toString().trim();
     }
 
     /**
--- a/src/test/java/net/sf/jabref/importer/fileformat/MedlineImporterTest.java
+++ b/src/test/java/net/sf/jabref/importer/fileformat/MedlineImporterTest.java
@@ -1,0 +1,100 @@
+package net.sf.jabref.importer.fileformat;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.DirectoryStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import net.sf.jabref.Globals;
+import net.sf.jabref.JabRefPreferences;
+import net.sf.jabref.importer.OutputPrinterToNull;
+import net.sf.jabref.model.entry.BibEntry;
+
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mockito;
+import org.mockito.runners.MockitoJUnitRunner;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Articles in the medline format can be downloaded from http://www.ncbi.nlm.nih.gov/pubmed/.
+ * 1. Search for a term and make sure you have selected the PubMed database
+ * 2. Select the results you want to export by checking their checkboxes
+ * 3. Press on the 'Send to' drop down menu on top of the search results
+ * 4. Select 'File' as Destination and 'XML' as Format
+ * 5. Press 'Create File' to download your search results in a medline xml file
+ *
+ * @author Daniel Mair/Bruehl
+ *
+ */
+@RunWith(MockitoJUnitRunner.class)
+public class MedlineImporterTest {
+
+    private MedlineImporter medlineImporter;
+    private static final String FILEFORMAT_PATH = "src/test/resources/net/sf/jabref/importer/fileformat";
+
+    /**
+     * Generates a List of all files in the package "/src/test/resources/net/sf/jabref/importer/fileformat"
+     * @return A list of Names
+     * @throws IOException
+     */
+    public List<String> getTestFiles() throws IOException {
+        List<String> files = new ArrayList<>();
+        try (DirectoryStream<Path> stream = Files.newDirectoryStream(Paths.get(FILEFORMAT_PATH))) {
+            stream.forEach(n -> files.add(n.getFileName().toString()));
+        }
+        return files;
+    }
+
+    @Before
+    public void setUp() throws Exception {
+        Globals.prefs = JabRefPreferences.getInstance();
+        this.medlineImporter = new MedlineImporter();
+    }
+
+    @Test
+    public void testExceptionOnInputStream() throws IOException {
+        try (InputStream is = Mockito.mock(InputStream.class)) {
+            Mockito.doThrow(new IOException()).when(is).read();
+            List<BibEntry> entry = medlineImporter.importEntries(is, new OutputPrinterToNull());
+            Assert.assertTrue(entry.isEmpty());
+        }
+    }
+
+    @Test
+    public void testGetItemsEmpty() {
+        MedlineHandler handler = new MedlineHandler();
+        assertEquals(Collections.emptyList(), handler.getItems());
+    }
+
+    @Test
+    public void testGetFormatName() {
+        assertEquals("Medline", medlineImporter.getFormatName());
+    }
+
+    @Test
+    public void testGetCLIId() {
+        assertEquals("medline", medlineImporter.getCLIId());
+    }
+
+    @Test
+    public void testIsRecognizedFormatReject() throws IOException {
+       List<String> list = getTestFiles().stream().filter(n -> !n.startsWith("MedlineImporter"))
+                .collect(Collectors.toList());
+
+        for (String str : list) {
+            try (InputStream is = MedlineImporter.class.getResourceAsStream(str)) {
+                Assert.assertFalse(medlineImporter.isRecognizedFormat(is));
+            }
+        }
+    }
+}
--- a/src/test/resources/net/sf/jabref/importer/fileformat/MedlineImporterTestArticleID.bib
+++ b/src/test/resources/net/sf/jabref/importer/fileformat/MedlineImporterTestArticleID.bib
@@ -1,0 +1,14 @@
+% Encoding: UTF-8
+
+@article{,
+  abstract = {This abstract is a dummy.},
+  author = {Rhein, Joshua and Bahr, Nathan C. and Hemmert, Andrew C. and Cloud, Joann L. and Bellamkonda, Satya and Oswald, Cody and Lo, Eric and Nabeta, Henry and Kiggundu, Reuben and Akampurira, Andrew and Musubire, Abdu and Williams, Darlisha A. and Meya, David B. and Boulware, David R. and , A. S. T. R. O.-C. M. Team},
+  institution = {Division of Infectious Disease and International Health, Department of Medicine, University of Minnesota, Minneapolis, MN, USA.},
+  journal = {Diagn Microbiol Infect Dis},
+  language = {eng},
+  medline-pst = {aheadofprint},
+  month = {Dec},
+  pmid = {26711635},
+  title = {Diagnostic performance of a multiplex PCR assay for meningitis in an HIV-infected population in Uganda.},
+  year = {2015}
+}
--- a/src/test/java/net/sf/jabref/importer/fileformat/MedlineImporterTestFiles.java
+++ b/src/test/java/net/sf/jabref/importer/fileformat/MedlineImporterTestFiles.java
@@ -15,17 +15,19 @@ import java.util.stream.Collectors;
 import net.sf.jabref.Globals;
 import net.sf.jabref.JabRefPreferences;
 import net.sf.jabref.importer.OutputPrinterToNull;
+import net.sf.jabref.logic.bibtex.BibEntryAssert;
 import net.sf.jabref.model.entry.BibEntry;
 
 import org.junit.Assert;
 import org.junit.Before;
+import org.junit.Ignore;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
 import org.junit.runners.Parameterized.Parameter;
 import org.junit.runners.Parameterized.Parameters;
 
-import static net.sf.jabref.bibtex.BibEntryAssert.assertEquals;
+import static org.junit.Assert.assertEquals;
 
 @RunWith(Parameterized.class)
 public class MedlineImporterTestFiles {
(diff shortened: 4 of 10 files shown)


oscargus:
BibtexEntry is called BibEntry now. Not sure how the CircleCI tests passes, but that is the reason that the Travis CI tests won't.

bruehldev:
I'm afraid i cant use the BibEntry. Somehow i can't use the libary ("BibEntry cannot be resolved to a type"), but i fetched my upstream.

oscargus:
Do you have a package net.sf.jabref.model.entry.BibEntry? model/entry/BibEntry.java
I'm more surprised that you still have BibtexEntry.java.

bruehldev:
I updated the branch with BibEntry. We will solve the package problem with our  tutor. Thanks Oscar

src/test/java/net/sf/jabref/importer/fileformat/MedlineImporterTest.java
```diff
@@ -0,0 +1,93 @@
...
+import org.junit.Test;
+import org.junit.Assert;
+
+/**
+ * @author Daniel Mair/Bruehl
+ *
+ */
+public class MedlineImporterTest {
+
+    MedlineImporter medlineImp;
```
simonharrer: Do not use abbreviations. Either use `medlineImporter` or `importer` as the context is clear by being in the `MedlineImporterTest` class.

src/test/java/net/sf/jabref/importer/fileformat/MedlineImporterTest.java
```diff
@@ -0,0 +1,93 @@
...
+                Assert.assertFalse(medlineImp.isRecognizedFormat(is));
+            }
+        }
+    }
+
+    @Test
+    public void testImportEntries() throws IOException {
+        try (InputStream is = MedlineImporter.class.getResourceAsStream("MedlineImporterTest.xml")) {
+            List<BibEntry> entries = medlineImp.importEntries(is, new OutputPrinterToNull());
+            Assert.assertEquals(1, entries.size());
```
simonharrer: This is not a good assertion, you need to verify as well all the fields of the BibEntry.

src/test/java/net/sf/jabref/importer/fileformat/MedlineImporterTest.java
```diff
@@ -0,0 +1,93 @@
...
+ */
+public class MedlineImporterTest {
+
+    MedlineImporter medlineImp;
+
+    @Before
+    public void setUp() throws Exception {
+        if (Globals.prefs == null) {
+            Globals.prefs = JabRefPreferences.getInstance();
+        }
```
simonharrer: You do not need the if-condition here, just set the `Globals.prefs`

src/test/java/net/sf/jabref/importer/fileformat/MedlineImporterTest.java
```diff
@@ -0,0 +1,93 @@
...
+        try (InputStream is = MedlineImporter.class.getResourceAsStream("CopacImporterTest1.txt")) {
+            // Trigger a catchblock
+            List<BibEntry> entries = medlineImp.importEntries(is, new OutputPrinterToNull());
+            //Assert.fail("Error during XML parsing");
+        }
+    }
+
+    @Test
+    public void testGetFormatName() {assertEquals("testGetFormatName failed", "Medline", medlineImp.getFormatName());
+    }
```
simonharrer: Mind the formatting, please.

src/test/java/net/sf/jabref/importer/fileformat/MedlineImporterTest.java
```diff
@@ -0,0 +1,93 @@
...
+
+    @Test
+    public void testGetCLIId() {
+        assertEquals("testGetFormatName failed", "medline", medlineImp.getCLIId());
```
simonharrer: Beware of copy-and-paste errors, as the message here is completely wrong. A message of an assert should add additional information, which in this case, it does not, it adds misleading information.

mairdl:
For now I took care of the smaller changes, more changes coming soon

boceckts:
Added more test files and added testing against bibtex files.

koppor:
@boceckts Is this ready for review? Then please add this label?

boceckts:
Testcoverage is at only ~80% so far for MedlineImporter and at ~90% for MedlineHandler

stefan-kolb:
What is the status here? DL was beginning of match.

src/test/java/net/sf/jabref/importer/fileformat/MedlineImporterTest.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    @Test
+    public void testImportEntryWithArticleID() throws IOException {
+        checkImportEntriesAgainstBibFileEntries("MedlineImporterArticleIDTest.xml",
+                "MedlineImporterArticleIDTestBib.bib");
+        try (InputStream shouldBeIs = MedlineImporter.class.getResourceAsStream("MedlineImporterArticleIDTest.xml");
+                InputStream bibIn = MedlineImporter.class.getResourceAsStream("MedlineImporterArticleIDTestBib.bib")) {
+            List<BibEntry> entries = medlineImporter.importEntries(shouldBeIs, new OutputPrinterToNull());
+            Assert.assertEquals(1, entries.size());
```
tschechlovdev: this assertion is not necessary, it is done in BibtexEntryAssert(InputStream, List<BibEntry>)

src/test/java/net/sf/jabref/importer/fileformat/MedlineImporterTest.java
```diff
@@ -0,0 +1,132 @@
...
+            BibEntry entry = entries.get(0);
+            BibtexEntryAssert.assertEquals(bibIn, entry);
```
tschechlovdev: use BibtexEntryAssert.assertEquals(InputStream, List<BibEntry>)

src/test/java/net/sf/jabref/importer/fileformat/MedlineImporterTest.java
```diff
@@ -0,0 +1,132 @@
...
+                "MedlineImporterMinimalEntryTestBib.bib");
+    }
+
+    private void checkImportEntriesAgainstBibFileEntries(String medlineFile, String bibTexFile) throws IOException {
+        try (InputStream is = MedlineImporter.class.getResourceAsStream(medlineFile);
+                InputStream nis = MedlineImporter.class.getResourceAsStream(bibTexFile)) {
+            List<BibEntry> entries = medlineImporter.importEntries(is, new OutputPrinterToNull());
+            Assert.assertNotNull(entries);
+            Assert.assertEquals(1, entries.size());
+            BibtexEntryAssert.assertEquals(nis, entries.get(0));
```
tschechlovdev: See above. 
Assert.assertNotNull is not necessary as well if you use `assertEquals(InputStram, List<BibEntry>)`

koppor:
Please rebase everything on master and create on commit. Description is here: https://github.com/JabRef/jabref/wiki/Tools#rebase-everything-as-one-commit-on-master Please use `git gui` for your commit. Please do not delete other test files.

There should no commits handling the InspecImporter in this pull request. Inspec has been handled in #356

mairdl:
I am running into a problem in the "GrammerBasedSearchRule.java" it is missing some required imports. The missing one are "net.sf.jabref.search.SearchBaseVisitor", "net.sf.jabref.search.SearchLexer" and "net.sf.jabref.search.SearchBaseParser"

tobiasdiez:
@mairdl You have to run `gradlew generateSource`.

mairdl:
Now it seems to work, thank you.

src/test/java/net/sf/jabref/importer/fileformat/MedlineImporterTest.java
```diff
@@ -0,0 +1,128 @@
...
+    public void testGetCLIId() {
+        assertEquals("medline", medlineImporter.getCLIId());
+    }
+
+    @Test
+    public void testFetchMedline() {
+        List<BibEntry> validMedLine = medlineImporter.fetchMedline("1", new OutputPrinterToNull());
+        List<BibEntry> emptyMedLine = medlineImporter.fetchMedline("zzzzzz", new OutputPrinterToNull());
+        assertNotNull(validMedLine);
+        assertEquals(0, emptyMedLine.size());
```
tschechlovdev: Use assertEquals(Collections.EMPTY_LIST, emptyMedline) instead

src/test/java/net/sf/jabref/importer/fileformat/MedlineImporterTest.java
```diff
@@ -0,0 +1,128 @@
...
+
+    @Test
+    public void testIsRecognizedFormatEmptyFile() throws IOException {
+        Assert.assertFalse(medlineImporter.isRecognizedFormat(emptyFileStream));
+    }
+
+    @Test
+    public void testImportMalformedEntry() throws IOException {
+        try (InputStream is = MedlineImporter.class.getResourceAsStream("MedlineImporterMalformedEntryTest.xml")) {
+            medlineImporter.importEntries(is, new OutputPrinterToNull());
```
tschechlovdev: why is here no assertion?

src/test/resources/net/sf/jabref/importer/fileformat/MedlineImporterArticleIDTestBib.bib
```diff
@@ -0,0 +1,15 @@
...
...ial to optimize survival. We evaluated the utility of a multiplex PCR test in differentiating possible etiologies of meningitis. Cerebrospinal fluid (CSF) from 69 HIV-infected Ugandan adults with meningitis was collected at diagnosis (n=51) and among persons with cryptococcal meningitis during therapeutic lumbar punctures (n=68). Cryopreserved CSF specimens were analyzed with BioFire FilmArray® Meningitis/Encephalitis panel, which targets 17 pathogens. The panel detected Cryptococcus in the CSF of patients diagnosed with a first episode of cryptococcal meningitis by fungal culture with 100\% sensitivity and specificity and differentiated between fungal relapse and paradoxical immune reconstitution inflammatory syndrome in recurrent episodes. A negative FilmArray result was predictive of CSF sterility on follow-up lumbar punctures for cryptococcal meningitis. EBV was frequently detected in this immunosuppressed population (n=45). Other pathogens detected included: cytomegalovirus (n=2), varicella zoster virus (n=2), human herpes virus 6 (n=1), and Streptococcus pneumoniae (n=1). The FilmArray Meningitis/Encephalitis panel offers a promising platform for rapid meningitis diagnosis.},
```
boceckts: Please use a dummy abstract instead of an actual one

src/test/resources/net/sf/jabref/importer/fileformat/MedlineImporterTestArticleID.bib
```diff
@@ -0,0 +1,14 @@
...
...ial to optimize survival. We evaluated the utility of a multiplex PCR test in differentiating possible etiologies of meningitis. Cerebrospinal fluid (CSF) from 69 HIV-infected Ugandan adults with meningitis was collected at diagnosis (n=51) and among persons with cryptococcal meningitis during therapeutic lumbar punctures (n=68). Cryopreserved CSF specimens were analyzed with BioFire FilmArray® Meningitis/Encephalitis panel, which targets 17 pathogens. The panel detected Cryptococcus in the CSF of patients diagnosed with a first episode of cryptococcal meningitis by fungal culture with 100\% sensitivity and specificity and differentiated between fungal relapse and paradoxical immune reconstitution inflammatory syndrome in recurrent episodes. A negative FilmArray result was predictive of CSF sterility on follow-up lumbar punctures for cryptococcal meningitis. EBV was frequently detected in this immunosuppressed population (n=45). Other pathogens detected included: cytomegalovirus (n=2), varicella zoster virus (n=2), human herpes virus 6 (n=1), and Streptococcus pneumoniae (n=1). The FilmArray Meningitis/Encephalitis panel offers a promising platform for rapid meningitis diagnosis.},
```
tschechlovdev: use a dummy abstract

mairdl:
The test are now parameterized as suggested by @tschechlovdev. Also I removed the copyrighted abstract.

simonharrer:
Please fix codacy issues.

src/main/java/net/sf/jabref/logic/util/strings/StringUtil.java
```diff
@@ -646,7 +646,14 @@ public static String expandAuthorInitials(String name) {
...
                     if (j == 1) {
-                        sb.append(Author.addDotIfAbbreviation(names[j]));
+                        String[] splittedNames = names[j].split(" ");
+                        if (splittedNames.length > 0) {
+                            sb.append(Author.addDotIfAbbreviation(splittedNames[0]));
+                        }
+                        for (int k = 1; k < splittedNames.length; k++) {
+                            sb.append(' ');
+                            sb.append(Author.addDotIfAbbreviation(splittedNames[k]));
+                        }
```
simonharrer: If you split, you should join afterwards using `String.join(" ", elems);`

mairdl: I am not sure if that's needed here, because the whole method `expandAuthorInitials()` doesn't use it all. This might be the case because the string that gets splitted is used to build a new string via `StringBuilder`.

mairdl:
Some tests outside the `MedlinImporterTest` and `MedlineImporterTestFiles` are failing. 
The class is the `ContainsAndRegexBasedSearchRuleDescriberTest`  class and the `GrammarBasedSearchRuleDescriberTest` class. The expected value is a `String` in english but the actual `String` is written in german.

src/main/java/net/sf/jabref/logic/util/strings/StringUtil.java
tobiasdiez: Can you please shortly outline why this change is necessary?

mairdl: A test case revealed this issue. The importent lines were these:
 `<Author>
                    <CollectiveName>ASTRO-CM Team</CollectiveName>
                </Author>` 
Bewteen ASTRO-CM and Team it would add an additional whitespace. With this change the whitespace won't be added.

mairdl:
Codacy issues should be fixed now.

src/main/java/net/sf/jabref/logic/util/strings/StringUtil.java
koppor: @mairdl The code you provided is the same as

```java
List<String> innerNameList = Arrays.asList(splittedNames);
String innerNames  innerNameList.stream().map(n -> Author.addDotIfAbbreviation(n)).collect(Collectors.joining(" "));
sb.append(innerNames);
```

(untested, please double check if it really works :innocent:)

This is much shorter than your code, probably more readable (if one is used to Java8) and less error prone.

See also http://stackoverflow.com/a/22577565/873282

mairdl: The first verson passes all test. Can't say that for you 2nd example.

I agree  that it's more readable, shall i change it? The code i used was inspired by the `else `part of this method.

simonharrer: For simplicity, I would probably use the StringJoiner class without streams.

koppor: Ah, I see. According to [the JAvaDoc of StringJoiner](https://docs.oracle.com/javase/8/docs/api/java/util/StringJoiner.html), one can also transform the inputs (achieved by `map` in my proposal). Since the iteration on the split result (an array) is easy via an foreach, this should easily work without some if checking for non-emptyness (`splittedNames.length > 0`) etc.

mairdl:
I'd like to move the issue #1267 to a seperate PR, to have a cleaner cut

matthiasgeiger:
So you have created a test for importing two or more entries which fails?

mairdl:
No, but @koppor wants tests even if it not fails

matthiasgeiger:
Yes. And this tests should be included here, I think. 

If the tests fail, the failing tests in conjunction with the bugfix can be in another PR.

koppor:
The failed tests can be included here with `@Ignore("Will be fixed in a future PR")` so that there is no huzzle with git, can't they?

matthiasgeiger:
Yes, of course and this is an even better solution. :wink:

src/main/java/net/sf/jabref/logic/util/strings/StringUtil.java
mairdl: The problem here is that atleast to my experience, that a `StringJoiner` and `StringBuilder` or another `StringJoiner` has to be used. This is due to the reason that it first has to put the last name and first name together with a `", "` after that it should add that `"last name, first name"` with a `" and "` to the `String` that gets returned. I don't know if that increases simplicity. I suggest using one of koppor's propasals.

PS: Sorry for the stall, I had an operation this week and was resting.

mairdl:
After some testing i came to the conclusion that this bug/issue is not a problem in the current development version. At the moment I am struggeling to coming up with the right test case. At the moment I am thinking of using a `file.bib` wit some pubmed entries and then importing it with the RISimporter and checking if it works properly. But that seems a bit unfitting for the MedlineImporter.

simonharrer:
What is the current status of this PR?

mairdl:
waiting for some feedback regardnig the issue #1267 and my comment and also on comments from me regarding the use of a `StringJoiner`

src/main/java/net/sf/jabref/logic/util/strings/StringUtil.java
simonharrer: The StringBuilder has nothing to do with the StringJoiner in that case. Just use this solution:

```java
String innerNames = Arrays.stream(names[j].split(" "))
    .map(Author::addDotIfAbbreviation)
    .collect(Collectors.joining(" "));
sb.append(innerNames);
```

mairdl: This leads to some errors in the `StringUtilTest.java` because the suffix gets ignored.

mairdl:
@simonharrer your suggested version leads to some problems

src/main/java/net/sf/jabref/logic/util/strings/StringUtil.java
tobiasdiez: We decided to completely remove the "expand author initials" functionality in #1373.

tobiasdiez:
Since this PR blocks #1207, I would prefer that we merge this as quickly as possible. So just ignore the issue #1267 for the moment.

simonharrer:
@mairdl hm, then revert it to the state without my suggestion, and rebase. We try to merge this in quickly as it blocks #1207

mairdl:
ok i will revert it asap

mairdl:
I reverted the changes of StringUtil and added a ignored label to the tests that would fail.

simonharrer:
Thanks. Please open a new PR then for the fix of the failing test. :)

PR #553, opened 2015-12-21
Scifinder importer test deleted

Braunch:
Scifinder Importer is deleted. We could not get any example file to test the importer. The website www.cas.org needs login credentials, which we do not have. Even after asking the users on the sourceforge mailing list for a scifinder resource file, we could not get any information on how the scifinder files should look like. It seems the scifinder importer is redundant.


simonharrer:
Please rebase on master, squash the four commits into a single one and add a CHANGELOG entry.

Issue #544, opened 2015-12-20
List all non-existent linked files


tobiasdiez:
It would be nice if JabRef could give a list of all files which are linked in entries but which do not exists/are not available.

simonharrer:
This could be done in an integrity check.

stefan-kolb:
:+1:

tobiasdiez:
This was implemented in #863.

PR #557, opened 2015-12-21
Moved methods from Util

oscargus:
* Moved `getListOfLinkedFiles` and `findAssociatedFiles` from Util to FileUtil
* Rewrote `getListOfLinkedFiles` to be independent of `FileListTableModel`, introducing a new method `decodeFileField` which does most of what  `FileListTableModel.setContent()` does (and later can be used there)
* Used the `decodeFileField` method in `DatabaseFileLookup`
* Replaced `"file"`with `Globals.FILE_FIELD` where appropriate
* Moved `intValueOf` from Util to StringUtil
* Added tests for `intValueOf`

### DIFF
--- a/src/main/java/net/sf/jabref/JabRefPreferences.java
+++ b/src/main/java/net/sf/jabref/JabRefPreferences.java
@@ -1283,6 +1283,21 @@ public final class JabRefPreferences {
     }
 
     /**
+     * Check if there is an external file type registered with this name.
+     *
+     * @param name The file type name.
+     * @return true if there is an ExternalFileType with this name, false if not.
+     */
+    public Boolean isExternalFileTypeName(String name) {
+        for (ExternalFileType type : externalFileTypes) {
+            if (type.getName().equals(name)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    /**
      * Look up the external file type registered for this extension, if any.
      *
      * @param extension The file extension.
@@ -1292,6 +1307,21 @@ public final class JabRefPreferences {
         for (ExternalFileType type : externalFileTypes) {
             if ((type.getExtension() != null) && type.getExtension().equalsIgnoreCase(extension)) {
                 return type;
+            }
+        }
+        return null;
+    }
+
+    /**
+     * Look up the external file type name registered for this extension, if any.
+     *
+     * @param extension The file extension.
+     * @return The name of the ExternalFileType registered, or null if none.
+     */
+    public String getExternalFileTypeNameByExt(String extension) {
+        for (ExternalFileType type : externalFileTypes) {
+            if ((type.getExtension() != null) && type.getExtension().equalsIgnoreCase(extension)) {
+                return type.getName();
             }
         }
         return null;
@@ -1331,6 +1361,26 @@ public final class JabRefPreferences {
         }
         if ("text/html".equals(mimeType)) {
             return HTML_FALLBACK_TYPE;
+        } else {
+            return null;
+        }
+    }
+
+    /**
+     * Look up the external file type name registered for this MIME type, if any.
+     *
+     * @param mimeType The MIME type.
+     * @return The name of the ExternalFileType registered, or null if none. For the mime type "text/html", a valid file type name is
+     *         guaranteed to be returned.
+     */
+    public String getExternalFileTypeNameByMimeType(String mimeType) {
+        for (ExternalFileType type : externalFileTypes) {
+            if ((type.getMimeType() != null) && type.getMimeType().equals(mimeType)) {
+                return type.getName();
+            }
+        }
+        if ("text/html".equals(mimeType)) {
+            return HTML_FALLBACK_TYPE.getName();
         } else {
             return null;
         }
--- a/src/main/java/net/sf/jabref/JabRefPreferences.java
+++ b/src/main/java/net/sf/jabref/JabRefPreferences.java
@@ -53,16 +53,13 @@ import org.apache.commons.logging.LogFactory;
 import net.sf.jabref.exporter.CustomExportList;
 import net.sf.jabref.exporter.ExportComparator;
 import net.sf.jabref.external.DroppedFileHandler;
-import net.sf.jabref.external.ExternalFileType;
-import net.sf.jabref.external.UnknownExternalFileType;
 import net.sf.jabref.importer.CustomImportList;
 import net.sf.jabref.logic.remote.RemotePreferences;
 import net.sf.jabref.specialfields.SpecialFieldsUtils;
-import net.sf.jabref.logic.util.strings.StringUtil;
 
 public final class JabRefPreferences {
     private static final Log LOGGER = LogFactory.getLog(JabRefPreferences.class);
-    private static final String EXTERNAL_FILE_TYPES = "externalFileTypes";
+    public static final String EXTERNAL_FILE_TYPES = "externalFileTypes";
 
     /**
      * HashMap that contains all preferences which are set by default
@@ -348,9 +345,6 @@ public final class JabRefPreferences {
     private static final String CUSTOM_TYPE_PRIOPT = "customTypePriOpt_";
     public static final String PDF_PREVIEW = "pdfPreview";
 
-    // This String is used in the encoded list in prefs of external file type
-    // modifications, in order to indicate a removed default file type:
-    private static final String FILE_TYPE_REMOVED_FLAG = "REMOVED";
 
     private static final char[][] VALUE_DELIMITERS = new char[][] { {'"', '"'}, {'{', '}'}};
 
@@ -376,10 +370,6 @@ public final class JabRefPreferences {
 
     // Object containing info about customized entry editor tabs.
     private EntryEditorTabList tabList;
-    // Map containing all registered external file types:
-    private final Set<ExternalFileType> externalFileTypes = new TreeSet<>();
-
-    private final ExternalFileType HTML_FALLBACK_TYPE = new ExternalFileType("URL", "html", "text/html", "", "www", IconTheme.JabRefIcon.WWW.getSmallIcon());
 
     // The following field is used as a global variable during the export of a database.
     // By setting this field to the path of the database's default file directory, formatters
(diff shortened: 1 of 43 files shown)


src/test/java/net/sf/jabref/logic/util/strings/StringUtilTest.java
```diff
@@ -186,4 +186,41 @@ public void testBooleanToBinaryString() {
...
+    public void testIntValueOfExceptionLetter() {
+        try {
+            StringUtil.intValueOf("12A2");
+            fail();
+        } catch (NumberFormatException ignored) {
+            // Ignored
+        }
+    }
+
+    @Test
```
simonharrer: just add a `@Text(expected = NumberFormatException.class)` to test that a specific exception has to occur. The test itself is only one line long then.

oscargus: Ahh, didn't know that!

src/main/java/net/sf/jabref/logic/util/io/FileUtil.java
```diff
@@ -318,4 +321,125 @@ private static File shortenFileName(File fileName, String dir) {
...
+     * @param bes list of BibTeX entries
+     * @param fileDirs list of directories to try for expansion
+     *
+     * @return list of files. May be empty
+     */
+    public static List<File> getListOfLinkedFiles(BibEntry[] bes, String[] fileDirs) {
+        ArrayList<File> result = new ArrayList<>();
+        for (BibEntry entry : bes) {
+            List<List<String>> fileList = decodeFileField(entry.getField(Globals.FILE_FIELD));
+            for (int i = 0; i < fileList.size(); i++) {
```
simonharrer: use a for-each-loop instead.

simonharrer:
May help in implementing #544

src/main/java/net/sf/jabref/logic/util/strings/StringUtil.java
```diff
@@ -500,4 +500,35 @@ private static String removeSingleBracesAroundCapitals(String s) {
...
+    /**
+     * Optimized method for converting a String into an Integer
+     *
+     * From http://stackoverflow.com/questions/1030479/most-efficient-way-of-converting-string-to-integer-in-java
+     *
+     * @param str the String holding an Integer value
+     * @throws NumberFormatException if str cannot be parsed to an int
+     * @return the int value of str
+     */
+    public static int intValueOf(String str) {
```
tobiasdiez: Do we really call this method so often that a optimized version is needed? `Integer.parseInt()` seems to be efficient enough.

oscargus: See #270

oscargus: Or rather #269

tobiasdiez: Ok, didn't know about its history :+1:

src/test/java/net/sf/jabref/logic/util/strings/StringUtilTest.java
```diff
@@ -186,4 +186,41 @@ public void testBooleanToBinaryString() {
         assertEquals("1", StringUtil.booleanToBinaryString(true));
     }
 
+    @Test
+    public void testIntValueOf() {
```
tobiasdiez: Can you rename and split the test please. For example, `InfValueOfParses Small/Large NumberCorrectly`.

src/test/java/net/sf/jabref/logic/util/strings/StringUtilTest.java
```diff
@@ -186,4 +186,41 @@ public void testBooleanToBinaryString() {
...
+        assertEquals(1, StringUtil.intValueOf("1"));
+        assertEquals(1234567890, StringUtil.intValueOf("1234567890"));
+    }
+
+    @Test
+    public void testIntValueOfExceptionLetter() {
```
tobiasdiez: Rename test, e.g. `InfValueOfThrowsExceptionIfStringContainsLetter`

src/main/java/net/sf/jabref/logic/util/io/FileUtil.java
```diff
@@ -318,4 +321,125 @@ private static File shortenFileName(File fileName, String dir) {
...
+                        }
+                    }
+                }
+            }
+        }
+
+        return result;
+    }
+
+    public static List<List<String>> decodeFileField(String value) {
```
tobiasdiez: If I understood you correctly, then this method is extracted from the `FileListTableModel` right? Then probably you should delete the code there, too?

Moreover, is there no better way to return a meaningful value here. A list of a list sounds rather cryptic. I would add a `LinkedFile` class which has the right properties (description, link, file type) and return here a list of these objects.

oscargus: It can't be deleted (but you are otherwise correct), as the added code is much more limited. However, with some thinking the code in FileListTableModel can be simplified and the added code used (primary problem is the code "optimized" for getting the first item only).

The code in FileListTableModel uses FileListEntry which consists of description, link, and ExternalFileType. Although ExternalFileType resides in external, so FileListEntry only includes from there, ExternalFileType has both a JPanel and an Icon. Although, it was my original plan to extract the whole thing setting up a list of FileListEntry, I thought that introducing (hidden) dependencies on Swing/GUI components in logic was a bad idea (although the independence tests might have passed).

One could possibly introduce another class, FileItem(?) which consists of the description, link, and filetype name. This is really enough for serialization and may be enough in some of the cases where now the whole FileListTableModel machinery is included (as you state in #572).

oscargus:
Will rebase and honor the comments (where appropriate), but not in a few days.

Merry Christmas! :christmas_tree: :santa:

oscargus:
Seems like I got a bit of time on my hands anyway.

Rebased, fixed, and a few more tests added.

oscargus:
Not sure when I will have time to continue, so feel free to merge if you think so.

@tobiasdiez : The new `SimpleFileList` will enable making file related cleanup actions independent of the gui package. Some testing may be good though, although it is basically a copy of `FileListTableModel` with all the non-required stuff removed.

I will add tests eventually and also change all (remaining) code which now use `FileListTableModel` without the need of a complete `ExternalFileType`.

oscargus:
OK, now I will try to leave this for a while so that it can be reviewed and possibly merged. Otherwise, it will just increase and increase. :-)

src/main/java/net/sf/jabref/logic/util/io/SimpleFileList.java
```diff
@@ -0,0 +1,117 @@
...
+import net.sf.jabref.Globals;
+import net.sf.jabref.logic.util.io.FileUtil;
+import net.sf.jabref.logic.util.io.SimpleFileListEntry;
+import net.sf.jabref.logic.util.strings.StringUtil;
+
+/**
+ * Data structure to contain a list of file links, parseable from a coded string.
+ * To be used instead of FileListTableModel when no GUI parts are required.
+ */
+public class SimpleFileList {
```
tobiasdiez: I wouldn't use a new class to wrap `List<SimpleFileListEntry>`. All the decode methods should be moved to the `FileUtil` class.

src/main/java/net/sf/jabref/logic/util/io/SimpleFileListEntry.java
```diff
@@ -0,0 +1,94 @@
...
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+package net.sf.jabref.logic.util.io;
+
+import java.util.List;
+
+/**
+ * This class represents a simplified version of the file link for a Bibtex entry.
+ */
+public class SimpleFileListEntry {
```
tobiasdiez: Why it is just a simplified version? I thought it contains all the data of a file link. Should probably be moved to the `model` package. 
I don't like the class name. What do you think about `LinkedFile` or `FileItem` (not sure if they are better :-) ).

oscargus: A `FileListEntry`, as used in `FileListTableModel` has an `ExternalFileType` as the third parameter. This results in bindings to the GUI package etc. A `SimpleFileListEntry` has the "correct" type name (as specified by an existing `ExternalFileType`) as the third parameter, which is enough for serialization but not for GUI.

As stated above, the difference between a plain `List<SimpleFileListEntry>` and `SimpleFileList` is that `ExternalFileType` is consulted via the preferences when the entries are generated and that it has methods required for simple drop-in replacement of `FileListTableModel`, including serialization.

Yes, maybe these methods should go to `model`. However, it is not clear exactly what should go where. Also, looking at the time perspective, it wasn't obvious from the start that I would end up here. Finally, there's a bit of discussion how many methods should be added to the entries. Now, parts of this is rather JabRef specific and should maybe stay away from `model` (which I think is aimed to be as "clean" as possible). 

Naming: yes, I agree that they are not exceptionally well chosen. One should also take into consideration that `FileListEntry` now extends`SimpleFileListEntry`, and that there's a `FileListEntryEditor` (and a `FileListTableModel`).

tobiasdiez: Ok, thanks for you clarification. 
So `FileListEntry` is the gui version of `SimpleFileListEntry`. Ok this makes sense and it is good that these two purposes are clearly separated. 

So the only difference between `List<SimpleFileListEntry>` and `SimpleFileList` is the method `setContent`, right? In other words, just the initialization differs. Thus my idea was to use the Builder design pattern and move `setContent` to a builder class (along with changing its signature to `String -> List<SimpleFileListEntry>`). In this way you can separate the initialization from the data-holding and, moreover, move the dependence on the preferences to a separate class.

oscargus: `setContent` (especially calling the method `decodeEntry`) and `getStringRepresentation`.

To be honest, I have no idea about Builder design patterns, so I cannot really say anything about it. (I'm actually quite bad at programming Java, but have picked up a bit at least recently.)

tobiasdiez: The `SimpleFileList` class right now has two purposes. For one, it contains the list of linked files, and secondly, it has the `setContent` method which parses a string to proper list of `SimpleFileListEntries`. So the idea was to move the latter functionality to a separate class. In detail:
1. Move `setContent` to a new class (maybe called `FileListBuilder` or `FileListFactory`) or move it to `FileUtil` as you want. In order to make the method work in its new home, you need to return the generated list of file entries (instead of writing it to a private field).
2. Now nothing is left in `SimpleFileList`. So it can be replaced by `List<SimpleFileListEntry>`.

oscargus: Except for `getStringRepresentation` (and the fact that it relies on static methods rather than keeping things as a class). I'll wait and see what the other guys have to say about it. I could in fact think that it may make sense to provide the file field directly to the constructor. May not change the fundamental discussion, but from other perspectives one probably would like that anyway.

oscargus:
Rebased

oscargus:
Rebased again. Anything missing here?

src/main/java/net/sf/jabref/JabRefPreferences.java
```diff
@@ -1337,6 +1368,26 @@ public ExternalFileType getExternalFileTypeByMimeType(String mimeType) {
...
+            if ((type.getMimeType() != null) && type.getMimeType().equals(mimeType)) {
+                return type.getName();
+            }
+        }
+        if ("text/html".equals(mimeType)) {
+            return HTML_FALLBACK_TYPE.getName();
+        } else {
+            return null;
+        }
+    }
```
simonharrer: From a refactoring point of view, I would suggest to create a class which contains all the external file type methods as JabRefPreferences should manage the preferences, but not external file types.

oscargus: Check koppor/jabref#115 :-)

oscargus: I'm giving it a go within this PR.

oscargus:
Not sure if this is the correct/best way to do it, but it seems to work.

src/main/java/net/sf/jabref/logic/util/io/SimpleFileListEntry.java
simonharrer: I do not like the current implementation of FileListEntry, SimpleFileListEntry and SimpleFileList. 

The thing is that the current implementation suffers as it comes directly from dependency hell. ;)

What I would expect to be there as classes:
- ExternalFileType defines the type of the ExternalFile (currently named FileListEntry). A class (could be bibenty, but I am not sure if it should not be another class) should be able to decode a List<ExternalFile> from its file attribute. 
- ExternalFileType should be independent from the GUI, and then the ExternalFile would be GUI independent as well - then both ExternalFile and ExternalFileType can be put in the model. At the moment ExternalFileType contains a label and an icon field. Hence, this should be split up into a GUIDependentExternalFileType which contains both the JLabel and the Icon plus a reference to the ExternalFileType which contains all the information which is serialized. Then, the SimpleFileListEntry is not necessary anymore. The issue here is that the current mapping of an ExternalFileType to an Icon is currently hard-coded, so we need a large switch statement in the convertion code of an ExternalFileType to a GUIDependentExternalFileType.

oscargus: I'll get back with a more detailed answer, but that is basically the
current implementation, subject to naming and a few details.

When it comes to the actual lists, should there be classes or static
utility methods filling the lists? Main benefit of a class is that the
serialisation formatting can be included.

simonharrer: When a utility method is only used once, it should not be in a utility class. And that case, it would only be used for this call. 

What is more, I do not like utility classes for domain specific things like external file types. It is OK for IO methods like reading from a file or interfacing with the file system, or for type conversion such as String to int.

simonharrer:
I do not like having singletons or global variables, as code with singletons or global variables becomes extremely hard to test. But I do not have any idea how to solve this issue at the moment - maybe JabRef is stuck with such things for eternity. ;)

simonharrer:
This has been merged manually, plus 2 hours of rework by @stefan-kolb and myself. 

One note for the future @oscargus : it contained a lot of ideas, but please keep PRs small :) then, we can merge small ideas quickly

But thanks for the effort on creating this PR @oscargus :smile:

PR #560, opened 2015-12-21
Write all field keys in lower case

koppor:
Write all field keys in lower case. No more camel casing of field names. E.g., `title` is written instead of `Title`, `howpublished` instead of `HowPublished`, and `doi` instead of `DOI`. The configuration option `Use camel case for field names (e.g., "HowPublished" instead of "howpublished")` is gone.

Refs #116


src/main/java/net/sf/jabref/bibtex/CamelCaser.java
```diff
@@ -1,101 +0,0 @@
-package net.sf.jabref.bibtex;
-
-import java.util.HashMap;
-import java.util.Map;
-
-class CamelCaser {
```
lenhard: That class might have been useful as a future formatter, but nevermind. We can rewrite it if needed.

src/test/java/net/sf/jabref/bibtex/BibEntryWriterTest.java
```diff
@@ -53,24 +51,28 @@ public void testSerialization() throws IOException {
...
-                "  Author                   = {Foo Bar}," + Globals.NEWLINE +
-                "  Journal                  = {International Journal of Something}," + Globals.NEWLINE +
-                "  Note                     = {some note}," + Globals.NEWLINE +
-                "  Number                   = {1}" + Globals.NEWLINE +
+                "  author                   = {Foo Bar}," + Globals.NEWLINE +
+                "  journal                  = {International Journal of Something}," + Globals.NEWLINE +
+                "  note                     = {some note}," + Globals.NEWLINE +
+                "  number                   = {1}" + Globals.NEWLINE +
                 "}";
+        // @formatter:on
```
lenhard: Please do not simply rewrite the tests this way! I used this test to ensure that a rewrite without modification does not change the serialization. Your change eliminates this information! Please only change the test that modify an entry (which results in a reformatting on write)

src/test/java/net/sf/jabref/bibtex/BibEntryWriterTest.java
```diff
@@ -95,12 +97,14 @@ public void roundTripTest() throws IOException {
...
```
lenhard: See above.

src/test/java/net/sf/jabref/bibtex/BibEntryWriterTest.java
```diff
@@ -151,24 +157,28 @@ public void roundTripWithModification() throws IOException {
...
         assertEquals(expected, actual);
     }
 
     @Test
     public void roundTripWithCamelCasing() throws IOException {
+        // @formatter:off
```
lenhard: If camel casing is turned off, you should propably also rename this method.

src/test/java/net/sf/jabref/bibtex/BibEntryWriterTest.java
```diff
@@ -190,24 +200,28 @@ public void roundTripWithCamelCasing() throws IOException {
...
-                "  Author                   = {Foo Bar}," + Globals.NEWLINE +
-                "  Journal                  = {International Journal of Something}," + Globals.NEWLINE +
-                "  Note                     = {some note}," + Globals.NEWLINE +
-                "  Number                   = {1}" + Globals.NEWLINE +
+                "  author                   = {Foo Bar}," + Globals.NEWLINE +
+                "  journal                  = {International Journal of Something}," + Globals.NEWLINE +
+                "  note                     = {some note}," + Globals.NEWLINE +
+                "  number                   = {1}" + Globals.NEWLINE +
                 "}\n\n";
+        // @formatter:on
```
lenhard: See above. Please do not modify casing here!

src/test/java/net/sf/jabref/bibtex/BibEntryWriterTest.java
```diff
@@ -233,12 +247,14 @@ public void roundTripWithAppendedNewlines() throws IOException {
...
-                "  Number                   = {1}" + Globals.NEWLINE +
+                "  author                   = {Foo Bar}," + Globals.NEWLINE +
+                "  journal                  = {International Journal of Something}," + Globals.NEWLINE +
+                "  note                     = {some note}," + Globals.NEWLINE +
+                "  number                   = {1}" + Globals.NEWLINE +
                 "}";
+        // @formatter:on
 
         String result = testSingleWrite(bibtexEntry);
         result = testSingleWrite(result);
```
lenhard: See above. please do not modify casing here!

src/main/java/net/sf/jabref/bibtex/CamelCaser.java
koppor: I also found the class a good example for good code. However, we decided to remove unused code. :neckbeard:

src/test/java/net/sf/jabref/bibtex/BibEntryWriterTest.java
koppor: The current PR changes the serialization. All fields will be written lower case. So I have to change `roundTripTest`, don't I?

src/test/java/net/sf/jabref/bibtex/BibEntryWriterTest.java
koppor: Done. `roundTripWithCamelCasingInTheOriginalEntryAndResultInLowerCase`

koppor:
Maybe, the UI also has to display lower case values or should it use the (currently removed) CamelCase functionality? I don't like the current `EntryUtil.capitalizeFirst` solution (found in MainTableColumn.java).

lenhard:
In my opinion, it should use lower case, since we are standardizing on that. I'd say you are free to implement what you find more beautiful :)

simonharrer:
:+1: on lower case.

PR #566, opened 2015-12-22
Remove alternative display name since it is never set

koppor:
I think, we don't need alternative display names. This will surely confuse users.


simonharrer:
Hm, from the code it seems that this was never used or set. :+1:

koppor:
One use case was presented in `src/main/java/net/sf/jabref/gui/BibtexFields.java` (removed in #581):

    // a alternative displayname, e.g. used for
    // "citeseercitationcount"="Popularity"

PR #568, opened 2015-12-22
Fixed #564: Cite command changes are immediately reflected in the pus…

### LINKED ISSUE #564: Cite command seems to have no effect at TeXstudio
I have configured "Cite command" to be `\citep`, but TeXstudio seems to ignore this setting.

grabbed_20151221-235016

I assume that it is impossible to set a cite command for TeXstudio, is it?

simonharrer:
…h-to-application actions, and not only after restart


oscargus:
:+1:

PR #572, opened 2015-12-23
Refactor cleanup code

tobiasdiez:
This PR (again) refactors the cleanup department. The motivation was to extract as much logic as possible from the cleanup action to make it easier to use cleanup operations also at different points (like on save).

Biggest changes:
- Move preferences controlling the cleanup process to own data class `CleanupPreset` and the execution of the cleanup process to `CleanupWorker`. Thus starting a cleanup process is reduced to setting up a proper preset and passing it to the worker.
- Move code displaying and changing a preset to a separate class. Hence it is now easy to show and change a preset at a different place (like in the preferences). 
- Move preferences code to `JabRefPreferences`. Well, this is the place where it belongs.
- Add tests for the `CleanupWorker`. More detailed tests of the different cleanup processes are still missing.

Todo's/Breaking changes: These are all related to the fact that there is no nice interface to get/set the file field of an entry (all the code is in the gui class FileListTableModel).
- The counting of unsuccessful renames does not work anymore (I vote for removing this completely)
- Remove CleanupUtil and incorperate the code in the respective cleanup classes. Currently this code relies on gui code (FileListTableModel) and thus cannot easily be moved to logic.
- Test the file related cleanups.


tobiasdiez:
By the way, does somebody know why the tests fail on the ci server but pass locally? Is there a secretly hidden preference controlling whether the file type is added?

tobiasdiez:
Any comments on this one? Should I rebase?

src/main/java/net/sf/jabref/gui/actions/CleanupAction.java
```diff
@@ -0,0 +1,174 @@
...
+            cancelled = true;
+            return;
+        }
+        frame.block();
+        panel.output(Localization.lang("Doing a cleanup for %0 entries...", Integer.toString(numSelected)));
+    }
+
+    @Override
+    public void run() {
+        if (cancelled) {
```
simonharrer: During the run method, NO swing class must be accessed. You use panels and option panes.

src/main/java/net/sf/jabref/logic/cleanup/RelativePathsCleanup.java
```diff
@@ -0,0 +1,37 @@
...
+    private final List<String> paths;
+
+
+    public RelativePathsCleanup(List<String> paths) {
+        this.paths = paths;
+    }
+
+    @Override
+    public List<FieldChange> cleanup(BibEntry entry) {
+        return CleanupUtil.makePathsRelative(entry, paths.toArray(new String[] {}));
```
simonharrer: Is the method makePathsRelative used more than once? Otherwise it could be inlined here

src/main/java/net/sf/jabref/logic/cleanup/CleanupPreset.java
```diff
@@ -0,0 +1,210 @@
...
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+    You should have received a copy of the GNU General Public License along
+    with this program; if not, write to the Free Software Foundation, Inc.,
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+
+package net.sf.jabref.logic.cleanup;
+
+public class CleanupPreset {
```
simonharrer: I do not get why a CleanupPreset class is necessary. Why not use a CleanupPreferences class which is a thin wrapper around the preferences as this has been done before?

tobiasdiez: The idea was to allow for multiple presets. So for example, there could be a preset for the cleanup upon save action and a completely different preset for the cleanup dialog. Both have to be stored in the preferences...

src/main/java/net/sf/jabref/logic/cleanup/RelativePathsCleanup.java
tobiasdiez: It is not used everywhere else. However it contains Swing code and for this reason could not be inlined here.

src/main/java/net/sf/jabref/gui/actions/CleanupAction.java
tobiasdiez: The method contains nearly the same code as the previous run method. What would be the proper way to do this?

simonharrer: I would move some of the code in the init part and some of it in the update part. You can have a look at the `SearchWorker`, which separate this correctly. The issue is that the run method is always executed outside of the EDT, and the update method always on the EDT. And swing only works correctly if all methods on all swing classes are executed on the EDT. 

This change may require a lot of work. If you think this is too much, we can leave it as is and improve that in another PR.

src/main/java/net/sf/jabref/logic/cleanup/CleanupPreset.java
simonharrer: Ah ok. 

I would not add the flexibility before it is actually used. I would only introduce something like this if the current code does not handle the situation perfect anymore. Do it as simple as possible until you feel the pain - and only then change it. 

To me, it feels more complex to manage a class with state (multiple booleans) in addition to the preferences, which have to be kept in sync. The JabRefPreferences class must also know this CleanupPreset, making the JabRefPreferences class also more complicated to understand as we would introduce another dependency. And the JabRefPreferences class should be kept as simple as possible as it is so core to JabRef - it is used by almost the whole JabRef code, and this code now all depend on the CleanupPreset as well. 

My suggestion is, use a simple wrapper class and when we want to store these values also for onSave and onOpen, we can simply use different versions of the wrapper class.

src/main/java/net/sf/jabref/logic/cleanup/RelativePathsCleanup.java
simonharrer: Hm, ok, not good. We should rewrite the code so that it no longer contains swing code to get the dependencies right. Could be done in another PR.

src/main/java/net/sf/jabref/gui/actions/CleanupAction.java
tobiasdiez: I'll have a look at it tomorrow.

src/main/java/net/sf/jabref/logic/cleanup/CleanupPreset.java
tobiasdiez: Hehe, actually the whole point of the PR was that I felt the pain :). I wanted to implement cleanup actions for onSave but then quickly realized that I would replicate a lot of code. Thus first this refactoring. 

In some way or another the cleanup process has to know a bunch of boolean values and so I figured it would be best to encapsulate them in a simple data class.

simonharrer: :) I agree that the booleans should be encapsulated in a class. This really helps. But I would use a wrapper instead.

tobiasdiez: What do you think about a hybrid wrapper similar to #658 ? That is, still store the boolean values in a class (with "state"), but move the methods to load/save from the preferences also to this new CleanupPreset class. This has the advantage that the Cleanup process has no dependency on the preferences and it brings some performance improvements with it (no 10*number of entries calls to getBoolean).

simonharrer:
- model.FileField.parse
- Preferences rework

src/main/java/net/sf/jabref/gui/actions/CleanupAction.java
tobiasdiez: The run method asks the user if he wants to continue with autogenerating pdf names. I'm not sure how to implement this properly with the Worker interface. Maybe the problem disappears automatically after migration to javafx?

src/main/java/net/sf/jabref/logic/cleanup/RelativePathsCleanup.java
tobiasdiez: It was relatively easy to implement them using the new FileField class! Thank you for pointing that out.

tobiasdiez:
I implemented everything we had discussed yesterday. In particular:
- Remove CleanupUtil class by moving its methods to appropriate cleanup classes (and use FileField instead of FileListTableModel)
- Reimplement counting of unsuccessful renames
- Add tests for rename files and make relative cleanups
- Rework cleanup preferences

However, I didn't make CleanupPreset stateless (i.e. it still contains setter) because I don't want to create a constructor accepting 19 booleans...Any idea how to proceed here?

simonharrer:
Why not use a list of enums, each enum stands for a cleanup that is active aka true.

tobiasdiez:
Because this would be too simple :smiley: . Nice idea :+1:

simonharrer:
Awesome PR. Thanks for the contribution!

PR #575, opened 2015-12-23
Fix SF bug 482 partly: escaped brackets are now properly parsed

tobiasdiez:
Field values as `escaped \{ bracket` are now parsed properly by the BibtexParser. However, there is still a warning about non-matching brackets if the same string is entered directly in the review panel.


koppor:
LGTM :+1:

PR #577, opened 2015-12-24
Made better use of getFileExtension and added tests

oscargus:
* Found code that extracts the extension from a file or a file name and used the already existing method for that
* Added tests for that method


oscargus:
I realized that it makes sense to add a method that returns the FileType directly (as Optional), as this is a quite common case. Hence, I changed to WIP.

koppor:
LGTM :+1:

PR #579, opened 2015-12-24
Implemented BibTexEntryAssert for Lists

boceckts:


boceckts:
I'm using the assertEquals(BibEntry shouldBeEntry, BibEntry entry) that was already present to compare each entry of the lists.

koppor:
Yeah, I saw the next lines, but what is `Assert.assertTrue(expectedEntries.containsAll(actualEntries));` doing?

boceckts:
First I wanted to check if both lists contain the same entries, later I check if the order of the entries is the same. Bu this line is probabbly redundant.

koppor:
How is `same` defined? If `BibEntry.equals()` is used, this will fail.

boceckts:
Yes you are right contains uses equals, it therefore fails.

boceckts:
I don't know why I even wanted to compare a input stream with a list. The methode I need for the medlineplain importer test should compare two lists of bibtex entries...
This compares all entries of the lists with their canonical representation.

tobiasdiez:
I think a `Assert.assertEquals(list one, list two)` already exists in JUnit and does exactly what you implemented in the new `assertEquals`. See http://stackoverflow.com/a/12495556

boceckts:
Yes but it uses the equals methode of the BibEntry. As stated by @koppor above we do not want to use equals but rather check the entries canonical representations for equality.

tobiasdiez:
In my opinion it would be cleaner to overwrite the equals method in `BibEntry` (checking id, type and fields directly) and then use the built-in methods instead of comparing based on `canonicalRepresnation`. @koppor what was the reason to dismiss this option?

simonharrer:
Unknown side effects which need to be investigated time consumingly. Because of this we first wanted to have a solution which can later be switched.

PR #580, opened 2015-12-25
Used a scrollable text area for parser warnings

oscargus:
Now, all warnings when opening a file is shown in a scrollable text area instead of as plain warning text (with all the consequences, try opening a file with ~1000 items without keys...).


tobiasdiez:
:+1:

PR #581, opened 2015-12-25
Small fixes

oscargus:
Was the original purpose at least.

* Code improvements
* Removal of unused variables and in some case propagation of constants to avoid having these as parameters


koppor:
Besides my comments, everything looks good!

oscargus:
Apart from the dropped database, the comments are handled. 

I couldn't resist cleaning up a bit in BibDatabase (committing only two changed comments isn't enough...). But hopefully nothing controversial there.

lenhard:
The changes in `BibDatabase` are reasonable. As is the rest of this PR.

PR #583, opened 2015-12-25
Changes in customized entry types are now directly reflected

oscargus:
When customizing entry types, the visible panel wasn't directly updated on "Apply" or "OK". Now it is. (And a Coverity warning leading to the solution of this is removed.)

It is only the final changed line in `updateTables` that is the solution, the rest is just using nicer methods.


koppor:
LGTM :+1:

PR #584, opened 2015-12-26
BasePanel cleanup

oscargus:
I installed PMD https://github.com/pmd and cleaned up BasePanel based on the issues there. I didn't add `final` to all the cases where it could (partly out of boredom, partly that I wasn't sure if it was appreciated).

Some interesting things emerged though, such as unused GridBagLayouts. 

Positive is that all printStackTrace are removed.

The cases where if-statements are modified, it is basically have the "positive" (i.e. not using !) case first.

### DIFF
--- a/src/main/java/net/sf/jabref/gui/BasePanel.java
+++ b/src/main/java/net/sf/jabref/gui/BasePanel.java
@@ -130,9 +130,6 @@ public class BasePanel extends JPanel implements ClipboardOwner, FileUpdateListe
     private boolean updatedExternally;
     private Charset encoding;
 
-    GridBagLayout gbl = new GridBagLayout();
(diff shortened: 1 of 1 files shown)


lenhard:
Looks good. In my opinion, final does not need to be used everywhere, e.g. for method parameters it does not really matter, but it also does no harm. 

Removals of printStackTrace are always appreciated :-)

oscargus:
Yes, I slowed down a bit on the final parts in later PRs. Now, I've set up the filters on some things I think is relevant and makes an actual difference. :-)

koppor:
LGTM :+1:

Unsure about the `final` declarations. We can also configure Eclipse to add them where appropriate. Found it a bit irritating. Don't know, what the others think.

Issue #593, opened 2015-12-27
JabRef 3.0 cannot insert reference list to LibreOffice 5.0


daniel-dong:
I use the latest master build of JabRef (http://builds.jabref.org/master/JabRef-3.1dev--snapshot--2015-12-23--master--1770f5c.jar) with LibreOffice 5.0 on Ubuntu 15.10 (x86_64).

JabRef can connect to LibreOffice via 'manual connect', and I can insert citations. However, when I click the 'refresh' button on the OpenOffice pane to generate a reference list, only the title "References" is inserted without any entries. Below is the console output:

java.lang.NullPointerException: field value must not be null
	at java.util.Objects.requireNonNull(Objects.java:228)
	at net.sf.jabref.model.entry.BibEntry.setField(BibEntry.java:328)
	at net.sf.jabref.openoffice.OOUtil.insertFullReferenceAtCurrentLocation(OOUtil.java:74)
	at net.sf.jabref.openoffice.OOBibBase.insertFullReferenceAtCursor(OOBibBase.java:1065)
	at net.sf.jabref.openoffice.OOBibBase.populateBibTextSection(OOBibBase.java:1187)
	at net.sf.jabref.openoffice.OOBibBase.rebuildBibTextSection(OOBibBase.java:842)
	at net.sf.jabref.openoffice.OpenOfficePanel$11.actionPerformed(OpenOfficePanel.java:319)
	at javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:2022)
	at javax.swing.AbstractButton$Handler.actionPerformed(AbstractButton.java:2348)
	at javax.swing.DefaultButtonModel.fireActionPerformed(DefaultButtonModel.java:402)
	at javax.swing.DefaultButtonModel.setPressed(DefaultButtonModel.java:259)
	at javax.swing.plaf.basic.BasicButtonListener.mouseReleased(BasicButtonListener.java:252)
	at java.awt.AWTEventMulticaster.mouseReleased(AWTEventMulticaster.java:289)
	at java.awt.Component.processMouseEvent(Component.java:6535)
	at javax.swing.JComponent.processMouseEvent(JComponent.java:3324)
	at java.awt.Component.processEvent(Component.java:6300)
	at java.awt.Container.processEvent(Container.java:2236)
	at java.awt.Component.dispatchEventImpl(Component.java:4891)
	at java.awt.Container.dispatchEventImpl(Container.java:2294)
[...]

PR #594, opened 2015-12-27
Solved #593 and cleaned up the OO-code a bit

oscargus:
Hopefully solved #593.

Try it out at http://builds.jabref.org/ooreferencelist/

(Also cleaned up the code a bit based on PMD. The actual fix is on lines 76-90 in OOUtil.java.)


koppor:
Looks good. Is it possible, that the dependency on "unoil.jar" can be removed?

oscargus:
unOil.jar was in a separate directory before (changing the default paths to be the same for unOil and the other jars, where jurt.jar is the one searched for, was the biggest change to get it to work with the most recent versions of OO and LO).

PR #596, opened 2015-12-27
Replaced explicit strings with constants for all(?) remaining prefere…

oscargus:
…nces


oscargus:
With all mean all except the OpenOffice ones, which are considered in #594

lenhard:
Looks good. This is long overdue and we get more compile-time checking regarding the preferences.

PR #597, opened 2015-12-27
Reverted a fix that lead to new Coverity warnings and more PMD cleanups

oscargus:
Sometimes comparing objects with `==` and `!=` is better than `equals`...

### DIFF
--- a/src/main/java/net/sf/jabref/gui/BasePanel.java
+++ b/src/main/java/net/sf/jabref/gui/BasePanel.java
@@ -123,7 +123,7 @@ public class BasePanel extends JPanel implements ClipboardOwner, FileUpdateListe
 
     private JSplitPane splitPane;
 
-    JabRefFrame frame;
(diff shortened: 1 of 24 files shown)


lenhard:
I think the underlying problem is that we have no good implementation of equals for `BibEntry`. Anyways, this looks fine for me!

oscargus:
In this case it wouldn't have helped as both objects can be null, so "hard"
to figure out which to use equals on anyway (which is why the warning
returned).

I've seen quite a few cases in the code where two objects are operated on
and either one of them may be null. Is there a good way to write this?
Somehow one would like to have four cases, which of course can be achieved
by four `if else`, but is there another way?

lenhard:
If both objects can be null, then there is no way to avoid explicit null checks, I think. If at least one of them was guaranteed to be non-null, a single equals comparision would do. But, alas, that doesn't help here.

There is far too much returning of null in the code...

PR #600, opened 2015-12-28
Avoid ConcurrentModificationException when reading column names

### LINKED ISSUE #599: Import from external database fails for 3.1 and PostgreSQL
Hi,

Just tried out 3.1.  Thank you for all of the enhancements, especially the new look!.

I tried to import from an existing database but it failed with the following error in a pop-up window:

"Could not import from SQL database for the following reason:
java.util.ConcurrentModificationException"

There were no obvious command line messages.  I used ./gradlew run to test.

Cheers,

Fred

lenhard:
Fixes #599

### DIFF
--- a/src/main/java/net/sf/jabref/sql/importer/DBImporter.java
+++ b/src/main/java/net/sf/jabref/sql/importer/DBImporter.java
@@ -104,8 +104,9 @@ public abstract class DBImporter extends DBImporterExporter {
                         rsEntryType.getStatement().close();
                     }
 
-                    List<String> colNames = this.readColumnNames(conn);
-                    for (String next : colNames) {
+                    List<String> allColNames = this.readColumnNames(conn);
+                    List<String> colNames = new ArrayList<>(allColNames.size());
+                    for (String next : allColNames) {
                         if (!columnsNotConsideredForEntries.contains(next)) {
                             colNames.add(next);
                         }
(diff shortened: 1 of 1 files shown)


src/main/java/net/sf/jabref/sql/importer/DBImporter.java
```diff
@@ -104,8 +104,9 @@
                         rsEntryType.getStatement().close();
                     }
 
-                    List<String> colNames = this.readColumnNames(conn);
```
tobiasdiez: Does something like 
`List<String> colNames = this.readColumnNames(conn).stream().filter(!columnsNotConsideredForEntries.contains);`
works?

lenhard:
Sure, a lambda does the job as well. If I recall it correctly, the syntax for your first example should be:
`List<String> colNames = this.readColumnNames(conn).stream().filter(!columnsNotConsideredForEntries::contains);`

But let's first see if this really solves the issue of @fishforbreakfast. http://builds.jabref.org/ seems to be down at the moment, so I cannot provide a jar. Once this is fixed, we'll get back at this issue.

koppor:
@lenhard You can provide a deep link to the jar using CircleCI: https://circle-artifacts.com/gh/JabRef/jabref/2214/artifacts/0/home/ubuntu/jabref/build/releases/JabRef-3.2dev--snapshot--2015-12-28--fix-599--372d06d.jar

I'm working on getting builds.jabref.org up and running again. We migrated to a new server, but the migration did not fully succeed. :disappointed:

lenhard:
Ok, thanks!

lenhard:
As a side note: We forgot the terminal operation. A correct lambda syntax is `List<String> colNames = this.readColumnNames(conn).stream().filter(column -> !columnsNotConsideredForEntries.contains(column)).collect(Collectors.toList());`

PR #601, opened 2015-12-28
Fixes and cleanup for OpenOffice

oscargus:
Revert a minor error introduced in earlier fix.

Further cleanup, primarily getting rid of duplicated code and some Coverity warnings.


koppor:
LGTM. I don't use the plugin and I think, no one else of the other developers do. Therefore, I fully trust you :palm_tree:

Issue #534, opened 2015-12-17
Nothing happens in options dialog "settings for open office"


teertinker:
Got to "Settings"-> "External Programs" click on "Settings for open office"

-> nothing happens. Clicking on the other external programs opens a dialog. Should probably be grayed or something else.

matthiasgeiger:
Confirmed. Another `NullPointerException`...

oscargus:
Which isn't surprising considering the code:

``` java
    @Override
    public JPanel getSettingsPanel() {
        return null;
        /*if (settings == null)
            initSettingsPanel();
        return settings;*/
    }
```

I do not have time to look more into it (as in creating a branch, uncomment, and test) but should be rather straightforward (although I guess it might be commented for a reason...).

matthiasgeiger:
:scream:

stefan-kolb:
@oscargus Is this treated in #602 somehow?

oscargus:
Yes, the button is removed from the preference dialog and there's no `PushToApplication` for OpenOffice anymore (only through the sidepane), so I close it now.

Issue #119, opened 2015-08-21
Add support for CitationStyles


koppor:
As user, I quickly want to copy and paste an entry rendered in a specified format. Either, because I want to paste a reference in a review, I work with Microsoft Word managing the references by hand, or I just quickly send around a reference to users not using BibTeX as serialization format. Currently, JabRef cannot be configured quickly to render Springer or IEEE styled entries: One has to use a combination of HTML and the formatters to get the desired result.

[CitationStyles](http://citationstyles.org/) is the new way to offer rendering of bibliographic data. Currently JabRef offers following two possibilities to format BibTeX entries

 - Through [JabRef's formatters](http://jabref.sourceforge.net/help/CustomExports.php), which are used at the export or at the preview panel
 - Through bibtex (`bst`) and biblatex ([biblatex-styles](https://de.sharelatex.com/learn/Biblatex_citation_styles)) - when using in a latex document

I am aware that another options causes more trouble. Nevertheless, the available formats of CitationStyles is huge (over 7500) and we can't keep maintaining all of them in a JabRef / [OpenOffice style](http://jabref.sourceforge.net/OOPlugin-styles.php) variant.

Possibly, we drop JabRef's formatters? If we decide against support of CitationStyles, we should enable displaying/exporting references formatted by bibtex (and eventually biblatex). As far as I understood, JabRef offers a [bibtex VM](https://github.com/JabRef/jabref/blob/v_2.11b3/src/main/java/net/sf/jabref/bst/VM.java), which interprets `bst` files and generates text output. Meaning: We have a bibtex program implemented in Java, dont' we?

## Background information
* [mapping from BibTeX to CSL](http://www.docear.org/2012/08/08/docear4word-mapping-bibtex-fields-and-types-with-the-citation-style-language/)
* With [citeproc-java](http://michel-kraemer.github.io/citeproc-java/) it should be fairly easy to implement the feature. It also offers a [bibtex importer](http://michel-kraemer.github.io/citeproc-java/using/bibtex-converter/).

mlep:
CitationStyles looks impressive! If programming-wise, this is relevant and efficient, I vote for this feature.

koppor:
With [citeproc-java](http://michel-kraemer.github.io/citeproc-java/) it should be fairly easy to implement the feature. It also offers a [bibtex importer](http://michel-kraemer.github.io/citeproc-java/using/bibtex-converter/).

stefan-kolb:
:+1:

[...]

Issue #110, opened 2015-08-18
Extract libraries out of JabRef ("Toolkit") / Offer JabRef as library / jar / maven central


koppor:
[Bioclipse](http://www.bioclipse.net/) has used JabRef in the past for the BibTeX data model, and there has been talk for reusing JabRef in PathVisio too for working with biobliograph data... do you see any chance of splitting out a JabRef functionality library with core functionality (data model, look up of data based on DOI or PubMed ID, ...)?

Source [message 32103878](https://sourceforge.net/p/jabref/mailman/message/32103878/) from @egonw
- [ ] [org.jabref.model](https://github.com/JabRef/jabref/tree/master/src/main/java/org/jabref/model)
- [ ] [org.jabref.logic](https://github.com/JabRef/jabref/tree/master/src/main/java/org/jabref/logic)
- [ ] XMPUtil (see #266)
- [ ] `net.sf.jabref.bst.VM` (see https://github.com/JabRef/jabref/issues/119#issuecomment-152326777) - refs https://github.com/JabRef/jabref/pull/5702
- [ ] many more (to be defined)
- [ ] Re-include https://github.com/JabRef/jabref/pull/9427/files#diff-7610f76c2991f2f5d8da7a008a7bab7a77d41f7dbfc6ed91145bada44b9cffa6

lenhard:
Yes, I can see a chance for this. Especially because of all the cleanup and class sorting we are doing right now. The data model is now placed into the `model` package and core functionality is moved into `logic`. These two packages could be turned into a JabRef functionality library. 

JabRef itself could benefit a lot from this separation of concerns. Unfortunately, there is still quite some way to go...

lenhard:
As of 86f635c855b83ee94230a72afcab15c7f3cd51e8 the dependencies of the model package are limited to logic classes, classes from the Java API and the infamous global classes (`Globals`, `JabRefPreferences`). This is one step towards extracting a library

stefan-kolb:
@lenhard Didn't you remove the dependency on Globals in `JabRefEntryType` before?! It breaks my tests as I now? need an instance of `JabRefPreferences` in any test.

lenhard:
You mean BibtexEntryType, right?

In that case, no I did not. Neither now nor before. Not sure why JabRefPreferences should be a sudden dependency.

stefan-kolb:
[...]

PR #602, opened 2015-12-28
Major OpenOffice changes

oscargus:
Assuming that the current state of the OpenOffice/LibreOffice connection is good enough (and that no one will improve it significantly in the near future), this PR removes currently unused code.

The major removed thing (from some perspective) is the .bst-support. However, as far as I can tell, it wasn't finished (and therefore not enabled) and while I really think it is a nice feature, I cannot really imagine someone finishing it.

As a second change, the push-to-application button support is removed and only the side panel can be used (which is much more advanced anyway). There are probably further things that can be removed, such as a few preferences related to push-to-application, but that will require further analysis.

This PR builds on #601, so merging that first is probably the best (and this may require some thinking).


oscargus:
Also "solves" #534.

koppor:
I need to investigate further. In #110 I identified the bibtex VM (`net.sf.jabref.bst.VM`), which can handle bsts. Seems to be very complete. I am aware that citastionstyles (see #119) is the successor of `bst` files. I am currently unsure, whether citation styles really cover all cases and if users can easily switch.

In the case of biblatex vs. citationstyles, I am pretty sure that biblatex is more powerful. When reading the documentation of [biblatex-apa](https://www.ctan.org/pkg/biblatex-apa), I don't think that a [621 line XML](https://www.zotero.org/styles/apa) can have the same functionality. But you can convince me of the opposite. - I know that this doesn't help in the bst vs. citation styles case though.

oscargus:
The BST engine is intact (and should probably be used more, e.g. in
generating  author formats). It was just the OO connection  (method
insertBST in OpenOfficePanel (?)) that was removed.

koppor:
Could you say a few words about the expected funtionality behind insertBST or is it a mystery? :innocent:

oscargus:
Well, it's got a hard coded .bst file in it:

``` java
wrapper.loadBstFile(new File("/home/usr/share/texmf-tetex/bibtex/bst/ams/amsalpha.bst"));
```

The other indication of incompleteness is `BstWrapper.java` with the following method:

``` java
public Map<String, String> processEntries(Collection<BibEntry> entries, BibDatabase database) {
        // TODO: how to handle uniquefiers?

        // TODO: need handling of crossrefs?
        String result = vm.run(entries);
        return parseResult(result);
    }
```

In addition, IEEEtran.bst introduces lots of additional LaTeX-commands, so the cleaning provided in `parseResult` will not take you that far. While the idea of using .bst is excellent, in this case I think it is still better to define reference style through the other definition files.

(I've been thinking about generating previews using .bst, but cleaning up the .bbl is a mess in the general, or at least the IEEEtran.bst, case and the same thing is bound to happen here. For previews, I think a feasible way is to generate a PNG/PDF and display it, but that won't really be of any use here.)

oscargus:
Also, there's not even commented out menu items etc or any configuration.

An option would also be to keep the .bst-stuff but get rid of the push-to-application-button (and other commented out things, there are three potential buttons which are not used). Will also simplify things a bit.

oscargus:
Seems like the final commit solves #607 if nothing else.

koppor:
Should we open an issue then which reads "Add bst support for OpenOffice" referencing this PR?

Regarding the previews: I assume, you mean the preview panel? I thought, the BibtexVM can be used as is for that?

oscargus:
Yes, I think it makes sense to open an issue.

Yes, the VM can be used, but for some .bst-files (e.g. IEEEtran.bst) there will be quite a bit of non-trivial latex code to parse...

PR #603, opened 2015-12-29
Removed Coverity warning in bst and cleaned up all files there

oscargus:
Fixed all but one of the possible* Coverity warnings for the bst package.

Also a PMD cleanup of all those files. Please see Github-comments in VM.java and see if you agree with me (the checks pass).

* The warnings for the automatically generated files seems more challenging.

### DIFF
--- a/src/main/java/net/sf/jabref/bst/VM.java
+++ b/src/main/java/net/sf/jabref/bst/VM.java
@@ -17,14 +17,15 @@ package net.sf.jabref.bst;
 
 import java.io.File;
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
+import java.util.List;
 import java.util.ListIterator;
 import java.util.Map;
 import java.util.Stack;
-import java.util.Vector;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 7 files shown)


src/main/java/net/sf/jabref/bst/VM.java
```diff
@@ -168,11 +179,6 @@ public void execute(BstEntry context) {
                     throw new VMException("Can only compare two integers with <");
                 }
```
oscargus: Both o1 and o2 are Integers (otherwise an Exception would have been thrown on line 179), so should be safe to only compare them using `compareTo`.

src/main/java/net/sf/jabref/bst/VM.java
```diff
@@ -144,11 +160,6 @@ public void execute(BstEntry context) {
                     throw new VMException("Can only compare two integers with >");
                 }
```
oscargus: Both o1 and o2 are Integers (otherwise an Exception would have been thrown on line 160), so should be safe to only compare them using `compareTo`.

src/main/java/net/sf/jabref/bst/VM.java
```diff
@@ -197,7 +203,7 @@ public void execute(BstEntry context) {
                     return;
                 }
 
-                if (o1 == o2) {
```
oscargus: The if statement on line 201 checks if one of o1 and o2 are null. If both are non-null line 211 takes care of the comparison, so the only remaining case is if both are null. (Trying to avoid Object == Object.)

simonharrer:
Automatically generated files should not be changed. If they produce warnings, we have to live with it at the moment.

PR #604, opened 2015-12-29
Removed Coverity warning in collab and cleaned up all files there

oscargus:


oscargus:
I need to double-check a few things here...

oscargus:
Should be good to go again.

src/main/java/net/sf/jabref/collab/ChangeScanner.java
```diff
@@ -64,6 +65,7 @@
 
     private static final Log LOGGER = LogFactory.getLog(ChangeScanner.class);
 
+    private final double MATCH_THRESHOLD = 0.4;
```
simonharrer: why not static?

src/main/java/net/sf/jabref/collab/EntryChange.java
```diff
@@ -85,7 +85,7 @@ public boolean makeChange(BasePanel panel, BibDatabase secondary, NamedCompound
         boolean allAccepted = true;
 
         Enumeration<Change> e = children();
-        for (; e.hasMoreElements();) {
+        while (e.hasMoreElements()) {
```
simonharrer: Maybe use a foreach loop there instead?

simonharrer:
:+1: except two minor comments.

src/main/java/net/sf/jabref/collab/ChangeScanner.java
oscargus: No idea. :-)

src/main/java/net/sf/jabref/collab/EntryChange.java
oscargus: Didn't know that one can have a foreach loop for Enumerations (there are these iterator-whiles everywhere in the code), but now I googled and it wasn't that hard...

PR #606, opened 2015-12-29
Logic cleanup

oscargus:
The cleanup continues...

### DIFF
--- a/src/main/java/net/sf/jabref/logic/journals/AbbreviationParser.java
+++ b/src/main/java/net/sf/jabref/logic/journals/AbbreviationParser.java
@@ -72,7 +72,7 @@ public class AbbreviationParser {
     }
 
     private void addLine(String line) {
-        if (line.startsWith("#")) {
+        if (!line.isEmpty() && (line.charAt(0) == '#')) {
             return;
         }
         String[] parts = line.split("=");
--- a/src/main/java/net/sf/jabref/logic/journals/AbbreviationParser.java
+++ b/src/main/java/net/sf/jabref/logic/journals/AbbreviationParser.java
@@ -72,7 +72,7 @@ public class AbbreviationParser {
     }
 
     private void addLine(String line) {
-        if (!line.isEmpty() && (line.charAt(0) == '#')) {
+        if (line.startsWith("#")) {
             return;
         }
         String[] parts = line.split("=");
--- a/src/main/java/net/sf/jabref/logic/net/Cookie.java
+++ b/src/main/java/net/sf/jabref/logic/net/Cookie.java
@@ -23,8 +23,7 @@ class Cookie {
 
     private final String name;
     private final String value;
-    private final URI uri;
-    String domain;
+    private String domain;
     private Date expires;
     private String path;
 
@@ -45,7 +44,6 @@ class Cookie {
     public Cookie(URI uri, String header) {
         String[] attributes = header.split(";");
         String nameValue = attributes[0].trim();
-        this.uri = uri;
         this.name =
                 nameValue.substring(0, nameValue.indexOf('='));
         this.value =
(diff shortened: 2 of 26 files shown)


src/main/java/net/sf/jabref/logic/journals/AbbreviationParser.java
```diff
@@ -72,7 +72,7 @@ private void readJournalList(Reader in) {
     }
 
     private void addLine(String line) {
-        if (line.startsWith("#")) {
+        if (!line.isEmpty() && (line.charAt(0) == '#')) {
```
simonharrer: Why is this better? I do like the previous version much better.

src/main/java/net/sf/jabref/logic/net/Cookie.java
```diff
@@ -123,12 +112,16 @@ public boolean matches(URI uri) {
...
+        if (uriPath == null) {
+            uriPath = "/";
         }
 
-        return path.startsWith(this.path);
+        return uriPath.startsWith(this.path);
+    }
+
+    public boolean equals(Cookie cookie) {
+        return ((domain.equals(cookie.domain)) && (name.equals(cookie.name)));
```
simonharrer: I do not like this equals implementation as only two of the five attributes are used. What is more, when you override equals, you should also override hashcode.

src/main/java/net/sf/jabref/logic/net/Cookie.java
```diff
@@ -123,12 +112,16 @@ public boolean matches(URI uri) {
             return false;
         }
 
-        String path = uri.getPath();
-        if (path == null) {
-            path = "/";
+        String uriPath = uri.getPath();
+        if (uriPath == null) {
+            uriPath = "/";
         }
```
simonharrer: You could use optional here: String uriPath = Optional.ofNullable(uri.getPath()).or("/");

src/main/java/net/sf/jabref/logic/journals/AbbreviationParser.java
oscargus: It may not be better... PMD claims that line.charAt(0) is "better" than startsWith, but in this case the additional line.isEmpty may cancel the benefit.

src/main/java/net/sf/jabref/logic/net/Cookie.java
oscargus: For unknown reasons (as in I didn't bother to dig deeper into it), it seems like equals is not overridden. It may be better to call it something different though. The main purpose was to make the fields private and the only place the fields were accessed was a when checking name and domain for equality.

src/main/java/net/sf/jabref/logic/net/Cookie.java
oscargus: Ahh, nice! I'll think about that when I'm not so tired... :zzz:

simonharrer: :) Good night.

src/main/java/net/sf/jabref/logic/journals/AbbreviationParser.java
simonharrer: I suggest to use startsWith instead, and also in the other places of this PR.

src/main/java/net/sf/jabref/logic/net/Cookie.java
simonharrer: Hm, then maybe name it equalsNameAndDomain to make this clear.

src/test/java/net/sf/jabref/logic/util/strings/StringUtilTest.java
```diff
@@ -94,20 +94,10 @@ public void testJoin() {
     public void testStripBrackets() {
         assertEquals("foo", StringUtil.stripBrackets("[foo]"));
         assertEquals("[foo]", StringUtil.stripBrackets("[[foo]]"));
-        assertEquals("foo", StringUtil.stripBrackets("foo]"));
```
oscargus: Looking at the use case for the function, it seems a bit weird that single square brackets are removed. So I removed these tests as the current implementation will not remove single brackets.

koppor: Could you keep the line as `assertNotEquals`? - The documentation in the tests should be self-contained and not rely on the discussion we do on GitHub ^^

oscargus: I readded similar tests in a later commit.

PR #611, opened 2015-12-31
Cleanup and some logging of importer

oscargus:


src/main/java/net/sf/jabref/importer/CustomImporter.java
```diff
@@ -0,0 +1,137 @@
...
+ * Object with data for a custom importer.
+ *
+ * <p>Is also responsible for instantiating the class loader.</p>
+ */
+public class CustomImporter implements Comparable<CustomImporter> {
+
+    private String name;
+    private String cliId;
+    private String className;
+    private String basePath;
```
simonharrer: Why not make them final?

src/main/java/net/sf/jabref/importer/ZipFileChooser.java
```diff
@@ -283,4 +190,100 @@ public void actionPerformed(ActionEvent e) {
...
     }
+
+
+    /**
+     * Table model for the ZIP archive contents.
+     *
+     * <p>Contains one row for each entry.
+     * Does not contain rows for directory entries.</p>
+     *
+     * <p>The columns contain information about ZIIP file entries:
```
simonharrer: ZIIP typo

src/main/java/net/sf/jabref/importer/fetcher/INSPIREBibtexFilterReader.java
```diff
@@ -71,16 +71,16 @@ private String readpreLine() throws IOException {
         return l;
     }
 
-    private String fixBibkey(String in) {
-        if (in == null) {
+    private String fixBibkey(final String _in) {
```
simonharrer: do not use variables starting with underscore. What does the `_in` variable contain? Is it the bibtex key? Is it the whole bibtex souce of a field?

src/main/java/net/sf/jabref/importer/fetcher/INSPIREBibtexFilterReader.java
```diff
@@ -71,16 +71,16 @@ private String readpreLine() throws IOException {
...
+        if (_in == null) {
             return null;
         }
         //System.out.println(in);
-        if (in.matches("@Article\\{.*,")) {
+        if (_in.matches("@Article\\{.*,")) {
```
simonharrer: As this is a regex, it is better to extract this is a pattern which is then used for the matching. Because then the pattern can be reused and must not be compiled every thing it is used.

src/main/java/net/sf/jabref/importer/fileformat/BiblioscapeImporter.java
```diff
@@ -264,15 +264,13 @@ public boolean isRecognizedFormat(InputStream in) throws IOException {
...
-                            (pages[0] != null ? pages[0] : "")
-                            + (pages[1] != null ? "--" + pages[1] : ""));
+                    hm.put("pages", (pages[0] == null ? "" : pages[0]) + (pages[1] == null ? "" : "--" + pages[1]));
                 }
 
                 // concatenate address and country
                 if (address != null) {
                     hm.put("address", address
-                            + (country != null ? ", " + country : ""));
+ + (country == null ? "" : ", " + country));
```
simonharrer: indention is broken here

src/main/java/net/sf/jabref/importer/fileformat/BibteXMLImporter.java
```diff
@@ -110,7 +110,7 @@ public boolean isRecognizedFormat(InputStream stream) throws IOException {
         } catch (org.xml.sax.SAXException e2) {
             LOGGER.error("Error during XML parsing", e2);
             status.showMessage(e2.getLocalizedMessage());
-        } catch (java.io.IOException e3) {
+        } catch (IOException e3) {
```
simonharrer: why e3? und e2? All exception can be called `e`.

src/main/java/net/sf/jabref/importer/CustomImporter.java
oscargus: Because there are set-methods for them. (And Eclipse does it automatically if possible.)

oscargus:
Fixed comments and a few minor other changes.

src/main/java/net/sf/jabref/importer/OpenDatabaseAction.java
```diff
@@ -261,6 +260,8 @@ private void openTheFile(File file, boolean raisePanel) {
                     }
 
                 }
+
```
tobiasdiez: This is not used?!

tobiasdiez:
LGTM

src/main/java/net/sf/jabref/importer/CustomImporter.java
simonharrer: Follow up question: are the set methods used? Can the code be rewritten so that they are not needed anymore? The class is much simpler to use when you know that the fields do not change. And to me the class looks like it can be achieved without too much effort.

oscargus: I can have a look at it. I do not think I checked if they were used (sometimes I check all public methods to see if they are used somewhere).

src/main/java/net/sf/jabref/importer/CustomImporter.java
```diff
@@ -0,0 +1,137 @@
...
+
+
+    public CustomImporter() {
+        super();
+    }
+
```
oscargus: Also, this String array constructor may be reevaluated.

src/main/java/net/sf/jabref/importer/CustomImporter.java
oscargus: It wasn't that easy it turned out (at least not for me). The reason is that the imported class is loaded within the class and the result of that is used to set the remaining fields.

Issue #393, opened 2015-11-26
Jabref rendering problems with GTK Look and Feel on Linux and OpenJDK


nswitte:
My system:
Ubuntu 15.10

Linux version 4.2.0-18-generic (buildd@lgw01-38) (gcc version 5.2.1 20151010 (Ubuntu 5.2.1-22ubuntu2) ) #22-Ubuntu SMP Fri Nov 6 18:25:50 UTC 2015 (Ubuntu 4.2.0-18.22-generic 4.2.3) 

java version "1.7.0_91"
OpenJDK Runtime Environment (IcedTea 2.6.3) (7u91-2.6.3-0ubuntu0.15.10.1)
OpenJDK 64-Bit Server VM (build 24.91-b01, mixed mode)

All three versions of Jabref hang (after starting up OK, reading the database and selecting an entry), say when editing a file link or some other operation. Once this happens all buttons are inactive and I have to kill the process. This started today after Java got upgraded, so maybe this has something to do with it. Previously this was not a problem. I have looked at the FAQ but it doesn't seem relevant.

Any solutions would be really handy as I am pretty reliant of Jabref to manage my bibliographies.

Thanks, Nicholas

koppor:
We did not change any code in JabRef 2.10 since its [release  in March 2014](http://jabref.sourceforge.net/revisionhistory.php#v2.10). Therefore, this should be an JRE issue. Could you downgrade your JRE? As soon as you know the JRE version where JabRef runs, could you report this as an issue to the [ubuntu bug tracker](https://bugs.launchpad.net/ubuntu/+source/jabref)?

Meanwhile, you can try Java8 and [the latest build](http://builds.jabref.org/master/). I am afraid, that this won't succeed, too.

The only thing, that currently helps is to upgrade to Oracle's JRE using the [oracle-java8-installer](https://launchpad.net/~webupd8team/+archive/ubuntu/java).

matthiasgeiger:
At least 2.11 should also work with OpenJDK 8 - see #168.

pviotti:
As reported above, I too use Ubuntu 15.10, and I have tried both the version in the repository (2.10) and the last stable version available (2.11.1).

 * **Version 2.10 using java 1.7 (openjdk)**: cannot test it because even if I change the default java from 1.8 to 1.7 system wide, it uses java 1.8 (this should probably be a separate bug..)
[...]

PR #616, opened 2016-01-01
Test both OpenJDK8 and OracleJDK8

koppor:
We should support both testing OpenJDK and OracleJDK to avoid incompatibilities because of using different JREs. This helps Linux users to run JabRef without any issues (see #393).

Currently, neither TravisCI (https://docs.travis-ci.com/user/languages/java/#Overview) nor CircleCI (https://circleci.com/docs/environment#java) officially support OpenJDK8.

Switching to OpenJDK at CircleCI seemed to be easer in CircleCI than in TravisCI. Therefore, I switched in CircleCI.

In my opinion, this switch shouldn't affect anything as we do not rely to internal classes of the OracleJDK.

Side comment: Oracle offers a scanner for such internal usages: https://wiki.openjdk.java.net/display/JDK8/Java+Dependency+Analysis+Tool

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,11 +1,16 @@
 machine:
   java:
+    # We switch to OpenJDK8 at dependencies/pre
     version: oraclejdk8
 
 dependencies:
   pre:
     - scripts/prepare-install4j.sh
     - install4j6/bin/install4jc --verbose --license=$INSTALL4J_KEY
+    # Switch to OpenJDK8 as described at https://discuss.circleci.com/t/using-openjdk-8/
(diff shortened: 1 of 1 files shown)


simonharrer:
The issues are mainly related to GUI concerns, hence, the switch currently would have no effect on the test results. I would postpone this until we have GUI tests in place.

koppor:
As mentioned at https://github.com/travis-ci/travis-ci/issues/2968#issuecomment-149164058, we could switch to [Trusty beta](https://docs.travis-ci.com/user/trusty-ci-environment/).

simonharrer:
https://travis-ci.org/JabRef/jabref/jobs/101143676 hm, does not work with trusty beta.

PR #619, opened 2016-01-01
Extracted the object array to its own class

oscargus:
@koppor something like this?  (Also a StringBuffer instead of String-concatenation.)

### DIFF
--- a/src/main/java/net/sf/jabref/sql/importer/DBImporterResult.java
+++ b/src/main/java/net/sf/jabref/sql/importer/DBImporterResult.java
@@ -21,14 +21,26 @@ import net.sf.jabref.model.database.BibDatabase;
 
 public class DBImporterResult {
 
-    public final BibDatabase database;
-    public final MetaData metaData;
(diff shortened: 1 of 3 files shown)


src/main/java/net/sf/jabref/sql/importer/DBImporterResult.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package net.sf.jabref.sql.importer;
+
+import net.sf.jabref.MetaData;
+import net.sf.jabref.model.database.BibDatabase;
+
+public class DBImporterResult {
+
+    public final BibDatabase database;
```
tobiasdiez: Make them private and generate getter methods.

oscargus:
I was considering that (and can of course do it). However, as a very temporary object I thought it was a bit overkill. But easily done. Will fix it.

oscargus:
Didn't know that Eclipse sorted everything out in just a few clicks. Nice!

Issue #608, opened 2015-12-30
Export - error stream closed


melTr:
Hi,
On version 3.1, I'm unable to export. I try either via cut-paste or via file menu. I try many formats (some of my own or those by default). I'm running Ubuntu 14.04, with Java 8.
I get this exception :

```
+1: AVERTISSEMENT: [ResolvePDF]
déc. 30, 2015 4:50:29 PM net.sf.jabref.exporter.ExportFormat performExport
AVERTISSEMENT: The following formatters could not be found: ResolvePDF, ResolvePDF
java.io.IOException: Stream closed
	at sun.nio.cs.StreamEncoder.ensureOpen(StreamEncoder.java:45)
	at sun.nio.cs.StreamEncoder.flush(StreamEncoder.java:140)
	at java.io.OutputStreamWriter.flush(OutputStreamWriter.java:229)
	at net.sf.jabref.exporter.ExportFormat.finalizeSaveSession(ExportFormat.java:365)
	at net.sf.jabref.exporter.ExportFormat.performExport(ExportFormat.java:304)
	at net.sf.jabref.exporter.ExportFormats$1ExportAction$1.run(ExportFormats.java:230)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:497)
	at spin.Invocation.evaluate(Invocation.java:175)
	at spin.off.SpinOffEvaluator$1.run(SpinOffEvaluator.java:108)
	at java.lang.Thread.run(Thread.java:745)
```

oscargus:
It seems like the exporter ResolvePDF is missing. When browsing the source, I can not find any references to it. I might have gone missing when the pdf-field was made obsolete (and should therefore be removed from the export formats).

Ferry128:
Hi, 

I have the same problem but without the export format missing:

[Fatal Error] :1:1: Content is not allowed in prolog.
java.io.IOException: Stream closed
	at sun.nio.cs.StreamEncoder.ensureOpen(StreamEncoder.java:45)
	at sun.nio.cs.StreamEncoder.flush(StreamEncoder.java:140)
	at java.io.OutputStreamWriter.flush(OutputStreamWriter.java:229)
[...]

Issue #417, opened 2015-11-30
Grouping: Switching from one group to another does not update table


matthiasgeiger:
Main table is not updated on selection of another group.

Workaround for users: Show "All entries" first before selecting the other group.

PR #633, opened 2016-01-03
Fix for issue #417

nivieru:
Removed checks that prevented the table from being updated when switching between groups.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ to [sourceforge feature requests](https://sourceforge.net/p/jabref/features/) by
 - ACM fetcher works again, fixes #545
 - Use default file icon for custom external file types, fixes #598
 - Fixed #608: Export works again
+- Table now updates when switching groups, fixes #417
 
 ### Removed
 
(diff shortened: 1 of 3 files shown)


simonharrer:
Thank you for the contribution. Can you please indent the lines correctly which were in the then-block of the removed if-condition? Apart from that, this can be merged in from my point of view.

simonharrer:
Hm, would like to merge it, but github says it has conflicts. Can you have a look into this, please?

oscargus:
Probably ChangeLog-conflicts...

oscargus:
By the way, when merging the ChangeLog (if that is the one with the conflict), note that the entry starting `Fixed #608` is actually the correct format. My bad for starting the new version with a made up format (but I'll take care of it eventually, a PR fixing most of it is in #602).

nivieru:
Ok, I think it should be fine now

oscargus:
I thought I'd give it a try to see how merging works. :-) Hope I didn't break anything...

simonharrer:
:+1:

PR #641, opened 2016-01-04
Refactor journal abbreviations and rewrite tests

tobiasdiez:
Injecting the preferences via the constructor and thus making it easy to the class via Mockito.

### DIFF
--- a/src/main/java/net/sf/jabref/exporter/IExportFormat.java
+++ b/src/main/java/net/sf/jabref/exporter/IExportFormat.java
@@ -15,13 +15,12 @@
 */
 package net.sf.jabref.exporter;
 
-import net.sf.jabref.model.database.BibDatabase;
 import net.sf.jabref.MetaData;
-
-import java.nio.charset.Charset;
-import java.util.Set;
+import net.sf.jabref.model.database.BibDatabase;
 
 import javax.swing.filechooser.FileFilter;
+import java.nio.charset.Charset;
+import java.util.Set;
 
 public interface IExportFormat {
 
@@ -57,9 +56,7 @@ public interface IExportFormat {
      *            should be exported. If null, all entries will be exported.
      * @throws Exception
      */
-    void performExport(BibDatabase database, MetaData metaData,
- String file, Charset encoding,
- Set<String> entryIds)
+    void performExport(BibDatabase database, MetaData metaData, String file, Charset encoding, Set<String> entryIds)
             throws Exception;
 
 }
--- a/src/main/java/net/sf/jabref/Globals.java
+++ b/src/main/java/net/sf/jabref/Globals.java
@@ -52,7 +52,6 @@ public class Globals {
     public static final String SELECTOR_META_PREFIX = "selector_";
     public static final String PROTECTED_FLAG_META = "protectedFlag";
     public static final String NONE = "_non__";
-    public static final String FORMATTER_PACKAGE = "net.sf.jabref.exporter.layout.format.";
 
     // In the main program, this field is initialized in JabRef.java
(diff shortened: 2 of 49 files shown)


oscargus:
The `JournalAbbreviator` formatter could borrow (or "inherit") some of those tests while you're at it, increasing the coverage three more lines. :-)

(To me, with limited review knowledge, it looks good!)

simonharrer:
The thing is that each time an abbreviation is required, the whole abbreviations have to be read from the disk. When using an abbreviation formatter in the output which is used for each entry, this can lead to a significant performance impact.

tobiasdiez:
Ok, then the abbreviation loader has to be passed down the whole dependency tree....
I don't like that the main frame is the central object providing access to the loader, but, well, it was the only omnipresent class. 

By the way, I also removed the singleton instance in `EntryFetchers`. I think it is ok to init them every time they are needed.

simonharrer:
Hmmm, the formatters are loaded by Class.forName. Because of this, Eclipse/IntelliJ cannot detect that they are used, but they are used. And another requirement: the need a default constructor. Because of this, your solution does not work. You still need a global variable for the abbreviations which is referenced from this formatter. 

Maybe we should add a test that every class which implements the formatter interface has a default constructor available. Could be done through reflection maybe.

tobiasdiez:
Now the `JournalAbbreviator` is no longer loaded via `Class.forName()`. So now the approach via dependency injection does work but there were quite a few places using layouts...

So now I know why JabRef has so many globals and statics... removing one (which I thought would be a small change) resulted in 400+ line changes...bah

tobiasdiez:
As discussed, a new global variable for the abbreviation loader is introduced (only used by GUI code). The logic & exporter code still uses dependency injection.

Rebased & ready to be merged

src/main/java/net/sf/jabref/exporter/IExportFormat.java
```diff
@@ -57,8 +58,7 @@
      *            should be exported. If null, all entries will be exported.
      * @throws Exception
      */
-    void performExport(BibDatabase database, MetaData metaData,
- String file, Charset encoding,
-            Set<String> entryIds) throws Exception;
+    void performExport(BibDatabase database, MetaData metaData, String file, Charset encoding, Set<String> entryIds,
+            JournalAbbreviationRepository repository) throws Exception;
```
simonharrer: Is this still required for the exporter?

tobiasdiez: Yes, the ExportFormat class uses the Layout class which itself needs the JournalAbbreviation. Thus also the interface has to know about it... I don't like it since also all the other exports have now a dependency on JournalAbbreviation. Any idea how to solve it? Pass the JournalAbbreviationLoader to the constructor of ExportFormat?

simonharrer: As neither Exporter nor Layouter are part of the logic package yet, I would not pass the loader around there.

tobiasdiez: With this attitude they will never accomplish to be in the logic department :)
But you are right, for now this is the easier solution.

simonharrer:
Hm, can you have a look at the unused imports? I think there are a lot of them right now.

tobiasdiez:
Yes your are right, there were quite a few unused imports.
Removed them now and rebased.

tobiasdiez:
Should I rebase and merge it in or do you have any other remarks?

simonharrer:
:+1: from my side. Rebase and you can merge this in directly.

Issue #647, opened 2016-01-07
copy formatted


Olifair:
It would be great if one could (right-click on a reference) copy a reference formatted in the selected style. This is something Mendeley does and it is very helpful, for example for sharing references with people who do not necessarily want the complete Bibtex reference.
It would be nice if you could consider this
thanks in advance

Oliver

oscargus:
Something like what is shown in the entry preview? I guess a menu entry "Copy preview" there could make sense.

Olifair:
Yes, something like that would be great, but style would also be great ;-)

oscargus:
What do you mean by style? Using a .bst?

oscargus:
(Better not close it, if you want it to happen. :-))

Olifair:
I'm new to GitHub, used the wrong button :-)
I'm talking about the style that is chose when for example connecting to LibreOffice.

oscargus:
Ah, OK! That's a third style (I think). :-)

oscargus:
I implemented copy of the preview in #648. Not exactly what you asked for, but a start and maybe substitute.

Olifair:
Wow great, you are really rapid. I think copying the preview is largely sufficient. To get the formatted citation one can always use as a workaround a connection to LL. Thanks again.

oscargus:
I think it is possible to copy the layout from the LL style to the preview (more or less, there are probably a few more tweaks available in the LL style and it can be adapted for different entry types).

teertinker:
Isn't this what "Export to clipboard" after right-click does? It opens a window, where you can choose the style you need.

Olifair:
No, these are export formats, not LL styles

oscargus:
Oh, really. Didn't know about that feature. :-) I don't think that the "Copy preview" hurts, but that one is clearly useful for a similar purpose as well!

Olifair:
As far as I can see most of the export filters (haven't tested all) copy a reference formatted with tags, symbols etc in order to import into another program. So Copy preview is really quite different.

koppor:
Refs #119 - when citation styles are supported, the entry preview should be more nicely formatted.

teertinker:
@Olifair I think "Export to clipboard" is almost exactly the thing you asked for. There are export filters. Some of them are created to import it into another program, other were created to build bibliographies according to a certain paper style (MIS Quarterly, Harvard RTF...). See the description at: http://jabref.sourceforge.net/resources.php#export_filters 
[...]

PR #648, opened 2016-01-07
Partly implemented #647

oscargus:
Added the possibility to copy the preview panel text. Not exactly what is requested in #647, but probably quite useful.

Note that the key binding is not working (nor is the print preview key working) so I'll open an issue for that.


stefan-kolb:
I would change the changelog entry to just describe the added functionality without refering to the other issue as it suggest another approach in my opinion.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ to [sourceforge feature requests](https://sourceforge.net/p/jabref/features/) by
 
 ### Changed
 - All import/open database warnings are now shown in a scrolling text area
+- Implemented #647 (partly): The preview can now be copied
```
tobiasdiez: Since Olifair in #647 in content with the `Copy of preview` (and the other suggestions are now tracked at #119), I would just remove the "partly" and close #647 after merging this PR.

tobiasdiez:
LGTM :+1:

src/main/java/net/sf/jabref/gui/PreviewPanel.java
```diff
@@ -422,4 +431,21 @@ public void actionPerformed(ActionEvent e) {
 
     }
 
+    class CopyAction extends AbstractAction {
```
simonharrer: Can you name it CopyPreviewAction as we have a lot of other copy actions there as well.

simonharrer:
Please rebase, then this can be merged in.

oscargus:
Fixed and rebased

PR #652, opened 2016-01-07
Replaced getStringArray and putStringArray with getStringList and putStringList in preferences

oscargus:
Also fixed an NPE in the XMP metadata preference tab.

### DIFF
--- a/src/main/java/net/sf/jabref/exporter/CustomExportList.java
+++ b/src/main/java/net/sf/jabref/exporter/CustomExportList.java
@@ -126,7 +126,7 @@ public class CustomExportList {
 
     private void purge(int from) {
         int i = from;
-        while (Globals.prefs.getStringArray(JabRefPreferences.CUSTOM_EXPORT_FORMAT + i) != null) {
+        while (!Globals.prefs.getStringList(JabRefPreferences.CUSTOM_EXPORT_FORMAT + i).isEmpty()) {
             Globals.prefs.remove(JabRefPreferences.CUSTOM_EXPORT_FORMAT + i);
             i++;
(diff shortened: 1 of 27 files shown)


src/main/java/net/sf/jabref/exporter/CustomExportList.java
```diff
@@ -69,12 +71,12 @@ private void readPrefs() {
...
         int i = 0;
-        String[] s;
-        while ((s = Globals.prefs.getStringArray(JabRefPreferences.CUSTOM_EXPORT_FORMAT + i)) != null) {
+        List<String> s;
+        while (!((s = Globals.prefs.getStringList(JabRefPreferences.CUSTOM_EXPORT_FORMAT + i)).isEmpty())) {
             Optional<ExportFormat> format = createFormat(s);
             if (format.isPresent()) {
                 formats.put(format.get().getConsoleName(), format.get());
-                list.add(s);
+                list.add(s.toArray(new String[s.size()]));
```
simonharrer: you can use addAll

simonharrer:
:+1: less nulls :)

src/main/java/net/sf/jabref/exporter/CustomExportList.java
oscargus: I do not think so. I do not want to add the elements but the array. `EventList<String[]>`. Or am I missing something?

simonharrer: Ah, sorry. You are right. But why are arrays stored in a list? But this is probably due for another PR. ;)

PR #657, opened 2016-01-09
English localization: correction of a glitch

mlep:


simonharrer:
In the english translation, the key and the value must always be the same. To fix the failing tests, you need to change the key as well for all language files and the Localization.lang usage of this key in the Java source code.

mlep:
OK. Well, I believe it is better for the project I do not touch the java code. Do you want me to submit an issue?

simonharrer:
Please go ahead and change the code directly in here. We will review it to make sure nothing is broken. :)

koppor:
I am doing it currently with Notepad++. "Find in Files." I will commit directly in master as this a minor issue

Refs https://github.com/koppor/jabref/issues/73

koppor:
Blindly replacing "bibtex" by "BibTeX" in `*.java` and `*.properties` cannot be done. Each change has to be reviewed manually. Although `100%` look good in the English language file, some languages such as Brazilian seem to like to write everything in lower case:

    no_base-BibTeX-file_specified=nenhum_arquivo_bibtex_de_base_especificado

PR #658, opened 2016-01-09
Refactor proxy registration and preferences

tobiasdiez:
Moves code from `JabRef` to own class and introduces new class encapsulating the proxy preferences.

### DIFF
--- a/src/main/java/net/sf/jabref/JabRef.java
+++ b/src/main/java/net/sf/jabref/JabRef.java
@@ -40,6 +40,8 @@ import net.sf.jabref.gui.*;
 import net.sf.jabref.importer.fetcher.EntryFetcher;
 import net.sf.jabref.importer.fetcher.EntryFetchers;
 import net.sf.jabref.logic.CustomEntryTypesManager;
+import net.sf.jabref.logic.ProxyPreferences;
+import net.sf.jabref.logic.ProxyRegisterer;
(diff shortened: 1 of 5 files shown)


oscargus:
In the spirit of moving things from JabRefPreferences, wouldn't it make sense to let the ProxyPreferences class access the preferences directly?

So: `ProxyPreferences proxyPrefs = new ProxyPreferences()` and `proxyPrefs.save()` (possibly with a JabRefPreferences as an argument).

tobiasdiez:
Yes you are right, this would also be a possibility. 

What I aim at in the end is to reduce the public interface of the `JabRefPreferences` class to like 20 methods which store/retrieve only complex types. In this way also the serialization is hidden in the class and the caller doesn't have to worry how to convert to string/boolean etc. Then it would be also very simple to replace the whole class with something completely different, for example writing the preferences as a json file.

tobiasdiez:
I now moved some code from the JabRefPreferences class to the new ProxyPreferences class as @oscargus suggested. Looks indeed cleaner.

simonharrer:
- no state in ProxyPreferences
- IDEA: logic and model should not know JabRefPreferences
- JabRefPreferences must not know ProxyPreferences; defaults and keys in JabRefPrefences

tobiasdiez:
Done & rebased.

PR #663, opened 2016-01-11
Fixed some issues in exporter

### LINKED ISSUE #627: Remove references to ResolvePDF in the default exports
With the removed pdf field, the export formatter ResolvePDF is removed, although it still is referenced in the export files.

oscargus:
Fixed an issue in the MS Office 2007 exporter and some code cleanups.

Removed all(?) references to the pdf field and the non-existing ResolvePDF export formatter.

Removed duplicate lines in the ChangeLog for 3.2, fixes #627.


CHANGELOG.md
```diff
@@ -15,9 +15,10 @@ to [sourceforge feature requests](https://sourceforge.net/p/jabref/features/) by
 ### Changed
 
 ### Fixed
+- MS Office 2007 export is working again
 
 ### Removed
-
+- Fixed #627: The pdf field is removed from the export formats, use the file field
```
tobiasdiez: I think the new format is now `Fixed [#num](link to github)`

CHANGELOG.md
```diff
@@ -45,9 +46,6 @@ to [sourceforge feature requests](https://sourceforge.net/p/jabref/features/) by
 
 ### Fixed
 - Changes in customized entry types are now directly reflected in the table when clicking "Apply" or "OK"
-- Fixed [#593](https://github.com/JabRef/jabref/issues/593): Reference list generation works for OpenOffice/LibreOffice again
```
tobiasdiez: Why did you deleted this entries?

src/main/java/net/sf/jabref/exporter/MSBibExportFormat.java
```diff
@@ -61,13 +61,9 @@ public void performExport(final BibDatabase database, final MetaData metaData, f
             } catch (Exception e) {
```
tobiasdiez: Can this catch block also be merged with the next one?

src/main/java/net/sf/jabref/exporter/MSBibExportFormat.java
```diff
@@ -61,13 +61,9 @@ public void performExport(final BibDatabase database, final MetaData metaData, f
...
                 throw new Error(e);
             }
-        }
-        try {
             finalizeSaveSession(ss);
-        } catch (SaveException ex) {
+        } catch (Exception ex) {
```
tobiasdiez: Is it possible to catch only a few specific exceptions then everything?

tobiasdiez:
Just a few minor remarks, otherwise :+1:

CHANGELOG.md
oscargus: Because there are identical ones further down (sort of my bad that they showed up there, as I changed the format of them, but somehow some other PR managed to restore the old ones...).

src/main/java/net/sf/jabref/exporter/MSBibExportFormat.java
oscargus: Probably. Not sure why an Error is thrown here though and if there's a practical difference.

src/main/java/net/sf/jabref/exporter/MSBibExportFormat.java
oscargus: Sure. However, as the same thing was done both for IOException and Exception, it seemed to make more sense to only catch Exception (which, of course, is bad in itself). Not sure why the code looks like it does and which exceptions make sense to catch.

CHANGELOG.md
oscargus: OK! Wasn't sure if that was something that was edited before a release only.

oscargus:
Edited the Changelog (and added the issue that reported the thing I fixed). The exception things I do not really know what to do about. Of course the exception handling can be improved further, but compared to the previous code, this is at least a bit improved.

oscargus:
Rebased again. Anything else that needs to be fixed? I know that the Exception handling is not brilliant, but nothing I introduced (it is probably slightly more brilliant now though...).

Issue #669, opened 2016-01-13
JabRef 3.1/3.2 writes bib file in a format that it will not read


LonelyProf:
If the @PREAMBLE field is present, JabRef 3.1/3.2 omits the newline between that line and the preceding Encoding line when saving the file. The resulting file is not readable by JabRef. Sample attached. Problem seems to have been introduced at v3.1.

[test.txt](https://github.com/JabRef/jabref/files/89065/test.txt)

lenhard:
related #621

stefan-kolb:
@tobiasdiez Is this fixed?

tobiasdiez:
@stefan-kolb It will be fixed with #718

stefan-kolb:
Too late in the evening, should have seen that myself :smile: :+1:

Issue #621, opened 2016-01-02
Ver 3.1 comment not written correctly.


zl5877761:
I have a file started with bibtex item, e.g.,

1  @ARTICLE{XXXXXX,
2      author = {XXXXXXXX},
3  .........
4  }

when I use 3.1 to save the database, one line break between encoding comment and items is missing, e.g.,

1  % Encoding: UTF-8@ARTICLE{XXXXXX,
2      author = {XXXXXXXX},
3  .........
4  }

please fix it.

lenhard:
This is not a bug, but the exact desired behavior.

We only modify entries that have been changed by the user, because there was a lot of complaining about this functionality. Every unchanged entry is written back exactly as it was read. So if there was no line break before your entry when loading the file, there will be no line break when saving it back.

You can "fix" your database yourself extremly easy. Just add a line break.

koppor:
@lenhard Sorry to disagree here. When opening a valid bib file in JabRef and saving it, it should not get invalid. I am aware that we should not add additional newlines at each save  (which IMHO would happen if we just wrote a newline after the encoding comment). I nevertheless wonder, why we don't write a newline after that comment :innocent:.

Does this relate to https://github.com/koppor/jabref/issues/114?

lenhard:
Ok, I see. The encoding parser throws an exception (which is utterly unrelated to the entry itself).

Unfortunately, there is no good way to fix this. Any optional newline writing will depend on a lot of conditions (first entry unchanged, etc. etc.) and will be quite hacky. I don't have time to do this, so a contribution from anybody else would be more than welcome.

Alternatively: Can't we just eliminate the file header completely? I though it wasn't used anyway?

tobiasdiez:
The easiest way would be to normalize all new lines in the file (i.e don't save empty lines in the parsed serialization). But I suspect there is a reason why newlines are stored, right?

The file header is used while opening a file (not on importing one - but this is more a bug than a feature).

lenhard:
Correct, if we do not store the newlines, then we end up reformatting the file on save. Avoiding this was the whole point. (And newlines are the most tricky part that leads to most of the effort regarding file format...).

[...]

PR #677, opened 2016-01-16
Make BibTex parser more robust against missing newlines

tobiasdiez:
As described in #669 and #621 the parser has some problems when an entry (or preamble) follows directly the file encoding header (on the same line), i.e. 
`% Encoding: myEncoding @article{...}`.
Actually, as the tests reveal it was not a problem with the actual parser but within the file opening logic. 
I don't view this PR as a complete fix for #669 and #621 since not just the parser but also the serialization should be improved.

Also fix a bug that prevented the import of BibTex entries having only a key as its content.

### DIFF
--- a/src/main/java/net/sf/jabref/importer/OpenDatabaseAction.java
+++ b/src/main/java/net/sf/jabref/importer/OpenDatabaseAction.java
@@ -62,7 +62,6 @@ public class OpenDatabaseAction extends MnemonicAwareAction {
     // upgrade actions etc. that may depend on the JabRef version that wrote the file:
     private static final List<PostOpenAction> POST_OPEN_ACTIONS = new ArrayList<>();
 
-
     static {
         // Add the action for checking for new custom entry types loaded from
         // the bib file:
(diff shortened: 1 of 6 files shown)


src/main/java/net/sf/jabref/importer/OpenDatabaseAction.java
```diff
@@ -447,7 +447,14 @@ private static InputStreamReader openFile(File fileToOpen, Optional<Charset> enc
...
                 } else if (line.startsWith(Globals.encPrefix)) {
                     // Line starts with "Encoding: ", so the rest of the line should contain the name of the encoding
-                    String encoding = line.substring(Globals.encPrefix.length());
+                    // Except if there is already a @ symbol signalising the starting of a BibEntry
+                    Integer atSymbolIndex = line.indexOf("@");
+                    String encoding;
+                    if(atSymbolIndex > 0)
+                        encoding = line.substring(Globals.encPrefix.length(), atSymbolIndex);
+                    else
+                        encoding = line.substring(Globals.encPrefix.length());
```
simonharrer: Please always use curly braces as this makes the semantics explicit.

simonharrer:
:+1: (except my minor comment)

tobiasdiez:
Corrected... for some reason the code style was not imported correctly in Intellij.

simonharrer:
Hm, I just use the default Intellij code style. :)

lenhard:
@tobiasdiez: what do you think should be done to fix the above issues completely?

tobiasdiez:
@lenhard In my opinion JabRef should add new lines after the encoding and preamble. I'm working on a solution right now (including a 'bit' of refactoring to make testing easier).

lenhard:
@tobiasdiez Ok, good to hear! I'll concentrate on other issues then.

PR #681, opened 2016-01-19
Update JabRef_de.properties

teertinker:
Just shortening one German string in search bar. The new string equals the one used in Libreoffice search dialog.


tobiasdiez:
:+1:

Issue #667, opened 2016-01-12
hash/pound sign causes URL link to fail


wujastyk:
Entering a URL into the URL field.  If the URL has a # in it, then when on the main table page, and clicking the url/doi link, the # is translated into %23 (or something like) and the URL connection fails.

Using Jabref 3.1 on Linux Mint.  Everything up to date.

PR #683, opened 2016-01-19
Remove URL sanitizer as it is malfunctioning and unclear where needed…

stefan-kolb:
Also see #667.

@JabRef/developers If someone knows a use case where this is needed and why we can try to reimplement it properly. But this way it really makes no sense IMHO.

### DIFF
--- a/src/main/java/net/sf/jabref/gui/desktop/JabRefDesktop.java
+++ b/src/main/java/net/sf/jabref/gui/desktop/JabRefDesktop.java
@@ -71,8 +71,6 @@ public class JabRefDesktop {
         } else if ("eprint".equals(fieldName)) {
             fieldName = "url";
 
-            link = URLUtil.sanitizeUrl(link);
-
             // Check to see if link field already contains a well formated URL
             if (!link.startsWith("http://")) {
                 link = Util.ARXIV_LOOKUP_PREFIX + link;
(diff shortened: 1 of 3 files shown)


oscargus:
So then either `URLDecoder.decode` or `URI.toASCIIstring` converts `#` to `%23`? Of course not impossible, but seems like one has used the incorrect methods in that case.

I think the main things it does is replacing space with %20 in the URLs and removing any `\url{...}` commands. There will probably be users that will run into problems, but maybe one can add a URL cleanup that does the corresponding thing?

I think an alternative solution would just be to do a `replaceAll("%23","#")` at the end.

stefan-kolb:
The problem is that is really hard to correctly encode a URL if we are not aware of its structure. Regardless of what methods we use.
So if we are not completely sure we need it, we should probably not do this.

simonharrer:
:+1: sounds reasonable.

simonharrer:
@koppor what do you think about this? Especially is it possible that users have a `url={\url{http://google.com}}` in their bibfile?

koppor:
Unsure. The method comment states following:

> make sure an URL is "portable", in that it doesn't contain bad characters that break the open command in some  OSes. A call to this method will also remove \\url{} enclosings.

That reads perfectly fine.

OK, I also read the following

> // FIXME: everything below is really flawed atm

Maybe, we should cleanup the URL fields and not do the cleanup at the place of opening. Although I see the issue of malicious bib files which could have an URL executing malicious code. However, that risk is so low that I would not count on it.

Meaning: :+1: and open a new issue for CleanUp.

Issue #271, opened 2015-10-27
GUI flashes when resizing


lenhard:
In its current state on master, the GUI window of JabRef flashes when resizing the window. This effect is quite annoying (at least to my eye).  It's hard to capture the flash in screenshot, but to reproduce it, it is sufficient to start up the GUI and resize the window by dragging at its borders.

OS: Win 7
JRE: Oracle 1.8.0_60

simonharrer:
This is a regression as this issue is not present in JabRef 2.10.

simonharrer:
And it is not present in the current dev2.11 branch.

simonharrer:
After doing `git bisect` and running `rm -r src/main/gen/ && gradlew clean run` for each commit the culprit seems to be f24233f86c8d26a95a1aa2d563ac1a0051949d1b

lenhard:
That was an awesome roundtrip time for a bug fix. Great work!

simonharrer:
:trophy:

PR #688, opened 2016-01-21
Updated versions for pdfbox, fontbox, and jempbox

oscargus:
Updated dependencies.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -71,9 +71,9 @@ dependencies {
 
     compile 'org.swinglabs:swingx:1.6.1' // do not update, 1.6.5.1 is broken
 
-    compile 'org.apache.pdfbox:pdfbox:1.8.10'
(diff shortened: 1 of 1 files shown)


koppor:
I think, minor version updates can always be directly made in the master branch. Also with generally mayor version updates, but this refs #271 and #507.

I think, the [PdfContentImporter](https://github.com/JabRef/jabref/blob/master/src/main/java/net/sf/jabref/importer/fileformat/PdfContentImporter.java) is the only component using these libraries, which is still [untested](https://codecov.io/github/JabRef/jabref/src/main/java/net/sf/jabref/importer/fileformat/PdfContentImporter.java?ref=f53cb380ef8f1ae7bd5cf4a159ddc7b02634e14b).

oscargus:
Yeah, I just wasn't sure how to not mess up, so I thought this was a safer way. :-)

PR #691, opened 2016-01-22
Cleanup in aux import

oscargus:
Cleanup and warning removal.


simonharrer:
Did you test whether the functionality is still working?

oscargus:
Yes, it is working. I'm thinking that maybe one should move the files into logic and gui, respectively?

oscargus:
I have not been able to find nested .aux-files to try on, but generating a .bib-file based on a larger .bib-file and an .aux-file works well.

PR #692, opened 2016-01-22
Cleanup in model

oscargus:


src/main/java/net/sf/jabref/model/entry/AuthorList.java
```diff
@@ -1250,16 +1250,17 @@ public String getLastFirst(boolean abbr) {
          * von Last, Jr.' (if <CODE>abbr==true</CODE>)
          */
         public String getFirstLast(boolean abbr) {
-            String res = getLastOnly();
+            StringBuffer res = new StringBuffer();
+            getLastOnly();
```
tobiasdiez: I think this should  be res.append(getLastOnly) ?

oscargus: Actually, it should be removed (as it is appended later). Just kept it in the code to remember what to write later and forgot to remove it... :-)

Thanks! Will update in a while.

PR #693, opened 2016-01-22
Cleanup in logic.mods and logic.msbib

oscargus:
Additional change: added method `hasField` to `BibEntry`.

### DIFF
--- a/src/main/java/net/sf/jabref/logic/msbib/MSBibEntry.java
+++ b/src/main/java/net/sf/jabref/logic/msbib/MSBibEntry.java
@@ -632,19 +632,6 @@ class MSBibEntry {
         return result;
     }
 
-    /* construct a MSBib date object */
-    protected String getDate(BibEntry bibtex) {
-        StringBuffer result = new StringBuffer();
-        if (bibtex.hasField("year")) {
-            result.append(bibtex.getField("year"));
-        }
-        if (bibtex.hasField("month")) {
-            result.append('-').append(bibtex.getField("month"));
-        }
-
-        return result.toString();
-    }
-
     private String getMSBibSourceType(BibEntry bibtex) {
         String bibtexType = bibtex.getType().getName();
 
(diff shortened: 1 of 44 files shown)


src/main/java/net/sf/jabref/logic/msbib/MSBibEntry.java
```diff
@@ -620,28 +621,28 @@ private void getAuthors(Element authorsElem, String bcol) {
...
         }
         return result;
     }
 
     /* construct a MSBib date object */
     protected String getDate(BibEntry bibtex) {
-        String result = "";
-        if (bibtex.getField("year") != null) {
-            result += bibtex.getField("year");
+        StringBuffer result = new StringBuffer();
```
tobiasdiez: I think StringBuilder is enough here (no mulit-threading required)

tobiasdiez:
:+1: 
One could also consider to return an optional from getField but this probably requires more changes.

oscargus:
I was also thinking about that. Would it make sense to add two more methods:
``` java
Optional<String> getOptionalField(String field)
String getDefaultField(String field, String defaultReturnValue)
```

These would make the code look pretty much nicer in quite a few places.

src/main/java/net/sf/jabref/logic/msbib/MSBibEntry.java
oscargus: I confuse those... I'll have a look!

oscargus:
Ended up adding `getFieldOptional` as the default value can be easily obtained through `orElse`...

koppor:
Why keeping both `getField` and `getFieldOptional`? Why not really replacing the return type of `getField`? That would make `BibEntry.java` consistent within itself. (And I'd remove the `hasField` method, too).

oscargus:
Because getField is used in 600 places... Also hasField should be faster than getField and in enough places one is actually just looking for that information.

koppor:
I'll discuss this at JabCon. :runner:

oscargus:
Do that. There may be refinements to the PR in the meantime. You can only
convert that many getField requests per day...

simonharrer:
Long term: getField should return optional and getFieldOptional should be removed  as this causes confusion with the getOptionalFields
Short term: merge this PR as is

PR #699, opened 2016-01-25
Cleanup in gui

oscargus:
Lots of minor cleanup in gui.


simonharrer:
Can you do such cleanup directly in master? I think this makes more sense, as we do not need to review every small change and improvement. :+1: :smile:

oscargus:
Sure. Hopefully I won't mess up... Or I do it in a two-step approach as it is easier to double-check the changes here and good to run all tests. But I can take care of the merging myself.

simonharrer:
:+1: :trophy:

PR #700, opened 2016-01-25
Cleanup in util

oscargus:


src/main/java/net/sf/jabref/logic/util/io/FileHistory.java
```diff
@@ -49,9 +49,9 @@ public int size() {
 
     public void newFile(String filename) {
         history.remove(filename);
-        history.addFirst(filename);
+        ((LinkedList<String>) history).addFirst(filename);
         while (history.size() > HISTORY_SIZE) {
-            history.removeLast();
+            ((LinkedList<String>) history).removeLast();
```
simonharrer: Hm, this looks strange.

simonharrer:
Hm, it does not compile.

oscargus:
Yeah, fixed now...

src/main/java/net/sf/jabref/logic/util/io/FileHistory.java
oscargus: Indeed. Should one keep the definition as a LinkedList then? (I have to cast it to get access to removeLast().)

simonharrer:
:+1:

src/main/java/net/sf/jabref/logic/util/io/FileHistory.java
simonharrer: In that case, I would leave it at LinkedList as we make explicit use of this data structure.

PR #703, opened 2016-01-26
Remove session functionality

stefan-kolb:
Open tabs are saved by default. Could be extended to Workspace functionality later on but should be removed for now as it provides no benefits at the moment.


simonharrer:
:+1:

PR #709, opened 2016-01-26
replaced some e.printstacktrace

chochreiner:


oscargus:
Thanks for your contribution! Two quick things:

1) Please create a separate branch for each pull request instead of working in your master branch.

2) Always include the exception in the log command, e.g., `LOGGER.error("File could not be opened", e);` In that way there's no need to do any `getMessage()`.

chochreiner:
@1) I'll use feature branches for the next pull request.

@2) should be fixed with the new commit

Issue #520, opened 2015-12-15
Move file sorting to button


koppor:
Currently, JabRef offers sort-at-save. This should be changed to sort-on-demand.

The sort functionality should be moved to a sort button, where following popup appears: 

![grabbed_20151215-175228](https://cloud.githubusercontent.com/assets/1366654/11817302/9ddef6d0-a354-11e5-96e0-0507e54b3344.png)

 - [ ] Add button "Sort and Save" which sorts the file accordingly and directly saves it
 - [ ] The popup stores its last selected options
 - [ ] "Descending" changes to "Ascending" and is selected as default
 - [ ] Enable per-database sorting options

oscargus:
I think it also can make sense to provide multi-column sorting for the main table in an easier way than changing the default sort order in the preferences.

koppor:
The current main table can do a multi sort if you use `CTRL` when selecting a table heading. Is that what you search for?

oscargus:
Might be! It seems like it works as expected anyway. However, the feedback is quite limited so really hard to know exactly which criteria is active. :-)

lenhard:
Currently, if we resort the file, the entries themselves are not reformatted (but the parsed serialization is used). Maybe we should format all entries on resort? I think this would be appropriate for the new "sort on demand"-behavior.

tobiasdiez:
I rather like the "sort on save" feature. Could it not happily coexists with the sort & save button?

matthiasgeiger:
We are planning to introduce configurable "save actions" (such as cleanups, sorting, etc.) - but the default behavior should be: change as little as possible when reading, editing and saving a file in order to allow sharing a *.bib database in groups and using version control.

tobiasdiez:
One suggestion about the auto-sort (since judging from #614 I'm not the only one which likes this feature): As far as I have understood, the main problem with auto-sort is that a bib file (under revision control) is constantly changed if the sort preferences differ from user to user. This issue can be circumvented by making the sort preferences file-based (instead of global) and serialize them in the file. In this way all users would sort the file in the same way.

koppor:
@tobiasdiez This is https://sourceforge.net/p/jabref/feature-requests/892/, I close it and point to here.

lenhard:
#711 solves this. We decided to skip the implementation of a manual sort button, since auto-sorting is still possible. Sorting options are now stored locally in each file and can be configured for each file. This makes it much easier to share files and sort them in the same fashion.

PR #711, opened 2016-01-26
Remove global sorting functionality

lenhard:
Refs #520 

Global file sorting options are removed. Sort functionality is still available but no localized to every file and the file sorting options are stored in the file.


tobiasdiez:
@lenhard: I have already implemented the other half of #520 (Make sorting configurable in database properties & write it to file). Code is coming soon...

lenhard:
@tobiasdiez: that feature has always been available :)

PR #717, opened 2016-01-26
Renaming

chochreiner:
fixed Naming of JabRef and BibTex in some keys as well as their translations


chochreiner:
will be reopened

oscargus:
I think it should be BibTeX if we should be really consistent:

https://en.wikipedia.org/wiki/BibTeX

PR #718, opened 2016-01-26
Improve database writer

### LINKED ISSUE #621: Ver 3.1 comment not written correctly.

tobiasdiez:
The aim of this PR is to improve the database writer, in particular fix #621 and #669. 
Also add tests to the writer and parser (and do some refactorings to make test writing easier).

Changes (to serialization):
- Fix #621 and #669 by adding new lines after encoding and preamble. The idea is that everything is terminated by a new line and the parsed serialization only safes newlines which separate it from the previous entry. 
- Comments and preamble are serialized with capitalized first letter, i.e. `@Comment` instead of `@comment` and `@Preamble` instead of `@PREAMBLE`. This should follow the convention for entries.
- There is no space anymore between `@String` and the bracket and after the bracket, i.e. `@String{name = {string}}` instead of `@String { name = {string} }`
- Implemented [#756](https://github.com/JabRef/jabref/issues/756): Add possibility to reformat all entries on save (under Preferences, File)

Refactorings:
- Renamed FileActions to BibDatabaseWriter
- Merged code for export and save in BibDatabaseWriter
- Extracted SaveSettings class from FileActions to own file and extended it (& renamed to SavePreferences). Note that this class has some setters (against the recently established convention for preference classes) since there are quite a few cases where the values from the preferences have to be overriden. 
- Change usage of arrays to list in many places
- Remove unused boolean parameters checkSearch, checkGroup
- Moved writing of metadata and CustomEntryTypesManager to BibDatabaseWriter
- Extracted and reused gui code for displaying and editing sort preferences (SaveOrderConfigDisplay)


simonharrer:
Could you rebase this on master as both pre-requisite PRs are already merged in. So this could be completed now.

tobiasdiez:
One question: currently the save methods take parameters `boolean checkSearch, boolean checkGroup` which specify that only entries satisfying the active search / group selection are exported. As far as I can see, these boolean parameters are always set to false (which means all entries are exported). Should I remove both parameters?

oscargus:
I guess that if this is merged in first, it should be quite OK to rebase #762 (although many changes are redundant).

Regarding your question: it would be nice to be able to save like this, but I guess it is another of JabRefs almost finished/commented out and then removed features... Considering that one can save selected entries and it is quite easy to see which entries are in the active group/search selection, I'd say remove.

tobiasdiez:
I think the code is now mature enough to be reviewed. I added some notes above in the first post. Sorry that this PR is rather big.

src/main/java/net/sf/jabref/BibDatabaseContext.java
```diff
@@ -32,8 +32,6 @@ public BibDatabaseContext(BibDatabase database, MetaData metaData, Defaults defa
         this.defaults = Objects.requireNonNull(defaults);
         this.database = Objects.requireNonNull(database);
         this.metaData = Objects.requireNonNull(metaData);
-
-        this.setMode(getMode());
```
simonharrer: Why did you delete this? What happens if you load a bib file and save it afterwards - is there a comment about the type stored in the file then still?

CHANGELOG.md
```diff
@@ -26,6 +27,7 @@ to [sourceforge feature requests](https://sourceforge.net/p/jabref/features/) by
 
 
 ### Fixed
+- Fixed [#621](https://github.com/JabRef/jabref/issues/621) and [#666](https://github.com/JabRef/jabref/issues/669) #669: Encoding and preamble now end with newline.
```
oscargus: 666/669

src/main/java/net/sf/jabref/BibDatabaseContext.java
tobiasdiez: I removed the line because otherwise the biblatex mode would be written every time even if it is the default one. In addition, I found it irritating that a constructor changed values in the parameters passed to it (just using a metadata in a specific context shouldn't change it values).

But yes, the biblatex mode is correctly written. I added a test for it now.

CHANGELOG.md
tobiasdiez: Thanks. Corrected.

tobiasdiez:
I also pushed it to the branch [JabRef/databaseWriter](https://github.com/JabRef/jabref/tree/databaseWriter) so that a build should be available under http://builds.jabref.org/.

simonharrer:
Do we have the "only update if it changed" functionality for the preamble, strings and comments as well?

tobiasdiez:
For strings and the epilogue yes, for the preamble kind of (the text in the preamble is stored and exactly rewritten, but the preamble tag and newlines around it are not stored). For the metadata there is no such functionality.

koppor:
@tobiasdiez Regarding the two parameters (https://github.com/JabRef/jabref/pull/718#issuecomment-178262570): Aren't they used for "Save selected as plain BibTeX..."?

tobiasdiez:
@koppor No it is not used for this function. "Save selected..." gets all the selected items and pass them to the save function. The code in question reads
```java
if (selectedOnly) {
                session = FileActions.savePartOfDatabase(
                        panel.getBibDatabaseContext(),
                        file, Globals.prefs,
                        panel.getSelectedEntries(), encoding, FileActions.DatabaseSaveType.DEFAULT);
            } else {
                session = FileActions.saveDatabase(panel.getBibDatabaseContext(),
                        file, Globals.prefs, false, false, encoding, false);
            }
```

simonharrer:
Please rebase. I think this should be merged. 

One suggestion regarding the SavePreferences: You could create a new instance after the call of a setter. This would result in the benefits of having final variables and it still allows to create derivations of this. 

```
SavePreferences prefs = new SavePreferences(Globals.prefs);
SavePreferences otherInstance = prefs.withEncoding("utf8");
assert prefs != otherInstance && !prefs.equals(otherInstance)
```

tobiasdiez:
Rebased and implemented the `withEncoding` suggestion (i.e. removed all setters in the preferences...you really don't like them, right? :smile: )

jkomelj:
In my case (JabRef 3.2 on 32- and 64-bit Windows the problem is not fixed. Original lines
% Encoding: windows-1250

@PREAMBLE{{\providecommand{\noopsort}[1]{}}}   (or @Preamble ...)

are always changed to
% Encoding: windows-1250@PREAMBLE{{\providecommand{\noopsort}[1]{}}}

after saving changed bib file.

Kind regards,
Janez Komelj

tobiasdiez:
@jkomelj You have to use the [latest development version](http://builds.jabref.org/master/) or wait for the release of 3.3 (probably in 1-2 weeks)

jkomelj:
Thank you

Issue #705, opened 2016-01-26
Keyword separator


oscargus:
It is possible to specify keyword separator for the automatically generated groups. However, the methods in BibEntry assumes that "," is used.

If we should keep the possibility to specify keywords separator, it should be used consistently and probably not be specified in the Groups preference tab. Biblatex specify "," as the default separator.

simonharrer:
Good point.

oscargus:
I tried to implement it and it turns out that automatic keyword grouping can be done for any field, not just the `keywords` field. Is it worthwhile keeping this? (Most definitely someone have defined their own magic fields...)

koppor:
Preference setting:

![grabbed_20160126-163648](https://cloud.githubusercontent.com/assets/1366654/12585148/01c9868a-c44b-11e5-9ce8-3bd02306a26a.png)

`net.sf.jabref.model.entry.EntryUtil.getSeparatedKeywords(String)`  uses `net.sf.jabref.model.entry.EntryUtil.SEPARATING_CHARS_NOSPACE`. `net.sf.jabref.model.entry.BibEntry.putKeywords(List<String>)` always uses `,`. Therefore, I would also vote to remove that configuration option and double check whether the group logic can use `EntryUtil.getSeparatedKeywords(String)`, too.

Regarding `keywords` vs. other field names. The usecase is that the a user wants too keep the keywords retrieved from an external database separate from his own keywords. And therefore, he uses a new field `labels` (or similar).

oscargus:
I realize that #719 is not completely done yet as `getSeparatedKeywords(String)` is not changed. The group logic can use these methods, but the undo requires a bit of additional thinking...

koppor:
Regarding undo/redo: I think, it is OK if we don't care in this case.

ambro2:
I had frustrating experiences in trying to use the dynamic groups functionality because of the comma as a keyword separator. In chemistry it not unusual to find keywords containing commas, like "1,2-diol" or "1,3-dipolar cycloaddition" or "α,β-unsaturated carbonyl".
Actually the keyword separator is the group of characters "," and " ". Indeed on importing from a RIS file, the keywords are combined in the bibtex "keywords" field, separated by ", ".
I could not find a way to tell JabRef to consider a _comma not followed by a space_ as part of the keyword.
Maybe there are users who managed to set a different keyword separator (semicolon, pipe or other character) and have their "chemical" keywords handled correctly. I gave up.

koppor:
[...]

Issue #628, opened 2016-01-03
Feature: Hierarchical Keywords


marc-reichenbach:
Hi,

I'm using jabref quite a long time and I'm very satisfied with it. Thanks for your work. Now, while I'm writing my Phd thesis I miss one very important feature: Hierarchical Keywords

I store in bibtex my keywords as: group1|subgroup1|subsubgroup1, group2|subgroup2, group3, group4|subgroup4

where (,) is the delimiter of the keywords and (|) the delimiter of hierarchical levels. 

In the current version, I can only parse regarding (,) resulting in very long keywords (which cannot be displayed as a whole). I would be very happy if somebody can implement this feature. What do you think about that? I'm willing to pay a donation, if somebody could implement it, because I have to submit the thesis in 3 month and the approx. 300 papers in my bibtex drives me to hell :)

Regards,

Marc

simonharrer:
You can use groups which can be put into a hierarchy. This should solve your issue.

koppor:
Marc, could you elaborate more on the concrete desired functionality? Would you be willing to join [JabCon](http://jabcon.jabref.org/), where we could work on that?

marc-reichenbach:
Thank you very much for your anwser and time.

@simonharrer: Thanks, for your recommondation: that will work, but then I have to create the groups by hand. I like the usage of standard file formats (like *.bib) and i like the intelligent functionality of creating groups by parsing keywords.

@koppor: The short anwser is: like XMP will do that for fotos. There also exisits hierarchies like: landscape|animal|bird|parrot

The long anwser is: Beginning with scientific work (first 3 years), I started with jabref and using the keywords by IEEExplor, ACM, etc. But then, I mentioned that the keywords have noting to do with the personal point of view, evaluating a paper. Therefore, I switch to docear (2 years usage). Using the mindmap, I have exactly the structured view as I wish. The problem is, a mind map with 300+ papers is not very clear, moreover I need to use links to have a paper in 2 "categories". Therefore, I switched back to jabref to combine hierachies and keywords together. As I mentioned, I was inspired by XMP and image. There I heavyly use hierarchy keywords. Like the example above: a parrot is a bird, a bird is an animal, etc. I understand, that this functionality can also be applied by using 4 single keywords than using 1 hierachical: but I think it makes life easier to combine these both approachs.

[...]

PR #719, opened 2016-01-26
Removed preference setting

oscargus:
Solves #705. It can be discussed if the setting in the auto generate group dialog should be removed as well.


oscargus:
I think that there are good arguments to have a configurable separator and even have two different separators, one "fundamental" and one for hierarchical keywords, see #628 and #705. Therefore, I close this PR.

PR #726, opened 2016-01-27
Merge from DOI now honors removed fields

oscargus:
Earlier the field wasn't removed when "None" was selected in the Merge from DOI dialog.


simonharrer:
:+1:

PR #728, opened 2016-01-28
Remove PDF preview functionality

stefan-kolb:
I suggest to remove the PDF preview functionality in the preview panel

Why?
- The preview is not working all the time
- the preview is too small to provide valuable information
- opening the linked PDF via a simple click on the PDF icon is way faster and more convenient

@JabRef/developers WDYT?


simonharrer:
:+1: makes sense.

oscargus:
Would probably be better to integrate a proper PDF viewer in that case.

I must say though that on my 2560x1440 screen, the preview is sort of OK sized.

oscargus:
What about adding a ChangeLog entry?

matthiasgeiger:
@stefan-kolb @oscargus is right: Should be added to the CHANGELOG

stefan-kolb:
Is already done https://github.com/JabRef/jabref/blob/master/CHANGELOG.md

matthiasgeiger:
Oh... :astonished:

Issue #738, opened 2016-01-28
URLDownload vs. getResults


oscargus:
There are currently two ways to fetch text from the internet: `URLDownload` and `getResults` (now in `logic.net.NetUtil`. Any particular reason to have two ways? Currently, it seems like `URLDownload` doesn't support post, but is "nicer" for standard get.

oscargus:
getResults is no longer needed with #739.

stefan-kolb:
Probably both classes are poorly implemented. I can at least confirm that for URLDown load which is...

oscargus:
I'd say the static getResults methods were probably worse then... Anyway, now URLDownload supports additional parameters and posting, so slightly better and at least only one home made solution for that. 

Not sure if one either should convert URLDownload to use Unirest or convert the use directly?

oscargus:
I checked the code a bit more and there are basically three ways now:

* Unirest
Simple, especially if you want json etc.

* URLDownload
Simple to get a String or download to a file

* Directly accessing the URLConnection class
Good to get a reader to be used for BibtexParser

Not sure if it is worthwhile unifying further?

simonharrer:
I think this is enough.

PR #739, opened 2016-01-28
Use URLDownload instead of getResults for fetching stuff

oscargus:
Relates to #738 

Fixed some things in GVKParser to hopefully get rid of the final Coverity warnings there.


oscargus:
Not sure why the tests fail. Although it is related to internet connections, I cannot really see that it has anything to do with the things I've modified since those fetches rely on Unirest and not URLDownload...

oscargus:
Seems like the tests pass now (without changing anything). I removed some Coverity warnings as well so that I just didn't repush the same stuff twice.

PR #744, opened 2016-01-29
Move edit custom entry types to bibtex menu as it is directly bibtex-related

stefan-kolb:
@JabRef/developers WDYT?


simonharrer:
:+1: makes sense

PR #746, opened 2016-01-29
A few less nulls

oscargus:


src/main/java/net/sf/jabref/specialfields/SpecialFieldValue.java
```diff
@@ -86,8 +88,8 @@ public JLabel createLabel() {
         return label;
     }
 
-    public String getFieldValue() {
-        return this.keyword;
+    public Optional<String> getFieldValue() {
+        return Optional.ofNullable(this.keyword);
```
oscargus: Why is getFieldValue returning keyword and not field?

simonharrer: @koppor maybe you can shed some light on this?

PR #755, opened 2016-01-31
Fix NPE at saving does not reset save flag #741

### LINKED ISSUE #741: NPE at saving does not reset save flag
I've got an NPE during the save (have to investigate)

```
%d [%thread] %-5level %logger - %msg%n java.lang.NullPointerException
        at net.sf.jabref.exporter.FileActions.saveDatabase(FileActions.java:232)
 ~[JabRef-dev-fat.jar:?]
```

When I try close JabRef, following dialog appears:

grabbed_20160129-050058

After clicking "Discard changes", following dialog appears:

grabbed_20160129-050124

A click on "cancel" returns to JabRef. I have to kill JabRef. This should not happen - or is it an edge case which will never appear again?

tobiasdiez:
Fix #741 (and a lot of other null pointer exceptions). getType and getStandardType now return an optional instead of possible null values.

### DIFF
--- a/src/main/java/net/sf/jabref/exporter/FileActions.java
+++ b/src/main/java/net/sf/jabref/exporter/FileActions.java
@@ -224,11 +224,12 @@ public class FileActions {
 
                 // Check if we must write the type definition for this
                 // entry, as well. Our criterion is that all non-standard
-                // types (*not* customized standard types) must be written.
-                EntryType entryType = EntryTypes.getType(entry.getType(), bibDatabaseContext.getMode());
-
-                if (EntryTypes.getStandardType(entryType.getName(), bibDatabaseContext.getMode()) == null) {
-                    types.put(entryType.getName(), entryType);
+                // types (*not* all customized standard types) must be written.
+                if (! EntryTypes.getStandardType(entry.getType(), bibDatabaseContext.getMode()).isPresent()) {
+                    // If user-defined entry type, then add it
+                    // Otherwise (getType returns empty optional) it is a completely unknown entry type, so ignore it
+                    EntryTypes.getType(entry.getType(), bibDatabaseContext.getMode()).ifPresent(
+                            entryType -> types.put(entryType.getName(), entryType));
                 }
 
                 // Check if the entry should be written.
@@ -428,10 +429,12 @@ public class FileActions {
 
                 // Check if we must write the type definition for this
                 // entry, as well. Our criterion is that all non-standard
-                // types (*not* customized standard types) must be written.
-                EntryType tp = EntryTypes.getType(be.getType(), bibDatabaseContext.getMode());
-                if (EntryTypes.getStandardType(tp.getName(), bibDatabaseContext.getMode()) == null) {
-                    types.put(tp.getName(), tp);
+                // types (*not* all customized standard types) must be written.
+                if (! EntryTypes.getStandardType(be.getType(), bibDatabaseContext.getMode()).isPresent()) {
+                    // If user-defined entry type, then add it
+                    // Otherwise (getType returns empty optional) it is a completely unknown entry type, so ignore it
+                    EntryTypes.getType(be.getType(), bibDatabaseContext.getMode()).ifPresent(
+                            entryType -> types.put(entryType.getName(), entryType));
                 }
 
                 bibtexEntryWriter.write(be, fw, bibDatabaseContext.getMode());
--- a/src/main/java/net/sf/jabref/model/entry/TypedBibEntry.java
+++ b/src/main/java/net/sf/jabref/model/entry/TypedBibEntry.java
@@ -26,16 +26,16 @@ public class TypedBibEntry {
 
     private final BibEntry entry;
     private final Optional<BibDatabase> database;
-    private final BibDatabaseMode type;
+    private final BibDatabaseMode mode;
 
-    public TypedBibEntry(BibEntry entry, BibDatabaseMode type) {
-        this(entry, Optional.empty(), type);
+    public TypedBibEntry(BibEntry entry, BibDatabaseMode mode) {
+        this(entry, Optional.empty(), mode);
     }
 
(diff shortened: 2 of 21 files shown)


src/main/java/net/sf/jabref/exporter/FileActions.java
```diff
@@ -224,11 +224,12 @@ public static SaveSession saveDatabase(BibDatabaseContext bibDatabaseContext, Fi
 
                 // Check if we must write the type definition for this
                 // entry, as well. Our criterion is that all non-standard
-                // types (*not* customized standard types) must be written.
-                EntryType entryType = EntryTypes.getType(entry.getType(), bibDatabaseContext.getMode());
-
-                if (EntryTypes.getStandardType(entryType.getName(), bibDatabaseContext.getMode()) == null) {
-                    types.put(entryType.getName(), entryType);
+                // types (*not* all customized standard types) must be written.
+                if (! EntryTypes.getStandardType(entry.getType(), bibDatabaseContext.getMode()).isPresent()) {
```
oscargus: No need for double check, so you can skip this one and only use ifPresent.

src/main/java/net/sf/jabref/exporter/FileActions.java
```diff
@@ -428,10 +429,12 @@ public static SaveSession savePartOfDatabase(BibDatabaseContext bibDatabaseConte
 
                 // Check if we must write the type definition for this
                 // entry, as well. Our criterion is that all non-standard
-                // types (*not* customized standard types) must be written.
-                EntryType tp = EntryTypes.getType(be.getType(), bibDatabaseContext.getMode());
-                if (EntryTypes.getStandardType(tp.getName(), bibDatabaseContext.getMode()) == null) {
-                    types.put(tp.getName(), tp);
+                // types (*not* all customized standard types) must be written.
+                if (! EntryTypes.getStandardType(be.getType(), bibDatabaseContext.getMode()).isPresent()) {
```
oscargus: Hmm, maybe almost the same here, but I do not really get the logic. First checking StandardType and then type?

src/main/java/net/sf/jabref/exporter/FileActions.java
oscargus: Sorry, I was wrong, will remove the comment....

oscargus:
Looks good to me! :+1:

src/main/java/net/sf/jabref/wizard/text/gui/TextInputDialog.java
```diff
@@ -479,9 +482,13 @@ private void updateSourceView() {
     }
 
     private String[] getAllFields() {
-        List<String> texFields = EntryTypes.getType(entry.getType(), frame.getCurrentBasePanel().getBibDatabaseContext().getMode()).getAllFields();
+        List<String> texFields = new ArrayList<>();
+        Optional<EntryType> type = EntryTypes.getType(entry.getType(),
+                frame.getCurrentBasePanel().getBibDatabaseContext().getMode());
+        if (type.isPresent()) {
+            texFields = type.get().getAllFields();
```
oscargus: You will need to do `texFields.addAll(type.get().getAllFields());` as the List returned from getAllFields can not be edited. I fixed this bug in the old version in #760 (which will collide with this PR...).

tobiasdiez: Good point! I edited the PR accordingly. So you can just remove the corresponding statement in #760 and both PRs should be merged in without problems.

oscargus: Not sure how I do that in practice. Rebasing once this is merged seems to be easier. :-)

src/main/java/net/sf/jabref/model/entry/TypedBibEntry.java
```diff
@@ -43,17 +43,21 @@ public TypedBibEntry(BibEntry entry, Optional<BibDatabase> database, BibDatabase
      * complete.
      */
     public boolean hasAllRequiredFields() {
-        EntryType type = EntryTypes.getType(entry.getType(), this.type);
-        return entry.allFieldsPresent(type.getRequiredFields(), database.orElse(null));
+        Optional<EntryType> type = EntryTypes.getType(entry.getType(), this.mode);
+        if(type.isPresent()) {
```
stefan-kolb: Maybe we should add a test for this branch evaluation.

src/main/java/net/sf/jabref/model/entry/TypedBibEntry.java
```diff
@@ -43,17 +43,21 @@ public TypedBibEntry(BibEntry entry, Optional<BibDatabase> database, BibDatabase
...
+            return entry.allFieldsPresent(type.get().getRequiredFields(), database.orElse(null));
+        } else {
+            return true;
```
stefan-kolb: See above.

src/main/java/net/sf/jabref/model/entry/TypedBibEntry.java
```diff
@@ -43,17 +43,21 @@ public TypedBibEntry(BibEntry entry, Optional<BibDatabase> database, BibDatabase
...
 
     /**
      * Gets the display name for the type of the entry.
      */
     public String getTypeForDisplay() {
-        EntryType entryType = EntryTypes.getType(entry.getType(), type);
-        if (entryType == null) {
-            return "";
+        Optional<EntryType> entryType = EntryTypes.getType(entry.getType(), mode);
+        if (entryType.isPresent()) {
```
stefan-kolb: See above.

stefan-kolb:
Would be cool if we can increase the test coverage, especially for the newly created branches. Otherwise LGTM :+1:

tobiasdiez:
I added tests for all uncovered branches in TypedBibEntry and after two LGTM will merge this in. @oscargus have fun rebasing :smile:

PR #758, opened 2016-02-01
Made all static regex patterns private static final fields

oscargus:
Should not be worse than the previous solution...


stefan-kolb:
LGTM :+1:

PR #760, opened 2016-02-01
Code improvements

oscargus:
Some refactoring, an odd bug, and some cleanups.


stefan-kolb:
For me there are too many things done inside one PR. Cleanups, even fixing a bug! It is hard to check whether it is ok. If you are sure stuff is alright (renaming a variable, smaller stuff) just commit it to master and create a PR for the stuff that other people should take a look at (e.g. the bug).

src/main/java/net/sf/jabref/sql/DBStrings.java
```diff
@@ -30,14 +34,14 @@
...
-    private String[] serverTypes;
+    private List<String> serverTypes;
     private boolean isInitialized;
     private boolean configValid;
 
 
     /** Creates a new instance of DBStrings */
     public DBStrings() {
-        this.setServerType(null);
+        this.serverTypes = Collections.EMPTY_LIST;
```
simonharrer: Collections.emptyList(); should be used instead as it uses the correct generics.

src/main/java/net/sf/jabref/sql/DBStrings.java
```diff
@@ -50,8 +54,7 @@ public DBStrings() {
      * Initializes the variables needed with defaults
      */
     public void initialize() {
-        String[] servers = {"MySQL", "PostgreSQL"};
-        setServerTypes(servers);
+        this.serverTypes = Arrays.asList(new String[] {"MySQL", "PostgreSQL"});
```
simonharrer: You can simply write `Arrays.asList("A", "B");`

Issue #763, opened 2016-02-02
Group right click menu


oscargus:
In `RightClickMenu.java` there's a method `buildGroupMenu()` which generates a submenu for dealing with groups. This has been disabled for quite some time and now the earlier commented out code is not even there anymore.

Does anyone remember what it does and should we try to enable it again or just remove it completely?

lenhard:
Since it didn't work anyway, I am very much for cleaning up the code (that is, for merging your pull request).

@JabRef/developers Does anybody have a different opinion?

koppor:
According to the help, there is a context menu for dealing with groups: http://help.jabref.org/en/GroupsHelp/

I think, we disabled it because of https://sourceforge.net/p/jabref/bugs/1110/. I close the issue there since the follow up is here.

    For large databases (>500 entries) JabRef becomes pretty slow. In particular, opening the context menu on an entry takes several seconds, which is completely unacceptable.

On the one hand, the title of this issue should be "Reenable context menu at entries for groups". On the other hand, no user complained. Maybe, we should ask at the mailing list?

tobiasdiez:
I'm also in favor of removing the code. In my opinion, the current solution (Right click -> Add group opens a new dialog) is well-arranged and more clean then an endless submenu with all groups.

matthiasgeiger:
:+1: 
... especially as it is possible to drag&drop entries into groups as an alternative solution without using the right-click menu.

PR #765, opened 2016-02-02
Removed unused group submenu from RIghtClickMenu

oscargus:
All code is now gone (and the Move to group menu item follows the same logic as the add to and remove from group).


oscargus:
This implements what is discussed in #763 so depending on the answer to that question this should be merged or not.

tobiasdiez:
:+1:

PR #769, opened 2016-02-04
Added some pre-requisites for testing custom imports

oscargus:
Added the custom import example source code (http://help.jabref.org/en/CustomImports/) to the repository and added a test for it. Also created a new directory `src/main/nonpackagejava` to eventually enable testing of custom import and export and added the "same" file there.

Not sure what will happen if the same class name is used, so I avoided that.

See https://github.com/JabRef/help.jabref.org/issues/3 for a bit more background.

### DIFF
--- a/src/main/nonpackagejava/CustomImportExample.java
+++ b/src/main/nonpackagejava/CustomImportExample.java
@@ -1,0 +1,45 @@
+import java.io.*;
+import java.util.*;
+
+import net.sf.jabref.importer.ImportFormatReader;
+import net.sf.jabref.importer.OutputPrinter;
+import net.sf.jabref.importer.fileformat.ImportFormat;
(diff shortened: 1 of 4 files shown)


simonharrer:
Maybe better implemented in https://github.com/JabRef/custom-imports

oscargus:
Indeed! Didn't know of it.

From a testing perspective it may be interesting to somehow connect that
repository, but I agree that the current update is better off there. Should
I add the test as well? I guess that eventually it would be nice to have
tests for the custom importers stored there.

simonharrer:
Yes, please add it directly to the other repository. There are integration possibilities, but I am not sure if we should invest much time on this at the moment.

simonharrer:
Closed as should be done in other repository.

PR #771, opened 2016-02-04
Using short circuit logic in boolean context

AymanDF:
This pull request is focused on resolving occurrences of Sonar rule squid:S2178 - “ Short-circuit logic should be used in boolean contexts”. You can find more information about the issue here:
https://dev.eclipse.org/sonar/rules/show/squid:S2178
Please let me know if you have any questions.
Ayman Abdelghany.

### DIFF
--- a/src/main/java/net/sf/jabref/bibtex/BibEntryWriter.java
+++ b/src/main/java/net/sf/jabref/bibtex/BibEntryWriter.java
@@ -72,7 +72,7 @@ public class BibEntryWriter {
         List<String> fields = type.getRequiredFieldsFlat();
         if (fields != null) {
             for (String value : fields) {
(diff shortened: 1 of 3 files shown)


oscargus:
The problem is that the right hand side MUST be evaluated here. An option would be to write:

`hasWritten =  writeField(entry, out, value, hasWritten, indentation) || hasWritten;`

but right now the tests are failing because of this change. (And the above wouldn't arguably be an improvement really.)

For the latter two cases (`isEmpty()`) it is OK to do like you have proposed, but not really a major benefit. I guess this is primarily important for statements like `if (a != null && a.someMethod())`.

Do not let this discourage you though! Thanks for the interest in improving the code quality, but sometimes these automatic tools are not really able to figure out the purpose of the code. (Trust me, I've cleaned up lots of JabRef code and sometimes the "obvious" improvement breaks something...)

PR #778, opened 2016-02-07
NPE when exporting the database to an .sql file

jensdoecke:
Couldn't resolve the root cause, because of a mixup between Connection and Filehandling (disregards SoC)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ to [sourceforge feature requests](https://sourceforge.net/p/jabref/features/) by
 - IEEEXplore fetcher replaces keyword separator with the preferred
 - Fixed [#710](https://github.com/JabRef/jabref/issues/710): Fixed quit behaviour under OSX
 - Merge from DOI now honors removed fields
+- Fixed [#778](https://github.com/JabRef/jabref/issues/778): Fixed NPE when exporting to .sql File
 
 
 ### Removed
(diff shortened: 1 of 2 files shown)


oscargus:
Thanks for the contribution. Makes sense!

Next time, please create a branch for your PR. I think that there may be some problems merging from your master branch so I let someone more knowledgeable do it.

simonharrer:
Could you add a CHANGELOG entry, please?

For us, it is not soo much an issue if this is on the master branch. So when a CHANGELOG entry is added, we can merge this in. 

But maybe we should refactor the code to avoid these if statements whether we have a database connection or a file handle.

jensdoecke:
I added the Fix to the CHANGELOG. 
These are my first steps in open source contribution so please forgive me the beginnner faults. ;-)

tobiasdiez:
Thanks @jensdoecke. 
Can you please shortly explain why you tested that `out` is a connection and not that `val` is not null?

jensdoecke:
In my opinion is the code more readable.

I prefer something like this:
When the object is a connection, I want to check if it is a MySql DB.

Instead of the assumption:
When the object is null, it must be a FileOutputStream and no Connection, so I don't check the dbParams.

tobiasdiez:
Ok thanks for this clarification and your pullrequest. In my opinion, this can be merged in. :+1:

stefan-kolb:
Thanks for the PR :+1:

PR #782, opened 2016-02-08
Fix a small bug in citekey generator

tobiasdiez:
Generating a new key would prepend the pattern which resulted for example in `[author]Weinstein` instead of the expected `Weinstein`. The reason for the bug was that the list returned by
`getLabelPattern().getValue(entryType)` contains as the first item the complete pattern.

This bug was introduced in https://github.com/oscargus/jabref/commit/81f2a687010c4da31f1808133ae03fb3ef8cd0b9.


oscargus:
Ah, sorry and thanks!

koppor:
Where is the test case? :innocent:

tobiasdiez:
@koppor Apparently there are no tests for this method (although there is bunch of tests for the class...), otherwise they would have catched the bug. But feel free to add tests :smiling_imp:

PR #784, opened 2016-02-09
More OpenOffice fixes

oscargus:
* Extracted strings
* More detailed exception handling
* Simplified/improved code

Note that there's a new file and one removed, but here it looks like it is edited and renamed...

### DIFF
--- a/src/main/java/net/sf/jabref/openoffice/CitationManager.java
+++ b/src/main/java/net/sf/jabref/openoffice/CitationManager.java
@@ -45,6 +45,7 @@ import java.awt.*;
 import java.awt.event.ActionEvent;
 import java.awt.event.MouseAdapter;
 import java.awt.event.MouseEvent;
+import java.util.Optional;
 
 /**
  * Dialog for modifying existing citations.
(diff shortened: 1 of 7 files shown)


src/main/java/net/sf/jabref/openoffice/CitationManager.java
```diff
@@ -58,13 +61,14 @@
...
-    public CitationManager(final JabRefFrame frame, OOBibBase ooBase) throws Exception {
+    public CitationManager(final JabRefFrame frame, OOBibBase ooBase)
+            throws NoSuchElementException, WrappedTargetException, UnknownPropertyException {
         diag = new JDialog(frame, Localization.lang("Manage citations"), true);
         this.ooBase = ooBase;
 
         list = new BasicEventList<>();
         XNameAccess nameAccess = ooBase.getReferenceMarks();
-        String[] names = ooBase.getJabRefReferenceMarks(nameAccess);
+        java.util.List<String> names = ooBase.getJabRefReferenceMarks(nameAccess);
```
lenhard: Is there a reason to the fully qualified class name here? Why not add `import java.util.List;` to the class header? Is there a conflict with `java.awt.List`?

oscargus: Not really. I guess it was Eclipse's choice as I usually try to avoid it. But you are probably right about java.awt.List.

lenhard:
Ok, then I think this is fine to merge!

PR #787, opened 2016-02-10
Fixed some warnings

oscargus:
The most important consequences:

* The entry count when synchronizing file links is correct again (and the database is marked as modified).

* The correct (i.e. current) encoding is hard coded for ACM, ADS and Inspire fetchers.

### DIFF
--- a/src/main/java/net/sf/jabref/importer/fetcher/INSPIREFetcher.java
+++ b/src/main/java/net/sf/jabref/importer/fetcher/INSPIREFetcher.java
@@ -33,6 +33,7 @@ import java.io.UnsupportedEncodingException;
 import java.net.HttpURLConnection;
 import java.net.URL;
 import java.net.URLEncoder;
+import java.nio.charset.Charset;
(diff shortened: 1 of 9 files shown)


src/main/java/net/sf/jabref/importer/fetcher/INSPIREFetcher.java
```diff
@@ -116,8 +117,10 @@ private BibDatabase importInspireEntries(String key, OutputPrinter frame) {
             HttpURLConnection conn = (HttpURLConnection) new URL(url).openConnection();
             conn.setRequestProperty("User-Agent", "JabRef");
             InputStream inputStream = conn.getInputStream();
+            System.err.println(conn.getContentType());
```
oscargus: This should have been removed...

simonharrer: Indeed. :)

oscargus:
Removed the forgotten print-out...

PR #790, opened 2016-02-11
Web Search: display only the name of the database

mlep:
Currently, in the Web Search, "Fetch Arxiv.org" and "Fetch INSPIRE" were displayed. For the other databases, only the name is displayed, which is sufficient

### DIFF
--- a/src/main/resources/l10n/Menu_en.properties
+++ b/src/main/resources/l10n/Menu_en.properties
@@ -108,7 +108,7 @@ Write_XMP-metadata_to_PDFs=Write_XMP-metadata_to_PDFs
 Set/clear_fields=Set/clear_fields
 
 Export_selected_entries=Export_selected_entries
-Fetch_ArXiv.org=Fetch_ArXiv.org
+Fetch_ArXiv.org=ArXiv.org
 Sessions=Sessions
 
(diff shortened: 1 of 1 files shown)


mlep:
It looks like I forgot that, in the English version, the key and the translation should be identical.
Another way to solve the issue?

simonharrer:
I would suggest to not use a translation here as INSPIRE and Archive.org cannot be translated.

mlep:
Good idea. Note: more database are listed in the Menu_en.properties

simonharrer:
@mlep need any help?

mlep:
Thank you for the reminder.
To solve this issue, the java code has to be modified. I do not think I have the knowledge for this...

simonharrer:
Well, it is really not so complicated. Maybe you want to try? Otherwise create an issue where you mention all the translations which do contain names which cannot be translated, so someone else can implement this. 

In the Java files, you only need to replace `Localization.lang("Search JSTOR")` with `Localization.lang("Search %0", "JSTOR")` which requires a `Search_%0` in the property files.

mlep:
I have just looked at it, but could not find the proper Localization.lang() commands...
So, I have opened the issue #941. I hope this help.

PR #797, opened 2016-02-11
Some more use of Java 8 features

oscargus:
Streams are awesome!

### DIFF
--- a/src/main/java/net/sf/jabref/JabRefPreferences.java
+++ b/src/main/java/net/sf/jabref/JabRefPreferences.java
@@ -28,6 +28,7 @@ import java.util.*;
 import java.util.prefs.BackingStoreException;
 import java.util.prefs.InvalidPreferencesFormatException;
 import java.util.prefs.Preferences;
+import java.util.stream.Collectors;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.nio.charset.Charset;
@@ -944,17 +945,7 @@ public class JabRefPreferences {
             return;
         }
 
-        if (value.isEmpty()) {
-            put(key, "");
-        } else {
-            StringBuilder linked = new StringBuilder();
-            for (int i = 0; i < (value.size() - 1); i++) {
-                linked.append(makeEscape(value.get(i)));
-                linked.append(';');
-            }
-            linked.append(makeEscape(value.get(value.size() - 1)));
-            put(key, linked.toString());
-        }
+        put(key, String.join(";", value.stream().map(val -> makeEscape(val)).collect(Collectors.toList())));
     }
 
 
--- a/src/main/java/net/sf/jabref/JabRefPreferences.java
+++ b/src/main/java/net/sf/jabref/JabRefPreferences.java
@@ -945,7 +945,7 @@ public class JabRefPreferences {
             return;
         }
 
-        put(key, String.join(";", value.stream().map(val -> makeEscape(val)).collect(Collectors.toList())));
+        put(key, value.stream().map(val -> EntryUtil.quote(val, ";", '\\')).collect(Collectors.joining(";")));
     }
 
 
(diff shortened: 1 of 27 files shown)


src/main/java/net/sf/jabref/JabRefPreferences.java
```diff
@@ -944,17 +945,7 @@ public void putStringList(String key, List<String> value) {
...
-        } else {
-            StringBuilder linked = new StringBuilder();
-            for (int i = 0; i < (value.size() - 1); i++) {
-                linked.append(makeEscape(value.get(i)));
-                linked.append(';');
-            }
-            linked.append(makeEscape(value.get(value.size() - 1)));
-            put(key, linked.toString());
-        }
+        put(key, String.join(";", value.stream().map(val -> makeEscape(val)).collect(Collectors.toList())));
```
simonharrer: You can even use a Collectors.joining(";") instead of the String.join

src/main/java/net/sf/jabref/model/entry/FileField.java
```diff
@@ -73,17 +62,21 @@ public ParsedFileField(String description, String link, String fileType) {
...
+            }
 
             FileField.ParsedFileField that = (FileField.ParsedFileField) o;
 
-            if (this.description != null ? !this.description.equals(that.description) : that.description != null)
+            if (this.description != null ? !this.description.equals(that.description) : that.description != null) {
                 return false;
-            if (this.link != null ? !this.link.equals(that.link) : that.link != null)
+            }
+            if (this.link != null ? !this.link.equals(that.link) : that.link != null) {
```
simonharrer: Hm, I do not like a ternary construct within an if condition, as it makes this code extremely hard to read. Could you extract the result of the ternary expression into a separate variable, please?

oscargus: Sure! I don't like it either. (The "edit" was automatically done by
Eclipse.)

oscargus:
I moved the `quote()` method from StringUtil to EntryUtil to make the model part independent again. Ideally one should also move (and use) unquote and maybe some more method.

src/main/java/net/sf/jabref/model/entry/FileField.java
oscargus: Since these are always non-null (see the constructor), the statements could be significantly simplified.

simonharrer:
Hm, I do not like it that the quote method is moved. I vote for moving everything back as before as it creates strange dependencies. A little bit of duplication is not a large issue, but I did like the FileField class before much better than the proposed version.

tobiasdiez:
Why not move all the string related helper methods (quote and encodeStringArray) to the StringUtil class?

src/test/java/net/sf/jabref/model/entry/EntryUtilTest.java
```diff
@@ -23,4 +25,12 @@ public void getSeparatedKeywords() {
         Assert.assertArrayEquals(expected, separatedKeywords.toArray());
     }
 
+    @Test
+    public void testQuote() {
```
tobiasdiez: I would separate these test cases in separate tests.

oscargus:
Because it will create a dependency from model to util, which is not
allowed. I'll move things back and duplicate a bit of the code.

src/test/java/net/sf/jabref/model/entry/EntryUtilTest.java
oscargus: I'll have a look at it. There will be a few more tests as well as there
will be more than one quote method (these tests were only moved from
StringUtil).

tobiasdiez:
Mhhh... but wouldn't it be cleaner to move the code in question from model to logic if it relies on StringUtil?  (For example the parse method in FileField definitely classifies as logic).

oscargus:
I agree with that, but then model won't be independent, which is also a
nice feature.

tobiasdiez:
No, what I meant is the following:
Move all the code which directly or indirectly relies on StringUtil (quote, encodeArray, ...) from the models package to logic. Hence there will be no reference to StringUtil in the model. For example, move the FielField.parse() method to some class in logic. In this way the model still is independent from logic, but logical code which was previously in model is now in logic.

oscargus:
Sorry, I meant logic, not util. StringUtil resides in logic.

I may still be missing something here though...

simonharrer:
I do not want to introduce dependencies for the model (which currently has no dependencies on other code in JabRef) just because of one or two very simple methods. This, from my point of view, is simply not worth it, and therefore, I would take code duplication into account.

oscargus:
Things are restored, so a generic quote method in StringUtil and a hard-coded in FileField.

Issue #789, opened 2016-02-10
Preference for default file-based mode


stefan-kolb:
- [x] Description should be rewritten to say "Default Tex mode".
- [x] select/radio button/dropdown for bibtex & biblatex
- [x] move to general tab?

![image](https://cloud.githubusercontent.com/assets/2141507/12965811/78138484-d059-11e5-8689-d0355f80ea7e.png)

simonharrer:
:+1: good idea

PR #799, opened 2016-02-11
Better preference for default file-based mode

stefan-kolb:
See #789.

- Description rewritten to say "Default TeX mode".
- Dropdown for bibtex & biblatex
- Moved to general tab
- Decided to keep old boolean preference to avoid preference migration for now

image

@JabRef/developers WDYT?


koppor:
`TeX` is misleading as this lets me think about pdflatex and latex or pure tex. Maybe "Default bibliography mode" as we treat the bibliography?

tobiasdiez:
:+1: 
Two minor remarks:
- I would write "BibTeX" and "BibLaTex" instead of everything upper case (what are our convention on this?)
- You can rename JabRefPreferences.BIBLATEX_MODE to something containing 'defaut' without forcing migration (i.e. leave the underlying string unchanged)

koppor:
The spelling is discussed at https://github.com/koppor/jabref/issues/73. BibTeX is set, but Biblatex or BibLaTeX is not yet finally decided.

simonharrer:
Looks good to me.

PR #804, opened 2016-02-12
Some more cleanups in logic

oscargus:


src/main/java/net/sf/jabref/logic/formatter/bibtexfields/PageNumbersFormatter.java
```diff
@@ -18,6 +18,9 @@
 
     private static final Pattern PAGES_PATTERN = Pattern.compile("\\A(\\d+)-{1,2}(\\d+)\\Z");
 
+    private static final String REJECT_LITERALS = "[^0-9,\\-\\+]";
+    private static final String REPLACE = "$1--$2";
```
tobiasdiez: Replace is not very descriptive. What about PagePattern or if you want ReplacePattern?

src/main/java/net/sf/jabref/logic/formatter/bibtexfields/SuperscriptFormatter.java
```diff
@@ -28,6 +28,8 @@
     private static final Pattern PATTERN = Pattern.compile("\\b(\\d+)(st|nd|rd|th)\\b",
             Pattern.CASE_INSENSITIVE | Pattern.MULTILINE);
 
+    // adds superscript tag
+    private static final String REPLACE = "$1\\\\textsuperscript{$2}";
```
tobiasdiez: Same as above. Also the comment would be better for line 63 I think

src/main/java/net/sf/jabref/logic/search/describer/GrammarBasedSearchRuleDescriber.java
```diff
@@ -109,8 +109,8 @@ public String visitComparison(SearchParser.ComparisonContext context) {
             }
 
         }.visit(parseTree));
-        stringBuilder.append(". ");
-        stringBuilder.append(caseSensitive ? Localization.lang("The search is case sensitive.") :
+        stringBuilder.append(". ").append(caseSensitive ? Localization
+                .lang("The search is case sensitive.") :
```
tobiasdiez: I actually found the previous code more readable.

tobiasdiez:
:+1: 
Good to go, except some minor remarks.

oscargus:
Thanks! I've updated accordingly.

PR #805, opened 2016-02-13
Lots of minor cleanups

oscargus:
Yet another massive cleanup... Sorry about that...

Major thing:
* Replaced the basic `StringUtil.join()` with `String.join()`

Most of the other changes are just plain edits (like not calling Assert.assertEquals() but instead assertEquals() since it is already imported statically).

### DIFF
--- a/src/main/java/net/sf/jabref/logic/autocompleter/ContentAutoCompleters.java
+++ b/src/main/java/net/sf/jabref/logic/autocompleter/ContentAutoCompleters.java
@@ -15,14 +15,14 @@ public class ContentAutoCompleters extends AutoCompleters {
     private final JournalAbbreviationLoader abbreviationLoader;
 
 
-    public ContentAutoCompleters(AutoCompletePreferences preferences, JournalAbbreviationLoader abbreviationLoader) {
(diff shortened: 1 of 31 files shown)


src/main/java/net/sf/jabref/logic/autocompleter/ContentAutoCompleters.java
```diff
@@ -12,12 +12,11 @@
 
 public class ContentAutoCompleters extends AutoCompleters {
 
-    AutoCompletePreferences preferences;
     private final JournalAbbreviationLoader abbreviationLoader;
 
 
     public ContentAutoCompleters(AutoCompletePreferences preferences, JournalAbbreviationLoader abbreviationLoader) {
-        this.preferences = Objects.requireNonNull(preferences);
+        Objects.requireNonNull(preferences);
```
simonharrer: You can remove the preferences argument if it is not stored and used.

oscargus: OK! Now, I see the bigger picture of how it is used. But could indeed be remove here.

Issue #685, opened 2016-01-20
Can't export to MySQL 5.5 in JabRef 3.2


yuyaoshen:
I'm using Jabref 3.2 and MySQL 5.5, but now I can't export my .bib to SQL database. It is fine when i am using 2.10, I don't know why.
The error is: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'INSERT INTO entries (jabref_eid, entry_types_id, cite_key, abstract, address, an' at line 1.
Can someone help me with this?

koppor:
Is it possible to post a [minimal, complete, and verifiable example](https://en.wikipedia.org/wiki/Minimal_Working_Example)? Is there a minimal bib file which is not working at your side? Which OS are you running?

yuyaoshen:
This is the example. If there is only one record, the export will fine, but if there two, it error.
I'm running Ubuntu 14.04.
[test.bib.txt](https://github.com/JabRef/jabref/files/99150/test.bib.txt)

Siedlerchr:
I'll take a look at that.

Siedlerchr:
I looked into the code and interestingly the generated statement works when it's executed directly on the Mysql Database. I will look further into this issue-

Siedlerchr:
I found the reason for the error and a fix!
The problem is, that the usage of multiple statements at once has to be specified in the Connection String. So, the solution is to simply add the parameter `?allowMultiQueries=true`.

I'll prepare a Pull Request.

[StackOverlow Answer to the Problem](http://stackoverflow.com/questions/10797794/multiple-queries-executed-in-java-in-single-statement

yuyaoshen:
so where should I add this parameter, I'm using .jar file. Thanks very much.

Siedlerchr:
@yuyaoshen  It has to be fixed in the code. Wait for a new release.

yuyaoshen:
@Siedlerchr Thanks again, waiting for the new release

PR #807, opened 2016-02-13
Fixes MySQL export for more than one entry 

Siedlerchr:
Fixes the export of more than one entry through the connection option #685 
Added ability to add server specific connection parameters in dbStrings, because MySQL and PostreSQL parameter names differ.

Using setCatalog instead of use database (recommended in mysql doc see: https://dev.mysql.com/doc/connector-j/en/connector-j-reference-configuration-properties.html , Section  "Initial Database for Connection" )


oscargus:
Thanks for this PR!

Ideally, I guess, one would like to set parameters pairwise as `dbStrings.addParameter("allowMultiQueries", "true")` and then have `dbStrings.getParameters()` form the correct connection string (see e.g. https://github.com/JabRef/jabref/blob/master/src/main/java/net/sf/jabref/logic/net/URLDownload.java ). However, I assume that this may be the only parameter of interest to add...

lenhard:
@Siedlerchr could you rebase this on master? There is also a good chance that this fixes your failing tests.

Siedlerchr:
@lenhard  Yes, I wil check it this evening.

src/main/java/net/sf/jabref/sql/exporter/MySQLExporter.java
```diff
@@ -36,6 +36,7 @@
 final public class MySQLExporter extends DBExporter {
 
     private static MySQLExporter instance;
+    private static String optAllowMultiQueries = "?allowMultiQueries=true";
```
simonharrer: Please make this final and use UPPER_CASE_NAMING.

Siedlerchr:
I have recreated the PR on the current upstream master branch, so ignore this one here.
(I'm still figuring out that whole git mergin stuff)

PR #812, opened 2016-02-13
Removed unused imports etc.

oscargus:


src/main/java/net/sf/jabref/external/DroppedFileHandler.java
```diff
@@ -560,7 +555,7 @@ private boolean doCopy(String fileName, ExternalFileType fileType, String toFile
         if (found < 0) {
             // OOps, we don't know which directory to put it in, or the given
             // dir doesn't exist....
-            System.out.println("dir: " + dirs.get(found) + "\t ext: " + fileType.getExtension());
+            // This should not happen!!
```
tobiasdiez: Why not log it in case it does happen? Actually, it is relatively probable that no folder is found: just copy the file to a different pc with completely different file disk layout.

src/main/java/net/sf/jabref/gui/JabRefFrame.java
```diff
@@ -162,34 +162,6 @@ public void valueChanged(ListSelectionEvent e) {
         }
     };
 
-    class ToolBar extends OSXCompatibleToolbar {
```
tobiasdiez: Is this really not used on OSX systems? I think there were some OSX classes which were apparently not used but on Mac there were referenced using reflection...

src/main/java/net/sf/jabref/external/DroppedFileHandler.java
oscargus: You're probably correct. I just copied the "The this should not happen from the corresponding place in `doMove`.

src/main/java/net/sf/jabref/importer/UnlinkedFilesCrawler.java
```diff
@@ -60,12 +61,24 @@ public CheckableTreeNode searchDirectory(File directory, UnlinkedPDFFileFilter f
             return null;
         }
 
-        List<File> files = Arrays.asList(directory.listFiles(ff));
+        File[] filesArray = directory.listFiles(ff);
```
tobiasdiez: What is wrong with the the previous code? I think there is no need to convert the array to a list since we just iterate over it anyway.

src/main/java/net/sf/jabref/openoffice/StyleSelectDialog.java
```diff
@@ -439,7 +440,13 @@ private void selectLastUsed() {
     private void addStyles(String dir, boolean recurse) {
         File dirF = new File(dir);
         if (dirF.isDirectory()) {
-            List<File> files = Arrays.asList(dirF.listFiles());
+            File[] fileArray = dirF.listFiles();
```
tobiasdiez: As above.

tobiasdiez:
:+1:

src/main/java/net/sf/jabref/importer/UnlinkedFilesCrawler.java
oscargus: There's a problem if `directory.listFiles(ff)` returns `null`. If we know that it won't happen then the previous code can indeed be simplified further (but as I recall it, iterating over `null` doesn't work as well as iterating over an empty list). I got a warning in Eclipse, so that is why I tried to solve it.

tobiasdiez: Ah ok, I see...maybe replace it completely by DirectoryStream.

oscargus: Would make sense, but also requires rewriting `UnlinkedPDFFileFilter`, so I will not get into this in the current PR.

PR #814, opened 2016-02-13
Got rid of null return values

oscargus:


src/main/java/net/sf/jabref/external/TransferableFileLinkSelection.java
```diff
@@ -43,7 +43,7 @@ public TransferableFileLinkSelection(BasePanel panel, List<BibEntry> selection)
         if (tm.getRowCount() > 0) {
             // Find the default directory for this field type, if any:
             List<String> dirs = panel.getBibDatabaseContext().getMetaData().getFileDirectory(Globals.FILE_FIELD);
-            File expLink = FileUtil.expandFilename(tm.getEntry(0).link, dirs);
+            File expLink = FileUtil.expandFilename(tm.getEntry(0).link, dirs).get();
```
tobiasdiez: I would add here a isPresent check. (the previous code did not throw an exception if expandFilename returned null, but now it does throw an error in this case)

src/main/java/net/sf/jabref/importer/DatabaseFileLookup.java
```diff
@@ -88,8 +88,8 @@ public boolean lookupDatabase(File file) {
                 continue;
             }
 
-            File expandedFilename = FileUtil.expandFilename(link, possibleFilePaths);
-            fileLinks.add(expandedFilename);
+            Optional<File> expandedFilename = FileUtil.expandFilename(link, possibleFilePaths);
+            fileLinks.add(expandedFilename.get());
```
tobiasdiez: Same as above

tobiasdiez:
:+1:

oscargus:
Thanks! I fixed those things now.

simonharrer:
Can you please double check as there are a few occasions where you simply call get without verifying if the value is null.

oscargus:
I have double-checked and expanded the potential cases. It was a bit confusing in `DroppedFileHandler` as the code clearly assumes that null is not returned, so it wasn't obvious what to do (nor how to access the critical part to test it). Found an unused method though...

oscargus:
Oh, in the tests, I didn't handle it since the tests would fail from a null return anyway, so might just as well fail in an NPE as in an incorrect assertion?

simonharrer:
Hm, in the tests, you should create an assertion before, ensuring that it is present.

oscargus:
OK! The tests are fixed now (and, yes, it makes much more sense now). However, they are all ignored...

PR #821, opened 2016-02-15
Save actions gui

lenhard:
This is the first (functional) shot at integrating a GUI for configuring save actions. I think it is now time for feedback on two aspects:
 1. the save action feature itself
 2. the look of the GUI

Regarding 1.: Currently, any class implementing `net.sf.jabref.logic.formatter.Formatter` can be used as a save action and can be configured for a field of any non-null and non-empty name (so you can configure actions for your custom fields). That is quite flexible. It would be trivial to extend that to `net.sf.jabref.logic.cleanup.CleanupJob`. The problem with integrating cleanup actions is that a `CleanupJob` seems to represent a batch job of cleanups that are preconfigured for certain fields. It is generally not intended to execute a `CleanupJob` for a specific and selectable field. Hence, I am not sure if cleanups and save actions go well together. A downside of the current save action feature is that save actions cannot be undone (like the sorting feature).

Regarding 2.: The GUI is functional, *not* pretty. 
save actions
It has a check box to enable or disable all actions, a set of controls for adding a new action, a list for displaying actions, and a button for deleting a selected action. Do you have suggestions for a better set of controls?

Finally, the layout of the controls definitely needs improvement and for some reason the size of the controls increases when adding an action. These things need to be fixed. Alas, layouting is not my favourite task...

I would be grateful if someone takes a look at this and provides some feedback!


tobiasdiez:
Very nice!
A few remarks or suggestions:
1. For the GUI: Maybe add a line break after the checkbox, so that the controls for adding a new save actions are below the checkbox (and maybe add some indention). Write "field" in the textbox and delete it upon focus (similar to how the search box works). Decrease width of delete button. Besides that It is already prettier then the usual JabRef UI :smile: 
2. I think the `SaveAction` and `CleanupFormatter` classes can be merged with `FieldFormatterCleanup` (all of them essentially just store a field and a formatter). Is `CleanupFormatter` even used? Moreover, the `SaveActions.applyActionForField` method is covered by `FieldFormatterCleanup.cleanup`. Replacing `SaveAction` by `FieldFormatterCleanup` has also the advantage that the `SaveActionsPanel` can be used to add arbitrary formatters as cleanup operations in the cleanup dialog.

src/main/java/net/sf/jabref/exporter/SaveActions.java
```diff
@@ -7,41 +7,91 @@
...
 
     private List<Formatter> availableFormatters;
 
     public static final String META_KEY = "saveActions";
 
+    private boolean enabled;
+
     public SaveActions(MetaData metaData) {
-        actions = new HashMap<>();
+        if (metaData == null) {
```
tobiasdiez: Objects.requireNonNull?

src/main/java/net/sf/jabref/gui/databaseProperties/SaveActionsListModel.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.List;
+
+class SaveActionsListModel<SaveAction> implements ListModel<SaveAction> {
+
+    private List<SaveAction> saveActions;
+
+    private List<ListDataListener> listeners;
+
+    public SaveActionsListModel(List<SaveAction> saveActions) {
+        if (saveActions == null) {
```
tobiasdiez: Replace by Objects.requireNonNull ?

src/main/java/net/sf/jabref/exporter/SaveActions.java
```diff
@@ -7,41 +7,91 @@
...
+    @Override
+    public int hashCode() {
+        int result = actions.hashCode();
+        result = 31 * result + (enabled ? 1 : 0);
+        return result;
+    }
+
+    private void parseSaveActions(List<String> formatters) {
+        //read concrete actions
+        for (int i = 1; i < formatters.size(); i += 2) {
```
oscargus: Wouldn't it be better to save this "2D"-array as a 2D-array instead of the even/odd thing? I'm thinking that the save format to the preferences is quite sensitive to get right first time to avoid migration hassle.

src/main/java/net/sf/jabref/gui/databaseProperties/SaveActionsPanel.java
```diff
@@ -0,0 +1,135 @@
...
+
+    private JComboBox<String> formatters;
+
+    public SaveActionsPanel() {
+
+        enabled = new JCheckBox(Localization.lang("Enable save actions"));
+    }
+
+    public void setValues(MetaData metaData) {
+        // first clear existing content
```
tobiasdiez: Require metaData to be non null?

src/main/java/net/sf/jabref/exporter/SaveActions.java
oscargus: Or maybe the meta data format is not so suitable for this...

lenhard: Yes, it would make sense to get that right the first time :) I am open to improvements regarding the format. Currently, it looks like this 
```
@comment{jabref-meta: saveActions:enabled;title;LowerCaseChanger;}
```

Do you have a suggestion for a better format?

lenhard:
@tobiasdiez Thanks! `FieldFormatterCleanup` and `SaveAction` indeed seem very replaceable. I'll try to merge the two. This means I have to add implementations of `equals`, `hashCode`, and `toString` to `FieldFormatterCleanup`.

I'll also try to add your remaining suggestions regarding layout and `Objects.requireNonNull`.

lenhard:
Ok, here's the new UI:
save actions2
The formatting changes also killed the prior resizing bug :) Any other suggestions?

As a side note: I seems that not all formatters that have been implemented have been added to the respective management classes. For instance, `net.sf.jabref.logic.formatter.bibtexfields.LatexFormatter` is not available in `BibtexFieldFormatters`. Is there a reason for that? If not, I would simply add the missing formatters, which would make them available as a safe action.

matthiasgeiger:
Looks good to me! :+1: 

Just a quick note from the User perspective: This should really be explained somewhere... we need a help page and a documenation for each formatter, as they may not be as selfexplanatory as needed (e.g., the mentioned "LatexFormatter")

src/main/java/net/sf/jabref/exporter/SaveActions.java
tobiasdiez: mhh... maybe something like 
`title [LowerCaseChanger, SecondFormatter], abstract [Formatter1, Formatter2]` but this is harder to parse. Or maybe use a different symbol for separating the field and the formatter. Something like `title: LowerCaseChanger, title: SecondFormatter`. 
But I'm not sure that these suggestions are really improvements.

tobiasdiez:
Looks good to me too! 
Maybe the display names of the formatter can be improved, but this is just a minor point.
:+1: to merge this in.

src/main/java/net/sf/jabref/exporter/SaveActions.java
lenhard: The thing is, we would have to change `MetaData` for that. I am not explicitly creating that structure, but just using the access via `MetaData` to the file. I can set a key (`saveActions`) and a list of values. `MetaData` takes care of the separators (`:`,`;`). All I could do would be to encode everything in a custom string or to use multiple `jabref-meta` entries.

lenhard:
Currently, I am displaying the formatter keys, not the display names. I'll have a look at that :)

lenhard:
I had a look at the formatter display names and, to be honest, they are quite bad ("bad" as in "non-informative"). The keys are much better. For instance "TitleCaseChanger" is better than "Title". Therefore, I will keep using keys for the display.

Anyways, @matthiasgeiger is of course right. There is little use for this feature if it is not documented somewhere in the help files. I would suggest to do that in a separate PR, however. As far as I understand it needs to be done in another repo as well?

When we have concluded the discussion on the meta-data format for save actions, this PR can be merged from my point of view.

koppor:
The formatters are explained at http://help.jabref.org/en/CustomExports/. Section "Built-in export formatters". Unfortunately, we do not offer anchors to headings (see https://github.com/JabRef/www.jabref.org/issues/24).

lenhard:
Ok, I changed the serialization format of save actions according to the description by @tobiasdiez The new format looks like this:
```
@Comment{jabref-meta: saveActions:enabled;pages[PageNumbersFormatter,DateFormatter,]title[LowerCaseChanger,];}
```
Formatters are grouped for the fields within `[...]` and separated by commas.

The parsing logic is now considerably more complex, but for the sake of serialization beauty... :) 

Any other comments on the format? 

I have written a few more unit tests and done some manual bug testing, but it would probably be good if someone else did some manual testing, too.

koppor:
Is it possible to omit the last `,`?

simonharrer:
Nice PR! Here my few comments:

- Why do we need the enabled flag stored in the file? For me, it is enabled if there are save actions, and disabled when there are no save actions. 
- Serialization format: we could put each field on a separate line, similarly to the groups serialization. 
- Should we provide some recommendation which save actions are sensible? Maybe something like a button that activates the recommended save actions automatically?
- Should we translate/i10n the names of the formatters?

src/main/java/net/sf/jabref/exporter/SaveActions.java
```diff
@@ -72,4 +148,36 @@ private Formatter getFormatterFromString(String formatterName) {
...
+            //add the formatter to the map if it is not already there
+            List<String> formattersForKey = groupedByField.get(key);
+            if (!formattersForKey.contains(cleanup.getFormatter().getKey())) {
+                formattersForKey.add(cleanup.getFormatter().getKey());
+            }
+        }
+
+        // convert the contents of the hashmap into the correct serialization
+        StringBuilder result = new StringBuilder();
+        for (String fieldKey : groupedByField.keySet()) {
```
oscargus: Coverity will complain about inefficient iterator (and I will change it...), so you might want to do (something like):

``` java
for (Entry<String, List<String>> entry : groupedByField.entrySet()) {
result.append(entry.getValue().stream().collect(Collectors.joining(",", entry.getKey() + "[" , "]")));  // Or use StringJoiner, not sure if stream() is really needed here.
}
```

src/main/java/net/sf/jabref/exporter/SaveActions.java
```diff
@@ -72,4 +148,36 @@ private Formatter getFormatterFromString(String formatterName) {
         return new IdentityFormatter();
     }
 
+    public static String getMetaDataString(List<FieldFormatterCleanup> actionList) {
+        //first, group all formatters by the field for which they apply
+        HashMap<String, List<String>> groupedByField = new HashMap<>();
```
oscargus: Enough to use `Map` for the variable.

lenhard:
Ok, I have again integrated most of the remarks by @koppor, @oscargus, and @simonharrer .

@koppor: Now, it is mandatory to omit the last `,`

Regarding the comments by @simonharrer 
- I introduced the enable flag to mimic Eclipse save actions. Say, you want to quickly turn off save actions to do some weird testing stuff in your bib file. Without the enable flag, you have to delete your save actions. For reenabling them, you have to build all actions from scratch. The enable flag is a very easy way to take care of this.
-  The save actions are now serialized in separate lines per field
- A recommended set of actions sounds reasonable (we could hardcode a default set in case there is nothing configured in the file). But what kind of actions should that be? Personal preferences on that surely vary a lot, so I am quite unsure on what to recommend. Maybe we should give this some time and see which actions we use personally?
- Formatter naming is still a mess, so I would avoid adding unnecessary translations at the moment. When we build help pages, we can come up with better formatter names and, then, introduce a translation.

Once again, the PR is ready for review and/or merge from my point of view.

src/main/java/net/sf/jabref/exporter/SaveActions.java
```diff
@@ -1,47 +1,123 @@
...
-                try {
-                    String field = formatters.get(i);
-                    Formatter formatter = getFormatterFromString(formatters.get(i + 1));
-
-                    actions.put(field, formatter);
-                } catch (IndexOutOfBoundsException e) {
-                    // the meta data string in the file is broken. -> Ignore the last item
-                    break;
-                }
+            parseEnabledStatus(formatters.get(0));
```
tobiasdiez: Suggestion: change the constructor so that it accepts a boolean enabled and a list of save actions. Then the parse methods could be implemented as a static method returning a new SaveActions class.
(would make testing easier since no metadata class has to be constructed)

simonharrer:
if you have only one save action, the serialization looks a little bit awkward:

```
@Comment{jabref-meta: saveActions:enabled;booktitle[SuperscriptFormatter]
;}
```

What about something like which would be more git-friendly
```
@Comment{jabref-meta: saveActions:enabled;
booktitle[SuperscriptFormatter]
;}
```

---

Another idea for the enable-flag: you could disable all controls when the flag is disabled, so that they can only be edited if they are enabled, similar to the Eclipse settings. Would this make sense?

---

For me, they do not work. Nothing is changed upon save. I use the following configuration:

```
@Comment{jabref-meta: saveActions:enabled;pages[PageNumbersFormatter]
booktitle[SuperscriptFormatter]
;}

@InProceedings{Pautasso2015TowardsWorkflowBenchmarking,
  author =    {Pautasso, Cesare and Ferme, Vincenzo and Roller, Dieter and Leymann, Frank and Skouradaki, Marigianna},
  title =     {{Towards Workflow Benchmarking: Open Research Challenges}},
  booktitle = {Proceedings of the 16th Conference on Database Systems for Business, Technology, and Web},
  year =      {2015},
  pages =     {1-20},
  address =   {Hamburg, Germany},
  month =     {March 2-6},
  file =      {:Pautasso2015TowardsWorkflowBenchmarking.pdf:PDF},
  keywords =  {benchflow}
}
```

and the save actions should change some things in there.

src/main/java/net/sf/jabref/exporter/SaveActions.java
```diff
@@ -52,17 +128,18 @@ private void setAvailableFormatters() {
     }
 
     public BibEntry applySaveActions(BibEntry entry) {
-        for (String key : actions.keySet()) {
-            Formatter formatter = actions.get(key);
-
-            String fieldContent = entry.getField(key);
-            String formattedContent = formatter.format(fieldContent);
-            entry.setField(key, formattedContent);
+        if (enabled) {
```
tobiasdiez: I would change the method signature to return nothing (void). Otherwise the signature suggests that the input entry is not changed and the returned BibEntry contains all the changes.

src/test/java/net/sf/jabref/exporter/SaveActionsTest.java
```diff
@@ -36,14 +38,14 @@ public void checkSaveActionsParsing() throws IOException {
                 "  ISSN                     = {0886-5930}\n" +
                 "}\n" +
                 "\n" +
-                "@comment{jabref-meta: saveActions:title;IdentityFormatter;}"));
+                "@comment{jabref-meta: saveActions:enabled;title[IdentityFormatter]}"));
 
         ParserResult parserResult = parser.parse();
 
         List<String> saveActions = parserResult.getMetaData().getData(SaveActions.META_KEY);
```
tobiasdiez: I think the first part of this test should be moved to the parser (since it tests the parser more then the save actions). It is more an integration test, that the parser invokes the right methods. 
Here in this test file I would suspect tests like 
`SaveActions saveActions = SaveActions.parse("some test string"); 
Assert.something(expected, saveActions.theRealThing); `

src/test/java/net/sf/jabref/exporter/SaveActionsTest.java
```diff
@@ -36,14 +38,14 @@ public void checkSaveActionsParsing() throws IOException {
...
 
-        assertEquals("title", saveActions.get(0));
-        assertEquals("IdentityFormatter", saveActions.get(1));
+        assertEquals("enabled", saveActions.get(0));
+        assertEquals("title[IdentityFormatter]", saveActions.get(1));
 
         SaveActions actions = new SaveActions(parserResult.getMetaData());
```
tobiasdiez: Move to new test (and construct the save actions as well as the test entry by hand)

src/test/java/net/sf/jabref/exporter/SaveActionsTest.java
```diff
@@ -65,7 +67,7 @@ public void invalidSaveActionSting() throws IOException {
                 "  ISSN                     = {0886-5930}\n" +
                 "}\n" +
                 "\n" +
-                "@comment{jabref-meta: saveActions:title;}"));
+                "@comment{jabref-meta: saveActions:enabled;title}"));
```
tobiasdiez: As above.

src/test/java/net/sf/jabref/exporter/SaveActionsTest.java
```diff
@@ -89,14 +91,14 @@ public void checkLowerCaseSaveAction() throws IOException {
...
-                "@comment{jabref-meta: saveActions:title;LowerCaseChanger;}"));
+                "@comment{jabref-meta: saveActions:enabled;title[LowerCaseChanger]}"));
 
         ParserResult parserResult = parser.parse();
 
         List<String> saveActions = parserResult.getMetaData().getData(SaveActions.META_KEY);
 
-        assertEquals("title", saveActions.get(0));
-        assertEquals("LowerCaseChanger", saveActions.get(1));
+        assertEquals("enabled", saveActions.get(0));
```
tobiasdiez: As above

lenhard:
The comments by @simonharrer and @tobiasdiez are implemented. 

Now, there are newlines in the serialization as desired and the GUI components are disabled if the checkbox is unchecked.

Furthermore, an embarrasing bug discoverd by @simonharrer has been fixed and thanks to @tobiasdiez the tests look way better. 

Once again, this PR is open for review.

simonharrer:
Idea: would a TrimFormatter make sense? One that trims the value of the field? Or is this done by the serialization algorithm?

Defaults:
```
@Comment{jabref-meta: saveActions:enabled;
pages[PageNumbersFormatter]
month[MonthFormatter]
booktitle[SuperscriptFormatter]
;}
````
 I wanted to add `author[AuthorsFormatter]` as well but it has problems with author names having latex symbols in it, hence, I cannot recommend this.

**Comment by tobiasdiez** [comment]
I searched hard but I have nothing more to complain about :cry: 
So :+1: for merging it in.

@simonharrer Yes a trim formatter would be sensible (and it is somewhere on my idea-todo-list). 
I kind of remember that months should be just a number in BibLatex so I wouldn't recommend the MonthFormatter (in BibLatex mode).

**Comment by simonharrer** [comment]
It is pretty annoying if you add a save action, then see what it did (as there is no documentation, and even with documentation, you probably want to try it out) and you cannot make it reverse, only using version control. Before this is released, we need an undo functionality, I think. 

Instead of a documentation, we could simply define a few "test strings" per formatter which can be returned with an additional getter in the `Formatter` interface, and present them as well as their result on calling the formatter in the UI to display what a formatter is doing. 

Maybe using the `JSearchTextField` would be a little bit nicer for the UI, as it automatically shows text and when the focus is removed and the field is empty, this text is shown again. You could use something like `enter field, e.g., title or author`. We could rename `JSearchTextField` to something without search in its name and move it to a ui util package. 

With save actions, we need to ensure that if we call them twice (save again), they add more stuff. This should be tested somehow, as otherwise, users will complain. Maybe use the same example strings used as a documentation replacement?

Duplicate entries can occur in the UI, e.g., I can add pages[PagesNumberFormatter] multiple times. Upon close, this is cleaned up. Could be left as it is, I think. Just a remark from my side. 

Icons (+, -) could be used instead of the text on the buttons.

**Comment by lenhard** [comment]
I'd like to sum up what needs to be done to get this PR merged:
 - *Undo functionality*: In principle, this is supported. The `SaveActions` use `FieldFormatterCleanup` and I can easily return a `List<FieldChange>` upon applying save actions, like for the cleanup functionality. Just how should an undo be integrated here? How can that be integrated with `BibDatabaseWriter`? As far as I know, JabRef currently cannot undo a save operation. How should that be achieved?

- *Better text field*: @simonharrer Can you refactor the `JSearchTextField` as you described, so I can use it for save actions?

- *Duplicate detection*: I can add a duplicate check when adding save actions.

- *Button icons*: This should be a trivial change in `SaveActionsPanel`. I'll look for something.

**Comment by simonharrer** [comment]
@lenhard see 702913bed084ce3684db337320bf3b801f3cca96 for the refactored text field

**Comment by lenhard** [comment]
Again, the things I stated above are implemented and, hopefully, this can soon be merged.

Just a few remarks:
 - Personally, I like the text "Add" and "Delete" better than the Icons that are displayed now, but whatever...

- Undoing save actions after save works via the menu Edit -> Undo. For some reason the key binding CTRL + Z does not trigger an action after save. I consider this a bug that should be fixed outside of this PR.

**Comment by simonharrer** [comment]
- Strange side effect: if I add save actions, the save sort order is added as well. I just enabled the save actions, and this was added:

```
+@Comment{jabref-meta: saveOrderConfig:original;;false;;false;;false;}
+
+@Comment{jabref-meta: saveActions:enabled;
+pages[PageNumbersFormatter]
+;}
+
```

- Removing all save actions and disabling it does not remove them from the file. 

```
+@Comment{jabref-meta: saveOrderConfig:original;;false;;false;;false;}
+
+@Comment{jabref-meta: saveActions:disabled;;}
+
```

Even when they are enabled but have no configuration, they are stored.

```
+@Comment{jabref-meta: saveOrderConfig:original;;false;;false;;false;}
+
+@Comment{jabref-meta: saveActions:enabled;
+;}
+
```

I propose to only store them if they are really used. Otherwise, nothing should be stored in the bib file.

**Comment by lenhard** [comment]
And it's implemented.

I guess this PR will never end, will it? ;-)

**Comment by tobiasdiez** [comment]
Mhhh... maybe it is better to apply the save actions not in the DatabaseWriter but in another place (savesession?). Because of separation of concerns and so... What do you think? 
(I would also remove the `undoable` prefix in the field changes name. The field changes are apriori just a list of changes and have no ability to be reverted. The revert functionality is added by the UndoableFieldChange class)

File: src/main/java/net/sf/jabref/gui/databaseProperties/DatabasePropertiesDialog.java [hunk 13]
```diff
@@ -13,9 +13,9 @@
     with this program; if not, write to the Free Software Foundation, Inc.,
     51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
-package net.sf.jabref.gui;
+package net.sf.jabref.gui.databaseProperties;
```
oscargus: Please avoid capitals in package names.

tobiasdiez:
Could you also please add the TrimFormatter and RemoveBracesFormatter from #853. Thanks!

lenhard:
Once more, the above comments are implemented and this PR is ready for review. Discussion points:
 - I added the default actions as recommend by @simonharrer 

 - @tobiasdiez Regarding the relocation of the application the save actions: I am not convinced why `SaveSession` would be a better place for applying the save actions. Right now `SaveSession` is more of a storage object for information regarding a save. If we integrate the application of save actions, `SaveSession` would need to know about `BibEntry`, which it currently doesn't. Moreover, the `BibDatabaseWriter` also resorts the entries, so why not apply the save actions there? It also requires the application of save actions before writing. All in all, I currently could not find a better place for applying save actions, but I am still open for suggestions.

- I took the liberty to implement a new feature: I you choose `all` as field name in the save actions, the selected formatter will be applied to all fields of all entries in the database. Say you want to always trim all fields, this sounded useful. However, the `TrimFormatter` in its current fashion does not work, since `BibEntry.getField` for some reason returns a trimmed version of the field content regardless of the content of the GUI. But that is a bug unrelated to this PR, which *won't* be fixed in this PR. As another example, lower casing all fields works just fine.

(Note: I will take care about the codacy stuff in a final step before we aggree on merging the current feature set, but not now when things might still change)

tobiasdiez:
I'm in favor of merging this in! (Especially since I want to use the feature :smiley: ).

@lenhard It was more a feeling that the application of save actions might have a better home somewhere else. But your are right, the DataBaseWriter is probably the right place for now.

simonharrer:
About the icons: why not use the icons for adding and removing files in the Entry Editor -> General?

Otherwise this seems good to be merged.

stefan-kolb:
Nice job :+1:

PR #827, opened 2016-02-18
Added test for AuxSubGenerator

oscargus:
Added a basic test for the AuxSubGenerator.

### DIFF
--- a/src/test/java/net/sf/jabref/wizard/auximport/AuxSubGeneratorTest.java
+++ b/src/test/java/net/sf/jabref/wizard/auximport/AuxSubGeneratorTest.java
@@ -1,0 +1,105 @@
+package net.sf.jabref.wizard.auximport;
+
+import static org.junit.Assert.*;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.util.List;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import net.sf.jabref.Globals;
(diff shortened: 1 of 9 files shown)


src/test/java/net/sf/jabref/wizard/auximport/AuxSubGeneratorTest.java
```diff
@@ -0,0 +1,105 @@
...
+            assertEquals(2, auxSubGenerator.getResolvedKeysCount());
+            assertEquals(2, auxSubGenerator.getFoundKeysInAux());
+            assertEquals(auxSubGenerator.getFoundKeysInAux(),
+                    auxSubGenerator.getResolvedKeysCount() + auxSubGenerator.getNotResolvedKeysCount());
+            assertEquals(0, auxSubGenerator.getNestedAuxCounter());
+            assertEquals(0, auxSubGenerator.getCrossreferencedEntriesCount());
+
+            // Clear
+
+            auxSubGenerator.clear();
```
simonharrer: Why do we need a reusable AuxSubGenerator? It would be easier to maintain a one time usable class with final fields instead of passing them to the `generate` method. Then, the AuxSubGenerator could return a new class like AuxSubGeneratorResult which contains all the methods to query it, e.g., getFoundKeysInAux.

oscargus: I agree. Not sure if I'm up for the refactoring at the moment though...

Also, I do not really see why the wizard package exists.

Issue #824, opened 2016-02-16
MimeTypeDetector fails for local URIs


henini:
When a URL encoded local URI is passed to the function

`MimeTypeDetector.isPdfContentType(String url)`

the return value turns out to be False if `url` is a local URI. 
If one likes to implement `FullTextFinder` using a temporary copy of some fetched PDF this needs to be changed.

stefan-kolb:
You are thinking of URIs like `file:/C:/Users/Stefan/Desktop/Github/jabref/build/resources/test/net/sf/jabref/logic/io/empty.pdf`?

henini:
Exactly, the current implementation seems to work for http URLs only.

I also tried to write a test, but trying to get absolute paths in Java annoyed me a bit.

stefan-kolb:
The question is what URLs/Paths should MimeTypeDetector be able to categorize.
Is `C:/Users/Stefan/Desktop/Github/jabref/build/resources/test/net/sf/jabref/logic/io/empty.pdf` also a valid input here? Do we really need to check local URLs and paths? At the moment there is no use case (FullTextFinder) for this.

henini:
URLConnection works for local URIs as well as http URLs.
Some institutional APIs return a PDF file instead of a link to a PDF. Generalising MimeTypeDetector eases writing custom fetchers a bit.

simonharrer:
:+1: for adding a test :)

tobiasdiez:
Mhhh... the code checks the HTTP header `Content-Type` and thus fails for local URIs since these don't have a HTTP header.

To be honest, I don't see the use-case to allow local URIs. Are you trying to check that a file with `.pdf` ending is really a pdf? Can you please elaborate a bit more on what you are trying to achieve.

henini:
`java.net.URLConnection` has `getContentType()`, which returns the Content-Type of a file:// URI as well as the Content-Type of a http:// URL, replacing the code in MimeTypeDetector does not break the current tests

stefan-kolb:
I already created a test, but will only push it later so it doesn't break the build.

PR #828, opened 2016-02-18
MimeTypeDetector Improvements

stefan-kolb:
See #824.

- added a http stubbing library
- added a redirection test
- added a mime type variation test
- changed detection implementation to HTTPUrlConnection

Looks like it works...


simonharrer:
@koppor do you know how to host files and redirection on www.jabref.org?

koppor:
File hosting: Even though it is possible to add files by adding them to https://github.com/JabRef/www.jabref.org (gh-pages branch), I would propose to create a new URL `test.jabref.org`, where we host such files. We should not mix pages.

Redirections: They are possible at GitHub using HTML redirects: https://help.github.com/articles/redirects-on-github-pages/. However, these are not the redirects, we are aiming for. I can setup `test.jabref.org` on my private server and create some redirects.

stefan-kolb:
Ok, we can mock the HTTP request via WireMock. Should be way better than having an external dependency to our homepage.

PR #830, opened 2016-02-19
Added Ordinal formatter

oscargus:
Motivated by http://tex.stackexchange.com/questions/292491/numerals-to-ordinals-in-jabref-jstyle-layout-file


simonharrer:
Should be documented in the documentation that we have an additional LayoutFormatter available.

oscargus:
Yeah, that will be a separate PR (or can one do a joint PR to two different repositories?). Added a ChangeLog entry though.

simonharrer:
:+1: I would link the other PR for the documentation.

oscargus:
(English) Documentation at JabRef/help.jabref.org#5

PR #832, opened 2016-02-19
Warning removals and fixed an issue with pushing citations to OpenOffice

oscargus:
Removed some Coverity warnings and fixed an issue in OpenOffice (and possibly elsewhere) where getSelectedRows() did not return a copy of the BibEntries (I introduced that when changing arrays to lists), and, hence, some entries was deselected and never pushed.

Another old major issue in OpenOffice was also solved, which relates to crossrefs and strings. Not sure if it was ever triggered, but there was one fundamentally wrong access to a Map.


src/main/java/net/sf/jabref/exporter/layout/LayoutHelper.java
```diff
@@ -80,7 +80,7 @@ public static void setCurrentGroup(String newGroup) {
         LayoutHelper.currentGroup = newGroup;
     }
 
-    private String getBracketedField(final int field) throws IOException {
+    private void getBracketedField(final int field) throws IOException {
```
tobiasdiez: I would rename these methods since they are not returning anything and so "get" is misleading.

src/main/java/net/sf/jabref/exporter/layout/LayoutHelper.java
```diff
@@ -119,13 +119,13 @@ private String getBracketedField(final int field) throws IOException {
             }
         }
 
-        return null;
+        return;
```
tobiasdiez: can be deleted completely ;-)

PR #835, opened 2016-02-19
Remove ide-settings as they are automatically generated

koppor:
I removed the files in `ide-settings` as they differ from the ones generated with `gradlew eclipse`.

Are they needed for IntelliJ or can they be removed?


simonharrer:
They could be imported from IntelliJ IDEA, but I think nobody does this. Hence, they can be removed.

PR #837, opened 2016-02-19
Added tests for some formatters

oscargus:


src/test/java/net/sf/jabref/exporter/layout/format/DefaultTest.java
```diff
@@ -0,0 +1,37 @@
...
+import net.sf.jabref.exporter.layout.ParamLayoutFormatter;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+public class DefaultTest {
+
+    @Test
+    public void testSimpleText() {
+        ParamLayoutFormatter a = new Default();
```
tobiasdiez: I would use here Default as the type instead of ParamLayoutFormatter. This makes is very immediate which class is tested...
Applies also to other tests.

src/test/java/net/sf/jabref/exporter/layout/format/WrapContentTest.java
```diff
@@ -0,0 +1,57 @@
...
+
+import org.junit.Assert;
+import org.junit.Test;
+
+public class WrapContentTest {
+
+    @Test
+    public void testSimpleText() {
+        ParamLayoutFormatter a = new WrapContent();
+        a.setArgument("Bob,Ben");
```
tobiasdiez: Could you please use more real-world examples as arguments, like " < , > ". Makes the tests more readable.

tobiasdiez:
:+1: 
Looks good to me. I made some very small remarks to improve readability of the tests. The naming could be improved at some points to also include the expected behavior (i.e. naming along methodUnderTest_TestCase_ExpectedBehavior...for example removeWhitespaceFromNullThrowsException)

src/test/java/net/sf/jabref/exporter/layout/format/DefaultTest.java
oscargus: I see the point. However, I believe that the file name naming makes it quite clear. (And the reason for keeping it is to quicker test new formatters...)

src/test/java/net/sf/jabref/exporter/layout/format/WrapContentTest.java
oscargus: I've tried to make it more relevant/real world/readable now. :-)

oscargus:
Thanks for the input! I've renamed some of the tests. While I agree with you that accurate naming is good, here there would be lots of "FormatTextExpectFormatted"... I've changed the name where it might not be "obvious" what the expected result would be (for some definition of obvious...).

tobiasdiez:
I know the naming can be a hassle. Normally, one wants descriptive test names so that the name of a failing test directly tells you what went wrong (without looking at the source code of the test). I think this is now the case with (most) tests you added.

Issue #839, opened 2016-02-20
Move all selected groups


tobiasdiez:
There is a small bug in the group interface.

Steps to reproduce: Select multiple groups -> Rightclick on one group -> Move -> Down
Behavior: only the rightclicked group is moved down
Expected behavior: all selected groups are moved down by one

tobiasdiez:
I think this is not that important right now.

tobiasdiez:
Move command was removed completely.

PR #847, opened 2016-02-20
Refactor group package

tobiasdiez:
The main aim of this PR is to separate the groups package in gui and logic parts. To accomplish this, the following changes were made:
- All the UI related code from the GroupTreeNode class is moved to a new class GroupTreeNodeViewModel, which functions as an adapter from GroupTreeNode to Swing's TreeNode class.
- Add classes EntriesGroupChange and MoveGroupChange which store the changes made to the group. The corresponing UI classes UndoableChangeAssignment and UndoableMoveGroup now work as converter to Swing's AbstractUndoableEdit.
- Add TreeNode class which is more a less a (typed) copy of Swing's DefaultMutableTreeNode and which contains all the logic for traversing the group tree.
- Implement observer interface for changes in the group tree
- Some smaller refactorings like moving method from class A to B.
- Add tests for the logic classes GroupTreeNode and TreeNode

Since this PR makes quite some changes and there are no tests in place, I wouldn't merge this PR if a new version should be released the next day :smiley: . All the usual functions should work though. 
For some reason github shows some of the moved files as removed + added. Locally, gitk shows a correct diff.


oscargus:
If you push to upstream there will be an automatic build which will be easier to try out.

simonharrer:
I think we should put all the groups classes in the logic part now under test.

tobiasdiez:
Yes of course there should be tests.... but I think this will be something for a next PR.

simonharrer:
:) I wanted to stop the mentality to postpone tests as the consequences are that if we want to release, we release untested code or have to keep track of the pull requests to come for the tests. Maybe this is something for the devcall.

tobiasdiez:
The new TreeNode class should now have a test coverage of 100%. Of course there are still many untested lines of code in the groups package...but one could not hope to change the whole JabRef world in one PR :smile: 

Anyway this PR now ready for review

src/main/java/net/sf/jabref/exporter/BibDatabaseWriter.java
```diff
@@ -289,7 +289,7 @@ private void writeMetaData(Writer out, MetaData metaData) throws IOException {
         // write groups if present. skip this if only the root node exists
         // (which is always the AllEntriesGroup).
         GroupTreeNode groupsRoot = metaData.getGroups();
-        if ((groupsRoot != null) && (groupsRoot.getChildCount() > 0)) {
+        if ((groupsRoot != null) && (groupsRoot.getNumberOfChildren() > 0)) {
```
oscargus: Introduce isEmpty()?

tobiasdiez: There is already `isLeaf()` which indicates whether a node as children or not. Not sure if `root.isLeaf()` is easier to understand than `root.numberOfChildren > 0`. Will keep your comment in mind when I`ll revisit the code.

oscargus: OK! You are probably right that isLeaf is harder to understand. Since at
least Collection and String use isEmpty it may be worthwhile. But then it
is not that crucial as one clearly can understand the current code
(obviously).

(Saw at least one more place where isEmpty/isLeaf can be used.)

simonharrer:
Very nice. Looks good to me. If you have time, have a look at the logic.group classes and try to make everything final, especially the *Change classes. Also have a look at the codacy warnings, as some of them make sense.

tobiasdiez:
Ok, I will do the code cleanup later. Should I merge this PR then in or wait until 3.3 is released?

simonharrer:
I think we should wait for 3.3 to be released, which will happen soon I think.

Siedlerchr:
Does this also fix the issue with moving groups in #839 ?

tobiasdiez:
No, I think the bug is still not fixed with this PR.

simonharrer:
This can be merged now.

tobiasdiez:
Done.

simonharrer:
:+1:

PR #848, opened 2016-02-20
Code cleanups

oscargus:
Yet another bunch of code cleanups. Some logging, some use of high level methods, some refined exception catching.


oscargus:
As an added bonus, the special field buttons are enabled and disabled depending on if any database is open.

PR #849, opened 2016-02-20
Even more cleanups

oscargus:
Installed SonarLint, so more hints and warnings appear...

### DIFF
--- a/src/main/java/net/sf/jabref/bst/BibtexCaseChanger.java
+++ b/src/main/java/net/sf/jabref/bst/BibtexCaseChanger.java
@@ -54,13 +54,14 @@ public final class BibtexCaseChanger {
 
         private final char asChar;
 
+        FORMAT_MODE(char asChar) {
+            this.asChar = asChar;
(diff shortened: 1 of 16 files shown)


src/main/java/net/sf/jabref/bst/BibtexCaseChanger.java
```diff
@@ -157,8 +158,10 @@ private String doChangeCase(String s, FORMAT_MODE format) {
      * @param format
      * @return
      */
-    private int convertSpecialChar(StringBuffer sb, char[] c, int i, FORMAT_MODE format) {
-        assert(c[i] == '{');
+    private int convertSpecialChar(StringBuilder sb, char[] c, int start, FORMAT_MODE format) {
+        assert (c[start] == '{');
```
simonharrer: asserts can be removed.

oscargus: OK! I'll do that.

Issue #574, opened 2015-12-23
Rethink displayable fields vs. writeable fields


koppor:
JabRef has following flags for fields to distinguish between to be shown in the BibTeX source tab and to be written to a file.

In `net.sf.jabref.gui.BibtexFields` this is shown as follows:

``` java
private static final int STANDARD = 0x01; // it is a standard bibtex-field
private static final int PRIVATE = 0x02; // internal use, e.g. owner, timestamp
private static final int DISPLAYABLE = 0x04; // These fields cannot be shown inside the source editor panel
private static final int WRITEABLE = 0x08; // These fields will not be saved to the .bib file.
```

The comment at `DISPLAYABLE` and `WRITEABLE` should IMHO be the other way round.

With #391, the source editor tab also shows write-only fields when the entry was not modified:

![grabbed_20151223-181053](https://cloud.githubusercontent.com/assets/1366654/11981192/b43288f0-a9a0-11e5-9e61-ae1b51177bab.png)

A possible solution is to rethink non-displayable fields. My current opinion is to write these markings in a separate `@comment` entry (`jabref-meta`) or as comment above the entry itself to keep the original entry clean. Reason: (i) For me as user it is not understandable why the displayed rendering is different from the writing in the `.bib` file. (ii) For me as developer I do not want to offer easy change JabRef's internals. Therefore, I'd like to separate JabRef's markings from the entry data. This should remove the `DISPLAYABLE` and `WRITEABLE` flags.

Finally, the flags `STANDARD` and `PRIVATE` should be obsolete and removed.

tobiasdiez:
In #629 I actually propose the converse for groups: serialize the explicit group membership as a field `groups` in the entry instead of a comment at the end. In my opinion the advantages of such an approach (as listed in #629) outweigh your concerns. In particular, concerning your remarks in the last paragraph:
1) This can be solved easily by displaying all the fields in the "Bibtex" tab.
2) From time to time I find it quite convenient to change the properties of an entry using the bibtex code instead of the UI. 

I think this is something for the next dev-call.

koppor:
Decision: show all fields being written. It has to be the same serialization as being written.

TypedBibEntry has no own state. State only at BibEntry.

If possible, remove `isSearchHit()`.

PR #851, opened 2016-02-20
Ignore non-displayable fields for indent during writing

### LINKED ISSUE #834: Writeable fields should not cause rewriting of the whole entry
When an entry is unmarked, the field `__markedentry` is removed. If all other fields are shorter, the complete entry changes as the alignment changes. I am aware that we decided that at #116. We have, however, not thought about such exceptional cases.

Example:
grabbed_20160219-201705
(I know that a little bit more changed here. My database is written as BibLaTeX database).

Suggestion: Do not include non-displayable fields in the calculation of the amount of spaces to write.

Refs #574

koppor:
Fixes #834

### DIFF
--- a/src/main/java/net/sf/jabref/bibtex/BibEntryWriter.java
+++ b/src/main/java/net/sf/jabref/bibtex/BibEntryWriter.java
@@ -154,9 +154,16 @@ public class BibEntryWriter {
         }
     }
 
+    /**
+     * Determine length of longest field key.
+     * Only displayable fields are regarded as writable-only fields are not shown in the UI, but the user expects the entry to be rendered as shown in the UI
+     * This is a quick fix for https://github.com/JabRef/jabref/issues/834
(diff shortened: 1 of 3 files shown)


simonharrer:
Long term fix would be to not indent at all, and just write `key = value`.

koppor:
The indent has been agreed upon during the discussion at #116 and it seems to be liked by users [as there was a question regarding it on tex.stackexchange](http://tex.stackexchange.com/q/294906/9075).

No indent would us bring closer to original bibtex format, but away from the biblatex format.

Example from page 32 of biblatex's manual:

grabbed_20160228-152131

koppor:
Decision: Do not merge. Do the long term fix described at https://github.com/JabRef/jabref/issues/574#issuecomment-193718235.

koppor:
Refs #1043

PR #852, opened 2016-02-20
[WIP] Add Quick cleanup

tobiasdiez:
Adds the ability to run preconfigured cleanup procedures on the currently edited entry.

In detail:
- Add button in entry edit toolbar, which directly runs a preconfigured
cleanup on the  current entry (without asking for confirmation)
- Add "Cleanup" tab in preferences to configure the quick cleanup


simonharrer:
Hm, why should we store two cleanup preference sets? I would propose to simplify it to use the same cleanup preferences for the quick cleanup and the full cleanup.

tobiasdiez:
Actually, two separate cleanup sets was the main selling point for me. I often change between a "full" cleanup (This makes however sometimes some wrong changes which I have to revert using version control) and a smaller cleanup which I run after every import. In the long term I suspect we will have quite a bunch of cleanup presets for import, save, normal cleanup, export, ...

tobiasdiez:
So what happens with this PR? Should I rebase?

tobiasdiez:
- Add option under Quality
- Change to apply save actions

simonharrer:
@tobiasdiez could you implement this maybe targetting 3.3? Could be a low hanging fruit, and reducing the large amount of open PRs. :)

tobiasdiez:
No, not for 3.3.
I think it is better to wait with this PR until the SaveActions are completely converted to cleanup operations (right now they are only a subset).

koppor:
I think, it can be merged, even though not all SaveActions are converted to cleanup operations. Is there an issue for it?

tobiasdiez:
We decided in a devcall (or private call with @simonharrer?) that an additional quick cleanup causes confusion. Thus the idea was to have a button in the entry editor which directly applies the save actions.

stefan-kolb:
Closing this issue due to inactivity :zzz: 
Please reopen the issue with additional information if the problem persists.

PR #853, opened 2016-02-21
Add a few cleanup formatters

tobiasdiez:
This PR adds the following formatters, which are suitable for the new save actions functionality #821:
- TrimFormatter: trims whitespace (as suggested in https://github.com/JabRef/jabref/pull/821#issuecomment-185801238)
- RemoveBracesFormatter: removes enclosing braces (i.e. "{{test}}" -> "test"). Partly implements [kooper #60](https://github.com/koppor/jabref/issues/60). This functionality was previously available under the option "Remove double braces around BibTeX fields when
loading" but was removed from the parser in c2928d2153301b3c5e90db468b53199bc8b1d521.

Also add some tests for all formatters.

### DIFF
--- a/src/test/java/net/sf/jabref/importer/HTMLConverterTest.java
+++ b/src/test/java/net/sf/jabref/importer/HTMLConverterTest.java
@@ -25,11 +25,6 @@ public class HTMLConverterTest {
     }
 
     @Test
-    public void testHTMLNull() {
-        assertEquals(null, conv.format(null));
-    }
-
-    @Test
     public void testHTMLEmpty() {
         assertEquals("", conv.format(""));
     }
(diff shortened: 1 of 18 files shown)


src/test/java/net/sf/jabref/importer/HTMLConverterTest.java
```diff
@@ -25,11 +25,6 @@ public void testBasic() {
     }
 
     @Test
```
oscargus: Keep and add (expected...)?

oscargus: Ahh, now I saw the next file... ;-)

oscargus:
:+1:

Now I only want a formatter that removes ' around fields. A colleague does that as that also keeps the casing... However, I guess it may be a rare use case for the general public...

simonharrer:
:+1: nice. I especially like the data driven test of all formatters! Good idea.

stefan-kolb:
LGTM :+1:

Issue #454, opened 2015-12-03
Entry Editor: Tab for remaining fields


simonharrer:
At the moment, each BibtexEntry has required and optional fields. The remaining fields can only be seen in the source tab of the entry editor. 

We should add a tab to the entry editor which shows all the remaining fields and provide the ability to change them. 

This is a good task to start developing for JabRef.

PR #858, opened 2016-02-23
Tab for remaining fields

stefan-kolb:
See #454.


simonharrer:
When is this updated? Will this update on the fly if I change the config?

stefan-kolb:
Don't think so as the FieldEditor is somehow cached i think.

lenhard:
What is the state here? What is missing to get this merged?

simonharrer:
@stefan-kolb is working on it. I removed the label to make it clear that is currently under work.

stefan-kolb:
Think its ok to merge.

PR #863, opened 2016-02-24
More integrity checks

simonharrer:
Adds more integrity checks:
- detect abbreviations in journal field
- detect abbreviations in booktitle field
- detect wrong usage of proceedings entry type
- detect broken file links

- [x] Tests green?


src/test/java/net/sf/jabref/logic/integrity/IntegrityCheckTest.java
```diff
@@ -72,6 +78,33 @@ public void testTitleChecks() {
...
+        assertCorrect(":build.gradle:gradle", IntegrityCheck.FILE_CHECKER, context);
+        assertWrong(":asflakjfwofja:PDF", IntegrityCheck.FILE_CHECKER, context);
+    }
+
+    @Test
+    public void testTypeChecks() {
+        LinkedList<IntegrityMessage> messages = new LinkedList<>();
+        IntegrityCheck.TYPE_CHECKER.check("11--15","pages",new BibEntry("asdfasdf", "inproceedings"), messages, new BibDatabaseContext());
+        assertEquals(Collections.emptyList(), messages);
+
```
tobiasdiez: Please split the test

src/main/java/net/sf/jabref/logic/integrity/IntegrityCheck.java
```diff
@@ -43,36 +48,83 @@
...
+        entry.getFieldOptional("pages").ifPresent(data -> TYPE_CHECKER.check(data, "pages", entry, result, bibDatabaseContext));
 
-        entry.getFieldOptional("url").ifPresent(data -> URL_CHECKER.check(data, "url", entry, result));
+        entry.getFieldOptional("journal").ifPresent(data -> ABBREVIATION_CHECKER.check(data, "journal", entry, result, bibDatabaseContext));
+        entry.getFieldOptional("booktitle").ifPresent(data -> ABBREVIATION_CHECKER.check(data, "booktitle", entry, result, bibDatabaseContext));
 
         return result;
     }
 
     public interface Checker {
```
tobiasdiez: Wouldn't it make sense to extract the interface and the checker classes to external files?

src/main/java/net/sf/jabref/logic/integrity/IntegrityCheck.java
```diff
@@ -43,36 +48,83 @@
...
+        void check(String value, String fieldName, BibEntry entry, List<IntegrityMessage> collector, BibDatabaseContext bibDatabaseContext);
```
tobiasdiez: I would simplify the interface to 
List<IntegrityMessage> check(BibEntry, BibDatabaseContext)
the rest of parameters can be inserted via the constructor.

This would make some of the checkers more readable. For example the ProceedingsChecker can then check if the page field is present and the type is proceedings and don't have to outsource one part of the check.

src/main/java/net/sf/jabref/logic/integrity/IntegrityCheck.java
```diff
@@ -18,18 +21,20 @@
     public static final Checker BRACKET_CHECKER = new BracketChecker();
```
tobiasdiez: Is there a need for these public static fields? There seems to be no value in keeping a static reference...

src/test/java/net/sf/jabref/logic/integrity/IntegrityCheckTest.java
simonharrer: At the moment, we do have a nice parallel structure, as each integrity check has its own test method in a single test class. If I would split this one test, I would loose this structure. Hence, I think it would make matters worse.

src/main/java/net/sf/jabref/logic/integrity/IntegrityCheck.java
simonharrer: At the moment, the checker classes are all private, hiding their implementation. I do not want to expose this currently private information, except if I have to.

src/main/java/net/sf/jabref/logic/integrity/IntegrityCheck.java
simonharrer: They are currently needed, as the caller cannot access the class directly because it is private.

src/main/java/net/sf/jabref/logic/integrity/IntegrityCheck.java
simonharrer: The problem is that the ProceedingsChecker is the exception, not the rule. With your change, the ProceedingsChecker would be more readable, but I would loose the statelessness of the checkers. I am not sure, if this improves the situation. What is more, the check whether the field is there as to be implemented in each checker over and over again, creating noise in the check logic. 

Passing the list as a collector I also like because otherwise one has to create a lot of lists and combine them. The collector pattern is used by most test frameworks as well, e.g., JUnit does this internally.

tobiasdiez: Mhhh...the only checker which operates on two fields is the AuthorNameChecker (and maybe the BracketChecker), so the parameter "fieldName" is in most cases redundant. Moreover, the value parameter is always entry.getField(fieldName). I would also move the code
entry.getFieldOptional("year").ifPresent( ... )
to the checker.

src/main/java/net/sf/jabref/logic/integrity/IntegrityCheck.java
tobiasdiez: But right now the checker are not completely private but exposed via the public fields. So a user of IntegretyCheck does know about underlying Checker. 
If these checker are in separate files, maybe with package-private, then nobody in the outside world knows about these checkers.

But anyway, this is just a small thing...

oscargus:
Maybe it should be possible to configure which integrity checks one would like to use? (Not in this PR, but as a general idea.)

What happens if there's a string? Is it resolved?

src/main/java/net/sf/jabref/logic/integrity/IntegrityCheck.java
simonharrer: I will try it out locally and see how it compares to the previous code.

simonharrer:
We could add an option to the Integrity Check UI which ones to display or which type of check to ignore. I would always do every check as they do not take very long.

simonharrer:
@tobiasdiez thank you for all your comments here. I think I got a much better design now - which I think is better than the one before. What do you think?

tobiasdiez:
Looks really good! Two very small comments:
- The Checker interface can now be made private, right?
- I don't like the naming assertWrong / Correct. How about assertHasErrors? But this is a little bit longer...

Can be merged in from my point of view.

simonharrer:
Hm, good points. But I currently have no better naming in mind. I want to keep the word *assert* and keep it short. What about assertValid vs. assertInvalid?

Nevertheless will I merge this as this is an interesting improvement of the integrity check functionality.

Issue #859, opened 2016-02-24
Organizing the Preferences


simonharrer:
Currently, each entry in the preferences follows the pseudo model below:

```
class PreferenceEntry {
   String key;
   Object value;
   Type type; // can be String, Integer or Boolean
   Optional<Object> defaultValue; // must be of same type, is optional
}
```

The current issues with this implementation
- have no control over all keys; most keys are constants in the JabRefPreferences class, but some keys are computed at runtime with user-defined parts (either starts or ends with a specific key)
- we do not know which keys are still in use with jabref and which are not; I tried to define an unused key by relying on the absence of a default value, but as we do not have control over all the keys, this is doomed to fail
- no tracking on when a preference is added and when it has been removed, making migration really hard.
- if you try to read a boolean as a string, an exception will occur. 

What we could do:
- make everything more explicit through a class which stores for each PreferenceEntry the version in which it was introduced and also removed 
- regain control over all preference keys making a check routine using equals and for some specific ones startswith and endswith
- show to the user which preferences he has overridden, i.e., which preference entries do deviate from their default values. These are the interesting ones, and only they really help in debugging issues. Maybe through a table in the UI.
- use the type information to reduce errors caused by getting prefs using the wrong type

We are open for help in that case. :)

PR #864, opened 2016-02-24
Add dialog to show all preferences in their raw form plus some filtering

simonharrer:
Relates to #859.

- Adds a dialog which shows all preferences in a table by their type, key, value and default value. 
- Table can be sorted
- Table can be filtered to show only the preferences which deviate from their default value. This is the most helpful one - and it revealed various bugs when having a closer look. :)
- Count is shown as well
- Reachable through Options -> Preferences -> Show Preferences (button is below import/export preferences)
- Converts default values of null to empty string

- [ ] Tests created for changes? Mostly GUI changes, testing makes not that much sense here
- [x] Tests green?


lenhard:
Looks good to me! Just rebase on master and this can be merged.

src/main/java/net/sf/jabref/JabRefPreferencesFilterDialog.java
```diff
@@ -0,0 +1,115 @@
...
+        table = new JTable();
+        table.setAutoCreateRowSorter(true);
+        updateModel();
+        panel.add(new JScrollPane(table), BorderLayout.CENTER);
+
+        this.getContentPane().add(panel);
+    }
+
+    private void updateModel() {
+        java.util.List<JabRefPreferencesFilter.PreferenceOption> preferenceOptions;
```
oscargus: No need for fully qualified path?

src/main/java/net/sf/jabref/JabRefPreferencesFilterDialog.java
```diff
@@ -0,0 +1,115 @@
...
+
+        private final List<JabRefPreferencesFilter.PreferenceOption> preferences;
+
+        public PreferencesTableModel(List<JabRefPreferencesFilter.PreferenceOption> preferences) {
+            this.preferences = Objects.requireNonNull(preferences);
+        }
+
+        @Override
+        public String getColumnName(int column) {
+            if (column == 0) {
```
oscargus: switch?

src/main/java/net/sf/jabref/JabRefPreferencesFilterDialog.java
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+        @Override
+        public Object getValueAt(int rowIndex, int columnIndex) {
+            if (rowIndex < 0 || rowIndex - 1 > preferences.size()) {
+                return "n/a";
+            }
+
+            JabRefPreferencesFilter.PreferenceOption preferenceOption = preferences.get(rowIndex);
+            if (columnIndex == 0) {
```
oscargus: switch?

src/main/java/net/sf/jabref/JabRefPreferences.java
```diff
@@ -543,8 +550,8 @@ private JabRefPreferences() {
         defaults.put(ENTRY_TYPE_FORM_WIDTH, 1);
         defaults.put(BACKUP, Boolean.TRUE);
         defaults.put(OPEN_LAST_EDITED, Boolean.TRUE);
-        defaults.put(LAST_EDITED, null);
-        defaults.put(LAST_FOCUSED, null);
+        defaults.put(LAST_EDITED, "");
```
oscargus: Added twice (subject to only seeing fractions of the code...).

oscargus:
Looks good! Some minor comments, which probably can be ignored/fixed later.

src/main/java/net/sf/jabref/JabRefPreferencesFilterDialog.java
simonharrer: Good point. Will fix this.

src/main/java/net/sf/jabref/JabRefPreferences.java
simonharrer: Will have a look, thanks.

src/main/java/net/sf/jabref/JabRefPreferencesFilterDialog.java
simonharrer: Na, I don't like switch statements, as one can miss the break.

Issue #496, opened 2015-12-10
Main PDF Directory: and Main PS directory: preference settings removed. Needed for heterogeneous repositories.


ajbelle:
As part of checking "issue Attached File Delete functionality #245" I note the Options>Preferences>[External programs] Legacy file fields --------------- are no longer showning, nor do they appear in File>Database Properties.  Settings for my repositories have carried across from my ver2.10 install, BUT WHAT DO I DO IN FUTURE? If I move the Endnote file repository I have in "Main PDF Directory:" and "Main PS directory:" I can edit the directory directly in "Exported preferences.xml” with PFE and reimport, but is all support going to terminate!
I have used these fields for heterogeneous linking to a large Endnote Database (not the intended purpose), so retention of these additional paths is useful functionality to me. All I found on the internet were changes to the bibtex file PS and PDF entries. Is there an alternative method to link a second file repository? The helpfile of my V.30 install is not getting past the second page and https://fossies.org/linux/misc/JabRef-3.0.jar/help/en/ExternalFiles.html is the old information.
This maybe should be a forum post.

lenhard:
Yes, all support for these things is terminated. You can put multiple file links into the `file` field of a bibtex entry. You just need to separate the path via a semicolon.

ajbelle:
Thanks lenhard:-( I am not referring to having multiple files linkedin the file field, I was asking after the additional paths to directory locations other than set in options>Preferences>External programs"Main file Directory" that the  "Main PDF Directory:" and "Main PS directory:" settings allowed (VERY HANDY EVEN THOUGH NOT THEIR INTENT). 
Can you please contirm your reply means that __I have to apply the full file specification from the root__ for the thousand+ files I have in my separate Endnote directory (outside the "Main file Directory") . Am I missing something?

matthiasgeiger:
I'm not sure whether I understand your use case correctly... You have used the "legacy fields" `pdf` / `ps` to define file links which are based outside your set "main file directory"?

So your entries have both: a file field **AND** a pdf/ps field pointing relatively to different files (in different folders)?

If so, then you cannot retain this behavior with JabRef 3.0+.

Or do you have only the pdf/ps field **but no additional file link**? So what you basically want is to define database specific "main directories" which are used for relative linking? 

If so, you should use the functionality to define a database specific main directory. Accessible via right-click on the database title tab -> "Database properties".

simonharrer:
[...]

PR #865, opened 2016-02-25
Deleted code related to the pdf and ps fields

oscargus:
Dropping the support for ps and pdf fields (which seems dropped already as hardcoding a pdf-field in a .bib-file doesn't even show up in the entry editor) seemed to remove quite a bit of code.


simonharrer:
I am suprised that so much code is unused. But I do like it that we can throw so much unused code away! :)

ajbelle:
I suspect that in all that code JabRef was providing the multiple root directory access functionality discussed in https://github.com/JabRef/jabref/issues/496 :-)

oscargus:
Could be like that, yes. Doesn't it work for you anymore?

ajbelle:
Sorry, so near end of my PhD I don't dare update, for fear of loosing access to thousands of references. My understanding is that the functionality I am using was 'never there' (none on the team understood what I was on about, and hence the lengthy  [#496](https://github.com/JabRef/jabref/issues/496)). I  thought you may have seen something that may have been enabling it, and was forewarning.

ajbelle:
@oscargus Over Easter I updated to ver3.2 to try fix a Search issue. My install still finds the files I have residing in the old legacy pdf file structure that is not in the same tree as the JabRef files! No symbolic or hardlinks present, so I believe in magic! :-)  It must be coded elsewhere. happy Easter.

oscargus:
@ajbelle this PR is not in 3.2 so maybe a bit too early to be fully relieved. ;-) However, I know of some code which currently supports different directories for different file types and I won't remove that. Not sure if that code is enough, but it might in fact be.

oscargus:
This code was primarily related to the UI side, so any buttons magically showing up next to the PDF-field etc.

PR #874, opened 2016-02-26
Fixing OpenOffice Citation with curly braces and manual connection browsing

### LINKED ISSUE #815: Curly braces ignored in in-text citations
In the reference,

```
@Manual{JabRef_software,
  Title                    = {JabRef},
  Author                   = {{JabRef Development Team}},
  Year                     = {2015},
  Version                  = {Version 2.11.1},
  Url                      = {http://www.jabref.org}
}
```

The curly braces in the Author field are ignored, causing, for example, its associated in-text citation to be presented as "(Team, 2015)", rather than, "(JabRef Development Team, 2015)". Curly braces should be honored in in-text citations.

### LINKED ISSUE #856: OpenOffice Manual connect  - Clicking on browse does nothing
While setting up my Libre Ofifce connection around for #815 I noticed that Browse does nothing.
However, directly inserting the correct path works.

There is also some odd behaviour with Autodetection.
When I chose the correct path for Autodection it displays an error message " Autodection failed", but it still connects.

Siedlerchr:
- [X] Change in CHANGELOG.md described?
- [X] Changes in pull request outlined? (What, why, ...)
- [X] Tests green?

Fixed #856: Clicking on browse for manual OpenOffice does now work
Fixed #815: Curly braces in OO/LO citation no longer ignored
- Code cleanup: Replaced ActionListener with Lamdbas

One point for discussion: DefaultFormBuilder (jgoodies) is deprecated. Is there already a replacement for the creation? I found this thread:
http://stackoverflow.com/questions/33044573/jgoodies-what-is-the-nextline-equivalent-in-formbuilder
@oscargus  @koppor

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,8 +56,9 @@ to [sourceforge feature requests](https://sourceforge.net/p/jabref/features/) by
 - Fixed [#803](https://github.com/JabRef/jabref/issues/803): Fixed dynamically group, free-form search
 - Fixed [#743](https://github.com/JabRef/jabref/issues/743): Logger not configured when JAR is started
 - Fixed [#822](https://github.com/JabRef/jabref/issues/822):OSX - Exception when adding the icon to the dock
-
 - Fixed [#685](https://github.com/JabRef/jabref/issues/685): Fixed MySQL exporting for more than one entry
+- Fixed [#815](https://github.com/JabRef/jabref/issues/815): Curly Braces no longer ignored in OpenOffice/LibreOffice citation
+- Fixed [#855](https://github.com/JabRef/jabref/issues/856): Fixed OpenOffice Manual connect - Clicking on browse does now work correctly
 
 ### Removed
 - Fixed [#627](https://github.com/JabRef/jabref/issues/627): The pdf field is removed from the export formats, use the file field
--- a/src/main/java/net/sf/jabref/logic/util/strings/StringUtil.java
+++ b/src/main/java/net/sf/jabref/logic/util/strings/StringUtil.java
@@ -411,12 +411,35 @@ public class StringUtil {
         return LINE_BREAKS.matcher(s).replaceAll(Globals.NEWLINE);
     }
 
+    /**
+     * Checks if the given String has exactly one pair of surrounding curly braces <br>
+     * Strings with escaped characters in curly braces at the beginning and end are respected, too
+     * @param toCheck The string to check
+     * @return True, if the check was succesful. False otherwise.
+     */
     public static boolean isInCurlyBrackets(String toCheck) {
+        int count = 0;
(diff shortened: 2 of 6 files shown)


oscargus:
Regarding DefaultFormBuilder: there is an alternative JGoodies approach. However, it has been decided to leave it as is as a Javax conversion is planned (any year now...).

I haven't looked at the code carefully, but what happens with {} around e.g. capitals?

Siedlerchr:
Ah thanks for your hint. Yeah, it seems like it needs more checking for special cases. I will try to come up with a solution to this. 
I noticed that the authors name Arp{\'\i}rez, then appears as Arp{i}rez et al. [2003]

Siedlerchr:
I tried some variants, e.g. adding the curly braces "manually" before and end.

` if (StringUtil.isInCurlyBrackets(field)) { //braces at the start and the end of the string
        
            sb.insert(0, "{");
            sb.append("}");
            System.out.println("In Curly Braces SB " + sb);
        }
        return sb.toString().replace("&dollar;", "$"); // Replace &dollar; with $`

this works well for Authors. However ,the problem is that I have no control over the field type. This creates then a problem when you have an entry with curly braces around the Title Field. Those curly Braces would then be printed in the generated Bibliography...
Any ideas appreciated.

oscargus:
I haven't tried it, but I think that OOBibBase.getCitationMarkerField is the place to try. Basically:

1. Check if there are surrounding braces.
2. Format (using the old formatter that removes all braces).
3. If there were surrounding  braces, add them back before returning the formatted string.

A potential problem is author lists which starts and ends with braces, but are not enclosed, say `{\AA}sa Stor{\aa}`, but there are methods for checking that in #863 (IntegrityCheck.BracketChecker), so as long as the basic mechanism is working one can always improve step 1 later.

Siedlerchr:
Thanks for your tip. I think I found a working solution. 
In fact I copied the method and added a Check for curly braces (code is based on the integrity checker), like you proposed.
The copied method is 
Manual tests look promising. I will add some unit tests.

Siedlerchr:
Found a solution by enhancing the StringUtil method and by adding a check for Curly brackets in
`getAuthorCitationMarkerField`

The only thing which bothers me, is that it is kind of "double code". I thought about this, but did not come up with a nice solution. Maybe one of you has an idea.

oscargus:
I think you can simply check if field is "author" or "editor" in the old method. Not 100% correct but there are other methods using similar logic.

(I guess ideally each field of the predefined entry types should have properties, e.g. if it is a name field.)

oscargus:
Apart from that I think the PR looks good to go!

(I believe that one probably can add a bit of code to handle escaped curly brackets for the rare cases that someone needs to put a single `\{` in a field... But hardly anything important.)

PR #875, opened 2016-02-26
Fix bug in parsing custom entry types

tobiasdiez:
Fix unreported bug which prevented the parsing of custom entry types.
Also move the custom entry type parsing logic to the parser.

- [ ] Change in CHANGELOG.md described? - No, the bug was not present in 3.2.
- [x] Tests green?

### DIFF
--- a/src/main/java/net/sf/jabref/importer/fileformat/BibtexParser.java
+++ b/src/main/java/net/sf/jabref/importer/fileformat/BibtexParser.java
@@ -22,7 +22,6 @@ import java.io.StringReader;
 import java.util.*;
 
 import net.sf.jabref.*;
-import net.sf.jabref.logic.CustomEntryTypesManager;
 import net.sf.jabref.model.database.KeyCollisionException;
 import net.sf.jabref.importer.ParserResult;
 import net.sf.jabref.logic.l10n.Localization;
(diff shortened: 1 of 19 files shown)


oscargus:
Looks good!

I guess one can argue where the method should reside, but since there are arguments for both locations I do not have a strong opinion.

No serious Codacy issues, but one the other hand quite easy to fix.

tobiasdiez:
@oscargus Thanks for the feedback. I now fixed the codacy errors.

simonharrer:
Would it be possible to create a converter class which can transform String <-> CustomEntryType in both directions?

tobiasdiez:
This should be possible yes... I created #888 to discuss where to put the parsing and serialization logic.

src/main/java/net/sf/jabref/importer/fileformat/BibtexParser.java
```diff
@@ -928,4 +927,26 @@ private void consume(char firstOption, char secondOption) throws IOException {
...
+    private CustomEntryType parseEntryType(String comment) {
+        try {
+            String rest = comment.substring(CustomEntryType.ENTRYTYPE_FLAG.length());
+            int indexEndOfName = rest.indexOf(':');
+            String fieldsDescription = rest.substring(indexEndOfName + 2);
+
+            int indexEndOfRequiredFields = fieldsDescription.indexOf(']');
+            int indexEndOfOptionalFields = fieldsDescription.indexOf(']', indexEndOfRequiredFields + 1);
+            if (indexEndOfRequiredFields < 4 || indexEndOfOptionalFields < indexEndOfRequiredFields + 6) {
+                throw new IndexOutOfBoundsException();
```
stefan-kolb: Do not throw an Exception and catch it later on to program control flow.

tobiasdiez:
Based on the discussion during the dev call this morning, I moved the code back to the CustomEntryTypesManager. So this PR now just contains the bug fix and nothing more.

simonharrer:
Looks good to me.

PR #877, opened 2016-02-27
Add ability to run arbitrary formatters as cleanup actions

### LINKED ISSUE #295: "change case" for multiple references
I dont know if this is relevent or not. 
When we want to change the case of titles of serveral references, we need to do "right-click, change case" for every reference, I was wondering if it is possible to change case in a batch way, or for a multiple selection.
This could be added in "Cleanup entries" for example,
Thank you all :)

tobiasdiez:
Add ability to run arbitrary formatters as cleanup actions by reusing save actions. Thus some old cleanup entries could be removed. This also implements/fixes #295. Moreover, an erase formatter is added which just deletes the corresponding field.

untitled

Code changes:
- Rename SaveActions to FieldFormatterCleanups
- Some smaller refactorings to reuse SaveActions (mainly to separate them from MetaData)

:
- [x] Tests green?


lenhard:
I think it would look better if the check boxes were aligned vertically.

tobiasdiez:
untitled
You are right!

simonharrer:
When I compare both UIs (before/after) this PR, we loose a lot of documentation and guidance for the user. Because how, the user has to think which formatters would be well suited for which fields to be helpful, making it harder to configure it correctly. We need to find a way to fight this. Why not add a section about suggestions which can be checked/unchecked and upon change change the table with the actions as well?

tobiasdiez:
All the previous cleanup actions appear as the default value for the field formatters:
```
activeFormatterCleanups.add(new FieldFormatterCleanup("date", BibtexFieldFormatters.DATE));
activeFormatterCleanups.add(new FieldFormatterCleanup("month", new MonthFormatter()));
activeFormatterCleanups.add(new FieldFormatterCleanup("title", new CaseKeeper()));
activeFormatterCleanups.add(new FieldFormatterCleanup("title", new UnitFormatter()));
activeFormatterCleanups.add(new FieldFormatterCleanup("title", new LatexFormatter()));
activeFormatterCleanups.add(new FieldFormatterCleanup("title", new HTMLToLatexFormatter()));
```
So opening the cleanup dialog for the first time shows essentially the same information as previously (this is not apparent from the screenshots since I already removed some of the default values).

I could add a button which resets the field formatters to their default value (or if you prefer adds the default formatters).

src/test/java/net/sf/jabref/logic/formatter/FormatterTest.java
```diff
@@ -85,6 +85,8 @@ public FormatterTest(Formatter formatter) {
                 new Object[]{new SuperscriptFormatter()},
                 new Object[]{new UnitFormatter()},
                 new Object[]{new RemoveBracesFormatter()}
+                // The EraseFormatter is special since it always returns null to indicate that the field should be deleted.
+                // new Object[]{new EraseFormatter()}
```
simonharrer: I do not like the EraseFormatter. I think this should not be seen as a formatter as it clearly violates the interface.

tobiasdiez: You are right, clearing a field is not really a Formatter. I could introduce an interface SingleFieldCleanup which the FormatterCleanup and RemoveFieldCleanup can implement. This new interface can then be used by the SaveActions. Does this sounds like a good solution?

simonharrer: Hm, what happens when a field is empty? Is it removed? I could live with an EraseFormatter if it would always return `""`. Does this make sense?

I do not have all the classes and their reponsibilities in mind for both the formatter and the cleanup. If you need more feedback, you will have to wait as I have to look this stuff up.

oscargus: Not really following the discussion, but it struck me that a functionality that removes all empty fields, i.e., where the field is `""` (or even whitespace) could be useful...

lenhard: I aggree with @oscargus That might be useful as a cleanup action.

src/main/java/net/sf/jabref/MetaData.java
```diff
@@ -30,9 +30,10 @@
 
     public static final String META_FLAG = "jabref-meta: ";
     public static final String SAVE_ORDER_CONFIG = "saveOrderConfig";
+    public static final String SAVE_ACTIONS = "saveActions";
```
lenhard: The save actions are not widely used yet (no UI in a official release), so we might get away with changing the meta key. If we unify the cleanup and save actions, maybe something like "cleanupActions" would be better?

src/test/java/net/sf/jabref/logic/formatter/FormatterTest.java
tobiasdiez: I now changed the eraseFormatter to always returns an empty string. Thus it now meets the Formatter contract.

The RemoveEmptyFieldsCleanup is a good suggestion, which is now tracked at #911.

src/main/java/net/sf/jabref/MetaData.java
tobiasdiez: But nonetheless the cleanup actions would be only invoked when the database is saved. So saveActions is also appropriate. If you want I can change it to the common denominator `saveCleanupActions` ?

lenhard: You are right. In that case it is probably best to leave the name as it is and to avoid the creation of potential migration problems.

simonharrer:
- Make paths of linked files relative if possible could also be a Formatter, and hence, added to the save actions.
- A button which resets to the defaults is a good idea.
- We really need to think how the formatters can be better explained to the user. Would it make sense if each formatter itself has a short localized description how it works, similar to the sentences used before which are shown in the table as well?

tobiasdiez:
New update with:
- Reset button (in cleanup dialog as well as in database properties dialog)
- Add descriptions of the formatters as tooltips in the list and in the combo box (not ideal, but better then nothing).
untitled

Adding the "Make paths relative" cleanup as a save action should be covered with #878. I don't want to implement this cleanup as a formatter (formatters are just in-out converters without any side-effects).

koppor:
- [x] keep name concept. Key is for serialization. Name is for display. Description is a long text.
- [ ] improve keys and names of formatters - separate issue with a table: key and name
  - e.g., `unit` as key
  - e.g., Name: DateFormatter -> NormalizeDateToYYYYMMDD
- [ ] Add description text as box below the line of "Enter field name" updated upon change of dropdown
  - e.g., "Normalizes dates ..."

tobiasdiez:
Added now a description text box. From my side this PR can be merged in. The only missing thing is the naming of the formatters, but I would prefer if this would be a separate PR (not sure how much time I have the next days and since this is a blocker for 3.3....)
untitled

simonharrer:
Last point: what happens when the description is longer than this line? Does it wrap automatically? Does it allow to use newlines for formatting?

tobiasdiez:
Since this is just an JLabel, all the nice limitations and/or features come along. So no, there are no automatic line breaks except if you put everything around <html> tags and you also need to use `<br>` for a new line.

simonharrer:
Hm, this limits the description we can display there. Nevertheless, I think it is best to merge it for now and then do the improvements in another PR. Just rebase and get this in.

koppor:
Is it possible to display a description with "the given field" in place of the field name if no field is given? Reason: I find the empty description irritating and thought that the content of the white box was the description.

Above at "Save sort order", one can select a field. I would have expected the same choice for the selection of the fields to apply the save action. Reason: I am too lazy to input the field name manually.

I experience layout issues at the end - The dialog height should be the size of all content including the white box, shouldn't it?

grabbed_20160313-212741

tobiasdiez:
I added your suggestion about adding "the given field". Could you please open an issue for the other things? Thanks!
I'll now merge this in.

PR #879, opened 2016-02-27
Fix #466: Rename PDF cleanup now also changes case of file name

### LINKED ISSUE #466: Cleanup Entries -> Rename PDFs Doesn't Preserve Key Case
If the filename of a linked PDF is converted to lowercase outside of JabRef, when JabRef runs a Cleanup -> Rename PDFs, it does not rename/move the file back to "proper" upper/lower case consistent with its associated key.

For example, a file with key `Article_2015_NT_Woolley_191--207` will not be renamed to `Article_2015_NT_Woolley_191--207.pdf` from `article_2015_nt_woolley_191--207.pdf`.

tobiasdiez:
Fixes #466.

- [x] Tests green?


simonharrer:
Please remove the unused imports according to the codacy warnings. Otherwise I think it is fine.

PR #884, opened 2016-02-28
Fixed #883

oscargus:
Should solve #883.

- [x] Change in CHANGELOG.md described?
- [x] Changes in pull request outlined? (What, why, ...)
- [x] Tests created for changes? No, tests already passed, this is a GUI issue.
- [x] Tests green? Yes.

PR #887, opened 2016-02-28
Pass database to doCleanup to enable resolving of crossref fields at renaming of files

koppor:
This addresses https://github.com/JabRef/jabref/pull/884#issuecomment-189891290.

Replaces `null` by the concrete database at `CleanWorker`.

No tests as #507 is not yet finished.

### DIFF
--- a/src/main/java/net/sf/jabref/gui/actions/CleanupAction.java
+++ b/src/main/java/net/sf/jabref/gui/actions/CleanupAction.java
@@ -15,6 +15,7 @@
  */
 package net.sf.jabref.gui.actions;
 
+import net.sf.jabref.BibDatabaseContext;
 import net.sf.jabref.Globals;
 import net.sf.jabref.JabRefPreferences;
 import net.sf.jabref.gui.BasePanel;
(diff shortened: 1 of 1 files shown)


src/main/java/net/sf/jabref/gui/actions/CleanupAction.java
```diff
@@ -155,10 +156,10 @@ private int showDialog() {
     /**
      * Runs the cleanup on the entry and records the change.
      */
-    private void doCleanup(CleanupPreset preset, BibEntry entry, NamedCompound ce) {
+    private void doCleanup(CleanupPreset preset, BibEntry entry, BibDatabase bibDatabase, NamedCompound ce) {
         // Run cleaner
         CleanupWorker cleaner = new CleanupWorker(preset,
-                panel.getBibDatabaseContext().getMetaData().getFileDirectory(Globals.FILE_FIELD), null,
+                panel.getBibDatabaseContext().getMetaData().getFileDirectory(Globals.FILE_FIELD), bibDatabase,
```
oscargus: Or should one simply add panel.getDatabase() here? Or pass Globals.journalAbbreviationLoader.getRepository() to doCleanup() as well?

Guess it doesn't really matter...

oscargus:
:+1:

src/main/java/net/sf/jabref/gui/actions/CleanupAction.java
koppor: You are right! The new `BibDatabaseContext` is nicer than that quick hack.

PR #894, opened 2016-02-29
Some use of forEach

oscargus:
Made use of the nice method forEach in a number of places.


simonharrer:
I am against this PR as forEach is in my opinion not helping here. It makes debugging much more complicated (hard to debug and step through), the stacktrace is getting polluted and it does not save much space. 

To me, the forEach is the least useful of the streaming functions. ForEach is also technically something which does not exist in the purely functional world, as a forEach explicitly aims to create side effects. To me, I always want to work with streams without any side effects, I only want to create new elements in the stream based on existing elements (map), filter them or merge them together (reduce).

stefan-kolb:
I second Simon here.

oscargus:
Fair enough.

PR #897, opened 2016-03-01
Fixed some coverity issues 

Siedlerchr:
- [x] Tests green?

(Serializable interface is recommended) and resource leaks in the database exporter
Use of ! equals instead of != for strings in EntryEditor
Replaced some ActionListener code with Lamdbas

### DIFF
--- a/src/main/java/net/sf/jabref/bibtex/comparator/BibtexStringComparator.java
+++ b/src/main/java/net/sf/jabref/bibtex/comparator/BibtexStringComparator.java
@@ -17,9 +17,11 @@ package net.sf.jabref.bibtex.comparator;
 
 import net.sf.jabref.model.entry.BibtexString;
 
+import java.io.Serializable;
 import java.util.Comparator;
 
-public class BibtexStringComparator implements Comparator<BibtexString> {
+
+public class BibtexStringComparator implements Comparator<BibtexString>, Serializable {
 
     private final boolean considerRefs;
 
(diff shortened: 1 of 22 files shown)


simonharrer:
- When you use try-finally and then return the connection, the connection you return in these methods is always closed. So these changes are probably wrong. 
- Why do you add the Serializable interface? This does not make sense to me. Can you please explain your reasons?

oscargus:
Regarding Serializable, it is to get rid of the corresponding Coverity warning. Maybe better to remove the check?

I can confirm that using try-with on the Connection doesn't work. The code must be significantly refactored to get around this.

matthiasgeiger:
> Maybe better to remove the check?

:+1:

Siedlerchr:
Ah shit, I didn't thought about that with the connection.
The serialiable thing is not necessary, but it is recommend in the Api Doc:
http://stackoverflow.com/questions/8642012/why-should-a-comparator-implement-serializable
And here:
http://docs.oracle.com/javase/8/docs/api/java/util/Comparator.html

>  Note: It is generally a good idea for comparators to also implement java.io.Serializable, as they may be used as ordering methods in serializable data structures (like TreeSet, TreeMap). In order for the data structure to serialize successfully, the comparator (if provided) must implement Serializable.

simonharrer:
The serializable thing only makes sense if the comparator is implemented in a separate class, but not just part of a really big class with lots of other fields. In our case, we do not use serialization, hence, we can skip these issues.

Siedlerchr:
Okay. Then it should be marked as suppressed in coverity, this would then clear the overview.
https://ondemand.coverity.com/help/suppressing-issues  (filter by issue type fits best, I think)

For the connection thing I will try to come up with another solution that works.

simonharrer:
Yes, we should suppress them. If you like, you can have a go at adding this yml file as part of this PR.

Siedlerchr:
Reworked as new PR.

PR #899, opened 2016-03-02
Fix bug in "Find broken file links" integrity check: files relative to the bibfile are now found

tobiasdiez:
Suppose you have a linked file test.pdf in the same folder as the bib file. Then "Open linked file" finds the file. However the integrity check complained. This is fixed now.

Solution was to replace FileUtil.expandFilename(p.link, fileDirectory) with FileUtil.expandFilename(context.getMetaData(), p.link), which is a bit smarter.

- [x] Change in CHANGELOG.md described? (no, since the feature was not yet released)
- [x] Tests green?

### DIFF
--- a/src/main/java/net/sf/jabref/logic/integrity/IntegrityCheck.java
+++ b/src/main/java/net/sf/jabref/logic/integrity/IntegrityCheck.java
@@ -1,7 +1,6 @@
 package net.sf.jabref.logic.integrity;
 
 import net.sf.jabref.BibDatabaseContext;
-import net.sf.jabref.Globals;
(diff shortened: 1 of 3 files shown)


matthiasgeiger:
codacy complains about an unused import ```Globals``` in ```IntegrityCheck.java```.

Apart from that: :+1:

simonharrer:
Hm, I took the code from `SynchronizeFileField` line 129 and forward. If this is the correct way, the code needs to be fixed there as well.

tobiasdiez:
The SynchronizeFileField works for some reason although the code is in the end the same... I have no idea why. 

I removed the unused import.

PR #900, opened 2016-03-03
Rework Aux dialog and logic

stefan-kolb:


src/main/java/net/sf/jabref/wizard/auximport/AuxFileParser.java
```diff
@@ -75,196 +71,110 @@ public final int getNotResolvedKeysCount() {
...
-        if (path == null) {
-            path = "";
-        } else {
-            path = path + File.separator;
-        }
-
-        nestedAuxCounter = -1; // count only the nested reads
+    private void parseAuxFile(String filename) {
+        // nested aux files
+        List<String> fileList = Arrays.asList(filename);
```
oscargus: asList(T... a)
    Returns a fixed-size list backed by the specified array.

So it won't work to add file names later on.

oscargus: Looks good! Except that it won't work with nested aux-files...

oscargus:
Looks good! Except that it won't work with nested aux-files...

Maybe it is time to move this to logic.auxparser and gui.auxdialog? (Or use the same naming in both packages.)

src/main/java/net/sf/jabref/wizard/auximport/AuxFileParser.java
```diff
@@ -274,6 +189,33 @@ private void resolveTags() {
...
+    private void insertEntry(BibEntry entry) {
+        BibEntry clonedEntry = (BibEntry) entry.clone();
+        clonedEntry.setId(IdGenerator.next());
+        auxDatabase.insertEntry(clonedEntry);
+    }
+
+    /*
      *  Copy the database's configuration, i.e., preamble and strings.
      */
     private void copyDatabaseConfiguration() {
```
tobiasdiez: I think it would make sense to move the copyDatabaseConfiguration method to the BibDatabase class (since the db class knows best what to copy).

tobiasdiez:
Looks good to me.
One remark: I don't like that the parse method is invoked in the constructor. Maybe it is worthwhile to adopt the structure of the BibParser. There the parser has essentially one public method `parse()` which returns a `ParseResult` containing all the information.

oscargus:
:+1: Anything to wait for or should we merge?

tobiasdiez:
:+1:

PR #904, opened 2016-03-04
Removed some switch warnings and restructured code

oscargus:
Extracted methods from large switch, rewrote as switch and added default labels to switch.

- [x] Tests green?


lenhard:
Personally, I prefer if/elseif constructs over switch, since it is so easy to introduce a fall-through bug. However, in the case of the `Layout` class, the switch increases readability in my point of view. So this is fine, as are the remaining changes.

PR #905, opened 2016-03-04
Made sure all ImportFormat classes never return null, but Collections…

oscargus:
….emptyList() instead

* ImportFormat.importEntries now never return null (still not propagated to calling code)
* All(?) ImportFormat readers are handled with try-with
* Replaced StringBuffer with StringBuilder where possible
* Used interface where applicable

- [ ] Change in CHANGELOG.md described? Not relevant
- [ ] Tests created for changes? Not relevant
- [x] Tests green?


src/main/java/net/sf/jabref/importer/fileformat/EndnoteImporter.java
```diff
@@ -85,26 +86,26 @@ public boolean isRecognizedFormat(InputStream stream) throws IOException {
      */
     @Override
     public List<BibEntry> importEntries(InputStream stream, OutputPrinter status) throws IOException {
-        ArrayList<BibEntry> bibitems = new ArrayList<>();
+        List<BibEntry> bibitems = new ArrayList<>();
         StringBuilder sb = new StringBuilder();
-        BufferedReader in = new BufferedReader(ImportFormatReader.getReaderDefaultEncoding(stream));
+        try (BufferedReader in = new BufferedReader(ImportFormatReader.getReaderDefaultEncoding(stream))) {
```
simonharrer: Does this really work? Because you close the input stream afterwards and I am not sure whether the `stream` is reused for different importers. Can you please investigate upon that?

simonharrer:
Looks good to me, except for the stream closing issue due to possible reuse.

simonharrer:
Did you had time to check the issue with the stream?

oscargus:
No, not yet. Hopefully this weekend...

oscargus:
I imported some databases of different types and it seems like it works. I therefore seems like closing Readers does not close the parent Stream, although it apparently is enough to close the Stream to also close the Reader.

lenhard:
As far as I can see, the only thing that is missing here is a rebase on master, right?

oscargus:
Yes , it should be OK.

PR #910, opened 2016-03-05
Fixes coverity issues, suppress Serialization issues

Siedlerchr:
- [X ] Changes in pull request outlined? (What, why, ...)

Used Lambdas for ActionListener
Fixed resource Leak in DB Exporter Methods

### DIFF
--- a/src/main/java/net/sf/jabref/external/push/PushToLyx.java
+++ b/src/main/java/net/sf/jabref/external/push/PushToLyx.java
@@ -75,6 +75,7 @@ public class PushToLyx extends AbstractPushToApplication implements PushToApplic
 
     @Override
     protected void initSettingsPanel() {
+        super.initSettingsPanel();
         settings = new JPanel();
         settings.add(new JLabel(Localization.lang("Path to LyX pipe") + ":"));
         settings.add(path);
--- a/src/main/java/net/sf/jabref/gui/DragDropPane.java
+++ b/src/main/java/net/sf/jabref/gui/DragDropPane.java
@@ -152,6 +152,7 @@ class DragDropPane extends JTabbedPane {
 
         @Override
         public void paintComponent(Graphics g) {
+            super.paintComponent(g);
             Graphics2D g2 = (Graphics2D) g;
             g2.setComposite(AlphaComposite.getInstance(
                     AlphaComposite.SRC_OVER, 0.9f)); // Set transparency
--- a/src/main/java/net/sf/jabref/gui/groups/GroupAddRemoveDialog.java
+++ b/src/main/java/net/sf/jabref/gui/groups/GroupAddRemoveDialog.java
@@ -1,3 +1,18 @@
+/*  Copyright (C) 20013-2016 JabRef contributors.
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License along
+    with this program; if not, write to the Free Software Foundation, Inc.,
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
 package net.sf.jabref.gui.groups;
 
 import java.awt.BorderLayout;
@@ -17,8 +32,6 @@ import javax.swing.JDialog;
 import javax.swing.JScrollPane;
 import javax.swing.JTree;
 import javax.swing.SwingUtilities;
-import javax.swing.event.TreeSelectionEvent;
-import javax.swing.event.TreeSelectionListener;
 import javax.swing.tree.DefaultTreeModel;
 import javax.swing.tree.TreeNode;
 import javax.swing.tree.TreePath;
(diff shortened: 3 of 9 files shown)


src/main/java/net/sf/jabref/groups/structure/ExplicitGroup.java
```diff
@@ -15,23 +15,23 @@
...
+import net.sf.jabref.groups.UndoableChangeAssignment;
+import net.sf.jabref.logic.l10n.Localization;
+import net.sf.jabref.logic.util.strings.QuotedStringTokenizer;
+import net.sf.jabref.logic.util.strings.StringUtil;
+import net.sf.jabref.model.database.BibDatabase;
+import net.sf.jabref.model.entry.BibEntry;
+
 /**
- * Select explicit bibtex entries. It is also known as static group.
+ * Select explicit bibtex groupEntries. It is also known as static group.
```
tobiasdiez: Why did you rename entries to group entries? It is relatively clear from the context that these entries belong to the explicit group.

src/main/java/net/sf/jabref/groups/structure/ExplicitGroup.java
```diff
@@ -217,7 +217,7 @@ public String getDescription() {
     }
 
     public static String getDescriptionForPreview() {
-        return Localization.lang("This group contains entries based on manual assignment. "
+        return Localization.lang("This group contains groupEntries based on manual assignment. "
```
tobiasdiez: you definitely don't want this change.

src/main/java/net/sf/jabref/gui/GroupAddRemoveDialog.java
```diff
@@ -51,16 +57,16 @@ public GroupAddRemoveDialog(BasePanel panel, boolean add, boolean move) {
     @Override
     public void action() throws Throwable {
         GroupTreeNode groups = panel.getBibDatabaseContext().getMetaData().getGroups();
-        if (groups == null) {
+        if (Objects.isNull(groups)) {
```
tobiasdiez: Objects.isNull exists more or less just to be used as a predicate (e.g. filter(Objects::isNull). The previous version with == null is more readable in general.

src/main/java/net/sf/jabref/groups/structure/ExplicitGroup.java
oscargus: I'd guess it is because some methods use `entries` as variables, see e.g. line 82.

@Siedlerchr I've tried to avoid touching the groups package since there's a big PR in the waiting to be merged after 3.3 is released, #847, just to avoid conflicts there.

src/main/java/net/sf/jabref/gui/util/comparator/IconComparator.java
```diff
@@ -17,13 +17,14 @@
...
 
+import java.io.Serializable;
 import java.util.Comparator;
 import java.util.List;
 
 /**
  * Comparator that handles icon columns.
  */
-public class IconComparator implements Comparator<BibEntry> {
+public class IconComparator implements Comparator<BibEntry>, Serializable {
```
tobiasdiez: I think the Serializable is not necessary (same also below)

src/main/java/net/sf/jabref/gui/dbproperties/SaveActionsPanel.java
```diff
@@ -138,19 +134,22 @@ public void storeSettings(MetaData metaData) {
     }
 
     public boolean hasChanged() {
-        List<FieldFormatterCleanup> newActions = ((SaveActionsListModel) actionsList.getModel()).getAllActions();
+        List<FieldFormatterCleanup> newActions = ((SaveActionsListModel<FieldFormatterCleanup>) actionsList.getModel())
```
oscargus: Do you need to specify `FieldFormatterCleanup` on the right hand side or is it enough to write `<>`? I have no idea...

tobiasdiez:
In general, this looks good to me. But I would welcome if you wait with merging this in until #847 is accepted (because otherwise rebasing this other PR will be a hell)

src/main/java/net/sf/jabref/groups/structure/ExplicitGroup.java
oscargus: That is probably what is causing the failing test as well.

src/main/java/net/sf/jabref/groups/structure/ExplicitGroup.java
Siedlerchr: I renamed it, because nearly every method uses entries and therefore would be hiding/overriding this.

src/main/java/net/sf/jabref/groups/structure/ExplicitGroup.java
Siedlerchr: Oops. I oversaw this from the Rename. Will correct it.

src/main/java/net/sf/jabref/gui/GroupAddRemoveDialog.java
Siedlerchr: Okay. Noted. Will change it back.

src/main/java/net/sf/jabref/gui/util/comparator/IconComparator.java
Siedlerchr: I meant not to change this. Seems like the files were not correctly overwritten when I deleted the old branch from the previous PR.

src/main/java/net/sf/jabref/gui/dbproperties/SaveActionsPanel.java
Siedlerchr: In this case it's necessary, it can't be derived automatically.
Previously all references to the type where non-generic

Siedlerchr:
@tobiasdiez  No problem.  What about including the few changes to the group related classes here in your PR?

src/main/java/net/sf/jabref/groups/structure/ExplicitGroup.java
simonharrer: Javadoc should be human readable, and not contain internal field names. Hence, I vote for entries instead of the internal groupEntries.

stefan-kolb:
We should be able to merge this now.

oscargus:
Yes, but quite a few of the files has been moved since the original PR... I would almost suggest redoing compared to rebasing/merging... Good luck @Siedlerchr! :+1:

Siedlerchr:
Yeah, I will look at this in the next days. I think many of these things already have been taken into account.

koppor:
Bump :innocent:

Siedlerchr:
@koppor  Re-added the changes. Some issue were already fixed.

Issue #455, opened 2015-12-03
reset Preference or tool for clear JabRef registry


g10996694:
I noticed that when I use different versions of JabRef, some times they load different files. 
This means that different versions read perhaps different registry keys, or the jabref preference are saved in different registry locations.
on the same machine, the 3.x version preference is different to 2.x version. 
perhaps I've tested too many versions ?

Is there an option to reset JabRef preference ?
Or, perhaps it would be better a tool which deletes all previous JabRef registries.

stefan-kolb:
You can reset the preferences by starting JabRef with `java -jar JabRef-2.11.jar --prdef all -n`.

lenhard:
We don't really need two issues in different repositories for the very same thing, do we?

https://github.com/koppor/jabref/issues/47 is enough

PR #915, opened 2016-03-06
Add button in preference dialog to reset preferences

tobiasdiez:
Implements #455 and/or https://github.com/koppor/jabref/issues/47.

- [x] Tests created for changes? - only UI change
- [x] Tests green?


simonharrer:
Nice. Looks good. But I would add a note that resetting the perferences requires a restart (see change the language or the remote settings).

tobiasdiez:
Good point. An info message is now shown.

simonharrer:
Looks good to me.

PR #918, opened 2016-03-07
Cleanups

oscargus:
See commit messages

### DIFF
--- a/src/main/java/net/sf/jabref/importer/fetcher/GVKParser.java
+++ b/src/main/java/net/sf/jabref/importer/fetcher/GVKParser.java
@@ -220,15 +220,15 @@ public class GVKParser {
 
             //isbn
             if ("004A".equals(tag)) {
-                final String isbn_10 = getSubfield("0", datafield);
-                final String isbn_13 = getSubfield("A", datafield);
+                final String isbn10 = getSubfield("0", datafield);
+                final String isbn13 = getSubfield("A", datafield);
 
-                if (isbn_10 != null) {
-                    isbn = isbn_10;
+                if (isbn10 != null) {
+                    isbn = isbn10;
                 }
 
-                if (isbn_13 != null) {
-                    isbn = isbn_13;
+                if (isbn13 != null) {
+                    isbn = isbn13;
                 }
 
             }
@@ -244,10 +244,8 @@ public class GVKParser {
                 }
 
                 String st = getSubfield("a", datafield);
-                if (st != null) {
-                    if (st.contains("Diss")) {
-                        entryType = "phdthesis";
-                    }
+                if ((st != null) && st.contains("Diss")) {
+                    entryType = "phdthesis";
                 }
             }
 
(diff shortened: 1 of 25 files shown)


src/main/java/net/sf/jabref/importer/fetcher/GVKParser.java
```diff
@@ -497,8 +485,7 @@ private Element getChild(String name, Element e) {
     }
 
     private String removeSortCharacters(String input) {
```
tobiasdiez: Can you please check that no one calls this code expecting that the input string is changed, i.e simply as `removeSortCharacters(input)` instead of `output = removeSortCharacters(input)`.

src/main/java/net/sf/jabref/logic/auxparser/AuxParserResult.java
```diff
@@ -50,6 +51,14 @@ public int getCrossRefEntriesCount() {
         return crossRefEntriesCount;
     }
 
+    public void countCrossRefEntry() {
+        crossRefEntriesCount++;
+    }
+
+    public void countNestedAuxFile() {
```
tobiasdiez: I don't like the name. Maybe something more expressive as `increaseCountOfNestedAuxFiles()`?

src/main/java/net/sf/jabref/importer/fetcher/GVKParser.java
oscargus: Done!

src/main/java/net/sf/jabref/logic/auxparser/AuxParserResult.java
oscargus: To long to type with one hand (in my bed)... :-) I can sort out tomorrow though.

src/test/java/net/sf/jabref/logic/xmp/XMPUtilTest.java
```diff
@@ -189,7 +189,7 @@ public String t3XMP() {
                 + "</rdf:Seq></bibtex:editor>" + "<bibtex:bibtexkey>Clarkson06</bibtex:bibtexkey>"
                 + "<bibtex:journal>International Journal of High Fidelity</bibtex:journal>"
                 + "<bibtex:booktitle>Catch-22</bibtex:booktitle>" + "<bibtex:pdf>YeKis03 - Towards.pdf</bibtex:pdf>"
-                + "<bibtex:keywords>peanut,butter,jelly</bibtex:keywords>"
+                + "<bibtex:keywords>peanut, butter, jelly</bibtex:keywords>"
```
tobiasdiez: Why did you changed the test input?

oscargus: To make the tests pass... But I might not have analysed that carefully exactly where it should be changed, but changed all keywords to contain ', ' as separator. You are probably correct that here it is better not to change it.

oscargus: To simplify the tests I kept it like that. If not, one would need to have two different "correct" results. As we use `, ` as separator at the moment it seems like an OK thing to do.

oscargus: The problem is that one of the test simply copies the keywords field as is and one parses it resulting in two differently looking fields at the end.

simonharrer:
:+1:

src/test/java/net/sf/jabref/logic/xmp/XMPUtilTest.java
tobiasdiez: Ok, sounds reasonable!

PR #922, opened 2016-03-08
Some more fixes

oscargus:
Small change of behaviour in LabelPatternUtil.

However, I think that the previous code didn't even work... How can e.g.`"univ".equalsIgnoreCase(k.substring(0, 5))` ever evaluate to true? The substring is five characters, right? This is something for static code checkers to find...

### DIFF
--- a/src/main/java/net/sf/jabref/importer/fileformat/BibtexParser.java
+++ b/src/main/java/net/sf/jabref/importer/fileformat/BibtexParser.java
@@ -616,8 +616,7 @@ public class BibtexParser {
                 return token.toString();
             }
 
-            if (Character.isLetterOrDigit((char) character) || (character == ':') || (character == '-') || (character == '_')
-                    || (character == '*') || (character == '+') || (character == '.') || (character == '/') || (character == '\'')) {
+            if (Character.isLetterOrDigit((char) character) || (":-_*+./'".indexOf(character) >= 0)) {
                 token.append((char) character);
             } else {
                 unread(character);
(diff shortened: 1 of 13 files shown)


oscargus:
The background of the regex-solution is the issues pointed out by Codacy, requiring a locale to do "proper" toLowerCase/toUpperCase. This doesn't solve every issue related to that, but at least a few... Shouldn't be that frequently used so performance is an issue.

simonharrer:
Why not just pass Locale.EN to the to*Case methods?

oscargus:
Yes, that is of course a simple solution. :-) However, there's no need (in the general case) to support other languages? Considering that we still have Unicodesupport. What happens to ß?

simonharrer:
yes, but when comparing it to a constant later, it is not that much of an issue, I think.

simonharrer:
Could you please update this and just use toLowerCase with Locale.EN? Then, this can be merged in.

oscargus:
Changed back (but kept the uni/dep/etc detection, which IMHO is better now).

src/main/java/net/sf/jabref/importer/fileformat/BibtexParser.java
```diff
@@ -616,8 +616,7 @@ private String parseTextToken() throws IOException {
                 return token.toString();
             }
 
-            if (Character.isLetterOrDigit((char) character) || (character == ':') || (character == '-') || (character == '_')
-                    || (character == '*') || (character == '+') || (character == '.') || (character == '/') || (character == '\'')) {
+            if (Character.isLetterOrDigit((char) character) || (":-_*+./'".indexOf(character) >= 0)) {
```
tobiasdiez: Maybe replace with Arrays.asList(':', '-', .... ).contains(character). But I'm not sure which one is more readable.

tobiasdiez:
LGTM

Issue #707, opened 2016-01-26
Add tests checking parsing of JabRef's metadata (groups etc.)


lenhard:
We should add tests to `BibtexParserTest` to make sure that we do not break the meta data format.

lenhard:
Seems like this issue has already been fixed 8 months ago. High time for closing.

PR #925, opened 2016-03-08
Add tests for metadata parsing

tobiasdiez:
This solves #707 and prepares the way to refactor the metadata class.

I added a few methods to the MetaData class to make testing more convenient. Many calls to getData() in fact can be replaced by these new methods. I think this is something for a new PR.

One question: to write the tests in a nice way I had to overwrite a few equals methods. In particular for the LowerCaseChanger. Should I also override equals for all other formatters?


simonharrer:
:+1:

oscargus:
When is the equals required (except for simplifying this test)? Seems like one can add them later if required.

The PR looks good!

tobiasdiez:
No as far as I can see equals is never used for the formatters (and also no getKey().equals which could be replaced with an internal equals).

PR #926, opened 2016-03-09
Key bindings are now working in preview panel

### LINKED ISSUE #649: Key bindings not working in preview panel
The key binding for print preview is not working (nor is the in #648 introduced key binding for copy preview working). Maybe it has something to do with the toolbar not being visible in the preview panel. Not sure how to enable it though...

bartsch-dev:
Fixes #649.
Until now the toolbar within the Preview panel held the key bindings which weren't active when the toolbar wasn't created. Now the Preview panel itself carries the key bindings.

- [X] Changes in pull request outlined? (What, why, ...)
- [x] Tests green?


boceckts:
Looks good to me :+1:

oscargus:
Looks good! :+1:

simonharrer:
- Please extract the setting of the key bindings as a method. Otherwise, looks good to me too. 
- For each action, one has to register a string two times, one at the input map and one in the action map, e.g., "print". Can you extract variables so that this is made more clear? A local variable for each of the blocks should suffice, making it a constant moves it away from the code in question, hence, just create a local one.

bartsch-dev:
Extracted the key bindings and the actionMapKeys.

simonharrer:
:+1:

PR #928, opened 2016-03-10
When changing the type of an entry, the editor does not update

stefan-kolb:
Regression bug introduced by https://github.com/JabRef/jabref/commit/51685e85a39cb6dd998e6f4fb994ff2dd15b8fb8.
JabRef v3.2 is fine, so no changelog entry.

EntryEditor must keep a field for the currentlyDisplayed BibEntry type in order to update the editor correctly onChange to the entry type.

### DIFF
--- a/src/main/java/net/sf/jabref/gui/BasePanel.java
+++ b/src/main/java/net/sf/jabref/gui/BasePanel.java
@@ -1859,7 +1859,7 @@ public class BasePanel extends JPanel implements ClipboardOwner, FileUpdateListe
 
     public void updateEntryEditorIfShowing() {
         if (mode == BasePanel.SHOWING_EDITOR) {
-            if (currentEditor.getType().equals(currentEditor.getEntry().getType())) {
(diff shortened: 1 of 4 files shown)


simonharrer:
:+1:

src/main/java/net/sf/jabref/gui/BasePanel.java
```diff
@@ -2025,40 +2025,36 @@ public void stringsClosing() {
...
+    public void changeTypeOfSelectedEntries(String newType) {
         List<BibEntry> bes = mainTable.getSelectedEntries();
-        changeType(bes, type);
+        changeTypeOfSelectedEntries(bes, newType);
     }
 
-    private void changeType(List<BibEntry> bes, String type) {
-
-        if ((bes == null) || (bes.isEmpty())) {
+    private void changeTypeOfSelectedEntries(List<BibEntry> entries, String newType) {
```
tobiasdiez: maybe just name it changeTypeOfEntries or changeType

tobiasdiez:
:+1: 
and sorry for introducing this bug in the first place.

PR #929, opened 2016-03-10
"Find unlinked files" freezes when using for PDF

### LINKED ISSUE #410: "Find unlinked files" freezes when using for PDF
Using "Find unlinked files" does not work. 
Scanning the directory works fine, but when I press "Apply" it displays "Importing into Database..." without ever making progress.

The problem occurs only for PDFs. It is independent of filename and directory. It is also independent of file content, i.e. it even occurs for empty PDFs.

I am using Windows 7 and the problem occurs for both JabRef 2.11.1 and JabRef 3.0

stefan-kolb:
1. Fixed #410 Reason was an IllegalArgumentException thrown by DOI class when pasting the contents of the whole first PDF page to the constructor.
2. The DOI class can only handle real DOIs and does not extract Strings that might look like a DOI from the text, therefore the whole DOI scraper was removed in https://github.com/JabRef/jabref/commit/acdcd829c304e03b9bfd92d6eba383b5a1d52108
3. The whole PDFImporter class is screwed and a monkeypatch. It should be reimplemented from scratch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,7 @@ to [sourceforge feature requests](https://sourceforge.net/p/jabref/features/) by
 - Fixed [#815](https://github.com/JabRef/jabref/issues/815): Curly Braces no longer ignored in OpenOffice/LibreOffice citation
 - Fixed [#855](https://github.com/JabRef/jabref/issues/856): Fixed OpenOffice Manual connect - Clicking on browse does now work correctly
 - Fixed [#649](https://github.com/JabRef/jabref/issues/649): Key bindings are now working in the preview panel
+- Fixed [#410](https://github.com/JabRef/jabref/issues/410): Find unlinked files no longer freezes when extracting entry from PDF content
 
 ### Removed
 - Fixed [#627](https://github.com/JabRef/jabref/issues/627): The pdf field is removed from the export formats, use the file field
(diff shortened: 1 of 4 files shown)


koppor:
1./2.: The usage of DOIs was requested by a user: If the PDF contains a DOI string, the data of the DOI should be used and not the guessed content. - With old JabRef code, that code worked. Do you have a proposal to bring that functionality back?

3.: The parsing results are better for Springer and IEEE papers than with other existing open source scrapers and Mr. DLib. That fact has to be regarded, too.

stefan-kolb:
The DOI finding should work again now, but the DOI class has quite some redundant code now, as I need to exclude spaces from the findInText regex. I don't know if there are any DOIs out there that include spaces but theoretically it seems to be allowed...

stefan-kolb:
For me `src/test/resources/encrypted.pdf`is not raising any exception. Also, `importEntries` is not throwing any `IOException`. Which is propagated to other methods later on. The functionality is really flawed....

oscargus:
Spaces should be encoded as %20 in a DOI, so I do not think it will be a problem, see https://www.doi.org/doi_handbook/2_Numbering.html#2.5.2.4

I seem to recall that we changed the encoding of DOIs at some stage, maybe this should be revisited?

stefan-kolb:
So you mean we can replace the EXACT_DOI pattern with the DOI_PATTERN?!

PR #930, opened 2016-03-10
Fix 845

simonharrer:
- better German translation of the menu items groups -> highlight groups ...
- add radio button functionality to the menu items groups -> highlight groups ... which is mutually exclusive plus add a disable menu entry which was not explicit before
- enable showing what is currently toggled in the menu as well (and not only in the toolbar)


oscargus:
Looks good!

simonharrer:
Change in the database? Why should there be a change in the database?

What are you not seeing in the toolbar? All toolbar stuff works on my machine.

koppor:
Yeah, only the group window should have a change.

Could we implement "highlight" as change in the background color (maybe the JabRef colour) and not via underline? :innocent: 

Where in the toolbar should I see something?

grabbed_20160311-081714

simonharrer:
I do not want to mess with the groups code at the moment, as this is done by @tobiasdiez currently - avoding any merge conflicts. 

The toolbar has not been changed, but the menu itself now shows the state of the toogle-actions.

tobiasdiez:
Please rebase, then this can be merged in my opinion. 
@koppor can you please open a new issue for your suggestion regarding highlighting.

koppor:
@simonharrer Could you briefly state what you meant by "and not only in the toolbar" in your PR description? What should I see where? :innocent:

simonharrer:
I had to improve this fix: previously, two boolean preferences were used to describe the three states DISABLED, ANY and ALL. The combination (true, true) of the booleans was not allowed. I replaced this by using a single string preference which is either "all", "any" or anything else which resorts to DISABLED. I did not implement any migrations as I would take it into account if the behaviour changes upon upgrade as the user can easily change the setting manually. Should not be a hassle.

tobiasdiez:
Would it make sense to create an enum for these states?

simonharrer:
I tried to create an enum, but ended up creating a separate preferences class with methods for each setting.

koppor:
I wonder whether MDSD would have been of any help here :innocent:

simonharrer:
???

Issue #462, opened 2015-12-04
Action to open console in folder which contains .bib file


simonharrer:
I always have my .bib files under version control with git. Because of this, I often have to open a console in the folder with the .bib file. If I could open a console right from JabRef, this would be nice, helping me to commit my changes faster.

koppor:
Code hint: "Open folder".

Featurewish: configurable console application such as [ConEmu](http://conemu.github.io/).

mairdl:
The JabRef StuPro feature team will take care of this issue.

obraliar:
The primary branch for this issue is located at https://github.com/obraliar/jabref/tree/OpenConsoleFeature

koppor:
Thank you for lettings us know. Please open a pull request if you want us to look into code. :innocent: This also allows @JabRef/stupro-qa to look into and to help avoiding git pitfalls such as starting at the wrong commit.

obraliar:
Now the feature is working on all supported operating systems :) You can start it with a mouse-click on the apropriate button (see image) or by using the shortcut CTRL+SHIFT+J.
![jabref_open_console_feature](https://cloud.githubusercontent.com/assets/15609468/13709271/4e9e232c-e7b3-11e5-9e6b-da3dd189102b.jpg)

Consider that a new terminal can be opened only if your database file is opened and saved.

obraliar:
Can be closed.

matthiasgeiger:
:+1: Implemented in #931

PR #931, opened 2016-03-10
Open console feature

obraliar:
- [ ] Tests green?
- [x] Commits squashed?


obraliar:
Coverage test fails (-0.01%) due to new written and untested code.

obraliar:
Issue: https://github.com/JabRef/jabref/issues/462

boceckts:
You should also include an else case for unsupported operating systems giving some sort of information message

src/main/java/net/sf/jabref/gui/BasePanel.java
```diff
@@ -874,6 +877,39 @@ public void update() {
...
+                    } else if (def_term_emul.contains("xfce4")) {
+                        runtime.exec("xfce4-terminal --working-directory=" + abs_dir);
+                    } else if (def_term_emul.contains("konsole")) {
+                        runtime.exec("konsole --workdir=" + abs_dir);
+                    } else {
+                        runtime.exec(def_term_emul, null, new File(abs_dir));
+                    }
+                }
+            } else if (OS.WINDOWS) {
+                Runtime.getRuntime().exec("cmd.exe /c start", null, new File(abs_dir));
```
boceckts: Why don't you use the runtime variable you created at the start of the methode? Same goes for os_x case

boceckts:
Please squash commits and try to limit your first line of the commit message to 50 characters or less

obraliar:
All done. Thank you for your advices :+1: .

src/main/java/net/sf/jabref/gui/JabRefFrame.java
```diff
@@ -243,6 +243,12 @@ public void actionPerformed(ActionEvent e) {
             Localization.lang("Paste"), Globals.getKeyPrefs().getKey(KeyBinding.PASTE), IconTheme.JabRefIcon.PASTE.getIcon());
     private final AbstractAction cut = new EditAction(Actions.CUT, Localization.menuTitle("Cut"),
             Localization.lang("Cut"), Globals.getKeyPrefs().getKey(KeyBinding.CUT), IconTheme.JabRefIcon.CUT.getIcon());
+    // ##*#
+    private final AbstractAction openConsole = new GeneralAction(Actions.OPEN_CONSOLE,
+            Localization.menuTitle("Open terminal here"),
+ Localization.lang("Open terminal here"),
+            Globals.getKeyPrefs().getKey(KeyBinding.OPEN_CONSOLE),
+ IconTheme.JabRefIcon.CONSOLE.getIcon());
```
boceckts: Could you pls fix the formatting and remove the comment?

boceckts:
Very good :+1:

src/main/java/net/sf/jabref/exporter/SaveDatabaseAction.java
```diff
@@ -389,7 +390,7 @@ public void saveAs() throws Throwable {
             }
             frame.getFileHistory().newFile(panel.getBibDatabaseContext().getDatabaseFile().getPath());
         }
-
+        frame.updateEnabledState();
```
simonharrer: Why was this line moved? Is not clear for me.

src/main/java/net/sf/jabref/gui/desktop/JabRefDesktop.java
```diff
@@ -432,4 +434,38 @@ public static void openBrowser(String url) throws IOException {
         ExternalFileType fileType = ExternalFileTypes.getInstance().getExternalFileTypeByExt("html");
         openExternalFilePlatformIndependent(fileType, url);
     }
+
+    public static void openConsole(JabRefFrame frame) throws IOException {
```
simonharrer: Why should the JabRefDesktop class has a reference to the whole frame? Would a File reference not suffice of the folder in which the console should be openend?

src/main/java/net/sf/jabref/gui/desktop/JabRefDesktop.java
```diff
@@ -432,4 +434,38 @@ public static void openBrowser(String url) throws IOException {
...
+        String path = bp.getBibDatabaseContext().getDatabaseFile().getAbsolutePath();
+        String abs_dir = path.substring(0, path.lastIndexOf(File.separator) + 1);
+        if (OS.LINUX) {
+            Process p = runtime.exec("readlink /etc/alternatives/x-terminal-emulator");
+            BufferedReader reader = new BufferedReader(new InputStreamReader(p.getInputStream()));
+
+            String def_term_emul = reader.readLine();
+
+            if (def_term_emul != null) {
+                def_term_emul = def_term_emul.substring(def_term_emul.lastIndexOf(File.separator) + 1);
```
simonharrer: no snake_case for variable names

src/main/java/net/sf/jabref/gui/desktop/JabRefDesktop.java
```diff
@@ -432,4 +434,38 @@ public static void openBrowser(String url) throws IOException {
         ExternalFileType fileType = ExternalFileTypes.getInstance().getExternalFileTypeByExt("html");
         openExternalFilePlatformIndependent(fileType, url);
     }
+
+    public static void openConsole(JabRefFrame frame) throws IOException {
+        Runtime runtime = Runtime.getRuntime();
+        int selectedIndex = frame.tabbedPane.getSelectedIndex();
+        BasePanel bp = (BasePanel) frame.tabbedPane.getComponentAt(selectedIndex);
+        String path = bp.getBibDatabaseContext().getDatabaseFile().getAbsolutePath();
+        String abs_dir = path.substring(0, path.lastIndexOf(File.separator) + 1);
```
simonharrer: please no abbreviations, use absolutePath instead

src/main/java/net/sf/jabref/gui/IconTheme.java
```diff
@@ -154,9 +154,8 @@
         FILE_MULTIPLE("\uf2ed"), /*css: file-multiple */
         KEY_BINDINGS("\uf3c6"), /*css: keyboard */
         FIND_DUPLICATES("\uf23e"), /*css: code-equal */
-
         OPEN_IN_NEW_WINDOW("\uf47b"), /*css: open-in-new */
-
+        CONSOLE("\uf25e"),
```
tobiasdiez: Could you please add the css code here (as for the other icons)

src/main/java/net/sf/jabref/gui/JabRefFrame.java
```diff
@@ -1573,6 +1584,17 @@ public void updateEnabledState() {
             getBackAction().setEnabled(false);
             getForwardAction().setEnabled(false);
         }
+
+        boolean saved = false;
+
+        if (tabCount > 0) {
```
tobiasdiez: Can you extract the next two lines to a new method `getSelectedPanel` any maybe replace other occurrences of the same code with this method.

src/main/java/net/sf/jabref/gui/JabRefFrame.java
```diff
@@ -1573,6 +1584,17 @@ public void updateEnabledState() {
...
+            int selectedIndex = tabbedPane.getSelectedIndex();
+            BasePanel bp = (BasePanel) tabbedPane.getComponentAt(selectedIndex);
+            saved = bp.getBibDatabaseContext().getDatabaseFile() != null;
```
tobiasdiez: No abbreviation. activePanel

src/main/java/net/sf/jabref/gui/desktop/JabRefDesktop.java
tobiasdiez: Or maybe pass the database context. This would correspond more to the semantic `open console for this database`

src/main/java/net/sf/jabref/gui/desktop/JabRefDesktop.java
tobiasdiez: and process

src/main/java/net/sf/jabref/gui/IconTheme.java
obraliar: Done.

src/main/java/net/sf/jabref/gui/JabRefFrame.java
obraliar: Done. New method extracted. Unfortunately there were other occurrences where it could be used.

src/main/java/net/sf/jabref/gui/JabRefFrame.java
obraliar: Done.

src/main/java/net/sf/jabref/gui/desktop/JabRefDesktop.java
obraliar: Done.

src/main/java/net/sf/jabref/gui/BasePanel.java
```diff
@@ -880,6 +880,11 @@ public void update() {
             });
         });
 
+        actions.put(Actions.OPEN_CONSOLE, (BaseAction) () -> {
+            JabRefDesktop.openConsole(((BasePanel) frame.tabbedPane.getComponentAt(frame.tabbedPane.getSelectedIndex()))
```
tobiasdiez: use getSelectedPanel ?

src/main/java/net/sf/jabref/gui/JabRefFrame.java
tobiasdiez: I just discovered that there is already a getCurrentBasePanel which does the same as the newly created `getSelectedPanel`.

src/main/java/net/sf/jabref/gui/BasePanel.java
obraliar: Done. Yes, I only looked for appearances in JabRefFrame.java, I apologize for that.

src/main/java/net/sf/jabref/gui/JabRefFrame.java
obraliar: Done replacing.

tobiasdiez:
Now just merge all commits into one and then this can be merged in my opinion. Good work!

obraliar:
What about the coverage? I just rebased the branch.. I think it's not up on git.

tobiasdiez:
Good point. You should definitely add tests. Some hints:
- writing tests right now is not practical, since invoking the method results in opening a console -> hard to test
- so the only thing we can test is that the right command is invoked -> need to mock `Runtime` -> need to move the code to a new class so that we can inject the runtime as a dependency in the constructor. (by the way, the new class should live in `logic`)
-  now you want to achieve 100% coverage, so you need to change the OS in your tests. Hard to do if the corresponding fields are static final fields. So the members of the `OS` class should be converted to be non-static instance members in order to just mock them. 

Yes looks like a lot of work just to test 10 lines...but these changes also result in structural improvements like less static methods, clear dependencies etc.

Feel free to ask you if you need more help.

obraliar:
I'm going to create new ticket about this test for the second sprint, cause the first one with this issue is over.

src/main/java/net/sf/jabref/gui/desktop/JabRefDesktop.java
```diff
@@ -432,4 +434,37 @@ public static void openBrowser(String url) throws IOException {
         ExternalFileType fileType = ExternalFileTypes.getInstance().getExternalFileTypeByExt("html");
         openExternalFilePlatformIndependent(fileType, url);
     }
+
+    public static void openConsole(File file) throws IOException {
```
simonharrer: please add if(file == null) { return;} so that no exception occurs if this method is called wrongly.

obraliar: Done.

simonharrer:
I think the effort to test these 10 lines of code is not worth the effort in that case. I would use the time for more important tasks. 

Otherwise, :+1: from me to merge this in. 

PS: Be aware that #961 may require you to use another icon code as it has changed since then.

koppor:
Please rebase on master (resolve the conflicts) [and squash everything in one commit](https://github.com/JabRef/jabref/wiki/Tools#rebase-everything-as-one-commit-on-master for a short howto). Please doublecheck the icon because of #961.

obraliar:
Do not merge, the icon code has changed. Solving...

tobiasdiez:
Even if you decide not to test the code, then I still prefer a move to a new class with non-static method in the `logic` package. @simonharrer what do you think?

obraliar:
@koppor Done resolving conflicts and updating icon code.

koppor:
@tobiasdiez I don't get your wish: `openConsole` is a `logic`?

I would leave as is and do a complete refactoring of JabRefDesktop if needed.

tobiasdiez:
Yes for me `openConsole` is logic or, if you want, creating the `open console command` is logic and maybe invoking it (i.e. runtime.exec) is kind of gui. I think if it is added to JabRefDesktop, then it will stick there forever. So better place it right the first time. But I have no strong opinion about this...

obraliar:
So ready to merge?
@koppor Or should we port `openConsole` to `logic` package? In my opinion there is no suitable class yet to get it in there. We would have to create a new one, probably in `logic/util`.

PR #933, opened 2016-03-11
Test for encrypted PDF does not throw an Exception

stefan-kolb:
I don't know if this sufficies the requirements, but for me it does not throw the strange exception.


stefan-kolb:
@koppor WDYT?

koppor:
Ok for me. Refs #935.

PR #939, opened 2016-03-11
Modify canonical representation to normalize authors

koppor:
This is a PR to start discussions on our canonical representation.

When are two entries are equal. Is "Firstname Lastname" equal to "Lastname, Firstname"? Is "Cool and the Gang" equal to "Cool And The Gang"? How strict should we be?

At first, I thought, we have to be very strict. Then, I saw that our XMP utilities normalize the content of the author field for writing into the PDF.

Maybe we need a "normalized format" and a "canonical representation". The canonical representation is kept as is and the "normalized format" normalizes as much as possible?

This PR currently just normalizes the entry in the canonical representation, which feels wrong, doesn't it?

### DIFF
--- a/src/main/java/net/sf/jabref/model/entry/CanonicalBibtexEntry.java
+++ b/src/main/java/net/sf/jabref/model/entry/CanonicalBibtexEntry.java
@@ -15,6 +15,8 @@ public class CanonicalBibtexEntry {
      * This returns a canonical BibTeX serialization. Special characters such as "{" or "&" are NOT escaped, but written
      * as is
      *
+     * authors and editors are normalized
+     *
      * Serializes all fields, even the JabRef internal ones. Does NOT serialize "KEY_FIELD" as field, but as key
      */
     public static String getCanonicalRepresentation(BibEntry e) {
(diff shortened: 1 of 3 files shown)


tobiasdiez:
I wouldn't change the canonical serialization. It is canonical because we don't normalize anything. 

In my opinion the tests for the XMP class should be improved. Proposed solution:
```java
BibEntry entry = new Entry();
entry.setField("author", "K. Crowston");
entry.setField("field", "data");  // or if you prefer use the parser, but I think the code will be shorter and better maintainable in this way 

XMPUtil.writeXMP(pdfFile, entry, null);

List<BibEntry> result = XMPUtil.readXMP(pdfFile.getAbsoluteFile());
BibEntry expected = entry.clone(); // or new BibEntry(entry), whatever the method is called
expected.setField("author", "Crowston, K."); // we expect normalized author
BibEntryAssert.equals(Arrays.listOf(expected), result);
```

simonharrer:
I think so, too. One could add a normalized format, but at the moment I do not see any help in there right now. Other things are much much much more pressing. :)

PR #942, opened 2016-03-11
Add support for write-protected PDFs

### LINKED ISSUE #935: Add support for parsing of encrypted PDFs
A refactoring revealed that there is a TODO at the PDFContentImporter. It seems that [Bouncy Castle](https://www.bouncycastle.org/documentation.html) was not included during development and a quick hack was not to support encrypted PDFs at all. Some papers, however, are encrypted should be supported.

koppor:
This PR shows how support for write-protected PDFs can be added to JabRef.

If we do not want to be a "cryptographic" software, I can just remove the dependency on `org.bouncycastle:bcprov-jdk15on:1.54`. Then, JabRef will compile, but throw an exception if write-protected PDFs are opened. I can adjust the test cases accordingly.

## Crypto Discussion

This fixes #935.

When importing PDFs, which carry a password for editing, both the XMP Metadata parser and the PDFContentImporter can handle this.

Now, the [Legion of the Bouncy Castle Java cryptography APIs](http://bouncycastle.org/java.html) are part of JabRef, which makes JabRef subject to export restrictions. [Bouncy Castle is approved classified under ECCN code 5D002 and approved for export under License Exception TSU](http://www.bouncycastle.org/wiki/display/JA1/Frequently+Asked+Questions#FrequentlyAskedQuestions-11.WhatisBouncyCastle%27sexportclassificationintheUnitedStatesofAmerica?). According to https://www.bis.doc.gov/index.php/policy-guidance/encryption/classification, JabRef can now only be exported "to any end-user located or headquartered in a country listed in Supplement No. 3 to Part 740", which is [AustriaAustraliaBelgiumBulgariaCanadaCyprusCzech RepublicEstoniaDenmarkFinlandFranceGermanyGreeceHungaryIcelandIrelandItalyJapanLatviaLithuaniaLuxembourgMaltaNetherlandsNew ZealandNorwayPolandPortugalRomaniaSlovakiaSloveniaSpainSwedenSwitzerlandTurkeyUnited Kingdom](https://www.law.cornell.edu/cfr/text/15/part-740/appendix-SupplementNo3).

The current code, however, also allows BouncyCastle to be removed and JabRef still compiles. It internally throw a NoClassDefFound Exception, which is handled properly.

I am unsure, how to proceed. [Debian also distributes bouncy-castle](https://packages.debian.org/de/sid/libbcprov-java).


Siedlerchr:
From what I understand in this discussion, that the US exporting laws would apply if you distribute the software from an US Server:
http://security.stackexchange.com/questions/7015/us-export-control-on-open-source-libraries

The best solution would be to load it dynamically and not distribute it with JabRef.
Then users can download the library, put them in the folder and use it.

stefan-kolb:
To be honest I don't think it is worth to make such a hassle just because of a partially working PDF content importer functionality. I mean how often do we really need this? Isn't it acceptable if we need to get the BibTex entry by a DOI or set it manually if the PDF is encrypted? It really just is a problem if we do not already have the BibTEx entry (which is probably often the case) or don't supply a DOI etc. Think we could discuss this in a call if I'm not the only one concernced about such a change.

koppor:
It's also about XMPUtil, which is really used in the wild. I agree that write-protected PDFs are not the majority of the PDFs in the wild.

With this patch, we could distribute JabRef without BouncyCaslte and users requiring that feature can compile a crypto-aware JabRef for themselves. Linux distributions can decide for themselves whether they want to enable that feature in their binary distributions.

src/test/java/net/sf/jabref/logic/xmp/XMPUtilTest.java
```diff
@@ -860,6 +861,28 @@ public void testWriteMultiple() throws IOException, TransformerException {
...
+
+    /**
+     * Tests whether a edit-protected PDF can be read
+     */
+    @Test
+    public void testReadProtectedPDFHasCorrectMetaData() throws Exception {
+        try (InputStream is = XMPUtilTest.class.getResourceAsStream("/pdfs/write-protected.pdf")) {
+            List<BibEntry> bibentries = XMPUtil.readXMP(is);
+            Assert.assertEquals(1, bibentries.size());
+            BibtexEntryAssert.assertEquals(this.getClass(), "write-protected.bib", bibentries.get(0));
```
tobiasdiez: No need to define an external bib file here...

koppor: Fixed :innocent:. - I've been in a mode that I read the resulting bib files and not interpret code. The code is ca. 6 lines longer, but we have one file less. Hope, that is OK now. https://github.com/JabRef/jabref/pull/942/commits/a39f11f8c4e66eef2921e16dbe01d325662228d4#diff-4538b7a42e8c15776dcbb34bca7c25d3R882

src/test/java/net/sf/jabref/logic/xmp/XMPUtilTest.java
```diff
@@ -860,6 +861,34 @@ public void testWriteMultiple() throws IOException, TransformerException {
...
+     */
+    @Test
+    public void testReadProtectedPDFHasCorrectMetaData() throws Exception {
+        try (InputStream is = XMPUtilTest.class.getResourceAsStream("/pdfs/write-protected.pdf")) {
+            List<BibEntry> readEntries = XMPUtil.readXMP(is);
+
+            BibEntry entry = new BibEntry();
+            entry.setType("misc");
+            entry.setField("author", "Firstname Lastname");
+            List<BibEntry> expected = new ArrayList<>(1);
```
tobiasdiez: Use Collections.singletonList

tobiasdiez:
LGTM :+1:

koppor:
DevCall result: We want to enable reading of write-protected PDFs as they are not that seldom. In the future, JabRef [will have better support for PDF text and PDF comments](https://github.com/JabRef/jabref/wiki/FeatureRequests-Sorted#make-a-better-use-of-my-linked-files-please). Thus, this PR enables treating of more PDFs.

src/test/java/net/sf/jabref/logic/xmp/XMPUtilTest.java
koppor: Fixed in 82b5b7c

PR #945, opened 2016-03-12
Tweak PDFContentImporter

koppor:
This is a minor update on PdfContentImporter, which fixes minor issues and enables a broader test.


src/test/java/net/sf/jabref/importer/fileformat/PdfContentImporterTest.java
```diff
@@ -17,4 +31,19 @@ public void doesNotHandleEncryptedPdfs() throws IOException {
...
     }
-}
\ No newline at end of file
+
+    @Test
+    public void correctContent() throws IOException {
+        PdfContentImporter importer = new PdfContentImporter();
+        // The test folder contains pairs of PDFs and BibTeX files. We check each pair.
+        // Currently only one pair available
+        List<String> prefixes = Arrays.asList("LNCS-minimal");
```
tobiasdiez: I wouldn't write this as a for loop and use parametrized tests instead.

src/test/java/net/sf/jabref/importer/fileformat/PdfContentImporterTest.java
```diff
@@ -17,4 +31,19 @@ public void doesNotHandleEncryptedPdfs() throws IOException {
...
+        for (String prefix : prefixes) {
+            try (InputStream is = PdfContentImporter.class.getResourceAsStream(prefix + ".pdf")) {
+                List<BibEntry> result = importer.importEntries(is, null);
+                Assert.assertEquals(1, result.size());
+                BibtexEntryAssert.assertEquals(PdfContentImporterTest.class, prefix + ".bib", result.get(0));
```
tobiasdiez: Do you really think it is more readable to have the entry as a bib-file? 
I would just construct it via `enty = new BibEntry(); entry.setFields`.

src/test/java/net/sf/jabref/importer/fileformat/PdfContentImporterTest.java
```diff
@@ -1,15 +1,29 @@
...
 import java.util.Collections;
 import java.util.List;
 
 import static org.junit.Assert.*;
 
 public class PdfContentImporterTest {
+
+    @BeforeClass
+    public static void setUp() {
+        Globals.prefs = JabRefPreferences.getInstance();
```
tobiasdiez: Is this necessary?

src/test/java/net/sf/jabref/importer/fileformat/PdfContentImporterTest.java
koppor: Yes, I think so. :innocent:. Think, it is easier for writing tests and debugging. The should-be-data is more close to the thing the user sees, but also close to the data generated by the importer.

src/test/java/net/sf/jabref/importer/fileformat/PdfContentImporterTest.java
koppor: Like

```
public void testImport(sourcePdfName, shouldBeBibFileName) {
...
}

public void testLNCS() {
    testImport("LNCS-minimal.pdf", "LNCS-minimal.bib")
}
```

src/test/java/net/sf/jabref/importer/fileformat/PdfContentImporterTest.java
koppor: Yes, otherwise the resolution of fields in the database fails.

src/test/java/net/sf/jabref/importer/fileformat/PdfContentImporterTest.java
tobiasdiez: Like https://github.com/junit-team/junit/wiki/Parameterized-tests and see the FormatterTest for an example.

koppor: Done.

src/main/java/net/sf/jabref/importer/fileformat/PdfContentImporter.java
```diff
@@ -3,13 +3,14 @@
...
 import net.sf.jabref.importer.fetcher.DOItoBibTeXFetcher;
-import net.sf.jabref.logic.l10n.Localization;
 import net.sf.jabref.logic.util.DOI;
 import net.sf.jabref.model.entry.BibEntry;
 import net.sf.jabref.model.entry.BibtexEntryTypes;
 import net.sf.jabref.model.entry.EntryType;
+
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.apache.logging.log4j.util.Strings;
```
simonharrer: I would not use a class which is part of the util package of log4j.

src/test/java/net/sf/jabref/importer/fileformat/PdfContentImporterTestFiles.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @Parameters(name = "{index}: {0}")
+    public static Collection<Object[]> fileNames() {
+        // The test folder contains pairs of PDFs and BibTeX files. We check each pair.
+        // This method returns the basenames of the available pairs
+
+        // currently, only one test case "LNCS-minimal" is existing
+        // code review demanded that @Paramters should be used
+        // Arrays.asList only works for multiple parameters, therefore, there are two parameters written
+        return Arrays.asList(new Object[] {"LNCS-minimal"}, new Object[] {"LNCS-minimal"});
```
simonharrer: You can use something like here https://github.com/uniba-dsg/BPELlint/blob/master/src/test/java/bpellint/SAViolationTestData.java which allows for a much nicer structure of the input paramters.

oscargus:
:+1:

simonharrer:
:+1: rebase and merge

Issue #796, opened 2016-02-11
Undo of cut and delete entries doesn't work


oscargus:
Somehow, only one entry is restored when undoing cut or delete of more than one entry. I thought I had messed it up in #793, but it behaved like this before merging it in...

tobiasdiez:
I can't replicate this bug. Can you please describe in more detail when this bug occurs (e.g. cut via context menu or normal menu, keybord shortcut,...).

bartsch-dev:
When you delete 2 Entries on the same time and undo the action only one entry is recovered and an error is throw.
I'm working on a fix.

oscargus:
As I recall it was indeed related to multiple entries.

oscargus:
I tried it now and it seems like it is working properly. I can at least not trigger anything. I think it was related to not returning a copy of the list of selected entries, so when starting to manipulate the entries, the list changed.

Unless someone else has a problem, I believe it can be closed.

tobiasdiez:
I close it now. If somebody experience the problem, then this issue can be reopened.

bartsch-dev:
I still experience that bug.
tested with the current master at 7aa371c931ea51e7c092632e3e3f817c4861f66b.

![issue-796](https://cloud.githubusercontent.com/assets/15333371/13703149/05664cf8-e794-11e5-8b52-12922b6c33b4.gif)

bartsch-dev:
This can now be closed!

matthiasgeiger:
Thx!

PR #947, opened 2016-03-12
[WIP] Fix for undoing deletion/cutting multiple entries

bartsch-dev:
Workaround for #796.
When an undo with multiple entries is undone, each entry is undone one at a time. The Problem herein lies, that when the first entry is undone, the insertion in the database triggers the creation and insertion of an NamedCompund into the UndoManager which kills the main Undo (the one with the multiple entries), which result in just one entry being undone.

With this Workaround each entry has its own NamedCompund which solves the Problem but the user has to call the UndoAction several times (Ctrl-Z for each entry).

- [x] Tests green?

### DIFF
--- a/src/main/java/net/sf/jabref/specialfields/SpecialFieldDatabaseChangeListener.java
+++ b/src/main/java/net/sf/jabref/specialfields/SpecialFieldDatabaseChangeListener.java
@@ -1,6 +1,20 @@
+/*  Copyright (C) 2003-2016 JabRef contributors.
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License along
+    with this program; if not, write to the Free Software Foundation, Inc.,
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
 package net.sf.jabref.specialfields;
 
-import net.sf.jabref.JabRefGUI;
 import net.sf.jabref.event.EntryAddedEvent;
 import net.sf.jabref.gui.undo.NamedCompound;
 import net.sf.jabref.logic.l10n.Localization;
@@ -21,14 +35,15 @@ public class SpecialFieldDatabaseChangeListener {
     }
 
     @Subscribe
-    public void listen(EntryAddedEvent EntryAddedEvent) {
+    public void listen(EntryAddedEvent event) {
         if (SpecialFieldsUtils.keywordSyncEnabled()) {
-            final BibEntry entry = EntryAddedEvent.getBibEntry();
+            final BibEntry entry = event.getBibEntry();
             // NamedCompount code similar to SpecialFieldUpdateListener
             NamedCompound nc = new NamedCompound(Localization.lang("Synchronized special fields based on keywords"));
             SpecialFieldsUtils.syncSpecialFieldsFromKeywords(entry, nc);
-            nc.end();
-            JabRefGUI.getMainFrame().getCurrentBasePanel().undoManager.addEdit(nc);
+            // Don't insert the compound into the undoManager,
+            // it would be added before the component which undoes the insertion of the entry and creates heavy problems
+            // (which prohibits the undo the deleting multiple entries)
         }
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@ to [sourceforge feature requests](https://sourceforge.net/p/jabref/features/) by
 - Fixed [#1245](https://github.com/JabRef/jabref/issues/1245): Empty jstyle properties can now be specified as ""
 - Fixed [#1364](https://github.com/JabRef/jabref/issues/1364): Windows: install to LOCALAPPDATA directory for non-admin users
 - Fixed [#1365](https://github.com/JabRef/jabref/issues/1365): Default label pattern back to "[auth][year]"
+- Fixed [#796](https://github.com/JabRef/jabref/issues/796): Undoing more than one entry at the same time is now working
 
 ### Removed
 - Removed possibility to export entries/databases to an `.sql` file, as the logic cannot easily use the correct escape logic
--- a/src/integrationTest/java/net/sf/jabref/gui/UndoTest.java
+++ b/src/integrationTest/java/net/sf/jabref/gui/UndoTest.java
@@ -1,0 +1,72 @@
+package net.sf.jabref.gui;
+
+import java.io.File;
+
+import net.sf.jabref.JabRefMain;
+
+import org.assertj.swing.finder.JFileChooserFinder;
+import org.assertj.swing.fixture.FrameFixture;
+import org.assertj.swing.fixture.JFileChooserFixture;
+import org.assertj.swing.fixture.JTableFixture;
+import org.assertj.swing.junit.testcase.AssertJSwingJUnitTestCase;
+import org.junit.Test;
+
+import static org.assertj.swing.finder.WindowFinder.findFrame;
+import static org.assertj.swing.launcher.ApplicationLauncher.application;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+public class UndoTest extends AssertJSwingJUnitTestCase {
+
+    private AWTExceptionHandler awtExceptionHandler;
+
+    @Override
+    protected void onSetUp() {
+        awtExceptionHandler = new AWTExceptionHandler();
+        awtExceptionHandler.installExceptionDetectionInEDT();
+        application(JabRefMain.class).start();
+
+        robot().waitForIdle();
+
+        robot().settings().timeoutToFindSubMenu(1_000);
+        robot().settings().delayBetweenEvents(50);
+    }
+
+    private void exitJabRef(FrameFixture mainFrame) {
+        mainFrame.menuItemWithPath("File", "Quit").click();
+        awtExceptionHandler.assertNoExceptions();
+    }
+
+    private String getTestFilePath(String fileName) {
+        return new File(this.getClass().getClassLoader().getResource(fileName).getFile()).getAbsolutePath();
+    }
+
+    private void importBibIntoNewDatabase(FrameFixture mainFrame, String path) {
+        mainFrame.menuItemWithPath("File", "Import into new database").click();
+
+        JFileChooserFixture openFileDialog = JFileChooserFinder.findFileChooser().withTimeout(10_000).using(robot());
+        robot().settings().delayBetweenEvents(1);
+        openFileDialog.fileNameTextBox().enterText(path);
+        robot().settings().delayBetweenEvents(50);
+        openFileDialog.approve();
+    }
+
+    @Test
+    public void undoCutOfMultipleEntries() {
+        FrameFixture mainFrame = findFrame(JabRefFrame.class).withTimeout(10_000).using(robot());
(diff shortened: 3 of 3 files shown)


src/main/java/net/sf/jabref/gui/BasePanel.java
```diff
@@ -355,24 +352,21 @@ private void setupActions() {
...
-                            (bes.size() > 1 ?
-                        Localization.lang("delete entries") :
-                        Localization.lang("delete entry")));
-                    // Loop through the list of entries, and delete them.
-                    for (BibEntry be : bes) {
+                    // can't be easily put in one NamedCompund because when the first entry is undone it triggers the
+                    // creation of another NamedCompound which kills THIS one,
+                    // which results in just one entry being restored.
+                    for (BibEntry be : bes){
+                        NamedCompound ce = new NamedCompound(Localization.lang("delete entry"));
```
tschechlovdev: why ce here and not de for delete entry? think de would be a better name

bartsch-dev: `ce` is the abbreviation for `CompoundEdit`, the super class for `NamedCompound`.

tschechlovdev: ok then it looks good for me

koppor:
I'm unsure whether this is the correct behavior: If I delete multiple entries at once, all of them should be restored. I thought that the undo/redo concept of JabRef is that always undo and redo actions are packaged as a whole.

Is it a valid workaround to globally flag whether an redo/undo action is in progress and thus no new undo thing is put on the undo stack.

src/main/java/net/sf/jabref/gui/BasePanel.java
simonharrer: I vote for nc instead, as this is the abbreviation for NamedCompound. The typical reader will be confused with ce, as he does not realize that there is an inheritance relationship.

simonharrer:
I would also prefer that the undo stuff is fixed so restoring several bibentries works.

src/main/java/net/sf/jabref/gui/BasePanel.java
stefan-kolb: I don't like such short names in general, why not namedCompound or compound?

oscargus:
But good that you found out the reason for the failing multiple entries! :+1:

src/main/java/net/sf/jabref/gui/BasePanel.java
bartsch-dev: I used this abbreviation because they are used everywhere in this class.
But I'll change them when they're creating such great discomfort.

bartsch-dev:
OK, I'll look further into it an look for a better solution!

tobiasdiez:
@bartsch-dev can you point me to the code where the insertion adds a new change in the undomanager? As far as I see, the revert action just calls database.insert() which does not call the undo manager. Thanks.

bartsch-dev:
When an entry is about to being restored [`UndoableRemoveEntry#undo()`](https://github.com/JabRef/jabref/blob/master/src/main/java/net/sf/jabref/gui/undo/UndoableRemoveEntry.java#L62) is called.
The entry is inserted back into the database `BibDatabase#insertEntry(BibEntry entry)`, which fires a [`DatabaseChangeEvent`](https://github.com/JabRef/jabref/blob/master/src/main/java/net/sf/jabref/model/database/BibDatabase.java#L185) and calls [`SpecialFieldDatabaseChangeListener#databaseChanged(DatabaseChangeEvent e)`](https://github.com/JabRef/jabref/blob/master/src/main/java/net/sf/jabref/specialfields/SpecialFieldDatabaseChangeListener.java#L25) and adds another Edit into the `undoManager`.
If we follow that trail we see that the `UndoManager` trims the Edits (line 502) which calls `die()` (line 244) and kills the `NamedCompound`which contains the other entries we haven't undone yet.

koppor:
I wonder whether it is a solution to add a second paramter to `insertEntry` indicating whether the action should be a silent change or not. - Or add a second method `insertEntrySilently`.

tobiasdiez:
Thank you @bartsch-dev for you helpful explanation! I think the cleanest way is to remove the `SpecialFieldDatabaseChangeListener` and let the calling code handle the undo-management. I.e. if I call database.insert() then I also have to create the undo event there. Otherwise we have problems everywhere with handling multiple changes (undo an import?). I think this strategy is also applied everywhere else, at least the group interface handles the undo thing itself.

koppor:
When removing the `SpecialFieldDatabaseChangeListener`, it should be taken care that the special fields still work. 

Currently, the fields are stored in the keyword field and are put into normal fields to enable rendering in the table. If the `SpecialFieldDatabaseChangeListener` is removed, this rendering IMHO won't take place any more.

When we rewrite the special fields (https://github.com/koppor/jabref/issues/84), this might change.

bartsch-dev:
Another working solution I found was to create a new `UndoableEdit` which manages multiple Entries with a List. When undoing the removal it uses the new `addEntries()`-Method inside `BibDatabase` which adds all the Entries and after that finally fires the DatabaseChanges.

bartsch-dev:
@koppor @tobiasdiez WDYT?

src/main/java/net/sf/jabref/gui/undo/MultipleUndoableRemoveEntries.java
```diff
@@ -0,0 +1,75 @@
...
+    @Override
+    public String getRedoPresentationName() {
+        return "Redo: remove "+ (entries.size() == 1 ?  "entry" : "entries");
+    }
+
+    @Override
+    public void undo() {
+        super.undo();
+
+        entries.stream().forEach(entry -> entry.setId(IdGenerator.next()));
```
simonharrer: why is this line necessary?

bartsch-dev: Ultimately it shouldn't.
I took it over from `UndoableRemoveEntry`.

But the main Idea is OK?

bartsch-dev:
As I explored deeper into the code I noticed that the Undo which the `SpecialFieldDatabaseChangeListener` adds doesn't do anything. Basically it's an empty `NamedCompund` (which gets discarded by the `UndoManager`).

Can that confirm anyone?

koppor:
Currently without code here. But double check that the stars in the column
rating are displayed in all undo cases. Deleting/adding an entry is OK, but
what about undo/redo of adding keyword "rank1" in the keyword field? - My
feeling is that undo/redo if that won't work when the updatelistener is
changed.

Also if an entry with keyword "rank1" is added -> is the column "ranking"
updated to show one star?

bartsch-dev:
The [`SpecialFieldDatabaseChangeListener`](https://github.com/JabRef/jabref/blob/master/src/main/java/net/sf/jabref/specialfields/SpecialFieldDatabaseChangeListener.java#L18) only reacts on the insertion of entries.

bartsch-dev:
I deleted the `SpecialFieldDatabaseChangeListener` and adjusted l10n.
As far as I can tell everything works just fine (tested it with deleting/undoing entries with rank/marked/color marked; @tschechlovdev maybe you can do some additional testing?).

I have NO idea why circleCI fails and how to fix it.

matthiasgeiger:
```
:compileJava/home/travis/build/JabRef/jabref/src/main/java/net/sf/jabref/gui/BasePanel.java:1324: error: cannot find symbol

        database.addDatabaseChangeListener(SpecialFieldDatabaseChangeListener.getInstance());

                                           ^

  symbol:   variable SpecialFieldDatabaseChangeListener

  location: class BasePanel
```
Should be quite obvious what is wrong and not be that hard to fix?!

bartsch-dev:
Woops, I slipped there. I extracted some code for another Bugfix which will get its own PR and extracted too much.
Now all checks pass but before that the build on circleCI exceeded the memory of 4G (no typing error, but I think circleCI meant 4GB) and failed.

tobiasdiez:
LGTM :+1:
Its nice that a bug can be fixed by removing code.

koppor:
Special fields functionality is disturbed by this patch.

## Steps to reproduce

1. open [special-fields.bib](https://github.com/JabRef/jabref/blob/master/src/test/resources/testbib/special-fields.bib)
2. double click on second entry (key: entry-with-rank2)
3. Go to "BibTeX source"
4. Copy the complete string into notepad++
5. Change `rank2` to `rank4` in both the key and the keywords field
6. Mark everything
7. Copy into the clipboard
8. Focus JabRef
9. Click into the white area
10. paste the entry.
11. expected: four stars shown at ranking column; as is: no stars shown at ranking column

## Intermediate solution
1. Open the editor
2. Go to the "General" tab
3. Change `rank4` to `rank3`
4. Save
5. Change `rank3` to `rank4`

## General remarks
I know that the keyword-based synchronization is a code which is not that straight forward as other code in JabRef. I personally like the functionality to have columns shown which are based on keywords I made. The reason might be that I sometimes edit my bibtex files manually, which also [nearly half of the survey participants do](http://www.jabref.org/surveys/2015/analysis).

bartsch-dev:
Thanks for that detailed explanation.

IMO the addition of the Undo-Object in [`SpecialFieldDatabaseChangeListener`](https://github.com/JabRef/jabref/blob/master/src/main/java/net/sf/jabref/specialfields/SpecialFieldDatabaseChangeListener.java#L25) is not useful.
Mostly because its added before the more important Undo-Object, the undo of the entry paste, and when you undo the paste this undo-Object is irrelevant.

koppor:
@bartsch-dev Could you add an integrationTest, meaning a test for the UI demonstrating the undo/redo behavior? Then, we don't have to rely on the [above manual steps](https://github.com/JabRef/jabref/pull/947#issuecomment-208730863) for testing. @bruehldev and @Braunch should be the experts in UI testing.

bartsch-dev:
I'll do it it the next sprint, which starts today.

src/main/java/net/sf/jabref/specialfields/SpecialFieldDatabaseChangeListener.java
```diff
@@ -1,28 +1,42 @@
...
-            SpecialFieldsUtils.syncSpecialFieldsFromKeywords(entry, nc);
-            nc.end();
-            JabRefGUI.getMainFrame().getCurrentBasePanel().undoManager.addEdit(nc);
+    public void databaseChanged(DatabaseChangeEvent event) {
+        if (event.getType() == DatabaseChangeEvent.ChangeType.ADDED_ENTRY && SpecialFieldsUtils.keywordSyncEnabled()) {
+            final BibEntry entry = event.getEntry();
+            // NamedCompound code similar to SpecialFieldUpdateListener
+            NamedCompound compound = new NamedCompound(Localization.lang("Synchronized special fields based on keywords"));
+            SpecialFieldsUtils.syncSpecialFieldsFromKeywords(entry, compound);
+            compound.end();
```
koppor: I think, compound.end() is not necessary. The object is thrown away anyhow.

simonharrer:
LGTM

Please rebase to allow it to be marged

koppor:
Please resolve the conflicts.

bartsch-dev:
I rebased and resolved the conflicts.

simonharrer:
Tests fail in master. Can you have a look? Otherwise, I will revert this merge.

bartsch-dev:
Hm, that's very strange (the tests in this PR passed).
I will look into it.

bartsch-dev:
After a quick study I noticed that assertJ is simply to fast for JabRef, meaning that the entry table after opening the database is simply not loaded and thus cannot be found. By increasing the delay between events from assertJ (```robot().settings().delayBetweenEvents(500);```) it can be averted but I don't like it that way as it depends on the power of the executing machine.
I will look further into it tomorrow.

You can revert the merge if you like, that way we can have a clean commit without a mess in between.

simonharrer:
The test seems to fail randomly. :(

Making it stable enough is fine for me.

Braunch:
@bartsch-dev  have you tried using robot().waitForIdle() ? 
This should wait until the Ui loaded the table without defining a timeout that slows the tests down on faster machines.

bartsch-dev:
That doesn't work.
A simple solution was to open an empty table at the very beginning, that way the loaded database opens without delay.

PR #950, opened 2016-03-12
Demo of AuthorList instead of AuthorsFormatter

koppor:
This PR demonstrates a straight-forward replacement of AuthorsFormatter by AuthorList. It shows that *none* of the cases treated by the current AuthorsFormatterTest can be treated by AuthorList. Only two cases not being handled by AuthorsFormatter can be handled by AuthorList.

I identified three things what the AuthorsFormatter does:

1. It adds dots after Initials. Meaning: A L Some gets Some, A. L.
2. It separates initials: Meaning: AL Some gets Some, A. L.
3. It knows what initials are: Olsen Y.Y. gets Olsen, Y.Y. (and not Y.Y. Olsen)

grabbed_20160313-123654

### DIFF
--- a/src/main/java/net/sf/jabref/logic/formatter/bibtexfields/AuthorsFormatter.java
+++ b/src/main/java/net/sf/jabref/logic/formatter/bibtexfields/AuthorsFormatter.java
@@ -16,19 +16,12 @@
 package net.sf.jabref.logic.formatter.bibtexfields;
 
 import net.sf.jabref.logic.formatter.Formatter;
-
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
+import net.sf.jabref.model.entry.AuthorList;
 
 /**
  * Class for normalizing author lists to BibTeX format.
  */
 public class AuthorsFormatter implements Formatter {
-    private static final Pattern LAST_F_F = Pattern.compile("(\\p{javaUpperCase}[\\p{javaLowerCase}]+) (\\p{javaUpperCase}+)");
-    private static final Pattern LAST_FDOT_F = Pattern.compile("(\\p{javaUpperCase}[\\p{javaLowerCase}]+) ([\\. \\p{javaUpperCase}]+)");
-    private static final Pattern F_F_LAST = Pattern.compile("(\\p{javaUpperCase}+) (\\p{javaUpperCase}[\\p{javaLowerCase}]+)");
-    private static final Pattern FDOT_F_LAST = Pattern.compile("([\\. \\p{javaUpperCase}]+) (\\p{javaUpperCase}[\\p{javaLowerCase}]+)");
-    private static final Pattern SINGLE_NAME = Pattern.compile("(\\p{javaUpperCase}[\\p{javaLowerCase}]*)");
 
     @Override
     public String getName() {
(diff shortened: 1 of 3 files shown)


oscargus:
I'm a bit ambiguous regarding if the formatter really should add . and guess that initials should be separated... I think we should decide on this.

Of course it helps improving the references from some perspective. Still it seems like two different use cases: 1) good author formatting in -> unified formatting out, 2) shit author formatting in -> make some guesses and provide most likely better formatting out. Not sure if they should be the same formatter. I would like 1) automatically, but not necessarily 2).

Not sure how Bibtex will handle A L Some... Probably not as initials.

Shouldn't it be Olsen, Y. Y.? (But then Y.-Y. Olsen shouldn't be separated.)

Two more (actual) author names which might be good for tests: Kenneth O and fred harris.

koppor:
Maybe we should rename the current code to "PersonListFixer" and use the AuthorList functionality for "PersonsFormatter" (which I find a better name than AuthorFormatter as also editor lists can be formatted)

oscargus:
Makes sense. And maybe get rid of the ordering and only focus on initials (which I guess is the main thing it does when ordering is excluded). Agree with "person".

tobiasdiez:
I have absolutely no idea how the AuthorsFormatter works, but in my humble opinion it should be a two liner: parse author string to PersonList and then call `getNormalizedString` or something like that on the PersonList object (so this is what you have changed here). Having two classes essentially doing the same thing fells redundant. So I would propose that you now add all the functionality of the old formatter to the PersonList class to fix the tests.

koppor:
@tobiasdiez I think, your suggestion is done at f8d33fd7e9ae93766b036fc51db6d0967d09e75b. I assume you meant `AuthorList` as there is no `PersonList`.

I discovered `net.sf.jabref.logic.mods.PersonName`, which also does some magic parsing, but not as magic as `net.sf.jabref.logic.formatter.bibtexfields.AuthorsFormatter.format(String)`.

With my screenshot above, I wanted to highlight that AuthorsFormatter is trying to *fix* names.

If I get you right, you talk about usecase "1)". AuthorsFormatter is mainly for usecase "2)". This demonstration PR makes AuthorsFormatter to fulfill usecase "1)".

For an import of data, for example via [FreeCite](http://freecite.library.brown.edu/), the authors have to be post-processed. (use-case 2)). 

Maybe, the current AuthorsFormatter should be called `ArbitraryPersonsToBibtexPerons`? And a new formatter `NormalizePersons` should be introduced. The NormalizePersons is also available as save action, but the ArbitraryPersonsToBibtexPerons is not. This does too much and should only be called "controlled".

tobiasdiez:
I suspect with use case 1 and 2 you mean the different cases as outlined above by @oscargus. While I see that these two different use cases exist in practice, I wouldn't separate them that clearly in the code. 
In my opinion all the author related stuff should be handled by the AuthorList class (or as you proposed PersonList). This means:
input -> parse by AuthorList -> output to sensible format

The first step should be the same for both use cases and only the second step might be different for "good" vs "bad" data. 

But as I said above, I have not a very deep understanding of this Author-formatting business.

oscargus:
The main reason to have two cases is that one of them is based on guessing
and basically cannot be systematically reversed.

koppor:
Since the code of the two use cases fundamentally differs and the guessing could go wrong, I would vote for keeping it separated. I understand, however, the user perspective to have only one formatter.

In the current state, I don't have resources to work on improving the fix algorithm. Would it be OK to keep them separated for now and to aim for an integration later?

tobiasdiez:
Small note: the recently released version 3.3 of biblatex changed a few things related to naming. Citing their changelog
```
INCOMPATIBLE CHANGE - There have been major changes to the handling of
  name lists in order to generalise name handling beyond the narrow
  confines of the legacy bibtex name parts of "last, first, prefix,
  suffix".
```
So they were apparently not happy with the standard ("legacy") bibtex format. Me might have a closer look at how biblatex handles author lists.

simonharrer:
What is the current status here? @koppor Should we close this for now and work on it later if we have time? I think this is not that pressing at the moment, isn't it?

oscargus:
I found the method expandAuthorInitials in ImportFormatReader(!) which does AL -> A. L. etc. Moved it to logic.util.StringUtil in #1138

koppor:
This was only a demo of making us aware of the differences between AuthorList and AuthorsFormatter. Think, this is really not urgent.

PR #956, opened 2016-03-13
Make tabbedPane in mainframe private

tobiasdiez:
Triggered by #931. 
- Makes tabbedPane in mainframe private
- Set tab tooltips in consistent way (for example they should now also show up after the tab list is sorted)

-
- [x] Change in CHANGELOG.md described? not worth it
- [x] Tests created for changes? well....
- [x] Tests green?


simonharrer:
:+1:

PR #230, opened 2015-10-12
New Iconset

stefan-kolb:
Also see #130 

- [x] Unlike expected, PNG files do not look sharper than the font version for me
- [x] Anyhow, we should go up to 24px as 16px seems way to small
- [x] ~~Strangely, some icons are missing in the materialize Github repository, e.g. `settings`~~ not relevant
- [x] We still have duplicated and unused icons in the properties file
- [x] Open file icon should be either `file-pdf` or `file-document` (or `attachment`) depending on actual file type.
- [x] How are icons for external file types handled? users cannot set this anymore, they have to use the default icons instead. 
- [x] squash commits so that temporary binary files are not in the history
- [x] separator are not removed when the icons are hidden (see special fields like rank)
- [x] add information about used font into `external-libraries.txt` and update "About" content

PR #961, opened 2016-03-14
Icon update to v1.5.54 + usage of regEx and case sensitivity icon

matthiasgeiger:
I just updated to the new material design icons version 1.5.54 - which is the first version to use fixed code points - so from now on it should not be necessary to recreate the whole mapping from the css file on each font update.

Moreover I added the Icons for RegEx and case sensitivity to the the searchbar to reduce the space needed by the search bar (description of the icons is available as a tooltip):
searchbar-old
searchbar-new

- [ ] Change in CHANGELOG.md described? - not relevant
- [ ] Tests created for changes? - only UI affected
- [x] Tests green?


koppor:
This is a follow up to #230.

PR #966, opened 2016-03-14
Fixes #962 Do not split Regexes and only match on entire fields

### LINKED ISSUE #962: Quantifying spaces in regex results in illegal search expression
* JabRef version (available in the About box): 3.3dev snapshot 29/02/2016
* Operating system and version: Ubuntu 15.10 Wily
* Steps to reproduce:
Quantifying spaces in regex doesn't work. It used to work. All my regex filter are not functionning anymore (and the groups are empty  because the regex is seen as faulty). I do not know what version caused this, however, I suspect it was fine in the version 2.

  1. Type the following regexes in either the search field or a filter regex field. All should work :
     * `a[^b]` : a followed by not b
     * `ab+`: a followed by at least one b
     * `ab*`: a followed by any number of b
     * `ab?`: a optionally followed by a b
     * `a `: a followed by a space
     * `(ab)`: ab in a group
  2. Those won't work and give a `Search failed: illegal search expression`:
     * `a[^ ]`: a followed by not space
     * `a +`: a followed by at least one space
     * `a *`: a followed by any number of spaces
     * `a ?`: a optionally followed by a space
     * `(a b)`: a b in a group, with space in the middle

Is this a desired behaviour ? I frequently use `[^ ]` to designate anything but a space and frequently put in alternatives such as : `(word and paradigm)|(item and arrangment)`.

stefan-kolb:
Attempts to fix the issue with whitespaces in #962.
Currently, all search queries are split on whitespaces and then matched separately on all fields.
This would change with this PR as Regex based search queries are only matched entirely on every field.
- [x] Matching looks fine, 
- [x] the higlighting is still based on splitting the string as far as I can see.

This PR must be assessed carefully to avoid unwanted effects.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,7 @@ to [sourceforge feature requests](https://sourceforge.net/p/jabref/features/) by
 - [#454](https://github.com/JabRef/jabref/issues/454) Add a tab that shows all remaining entry fields that are not displayed in any other tab
 - The LaTeX to Unicode/HTML functionality is much improved by covering many more cases
 - Ability to convert from LaTeX to Unicode in right-click field menu
-
+- Regex-based search is know only applied entirely and not split up to different Regexes on whitespaces
 - [#492](https://github.com/JabRef/jabref/issues/492): If no text is marked, the whole field is copied. Preview of pasted text in tool tip
 - Integrity check now also checks broken file links, abbreviations in journal and booktitle, and incorrect use of proceedings with page numbers
 - PdfContentImporter does not write the content of the first page into the review field any more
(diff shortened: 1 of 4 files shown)


koppor:
:+1: The alternative is to parse the regex and guess when a space is used as separator and when not. I would opt to be strict on regex parsing (as you propose) and rely on the users that they are able to use `|` when they want to search for alternatives.

simonharrer:
Have a look at SearchQueryHighlightObservable at method fireSearchlistenerEvent. There, you can add another elseif which handles regex based searches which does not use the sentence analyzer.

PR #968, opened 2016-03-15
Implemented about dialog in Java FX

boceckts:
This is the first dialog in JabRef being implemented in JavaFX. The Dialog is highly flexible in positioning and styling its components through a fxml and corresponding css stylesheet. If JavaFX should be a part of JabRef it is probabbly usefull to define a global stylesheet as well to become a consistent look and feel.
To include JavaFX in JabRef I had to adjust the starting and closing behaviour of Jabref.

In order to run these commits, Java 8_40 or newer is required!

- [ ] Tests green?


simonharrer:
Travis CI uses Jdk 1.8.0_31, and because of this cannot compile this.

boceckts:
As it says in the error message the class javafx.scene.control.Alert can not be found since this has been introduced in Java 8_40 along with javafx.scene.control.Dialog. However this is a essential part in creating a javafx gui since it is basically the equivalent class to JOptionPane / JDialog classes in swing.
Is it possible to update the jdk version of the Travis CI?

simonharrer:
Please do investigate on this.

Siedlerchr:
@boceckts  This is currently a problem with Travis,but it can be updated:
https://github.com/travis-ci/travis-ci/issues/3259

This one looks the most promising:
https://github.com/travis-ci/travis-ci/issues/3259#issuecomment-130860338

src/main/java/net/sf/jabref/gui/help/AboutDialogController.java
```diff
@@ -0,0 +1,155 @@
...
+
+        librariesHeading.setText("External Libraries:");
+        librariesContent.setText("https://github.com/JabRef/jabref/blob/master/external-libraries.txt");
+
+        codeHeading.setText("Code:");
+        codeContent.setText("https://github.com/JabRef/jabref");
+
+        closeButton.setText(Localization.lang("Close"));
+
+        copyButton.setText(Localization.lang("Copy_version_to_clipboard"));
```
tobiasdiez: Just use normal spaces instead of underscore. Then maybe the circleci test should pass.

koppor:
@boceckts As JabRef uses both CircleCI and Travis and tests for both OpenJDK and JRE, there won't be a simple solution.

Siedlerchr:
Higher jdk versions than Build 25 (this was from one blog post) seem to be not available, a workaound seems to be this hack:
https://discuss.circleci.com/t/java-version/2435

boceckts:
Thank you for all your feedback, I'm goinig to further investigate on this.

src/main/java/net/sf/jabref/gui/JabRefFrame.java
```diff
@@ -638,6 +647,24 @@ public void componentMoved(ComponentEvent e) {
...
         }
     }
+    
+    private void initAboutDialog() {
+        Platform.runLater(() -> {
+            FXMLLoader loader = new FXMLLoader();
+            try (InputStream imageStream = JabRef.class.getResourceAsStream("/images/external/JabRef-icon-48.png");
+                    InputStream fxmlStream = JabRef.class.getResourceAsStream("/gui/help/AboutDialogLayout.fxml");) {
+                DialogPane dialogContentPane = (DialogPane) loader.load(fxmlStream);
+                javafx.scene.image.Image img = new javafx.scene.image.Image(imageStream);
```
tschechlovdev: don't use abbreviation here. img-> image

src/main/java/net/sf/jabref/gui/help/AboutDialogController.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    @FXML
+    private void initialize() {
+        String header = String.format("JabRef %s", Globals.BUILD_INFO.getVersion());
+        aboutHeading.setText(header);
+
+        try (InputStream imageStream = JabRef.class.getResourceAsStream("/images/external/JabRef-icon-48.png");) {
+            jabrefLogo.setImage(new Image(imageStream));
+        } catch (Exception e) {
```
tschechlovdev: specify which Exception should be catched

src/main/java/net/sf/jabref/gui/help/FXAlert.java
```diff
@@ -0,0 +1,116 @@
...
+            }
+        }
+    }
+
+    private void setLocationRelativeToMainWindow() {
+        double mainX = JabRef.jrf.getLocationOnScreen().getX();
+        double mainY = JabRef.jrf.getLocationOnScreen().getY();
+        double mainWidth = JabRef.jrf.getSize().getWidth();
+        double mainHeight = JabRef.jrf.getSize().getHeight();
+
```
tschechlovdev: mainWidth is not such a meaningful name. I think that mainWindowWidth for example would be better

src/main/java/net/sf/jabref/gui/help/AboutDialogController.java
```diff
@@ -0,0 +1,154 @@
...
+
+    @FXML
+    private Button closeButton;
+
+    @FXML
+    private Button copyButton;
+
+
+    public AboutDialogController() {
+        // TODO Auto-generated constructor stub
```
tschechlovdev: remove the comment

boceckts:
I created a new AboutDialog class which handles its initialization itself as previously. This way there is less changes and imports in the JabrefFrame class.
The improvements mentioned in the previous comments are also included.

Siedlerchr:
Interestingly I found out that circle-ci is already using java8 u40:

> Using Java 1.8.0_40 from /usr/lib/jvm/jdk1.8.0/jre
From the log of:
`./gradlew -Pdev=true -Pinstall4jDir="install4j6" clean release --stacktrace || exit 0`

boceckts:
Thanks for pointing this out, I already wondered since I didn't touch the circle.yml yet the tests are passing on CircleCI.

src/main/java/net/sf/jabref/gui/help/AboutDialog.java
```diff
@@ -15,54 +15,40 @@
...
+        try (InputStream imageStream = JabRef.class.getResourceAsStream("/images/external/JabRef-icon-48.png");
+                InputStream fxmlStream = JabRef.class.getResourceAsStream("/gui/help/AboutDialogLayout.fxml");) {
+            DialogPane aboutDialogContentPane = (DialogPane) fxmlLoader.load(fxmlStream);
+            Image jabRefIcon = new Image(imageStream);
+            setDialogIcon(jabRefIcon);
+            setDialogPane(aboutDialogContentPane);
+            setDialogStyle(JabRefMain.class.getResource("/gui/help/AboutDialog.css").toExternalForm());
+        } catch (IOException e) {
+            logger.debug("AboutDialog could not be completely loaded!", e);
+        } catch (NullPointerException e) {
```
simonharrer: Do not catch NPEs.

koppor:
Please ensure that this version of JabRef also runs under Linux with OpenJDK.

boceckts:
As this states https://docs.travis-ci.com/user/trusty-ci-environment/ it should be possible to switch to the trusty beta build environment where we should be able to use the open jdk 1.8.0_60. I'm going to try this first before manually testing it.

boceckts:
Hm from the log of the build with the changed travis.yml it seems that the build uses the open jdk 1.8.0_45 but can not find the javafx packages. Is javafx even supported by the open jdk?

Siedlerchr:
@boceckts  JavaFX is part of the OpenJDK.
http://openjdk.java.net/projects/openjfx/

From what I see it is a problem of Ubuntu Trusty, the javafx is a kind of separate package, which is not yet backported to trusty:
https://bugs.launchpad.net/ubuntu/+source/openjdk-8/+bug/1398660


You can also spefiy multiple sdks:
Please do this to add oraclejdk (and remove openjdk until it gets implemented)
https://docs.travis-ci.com/user/languages/java#Testing-Against-Multiple-JDKs

boceckts:
I have changed the travis configuration file to use the newest version of the oracle jdk with the trick mentioned by @Siedlerchr .

lenhard:
Hm, this is a real problem. I'm not sure if we should continue with moving to JavaFX if it does not come natively with the openjdk in ubuntu. We can require Ubuntu users to install the oraclejdk, but we will get a lot of discussion and flaming for that.

matthiasgeiger:
If I get it right, the problem is mainly Ubuntu 14.04 LTS - the newer non-LTS versions should support openjdk8 (and hopefully JavaFX ;-))

Siedlerchr:
@matthiasgeiger  @lenhard 
Yes, the javaFX is official part of the openjdk, but it is a separate package.
 However, it has not been back ported to Ubuntu 14.04 LTS.

From what I read one could use the debian Version:
https://packages.qa.debian.org/o/openjfx.html

I am not that deep into Linux, so no idea.

koppor:
The next LTS is 16.04, which will have openfx support. (double check needed
- I only assume it, because ubuntu used to use debian/unstable as basis)

Backwards compatibility concerns did not stop us to make all plugins stop
working and to switch from Java6 to Java8. I am very optimistic in our move
to JavaFX. Especially because of supporting high res displays and because I
don't see how we can quickly make JabRef a high-quality webapp - code-wise
and usability-wise.

Siedlerchr:
And of coruse, if you install the orcaclejdk than it's not a problem in
14.04, too.

2016-03-24 0:03 GMT+01:00 Oliver Kopp <notifications@github.com>:

lenhard:
World upside down: @koppor is enthusiastic about a breaking change and I am sceptical? You convinced me. 

So this PR is ready to merge, right? Locally, I am still getting a compilation error, because it cannot find `AlertType`, although I am using the oracle jdk 1.8.073 under Win10, but that might also be due to some other local issues.

Siedlerchr:
You may need to conifgure your Eclipse to include the javafx libs correctly:
See the second answer:
http://stackoverflow.com/questions/22812488/using-javafx-in-jre-8

2016-03-24 9:30 GMT+01:00 Jörg Lenhard <notifications@github.com>:

lenhard:
I am using IntelliJ, but yes it was just a configuration issue. To me, the dialog looks like this:
jabref-about-dialog
Looks like variables are not read correctly?

Siedlerchr:
I have the same thing with the old about dialog, too when I run it from eclipse directly.

boceckts:
I think you need to configure eclipse/IntelliJ to run/build via gradle.

lenhard:
Yes, if I build the release jar and start it from the console, the about dialog looks correct on master. However, if I build the release for this branch it looks like this:
jabref-about-dialog
Also nothing happens if I click on the empty menu item. What else am I doing wrong?

Siedlerchr:
@lenhard  Yes, I experience the same after I applied the changes.  I'll take a look

boceckts:
This is a really strange behaviour as it works fine for me. However I have seen a fellow student having the same issue only that it was not reproduceable and appeared sometimes on the first execution after he pulled my changes. Further executions always worked for him though. Since there was no error message in the console it might be an issue on how/where you start it? I will further investigate and try to reproduce the effect on different operating systems. Which os and java version did you use?

lenhard:
Your comments make this sound very much like a threading issue. Platform.runLater also executes on the FX thread, right? @Siedlerchr perhaps adding the sysout leads to a synchronization that doesn'the necessarily happen otherwise.

Siedlerchr:
I tried a bit around and for me adding this line in AboutDialog:
`            fxmlLoader.setController(new AboutDialogController()) `
fixes the problem.
I can run it from Eclipse and it shows the code directly.
After doing a gradlew release I can run it with java -jar and it works, too.  (Win 7, x64 java 8u72
The same with gradlew run

boceckts:
@Siedlerchr this line should be called implicitly since the fx:controller attribute is set in the AboutDialogLayout.fxml file. This happens on my windows 7, x64 and java 8u73 after I pulled changes from the remote. Strangely if I don't change any files at all and re-run the execution it works fine and stays that way. @Siedlerchr Can you remove the line and check if execution works as normal now?

Siedlerchr:
> Strangely if I don't change any files at all and re-run the execution it works fine and stays that way. 
> @Siedlerchr Can you remove the line and check if execution works as normal now?

Yes, this is really odd behavior. I now experience the same.
Seems like it is only the first run which is problematic. All further runs seem to be ok

boceckts:
After investing quite some time to set up virtual machines I tried a lot to reproduce this odd behaviour on different os' and came to the following conclusion. 

**How I tested:**

1. I updated my Java Version to 1.8.0_77 (execpt for openjdk 1.8.0_66  and openjfx 8u60 which were the newest versions I could find)
2. I cloned my repository (if not already available)
3. If I had a local working branch named "AboutDialogFX" I first deleted this one.
4. Executed 'git checkout origin/AboutDialogFX'
5. Executed './gradlew run'
6. Check for menu item and open the about dialog if applicable
7. Close Application and repeat step 5-7, four times (and more)


**Results:**

Windows 7, 64bit, java8u77
1. run: About Jabref menu item available and dialog opens normally
2. run: **Menu item not available and dialog won't open, no exception**
3. run: About Jabref menu item available and dialog opens normally
4. run: About Jabref menu item available and dialog opens normally
**After trying a trying more runs, the 9th run showed the same behaviour as the 2nd one**

MacOsX 10.11.1, java8u77
1. run: About Jabref menu item available and dialog opens normally
2. run: About Jabref menu item available and dialog opens normally
3. run: About Jabref menu item available and dialog opens normally
4. run: About Jabref menu item available and dialog opens normally
_After trying a trying more runs, the same behaviour occured_

Ubuntu 14.04, 64bit, oracle java8u77
1. run: About Jabref menu item available and dialog opens normally
2. run: About Jabref menu item available and dialog opens normally
3. run: **Menu item not available and dialog won't open, no exception**
4. run: About Jabref menu item available and dialog opens normally
**After trying a trying more runs, the 6th, 7th and 8th run showed the same behaviour as the 3rd one**

Ubuntu 14.10, 64bit, oracle java8u77
1. run: About Jabref menu item available and dialog opens normally
2. run: About Jabref menu item available and dialog opens **but does not show elements and doesn't respond**
3. run: About Jabref menu item available and dialog opens normally
4. run: About Jabref menu item available and dialog opens normally
**After trying a trying more runs, the 6th run showed the same behaviour as the 2nd one**

Ubuntu 14.10, 64bit, openjdk8u66 openjfx8u60
1. run: About Jabref menu item available and dialog opens normally, **however build does not finish when closing JabRef, console stays at '> Building 85% > :run'**
2. run: About Jabref menu item available and dilalog opens normally, **however build does not finish when closing JabRef, console stays at '> Building 85% > :run'**
3. run: About Jabref menu item available and dialog opens normally, **however build does not finish when closing JabRef, console stays at '> Building 85% > :run'**
4. run: About Jabref menu item available and dialog opens normally, **however build does not finish when closing JabRef, console stays at '> Building 85% > :run'**
_After trying a trying more runs, the same behaviour occured_


**My Conculsion**

From my perspective this strange behaviour looks like something in java is not always 'initialized' when the application launches. As there is no exception even when the menu item is not shown I doubt the fault is in my commits. It also seems like the behaviour appears at random and cross platform on (almost) all oracle jdks (at least with version 8u77). I don't really know how to proceed as there is clearly something wrong but I have no idea how to fix this...

oscargus:
I have no idea, but good work in trying out the different configurations! :+1:

tobiasdiez:
I tried it with the [modified version](https://github.com/boceckts/jabref/pull/1) and it seems to be fine there. I think the issue is the method initComponentsInFXThread. As far as I can see there is no need to init the about action in the FXThread. So just add the menu entry as usual and then init the FX dialog in AboutAction.actionPerformed. Maybe this is worth a shot. 
Moreover, when you write `no exception`, did you also had a look at the help > error console? I had it quite a few times that no exception dialog was shown but an error message was logged there.

boceckts:
@tobiasdiez I will have a look at the modified version soon. Since I started the build from the console I looked at the os specific terminal/console and yes there is about 16xtimes this sentence in the console "OpenGL Warning: crPixelCopy3D:  simply crMemcpy'ing from srcPtr to dstPtr" on windows and ubuntu. But regardless if the about JabRef menu item is there or not. On ubuntu there is an additionally "libGL error: core dri or dri2 extension not found" and "libGL error: failed to load driver: vboxvideo" in the console. With "no exception" I meant something like a java Exception with a stacktrace pointing to a faulty line of code.

tobiasdiez:
@boceckts Thanks for the details. What I meant is the JabRef internal error console available under the help menu. While trying out some JavaFX stuff I had the feeling that some of the error messages just got lost (for example, change the fxml file to something invalid...then the about menu item does not show up but there is no exception), but some of them showed up in the internal error console.

boceckts:
@tobiasdiez Thanks for the tip! I thought everything that gets written to the internal JabRef Error Console also gets printed to the console I started it from? But I had a look at it and still didn't find any exceptions.
This new version creates a new dialog each time the user clicks on the About JabRef menu item. In the previous commit i only wanted to create the dialog once and store it in an object to speed up things when opening the dialog multiple times. This worked just fine for mac and windows but strangely enough the dialog looked different for every second time I opened it on ubuntu 14.04/15.10 with the latest oracle jdk. I didn't test this commit as excessive as the first one but it seems like the "hidden menu item" bug doesn't appear anymore and the window looks the same even when opening it multiple times on all my test systems.
The big disadvantage I see is that each time I want to open the dialog it takes time to create it first. It may not matter for such a small dialog but needs to be considered when porting bigger dialogs such as the preferences one in the same way.

tobiasdiez:
We recently discussed in the devcall how to further proceed with this PR. 
Please implement the following points:
- Use the [afterburner.fx library](https://github.com/AdamBien/afterburner.fx) to hard-coding of fxml or css file names. See https://github.com/boceckts/jabref/pull/1 (no need to use the dependency injection stuff)
- Add static strings to the fxml file, so for example `text="%Close"`. The preceding percentage sign makes sure that the text is automatically translated.
- As you mentioned above, a central css file should be created to define the general look (e.g. background color)
- The design of the dialog can be improved a bit. You can take the about dialog of intelliJ or firefox (or whatever you like) as a source of inspiration. There is no need to stick to the old JabRef layout.

-- 
Remarks for the next dialogs implemented in JavaFX:
- You might try out the MVVM pattern as sketched in https://github.com/boceckts/jabref/pull/1. We want to evaluate which pattern works better with JavaFX.
- A quick mockup of how you plan to design the dialog would be nice before you actually implement it. Most dialogs are ~~horrible~~ hard to use and should be improved.

boceckts:
@tobiasdiez I updated my branch against JabRef/master but the pull request is on JabRef/javafx which hasn't been updated. I think updating the JabRef/javafx branch would resolve the commit chaos.

As for the design I thought about using an accordion element to display the different section titles e.g. authors, developers and the user can expand whatever he wants to see.

CHANGELOG.md
```diff
@@ -34,13 +43,24 @@ to [sourceforge feature requests](https://sourceforge.net/p/jabref/features/) by
 - [#454](https://github.com/JabRef/jabref/issues/454) Add a tab that shows all remaining entry fields that are not displayed in any other tab
 - The LaTeX to Unicode/HTML functionality is much improved by covering many more cases
 - Ability to convert from LaTeX to Unicode in right-click field menu
+- Regex-based search is know only applied entirely and not split up to different Regexes on whitespaces
+- Application now starts from the JavaFX thread and the about dialog is now implemented in JavaFX
```
tobiasdiez: Since the Changelog should be understandable for the average user, I would propose something like `Partly switched to new UI technology (JavaFX). Redesigned about dialog.̀

src/main/java/net/sf/jabref/gui/FXAlert.java
```diff
@@ -0,0 +1,124 @@
...
+                Stage fxDialogWindow = (Stage) getDialogPane().getScene().getWindow();
+                fxDialogWindow.toFront();
+                fxDialogWindow.requestFocus();
+            });
+        }
+
+        @Override
+        public void windowGainedFocus(WindowEvent e) {
+            Platform.runLater(() -> {
+                Stage fxDialogWindow = (Stage) getDialogPane().getScene().getWindow();
```
tobiasdiez: I think it is worthwhile to extract the previous line to a new method `getDialogWindow/stage`

src/main/java/net/sf/jabref/gui/help/AboutAction.java
```diff
@@ -33,7 +38,12 @@ public AboutAction(String title, String tooltip, Icon iconFile) {
     @Override
     public void actionPerformed(ActionEvent e) {
         Platform.runLater(() -> {
-            new AboutDialog().show();
+            FXAlert aboutDialog = new FXAlert(Alert.AlertType.INFORMATION, Localization.lang("About JabRef"));
+
+            AboutDialogView view = new AboutDialogView();
```
tobiasdiez: In my opinion it would be slightly cleaner if all this FXAlert stuff is moved to a new `show` method in `AboutDialogView`.

src/main/java/net/sf/jabref/gui/help/AboutDialogViewModel.java
```diff
@@ -0,0 +1,146 @@
...
+    private final ReadOnlyStringWrapper developers = new ReadOnlyStringWrapper();
+    @Inject
+    protected BuildInfo buildInfo;
+    @FXML
+    private Button closeButton;
+
+    @FXML
+    private void initialize() {
+
+        header.set(String.format("JabRef %s", Globals.BUILD_INFO.getVersion()));
```
tobiasdiez: Can you please check if Globals.Build_Info is used somewhere else? If not then this global variable can be removed and its initialization moved here.

src/main/java/net/sf/jabref/gui/help/AboutDialogViewModel.java
```diff
@@ -0,0 +1,146 @@
...
+    @FXML
+    private void closeAboutDialog() {
+        Stage stage = (Stage) closeButton.getScene().getWindow();
+        stage.close();
+    }
+
+    @FXML
+    private void copyVersionToClipboard() {
+        ClipBoardManager.CLIPBOARD.setClipboardContents(Globals.BUILD_INFO.getVersion());
+        String message = String.format("%s - %s", Localization.lang("Copy version"), Globals.BUILD_INFO.getVersion());
```
tobiasdiez: Copied?

src/main/resources/net/sf/jabref/gui/help/AboutDialog.fxml
```diff
@@ -0,0 +1,111 @@
...
+                        <Hyperlink onAction="#openExternalLibrariesWebsite" text="${controller.websiteLibraries}" />
+                     </content>
+						</TitledPane>
+                  <TitledPane expanded="false" text="%Contribution">
+                     <content>
+                        <Hyperlink onAction="#openGithub" text="${controller.websiteGithub}" />
+                     </content>
+                  </TitledPane>
+					</panes>
+               <BorderPane.margin>
```
tobiasdiez: The margin can be handled via css, right?

tobiasdiez:
I updated the javafx branch to the same commit you are based. So once you update the PR hopefully the large amount of false changes disappears. 

I made some small comments to the code (by default, these are only visible in a collapsed state here on github for some reason). In general, the code really looks good! :+1: I especially like the FXAlert class.

So codewise this PR can be merged in from my point of view. Regarding the design, I liked the previous version a bit more. @JabRef/developers any ideas how the dialog should look like? Our dialog shows more infos than in other programs (like IntelliJ) so it hard to copy their design...

koppor:
I just updated the javafx branch to the latest HEAD (in master).

boceckts:
The comments are addressed now @tobiasdiez .
I have to note two things though, before someone else finds them.. :)
For me the localization of strings specified in the fxml file does not work for strings with an underscore (_). If I specified "%Copy_Version" in the fxml file it would give me a button text saying "Copy_Version" instead of "Copy Version". That's why I implemented the workaround with setting the button text to a string property of the controller ("${controller.copyButtonText}") and translating it there.
Second thing to note is that when using the afterburnerfx library the resource files (.css, .fxml) need to be in the same directroy as the View (e.g. AboutDialogView) in the resource folder. I first wanted to put them in the resources in a package gui.help instead of net.sf.jabref.gui.help but it could not be found and so far I haven't seen a possibility to set the path of the fxml and css files in the FXMLView class or its subclass.

src/main/java/net/sf/jabref/gui/help/AboutDialogView.java
```diff
@@ -0,0 +1,25 @@
...
+public class AboutDialogView extends FXMLView {
+
+    public AboutDialogView() {
+        super();
+        bundle = Localization.getMessages();
+    }
+
+    public void show() {
+        FXAlert aboutDialog = new FXAlert(AlertType.INFORMATION, Localization.lang("About JabRef"));
+
```
tobiasdiez: Remove empty lines.

tobiasdiez:
LGTM :+1:
Regarding the underscore in the localization string: this is strange.
As far as I can see, the FXML loader simply calls `resources.getString(aValue);` with the provided key, see http://hg.openjdk.java.net/openjfx/8/master/rt/file/tip/modules/fxml/src/main/java/javafx/fxml/FXMLLoader.java. If the key is not found, then an exception should be thrown. Probably, the resource is found, but not fixed (so `Copy_Version` is returned). Does it work if you change the value to `Copy\ version` in the language file?

Did you tried out if `%Copy version` (with space) or with escaped space `%Copy\ version` in the fxml file works? Probably not...

boceckts:
I tried writing "%Version_Copy", "%Version Copy" and "%Version\ Copy", the first gave me "Copy_Version" and the last two gave me an exception with the message "Resource "Copy Version" not found." and "Resource "Copy\ Version" not found."

tobiasdiez:
Ok, thanks for testing. One last shot: does `Copy_Version = Copy\ Version` (or `Copy_Version = justToTest`) in the resource file and `%Copy_Version` work?

boceckts:
The first one gives the desired result. the second one works as well.

boceckts:
Should I change it and use "Copy_Version=Copy\ Version" in the Localization file?
Also this is an issue when translating single words to multiple words in a different languages. For example I translated "Contribution=Beitrag_leisten" and used text="%Contribution" in the fxml file. Setting the language to german gives a tabbed pane named "Beitrag_leisten" instead of "Beitrag leisten". So maybe it is best to always use the controller injected methode for translating strings?

tobiasdiez:
Probably the cleanest version would be to derive a new class from ResourceBundle which incorporates the same logic as the translate method, i.e. something along the lines
````java
public Object handleGetObject(String key) {
  String propertiesKey = new LocalizationKey(key).getPropertiesKeyUnescaped();
  translation = resBundle.getString(propertiesKey);
  return new LocalizationKey(translation).getTranslationValue();
}
````
We will discuss this matter in the next devcall. But if you are motivated, then you are of course allowed to implement such a solution. Or you channel your motivation to the next dialog and create a [mockup](https://jabref.mybalsamiq.com/projects/ui/grid), as you prefer.

**Comment by boceckts** [comment]
Well I wanted to port the customize key bindings dialog but I wouldn't change its layout since in my opinion it's quite practical.

File: build.gradle [hunk 14]
```diff
@@ -95,6 +95,8 @@ dependencies {
     compile fileTree(dir: 'lib', includes: ['*.jar'])
 
     compile 'com.google.guava:guava:19.0'
+    
+    compile 'com.airhacks:afterburner.fx:1.6.3'
```
koppor: When adding a new library, please also add it to [external-libraries.txt](https://github.com/JabRef/jabref/blob/master/external-libraries.txt).

src/main/java/net/sf/jabref/gui/FXAlert.java
```diff
@@ -0,0 +1,128 @@
+/*  Copyright (C) 2003-2015 JabRef contributors.
```
koppor: Are the years correct?

src/main/resources/l10n/JabRef_en.properties
```diff
@@ -2259,4 +2259,10 @@ Show_diff=Show_diff
...
 word=word
 Show_symmetric_diff=Show_symmetric_diff
-
+Copy_Version=Copy_Version
+Copied_version_information_to_clipboard=Copied_version_information_to_clipboard
+Developers=Developers
+Authors=Authors
+License=License
+Libraries=Libraries
+Contribution=Contribution
```
koppor: I would call it "Contributing"

src/main/resources/net/sf/jabref/gui/help/AboutDialog.fxml
```diff
@@ -0,0 +1,96 @@
...
+                        <Hyperlink alignment="CENTER" contentDisplay="CENTER" onAction="#openJabrefWebsite" text="${controller.website}" />
+                     </children>
+                  </VBox>
+               </bottom>
+            </BorderPane>
+         </top>
+		</BorderPane>
+	</content>
+   <header>
+		<Label alignment="CENTER" contentDisplay="CENTER" onMouseClicked="#copyVersionToClipboard" styleClass="about-heading" text="${controller.heading}">
```
koppor: Indent messed up?

src/main/resources/net/sf/jabref/gui/help/AboutDialog.fxml
```diff
@@ -0,0 +1,96 @@
...
+                  <VBox styleClass="general-info" alignment="CENTER" BorderPane.alignment="CENTER">
+                     <children>
+                        <Label text="${controller.years}" />
+                        <Hyperlink alignment="CENTER" contentDisplay="CENTER" onAction="#openJabrefWebsite" text="${controller.website}" />
+                     </children>
+                  </VBox>
+               </bottom>
+            </BorderPane>
+         </top>
+		</BorderPane>
```
koppor: Indent?

koppor:
We should sit together and think about a good design. I would like to have tabs instead of these strange pull down menus. Reason: responsiveness of the UI. If I click something, I want to read the things instantly. Proposal: Fix the minor issues here, then merge and then think about a good design of the dialog.

I liked the JabRef 3.2 About dialog more - maybe we can put each heading into a tab and render the first two paragraphs above that.

grabbed_20160425-090444
grabbed_20160425-090519

src/main/java/net/sf/jabref/gui/help/AboutDialogViewModel.java
```diff
@@ -0,0 +1,186 @@
...
+
+        developers.set(Globals.BUILD_INFO.getDevelopers());
+
+        authors.set(Globals.BUILD_INFO.getAuthors());
+
+        websiteLibraries.set("https://github.com/JabRef/jabref/blob/master/external-libraries.txt");
+
+        websiteGithub.set("https://github.com/JabRef/jabref");
+
+        copyButtonText.set(Localization.lang("Copy_Version"));
```
tobiasdiez: Please change it back to `%Copy Version` in the fxml file. We will provide a proper fix for this issue.

koppor:
For reference: The current state of the dialog:

grabbed_20160425-090724

boceckts:
I changed the code accordingly to your comments.
Note that at the current stage the dialog can not be shown as it throws an exception with the message `Resource "Copy Version" not found`.

tobiasdiez:
I will now merge this PR in (in the JavaFX branch) so that you can work on further dialogs. 

You are right, the keyboard shortcuts dialog looks relatively good. The "grab" button is probably superfluous and "Default" should specify if it resets just one shortcut or all. Moreover, I like the design in the Vivaldi browser:
pic
See also the description https://help.vivaldi.com/article/keyboard-shortcuts/.
I especially like that the shortcuts are grouped.

koppor:
Current state is still not that nice.

grabbed_20160907-185247

I know that removing code is hard, but could you please make the dialog appearing more than the old one and showing all information **nice** in one place? I think, we can show it on one page (maybe not the developers). We do not need collapsing for one liners.

boceckts:
I will simplyfy it and remove the tab view.

boceckts:
@JabRef/developers opinions to the following design?

aboutdialogfx

Siedlerchr:
Looks good! Just one minor thing, the URL for contributing seems to be not aligned at the same height as the label

tobiasdiez:
Looks way better!
Only criticism:
- The background color of the header is...well...ugly :smile: 
- Display the URLs is unnecessary, better link text 
- Remove border around the authors field (if easily possible)
- The license is wrong ( ? @koppor ? )
- The sub-header "Developer" and "Authors" don't stand out.

I would slightly reorganize everything, but your version is also fine
image
(without the underlined text, that's a relict from the markup software)

Siedlerchr:
@tobiasdiez  and no Comic Sans MS style (as from the mockup) http://www.comicsanscriminal.com/  ;)

boceckts:
So there are two new versions similar to what @tobiasdiez suggested.
(For some reason the authors contain weird characters for me)

screen shot 2016-09-09 at 12 40 38
screen shot 2016-09-09 at 12 45 01

It really looks a bit weird to me with so much space in the top but the information to display is rather small, maybe resize the dialog a bit smaller?

boceckts:
I tried a smaller width for the dialog and personally this is my favorite, all the headings and information are left aligned and the icon has a small reflection which does not interfere with any text. Opinions are welcomed.

screen shot 2016-09-09 at 15 54 17

Siedlerchr:
The small dialog looks good.  :heart:

tobiasdiez:
Yes I really love the smaller dialog. Good work! Where is the PR?? I want to merge it right-away :smile: 
(small suggestion: maybe move the "Developers" headline a bit down)

Siedlerchr:
@boceckts  Could you please test with the gradle change  I introduced in #1945 and see if authors now are displayed correctly`?. And please check the encoding of the Authors file (should be UTF-8 without BOM)

boceckts:
@Siedlerchr the encoding is set to UTF-8 without BOM but unfortunately your changes have no effect for me.
@tobiasdiez There it is 😄 and I moved down the sub heading a little bit.

Siedlerchr:
Regarding the encoding problems, I just fixed it with #1954.  Please try again with the changes I made (gradle changes + version info)
And check: `src\main\resources\build.properties` if it is correctly encoded as utf8 without bom.

Issue #920, opened 2016-03-08
Table of formatters


tobiasdiez:
This issue summarizes all the available formatters.
The aim is to improve their names and descriptions.
I think the most practicable way is to directly edit this post with the new proposed version.

Class Name | Key | Name | Description
------------ | -------------| -------------| -------------
AuthorsFormatter | AuthorsFormatter | BibTex authors format | Normalizes lists of persons in %s to the BibTeX standard
UpperEachFirst CaseChanger |  UpperEachFirst CaseChanger | Upper Each First | Converts the first character of each word in %s to upper case (and all others to lower case), but does not change words starting with "{".
UpperCaseChanger | UpperCaseChanger | UPPER | Converts all characters in %s to upper case, but does not change words starting with "{"
MonthFormatter | MonthFormatter | Month | Normalizes content of %s to the format #mon#.
LatexFormatter | LatexFormatter | Latex | Cleans latex code in %s.
UpperFirst CaseChanger | UpperFirst CaseChanger | Upper first | Converts the first character of the first word in %s to upper case (and the remaining characters of the first word to lower case), but does not change anything if word starts with "{".
AuthorsMinifier | MinifyAuthors | Minify authors | Replaces three or more authors with "and others".
DateFormatter | DateFormatter | Date | Normalizes dates in %s to the BibLaTeX standard yyyy-mm-dd or yyyy-mm.
TitleCaseChanger | TitleCaseChanger | Title | Converts all words in %s to upper case, but converts articles, prepositions, and conjunctions to lower case.
CaseKeeper | CaseKeeper | CaseKeeper | Adds {} brackets around acronyms, month names and countries in %s to preserve their case.
PageNumbers Formatter | PageNumbers Formatter | Page numbers | Ensures that pages numbers in %s are of the form num--num.
LowerCaseChanger | LowerCaseChanger | lower | Converts all characters in %s to lower case, but does not change words starting with "{"
TrimFormatter | TrimFormatter | Trim whitespace | Removes any leading and trailing whitespace in %s.
HTMLToLatex Formatter | HtmlConverter **HtmlToLatex** | HTMLConverter **HTML to LaTeX** | Changes HTML expressions in %s to their LaTeX equivalent.
SuperscriptFormatter | SuperscriptFormatter | Superscripts | Transforms ordinal numbers in %s into LaTex superscripts.
UnitFormatter | UnitFormatter | UnitFormatter | Converts units in %s to LaTeX code.
[...]

PR #974, opened 2016-03-15
Always trim fields on save and remove TrimFormatter

tobiasdiez:
The consensus in #920 was that fields should always be written without leading or trailing whitespace. This is done in this PR and the TrimFormatter is removed.
(by the way, empty fields are never written as one of the added tests shows) 

- [x] Tests green?


matthiasgeiger:
LGTM

PR #975, opened 2016-03-15
Simplify github templates

tobiasdiez:
I find the checkboxes `Changes in pull request outlined? (What, why, ...)` and `Tests green?` a bit redundant. Especially since the build result is displayed directly under the PR message. So with this PR the template is even more simplified.

- [x] Change in CHANGELOG.md described? - pointless
- [x] Tests created for changes? - not applicable
- [x] Tests green?


.github/ISSUE_TEMPLATE.md
```diff
@@ -1,7 +1,7 @@
-* JabRef version (available in the About box):
-* Operating system and version:
 * Steps to reproduce:
   1. ...
   2. ...
   3. ...
-  * If applicable, excerpt of the bibliography file, screenshot, and excerpt of log (available in the error console)
+* If applicable, excerpt of the bibliography file, screenshot, and excerpt of log (available in the error console)
+* JabRef version (available in the About box):
```
koppor: I liked to read the JabRef version first. Maybe we should reuse your idea from the PR template:

```
JabRef version <version as shown in the about box> on <Windows 10|Ubuntu 14.04|Mac OS X 10.8|...>
```

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,4 +1,4 @@
+<describe the changes you have made here>
```
koppor: I like that. May be we should add `motivation, considered alternatives, ...`. Not sure if that is too much.

matthiasgeiger:
The updated pull request template is fine for me. 

But in my opinion the previous Issue-Template was better ;-)

.github/PULL_REQUEST_TEMPLATE.md
simonharrer: In favor of also the why, because this is often forgotten ...

Siedlerchr: Describe the changes is enough, maybe just add a simple:
Why did you make the changes?  -> The one could add here the issue ref or just describe a one liner.
I think motivation, considered alternatives are only usefuil for major architecture changes

tobiasdiez:
@koppor when you change this, can you also add `Screenshots added (for bigger UI changes)` to the PR template? Thanks. :+1:

stefan-kolb:
We should add a hint that if they are using a dev version of JabRef they should always first try the latest version before reporting a bug. While it is good that they use the dev version and try new stuff before it is released, in the last days it happened more often that they reported bugs that were already fixed in the most recent version.

koppor:
Done.

oscargus:
Wouldn't it make sense to ask the user to try the latest development version as well? Or at least look at the ChangeLog of the development version?

stefan-kolb:
For full versions I think its ok to shift this responsibility on our side. For development versions a user should make sure it is broken on the latest dev build.

tobiasdiez:
Thanks @koppor for the changes. I merge this in for now. Further modifications can be made in a new PR.

Issue #978, opened 2016-03-16
Field properties - future implementation


oscargus:
Based on some earlier discussions, e.g., koppor/jabref#59 #967 koppor/jabref#32 , it seems to make sense to redefine and restructure the properties of each field. At the moment it is handled by https://github.com/JabRef/jabref/blob/master/src/main/java/net/sf/jabref/gui/InternalBibtexFields.java and especially the class BibtexSingleField in that file.

Currently the following properties are defined:
- `final String name;` - the name of the field
-  `int flag` - a combination of 

``` java
        private static final int STANDARD = 0x01; // it is a standard bibtex-field
        private static final int PRIVATE = 0x02; // internal use, e.g. owner, timestamp
        private static final int DISPLAYABLE = 0x04; // These fields cannot be shown inside the source editor panel
        private static final int WRITEABLE = 0x08; // These fields will not be saved to the .bib file.
```
- `int length` - field length in characters(?)
- `double weight` - weight factor used for relative row size in the entry editor columns(?)
- `int editorType` - may be one of

``` java
    // constants for editor types:
    public static final int STANDARD_EDITOR = 1;
    public static final int FILE_LIST_EDITOR = 2;
```
- `String extras;` - Determines any extra component in the entry editor, currently one of:

``` java
    public static final String EXTRA_YES_NO = "yesNo"; // Blank/Yes/No Combo-box
    public static final String EXTRA_URL = "url"; // Drop target for URL
    public static final String EXTRA_DATEPICKER = "datepicker"; // Calendar button and double-click in field to set current date
    public static final String EXTRA_JOURNAL_NAMES = "journalNames"; // Journal abbreviation button
    public static final String EXTRA_EXTERNAL = "external"; // Open external viewer on double-click
    public static final String EXTRA_BROWSE = "browse"; // Browse button, file dialog
    public static final String EXTRA_SET_OWNER = "setOwner"; // Button to set owner to current used
    public static final String EXTRA_MONTH = "month"; // Button to show the months and set abbreviation
```
- `boolean numeric;` - true if the field is expected to be numeric and in that way one can sort based on that

[...]

PR #979, opened 2016-03-16
Refactored Singlebibtexfield

oscargus:
Based on the initial ideas in #978 the class SingleBibtexField is extracted and two EnumSets are used for implementing the existing functionality.

### DIFF
--- a/src/main/java/net/sf/jabref/bibtex/BibtexSingleField.java
+++ b/src/main/java/net/sf/jabref/bibtex/BibtexSingleField.java
@@ -3,20 +3,22 @@ package net.sf.jabref.bibtex;
 import java.util.EnumSet;
 import java.util.Set;
 
-import net.sf.jabref.gui.GUIGlobals;
-
-// --------------------------------------------------------------------------
-// a container class for all properties of a bibtex-field
-// --------------------------------------------------------------------------
+/**
+ *
+ * Class for keeping properties of a single BibTeX/BibLatex field
+ *
+ */
 public class BibtexSingleField {
 
-    // some fieldname constants
+    // some field constants
     public static final double DEFAULT_FIELD_WEIGHT = 1;
     public static final double MAX_FIELD_WEIGHT = 2;
 
     public static final double SMALL_W = 0.30;
     public static final double MEDIUM_W = 0.5;
     public static final double LARGE_W = 1.5;
+
+    public static final int DEFAULT_FIELD_LENGTH = 100;
 
 
     private enum Flag {
@@ -32,9 +34,9 @@ public class BibtexSingleField {
 
     // contains the standard, private, displayable, writable infos
     // default is: not standard, public, displayable and writable
-    private final Set<Flag> flag = EnumSet.of(Flag.DISPLAYABLE, Flag.WRITEABLE);
+    private final Set<Flag> flags = EnumSet.of(Flag.DISPLAYABLE, Flag.WRITEABLE);
 
-    private int length = GUIGlobals.DEFAULT_FIELD_LENGTH;
+    private int length = DEFAULT_FIELD_LENGTH;
     private double weight = DEFAULT_FIELD_WEIGHT;
 
     // the extras data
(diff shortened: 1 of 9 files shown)


src/main/java/net/sf/jabref/bibtex/BibtexSingleField.java
```diff
@@ -0,0 +1,180 @@
+package net.sf.jabref.bibtex;
+
+import java.util.EnumSet;
+import java.util.Set;
+
+import net.sf.jabref.gui.GUIGlobals;
+
+// --------------------------------------------------------------------------
+// a container class for all properties of a bibtex-field
+// --------------------------------------------------------------------------
```
simonharrer: Have a look at the comments and try to use JavaDoc instead.

src/main/java/net/sf/jabref/bibtex/BibtexSingleField.java
```diff
@@ -0,0 +1,180 @@
...
+
+    public BibtexSingleField(String fieldName, boolean pStandard, double pWeight, int pLength) {
+        name = fieldName;
+        setFlag(pStandard, Flag.STANDARD);
+        weight = pWeight;
+        length = pLength;
+    }
+
+    // -----------------------------------------------------------------------
+    // -----------------------------------------------------------------------
```
simonharrer: Such lines can simply be removed

src/main/java/net/sf/jabref/bibtex/BibtexSingleField.java
```diff
@@ -0,0 +1,180 @@
...
+        WRITEABLE;
+    }
+
+
+    // the field name
+    private final String name;
+
+    // contains the standard, private, displayable, writable infos
+    // default is: not standard, public, displayable and writable
+    private final Set<Flag> flag = EnumSet.of(Flag.DISPLAYABLE, Flag.WRITEABLE);
```
simonharrer: should be flags as a set contains more than one flag

src/main/java/net/sf/jabref/bibtex/BibtexSingleField.java
```diff
@@ -0,0 +1,180 @@
...
+    // -----------------------------------------------------------------------
+    // -----------------------------------------------------------------------
+
+    /**
+     * Sets or onsets the given flag
+     * @param setToOn if true, set the flag; if false, unset the flat
+     * @param flagID, the id of the flag
+     */
+    private void setFlag(boolean setToOn, Flag flagID) {
+        if (setToOn) {
```
simonharrer: would it be possible to only set the flags in the constructor and not allow it to be changed later?

src/main/java/net/sf/jabref/gui/InternalBibtexFields.java
```diff
@@ -313,20 +307,12 @@ private static BibtexSingleField getField(String name) {
         return null;
```
simonharrer: You could introduce optional here

src/main/java/net/sf/jabref/bibtex/BibtexSingleField.java
```diff
@@ -0,0 +1,180 @@
...
+
+
+    // the field name
+    private final String name;
+
+    // contains the standard, private, displayable, writable infos
+    // default is: not standard, public, displayable and writable
+    private final Set<Flag> flag = EnumSet.of(Flag.DISPLAYABLE, Flag.WRITEABLE);
+
+    private int length = GUIGlobals.DEFAULT_FIELD_LENGTH;
```
simonharrer: could the default field lenght moved to this class as well?

oscargus:
Very good points! Too much copy-paste (as usual). Will sort it out tomorrow.

src/main/java/net/sf/jabref/bibtex/BibtexSingleField.java
oscargus: Possibly. However, the way the current code looks it is a bit of work and since this is primarily used for the special fields, I think it is better to wait until that is reworked.

simonharrer:
Looks good. Other improvements can be made in separate PRs.

src/main/java/net/sf/jabref/bibtex/BibtexSingleField.java
```diff
@@ -0,0 +1,174 @@
...
+        if (setToOn) {
+            // set the flag
+            flags.add(flagID);
+        } else {
+            // unset the flag
+            flags.remove(flagID);
+        }
+    }
+
+    // -----------------------------------------------------------------------
```
simonharrer: Can you remove these lines? There are some others in this file similar to this as well.

oscargus: Ahh, I missed some...

PR #980, opened 2016-03-16
Replaced Bitset with EnumSet in OOUtil

oscargus:
Clearly EnumSet is superior.

### DIFF
--- a/src/main/java/net/sf/jabref/openoffice/OOUtil.java
+++ b/src/main/java/net/sf/jabref/openoffice/OOUtil.java
@@ -16,7 +16,8 @@
 package net.sf.jabref.openoffice;
 
 import java.util.List;
-import java.util.BitSet;
+import java.util.Set;
+import java.util.EnumSet;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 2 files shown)


simonharrer:
LGTM

src/main/java/net/sf/jabref/openoffice/OOUtil.java
```diff
@@ -236,7 +239,7 @@ public static void insertTextAtCurrentLocation(XText text, XTextCursor cursor, S
 
         // TODO: the <monospace> tag doesn't work
         /*
-        if (formatting.get(FORMAT_MONOSPACE)) {
+        if (formatting.contains(Formatting.MONOSPACE)) {
```
Siedlerchr: Also this PR is already merged, does this work with newer OO/LO Versions?
Anyone tried this?

oscargus: I think I tried uncommenting that part at some stage, but do not remember exactly what happens. As I recall it nothing. So it seems like the part that doesn't work is that setting the OO/LO property the way it is done in the code doesn't change the font to a monospace font. The JabRef part works though as far as I can tell, it is only how to make OO/LO change it that is the problem.

PR #981, opened 2016-03-16
No more printStackTrace()

oscargus:
The last printStackTrace() are replaced with loggers. (Except one in a test.)


simonharrer:
Awesome

PR #982, opened 2016-03-16
Preserve bibtex import order

tobiasdiez:
@obraliar https://github.com/JabRef/jabref/pull/504#discussion_r56398599 noted that the order of the bibentries in `parserResult.getDatabase().getEntries` is not the same as in the bib file.
As it turned out, fixing the order required to store the bibentries not as `Map<id, BibEntry>` but simply as a list. So I removed almost everything which is connected to ID's. Now they are only used for writing bibentries (to determine the original order) and for the sql export. 

- [x] Change in CHANGELOG.md described? Not relevant
- [x] Tests green?

### DIFF
--- a/src/main/java/net/sf/jabref/exporter/ExportFormats.java
+++ b/src/main/java/net/sf/jabref/exporter/ExportFormats.java
@@ -200,13 +200,9 @@ public class ExportFormats {
                         }
                     }
                     final IExportFormat format = eff.getExportFormat();
-                    Set<String> entryIds = null;
+                    List<BibEntry> entries = null;
                     if (selectedOnly) {
-                        List<BibEntry> selected = frame.getCurrentBasePanel().getSelectedEntries();
-                        entryIds = new HashSet<>();
-                        for (BibEntry bibtexEntry : selected) {
-                            entryIds.add(bibtexEntry.getId());
-                        }
+                        entries = frame.getCurrentBasePanel().getSelectedEntries();
                     }
 
                     // Set the global variable for this database's file directory before exporting,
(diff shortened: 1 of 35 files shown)


src/main/java/net/sf/jabref/exporter/ExportFormats.java
```diff
@@ -223,7 +219,7 @@ public void actionPerformed(ActionEvent e) {
                     Globals.prefs.put(JabRefPreferences.EXPORT_WORKING_DIRECTORY, file.getParent());
 
                     final File finFile = file;
-                    final Set<String> finEntryIDs = entryIds;
+                    final List<BibEntry> finEntries = entries;
```
oscargus: Why is a new variable introduced?

src/main/java/net/sf/jabref/logic/labelpattern/LabelPatternUtil.java
```diff
@@ -499,11 +499,11 @@ public static void makeLabel(MetaData metaData, BibDatabase dBase, BibEntry entr
         if (!alwaysAddLetter && (occurrences == 0)) {
             // No dupes found, so we can just go ahead.
             if (!key.equals(oldKey)) {
-                if (database.getEntryById(entry.getId()) == null) {
+                if (! database.containsEntryWithId(entry.getId())) {
```
oscargus: No space...

oscargus:
:+1: (Just one question and one space...)

I guess an intermediate step would be to have a List of ids instead of a Set as that list should probably have a smaller memory footprint compared to bibentry, but as long as it is not a problem (and I guess it isn't, even with thousands of entries), this seems to be the way to go!

src/main/java/net/sf/jabref/exporter/ExportFormats.java
tobiasdiez: Further down it is used in an anonymous method, which is only possible if it final.

simonharrer:
:+1:

Issue #958, opened 2016-03-14
Size of database properties dialog


koppor:
The size of the database properties dialog is not fixed. When opening the dialog the first time, it seems to be not high enough.

The dialog should be of a fixed size or at least high enough at the first start.

Siedlerchr:
I did not notice a problem on Win7. The only thing is that the space between the last two buttons (bottom center) and the margin looks too small.

simonharrer:
If you are already at this, maybe you could have a look at replacing the JLabel for the description of the currently selected save action with something like a text area which is used in the about dialog.

matthiasgeiger:
The problem is visible in the screenshot @koppor posted at #957: With win10 the buttons are cut a few pixels. Increasing the size of the dialog for 10-20px should be sufficient to show the buttons correctly.

Siedlerchr:
Another thing I noticed is that the size of the "Minus" Button (CSS-Icon) is bigger than the Reset Button.
![iconbuttondbproperties](https://cloud.githubusercontent.com/assets/320228/13775999/f83c444e-eaa7-11e5-8b9e-eb9a2ee10b87.png)

Any idea how to scale the Buttons correctly? Never worked with jgoodies before.

Siedlerchr:
This can be closed.

Issue #957, opened 2016-03-14
Save actions: better field selection


koppor:
At "Save sort order" in the same dialog, one can also select a field. I would have expected the same choice for the selection of the fields to apply the save action. Is that possible? Reason: I am too lazy to input the field name manually.

![grabbed_20160313-212741](https://cloud.githubusercontent.com/assets/1366654/13731147/ab897bee-e962-11e5-84ff-a9546b62021e.png)

simonharrer:
Please not that there is the "special field" "all" which applies the save action on all fields.

Siedlerchr:
I will look into this, Together with #958

koppor:
@Siedlerchr Yes, I know. That was sort of a quick hack to enable quick selection of common fields, but also the input of arbitrary fields (not contained in the combobox list). Don't know, whether that's possible with a normal JCombobox, too.

Siedlerchr:
For the sort criterion:
A solution to this would be to use an Editable Combobox (Default Combobox with Editable=true)
https://docs.oracle.com/javase/tutorial/uiswing/components/combobox.html#editable
- Enables the selection of items
- Add new items not in the list of the combobox.
=> This would make the textfield unnecessary.

For the Save Action I would suggest the same solution

Siedlerchr:
This save sort order selection with the Combobox and the Textfield is also used in the Preferences->Export sorting, it is the same class which is used for both.

A rework would then profit in both cases.

PR #983, opened 2016-03-16
Some random cleanups (and probably a bug-fix)

oscargus:
* Removed unused imports
* Some code style fixes
* Most likely fixed a potential NPE in SaveDataBaseAction (chosenFile will never be null here, but f might be and is later dereferenced)

### DIFF
--- a/src/main/java/net/sf/jabref/importer/EntryFromPDFCreator.java
+++ b/src/main/java/net/sf/jabref/importer/EntryFromPDFCreator.java
@@ -17,7 +17,6 @@ import net.sf.jabref.pdfimport.PdfImporter.ImportPdfFilesResult;
 import net.sf.jabref.JabRef;
 import net.sf.jabref.external.ExternalFileType;
 import net.sf.jabref.external.ExternalFileTypes;
-import net.sf.jabref.logic.xmp.EncryptionNotSupportedException;
 import net.sf.jabref.logic.xmp.XMPUtil;
 
 /**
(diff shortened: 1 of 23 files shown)


simonharrer:
:+1:

src/main/java/net/sf/jabref/gui/dbproperties/FieldFormatterCleanupsPanel.java
```diff
@@ -44,7 +44,7 @@
 
     private JButton resetButton;
 
-    private FieldFormatterCleanups defaultFormatters;
+    private final FieldFormatterCleanups defaultFormatters;
```
Siedlerchr: Could you please exclude this file from your commit as I am currently working with this file for #957  and #958?
I included your changes already.

oscargus: Sure, it was only line 233 which I personally changed, Eclipse took care of the rest.

oscargus:
Reverted the change the @Siedlerchr asked for, but did a force push so that it wasn't changed back and forth...

src/main/java/net/sf/jabref/importer/EntryFromPDFCreator.java
```diff
@@ -68,12 +67,12 @@ public boolean accept(File f) {
...
-        addEntyDataFromXMP(pdfFile, entry);
-
+        addEntryDataFromXMP(pdfFile, entry);
+        
         if (entry.getField("title") == null) {
         	entry.setField("title", pdfFile.getName());
         }
-
+        
         return entry;*/
```
simonharrer: I would suggest just delete this commented out code.

oscargus: Maybe. However, in that case we can also remove addEntryDataFromPDDocumentInformation and  addEntryDataFromXMP (if I recall correctly) and it seems like those could be useful if they are used. Of course we can recreate those if/when needed.

simonharrer: Ok, then we leave it in for now.

PR #985, opened 2016-03-17
More small fixes and cleanups

oscargus:


src/main/java/net/sf/jabref/importer/fetcher/IEEEXploreFetcher.java
```diff
@@ -172,16 +180,10 @@ public boolean processQuery(String query, ImportInspector dialog, OutputPrinter
 
         } catch (MalformedURLException e) {
             LOGGER.warn("Bad URL", e);
-        } catch (ConnectException e) {
+        } catch (ConnectException | UnknownHostException e) {
             status.showMessage(Localization.lang("Connection to IEEEXplore failed"), DIALOG_TITLE,
```
koppor: Could we introduce the typical `"Connection to %0 failed", "IEEE"`? - See https://github.com/JabRef/jabref/pull/948. Maybe in a separate PR, but I just wanted that this is not forgotten.

src/main/java/net/sf/jabref/importer/fetcher/IEEEXploreFetcher.java
```diff
@@ -279,15 +281,15 @@ private BibEntry cleanup(BibEntry entry) {
             title = title.replaceAll("/[sS]pl ([^/]+)/", "\\$\\\\$1\\$");
             // Deal with subscripts and superscripts
             if (Globals.prefs.getBoolean(JabRefPreferences.USE_CONVERT_TO_EQUATION)) {
-                title = title.replaceAll("/sup ([^/]+)/", "\\$\\^\\{$1\\}\\$");
-                title = title.replaceAll("/sub ([^/]+)/", "\\$_\\{$1\\}\\$");
-                title = title.replaceAll("\\(sup\\)([^(]+)\\(/sup\\)", "\\$\\^\\{$1\\}\\$");
-                title = title.replaceAll("\\(sub\\)([^(]+)\\(/sub\\)", "\\_\\{$1\\}\\$");
+                title = title.replaceAll(SUP1, SUP_EQ_RESULT);
```
koppor: Would it be possible to use constant names, which differ more than one character. `P` vs. `B`? :innocent:

src/main/java/net/sf/jabref/model/database/BibDatabase.java
```diff
@@ -83,8 +83,8 @@ public int getEntryCount() {
     /**
      * Checks if the database contains no entries.
      */
-    public boolean isEmpty() {
-        return entries.size() == 0;
+    public boolean hasNoEntries() {
```
koppor: Why not keeping `isEmpty`, because that is more common than `hasNoEntries`?

koppor:
Other than the micro comments: LGTM :+1:

src/main/java/net/sf/jabref/importer/fetcher/IEEEXploreFetcher.java
oscargus: I could change SUP to SUPER. :-)

src/main/java/net/sf/jabref/importer/fetcher/IEEEXploreFetcher.java
oscargus: Changed it to (already existing) "Could not connect to %0". Waiting with removal until the automatic one is done. ;-)

src/main/java/net/sf/jabref/importer/fetcher/IEEEXploreFetcher.java
```diff
@@ -279,15 +281,15 @@ private BibEntry cleanup(BibEntry entry) {
...
             // Deal with subscripts and superscripts
             if (Globals.prefs.getBoolean(JabRefPreferences.USE_CONVERT_TO_EQUATION)) {
-                title = title.replaceAll("/sup ([^/]+)/", "\\$\\^\\{$1\\}\\$");
-                title = title.replaceAll("/sub ([^/]+)/", "\\$_\\{$1\\}\\$");
-                title = title.replaceAll("\\(sup\\)([^(]+)\\(/sup\\)", "\\$\\^\\{$1\\}\\$");
-                title = title.replaceAll("\\(sub\\)([^(]+)\\(/sub\\)", "\\_\\{$1\\}\\$");
+                title = title.replaceAll(SUPER_DETECTION_1, SUPER_EQ_RESULT);
+                title = title.replaceAll(SUB_DETECTION_1, SUB_EQ_RESULT);
+                title = title.replaceAll(SUPER_DETECTION_2, SUPER_EQ_RESULT);
+                title = title.replaceAll(SUB_DETECTION_2, SUB_EQ_RESULT);
```
simonharrer: For performance reasons, you should create Patterns as static variables and then use them to perform the replace all.

oscargus: Good idea! I'm actually wondering if these replacements (and quite a bit of the other cleanups) are useful any more, after the HTML-parsing has been replaced... Not sure how to fully determine it...

simonharrer:
@koppor I am not sure whether they were micro comments or nano comments. Why would you say that they are micro comments?

PR #986, opened 2016-03-17
Replaced replaceAll with replace

oscargus:
replace is faster, so if no regular expression is needed it is better to use that.

There are a few more places where it can be replaced, but I got confused when it came to escaping...


simonharrer:
:+1:

PR #987, opened 2016-03-17
Refactors database code

simonharrer:
Refactoring of the database SQL stuff. 

- Favor composition over inheritance
- Make a lot of stuff more explicit, e.g., DBStringsPreferences are the values which are stored in the preferences (a subset of the DBStrings basically), and DatabaseType defines the types
- MySQL and PostgreSQL specific code is now in the `MySQL` and the `PostgreSQL` class and can be accessed, both implement the `Database` interface which defines all DB specific code.

There is still a lot to do in that package, but this should be a good start, I think.


src/main/java/net/sf/jabref/gui/BasePanel.java
```diff
@@ -837,31 +838,31 @@ public void update() {
         actions.put(Actions.ADD_FILE_LINK, new AttachFileAction(this));
 
         actions.put(Actions.OPEN_EXTERNAL_FILE, (BaseAction) () ->
-            JabRefExecutorService.INSTANCE.execute((Runnable) () -> {
-                final List<BibEntry> bes = mainTable.getSelectedEntries();
-                if (bes.size() != 1) {
-                    output(Localization.lang("No entries or multiple entries selected."));
-                    return;
-                }
+                JabRefExecutorService.INSTANCE.execute((Runnable) () -> {
```
oscargus: You are using IntelliJ(?), right? I'm wondering if the settings there and Eclipse's are the same? I think I formatted that complete block when I introduced the lambda... But not fully sure.

src/main/java/net/sf/jabref/gui/BasePanel.java
```diff
@@ -97,8 +97,10 @@
 import java.nio.charset.Charset;
 import java.nio.charset.UnsupportedCharsetException;
 import java.util.*;
+import java.util.List;
```
oscargus: Can be removed.

oscargus:
Looks good!

src/main/java/net/sf/jabref/sql/DBStringsPreferences.java
```diff
@@ -0,0 +1,55 @@
...
+    private final String database;
+
+    public DBStringsPreferences(String serverType, String serverHostname, String username, String database) {
+        this.serverType = serverType;
+        this.serverHostname = serverHostname;
+        this.username = username;
+        this.database = database;
+    }
+
+    public String getServerType() {
```
tobiasdiez: I think it is better to change the return type to `DatabaseType` instead of the string value.

src/main/java/net/sf/jabref/sql/DatabaseType.java
```diff
@@ -0,0 +1,40 @@
...
+
+    DatabaseType(String dbType) {
+        this.serverType = dbType;
+    }
+
+    public String getServerType() {
+        return serverType;
+    }
+
+    public static Optional<DatabaseType> build(String serverType) {
```
tobiasdiez: Is the internal string serverType and `build` really necessary? Why is it not enough to use the `valueOf` method?

src/main/java/net/sf/jabref/sql/DatabaseType.java
```diff
@@ -0,0 +1,40 @@
...
+        for(DatabaseType type : values()) {
+            if(type.getServerType().equalsIgnoreCase(serverType)) {
+                return Optional.of(type);
+            }
+        }
+        return Optional.empty();
+    }
+
+    public static List<String> SERVER_TYPES = Arrays
```
tobiasdiez: Return an enumSet here?

src/main/java/net/sf/jabref/sql/DBStringsPreferences.java
simonharrer: Good idea.

src/main/java/net/sf/jabref/sql/DatabaseType.java
simonharrer: In the GUI, I need a list of strings.

src/main/java/net/sf/jabref/sql/DatabaseType.java
simonharrer: valueOf is case sensitive and throws exceptions if it cannot find an entry. Mine returns an optional and handles case insensitive comparison.

src/main/java/net/sf/jabref/gui/BasePanel.java
simonharrer: Hm, compiler requires it. I think this may be related to the * imports, and probably two List types in two * imports requiring the specific import.

src/main/java/net/sf/jabref/gui/BasePanel.java
simonharrer: Yes, I am an IntelliJ addict. Formatting seems to differ. I reverted it.

src/main/java/net/sf/jabref/sql/DatabaseType.java
tobiasdiez: Ok, then I would maybe move this part to the UI and/or rename it to DisplayNames or something like that.

simonharrer: done

src/main/java/net/sf/jabref/sql/database/MySQL.java
```diff
@@ -0,0 +1,129 @@
...
+    @Override
+    public String getReadColumnNamesQuery() {
+        return "SHOW columns FROM entries;";
+    }
+
+    @Override
+    public String getCreateTableSQL(Table table) {
+        switch (table) {
+
+        case JABREF_DATABASE:
```
Siedlerchr: At this point I would suggest better creating a method for each case  
e.g. createhabRefDatabase 
That would better follow the principle of a single purpose method and additionally reduces the length of the method. Furthermore better readability

simonharrer: Hm, to me, the method is still quite simple despite being long in lines as it comes with almost no complexity. Hence, I would not gain that much by splitting it up. From my perspective, I would have the hassle of maintaining 9 methods, which have to be put in the interface, and I cannot iterate over these 9 methods in the caller (which is currently done as I can iterate over the Table enum to call the method for each table).

src/main/java/net/sf/jabref/sql/DBConnectDialog.java
```diff
@@ -37,6 +37,11 @@
  */
 public class DBConnectDialog extends JDialog {
 
+    private static List<String> FORMATTED_NAMES = Arrays
```
tobiasdiez: Make it non-static and move to the constructor since it is only used there as far as I can see?

src/main/java/net/sf/jabref/sql/DBStringsPreferences.java
```diff
@@ -0,0 +1,52 @@
...
+                preferences.get(JabRefPreferences.DB_CONNECT_HOSTNAME),
+                preferences.get(JabRefPreferences.DB_CONNECT_USERNAME),
+                preferences.get(JabRefPreferences.DB_CONNECT_DATABASE));
+    }
+
+    /**
+     * Store these db strings into JabRef preferences.
+     */
+    public void storeToPreferences(JabRefPreferences preferences) {
+        preferences.put(JabRefPreferences.DB_CONNECT_SERVER_TYPE, getServerType().getFormattedName());
```
tobiasdiez: I would suggest to store the enum by name and not by formatted name (the latter might change in the future...although not likely)

src/main/java/net/sf/jabref/sql/DatabaseUtil.java
```diff
@@ -9,9 +9,9 @@
 
 import net.sf.jabref.MetaData;
 
-public class DBImporterExporter {
+public class DatabaseUtil {
```
tobiasdiez: I don't like these static Util classes... couldn't you find a better home for these methods?

tobiasdiez:
In principle LGTM

src/main/java/net/sf/jabref/sql/DBConnectDialog.java
simonharrer: Hm, to me, this is a constant, and therefore should be static (and even made final as I see now).

src/main/java/net/sf/jabref/sql/DatabaseUtil.java
simonharrer: Yes, good point. Previously, these methods I changed to static ones have been in the root class of a three layered inheritence hierarchy. I just moved them out. I think this is subject to another PR someday by someone. 

And I do not have a good idea what to do with them at the moment :)

src/main/java/net/sf/jabref/sql/DBStringsPreferences.java
simonharrer: Yes, y not.

src/main/java/net/sf/jabref/sql/database/MySQL.java
```diff
@@ -0,0 +1,135 @@
...
+    public Connection connectAndEnsureDatabaseExists(DBStrings dbStrings)
+            throws SQLException, IllegalAccessException, ClassNotFoundException, InstantiationException {
+
+        dbStrings.setDbParameters(OPT_ALLOW_MULTI_QUERIES);
+        String url = SQLUtil.createJDBCurl(dbStrings, false);
+
+        Connection conn = connect(url, dbStrings.getDbPreferences().getUsername(), dbStrings.getPassword());
+
+        SQLUtil.processQuery(conn, "CREATE DATABASE IF NOT EXISTS `" + dbStrings.getDbPreferences().getDatabase() + '`');
+        conn.setCatalog(dbStrings.getDbPreferences().getDatabase());
```
Siedlerchr: Just one thing I got in mind while looking through an old PR I created a while ago:
There is a potential resource leak in here:
If the Create Database query fails and throws an SQLException, the connection does not get closed and remains open.
That was  one thing I fixed in my old PR #910 because it was discovered from coverity amalysis.
My solution was to catch the exception and close the conn. Then re throw the ex.
https://github.com/Siedlerchr/jabref/blob/reworkedCoverityFixes/src/main/java/net/sf/jabref/sql/exporter/MySQLExporter.java

simonharrer: Hm, you are talking about managing the connection itself. I am not sure if we do handle the connection resource at all, mostly only the statements and result sets.

stefan-kolb:
Can you merge this in?

simonharrer:
Hm, I am not completely sure this introduces some bugs. Should I wait for directly after the release?

PR #992, opened 2016-03-18
Better layout of DatabasePropertiesDialog

oscargus:
Does not fully solve #958 but at least better...

I do not fully understand why the formatter part doesn't scale horizontally.

### DIFF
--- a/src/main/java/net/sf/jabref/gui/SaveOrderConfigDisplay.java
+++ b/src/main/java/net/sf/jabref/gui/SaveOrderConfigDisplay.java
@@ -92,7 +92,7 @@ public class SaveOrderConfigDisplay {
         saveTerDesc = new JCheckBox(Localization.lang("Descending"));
 
 
-        FormLayout layout = new FormLayout("right:pref, 8dlu, fill:pref, 4dlu, fill:60dlu, 4dlu, left:pref",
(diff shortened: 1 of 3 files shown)


matthiasgeiger:
Uh... isn't @Siedlerchr currently working on this dialog (see #958 and #957)?

Siedlerchr:
Öhm, you didn't need to do that, as I am already working at this ;)
See assigned label.

oscargus:
I know. However, since I've struggled with JGoodies earlier I thought you
could use some input (regarding your question in that issue). Use it as
inspiration. :-)

Siedlerchr:
Yeah, it was the scaling of the Icons to the buttons. I will try later, I think the solution is to putt the Icon Button and the other Buttons in a button bar.

oscargus:
I think I solved that part (only tested on Linux where it looked OK to
start with though).

The key thing I did was to use "fill".

matthiasgeiger:
@Siedlerchr Shall we merge this in and you work with @oscargus' improvements as a basis or do you prefer to create another PR beginning from scratch?

Siedlerchr:
I will manually try to apply the fixes because I changed some more in the
layout (removed  Textfields etc)


2016-03-18 9:59 GMT+01:00 Matthias Geiger <notifications@github.com>:

PR #994, opened 2016-03-18
Replaced some strings

oscargus:
Unified some translations to reduce the number of translations.

Related to #869 : It seems like some encoding changed. I do not fully understand why since I didn't do anything active here. Any ideas which is the better approach? Are they even different?

(There are some obsolete strings which will be removed once #869 is merged.)


src/main/resources/l10n/JabRef_fr.properties
```diff
@@ -1465,63 +1465,63 @@ Show_deprecated_bibtex_fields=Montrer_les_champs_BibTeX_obsol\u00e8tes
...
-Set_priority_to_'%0'_for_%1_entries=Priorité_mise_à_'%0'_pour_%1_entrées
-Toggled_quality_for_%0_entries=Vérification_de_qualité_modifiée_pour_%0_entrées
-Toggled_print_status_for_%0_entries=Statut_d'impression_modifié_pour_%0_entrées
-Toggled_relevance_for_%0_entries=Pertinence_modifiée_pour_%0_entrées
-Set_read_status_to_'%0'_for_%1_entries=Statut_de_lecture_mis_à_'%0'_pour_%1_entrées
-
-Marked_all_%0_selected_entries=%0_entrées_sélectionnées_étiquetées
-Marked_selected_entry=Entrées_sélectionnées_étiquetées
+Table_row_height_padding=Espacement_en_hauteur_des_rang\u00e9es_de_tableau
+Set_rank_to_'%0'_for_%1_entries=Rang_mis_\u00e0_'%0'_pour_%1_entr\u00e9es
```
koppor: Is it possible to keep the plain  UTF-8 characters instead of using `\u....`?

oscargus:
Indeed it is. I have figured out a method for this, so I'll apply it to all files when I rebase.

oscargus:
Now (almost) everything is actually encoded in UTF-8. There are some odd flaws and I had to sort the Chinese translations to sort a few things out. There are probably duplicates for some translations caused by the merging, but as far as I know that doesn't cause any problems. Maybe we should add a check for that later on as well?

tobiasdiez:
LG in my opinion. I'm not sure if we really want to merge this before the release of 3.3 since it is a rather large change.

oscargus:
I see your point, but let me tell you how the change is made (can be good for later reference): 

1. Load the file in Eclipse (I have some plugin which shows the escaped Unicode as actual Unicode, I think it helps).
2. Ctrl-a Ctrl-c
3. Open the same file in NotePad++
4. Ctrl-a Ctrl-v
5. Save in NotePad++

The only potential drawback of not merging now is that there will be more duplicates when rebasing later with files possibly not entirely in UTF-8. Not a major issue.

matthiasgeiger:
There are indeed some files with duplicates... IntelliJ is highlighting them as errors, so its rather easy for me to get rid of them.

I'll do this and push the result in the repository here.

As it is rather easy to check whether the changes work by simply running JabRef in all languages I would merge this in for the 3.3 release.

matthiasgeiger:
Result is [here](https://github.com/JabRef/jabref/compare/oscargus-fewertranslations)... if there is no veto I'll merge this directly into `master` without creating another pull request after lunch :wink:

Siedlerchr:
Looks good!

2016-03-31 9:28 GMT+02:00 Matthias Geiger <notifications@github.com>:

PR #996, opened 2016-03-18
fix entry deletion from database

### LINKED ISSUE #991: Insert entry in an empty database leads to KeyCollisionException
JabRef 3.3dev--snapshot--2016-03-18--master--57963f7 on Windows 10

When I delete an entry at a database and refetch the entry via the DOI to BibTeX converter, the entry is marked as duplicate in the fetch dialog.

Steps to repoduce:
1. Create new database
2. Search for `10.1007/978-3-642-36926-1_34` in the DOI to BibTeX fetcher.
3. In the untitled dialog for adding entries, the arcticle appears
4. Click "OK"
5. Wrong behavior: The checkmark in "Keep" disappears
6. Add the checkmark at "Keep"
7. Click "OK" again
8. Wrong beheavior: nothing shown at the UI
9. Click "Cancel"
10. Open Error console

```
Exception in thread "AWT-EventQueue-0" net.sf.jabref.model.database.KeyCollisionException: ID is already in use, please choose another
	at net.sf.jabref.model.database.BibDatabase.insertEntry(BibDatabase.java:163)
	at net.sf.jabref.gui.ImportInspectionDialog.generateKeys(ImportInspectionDialog.java:489)
```

bartsch-dev:
Fixes #991, a Bug introduced in b3aee00bae8b589aa1252ef182f8b3293d06bd69.

While moving from `Map` to `List` as the data structure for the stored `BibEntries` is has been forgotten to adjust the `removeEntry`-Method.

- [x] Tests green?


src/main/java/net/sf/jabref/model/database/BibDatabase.java
```diff
@@ -173,15 +171,14 @@ public synchronized boolean insertEntry(BibEntry entry) throws KeyCollisionExcep
...
      */
-    public synchronized void removeEntry(BibEntry oldValue) {
-        if (oldValue == null) {
+    public synchronized void removeEntry(BibEntry toBeDeleted) {
+        if (toBeDeleted == null) {
             return;
         }
 
-        entries.remove(oldValue.getId());
-
```
oscargus: Ahh, I just fixed that (but worse, just removed the getId()-part). Will removed that commit from my PR.

CHANGELOG.md
```diff
@@ -73,6 +73,7 @@ to [sourceforge feature requests](https://sourceforge.net/p/jabref/features/) by
 - Fixed [#649](https://github.com/JabRef/jabref/issues/649): Key bindings are now working in the preview panel
 - Fixed [#410](https://github.com/JabRef/jabref/issues/410): Find unlinked files no longer freezes when extracting entry from PDF content
 - Fixed [#936](https://github.com/JabRef/jabref/issues/936): Preview panel is now updated when an entry is cut/deleted
+- Fixed Bug introduced with Commit [b3aee00](https://github.com/JabRef/jabref/commit/b3aee00bae8b589aa1252ef182f8b3293d06bd69): Entry is now correctly removed from BibDatabase
```
boceckts: I think you should create a issue and reference to it as this would be more consistent and compact.

koppor: There is no need for a changelog entry in this case at all: The bug does not appear in a released version. Development versions may be broken. The text is OK for the commit. For a changelog, only the text is OK (without the commit id). However, the commit happend **after** the last release of JabRef and thus did not appear to the end user. If an enduser wondered, he would report an issue. Since no issue has been reported, one does not need to create a "fake" issue.

Nevertheless, it might be that #991 could be fixed with that PR.

koppor:
Is it really impossible to add a test case?

CHANGELOG.md
bartsch-dev: Yes, will fix #991 (just tested it, I overlooked that issue, thanks for mentioning).
I'll adjust the Changelog.

src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
```diff
@@ -0,0 +1,89 @@
...
+        db.insertEntry(entry1);
+    }
+
+    /**
+     * Inserts and removes a {@link BibEntry} and runs several tests accordingly
+     */
+    @Test
+    public void testRemoveEntry() {
+        BibDatabase db = new BibDatabase();
+        assertEquals(Collections.emptyList(), db.getEntries());
```
koppor: Possibly, this "assertion" for the subsequent test should be made as separate test. This test is repeated in `testInsertEntryKeyCollision`. Therefore, I vote for extracting these lines.

src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
```diff
@@ -0,0 +1,89 @@
...
+     * expects a {@link KeyCollisionException}
+     */
+    @Test(expected = KeyCollisionException.class)
+    public void testInsertEntryKeyCollision() {
+        BibDatabase db = new BibDatabase();
+        assertEquals(Collections.emptyList(), db.getEntries());
+
+        BibEntry entry0 = new BibEntry();
+        db.insertEntry(entry0);
+        assertEquals(db.getEntries().size(), 1);
```
koppor: This should be a separate test case so that this test case can really concentrate on the KeyCollisionException.

src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
```diff
@@ -0,0 +1,89 @@
...
+     */
+    @Test
+    public void testRemoveEntry() {
+        BibDatabase db = new BibDatabase();
+        assertEquals(Collections.emptyList(), db.getEntries());
+
+        BibEntry entry0 = new BibEntry();
+        db.insertEntry(entry0);
+        assertEquals(db.getEntries().size(), 1);
+        assertTrue(db.containsEntryWithId(entry0.getId()));
```
koppor: See above :innocent:

bartsch-dev:
Enhanced the tests and added a null-check while inserting an object into the database.

src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
```diff
@@ -0,0 +1,119 @@
...
+
+    /**
+     * Tests the removal of a {@link BibEntry}
+     */
+    @Test
+    public void testInsertRemoveNullEntry(){
+        BibDatabase db = new BibDatabase();
+        assertEquals(Collections.emptyList(), db.getEntries());
+
+        db.insertEntry(null);
```
koppor: Why not two different test cases? One for insert null and one for null removal?

bartsch-dev: Because the Test is very small and tests similar things, I'll extract it.

src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    /**
+     * Tests the removal of a {@link BibEntry}
+     */
+    @Test
+    public void testRemoveEntry() {
+        BibDatabase db = new BibDatabase();
+
+        BibEntry entry0 = new BibEntry();
```
koppor: nitpicking: There is no `entry1` in this test case. Therefore, I propose to name this variable `entry`.

bartsch-dev: done

koppor:
LGTM :+1:

src/main/java/net/sf/jabref/model/database/BibDatabase.java
```diff
@@ -158,30 +156,33 @@ public synchronized BibEntry getEntryByKey(String key) {
      * use Util.createId(...) to make up a unique ID for an entry.
      */
     public synchronized boolean insertEntry(BibEntry entry) throws KeyCollisionException {
+        if (entry == null){
```
tobiasdiez: I think Objects.requireNonNull is more appropriate here.

src/main/java/net/sf/jabref/model/database/BibDatabase.java
```diff
@@ -158,30 +156,33 @@ public synchronized BibEntry getEntryByKey(String key) {
...
         return checkForDuplicateKeyAndAdd(null, entry.getCiteKey());
     }
 
     /**
      * Removes the given entry.
      */
-    public synchronized void removeEntry(BibEntry oldValue) {
-        if (oldValue == null) {
+    public synchronized void removeEntry(BibEntry toBeDeleted) {
+        if (toBeDeleted == null) {
```
tobiasdiez: Also Objects.requireNonNull

src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
```diff
@@ -0,0 +1,125 @@
...
+    @Rule
+    public ExpectedException thrown = ExpectedException.none();
+
+    @Before
+    public void setUp() {
+        Globals.prefs = JabRefPreferences.getInstance(); // set preferences for this test
+    }
+
+    @After
+    public void tearDown() {
```
tobiasdiez: I think there is no need for this.

bartsch-dev: So do I.

src/main/java/net/sf/jabref/model/database/BibDatabase.java
bartsch-dev: Done, didn't knew that trick.

src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
tobiasdiez: I second this comment. One test should check that insertEntry really inserts the entry and a second test to check that containsEntryWithId works, finally a third test with 2 items.

tobiasdiez: Use the BibtexEntryAssert.assertEquals to check for equality.

src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
```diff
@@ -0,0 +1,125 @@
...
+        assertTrue(db.containsEntryWithId(entry0.getId()));
+        assertTrue(db.containsEntryWithId(entry1.getId()));
+    }
+
+
+    /**
+     * Insert two instances of {@link BibEntry} with the same ID into a {@link BibDatabase},
+     * expects a {@link KeyCollisionException}
+     */
+    @Test
```
tobiasdiez: I would prefer a `expected` argument to the test instead of the thrown.expect construction.

tobiasdiez:
Thanks for fixing the bug I had introduced. 
The changes look good.

Please try to adhere to the convention for tests outlined in the wiki https://github.com/JabRef/jabref/wiki/Code-Howtos#test-cases, especially the naming of the tests and only one test per test method.

src/main/java/net/sf/jabref/model/database/BibDatabase.java
```diff
@@ -158,30 +156,29 @@ public synchronized BibEntry getEntryByKey(String key) {
...
 
         entries.add(entry);
-
         fireDatabaseChanged(new DatabaseChangeEvent(this, DatabaseChangeEvent.ChangeType.ADDED_ENTRY, entry));
-
         return checkForDuplicateKeyAndAdd(null, entry.getCiteKey());
     }
 
     /**
      * Removes the given entry.
```
tobiasdiez: Please add in the javadoc comment that the entry is removed based on the id.

src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
bartsch-dev: The advantage of the `thrown.expect? construct is, that the test expects the exception after the statement.
When using `@Test(expected = ... )` the exception can occur anywhere in the Method.

src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
koppor: I thought so too, but I'm unsure whether JUnit really cleans up everything. Is it possible that `Globals.prefs` survives each test run?

bartsch-dev: @koppor is right.
I just printed `Globals.prefs` right before it is set in the `setUp()`-Method. Before the first test it is `null`, but at any test after that is alredy set.
I undo it.

src/main/java/net/sf/jabref/model/database/BibDatabase.java
bartsch-dev: done

src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
tobiasdiez: Yes @koppor is right, it survives the test. But this is no reason to worry since you overwrite it anyway in the setup method.

bartsch-dev: Yeah, but as I just noticed that doesn't matter.
`JabRefPreferences` is a Singleton; every test gets the same instance, every change in the prefs are carried to the next test.

src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
```diff
@@ -0,0 +1,124 @@
...
+    public void containsEntryId() {
+        BibDatabase db = new BibDatabase();
+        BibEntry entry = new BibEntry();
+        assertFalse(db.containsEntryWithId(entry.getId()));
+        db.insertEntry(entry);
+        assertTrue(db.containsEntryWithId(entry.getId()));
+    }
+
+    @Test
+    public void insertEntry_doubleValue_exceptionThrown() {
```
tobiasdiez: use names without underscore, for example `insertEntryWithSameIdThrowsException`

src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
tobiasdiez: Well, in the end it is not that important. By the way, which of these methods actually needs an initialized preference class, i.e. which test fails if you remove the setup method

src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
```diff
@@ -0,0 +1,124 @@
...
+        assertEquals(db.getEntries().size(), 1);
+        BibtexEntryAssert.assertEquals(entry, db.getEntries().get(0));
+    }
+
+    /**
+     * Tests {@link BibDatabase#containsEntryWithId(String)}
+     */
+    @Test
+    public void containsEntryId() {
+        BibDatabase db = new BibDatabase();
```
tobiasdiez: do not abbreviate database

tobiasdiez:
After some minor naming improvements, this can be merged in in my opinion.

src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
bartsch-dev: Only the test-method for resolveStrings (which was already present) needs the preferences set.

PR #997, opened 2016-03-18
Coverity and Codacy cleanups

oscargus:
Actually found a bug in the ChangeScanner leading to NPEs while trying to remove a Codacy warning (not sure if that part succeeded though).

Got rid of unused imports (don't you have your editors configured to do that automatically :question: ).

Got rid of maybe the last unwanted println.

### DIFF
--- a/src/main/java/net/sf/jabref/model/database/EntrySorter.java
+++ b/src/main/java/net/sf/jabref/model/database/EntrySorter.java
@@ -90,8 +90,10 @@ public class EntrySorter implements DatabaseChangeListener {
             int pos;
             switch (e.getType()) {
             case ADDED_ENTRY:
-                pos = -Collections.binarySearch(set, e.getEntry(), comp) - 1;
-                set.add(pos, e.getEntry());
+                pos = Collections.binarySearch(set, e.getEntry(), comp);
+                if (pos < 0) {
+                    set.add(0, e.getEntry());
+                }
                 break;
             case REMOVED_ENTRY:
                 set.remove(e.getEntry());
@@ -101,8 +103,7 @@ public class EntrySorter implements DatabaseChangeListener {
                 pos = Collections.binarySearch(set, e.getEntry(), comp);
                 int posOld = set.indexOf(e.getEntry());
                 if (pos < 0) {
-                    set.remove(posOld);
-                    set.add(-pos - 1, e.getEntry());
+                    set.add(0, e.getEntry());
                 }
                 break;
             }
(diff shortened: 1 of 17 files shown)


src/main/java/net/sf/jabref/model/database/EntrySorter.java
```diff
@@ -90,8 +90,10 @@ public void databaseChanged(DatabaseChangeEvent e) {
             int pos;
             switch (e.getType()) {
             case ADDED_ENTRY:
-                pos = -Collections.binarySearch(set, e.getEntry(), comp) - 1;
-                set.add(pos, e.getEntry());
+                pos = Collections.binarySearch(set, e.getEntry(), comp);
```
tobiasdiez: I don't understand this change. Previously, the coded added always an entry, now only if it would be position < 0 ?

oscargus: This relates to a rather strange thing that I do not really understand. When I played around with changing the file in a separate editor to trigger change detection, I found out that lead to the fix in EntryChange. After fixing that, the detection code behaved differently (there's a potential reason though). Before, removing a field leads to a changed entry. After, removing a field leads to a removed and an added entry. I *think* it depends on which field is removed since duplicate detection is used to determine if the entry is similar enough to be considered as modified or two completely different entries.

To answer your question, I think this makes more sense. If an ADDED_ENTRY event is triggered, it is checked if the entry already exists. If not, it is added. The problem with the previous code was that it only worked if the entry didn't exist (otherwise pos would be negative and an exception was thrown). I guess a fundamental reason behind the problem here is that similarity is checked with duplicate detection in the other place and a comparator here. Note that the only time the entry is not added is when it already exists...

tobiasdiez: Thank you for your explanation. Now it makes kind of sense. 
A minimal test just would call databaseChanged. A more elaborate way is to register the EntrySorter to a test database and then remove or add items there. Both tests should be more or less straightforward.

oscargus: Ahh, I was a bit wrong here. The old code didn't work and the current code at least doesn't throw IndexOutOfBoundExceptions but doesn't behave as the old code did (when it worked)...

oscargus:
The current DatabaseChangeEvent system can be improved. At the moment only ADDED_ENTRY and REMOVED_ENTRY events are fired, not CHANGED_ENTRY or CHANGING_ENTRY. It also seems like an event is fired multiple times(?). For example, adding an entry in an external editor triggers at least three events...

PR #1002, opened 2016-03-20
Increase default window size to show complete toolbar

stefan-kolb:
Is 1080px too wide as default?


koppor:
The width 1024px is near and would match a low resolution. I think, however, that nearly no one is using that low screen resolution anymore. I assume that 1080 is the minimum width required to show the toolbar. Thus, this change is OK.

oscargus:
If 1024 is enough, I would go with that. Unlikely, but could be people with older/smaller laptops that would appreciate it. If not, go ahead and merge it.

stefan-kolb:
1024px is not enough. 1080px is needed which is why I chose it (at least for my tests, correct me if I'm wrong).

oscargus:
I can use 978 on my Windows 7 laptop. The buttons do move slightly closer, but all are visible. 

Interesting thing though: the external application button becomes wider when making the window more narrow...

stefan-kolb:
Hm, I tested it on Ubuntu...

oscargus:
Yes, it is probably OS dependent... Not sure what the best approach is here... More than 840 for sure though...

oscargus:
I'd say try it. If someone complains pick another value. Still, it will only be used for the first start...

stefan-kolb:
@JabRef/developers Can you guys check on different OS and give your minimum numbers?

stefan-kolb:
@JabRef/developers @chochreiner Can you guys check on different OS and give your minimum numbers? Can be shown `Options -< Preferences -> Show Preferences`

oscargus:
On the CentOS machine in my office 1050 is the minimum and the external application button-thing doesn't happen. (I have a 1050 x 1680 screen and it is OK to run it maximized there, which is a good thing in itself.)

stefan-kolb:
Windows: 
- WindowsLAF: ~820px
- MetalLAF: ~890px
- Plastic: ~890px

mlep:
Linux Mint LMDE: 1024 is fine. With lower values (down to 800), the right part of the toolbar is missing (because icons are big).

mlep:
BTW: The 2015 survey outlined an "awful display with HiDPI screens". Anyone able to test it?

oscargus:
I have a 2560 x 1440 (which is maybe not HiDPI anymore since it is a 27") and I think that with the current, larger, icons it is OK. Of course, being able to change the size of the icons would be nice, especially when the smaller icons are used in buttons.

Siedlerchr:
Hm, I don't see any big problems:
1920x1080 on 24" with Win7
INTEGER	mainWindowSizeX	865	| 840
Still seeing all Icon in the Toolbar, that is the minimum size. If I resize it further, the Donate button gets lost.
toolbarjabref

chochreiner:
for OSX I need ~1111px to list all icons in the toolbar on a 13,3-inch (2560 x 1600) display (native UI)

stefan-kolb:
Ouch, this is huge...

chochreiner:
maybe it is possible to mark some activities as optional and let the users design their custom toolbars  or at least hide not needed activities

simonharrer:
We proposed to use 1024 x 768 according to http://www.w3schools.com/browsers/browsers_display.asp which should cover every display size.

PR #1003, opened 2016-03-20
Used Optional for the getExternalFileType-methods and fixed an issue …

oscargus:
Better to avoid null assignments and null checks...

### DIFF
--- a/src/main/java/net/sf/jabref/external/DroppedFileHandler.java
+++ b/src/main/java/net/sf/jabref/external/DroppedFileHandler.java
@@ -116,8 +116,7 @@ public class DroppedFileHandler {
      * @param mainTable The MainTable the file was dragged to.
      * @param dropRow   The row where the file was dropped.
      */
-    public void handleDroppedfile(String fileName, Optional<ExternalFileType> fileType, MainTable mainTable,
-            int dropRow) {
+    public void handleDroppedfile(String fileName, ExternalFileType fileType, MainTable mainTable, int dropRow) {
 
         BibEntry entry = mainTable.getEntryAt(dropRow);
         handleDroppedfile(fileName, fileType, entry);
@@ -128,8 +127,8 @@ public class DroppedFileHandler {
      * @param fileType  The FileType associated with the file.
      * @param entry     The target entry for the drop.
      */
-    public void handleDroppedfile(String fileName, Optional<ExternalFileType> fileType, BibEntry entry) {
-        NamedCompound edits = new NamedCompound(Localization.lang("Drop %0", fileType.get().getExtension()));
+    public void handleDroppedfile(String fileName, ExternalFileType fileType, BibEntry entry) {
+        NamedCompound edits = new NamedCompound(Localization.lang("Drop %0", fileType.getExtension()));
 
         if (tryXmpImport(fileName, fileType, edits)) {
             edits.end();
@@ -178,8 +177,14 @@ public class DroppedFileHandler {
     }
 
     public void linkPdfToEntry(String fileName, BibEntry entry) {
-        Optional<ExternalFileType> fileType = ExternalFileTypes.getInstance().getExternalFileTypeByExt("pdf");
+        Optional<ExternalFileType> optFileType = ExternalFileTypes.getInstance().getExternalFileTypeByExt("pdf");
 
+        if (!optFileType.isPresent()) {
+            LOGGER.warn("No file type with extension 'pdf' registered.");
+            return;
+        }
+
+        ExternalFileType fileType = optFileType.get();
         // Show dialog
         if (!showLinkMoveCopyRenameDialog(fileName, fileType, entry, panel.database())) {
             return;
@@ -192,7 +197,7 @@ public class DroppedFileHandler {
 
         boolean success = true;
         String destFilename;
-        NamedCompound edits = new NamedCompound(Localization.lang("Drop %0", fileType.get().getExtension()));
+        NamedCompound edits = new NamedCompound(Localization.lang("Drop %0", fileType.getExtension()));
 
         if (linkInPlace.isSelected()) {
             destFilename = FileUtil.shortenFileName(new File(fileName), panel.getBibDatabaseContext().getMetaData().getFileDirectory(Globals.FILE_FIELD)).toString();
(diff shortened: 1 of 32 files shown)


src/main/java/net/sf/jabref/external/DroppedFileHandler.java
```diff
@@ -129,8 +130,8 @@ public void handleDroppedfile(String fileName, ExternalFileType fileType, MainTa
      *                  to a local temporary file.
      * @param entry     The target entry for the drop.
      */
-    public void handleDroppedfile(String fileName, ExternalFileType fileType, BibEntry entry) {
-        NamedCompound edits = new NamedCompound(Localization.lang("Drop %0", fileType.getExtension()));
+    public void handleDroppedfile(String fileName, Optional<ExternalFileType> fileType, BibEntry entry) {
```
tobiasdiez: I think this shouldn't accept an optional, or if it accepts one that there should be some safeguard instead of a simple `get` call.

src/main/java/net/sf/jabref/external/DroppedFileHandler.java
```diff
@@ -218,9 +219,9 @@ public void linkPdfToEntry(String fileName, BibEntry entry) {
 
     // Done by MrDlib
 
-    private boolean tryXmpImport(String fileName, ExternalFileType fileType, NamedCompound edits) {
+    private boolean tryXmpImport(String fileName, Optional<ExternalFileType> fileType, NamedCompound edits) {
```
tobiasdiez: same as above.

src/main/java/net/sf/jabref/external/DroppedFileHandler.java
```diff
@@ -305,7 +306,7 @@ private boolean tryXmpImport(String fileName, ExternalFileType fileType, NamedCo
     //
     // @return true if user pushed "OK", false otherwise
     //
-    private boolean showLinkMoveCopyRenameDialog(String linkFileName, ExternalFileType fileType,
+    private boolean showLinkMoveCopyRenameDialog(String linkFileName, Optional<ExternalFileType> fileType,
```
tobiasdiez: same

src/main/java/net/sf/jabref/external/ExternalFileTypes.java
```diff
@@ -184,7 +170,7 @@ public ExternalFileType getExternalFileTypeForName(String filename) {
                 foundType = type;
```
tobiasdiez: Follow the pattern of the other methods here and return Optonial.of(...)

src/main/java/net/sf/jabref/util/Util.java
```diff
@@ -498,12 +498,12 @@ public void run() {
                         }
                         if (!alreadyHas) {
                             foundAny = true;
-                            ExternalFileType type;
+                            Optional<ExternalFileType> type;
                             Optional<String> extension = FileUtil.getFileExtension(f);
                             if (extension.isPresent()) {
                                 type = ExternalFileTypes.getInstance().getExternalFileTypeByExt(extension.get());
```
tobiasdiez: Would it make sense to use .orElse(UnknownExternalFileType) here?

src/main/java/net/sf/jabref/external/DroppedFileHandler.java
oscargus: I'll look more into it, but there is not less checking now. With the
assumption, where checking is not explicit, that if it worked before it
will work as well now.

src/main/java/net/sf/jabref/external/ExternalFileTypes.java
oscargus: Hmm, I must have missed that. (Or it was a method only called once with a constant input, do not remember fully.)

src/main/java/net/sf/jabref/util/Util.java
oscargus: I just tried to copy the current behavior. One could also return
UnknownExternalFileType instead of empty() in the case where it is not
found. This is done in one (?) method, but I wasn't sure enough of the
exact mechanisms to do this right now. Drawback is maybe that one will need
to check instance type rather than is present, which somehow voids the use
of Optional...
Den 20 mar 2016 19:14 skrev "Tobias Diez" <notifications@github.com>:

tobiasdiez:
I really like that you removed all those null's. What I wondered after reading the last change in `Util`: can all these null return values (and now Optional.empty) be replaced by `UnknownExternalFileType`?

oscargus:
Yes, that is a relevant question. I do not fully understand the workings, but it seems to make sense and I pondered around that for a while.

src/main/java/net/sf/jabref/external/DroppedFileHandler.java
tobiasdiez: Yes, for the moment it is not a problem. However, the method is public and so in the future somebody might pass `Optional.empty` which fails immediately.

oscargus: Sure, in the same way it would fail if someone passed null.

src/main/java/net/sf/jabref/external/ExternalFileTypes.java
oscargus: No. First, I missed the point of your comment, got confused by the shaded text stating that ExternalFileType was returned. Here there's a mechanism (unknown why), which honours the longest extension string. Therefore, all types must be checked.

src/main/java/net/sf/jabref/external/DroppedFileHandler.java
oscargus: Won't happen with added check in handleDroppedfile

src/main/java/net/sf/jabref/external/DroppedFileHandler.java
oscargus: Only called with an ExternalFileType obtained from hardcoded "pdf", but I added a test that we have a valid type.

oscargus:
I think that the current logic is that when we want to show a table or get an icon, UnknownExternalFileType is used. If not, Optional.empty()/null.

src/main/java/net/sf/jabref/external/DroppedFileHandler.java
tobiasdiez: I still don't get why an optional argument is necessary here instead of a simple ExternalFileType (which has to be non-null). The method apparently needs the fileType, so the argument is not optional...

oscargus: It just seemed to make sense to use Optional as default instead of using Optional.of and get() just to pass ExternalFileType around (since all(?) ExternalFileTypes are obtained as Optionals now). For these two methods one may just as well do the get() in the previous method (since the ExternalFileType is obtained from one of the ExternalFileTypes methods, returning an Optional, which is already checked for content). Not sure if it matters.

oscargus:
I looked over the code again and reverted some of the Optional arguments. So now, instead of passing an optional and then at the end realising it is empty the data is never passed in the first place. Major impact on the DroppedFile-stuff which is almost back to the initial state.

Issue #1001, opened 2016-03-20
error when exporting to endote and ris format


csucat:
* JabRef version (available in the About box):JabRef_windows_3_3dev--snapshot--2016-03-18--master--70a65a6
* Operating system and version: windows 8.1 enterprise edition
* Steps to reproduce:
  1. open and bibtex file
  2. export(choose the endote or ris format)
  3. the program will tell you that there is an error "could not save the file, java.lang.NullPointerException"

Siedlerchr:
Confirmed in latest development version:
Seems to have sth to do with the PR #982 
I just looked into it, it seems to only happen when no Entries are selected.
However, selecting all entries and then clicking export works.

Full error:

> WARN  net.sf.jabref.exporter.ExportFormats - Problem exporting
> java.lang.NullPointerException
> 	at java.util.ArrayList.addAll(Unknown Source) ~[?:1.8.0_74]
> 	at net.sf.jabref.exporter.BibDatabaseWriter.getSortedEntries(BibDatabaseWriter.java:97) ~[bin/:?]
> 	at net.sf.jabref.exporter.ExportFormat.performExport(ExportFormat.java:230) ~[bin/:?]
> 	at net.sf.jabref.exporter.ExportFormats$1ExportAction$1.run(ExportFormats.java:231) [bin/:?]
> 	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[?:1.8.0_74]
> 	at sun.reflect.NativeMethodAccessorImpl.invoke(Unknown Source) ~[?:1.8.0_74]
> 	at sun.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source) ~[?:1.8.0_74]
> 	at java.lang.reflect.Method.invoke(Unknown Source) ~[?:1.8.0_74]
> 	at spin.Invocation.evaluate(Invocation.java:175) [spin.jar:?]
> 	at spin.off.SpinOffEvaluator$1.run(SpinOffEvaluator.java:108) [spin.jar:?]
> 	at java.lang.Thread.run(Unknown Source) [?:1.8.0_74]
>

oscargus:
Could you try the latest master (available in 20 minutes or so)? I had to push directly to the master as I forgot to create a separate branch...

oscargus:
And, yes, it was only when no entries were selected. I also fixed another problem where the complete database was exported if no MetaData was present, even when entries where selected.

PR #1005, opened 2016-03-20
Disabled NPE test

oscargus:
Disabled the dubious test. 

(The roundtrip test doesn't pass on my computer although I can not tell any difference between the outputs... Let's see what happens here...)


tobiasdiez:
What was the problem with this test?

oscargus:
See comment in the commit that closed #1001 (cannot easily link it from the phone). Basically that in an ExportFormat passing null as a list of BibEntry means that all entries should be exported.

PR #1006, opened 2016-03-20
Reworked and checked the constants in GUIGlobals

oscargus:
There were a few unused constants in GUIGlobals and a few which shouldn't be there.


tobiasdiez:
LGTM

PR #1007, opened 2016-03-20
Fixed problem with whitespaces in file paths for tests

Siedlerchr:
After I relocated my workspace, I notice some problems with white spaces in file paths. The whitespaces got encoded as %20 and failed on Windows.
Problem was Url.toFile()
`Paths.get(url.toUri()).toFile()` is now the correct way to handle Files from URLs.
See  http://stackoverflow.com/questions/6164448/convert-url-to-normal-windows-filename-java

- [x] Tests green?


tobiasdiez:
LGTM :+1:.
This should also be added to the wiki.

Siedlerchr:
I just described it in the wiki:
https://github.com/JabRef/jabref/wiki/Code-Howtos#loading-files-from-resources

PR #1008, opened 2016-03-21
Used PositionWindow for PreambleEditor dialog size

oscargus:
Now the preamble editor opens with the same size and position as last time. (Plus that some unused preferences and GUIGlobals variables were removed.)


matthiasgeiger:
:+1: LGTM

PR #1009, opened 2016-03-21
Added enums

oscargus:
Used enums instead of explicit ints for the mode of BasePanel, how to render cells in MainTable, and mode and result for DuplicateResolverDialog.


matthiasgeiger:
:+1: LGTM

PR #1010, opened 2016-03-21
Added initial tests for ExportFormat

oscargus:
Really trivial and early test, but could maybe avoid bugs like #1001 to happen (and form a foundation to extend on).

### DIFF
--- a/src/test/java/net/sf/jabref/exporter/ExportFormatTest.java
+++ b/src/test/java/net/sf/jabref/exporter/ExportFormatTest.java
@@ -38,41 +38,60 @@ public class ExportFormatTest {
     }
 
     @Test
-    public void testExportingEmptyDatabaseLayoutBasedFormat() throws Exception {
+    public void testExportingEmptyDatabaseYieldsEmptyFile() throws Exception {
         BibDatabase database = new BibDatabase();
-        Map<String, IExportFormat> exportFormats = ExportFormats.getExportFormats();
-        IExportFormat exportFormat = exportFormats.get("ris");
-        File tmpFile = testFolder.newFile();
-        tmpFile.deleteOnExit();
-        String filename = tmpFile.getCanonicalPath();
-        List<BibEntry> entries = Collections.emptyList();
-        Charset charset = Charsets.UTF_8;
-        MetaData metaData = new MetaData();
-        exportFormat.performExport(database, metaData, filename, charset, entries);
-        assertTrue(tmpFile.exists());
-        try (FileReader fileReader = new FileReader(tmpFile)) {
-            char[] buffer = new char[512];
-            assertEquals(-1, fileReader.read(buffer)); // Empty file
+        for (Map.Entry<String, IExportFormat> exportFormats : ExportFormats.getExportFormats().entrySet()) {
+            IExportFormat exportFormat = exportFormats.getValue();
+            File tmpFile = testFolder.newFile();
+            tmpFile.deleteOnExit();
+            String filename = tmpFile.getCanonicalPath();
+            List<BibEntry> entries = Collections.emptyList();
+            Charset charset = Charsets.UTF_8;
+            MetaData metaData = new MetaData();
+            exportFormat.performExport(database, metaData, filename, charset, entries);
+            assertTrue(exportFormats.getKey() + " failed -- no file", tmpFile.exists());
+            try (FileReader fileReader = new FileReader(tmpFile)) {
+                char[] buffer = new char[512];
+                assertEquals(exportFormats.getKey() + " failed -- non-empty file", -1, fileReader.read(buffer)); // Empty file
+            }
         }
     }
 
     @Test
-    public void testExportingEmptyDatabaseClassBasedFormat() throws Exception {
-        BibDatabase database = new BibDatabase();
-        Map<String, IExportFormat> exportFormats = ExportFormats.getExportFormats();
-        IExportFormat exportFormat = exportFormats.get("oocalc");
-        File tmpFile = File.createTempFile("exporttest", "oocalc");
-        tmpFile.deleteOnExit();
-        String filename = tmpFile.getCanonicalPath();
+    public void testExportingNullDatabaseThrowsNPE() throws Exception {
         List<BibEntry> entries = Collections.emptyList();
         Charset charset = Charsets.UTF_8;
         MetaData metaData = new MetaData();
-        exportFormat.performExport(database, metaData, filename, charset, entries);
-        assertTrue(tmpFile.exists());
-        try (FileReader fileReader = new FileReader(tmpFile)) {
-            char[] buffer = new char[512];
-            assertEquals(-1, fileReader.read(buffer)); // Empty file
+        for (Map.Entry<String, IExportFormat> exportFormats : ExportFormats.getExportFormats().entrySet()) {
+            try {
+                IExportFormat exportFormat = exportFormats.getValue();
+                File tmpFile = testFolder.newFile();
+                tmpFile.deleteOnExit();
+                String filename = tmpFile.getCanonicalPath();
+                exportFormat.performExport(null, metaData, filename, charset, entries);
+                fail();
+            } catch (NullPointerException expected) {
+                // Ignore -- this is what we want!
+            }
         }
     }
 
+    @Test
+    public void testExportingNullEntriesThrowsNPE() throws Exception {
+        BibDatabase database = new BibDatabase();
+        Charset charset = Charsets.UTF_8;
+        MetaData metaData = new MetaData();
+        for (Map.Entry<String, IExportFormat> exportFormats : ExportFormats.getExportFormats().entrySet()) {
+            try {
+                IExportFormat exportFormat = exportFormats.getValue();
+                File tmpFile = testFolder.newFile();
+                tmpFile.deleteOnExit();
+                String filename = tmpFile.getCanonicalPath();
+                exportFormat.performExport(database, metaData, filename, charset, null);
+                fail();
+            } catch (NullPointerException expected) {
+                // Ignore -- this is what we want!
+            }
+        }
+    }
 }
--- a/src/test/java/net/sf/jabref/exporter/ExportFormatTest.java
+++ b/src/test/java/net/sf/jabref/exporter/ExportFormatTest.java
@@ -1,0 +1,69 @@
+package net.sf.jabref.exporter;
+
+import java.io.File;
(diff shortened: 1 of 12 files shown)


src/test/java/net/sf/jabref/exporter/ExportFormatTest.java
```diff
@@ -0,0 +1,69 @@
...
+        try {
+            File tmpFile = File.createTempFile("exporttest", "html");
+            tmpFile.deleteOnExit();
+            String filename = tmpFile.getCanonicalPath();
+            List<BibEntry> entries = Collections.emptyList();
+            Charset charset = Charsets.UTF_8;
+            MetaData metaData = new MetaData();
+            exportFormat.performExport(db, metaData, filename, charset, entries);
+        } catch (IOException e) {
+            Assert.fail("Exception caught: " + e.toString() + e.getMessage());
```
lenhard: It would probably be more readable code if you just skip the try-catch here. The tests will fail anyway if the method is terminated by an exception and we will get all the exception info we need. There is no need to explicitly build an exception string as you do here. 

Some code quality tools might unnecessarily complain about missing assertions, though...

src/test/java/net/sf/jabref/exporter/ExportFormatTest.java
```diff
@@ -0,0 +1,69 @@
...
+        try {
+            File tmpFile = File.createTempFile("exporttest", "oocalc");
+            tmpFile.deleteOnExit();
+            String filename = tmpFile.getCanonicalPath();
+            List<BibEntry> entries = Collections.emptyList();
+            Charset charset = Charsets.UTF_8;
+            MetaData metaData = new MetaData();
+            exportFormat.performExport(db, metaData, filename, charset, entries);
+        } catch (IOException e) {
+            Assert.fail("Exception caught: " + e.toString() + e.getMessage());
```
lenhard: Also here the try-catch can be omitted from my point of view.

src/test/java/net/sf/jabref/exporter/ExportFormatTest.java
oscargus: Makes sense. I guess one can figure out some sort of assertion, such as the
generated file actually existing...

src/test/java/net/sf/jabref/exporter/ExportFormatTest.java
```diff
@@ -0,0 +1,69 @@
...
+    public void setUp() {
+        Globals.prefs = JabRefPreferences.getInstance();
+        ExportFormats.initAllExports();
+        Globals.journalAbbreviationLoader = new JournalAbbreviationLoader(Globals.prefs);
+
+    }
+
+    @Test
+    public void testExportingEmptyDatabaseLayoutBasedFormat() throws Exception {
+        BibDatabase db = new BibDatabase();
```
tobiasdiez: Not sure how strict we are, but `db` is an abbreviation.

src/test/java/net/sf/jabref/exporter/ExportFormatTest.java
```diff
@@ -0,0 +1,69 @@
...
+        Map<String, IExportFormat> exportFormats = ExportFormats.getExportFormats();
+        IExportFormat exportFormat = exportFormats.get("html");
+        try {
+            File tmpFile = File.createTempFile("exporttest", "html");
+            tmpFile.deleteOnExit();
```
tobiasdiez: There is a slightly more preferable way to deal with temporary files in form of a rule, see the wiki.

src/test/java/net/sf/jabref/exporter/ExportFormatTest.java
oscargus: I probably copies that line from the exactly same code that was later
commented on. I'll fix it. :-)
Den 21 mar 2016 18:03 skrev "Tobias Diez" <notifications@github.com>:

tobiasdiez:
Mhh....I am missing an assertion in these tests, i.e. reading the exported file and checking that it is empty.

src/test/java/net/sf/jabref/exporter/ExportFormatTest.java
Siedlerchr: @lenhard  That should be a general rule in my opinion, not to catch Exceptions ins Tests.

oscargus:
I have changed the code based on all(?) comments. Major implications:

* Now, all exporters returns an empty file for an empty set of entries and are tested for this. Earlier, some heading information was generated
* All exporters are now tested that they throw an NPE with a null database or null entries list
* The change that fixes #1001 is in practice reverted and solved such that the ExportFormat calls add all entries instead of null for all entries
* The test disabled in #1005 is active again
* I had to break the new guideline in #1012 to be able to test all exporters for NPEs in a simple way, if there's a better way please let me know

(This shows that adding a quite crappy initial test can lead to good things. :-))

oscargus:
As the export is now not performed for empty databases (entries lists), as a side effect, the coverage is more accurate now as well. Earlier it increased quite a bit, but no actual testing was done (well, apart from that the code executed without any errors).

src/test/java/net/sf/jabref/exporter/ExportFormatTest.java
```diff
@@ -0,0 +1,97 @@
...
+        Globals.prefs = JabRefPreferences.getInstance();
+        ExportFormats.initAllExports();
+        Globals.journalAbbreviationLoader = new JournalAbbreviationLoader(Globals.prefs);
+
+    }
+
+    @Test
+    public void testExportingEmptyDatabaseYieldsEmptyFile() throws Exception {
+        BibDatabase database = new BibDatabase();
+        for (Map.Entry<String, IExportFormat> exportFormats : ExportFormats.getExportFormats().entrySet()) {
```
tobiasdiez: You can use parametrized tests instead of the for loop. Then also the the string argument to assertEquals is no longer needed.

src/test/java/net/sf/jabref/exporter/ExportFormatTest.java
```diff
@@ -0,0 +1,97 @@
...
+            IExportFormat exportFormat = exportFormats.getValue();
+            File tmpFile = testFolder.newFile();
+            tmpFile.deleteOnExit();
```
tobiasdiez: Not necessary. Will be deleted after the test.

src/test/java/net/sf/jabref/exporter/ExportFormatTest.java
```diff
@@ -0,0 +1,97 @@
...
+        MetaData metaData = new MetaData();
+        for (Map.Entry<String, IExportFormat> exportFormats : ExportFormats.getExportFormats().entrySet()) {
+            try {
+                IExportFormat exportFormat = exportFormats.getValue();
+                File tmpFile = testFolder.newFile();
+                tmpFile.deleteOnExit();
+                String filename = tmpFile.getCanonicalPath();
+                exportFormat.performExport(null, metaData, filename, charset, entries);
+                fail();
+            } catch (NullPointerException expected) {
```
tobiasdiez: with parametized tests a simple @Test(expected = ...) should work

tobiasdiez:
This looks way better then the initial PR. Good job!

src/test/java/net/sf/jabref/exporter/ExportFormatTest.java
oscargus: Any good pointer/example? (It seemed like it should be possible somehow...)

src/test/java/net/sf/jabref/exporter/ExportFormatTest.java
oscargus: I sort of assumed it, but better safe than sorry. :-) Will remove.

src/test/java/net/sf/jabref/exporter/ExportFormatTest.java
tobiasdiez: https://github.com/junit-team/junit4/wiki/Parameterized-tests and the tests in FormatterTests

src/test/java/net/sf/jabref/exporter/ExportFormatTest.java
Siedlerchr: @Test(expected = NullPointerException.class)
Have a look at the CleanUpWorkerTest

oscargus:
Not obvious that Parameterized is executed before BeforeClass, but once I figured that out...

src/test/java/net/sf/jabref/exporter/ExportFormatTest.java
```diff
@@ -23,75 +24,68 @@
...
+
     @Rule
     public TemporaryFolder testFolder = new TemporaryFolder();
 
 
-    @Before
-    public void setUp() {
-        Globals.prefs = JabRefPreferences.getInstance();
-        ExportFormats.initAllExports();
+    @BeforeClass
```
tobiasdiez: I think before is preferable to beforeclass since then every test gets a fresh database and the potentially modifications of the previous test do not influence the next test.

oscargus: OK! I do not fully understand the difference and usually use Before. Now I just mimicked FormatterTest to see if it helped (but the problem was not related to that).

tobiasdiez:
LGTM

src/test/java/net/sf/jabref/exporter/ExportFormatTest.java
tobiasdiez: http://stackoverflow.com/questions/20295578/difference-between-before-and-beforeclass

oscargus: And regarding providing an actual database, the current method is not the best approach (with statics). However, it seems like the tests needs to be rather specific anyway for the different export formats so maybe they can't be parameterized in a good way anyhow.

tobiasdiez: Yes, I second this. Every export format should have its own test class, which tests the actually export.

oscargus: Btw, is there a way to add a test which is not included in the coverage check? It would make sense to try to export a non-empty database for every export format just to make sure that no exceptions are thrown, but probably one wouldn't like to include these test in the coverage report...

oscargus:
I think this is probably as far as we get right now and that the code is quite OK since it is rather general and only tests what is actually tested (or how to formulate it...).

PR #1016, opened 2016-03-22
Added info about current font size when changing it using key bindings

oscargus:
Shows the new font size in the status bar after changing it with increase/decrease table font size.


tobiasdiez:
LGTM :+1:

matthiasgeiger:
:+1:

PR #1020, opened 2016-03-22
Simplified translations in SynchronizeFileField

oscargus:
See koppor/Jabref#118 for background. Basically, now that only the file field is supported for linking to files, there's no need for generic translations strings.

I'm not sure that all the translations are grammatically correct, I basically replaced the translation of "file" where previously the translation of "file" was.


matthiasgeiger:
:+1:

Can you rebase and merge this in?

PR #1022, opened 2016-03-22
Added highlighting of differences in merge entries dialogs and made s…

oscargus:
Based on an issue I cannot find at the moment.

* Show all information in the merge entries fields
* Show the difference between fields

May not be the worlds most advanced formatting and diff algorithm but it works (it seems).

clipboard01

- [x] Tests green?


stefan-kolb:
Looks nice :+1: 

But the usage here is not intuitive to me. We already have a display of the diff here as both values (former and the fetched entry) are displayed next to each other. 
Also, the usage of the dialog is not clear now as you always select to keep the left or the right version.
However, currently the right version is a complex diff, but not a display of the version we select.

I also have a field that shows the result of the merged entry to the user, which lets me wonder what the advantage of the displaying a diff is.

oscargus:
It was requested at some stage (by @koppor ?) and I think the main point is
that it sometimes can be tricky to see the difference between two fields.

If the difference is significant it is indeed more confusing with the
current format. Assuming that we do not come up with a good heuristic for
"significant", the resulting entry can be quite useful.

One thing that struck me now is to always show the diff in the non-selected
field. Would that make more sense?

oscargus:
Another "solution" might be to add a checkbook to the far right that enables the diff view (and possibly a preference if it should be on or off by default).

tobiasdiez:
Nice idea.

I would opt for a symmetric version. I.e. highlight all diffs on both sides similar to how kdiff3 handles this. So the red diff in the above picture would only show on the left side.
kdiff3

stefan-kolb:
Don't get me wrong, but I'm not quite sure if we really need such functionality. Seems a bit overpowered for such an easy merge task from my point of view. In the end it's just a Bibtex entry and not a huge source code diff.

oscargus:
@tobiasdiez I was thinking about that. However, in most cases the left entry is the original (well, not in the plain merge entries, but when merging DOI information and when called from the ImportInspectionDialog) and this will make it even more cluttered (or how would it look, I cannot really generalize the idea from your screen shot, partly because line differences are not really relevant here).

The current formatting is quite similar to latexdiff so frequent users of that might appreciate it.

@stefan-kolb I think that when you get two almost identical abstracts and need to pick one, you will appreciate it. :-) But I agree that for large differences it is of less use and also that in some cases (e.g. differently formatted names), it is probably more of a problem.

Btw, if you look carefully at the screen shot you can see that it is not fully working. That should be fixed now.

tobiasdiez:
Its hard to explain in words and I have to admit the screenshot is also not too helpful. Let me try again:
The idea was to highlight additions. So 
left: **This text only occurs left,** same text.
right: same text**, except that this is right.**
(in your version it would be for right ~~This text only occurs left,~~ same text**, except that this is right.**). Changes (i.e. `A` vs `a`) could be highlighted using a different color.
In this way you have a more symmetric view and don't run into the problems mentioned by @stefan-kolb.

oscargus:
@tobiasdiez Thanks for the explanation, I believe I got it now. I think that it can be done like that as well. However, I believe that both methods are good, so in that case I will support both (I assume that there could be people that want to turn it off completely as well...).

oscargus:
OK, I have added a diff mode based on what I understood from @tobiasdiez. The code structure is at least slightly better now, but the fundamental code is indeed quite ugly.

What should one call the three different diff modes? "Plain text" seems quite obvious, but then? "Latexdiff style" is a possible alternative for the first example, but what about the one below?

Input on formatting is also welcome. I'm thinking about using CSS, which seems to be a slightly better idea than old school tags I guess.


clipboard02

oscargus:
There's a bit of a problem to push the latest version now. Not sure what happens, but I'll try again tomorrow.

oscargus:
I added the option to select between word and character based diffs now, so one can select between five different modes. Builds available at http://builds.jabref.org/diff/ if you want to try it out.

Still to be done:
[ ] Store the last diff mode (will do once the number of them are decided)
[ ] Determine suitable names for the diff modes

simonharrer:
I am against having five different modes - too much choice for the user. I propose to remove the symmetric ones, and maybe use either latex word or character diff. For me, the word diff seems to be more helpful.

stefan-kolb:
I propose plain text and latex diff character. I see a use for larger text passages like abstract or review where I think diff style character is best. In general I wouldnt include such hacker functionality at all since I think my eyes can see the differences between two strings best in plain text except for the two exceptions mentioned. I would be OK with such a compromise. Let's see what the majority decides.

mlep:
Thanks @oscargus for this feature. Very useful!

To me, the "LaTeXdiff style - character" is the most useful: I can quickly grasp the differences. However for a small difference in a long text (such as an abstract), the "LaTeXdiff style - single word" may better show it.

I believe Plain text is not useful. My rational is that the software should help me see the differences: Why should I have a mode that make me spend time on searching for differences?
Just my opinion.

Also (please, let me know if I should open a new issue):
- mouse scrolling is very very slow for the top part of the window (while it works as expected on the lower part --- both right and left sides).
- I never got the abstract from the DOI...  :cry:
- It would be convenient if pressing "ESC" could close this window (like for other windows).

oscargus:
Not sure what the consensus is here, as I understand it these are the standpoints:

Person   | Preference
---------|----------
@stefan-kolb | Plain text (maybe latexdiff, character)
@simonharrer | latexdiff, word (maybe character)
@tobiasdiez | symmetric
@mlep | latexdiff, character (maybe word)

Remaining @JabRef/developers: do you want to have a say?

I've updated with a new version with the latexdiff version in both modes, so:
* Plain text
* Show diff - word
* Show diff - character

Haven't removed the symmetric code yet though.

@mlep : will try to add ESC-closing! The mouse scrolling I'm not really sure how to fix (and do not notice).

simonharrer:
To me, the current state looks good enough(tm).

oscargus:
OK! I went for that.

mlep:
About  ESC-closing and mouse scrolling, shall I open 2 new issues (for later triage)?

oscargus:
@mlep, it might be the safest...

tobiasdiez:
I really like the symmetric version you posted in https://github.com/JabRef/jabref/pull/1022#issuecomment-200574905.

oscargus:
I do not have any strong opinions. I like the idea to be able to see the differences (obviously) and I do not really see it as a problem if there are quite a few variants. Most people would try them out and find one or two useful and leave it there. The overhead in terms of preferences is the same as well.

It is more or less just to revert the last commit, add translations for the symmetric case (and come up with good names, which seems to be the most tricky part if we have two variants), and change the default (as that was part of the commit).

mlep:
FYI: I preferred the "Show diff" because the "Symmetric" requires me to look at both sides of the window to find the differences. It seems I am the laziest of you all...  :flushed:

simonharrer:
Can be merged as well. Go ahead. :)

mlep:
Suggestion: in a few weeks, encourage users to test this feature and give us feedback (via blog post and email to user list)

stefan-kolb:
Can we please make `plain`default? Do you find this intuitive:
image
Or this:
image

mlep:
This or this? I prefer this.

**Suggestion:** since, this is not urgent to decide about, let's ask the community of users to test this feature and vote for the preferred one and the ugliest one. A good opportunity to have some input from users.

stefan-kolb:
Seems like I'm alone here :cry: ? I just cannot see the output in the first pic...

oscargus:
I thought that someone suggested to use that as the default, but maybe not.
Anyway, there was a point of not having plain text as default, namely to
show the users that something had happened. Now they will at least notice
that something did change and be able to turn it off (as opposed to
noticing the drop-down when the text looks more or less the same.)

Names are quite interesting though as there are huge changes if the name
formats are different and it may be tricky to see the actual difference.
Not sure if we should normalize the names before showing them? Then it
would be easier to spot actual differences.

(Also, if I were to edit your references there would be fewer
differences... ;-))

stefan-kolb:
@oscargus The pictures are just from auto retrieved entries, didn't set anything there in my bib :smile:

tobiasdiez:
If I remember correctly, the decision of the devcall was to have `plain` as default...

oscargus:
@stefan-kolb I was primarily thinking about all the {} in the left hand entry, but I know that we have different opinions there. :-)

@tobiasdiez OK, I had no idea. Must be something I made up myself then.

stefan-kolb:
@oscargus Yeah, in that case it wasn't me. Think the entry comes from Zotero or so... :smile:

PR #1025, opened 2016-03-23
Added IEEE PDF fetcher

oscargus:
PDFs can now be automatically downloaded from IEEE Xplore, given that you have IP-based access (if not, nothing should happen).

- [x] Tests green?

### DIFF
--- a/src/main/java/net/sf/jabref/logic/fetcher/IEEE.java
+++ b/src/main/java/net/sf/jabref/logic/fetcher/IEEE.java
@@ -1,6 +1,7 @@
 package net.sf.jabref.logic.fetcher;
 
 import java.io.IOException;
+import java.net.URI;
 import java.net.URL;
 import java.util.Objects;
 import java.util.Optional;
@@ -13,7 +14,6 @@ import net.sf.jabref.logic.net.URLDownload;
 import net.sf.jabref.logic.util.DOI;
 import net.sf.jabref.model.entry.BibEntry;
 
-
 public class IEEE implements FullTextFinder {
 
     private static final Log LOGGER = LogFactory.getLog(IEEE.class);
@@ -22,28 +22,29 @@ public class IEEE implements FullTextFinder {
             .compile("\"(http://ieeexplore.ieee.org/ielx[0-9/]+\\.pdf[^\"]+)\"");
     private static final String IEEE_DOI = "10.1109";
     private static final String BASE_URL = "http://ieeexplore.ieee.org";
+
+
     @Override
     public Optional<URL> findFullText(BibEntry entry) throws IOException {
         Objects.requireNonNull(entry);
         Optional<URL> pdfLink = Optional.empty();
 
-            Optional<DOI> doi = DOI.build(entry.getField("doi"));
+        Optional<DOI> doi = DOI.build(entry.getField("doi"));
         if (doi.isPresent() && doi.get().getDOI().startsWith(IEEE_DOI)) {
-            URLDownload urlDownload = new URLDownload(doi.get().getURI().get().toURL());
-            String firstResult = urlDownload.downloadToString();
-            Matcher matcher = STAMP_PATTERN.matcher(firstResult);
-            if (matcher.find()) {
-                URL stampURL = new URL(BASE_URL + matcher.group(1));
-                String secondResult = new URLDownload(stampURL).downloadToString();
-                matcher = PDF_PATTERN.matcher(secondResult);
+            Optional<URI> uri = doi.get().getURI();
+            if (uri.isPresent()) {
+                String firstResult = new URLDownload(uri.get().toURL()).downloadToString();
+                Matcher matcher = STAMP_PATTERN.matcher(firstResult);
                 if (matcher.find()) {
-                    String pdfURLText = matcher.group(1);
-                    pdfLink = Optional.of(new URL(pdfURLText));
-                    LOGGER.debug("Full text document found on IEEE Xplore");
+                    String secondResult = new URLDownload(new URL(BASE_URL + matcher.group(1))).downloadToString();
+                    matcher = PDF_PATTERN.matcher(secondResult);
+                    if (matcher.find()) {
+                        pdfLink = Optional.of(new URL(matcher.group(1)));
+                        LOGGER.debug("Full text document found on IEEE Xplore");
+                    }
                 }
             }
         }
-
         return pdfLink;
     }
 
--- a/src/main/java/net/sf/jabref/logic/fetcher/IEEE.java
+++ b/src/main/java/net/sf/jabref/logic/fetcher/IEEE.java
@@ -10,6 +10,9 @@ import java.util.regex.Pattern;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+
+import com.google.common.base.Charsets;
+
 import net.sf.jabref.logic.net.URLDownload;
 import net.sf.jabref.logic.util.DOI;
 import net.sf.jabref.model.entry.BibEntry;
@@ -33,10 +36,11 @@ public class IEEE implements FullTextFinder {
         if (doi.isPresent() && doi.get().getDOI().startsWith(IEEE_DOI)) {
             Optional<URI> uri = doi.get().getURI();
             if (uri.isPresent()) {
-                String firstResult = new URLDownload(uri.get().toURL()).downloadToString();
+                String firstResult = new URLDownload(uri.get().toURL()).downloadToString(Charsets.UTF_8);
                 Matcher matcher = STAMP_PATTERN.matcher(firstResult);
                 if (matcher.find()) {
-                    String secondResult = new URLDownload(new URL(BASE_URL + matcher.group(1))).downloadToString();
+                    String secondResult = new URLDownload(new URL(BASE_URL + matcher.group(1)))
+                            .downloadToString(Charsets.UTF_8);
                     matcher = PDF_PATTERN.matcher(secondResult);
                     if (matcher.find()) {
                         pdfLink = Optional.of(new URL(matcher.group(1)));
--- a/src/main/java/net/sf/jabref/logic/fetcher/IEEE.java
+++ b/src/main/java/net/sf/jabref/logic/fetcher/IEEE.java
@@ -1,7 +1,6 @@
 package net.sf.jabref.logic.fetcher;
 
 import java.io.IOException;
-import java.net.URI;
 import java.net.URL;
 import java.util.Objects;
 import java.util.Optional;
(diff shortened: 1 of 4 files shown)


src/main/java/net/sf/jabref/logic/fetcher/IEEE.java
```diff
@@ -0,0 +1,50 @@
...
+    private static final Pattern PDF_PATTERN = Pattern
+            .compile("\"(http://ieeexplore.ieee.org/ielx[0-9/]+\\.pdf[^\"]+)\"");
+    private static final String IEEE_DOI = "10.1109";
+    private static final String BASE_URL = "http://ieeexplore.ieee.org";
+    @Override
+    public Optional<URL> findFullText(BibEntry entry) throws IOException {
+        Objects.requireNonNull(entry);
+        Optional<URL> pdfLink = Optional.empty();
+
+            Optional<DOI> doi = DOI.build(entry.getField("doi"));
```
stefan-kolb: Indentation

stefan-kolb:
You might also access the IEEEXplore API directly to ommit the three URLDownloads?
http://ieeexplore.ieee.org/gateway/
We also need a bunch of tests for this, see the other fetchers for examples.

src/main/java/net/sf/jabref/logic/fetcher/IEEE.java
oscargus: Whoops!

oscargus:
I was considering that, but parsing HTML seemed like a ~~better~~ easier solution (for me) compared to XML.

Regarding tests, I have no idea how to make a proper test since the outcome of the test will be IP-number dependent... It seems quite useless to have a test that pass independent of if the PDF was downloaded or not. Input is warmly welcome!

stefan-kolb:
The result of this stamp link http://ieeexplore.ieee.org/stamp/stamp.jsp?tp=&arnumber=7421926 is IP dependent?
-> http://ieeexplore.ieee.org/ielx7/6287639/7419931/07421926.pdf?tp=&arnumber=7421926&isnumber=7419931

The tests for the fetchers are not for checking if a PDF is downloaded, but if the correct URL is transferred?!

stefan-kolb:
Regarding the HTML solution: In my opinion it is more likely that the API remains stable than the HTML structure. Therefore, I always tend to use some API if possible.

oscargus:
The URL that is provided from the API (and the first step of my method) is an URL to a HTML page which loads the PDF in a frame (to show copyright etc and probably prohibit exactly what we are doing). Therefore, the second regexp is still needed and if you do not have access to the PDF you will not be able to get the actual PDF URL which is what the method should return.

I fully agree with your point regarding API, it was more a matter of being able to add the functionality at all that lead me down the HTML road, of course I can regexp the first URL from the XML using the exactly same code just as well ;-).

oscargus:
Yes, try that link from a non-University computer... (Such as your phone.)

oscargus:
OK, let me rephrase that. That particular link was open access, so, well, that was a very good point! Thanks!

stefan-kolb:
I thought so...you made me wonder :smile:

stefan-kolb:
I also test the other fetchers with open access articles, its the only way to make it work consistently everywhere and still be able to get a failing test if something changed.

oscargus:
Tests are added. :-)

The somehow positive argument for the current approach is that it isn't even dependent on the API. Since it is based on DOI lookup and a rather generic reg exp, the DOI resolution page can be changed and it should still succeed. Then, of course, it is probably more likely that the intermediate frame page and the stamp mechanism is changed than the API...

src/test/java/net/sf/jabref/logic/fetcher/IEEETest.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @Test
+    public void doiNotPresent() throws IOException {
+        Assert.assertEquals(Optional.empty(), finder.findFullText(entry));
+    }
+
+    @Test(expected = NullPointerException.class)
+    public void rejectNullParameter() throws IOException {
+        finder.findFullText(null);
+        Assert.fail();
```
stefan-kolb: We don't need assert.fail() here, do we?

src/test/java/net/sf/jabref/logic/fetcher/IEEETest.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    @Test
+    public void findByDOI() throws IOException {
+        // CI server is unreliable
+        Assume.assumeFalse(DevEnvironment.isCIServer());
```
stefan-kolb: Is the CI unreliable for IEEE? Only include this if we get problems with the test on CI.

src/test/java/net/sf/jabref/logic/fetcher/IEEETest.java
```diff
@@ -0,0 +1,59 @@
...
+        Assert.assertEquals(
+                Optional.of(
+                        new URL("http://ieeexplore.ieee.org/ielx7/6287639/7419931/07421926.pdf?tp=&arnumber=7421926&isnumber=7419931")),
+                finder.findFullText(entry)
+        );
+    }
+
+    @Test
+    public void notFoundByDOI() throws IOException {
+        // CI server is unreliable
```
stefan-kolb: Same as above.

src/test/java/net/sf/jabref/logic/fetcher/IEEETest.java
```diff
@@ -0,0 +1,59 @@
...
+        finder.findFullText(null);
+        Assert.fail();
+    }
+
+    @Test
+    public void findByDOI() throws IOException {
+        // CI server is unreliable
+        Assume.assumeFalse(DevEnvironment.isCIServer());
+
+        entry.setField("doi", "10.1109/ACCESS.2016.2535486"); // Open access paper
```
stefan-kolb: Doesn't matter for the test what kind of paper it is, just keep the code clean.

src/test/java/net/sf/jabref/logic/fetcher/IEEETest.java
```diff
@@ -0,0 +1,59 @@
...
+                finder.findFullText(entry)
+        );
+    }
+
+    @Test
+    public void notFoundByDOI() throws IOException {
+        // CI server is unreliable
+        Assume.assumeFalse(DevEnvironment.isCIServer());
+
+        entry.setField("doi", "10.1021/bk-2006-WWW.ch014"); // Arbitrary non-IEEE DOI
```
stefan-kolb: Unnecessary comment for the test.

stefan-kolb:
I think it is a cool addition :smile: ! However, I'm still not really happy with the regex approach, let's see what the others think!

src/test/java/net/sf/jabref/logic/fetcher/IEEETest.java
oscargus: This test is very similar to ACSTest. ;-)

oscargus: Not really, but then some static code analysis tools complains, so no harm I guess.

src/test/java/net/sf/jabref/logic/fetcher/IEEETest.java
oscargus: I thought that the point was that the CI server was unreliable for external net access, but I have removed it.

src/test/java/net/sf/jabref/logic/fetcher/IEEETest.java
oscargus: It sort of does. I just wanted to state what the requirements for selecting that particular DOI was (in case it must be changed in the future). I have removed it though.

src/main/java/net/sf/jabref/logic/fetcher/IEEE.java
```diff
@@ -0,0 +1,55 @@
...
+    private static final String BASE_URL = "http://ieeexplore.ieee.org";
+
+
+    @Override
+    public Optional<URL> findFullText(BibEntry entry) throws IOException {
+        Objects.requireNonNull(entry);
+        Optional<URL> pdfLink = Optional.empty();
+
+        Optional<DOI> doi = DOI.build(entry.getField("doi"));
+        if (doi.isPresent() && doi.get().getDOI().startsWith(IEEE_DOI)) {
```
tobiasdiez: I would invert some if statements and return an empty optional, so that you do not have 4 nested ones.

src/test/java/net/sf/jabref/logic/fetcher/IEEETest.java
oscargus: Same as above.

src/main/java/net/sf/jabref/logic/fetcher/IEEE.java
```diff
@@ -0,0 +1,55 @@
...
+            Optional<URI> uri = doi.get().getURI();
+            if (uri.isPresent()) {
+                String firstResult = new URLDownload(uri.get().toURL()).downloadToString(Charsets.UTF_8);
```
tobiasdiez: What is `firstResult`?

src/main/java/net/sf/jabref/logic/fetcher/IEEE.java
```diff
@@ -0,0 +1,55 @@
...
+    private static final Pattern PDF_PATTERN = Pattern
+            .compile("\"(http://ieeexplore.ieee.org/ielx[0-9/]+\\.pdf[^\"]+)\"");
+    private static final String IEEE_DOI = "10.1109";
+    private static final String BASE_URL = "http://ieeexplore.ieee.org";
+
+
+    @Override
+    public Optional<URL> findFullText(BibEntry entry) throws IOException {
+        Objects.requireNonNull(entry);
+        Optional<URL> pdfLink = Optional.empty();
```
tobiasdiez: The local variable is actually not needed, just return the appropriate value at the right place.

src/main/java/net/sf/jabref/logic/fetcher/IEEE.java
oscargus: The HTML page after resolving the DOI.

tobiasdiez: Ah ok. Then I would suggest a rename of the variable.

tobiasdiez:
LGTM

stefan-kolb:
I'm not so happy with the addition of the download via the URL. Whereas it is convenient, I'd like to keep the Fetchers based on "immutable" values. Something like DOI or Eprint should normally stay the same. A URL is not a goog identifier in my opinion. This suggestion is more like a general discussion for keeping the implementations of the fetchers focused and not cluttering them with a lot of download options (by DOI, Eprint, URL, Title,...). WDYT?

stefan-kolb:
What is great in your approach is that you check for a IEEE DOI and skip if it is not issued by the vendor. This could be nice for the other fetchers too as we save HTTP request etc. However, I'm not sure if we can savely find out all DOIs that can be issued by a vendor.

oscargus:
The main reason for adding URL support is that IEEE also have some papers
without IEEE DOI. That was the main reason for adding the URL-part. Adding
more DOI doesn't work as those publishers sends elsewhere (or to an
intermediate page where one can select provider).

It is primarily only useful for older bib-files as the current IEEE search
doesn't add it. Not sure what e.g. Google provides. However, with better
meta data search it can be quite useful. At the moment it was more a
convenience thing to be able to populate my old bib-files with PDFs.

lenhard:
Is anything missing here or can this be merged?

oscargus:
Apart from the disagreement on using the URL or not it can be merged.

oscargus:
The thing is that even using the DOI one will currently need to rely on a given URL as the DOI points to a HTML page with a frame in which the PDF is loaded. If they would change that, one would also need to change the code.

The benefit, as stated above, is that papers with non-IEEE DOI on IEEE Xplore, can now be downloaded. This hold, e.g., for IET papers (Electronic Letters among others). Now, for this a better solution would indeed be to use the API as that provides the URL (as far as I can tell). Still, the HTML parsing must be done, but then one could get around the non-IEEE on IEEE Xplore issue. (It is not as easy as adding more DOI series, see http://dx.doi.org/10.1049/el.2012.0577 as an example.)

Issue #970, opened 2016-03-15
Support for a shared database


koppor:
JabRef should support a database (MySQL, Postgres, Oracle, ...) as backend. As far as I understand the current implementation, the DB export simply overwrites all data. This is a show stopper for collaborative work.

JabRef should support collaborative work where a database (MySQL, Postgres, or Oracle) is shared between multiple users. If an entry is added (or even changed) to the DB, all other users should see it.

If there are concurrent edits, the user should be warned and the window "Merge duplicate entries" should be shown. With the event system (#969), this might be obsolete as JabRef will instantly synchronize the changes.

The implementation heavily depends on the event system of JabRef (#969).

As soon as the feature is implemented, it should also be possible to import a bib into the database and export the database to a bib. This should work with the current functionality, but we have to add documentation and a test case.

Background: For many users, local installation of desktop software is OK. They are not on the everything-in-the-web-train.

- [x] conflict handling easy - at most [optimistic offline lock](http://martinfowler.com/eaaCatalog/optimisticOfflineLock.html)

## Dropped wishes:

- show online users globally
- show users editing the current entry

This is (currently) too much work for too little gain.

## Implementation hints:

 * http://stackoverflow.com/q/12618915/873282
    * Oracle: https://docs.oracle.com/cd/E11882_01/java.112/e16548/dbchgnf.htm#JJDBC28818
    * Postges: http://www.postgresql.org/docs/9.0/static/sql-listen.html
    * not available for other databases.

Decision: Only implement it for Oracle and Postgres.

obraliar:
It would be nice if the new EventSystem would get merged in, so I could use the existing event bus and its event objects.

Issue #969, opened 2016-03-15
Add event system to JabRef


koppor:
JabRef needs an event system.

Idea: https://github.com/JabRef/jabref/pull/937#issuecomment-195420018

Details are discussed at https://github.com/JabRef/jabref/pull/1028/#issuecomment-201400440.

PR #1028, opened 2016-03-23
[WIP] AddEventSystem

obraliar:
- [x] Introduce an EventBus object being passed around. This enables better testing.
- [x] Make `DatabaseChangeEvent` abstract and add subclasses according to `DatbaseChangeEvent`
- [x] Rewrite the currently existing code to use that event bus instead of `net.sf.jabref.model.database.BibDatabase.addDatabaseChangeListener(DatabaseChangeListener)` and `net.sf.jabref.model.database.BibDatabase.removeDatabaseChangeListener(DatabaseChangeListener)` 
- [x] Changes described in CHANGELOG.md?
- [x] Tests green?


obraliar:
Issue: https://github.com/JabRef/jabref/issues/969

obraliar:
This commit should also demonstrate the structure of the new event system.

Discussions are welcome.

oscargus:
I just want to point out that there is an event system for BibDatabase. Which is not that much used. Maybe it can be used as inspiration/replaced.

koppor:
This does demonstrate how the API of Google Guava works, but not how it can really be used at JabRef. Should we use one bus for everything? Should we use one bus per "thing" such as BibDatabase etc. - This is also [discussed at the documentation](https://github.com/google/guava/wiki/EventBusExplained#why-must-i-create-my-own-event-bus-rather-than-using-a-singleton).

Please double check the usage of `net.sf.jabref.model.database.DatabaseChangeEvent`. I think, the whole code around `net.sf.jabref.model.database.DatabaseChangeEvent` should be replaced with the Guava thing. Instead of `net.sf.jabref.model.database.DatabaseChangeEvent.ChangeType` there should be different classes. Then, it is possible to use the power of the [EventBus](https://google.github.io/guava/releases/snapshot/api/docs/com/google/common/eventbus/EventBus.html). Important excerpts of the documentation: "which accepts a single argument of the type of event desired; " and "The EventBus instance will determine the type of event and route it to all registered listeners.". This enables to use a single bus with dedicated events such as `FieldContentChanged` etc.

Thus, please update this PR to
- [ ] Introduce a Global EventBus into `net.sf.jabref.Globals`
- [ ] Make `DatabaseChangeEvent` abstract and add subclasses according to `DatbaseChangeEvent`
- [ ] Rewrite the currently existing code to use that event bus instead of `net.sf.jabref.model.database.BibDatabase.addDatabaseChangeListener(DatabaseChangeListener)` and `net.sf.jabref.model.database.BibDatabase.removeDatabaseChangeListener(DatabaseChangeListener)`

obraliar:
Yes, it actually only demonstrates the base function. I think we should have a separate package for that. So I created a new one and created a interface, which should normalize the usage of Google guava. Now we are going to replace the existing listeners and as @koppor  already observed to prepare a global event bus.

tobiasdiez:
Mhh...I'm not sure that a global event bus is the right way to go. It adds another global object, everyone gets informed about everything (it is hard to listen only to changes to a certain database) etc. But of course it also has its advantages. What is the opinion of the other @JabRef/developers here?

obraliar:
I think its OK to use one. Therefore we use different `Event` objects, which should prevent the chaos.

koppor:
On the one hand, I don't like to pass around the EventBus object everywhere. It is a global thing and used everywhere. Maybe we should place it in `net.sf.jabref.gui.GUIGlobals`.

On the other hand, I agree that a global object makes testing difficult.

Let's hear about design ideas by the others. This, however, doesn't stop @obraliar replacing the old event system. The "location" of the EventBus object can be changed later on.

src/main/java/net/sf/jabref/gui/event/Event.java
```diff
@@ -0,0 +1,21 @@
+package net.sf.jabref.gui.event;
+
+/**
+ * Represents a medium which can be used
+ * to pass objects on event occurrence.
+ */
+
+public class Event {
```
tobiasdiez: Why not use https://docs.oracle.com/javase/7/docs/api/java/util/EventObject.html for this?

src/main/java/net/sf/jabref/gui/event/listener/EventListener.java
```diff
@@ -0,0 +1,11 @@
+package net.sf.jabref.gui.event.listener;
+
+import com.google.common.eventbus.Subscribe;
+
+import net.sf.jabref.gui.event.Event;
+
+public interface EventListener {
```
tobiasdiez: Should probably inherit from https://docs.oracle.com/javase/7/docs/api/java/util/EventListener.html

src/main/java/net/sf/jabref/gui/event/listener/EventListener.java
```diff
@@ -0,0 +1,11 @@
...
+
+    @Subscribe
```
tobiasdiez: What is the purpose of the subscribe tags

src/main/java/net/sf/jabref/gui/event/Event.java
obraliar: Oh, I didn't know that there is a existing class for this purpose :)

koppor: I see no reason for using java-util. We are implementing something contrary when using Guava.

src/main/java/net/sf/jabref/gui/event/listener/TestEventListener.java
```diff
@@ -0,0 +1,19 @@
+package net.sf.jabref.gui.event.listener;
+
+
+import net.sf.jabref.gui.event.Event;
+
+public class TestEventListener implements EventListener {
```
tobiasdiez: How does the TestEventListener signals that it only wants to listen for testevents and not arbitrary events?

koppor:
@obraliar Please do not take further action until @JabRef/developers decided which way to go.

src/main/java/net/sf/jabref/gui/event/listener/TestEventListener.java
obraliar: `TestEventListener` does not signal anything cause it implements the interface `EventListener`. `EventListener` signals with `@Subscribe` that it's accepting all events of type `Event` (subclasses included).

src/main/java/net/sf/jabref/gui/event/Event.java
Siedlerchr: I agree with @koppor  From the api doc the EventObject looks like that is's main purpose is to be used in GUI swing/awt events.

obraliar:
No opinions? If not, we would begin to discuss the details internally within the stupro group.

lenhard:
Sorry, but (in contrast to other developers) I do not see why we need an event system. Everything we need should be doable with the plain old observer pattern and the benefit of an event bus is not clear to me. 

This will be discussed in a devcall, which may or may not take place soon.

obraliar:
Initially I commented out the code of the old event solution.

obraliar:
Now the Software Events only base upon Google's Guava Bus system.

obraliar:
Now all entry events only base upon Google's Guava Bus System.
The integrated event system (listener registration, event posts & receivements, etc.) works.

Here are some new findings:
- We have to use one EventBus for each BibDatabase, cause there may be more than one opened. A main event such as `AddEntryEvent` should not affect all opened databases.
- For all other events that do not affect BibDatabase we could introduce one general EventBus. This part of introduction is very easy, but I see no need for that yet.

src/main/java/net/sf/jabref/event/Event.java
```diff
@@ -0,0 +1,12 @@
+package net.sf.jabref.event;
+
+/**
+ * Represents an interface of a medium which can be used
+ * to pass objects on event occurrence.
+ */
+public interface Event {
```
koppor: Do not introduce a generic Event object. This is not necessary.

src/main/java/net/sf/jabref/event/ChangeEntryEvent.java
```diff
@@ -0,0 +1,9 @@
+package net.sf.jabref.event;
+
+public class ChangeEntryEvent extends AddOrChangeEntryEvent {
+
+    public ChangeEntryEvent(Object object) {
```
koppor: An Entry is always a BibEntry. Please rely on Java's type checking and do not use generic types and cast internally.

src/main/java/net/sf/jabref/model/database/BibDatabase.java
```diff
@@ -84,9 +87,10 @@
     private boolean followCrossrefs = true;
 
     /**
-     * Behavior
+     * Event bus
```
koppor: Do not add comments, which an automatically be derived from the variable name.

src/main/java/net/sf/jabref/event/AddOrChangeEntryEvent.java
```diff
@@ -0,0 +1,28 @@
+package net.sf.jabref.event;
+
+import net.sf.jabref.model.entry.BibEntry;
+
+public class AddOrChangeEntryEvent implements Event {
```
koppor: Do not introduce this event. -> No hierarchy.

If one wants to listen to both Add and Change, then one implements two subscribe methods internally calling a generic method.

This makes it much more easier for the publisher of the event. In your case, if one adds an entry, one has to call both, AddEntryEvent and AddOrChangeEntryEvent.

src/test/java/net/sf/jabref/event/RemoveEntryEventTest.java
```diff
@@ -0,0 +1,38 @@
...
+    public void testReceivingEvent() {
+        BibEntry shouldBeBibEntry = new BibEntry();
+        shouldBeBibEntry.setId("testkey2");
+        RemoveEntryEvent aocee = new RemoveEntryEvent(shouldBeBibEntry);
+        eventBus.post(aocee);
+
+        Assert.assertEquals(shouldBeBibEntry.getId(), testEventListener.getBibEntry().getId());
+        Assert.assertEquals(shouldBeBibEntry.getId(), aocee.getEntry().getId());
+    }
+
```
koppor: Too many empty lines ;)

src/test/java/net/sf/jabref/event/RemoveEntryEventTest.java
```diff
@@ -0,0 +1,38 @@
...
+    @Test
+    public void testReceivingEvent() {
+        BibEntry shouldBeBibEntry = new BibEntry();
+        shouldBeBibEntry.setId("testkey2");
+        RemoveEntryEvent aocee = new RemoveEntryEvent(shouldBeBibEntry);
+        eventBus.post(aocee);
+
+        Assert.assertEquals(shouldBeBibEntry.getId(), testEventListener.getBibEntry().getId());
+        Assert.assertEquals(shouldBeBibEntry.getId(), aocee.getEntry().getId());
```
koppor: Why not directly comparing the entries?

src/test/java/net/sf/jabref/event/TestHierarchicalEventListener.java
```diff
@@ -0,0 +1,36 @@
+package net.sf.jabref.event;
+
+import net.sf.jabref.model.entry.BibEntry;
+
+import com.google.common.eventbus.Subscribe;
+
+public class TestHierarchicalEventListener {
```
koppor: I think, this is calls is only used in testing and does not test code in the main JabRef code. Just remove it.

src/main/java/net/sf/jabref/event/Event.java
obraliar: Done.

src/main/java/net/sf/jabref/event/ChangeEntryEvent.java
obraliar: Done.

src/main/java/net/sf/jabref/model/database/BibDatabase.java
obraliar: Done.

src/main/java/net/sf/jabref/event/AddOrChangeEntryEvent.java
obraliar: Done. (Hierarchy & hierarchy tests dissolved)

src/test/java/net/sf/jabref/event/RemoveEntryEventTest.java
obraliar: Done.

src/test/java/net/sf/jabref/event/RemoveEntryEventTest.java
obraliar: I thought about that, but the codacy test fails every time, when not using JUnit-Assert. Can this be fixed?

src/test/java/net/sf/jabref/event/TestHierarchicalEventListener.java
obraliar: Done.

src/main/java/net/sf/jabref/event/AddEntryEvent.java
```diff
@@ -0,0 +1,21 @@
...
+
+public class AddEntryEvent {
+
+    private BibEntry bibEntry;
+
+    public AddEntryEvent(BibEntry bibEntry) {
+        setBibEntry(bibEntry);
+    }
+
+    public void setBibEntry(BibEntry bibEntry) {
```
koppor: Is this method necessary? I would remove it and inline the setting in the constructor.

Reason: the event is set in the constructor

src/main/java/net/sf/jabref/event/AddEntryEvent.java
```diff
@@ -0,0 +1,21 @@
+package net.sf.jabref.event;
+
+import net.sf.jabref.model.entry.BibEntry;
+
+public class AddEntryEvent {
+
+    private BibEntry bibEntry;
```
koppor: Please make this final

src/main/java/net/sf/jabref/event/ChangeEntryEvent.java
```diff
@@ -0,0 +1,21 @@
+package net.sf.jabref.event;
+
+import net.sf.jabref.model.entry.BibEntry;
+
+public class ChangeEntryEvent {
```
koppor: The comments for AddEntryEvent also apply here.

src/main/java/net/sf/jabref/event/RemoveEntryEvent.java
```diff
@@ -0,0 +1,21 @@
+package net.sf.jabref.event;
+
+import net.sf.jabref.model.entry.BibEntry;
+
+public class RemoveEntryEvent {
```
koppor: The comments for AddEntryEvent also apply here.

koppor:
Please add a test case where you subscribe to the eventbus of the database and test if the events added, removed, and changed are properly raised?

src/main/java/net/sf/jabref/event/AddEntryEvent.java
obraliar: The idea was to leave possibility for setting `bibEntry` later. I can imagine that updating the `bibEntry` to another (within one event) is needed. But this situation didn't occur yet, so I'm going to remove it.

src/main/java/net/sf/jabref/event/AddEntryEvent.java
obraliar: Done.

src/main/java/net/sf/jabref/event/ChangeEntryEvent.java
obraliar: Done.

src/main/java/net/sf/jabref/event/RemoveEntryEvent.java
obraliar: Done.

src/main/java/net/sf/jabref/gui/BasePanel.java
```diff
@@ -1279,27 +1278,34 @@ public void databaseChanged(final DatabaseChangeEvent e) {
...
+        @Subscribe
+        public void listen(AddEntryEvent aocee) {
+            searchAutoCompleter.addBibtexEntry(aocee.getBibEntry());
+        }
+
+        @Subscribe
+        public void listen(ChangeEntryEvent aocee) {
+            searchAutoCompleter.addBibtexEntry(aocee.getBibEntry());
         }
+
```
tschechlovdev: please remove this empty line.

src/main/java/net/sf/jabref/model/database/BibDatabase.java
```diff
@@ -83,10 +86,8 @@
      */
     private boolean followCrossrefs = true;
 
-    /**
-     * Behavior
-     */
-    private final Set<DatabaseChangeListener> changeListeners = new HashSet<>();
+    private final EventBus eventBus;
+
```
tschechlovdev: remove empty line

src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
```diff
@@ -187,4 +189,35 @@ public void addSameStringIdTwiceThrowsKeyCollisionException() {
...
+        BibDatabase database = new BibDatabase();
+        BibEntry shouldBeEntry = new BibEntry(IdGenerator.next());
+        TestEventListener tel = new TestEventListener();
+        EventBus eventBus = database.getEventBus();
+
+        eventBus.register(tel);
+        database.insertEntry(shouldBeEntry);
+
+        BibEntry isEntry = tel.getBibEntry();
+
```
tschechlovdev: too many empty lines

src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
```diff
@@ -187,4 +189,35 @@ public void addSameStringIdTwiceThrowsKeyCollisionException() {
...
+        BibEntry shouldBeEntry = new BibEntry(IdGenerator.next());
+        TestEventListener tel = new TestEventListener();
+        EventBus eventBus = database.getEventBus();
+
+        eventBus.register(tel);
+        database.insertEntry(shouldBeEntry);
+        database.removeEntry(shouldBeEntry);
+
+        BibEntry isEntry = tel.getBibEntry();
+
```
tschechlovdev: see above

src/test/java/net/sf/jabref/event/AddEntryEventTest.java
```diff
@@ -0,0 +1,32 @@
...
+
+    @Test
+    public void testEventReceivement() {
+        BibEntry shouldBeBibEntry = new BibEntry();
+        shouldBeBibEntry.setId("testkey1");
+        AddEntryEvent aocee = new AddEntryEvent(shouldBeBibEntry);
+        eventBus.post(aocee);
+
+        Assert.assertEquals(shouldBeBibEntry.getId(), testEventListener.getBibEntry().getId());
+        Assert.assertEquals(shouldBeBibEntry.getId(), aocee.getBibEntry().getId());
```
tschechlovdev: why not using BibtexEntryAssert.assertEquals?

obraliar: see https://github.com/JabRef/jabref/pull/1028#discussion_r60878805

tschechlovdev: This also appeared in other PR's and they still were merged. So I don't think that this should be a problem.

obraliar: It would be nice if we could fix that, because BibtexEntryAssert uses JUnit-Assert internally too. @koppor: Is it possible?

tobiasdiez: What is the test failure message here? Did you tried out the with the latest master branch because the BibtexEntryAssert was changed recently. There is also an open PR by Olly about this.

obraliar: Can't remember exactly, but the message was about using a test method without Assert (that means JUnit-Assert). I just fetched and rebased on current master... Can you post the link to the PR, please?

tobiasdiez: Already merged https://github.com/JabRef/jabref/pull/1296

src/test/java/net/sf/jabref/event/ChangeEntryEventTest.java
```diff
@@ -22,11 +23,11 @@ public void setUp() {
     @Test
     public void testEventReceivement() {
         BibEntry shouldBeBibEntry = new BibEntry();
-        shouldBeBibEntry.setId("testkey2");
+        shouldBeBibEntry.setId(IdGenerator.next());
```
koppor: When constructing an entry using the default constructor, `IdGenerator.next()` is already called. No need to do an additional call.

src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
```diff
@@ -187,4 +189,29 @@ public void addSameStringIdTwiceThrowsKeyCollisionException() {
...
+    }
+
+    @Test
+    public void testRemoveEntryEventReceivement() {
+        BibDatabase database = new BibDatabase();
+        BibEntry shouldBeEntry = new BibEntry(IdGenerator.next());
+        TestEventListener tel = new TestEventListener();
+        EventBus eventBus = database.getEventBus();
+        eventBus.register(tel);
+        database.insertEntry(shouldBeEntry);
```
koppor: insertEntry and removeEntry are two different events. The event listener should be reset inbetween. Otherwise, the local variable storing the bibentry might not be changed.

Adapt (and rename) this combined test and add another test testing the removal only.

koppor:
Besides the minor comment, it looks good to me.

@obraliar Can you also replace `net.sf.jabref.model.entry.BibEntry.addPropertyChangeListener(VetoableChangeListener)? As far as I could see, there is no need for the vetoing so a normal event bus can be used per entry.

src/main/java/net/sf/jabref/event/AddEntryEvent.java
```diff
@@ -0,0 +1,24 @@
...
+ * A <code>BibEntry</code> object the changes were applied on is required.
+ * <code>AddEntryEvent</code> should be used, when new <code>BibEntry</code>
+ * was added to the database.
+ */
+
+public class AddEntryEvent {
+
+    private final BibEntry bibEntry;
+
+    public AddEntryEvent(BibEntry bibEntry) {
```
tobiasdiez: Please introduce a new abstract class `EntryEvent` which has a constructor accepting a BibEntry and contains the getBibEntry method. Use this new class as the base for the Add/Change/RemoveEntryEvent.

src/main/java/net/sf/jabref/gui/BasePanel.java
```diff
@@ -1279,27 +1278,33 @@ public void databaseChanged(final DatabaseChangeEvent e) {
...
-    private class SearchAutoCompleterUpdater implements DatabaseChangeListener {
+    private class SearchAutoCompleteListener {
 
-        @Override
-        public void databaseChanged(DatabaseChangeEvent e) {
-            if ((e.getType() == ChangeType.CHANGED_ENTRY) || (e.getType() == ChangeType.ADDED_ENTRY)) {
-                searchAutoCompleter.addBibtexEntry(e.getEntry());
-            }
+        @Subscribe
+        public void listen(AddEntryEvent aocee) {
```
tobiasdiez: dont't abbreviate, just use `event` as name. Applies also to many other cases.

src/main/java/net/sf/jabref/model/database/BibDatabase.java
```diff
@@ -562,4 +554,8 @@ public void setEpilog(String epilog) {
     public String getEpilog() {
         return epilog;
     }
+
+    public EventBus getEventBus() {
```
tobiasdiez: I wouldn't expose the EventBus but provide a `register` (or `listenForChanges` or `subscribeToChanges` or whatever) method.

src/test/java/net/sf/jabref/event/AddEntryEventTest.java
```diff
@@ -0,0 +1,33 @@
...
+        testEventListener = new TestEventListener();
+        eventBus.register(testEventListener);
+    }
+
+    @Test
+    public void testEventReceivement() {
+        BibEntry shouldBeBibEntry = new BibEntry();
+        shouldBeBibEntry.setId(IdGenerator.next());
+        AddEntryEvent aocee = new AddEntryEvent(shouldBeBibEntry);
+        eventBus.post(aocee);
```
tobiasdiez: You are actually retesting the EventBus class which is not required. Only test our code.

src/main/java/net/sf/jabref/model/database/BibDatabase.java
```diff
@@ -562,4 +554,8 @@ public void setEpilog(String epilog) {
...
+        return this.eventBus;
```
tobiasdiez: Please all add in the JavaDoc which events are thrown

src/main/java/net/sf/jabref/event/AddEntryEvent.java
koppor: Yeah, this is the really right implementation. I discussed away the old inheritance, because there was Object casting, but this proposal sounds great.

src/main/java/net/sf/jabref/model/database/BibDatabase.java
koppor: I'm not sure whether we should really do information hiding here. I think, we also want to publish changes on the bus. Maybe, if the entry editor changes something?

Possibly, its the best way not to expose the event bus here, but expose it when working on the eventBus for a specific entry.

Currently, I don't know what requirements we have because of #970.

tobiasdiez: For me the Eventbus is an implementation-specific thing and so far nobody outside of the database class has to know about it. Thus hide it. If it is really needed, then the getEventBus can be readded later.

obraliar:
@koppor Done replacing VetoableChangeListener.

Can not understand, why codecov is failing.

src/test/java/net/sf/jabref/event/ChangeEntryEventTest.java
obraliar: Done.

src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
obraliar: Done.

src/main/java/net/sf/jabref/event/AddEntryEvent.java
obraliar: Done.

src/main/java/net/sf/jabref/gui/BasePanel.java
obraliar: Done. (for all)

src/main/java/net/sf/jabref/model/database/BibDatabase.java
obraliar: register/unregister provided.

src/test/java/net/sf/jabref/event/AddEntryEventTest.java
obraliar: Removed.

src/main/java/net/sf/jabref/model/database/BibDatabase.java
obraliar: Done.

src/main/java/net/sf/jabref/gui/PreviewPanel.java
```diff
@@ -275,16 +275,27 @@ public void setLayout(Layout layout) {
...
+            newEntry.registerListener(this.currentChangeFieldUpdateEvent);
         }
+
         entry = Optional.ofNullable(newEntry);
 
         updateLayout();
         update();
     }
 
+    private class ChangeFieldUpdateEvent {
```
tobiasdiez: Is the private class really necessary or can `PreviewPanel` contain the listen method directly and above `this` is passed to the newEntry.registerListener?

src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
```diff
@@ -187,4 +188,29 @@ public void addSameStringIdTwiceThrowsKeyCollisionException() {
...
+    }
+
+    @Test
+    public void testRemoveEntryEventReceivement() {
+        BibDatabase database = new BibDatabase();
+        BibEntry shouldBeEntry = new BibEntry();
+        TestEventListener tel = new TestEventListener();
+        database.registerListener(tel);
+        database.insertEntry(shouldBeEntry);
+        tel = new TestEventListener();
```
tobiasdiez: Just register the listener after inserting the entry, then no reinit of the listener is needed.

tobiasdiez:
The code looks good. 
@JabRef/developers do we really need a devcall here? It is not an eventbus in the sense that there exists a global event dispatcher but the eventbus just presents a simpler way to throw events per object. 

I'm :+1: for merging.

src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
obraliar: The problem here was to avoid already set `bibEntry` in `TestEventListener` while calling `removeEntry()`. Is we register it again, the same listener would be used again.

src/main/java/net/sf/jabref/gui/PreviewPanel.java
obraliar: Yes it is necessary, cause there is only one `PreviewPanel` for all entries. There should be a possibility to replace the listener and to set new listener for new entries. Furthermore passing small objects (like `ChangeFieldUpdateEvent`) is more efficient than passing the hole `PreviewPanel` object.

src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
tobiasdiez: Yes, but this problem can also be circumvented by calling the methods in this order
````
database.insertEntry(...);
TestEventListener listener = new ...;
database.registerListener()

database.remove()

assertEquals()
````

**Comment by obraliar** [comment]
Redy for squashing the commits?

File: src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java (hunk 29 above)
Comment (obraliar) [inline src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java]: Done.

File: src/main/java/net/sf/jabref/event/ChangeFieldEvent.java [hunk 30]
```diff
@@ -0,0 +1,22 @@
+package net.sf.jabref.event;
+
+
+public class ChangeFieldEvent {
+
+    private final String fieldName;
+    private final String value;
+
+    public ChangeFieldEvent(String fieldName, String value) {
```
koppor: Could you rename `value` to `newValue`? Also at the getter? Otherwise, it is not straight forward understandable whether it is the new value or the old value.

(One could ask for adding `oldValue` and `bibEntry` as paramter. Since this is not yet required, we can add it )later.

src/main/java/net/sf/jabref/gui/PreviewPanel.java
koppor: Maybe add this as JavaDoc comment above this place.

src/main/java/net/sf/jabref/event/ChangeFieldEvent.java
obraliar: Done.

src/main/java/net/sf/jabref/gui/entryeditor/EntryEditor.java
```diff
@@ -921,15 +927,13 @@ public void updateAllContentSelectors() {
...
-    /**
-     * Update the JTextArea when a field has changed.
-     *
-     * @see java.beans.VetoableChangeListener#vetoableChange(java.beans.PropertyChangeEvent)
-     */
-    @Override
-    public void vetoableChange(PropertyChangeEvent e) {
-        String newValue = e.getNewValue() == null ? "" : e.getNewValue().toString();
-        setField(e.getPropertyName(), newValue);
+    private class ChangeFieldListener {
```
koppor: Why not keeping the old JavaDoc comment "Update the JTextArea when a field has changed."?

src/main/java/net/sf/jabref/model/database/BibDatabase.java
```diff
@@ -119,6 +119,14 @@ public boolean containsEntryWithId(String id) {
         return internalIDs.contains(id);
     }
 
+    /**
+     * Constructor
```
koppor: This is obsolete, isn't it?

src/main/java/net/sf/jabref/model/entry/BibEntry.java
```diff
@@ -189,6 +167,20 @@ public void setType(EntryType type) {
     }
 
     /**
+     * Sets this entry's ID, provided the database containing it
```
koppor: Why is the code moved in this class? Could you keep the old position? I think, the move was on purpose. See https://github.com/JabRef/jabref/pull/1309

src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
```diff
@@ -187,4 +188,27 @@ public void addSameStringIdTwiceThrowsKeyCollisionException() {
...
+    @Test
+    public void testRemoveEntryEventReceivement() {
+        BibDatabase database = new BibDatabase();
+        BibEntry shouldBeEntry = new BibEntry();
+        TestEventListener tel = new TestEventListener();
+        database.insertEntry(shouldBeEntry);
+        database.registerListener(tel);
+        database.removeEntry(shouldBeEntry);
+        BibEntry isEntry = tel.getBibEntry();
+        assertEquals(shouldBeEntry, isEntry);
```
koppor: could you use `expected` and `actual` as prefixes instead of `shouldBe` and `is`? I know, that old JabRef code uses `shouldBe` and `asIs`, but this is not in line with JUnit.

src/main/java/net/sf/jabref/gui/entryeditor/EntryEditor.java
obraliar: Done.

src/main/java/net/sf/jabref/model/database/BibDatabase.java
obraliar: Removed constructor an added attribute definition instead.

src/main/java/net/sf/jabref/model/entry/BibEntry.java
obraliar: Done. Moved to old position.

src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
obraliar: Done.

src/main/java/net/sf/jabref/gui/PreviewPanel.java
tobiasdiez: Sorry, but I still don't get why the private class is necessary. As soon as some entry is changed you call PreviewPanel.update. Why should this call be in a private subclass? Can you please provide an example.

src/main/java/net/sf/jabref/event/ChangeFieldEvent.java
```diff
@@ -0,0 +1,22 @@
...
+    public ChangeFieldEvent(String fieldName, String newValue) {
+        this.fieldName = fieldName;
+        this.value = newValue;
+    }
+
+    public String getFieldName() {
+        return fieldName;
+    }
+
+    public String getValue() {
```
tobiasdiez: getNewValue and rename private variable.

src/main/java/net/sf/jabref/event/AddEntryEvent.java
```diff
@@ -0,0 +1,18 @@
+package net.sf.jabref.event;
+
+import net.sf.jabref.model.entry.BibEntry;
+
+/**
+ * Instances of this class are suitable for passing through the event bus.
```
tobiasdiez: The first comment is a bit superfluous. The second is more appropriate as a `@param` comment for the constructor. I would just keep the last line, i.e. `Event fired when a BibEntry was added to the db`

src/main/java/net/sf/jabref/event/ChangeEntryEvent.java
```diff
@@ -0,0 +1,18 @@
+package net.sf.jabref.event;
+
+import net.sf.jabref.model.entry.BibEntry;
+
+/**
+ * Instances of this class are suitable for passing through the event bus.
+ * A <code>BibEntry</code> object the changes were applied on is required.
```
tobiasdiez: Same as above.

src/main/java/net/sf/jabref/event/ChangeFieldEvent.java
```diff
@@ -0,0 +1,22 @@
+package net.sf.jabref.event;
+
+
+public class ChangeFieldEvent {
```
tobiasdiez: Add doc.

src/main/java/net/sf/jabref/specialfields/SpecialFieldUpdateListener.java
```diff
@@ -15,30 +15,28 @@
...
 
     private static SpecialFieldUpdateListener INSTANCE;
 
-
-    @Override
-    public void vetoableChange(PropertyChangeEvent e) throws PropertyVetoException {
-        final BibEntry entry = (BibEntry) e.getSource();
-        final String fieldName = e.getPropertyName();
+    @Subscribe
+    public void listen(ChangeSpecialFieldEvent changeSpecialFieldEvent) {
```
tobiasdiez: I cant find any instance where this event is actually fired.

src/main/java/net/sf/jabref/event/EntryEvent.java
```diff
@@ -0,0 +1,16 @@
+package net.sf.jabref.event;
+
+import net.sf.jabref.model.entry.BibEntry;
+
+public abstract class EntryEvent {
```
tobiasdiez: doc

src/main/java/net/sf/jabref/event/RemoveEntryEvent.java
```diff
@@ -0,0 +1,18 @@
+package net.sf.jabref.event;
+
+import net.sf.jabref.model.entry.BibEntry;
+
+/**
+ * Instances of this class are suitable for passing through the event bus.
```
tobiasdiez: same as above

src/main/java/net/sf/jabref/gui/PreviewPanel.java
obraliar: Have a look at `public void setEntry(BibEntry newEntry)`. This could help to understand why this private class is needed.

src/main/java/net/sf/jabref/gui/PreviewPanel.java
```diff
@@ -275,16 +275,35 @@ public void setLayout(Layout layout) {
...
 
     public void setEntry(BibEntry newEntry) {
-        if(entry.isPresent() && (entry.get() != newEntry)) {
-            entry.ifPresent(e -> e.removePropertyChangeListener(this));
-            newEntry.addPropertyChangeListener(this);
+
+        if (entry.isPresent() && (entry.get() != newEntry)) {
+            entry.ifPresent(e -> e.unregisterListener(this.currentChangeFieldUpdateEvent));
+            this.currentChangeFieldUpdateEvent = new ChangeFieldUpdateEvent();
+            newEntry.registerListener(this.currentChangeFieldUpdateEvent);
```
tobiasdiez: I think, the registerListener should be moved after the if statement.

src/main/java/net/sf/jabref/gui/PreviewPanel.java
tobiasdiez: No it doesnt. There I simply would write 
````java
entry.ifPresent(e -> e.unregisterListener(this));
newEntry.registerListener(this);
````
The same comment applies also to the EntryEditor.

Comment (obraliar) [inline src/main/java/net/sf/jabref/gui/PreviewPanel.java]: Now I can understand You. Well this is possible but is the PreviewPanel object not too big for passing? I don't know how guava manages this... However I saw the following comment (https://github.com/JabRef/jabref/blob/master/src/main/java/net/sf/jabref/gui/PreviewPanel.java#L322). Now I cannot reconstruct the case where `vetoableChange(...)` method is called here. Probably we don't need such a listener here at all.

File: src/main/java/net/sf/jabref/event/ChangeFieldEvent.java (hunk 38 above)
Comment (koppor) [inline src/main/java/net/sf/jabref/event/ChangeFieldEvent.java]: Should we rename it to `FieldAddedChangedRemovedEvent`?

File: src/main/java/net/sf/jabref/gui/PreviewPanel.java (hunk 42 above)
Comment (koppor) [inline src/main/java/net/sf/jabref/gui/PreviewPanel.java]: With a check if newEntry is not null.

File: src/main/java/net/sf/jabref/gui/PreviewPanel.java [hunk 43]
```diff
@@ -275,16 +275,35 @@ public void setLayout(Layout layout) {
...
+
         entry = Optional.ofNullable(newEntry);
 
         updateLayout();
         update();
     }
 
+
+    /**
+    * Listener class for ChangeFieldUpdateEvent.
```
koppor: The name is confusing. Why is "Change" and "Update" in the name? Is it only "Change"? The event should be called `FieldAddedChangedRemovedEvent` and this listener `FieldAddedChangedRemovedEventListener` (meaning `Listener` added to the name of the event class).

src/main/java/net/sf/jabref/event/ChangeFieldEvent.java
koppor: I vote for renaming to `FieldAddedChangedRemovedEvent`, because this event is also fired when a field is removed or added.

src/main/java/net/sf/jabref/specialfields/SpecialFieldUpdateListener.java
koppor: It should listen for `ChangeFieldEvent`, not for this special class. Please remove the class `ChangeSpecialFieldEvent`.

src/main/java/net/sf/jabref/event/ChangeSpecialFieldEvent.java
```diff
@@ -0,0 +1,22 @@
+package net.sf.jabref.event;
+
+import net.sf.jabref.model.entry.BibEntry;
+
+public class ChangeSpecialFieldEvent {
```
koppor: This class can be removed. See https://github.com/JabRef/jabref/pull/1028/files#r62419955

src/main/java/net/sf/jabref/event/ChangeFieldEvent.java
tobiasdiez: Vote against it. Adding a field is changing it from nothing to something, similar for removed. Thus I think `FieldChangedEvent` is enough.

src/main/java/net/sf/jabref/gui/PreviewPanel.java
tobiasdiez: No its not to big...java passes objects by reference (kind of).

src/main/java/net/sf/jabref/event/ChangeSpecialFieldEvent.java
obraliar: Removed and replaced with existing `ChangedFieldEvent`.

src/main/java/net/sf/jabref/gui/PreviewPanel.java
obraliar: Done.

src/main/java/net/sf/jabref/event/RemoveEntryEvent.java
obraliar: All done.

src/main/java/net/sf/jabref/event/ChangedFieldEvent.java
```diff
@@ -0,0 +1,37 @@
+package net.sf.jabref.event;
+
+import net.sf.jabref.model.entry.BibEntry;
+
+/**
+ * <code>ChangedFieldEvent</code> is fired when a field of <code>BibEntry</code> has been modified.
```
koppor: Or removed. Or added. See `eventBus.post(new ChangedFieldEvent(this, fieldName, null));`. Please add this as comment.

src/main/java/net/sf/jabref/event/ChangedFieldEvent.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+public class ChangedFieldEvent {
```
koppor: Please extend ChangedEntryEvent and do the necessary modifications (calling super, removing the local bibEntry variable and removing getBibEntry).

src/main/java/net/sf/jabref/event/ChangedFieldEvent.java
obraliar: Done.

src/main/java/net/sf/jabref/event/ChangedFieldEvent.java
obraliar: Done.

koppor:
[LGTM](http://lgtm.in/i/qjdMkyEnq)

src/main/java/net/sf/jabref/gui/PreviewPanel.java
```diff
@@ -275,16 +274,28 @@ public void setLayout(Layout layout) {
     }
 
     public void setEntry(BibEntry newEntry) {
-        if(entry.isPresent() && (entry.get() != newEntry)) {
-            entry.ifPresent(e -> e.removePropertyChangeListener(this));
-            newEntry.addPropertyChangeListener(this);
+
+        if (entry.isPresent() && (entry.get() != newEntry)) {
+            entry.ifPresent(e -> e.unregisterListener(this));
         }
```
simonharrer: could be written simpler: `entry.filter(e -> e != newEntry).ifPresent(e -> e.unregisterListener(this));`

tobiasdiez:
@JabRef/stupro could you please add some short notes to the wiki on how to push and subscribe to events (using the eventbus). Thanks!

obraliar:
Done. See https://github.com/JabRef/jabref/wiki/Code-Howtos#using-the-eventsystem

PR #1029, opened 2016-03-23
Better main table rendering

oscargus:
The main table doesn't always look that nice with latex commands in it. This PR makes things much better.

Potential drawback is rendering time, but I am not sure how much of a problem it actually is.

clipboard03


tobiasdiez:
LGTM

matthiasgeiger:
Have not that much time but there might be one problem: Some users have configured the table to show "author/editor". After a quick look at your changes I suppose that this might be troublesome?

oscargus:
If anything it will work even better. At least once InternalBibtexFields is updated to cover all the different Biblatex fields that contains a list of person names. At the moment there shouldn't be any difference as I can see.

matthiasgeiger:
Okay. Then: Go for it! :wink:

PR #1030, opened 2016-03-24
Got rid of some null return values and added a small test

oscargus:
* Replaced null return values with Optional (and emptyList())
* Added test for openoffice.CitationManager.CitEntry
* Removed unused method

- [x] Tests green?

### DIFF
--- a/src/main/java/net/sf/jabref/openoffice/CitationManager.java
+++ b/src/main/java/net/sf/jabref/openoffice/CitationManager.java
@@ -145,11 +145,21 @@ class CitationManager {
         private final Optional<String> origPageInfo;
 
 
-        public CitEntry(String refMarkName, String context, Optional<String> optional) {
+        // Only used for testing...
+        public CitEntry(String refMarkName, String context) {
+            this(refMarkName, context, Optional.empty());
+        }
+
+        // Only used for testing...
+        public CitEntry(String refMarkName, String context, String pageInfo) {
+            this(refMarkName, context, Optional.ofNullable(pageInfo));
+        }
+
+        public CitEntry(String refMarkName, String context, Optional<String> pageInfo) {
             this.refMarkName = refMarkName;
             this.context = context;
-            this.pageInfo = optional;
-            this.origPageInfo = optional;
+            this.pageInfo = pageInfo;
+            this.origPageInfo = pageInfo;
         }
 
         public Optional<String> getPageInfo() {
@@ -196,9 +206,8 @@ class CitationManager {
             return context;
         }
 
-        public void setPageInfo(Optional<String> trim) {
-            pageInfo = trim;
-
+        public void setPageInfo(String trim) {
+            pageInfo = Optional.ofNullable(trim);
         }
     }
 
@@ -278,9 +287,9 @@ class CitationManager {
 
             okButton.addActionListener(e -> {
                 if (pageInfo.getText().trim().isEmpty()) {
-                    entry.setPageInfo(Optional.empty());
+                    entry.setPageInfo(null);
                 } else {
-                    entry.setPageInfo(Optional.of(pageInfo.getText().trim()));
+                    entry.setPageInfo(pageInfo.getText().trim());
                 }
                 tableModel.fireTableDataChanged();
                 singleCiteDialog.dispose();
--- a/src/main/java/net/sf/jabref/openoffice/CitationManager.java
+++ b/src/main/java/net/sf/jabref/openoffice/CitationManager.java
@@ -140,20 +140,20 @@ class CitationManager {
     static class CitEntry implements Comparable<CitEntry> {
 
         private final String refMarkName;
-        private String pageInfo;
+        private Optional<String> pageInfo;
         private final String context;
-        private final String origPageInfo;
+        private final Optional<String> origPageInfo;
 
 
-        public CitEntry(String refMarkName, String context, String pageInfo) {
+        public CitEntry(String refMarkName, String context, Optional<String> optional) {
             this.refMarkName = refMarkName;
             this.context = context;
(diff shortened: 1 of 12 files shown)


src/main/java/net/sf/jabref/openoffice/CitationManager.java
```diff
@@ -140,20 +140,20 @@ public void showDialog() {
...
         private final String refMarkName;
-        private String pageInfo;
+        private Optional<String> pageInfo;
         private final String context;
-        private final String origPageInfo;
+        private final Optional<String> origPageInfo;
 
 
-        public CitEntry(String refMarkName, String context, String pageInfo) {
+        public CitEntry(String refMarkName, String context, Optional<String> optional) {
```
tobiasdiez: rename back to pageInfo?

src/test/java/net/sf/jabref/importer/EntryFromPDFCreatorTest.java
```diff
@@ -47,13 +48,13 @@ public void testPDFFileFilter() {
     @Test
     @Ignore
     public void testCreationOfEntry() {
-        BibEntry entry = entryCreator.createEntry(ImportDataTest.NOT_EXISTING_PDF, false);
-        Assert.assertNull(entry);
+        Optional<BibEntry> entry = entryCreator.createEntry(ImportDataTest.NOT_EXISTING_PDF, false);
+        Assert.assertFalse(entry.isPresent());
 
         entry = entryCreator.createEntry(ImportDataTest.FILE_NOT_IN_DATABASE, false);
```
tobiasdiez: Could you please use the opportunity and split these tests into two. Thanks.

src/test/java/net/sf/jabref/openoffice/CitationManagerTest.java
```diff
@@ -0,0 +1,77 @@
...
+        assertEquals("Other info", citEntry.getPageInfo().get());
+    }
+
+    @Test
+    public void testCitEntryEquals() {
+        CitEntry citEntry1 = new CitEntry("RefMark", "Context", Optional.of("Info"));
+        CitEntry citEntry2 = new CitEntry("RefMark2", "Context", Optional.of("Info"));
+        CitEntry citEntry3 = new CitEntry("RefMark", "Other Context", Optional.of("Other Info"));
+        assertTrue(citEntry1.equals(citEntry1));
+        assertTrue(citEntry1.equals(citEntry3));
```
tobiasdiez: 1 equals 3? Looks kind of strange.

tobiasdiez:
LGTM

src/main/java/net/sf/jabref/openoffice/CitationManager.java
stefan-kolb: I think this is no good use of Optional. This should be done by overlaoding the method.

stefan-kolb: https://dzone.com/articles/java-8-optional-use-cases

stefan-kolb:
Other than the use of Optional in CitationManager.java OK to merge.

src/main/java/net/sf/jabref/openoffice/CitationManager.java
oscargus: I wouldn't have changed it unless I changed the method generating the argument. If I get you correct there should be two different constructors? One with Optional (for convenience) and one for String (for actual use)?

oscargus: Indeed it should read pageInfo...

src/test/java/net/sf/jabref/openoffice/CitationManagerTest.java
oscargus: Only the first argument is used for comparison, so I thought it was a good idea to create two different CitEntry with identical first arguments...

src/main/java/net/sf/jabref/openoffice/CitationManager.java
tobiasdiez: No, I think Stefan meant one constructor with String pageInfo and one without any pageInfo. I can only second this suggestion.

src/test/java/net/sf/jabref/openoffice/CitationManagerTest.java
tobiasdiez: Ok...if it makes sense to compare citentries only by the first key...

src/main/java/net/sf/jabref/openoffice/CitationManager.java
oscargus: OK, the thing is this:

``` java
            list.add(new CitEntry(name,
                    "<html>..." + ooBase.getCitationContext(nameAccess, name, 30, 30, true) + "...</html>",
                    ooBase.getCustomProperty(name)));
```

since getCustomProperty now returns an Optional it makes sense to have an Optional as a parameter, if nothing else for brevity.

src/test/java/net/sf/jabref/openoffice/CitationManagerTest.java
oscargus: Yeah, that I have no idea about... My guess is that it is used for sorting a table and then this is a quicker way (the refMarks should be separate anyway, although I'm not sure what happens if you create two with the same name...).

src/main/java/net/sf/jabref/openoffice/CitationManager.java
oscargus: I have added two more constructor, one (String, String), one (String, String, String). However, those are only used for testing. The only place where a constructor is used in the code is the above one and then I do not think it makes sense to add a .orElse(null) to remove the Optional-constructor.

PR #1032, opened 2016-03-24
Extracted field updating methods out of util

oscargus:
Moved methods related to updating fields and timestamp from Util to logic. Since NamedCompounds were passed, the methods were changed to return FieldChange.

I am not convinced that the current naming/position of the new classes is optimal, so input is welcome.

- [x] Tests green?


Siedlerchr:
Looks good. As there have been Date and  Time API Changes in Java 8, maybe you can use some of them:
http://javarevisited.blogspot.de/2015/03/20-examples-of-date-and-time-api-from-Java8.html

oscargus:
I added an issue regarding replacing SimpleDateFormat. It was used in 23 places if I remember correctly. Good link!

PR #1035, opened 2016-03-25
Tests of BibEntry and BibDatabase

oscargus:
Added more tests and fixed a few issues.

Fixed(?) a Coverity issue in AboutDialog.

### DIFF
--- a/src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
+++ b/src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
@@ -49,7 +49,7 @@ public class BibDatabaseTest {
     @Test
     public void resolveStrings() throws IOException {
         try (FileInputStream stream = new FileInputStream("src/test/resources/net/sf/jabref/util/twente.bib");
-                InputStreamReader fr = new InputStreamReader(stream, StandardCharsets.ISO_8859_1)) {
+                InputStreamReader fr = new InputStreamReader(stream, StandardCharsets.UTF_8)) {
             ParserResult result = BibtexParser.parse(fr);
 
             BibDatabase db = result.getDatabase();
@@ -84,7 +84,7 @@ public class BibDatabaseTest {
         assertTrue(database.containsEntryWithId(entry.getId()));
     }
 
-    @Test
+    @Test(expected = KeyCollisionException.class)
     public void insertEntryWithSameIdThrowsException() {
         BibDatabase database = new BibDatabase();
 
@@ -92,8 +92,8 @@ public class BibDatabaseTest {
         database.insertEntry(entry0);
 
         BibEntry entry1 = new BibEntry(entry0.getId());
-        thrown.expect(KeyCollisionException.class);
         database.insertEntry(entry1);
+        fail();
     }
 
     @Test
@@ -108,25 +108,30 @@ public class BibDatabaseTest {
         assertFalse(database.containsEntryWithId(entry.getId()));
     }
 
-    @Test
+    @Test(expected = NullPointerException.class)
     public void insertNullEntryThrowsException() {
         BibDatabase database = new BibDatabase();
-        thrown.expect(NullPointerException.class);
         database.insertEntry(null);
+        fail();
+    }
+
+    @Test(expected = NullPointerException.class)
+    public void removeNullEntryThrowsException() {
+        BibDatabase database = new BibDatabase();
+        database.removeEntry(null);
+        fail();
     }
 
     @Test
-    public void removeNullEntryThrowsException() {
+    public void emptyDatabaseHasNoStrings() {
         BibDatabase database = new BibDatabase();
-        thrown.expect(NullPointerException.class);
-        database.removeEntry(null);
+        assertEquals(Collections.emptySet(), database.getStringKeySet());
+        assertTrue(database.hasNoStrings());
     }
 
     @Test
     public void insertString() {
         BibDatabase database = new BibDatabase();
-        assertEquals(Collections.emptySet(), database.getStringKeySet());
-        assertTrue(database.hasNoStrings());
         BibtexString string = new BibtexString(IdGenerator.next(), "DSP", "Digital Signal Processing");
         database.addString(string);
         assertFalse(database.hasNoStrings());
@@ -161,7 +166,7 @@ public class BibDatabaseTest {
     }
 
     @Test(expected = KeyCollisionException.class)
-    public void addSameStringLabelTwice() {
+    public void addSameStringLabelTwiceThrowsKeyCollisionException() {
         BibDatabase database = new BibDatabase();
         BibtexString string = new BibtexString(IdGenerator.next(), "DSP", "Digital Signal Processing");
         database.addString(string);
@@ -171,7 +176,7 @@ public class BibDatabaseTest {
     }
 
     @Test(expected = KeyCollisionException.class)
-    public void addSameStringIdTwice() {
+    public void addSameStringIdTwiceThrowsKeyCollisionException() {
         BibDatabase database = new BibDatabase();
         String id = IdGenerator.next();
         BibtexString string = new BibtexString(id, "DSP", "Digital Signal Processing");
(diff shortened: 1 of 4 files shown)


src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
```diff
@@ -43,7 +48,8 @@ public void tearDown() {
      */
     @Test
     public void resolveStrings() throws IOException {
-        try (FileReader fr = new FileReader("src/test/resources/net/sf/jabref/util/twente.bib")) {
+        try (FileInputStream stream = new FileInputStream("src/test/resources/net/sf/jabref/util/twente.bib");
+                InputStreamReader fr = new InputStreamReader(stream, StandardCharsets.ISO_8859_1)) {
```
Siedlerchr: Is the StandardCharset really ISO and not UTF of that file?

oscargus: I see now that the encoding is stated as UTF in the file. However since ISO is the first code page of UTF it is hard to actually determine the difference here... (And if I'm not wrong US_ASCII would also have worked for this file...)

I'll change it if there are other things to change as it is working and is sort of correct.

src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
```diff
@@ -115,4 +122,62 @@ public void removeNullEntryThrowsException() {
         database.removeEntry(null);
     }
 
+    @Test
+    public void insertString() {
+        BibDatabase database = new BibDatabase();
+        assertEquals(Collections.emptySet(), database.getStringKeySet());
```
tobiasdiez: I would move these two initial assert to a new test `newDatabaseHasNoStrings`

src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
```diff
@@ -115,4 +122,62 @@ public void removeNullEntryThrowsException() {
...
+    }
+
+    @Test(expected = KeyCollisionException.class)
+    public void addSameStringLabelTwice() {
+        BibDatabase database = new BibDatabase();
+        BibtexString string = new BibtexString(IdGenerator.next(), "DSP", "Digital Signal Processing");
+        database.addString(string);
+        string = new BibtexString(IdGenerator.next(), "DSP", "Digital Signal Processor");
+        database.addString(string);
+        fail();
```
tobiasdiez: fail is not required here I think

src/test/java/net/sf/jabref/model/entry/BibtexEntryTests.java
```diff
@@ -84,6 +86,22 @@ public void isEmptyCiteKey() {
     }
 
     @Test
+    public void typeOfBibEntryIsMiscAfterSettingToNullString() {
```
tobiasdiez: I would prefer if the entry is setup in the SetUp method using `setField` methods and not parsed from a string.

src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
oscargus: I will consistently add those as Codacy reports it as an issue. Conceptually you are correct though.

src/test/java/net/sf/jabref/model/entry/BibtexEntryTests.java
```diff
@@ -108,6 +126,78 @@ public void testGetPublicationDate() {
...
+    public void getFieldOrAliasMonthWithDateYYYYMMDD() {
+        BibEntry be = BibtexParser.singleFromString("@ARTICLE{HipKro03, date = {2003-03-30}}");
+        Assert.assertEquals("3", be.getFieldOrAlias("month"));
+    }
+
+    @Test(expected = NullPointerException.class)
+    public void setNullField() {
+        BibEntry be = BibtexParser.singleFromString("@ARTICLE{Key15, keywords = {Foo, Bar}}");
+        be.setField(null);
+        Assert.fail();
```
tobiasdiez: fail should not be required.

src/test/java/net/sf/jabref/model/entry/BibtexEntryTests.java
```diff
@@ -116,40 +206,116 @@ public void addNullKeyword() {
...
+    }
+
     @Test
-    public void testKeywordMethods() {
+    public void testGetSeparatedKeywordsAreCorrect() {
         BibEntry be = BibtexParser.singleFromString("@ARTICLE{Key15, keywords = {Foo, Bar}}");
 
         String[] expected = {"Foo",  "Bar"};
         Assert.assertArrayEquals(expected, be.getSeparatedKeywords().toArray());
+        Assert.assertFalse(be.hasChanged());
```
tobiasdiez: I would move these hasChanged test to a new test

src/test/java/net/sf/jabref/model/entry/BibtexEntryTests.java
```diff
@@ -116,40 +206,116 @@ public void addNullKeyword() {
...
 
-        List<String> kw = be.getSeparatedKeywords();
+    @Test
+    public void testAddKeywordIsCorrect() {
+        BibEntry be = BibtexParser.singleFromString("@ARTICLE{Key15, keywords = {Foo, Bar}}");
 
         be.addKeyword("FooBar");
         String[] expected2 = {"Foo", "Bar", "FooBar"};
         Assert.assertArrayEquals(expected2, be.getSeparatedKeywords().toArray());
+        Assert.assertTrue(be.hasChanged());
```
tobiasdiez: Same

src/test/java/net/sf/jabref/model/entry/BibtexEntryTests.java
```diff
@@ -116,40 +206,116 @@ public void addNullKeyword() {
...
+    }
+
+    @Test
+    public void testAddKeywordTwiceYiedsOnlyOne() {
+        BibEntry be = BibtexParser.singleFromString("@ARTICLE{Key15, keywords = {Foo, Bar}}");
 
         be.addKeyword("FooBar");
+        be.addKeyword("FooBar");
+        String[] expected2 = {"Foo", "Bar", "FooBar"};
```
tobiasdiez: nameing without the last 2

src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
tobiasdiez: Couldn't you disable the corresponding codacy rule instead?

oscargus: That is indeed an idea. The question is just what is worse? Adding tests without asserts or an additional line?

oscargus:
I've fixed all(?) comments. One thing happened though. What is the difference between the two BibEntry under test here? (The test shouldn't look like this, but I kept both for now.)

``` java
    @Test
    @Ignore
    public void getFieldOrAliasDateWithYearNumericalMonth() {
        Assert.assertEquals("2003-03",
                (BibtexParser.singleFromString("@ARTICLE{HipKro03, year = {2003}, month = 3 }")).getFieldOrAlias("date"));
        emptyEntry.setField("year", "2003");
        emptyEntry.setField("month", "#3#");
        Assert.assertEquals("2003-03", emptyEntry.getFieldOrAlias("date"));
    }
```
The first assertion passes, but not the second. Seems to be something with the parser?

tobiasdiez:
What happens if you remove the # in the month field?

oscargus:
Then it works, but that is a different test case. (I was surprised that the original test case worked, I only saw it in another test a few lines above.)

oscargus:
I have the other test case `getFieldOrAliasDateWithYearNumericalMonthString`

tobiasdiez:
But `#3#` is not a valid month format, or is it? as `getFieldOrAliasDateWithYearAbbreviatedMonth` works, I don't see a problem here.

oscargus:
Maybe there isn't a problem. However, if the corresponding statement was in
a .bib-file it would apparently be parsed as 3 and not #3# .

I remove that test then.

PR #1039, opened 2016-03-25
Added more testing to OOBibStyle and fixed a few issues

oscargus:
More testing leading to more understanding and a few fixes. (And, no, the tests are maybe not the most elegant, but it was more a matter of figuring out how the code behaves in certain situations and try to trigger as many situations as possible...).

### DIFF
--- a/src/main/java/net/sf/jabref/openoffice/OOBibStyle.java
+++ b/src/main/java/net/sf/jabref/openoffice/OOBibStyle.java
@@ -432,7 +432,7 @@ class OOBibStyle implements Comparable<OOBibStyle> {
             int i1 = lNum.get(i);
             if (combineFrom < 0) {
                 // Check if next entry is the next in the ref list:
-                if ((i < (lNum.size() - 1)) && (lNum.get(i + 1) == (i1 + 1))) {
+                if ((i < (lNum.size() - 1)) && (lNum.get(i + 1) == (i1 + 1)) && (i1 > 0)) {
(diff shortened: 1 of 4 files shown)


src/main/java/net/sf/jabref/openoffice/OOBibStyle.java
```diff
@@ -705,15 +705,10 @@ private String getCitationMarkerField(BibEntry entry, BibDatabase database, Stri
...
 
             if ((content != null) && !content.trim().isEmpty()) {
-                if (fieldFormatter != null) {
-
-                    if (field.equals(authorField) && StringUtil.isInCurlyBrackets(content)) {
-                        return "{" + fieldFormatter.format(content) + "}";
-                    }
-                    return fieldFormatter.format(content);
+                if (field.equals(authorField) && StringUtil.isInCurlyBrackets(content)) {
+                    return "{" + fieldFormatter.format(content) + "}";
```
Siedlerchr: The check `if fieldFormatter != null` was obsolet?

oscargus: Yes, the fieldFormatter is initialized on construction. I think it is a
left-over from the plugin time.

(Of course, the initialization can fail, but then one should null check
after every new-call...)

Siedlerchr: Ah okay! Thanks for clearing that out.

PR #1045, opened 2016-03-26
Move Set/clear/rename fields to edit

tobiasdiez:
Move the function `Set clear rename fields` to edit from search. Since it is definitely an edit function and has nothing to do with searches.


oscargus:
If it was called "Find/replace" it would belong in search. ;-) (I guess that is the historical reason, but I agree that Edit is better.)

lenhard:
No objections here. Can you do a rebase, so we can merge it for v3.3?

PR #1046, opened 2016-03-26
Refactor AuthorList

tobiasdiez:
- Refactor AuthorList by extracting Author class and move parsing logic to new AuthorListParser
- Rename a few methods in AuthorList
- Add a few tests taken from biber https://github.com/plk/biber/blob/dev/t/names.t (two of them actually fail, so maybe somebody with better understanding of the inner workings of AuthorList may fix them)

--


src/main/java/net/sf/jabref/logic/autocompleter/NameFieldAutoCompleter.java
```diff
@@ -93,7 +94,7 @@ public void addBibtexEntry(BibEntry entry) {
             if (entry.hasField(fieldName)) {
                 String fieldValue = entry.getField(fieldName);
                 AuthorList authorList = AuthorList.getAuthorList(fieldValue);
-                for (AuthorList.Author author : authorList.getAuthorList()) {
+                for (Author author : authorList.getAuthorList()) {
```
Siedlerchr: AuthorList.getAuthorList() sounds really redundant from naming.
Together with the next line it is confusing on the first read.
What about a simple  `getAuthors()`

src/main/java/net/sf/jabref/logic/xmp/XMPUtil.java
```diff
@@ -626,11 +626,10 @@ private static void writeToDCSchema(XMPSchemaDublinCore dcSchema,
              */
             if ("author".equals(field)) {
                 String authors = resolvedEntry.getField(field);
-                AuthorList list = AuthorList.getAuthorList(authors);
+                AuthorList list = AuthorList.getAuthors(authors);
 
-                int n = list.size();
-                for (int i = 0; i < n; i++) {
-                    dcSchema.addCreator(list.getAuthor(i).getFirstLast(false));
```
oscargus: So the one-liner to get a List <Auhtor> is

AuthorList.getAuthors(String).getAuthors()?

Maybe not a huge difference but I sort of liked the previous naming better...

src/main/java/net/sf/jabref/logic/xmp/XMPUtil.java
```diff
@@ -626,11 +626,10 @@ private static void writeToDCSchema(XMPSchemaDublinCore dcSchema,
...
+                for (Author author : list.getAuthors()) {
```
oscargus: At least the first call could still well be getAuthorList() as that is more accurate.

src/main/java/net/sf/jabref/model/entry/AuthorList.java
```diff
@@ -768,18 +392,18 @@ public String getAuthorsLastFirst(boolean abbreviate, boolean oxfordComma) {
         }
 
         StringBuilder result = new StringBuilder();
-        if (size() > 0) {
+        if (getNumberOfAuthors() > 0) {
```
oscargus: Is there an `isEmpty()` (or corresponding) method? Maybe there should be.

oscargus:
Looks good! Have you tried the formatter in bst and seen if it formats incorrectly as well? I think that we should try to use that as much as possible instead of trying to reinvent it.

The only major comment is that getAuthors return different types. I think it makes sense to keep getAuthorList (or use a different name).

src/main/java/net/sf/jabref/logic/autocompleter/NameFieldAutoCompleter.java
tobiasdiez: This is actually done in a latter commit.

src/main/java/net/sf/jabref/logic/xmp/XMPUtil.java
tobiasdiez: previously it was AuthorList.getAuthorList.getAuthorList...has also a duplication. Mhh...what do you think about renaming the first getAuthor to parse? I.e `AuthorList.parse(String).getAuthors`. That the parsing is actually cached shouldn't be a worry for the name.

src/main/java/net/sf/jabref/logic/autocompleter/NameFieldAutoCompleter.java
oscargus: OK! Sounds good! I use some tool on the phone to check it so maybe that has
problems. I can for example not see your comment in the tool...

src/main/java/net/sf/jabref/model/entry/AuthorList.java
tobiasdiez: Yes there is an isEmpty...should be used here you are right.

src/main/java/net/sf/jabref/logic/xmp/XMPUtil.java
oscargus: I think I introduced the second getAuthorList... Didn't think of it then...

Or getAuthorList().getListOfAuhtors() which in some sense is more
self-explanatory but maybe also open for more confusion. Your suggestion
works for me, so as long as they are different I'm OK with it.

tobiasdiez: I now renamed it to `parse` which reads quite well in my opinion.

oscargus: :+1:

PR #1047, opened 2016-03-26
Refactoring of OO connection

oscargus:
I'm restructuring the OO connection with the following goals:

 - [x] Split the files between gui and logic
 - [x] Change the style selection dialog so that it is easier to support many internal styles as well as adding/removing external styles

Long term (probably not in this PR)
 - [x] Improve the mechanism that searches for binaries and .jars

A question:
Should classes that accesses the OO-specific jars reside in gui or logic (assuming that they do not access gui or swing, so they could reside in logic from a test perspective)?


src/main/java/net/sf/jabref/openoffice/OpenOfficeFileSearch.java
```diff
@@ -0,0 +1,116 @@
...
+     * style files, and add them. The parameter recurse determines whether we should
+     * recurse into subdirectories.
+     * @param dir the directory or file to handle.
+     * @param recurse true indicates that we should recurse into subdirectories.
+     * @param encoding
+     */
+    public void addStyles(String dir, boolean recurse, Charset encoding, List<OOBibStyle> styles) {
+        File dirF = new File(dir);
+        if (dirF.isDirectory()) {
+            File[] fileArray = dirF.listFiles();
```
Siedlerchr: You could make use of the new JAVA 8 Feature for a Directory as Stream:
 java.nio.file.Files list(Path) returns a Stream<Path> and loads the content successive on demand, only when needed. Also useful for searching specified files:
http://eherrera.net/ocpj8-notes/09-java-file-io-%28NIO.2%29.html#use-stream-api-with-nio-2

oscargus: Thanks!

My plan right now is to skip this "get all files in a directory subtree"
and require the user to point at every file (once). All available style
files (including those supplied by default, which are planned to increase
in number) will be available in a single list with add/remove buttons.

oscargus:
Now I think it is time to have this checked.

The split between gui and logic is not done yet, but is rather straightforward (waiting for some opinions on com.sun.star in logic). Also, the improved detection is not done, but seems like a separate issue somehow (and currently I do not have any good idea how to do that nor do I have access to all platforms...).

The major change is the new style selection dialog and underlying mechanism. Now, all files are shown in a single dialog, so no automatic searching of all subdirectories etc. This will also allow for easier integration of more JabRef-provided style files (just add the resource name to StyleLoader). The drawback is that users will have to add all their style files again.

newstyledialog

These changes made some preferences obsolete. Is there a mechanism to remove obsolete preference keys from the preference file?

oscargus:
OK, seems like I did a bit of work on the autodetection anyway...

Both Windows and OS X is restructured and something similar can be done for Linux. However, I'm not convinced with the suitability of the current default directories for Linux. Does anyone have some information about where different distributions tends to install OO/LO? My Linux machine is centrally administered and have all software in /sw, but I guess that isn't really standard...

Siedlerchr:
On Linux there a few standard dirs for software.
I could later check on my other Ubuntu pcs where it is installed.
At the meantime this post lists all the basic (standard) bin dirs
http://askubuntu.com/questions/60826/where-is-the-default-folder-for-apps

lenhard:
Regarding your question on structuring: There should be no dependencies to any GUI technology in the classes in logic. That is, no class in logic should have a UI class in its import statements (that is ensured automatically by the architecture tests). 

Transitive dependencies that come via external jars are a gray area to some degree (for reasons of practicality). If you have a jar that bundles UI and logic code, it is fine to use it in our logic package as long as you do not access the UI classes in the jar. Otherwise, better keep these dependencies in the gui package. 

You have mixed types of dependencies. `com.sun.star.beans` and `com.sun.star.container` can be referenced in logic without problems. `com.sun.star.uno` should also be fine. `com.sun.star.text` has GUI and non-GUI content, so it depends on the usage, and `com.sun.star.frame` should definitly not be referenced from logic. Unless you can separate these dependencies, rather keep the referencing classes in GUI.

oscargus:
Thanks for clarifying. I guess it is really OOBibBase and OOUtil that require a bit of thinking. OOUtil is primarily a set of static methods, one which contains swing stuff, so that can easily be moved to a separate class (or just into a calling class). Regarding the sun stuff it is indeed not obvious where to draw the line, but I'll make an attempt based on your discussion.

Exceptions will go based on where they are thrown unless that somehow breaks the logic vs gui rule (although I do not think that will/can happen).

lenhard:
I had a look at your distribution of oo classes to the gui/logic packages and it seems fine! I also aggree that we could merge this now and improve linux support later. The current PR is already quite big and, thus, quite hard to review. Most codacity issues are uncritical from my point of view.

Unless you do not think that something is missing, we can rebase and merge.

oscargus:
I think it should be as good as/slightly better than before so I'll try to
rebase later on tonight and then merge it in.

Siedlerchr:
@oscargus  The only problematic I see from the codacy is the catching of a Throwable in 
`src/main/java/net/sf/jabref/gui/openoffice/AutoDetectPaths.java`
`public boolean runAutodetection() { `

Avoid  catching a Throwable or a general Exception as it may include non-related errors, e.g. OutOfMemory or others.
http://javarevisited.blogspot.de/2014/02/why-catching-throwable-or-error-is-bad.html

lenhard:
@Siedlerchr To be fair to @oscargus: That issue did exist before. Codacy sometimes just detects issues as new, although they are old, if a related class is modified.

Anyway, it is certainly an issue and is also easily corrected, so @oscargus please change it to a proper exception before merge.

Btw.: JabRef is full of such issues. Just search for usages of Throwable. If anybody has the time and is in the mood for easy and small code improvements, go ahead :-)

oscargus:
Yes, there are a number of Throwables caught in JabRef, but I haven't dared
to change them as I do not know enough about it. But I've been tempted to
just replace all Throwable with Exception throughout the code...

oscargus:
And, yes, moving the files shows all issues as new and no fixed issues...

lenhard:
Ok, Exception is better than Throwable, but still quite bad :-) You should always catch the most concrete type (IOException, etc.) to avoid catching other Exceptions accidently. That really depends on the code inside the try-catch-block.

oscargus:
One of the sun methods throw Exception. Apart from the Throwable I believe
that most of the code catches as detailed exceptions as possible  (as I
spent some time earlier tuning that). If I remember correctly quite a bit
of the Exception/Throwable catching in JabRef is of a rather generic type:
run some methods which doesn't explicitly throw something and surround them
with a try block just in case so that nothing breaks if an exception is
thrown (to avoid halting the application I assume). Most old GUI-build code
is like this.

I'll have a look at the Throwable and see what I can do.

Siedlerchr:
Good to know about the Codacy issues thing. Already wondered a couple of times about this :confused: 

For the concrete case the init method which shows the dialog seems to be responsible for it.
A quick look shows that there is no Exception Handling necessary.
To be on the safe side you could initialize the foundPahts to false.


One minor thing in the long Autodetect method : I would suggest to split up each case specifid (OSX, Windows, Linux) in an own method and then simply returning true/false. 
Apart from this: Great work :+1:

oscargus:
Thanks!

My plan is to split it in another way, selecting a class for the commands which are different, but we'll see what happens in the end. Separate functions would have been a good idea now indeed.

Issue #1038, opened 2016-03-25
Improve layout of "Cleanup entries" dialog


koppor:
JabRef 3.3dev--snapshot-2016-0-24--master--3d8cc32 on Windows 10.

I have some usability remarks on the new "Cleanup entries" dialog:

- [ ] The position of the button `-`is unclear. When reading it, it removes the thing left of it, because this is the context. But that context doesn't change when selecting something different in the box of selected formatters. Proposal: Move below the box
- [ ] "Description": is always empty at current master
- [ ] The box of selected formatters should have a heading
- [ ] "Run field formatter:" should be "Run configured field formatters:"
- [ ] After removing an entry in the list, the next entry should be selected
- [ ] There should be a context menu for each enetry in the table. It should contain "Remove selected entry" (or similar)

As is dialog:

![grabbed_20160325-131223](https://cloud.githubusercontent.com/assets/1366654/14044196/e8d1e71e-f28d-11e5-94ca-07464a1befad.png)

Should be dialog - only the last part, the upper part is addressed in #959 

We are modifying a table somehow. However, it seems to be difficult to add a complete table editing. At the beginning, I thought about "Add", "Remove", and "Change" as buttons. That seemed to be too complicated. Therefore, I stick with the current way.

Possibly, we should add move-up and move-down buttons.

The text at "Description" should be updated either when an entry in the table is selected or the combo box for formatters is changed.

![grabbed_20160325-161018](https://cloud.githubusercontent.com/assets/1366654/14046815/16b97208-f2a4-11e5-8c1f-0feb2912fb34.png)

(Tool used for UI sketch: [Evolus Pencil](http://pencil.evolus.vn/))

Siedlerchr:
There is also one thing I noticed during the changes in #999  which is not yet resolved.
When one item gets deleted/remoed the selection-focus should move to the next item in the list.
However, there have to be considered a few cases. E.g. 
OnRemoving:
- last item: Selection should go the prev one
- first item : Selection should go to the next one
- only item: No more selection
- "middle" item: Prev or Next one?

Siedlerchr:
I'm onto that.

tobiasdiez:
I would leave the bigger restruction to the javafx rework, so for the moment only fix the description field and maybe that the next item is selected on delete. In the long-term I would prefer a completely inline editing of the formatters without any buttons and no separation between addition and display of existing formatters.

Siedlerchr:
See PR for more details
[...]

PR #1050, opened 2016-03-27
Restructrued CleanUp Formatters dialog  #1038

Siedlerchr:
Reworked CleanUp Formatter dlg:
- [x] Description updates when and Item is selected. Also on startup. The field is now also multilane with auto resizing and word wrap.
- [x] After deleting a selected item in the list, the next one (previous) is selected
- [x] Changed layout to be more consistent 

Renamings to be more precise and moving in new package.
Added tests.
- [ X] Change in CHANGELOG.md described
- [ X] Tests created for changes
- [ X] Screenshots added (for bigger UI changes)

cleanupentriesdialog


src/main/java/net/sf/jabref/gui/cleanup/CleanupActionsListModel.java
```diff
@@ -10,34 +10,39 @@
...
+    public CleanupActionsListModel(List<FieldFormatterCleanup> cleanupAction) {
+        Objects.requireNonNull(cleanupAction);
+
+        this.cleanupActions = cleanupAction;
 
-        this.saveActions = saveActions;
-        listeners = new ArrayList<>();
     }
 
     public void addSaveAction(FieldFormatterCleanup action) {
```
tobiasdiez: Maybe also rename this.

src/main/java/net/sf/jabref/gui/cleanup/CleanupPresetPanel.java
```diff
@@ -51,15 +51,16 @@ private void init() {
         cleanUpUpgradeExternalLinks = new JCheckBox(
                 Localization.lang("Upgrade external PDF/PS links to use the '%0' field.", Globals.FILE_FIELD));
         cleanUpUnicode = new JCheckBox(Localization.lang("Run Unicode converter on title, author(s), and abstract"));
-        cleanUpBibLatex = new JCheckBox(Localization
-                .lang("Convert to BibLatex format (for example, move the value of the 'journal' field to 'journaltitle')"));
-        cleanUpFormatters = new FieldFormatterCleanupsPanel(Localization.lang("Run field formatter:"),
+        cleanUpBibLatex = new JCheckBox(Localization.lang(
+                "Convert to BibLatex format (for example, move the value of the 'journal' field to 'journaltitle')"));
+
+        cleanUpFormattersPanel = new FieldFormatterCleanupsPanel(Localization.lang("Run selected field formatter:"),
```
tobiasdiez: Do you think the ,selected, might lead to confusion since user think that only the selected formatter and not all listed ones are used?

src/main/java/net/sf/jabref/gui/cleanup/FieldFormatterCleanupsPanel.java
```diff
@@ -102,46 +101,108 @@ public void mouseMoved(MouseEvent e) {
...
         // make sure the layout is set according to the checkbox
         cleanupEnabled.addActionListener(new EnablementStatusListener(fieldFormatterCleanups.isEnabled()));
         cleanupEnabled.setSelected(fieldFormatterCleanups.isEnabled());
+
+    }
+
+    /**
+     * Create a TextArea that looks and behaves like a JLabel. Has the advantage of supporting multine and wordwrap
+     */
+    private void makeDescriptionTextAreLikeJLabel() {
```
tobiasdiez: There is an a missing in textarea

src/test/java/net/sf/jabref/gui/cleanup/CleanupActionsListModelTest.java
```diff
@@ -0,0 +1,90 @@
...
+        assertEquals(ListDataEvent.INTERVAL_REMOVED, argument.getValue().getType());
+
+    }
+
+    @Test
+    public void removedAtIndexMinus1() {
+
+        CleanupActionsListModel model = new CleanupActionsListModel(getDefaultFieldFormatterCleanups());
+        ListDataListener listener = mock(ListDataListener.class);
+        model.addListDataListener(listener);
```
tobiasdiez: I am missing an assertion here and in the next test. Probably you want to check that the model still contains the same items and use `verifyZeroInteractions(mockObject)` to verify that no change event was raised. Please also change the test name to reflect the expected outcome, i.e. removedAtIndexMinus1DoesNothing

tobiasdiez:
Looks good to me :+1: Just a few small remarks which should be fixed then this can be merged in.

src/main/java/net/sf/jabref/gui/cleanup/CleanupPresetPanel.java
Siedlerchr: Maybe: Run listed formatters?

Siedlerchr:
@tobiasdiez  Applied your proposed changes.
Instead of "Run selected field formatters" it is now  "Run listed field formatters"

tobiasdiez:
I think this looks good. @koppor what do you think?

simonharrer:
Why is this assigned to @koppor ? What is blocking this?

koppor:
#1038 is my request and I'd like to have the example displayed. For that, I need #1159 to get through.

My aim to have the dialog understandable to users.

Further, I need to test the thing, for which I didn't have time yet. :see_no_evil:

koppor:
- [ ] When changing the focus of the list, the description does not update. Is it possible to update the description when the focus of the list changes?

- [ ] When the formatters are disabled, it is not clear, where the description comes from. Maybe hide the field?

grabbed_20160410-140135

- [ ] When I enable the formatting, no formatter is selected. Could there the first formatter (if available) be enabled?

Change at the combobox works.

Siedlerchr:
@koppor  The focus of the description only updates when you select it in the combobox. It would be otherwise too confusing, if the combobox displays Formatter A and you select Formatter B in the list. Which one should be displayed then?
From my point of view the Workflow atm is: Select the field e.g. abstract and select the formatter to add to the list. Therefore the description is referring to the current selected formatter.

The other things regarding the enabling/disabling could be done easily

As tomorrow my new semester starts, I don't know if I can get this whole thing done in time for v3.3 Release on (Friday?).

koppor:
@Siedlerchr Do you still have time?

Siedlerchr:
Yep, wil look at it this evening or tomorrow morning.

koppor:
The thing is that a user might now, what a specific formatter does when opening the dialog the first time. As far as I remember, we have some default formatters enabled and the user doesn't know what to do.

Therefore, this quick suggestion. Note that I added the formatter string at the sketch at #1038. That was aimed to reduce confusion somehow :innocent:

koppor:
Decision: we merge as is. @Siedlerchr Thank you for working on it.

stefan-kolb:
@Siedlerchr Can you please rebase this, so we can merge?

Siedlerchr:
Will do this today evening.

2016-04-13 10:51 GMT+02:00 Stefan Kolb <notifications@github.com>:

koppor:
Unfortunately, nearly all localization files have been changed. How did that happen? Can you remove these changes from the commit?

grabbed_20160413-221435

Siedlerchr:
@koppor  I have no idea. I recreated the branch from master and then I ran the gradle lang scripts.
I only added one new string and it seems like the obsolete one got removed.

koppor:
Other than that: LGTM :+1:

koppor:
We decided to use existing translation strings instead of introducing new ones a few days before the release.

"Remove selected" (wished) -> "Remove selected entries" (existing)
"Run_listed_field_formatters" (wished) -> "Run field formatters" (existing)

koppor:
[LGTM](http://lgtm.in/i/4EbBDqGPn)

Issue #1051, opened 2016-03-28
Entry sorting for Export


halirutan:
JabRef version 3.3dev--snapshot--2016-03-28--master--f9f6cec on Ubuntu 14.04.

When I try to export a database to a different format (I tried simpleHTML and my own implementation of Markdown), the sorting order of the entries is not used. It doesn't matter if I set the current sorting by clicking on the columns or if I'm using the default sorting accessed through Preferences -> Entry Table.

The only thing the error console shows is

```
Mar 28, 2016 6:02:58 AM net.sf.jabref.exporter.ExportFormat performExport
WARNING: []
```

I found similar issues that were reported some years ago where the solution always was to set the *default sorting*. It would be helpful if someone could confirm that this indeed works.

Siedlerchr:
As there have been recently fixes with Entry exporting, did you try the latest version from master?
http://builds.jabref.org/master/

halirutan:
Let me add some *official* example data. I'm using this [test.bib](http://mirror.unicorncloud.org/CTAN/biblio/bibtex/contrib/test/test.bib). I have set the default sorting to year (and title as second) and it looks good in the interface

![bibtex1](https://cloud.githubusercontent.com/assets/511683/14091836/d145a7a4-f542-11e5-94cf-62b4deccb67d.png)

When I choose `File` -> `Export` and export to simple html, then the ordering is destroyed:

![bibtex2](https://cloud.githubusercontent.com/assets/511683/14091859/f66baf92-f542-11e5-842d-cd6f2193e65f.png)

halirutan:
@Siedlerchr Let me try the latest snapshot.

halirutan:
@Siedlerchr Same behavior with the latest snapshot. I updated the version in the original issue description.

Siedlerchr:
@halirutan  Thanks for your update and reproducing the issue.

koppor:
@halirutan halirutan Just to double check: Did you choose "Export in current table sort order" in the preferences as shown in the screenshot?

![grabbed_20160330-085407](https://cloud.githubusercontent.com/assets/1366654/14134125/05b15dae-f655-11e5-8611-8e972b91630c.png)

halirutan:
@koppor Thank you. Since this option was not available until recently (my ubuntu comes originally with 2.10beta) I did not notice this additional setting when testing version 3!

That solves the problem!

Issue #180, opened 2015-09-24
Improve sharing: Store serialization settings within the bibtex file


koppor:
JabRef is able to sort the `.bib` file according to a configuration. When users with different configuration options check in their `.bib` file into version control, the files have textual diffeences without any content difference.

I propose that JabRef writes its serialization settings into the bibtex file at each save. This ensures that the serialization options are always met. The other option is that JabRef asks the user whether he wants to prepare the `.bib` file for sharing. If not, JabRef stores the filename in its preferences and never asks the user again.

This covers at least
* entry sorting (alphabetically using the author, sorting by the bibtex key, or no sorting, just adding entries at the end)
* format of each entry (see #116)

## Related discussions
* This issue is closely related to the discussion going on at #116 - especially https://github.com/JabRef/jabref/issues/116#issuecomment-133311604.
* https://sourceforge.net/p/jabref/feature-requests/892/
* https://sourceforge.net/p/jabref/feature-requests/854/

lenhard:
But then the file contains text that is not bibtex and probably not in bibtex syntax. Might this not be a problem if the `.bib` file is shared with someone who uses a different editor?

Apart from that, I am all for file purity and against putting additional stuff into the `.bib` file. I'd prefer to define a fixed serialization format as in #116 :)

koppor:
Currently, JabRef uses standardized `@comment` entries, e.g. `@comment{jabref-meta: groupsversion:3;}`. The only exception is the encoding (see #118).

Maybe it would be good to
A) Have one defined serialization format (see your comment https://github.com/JabRef/jabref/issues/116#issuecomment-134121807)
B) Have the file sorting options not stored as setting, put as Tool to call explicitly. New entries are sorted in after the first alphabetical match. Only ONE sorting setting, not a configurable one.

Maybe, both things should but into our survey. @mlep (I'll write a private mail)

mlep:
Waiting for your input!  :-)

lenhard:
Fixed in master

PR #1054, opened 2016-03-28
Fix sorting of entries on save

tobiasdiez:
Fix sorting of entries on save (this shouldn't affect the sorting of export, so is unrelated to #1051). In the end it was just a wrong boolean in the preference.

- [ ] Change in CHANGELOG.md described - no, I think the issue was introduced in this version
- [ ] Tests created for changes - kind of, there is a test for the sorting now but it passed before any fix, no test added for the boolean in the preference

### DIFF
--- a/src/main/java/net/sf/jabref/exporter/SavePreferences.java
+++ b/src/main/java/net/sf/jabref/exporter/SavePreferences.java
@@ -51,7 +51,7 @@ public class SavePreferences {
     }
 
     public static SavePreferences loadForSaveFromPreferences(JabRefPreferences preferences) {
-        Boolean saveInOriginalOrder = true;
+        Boolean saveInOriginalOrder = false;
         SaveOrderConfig saveOrder = null;
         Charset encoding = preferences.getDefaultEncoding();
         Boolean makeBackup = preferences.getBoolean(JabRefPreferences.BACKUP);
--- a/src/test/java/net/sf/jabref/exporter/BibDatabaseWriterTest.java
+++ b/src/test/java/net/sf/jabref/exporter/BibDatabaseWriterTest.java
@@ -420,4 +420,92 @@ public class BibDatabaseWriterTest {
 
         assertEquals("", stringWriter.toString());
     }
+
+    @Test
+    public void writeEntriesSorted() throws IOException {
+        SaveOrderConfig saveOrderConfig = new SaveOrderConfig(false, new SaveOrderConfig.SortCriterion("author", false),
+                new SaveOrderConfig.SortCriterion("year", true),
+                new SaveOrderConfig.SortCriterion("abstract", false));
+        metaData.setSaveOrderConfig(saveOrderConfig);
+
+        BibEntry firstEntry = new BibEntry();
+        firstEntry.setType(BibtexEntryTypes.ARTICLE);
+        firstEntry.setField("author", "A");
+        firstEntry.setField("year", "2000");
+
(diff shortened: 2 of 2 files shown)


src/main/java/net/sf/jabref/exporter/SavePreferences.java
```diff
@@ -51,7 +51,7 @@ public static SavePreferences loadForExportFromPreferences(JabRefPreferences pre
     }
 
     public static SavePreferences loadForSaveFromPreferences(JabRefPreferences preferences) {
-        Boolean saveInOriginalOrder = true;
```
oscargus: I'm not really sure I'm commenting on the right thing, but I am sure that the consensus is that the default should be to save in original order. It might be that this switch cannot be changed in the preferences, but then that issue should be solved.

lenhard: That is true. If no explicit sort order is set in the meta data, the entries should be saved in the original order (to modify as little in the file as possible). 

As far as I can see there is currently no test for that. Could you perhaps add one in this PR?

koppor:
The idea is that a user can configure a `bib` template, which contains the default "preferences" for a database, such as the bibtex key patterns and the sort order configuration. When the user creates a new `bib` file, the template is used.

We discussed sometime ago the plethora of preferences. Maybe that approach could help to reduce that. However, the system of a `bib` template could increase complexity. Since I care for sharing and group editing of `bib` files, I want to a) have as much as possible bib-affecting configurations (directory, key patterns, ...) in the `bib` file itself (See for instance #180) and b) enable new users to have a properly configured `bib` file and not requiring an expert to configure JabRef.

a) is IMHO mostly achieved
b) is some thing we are working on to have correct default settings. My personal feeling is that some groups, however, do not want to follow our conventions (for instance, bibtex keys, sort order) and thus require different settings.

Maybe I should open a separate issue asking for support of "bib template". :innocent:

tobiasdiez:
@koppor commented on the wrong issue?

koppor:
@tobiasdiez No, not the wrong issue. I tried to give some context above. Hope, that helps to understand my text.

src/main/java/net/sf/jabref/exporter/SavePreferences.java
tobiasdiez: I added a corresponding test. If `saveInOriginalOrder` is set to true, then even the save order specified in the metadata is ignored.

lenhard: Hm, do I understand you correctly @tobiasdiez: The preference `saveInOriginalOrder` overwrites meta data in the file? If so, then this is a problem, because the serialization can differ between machines. That harms version-controllability, our top aim regarding the serialization.

I think meta data that is explicitly written into the file should always be followed above user-specific preferences.

tobiasdiez: Yes `saveInOriginalOrder = true` let JabRef ignore the save order specified in the metadata. It was a bug that this was set for the save procedure (fixed in this PR). So now the metadata is taken into account for saves. For exports, user can choose to always export in the original order (see Preference dialog) regardless of what was specified in the metadata (and this preference option is the whole purpose of the boolean variable).

lenhard: Ok, great! Then, I'll have a last check and merge this PR.

Issue #1055, opened 2016-03-28
NPE in getSortedEntries when exporting to external DB


Siedlerchr:
Jabref 3.3 dev, master, commit: ed989b069b768390240fd4a2946b9f4b13142cdc
Reproducable
Steps to reproduce:

1. Open Export to External SQL Database
2. Choose Mysql (doesn't matter)
3. Fill in the conn details
4 . Select Target DB/Create new one
5. Enter name 
6. Click OK

NPE is shown:
```
18:53:07.518 [Spin-0] INFO  net.sf.jabref.gui.BasePanel - Could not export to SQL database
java.lang.NullPointerException
	at java.util.Objects.requireNonNull(Unknown Source) ~[?:1.8.0_74]
	at net.sf.jabref.exporter.BibDatabaseWriter.getSortedEntries(BibDatabaseWriter.java:85) ~[bin/:?]
	at net.sf.jabref.sql.exporter.DBExporter.performExport(DBExporter.java:86) ~[bin/:?]
	at net.sf.jabref.sql.exporter.DBExporter.exportDatabaseToDBMS(DBExporter.java:430) ~[bin/:?]
	at net.sf.jabref.gui.BasePanel$1.run(BasePanel.java:544) [bin/:?]
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[?:1.8.0_74]
```

Siedlerchr:
I am looking into that, seems like I found the problem

tobiasdiez:
Yes, someone still passes `null` to the entries argument, which was the old version to indicate that all entries should be sorted.

Siedlerchr:
@tobiasdiez  Yep. there is missing a getSelectedEntries() as Parameter.

oscargus:
Hmm, I thought that I (i.e. Eclipse) had found all performExport calls, but obviously not...

PR #1056, opened 2016-03-28
Fixes NPE in Export to External DB

Siedlerchr:
Fixes NPE in Export to External DB
Add DB Type to Export Succesful message
Reformatting
#1055

No Changelog because I think it was new introduced by the redesign of the Entry sorting (passing null)

### DIFF
--- a/src/main/java/net/sf/jabref/gui/BasePanel.java
+++ b/src/main/java/net/sf/jabref/gui/BasePanel.java
@@ -98,8 +98,8 @@ import java.nio.charset.UnsupportedCharsetException;
 import java.util.*;
 
 public class BasePanel extends JPanel implements ClipboardOwner, FileUpdateListener {
-    private static final Log LOGGER = LogFactory.getLog(BasePanel.class);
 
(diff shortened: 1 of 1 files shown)


src/main/java/net/sf/jabref/gui/BasePanel.java
```diff
@@ -541,7 +543,8 @@ public void run() {
                         frame.output(Localization.lang("Attempting SQL export..."));
                         final DBExporterAndImporterFactory factory = new DBExporterAndImporterFactory();
                         final DBExporter exporter = factory.getExporter(dbs.getServerType());
-                        exporter.exportDatabaseToDBMS(database, bibDatabaseContext.getMetaData(), null, dbs, frame);
```
tobiasdiez: Should really only the selected entries be exported or all (getDatabase().getEntries)? Previously `null` indicated the latter.

Siedlerchr: Okay, then I think I take better all entries to be consistent

Siedlerchr:
Export now all entries by default

PR #1057, opened 2016-03-29
Fix wrong SQL db-export (same entry multiple times)

Siedlerchr:
Followup from #1056. I noticed that all entries where inserted multiple times in the entry table.
Moved processQuery out of the loops.

From what I remember 3.2 Export was working correct.

### DIFF
--- a/src/main/java/net/sf/jabref/sql/exporter/DBExporter.java
+++ b/src/main/java/net/sf/jabref/sql/exporter/DBExporter.java
@@ -67,6 +67,7 @@ public abstract class DBExporter extends DBImporterExporter {
 
     private static final Log LOGGER = LogFactory.getLog(DBExporter.class);
 
+
     /**
      * Method for the exportDatabase methods.
      *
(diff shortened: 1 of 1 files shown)


Siedlerchr:
@tobiasdiez  Coud you or someone else merge this in please? Only 1 line affected.

src/main/java/net/sf/jabref/sql/exporter/DBExporter.java
```diff
@@ -202,7 +201,7 @@ private void populateEntryTypesTable(Object out, BibDatabaseMode type) throws SQ
         List<String> existentTypes = new ArrayList<>();
         if (out instanceof Connection) {
             try (Statement sm = (Statement) SQLUtil.processQueryWithResults(out, "SELECT label FROM entry_types");
-                 ResultSet rs = sm.getResultSet()) {
+                    ResultSet rs = sm.getResultSet()) {
```
koppor: Why the indent here?

PR #1066, opened 2016-03-30
Changed preview font from Arial to Sans-Serif family

oscargus:
Questions on that?

### DIFF
--- a/src/main/java/net/sf/jabref/JabRefPreferences.java
+++ b/src/main/java/net/sf/jabref/JabRefPreferences.java
@@ -681,7 +681,7 @@ public class JabRefPreferences {
         defaults.put(PREVIEW_ENABLED, Boolean.TRUE);
         defaults.put(ACTIVE_PREVIEW, 0);
         defaults.put(PREVIEW_0,
-                "<font face=\"helvetica\">"
+                "<font face=\"sans-serif\">"
                 + "<b><i>\\bibtextype</i><a name=\"\\bibtexkey\">\\begin{bibtexkey} (\\bibtexkey)</a>"
                 + "\\end{bibtexkey}</b><br>__NEWLINE__"
                 + "\\begin{author} \\format[Authors(LastFirst,Initials,Semicolon,Amp),HTMLChars]{\\author}<BR>\\end{author}__NEWLINE__"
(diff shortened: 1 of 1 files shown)


stefan-kolb:
Were does the font come from? Helvetica is not a system font on WIndows for example.

Siedlerchr:
I would rather suggest to use one of the Open-Source Fonts from the LIberation Sans family.
They are an replacement to Arial and they are used in LO/OO, too. 
So Linux-users are not required to install the propietary fonts.

https://en.wikipedia.org/wiki/Liberation_fonts
Or for Helvetica a replacement of the Tyre Gyro fonts 
https://wiki.ubuntuusers.de/Schriftarten/

oscargus:
Helvetica and Arial will give the same font on all systems as they are both in the list of "standard sans-serif fonts", so on Windows Arial will be used and on Linux whatever the standard sans-serif font is (probably as Christoph says). To me Arial is just the ugly cousin of Helvetica (which is one of the standard 14 postscript fonts), so it is more a matter of principles and values rather than any practical difference (unless you have both installed, in which case you most likely work with typography and probably prefer Helvetica anyway).

oscargus:
An option is of course to go with "sans-serif", see http://www.w3schools.com/cssref/pr_font_font-family.asp

stefan-kolb:
Can you add a image showing the differences here? As the Fallback will probably work on every OS I am OK with your change.

matthiasgeiger:
the "default" solution for this in HTML/CSS would be `<font face="Arial, Helvetica, sans-serif">...</font>`

oscargus:
There shouldn't be any visible differences (although I think I do have
Helvetica installed). Maybe the best thing is to skip the personal
preferences and go with sans-serif.
Den 31 mar 2016 11:29 skrev "Stefan Kolb" <notifications@github.com>:

oscargus:
Yes, if we were convinced that Arial is the best font to be used. Which I do not think we are, but just put the standard sans-serif Windows font there.

matthiasgeiger:
> To me Arial is just the ugly cousin of Helvetica 

> There shouldn't be any visible differences

This is why I never really understood these *Arial vs. Helvetica* discussions... :stuck_out_tongue_winking_eye:

oscargus:
I think the main thing is that Arial is considered to be Microsoft (which
strictly isn't true) and therefore bad. Especially if you grew up with Unix
and Postscript and so on.

And they are visually not identical. It was more a statement of that the
HTML renderer will pick the other if the requested isn't available.

However, to avoid (more) purist discussions I will change to sans-serif and
then it is up to the OS to decide which is the "best"
Arial/Helvetica/Whatever font available.

matthiasgeiger:
:+1:

We should not waste more time on this ;-)

Siedlerchr:
While reading on Arial vs Helvetica I found this little quiz where one of the logos is in Arial the other in Helvetica. I could not see any differences...  :astonished:
http://www.ironicsans.com/helvarialquiz/

oscargus:
:-)

19 out of 20, but I must admit that a few days ago I would have ended up on
about 10 (random).

PR #1067, opened 2016-03-31
Added initial Swedish translations

oscargus:
I cannot really figure out what "entry" should be...


stefan-kolb:
Nice, when you're ready to merge you can also add a changelog entry :smile:

lenhard:
Tricky :-) With my limited non-native swedish, I can only be of limited help. Maybe an alternative? After all, a BibTeX entry is nothing but an *element* of a BibTeX database. So how does *ett BibTeX element* sound to you?

stefan-kolb:
There's no"Eintrag" in Swedish? :smile:

lenhard:
According to my dictionary, Eintrag translates to införande (~ Eintragung, Registrierung) or anteckning (~ Anmerkung), neither of which seem very fitting here.

matthiasgeiger:
Google proposes "entry - inträde, post, inresa, tillträde, artikel, entre".
But as I do not speak any swedish I cannot decide whether those make any sense ;-)

oscargus:
Element is OK but there might be better ones (strictly it can be used, but sounds a bit _krystat_). Same with post. Artikel is sort of interesting as well, but there's the duplicate meaning article.

oscargus:
I think post is probably the best. That is what is commonly used in databases, so from that perspective it should be OK. Thanks all for your input!

oscargus:
Just merge this before releasing 3.3. Most menu entries are translated and a decent part of the other strings (I use a plain text editor, so no idea what percentage). I will add a few more every now and then, but would need to install some standard software in Swedish to figure out certain terms, like "redo"... ;-)

Issue #1018, opened 2016-03-22
Importing to the correct mode when searching (BibTeX vs biblatex)


oscargus:
Most (all?) of the fetchers will now produce a BibTeX entry. When importing to a BibLaTeX database, this causes some issues, see e.g. #1015. It might make sense to add some mechanism to the search interface such that the correct format is returned. For example, this can be achieved by "search formatters", which will also remove some of the need of explicit cleanup code in the fetchers as well as some of the preference settings.

stefan-kolb:
I don't think this should be classified as a bug.
The question here is, who is responsible for outputting the correct format?
- the fetcher
- Jabref itself

I think it should be the fetcher that must be aware of what to output because he is the only one that is able to produce correct entries?!
On the other hand, this would force the fetcher to know the current mode which adds another dependency or at least a parameter to the call.

If we transform entries after fetching to the correct format this could be a hard thing.
This can only work if we always fetch Biblatex and rearrange the fields and strip unsupported fields etc.

tobiasdiez:
Fetcher should try to fetch in BibLaTex format, but can also return BibTex format. Afterwards JabRef should always run a converter which converts to the correct mode depending on user preferences (for this a BibLaTex > BibTex converter has to be written, based on the `BiblatexCleanup`).

Siedlerchr:
This issue is now finally closed and fixed in the latest development snapshot 


We would like to ask you to use a development build from https://builds.jabref.org/main and report back if it works for you. Please remember to make a backup of your library before trying-out this version.

PR #1069, opened 2016-03-31
GVKParser handle publication year correctly

teertinker:
exchanged the field "date" with "year" to import GVK entries correctly.

Publication year was previously put to date when fetching entry-> now to year

* Sorry if I broke any conventions of changing code/testing...

### DIFF
--- a/src/main/java/net/sf/jabref/importer/fetcher/GVKParser.java
+++ b/src/main/java/net/sf/jabref/importer/fetcher/GVKParser.java
@@ -388,7 +388,7 @@ public class GVKParser {
             result.setField("publisher", publisher);
         }
         if (date != null) {
-            result.setField("date", date);
+            result.setField("year", date);
         }
         if (address != null) {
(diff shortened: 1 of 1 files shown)


koppor:
GVK's `date` is always 4 digits?

teertinker:
I use it a lot, and never got other "dates" than 4-digits.

matthiasgeiger:
@koppor You changed this in b2c8db5458fb46b5581b7018897f67c8cf010bcc from year to date. 

The reason seems to be related to #1018: GVK returns Biblatex entries (regardless of the chosen mode)...

koppor:
OK for revert. Current JabRef displays "year" in the "Required fields", which it IMHO did not back in 2015.

simonharrer:
Reverted in 93fdb02ee5889d0e0cf9b41a2af763cc306fbe49

Thanks for the help, especially @teertinker :smile:

Issue #989, opened 2016-03-17
Translations - what to translate?


oscargus:
While struggling with #869 I've scrolled through the translation strings and there are some strings which is questionable to translate. However, before I remove them, I appreciate some feedback. The strings I think we should remove from translation are:

* HTML
* BibTeXML
* DocBook
* Endnote
* Harvard_RTF
* MODS
* MS_Office_2007
* OpenOffice_Calc
* OpenOffice_CSV
* BibO_RDF
* RIS
* DIN_1505
* OpenOffice (which we should consider to replace with "OpenOffice/LibreOffice" everywhere)
* ISO_690
* MIS_Quarterly
* CaseKeeper
* UnitFormatter
* BibTeX

In addition there are some strings which are inconsistently cased (typically written in Title Case). Not sure if I will change those, but eventually...

matthiasgeiger:
Yes. I think all of them (exept CaseKeeper and UnitFormatter, there is still #920 open) can be removed.

lenhard:
I really wonder why these things were localized to begin with...

Anyway, you solved it with #1071

oscargus:
Probably just copy paste behaviour.  If some file types should be
translated, all are etc.

CaseKeeper for sure as I did that back in the days without really
understanding...

PR #1013, opened 2016-03-21
Rename formatters

tobiasdiez:
The aim of this PR is to rename all formatters as discussed in https://github.com/JabRef/jabref/issues/920.

Changes:
- Remove Converter class which is superfluous with the improved formatter interface
- Enable LatexToUnicode as formatter (previously it was only a LayoutFormatter)
- Rename formatters (the key is always the class name without the `Formatter` suffix, a value in brackets indicates that I deviated from the last version in #920):

(I will add the necessary changes to the language files as soon as the names are accepted)

Format:
 * Basis is https://github.com/JabRef/jabref/issues/920#issuecomment-195779352
 * Plus later comments
 * name: chosen name, brackets show non chosen alternatives.

--

### Authors related
- [ ] AuthorsFormatter
  - classname: NormalizePersonNamesFormatter (NormalizeAuthorsFormatter)
  - name: normalize names of persons (normalize authors)
- [ ] AuthorsMinifier
  - classname: MinifyPersonNamesFormatter (MinifyAuthorsFormatter)
  - name: Minify authors

### Date related
- [ ] MonthFormatter
  - classname: NormalizeMonthFormatter
  - name: normalize month
- [ ] DateFormatter
  - classname: NormalizeDateFormatter
  - name: normalize date

### Page related
- [ ] PageNumbersFormatter
  - classname: NormalizePagesFormatter
  - name: normalize page numbers (normalize pages)

### Casing related
- [ ] TitleCaseChanger
  - classname: TitleCaseFormatter
  - name: title case (Titelize, EnforceTitleCaseFormatter)
- [ ] UpperEachFirstCaseChanger
  - classname: CapitalizeFormatter
  - name: capitalize
- [ ] UpperCaseChanger
  - classname: UpperCaseFormatter (UpcaseFormatter)
  - name: upper case (to upper case)
- [ ] UpperFirstCaseChanger
  - classname: SentenceCaseFormatter (CapitalizeFirstDowncaseOthersFormatter)
  - name: sentence case (capitalize first, downcase others)
- [ ] LowerCaseChanger
  - classname: LowerCaseFormatter (DowncaseFormatter)
  - name: lower case (to down case)
- [ ] CaseKeeper
  - classname: ProtectCapitalsFormatter (FreezeTermsFormatter)
  - name: protect capitals (freeze terms)
- [ ] RemoveBracesFormatter
  - key: RemoveBracesFormatter (UnfreezeWords)
  - name: remove enclosing braces (unfreeze words)
[...]

PR #1071, opened 2016-03-31
Unified translations strings

oscargus:
Removed:
 * 10 translations strings where a very similar one was already available
 * 3 unused keybindnings
 * 14 names which cannot be translated


oscargus:
Fixes #989

Siedlerchr:
Good work. When this is done I will add the missing German strings.

matthiasgeiger:
@Siedlerchr You should wait with translating as a) we'll discuss #1013 and 920 today in a DevCall which is highly affecting the l10n files as there will be a whole bunch of new/changed translations and b) we should ask our longstanding German translator @domwass if he has time to translate. (As I learned that he's more experienced than me producing better and more consistent translations :wink:)

domwass:
Regarding my time: there is none left at the moment (I’m writing a book which is due in two weeks). So go ahead, I will look into it as soon as I will have some time left.

matthiasgeiger:
Thanks for the feedback! And good luck with your book!

PR #1072, opened 2016-03-31
DOI field improvements

stefan-kolb:
## __Major changes of this PR__
- [x] Removed unused BROWSE action
- [x] Remove double click action for URL and DOI fields
- [x] Add open and fetch from buttons for URL and DOI
- [x] Only activate button if field includes valid DOI/URL

## __New Look__

image


oscargus:
I use doubleclick and think it is a quite neat feature. Especially when only a single field is available.

oscargus:
Regarding browse I believe it can be useful in combination with filespecific fields (which we currently try to avoid, so I'm fine with the removal).

stefan-kolb:
Double-click might be a neat feature, but is neither intuitive from a UI usage perspective nor shown anywhere. So it is just a feature that can be used by insiders which is really not what I think is our intention.

oscargus:
Maybe. What about a tool tip text then? I would expect requests to get it back if we remove it.

stefan-kolb:
Intuitive to me would be 2 options:
- Adding a open button to the right (ugly)
- Highlighting the text so that one can see it is clickable (problems when one wants to only edit it)

Siedlerchr:
I use that doubleclick feature, too. It is natural to expect Jabref to open the link etc then.
I agree with @oscargus, to me a tooltip à la "Double click to open URL/DOI"

oscargus:
Ahh, I realise that I have been talking about the entry table and not the
entry editor (which I now realise the discussion most likely is about).

Still, I think that a tooltip might do the trick. One can also double click
the date field to set the current date and I think the owner field to
update the owner.

stefan-kolb:
>  It is natural to expect Jabref to open the link etc then.

To be honest I've never experienced that double clicking a text input field opens a link or does anything with the text except marking it.

Siedlerchr:
@stefan-kolb  I thought you meant the Entry Table DOI Field

stefan-kolb:
I mean this DOI field:
image

tobiasdiez:
I would propose to add an inline-button, in the textbox, similar to something like the clear symbol here

which appears on mouse-over and opens the url. Not sure if this can be done with Swing or should wait until JavaFX. Until then an ugly `Open` button might be an interim solution.

matthiasgeiger:
@tobiasdiez We have already discussed with the "clear search" button. Unfortunately, this is no built in function of Swing and requires some quick ugly hacks to implement. But I agree, in general that would be the nicest solution.
However, for the moment I would add an "open" button - especially as @stefan-kolb will already add another button for the "fetch from DOI" logic.

stefan-kolb:
Ok, this is ready for your comments!

src/main/java/net/sf/jabref/gui/menus/RightClickMenu.java
```diff
@@ -171,14 +171,6 @@ public RightClickMenu(JabRefFrame frame, BasePanel panel) {
             }
         });
 
-        add(new GeneralAction(Actions.MERGE_DOI, Localization.lang("Get BibTeX data from DOI")) {
```
oscargus: I so prefer it to stay here (as well). This has been one of my most used features recently.

src/main/java/net/sf/jabref/gui/entryeditor/EntryEditor.java
```diff
@@ -463,7 +462,7 @@ public void rebuildPanels() {
                     fieldExtras.contains(BibtexSingleFieldProperties.DATEPICKER));
         } else if (fieldExtras.contains(BibtexSingleFieldProperties.EXTERNAL)) {
             // Add external viewer listener for "pdf" and "url" fields.
```
oscargus: I'm wondering if the comment is correct for "pdf" nowadays...

src/main/java/net/sf/jabref/gui/entryeditor/FieldExtraComponents.java
```diff
@@ -15,41 +15,40 @@
...
 import net.sf.jabref.logic.journals.JournalAbbreviationRepository;
 import net.sf.jabref.logic.l10n.Localization;
+import net.sf.jabref.logic.net.URLUtil;
+import net.sf.jabref.logic.util.DOI;
 import net.sf.jabref.logic.util.date.EasyDateFormat;
 import net.sf.jabref.model.database.BibDatabaseMode;
 import net.sf.jabref.model.entry.BibEntry;
 import net.sf.jabref.model.entry.MonthUtil;
 
+import javax.swing.*;
```
oscargus: This seems to be yet another ordering of imports. Not a problem as such, just interesting.

oscargus:
Looks good to me! I'd really like to keep the right-click menu though.

With this PR the use of the entry editor has slightly changed from editing to actually using the information (although it was hidden there earlier through double-click). Not sure if it is worth thinking about what we really want the entry editor for.

src/main/java/net/sf/jabref/gui/menus/RightClickMenu.java
stefan-kolb: Next step would be to add https://github.com/JabRef/jabref/issues/1098, so I prefer not to add it back in this context menu.

simonharrer: Hm, maybe rename it differently using "autocomplete with BibTeX data from DOI"?

simonharrer:
:+1: LGTM

src/main/java/net/sf/jabref/gui/menus/RightClickMenu.java
oscargus: I'm not really seeing the problem. A button somewhere else won't help the (superior) habit of right-clicking on an entry and selecting in a menu compared to double-clicking to open the editor, change to the right tab and clicking a button. Marking, click on another button is slightly better, but I still do not see what the problem with the right-click menu staying is. "New from DOI" is not going to end up in the right-click menu? 

I thought the purpose of #1098 was to add e new entry (which IMHO would be better splitting web search and "add/fetch") so I do not see marking and clicking that button as a replacement. 

The context here is that one do get slightly different information when searching and when adding information from DOI, so I usually do both at once. Not to mention improving all my old bib-files.

simonharrer: No worries. :) Action in right click menu stayed in there. :trophy: 

The right click menu now only contains actions related to selected entries in the table. Hence, the "new from doi" was moved to the menu.

PR #1077, opened 2016-04-01
[WIP] Refactor auto completion

stefan-kolb:
__Major changes in this PR:__
- [x] Only use auto-completion when at the end of a word (space after)
- [x] Detection modes: Only use case-sensitive detection for now (performance, logic, we can re-add functionality from here)
- [x] Remove method getPrefix() - Intention of this method was unknown
- [x] Use Nameautocompleter for every PERSONNAME BibTeX field, not only author & editor

### DIFF
--- a/src/main/java/net/sf/jabref/gui/autocompleter/AutoCompleteState.java
+++ b/src/main/java/net/sf/jabref/gui/autocompleter/AutoCompleteState.java
@@ -1,0 +1,65 @@
+package net.sf.jabref.gui.autocompleter;
+
+import java.util.List;
+
+public class AutoCompleteState {
+    private List<String> suggestions;
+    private String currentWord;
+    private int currentCompletion;
+    private int realCaretPosition;
+
+    public AutoCompleteState(String currentWord, List<String> suggestions, int caretPosition) {
+        this.currentWord = currentWord;
+        this.suggestions = suggestions;
+        realCaretPosition = caretPosition;
+        currentCompletion = 0;
+    }
+
+    public String nextSuggestion() {
+        return cycleSuggestion(1);
+    }
+
+    public String previousSuggestion() {
+        return cycleSuggestion(-1);
+    }
+
+    public String getCurrentWord() {
+        return currentWord;
+    }
+
+    public boolean continuesSuggestion(char newChar) {
+        if (!getSuggestedWord().startsWith(currentWord + newChar)) {
+            return false;
+        }
+        return true;
+    }
+
+    public void updateState(char newChar) {
+        if (continuesSuggestion(newChar)) {
+            currentWord = currentWord + newChar;
+            realCaretPosition++;
+        } else {
+            throw new IllegalArgumentException("Make sure continuesSuggestion(newChar) is true before using this method!");
+        }
+    }
+
+    public int getRealCaretPosition() {
+        return realCaretPosition;
+    }
+
+    public String getSuggestedWord() {
+        return suggestions.get(currentCompletion);
+    }
+
+    private String cycleSuggestion(int increment) {
+        currentCompletion += increment;
+
+        if (currentCompletion >= suggestions.size()) {
+            currentCompletion = 0;
+        } else if (currentCompletion < 0) {
+            currentCompletion = suggestions.size() - 1;
+        }
+
+        return getSuggestedWord();
+    }
+}
--- a/src/main/java/net/sf/jabref/gui/autocompleter/AutoCompleteState.java
+++ b/src/main/java/net/sf/jabref/gui/autocompleter/AutoCompleteState.java
@@ -29,7 +29,7 @@ public class AutoCompleteState {
     }
 
     public boolean continuesSuggestion(char newChar) {
-        return !getSuggestedWord().startsWith(currentWord + newChar);
+        return getSuggestedWord().startsWith(currentWord + newChar);
     }
 
     public void updateState(char newChar) {
--- a/src/main/java/net/sf/jabref/gui/autocompleter/AutoCompleteState.java
+++ b/src/main/java/net/sf/jabref/gui/autocompleter/AutoCompleteState.java
@@ -29,10 +29,7 @@ public class AutoCompleteState {
     }
 
     public boolean continuesSuggestion(char newChar) {
-        if (!getSuggestedWord().startsWith(currentWord + newChar)) {
-            return false;
-        }
-        return true;
+        return !getSuggestedWord().startsWith(currentWord + newChar);
     }
 
     public void updateState(char newChar) {
--- a/src/main/java/net/sf/jabref/gui/autocompleter/AutoCompleteState.java
+++ b/src/main/java/net/sf/jabref/gui/autocompleter/AutoCompleteState.java
@@ -1,16 +1,17 @@
 package net.sf.jabref.gui.autocompleter;
 
 import java.util.List;
+import java.util.Objects;
 
 public class AutoCompleteState {
-    private List<String> suggestions;
+    private final List<String> suggestions;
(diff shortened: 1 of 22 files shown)


stefan-kolb:
The current auto completion has two modes:
- a case sensitive mode
- a case insensitive mode

The case sensitive mode is activated when one types a word that has mixed case, e.g., `gZhhUh`.
If the word is only lowercase the auto completion will perform a case insensitive mode.

The case sensitive mode is straight-forward: It will lookup a word that can be completed from the current partial word.
Case insensitive lookup tries to find a word that can be completed from the input based on ignore case search. If it found one, it will still present the word in its original casing, e.g., `some -> Something`

__Discussion point__
What is the purpose of the case-insensitive mode? The detection (only lowercase in word) seems arbitrary to me. How should the completion be performed in general or what are typical algorithms for this? I would either suggest case-sensitive or a possibility to configure this but no auto detection?!

stefan-kolb:
Ok, I got something working here now. Massively refactored the code (god was this hard to read).
I'm still not quite happy with the result but I hope this will be a starting point for more improvements with a readable codebase.

@JabRef/developers WDYT?

src/main/java/net/sf/jabref/gui/autocompleter/AutoCompleteState.java
```diff
@@ -0,0 +1,65 @@
+package net.sf.jabref.gui.autocompleter;
+
+import java.util.List;
+
+public class AutoCompleteState {
+    private List<String> suggestions;
+    private String currentWord;
+    private int currentCompletion;
+    private int realCaretPosition;
```
simonharrer: make everything final?

src/main/java/net/sf/jabref/gui/autocompleter/AutoCompleteState.java
```diff
@@ -0,0 +1,65 @@
...
+
+    public AutoCompleteState(String currentWord, List<String> suggestions, int caretPosition) {
+        this.currentWord = currentWord;
+        this.suggestions = suggestions;
```
simonharrer: Use Objects.requireNonNull for both values like this.field = requireNonNull(field);

src/main/java/net/sf/jabref/logic/autocompleter/DefaultAutoCompleter.java
```diff
@@ -51,16 +48,19 @@ public boolean isSingleUnitField() {
...
             return;
         }
 
-        entry.getFieldOptional(fieldName).ifPresent(fieldValue -> {
-            StringTokenizer tok = new StringTokenizer(fieldValue, SEPARATING_CHARS);
-            while (tok.hasMoreTokens()) {
-                addItemToIndex(tok.nextToken());
+        Optional<String> fieldValue = entry.getFieldOptional(fieldName);
+
+        if(fieldValue.isPresent()) {
```
simonharrer: use fieldValue.ifPresent instead

src/main/java/net/sf/jabref/logic/autocompleter/NameFieldAutoCompleter.java
```diff
@@ -86,14 +79,16 @@ public boolean isSingleUnitField() {
...
             return;
         }
+
         for (String fieldName : fieldNames) {
-            if (entry.hasField(fieldName)) {
-                String fieldValue = entry.getField(fieldName);
-                AuthorList authorList = AuthorList.parse(fieldValue);
+            Optional<String> fieldValue = entry.getFieldOptional(fieldName);
+
+            if (fieldValue.isPresent()) {
```
simonharrer: use fieldValue.ifPresent here

stefan-kolb: Hm I don't like the code readability with ifPresent() because the body includes several line sof code.

src/main/java/net/sf/jabref/logic/autocompleter/DefaultAutoCompleter.java
stefan-kolb: See below.

src/main/java/net/sf/jabref/logic/autocompleter/NameFieldAutoCompleter.java
simonharrer: Hm, we have different opinions on that, I see. :)

simonharrer:
LGTM :+1:

stefan-kolb:
@JabRef/developers Does it work for you? Should we merge this in or any further suggestions?

oscargus:
Would it make sense for the NameAutoCompleter to work for all fields with BibtexSingleFieldProperties.PERSON_NAMES set? I do not really know how it works today, but I seem to recall that it wasn't obvious if the same AutoCompleter was used for both "author" and "editor". (Haven't had time to test it yet.)

stefan-kolb:
Would be possible, currently it onl works on author and editor.
```
if ("author".equals(fieldName) || "editor".equals(fieldName)) {
             return new NameFieldAutoCompleter(fieldName, preferences);
```

stefan-kolb:
I quick checked. Currently only those two fields have the property `BibtexSingleFieldProperties.PERSON_NAMES` set.

oscargus:
Correct but there's is a PR (don't know the number now) which adds 20 or so
fields.

stefan-kolb:
Ah ok. We could do something like this:
```
if (InternalBibtexFields.getFieldExtras(fieldName).contains(BibtexSingleFieldProperties.PERSON_NAMES)) {
            return new NameFieldAutoCompleter(fieldName, preferences);
        }
```

But I get NullpointerExceptions in the tests due to the InternalBibtexFields relying on Globals.prefs (InternalBibtexFields.java:125)

oscargus:
OK! These dependencies are annoying... Mock JabRefPreferences?

koppor:
> What is the purpose of the case-insensitive mode? The detection (only lowercase in word) seems arbitrary to me.

Is this still a valid question? If yes: I was thinking that casing causes more keys to type (shift key). Thus: being lazy should instruct JabRef to search for everything. Being non-lazy (capital letter), I want the tool to search only for these cases as I demand something specific.

In [powerline-shell](https://github.com/milkbikis/powerline-shell) or Windows CMD, the search IMHO also is case-insensitive.

I have not researched for typical algorithms the recent time. When refactoring the auto completion a few years ago, I found nothing useful. However, I did not document my search.

koppor:
Quick test:

 * author field
 * `Kwantes, PieterM. and Van Gorp, Pieter and Kleijn, Jetty and Rensink, Arend`
 * Change `V` in `Van Gorp` to `v`
 * go to the end of the line, type `and v`.
 * One cannot type `v`

 * title field
 * Text: `Towards Compliance Verification Between Global and Local Process Models`
 * Go to the end of the entry
 * type ` and`
 * ` about` is displayed and the accepted characters are `abo` and not the typed `and`.

The typed characters have to be kept! The autocompletion must not override the user's input.

I can send you my large bib file privatly. Don't want to share it publicly.

koppor:
```
Exception in thread "AWT-EventQueue-0" java.lang.StringIndexOutOfBoundsException: String index out of range: -67
        at java.lang.String.substring(String.java:1931)
        at net.sf.jabref.gui.autocompleter.AutoCompleteListener.insertKeyAndSuggestion(AutoCompleteListener.java:148)
        at net.sf.jabref.gui.autocompleter.AutoCompleteListener.keyTyped(AutoCompleteListener.java:104)
        at java.awt.Component.processKeyEvent(Component.java:6490)
```

koppor:
This PR is superseeded by #1136

PR #1014, opened 2016-03-21
Improved Exception Handling in Tests cases

Siedlerchr:
Never catch an Exception
Removed unnecessary Exceptions
Throw specific Exceptions, never only Exception
Ref #1012 

- [x] Changes in pull request outlined? (What, why, ...)
- [x] Tests created for changes?
- [x] Tests green?

PR #1079, opened 2016-04-01
Improded readability of some BibTex Test Strings

Siedlerchr:
Fixed encoding problem in Test when running gradle check
Follow up from #1014 

@tobiasdiez


tobiasdiez:
Thanks!

PR #1081, opened 2016-04-01
More Swedish translations and unified strings

oscargus:


src/main/java/net/sf/jabref/external/DroppedFileHandler.java
```diff
@@ -345,7 +345,7 @@ public void stateChanged(ChangeEvent arg0) {
 
         linkInPlace.setText(Localization.lang("Leave file in its current directory."));
         copyRadioButton.setText(Localization.lang("Copy file to file directory."));
-        moveRadioButton.setText(Localization.lang("Move file to file directory."));
+        moveRadioButton.setText(Localization.lang("Move file to file directory") + '.');
```
tobiasdiez: The punctuation should be part of the language string as some languages have different symbols for the full stop (Chinese comes to my mind as an example).

oscargus: I know the concept, but in this case I removed one of

"Move file to file directory"
"Move file to file directory."
"Move file to file directory?"

Not sure how to balance the number of strings and correctness...

tobiasdiez: I would remove the first one and keep the other ones. Some languages also have a different sentence structure for questions...

oscargus: Fair enough. A dialog title will end with a period, but I can live with that.

oscargus: Turned out it was a menu item, so better to actually remove the period for the radio button option.

PR #1082, opened 2016-04-01
Sorted out some exceptions

oscargus:
Fewer thrown RunTimeExceptions and fewer caught Throwables.

### DIFF
--- a/src/main/java/net/sf/jabref/gui/keyboard/KeyBindingPreferences.java
+++ b/src/main/java/net/sf/jabref/gui/keyboard/KeyBindingPreferences.java
@@ -65,7 +65,7 @@ public class KeyBindingPreferences {
             if (SHORTCUT_MASK == -1) {
                 try {
                     SHORTCUT_MASK = Toolkit.getDefaultToolkit().getMenuShortcutKeyMask();
-                } catch (Throwable ignored) {
+                } catch (AWTError | HeadlessException ignored) {
                     // Ignored
                 }
             }
--- a/src/main/java/net/sf/jabref/gui/keyboard/KeyBindingPreferences.java
+++ b/src/main/java/net/sf/jabref/gui/keyboard/KeyBindingPreferences.java
@@ -4,6 +4,10 @@ import net.sf.jabref.JabRefPreferences;
 import net.sf.jabref.logic.util.OS;
 
 import javax.swing.*;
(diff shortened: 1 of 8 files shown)


src/main/java/net/sf/jabref/gui/keyboard/KeyBindingPreferences.java
```diff
@@ -65,7 +65,7 @@ private KeyStroke getKeyForMac(KeyStroke ks) {
             if (SHORTCUT_MASK == -1) {
                 try {
                     SHORTCUT_MASK = Toolkit.getDefaultToolkit().getMenuShortcutKeyMask();
-                } catch (Throwable ignored) {
+                } catch (AWTError | HeadlessException ignored) {
```
Siedlerchr: I would recommend at least logging the exception instead of ignoring it.

src/main/java/net/sf/jabref/importer/fetcher/ScienceDirectFetcher.java
```diff
@@ -114,20 +113,16 @@ public boolean processQuery(String query, ImportInspector dialog, OutputPrinter
...
-            urlQuery = ScienceDirectFetcher.SEARCH_URL + URLEncoder.encode(query, StandardCharsets.UTF_8.name());
-            int count = 1;
-            String nextPage;
-            while (((nextPage = getCitationsFromUrl(urlQuery, ids)) != null)
-                    && (count < ScienceDirectFetcher.MAX_PAGES_TO_LOAD)) {
-                urlQuery = nextPage;
-                count++;
-            }
-            return ids;
-        } catch (UnsupportedEncodingException e) {
```
tobiasdiez: Does the caller of getCitations somewhere catches the runtimeException? If yes, then this should be adopted too.

oscargus:
@Siedlerchr: yes, I thought of that independently. :-)

@tobiasdiez: not that I've seen. I cannot really figure out why this was done in the first place. With the change, the UnsupportedEncodingException is caught among other potential fetching problems with IOException in the next higher method and it is returned that it wasn't able to fetch.  Not obvious why one would like to bypass this and send a RuntimeException up among the calls.

tobiasdiez:
Ok, then this can be merged in my opinion.

PR #1083, opened 2016-04-02
Metadata cleanup

tobiasdiez:
- Move file handling from MetaData to BibDatabaseContext which resulted in a lot of classes using the database context instead of the metadata
- Remove Globals.prefs.databaseFile

--


src/main/java/net/sf/jabref/JabRef.java
```diff
@@ -350,7 +350,7 @@ public void start(String[] args) {
                     }
                     BibDatabaseContext databaseContext = pr.getDatabaseContext();
                     databaseContext.setDatabaseFile(theFile);
-                    Globals.prefs.fileDirForDatabase = databaseContext.getFileDirectory(Globals.FILE_FIELD).toArray(new String[0]);
+                    Globals.prefs.fileDirForDatabase = databaseContext.getFileDirectory().toArray(new String[0]);
```
Siedlerchr: I don't like the toArray stuff. I suggest you change the Array in The GlobalPrefs to a type of List<String>, as Array also implements the List Interface. Then you can omit the toArray stuff. 
And for that 
`                            dirs = new String[] {Globals.prefs.get(Globals.FILE_FIELD + Globals.DIR_SUFFIX)};
`
simply use Arrays.asList

tobiasdiez: Yes you are right this should be a list. I even think this field should be removed completely. However, since its usage is hidden deeply in some layout logic, this will be a bigger project...

simonharrer:
Nice work.

Some comments:
- Why did you disable the test? Why did it fail on the CI server?
- Due to having a the BibDatabaseContext, one could remove a lot of unused fields for the metadata or the bibdatabase (see codacy checks)
- The new method about determining the file directories within the BibDatabaseContext uses the preferences a lot which we did not want in the logic part. I would not change this at the moment, but just to keep in mind for further changes, as they should be moved to the Defaults class which encapsulates these preferences for a particular BibDatabaseContext. 
- It would be nice to use `Path` instead of `File` which is the new API since Java 7. Maybe something for our cleanup advocate @oscargus ? :)

tobiasdiez:
To be honest, I have no idea why the test fails. Apparently it can't find the build.gradle file...I guess that the mocking fails, maybe because of different preferences. Is it ok if I merge this PR in and create a new issue for the test and the usage of global preferences in BibDatabaseContext (since I think both issues are related in the end)?

simonharrer:
Hm, I do not like that, as there is a reason why the test fails. And this functionality is critical. Master should be stable.

Siedlerchr:
@tobiasdiez Could you provide the full error message of the failing test? Maybe someone has an idea how this could be fixed

tobiasdiez:
As I thought, in the end it was only a problem with the mock. Should be fixed now.

PR #1086, opened 2016-04-02
Fixed some things related to localization

oscargus:
Fixed a number of TODOs/FIXMEs.


src/main/java/net/sf/jabref/external/AttachFileAction.java
```diff
@@ -25,7 +26,8 @@ public AttachFileAction(BasePanel panel) {
     @Override
     public void action() {
         if (panel.getSelectedEntries().size() != 1) {
-            return; // TODO: display error message?
+            panel.output(Localization.lang("No entries or multiple entries selected."));
```
tobiasdiez: Add that the action was aborted.

src/main/java/net/sf/jabref/external/FindFullTextAction.java
```diff
@@ -93,7 +95,7 @@ public void update() {
             basePanel.output(Localization.lang("Finished downloading full text document"));
         }
         else {
-            String message = Localization.lang("Full text article download failed");
+            String message = Localization.lang("Full text document download failed");
```
tobiasdiez: just remove `document`?

tobiasdiez:
LGTM :+1:

src/main/java/net/sf/jabref/external/AttachFileAction.java
oscargus: We should look over that in general. The next action (in the BasePanel (?) code) has exactly the same message. There are also a few related translations strings. For example ~"You must select exactly one entry".

Also, one can consider if a dialog should be shown, possibly with a "do not show this again check box", for this and many other issues. Drawback is that the check box selection must be stored somehow, but maybe one can create a dedicated dialog type for that?

src/main/java/net/sf/jabref/external/FindFullTextAction.java
oscargus: For consistency I think document should be there (and that document is the correct word). Many of the translations already use document by the way.

src/main/java/net/sf/jabref/external/AttachFileAction.java
tobiasdiez: I would prefer if you could clean up all the related messages as well in this PR. Right now the output is just a statement and not an error message.

src/main/java/net/sf/jabref/external/FindFullTextAction.java
oscargus: I changed this text to "This operation requires exactly one item to be selected." (new string) and also replaced "No entries selected." with "This operation requires one or more entries to be selected." (already existing).

simonharrer:
Are you done? @oscargus :) Stefan and myself are currently looking at the pull requests and trying to reduce the huge amount (>30) by merging them when suitable.

oscargus:
Now I think I'm done. :-) If nothing else I have teaching now...

PR #1089, opened 2016-04-03
FileList field size now dependent on its content

### LINKED ISSUE #672: Fix width of file fields
grabbed_20160115-023141

Refs #113, #98 

(request as of 20150326 by stekks)

bartsch-dev:
Fixes #672.
The FileList (within the EntryEditor) now automatically resizes its Columns dependent on the length of its content.

java 2016-04-03 12-13-50-65

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,8 +54,8 @@ to [sourceforge feature requests](https://sourceforge.net/p/jabref/features/) by
 - Moved "Get BibTeX data from DOI" from main table context menu to DOI field in entry editor
 - Added open buttons to DOI and URL field
 - Move Look & Feel settings from advanced to appearance tab in preferences
-
 - Move PDF file directory configuration from external tab to file tab in preferences
+- Implemented [#672](https://github.com/JabRef/jabref/issues/672): FileList now distributes its space dependent on the width of its columns
 
 ### Fixed
 - Fixed [#318](https://github.com/JabRef/jabref/issues/318): Improve normalization of author names
--- a/src/main/java/net/sf/jabref/gui/entryeditor/EntryEditor.java
+++ b/src/main/java/net/sf/jabref/gui/entryeditor/EntryEditor.java
@@ -34,40 +34,61 @@ import java.beans.VetoableChangeListener;
 import java.io.IOException;
 import java.io.StringReader;
 import java.io.StringWriter;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
-import javax.swing.*;
+import javax.swing.AbstractAction;
+import javax.swing.Action;
+import javax.swing.ActionMap;
+import javax.swing.InputMap;
+import javax.swing.JButton;
+import javax.swing.JComponent;
+import javax.swing.JLabel;
+import javax.swing.JOptionPane;
+import javax.swing.JPanel;
+import javax.swing.JPopupMenu;
+import javax.swing.JScrollPane;
+import javax.swing.JTabbedPane;
+import javax.swing.JTextArea;
+import javax.swing.JToolBar;
+import javax.swing.KeyStroke;
+import javax.swing.ScrollPaneConstants;
+import javax.swing.SwingConstants;
+import javax.swing.SwingUtilities;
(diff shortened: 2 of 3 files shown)


src/main/java/net/sf/jabref/gui/entryeditor/EntryEditor.java
```diff
@@ -15,70 +15,61 @@
...
+import java.beans.PropertyChangeEvent;
+import java.beans.VetoableChangeListener;
+import java.io.IOException;
+import java.io.StringReader;
+import java.io.StringWriter;
+import java.util.*;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
```
boceckts: Please change as little of the imports as possible. For example your remove the IOException and add it later again.

src/main/java/net/sf/jabref/gui/fieldeditors/FileListEditor.java
```diff
@@ -15,53 +15,44 @@
...
+import java.awt.event.MouseEvent;
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Optional;
+
+
```
boceckts: See comment above. Also please don't add unnecessary blank lines. Try to change only the lines that are really necessary.

src/main/java/net/sf/jabref/gui/entryeditor/EntryEditor.java
oscargus: My guess is that it is some switch in the IDE that does this as the result is sorted. A bit "annoying" now, but might simplify things in the longer run.

Personally, I do not have strong opinions on what is the correct thing here, but others might... (And I know it is only for internal StuPro review so far... ;-))

bartsch-dev: I'm using the standard style from IntelliJ IDEA, maybe someone should create a template for that (I thought there was one, but I can't find it right now).

Siedlerchr: Eclipse has an option to sort and manage imports (e.g. delete one).
And one important thing: Only add the specific imports, e.g. java.util.List instead of whole packages. 

@bartsch-dev  Have a look at here:
http://stackoverflow.com/questions/14716283/is-it-possible-for-intellij-to-organize-imports-the-same-way-as-in-eclipse
http://stackoverflow.com/questions/11704327/intellij-idea-how-to-optimize-imports-automatically-after-each-save

oscargus: I do not know what the "official" standpoint on * vs specific packages is.
I sort of prefer specific packages as you can see what dependencies you add
or remove, but I know that many of the other developers like (or at least
use) *.

(In this case both util.* and util.List looks a bit weird, but is required
as there is awt.List...)

Siedlerchr: I would recommend only using full qualified imports, easier to avoid conflicts, e.g. you import java.a* and google.com.b* and both contain different classes with the same name.

bartsch-dev:
I rearranged the Imports a bit (using the settings @tobiasdiez describes in #1091).

lenhard:
Looks good to me me! @JabRef/developers anybody else any comments? Maybe we could integrate that in v3.3 already?

CHANGELOG.md
```diff
@@ -96,6 +96,8 @@ to [sourceforge feature requests](https://sourceforge.net/p/jabref/features/) by
 - Fixed [#535](https://github.com/JabRef/jabref/issues/535): Add merge action to right click menu
 - Fixed [#1115](https://github.com/JabRef/jabref/issues/1115): Wrong warning message when importing duplicate entries
 - Fixed [#935](https://github.com/JabRef/jabref/issues/935): PDFs, which are readable, but carry a protection for editing, are treated by the XMP parser and the importer generating a BibTeX entry based on the content.
+- Fixed [#672](https://github.com/JabRef/jabref/issues/672): FileList now distributes its space dependent on the width of its columns
```
matthiasgeiger: As #672 is no bug this should be added to the "Changes" section.

matthiasgeiger:
Apart from my comment regarding the CHANGELOG this is fine and can be merged.

bartsch-dev:
Moved the changelog entry from `fixed` to `changed`.

PR #1090, opened 2016-04-03
Cleanup of basepanel and mainframe

tobiasdiez:
Just made some classes or methods private. Also deleted unused code.

### DIFF
--- a/src/main/java/net/sf/jabref/gui/BasePanel.java
+++ b/src/main/java/net/sf/jabref/gui/BasePanel.java
@@ -432,8 +432,6 @@ public class BasePanel extends JPanel implements ClipboardOwner, FileUpdateListe
                     }
                     ce.end();
                     undoManager.addEdit(ce);
-                    //entryTable.clearSelection();
-                    //entryTable.revalidate();
                     output(formatOutputMessage(Localization.lang("Pasted"), bes.size()));
                     markBaseChanged();
 
@@ -583,7 +581,6 @@ public class BasePanel extends JPanel implements ClipboardOwner, FileUpdateListe
         // The action for auto-generating keys.
         actions.put(Actions.MAKE_KEY, new AbstractWorker() {
 
-            //int[] rows;
             List<BibEntry> entries;
             int numSelected;
             boolean cancelled;
@@ -1256,7 +1253,6 @@ public class BasePanel extends JPanel implements ClipboardOwner, FileUpdateListe
                         ((GroupTreeNode) tree.getLastPathComponent()).addToGroup(entries);
                     }
                 }
-                //BasePanel.this.updateEntryEditorIfShowing(); // doesn't seem to be necessary
                 SwingUtilities.invokeLater(() -> BasePanel.this.getGroupSelector().valueChanged(null));
             }
         }
@@ -1335,7 +1331,6 @@ public class BasePanel extends JPanel implements ClipboardOwner, FileUpdateListe
     }
 
     private void createMainTable() {
-        //Comparator comp = new FieldComparator("author");
 
         final GlazedEntrySorter eventList = new GlazedEntrySorter(database.getEntries());
         // Must initialize sort columns somehow:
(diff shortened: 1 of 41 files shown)


src/main/java/net/sf/jabref/gui/BasePanel.java
```diff
@@ -227,7 +219,7 @@ public String getTabTitle() {
         if (getBibDatabaseContext().getDatabaseFile() == null) {
             title.append(GUIGlobals.UNTITLED_TITLE);
 
-            if (!database().getEntries().isEmpty()) {
+            if (!getDatabase().getEntries().isEmpty()) {
```
oscargus: hasEntries()? (Or whatever it is called...)

src/main/java/net/sf/jabref/gui/BasePanel.java
```diff
@@ -1872,17 +1856,6 @@ private synchronized void markChangedOrUnChanged() {
         frame.setWindowTitle();
     }
 
-    /**
-     * Selects a single entry, and scrolls the table to center it.
-     *
-     * @param pos Current position of entry to select.
-     */
-    public void selectSingleEntry(int pos) {
```
oscargus: Is there a corresponding method that does this? I assume that it was unused, but I guess that there are a number of situations where one would like to use it: click on ParserResult warnings, incorrect encoding on save etc.

On the other hand, it should be quite easy to reimplemented and ideally one would like to call it with an entry I assume.

src/main/java/net/sf/jabref/gui/actions/MassSetFieldAction.java
```diff
@@ -15,24 +15,24 @@
  */
 package net.sf.jabref.gui.actions;
 
-import java.awt.BorderLayout;
+import java.awt.*;
```
oscargus: I assume that your editor does this, but in this case I think java.util.* would make more sense.

oscargus:
Looks good! Just some minor comments.

src/main/java/net/sf/jabref/gui/BasePanel.java
tobiasdiez: Yes, this method was not used. Probably a similar logic is implemented in the places you mentioned and thus it would be easy to refactor-out a corresponding method. But even then BasePanel is probably not the right place for this method.

src/main/java/net/sf/jabref/gui/BasePanel.java
lenhard: Surprisingly, the method is called `hasNoEntries`. Maybe we should rename that to `isEmpty` directly? So we can call `getDatabase().isEmpty()`. @tobiasdiez Could you do this in this PR?

oscargus: No! There are also strings and a preamble in a database. :-)

(It was called isEmpty for a while though...)

src/main/java/net/sf/jabref/gui/BasePanel.java
```diff
@@ -2241,25 +2210,21 @@ public void fileUpdated() {
...
+            if (hasAlready) {
+                sidePaneManager.hideComponent(FileUpdatePanel.NAME);
+                sidePaneManager.unregisterComponent(FileUpdatePanel.NAME);
+            }
+            FileUpdatePanel pan = new FileUpdatePanel(BasePanel.this, sidePaneManager,
+                    getBibDatabaseContext().getDatabaseFile(), scanner);
+            sidePaneManager.register(FileUpdatePanel.NAME, pan);
+            sidePaneManager.show(FileUpdatePanel.NAME);
+            //setUpdatedExternally(false);
+            //scanner.displayResult();
```
lenhard: Since this is a cleanup PR, could you also remove those commented-out lines?

src/main/java/net/sf/jabref/gui/BasePanel.java
lenhard: Ah, good point, thanks! In that case just use `hasNoEntries`.

lenhard:
Apart from the minor comments, looks good to me. Just fix these and this PR is good to go.

Regarding the import statements: It reads a little strange in this PR, as sometimes the statements are coalesced to a `*` and sometimes they are expanded (which is probably done by your IDE). However, we have no policy on that and I am not aware of a generally-accepted policy or performance-related benefits of either version. So it doesn't really matter from my point of view.

oscargus:
I think expanded is quite nice since it shows exactly which dependencies
are added or removed. On the other hand they lead to quite a lot of "unused
import" warnings...

tobiasdiez:
Thanks for your comments. I will address them later this evening. 
Regarding the import stuff, I think we should discuss it at  #1091. I have absolutely no opinion about combining them or the right sort order (I never really cared about the imports so far), but I think we should come up with an accepted standard so that not every PR resorts imports...

src/main/java/net/sf/jabref/gui/BasePanel.java
tobiasdiez: I think the positive version `hasEntries` is more convenient and avoids the double negative `! hasNoEntries()`. If nobody veto against it, I will change it to `hasEntries` later.

PR #1093, opened 2016-04-03
Fix open folder action in Windows not working correctly...

Siedlerchr:
when direcory path contained whitespaces
Use Path-class and methods
"& Symbol" does not cause problems on Windows
Use Path.getParent instead of ugly substring-hack

### DIFF
--- a/src/main/java/net/sf/jabref/gui/desktop/JabRefDesktop.java
+++ b/src/main/java/net/sf/jabref/gui/desktop/JabRefDesktop.java
@@ -33,21 +33,17 @@ import java.util.regex.Pattern;
  * http://stackoverflow.com/questions/18004150/desktop-api-is-not-supported-on-the-current-platform
  */
 public class JabRefDesktop {
-    private static final NativeDesktop ND_LINUX = new Linux();
-    private static final NativeDesktop ND_WINDOWS = new Windows();
-    private static final NativeDesktop ND_MAC = new OSX();
-    private static final NativeDesktop ND_DEFAULT = new DefaultDesktop();
+
     private static final NativeDesktop NATIVE_DESKTOP = getNativeDesktop();
-
     private static final Log LOGGER = LogFactory.getLog(JabRefDesktop.class);
-
     private static final Pattern REMOTE_LINK_PATTERN = Pattern.compile("[a-z]+://.*");
+
 
     /**
      * Open a http/pdf/ps viewer for the given link string.
      */
-    public static void openExternalViewer(BibDatabaseContext databaseContext, String initialLink, String initialFieldName)
-            throws IOException {
+    public static void openExternalViewer(BibDatabaseContext databaseContext, String initialLink,
+            String initialFieldName) throws IOException {
         String link = initialLink;
         String fieldName = initialFieldName;
         if ("ps".equals(fieldName) || "pdf".equals(fieldName)) {
@@ -74,7 +70,7 @@ public class JabRefDesktop {
             }
         } else if ("doi".equals(fieldName)) {
             Optional<DOI> doiUrl = DOI.build(link);
-            if(doiUrl.isPresent()) {
+            if (doiUrl.isPresent()) {
                 link = doiUrl.get().getURLAsASCIIString();
             }
             // should be opened in browser
@@ -171,20 +167,20 @@ public class JabRefDesktop {
 
         String cancelMessage = Localization.lang("Unable to open file.");
         String[] options = new String[] {Localization.lang("Define '%0'", fileType.getName()),
-                Localization.lang("Change file type"),
-                Localization.lang("Cancel")};
+                Localization.lang("Change file type"), Localization.lang("Cancel")};
         String defOption = options[0];
-        int answer = JOptionPane.showOptionDialog(frame, Localization.lang("This external link is of the type '%0', which is undefined. What do you want to do?",
+        int answer = JOptionPane.showOptionDialog(frame,
+                Localization.lang("This external link is of the type '%0', which is undefined. What do you want to do?",
                         fileType.getName()),
                 Localization.lang("Undefined file type"), JOptionPane.YES_NO_CANCEL_OPTION,
                 JOptionPane.QUESTION_MESSAGE, null, options, defOption);
         if (answer == JOptionPane.CANCEL_OPTION) {
             frame.output(cancelMessage);
             return false;
-        }
-        else if (answer == JOptionPane.YES_OPTION) {
+        } else if (answer == JOptionPane.YES_OPTION) {
             // User wants to define the new file type. Show the dialog:
-            ExternalFileType newType = new ExternalFileType(fileType.getName(), "", "", "", "new", IconTheme.JabRefIcon.FILE.getSmallIcon());
+            ExternalFileType newType = new ExternalFileType(fileType.getName(), "", "", "", "new",
+                    IconTheme.JabRefIcon.FILE.getSmallIcon());
             ExternalFileTypeEntryEditor editor = new ExternalFileTypeEntryEditor(frame, newType);
             editor.setVisible(true);
             if (editor.okPressed()) {
@@ -226,8 +222,7 @@ public class JabRefDesktop {
             if (editor.okPressed()) {
                 // Store the changes and add an undo edit:
                 String newValue = tModel.getStringRepresentation();
-                UndoableFieldChange ce = new UndoableFieldChange(entry, Globals.FILE_FIELD,
-                        oldValue, newValue);
+                UndoableFieldChange ce = new UndoableFieldChange(entry, Globals.FILE_FIELD, oldValue, newValue);
                 entry.setField(Globals.FILE_FIELD, newValue);
                 frame.getCurrentBasePanel().undoManager.addEdit(ce);
                 frame.getCurrentBasePanel().markBaseChanged();
@@ -266,20 +261,19 @@ public class JabRefDesktop {
             return;
         }
 
-        String absolutePath = file.getAbsolutePath();
-        absolutePath = absolutePath.substring(0, absolutePath.lastIndexOf(File.separator) + 1);
+        String absolutePath = file.toPath().toAbsolutePath().getParent().toString();
         NATIVE_DESKTOP.openConsole(absolutePath);
     }
 
     // TODO: Move to OS.java
     public static NativeDesktop getNativeDesktop() {
-        if(OS.WINDOWS) {
-            return ND_WINDOWS;
-        } else if(OS.OS_X) {
-            return ND_MAC;
-        } else if(OS.LINUX) {
-            return ND_LINUX;
+        if (OS.WINDOWS) {
+            return new Windows();
+        } else if (OS.OS_X) {
+            return new OSX();
+        } else if (OS.LINUX) {
+            return new Linux();
         }
-        return ND_DEFAULT;
+        return new DefaultDesktop();
     }
 }
--- a/src/main/java/net/sf/jabref/gui/desktop/os/DefaultDesktop.java
+++ b/src/main/java/net/sf/jabref/gui/desktop/os/DefaultDesktop.java
@@ -6,6 +6,7 @@ import org.apache.commons.logging.LogFactory;
 import java.awt.*;
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.Paths;
 
 public class DefaultDesktop implements NativeDesktop {
     private static final Log LOGGER = LogFactory.getLog(NativeDesktop.class);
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -52,6 +52,7 @@ to [sourceforge feature requests](https://sourceforge.net/p/jabref/features/) by
 - [#1023](https://github.com/JabRef/jabref/issues/492) ArXiv fetcher now also fetches based on eprint id
 
 ### Fixed
+- Fixed "Open-folder" action in Windows not opening tge correct folder when folder path contained whitespaces
```
tobiasdiez: Spelling mistake

src/main/java/net/sf/jabref/gui/desktop/JabRefDesktop.java
```diff
@@ -265,19 +267,17 @@ public static void openConsole(File file) throws IOException {
         if (file == null) {
             return;
         }
-
-        String absolutePath = file.getAbsolutePath();
-        absolutePath = absolutePath.substring(0, absolutePath.lastIndexOf(File.separator) + 1);
+        String absolutePath = file.toPath().toAbsolutePath().getParent().toString();
```
tobiasdiez: Just out of interest: what is the difference between `file.toPath().toAbsolutePath()` and `file.getAbsolutePath()`?

simonharrer: `File` is part of the old Java IO API. `Path` is its new replacement in the Java NIO (New IO) API. With file.toPath() or path.toFile() one can convert between these two APIs. The Java NIO has a lot of benefits to offer, which may be the reason why it is used here.

src/main/java/net/sf/jabref/gui/desktop/os/Windows.java
```diff
@@ -3,24 +3,38 @@
...
 
 public class Windows implements NativeDesktop {
+
+    private static String DEFAULT_EXECUTABLE_EXTENSION = ".exe";
+
+
+    @Override
+    public void openFile(String filePath, String fileType) throws IOException {
+        Optional<ExternalFileType> type = ExternalFileTypes.getInstance().getExternalFileTypeByExt(fileType);
+        if (type.isPresent() && !type.get().getOpenWithApplication().isEmpty()) {
```
tobiasdiez: The functionality "if recognized type, then open with special application" should be added to all OS in a consistent way. I would propose to implement a corresponding method in `NativeDesktop`, which calls a new method `openFileDirectly` (or better named). This new method is then overwritten by the specific OS implementations.

tobiasdiez:
LGTM :+1:
Just some small remarks.

src/main/java/net/sf/jabref/gui/desktop/JabRefDesktop.java
Siedlerchr: Yep. The nio.Path was introduced with Java SE 7 and further extended with Java 8 (e,g Stream support). The Old File ist regarded as legacy

tobiasdiez: Thanks for the explanation!

stefan-kolb:
LGTM :+1:

Siedlerchr:
I further refactored the OS classes to remove redundant code.

src/main/java/net/sf/jabref/gui/desktop/JabRefDesktop.java
```diff
@@ -156,6 +158,7 @@ public static boolean openExternalFileAnyFormat(final MetaData metaData, String
     private static void openExternalFilePlatformIndependent(Optional<ExternalFileType> fileType, String filePath)
             throws IOException {
         if (fileType.isPresent()) {
+
```
koppor: Why the empty line?

CHANGELOG.md
koppor: The Changelog entry is NOT needed. JabRef worked fine in 3.2. Please remove the line.

src/main/java/net/sf/jabref/gui/desktop/os/DefaultDesktop.java
```diff
@@ -3,31 +3,37 @@
...
     public void openFolderAndSelectFile(String filePath) throws IOException {
-        File file = new File(filePath);
-        Desktop.getDesktop().open(file.getParentFile());
+        File file = Paths.get(filePath).toAbsolutePath().getParent().toFile();
+        Desktop.getDesktop().open(file);
     }
 
     @Override
-    public void openConsole(String absolutePath) throws IOException {
+    public void openConsoleInDir(String absolutePath) throws IOException {
```
stefan-kolb: I don't like the new method name. For me it is clear that the parameter opens the console inside this directory.

src/main/java/net/sf/jabref/gui/desktop/os/NativeDesktop.java
```diff
@@ -1,22 +1,74 @@
...
+     * @throws IOException
+     */
+    public abstract void openFolderAndSelectFile(String filePath) throws IOException;
+
+    /**
+     * Opens a console or terminal in the provided working directory
+     * @param workingDir The working directory
+     * @throws IOException
+     */
+    public abstract void openConsoleInDir(String workingDir) throws IOException;
```
stefan-kolb: See comment above

src/main/java/net/sf/jabref/gui/desktop/os/OSX.java
```diff
@@ -1,48 +1,47 @@
...
     }
 
     @Override
-    public void openFileWithApplication(String filePath, String application) throws IOException {
-        // Use "-a <application>" if the app is specified, and just "open <filename>" otherwise:
-        String[] cmd = (application != null) && !application.isEmpty() ?
-                new String[] {"/usr/bin/open", "-a", application, filePath} :
-                new String[] {"/usr/bin/open", filePath};
+    protected void openFileWithApplication(String filePath, String application) throws IOException {
+        String[] cmd = {"/usr/bin/open", "-a", application, filePath};
```
stefan-kolb: What happens when application is null?

src/main/java/net/sf/jabref/gui/desktop/os/OSX.java
```diff
@@ -1,48 +1,47 @@
...
+        File file = Paths.get(filePath).getParent().toFile();
+        Desktop.getDesktop().open(file);
     }
 
     @Override
-    public void openConsole(String absolutePath) throws IOException {
-        Runtime runtime = Runtime.getRuntime();
-        runtime.exec("open -a Terminal " + absolutePath, null, new File(absolutePath));
+    public void openConsoleInDir(String absolutePath) throws IOException {
+
```
stefan-kolb: Empty line

src/main/java/net/sf/jabref/gui/desktop/os/Windows.java
```diff
@@ -6,20 +6,20 @@
...
-        if (type.isPresent() && !type.get().getOpenWithApplication().isEmpty()) {
-            openFileWithApplication(filePath, type.get().getOpenWithApplication());
+    public void openFile(String filePath, Optional<ExternalFileType> fileType) throws IOException {
+
+        if (hasRegisteredApplication(fileType)) {
+            openFileWithApplication(filePath, fileType.get().getOpenWithApplication());
         } else {
-            Runtime.getRuntime().exec(new String[] {"cmd.exe", "/c", Paths.get(filePath).toString()});
+            //Use File because Path can't handle URLs
+            Runtime.getRuntime().exec(new String[] {"explorer.exe", new File(filePath).toString()});
```
stefan-kolb: Why explorer here now?

CHANGELOG.md
```diff
@@ -52,7 +52,7 @@ to [sourceforge feature requests](https://sourceforge.net/p/jabref/features/) by
 - [#1023](https://github.com/JabRef/jabref/issues/492) ArXiv fetcher now also fetches based on eprint id
 
 ### Fixed
-- Fixed "Open-folder" action in Windows not opening tge correct folder when folder path contained whitespaces
+- Fixed "Open-folder" action in Windows not opening the correct folder when folder path contained whitespaces
```
koppor: (Repetition of another comment - just to be sure, it doesn't get lost)

Please remove that line. It worked fine at the last released version 3.2. Thus, no CHANGELOG entry required.

src/main/java/net/sf/jabref/gui/desktop/os/Windows.java
Siedlerchr: Because explorer also can open URLs. e.g. explorer.exe "www.jabref.org"

stefan-kolb:
I liked the fix before. I don't like the refactoring so far. For me it would have been a better solution to first merge the fix and later discuss the changes in the logic and the general architecture.

src/main/java/net/sf/jabref/gui/desktop/os/OSX.java
Siedlerchr: application is never null. a) it is only called when it has a registered application and b) getOpenWithApplication returns an empty String if nothing is found

Siedlerchr:
@stefan-kolb  Well. That would be no problem. as I have split it up in 2 commits.
What do the others think?

src/main/java/net/sf/jabref/gui/desktop/os/Windows.java
stefan-kolb: Ok, verified that :+1:

src/main/java/net/sf/jabref/gui/desktop/os/OSX.java
stefan-kolb: :+1:

src/main/java/net/sf/jabref/gui/desktop/os/NativeDesktop.java
```diff
@@ -1,22 +1,71 @@
...
 
 import java.io.IOException;
+import java.util.Optional;
 
-public interface NativeDesktop {
-    void openFile(String filePath, String fileType) throws IOException;
+import net.sf.jabref.external.ExternalFileType;
+import net.sf.jabref.external.ExternalFileTypes;
+
+public abstract class NativeDesktop {
```
simonharrer: Can you use an interface with default methods instead?

src/main/java/net/sf/jabref/gui/desktop/os/DefaultDesktop.java
```diff
@@ -3,31 +3,37 @@
...
+public class DefaultDesktop extends NativeDesktop {
 
-public class DefaultDesktop implements NativeDesktop {
     private static final Log LOGGER = LogFactory.getLog(NativeDesktop.class);
 
+
     @Override
-    public void openFile(String filePath, String fileType) throws IOException {
-        Desktop.getDesktop().open(new File(filePath));
+    public void openFile(String filePath, Optional<ExternalFileType> fileType) throws IOException {
```
simonharrer: Never pass an optional. Create two methods instead, one with and one without a file type.

src/main/java/net/sf/jabref/gui/desktop/os/NativeDesktop.java
Siedlerchr: I  don't really like having an interface with default methods. I chose the abstract class because from my point of view a class hierarchy better represents the is-a hierarchy in this case. e.g. a OSX(Desktop) is a Native Desktop. An Interface is more a kind of Role-based thing and in this case not the best way.

simonharrer: In that case, the concrete method hasRegisteredApplication is never overridden by their sub classes, and could be made static. From my point of view, It should be a method of the class `ExternalFileType` (without the Optional parameter of course). 

The other concrete method is also never overridden and has only two very simple lines of code. It is only required in two places in the JabRefDesktop class which are only a few lines apart in the same method. Hence, I would move this method to the calling side, as it does not really help. 

With this said, I would convert the abstract class back to an interface without default methods which can be handled more easily (favor composition over inheritance).

Siedlerchr: Okay, convinced ;)

Siedlerchr:
I looked a bit further around and saw some more relating things to ExternalFileType which could be improved. However, that would take some more time to fix and be beyond this scope. Will do that in a different PR.
For this PR and the 3.3 I would use the first commit as base with minor modifications (cmd to explorer)

Siedlerchr:
Rebased the changes.

simonharrer:
:+1: LGTM

stefan-kolb:
LGTM :+1:

PR #1095, opened 2016-04-04
Plain text input refactoring

oscargus:
Moved the plain text input to gui and made some changes:
* Required and optional fields are marked with the corresponding background color
* Support for adding keywords

Questions:
* How should we handle the copyright notice here?
* "Work options"? Suggestions?

screenshot-2


oscargus:
This can be merged. The questions remain, but are not really affecting the PR as such.

PR #1097, opened 2016-04-04
Added all person name and date fields to InternalBibtexFields

oscargus:
* Added all person name and date fields to InternalBibtexFields and changed logic which typically detects "author" and "editor" to detect if the field is a person name list field. All date fields now get a calendar button.

* Moved default ImportInspectorDialog fields to ImportInspectorDialog and removed the argument to the constructor since the default was always used anyway.

* Added support for cleaning up dates of the form "uuuu-M-d", e.g. 2016.4.4.


src/main/java/net/sf/jabref/gui/search/SearchResultsDialog.java
```diff
@@ -480,7 +481,7 @@ public Object getColumnValue(BibEntry entry, int column) {
             }
             else {
                 String field = FIELDS[column - PAD];
-                if ("author".equals(field) || "editor".equals(field)) {
+                if (InternalBibtexFields.getFieldExtras(field).contains(BibtexSingleFieldProperties.PERSON_NAMES)) {
```
tobiasdiez: I think it would make sense to add a method `isPersonName(field)` to InternalBibtexFields since this is check occurs quite often.

src/main/java/net/sf/jabref/logic/integrity/IntegrityCheck.java
```diff
@@ -38,8 +40,11 @@ public IntegrityCheck(BibDatabaseContext bibDatabaseContext) {
             return result;
         }
 
-        result.addAll(new AuthorNameChecker("author").check(entry));
-        result.addAll(new AuthorNameChecker("editor").check(entry));
+        for (String fieldName : entry.getFieldNames()) {
+            if (InternalBibtexFields.getFieldExtras(fieldName).contains(BibtexSingleFieldProperties.PERSON_NAMES)) {
```
tobiasdiez: I think a slightly cleaner solution would be to add and use a method `InteralBibtexFields.getPersonFields()`.

src/main/java/net/sf/jabref/model/entry/BibLatexEntryTypes.java
```diff
@@ -871,4 +871,10 @@ public String getName() {
     public static Optional<EntryType> getType(String name) {
         return ALL.stream().filter(e -> e.getName().equalsIgnoreCase(name)).findFirst();
     }
+
+    public static final List<String> BIBLATEX_PERSON_NAME_FIELDS = Arrays.asList("author", "editor", "editora",
```
tobiasdiez: I would move these lists to InternalBibtexFields.

src/main/java/net/sf/jabref/logic/formatter/bibtexfields/NormalizeDateFormatter.java
```diff
@@ -59,7 +60,7 @@ public String getDescription() {
                 "uuuu-M-d", "uuuu-M",
                 "M/uu", "M/uuuu",
                 "MMMM d, uuuu", "MMMM, uuuu",
-                "d.M.uuuu"};
+                "d.M.uuuu", "uuuu.M.d"};
```
tobiasdiez: Mhhh...according to https://en.wikipedia.org/wiki/Date_format_by_country this format is only used by Korea and Mongolia. Do you think it is really worth adding?

src/main/java/net/sf/jabref/gui/search/SearchResultsDialog.java
oscargus: I was considering that, but at the same time there may be quite a few different methods of that type provided that one supports all properties. But I agree that the call is quite long considering that there are long class names etc and that this is probably the most frequent call.

src/main/java/net/sf/jabref/logic/integrity/IntegrityCheck.java
oscargus: That list would typically be longer, so more iterations. Also, if we have e.g. a DateChecker, one can add a simple if-clause below this. Probably one should add the JOURNAL_NAME property to a few more fields and run that checker for those fields.

src/main/java/net/sf/jabref/model/entry/BibLatexEntryTypes.java
oscargus: I was considering that and am open to input here. I can see the point of keeping model clean, but on the other hand, it seems to make sense to keep the information "close to" the source. (In IEEETranEntryTypes the corresponding fields are there.)

src/main/java/net/sf/jabref/logic/formatter/bibtexfields/NormalizeDateFormatter.java
oscargus: Almost all my .bib-files have that format (or rather uuuu.MM.dd, but this seems to cover that as well). Not sure why, but the files are created on both Linux and Windows and some are from my (non-Swedish) PhD students, so for some reason it seems quite common. Default in older versions of JabRef? (I would never select that format myself, so I haven't configured it for sure.)

src/main/java/net/sf/jabref/logic/integrity/IntegrityCheck.java
tobiasdiez: I think performance is no issue here as you iterate over like 10 items, so I would put my priority on readability. On second thought, I would even propose to move the code to AuthorNameChecker and remove its constructor argument.

oscargus: One way to enhance readability would obviously be to have shorter class names. I honestly think that it is the major concern here. `Internal.get(field).contains(NAME)` is pretty much more readable... However, in this case, moving to AuthorNameChecker clearly makes more sense and is probably what I had done with more insight into the Checkers. I'll give it a go.

src/main/java/net/sf/jabref/model/entry/BibLatexEntryTypes.java
Siedlerchr: Are not all kinds of BibTex fields already  somewhere defined  globally as
constants?
e.g. author, editor. Then it would be easy to just use them,. e.g
GlobalFields.AUTHOR or somehow...

And is "editora" a correct field?


2016-04-04 22:46 GMT+02:00 Tobias Diez <notifications@github.com>:

oscargus:
I moved the field arrays to InternalBibtexFields, renamed BibtexSingleFieldProperties to FieldProperties, made AuthorNameChecker check all field with PERSON_NAMES property, and added support for integer (partial) and gender fields. Screen shot for gender field (which is mocked because of #1101).

screenshot-3

src/main/java/net/sf/jabref/model/entry/BibLatexEntryTypes.java
oscargus: No, not all field, only some (configurable?). Yes, editora, editorb, editorc are defined in BibLatex. There are also namea, nameb, ... if one would like but these are not general fields in the standard but are defined by certain styles.

simonharrer:
Please rebase and merge.

src/main/java/net/sf/jabref/bibtex/BibtexSingleFieldProperties.java
```diff
@@ -5,11 +5,11 @@
 public enum BibtexSingleFieldProperties {
```
simonharrer: Is this class still in use?

oscargus: Yes, that is the new class for this. Now renamed to FieldProperties (it might be that the removal was not done, on my office machine it isn't enough to do git add . to include all changes...).

PR #1102, opened 2016-04-05
Fixed freezing when canceling Find unlinked files (for PDFs)

oscargus:
This can be significantly improved, but at least it won't freeze for now. Also, the output do make sense.


stefan-kolb:
LGTM :+1:

PR #1103, opened 2016-04-05
Add benchmarks

tobiasdiez:
This PR adds some basic benchmarks for parsing and writing a bib file. The results are as following for a database consisting of 1000 entries.
Benchmark                         Score     Error  Units
Benchmarks.parse                 49736.582 ± 788.879  ops/s
Benchmarks.write                0.706 ±   0.012  ops/s
Benchmarks.search                258.838 ±   5.604  ops/s
Benchmarks.inferBibDatabaseMode  1297.622 ±  22.910  ops/s

As one can see the parse operation is by many orders of magnitudes quicker then writing. I had a closer look at the write operation and it turned out that 66% of the time is spent in Database.getMode(). Some small changes improved the situation by a factor of 10
Benchmark                         Score      Error  Units
Benchmarks.parse                42031.971 ± 8188.833  ops/s
Benchmarks.write                 8.299 ±    0.304  ops/s
Benchmarks.search               248.093 ±    7.573  ops/s
Benchmarks.inferBibDatabaseMode  20759.711 ±  397.031  ops/s

I suspect the changes in #1100 improve the situation even more (since there the database mode is cached).

(By the way, `gradle jmh` runs the benchmarks. So its pretty simple to use.)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -19,6 +19,7 @@ plugins {
     id "com.github.youribonnaffe.gradle.format" version "1.3"
     id "com.coverity.ondemand" version "1.5.61"
     id "net.ltgt.errorprone" version "0.0.8"
+    id 'me.champeau.gradle.jmh' version '0.3.0'
 }
 
 apply plugin: "java"
@@ -27,6 +28,7 @@ apply plugin: "project-report"
 apply plugin: 'jacoco'
 apply plugin: 'com.github.kt3k.coveralls'
 apply plugin: 'install4j'
+apply plugin: 'me.champeau.gradle.jmh'
 
 apply from: 'eclipse.gradle'
 
--- a/build.gradle
+++ b/build.gradle
@@ -367,5 +367,4 @@ jmh {
     warmupIterations = 5
     iterations = 10
     fork = 2
-    benchmarkMode = 'avgt'
 }
--- a/build.gradle
+++ b/build.gradle
@@ -363,3 +363,9 @@ task getdeps(type: Copy) {
     into 'build/tmp/alldeps/'
 }
 
+jmh {
+    warmupIterations = 5
+    iterations = 10
+    fork = 2
+    benchmarkMode = 'avgt'
+}
--- a/src/jmh/java/net/sf/jabref/benchmarks/Benchmarks.java
+++ b/src/jmh/java/net/sf/jabref/benchmarks/Benchmarks.java
@@ -16,6 +16,8 @@ import net.sf.jabref.importer.ParserResult;
 import net.sf.jabref.importer.fileformat.BibtexParser;
 import net.sf.jabref.logic.search.SearchQuery;
 import net.sf.jabref.model.database.BibDatabase;
+import net.sf.jabref.model.database.BibDatabaseMode;
+import net.sf.jabref.model.database.BibDatabaseModeDetection;
 import net.sf.jabref.model.entry.BibEntry;
 import org.openjdk.jmh.Main;
 import org.openjdk.jmh.annotations.*;
@@ -23,6 +25,7 @@ import org.openjdk.jmh.runner.RunnerException;
 
 @State(Scope.Thread)
 public class Benchmarks {
+
     StringReader bibtexStringReader;
     BibDatabase database = new BibDatabase();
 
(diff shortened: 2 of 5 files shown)


tobiasdiez:
Ready for review.

src/jmh/java/net/sf/jabref/benchmarks/Benchmarks.java
```diff
@@ -0,0 +1,91 @@
...
+    StringReader bibtexStringReader;
+    BibDatabase database = new BibDatabase();
+
+    @Setup
+    public void init() throws IOException, SaveException {
+        Globals.prefs = JabRefPreferences.getInstance();
+
+        Random randomizer = new Random();
+        for(int i = 0; i < 1000; i++)
+        {
```
simonharrer: formatting

I would use more elements, e.g., 10k as this should be more helpful for finding the issues.

src/jmh/java/net/sf/jabref/benchmarks/Benchmarks.java
```diff
@@ -0,0 +1,91 @@
...
+import net.sf.jabref.model.database.BibDatabase;
+import net.sf.jabref.model.database.BibDatabaseMode;
+import net.sf.jabref.model.database.BibDatabaseModeDetection;
+import net.sf.jabref.model.entry.BibEntry;
+import org.openjdk.jmh.Main;
+import org.openjdk.jmh.annotations.*;
+import org.openjdk.jmh.runner.RunnerException;
+
+@State(Scope.Thread)
+public class Benchmarks {
```
simonharrer: can we define any thresholds?

build.gradle
```diff
@@ -361,3 +363,8 @@ task getdeps(type: Copy) {
     into 'build/tmp/alldeps/'
 }
 
+jmh {
+    warmupIterations = 5
+    iterations = 10
+    fork = 2
+}
```
simonharrer: Nice configuration in gradle :)

src/main/java/net/sf/jabref/model/database/BibDatabaseModeDetection.java
```diff
@@ -39,20 +38,15 @@ public static BibDatabaseMode inferMode(BibDatabase database) {
         }
     }
 
-    private static List<String> getEntryTypes(Collection<BibEntry> collection) {
-        return collection.stream().map(BibEntry::getType).collect(Collectors.toList());
+    private static Stream<String> getEntryTypes(Collection<BibEntry> collection) {
```
simonharrer: this method should be inlined, as it is very simple.

src/jmh/java/net/sf/jabref/benchmarks/Benchmarks.java
tobiasdiez: The writing operation takes per iteration over 2 minutes with 10k items. So this would result in > 30 min benchmark time.

src/jmh/java/net/sf/jabref/benchmarks/Benchmarks.java
tobiasdiez: Not that I'm aware of. In the end JMH only runs the benchmarks and writes a result file.

I found plugins for [jenkins](https://github.com/blackboard/jmh-jenkins) and for [teamcity](https://github.com/presidentio/teamcity-plugin-jmh) but apperently there is nothing similar for circleci or travis.

src/jmh/java/net/sf/jabref/benchmarks/Benchmarks.java
simonharrer: Hm, ok. Then leave it as is.

simonharrer:
Really like this. However, we cannot easily benchmark the GUI performance. But we can benchmark the MainTableDataModel from #1100 which does all the heavy lifting regarding sorting, filtering, etc. 

It would be awesome if we could track the progress of these benchmarks, but that would require something like a jenkins. Probably something for the future.

tobiasdiez:
I changed the code accordingly to your comments.

Yes it would be nice to have an overview of the performance for each PR. Since JMH writes the results to a text file in build\reports\jmh, it shouldn't be to hard to get the numbers.

simonharrer:
:+1: LGTM

PR #1105, opened 2016-04-05
Fixes

oscargus:
* Removed Codacy warnings sloppily introduced in TextInputDialog

### DIFF
--- a/src/main/java/net/sf/jabref/gui/plaintextimport/TextInputDialog.java
+++ b/src/main/java/net/sf/jabref/gui/plaintextimport/TextInputDialog.java
@@ -234,8 +234,9 @@ public class TextInputDialog extends JDialog {
 
         try {
             document.insertString(0, "", document.getStyle("regular"));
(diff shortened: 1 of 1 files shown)


tobiasdiez:
Mhh. I'm against reintroducing FormatChars as a class. Reasoning:
- I don't like the name at all and it is not obvious what a `FormatChars` - formatter does.
- Even if we would like to support FormatChars for backwards compatibility, then there are better solutions for that. Just insert a corresponding if statement in `LayoutEntry.getLayoutFormatterByClassName` (and even better, remove the reflection stuff completely).

oscargus:
Ok, as long as it works next Friday anything goes for me...

oscargus:
Regarding naming I conceptually agree but it's been around for 10+ years so... (even before Unicode was well established) Legacy etc...

oscargus:
I removed the FormatChars part and put it in #1114 so that there is a quick fix available before the release...

PR #1112, opened 2016-04-05
Move L&F configuration from advanced to appearance tab in preferences

stefan-kolb:
- [ ] Why don't we display LAF selection on OSX?

__Appearance__

image

__Advanced__

image


simonharrer:
@chochreiner any idea why the L&F selection is hidden on Mac OS X?

simonharrer:
LGTM :+1:

chochreiner:
I've only had a quick look at it, but i guess it is because only the default L&F shows the the OSX specific first dropdownmenu

at a first glance, there is no reason why this menu should be disabled - apart from the fact that IMHO only the default L&F looks good

stefan-kolb:
Seems reasonable to me :+1: So we keep it this way.

PR #1114, opened 2016-04-06
Added FormatChars LayoutFormatter back again (to be merged before 3.3 unless someone else solves it in a proper way)

oscargus:
Just to make sure that we do not release 3.3 with non-working export formats...

Also moves LatexToUnicodeFormatter to its proper place.

### DIFF
--- a/src/main/java/net/sf/jabref/gui/maintable/MainTableColumn.java
+++ b/src/main/java/net/sf/jabref/gui/maintable/MainTableColumn.java
@@ -2,8 +2,7 @@ package net.sf.jabref.gui.maintable;
 
 import net.sf.jabref.bibtex.BibtexSingleFieldProperties;
 import net.sf.jabref.bibtex.InternalBibtexFields;
-import net.sf.jabref.logic.layout.LayoutFormatter;
-import net.sf.jabref.logic.layout.format.LatexToUnicodeFormatter;
+import net.sf.jabref.logic.formatter.bibtexfields.LatexToUnicodeFormatter;
 import net.sf.jabref.model.database.BibDatabase;
 import net.sf.jabref.model.entry.BibEntry;
 import net.sf.jabref.model.entry.EntryUtil;
@@ -22,7 +21,7 @@ public class MainTableColumn {
 
     private final Optional<BibDatabase> database;
 
-    private final LayoutFormatter toUnicode = new LatexToUnicodeFormatter();
+    private final LatexToUnicodeFormatter toUnicode = new LatexToUnicodeFormatter();
 
     public MainTableColumn(String columnName) {
         this.columnName = columnName;
--- a/src/main/java/net/sf/jabref/logic/layout/format/FormatChars.java
+++ b/src/main/java/net/sf/jabref/logic/layout/format/FormatChars.java
@@ -1,0 +1,34 @@
+/*  Copyright (C) 2003-2016 JabRef contributors.
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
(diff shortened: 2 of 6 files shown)


src/main/java/net/sf/jabref/gui/maintable/MainTableColumn.java
```diff
@@ -22,7 +21,7 @@
 
     private final Optional<BibDatabase> database;
 
-    private final LayoutFormatter toUnicode = new LatexToUnicodeFormatter();
+    private final LatexToUnicodeFormatter toUnicode = new LatexToUnicodeFormatter();
```
lenhard: Is there a particular reason for using the exact type in the variable declaration? Why not just `Formatter`?

oscargus: Not really. I had some issue with it and decided to go down the easiest path... (It doesn't implement LayoutFormatter anymore, do not know what I tried as an intermediate solution.)

oscargus: Maybe I was simply confused which caused the problem solved by the exact type.

lenhard: The problem was a name clash with `java.util.Formatter` which could be solved by importing `net.sf.jabref.logic.formatter.Formatter` explicitly. I've done that and will merge the PR when the checks are completed.

lenhard:
Oh my, I can't commit into your branch and my push resulted in a new branch in the JabRef repo... Can you just apply this 2c8b6e428d7b023bd7657111893e223447f128f2 here and afterwards we can merge this PR and delete my duplicate branch.

oscargus:
I'm back home without computer but will sort it out tomorrow.

tobiasdiez:
Again, why do you think it is better to reintroduce the FormatChars formatter instead of editing the provided export layouts in resource.layout.iso690txt and .ris ?
In my opinion 
`\begin{author}\format[LatexToUnicode,Iso690NamesAuthors]{\author}`
more understandable then
`\begin{author}\format[FormatChars,Iso690NamesAuthors]{\author}`

oscargus:
Feel free to edit it. I'm not saying that it is better, but that it was a
quicker fix (for something I didn't break...) than editing those files,
updating the help files and handling issues posted by people having their
own export formats broken.

tobiasdiez:
@JabRef/developers what you do think about it? Move forward and use LatexToUnicode as the name or be backwards compatible?

src/main/java/net/sf/jabref/logic/layout/format/FormatChars.java
```diff
@@ -0,0 +1,34 @@
...
+import net.sf.jabref.logic.formatter.bibtexfields.LatexToUnicodeFormatter;
+import net.sf.jabref.logic.layout.LayoutFormatter;
+
+/**
+ * Formatter that outputs a sequence number for the current entry. The sequence number is
+ * tied to the entry's position in the order, not to the number of calls to this formatter.
+ */
+public class FormatChars implements LayoutFormatter {
+
+    private static final Formatter formatter = new LatexToUnicodeFormatter();
```
koppor: Here, I would use `LatexToUnicodeFormatter`, because this class really depends on that formatter.

koppor:
We have to be backwards compatible here. Users built `layout` files which depend on these names.

In the next step, we have to offer migrations for `layout` files. (Reference is IMHO http://help.jabref.org/en/CustomExports/).

lenhard:
I close this issue in favor of the alternative solution implemented in #1144

Issue #638, opened 2016-01-04
JabRef GUI blank after Ubuntu wakes up from sleep


antonkarl:
When my monitor goes out and Ubuntu goes to sleep and I log in again, the user interface looks like that, all grey. It does not recover by minimizing and maximizing but individual GUI elements become visible by using the mouse on them.

![jabrefbug](https://cloud.githubusercontent.com/assets/288033/12095041/1858b8a6-b304-11e5-9b20-d8f509be7ea1.png)

alefa:
I have the exact same problem in Debian 8 (Jessie) with Gnome. The JabRef version in Debian is an older one (2.10), though. I'll test with the newest version when I have a bit of time.

teertinker:
I can confirm this bug. For 3.1, Ubuntu 14.04. 64 bit 

    On Monday, January 4, 2016 7:58 PM, Gita Benadi <notifications@github.com> wrote:
 

 I have the exact same problem in Debian 8 (Jessie) with Gnome. The JabRef version in Debian is an older one (2.10), though. I'll test with the newest version when I have a bit of time.—

oscargus:
What JDK are you running? (And possibly look-and-feel?)

(I've never noticed this as I recall on:
`Linux urkua 2.6.32-573.3.1.el6.x86_64 #1 SMP Thu Aug 13 22:55:16 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux` (CentOS)

```
java version "1.8.0_65"
Java(TM) SE Runtime Environment (build 1.8.0_65-b17)
Java HotSpot(TM) 64-Bit Server VM (build 25.65-b01, mixed mode)
```
but will check it explicitly once back to the office. No idea how to solve it though...
)

alefa:
My Java version:

    openjdk version "1.8.0_66-internal"
    OpenJDK Runtime Environment (build 1.8.0_66-internal-b17)
    OpenJDK 64-Bit Server VM (build 25.66-b17, mixed mode)

I'm using the GTK look-and-feel of JabRef.

oscargus:
I can just confirm that I do not have this problem on my Linux machine. While there are quite a few potential reasons for the problem (windows manager etc), it would be interesting to know if OpenJDK is a common factor or not.

koppor:
Refs #184 and #425. Maybe this is really a duplicate of #425?

antonkarl:
Java version:

java version "9-ea"
[...]

PR #1116, opened 2016-04-06
Open Jdk problems with Look and Feels

stefan-kolb:
See #393 #638.

Still not 100% sure if this only applies to OpenJDK and GTK L&F or also to Oracle JDK and GTK L&F.
Metal L&F seems to work but looks ugly.

Idea is to try Metal on OpenJDK and wait if more bug reports pop up even for Oracle.

- [x] Maybe we should close at least one of the two bug reports that seem to be related to OpenJDK and GTK #393 #638 Or maybe even open a new one with possible solutions.

Also see https://bugs.launchpad.net/ubuntu/+source/openjdk-8/+bug/1510009

__Especially see__
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=813143

Maybe we can just ignore this and wait for it to be fixed in OpenJDK and add a hint to our FAQ and documentation?


stefan-kolb:
@JabRef/developers I think one open issue is enough for reference. Should we close #393 or #638 ?

lenhard:
Close the newer one, since it is the duplicate of the older one. (Don't forget to mark it as such)

stefan-kolb:
When merging this in please use squash merge to make it easy to revert this change when this is obsolete for OpenJDK later on.

simonharrer:
:+1:

koppor:
Oh, where did 4aa0d5b go? As far as I can see, only ac0bae5 (which is very short in comparison to the changes shown above) has been cherry-picked?

mlep:
#1145 : related?

Issue #783, opened 2016-02-09
installer fails without admin permissions


donovaly:
The new Win installer of JabRef 3.x does not check if the user that executes the installer has admin privileges or not.

It should be:
- if the user has admin permissions, the installer should propose to install for all users of the PC (with an option to install only for the current user)
- if the user has no admin permissions, the installer must install only for the current user

Currently the installer tries to install for all users, also if the user doesn't have admin permissions - which will of course fail. (As an expert one can already change the install path but the vast majority doesn't not the correct install location for a non-admin user.)

matthiasgeiger:
Which Windows Version are you using? 

At least for my Win10 System the 64bit installer asks for Admin rights using the usual UAC dialog.

donovaly:
> Which Windows Version are you using? 

Win 7 64bit

> At least for my Win10 System the 64bit installer asks for Admin rights using the usual UAC dialog.

Here not.
However, there is no reason to force Admin permissions if the user don't have them. It was possible to install JabRef 2.x as a non-admin user without problems. I bundle JabRef with the win installer of LyX (and once wrote the JabRef 2.x installer). If JabRef disallows to install as normal user, I can no longer bundle because installing as non-admin is a very often requested feature that the LyX installer provides. (especially in companies and universities one doesn't have admin permissions it should be possible to use LyX and JabRef there too)
From the technical position forcing admin permissions is only necessary to install a program for all users of a PC and/or if one needs to install Windows services or drivers. And JabRef doesn't install services nor drivers.

donovaly:
>>   At least for my Win10 System the 64bit installer asks for Admin rights using the usual UAC dialog.

> Here not.

This is also true for Win 10. Just run the JabRef installer as non-admin user and you will see this.

matthiasgeiger:
Okay - thanks for the clarification. I can see your point now. 
Using the install4j installer it is generally possible to install JabRef 3.X - but the default install directory is "wrong" as it points to the "Program Files" dir - where a normal user has no write priviledges.

So we should configure the new installer, that it is able to detect whether a user has admin rights - or not - and set the default install directory accordingly.

@koppor Can you do this?

koppor:
Maybe we need the admin rights to create file associations. I am unsure about that.

Otherwise, we [could simply create a launcher](http://stackoverflow.com/a/11261259/873282), would that help?

donovaly:
> Maybe we need the admin rights to create file associations.

[...]

PR #1125, opened 2016-04-06
Installer allows to select installation for all users or current user

stefan-kolb:
See #783 
All users needs admin privileges. Current user cannot install into Program file directory.
Installation directory will be set to `user.home` when installation only for current user is selected.
Please, check the installer on http://builds.jabref.org/install-privileges/ if its working correctly.

### DIFF
--- a/jabref.install4j
+++ b/jabref.install4j
@@ -180,7 +180,7 @@
             <preActivation />
             <postActivation />
             <actions>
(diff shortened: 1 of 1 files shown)


simonharrer:
- [ ] new screen is not automatically translated, while all other pages are translated

PR #1126, opened 2016-04-06
Made change to alleviate a bug in a multiuser environment with one ...

ScottTownsendNIST:
…centralized database file that near simultaneous database saves may cause changes from one save to be overwritten by changes from the other. This would occur when the second user save happened while the first user save was locking the database file or saving to the tmp file but prior to the copy of the tmp file to the database file. In this case the second user save would not yet see a modified timestamp or size change in the database file and thus start the save process. The save process would wait for the file lock to be removed then overwrite the database causing the changes from the previous save to be lost. This was especially noticeable when the database file resided on a network drive with frequent network latency issues.


src/main/java/net/sf/jabref/exporter/SaveDatabaseAction.java
```diff
@@ -1,4 +1,6 @@
 /*  Copyright (C) 2003-2015 JabRef contributors.
+    Copyright (C) 2016 Scott Townsend
```
tobiasdiez: Would you mind removing your name here?

src/main/java/net/sf/jabref/exporter/SaveDatabaseAction.java
```diff
@@ -167,6 +169,66 @@ public void run() {
...
+                                        cancelled = true;
+                                    }
+                                });
+                            }
+                        });
+
+                        return;
+                    } else { // User indicated to store anyway.
+                                 // See if the database has the protected flag set:
+                        List<String> pd = panel.getBibDatabaseContext().getMetaData()
```
tobiasdiez: There is a method `isProtected` (or something like that) in `MetaData` which is (now) the preferred way to receive this information.

src/main/java/net/sf/jabref/exporter/SaveDatabaseAction.java
```diff
@@ -167,6 +169,66 @@ public void run() {
             panel.autoGenerateKeysBeforeSaving();
 
             if (FileBasedLock.waitForFileLock(panel.getBibDatabaseContext().getDatabaseFile(), 10)) {
+                // Check for external modifications to alleviate multiuser concurrency issue when near
+                // simultaneous saves occur to a shared database file:
+                if (panel.isUpdatedExternally()
```
tobiasdiez: Could you please split your addition in multiple, smaller methods so that the code is easier to understand. Thanks.

tobiasdiez:
Thank you very much for your contribution :+1: .
I just made some comments. Please fix these small issues and rebase, then we will merge this in.

ScottTownsendNIST:
Sorry to take so long to address the comments and fix the code. I had other projects that needed attention. Thank you for your consideration of my code change.

tobiasdiez:
Thanks for your revision. 
LGTM :+1:

Issue #1134, opened 2016-04-07
Unable to import *.ris file


AlexanderZ11234:
JabRef version "JabRef_windows-x64_3_3dev--snapshot--2016-04-06--master--6488cf7" on Windows 10

I cannot import *.ris files with the message in the lower left corner: "Status: Could not find a suitable import format".

But I remember that version 3.2 is free of this problem.

oscargus:
Try the latest master available in 15-20 minutes or so. Sorry about that.

purplewakanda:
The same issue exists in the current version

PR #1138, opened 2016-04-07
ImportFormatReader test

oscargus:
Added initial tests for ImportFormatReader. Now, seven different formats are imported without specifying the format and the number of imported entries is checked. Should avoid bugs like #1134.

I also removed unused methods and moved a totally unrelated method to a better home.

### DIFF
--- a/src/test/java/net/sf/jabref/importer/ImportFormatReaderTest.java
+++ b/src/test/java/net/sf/jabref/importer/ImportFormatReaderTest.java
@@ -2,19 +2,55 @@ package net.sf.jabref.importer;
 
 import static org.junit.Assert.*;
 
+import java.util.ArrayList;
+import java.util.Collection;
+
+import org.junit.Before;
 import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import net.sf.jabref.Globals;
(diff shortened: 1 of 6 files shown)


src/test/java/net/sf/jabref/importer/ImportFormatReaderTest.java
```diff
@@ -0,0 +1,67 @@
...
+        this.resourceName = resource;
+        this.format = format;
+        this.count = count;
+    }
+
+    @Before
+    public void setUp() {
+        Globals.prefs = JabRefPreferences.getInstance();
+        reader = new ImportFormatReader();
+        reader.resetImportFormats();
```
tobiasdiez: Is the resetImportFormats really necessary? I would have expected that the constructor takes care of initializing a correct importformatreader.

tobiasdiez:
LGTM :+1: Nice work.

src/test/java/net/sf/jabref/importer/ImportFormatReaderTest.java
oscargus: Yes. There is no explicit constructor. Maybe has something to do with the
preferences not being initialized? resetImportFormat() is called quite
early in JabRef.start

PR #478, opened 2015-12-07
delete JstorImporter, edit Changelog

ayanai1:
JSTOR is exporting BibTeX and has no own format anymore.

PR #1140, opened 2016-04-07
Remove unused JSTOR code

zellerdev:
Remove unused JSTOR code, because it's unlikely that JSTOR comes back.  Refs #478 , https://github.com/koppor/jabref/issues/66


koppor:
LGTM

simonharrer:
Jstor as a database still exists, I think. Why should we want to delete the fetcher for that? Which Format does the fetcher return?

koppor:
Please see the referenced pull request: https://github.com/JabRef/jabref/pull/1140. 

> JSTOR is exporting BibTeX and has no own format anymore.

Should we check the statement at #1140 again?

mlep:
What is meant by "Remove unused JSTOR code"?
Is it
"remove the code specific to the old JSTOR format (because there is no need for it since JSTOR directly generates BibTeX format), but keep JSTOR as a fetcher"
or
"remove the code specific to the old JSTOR format and the JSTOR fetcher"
?

simonharrer:
To clarify things:

- JSTOR format is obsolete, hence, the JSTOR format importer was removed in a PR some time ago
- JSTOR fetcher fetchting entries from the JSTOR database still makes sense, as we can just use their bibtex export.
- BUT: we currently have two broken JSTOR fetchers

Hence, I second this removal. Hopefully, we can add a correctly working JSTOR fetcher in the future.

koppor:
The JSTOR fetcher is [removed since version 3.0](https://github.com/JabRef/jabref/blob/master/CHANGELOG.md#removed-3), because [it was not working](https://github.com/koppor/jabref/issues/66).

mlep:
Thank you for the clarification!

PR #1142, opened 2016-04-07
Removed unused imports and some other minor things

oscargus:
Oh, how the Codacy issue count will drop...


Siedlerchr:
Hm seems like Gradle encountered probs at CI server:
`./gradlew test died unexpectedly`

oscargus:
Yes, a few recent PRs have run into this. Would guess at something
temporary at their servers.

src/test/java/net/sf/jabref/logic/l10n/LocalizationConsistencyTest.java
```diff
@@ -56,6 +54,7 @@ public DuplicationDetectionProperties(Properties defaults) {
         /**
          * Overriding the HashTable put() so we can check for duplicates
          */
+        @Override
         public synchronized Object put(Object key, Object value) {
             // Have we seen this key before?
             if (get(key) != null) {
```
Siedlerchr: Please use containsKey(Key), because a HashMap allows to have a Null Key and this is therefore not a valid indicator.  I just read about that today.
There might be also other pieces in the code that use the same logic.

oscargus:
Thanks @Siedlerchr . I changed it based on your suggestion. We should try to look for similar code (seems like there is room for a PMD rule here almost).

PR #1144, opened 2016-04-07
Readd FormatChars support

tobiasdiez:
Alternative to  #1114 to add backwards compatibility for FormatChars (and a few other renamed LayoutFormatters). The idea was to replace the reflection code by an (admittedly big) switch statement. In this way we gain some flexibility about the location and naming of LayoutFormatter classes.

By the way,  according to http://jabref.sourceforge.net/help/CustomExports.php all LayoutFormatters starting with "Author" are deprecated and are replaced by "Authors" since 2.7. Should we just delete them now?

### DIFF
--- a/src/main/java/net/sf/jabref/logic/layout/LayoutEntry.java
+++ b/src/main/java/net/sf/jabref/logic/layout/LayoutEntry.java
@@ -78,9 +78,7 @@ class LayoutEntry {
 
     public LayoutEntry(List<StringInt> parsedEntries, int layoutType, JournalAbbreviationRepository repository) {
         this.repository = repository;
-        List<StringInt> blockEntries = null;
         List<LayoutEntry> tmpEntries = new ArrayList<>();
-        LayoutEntry le;
         String blockStart = parsedEntries.get(0).s;
         String blockEnd = parsedEntries.get(parsedEntries.size() - 1).s;
 
@@ -90,6 +88,7 @@ class LayoutEntry {
 
         type = layoutType;
         text = blockEnd;
+        List<StringInt> blockEntries = null;
         for (StringInt parsedEntry : parsedEntries.subList(1, parsedEntries.size() - 1)) {
             switch (parsedEntry.i) {
             case LayoutHelper.IS_FIELD_START:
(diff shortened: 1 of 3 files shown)


lenhard:
I am in favour of this solution, because it breaks the link between our class structure and hard-coded strings in the layout files of random people. The switch may be long, but it is very easy to read. 

As you suggest, we should delete the deprecated "Author" formatters.

Let's decide on this at the upcoming devcall!

oscargus:
Just make sure that http://help.jabref.org/en/CustomExports/ is updated accordingly and note:

"Custom formatters

If none of the available formatters can do what you want to achieve, you can add your own by implementing the net.sf.jabref.export.layout.LayoutFormatter interface. If you insert your class into the net.sf.jabref.export.layout.format package, you can call the formatter by its class name only, like with the standard formatters. Otherwise, you must call the formatter by its fully qualified name (including package name). In any case, the formatter must be in your classpath when running JabRef."

(I do not have any opinion on this, just want to make sure that things are not obviously broken.)

src/main/java/net/sf/jabref/logic/layout/LayoutEntry.java
```diff
@@ -358,40 +358,152 @@ private void doOptionField(String s) {
...
+        case "IfPlural":
+            return new IfPlural();
+        case "Replace":
+            return new Replace();
+        case "WrapContent":
+            return new WrapContent();
+        case "WrapFileLinks":
+            return new WrapFileLinks();
+        default:
+            return new NotFoundFormatter(name);
```
simonharrer: One could create a Map<String, Formatter> which contains the mapping.

src/main/java/net/sf/jabref/logic/layout/LayoutEntry.java
```diff
@@ -358,40 +358,152 @@ private void doOptionField(String s) {
...
-    // added section - end (arudert)
 
-    private static LayoutFormatter getLayoutFormatterByClassName(String className,
-            JournalAbbreviationRepository repostiory)
-            throws Exception {
+    private LayoutFormatter getLayoutFormatterByName(String name) throws Exception {
 
-        if (className.isEmpty()) {
+        if (name.isEmpty()) {
             return null;
```
simonharrer: I would return a NotFoundFormatter here as well, avoiding this evil null value :)

src/main/java/net/sf/jabref/logic/layout/LayoutEntry.java
tobiasdiez: At first, I thought so, too. However a map would require that all formatters are initialized although only one of them is returned (and a static map is not possible since some of the formatters have constructor arguments).

src/main/java/net/sf/jabref/logic/layout/LayoutEntry.java
tobiasdiez: Done.

src/main/java/net/sf/jabref/logic/layout/LayoutEntry.java
simonharrer: ok

simonharrer:
LGTM :+1:

koppor:
@tobiasdiez Could you open a new issue for your "By the way" thing and post the excerpt from the help page? I couldn't quickly find the paragraph you're referring to.

PR #1146, opened 2016-04-08
Replaced Map.get()  with containsKey() where appropriate

oscargus:
Based on a disussion in #1142, pointed out by @Siedlerchr , checking if a `Map` containts a key is better done using `Map.containsKey()` than checking if `Map.get()` returns null. (I have checked that null is not stored in the `Map` as well.)


simonharrer:
LGTM

PR #1147, opened 2016-04-08
Added test for StringUtil.expandAuthorInitials 

oscargus:
Added support for names with - and A.A.

(Could be parameterized but then a new file is required etc... and didn't have time for that right now...)


simonharrer:
:+1:

PR #1148, opened 2016-04-08
Added support for specific Biblatex fields

oscargus:
Support for editor type, pagination, and type fields.


simonharrer:
Would postpone it after Release because of the many new language Keys.

oscargus:
Makes sense.  I added the 3.4 tag and will rebase and merge once 3.3 is released.

PR #1149, opened 2016-04-08
Replace NormalizeNamesFormatter by AuthorList and so fix #318

### LINKED ISSUE #318: Normalize to Bibtex name format

tobiasdiez:
- Fixes #318 (and refs #950)
- Removes comma as delimiter of author names (in NormalizenamesFormatter).
 The problem is that comma separated authors cannot be distingued from one author in BibTeX format.
 For example `Name, della, first` could be three names separated by commas or if interpreted as valid BibTeX string it is one author with lastname=name, jrpart=della, firstname=first.
- Add colon as delimiter of author names (in AuthorListParser).

Thanks at all, but especially @oscargus, for creating many tests. Made implementing this PR way easier! 

--

### DIFF
--- a/src/main/java/net/sf/jabref/model/entry/Author.java
+++ b/src/main/java/net/sf/jabref/model/entry/Author.java
@@ -38,18 +38,98 @@ public class Author {
      * @param von       the von part of the author's name (may consist of several
      *                  tokens, like "de la" in "Charles Louis Xavier Joseph de la
      *                  Vall{\'e}e Poussin")
-     * @param last      the lats name of the author (may consist of several
+     * @param last      the last name of the author (may consist of several
      *                  tokens, like "Vall{\'e}e Poussin" in "Charles Louis Xavier
      *                  Joseph de la Vall{\'e}e Poussin")
      * @param jr        the junior part of the author's name (may consist of
      *                  several tokens, like "Jr. III" in "Smith, Jr. III, John")
      */
     public Author(String first, String firstabbr, String von, String last, String jr) {
-        firstPart = removeStartAndEndBraces(first);
+        firstPart = addDotIfAbbreviation(removeStartAndEndBraces(first));
         firstAbbr = removeStartAndEndBraces(firstabbr);
         vonPart = removeStartAndEndBraces(von);
         lastPart = removeStartAndEndBraces(last);
         jrPart = removeStartAndEndBraces(jr);
+    }
+
+    public static String addDotIfAbbreviation(String name) {
+        // Avoid arrayindexoutof.... :
+        if (name == null || name.isEmpty()) {
+            return name;
+        }
+        // If only one character (uppercase letter), add a dot and return immediately:
+        if ((name.length() == 1) && Character.isLetter(name.charAt(0)) &&
+                Character.isUpperCase(name.charAt(0))) {
+            return name + ".";
+        }
+
+        StringBuilder sb = new StringBuilder();
+        char lastChar = name.charAt(0);
+        for (int i = 0; i < name.length(); i++) {
+            if(i > 0) {
+                lastChar = name.charAt(i - 1);
+            }
+            char currentChar = name.charAt(i);
+            sb.append(currentChar);
+
+            if(currentChar == '.') {
+                // A.A. -> A. A.
+                if(i + 1 < name.length() && Character.isUpperCase(name.charAt(i + 1))) {
+                    sb.append(' ');
+                }
+            }
+
+            boolean currentIsUppercaseLetter = Character.isLetter(currentChar) && Character.isUpperCase(currentChar);
+            if(!currentIsUppercaseLetter) {
+                // No uppercase letter, hence nothing to do
+                continue;
+            }
+
+            boolean lastIsLowercaseLetter = Character.isLetter(lastChar) && Character.isLowerCase(lastChar);
+            if(lastIsLowercaseLetter) {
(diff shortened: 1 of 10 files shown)


src/main/java/net/sf/jabref/model/entry/Author.java
```diff
@@ -38,20 +38,100 @@
...
     public Author(String first, String firstabbr, String von, String last, String jr) {
-        firstPart = removeStartAndEndBraces(first);
+        firstPart = addDotIfAbbreviation(removeStartAndEndBraces(first));
         firstAbbr = removeStartAndEndBraces(firstabbr);
         vonPart = removeStartAndEndBraces(von);
         lastPart = removeStartAndEndBraces(last);
         jrPart = removeStartAndEndBraces(jr);
     }
 
+    public static String addDotIfAbbreviation(String name) {
```
oscargus: How does this method compare to expandAll (?) in StringUtil? Could that be replaced by this (later) as this seems more competent?

oscargus: Ahh, I know the answer for that... Somehow I didn't see StringUtil in the diff the first time. :-)

src/test/java/net/sf/jabref/logic/util/strings/StringUtilTest.java
```diff
@@ -313,6 +312,9 @@ public void testExpandAuthorInitialsDoNotAddDot() {
         assertEquals("Moore, O.", StringUtil.expandAuthorInitials("Moore, O."));
         assertEquals("Moore, A. O.", StringUtil.expandAuthorInitials("Moore, A. O."));
         assertEquals("Moore, A.-O.", StringUtil.expandAuthorInitials("Moore, A.-O."));
+        assertEquals("MEmre", StringUtil.expandAuthorInitials("MEmre"));
+        assertEquals("{\\'{E}}douard", StringUtil.expandAuthorInitials("{\\'{E}}douard"));
```
oscargus: Nice additional tests! Will "{\\'{E}}" work? As in a dot is added.

src/test/resources/net/sf/jabref/importer/fileformat/OvidImporterTestBib6.bib
```diff
@@ -1,6 +1,6 @@
 @Misc{,
   Title                    = {Correct positioning of percutaneous iliosacral screws with computer-navigated versus fluoroscopically guided surgery in traumatic pelvic ring fractures},
-  Author                   = {Verbeek, Jan MD and Hermans, Erik MD and van Vugt, Arie MD, PhD and Frolke, Jan Paul MD, PhD},
```
oscargus: Hmm, I'm wondering if it is possible to skip this expansion... Probably we will have to decide on either hardcoding MD (medical doctor) and disappoint people with Marie David as middle names or just live with it.

oscargus:
Looks very nice! :+1:

The MD thing is a bit challenging and either way we will do it wrong for one of the cases. If I were in medicine I would probably prefer not to expand MD (especially).

src/test/java/net/sf/jabref/logic/util/strings/StringUtilTest.java
tobiasdiez: No it will not. The dot will be added only if the uppercase letter is followed by a whitespace, dash or tilde. I think it would be to hard to detect all the latex stuff.

src/test/resources/net/sf/jabref/importer/fileformat/OvidImporterTestBib6.bib
tobiasdiez: Yeah... you are right it is a bit annoying. But I think the majority of author names does not contain the degree (MD, PHD, ...) of the author or set it up properly as the Jr part.

src/test/java/net/sf/jabref/logic/util/strings/StringUtilTest.java
oscargus: Makes sense. It might be possible to run RemoveLatexFormatter  (or whatever it is called) and do the detection on the stripped string but probably not worth it for such a corner case.

oscargus: It might (eventually) make sense to replace expandAuthorInitials with the name formatter. Even though they doesn't do exactly the same thing, if I recall correctly expandAuthorInitials is used in two importers only and it should be OK to possible change the complete formatting of those author fields.

src/test/resources/net/sf/jabref/importer/fileformat/OvidImporterTestBib6.bib
oscargus: Not obvious why they need to brag about their titles in the first place. :-) (And I know that some of you might end up to write Herr Prof. Dr. Ing. Hab. Prof. Dr.)

simonharrer:
LGTM

PR #1159, opened 2016-04-10
Cleanup formatter tests

koppor:
This is a minor cleanup to the formatter tests. Tests covered in FormatterTests were removed from the concrete formatter classes. A comment pointing to the general tests has been added to all concrete test classes.


src/test/java/net/sf/jabref/logic/formatter/CaseChangersTest.java
```diff
@@ -3,19 +3,20 @@
...
 
+/**
+ * Tests in addition to the general tests from {@link net.sf.jabref.logic.formatter.FormatterTest}
+ */
 public class CaseChangersTest {
 
     @Test
     public void testChangeCaseLower() {
-        Assert.assertEquals("", CaseChangers.TO_LOWER_CASE.format(""));
         Assert.assertEquals("lower", CaseChangers.TO_LOWER_CASE.format("LOWER"));
```
tobiasdiez: Since this is a cleanup PR, could you please also move these test cases to separate test files for the respective CaseChanger (e.g. LowerCaseFormatterTests)

src/test/java/net/sf/jabref/logic/formatter/OrdinalsToSuperscriptFormatterTest.java
```diff
@@ -2,29 +2,15 @@
...
 import org.junit.Test;
 
+/**
+ * Tests in addition to the general tests from {@link net.sf.jabref.logic.formatter.FormatterTest}
+ */
 public class OrdinalsToSuperscriptFormatterTest {
-    private OrdinalsToSuperscriptFormatter formatter;
 
-    @Before
-    public void setUp() {
```
tobiasdiez: Using the setup method is the preferred way to init objects under test (since then every test method gets a new object and possible modifications from a previous test are erased). This is probably not a big deal for the formatters but I wouldn't break the pattern here.

src/test/java/net/sf/jabref/logic/formatter/bibtexfields/ClearFormatterTest.java
```diff
@@ -1,18 +1,8 @@
...
+ */
 public class ClearFormatterTest {
 
-    @Test
-    public void formatReturnsEmptyForEmptyString() throws Exception {
-        assertEquals("", new ClearFormatter().format(""));
-    }
-
-    @Test
-    public void formatReturnsEmptyForSomeString() throws Exception {
```
tobiasdiez: This is an actual test :smile_cat:

src/test/java/net/sf/jabref/logic/layout/format/LatexToUnicodeFormatterTest.java
```diff
@@ -34,25 +34,27 @@
...
     @Test
     public void testFormatStripLatexCommands() {
-        assertEquals("-", new LatexToUnicodeFormatter().format("\\mbox{-}"));
+        assertEquals("-", formatter.format("\\mbox{-}"));
     }
 
     @Test
     public void testEquations() {
-        LayoutFormatter layout = new LatexToUnicodeFormatter();
+        LayoutFormatter layout = formatter;
```
tobiasdiez: Replace layout by formatter?

tobiasdiez:
LGTM :+1: just a few remarks.
For the formatters a better support for parametrized tests like [JUnitParams](https://github.com/Pragmatists/JUnitParams) would be nice.

src/test/java/net/sf/jabref/logic/formatter/OrdinalsToSuperscriptFormatterTest.java
koppor: Quoting [Junit's JavaDoc](http://junit.sourceforge.net/javadoc/org/junit/Test.html) (referenced from http://stackoverflow.com/a/16714168/873282).

> To run the method, JUnit first constructs a fresh instance of the class then invokes the annotated method

Thus, the effects should be the same and my proposal the one with less LOC for the same effects.

src/test/java/net/sf/jabref/logic/formatter/bibtexfields/ClearFormatterTest.java
koppor: Fixed (via changing the commit "Formatters tests: Remove methods already covered by FormatterTest" from 9de60b5b8cf04 into 1262c2551e266.

src/test/java/net/sf/jabref/logic/layout/format/LatexToUnicodeFormatterTest.java
koppor: Fixed by changing the commit "Unify formatter tests" from 4f8fe59423896e1645 to 529bdab7aa900563

src/test/java/net/sf/jabref/logic/formatter/OrdinalsToSuperscriptFormatterTest.java
koppor: See also commit note at "Unify formatter tests" (529bdab7aa900563)

koppor:
I want to get this into v3.3 as i have another commit based on that, which adds `getExampleInput()` to Formatters to improve  #1050

src/test/java/net/sf/jabref/logic/formatter/CaseChangersTest.java
koppor: Done in 8e0ce84

simonharrer:
LGTM as it is.

src/test/java/net/sf/jabref/logic/formatter/OrdinalsToSuperscriptFormatterTest.java
simonharrer: I am OK with the changes.

tobiasdiez: Yes, for each test method a new instance of the test class is used. However, all the different instances are created before any test is run. So if the constructor of the formatter modifies some global static objects, then these changes might get lost.
Some discussion about this can be found here http://stackoverflow.com/questions/512184/best-practice-initialize-junit-class-fields-in-setup-or-at-declaration.
But in the end it is just a minor conventional thing...

koppor:
I rebased onto master

PR #1160, opened 2016-04-10
Remove %s in formatter descriptions

koppor:
In the developer telco, we discussed that we don't want to have `%s` in the descriptions of formatters. This PR removes the last two remaining `%s`.


tobiasdiez:
LGTM :+1:
Sorry for having missed them.

PR #1161, opened 2016-04-10
Add method getExampleInput() to Format class

koppor:
I added `getExampleInput()` to the Formatter class to improve #1050.


src/test/java/net/sf/jabref/logic/formatter/bibtexfields/UnicodeConverterTest.java
```diff
@@ -27,6 +27,12 @@ public void testBasic() {
     }
 
     @Test
+    public void testEmpty() {
```
tobiasdiez: not formatExample? The empty string should be covered by the general tests.

tobiasdiez:
LGTM :+1:
Could you please also add a test in FormatterTests that  getExampleInput does not return null or an empty string?

PR #1167, opened 2016-04-11
PositionWindow automatically installs resize event listener

oscargus:
Instead of always having to copy and paste the same code, the ComponentAdapter is now automatically installed.


Siedlerchr:
Good, but I think you missed one: 
`FindUnlinkeFilesDialog->initalizeAction()`

oscargus:
Not really. That one can be replaced with `PositionWindow`. This holds for `ImportInspectionDialog` as well. At the moment those two use another mechanism for this. I won't change that now though.

PR #1168, opened 2016-04-11
Fix #1163 by removing the strange functionality which enabled sorting…

### LINKED ISSUE #1163: Year sorting doesn't work
Clicking on the year column doesn't affect the sorting of the table.

simonharrer:
We have some very strange logic in JabRef which allows to handle years using only the two last numbers, so we can detect that 68 means 1968. This causes a lot of issues. I do not think that JabRef should still support such strange logic, and just use the value which is in the year field. 

Hence, I removed it. 

This also solves the issue of having that YearUtil in the model package which duplicated a method in the StringUtil class. 

Refs #1163


tobiasdiez:
LGTM :+1:

stefan-kolb:
LGTM :+1:

koppor:
Should we readd YearUtil as a cleanup to convert two digit years to four digit years?

Siedlerchr:
If we add it, we should make use of the java 8 functionality instead of the own ugly parsing:
http://stackoverflow.com/questions/32505490/how-to-change-the-base-date-for-parsing-two-letter-years-with-java-8-datetimefor

and this one:
http://stackoverflow.com/questions/29490893/parsing-string-to-local-date-doesnt-use-desired-century

Issue #1164, opened 2016-04-11
Use CircleCI only for building the binaries


simonharrer:
At the moment, gradle has issues with executing the tests on CircleCI due to 4GB RAM limit. I propose to use travis ci for testing and coverage analysis, and CircleCI for building the binaries only. 

The downside is that CircleCI can show test failures much better (travis ci only uses a textual representation whereas CircleCI parses the result xml and provides a nice result page).

stefan-kolb:
#1169

PR #1169, opened 2016-04-11
Testing is done at travis-ci, packaging on circleci

simonharrer:
- Implements #1164 
- Removes coverity (which is no longer being used) in favor of codacy
- Coveralls is removed in favor of codecov due to quicker check

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ before_script:
   - sleep 3 # give xvfb some time to start
 
 script:
-  - ./gradlew test integrationTest
+  - ./gradlew check integrationTest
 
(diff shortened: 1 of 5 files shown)


matthiasgeiger:
There seems to be a flaw in your description...

Coveralls and Coverity are removed and replaced by codecov and codacy?

koppor:
Reasoning:

 * Most coverity have been fixed. 
  * 13 identified medium issues are false positives.
  * we needed gradle work arounds to get it working
 * Codacy invests in its future and will integrate other tooling; also offers a badge
 * Coveralls is slower as codecov; used

PR #1170, opened 2016-04-11
Some minor cleanups

oscargus:
A little bit of everything... Maybe one should wait until after 3.3 is released before merging this?

### DIFF
--- a/src/main/java/net/sf/jabref/model/entry/Author.java
+++ b/src/main/java/net/sf/jabref/model/entry/Author.java
@@ -54,7 +54,7 @@ public class Author {
 
     public static String addDotIfAbbreviation(String name) {
         // Avoid arrayindexoutof.... :
-        if (name == null || name.isEmpty()) {
+        if ((name == null) || name.isEmpty()) {
             return name;
         }
         // If only one character (uppercase letter), add a dot and return immediately:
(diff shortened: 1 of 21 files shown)


src/main/java/net/sf/jabref/model/entry/Author.java
```diff
@@ -74,7 +74,7 @@ public static String addDotIfAbbreviation(String name) {
 
             if(currentChar == '.') {
                 // A.A. -> A. A.
-                if(i + 1 < name.length() && Character.isUpperCase(name.charAt(i + 1))) {
+                if(((i + 1) < name.length()) && Character.isUpperCase(name.charAt(i + 1))) {
```
tobiasdiez: I find it more readable without the brackets...

tobiasdiez:
LGTM :+1:

src/main/java/net/sf/jabref/model/entry/Author.java
koppor: Think, this is the eclipse setting we agreed about half year ago. We add brackets as much as possible to help readers not familiar with operator precedence to understand the code.

oscargus: That is an automatic Eclipse addition... All brackets or outer brackets? Outer brackets I find quite useful, but I agree that the ones surrounding i + 1 could be skipped. There are even some more annoying cases of these brackets, e.g. (or a quite similar case)

```
j = k - l + m;
```
becomes
```
j = (k - l) + m;
```

tobiasdiez: Oh ok, you already had a discussion about it, never mind.

PR #1171, opened 2016-04-11
Updated Spanish translation

jtornero:
Updated Spanish translation strings both in menu and JabRef localization files


matthiasgeiger:
Awesome! Thank you!

jtornero:
@matthiasgeiger there is a not successful automatic test. Looks that I have nothing to do with it... something to do on my side? And, by the way, I missed one string... what means From Import string?

Cheers

matthiasgeiger:
I don't really know why the code coverage sinks - causing to fail this check. But this is no problem - we'll merge this anyway.

Regarding the missing string "From import" You can see the context in this screenshot:

import-dupblicates

If you find an entry using the web search which already exists in your database, you can open this dialog to check whether those entries are really duplicates and which information should be used. On the left side there is the existing entry - on the right side there is the new entry "from [the] import".

Do you want to add the missing string? Or you can provide me the translation here and I'll add it.

Regards

matthiasgeiger:
Oh... I just checked your translation. Unformatunately you escaped special characters using the "HTML syntax" (e.g., `&oacute;`). This does not work for "normal" Java strings. 

We'll be able to convert this, but I'm curious how this happend. Have you escaped this manually? Or has this happened automatically? If so, which tools have you used to create the translations?

jtornero:
Hi Matthisas,

I think that for a single entry it is better that you translate it. 
Tried to do it in github but I'm ending with the usual mess. Git/Github 
is a beast I haven't been able to cope with yet.

I think we should translate two strings, taking into account the context:

Line 570

Remove_entry_from_import=Eliminar_entrada_importada

Line 1302

From_import=Entrada_importada

Cheers


Jorge

mlep:
@matthiasgeiger  One of the difficulty in translating the entries is to know the context in which a specific string is used. I have always thought there was no easy way to find in which window a specific string was displayed; and here you come directly with a screen shot!
So: is there an easy way to find the proper window, or does one have to be wizard?

jtornero:
Well... annoying... I just edited the file which was cloned after git 
clone....

And I noticed that the strings were scaped like html. So I wrote the 
text normally (with accents and all) and then replaced the characters 
(¿,,¡ñ,+á,é,... etc) for the equivalente html encoding. I simply used 
kate for the edition.

Cheers

Jorge

mlep:
Ooops: there is a space in the translation "Entrada importada"....

matthiasgeiger:
@jtornero Thanks for the strings and the explanation. There some few cases where longer texts are styled using HTML - However, for most cases you can directly use unicode characters.

@mlep Depends on the point of view whether some magic is needed or not :wink: You need to work withe the code and even then it is not always easy to produce the screens (e.g., for error messages).  What I did here was: Performing a full text search on the whole jabref project for `"From import"` (note: with space here, as there are spaces in the code, `_` is only relevant for the properties files). I found the string in the dialog for duplicate resolution and then I looked up, how to create the dialog for "import duplicate detection". 
So, as you can see: Not so easy to correctly show the context of the translations - even for us developers who know where to look.

mlep:
@matthiasgeiger Lesson: translators can ask developers if, by chance, magic can operate...
Thanks! :+1:

matthiasgeiger:
Merged with https://github.com/JabRef/jabref/commit/7dee04666f0f0dfe5f7658e67f633541def997e1.

Thanks again, Jorge!

Issue #1182, opened 2016-04-12
Brazilian Portuguese: UI remains in English


mlep:
JabRef version JabRef-3.3dev--snapshot--2016-04-12--master--f8bce59.jar on Linux Mint LMDE

Steps to reproduce:
1. Options -> Preferences
2. Select "Brazilian Portuguese"
3. Restart JabRef
   ==> English interface...
   (all other languages work as expected)

Log:

22:03:37.907 [AWT-EventQueue-0] WARN  net.sf.jabref.logic.l10n.Localization - Could not find bundles for language pt_br, switching to full english language
java.util.MissingResourceException: Can't find bundle for base name l10n/JabRef, locale pt_br
    at java.util.ResourceBundle.throwMissingResourceException(ResourceBundle.java:1564) ~[?:1.8.0_66]
    at java.util.ResourceBundle.getBundleImpl(ResourceBundle.java:1387) ~[?:1.8.0_66]
    at java.util.ResourceBundle.getBundle(ResourceBundle.java:890) ~[?:1.8.0_66]
    at net.sf.jabref.logic.l10n.Localization.createResourceBundles(Localization.java:41) ~[JabRef-3.3dev--snapshot--2016-04-12--master--f8bce59.jar:?]
    at net.sf.jabref.logic.l10n.Localization.setLanguage(Localization.java:32) ~[JabRef-3.3dev--snapshot--2016-04-12--master--f8bce59.jar:?]
    at net.sf.jabref.JabRef.start(JabRef.java:84) ~[JabRef-3.3dev--snapshot--2016-04-12--master--f8bce59.jar:?]
    at net.sf.jabref.JabRefMain.lambda$main$0(JabRefMain.java:10) ~[JabRef-3.3dev--snapshot--2016-04-12--master--f8bce59.jar:?]
    at java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:311) [?:1.8.0_66]
    at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:756) [?:1.8.0_66]
    at java.awt.EventQueue.access$500(EventQueue.java:97) [?:1.8.0_66]
    at java.awt.EventQueue$3.run(EventQueue.java:709) [?:1.8.0_66]
    at java.awt.EventQueue$3.run(EventQueue.java:703) [?:1.8.0_66]
    at java.security.AccessController.doPrivileged(Native Method) ~[?:1.8.0_66]
[...]

PR #1183, opened 2016-04-12
Rename pt_BR to pt_br, because the locale is converted to lowercase

Siedlerchr:
in EncodingControl resource bundle loading
Fix for #1182 
(I could not test it directly because I only have Win atm, which is case insensitive, but from debugging it was clear that it was a case-sensity).
So please test that. 

@mlep  Please notice that I renamed the files to lowercase only. You can try this directly in your JabRef installation, it should then work.


Siedlerchr:
Could someone of the linux users please take a look at that failing test? Can't reproduce that on win.

mlep:
Indeed, this works on my Linux computer: I get the menus translated in Brazilian Portuguese.
Great!

matthiasgeiger:
@Siedlerchr The problem is again casing: In `Languages.java` Portuguese is added as `        LANGUAGES.put("Brazilian Portugese", "pt_BR");` - and this locale abbreviation is used in the tests to find the files...

However, the format `pt_BR` or `languange_COUNTRY` are the official abbreviations so I don't like to change all keys and files to lowercase but I'd prefer to remove the lowercasing on resource loading. Can you change this in this PR?

mlep:
This could also be the opportunity to correct the missing "u":
"Brazilian Portugese" --> "Brazilian Portuguese"

Siedlerchr:
@matthiasgeiger  I will look at it, however the problem with the case is implicit in the EncodingControl, there a Locale is passed in the ResourceBundle methods (which in fact is the JAVA SDK implementation). 
However, the Locale Object is internally converted to lowercase --at least that is what I saw.
I will setup my VM today for testing directly on linux.

matthiasgeiger:
Ah okay... I thought we do this lowercasing somewhere... if this is an internal JDK thing it is indeed easier to use lowercase for our files.

stefan-kolb:
This is caused by the missuse of the Locale class. See http://stackoverflow.com/questions/3318465/missingresourceexception-cant-find-bundle-for-base-name.

PR #1184, opened 2016-04-12
Use Paths.get() instead of Url.getfile() - Fix failing tests

Siedlerchr:
Fixes problem with failing tests due to whitespaces


oscargus:
:+1:

PR #1196, opened 2016-04-13
 Add possibility to lookup DOI from BibTeX entry content

stefan-kolb:
image


tobiasdiez:
LGTM :+1:

matthiasgeiger:
Is it possible to Align the Buttons horizontally? In your Screenshot the DOI field requires so much space but a DOI is always a single line with at most ~50 characters

simonharrer:
LGTM (the suggestion of matthias could be followed if you have time @stefan-kolb but I would merge this for 3.3 either in this or in the improved state.

koppor:
We should not do it into v3.3 because of two new translation strings:

``
Lookup_DOI=
No_DOI_found=
```

See also https://github.com/JabRef/jabref/pull/1050#issuecomment-209641393

simonharrer:
We should make an exception here, as this feature is really awesome.

stefan-kolb:
We can rethink the layout later on I think. it is flawed in general...

matthiasgeiger:
:+1: for making the exception - DOI is is language independenden and "lookup" should be understandable for most users :wink:

PR #1201, opened 2016-04-14
Added method to generate string with n spaces or n characters

oscargus:
Used in a number of places so makes sense to have a method.

### DIFF
--- a/src/main/java/net/sf/jabref/logic/util/strings/StringUtil.java
+++ b/src/main/java/net/sf/jabref/logic/util/strings/StringUtil.java
@@ -669,4 +669,32 @@ public class StringUtil {
         return sb.toString().trim();
     }
 
+    /**
+     * Return a String with n spaces
+     *
+     * @param n Number of spaces
+     * @return String with n spaces
+     */
+    public static String nSpaces(int n) {
+        return nCharacters(n, ' ');
+    }
+
+    /**
+     * Return a String with n copies of the char c
+     *
+     * @param n Number of copies
+     * @param c char to copy
+     * @return String with n copies of c
+     */
+    public static String nCharacters(int n, char c) {
+        StringBuilder resultSB = new StringBuilder();
+
+        for (int i = 0; i < n; i++) {
+            resultSB.append(c);
+        }
+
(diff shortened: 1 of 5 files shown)


simonharrer:
LGTM

src/main/java/net/sf/jabref/logic/util/strings/StringUtil.java
```diff
@@ -669,4 +669,32 @@ public static String expandAuthorInitials(String name) {
...
+    }
+
+    /**
+     * Return a String with n copies of the char c
+     *
+     * @param n Number of copies
+     * @param c char to copy
+     * @return String with n copies of c
+     */
+    public static String nCharacters(int n, char c) {
```
tobiasdiez: Such a method already exists in Guava: http://docs.guava-libraries.googlecode.com/git-history/v10.0.1/javadoc/com/google/common/base/Strings.html#repeat%28java.lang.String,%20int%29
Should we use this method instead?
If not I would propose to change the name of `nCharacters` to `repeat(char, int)` and similarly `repeatSpaces(int)`

src/main/java/net/sf/jabref/logic/util/strings/StringUtil.java
```diff
@@ -669,4 +669,32 @@ public static String expandAuthorInitials(String name) {
...
+        StringBuilder resultSB = new StringBuilder();
```
tobiasdiez: Init StringBuilder with the right capacity `n`

src/main/java/net/sf/jabref/logic/util/strings/StringUtil.java
oscargus: I've seen references to it when googling, but wasn't sure which libraries we already use (there's also one in Apache StringUtils).

src/main/java/net/sf/jabref/logic/util/strings/StringUtil.java
oscargus: Good idea!

tobiasdiez:
LGTM :+1:

Issue #1153, opened 2016-04-09
File>Import into current database> default still Windows 1252


ajbelle:
JabRef version <3.2> on <Windows 7>

Steps to reproduce:

1. File>Import into current database> select 'utf8 encoded'.bib (Use the file you a viewing)
2. The import inspection dialog, and Possible duplicate entries so the imported file with coding corruption. eg: ° = Â° ©=Â©   “=â€œ ...
3. My guess is that the direct import has been changed from the OS default on the Window build to utf8, but when importing into the current database it is routing via the OS default encoding.
4. Reformatting the import.bib with text editor to CP1252 fixes the problem.

Possible relevant post are issue #262 @oscargus ,  [UTF8 import from bib file ](https://sourceforge.net/p/jabref/discussion/318825/thread/92143b77/#cc76), and #1080 ,but resolution is beyond me. There seem to have been edits [here](https://github.com/JabRef/jabref/commit/312fc6f3465554d86d28cb129058bd80feca0dd8) by  @lenhard 

Repeat: file opens fine, but File>Import into current database> or File>Import into new database> sees utf8 encoded file read as CP1252.

Siedlerchr:
Maybe related? #1080

ajbelle:
Yes quite possible @Siedlerchr, and that was one of the links (I have corrected them, THX). My impression is the 'issue' is not a coding bug, but is a systems level ommision or fault.

koppor:
What is the encoding of the opened database? My guess is that it is CP1252 instead of UTF8.

Related #558.

lenhard:
If you are using JabRef 3.2 then it cannot be caused by #1080 since this change is only available in the current development version.

ajbelle:
Thank you @lenhard I am actually running JabRef_windows-x64_3_3dev--snapshot--2016-03-25--livesearch--55fc5f2 which exhibited the fault, but backloaded V3.2 and it is identical. I checked V2.1 and same problem. That is why I suspect it is an omisison in the Windows build that has been there for a long time. I didn't want to confuse my post that it is a development build bug.

Your point that it doesn't seem to be  #1080 is valid.

tobiasdiez:
It seems to be fixed in the current development version. Probably as a result of https://github.com/JabRef/jabref/pull/1207. @ajbelle can you please double check that the issue is indeed fixed? Thanks!

stefan-kolb:
@ajbelle Thank you for your report :+1:
[...]

Issue #1080, opened 2016-04-01
BibTextImporter - No encoding specified in importEntries


Siedlerchr:
While doing #1079  I noticed that there is an Encoding issue when running gradlew check from commandline in Windows. The specified line is the following in `BibtexImporter` where only default Charset is used for the Inputstreamreader.
However, I was not sure if I can change it  safely  to UTF8 or if this would create problems.

````java
    public List<BibEntry> importEntries(InputStream in, OutputPrinter status)
            throws IOException {
        ParserResult pr = BibtexParser.parse(new InputStreamReader(in));
  ParserResult pr = BibtexParser.parse(new InputStreamReader(in));
````

lenhard:
Thanks for opening this issue. I often experienced the same problem, but did not follow it up.

In 312fc6f3465554d86d28cb129058bd80feca0dd8 I modified the `BibtexImporter` to use the default encoding set in the preferences, as other importers do. In a clean install, the default encoding is `UTF_8`. The commit fixed the problem for me, but if it persists for others, we could modify the preferences in `BibtexImporterTest` to be 100% sure it uses `UTF_8`.

Does it work for you @Siedlerchr ? If so, please close this issue.

koppor:
Related: https://github.com/JabRef/jabref/pull/515#issuecomment-164766590 and #558

stefan-kolb:
Whats the current status here?

lenhard:
@stefan-kolb This is waiting for confirmation from @Siedlerchr

Siedlerchr:
If you have tested it and it works then it's okay for me. 
Otherwise I will look at it by tomorrow.

lenhard:
It is indeed working for me, but an additional check by someone else never hurts :) It is sufficient to check if there are no more encoding related test failures in your windows command line on current master.

Siedlerchr:
@lenhard  Will check the current master now.

Siedlerchr:
@lenhard  :+1:  
There is one thing more, that is the BibDatabaseWriterTest, it always failed for me in roundtrip
The problem is, the complex.bib uses a UNIX Line ending (\n), while on windows it should be \r\n.
The StringWriter and the BibTexWriter uses Globals.Newline, which is on windows \r\n.

This is not a big problem for me.

simonharrer:
Closing this as the fix is conformed twice.

PR #1207, opened 2016-04-14
Restructure importers 

tobiasdiez:
The aim of this PR is to reduce the overlap between the Importer package and some code in GUI (in particular, the class OpenDatabaseAction).
Moreover, #1080 and #1153 are fixed.

In detail, the following things were changed:
- The abstract `ImportFormat` class is almost completely rewritten. Please have a look at this class. Almost all other changes are a direct consequence of the edits made to the ImportFormat class. The biggest change is that the import method returns a `ParserResult` instead of a list of entries (in this way, metadata can also be imported and error messages are communicated more effectively instead of throwing IOExceptions). Moreover, the import method operates on a reader instead of an input stream in order to reduce some common code in all importers.
-  Add method `importDatabase(Path file, Charset encoding)` to the importer interface which opens the file and passes the stream to the other import method.
- The BibTexImporter overwrites this new import method and uses the logic from `OpenDatabaseAction.loadDatabase` to determine the encoding of the bib file.
- Change OpenDatabaseAction to operate against the BibTex importer. 
- Remove get/set isCustomImporter: was not really used (just for sorting - now everything is alphabetical)
- Empty entries are no longer imported (thus a few test had to be changed)

Remark: The methods getExtension and getDescription are not implemented by almost any importer. This should be done. Maybe this is something for the stupro?

(Will add the changelog entry and add a few tests later)


tobiasdiez:
Any comments or can I rebase and merge?

src/main/java/net/sf/jabref/importer/fileformat/RisImporter.java
```diff
@@ -38,56 +38,42 @@
...
     }
 
-    /*
-     *  (non-Javadoc)
-     * @see net.sf.jabref.imports.ImportFormat#getCLIId()
-     */
     @Override
-    public String getCLIId() {
-        return "ris";
+    public List<String> getExtensions() {
```
oscargus: Why return null here (and in similar places)?

tobiasdiez: Since I have no idea for which extensions these importers support. 

See above:
> Remark: The methods getExtension and getDescription are not implemented by almost any importer. This should be done. Maybe this is something for the stupro?

oscargus:
I do not fully understand why the getCLIId method is removed? How is the importer selected from CLI now? 

Also, since there is a method setCliId, maybe that should change name as well (as it appears as getCLIId is sometimes replaces with getId, somehow, it seems like getCLIId is a better name, but no strong opinions).

Many changes are related to imports and replacing extension strings with a list of extension strings, right?

src/main/java/net/sf/jabref/importer/fileformat/RisImporter.java
oscargus: Ah, OK. Tired... Somehow, the questions was also, why not an empty list?

oscargus:
Now I think I understand why. However, I think that getCLIId and getFormatName in general should return different things...

oscargus:
Well, maybe I do not understand the code fully (again, tired), but what I was worried about with getId doesn't seem to the case, so to me this looks good!

tobiasdiez:
There is a method getId (previously getCLIID) in ImportFormat which constructs the id based on the name of the formatter. Thus some of the hand-coded ids in the specific importers could be removed.

oscargus:
I saw `"REPEC New Economic Papers (NEP)"` and then that getId() used that string which caused one comment, but then I saw the tests and realized that one wasn't supposed to type something like that. :-) Now, I also know why.

:+1:

simonharrer:
Waiting for the last #547 to be merged - then, this PR can clean up all the importer code uniformly.

simonharrer:
This is unblocked now. Go 4 it. :)

Issue #1091, opened 2016-04-03
Structure of imports


tobiasdiez:
As discovered in #1089 there might be a problem with IntelliJ restructuring the imports vs how Eclipse does it. I just run into the same problem with #1090, so I would like to discuss here what the correct preferences for IntelliJ are. The following settings produce a somewhat satisfying result, but there are still quite a few difference to some imports in some files.
![untitled](https://cloud.githubusercontent.com/assets/5037600/14232775/2cbbfa04-f9b5-11e5-80ee-4f9743df5431.png)
.

Siedlerchr:
I just posted this at the PR:
Maybe you could evaluate that:
http://stackoverflow.com/questions/14716283/is-it-possible-for-intellij-to-organize-imports-the-same-way-as-in-eclipse
http://stackoverflow.com/questions/11704327/intellij-idea-how-to-optimize-imports-automatically-after-each-save

simonharrer:
We need something like checkstyle which enforces it using gradle. 

We should maybe activate checkstyle for this, making the build fail if the imports are not in order.

bartsch-dev:
I looked through some of the source files and have noticed that there is no clear order in which the classes are imported.

Thus I think we firstly need to decide on a proper order (like alphabetically with a blank line between).

Furthermore I prefer the fully qualified Imports over the generalized ones (*).

simonharrer:
Checkstyle checks are now in place for forcing one to remove unused imports. We can add additional checks, e.g., for the import order, BUT, we should be careful not to add to much, and only add additional checks as an evolution of the code base. See 9e9fc8c

simonharrer:
![image](https://cloud.githubusercontent.com/assets/375693/14459902/0597da6c-00bb-11e6-84e4-b0d4d978f2d1.png)

simonharrer:
![image](https://cloud.githubusercontent.com/assets/375693/14459998/7aeb7026-00bb-11e6-96aa-0adb1564b97e.png)

koppor:
In Eclipse automatically generated - use `gradlew eclipse`.

![grabbed_20160412-143351](https://cloud.githubusercontent.com/assets/1366654/14460040/9a9ce88c-00bb-11e6-83e7-52a482dad3d2.png)

simonharrer:
![image](https://cloud.githubusercontent.com/assets/375693/14460173/5113c518-00bc-11e6-89b8-e1f221001a7d.png)

[...]

PR #1214, opened 2016-04-15
BasePanel now accepts pasted DOIs and tries to retrieve the entry

stefan-kolb:
Just paste a DOI inside the main table and the fetcher will try to get the bibtex entry for it.
Works now for DOI, copying entries from other JabRef instances, and plain text BibTeX.


src/main/java/net/sf/jabref/gui/BasePanel.java
```diff
@@ -54,6 +52,9 @@
 import javax.swing.undo.CannotRedoException;
 import javax.swing.undo.CannotUndoException;
 
+import ca.odell.glazedlists.event.ListEventListener;
+import com.jgoodies.forms.builder.FormBuilder;
+import com.jgoodies.forms.layout.FormLayout;
 import net.sf.jabref.BibDatabaseContext;
```
tobiasdiez: According to https://github.com/JabRef/jabref/issues/1091, there should be an empty line before the JabRef-specific imports.

tobiasdiez: According to https://github.com/JabRef/jabref/issues/1091, there should be an empty line before the JabRef-specific imports.

tobiasdiez:
LGTM :+1:

Issue #405, opened 2015-11-27
Preserve capital Umlauts during cleanup


grimes2:
Ägypten in title field results in  {\"{A}}gypten. This should be the correct syntax, but the result of the pdf is ägypten (field title only!). {{\"{A}}}gypten produces Ägypten. This might be a bibtex bug, because the author field works without additional braces.

oscargus:
Oh, yeah. I was thinking of that earlier in another discussion. If I recall it correctly, there's a parity behavior, so one needs an odd number of brackets to keep the case.

grimes2:
I've checked the syntax: ü is {\"{u}}, Ü is {{\"{U}}} (to preserve capital letter two additional curley braces are needed). This bug should be fixed.

lenhard:
@oscargus What is the status of this issue?

oscargus:
One would need to add an extra set of {} around all capital entries (not characters) in the huge array, soon to be renamed in #841 .

stefan-kolb:
Has this been addressed in #841?

oscargus:
Not consistently, no. There may be capital letters that doesn't have double
braces around them.

PR #1215, opened 2016-04-15
Improved conversion maps and conversion speed

oscargus:
The HTML/Unicode <-> LaTeX conversion map source was based on regular expressions making it a jungle of backslashes. Now, replace is used instead of replaceAll so fewer backslashes and (theoretically) a bit faster.

I also added benchmarking for the following conversions:
* HTML to LaTeX (used at import, cleanup and save action)
* LaTeX to HTML (used for preview)
* LaTeX to Unicode (used for table rendering)
and I got a slight improvement, although I cannot really tell why the two latter are faster now as they are based on cleaned versions... One reason might be that the tests are based on LaTeX symbols with more backslashes in them for the old case...

Before:
Benchmarks.htmlToLatexConversion      5533.070 ▒   86.148  ops/s
Benchmarks.latexToHTMLConversion     141500.226 ▒ 8011.038  ops/s
Benchmarks.latexToUnicodeConversion  124527.930 ▒ 3984.776  ops/s

After:
Benchmarks.htmlToLatexConversion     6153.450 ▒   114.716  ops/s
Benchmarks.latexToHTMLConversion     168037.040 ▒ 13412.341  ops/s
Benchmarks.latexToUnicodeConversion  173674.308 ▒  7455.006  ops/s

Here one can see that only searching the string to be converted once is far better than running replace on all possible characters. Which is not surprising, but require quite complicated code.

Finally, with the simpler strings it was easier to add most (all?) of the curly brackets required to solve #405 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - Fixed [#1457](https://github.com/JabRef/jabref/issues/1457): Support multiple words inside LaTeX commands to RTF export
 - Entries retain their groupmembership when undoing their cut/deletion
 - Fixed [#1450](https://github.com/JabRef/jabref/issues/1450): EntryEditor is restored in the correct size after preference changes
+- Fixed [#405](https://github.com/JabRef/jabref/issues/405): Added more {} around capital letters in Unicode/HTML to LaTeX conversion to preserve them
 
 ### Removed
 - Removed possibility to export entries/databases to an `.sql` file, as the logic cannot easily use the correct escape logic
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -138,7 +138,10 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - [#969](https://github.com/JabRef/jabref/issues/969) Adding and replacing old event system mechanisms with Google Guava EventBus.
 
 ### Fixed
+<<<<<<< 973a39d9fa61ad9149db778e27112ecaeb13d712
 - Alleviate multiuser concurrency issue when near simultaneous saves occur to a shared database file
+=======
+>>>>>>> Added double {} around all symbols with capital letters to avoid lowercasing in titles. Fixed #405
 - Fixed [#318](https://github.com/JabRef/jabref/issues/318): Improve normalization of author names
 - Fixed [#598](https://github.com/JabRef/jabref/issues/598) and [#402](https://github.com/JabRef/jabref/issues/402): No more issues with invalid icons for ExternalFileTypes in global search or after editing the settings
 - Fixed [#883](https://github.com/JabRef/jabref/issues/883): No NPE during cleanup
--- a/src/jmh/java/net/sf/jabref/benchmarks/Benchmarks.java
+++ b/src/jmh/java/net/sf/jabref/benchmarks/Benchmarks.java
@@ -18,6 +18,7 @@ import net.sf.jabref.exporter.SaveException;
 import net.sf.jabref.exporter.SavePreferences;
 import net.sf.jabref.importer.ParserResult;
 import net.sf.jabref.importer.fileformat.BibtexParser;
+import net.sf.jabref.logic.formatter.bibtexfields.HtmlToLatexFormatter;
 import net.sf.jabref.logic.layout.format.HTMLChars;
 import net.sf.jabref.logic.layout.format.LatexToUnicodeFormatter;
 import net.sf.jabref.logic.search.SearchQuery;
@@ -39,7 +40,8 @@ public class Benchmarks {
 
     String bibtexString;
     BibDatabase database = new BibDatabase();
-    List<String> conversionStrings = new ArrayList<>();
+    List<String> latexConversionStrings = new ArrayList<>();
+    List<String> htmlConversionStrings = new ArrayList<>();
 
     @Setup
     public void init() throws IOException, SaveException {
@@ -73,7 +75,18 @@ public class Benchmarks {
         sb.append(latexSymbols.get(Math.abs(randomizer.nextInt()) % symbolcount));
         sb.append(latexSymbols.get(Math.abs(randomizer.nextInt()) % symbolcount));
         sb.append("} abc");
-        conversionStrings.add(sb.toString());
+        latexConversionStrings.add(sb.toString());
+
+        List<String> htmlSymbols = new ArrayList<>(HTMLUnicodeConversionMaps.HTML_LATEX_CONVERSION_MAP.keySet());
+        symbolcount = htmlSymbols.size();
+        sb = new StringBuilder();
+        sb.append("A <b>bold</b> ");
+        sb.append(htmlSymbols.get(Math.abs(randomizer.nextInt()) % symbolcount));
+        sb.append(" <it>italic</it> ");
+        sb.append(htmlSymbols.get(Math.abs(randomizer.nextInt()) % symbolcount));
+        sb.append(htmlSymbols.get(Math.abs(randomizer.nextInt()) % symbolcount));
+        sb.append("&#8211; abc");
+        htmlConversionStrings.add(sb.toString());
     }
 
     @Benchmark
@@ -112,7 +125,7 @@ public class Benchmarks {
     public List<String> latexToUnicodeConversion() {
         List<String> result = new ArrayList<>(1000);
         LatexToUnicodeFormatter f = new LatexToUnicodeFormatter();
-        for (String s : conversionStrings) {
+        for (String s : latexConversionStrings) {
             result.add(f.format(s));
         }
         return result;
@@ -122,7 +135,17 @@ public class Benchmarks {
     public List<String> latexToHTMLConversion() {
         List<String> result = new ArrayList<>(1000);
         HTMLChars f = new HTMLChars();
-        for (String s : conversionStrings) {
+        for (String s : latexConversionStrings) {
+            result.add(f.format(s));
+        }
+        return result;
+    }
+
+    @Benchmark
+    public List<String> htmlToLatexConversion() {
+        List<String> result = new ArrayList<>(1000);
+        HtmlToLatexFormatter f = new HtmlToLatexFormatter();
+        for (String s : htmlConversionStrings) {
             result.add(f.format(s));
         }
         return result;
--- a/src/jmh/java/net/sf/jabref/benchmarks/Benchmarks.java
+++ b/src/jmh/java/net/sf/jabref/benchmarks/Benchmarks.java
@@ -22,7 +22,6 @@ import net.sf.jabref.logic.formatter.bibtexfields.HtmlToLatexFormatter;
 import net.sf.jabref.logic.layout.format.HTMLChars;
 import net.sf.jabref.logic.layout.format.LatexToUnicodeFormatter;
 import net.sf.jabref.logic.search.SearchQuery;
-import net.sf.jabref.logic.util.strings.HTMLUnicodeConversionMaps;
 import net.sf.jabref.model.database.BibDatabase;
 import net.sf.jabref.model.database.BibDatabaseMode;
 import net.sf.jabref.model.database.BibDatabaseModeDetection;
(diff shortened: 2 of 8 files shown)


src/jmh/java/net/sf/jabref/benchmarks/Benchmarks.java
```diff
@@ -26,8 +30,10 @@
 @State(Scope.Thread)
 public class Benchmarks {
 
-    String bibtexString;
-    BibDatabase database = new BibDatabase();
+    private String bibtexString;
+    private final BibDatabase database = new BibDatabase();
+    private final List<String> latexConversionStrings = new ArrayList<>();
```
tobiasdiez: This can be a simple string, right?

src/jmh/java/net/sf/jabref/benchmarks/Benchmarks.java
```diff
@@ -52,6 +58,27 @@ public void init() throws IOException, SaveException {
                 new SavePreferences());
         bibtexString = stringWriter.toString();
 
+        List<String> latexSymbols = new ArrayList<>(HTMLUnicodeConversionMaps.UNICODE_LATEX_CONVERSION_MAP.values());
```
tobiasdiez: I would prefer a simple hard-coded input. Mainly because it reduces code complexity. Moreover, it serves as a good baseline and then you might add some variations (like format of strings which need cleanup vs strings which don't).

src/jmh/java/net/sf/jabref/benchmarks/Benchmarks.java
oscargus: Except that each run converts 1000 strings. But if a static string should
be used I guess it is enough to create a single (quite long) string.

tobiasdiez:
With the benchmarks in place, you might also try out the recommendations of http://stackoverflow.com/questions/4285083/fastest-way-to-perform-a-lot-of-strings-replace-in-java and http://stackoverflow.com/questions/16228992/commons-lang-stringutils-replace-performance-vs-string-replace.
. So using Apache's StringUtils.replace() and/or 
> Don't make a pattern for every single replaceAll line, but try to combine them (regex has a OR operator) and let Pattern optimize the regex. Do use the compiled pattern and don't compile it in every call, it's fairly expensive.

tobiasdiez:
With the benchmarks in place, you might also try out the recommendations of http://stackoverflow.com/questions/4285083/fastest-way-to-perform-a-lot-of-strings-replace-in-java and http://stackoverflow.com/questions/16228992/commons-lang-stringutils-replace-performance-vs-string-replace.
. So using Apache's StringUtils.replace() and/or 
> Don't make a pattern for every single replaceAll line, but try to combine them (regex has a OR operator) and let Pattern optimize the regex. Do use the compiled pattern and don't compile it in every call, it's fairly expensive.

oscargus:
At the moment there are (almost) only replace, so very few actual regexes. There are a few cases where the several characters are replaced with the same, so in those cases it may be worthwhile to use an or-regex.

Still, I think the way to go here (for HTML to LaTeX) would be to parse the string and once either HTML tags or HTML-encoded characters are found those are replaced, rather than blindly replacing all possible characters. In that way one can probably honor more tags as well.

src/jmh/java/net/sf/jabref/benchmarks/Benchmarks.java
tobiasdiez: Why 1000? I don't see a for loop in your code.

oscargus: You are right... Forgot to do that part... Anyway, maybe better to have a single longer hard-coded string.

simonharrer:
I would not prematurely optimize things which does not reduce the speed of JabRef.

oscargus:
Except that optimizing code is quite fun. :speedboat: 

I'm quite sure #405 should be fixed now.

src/test/java/net/sf/jabref/logic/formatter/bibtexfields/HtmlToLatexFormatterTest.java
```diff
@@ -45,7 +45,7 @@ public void testHTML() {
         assertEquals("{\\\"{a}}", formatter.format("&auml;"));
         assertEquals("{\\\"{a}}", formatter.format("&#228;"));
         assertEquals("{\\\"{a}}", formatter.format("&#xe4;"));
-        assertEquals("{$\\Epsilon$}", formatter.format("&Epsilon;"));
+        assertEquals("{{$\\Epsilon$}}", formatter.format("&Epsilon;"));
```
tobiasdiez: What is the function of all these brackets and escaped backslash? I would have expected `$\Epsilon$` (or even better `\( \Epsilon \)` )

oscargus: Well, ä is correctly written as  {\"{a}} which as a Java String becomes
that.

\Epsilon is only valid in math mode and for the \Epsilon to stay a capital
\Epsilon double {} are needed in the title field, see #405.

tobiasdiez: In my opinion the capitalization shouldn't be handled by the HtmlToLatex formatter; we have a special cleanup for preserving capitalization. BibLatex has absolutely no problem with a simple $\Epsilon$. Similarly, I don't see any reason to not write ä as `{\\"a}` (so removing the brackets around the a). Maybe, Biblatex don't even need the outer brackets.

oscargus: It is not actually capitalization. It is about replacing the HTML/Unicode symbol with the correct LaTeX command. Consider `\Leftrightarrow`. And see #405. Btw, the cleanup doesn't handle this case so it isn't a solution (try `\AA`). I cannot say anything about Biblatex, but for BibTeX it is an actual problem.

For the umlaut, see:
http://tex.stackexchange.com/questions/57743/how-to-write-%C3%A4-and-other-umlauts-and-accented-letters-in-bibliography
https://en.wikibooks.org/wiki/LaTeX/Special_Characters
and consider using another accent, like `\k`. so `{\ka}` based on your suggestion. I read a very good explanation about why `{\"{a}}` and the like is the correct way, although other cases work in specific situations, but I cannot find the link right now. I doubt Biblatex has solved the final case, unless it is at the end of a word (LaTeX will complain if nothing else. Again, imagine `{\ka}` (which you can write `{\k a}` to get around the first problem), inside a word, so `b\kab` or `b\k ab` or `b\k a b`. I'm quite sure none of these will give the correct result.

simonharrer: I do use a lot of `\"a` in the bin files without braces. Is this handled with this Mapping as well? Or do we need a Checker for detecting the usage without Curly braces?

tobiasdiez: Yes, and the correct latex command is `$\Epsilon$` without any braces! The reason I'm insisting on this is that BibLatex/Biber can automatically convert Latex code to Unicode, but then of course `{{\"a}}` is converted to `{ä}` which is not the desired output.

By the way, it is better to preserve the capitalization of `Ägypten` by `{{\"{A}}gypten}` (i.e. the braces enclose the whole word and not just the capital letter so that kerning works as expected), see http://tex.stackexchange.com/a/140071/34554. If the preserve capitalization cleanup is not able to do this right now, then it should be improved...

Finally, have a look at https://github.com/plk/biber/blob/dev/lib/Biber/LaTeX/recode_data.xml which provides a huge map from latex to unicode (and I like the data scheme more then our internal maps!). Finally, https://github.com/plk/biber/blob/dev/lib/Biber/LaTeX/Recode.pm may also be helpful.

simonharrer: We should also think about having a LaTeX free Version of the field value in the bibentry required for searching without braces. This class should be able to provide this.

oscargus: Writing `\"a` is OK, but it is not wise for a piece of software to replace ä with `\"a` as there are many situations where it will lead to errors. So, if a single context independent string should be used it is `{\"{a}}` as that will always give the correct output. Same thing with `&Epsilon;`, if we want to guarantee that Ε is produced under all circumstances, we should use `{{$\Epsilon$}}`. Yes, kerning will not be perfect, but I'd rather take that than ε.


One should probably differ between two cases here. The Ägypten example is actually quite good. Somehow, I would think that this should be converted to {\"{A}}gypten as we cannot know that the Ä should always be capital (in the general case). On the other hand, in most cases, if we get &Delta; we can be quite sure that we actually want  Δ  as this is most likely used as a symbol rather than a character in a word. So, yes, somehow it is inconsistent.

Automatically adding {} will never work properly as *it is completely OK to have capital letters without {} in a .bib-file.* One should just be aware of the fact that the .bst determines the outcome. I'm therefore quite doubtful about that checker and will never add {} to any capital letters as long as I know for sure that they always should be capital as the .bst-file will take care of it (and that is the purpose of the .bst-file). That is the reason of the CaseKeeper using a list of words that should be kept in capital form. An improvement here would be to either extend these lists and/or support a user-defined list (that would solve the Ägypten case in a clearly better way).

If someone wants to write a better converter feel free to do so, but please make sure that it is indeed better and produces consistent code yielding correct output in the final document. The current one is far from perfect, but it does its job quite well when looking at the end result (it has saved me an immense amount of time if nothing else) and not merging these things in because someone might write a better converter at some stage seems quite stupid to me.

Again, feel free to update the internal format, but make sure you also add the HTML stuff. And one can wonder how often ✎ is really needed...

oscargus: And one will need to distinguish between LaTeX and BibTeX. It is like writing "the correct way to write JabRef in LaTeX is `JabRef` not `{JabRef}`!".

Finally, do not forget that lots of people still use BibTeX and not Biblatex. For lots of reasons. And if Biblatex had a good enough converter it would produce `ä` not `{ä}`. JabRef will convert `{{{{{{{\"{{{{a}}}}}}}}}}}`to `ä`. Despite using a non-XML format of the conversion table.

tobiasdiez: Well, if you think these extra braces are that important then go for it and hope every biblatex user actually uses Unicode...

oscargus: But do these braces end up in the final (dvi/ps/pdf) document? The only way for that to happen is if Biblatex escapes them...

tobiasdiez: Yes, they end up in the output the same way as they do for bibtex: as kerning problems.

oscargus: OK. Talking about kerning problems one shouldn't use `{\"{a}}` etc at all
(which is why Biblatex converts to Unicode first). But again, choosing
between kerning problems and wrong symbol, I'd pick kerning problems any
day of the week.

And in that particular example the only potential kerning problems is for
Epsilon, not for ä, which is most likely not part of a word anyway (I have
a hard problem believing that Greek people write their references using
equations...).

It is indeed a bit annoying that the extra braces are only needed in the
title field though.

simonharrer:
What is the current state here? Can this be merged?

oscargus:
This should be good to go now. Only better compared to the earlier version although of course, the "correct" way to convert can be argued infinitely...

stefan-kolb:
Thanks @oscargus :+1:

CHANGELOG.md
```diff
@@ -137,7 +138,10 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - [#969](https://github.com/JabRef/jabref/issues/969) Adding and replacing old event system mechanisms with Google Guava EventBus.
 
 ### Fixed
+<<<<<<< 973a39d9fa61ad9149db778e27112ecaeb13d712
```
tobiasdiez: Please remove these conflict markers.

stefan-kolb: I did that. Manually merged this.

PR #1216, opened 2016-04-16
Refactored CLI code

oscargus:
Moved code related to the CLI to the cli package, including half of JabRef.java.


simonharrer:
Idea: at the moment we have a somewhat intermingled relationship between JabRefMain, JabRef, JabRefFrame and now the ArgumentProcessor. Especially the purpose of the JabRef class is not that clear. 

I think it would be best if we could make the relationship more clear, especially separating the GUI and the CLI. I am currently unsure how it should be done, but I think making the relationship between the CLI and the GUI more explicit with interfaces could help as well as working on this unclear JabRef class. 

At the moment the flow is directed using "Main -> CLI -> GUI" but does the CLI really need a dependency of the GUI? Maybe it should be converted to something like the following:

```
Main calls CLI and get a cliResult
Main calls GUI with cliResult
GUI decides whether to start up depending on the cliResult
```

This would allow to separate CLI and GUI, testing CLI without any GUI parts.

oscargus:
Good idea! I think that this PR, if nothing else, makes it a bit easier to see the connections. For example, the JabRef instance was passed in some placed where it actually wasn't needed since processArguments behaved as a static function, although it wasn't declared as one and the JabRefCLI field was reinitialized on each call anyway. I'll try to dig a bit more into it.

oscargus:
I've looked into it a bit more carefully now and it, sort of, already works like that.

In `JabRef.java`, first the ArgumentProcessor is called, processing all the CLI stuff.

Then, if we are happy there, JabRef closes. If not, the GUI is started providing the required information from the ArgumentProcessor.

I do not really see the dependency on the GUI in the CLI (especially now that the CLI is no longer in `JabRef.java`. The only thing that can be argued in that `JabRefMain` starts `JabRef` as `invokeLater`, but I have no idea why a new thread is used there and what is then executed in the main thread.

oscargus:
There is an import of `JabRef` in `ArgumentProcessor` but that is to the static method `openBibFile` which should not necessarily be in `JabRef` (and maybe not at all as there should be other ways to do the same thing, the special thing here seems to be if the auto save is considered or not).

Of course one can imagine moving `JabRef.openWindow()` to a separate class as that is where the GUI really starts.

src/main/java/net/sf/jabref/JabRef.java
```diff
@@ -118,400 +117,21 @@ public void start(String[] args) {
         // The preferences return the system newline character sequence as default
         Globals.NEWLINE = Globals.prefs.get(JabRefPreferences.NEWLINE);
 
-        Optional<Vector<ParserResult>> loaded = processArguments(args, true);
+        // Process arguments
+        ArgumentProcessor argumentProcessor = new ArgumentProcessor(args, true);
```
simonharrer: Instead of passing booleans to constructors, it is better to use an enum instead. Because I have no idea what the true in that case means.

src/main/java/net/sf/jabref/JabRef.java
```diff
@@ -736,6 +352,13 @@ private void openWindow(List<ParserResult> loaded) {
...
     }
 
+    /**
+     *
+     * @param name Name of the bib-file to open
+     * @param ignoreAutosave true if autosave version of the file should be ignored
+     * @return ParserResult which never is null
+     */
+
     public static ParserResult openBibFile(String name, boolean ignoreAutosave) {
```
simonharrer: Could this method be extracted into a separate class, something like a AutosaveAwareParser?

src/main/java/net/sf/jabref/JabRef.java
```diff
@@ -575,7 +196,7 @@ private void setLookAndFeel() {
         }
     }
 
-    private void openWindow(List<ParserResult> loaded) {
+    private void openWindow(List<ParserResult> loaded, boolean isBlank) {
```
simonharrer: We could move this method and other code to a JabRefGUI class which then loads the gui, sets the look and feel, etc.

src/main/java/net/sf/jabref/JabRef.java
oscargus: true ;-)

oscargus:
OK, the new structure looks like this:

* `JabRefMain` is replaced with `JabRef`.
* `JabRefGUI` starts up the GUI from the constructor and is the place to go for the static `JabRefFrame` `mainFrame` earlier in `JabRef`
* `AutosaveAwareDatabaseLoader` now resides the openBibFile method

src/main/java/net/sf/jabref/cli/ArgumentProcessor.java
```diff
@@ -0,0 +1,548 @@
...
+    private final Mode startupMode;
+
+
+    public ArgumentProcessor(String[] args, Mode startupMode) {
+        cli = new JabRefCLI(args);
+        this.startupMode = startupMode;
+        parserResults = processArguments();
+    }
+
+    public Optional<List<ParserResult>> getParserResults() {
```
tobiasdiez: I think it is cleaner to return an empty list instead of Optional.empty

tobiasdiez: Rename to `getFilesToOpen`?

src/main/java/net/sf/jabref/cli/ArgumentProcessor.java
```diff
@@ -0,0 +1,548 @@
...
+        importAndOpenFiles(loaded);
+
+        if (!cli.isBlank() && cli.isFetcherEngine()) {
+            fetch(cli.getFetcherEngine()).ifPresent(loaded::add);
+        }
+
+        if (cli.isExportMatches()) {
+            if (!loaded.isEmpty()) {
+                String[] data = cli.getExportMatches().split(",");
+                String searchTerm = data[0].replace("\\$", " "); //enables blanks within the search term:
```
tobiasdiez: Extract export matches to new method.

src/main/java/net/sf/jabref/cli/ArgumentProcessor.java
```diff
@@ -0,0 +1,548 @@
...
+        }
+
+        if (usageMsg) {
+            System.out.println(Localization.lang("no base-BibTeX-file specified") + "!");
+            System.out.println(Localization.lang("usage") + " :");
+            System.out.println("jabref --aux infile[.aux],outfile[.bib] base-BibTeX-file");
+        }
+    }
+
+    private void importAndOpenFiles(List<ParserResult> loaded) {
```
tobiasdiez: Return a list of loaded files instead of accepting it as a parameter?

src/main/java/net/sf/jabref/importer/AutosaveAwareDatabaseLoader.java
```diff
@@ -0,0 +1,78 @@
...
+
+
+    /**
+     *
+     * @param name Name of the bib-file to open
+     * @param ignoreAutosave true if autosave version of the file should be ignored
+     * @return ParserResult which never is null
+     */
+
+    public static ParserResult openBibFile(String name, boolean ignoreAutosave) {
```
tobiasdiez: Move to OpenDatabaseAction?

src/main/java/net/sf/jabref/cli/ArgumentProcessor.java
oscargus: I was thinking about that, but as there is a return in that part it was a bit of a hassle to rework the logic...

src/main/java/net/sf/jabref/cli/ArgumentProcessor.java
oscargus: Done (not the renaming though).

src/main/java/net/sf/jabref/importer/AutosaveAwareDatabaseLoader.java
oscargus: I moved it there, not sure if it is better or not.

src/main/java/net/sf/jabref/cli/ArgumentProcessor.java
oscargus: Good idea!

simonharrer:
I think this can be merged. I do like the new structure. However, why not rename JabRef to JabRefMain to make it explicit that this class is used for starting?

src/main/java/net/sf/jabref/cli/ArgumentProcessor.java
tobiasdiez: Returning a boolean which indicates whether the call was successful or not does not work?

src/main/java/net/sf/jabref/importer/OpenDatabaseAction.java
```diff
@@ -454,4 +454,64 @@ private static InputStreamReader openFile(File fileToOpen, Optional<Charset> enc
...
     }
+
+    /**
+    *
+    * @param name Name of the bib-file to open
+    * @param ignoreAutosave true if autosave version of the file should be ignored
+    * @return ParserResult which never is null
+    */
+
+    public static ParserResult openBibFile(String name, boolean ignoreAutosave) {
```
tobiasdiez: I think this method should be renamed and get a more meaningful javadoc.

src/main/java/net/sf/jabref/importer/ParserResult.java
```diff
@@ -24,15 +24,15 @@
...
+import net.sf.jabref.MetaData;
 import net.sf.jabref.model.database.BibDatabase;
 import net.sf.jabref.model.database.BibDatabases;
 import net.sf.jabref.model.entry.BibEntry;
 import net.sf.jabref.model.entry.EntryType;
 
 public class ParserResult {
-    public static final ParserResult INVALID_FORMAT = new ParserResult(null, null, null);
-    public static final ParserResult FILE_LOCKED = new ParserResult(null, null, null);
+    public static final ParserResult NULL_RESULT = new ParserResult(null, null, null);
```
tobiasdiez: Does this really have to be public?

oscargus: I think so as it is accessed and compared against in quite a few places. The alternative would be to provide a method that compares against it (or rather check that the three parameters are null). In general, I do not really get the logic for all these comparisons...

oscargus:
Oh, I thought JabRef made much more sense to start from than JabRefMain. :-) For sure that is what I would have guessed for if I was trying to figure out the main class.

src/main/java/net/sf/jabref/importer/OpenDatabaseAction.java
oscargus: Better name for sure. Any suggestion? Javadoc: I see the point, but not really sure what to write than, more or less, combine the argument descriptions "Open bib-file and possibly ignore autosave file", sort of.

src/main/java/net/sf/jabref/cli/ArgumentProcessor.java
oscargus: Btw, an empty list can be returned although the call is correct and valid, e.g. everytime JabRef is started without any argument, so some additional logic was required to deal with the erroneous cases (which is why the Optional was there in the first place).

src/main/java/net/sf/jabref/importer/ParserResult.java
tobiasdiez: +1 for encapsulating the check as a a pblic method in ParserResult.

src/main/java/net/sf/jabref/importer/OpenDatabaseAction.java
tobiasdiez: Something like `loadDatabaseOrAutoSave`? I have no idea what this method actually tries to do and how it compares against the other methods in the OpenDatabaseAction class. This was also the point of me wanting a more descriptive javadoc.

oscargus: I have no idea either. ;-) But let me see what I can figure out.

src/main/java/net/sf/jabref/cli/ArgumentProcessor.java
oscargus: Yes, that is indeed an option.

Issue #1212, opened 2016-04-14
Saved file cannot be read


oscargus:
Based on the discussion in #1188 I created an empty database with a single entry with a single field containing `&#8211;&#8211;`, which is saved as

```
% Encoding: UTF-8

@Article{,
  author = {\&} # 8211;& #{8211;},
}

@Comment{jabref-meta: databaseType:bibtex;}
```

Log messages:

WARN  net.sf.jabref.importer.fileformat.BibtexParser - Could not parse entry
java.io.IOException: Error in line 4 or above: Empty text token.
This could be caused by a missing comma between two fields.

WARN  net.sf.jabref.importer.fileformat.BibtexParser - Could not parse entry
java.io.IOException: Error in line 4: Expected { or ( but received }

I notice two things here, except that it doesn't work (I think one reason is that a string cannot start with a digit as I recall it, although that is probably not the reason here). First, & is escaped. Why? What if I want to have a & in my field? One reason could be in an abstract where a slightly more complicated equation or tabular might theoretically be included. Second, there should probably be a space after the second #. Although not strictly required, it would be nice for symmetry.

Suggestion: check if what JabRef believes to be a string, i.e., content between two # is a valid string label, if not make the user sort it out,

lenhard:
Hm, I wonder why and where all this escaping and reformatting takes place on save. Didn't we decide to do no automatic reformatting of entries on save except for save actions. Critically, a closing curly bracket is entered `}`, but no corresponding opening bracket (Note: the first opening bracket is just to signal the beginning of the field value). Afterwards there is also a missing closing bracket.

This is not a problem in the parser, which just quits after  `author = {\&}`. This is exactly what it should do since all content afterwards violates the BibTeX syntax. The serialization/escaping is just wrong and writes too few brackets.

oscargus:
I do not fully agree on the details. Assume a field in JabRef which reads `a#b#c`, this should be serialized as `{a} # b # {c}` (just as `#jan#` is serialized as `jan`, i.e. without `{}`). So that part is still OK (apart from the missing space for symmetry after the second #) and that will indeed load correctly, I tried the example in the first post with only characters inbetween the ## and it both saved and loaded OK. Hence, I conclude that the parser is checking for a string label (b or jan or 8211;&), where there are some formal requirements on how it should look.

[...]

Issue #1188, opened 2016-04-13
Warning should be displayed when using illegal/unescaped characters in bibtex fields (e.g., %, #, &)


ajbelle:
JabRef version <3.2 and 2.1> on <Windows 7>

Steps to reproduce:

1. Have \&#8211 (which I believe is html for dash) pulled into a title of a  bib file created with Ver 3.2 
2. Revert back to JabRef Ver2.1 and attempt to save
3. Message stating unescaped # not acceptable in BibTeX and will not save until the non-BibTeX character is removed.

It may not be a fault in BibLaTeX which I have selected, but I report it in case it indicates that the BibTeX save parsing algorithm is no longer working/deleted in ver3.2 and later.

koppor:
`biber` (or `bibtex`) processes the `bib` file without issues?

In this case, it is an issue in JabRef 2.1 (from August 9<sup>th</sup>, 2006), which we will not fix as we focus on JabRef 3.x. :innocent:

ajbelle:
I tested Ver2.1 because that is a version I have left installed. My though was "WHY DOES AN OLDER VERSION PICK UP A BibLaTeX ERROR" while the current one doesnt! The older version isn't the problem, but it is IDENTIFIYING AN ERROR in the bib file accepted by the current version. I am not expert, but assume it may cause issues if you attempt to use the entry in a paper citation. Lots of features are being removed, and I wondered if parsing the bib file for errors was one that should be retained. Maybe BibLateX handles non-BibTeX characters, so the check has been removed from the current versions. I hope someone who knows can identify if this is a problem version 3.2 and JabRef_windows-x64_3_3dev--snapshot--2016-04-05--fast-search--e0380b7, or an intetional removal of code.

koppor:
@ajbelle Please write me a personal email stating the features you miss. I personally fight for keeping all available issues and support everyone wanting to add an issue. The only thing I currently see being removed and really affecting someone is #496. - I know that there are other things being removed listed at https://github.com/JabRef/jabref/blob/master/CHANGELOG.md, but does anything affect you? Everything else seem to be issues being raised, because other issues (affecting other users). 

I made a minimal example:

```
\documentclass{scrartcl}

\usepackage{filecontents}
\begin{filecontents}{\jobname.bib}
@misc{A01,
  author = {Author, A.},
  year = {2001},
  title = {The &#8211 thing},
}
\end{filecontents}

\begin{document}

\cite[1]{A01}

\bibliography{\jobname}
[...]

PR #1217, opened 2016-04-16
Added integrity checker for odd number of #

oscargus:
Will sometimes help for #1212 and #1188

### DIFF
--- a/src/main/java/net/sf/jabref/logic/integrity/IntegrityCheck.java
+++ b/src/main/java/net/sf/jabref/logic/integrity/IntegrityCheck.java
@@ -1,5 +1,17 @@
 package net.sf.jabref.logic.integrity;
 
+import java.io.File;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.function.Predicate;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
 import net.sf.jabref.BibDatabaseContext;
 import net.sf.jabref.Globals;
 import net.sf.jabref.bibtex.FieldProperties;
@@ -9,17 +21,6 @@ import net.sf.jabref.logic.util.io.FileUtil;
 import net.sf.jabref.model.entry.BibEntry;
 import net.sf.jabref.model.entry.FileField;
 import net.sf.jabref.model.entry.ParsedFileField;
-
-import java.io.File;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-import java.util.Objects;
-import java.util.Optional;
-import java.util.function.Predicate;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
-import java.util.stream.Collectors;
 
 public class IntegrityCheck {
 
@@ -60,7 +61,7 @@ public class IntegrityCheck {
         result.addAll(new TypeChecker().check(entry));
         result.addAll(new AbbreviationChecker("journal").check(entry));
         result.addAll(new AbbreviationChecker("booktitle").check(entry));
-        result.addAll(new HashChecker().check(entry));
+        result.addAll(new BibStringChecker().check(entry));
 
         return result;
     }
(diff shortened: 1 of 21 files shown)


src/main/java/net/sf/jabref/logic/integrity/IntegrityCheck.java
```diff
@@ -10,7 +11,11 @@
 import net.sf.jabref.model.entry.ParsedFileField;
 
 import java.io.File;
-import java.util.*;
```
tobiasdiez: I think the sorting of the imports differs from https://github.com/JabRef/jabref/issues/1091

src/main/java/net/sf/jabref/logic/integrity/IntegrityCheck.java
```diff
@@ -320,4 +326,32 @@ private BracketChecker(String field) {
...
+        private static final Pattern UNESCAPED_HASH = Pattern.compile("(?<!\\\\)#|^#");
+
+
+        /**
+         * Checks, if there is an even number of unescaped #
+         */
+        @Override
+        public List<IntegrityMessage> check(BibEntry entry) {
+            List<IntegrityMessage> results = new ArrayList<>();
+            for (String fieldName : entry.getFieldNames()) {
```
tobiasdiez: I think there is a method BibEntry.getFields which returns the map.

src/main/java/net/sf/jabref/logic/integrity/IntegrityCheck.java
```diff
@@ -320,4 +326,32 @@ private BracketChecker(String field) {
         }
     }
 
+    private static class HashChecker implements Checker {
```
tobiasdiez: I would refer to # as the number sign, because "hash" reminds me too much of the hash code. Maybe use some more functional name like `BibStringChecker`

tobiasdiez:
LGTM :+1:
I think we slowly approach the point where it might make sense to extract the checker as separate classes. @simonharrer what do you think?

simonharrer:
Hm, yes, would make sense @tobiasdiez . I would merge this in with the current structure but for the next checker we should create separate classes.

src/main/java/net/sf/jabref/logic/integrity/IntegrityCheck.java
oscargus: I tried to find it, but either field names or field content as far as I can
tell. Would clearly be more convenient/cleaner.

tobiasdiez: If it not yet exists, then just create it.

src/main/java/net/sf/jabref/logic/integrity/IntegrityCheck.java
oscargus: Maybe. However, in this context "number sign" seems incorrect. HashMarkChecker?

src/main/java/net/sf/jabref/logic/integrity/IntegrityCheck.java
oscargus: I use Eclipse, so it should behave as we want it to if nothing else... I changed it manually though.

src/main/java/net/sf/jabref/logic/integrity/IntegrityCheck.java
oscargus: OK. I assumed that there was a reason that it didn't exist, but I created it now.

src/main/java/net/sf/jabref/logic/integrity/IntegrityCheck.java
oscargus: Now I read and understood the complete comment. BibStringChecker is good.

src/main/java/net/sf/jabref/logic/integrity/IntegrityCheck.java
tobiasdiez: The latest comment by @koppor in #1091 does not reflect the correct settings for Eclipse. See the comment by Simon for IntelliJ.

oscargus: OK! I didn't follow it that much, but got the impression that those were the correct settings (I use another version btw, so slightly different look).

simonharrer:
Go ahead an merge this in.

PR #1218, opened 2016-04-16
Use Path methods from nio in FindUnlinkeFiles

Siedlerchr:
May also avoid problems with whitespaces in Directories.
Remove unnecessary checks
Some Reformatting and organizing imports
Added Debug Logging for "ignored" Exception, better than ignoring it.

### DIFF
--- a/src/main/java/net/sf/jabref/gui/FindUnlinkedFilesDialog.java
+++ b/src/main/java/net/sf/jabref/gui/FindUnlinkedFilesDialog.java
@@ -36,8 +36,18 @@ import java.awt.event.WindowEvent;
 import java.io.File;
 import java.io.FileFilter;
 import java.io.IOException;
-import java.util.*;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Enumeration;
+import java.util.Iterator;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Vector;
 import java.util.concurrent.atomic.AtomicBoolean;
+
 import javax.swing.AbstractAction;
 import javax.swing.Action;
 import javax.swing.BorderFactory;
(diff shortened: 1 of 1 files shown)


src/main/java/net/sf/jabref/gui/FindUnlinkedFilesDialog.java
```diff
@@ -337,15 +354,13 @@ private void storeLastSelectedDirectory(File lastSelectedDir) {
...
+    private Path loadLastSelectedDirectory() {
         String workingDirectory = Globals.prefs.get(FindUnlinkedFilesDialog.GLOBAL_PREFS_WORKING_DIRECTORY_KEY);
         if (workingDirectory == null) {
             workingDirectory = Globals.prefs.get(JabRefPreferences.WORKING_DIRECTORY);
         }
-        lastSelectedDirectory = new File(workingDirectory);
-        if (lastSelectedDirectory.isFile()) {
-            lastSelectedDirectory = lastSelectedDirectory.getParentFile();
-        }
+        lastSelectedDirectory = Paths.get(workingDirectory);
```
simonharrer: The condition has been removed which may cause trouble. I think using the folder if a file has been deleted should still be done.

src/main/java/net/sf/jabref/gui/FindUnlinkedFilesDialog.java
```diff
@@ -223,7 +239,7 @@ private void restoreSizeOfDialog() {
                 String[] dim = store.split(";");
                 dimension = new Dimension(Integer.valueOf(dim[0]), Integer.valueOf(dim[1]));
             } catch (NumberFormatException ignored) {
-                // Ignored
+                LOGGER.debug("RestoreSizeDialog Ecxception ", ignored);
```
oscargus: Exception spelling

src/main/java/net/sf/jabref/gui/FindUnlinkedFilesDialog.java
```diff
@@ -495,6 +497,10 @@ private void startSearch() {
         if (Files.notExists(directory)) {
             directory = Paths.get(System.getProperty("user.dir"));
         }
+        if (!Files.isDirectory(directory)) {
```
Siedlerchr: @simonharrer  The thing is about the Search Directory, not individual files
If accidentally a file (e.g.  xyz.pdf)  is selected or stored in the settings, the parent (e..g the directory) of this file is used.
And the FileChooser dialog only allows to select a directory, not a single file (not visible).
I tested all kind of cases and also when the folder is deleted etc.
That is also why I decided to implement the checks here, when the scan Button is clicked

PR #1224, opened 2016-04-18
Remove duplicated BibteXMLImporter

bruehldev:
A problem on windows occured with creating the same files BibteXMLImporter and BibTeXMLImporter. @zellerdev renamed the file but it wasen't deleted. The file with the capital T is the newer one.


bartsch-dev:
To be more precise, `BibteXMLImporter` (the one with the small `t`) was added again in 4354e8d1c330f14f69228e639d4e2556b42101f0.

tobiasdiez:
Thanks for cleaning up my mess :smile_cat:

Siedlerchr:
The problem on windows is the case insensitivity regarding files...

PR #1227, opened 2016-04-18
Removed some unused fields and parameters

oscargus:

### DIFF
--- a/src/main/java/net/sf/jabref/gui/groups/GroupTreeNodeViewModel.java
+++ b/src/main/java/net/sf/jabref/gui/groups/GroupTreeNodeViewModel.java
@@ -142,7 +142,7 @@ public class GroupTreeNodeViewModel implements Transferable, TreeNode {
     @Override
     public Enumeration children() {
         Iterable<GroupTreeNode> children = node.getChildren();
-        return new Enumeration() {
+        return new Enumeration<Object>() {
 
             @Override
             public boolean hasMoreElements() {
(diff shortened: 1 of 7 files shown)


src/main/java/net/sf/jabref/gui/groups/GroupTreeNodeViewModel.java
```diff
@@ -142,7 +142,7 @@ public boolean isLeaf() {
     @Override
     public Enumeration children() {
         Iterable<GroupTreeNode> children = node.getChildren();
-        return new Enumeration() {
+        return new Enumeration<Object>() {
```
simonharrer: The type is GroupTreeNodeViewModel

oscargus: OK! I just let Eclipse decide. :-)

src/main/java/net/sf/jabref/gui/groups/UndoableMoveGroup.java
```diff
@@ -39,8 +38,7 @@
 
 
     public UndoableMoveGroup(GroupSelector groupSelector, MoveGroupChange moveChange) {
-        this.groupSelector = Objects.requireNonNull(groupSelector);
-
+        Objects.requireNonNull(groupSelector);
         Objects.requireNonNull(moveChange);
```
simonharrer: In that case, I would move the requireNonNull check to the first usage as a the requireNonNull also returns the passed object.

PR #1228, opened 2016-04-18
Removed unused classes

oscargus:
These classes are not referenced, so I do not see any obvious use of these. Anyone?


simonharrer:
:+1:

PR #1229, opened 2016-04-18
Fix #1091 apply and ensure correct import sort order

### LINKED ISSUE #1091: Structure of imports

simonharrer:
See #1091


oscargus:
:+1:

stefan-kolb:
LGTM :+1:

PR #1230, opened 2016-04-18
Removed some issues and encapsulated some fields

oscargus:
* Encapsulated fields and extracted methods in BasePanel
* Some minor code cleanups

### DIFF
--- a/src/main/java/net/sf/jabref/exporter/SaveDatabaseAction.java
+++ b/src/main/java/net/sf/jabref/exporter/SaveDatabaseAction.java
@@ -140,7 +140,7 @@ public class SaveDatabaseAction extends AbstractWorker {
             }
             panel.setSaving(false);
             if (success) {
-                panel.undoManager.markUnchanged();
+                panel.getUndoManager().markUnchanged();
 
                 if (!AutoSaveManager.deleteAutoSaveFile(panel)) {
                     //System.out.println("Deletion of autosave file failed");
@@ -191,10 +191,10 @@ public class SaveDatabaseAction extends AbstractWorker {
             if (ex.specificEntry()) {
                 // Error occured during processing of
                 // be. Highlight it:
-                int row = panel.mainTable.findEntry(ex.getEntry());
+                int row = panel.getMainTable().findEntry(ex.getEntry());
                 int topShow = Math.max(0, row - 3);
-                panel.mainTable.setRowSelectionInterval(row, row);
-                panel.mainTable.scrollTo(topShow);
+                panel.getMainTable().setRowSelectionInterval(row, row);
+                panel.getMainTable().scrollTo(topShow);
                 panel.showEntry(ex.getEntry());
             } else {
                 LOGGER.error("Problem saving file", ex);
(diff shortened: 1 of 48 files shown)


src/main/java/net/sf/jabref/gui/BasePanel.java
```diff
@@ -718,143 +626,19 @@ public void update() {
...
-                        // content in bibtex field is not readable
-                        new SearchAndOpenFile(entry, BasePanel.this).searchAndOpen();
-                        return;
-                    }
-                    FileListEntry flEntry = tableModel.getEntry(0);
-                    ExternalFileMenuItem item = new ExternalFileMenuItem(frame(), entry, "", flEntry.link,
-                            flEntry.type.get().getIcon(), bibDatabaseContext, flEntry.type);
-                    item.openLink();
-                }));
+        actions.put(Actions.OPEN_EXTERNAL_FILE, (BaseAction) () -> openExternalFile());
```
simonharrer: instead of extracted methods, I would prefer separate classes.

src/main/java/net/sf/jabref/exporter/SaveDatabaseAction.java
```diff
@@ -358,25 +358,23 @@ public void saveAs() throws Throwable {
             }
         }
 
-        if (f != null) {
```
simonharrer: Why can this check be removed?

src/main/java/net/sf/jabref/exporter/SaveDatabaseAction.java
```diff
@@ -358,25 +358,23 @@ public void saveAs() throws Throwable {
...
-            }
-            // Register so we get notifications about outside changes to the file.
-            try {
-                panel.setFileMonitorHandle(Globals.fileUpdateMonitor.addUpdateListener(panel,
-                        panel.getBibDatabaseContext().getDatabaseFile()));
-            } catch (IOException ex) {
-                LOGGER.error("Problem registering file change notifications", ex);
-            }
-            frame.getFileHistory().newFile(panel.getBibDatabaseContext().getDatabaseFile().getPath());
+        File oldFile = panel.getBibDatabaseContext().getDatabaseFile();
```
Siedlerchr: I just saw there is a throwable in thrown in saveAs() (some lines above, can't mark it here directly)
Can you please check if this is necessary or if it can be replaced by some more specific exception type?

src/main/java/net/sf/jabref/exporter/SaveDatabaseAction.java
```diff
@@ -358,25 +358,23 @@ public void saveAs() throws Throwable {
...
+        panel.getBibDatabaseContext().setDatabaseFile(f);
+        Globals.prefs.put(JabRefPreferences.WORKING_DIRECTORY, f.getParent());
```
Siedlerchr: Try using the new nio Path methods.
You could see my latest commit  #1218 as an example.

src/main/java/net/sf/jabref/exporter/SaveDatabaseAction.java
oscargus: The loop above will only exit when file != null.

src/main/java/net/sf/jabref/exporter/SaveDatabaseAction.java
oscargus: Oh, I prefer to not go into the Throwable-business... It seems to be easily replaceable, but there's a clear risk that eventually something will break...

simonharrer: You can always change a Throwable to an Exception, as Errors cannot be handled by JabRef.

tobiasdiez:
What is the status here?

simonharrer:
@oscargus *push* :)

oscargus:
Rebased. Still methods instead of classes. No Path. So basically some encapsulations of variables and replacing long lambdas with methods.

PR #1231, opened 2016-04-18
Fixes

oscargus:
Some random cleanups


src/main/java/net/sf/jabref/gui/preftabs/FontSelectorDialog.java
```diff
@@ -166,13 +166,13 @@ public FontSelectorDialog(Component comp, Font font) {
 
         familyList = new JList<>(getFontList());
 
-        JPanel familyPanel = createTextFieldAndListPanel(Localization.lang("Font Family"), familyField, familyList);
```
tobiasdiez: I don't understand the changes in this file.

oscargus: The added underscore was just an ugly way to do search and replace in both Java and properties-files. (The translator replaces spaces with underscores, finds the correct property and then replaces the underscores with spaces, so adding it initially doesn't matter.)

The translation changes was primarily finding better wording ("FontSelector" is probably a worse tile compared to "Font selection", most strings do not use title case, so "Font family" instead of "Font Family" etc).

Removed exceptions were not thrown by the methods.

UPPER/LOWER made sense in the old dialog, now LEFT/RIGHT is more intuitive.

Variable names flds, s1, s2 were not that descriptive.

Lambdas

Commented out code was removed.

src/main/java/net/sf/jabref/gui/StringDialog.java
```diff
@@ -58,7 +58,7 @@
     // The action concerned with closing the window.
     private final CloseAction closeAction = new CloseAction();
 
-    public static final String STRINGS_TITLE = Localization.lang("Strings for database");
+    private static final String STRINGS_TITLE = Localization.lang("Strings for database");
```
oscargus: String only used in this class.

src/main/java/net/sf/jabref/gui/preftabs/FontSelectorDialog.java
```diff
@@ -100,7 +100,7 @@ Portions copyright (C) 1999 Jason Ginchereau
  * Added more translations, and started work on a Replace string feature
  *
  * Revision 1.4  2004/02/17 09:14:02  mortenalver
- * Similar update in FontSelector preview.
+ * Similar update in Font_selection preview.
```
oscargus: Bad case of search and replace...

src/main/java/net/sf/jabref/gui/preftabs/LabelPatternPrefTab.java
```diff
@@ -50,8 +50,8 @@
     private final JRadioButton letterStartB = new JRadioButton(Localization.lang("Ensure unique keys using letters (b, c, ...)"));
     private final JRadioButton alwaysAddLetter = new JRadioButton(Localization.lang("Always add letter (a, b, ...) to generated keys"));
 
-    private final JTextField KeyPatternRegex = new JTextField(20);
-    private final JTextField KeyPatternReplacement = new JTextField(20);
+    private final JTextField keyPatternRegex = new JTextField(20);
+    private final JTextField keyPatternReplacement = new JTextField(20);
```
oscargus: fields should not start with capital letter.

src/main/resources/l10n/JabRef_fr.properties
```diff
@@ -314,11 +314,11 @@ First_select_the_entries_you_want_keys_to_be_generated_for.=Commencez_par_sélec
 Fit_table_horizontally_on_screen=Ajuster_horizontalement_la_table_à_l'écran
 Float=Flottante
 Float_marked_entries=Entrées_marquées_flottantes
-Font_Family=Famille_de_police
-Font_Preview=Prévisualisation_de_la_police
-Font_Size=Taille_de_police
-Font_Style=Style_de_police
```
oscargus: Most translations doesn't follow the capitalization anyway.

src/main/resources/l10n/JabRef_de.properties
```diff
@@ -609,15 +609,15 @@ Float=Oben_einsortieren
...
+Font_preview=Vorschau
 
-Font_Size=Schriftgröße
+Font_size=Schriftgröße
 
-Font_Style=Schriftstil
+Font_style=Schriftstil
 
-FontSelector=Schriften_wählen
+Font_selection=Schriften_wählen
```
oscargus: SchriftenWähler? (If it should have followed the old naming.)

src/main/java/net/sf/jabref/gui/preftabs/FontSelectorDialog.java
tobiasdiez: Ok, so can you please remove the underscore again?

simonharrer:
@oscargus LGTM

Can you fix the things you mentioned in your own comments and just merge this in?

oscargus:
I think this should be OK now.

Issue #1233, opened 2016-04-19
Groups sidebar element does not scale automatically with window height


tillschaefer:
JabRef version 3.2 on Gentoo Linux

The sidebar element "Groups" has a fixed height. This is problematic for larger displays as it wastes a lot of space and one looses the overview over the categories (see screenshot). It should scale with the remaining vertical space available in the sidebar. 

As this dynamic scaling is sometimes a bit problematic using multiple sidebar elements (form a programmer perspective) it would also be helpful (as an alternative solution), if the sidebar height could be configured.

![groups](https://cloud.githubusercontent.com/assets/3009428/14637866/716f2a50-0634-11e6-8d1f-3d56bb008f84.png)

matthiasgeiger:
Thanks for your report.

It is already possible to increase/decrease the size of the Group list manually by using the buttons beneath the list of groups next to "settings".

tillschaefer:
wow, thank you for the fast answer.

The settings are still somewhat suboptimal, because the buttons move with each resize operation.

tillschaefer:
I will leave this ticket open as a feature request for automatic scaling

matthiasgeiger:
Yes, we are also not happy with the current solution and the used icons :wink:

However, as you pointed out automatic scaling might be troublesome if more than one sidebar is used - but as the search sidebar was moved to the top of each tab this should not be a big issue.

tobiasdiez:
It is implemented in the latest development version, which should be available in a few minutes under http://builds.jabref.org/master/. Thanks for suggesting this improvement.

tillschaefer:
wow again for the speed :-) thank you, too.

simonharrer:
JabRef is powered by donations ;-)

tillschaefer:
done :-)

PR #1236, opened 2016-04-19
Group pane automatically takes up remaning space in side pane

tobiasdiez:
Implements https://github.com/JabRef/jabref/issues/1233. Also the buttons at the bottom to expand or shrink the group pane are now superfluous, so I removed them.


simonharrer:
Can you provide a screenshot? Are there any issues with the other side pane elements, e.g., Fetchers?

tobiasdiez:
Yes and no :smile: 
untitled

matthiasgeiger:
This looks rather strange at the bottom over the status bar... Is there some overlap?

simonharrer:
Hm, what I see: it could happen that the window size render the group function unusable. 

1. Shrink window so that there is only space for the fetcher
2. Groups pane shrinks as well, showing basically no entry anymore

Is this the case?

tobiasdiez:
@matthiasgeiger Yes, the status bar has no border and thus there is no separation to the side pane.
@simonharrer No, the groups pane does not shrink below a certain size (two groups displayed). If not enough space remains to show all groups, then a scrollbar is displayed.

simonharrer:
LGTM

PR #1238, opened 2016-04-19
Added integrity checker to detect HTML encoded characters in fields

oscargus:
Based on discussions in #1188


simonharrer:
:+1: LGTM

simonharrer:
@oscargus in the url field, your detection now finds html encoded characters. Maybe we should exclude the check for the url field?

simonharrer:
@oscargus same holds true for the odd number of unescaped '#' warning, which also occurs in a url field of mine.

PR #1241, opened 2016-04-19
add test for MsBibExportFormat

tschechlovdev:
In this PR the class MsBibExportFormat will be tested. The logic of this class will be found at MsBibEntry.java and MsBibDatabse.java. I also edited the ADDRESS_PATTERN in MsBibEntry.java so that the regex described in the comment will be matched.


stefan-kolb:
Please fix Codacy issues and reorganize imports. We fixed this to a new format lately, thats why the tests are failing. See https://github.com/JabRef/jabref/issues/1091

src/test/java/net/sf/jabref/exporter/MSBibExportFormatTestFiles.java
```diff
@@ -0,0 +1,104 @@
...
+    public String filename;
+
+    @Rule
+    public TemporaryFolder testFolder = new TemporaryFolder();
+
+
+    @Parameters(name = "{0}")
+    public static Collection<String> fileNames() throws IOException {
+        List<String> files = new ArrayList<>();
+        try (DirectoryStream<Path> stream = Files.newDirectoryStream(Paths.get(PATH_TO_FILE))) {
```
Siedlerchr: I think this is kind of double code:
Your first collect all files and then filter on them, but you are only searching for some specific files:
You could instead use: Stream stream = Files.list(Paths.get) and then apply the filters directly on it or you directly use Files.find(...) to only find the bib files

Have a look here at Listing Files / Finding Files:
http://winterbe.com/posts/2015/03/25/java8-examples-string-number-math-files/

tschechlovdev:
I fixed the codacy issues and addressed the comments.

src/test/java/net/sf/jabref/exporter/MSBibExportFormatTestFiles.java
```diff
@@ -0,0 +1,104 @@
...
+        msBibExportFormat.performExport(databaseContext, tempFilename, charset, entries);
+        try (FileInputStream stream = new FileInputStream(tempFile)) {
+            DocumentBuilder documentBuilder = DocumentBuilderFactory.newInstance().newDocumentBuilder();
+            Document exportedDocument = documentBuilder.parse(stream, tempFilename);
+            exportedDocument.normalizeDocument();
+            Document expectedDocument = documentBuilder.parse(new File(PATH_TO_FILE + xmlFileName));
+            expectedDocument.normalizeDocument();
+            assertTrue(expectedDocument.isEqualNode(exportedDocument));
+        } catch (ParserConfigurationException | SAXException | TransformerFactoryConfigurationError e) {
+            System.out.println(e.getMessage());
```
simonharrer: please just throw the exceptions

src/test/java/net/sf/jabref/exporter/MSBibExportFormatTestFiles.java
```diff
@@ -0,0 +1,104 @@
...
+        }
+        assertNotNull(entries);
+        msBibExportFormat.performExport(databaseContext, tempFilename, charset, entries);
+        try (FileInputStream stream = new FileInputStream(tempFile)) {
+            DocumentBuilder documentBuilder = DocumentBuilderFactory.newInstance().newDocumentBuilder();
+            Document exportedDocument = documentBuilder.parse(stream, tempFilename);
+            exportedDocument.normalizeDocument();
+            Document expectedDocument = documentBuilder.parse(new File(PATH_TO_FILE + xmlFileName));
+            expectedDocument.normalizeDocument();
+            assertTrue(expectedDocument.isEqualNode(exportedDocument));
```
simonharrer: what about trying to import the previously exported document and then just compare the databases? This should ensure that they behave correctly together.

src/test/java/net/sf/jabref/exporter/MsBibExportFormatTest.java
```diff
@@ -0,0 +1,58 @@
...
+    @Test
+    public final void testPerformExportWithNoEntry() throws IOException {
+        entries = Collections.EMPTY_LIST;
+        tempFile = testFolder.newFile();
+        String tempFileName = tempFile.getCanonicalPath();
+        msBibExportFormat.performExport(databaseContext, tempFileName, charset, entries);
+        try (FileInputStream stream = new FileInputStream(tempFile);
+                InputStreamReader reader = new InputStreamReader(stream, charset)) {
+            char[] buffer = new char[512];
+            assertEquals(-1, reader.read(buffer));
```
simonharrer: Why not use Files.readAllLines and just compare the list? This code is very hard to read as it is very technical.

src/test/resources/net/sf/jabref/exporter/MsBibExportFormatTest1.bib
```diff
@@ -0,0 +1,44 @@
+@unpublished{,
+  author =                      {Hanns-Jörg Stoß},
+  month =                       {July},
+  note =                        {a Note},
+  abstract =                    {Es werdenTuring-Maschinen mitk unabhängigen Lese-Schreib-Köpfen auf einem Band undTuring-Maschinen mit je einem Kopf aufk Bändern betrachtet. Es wird gezeigt, daß sich ohne Verändern der Rechenzeit jede Maschine von einem Typ durch eine Maschine vom andern Typ simulieren läßt. Daraus wird gefolgert, daß es für die Rechengeschwindigkeit vonTuring-Maschinen nur auf die Zahl der Köpfe, nicht aber auf deren Verteilung auf einzelne Bänder ankommt.},
```
simonharrer: abstracts are copyright protected ......... this has come up in so many of the stupro PRs, please do communicate such things within your group.

src/test/resources/net/sf/jabref/exporter/MsBibExportFormatTest6.xml
```diff
@@ -0,0 +1,21 @@
...
+<b:Volume>44</b:Volume>
+<b:StandardNumber> MRN: 0937-6429</b:StandardNumber>
+<b:Publisher>Gabler Verlag</b:Publisher>
+<b:JournalName>Wirtschaftsinformatik</b:JournalName>
+<b:Issue>3</b:Issue>
+<b:ThesisType>Master's thesis</b:ThesisType>
+<b:URL>http://dx.doi.org/10.1007/BF03250842</b:URL>
+<b:BIBTEX_KeyWords>software development processes; agile software development environments; time-to-market; Extreme Programming; Crystal methods family; Adaptive Software Development</b:BIBTEX_KeyWords>
+</b:Source>
+</b:Sources>
```
simonharrer: why no indention?

src/test/resources/net/sf/jabref/exporter/MsBibExportFormatTest1.bib
mlep: In the wiki, we have at the moment 2 databases that could be used for testing.
This list could be complemented with custom-made entries fitting the purpose of the tests (for example, the listed BibTeX database does not contains an "abstract" field). This would avoid having to create entries over and over again.

src/test/java/net/sf/jabref/exporter/MSBibExportFormatTestFiles.java
tobiasdiez: I think our importer and exporter are not good enough to handle such roundtrips. However, comparing the files just as plain strings should work without any problems and should reduce the complexity of the test.

src/test/java/net/sf/jabref/exporter/MsBibExportFormatTest.java
```diff
@@ -0,0 +1,58 @@
...
+        charset = Charsets.UTF_8;
+        msBibExportFormat = new MSBibExportFormat();
+    }
+
+    @Test
+    public final void testPerformExportWithNoEntry() throws IOException {
+        entries = Collections.EMPTY_LIST;
+        tempFile = testFolder.newFile();
+        String tempFileName = tempFile.getCanonicalPath();
+        msBibExportFormat.performExport(databaseContext, tempFileName, charset, entries);
```
tobiasdiez: Can you please add a overload to the performExport method in the export ExportFormats which accepts the file as an nio.Path instead of a stringy pathname.

src/test/java/net/sf/jabref/exporter/MsBibExportFormatTest.java
```diff
@@ -0,0 +1,58 @@
...
+        Globals.prefs = JabRefPreferences.getInstance();
+        databaseContext = new BibDatabaseContext(new BibDatabase(), new MetaData());
+        charset = Charsets.UTF_8;
+        msBibExportFormat = new MSBibExportFormat();
+    }
+
+    @Test
+    public final void testPerformExportWithNoEntry() throws IOException {
+        entries = Collections.EMPTY_LIST;
+        tempFile = testFolder.newFile();
```
tobiasdiez: The temp File can be created in the setUp method.

src/test/java/net/sf/jabref/exporter/MsBibExportFormatTest.java
```diff
@@ -0,0 +1,58 @@
...
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+
+import static org.junit.Assert.assertEquals;
+
+public class MsBibExportFormatTest {
+
+    public static BibDatabaseContext databaseContext;
+    public static Charset charset;
+    public List<BibEntry> entries;
```
tobiasdiez: Inline the entries field since it is not used across tests or initialized in the setup method.

src/test/resources/net/sf/jabref/exporter/MsBibExportFormatTest1.bib
tschechlovdev: The problem with using the database is, that when there are more than one entry, the order of the elements in the exported xml file is somehow changing everytime. So I cannot make one xml document for all entries in the database, but have to make a xml file for every entry. So if you're fine with it, I wouldn't use the database here.

tschechlovdev:
I've addressed the comments.

src/main/java/net/sf/jabref/exporter/ExportFormat.java
```diff
@@ -311,6 +312,22 @@ public void performExport(final BibDatabaseContext databaseContext, final String
     }
 
     /**
+     * @param databaseContext the database to export from.
+     * @param file       the file to write the resulting export to (as java.nio.Path)
```
tobiasdiez: No need to copy the javadoc comments when overwriting a method.

src/test/java/net/sf/jabref/exporter/MsBibExportFormatTest.java
```diff
@@ -0,0 +1,77 @@
...
+        Globals.prefs = JabRefPreferences.getInstance();
+        databaseContext = new BibDatabaseContext(new BibDatabase(), new MetaData());
+        charset = Charsets.UTF_8;
+        msBibExportFormat = new MSBibExportFormat();
+        tempFile = testFolder.newFile();
+    }
+
+    @Test
+    public final void testPerformExportWithNoEntry() throws IOException {
+        List<BibEntry> entries = Collections.EMPTY_LIST;
```
tobiasdiez: Use Collections.emtyList()

src/test/java/net/sf/jabref/exporter/MsBibExportFormatTest.java
```diff
@@ -0,0 +1,77 @@
...
+                InputStreamReader reader = new InputStreamReader(stream, StandardCharsets.UTF_8)) {
+            ParserResult result = BibtexParser.parse(reader);
+            BibDatabase database = result.getDatabase();
+            String tempFilename = tempFile.getCanonicalPath();
+            List<BibEntry> entries = database.getEntries();
+            assertNotNull(entries);
+            msBibExportFormat.performExport(databaseContext, tempFile.getPath(), charset, entries);
+            List<String> expected = Files.readAllLines(Paths.get("src/test/resources/net/sf/jabref/exporter/test.xml"));
+            List<String> exported = Files.readAllLines(Paths.get(tempFilename));
+            assertTrue(expected.containsAll(exported) && exported.containsAll(expected)
```
tobiasdiez: Use assertEquals(expected, exported)

tobiasdiez:
LGTM :+1: - just some minor remarks.

src/test/java/net/sf/jabref/exporter/MsBibExportFormatTest.java
tschechlovdev: This won't work here. The order of the entries, that are exported, is not always the same. So just assertEquals(expected, exported) sometimes goes through and sometimes fail.

src/test/java/net/sf/jabref/exporter/MSBibExportFormatTestFiles.java
```diff
@@ -0,0 +1,87 @@
...
+        Globals.prefs = JabRefPreferences.getInstance();
+        databaseContext = new BibDatabaseContext(new BibDatabase(), new MetaData());
+        charset = Charsets.UTF_8;
+        msBibExportFormat = new MSBibExportFormat();
+        tempFile = testFolder.newFile();
+    }
+
+    @Test
+    public final void testPerformExport() throws IOException {
+        xmlFileName = filename.replace(".bib", ".xml");
```
simonharrer: Why is this a field instead of a local variable?

src/test/java/net/sf/jabref/exporter/MsBibExportFormatTest.java
```diff
@@ -0,0 +1,77 @@
...
+    public static BibDatabaseContext databaseContext;
+    public static Charset charset;
+    public File tempFile;
+    public static MSBibExportFormat msBibExportFormat;
+
+    @Rule
+    public TemporaryFolder testFolder = new TemporaryFolder();
+
+
+    @Before
```
simonharrer: everything that you set in methods marked `@Before` should not be stored in static fields. Please check in your other tests as well.

src/test/java/net/sf/jabref/exporter/MsBibExportFormatTest.java
```diff
@@ -0,0 +1,77 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+
+public class MsBibExportFormatTest {
+
+    public static BibDatabaseContext databaseContext;
+    public static Charset charset;
+    public File tempFile;
+    public static MSBibExportFormat msBibExportFormat;
```
simonharrer: Do not mix static and non-static fields here. Group them together and add an empty line between them. Please check in your other tests as well.

src/test/resources/net/sf/jabref/exporter/MsBibExportFormatTest1.bib
```diff
@@ -0,0 +1,44 @@
+@unpublished{,
+  author =                      {Hanns-Jörg Stoß},
+  month =                       {July},
+  note =                        {a Note},
+  abstract = {there is something described about turing machines. They have k bands but still are as strong as turing machines that have just one. This will be proofed.},
```
simonharrer: Abstracts are copyright protected...

src/test/resources/net/sf/jabref/exporter/MsBibExportFormatTest1.xml
```diff
@@ -0,0 +1,67 @@
...
+<b:Broadcaster>broad</b:Broadcaster>
+<b:Station>stat</b:Station>
+<b:Type>type</b:Type>
+<b:PatentNumber>pat</b:PatentNumber>
+<b:Court>court</b:Court>
+<b:Reporter>rep</b:Reporter>
+<b:CaseNumber>case</b:CaseNumber>
+<b:AbbreviatedCaseNumber>abb</b:AbbreviatedCaseNumber>
+<b:BIBTEX_Series>ser</b:BIBTEX_Series>
+<b:BIBTEX_Abstract>there is something described about turing machines. They have k bands but still are as strong as turing machines that have just one. This will be proofed.</b:BIBTEX_Abstract>
```
simonharrer: Abstracts are copyright protected...

src/test/resources/net/sf/jabref/exporter/MsBibExportFormatTest1.bib
tschechlovdev: I've wrote the abstract myself, so I don't think that this should be copyrigthed?

src/test/java/net/sf/jabref/exporter/MSBibExportFormatTestFiles.java
tschechlovdev: I've changed it to a local variable.

src/test/java/net/sf/jabref/exporter/MsBibExportFormatTest.java
tschechlovdev: I've changed them to non-static.

src/test/java/net/sf/jabref/exporter/MsBibExportFormatTest.java
tschechlovdev: done.

src/test/java/net/sf/jabref/exporter/ExportFormatTest.java
```diff
@@ -58,11 +57,7 @@ public void testExportingEmptyDatabaseYieldsEmptyFile() throws Exception {
         File tmpFile = testFolder.newFile();
         String filename = tmpFile.getCanonicalPath();
         exportFormat.performExport(databaseContext, filename, charset, entries);
-        try (FileInputStream stream = new FileInputStream(tmpFile);
-                InputStreamReader reader = new InputStreamReader(stream, charset)) {
-            char[] buffer = new char[512];
-            assertEquals(-1, reader.read(buffer)); // Empty file
-        }
+        assertEquals(Files.readAllLines(tmpFile.toPath()), Collections.EMPTY_LIST);
```
simonharrer: expected is the first parameter
always use Collections.emptyList() instead

src/test/java/net/sf/jabref/exporter/MSBibExportFormatTestFiles.java
```diff
@@ -0,0 +1,88 @@
...
+
+    @Rule
+    public TemporaryFolder testFolder = new TemporaryFolder();
+
+
+    @Parameters(name = "{0}")
+    public static Collection<String> fileNames() throws IOException {
+        List<String> files = new ArrayList<>();
+        try (Stream<Path> stream = Files.list(Paths.get(PATH_TO_FILE))) {
+            files = stream.map(n -> n.getFileName().toString()).filter(n -> n.endsWith(".bib"))
```
simonharrer: replace `files =` with return

src/test/java/net/sf/jabref/exporter/MsBibExportFormatTest.java
simonharrer: Then sort them so that the order is the same.

src/test/resources/net/sf/jabref/exporter/MsBibExportFormatTest1.bib
simonharrer: This does not always hold true. Often, when you publish something, you have to sign a lot of paperwork which may influence your rights to publish parts of it anywhere. Please replace this with dummy text, as we do not want to keep track which abstracts in our test data are OK to be included and which are not.

tschechlovdev: ok I've changed it to a dummy text.

src/test/java/net/sf/jabref/exporter/ExportFormatTest.java
tschechlovdev: done.

src/test/java/net/sf/jabref/exporter/MSBibExportFormatTestFiles.java
tschechlovdev: done

src/test/java/net/sf/jabref/exporter/MsBibExportFormatTest.java
tschechlovdev: done.

tschechlovdev:
Codecov/changes fails, but I don't know why. Does someone know the reason?

PR #1243, opened 2016-04-19
Fixes #1234 Trigger entry update explicitly

### LINKED ISSUE #1234: NPE when getting information from newly found DOI
Latest version. Both Windows and Linux.

Reproduce:
1. Select entry without DOI
2. Click "Look for DOI" (running the Swedish translation...)
3. Click "Get information from DOI"
NPE

Resolved by clicking in the DOI text field. After that, no NPE. However, it seems like one will need to save (or at least do something more) to actually get the additional information when one clicks.

BTW, the "Look for DOI"-functionality is brilliant!

stefan-kolb:
I'm not quite sure if this fixes all issues. The real problem ist the update mechanism of the `StoreFieldAction` imho. Our GUI state is so inconsistent...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ to [sourceforge feature requests](https://sourceforge.net/p/jabref/features/) by
 - Implemented [#1233](https://github.com/JabRef/jabref/issues/1233): Group side pane now takes up all the remaining space
 
 ### Fixed
+- Fixed [#1234](https://github.com/JabRef/jabref/issues/1234): NPE when getting information from retrieved DOI
 
 ### Removed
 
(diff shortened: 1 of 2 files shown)


oscargus:
Worth a try! :+1:

There are probably more field components where something similar is required. I tested looking up DOI and switching to source view and it wasn't updated (not with your fix though), so probably we should add it consistently if it is the solution.

tobiasdiez:
Does setting the doi directly for the entry also update the display? I.e. does `entryEditor.getEntry().setField("doi", result)` instead of JTextComponent.setText works?

stefan-kolb:
Yeah, looks like this works too.

matthiasgeiger:
:+1:

PR #1246, opened 2016-04-19
Fixed #1245

### LINKED ISSUE #1245: Jstyle files - improper handling of quotation marks
JabRef version 3.3 on Windows 10

Steps to reproduce:

1. .Use the Citation file below/ rename to .jstyle
[test.txt](https://github.com/JabRef/jabref/files/226348/test.txt)

2. Have a look at:
BracketBefore=""
BracketAfter=""

They are defined with quoutation marks according to the example at: http://jabref.sourceforge.net/OOPlugin-jabref.php 

2. create an empty LO document and cite some entries
3. The following happens: ""DOE 1998""...

Expected was an entry without "brackets" before and after: DOE 1998
I know, the expected behavior can be triggered by the following code:
BracketBefore=
BracketAfter=

but for other cases quotation marks are needed, leaving me with inconsistent property definitions

oscargus:
This should fix #1245 so that it is possible to set empty strings is jstyle-files as `key=""`

### DIFF
--- a/src/test/java/net/sf/jabref/logic/openoffice/OOBibStyleTest.java
+++ b/src/test/java/net/sf/jabref/logic/openoffice/OOBibStyleTest.java
@@ -494,4 +494,26 @@ public class OOBibStyleTest {
         assertTrue(style1.compareTo(style2) > 0);
         assertFalse(style2.compareTo(style1) > 0);
     }
+
+
+    public void testEmptyStringPropertyAndOxfordComma() throws URISyntaxException, IOException {
+        String fileName = Paths.get(OOBibStyleTest.class.getResource("test.jstyle").toURI()).toString();
+        OOBibStyle style = new OOBibStyle(fileName, mock(JournalAbbreviationRepository.class));
+        Map<BibEntry, BibDatabase> entryDBMap = new HashMap<>();
+        List<BibEntry> entries = new ArrayList<>();
+        BibDatabase database = new BibDatabase();
+
+        BibEntry entry = new BibEntry();
+        entry.setType("article");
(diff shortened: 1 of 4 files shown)


tobiasdiez:
LGTM but can you please add a test for the fix.

oscargus:
Yes, but not right now. Good point!

stefan-kolb:
Could you please add a test for it so we can merge this in :smile: ? We already closed the related issue.

oscargus:
Done! Will merge when I see that the tests pass.

src/test/java/net/sf/jabref/logic/openoffice/OOBibStyleTest.java
```diff
@@ -494,4 +494,26 @@ public void testCompareToNotEqual() throws IOException {
         assertTrue(style1.compareTo(style2) > 0);
         assertFalse(style2.compareTo(style1) > 0);
     }
+
+
+    public void test1245AndOxfordComma() throws URISyntaxException, IOException {
```
stefan-kolb: Dont call it 1234 please :smile:

oscargus: Yeah, I do not really know what it checks. ;-) (But I'll figure something out.)

stefan-kolb: Something with empty strings as values?! maybe the oxford comma is even a separate test case?

oscargus: I was considering a separate test case, but it would (could) look exactly the same, so I thought it made more sense to have both in one.

PR #1250, opened 2016-04-19
Replace general get/put/remove metadata calls with typed versions

tobiasdiez:
Also removed support of legacy user-specific file paths (serialized with a @ symbol). JabRef stopped writing this format over 3 years ago (according to https://sourceforge.net/p/jabref/bugs/1153/).

### DIFF
--- a/src/main/java/net/sf/jabref/JabRefPreferences.java
+++ b/src/main/java/net/sf/jabref/JabRefPreferences.java
@@ -326,8 +326,6 @@ public class JabRefPreferences {
     public static final String ALLOW_FILE_AUTO_OPEN_BROWSE = "allowFileAutoOpenBrowse";
     public static final String CUSTOM_TAB_NAME = "customTabName_";
     public static final String CUSTOM_TAB_FIELDS = "customTabFields_";
-    public static final String USER_FILE_DIR_INDIVIDUAL = "userFileDirIndividual";
-    public static final String USER_FILE_DIR_IND_LEGACY = "userFileDirInd_Legacy";
     public static final String USE_UNIT_FORMATTER_ON_SEARCH = "useUnitFormatterOnSearch";
     public static final String USE_CASE_KEEPER_ON_SEARCH = "useCaseKeeperOnSearch";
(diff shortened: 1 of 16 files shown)


src/main/java/net/sf/jabref/JabRefPreferences.java
```diff
@@ -862,14 +860,14 @@ private JabRefPreferences() {
         defaults.put(USE_CONVERT_TO_EQUATION, Boolean.FALSE);
         defaults.put(USE_CASE_KEEPER_ON_SEARCH, Boolean.TRUE);
         defaults.put(USE_UNIT_FORMATTER_ON_SEARCH, Boolean.TRUE);
+    }
 
+    public String getUser() {
         try {
-            defaults.put(USER_FILE_DIR_IND_LEGACY, Globals.FILE_FIELD + Globals.DIR_SUFFIX + '-' + get(DEFAULT_OWNER) + '@' + InetAddress.getLocalHost().getHostName()); // Legacy setting name - was a bug: @ not allowed inside BibTeX comment text. Retained for backward comp.
-            defaults.put(USER_FILE_DIR_INDIVIDUAL, Globals.FILE_FIELD + Globals.DIR_SUFFIX + '-' + get(DEFAULT_OWNER) + '-' + InetAddress.getLocalHost().getHostName()); // Valid setting name
+            return get(DEFAULT_OWNER) + '-' + InetAddress.getLocalHost().getHostName();
```
simonharrer: Why do you remove the user file dir individual default values?

tobiasdiez: ̀USER_FILE_DIR_INDIVIDUAL` was never a real preference and only used as a key to some other preference, i.e. the code read `pref.get(pref.get(USER_FILE_DIR_INDIVIDUAL))` before and now takes the form `pref.get(pref.getUser())`. Moreover, `USER_FILE_DIR_INDIVIDUAL` was never set and so stayed at the default value.

simonharrer:
LGTM

PR #1265, opened 2016-04-20
Restructure Help Pages 

Siedlerchr:
Help Pages are now all listed in one file. 
See: https://github.com/JabRef/help.jabref.org/issues/11 (I cant' get the crosslinking right)
~~Use given encoding when writing SQL-Files~~ (no longer relevant)
Refactored some DB Exceptions

-  [X] Change in CHANGELOG.md described

@mlep  Have a look at the HelpFiles,java file There are all help pages listed which are atm used. The one with the empty Strings ("") atm don't have a help page associated, but should (at least the fetchers) (except for CONTENTS, as it links to the toc of the help pages.

@oscargus  As you worked on the CLI classes recently, is there a way to link to the help page like in the gui dialogs?
https://github.com/JabRef/help.jabref.org/blob/gh-pages/en/CommandLine.md

PR #1269, opened 2016-04-21
Create tests to fix #473

### LINKED ISSUE #473: Import/Export to external database fails for PostgreSQL in 3.0 of jabref

simonharrer:
- [x] Test which reproduces #473
- [x] fix for #473


simonharrer:
Possible solution: use prepared statements, which should solve this issue automatically. And they should be faster and safer to execute as well.

Siedlerchr:
@simonharrer  Yep, I would support that. And second you could use them for MYSQL as well

simonharrer:
Another issue: the we currently support the export of a JabRef database to an SQL file using the syntax for MySQL or PostgreSQL. But this SQL file does not contain the correct escape logic, and it is very hard to get this info out of JDBC without a database connection. Hence, I would suggest to remove these actions, as I cannot see how they are useful without the correct escape logic.

src/main/java/net/sf/jabref/sql/exporter/DatabaseExporter.java
```diff
@@ -170,36 +155,38 @@ private int populateEntryGroupsTable(GroupTreeNode cursor, int parentID, int cur
...
-                    currentID = populateEntryGroupsTable(child, myID, currentID, out, database_id);
+                    currentID = populateEntryGroupsTable(child, myID, currentID, connection, database_id);
                 }
-                //Unfortunatley, AutoCloseable throws only Exception
-            } catch (Exception e) {
-                LOGGER.warn("Cannot close resource", e);
-            }
 
+            }
+            //Unfortunatley, AutoCloseable throws only Exception
```
stefan-kolb: Remove comment

src/main/java/net/sf/jabref/sql/exporter/DatabaseExporter.java
```diff
@@ -305,28 +290,25 @@ private int populateGroupsTable(GroupTreeNode cursor, int parentID, int currentI
...
+                ResultSet rs = statement.executeQuery(
+                        "SELECT groups_id FROM groups WHERE label='" + cursor.getGroup().getName() + "' AND database_id='"
+                                + database_id + "' AND parent_id='" + parentID + "';")) {
+            // setting values to ID and myID to be used in case of textual SQL
+            // export
+            int myID = currentID;
+            rs.next();
+            myID = rs.getInt("groups_id");
+            for (GroupTreeNode child : cursor.getChildren()) {
+                ++currentID;
```
stefan-kolb: currentID++;

src/main/java/net/sf/jabref/sql/exporter/DatabaseExporter.java
```diff
@@ -305,28 +290,25 @@ private int populateGroupsTable(GroupTreeNode cursor, int parentID, int currentI
...
+                currentID = populateGroupsTable(child, myID, currentID, out, database_id);
             }
-
+            //Unfortunatley, AutoCloseable throws only Exception
```
stefan-kolb: You used SQLException before?!

stefan-kolb:
LGTM :+1:

src/main/java/net/sf/jabref/sql/exporter/DatabaseExporter.java
Siedlerchr: Maybe take a look at what I did in #1265
I replaced all throws Exception with their typed ones

PR #1273, opened 2016-04-21
Improved integrity check dialog

oscargus:
* Table can be sorted
* Editing disabled
* Resize properly if a different menu font size is used
* Extracted to its own class

Also made writing of "BibTeX key" consistent.

### DIFF
--- a/src/main/java/net/sf/jabref/gui/actions/Actions.java
+++ b/src/main/java/net/sf/jabref/gui/actions/Actions.java
@@ -11,7 +11,6 @@ public class Actions {
     public static final String ADD_TO_GROUP = "addToGroup";
     public static final String AUTO_SET_FILE = "autoSetFile";
     public static final String BACK = "back";
-    public static final String CHECK_INTEGRITY = "checkIntegrity";
     public static final String CLEANUP = "Cleanup";
     public static final String COPY = "copy";
     public static final String COPY_KEY = "copyKey";
--- a/src/main/java/net/sf/jabref/gui/actions/IntegrityCheckAction.java
+++ b/src/main/java/net/sf/jabref/gui/actions/IntegrityCheckAction.java
@@ -1,0 +1,85 @@
+package net.sf.jabref.gui.actions;
+
+import java.awt.event.ActionEvent;
+import java.util.List;
+
+import javax.swing.Action;
+import javax.swing.JDialog;
+import javax.swing.JOptionPane;
+import javax.swing.JScrollPane;
+import javax.swing.JTable;
+import javax.swing.ListSelectionModel;
+
+import net.sf.jabref.Globals;
+import net.sf.jabref.JabRefPreferences;
+import net.sf.jabref.gui.JabRefFrame;
+import net.sf.jabref.logic.integrity.IntegrityCheck;
+import net.sf.jabref.logic.integrity.IntegrityMessage;
+import net.sf.jabref.logic.l10n.Localization;
+
+public class IntegrityCheckAction extends MnemonicAwareAction {
+
+    private static final String ELLIPSES = "...";
+
+
+    private final JabRefFrame frame;
+
+
+    public IntegrityCheckAction(JabRefFrame frame) {
+        this.frame = frame;
+        putValue(Action.NAME, Localization.menuTitle("Check integrity") + ELLIPSES);
+    }
+
+    @Override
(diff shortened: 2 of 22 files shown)


src/main/java/net/sf/jabref/gui/actions/Actions.java
```diff
@@ -11,7 +11,6 @@
     public static final String ADD_TO_GROUP = "addToGroup";
     public static final String AUTO_SET_FILE = "autoSetFile";
     public static final String BACK = "back";
-    public static final String CHECK_INTEGRITY = "checkIntegrity";
```
simonharrer: Please do not remove the action here. Use it as it was used before.

oscargus: The thing is that it only makes sense if the new Action inherits BasicAction, which is a private class in JabRefFrame. Then it just is stored in a field. Since there seems to be little reason storing it in an unused field in the new Action and it was only used there, I didn't see any point in keeping the String.

oscargus: In GeneralAction, the idea is to run a command based on this (which you probably are aware of). However, the earlier implementation didn't use that as actionPerformed was overridden, so it didn't have any effect at all.

src/main/java/net/sf/jabref/gui/actions/IntegrityCheckAction.java
```diff
@@ -0,0 +1,85 @@
...
+        List<IntegrityMessage> messages = check.checkBibtexDatabase();
+
+        if (messages.isEmpty()) {
+            JOptionPane.showMessageDialog(frame.getCurrentBasePanel(), Localization.lang("No problems found."));
+        } else {
+            // prepare data model
+            Object[][] model = new Object[messages.size()][3];
+            int i = 0;
+            for (IntegrityMessage message : messages) {
+                model[i][0] = message.getEntry().getCiteKey();
```
tobiasdiez: I would prefer if the entry is used as a reference and not its cite key (in case of duplicate cite keys...)

src/main/java/net/sf/jabref/gui/actions/IntegrityCheckAction.java
```diff
@@ -0,0 +1,85 @@
...
+                model[i][1] = message.getFieldName();
+                model[i][2] = message.getMessage();
+                i++;
+            }
+
+            // construct view
+            JTable table = new JTable(model,
```
tobiasdiez: I would use a dedicated table model here, see http://stackoverflow.com/questions/7378013/connect-a-list-of-objects-to-a-jtable

Siedlerchr:
Please add a screenshot for the changes.

src/main/java/net/sf/jabref/gui/actions/IntegrityCheckAction.java
simonharrer: I think a simple model as this suffices in that case. If it gets bigger, we can switch a dedicated table model class.

oscargus:
Well, the visual change isn't much. What you can do to try to see the point
is to increase the menu font size (significantly) in JabRef and then invoke
the old integrity check. With the current PR it looks OK then, so if
anything I should add a screen shot what it looks like without this PR. ;-)

src/main/java/net/sf/jabref/gui/actions/IntegrityCheckAction.java
oscargus: This would require a dedicated table model (I think) and currently it seems like a bit of overkill. But, yes, it would make sense to have some reference to the entry. The method called will not open the entry if there are duplicate keys though.

oscargus:
Without this PR
capture2

With this PR
capture

However, I noted that this seems to be a Windows problem as my Linux machine at work doesn't behave like this. There, the table font size is not increased when increasing the menu and label size (in fact, the menu size is not increased, only the label size). So maybe one should bring some more clarity into this.

oscargus:
The same thing happens when viewing the preferences by the way. Once a good solution is found, the same should be applied there.

simonharrer:
LGTM

Please rebase and merge it

tobiasdiez:
Yes looks good.

simonharrer:
Please rebase and merge @oscargus :)

Issue #629, opened 2016-01-03
Serialization of groups in entry


tobiasdiez:
Currently the manual groups are written to the bib file as a jabref comment at the end of the file. I was thinking about changing this to an inline serialization similar to the keywords fields. That is, as `groups={mygroup1, mygroup2}` in the entry.

This would have the following advantages:
- One could easily see the groups an entry belongs to and could change them by hand. Related [SF 868](https://sourceforge.net/p/jabref/feature-requests/868/) and [SF 750](https://sourceforge.net/p/jabref/feature-requests/750/)
- The group information is accessible in latex (very easy with biblatex).
- Copying an entry to different db keeps group assignment (if group exists in new db). Related [SF 756](https://sourceforge.net/p/jabref/feature-requests/756/)
- Allows to search for group memberships in the search bar (would solve [SF 681](https://sourceforge.net/p/jabref/feature-requests/681/)).
- The logic for filtering by groups can be completely replaced by the search functionality.
- It would work also for entries without bibkey (ok not a big achievement) and there is no logic required if the bibkey is changed (but the very same code is required for group renames).

But since there were quite a few discussions recently about how to serialize the database I want to asked about your opinions before investigating time into this :).

koppor:
My first thought: This contradicts #574. I always have sharing in mind. On the one hand, when sending a bibtex entry, I don't want to send internal fields such das `__markedentry`. On the other hand, data such as groups or keywords belong to an entry and are not external to it.

Maybe, we have to add a functionality "copy entry without personal data to the clipboard".

koppor:
Refs #628

koppor:
Refs #628

lenhard:
Implemented by #1276

PR #1276, opened 2016-04-22
Groups inline

tobiasdiez:
Implement https://github.com/JabRef/jabref/issues/629 so explicit groups are now explicitly serialized in the "groups" field of the entry. 
An old bib file is of course migrated to the new format.


lenhard:
Looks good! No objections code-wise.

PR #1277, opened 2016-04-22
Fix 1270: autosave is working again

### LINKED ISSUE #1270: Delete temporary files after saving a database
JabRef version 3.3 on Linux 3.16.7-35-default (openSuSE 13.2)

Steps to reproduce:

1. Activate autosave.
2. Autosave files not wriiten into the same directory as the *.bib file but to /tmp.
3. Even after saving, autosave files persist.

tobiasdiez:
Fixed #1270. The auto save file was actually not saved at the desired place and thus the temp files were left over.


Siedlerchr:
Do we have any tests for that?`Would it make sense to add a test for thatß

tobiasdiez:
No there are no tests in place and I have no idea how to test the class without doing some bigger refactorings.

simonharrer:
LGTM

May be we could write a test for creating the auto save file by extracting a method which does not depend on the UI?

PR #1279, opened 2016-04-22
More tests for DOI cleanup and a small fix

oscargus:
Tests when DOI in URL-field.

Fix so that the FieldChange new value is null instead of "" after removing field contents.


tobiasdiez:
LGTM :+1:

Issue #625, opened 2016-01-02
[Code] Overwrite BibEntry.equals and use it in tests


tobiasdiez:
The `equals` method of a BibEntry should be overwritten and compare two entries based on their id, type and field values. Then the tests for the parser and importers should be rewritten to construct a list of entries and compare this list with the parsed/imported list.
Reasoning: Currently specific fields of imported entries are checked, however it is not verified that these fields are the only ones present in the entry.

koppor:
Refs  #579

PR #1281, opened 2016-04-22
Replace BibtexEntryAssert by overwriting BibEntry.equals method

tobiasdiez:
As proposed in #625 a proper BibEntry.equals is provided in this PR and used for the tests. 

This as a few advantages:
- detect differences which are not visible in the canonical string representation (so far this hit empty fields and different line breaks)
- get more meaningful test-failure messages for lists (instead of "expected 2 got 3" the added entry is displayed)
- Assert.assertEquals just works and no custom written assertEquals method has to be used

I clicked around a bit, but couldn't find any unexpected side effects due to the overwritten equals method. 

If this PR is accepted, then I would inline the BibtexEntryAssert.assertEquals methods.

### DIFF
--- a/src/main/java/net/sf/jabref/gui/BasePanel.java
+++ b/src/main/java/net/sf/jabref/gui/BasePanel.java
@@ -664,7 +664,7 @@ public class BasePanel extends JPanel implements ClipboardOwner, FileUpdateListe
                     for (BibEntry entry : entries) {
                         bes = entry;
                         // Store the old value:
-                        oldvals.put(bes, bes.getField(BibEntry.KEY_FIELD));
+                        oldvals.put(bes, bes.getCiteKey());
                         database.setCiteKeyForEntry(bes, null);
                     }
                 }
@@ -676,7 +676,7 @@ public class BasePanel extends JPanel implements ClipboardOwner, FileUpdateListe
                     bes = entry;
                     LabelPatternUtil.makeLabel(bibDatabaseContext.getMetaData(), database, bes);
                     ce.addEdit(new UndoableKeyChange(database, bes, (String) oldvals.get(bes),
-                            bes.getField(BibEntry.KEY_FIELD)));
+                            bes.getCiteKey()));
                 }
                 ce.end();
                 undoManager.addEdit(ce);
--- a/src/main/java/net/sf/jabref/importer/fileformat/MedlinePlainImporter.java
+++ b/src/main/java/net/sf/jabref/importer/fileformat/MedlinePlainImporter.java
@@ -26,6 +26,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.regex.Pattern;
 
+import net.sf.jabref.Globals;
 import net.sf.jabref.importer.ImportFormatReader;
 import net.sf.jabref.importer.OutputPrinter;
 import net.sf.jabref.model.entry.AuthorList;
@@ -213,7 +214,7 @@ public class MedlinePlainImporter extends ImportFormat {
                     if (oldAb == null) {
                         hm.put("abstract", val);
                     } else {
-                        hm.put("abstract", oldAb + "\n" + val);
+                        hm.put("abstract", oldAb + Globals.NEWLINE + val);
                     }
                 } else if ("DP".equals(lab)) {
                     String[] parts = val.split(" ");
--- a/src/main/java/net/sf/jabref/model/entry/BibEntry.java
+++ b/src/main/java/net/sf/jabref/model/entry/BibEntry.java
@@ -611,4 +611,21 @@ public class BibEntry {
         // TODO Auto-generated method stub
         return fields;
     }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        BibEntry entry = (BibEntry) o;
+        return Objects.equals(type, entry.type) && Objects.equals(fields, entry.fields);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(type, fields);
+    }
 }
--- a/src/main/java/net/sf/jabref/model/entry/BibEntry.java
+++ b/src/main/java/net/sf/jabref/model/entry/BibEntry.java
@@ -309,15 +309,11 @@ public class BibEntry {
     /**
      * Sets a number of fields simultaneously. The given HashMap contains field
      * names as keys, each mapped to the value to set.
-     * WARNING: this method does not notify change listeners, so it should *NOT*
-     * be used for entries that are being displayed in the GUI. Furthermore, it
-     * does not check values for content, so e.g. empty strings will be set as such.
      */
     public void setField(Map<String, String> fields) {
         Objects.requireNonNull(fields, "fields must not be null");
 
-        changed = true;
-        this.fields.putAll(fields);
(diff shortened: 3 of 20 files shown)


stefan-kolb:
Great stuff! Love it (y)

stefan-kolb:
LGTM :+1:

simonharrer:
LGTM

src/main/java/net/sf/jabref/importer/fileformat/MedlinePlainImporter.java
```diff
@@ -213,7 +214,7 @@ else if ("BTI".equals(lab) || "CTI".equals(lab)) {
                     if (oldAb == null) {
                         hm.put("abstract", val);
                     } else {
-                        hm.put("abstract", oldAb + "\n" + val);
+                        hm.put("abstract", oldAb + Globals.NEWLINE + val);
```
koppor: This commit should be places somewhere else, shouldn't it?

src/main/java/net/sf/jabref/model/entry/BibEntry.java
```diff
@@ -330,6 +326,11 @@ public void setField(String name, String value) {
         Objects.requireNonNull(name, "field name must not be null");
         Objects.requireNonNull(value, "field value must not be null");
 
+        if(value.isEmpty()) {
```
koppor: Sure? - Wie wanted to force the caller to use clearField and not this setField method. Therefore Objects.requireNonNull

src/main/java/net/sf/jabref/model/entry/BibEntry.java
```diff
@@ -611,4 +612,21 @@ public void addKeywords(List<String> keywords) {
...
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        BibEntry entry = (BibEntry) o;
+        return Objects.equals(type, entry.type) && Objects.equals(fields, entry.fields);
```
koppor: What about the comments before the entry? I thought, these are also stored in the entry, aren't they?

src/test/java/net/sf/jabref/bibtex/BibtexEntryAssert.java
```diff
@@ -133,21 +132,13 @@ public static void assertEquals(ImportFormat importer, InputStream shouldBeIs, I
...
     /**
-     * Compares to BibTeX entries using their canonical representation
+     * Compares to BibTeX entries
      */
     public static void assertEquals(BibEntry shouldBeEntry, BibEntry entry) {
-        // use the canonical string representation to compare the entries
-        String shouldBeEntryRepresentation = CanonicalBibtexEntry.getCanonicalRepresentation(shouldBeEntry);
-        String entryRepresentation = CanonicalBibtexEntry.getCanonicalRepresentation(entry);
-        Assert.assertEquals(shouldBeEntryRepresentation, entryRepresentation);
+        Assert.assertEquals(shouldBeEntry, entry);
```
koppor: When there are non equal entries, how do I see the differences easily? With the old implementation, I could see the differences in string representations. What about the new way?

src/main/java/net/sf/jabref/gui/BasePanel.java
```diff
@@ -664,7 +664,7 @@ public void run() {
                     for (BibEntry entry : entries) {
                         bes = entry;
                         // Store the old value:
-                        oldvals.put(bes, bes.getField(BibEntry.KEY_FIELD));
+                        oldvals.put(bes, bes.getCiteKey());
```
koppor: This is another change. Sure, that this should go into this PR?

src/main/java/net/sf/jabref/model/entry/BibEntry.java
```diff
@@ -611,4 +612,21 @@ public void addKeywords(List<String> keywords) {
         // TODO Auto-generated method stub
         return fields;
     }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
```
koppor: Why not (`!o instance of BibEntry`)?

src/test/java/net/sf/jabref/bibtex/BibtexEntryAssert.java
koppor: After thinking a bit more, I would keep the other changes, but not do this change here. This eases writing test cases. Possibly, one should add your assertEquals additionally to be 100% sure that both the canoncial representation and th eentry themselves equal.

src/main/java/net/sf/jabref/importer/fileformat/MedlinePlainImporter.java
tobiasdiez: No, every time JabRef writes a newline somewhere it uses Globals.NEWLINE...except here.

src/main/java/net/sf/jabref/model/entry/BibEntry.java
tobiasdiez: I do not see any value in storing empty fields. They are not written anyway. Throwing an exception complicates things as the user might set the text to the empty string in the bibentry editor.

src/test/java/net/sf/jabref/bibtex/BibtexEntryAssert.java
tobiasdiez: This change was the whole point of the PR :smiling_imp: 
If the entries are not equals, then JUnit calls the toString method and says `expected a.ToString but got b.ToString` so you still get a meaningful message in this case.

src/main/java/net/sf/jabref/gui/BasePanel.java
tobiasdiez: Yeah, I got some failing tests due to empty cite keys and while investigating replaced these calls. Its some additional code cleanup and I did not saw any value in opening an extra PR for this commit.

src/main/java/net/sf/jabref/model/entry/BibEntry.java
tobiasdiez: Since I do not want to include subclasses, these should implement a proper equals method themselves.

Siedlerchr: Yep, this way is the right approach. However I would prefer this template (recommend by java Expert Michael Inden in his book "Der Weg zum Java Profi")
````java
@Override
public boolean equals(Object other)
        {
            if (other == null) // Null-Akzeptanz
                return false;

            if (this == other) // Reflexivität 
                return true;

            if (this.getClass() != other.getClass()) // Typgleichheit sicherstellen 
                return false;
final BibEntry = (BibEntry) other;
return .....
````
@koppor  For more info see here: https://www.artima.com/intv/bloch17.html
Equals with instanceof would additionally violate symmetry.

src/test/java/net/sf/jabref/bibtex/BibtexEntryAssert.java
Siedlerchr: @tobiasdiez  Maybe it is useful to override the toString method implementation?

tobiasdiez: At the moment it is 
@Override
    public String toString() {
        return CanonicalBibtexEntry.getCanonicalRepresentation(this);
}
which is probably what we want.

PR #1296, opened 2016-04-25
Improve BibtexEntryAssert

koppor:
This is a follow up of  #1281 with removing obsolete BibtexEntry.assertEquals methods


koppor:
I have no clue where the missed files come from. :innocent:

obraliar:
Main question: Why there is no assertEquals with two comparative parameters of the same type like: `... assertEquals(BibEntry is, BibEntry shouldBe) { ...` ?
I noticed that such a function would be used verfy often.

tobiasdiez:
Because you can just use the assertEquals of JUnits Assert class for this case and for lists too.

Issue #1294, opened 2016-04-25
Small issues regarding the new style selection dialog for LibreOffice 


teertinker:
JabRef version 3.3 on Ubuntu 14.04
![image](https://cloud.githubusercontent.com/assets/8830305/14783017/3c06ad3a-0aed-11e6-9bec-5be4b00dc11d.png)

The new style dialog is beautiful, thank you very much!

What I miss from before is:

a) a file-filter, allowing to show only *.jstyle files in the window on the far right.
b) in case other users store the style files in one folder: that jabref remembers the folder selected the last time and one does not need to go through all the folders again

a further, real minor, issue might be the option to show the file definition.
- if you right click on the left window (choose style), there is the option to "show" the file definition.
-> the content should be refreshed as this click is triggered, because if there are file modifications between adding a new style and refreshing the bibliography, the file shown, is not the one saved on the hard drive and used to compile the list of references.

Siedlerchr:
I will look into that. @oscargus  You can assign this to me

Siedlerchr:
@teertinker  Would you please test it with one of the current master snapshots? If all is okay, then we can close this. http://builds.jabref.org/master/ 
Code has been merged a few days ago.

teertinker:
Checked with current master:
I am not able to open the "choose style" dialog anymore. The following exceptions seems to be raised:

```
Exception in thread "AWT-EventQueue-0" java.lang.NullPointerException
	at net.sf.jabref.logic.openoffice.OOBibStyle.compareTo(OOBibStyle.java:885)
	at net.sf.jabref.logic.openoffice.OOBibStyle.compareTo(OOBibStyle.java:68)
	at ca.odell.glazedlists.impl.sort.ComparableComparator.compare(ComparableComparator.java:27)
	at ca.odell.glazedlists.impl.sort.ComparableComparator.compare(ComparableComparator.java:18)
	at ca.odell.glazedlists.SortedList$ElementComparator.compare(SortedList.java:647)
	at ca.odell.glazedlists.impl.adt.barcode2.SimpleTree.insertIntoSubtreeInSortedOrder(SimpleTree.java:353)
	at ca.odell.glazedlists.impl.adt.barcode2.SimpleTree.addInSortedOrder(SimpleTree.java:324)
	at ca.odell.glazedlists.SortedList.insertByUnsortedNode(SortedList.java:368)
	at ca.odell.glazedlists.SortedList.listChanged(SortedList.java:352)
[...]

PR #1304, opened 2016-04-26
Implemented Jstyle file filter and opening in last dir

Siedlerchr:
Filter filter for jstyle files #1294 
Extended BrowseAction to display file filter
Fixed NPE when selecting invalid jstyle (prevented style dialog from opening)

- [X ] Change in CHANGELOG.md described

### DIFF
--- a/src/main/java/net/sf/jabref/gui/actions/BrowseAction.java
+++ b/src/main/java/net/sf/jabref/gui/actions/BrowseAction.java
@@ -38,33 +38,35 @@ public final class BrowseAction extends AbstractAction {
     private final JTextField comp;
     private final boolean dir;
     private final JComponent focusTarget;
+    private final String extension;
 
     public static BrowseAction buildForDir(JFrame frame, JTextField tc) {
-        return new BrowseAction(frame, tc, true, null);
+        return new BrowseAction(frame, tc, true, null, Globals.NONE);
     }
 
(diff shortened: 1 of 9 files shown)


src/main/java/net/sf/jabref/gui/actions/BrowseAction.java
```diff
@@ -38,33 +38,35 @@
...
     }
 
     public static BrowseAction buildForDir(JTextField tc) {
-        return new BrowseAction(null, tc, true, null);
+        return new BrowseAction(null, tc, true, null, Globals.NONE);
     }
 
     public static BrowseAction buildForFile(JTextField tc) {
-        return new BrowseAction(null, tc, false, null);
+        return new BrowseAction(null, tc, false, null, Globals.NONE);
```
simonharrer: Why Globals.NONE? Why not simply an empty string? Seems to be an unecessary dependency.

simonharrer:
LGTM

src/main/java/net/sf/jabref/gui/actions/BrowseAction.java
Siedlerchr: Yes, you are right. Interestingly the Globals.NONE was used before. I will look into that this evening.

Siedlerchr:
@simonharrer Would you merge this in?

PR #1309, opened 2016-04-27
Minor updates to BibEntry

koppor:
- <s>rename CiteKey to BibTeXKey to be consistent with wording of JabRef users</s>
- reorder methods in BibEntry to follow id, cite key, type, everything else
- add some JavaDoc
- remove obsolete setting of `changed=true`

### DIFF
--- a/src/main/java/net/sf/jabref/model/entry/BibEntry.java
+++ b/src/main/java/net/sf/jabref/model/entry/BibEntry.java
@@ -63,35 +63,90 @@ public class BibEntry {
     private String parsedSerialization;
 
     /*
-    * marks if the complete serialization, which was read from file, should be used.
-    * Is set to false, if parts of the entry change
+     * Marks whether the complete serialization, which was read from file, should be used.
+     *
+     * Is set to false, if parts of the entry change. This causes the entry to be serialized based on the internal state (and not based on the old serialization)
      */
     private boolean changed;
 
+
+    /**
+     * Constructs a new BibEntry. The internal ID is set to IdGenerator.next()
+     */
     public BibEntry() {
         this(IdGenerator.next());
     }
 
+    /**
+     * Constructs a new BibEntry with the given ID and DEFAULT_TYPE
+     *
+     * @param id The ID to be used
+     */
     public BibEntry(String id) {
         this(id, DEFAULT_TYPE);
     }
 
+    /**
+     * Constructs a new BibEntry with the given ID and given type
+     *
+     * @param id The ID to be used
+     * @param type The type to set. May be null or empty. In that case, DEFAULT_TYPE is used.
+     */
     public BibEntry(String id, String type) {
         Objects.requireNonNull(id, "Every BibEntry must have an ID");
 
         this.id = id;
-        changed = true;
         setType(type);
     }
 
     /**
-     * Returns an set containing the names of all fields that are
-     * set for this particular entry.
+     * Sets this entry's ID, provided the database containing it
+     * doesn't veto the change.
      *
-     * @return a set of existing field names
+     * @param id The ID to be used
      */
-    public Set<String> getFieldNames() {
-        return new TreeSet<>(fields.keySet());
+    public void setId(String id) {
+        Objects.requireNonNull(id, "Every BibEntry must have an ID");
+
+        try {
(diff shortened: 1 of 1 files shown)


src/main/java/net/sf/jabref/model/entry/BibEntry.java
```diff
@@ -68,14 +68,26 @@
...
+     */
     public BibEntry(String id) {
         this(id, DEFAULT_TYPE);
     }
 
+    /**
+     * Constructs a new BibEntry with the given ID and given type
+     *
+     * @param type The type to set. May be null or empty. In that case, DEFAULT_TYPE is used.
+     */
```
matthiasgeiger: :+1: for creating JavaDoc - but please add the second parameter, too :wink:

src/main/java/net/sf/jabref/model/entry/BibEntry.java
```diff
@@ -68,14 +68,26 @@
...
+    /**
+     * Constructs a new BibEntry. The internal ID is set to IdGenerator.next()
+     */
     public BibEntry() {
         this(IdGenerator.next());
     }
 
+    /**
+     * Constructs a new BibEntry with the given ID and DEFAULT_TYPE
+     */
```
matthiasgeiger: @param?

matthiasgeiger:
LGTM - apart from the small JavaDoc issues

src/jmh/java/net/sf/jabref/benchmarks/Benchmarks.java
```diff
@@ -44,7 +44,7 @@ public void init() throws IOException, SaveException {
         Random randomizer = new Random();
         for (int i = 0; i < 1000; i++) {
             BibEntry entry = new BibEntry();
-            entry.setCiteKey("id" + i);
+            entry.setBibTeXKey("id" + i);
```
tobiasdiez: I think set/getKey would work too and might be preferable.

tobiasdiez:
LGTM :+1:

src/jmh/java/net/sf/jabref/benchmarks/Benchmarks.java
koppor: I wanted to be explicit to avoid confusion, because someone might think that a key is an id. And keys are used in SQL-based databases, too. "primary key", "foreign key", etc.

simonharrer: Why not introduce classes like BibTeXKey vs. InternalJabRefKey to distinguish them? Both can wrap a String inside.

matthiasgeiger: Are you talking about the general naming of the method or about the concrete usage of `setBibTeXKey(...)` in this class? :confused: 

I think the proposed naming and reasoning of @koppor is fine. ID for internal usage - "BibTeXKey" used as in the UI...
Using additional classes for this seems to be an overkill for me.

stefan-kolb: Hm, what is your problem with getCiteKey or getKey and getID? The description makes it clear what are the differences. Also the entry is a BibEntry, so so why clutter the method name with BibTeX again? It's not a big deal, but I think a crisper method name is nicer here.

koppor: According to [WikiPedia's BibTeX article](https://en.wikipedia.org/wiki/BibTeX), there is a bibtex field `key`, which is ...

> A hidden field used for specifying or overriding the alphabetical order of entries (when the "author" and "editor" fields are missing). Note that this is very different from the key (mentioned just after this list) that is used to cite or cross-reference the entry.

BibTeX key is defined as follows:

> In addition, each entry contains a key (Bibtexkey) that is used to cite or cross-reference the entry. This key is the first item in a BibTeX entry, and is not part of any field

 * Biber offers the `id` field and refers to the bibtex key as "citatoin key" ("Citation key aliases for the main citation key.")
 * [citation key is more common than cite key](http://www.googlefight.com/cite+key-vs-citation+key.php)

Since "citation key" is longer than the old Cite key", I will undo this renaming and keep `CiteKey`.

lenhard:
Same here. Looks good. @koppor: Is this ready to merge or do you want to implement more improvements?

PR #1316, opened 2016-04-28
Remove support of old groups format

tobiasdiez:
Remove support of old groups format (used prior to JabRef version 1.6, more than 10 years ago)

- [x] Tests created for changes (Tests adopted)


oscargus:
Is it correct that the old format can be loaded with versions up to and including 3.3 and then saved in the new format? If then I think we can safely delete this now.

tobiasdiez:
@oscargus Yes, you are right.

src/main/java/net/sf/jabref/logic/groups/AbstractGroup.java
```diff
@@ -61,18 +61,18 @@
      * @throws Exception If an error occured and a group could not be created, e.g.
      *                   due to a malformed regular expression.
      */
-    public static AbstractGroup fromString(String s, int version) throws Exception {
+    public static AbstractGroup fromString(String s) throws Exception {
```
lenhard: Is it possible to throw a more concrete type of `Exception` here?

src/main/java/net/sf/jabref/logic/groups/AllEntriesGroup.java
```diff
@@ -32,22 +32,14 @@ public AllEntriesGroup() {
         super(Localization.lang("All entries"), GroupHierarchyType.INDEPENDENT);
     }
 
-    public static AbstractGroup fromString(String s, int version) throws Exception {
+    public static AbstractGroup fromString(String s) throws Exception {
         if (!s.startsWith(AllEntriesGroup.ID)) {
             throw new Exception(
```
lenhard: In reference to my above comment: We could certainly throw a better type of Exception here. What about an `IllegalArgumentException`?

src/main/java/net/sf/jabref/logic/groups/ExplicitGroup.java
```diff
@@ -40,34 +40,22 @@ public ExplicitGroup(String name, GroupHierarchyType context) {
         super(name, "groups", name, true, false, context);
     }
 
-    public static ExplicitGroup fromString(String s, int version) throws Exception {
+    public static ExplicitGroup fromString(String s) throws Exception {
         if (!s.startsWith(ExplicitGroup.ID)) {
             throw new Exception(
```
lenhard: same here

src/main/java/net/sf/jabref/logic/groups/GroupsParser.java
```diff
@@ -0,0 +1,59 @@
...
+        for (String string : orderedData) {
+            // This allows to read databases that have been modified by, e.g., BibDesk
+            string = string.trim();
+            if (string.isEmpty()) {
+                continue;
+            }
+
+            int spaceIndex = string.indexOf(' ');
+            if (spaceIndex <= 0) {
+                throw new Exception("bad format");
```
lenhard: Even better: There is `java.util.IllegalFormatException`

src/main/java/net/sf/jabref/logic/groups/KeywordGroup.java
```diff
@@ -74,7 +74,7 @@ private void compilePattern() throws IllegalArgumentException {
      * @param s The String representation obtained from
      *          KeywordGroup.toString()
      */
-    public static AbstractGroup fromString(String s, int version) throws Exception {
+    public static AbstractGroup fromString(String s) throws Exception {
```
lenhard: see above

lenhard:
This is a good opportunity to improve exception handling a little bit. I would go for `java.util.IllegalFormatException` in most cases. Apart from that, and the failing checkstyle warning, this looks good to me.

src/main/java/net/sf/jabref/logic/groups/GroupsParser.java
tobiasdiez: I introduced a new checked `ParseException` since the `IllegalFormatException` is "thrown when a format string contains an illegal syntax" according to the documentation (so the format string is illegal and not the string in an ill-formed format).

tobiasdiez:
Ok, I improved the exception handling.

lenhard:
I am unsure whether it might be smarter to turn `ParseException` into a `RuntimeException`. We would still have better typing and would not be forced to declare it everywhere.

But it is really hard to tell which way makes more sense here. I'll throw my hat in for merging this. We can easily change this in the future if we find unchecked exceptions to be more convenient here.

Siedlerchr:
@lenhard  We should avoid throwing RuntimeExceptions at all cost, as those are unchecked and not catachable. They should be only used if the situation is unrecoverable:
http://stackoverflow.com/questions/27578/when-to-choose-checked-and-unchecked-exceptions

lenhard:
@Siedlerchr Sorry, but I can't leave that standing as it is :-)

RuntimeExceptions are also Exceptions, they just do not need to be declared in method signatures and it is not mandatory to write exception handling code for them. Nevertheless, they *can be caught* and dealt with in the same fashion as regular checked exceptions. 

The debate between checked and unchecked exceptions is ages old and there are valid arguments for using both types. Usually one of the two sides is more "cool" at a certain point in time and API designers overuse it (see Java APIs designed at the turning of the millenial). Things are also not as simple as described in that stackoverflow post (and I could argue that the post says we should use unchecked exceptions). One perceived downside of checked exceptions is the following: if you overuse them, then all your code is cluttered with exception handling blocks that you get tired of writing and end up doing abysmal exception handling, such as `catch(Throwable e) {}`, which you so frequently find in JabRef.

Anyway, let's keep the checked exception in this PR.

simonharrer:
So I think we can rebase and merge this in.

CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ to [sourceforge feature requests](https://sourceforge.net/p/jabref/features/) by
 
 ### Removed
 - Removed possibility to export entries/databases to an `.sql` file, as the logic cannot easily use the correct escape logic
+- Removed support of old groups format (used prior to JabRef version 1.6)
```
simonharrer: Could you clarify that we can still read the old format, hence, automatically convert it to the new group format?

tobiasdiez: No this is not true. With this PR exactly this migration code is removed, so groups in 10 years old bib files are not read correctly. But every version between 1.7 and 3.3 converted the old format to the current one.

simonharrer: Ah ok. Can you provide a hint in the changelog that one can still use a JabRef version 3.3 or older to read it and convert the group format to a newer one so that it is usable for 3.4+?

tobiasdiez: Done.

PR #1325, opened 2016-04-29
Added ISSN integrity checker

oscargus:
Detects both that the content looks like an ISSN number and that the ISSN number is valid based on the control digit.

### DIFF
--- a/src/main/java/net/sf/jabref/logic/integrity/ISSNChecker.java
+++ b/src/main/java/net/sf/jabref/logic/integrity/ISSNChecker.java
@@ -1,0 +1,53 @@
+package net.sf.jabref.logic.integrity;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import net.sf.jabref.logic.integrity.IntegrityCheck.Checker;
+import net.sf.jabref.logic.l10n.Localization;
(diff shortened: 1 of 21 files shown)


src/main/java/net/sf/jabref/logic/integrity/ISSNChecker.java
```diff
@@ -0,0 +1,53 @@
...
+    public List<IntegrityMessage> check(BibEntry entry) {
+        if (!entry.hasField("issn")) {
+            return Collections.emptyList();
+        }
+
+        // Check that the ISSN is on the correct form
+        List<IntegrityMessage> result = new ArrayList<>();
+        String issn = entry.getField("issn").trim();
+        Matcher issn_matcher = ISSN_PATTERN.matcher(issn);
+        if (!issn_matcher.matches()) {
```
tobiasdiez: Don't use underscore.

tobiasdiez:
LGTM :+1:

src/main/java/net/sf/jabref/logic/integrity/ISSNChecker.java
oscargus: Bah! Quick hack, so clearly didn't spend much time thinking about those
things... Will fix!

I found an ISSN where the checker claims to give an incorrect control
digit. It happens when there's a zero control (final) digit and it makes
sense that the current code doesn't work then. One more if-statement
needed...

PR #1335, opened 2016-05-01
Improve symmetric diff

tobiasdiez:
Changes are only highlighted. See the doi field in the following screenshot. Previously the change
 was written as "MSP~~msp~~".
untitled
Also fixed some smaller bugs in the symmetric diff (like accidentantly removed whitespace).


koppor:
Although this is not common to MS Word users, it looks more readable. Could you remove the underline property and check if that looks better? - underline is the MS Word thing...

:+1: for the test cases. I would have used package-private, but I think, we discussed that and voted for `public` instead of using no visibility modifier.

tobiasdiez:
Without the underline some changes (especially small one character changes) are hard to spot.

koppor:
Then it's OK for me. :smiley:

simonharrer:
LGTM

Issue #1324, opened 2016-04-29
"Main file directory" should be relative to the bib file


vogler:
If I set "Preferences > File> Main file directory" to "lit", then the default "Link" after "Look up full text document" will be "/Applications/JabRef.app/Contents/java/app/lit/some.pdf" instead of "/path/to/bib/lit/some.pdf"

Version 3.3

Siedlerchr:
I can not really reproduce this in the current master: 1aa214510eee23cc586aad116780e4a875538e00
Please try again with v3.3 or the latest development snapshot from http://builds.jabref.org/master/

My settings are as follows: 
 1.  Open bib databse e.g . `\JabRef\Data\myBibs\test.bib`
 2.  Set Main file directory to "lit"  (so the relative path is `myBibs\lit` )
 3.  **Check "Allow file links relative to bib"** <- Maybe you did not check that?
 4. Auto ink files with names starting with bibtex keys 
 5. Select an entry in the database -> File -> Auto

And now the file is attached using its relative path:
![jabrefrelativefilepath](https://cloud.githubusercontent.com/assets/320228/14927147/867888c8-0e50-11e6-8218-bc5f4593802d.png)

vogler:
Same settings, same behavior with the current snapshot.
Linking files works. The only problem is that the path-prefix in the "Look up full text document" dialog is set to wherever the jar-file is, instead of where the bib-file is. E.g. now it is "/Users/voglerr/Downloads/lit/foo.pdf" instead of "/Users/voglerr/paper/lit/foo.pdf".

Siedlerchr:
@vogler  Ah yes. Now I understand what you mean.  You are referring to the Quality-> Lookup fulltext document dialog.
Yes, I  could reproduce  that behavior. I will look into that. I think I know where to look.

Could someone add bug label and assign it to me?

JabFabi:
Hi, I'm not sure if this is related to the issue you are working on, but as you mentioned the option "Allow file links relative to bib":
I would like to enable this option (as I had in an older version) but whenever I do so, it gets reset to the disabled state after restarting JabRef 3.3.
I had a look into the exported settings XML file and compared it to the old one; the corresponding entry is simply missing and it seems to be missing in the Windows registry as well. Also, I didn't find any entry for this option in the "View settings" dialog.
Let me know if you need further details.

Greetings and thanks for the great job!
Fabi

stefan-kolb:
@JabFabi Please open a new issue for this behavior.

[...]

PR #1336, opened 2016-05-01
WIP: Refactor Open and Save Dialogs

Siedlerchr:
Complete Refactoring of the ugly and propably outdated FileDialogs code
NewFileDialogs will be renamed when the work is done
Atm I want to keep it parallel.

TODO:
- [x] Refactor getMultpleFiles 
- [x] Provide methods for other Open and Save Dialogs 
- [x] Replace OpenFileFilter with FileNameExtensionFilter  http://docs.oracle.com/javase/8/docs/api/javax/swing/filechooser/FileNameExtensionFilter.html
- [x] Refactor BrowseAction to use new the Methods
- [x] Test #1324  again


tobiasdiez:
Good idea to refactor the file dialogs :+1:. Just one small comment so far: could you please make the methods non-static (and probably move the passed frame to the constructor). In this way the code becomes more testable because the file dialog can be mocked away.

Siedlerchr:
@tobiasdiez  Yes, I already came up with a solution for this. I will probably use some kind of FactorPattern to create the dialogs. That reduces the number of constructors/method overloads.

stefan-kolb:
What is the status here? I'd love to see these changes inside JabRef 3.5!

Siedlerchr:
Unfortunately I have not that much time atm. But I will try to see how far I get the next days. Mabye at the weekend I can get some time for it. Atm another project ist due on sunday...;)

Siedlerchr:
Got some time working on it.

src/main/java/net/sf/jabref/gui/FileExtensions.java
```diff
@@ -0,0 +1,34 @@
...
+    private final String[] extension;
+    private final String description;
+
+
+    private FileExtensions(String description, String... extension) {
+        this.extension = extension;
+        this.description = description;
+    }
+
+    public String[] getExtensions() {
```
oscargus: List?

Siedlerchr: Normally I would agree with you, but in this case an Array is better, because a) varargs (...) as parameter is handled as array and  b) the array is then passed again as varags parameter to the FileNameFilterExtension.
A list would have to be converted back to array again to be passed as paramter.
And the decison for varargs was to support  multiple file extension for one type in the enum e.g. `Import_Format("Sample input format",xml,bib,dat)`
http://docs.oracle.com/javase/1.5.0/docs/guide/language/varargs.html

oscargus: OK! Makes sense!

Siedlerchr:
Ready for review.

src/main/java/net/sf/jabref/gui/FileListEntryEditor.java
```diff
@@ -199,6 +202,8 @@ public void changedUpdate(DocumentEvent documentEvent) {
 
         });
```
oscargus: Get rid of print-debugging. :-)

src/main/java/net/sf/jabref/gui/FileListEntryEditor.java
```diff
@@ -210,7 +215,8 @@ public void changedUpdate(DocumentEvent documentEvent) {
             public void windowActivated(WindowEvent event) {
                 if (openBrowseWhenShown && !dontOpenBrowseUntilDisposed) {
                     dontOpenBrowseUntilDisposed = true;
-                    SwingUtilities.invokeLater(() -> browse.actionPerformed(new ActionEvent(browseBut, 0, "")));
+                    System.out.println("Window activated ");
```
oscargus: And here.

src/main/java/net/sf/jabref/external/MoveFileAction.java
```diff
@@ -143,23 +144,16 @@ public void actionPerformed(ActionEvent event) {
...
+                    chosenFile = new NewFileDialogs(frame, file.getPath()).saveNewFile().toString();
+
+
                 }
-                if (chosenFile == null) {
+                if (chosenFile.isEmpty()) {
                     return; // canceled
                 }
                 newFile = new File(chosenFile);
-                // Check if the file already exists:
```
oscargus: Now files are always overwritten?

src/main/java/net/sf/jabref/gui/actions/BrowseAction.java
```diff
@@ -77,19 +78,21 @@ public void actionPerformed(ActionEvent e) {
...
     }
 
     private String askUser() {
-        if (dir) {
-            return FileDialogs.getNewDir(frame, new File(comp.getText()), extension,
-                    JFileChooser.OPEN_DIALOG, false);
+        if (dirsOnly) {
+            String file = new NewFileDialogs(frame, comp.getText()).dirsOnly().withExtension(extensions)
+                    .getSelectedFile().toString();
+            System.out.println("Browse Action DIR " + file);
```
oscargus: Here

src/main/java/net/sf/jabref/gui/actions/BrowseAction.java
```diff
@@ -77,19 +78,21 @@ public void actionPerformed(ActionEvent e) {
...
+            return file;
         } else {
-            return FileDialogs.getNewFile(frame, new File(comp.getText()), extension,
-                    JFileChooser.OPEN_DIALOG, false);
+            String file = new NewFileDialogs(frame, comp.getText()).withExtension(extensions).getSelectedFile()
+                    .toString();
+            System.out.println("Browse Action file " + file);
```
oscargus: And here

src/main/java/net/sf/jabref/logic/integrity/ISSNChecker.java
```diff
@@ -3,6 +3,7 @@
 import java.util.Collections;
 import java.util.List;
 
+
```
oscargus: remove

src/main/java/net/sf/jabref/gui/FileExtensions.java
```diff
@@ -0,0 +1,61 @@
...
+            "Pubmed file", "fcgi"),
+    RIS(
+            "Ris file ", "ris"),
+    ENDNOTE(
+            "Endnote/Refer file", "ref"),
+
+    JSTYLE(
+            "Style file", "jstyle"),
+    LAYOUT(
+            "Custom Layout File", "layout"),
```
oscargus: Custom layout file

src/main/java/net/sf/jabref/gui/FileExtensions.java
```diff
@@ -0,0 +1,61 @@
...
+    XML(
+            "XML file", "xml"),
+    TERMS(
```
oscargus: terns

src/main/java/net/sf/jabref/gui/FileExtensions.java
```diff
@@ -0,0 +1,61 @@
+package net.sf.jabref.gui;
+
+public enum FileExtensions {
+
+    //importannt: No dot before the extension!
+    BIBTEX_DB(
+            "BibTeX Database", "bib"),
```
oscargus: database

src/main/java/net/sf/jabref/gui/FileExtensions.java
```diff
@@ -0,0 +1,61 @@
...
+    BIBTEXML(
+            "BibTexMl, bibx"),
```
oscargus: BibTeXML?

src/main/java/net/sf/jabref/gui/FileExtensions.java
```diff
@@ -0,0 +1,61 @@
...
+    AUX(
+            "Aux file", "aux"),
+    SILVER_PLATTERN(
+            "SilverPlatter", "dat"),
+    PUBMED(
+            "Pubmed file", "fcgi"),
```
oscargus: PubMed?

src/main/java/net/sf/jabref/gui/FileExtensions.java
```diff
@@ -0,0 +1,61 @@
+package net.sf.jabref.gui;
+
+public enum FileExtensions {
```
oscargus: Should these be translatable?

src/main/java/net/sf/jabref/external/MoveFileAction.java
```diff
@@ -143,23 +144,16 @@ public void actionPerformed(ActionEvent event) {
                     }
                     chosenFile = sb.toString();
                 } else {
-                    chosenFile = FileDialogs.getNewFile(frame, file, Collections.singletonList(extension),
-                            JFileChooser.SAVE_DIALOG, false);
+                    System.out.println("Entry extension " + extension);
+                    chosenFile = new NewFileDialogs(frame, file.getPath()).saveNewFile().toString();
```
tobiasdiez: Below you create a `File newFile` from it, so probably no need to first convert it to a string

src/main/java/net/sf/jabref/gui/FileExtensions.java
```diff
@@ -0,0 +1,61 @@
...
+
+    //importannt: No dot before the extension!
```
tobiasdiez: important with only one `n` (and probably add this as a javadoc comment)

src/main/java/net/sf/jabref/gui/FileExtensions.java
```diff
@@ -0,0 +1,61 @@
...
+    BIBTEX_DB(
+            "BibTeX Database", "bib"),
+    BIBTEXML(
+            "BibTexMl, bibx"),
+    AUX(
+            "Aux file", "aux"),
```
tobiasdiez: AUX file? I actually like "Aux" more but we recently had this discussion about uniform file type names

src/main/java/net/sf/jabref/gui/NewFileDialogs.java
```diff
@@ -0,0 +1,161 @@
...
+        }
+    };
+
+    private final JFrame parent;
+    private final String directory;
+    private FileNameExtensionFilter extFilter;
+    private Set<FileExtensions> extensions = EnumSet.noneOf(FileExtensions.class);
+
+
+    public NewFileDialogs(JFrame owner) {
```
tobiasdiez: rename owner -> parent

src/main/java/net/sf/jabref/gui/NewFileDialogs.java
```diff
@@ -0,0 +1,161 @@
...
+        this(owner, getWorkingDir());
+    }
+
+    public NewFileDialogs(JFrame owner, String dir) {
+        this.parent = owner;
+        this.directory = dir;
+        System.out.println("DIR in Constructor" + Paths.get(dir));
```
tobiasdiez: to cite Oscar: here too 💃

src/main/java/net/sf/jabref/gui/NewFileDialogs.java
```diff
@@ -0,0 +1,161 @@
...
+        fileChooser.setFileSelectionMode(JFileChooser.DIRECTORIES_ONLY);
+        return this;
+    }
+
+    public NewFileDialogs withExtension(FileExtensions singleExt) {
+        withExtension(EnumSet.of(singleExt));
+        return this;
+    }
+
+    public NewFileDialogs withExtension(Set<FileExtensions> fileExtensions) {
```
tobiasdiez: plural s: withExtensions and probably accept an arbitrary collection (not just a set)

src/main/java/net/sf/jabref/gui/NewFileDialogs.java
```diff
@@ -0,0 +1,161 @@
...
+    private boolean showDialogAndIsAccepted() {
+        return fileChooser.showDialog(parent, null) == JFileChooser.APPROVE_OPTION;
+    }
+
+    private static String getWorkingDir() {
+        return Globals.prefs.get(JabRefPreferences.WORKING_DIRECTORY);
+    }
+
+    private static void updateWorkingDirectorySetting(String dir) {
+        Globals.prefs.put(JabRefPreferences.WORKING_DIRECTORY, dir);
```
tobiasdiez: Inline this method

src/main/java/net/sf/jabref/gui/NewFileDialogs.java
```diff
@@ -0,0 +1,161 @@
...
+        for (FileExtensions ext : fileExtensions) {
+            extFilter = new FileNameExtensionFilter(ext.getDescription(), ext.getExtensions());
+            fileChooser.addChoosableFileFilter(extFilter);
+
+        }
+
+        return this;
+    }
+
+    public NewFileDialogs updateWorkingDirectory() {
```
tobiasdiez: I would make this a void method witch accepts a JabRefPreferences class

src/main/java/net/sf/jabref/gui/NewFileDialogs.java
```diff
@@ -0,0 +1,161 @@
...
+
+    public NewFileDialogs(JFrame owner, String dir) {
+        this.parent = owner;
+        this.directory = dir;
+        System.out.println("DIR in Constructor" + Paths.get(dir));
+        fileChooser.setCurrentDirectory(Paths.get(dir).toFile());
+    }
+
+    private void saveFileDialog() {
+        fileChooser.setDialogType(JFileChooser.SAVE_DIALOG);
```
tobiasdiez: Inline these 2 methods

src/main/java/net/sf/jabref/gui/NewFileDialogs.java
```diff
@@ -0,0 +1,161 @@
...
+
+        return this;
+    }
+
+    public NewFileDialogs updateWorkingDirectory() {
+        updateWorkingDirectorySetting(this.directory);
+        return this;
+    }
+
+    public List<String> getMultipleFileNames() {
```
tobiasdiez: I'm not sure if it clear from the names that some of the methods actually open a dialog. Do you think something like `showDialogToGetMutlipleFileNames` is to verbose?

src/main/java/net/sf/jabref/gui/NewFileDialogs.java
```diff
@@ -0,0 +1,161 @@
...
+                    .collect(Collectors.toList());
+
+            return files;
+        }
+
+        return Collections.emptyList();
+
+    }
+
+    public Path getSelectedFile() {
```
tobiasdiez: Return Optional<Path> to indicate if no file was selected / dialog was aborted? Applies to this method and the ones below.

src/main/java/net/sf/jabref/gui/NewFileDialogs.java
```diff
@@ -0,0 +1,161 @@
...
+        if (showDialogAndIsAccepted()) {
+            return fileChooser.getSelectedFile().toPath();
+        }
+        return Paths.get("");
+    }
+
+    public Path saveNewFile() {
+        saveFileDialog();
+        if (showDialogAndIsAccepted()) {
+            File f = fileChooser.getSelectedFile();
```
tobiasdiez: No abbreviations pls

src/main/java/net/sf/jabref/gui/NewFileDialogs.java
```diff
@@ -0,0 +1,161 @@
...
+            if (!extensions.isEmpty() && !fileChooser.accept(f)) {
+
+                System.out.println("Exts " + extensions);
```
tobiasdiez: Remove debug output

src/main/java/net/sf/jabref/gui/NewFileDialogs.java
```diff
@@ -0,0 +1,161 @@
...
+
+                return new File(f.getPath() + extensions.iterator().next().getFirstExtensionWithDot()).toPath();
```
tobiasdiez: I think there was a better way to add an extension to a path using nio path methods (i.e. something like `file.toPath().concatenate(extension)` )

src/main/java/net/sf/jabref/gui/NewFileDialogs.java
```diff
@@ -0,0 +1,161 @@
...
+        updateWorkingDirectorySetting(this.directory);
+        return this;
+    }
+
+    public List<String> getMultipleFileNames() {
+        openFileDialog();
+        fileChooser.setMultiSelectionEnabled(true);
+
+        if (showDialogAndIsAccepted()) {
+            List<String> files = Arrays.stream(fileChooser.getSelectedFiles()).map(f -> f.toString())
```
tobiasdiez: No abbreviation pls

src/main/java/net/sf/jabref/gui/NewFileDialogs.java
```diff
@@ -0,0 +1,161 @@
...
+            return f.toPath();
+        }
+        return Paths.get("");
+    }
+
+    private boolean showDialogAndIsAccepted() {
+        return fileChooser.showDialog(parent, null) == JFileChooser.APPROVE_OPTION;
+    }
+
+    private static String getWorkingDir() {
```
tobiasdiez: Accept JabRefPreferences instead of operating on Globals.prefs

src/main/java/net/sf/jabref/gui/exporter/SaveDatabaseAction.java
```diff
@@ -309,19 +310,27 @@ public void saveAs() throws Throwable {
         String chosenFile;
         File f = null;
         while (f == null) {
-            chosenFile = FileDialogs.getNewFile(frame, new File(Globals.prefs.get(JabRefPreferences.WORKING_DIRECTORY)),
-                    Collections.singletonList(".bib"), JFileChooser.SAVE_DIALOG, false, null);
+
+            Path p = new NewFileDialogs(frame).withExtension(FileExtensions.BIBTEX_DB).saveNewFile();
+
+            chosenFile = p.toString();
+            // chosenFile = FileDialogs.getNewFile(frame, new File(Globals.prefs.get(JabRefPreferences.WORKING_DIRECTORY)),
```
tobiasdiez: Remove comments if they are not needed anymore

src/main/java/net/sf/jabref/gui/exporter/SaveDatabaseAction.java
```diff
@@ -309,19 +310,27 @@ public void saveAs() throws Throwable {
         String chosenFile;
         File f = null;
         while (f == null) {
-            chosenFile = FileDialogs.getNewFile(frame, new File(Globals.prefs.get(JabRefPreferences.WORKING_DIRECTORY)),
-                    Collections.singletonList(".bib"), JFileChooser.SAVE_DIALOG, false, null);
+
+            Path p = new NewFileDialogs(frame).withExtension(FileExtensions.BIBTEX_DB).saveNewFile();
```
tobiasdiez: no abbreviation, please check this everywhere

tobiasdiez:
I really like the new FileDialog class and its fluent interface. The code needs some cleanup in general (remove commented-out code and debug prints, no abbreviations, ...) and then I think it is good to go.

src/main/java/net/sf/jabref/gui/FileExtensions.java
```diff
@@ -0,0 +1,61 @@
...
+    LAYOUT(
+            "Custom Layout File", "layout"),
+    XML(
+            "XML file", "xml"),
+    TERMS(
+            "Protected TERMS file", "terms"),
+    TXT(
+            "Plain text file", "txt"),
+    CLASS(
+            "Class file", "class"),
```
tobiasdiez: I think most of the description values should actually be localized.

src/main/java/net/sf/jabref/external/MoveFileAction.java
Siedlerchr: Nope, the overwrite dialog is implemented as part of the JFileChooser (overriden method), so it needs not to be handled everywhere

src/main/java/net/sf/jabref/gui/FileExtensions.java
Siedlerchr: I was not sure about this, so I wanted to wait for feedback. But I agree that it probably makes sense.

src/main/java/net/sf/jabref/gui/NewFileDialogs.java
Siedlerchr: Not sure, but you mean sth like this:
`JabRefPreferences.getInstance().get(JabRefPreferences.WORKING_DIRECTORY);`

src/main/java/net/sf/jabref/gui/NewFileDialogs.java
Siedlerchr: The decision against a void method is that you then can't chain it together.
You would have to create a second call.

src/main/java/net/sf/jabref/gui/NewFileDialogs.java
oscargus: This is gui code so no need to inject `JabRefPreferences`? (I think @tobiasdiez means `private static String getWorkingDir(JabRefPreferences prefs) {`)

Siedlerchr:
Adressed comments. @tobiasdiez  Not sure if I understand you right on how I should handle the WorkingDir-Prefs-stuff.
To me it does not make sense to pass the Working dir to the constructor from an external call. It would create unnecessary dependencies to the prefs.  Or what was your idea?

src/main/java/net/sf/jabref/gui/NewFileDialogs.java
Siedlerchr: Yes, all gui code, but still a bit confused about the Pref stuff :confused:  
The getWorkingDir() is only called from the constructor.
```
public NewFileDialogs(JFrame owner) {
        this(owner, getWorkingDir());
    }
```
So here I would need to pass the JabRefPreferences.getInstance()?

tobiasdiez: Yes I thought about injecting the JabRefPreferences class. So
̀`getWorkingDir(JabRefPreferences preferences)` and `NewFileDialogs(JFrame owner, JabRefPreferences preferences)` which then is called via `new NewFileDialogs(owner, Globals.prefs)`. 

Just because its GUI code doesn't mean that it is ok to work on `Globals.prefs`. We are just not as strict in enforcing the rule since sometimes it is very hard to implement. But here it is relatively simple.

src/main/java/net/sf/jabref/gui/NewFileDialogs.java
tobiasdiez: Yes you are right that you can't chain in it. But in for this method, I think chained code is harder to understand.
I have no idea what `updateWorkingDirectory` does here (since it looks like that the working directory of the dialog is updated, but thats not the case):
`new FileDialog(...).withExtensions(...).updateWorkingDir().showDialog()` 
versus
````
FileDialog dialog = new FileDialog(...).withExtensions(...);
dialog.updateWorkingDir();
dialog.showDialog();
````
is more understandable in my opinion. But I have no strong opinion on this.

File: src/main/java/net/sf/jabref/gui/NewFileDialogs.java (hunk 29 above)
Comment (oscargus) [inline src/main/java/net/sf/jabref/gui/NewFileDialogs.java]: I agree that it is simple here, but I'm still under the impression that it indeed is OK. With that said, I also happens to inject it in gui code every now and then.

I guess here one may actually inject the preference value rather than the complete preferences as well?

Comment (oscargus) [inline src/main/java/net/sf/jabref/gui/NewFileDialogs.java]: Oh, injecting the correct value in this particular case wouldn't make much sense. :-)

Comment (Siedlerchr) [inline src/main/java/net/sf/jabref/gui/NewFileDialogs.java]: I thought about this again and I see no advantages over injecting the Preferences or passing everytime the working directory.
The goal of the redesign was that the developer does not have to care about setting/passing the working directory every time.  Passing the Preferences every time would not make any sense to me. Addtionally, it would create some extra dependencies of Globals.prefs  to the caller class.

File: src/main/java/net/sf/jabref/gui/FileExtensions.java [hunk 33]
```diff
@@ -0,0 +1,68 @@
...
+    AUX(
+            Localization.lang("%0 file", "AUX"), "aux"),
+    SILVER_PLATTERN(
+            Localization.lang("%0 file", "SilverPlatter"), "dat"),
+    PUBMED(
+            Localization.lang("%0 file", "PubMed"), "fcgi"),
+    RIS(
+            Localization.lang("%0 file", "RIS"), "ris"),
+    ENDNOTE(
+            Localization.lang("Endnote/Refer file"), "ref"),
```
koppor: Please rewrite to `Localization.lang("%0 file", "Endnote/Refer"), "ref"),`

src/main/resources/l10n/JabRef_zh.properties
```diff
@@ -1661,3 +1658,10 @@ Add_selected_text_to_list=
 Add_{}_around_selected_text=
 Format_field=
 New_protected_terms_file=
+
+%0_file=
+BibTeX_database=
```
koppor: Remove this and everyhting below. See above for a code solution.

src/main/resources/l10n/JabRef_zh.properties
```diff
@@ -1661,3 +1658,10 @@ Add_selected_text_to_list=
...
+Custom_layout_file=
+Endnote/Refer_file=
```
koppor: Please remove this, too. "Endnote" and "Refer" are fixed names.

The other translations are OK.

Siedlerchr:
@oscargus  Could you please have a look why the `cleanupCasesAddsBracketAroundAluminiumGalliumArsenid ` test is failing?
Locally all is okay for me. I have no idea what it could cause on Travis.  And I didn't touch any of the cleanup code in this PR . :confused:

Issue #1340, opened 2016-05-02
Edit -> Mark Specific Color Dysfunctional on OSX


jkulesza:
JabRef version 3.3 on OSX 10.11.4

Steps to reproduce:

With open database, go to Edit -> Mark Specific Color.  The colors do not show up (see screenshot below).  Note that right-clicking on an entry shows the colors appropriately.

<img width="563" alt="screen shot 2016-05-02 at 11 26 28 am" src="https://cloud.githubusercontent.com/assets/9939059/14958463/e8dc4a90-1058-11e6-88a8-36b40d23ca42.png">

simonharrer:
@chochreiner could you have a look? On my windows machine everything works fine.

chochreiner:
I can reproduce this bug with the current master - i may have a look at it at the end of this week

stefan-kolb:
Closed by #1344

PR #1344, opened 2016-05-03
preliminary fix for #1340

chochreiner:
Added a textual description for the functionality "Mark Specific Color" since some Look and Feels (including the OSX one) do not consider the Background and Foreground color (at least for) JMenuItem

the current fix at least provides some kind of indicator of the level instead of a blank field as before


simonharrer:
Would it make sense to use this for the other menu as well, which can be used from within the right-click menu? Or is this not an issue there?

chochreiner:
the right-click menu works as intended - nevertheless this PR also adds the Level 0 - Level 5 notion there

simonharrer:
LGTM

stefan-kolb:
Can you please quickly rebase this, then I'll merge this in.

chochreiner:
done

stefan-kolb:
Merci :smile:

PR #1349, opened 2016-05-03
Fix #1347 Open URLs does not work with query strings

### LINKED ISSUE #1347: Open URLs does not work
JabRef version: dev on Windows 10
<Hint: If you use a [development version](http://builds.jabref.org/master/), ensure that you use the latest one.>

Steps to reproduce:

1. Create entry with url field `http://books.google.de/books?id=0yaM63MuqckC&lpg=PA153&ots=0H7EqpbM79&lr&hl=de&pg=PA153#v=onepage&q&f=false` simpler urls like `http://books.google.de` work as expected.
2. Click on the open button next to the url field
3. Windows Explorer opens (at some folder) instead of the browser. Sometimes you have to press the button multiple times before something happens.

<If applicable, excerpt of the bibliography file, screenshot, and excerpt of log (available in the error console)>

stefan-kolb:
This should work with query strings like
`http://books.google.de/books?id=0yaM63MuqckC&lpg=PA153&ots=0H7EqpbM79&lr&hl=de&pg=PA153#v=onepage&q&f=false`
and normal URLs.

However, the question is:
Does this impact the opening of any other file path or anything? We need to be careful as this is a generic open method. Maybe that's also the problem because it needs to handle any kind of URI.

@Siedlerchr @tobiasdiez Please take a look.

### DIFF
--- a/src/main/java/net/sf/jabref/gui/desktop/os/Windows.java
+++ b/src/main/java/net/sf/jabref/gui/desktop/os/Windows.java
@@ -19,8 +19,9 @@ public class Windows implements NativeDesktop {
         if (type.isPresent() && !type.get().getOpenWithApplication().isEmpty()) {
             openFileWithApplication(filePath, type.get().getOpenWithApplication());
         } else {
-            //filePath as string, because it could be an URL
-            Runtime.getRuntime().exec(new String[] {"explorer.exe", filePath});
+            // quote String so explorer handles URL query strings correctly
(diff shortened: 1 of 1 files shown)


Siedlerchr:
Tested it. Works, but requires for Urls a http:// to be present.
a simple `www.gooogle.de` does not work. But it also does not work with explorer "www.google.de" from the start menu. So I assume this is some windows behaviour.

It would be nice to have a cleanup for this then or maybe we should automatically do this.

PR #1350, opened 2016-05-03
Fix NPE on selecting invalid jstyle

Siedlerchr:
Follow up from #1294 

NPE was raised in Comparator because name was initialized with null instead of Empty string
Only store preferences if style is valid

@oscargus  The fix has the side effect that a file where the name can not be parsed correctly is displayed as "Empty style". I attached the problematic style file which was provided in #1294 
However: Valid still returns true. 
As you have worked recently with the styles things, do you know what is required for a jstyle? @koppor  Maybe you know this?
Then we could improved the validity check 

[campus_format.english.txt](https://github.com/JabRef/jabref/files/247561/campus_format.english.txt)
emptyjstyle


Siedlerchr:
I suggest merging this. For the discussion on a validity checker for ooStyle I will create a new Issue.

Issue #463, opened 2015-12-04
Some actions are active while no database is open. 


simonharrer:
In the UI, some menu entries are enabled despite the fact that there is no database loaded. For instance, the rank action. We should by default disable all actions and only enable actions based on different whitelists (no databases loaded, one entry selected, two entries selected, many entries selected, ...) to ensure that we only enable the actions when they fit the context.

koppor:
More examples stated at https://sourceforge.net/p/jabref/bugs/1280/.

Wait for #113?

koppor:
@motokito You addressed the issue with **no database** in #1743. Simon also talked about selection of multiple entries. I checked the context menu of entries. For instance, "Print Entry Preview" prints the currently selected entry and and "Attach File" does not do anything at all.

Can you please use [complex.bib](https://github.com/JabRef/jabref/blob/master/src/test/resources/testbib/complex.bib) (and other bib files if necessary) and test if all actions are disabled correctly a) in the context menu, b) in the toolbar, c) in the menu? If not, please fix. At least, Print Entry Preview and Attach File" have to be fixed.

boceckts:
I will close this as everything seems to be done.

melisolmez:
Hello I noticed an error yesterday. Maybe, You want to examine. the error is like this:

redo and undo buttons become active when the project is up and running:
![redo undo](https://github.com/user-attachments/assets/c179bdfd-0e31-42a0-8445-8062dc812e15)

When you click on these buttons, an error message appears like this: 
![image](https://github.com/user-attachments/assets/d2f8527b-d94a-47e2-b351-4a02310ffdc7)

Siedlerchr:
@melisolmez I moved this to a new issue. It's better to have it as a new issue instead of commenting on an old closed one

PR #1358, opened 2016-05-06
Fix enabled OpenConsole button when no tabs are opened.

obraliar:


obraliar:
I noticed that, while starting JabRef with no opened databases.

tobiasdiez:
LGTM :+1:

koppor:
This refs #463.

simonharrer:
Does it fix #463 completely?

PR #988, opened 2016-03-17
When clicking on a tab, the first field now has the focus

Braunch:
- [x] Change in CHANGELOG.md described?
- [x] Changes in pull request outlined? (What, why, ...)
- [x] Tests created for changes?
- [x] Tests green?

This is referes to #623 
I just had to add the activateVisible() Method to the ChangeListener to ensure the focus is set.

PR #1361, opened 2016-05-07
Scrollingthroughentrylisttest

Braunch:
I wrote a ui-test for the following use case:
I import a bib-file into a new database.
Then I double-click on the first entry in the database to open up the entry editor.
Then I click back into the table and start scrolling through the entries.
When I do that, the table should always keep the focus.

This test case tests the following pull-requests enhancement: #988
It also referes to the discussion about ui-testing: #507 
 
- [x] Tests created

### DIFF
--- a/src/integrationTest/java/net/sf/jabref/gui/AbstractUITest.java
+++ b/src/integrationTest/java/net/sf/jabref/gui/AbstractUITest.java
@@ -1,31 +1,91 @@
 package net.sf.jabref.gui;
 
 import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 
+import net.sf.jabref.JabRefMain;
+
+import org.assertj.swing.fixture.AbstractWindowFixture;
 import org.assertj.swing.fixture.FrameFixture;
 import org.assertj.swing.fixture.JFileChooserFixture;
+import org.assertj.swing.image.ScreenshotTaker;
 import org.assertj.swing.junit.testcase.AssertJSwingJUnitTestCase;
+import org.assertj.swing.timing.Pause;
 
+import static org.assertj.swing.finder.WindowFinder.findFrame;
+import static org.assertj.swing.launcher.ApplicationLauncher.application;
 
 public abstract class AbstractUITest extends AssertJSwingJUnitTestCase {
 
+    protected final static int SPEED_NORMAL = 50;
 
+    protected AWTExceptionHandler awtExceptionHandler;
+    protected FrameFixture mainFrame;
 
-    public String getTestFilePath(String fileName) {
-        return new File(this.getClass().getClassLoader().getResource(fileName).getFile()).getAbsolutePath();
+    @Override
+    protected void onSetUp() {
+        awtExceptionHandler = new AWTExceptionHandler();
+        awtExceptionHandler.installExceptionDetectionInEDT();
+        application(JabRefMain.class).start();
+
+        robot().waitForIdle();
+
+        robot().settings().timeoutToFindSubMenu(1_000);
+        robot().settings().delayBetweenEvents(SPEED_NORMAL);
+
+        mainFrame = findFrame(JabRefFrame.class).withTimeout(10_000).using(robot());
+        robot().waitForIdle();
     }
 
-    public void importBibIntoNewDatabase(FrameFixture mainFrame, String path) {
-        // have to replace backslashes with normal slashes b/c assertJ can't type the former one on windows
-        path = path.replace("\\", "/");
+    /**
+     * Returns the absolute Path of the given relative Path
+     * The backlashes are replaced with forwardslashes b/c assertJ can't type the former one on windows
+     */
+    protected String getAbsolutePath(String relativePath) {
+        return new File(this.getClass().getClassLoader().getResource(relativePath).getFile()).getAbsolutePath().replace("\\", "/");
+    }
 
+    /**
+     * opens a database and gives JabRef a second to open it before proceeding
+     */
+    protected void importBibIntoNewDatabase(String path) {
         mainFrame.menuItemWithPath("File", "Import into new database").click();
         JFileChooserFixture openFileDialog = mainFrame.fileChooser();
         robot().settings().delayBetweenEvents(1);
         openFileDialog.fileNameTextBox().enterText(path);
-        robot().settings().delayBetweenEvents(1_000);
         openFileDialog.approve();
-        robot().settings().delayBetweenEvents(50);
+        Pause.pause(1_000);
+    }
+
+    protected void exitJabRef() {
+        mainFrame.menuItemWithPath("File", "Quit").click();
+        awtExceptionHandler.assertNoExceptions();
+    }
+
+    protected void newDatabase() {
+        mainFrame.menuItemWithPath("File", "New BibTeX database").click();
+    }
+
+    protected void closeDatabase() {
+        mainFrame.menuItemWithPath("File", "Close database").click();
+    }
+
+    protected void takeScreenshot(AbstractWindowFixture<?, ?, ?> dialog, String filename) throws IOException {
+        ScreenshotTaker screenshotTaker = new ScreenshotTaker();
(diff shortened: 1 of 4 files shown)


src/integrationTest/java/net/sf/jabref/gui/EntryTableTest.java
```diff
@@ -0,0 +1,91 @@
...
+    public void scrollThroughEntryList() throws InterruptedException {
+        FrameFixture mainFrame = findFrame(JabRefFrame.class).withTimeout(10_000).using(robot());
+        robot().waitForIdle();
+
+        mainFrame.menuItemWithPath("File", "Import into new database").click();
+        JFileChooserFixture openFileChooser = mainFrame.fileChooser();
+
+        //reduce the delay to speed up loading the test resource
+        robot().settings().delayBetweenEvents(10);
+        openFileChooser.fileNameTextBox().enterText(getTestFilePath(TEST_FILE_NAME));
```
tschechlovdev: If I see right, you just want the file path as a string. So I don't think you need getTestFilePath method if you're using java.nio.path. 
See [here] (http://www.java2s.com/Tutorials/Java/java.nio.file/Path/Java_Path_getFileName_.htm) for an example.

src/integrationTest/java/net/sf/jabref/gui/EntryTableTest.java
```diff
@@ -0,0 +1,91 @@
...
+    private final static String TEST_FILE_NAME = "testjabref.bib";
+
+    private int down = KeyEvent.VK_DOWN;
+
+    @Override
+    protected void onSetUp() {
+        AWTExceptionHandler awtExceptionHandler = new AWTExceptionHandler();
+        awtExceptionHandler.installExceptionDetectionInEDT();
+        application(JabRefMain.class).start();
+
```
tschechlovdev: I think one empty line is enough, so delete one.

src/integrationTest/java/net/sf/jabref/gui/EntryTableTest.java
```diff
@@ -0,0 +1,91 @@
...
+        //delay has to be shortened so that double click is recognized
+        robot().settings().delayBetweenEvents(0);
+        firstCell.doubleClick();
+        robot().settings().delayBetweenEvents(50);
+
+        firstCell.click();
+
+        for(int i=0; i < SCROLL_ACTION_EXECUTION; i++)
+        {
```
tschechlovdev: why is the clip in a new line?

src/integrationTest/java/net/sf/jabref/gui/EntryTableTest.java
```diff
@@ -0,0 +1,91 @@
...
+        //reduce the delay to speed up loading the test resource
+        robot().settings().delayBetweenEvents(10);
+        openFileChooser.fileNameTextBox().enterText(getTestFilePath(TEST_FILE_NAME));
+        openFileChooser.approveButton().click();
+        robot().settings().delayBetweenEvents(50);
+
+        JTableFixture entryTable = mainFrame.table();
+
+        Pattern pattern = Pattern.compile("256.*");
+        JTableCellFixture firstCell = entryTable.cell(pattern);
```
tschechlovdev: would be nice if you would found out a way to always select the first cell and that it not depends on the content.

src/integrationTest/java/net/sf/jabref/gui/EntryTableTest.java
```diff
@@ -0,0 +1,91 @@
...
+        awtExceptionHandler.installExceptionDetectionInEDT();
+        application(JabRefMain.class).start();
+
+
+        robot().settings().timeoutToFindSubMenu(1_000);
+        robot().settings().delayBetweenEvents(50);
+    }
+
+    @Test
+    public void scrollThroughEntryList() throws InterruptedException {
```
tschechlovdev: ``throw ...`` is not necessary

src/integrationTest/java/net/sf/jabref/gui/EntryTableTest.java
```diff
@@ -0,0 +1,91 @@
...
+ * I import a database. Then I doubleclick on the first entry in the table to open the entry editor.
+ * Then I click on the first entry again, and scroll through all of the lists entries, without having to click
+ * on the table again.
+ */
+public class EntryTableTest extends AssertJSwingJUnitTestCase {
+
+    private final static int SCROLL_ACTION_EXECUTION = 5;
+    private final static String TEST_FILE_NAME = "testjabref.bib";
+
+    private int down = KeyEvent.VK_DOWN;
```
tschechlovdev: why is this field not declared final static?

src/integrationTest/java/net/sf/jabref/gui/EntryTableTest.java
Braunch: I tried using entryTable().selectRows(0) but that did not set the focus right. The pattern seems to be the easiest way to get the cell object.

Braunch:
I created an `abstract class AbstractUITest` since the Undo test and the entry table test were using the same functionality. The scrolling though entry table test should now also work on windows os. @tschechlovdev  can you confirm that?

tschechlovdev:
should work now on windows :+1: 
can you please squash the commits, so that it is easier for others to review this PR?

Braunch:
Is there any feedback on this Pull-Request?

src/integrationTest/java/net/sf/jabref/gui/AbstractUITest.java
```diff
@@ -0,0 +1,91 @@
...
+        mainFrame = findFrame(JabRefFrame.class).withTimeout(10_000).using(robot());
+        robot().waitForIdle();
+    }
+
+    /**
+     * Returns the absolute Path of the given relative Path
+     * The backlashes are replaced with forwardslashes b/c assertJ can't type the former one on windows
+     */
+    protected String getAbsolutePath(String relativePath) {
+        return new File(this.getClass().getClassLoader().getResource(relativePath).getFile()).getAbsolutePath().replace("\\", "/");
```
simonharrer: Please try to use Path instead of File

src/integrationTest/java/net/sf/jabref/gui/EntryTableTest.java
```diff
@@ -0,0 +1,58 @@
...
+        firstCell.click();
+
+        for (int i=0; i < SCROLL_ACTION_EXECUTION; i++) {
+            robot().pressAndReleaseKey(DOWN);
+            Assert.assertTrue(entryTable.selectionValue() != null);
+        }
+
+        for (int i = 0; i < SCROLL_ACTION_EXECUTION; i++) {
+            robot().pressAndReleaseKey(UP);
+            Assert.assertTrue(entryTable.selectionValue() != null);
```
simonharrer: can you assert that the selected value is the actual number X from the bib file?

src/integrationTest/java/net/sf/jabref/gui/AbstractUITest.java
```diff
@@ -0,0 +1,91 @@
...
+import org.assertj.swing.fixture.FrameFixture;
+import org.assertj.swing.fixture.JFileChooserFixture;
+import org.assertj.swing.image.ScreenshotTaker;
+import org.assertj.swing.junit.testcase.AssertJSwingJUnitTestCase;
+import org.assertj.swing.timing.Pause;
+
+import static org.assertj.swing.finder.WindowFinder.findFrame;
+import static org.assertj.swing.launcher.ApplicationLauncher.application;
+
+public abstract class AbstractUITest extends AssertJSwingJUnitTestCase {
```
simonharrer: :+1: for extracting this

src/integrationTest/java/net/sf/jabref/gui/EntryTableTest.java
Braunch: Is it sufficient to check if the value in the selected row is the same as in the row with the next index?

simonharrer: Yes, you need to be sure that the next entry is selected, not that any entry is selected.

PR #1373, opened 2016-05-10
Remove StringUtil.expandAuthorInitials

tobiasdiez:
Reasoning:
- only used by a few importers and not consistently so
- better handled by cleanup actions / converter
- duplicates logic to parse list of author names (while the AuthorList.parse method is far superior)
- results in discussion about string joiner and builder in #547 :smile_cat:

### DIFF
--- a/src/main/java/net/sf/jabref/logic/util/strings/StringUtil.java
+++ b/src/main/java/net/sf/jabref/logic/util/strings/StringUtil.java
@@ -23,7 +23,6 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import net.sf.jabref.Globals;
-import net.sf.jabref.model.entry.Author;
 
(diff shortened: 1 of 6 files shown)


src/main/java/net/sf/jabref/logic/util/strings/StringUtil.java
```diff
@@ -627,53 +626,6 @@ public static String replaceSpecialCharacters(String s) {
     }
 
     /**
-     * Expand initials, e.g. EH Wissler -> E. H. Wissler or Wissler, EH -> Wissler, E. H.
```
simonharrer: Can you add the example as a test?

simonharrer:
LGTM

Please add a changelog with the recommendation that save actions with the name of the appropriate save action should be used.

PR #1378, opened 2016-05-10
Add MathSciNet tab

tobiasdiez:
If the `MRNumber` field is present in the entry, then a new tab is added which shows the associated MathSciNet review. (Note that MathSciNet is probably _the_ reference for all mathematicians).

This PR is against the javafx branch since the javafx control WebView is used to show the webpage.
@JabRef/stupro also have a look at this PR, since you are the ones with JavaFX experience.

untitled

### DIFF
--- a/src/main/java/net/sf/jabref/gui/entryeditor/EntryEditor.java
+++ b/src/main/java/net/sf/jabref/gui/entryeditor/EntryEditor.java
@@ -64,6 +64,13 @@ import javax.swing.event.ChangeEvent;
 import javax.swing.event.ChangeListener;
 import javax.swing.text.JTextComponent;
 
+import javafx.application.Platform;
+import javafx.embed.swing.JFXPanel;
+import javafx.scene.Scene;
+import javafx.scene.control.ProgressIndicator;
+import javafx.scene.layout.StackPane;
+import javafx.scene.web.WebView;
+
 import net.sf.jabref.Globals;
 import net.sf.jabref.JabRefPreferences;
 import net.sf.jabref.bibtex.BibEntryWriter;
@@ -115,12 +122,6 @@ import net.sf.jabref.model.entry.EntryType;
 import net.sf.jabref.specialfields.SpecialFieldUpdateListener;
 
 import com.google.common.eventbus.Subscribe;
-import javafx.application.Platform;
-import javafx.embed.swing.JFXPanel;
-import javafx.scene.Scene;
-import javafx.scene.control.ProgressIndicator;
-import javafx.scene.layout.StackPane;
-import javafx.scene.web.WebView;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
--- a/src/main/java/net/sf/jabref/gui/entryeditor/EntryEditor.java
+++ b/src/main/java/net/sf/jabref/gui/entryeditor/EntryEditor.java
@@ -105,6 +105,7 @@ import net.sf.jabref.logic.l10n.Localization;
 import net.sf.jabref.logic.labelpattern.LabelPatternUtil;
 import net.sf.jabref.logic.search.SearchQueryHighlightListener;
 import net.sf.jabref.logic.util.date.TimeStamp;
+import net.sf.jabref.logic.util.strings.StringUtil;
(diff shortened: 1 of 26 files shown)


boceckts:
It depends on what you want to achieve, displaying the content of a webpage or giving the user access to a fully functioning web browser inside JabRef. Also note that javascipt, possibly java and even flash are also activated inside the `WebView` component. Also my JRE crashed once when naviagting through a pages with a lot of javascript. - So I don't know how reliable the component is.
screen shot 2016-05-11 at 12 27 33
If you want the user to be able to navigate but only on the given website you can add a `listener` to the `locationProperty()` of the engine of the `WebView` component and filter which websites should be loaded. Same goes if the user shouldn't be able to navigate on JavaFX.
If you only want to display the website with its correct html and css appearance on SWING I would prefer using the `javax.swing.JEditorPane` and use the methode `setPage(<aURLObject>)`.
Otherwise the JFXPanel is exactly the component you need to use for adding JavaFX components into SWING components.

src/main/java/net/sf/jabref/gui/entryeditor/EntryEditor.java
```diff
@@ -330,6 +340,41 @@ private void addGeneralTabs() {
...
+                browser.getEngine().load("http://www.ams.org/mathscinet-getitem?mr=" + mrNumber);
+
+                // Update progress according to browser
+                progress.progressProperty().bind(browser.getEngine().getLoadWorker().progressProperty());
+                // Hide progress indicator if finished
+                browser.getEngine().getLoadWorker().stateProperty().addListener((observable, oldState, newState) -> {
+                    if(newState == Worker.State.SUCCEEDED) {
+                        progress.setVisible(false);
+                    }
+                });
```
boceckts: The progress bar will never be shown, you need to add something like this
```java
else if (newState == Worker.State.RUNNING) {
    progress.setVisible(true);
}
```

src/main/java/net/sf/jabref/gui/entryeditor/EntryEditor.java
```diff
@@ -330,6 +340,41 @@ private void addGeneralTabs() {
...
+            // Take everything before whitespace or open bracket, so something like `619693 (82j:58046)` gets parsed correctly
+            String mrNumber = StringUtil.tokenizeToList(mrNumberRaw, " (").get(0);
+
+            JFXPanel reviewPane = new JFXPanel();
+            tabbed.addTab(Localization.lang("MathSciNet Review"), reviewPane);
+            tabs.add(reviewPane);
+
+            // Execute on JavaFX Application Thread
+            Platform.runLater(() -> {
+                StackPane root = new StackPane();
```
boceckts: Maybe not the best choice, the progress bar will always be in the center but also on top of the `WebView`. In my opinion a `BorderPane` with the browser in the center and the progress bar in the bottom area would be more suitable.

src/main/java/net/sf/jabref/gui/entryeditor/EntryEditor.java
```diff
@@ -330,6 +340,41 @@ private void addGeneralTabs() {
...
+                ProgressBar progress = new ProgressBar();
+                WebView browser = new WebView();
+                //root.getChildren().add(progress);
+                root.getChildren().addAll(browser, progress);
+                reviewPane.setScene(new Scene(root));
+
+                browser.getEngine().load("http://www.ams.org/mathscinet-getitem?mr=" + mrNumber);
```
boceckts: Maybe also display a message if there is no internet connection available (right now its just blank)

src/main/java/net/sf/jabref/gui/entryeditor/EntryEditor.java
tobiasdiez: I now switched to the ProgressIndicator (instead of bar) and already hide it at 70% completion.

tobiasdiez:
Thanks @boceckts for the review and the helpful comments. 
The swing component does not work since for some reason it is not able to render the embedded MathJax equations.

@JabRef/developers ready to be merged?

simonharrer:
LGTM

But fix l10n issues first.

Issue #872, opened 2016-02-26
The error console should be resizable


koppor:
The Window showing the Error Console (Window "Program output") should be resizable.

Refs #819

oscargus:
See #819 as well in case someone gives it a go.

koppor:
This is fixed at our javafx branch. The binary is available at https://builds.jabref.org/javafx/.

The PR tracking the javafx updates is available https://github.com/JabRef/jabref/pull/2290

Issue #882, opened 2016-02-28
Cleanup "Error console"


koppor:
JabRef as built on 2016-02-28.

### Consistent naming of "Error console"
Help / "Show error console" brings up a window called "Program output". This is inconsistent. To clean the thing up, I propose to 
 * rename "Show error console" to "Show developer information"
 * rename "Program output" to "Developer information"

Reason: I neither find "Error console" nor "Program output" fitting. "Program output" fits better, but it is not the UI, it is the console output somehow. The log, however, is not shown on the console.

### Remove tab "Exceptions"
I think with the current exception handling, we can remove the tab "Exceptions". In case the "Log" shows an exception (such as `java.lang.NullPointerException`), the tab "Exceptions" does shows "No exceptions have occurred." This feels inconsistent. Therefore, I propose to remove the tab "Exceptions".

**Update**: I had an exception today (#890). Maybe we can try to log all exceptions somehow and then remove that tab.

### Change the big red x into something nice.
The window of the error console shows a big X on inside a red circle. That indicates errors, but we also show useful logs without errors. Other dialogs of JabRef also don't have an icon. Therefore, I propose to remove the X completely. If we want to go with icons because they support usability, we should use an "i" somehow.

![grabbed_20160228-141703](https://cloud.githubusercontent.com/assets/1366654/13379430/f39e4dda-de25-11e5-841b-f3d8c231b9c6.png)

oscargus:
The X is because it is a JOptionPane(?) and will be automatically not there when a better solution (resizable and non-modal) is in place (unless someone adds it back explicitly).

simonharrer:
Why not call the tabs "Logs", "System.err" and "System.out"? And the action "Show logs"?

Siedlerchr:
I will try to rewrite this in JavaFX.

koppor:
Good idea. Please base on the javafx branch. There is also a label javafx
and a first pull request for the about dialog rewritten in JavaFX.

tobiasdiez:
I would remove all the tabs and just have one list for all messages. The different kinds of messages can then be differentiated via color or icons. This would also make debugging easier since debug messages are logged before and after the corresponding exception. 

Two further remarks:
- add button to toggle debug mode on/off 
- add button to copy log messages (necessary?)

motokito:
Hiho,
I will ask something:
@koppor Should i remove "Exception" Tab?
@tobiasdiez Should i add this buttons?

thx for tip and help ;-)

[...]

Issue #819, opened 2016-02-15
Error console window is modal


Ede123:
I just installed Jabref 3.2 (updated from 2.x) on Windows 7, java version "1.8.0_73" (32 bit).

While trying to figure out an issue with the full text fetcher I noticed that the error console is a modal window (e.g. no interaction with the JabRef UI is possible while it is opened) and also seems to block the UI thread somehow (the error message that the fetcher had failed in the meantime only appeared _after_ I closed the error console).

If this is "by design" feel free to close this issue right away.
It's however counter-intuitive (from my experience with other applications I'd have expected the normal workflow to be _"Open error console -> Do the steps to reproduce errors in UI -> Check if an error is logged along the way"_, which is obviously not possible with the current implementation).

simonharrer:
Note: then, the error console should also update itself when new errors arrive which I think is currently not the case.

koppor:
This should be fixed at our javafx branch. Please try the binary from https://builds.jabref.org/javafx/ and report back if that works for you.

PR #1383, opened 2016-05-11
Implemented error console in JavaFX

motokito:
This is the error console in JabRef being implemented in JavaFX. The Console is highly flexible in positioning and styling its components through a fxml and corresponding css stylesheet.

Also to show better the error console content (Log, Output, Exception), i have allowed to resize error console windows manually.

error console dialog

PS:
* update the error console screenshot
* new feature:
 * copy log: user can copy the log
 * report issue: user can report directly issues from console.

#872: have done


tobiasdiez:
Please also add a screenshot for these new javafx dialogs. 
Moreover, it would be nice if you could address the issues https://github.com/JabRef/jabref/issues/882, https://github.com/JabRef/jabref/issues/872 (I think, this is already covered with this PR...in this case please add a changelog entry) and https://github.com/JabRef/jabref/issues/819, which are all connected to the error console.

Siedlerchr:
Please adress the Language-key related errors, The first build failed because there were still some obsolete keys...

matthiasgeiger:
@tobiasdiez @Siedlerchr slowly, slowly... let the students check this PR first and wait until it is labeled with "ready-for-review" :wink:

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleViewModel.java
```diff
@@ -0,0 +1,120 @@
...
+    //maybe make it editable as a preference?
+    private static final long REFRESH_RATE = 1000;
+
+    private Thread refreshThread = new Thread(() -> {
+        boolean injectedCloseButton = false;
+        try {
+            while (true) {
+                Thread.sleep(REFRESH_RATE);
+
+                // IMO this is a really, really bad way to do this, but a saw no other way since the stage is
```
tobiasdiez: Suggestion: don't use a refresh thread but data binding.
So you add a property to ErrorConsoleViewModel to return a FX-observable list of output/errormessages which are always up-to-date and then bind the text property of the textarea to it.

motokito: how can i bind the observable list to this outputStream. The problem is that all output gets written to a stream which in my opinion is incompatibe with a observable List.

tobiasdiez: You can register a custom PrintStream as a TeeStream, similar to what is done in https://github.com/JabRef/jabref/blob/144c03c8140737e5c8cac86b6f9156c71ac2bda7/src/main/java/net/sf/jabref/logic/error/StreamEavesdropper.java#L49.
This custom PrintStream writes to a (observable) list. 

````
new PrintStream(System.out) {
List<String> messages = new blabla();
@Override  
public void println(String s) {
    messages.add(s);
    super.println(s); //needed?
  }
````

**Comment by motokito** [comment]
I am having trouble to hide the developer information. The css file probably needs some changes, I tried to set the visibilty but that didn't work out. @tobiasdiez Do you have some tips to handle this?

**Comment by tobiasdiez** [comment]
Sorry for my later answer. Normally you do this by using a PseudoClass `inactive` and change it with `pseudoClassStateChanged`, see for example https://github.com/JabRef/jabref/pull/1438/files#diff-9a2c5dd139caeeb5e3e3eebb9adb24b3R92. Then you can style them via `.list-cell:active` and `.list-cell:inactive`

File: src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleViewModel.java [hunk 2]
```diff
@@ -0,0 +1,141 @@
...
+
+        ButtonBar.setButtonData(developerButton, ButtonBar.ButtonData.LEFT);
+
+        ObservableList<ObservableMessageWithPriority> masterData = ObservableMessages.INSTANCE.messagesPropety();
+        FilteredList<ObservableMessageWithPriority> filteredData = new FilteredList<>(masterData, p -> p.getPriority() == 0);
+        allMessage.setItems(filteredData);
+
+        developerInformation.bind(developerButton.selectedProperty());
+        developerInformation.addListener((observable, oldValue, newValue) -> {
+            ObservableList<ObservableMessageWithPriority> emptyList = FXCollections.observableArrayList(Collections.EMPTY_LIST);
```
tobiasdiez: I think there is no need to replace the items in `allMessage`, just use `filteredData.setPredicate`. Have a look at http://stackoverflow.com/questions/30915007/javafx-filteredlist-filtering-based-on-property-of-items-in-the-list

src/main/java/net/sf/jabref/logic/error/ObservableMessageWithPriority.java
```diff
@@ -0,0 +1,29 @@
...
+    private String message;
+
+    private int priority;
+
+    public ObservableMessageWithPriority(String message, int priority) {
+        this.message = message;
+        this.priority = priority;
+    }
+
+    public int getPriority() {
```
tobiasdiez: Please use an enum as a return value.

motokito:
@tobiasdiez 
I have created FilterList to filter messages. But I have one problem to view correctly entries in Listview (have a look at my Screenshot). It only runs correctly if entries run out of the Listview. I have created a Test Button to test this behavior. This Test Button generates 1 entry in System.out, throws 1 exception and generates 1 entry in the Cache. This will be removed later.
Can you help me and tell me, why Listview doesn't displays the entry correctly?

notcorrectview

tobiasdiez:
Sorry but I don't understand where your problem is. It looks like it prints the Test Cache message as well as the (first line of) the error stack trace. The system.out is ignored as I would have expect since you hide low priority messages.

motokito:
Hi @tobiasdiez ,
So i will explain you, what is my problem.
- The ListView shoulds show you all log entries and other entries like system.out and exception, if the button "Development Information" disable.
- If the button on click, it shoulds hide all system.out and exception entries.
I have handle this will a filterlist. But the behavior by showing of entries is not correct.
- Reproduct:
 - Start error console
 - Click on "Test" Button (1 time or more times)
 - Click on  "Development Information" Button
 - Click on this again
If you do so like i have written, then you will understand my problem.
Sorry about my bad english. I hope that you understand what I mean. I need you help or some tipp how i can clear this problem :-)

Thx you

src/main/java/net/sf/jabref/logic/error/TeeStream.java
```diff
@@ -26,17 +28,32 @@
...
+
+    public TeeStream(PrintStream out1, PrintStream out2, MessagePriority priority) {
+        super(out1);
+        this.outStream = out2;
+        this.priority = priority;
+
+    }
 
     public TeeStream(PrintStream out1, PrintStream out2) {
         super(out1);
```
tobiasdiez: Call the other constructor using this(...) here.

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleViewModel.java
```diff
@@ -0,0 +1,141 @@
...
+        }
+        allMessage.refresh();
+
+    }
+
+    // Handler for copy of Log Entry in the List by click of Copy Log Button
+    @FXML
+    private void copyLogErrorDialog() {
+        String logContentCopy = "";
+        for (String message : GuiAppender.CACHE.get()) {
```
tobiasdiez: Only copy what is displayed in the list view

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleViewModel.java
```diff
@@ -0,0 +1,141 @@
...
+                            if (om.getPriority() == MessagePriority.HIGH) {
+                                if (developerInformation.getValue()) {
+                                    getStyleClass().add("exception");
+                                }
+                            } else if (om.getPriority() == MessagePriority.MEDIUM) {
+                                getStyleClass().add("output");
+                            } else {
+                                getStyleClass().add("log");
+                            }
+                        }
```
tobiasdiez: add here 
```
} else {
   setText("");
}
```
to solve your problem. Source: http://stackoverflow.com/questions/28384803/listview-using-custom-cell-factory-doesnt-update-after-items-deleted

tobiasdiez:
Thanks for the additional description. I added a comment to ErrorConsoleViewModel which should help you to solve the problem. 

For the next time, please try to add some tests first. There were quite a few things which could have been wrong:
- Are the messages correctly passed to the T-Stream?
- Do they have the correct priority?
- Is filtering of the list working properly? 
- And finally: is is just a problem with the way how messages are displayed in the list view.

The first 3 options could have been ruled out by automatic tests. It would be good if you could add corresponding tests (which also require some positve refactoring, like making `filteredList` a (public) class variable and moving its initialization to the constructor).

motokito:
There are three solution for this problem with create issue function:

* **Solution 1:** Using of a GitHub API to create issue directly from JabRef with log in data, which user not like to do it. But I think, that way will be nice solution and GitHub will get more member (take a look in create issue dialog screenshot).
From customer: it will be nice this solution with a default account, but it will be not work.

create issue with github api

* **Solution 2:** Using OAuth2 to create issue directly from JabRef without log in data, but the user must give the application access in their account to use it, That will be difficult for some user.

* **Solution 3:** Create issue over default browser with help of saved log and exception information in clipboard.  (take a look in create issue pop up screenshot).

create issue with default browser

After close or click on OK Button. The default browser will open and show create issue site with default title (automatic bug report-"current time with this format yyyyMMddHHmmss") and a default issue description (JabRef version, operating system version, Java version). During the show of this create issue site, the log and exception information will be saved in clipboard. So the user can be paste it for detail information in issue description. (take a look in create issue in default browser with paste detail information screenshot)

new issue

Braunch:
Which one did you implement?

tobiasdiez:
I really like the copied-to-clipboard-solution. Its simple yet effective.

Siedlerchr:
Can you please post a new screenshot of the actual look?

motokito:
This is the new look of error console dialog:

errorconsoledialog

with button "Developer information" it will show you the exceptions and outputs, if it is exist.
You can copy the log and create issue directly from errorconsole. :-)

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleView.java
```diff
@@ -0,0 +1,93 @@
...
+    private void copyLogButton() {
+        errorViewModel.copyLog();
+    }
+
+    @FXML
+    private void createIssueButton() {
+        errorViewModel.createIssue();
+    }
+
+    // handler for close of error console
```
boceckts: Remove comment

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleViewModel.java
```diff
@@ -0,0 +1,147 @@
...
+    private BooleanProperty developerInformation = new SimpleBooleanProperty();
+    private final String CREATEISSUEURL = "https://github.com/JabRef/jabref/issues/new";
+    private final DateFormat dateFormat = new SimpleDateFormat("yyyyMMddHHmmss");
+    private final Date date = new Date();
+
+    /**
+     * Create allMessage ListView, which shows the filtered entries (default at first only the Log entries),
+     * when the ToggleButton "developerButton" is disable.
+     * If ToggleButton "developerButton" is enable, then it should show all entries
+     * @param allMessage
```
boceckts: Add an empty line between description and parameters in java docs.

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleViewModel.java
```diff
@@ -0,0 +1,147 @@
...
+     * @param developerButton (default is disable at start)
+     */
+
```
boceckts: There shouldn't be an empty line between java doc comments and the method.

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleViewModel.java
```diff
@@ -0,0 +1,147 @@
...
+    public void setUpListView(ListView allMessage, ToggleButton developerButton) {
```
boceckts: The view model should only provide the data so the view class can decide how to display it.
Therefore I would change the method signature to something like `public ObservableList<ObservableMessageWithPriority> setUpListView(boolean isDeveloperInformationActive)` and set the returned list in the view class.

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleViewModel.java
```diff
@@ -0,0 +1,147 @@
...
+                            } else {
+                                getStyleClass().add("log");
+                            }
+                        } else {
+                            setText("");
+                        }
+                    }
+                };
+            }
+        });
```
boceckts: extract to a new method like setUpListView and move to the view class.

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleViewModel.java
```diff
@@ -0,0 +1,147 @@
...
+            ObservableList<ObservableMessageWithPriority> masterData = ObservableMessages.INSTANCE.messagesPropety();
+            String listViewContent = "";
+            for (ObservableMessageWithPriority message : masterData) {
+                listViewContent += message.getMessage() + System.lineSeparator();
+            }
+            // format the contents of listview in Issue Description
+            String issueDetails = ("<details>\n" + "<summary>" + Localization.lang("Detail information:") + "</summary>\n```\n" + listViewContent + "\n```\n</details>");
+            new ClipBoardManager().setClipboardContents(issueDetails);
+        } catch (IOException e) {
+            e.printStackTrace();
```
boceckts: Don't print stacktraces, use the Logger class to log errors.

src/main/java/net/sf/jabref/logic/error/MessagePriority.java
```diff
@@ -0,0 +1,5 @@
+package net.sf.jabref.logic.error;
```
boceckts: Add the copyright notice to the beginning of every file you created.

src/main/java/net/sf/jabref/logic/error/ObservableMessageWithPriority.java
```diff
@@ -0,0 +1,48 @@
+package net.sf.jabref.logic.error;
+
+import javafx.beans.property.BooleanProperty;
+import javafx.beans.property.SimpleBooleanProperty;
+
+public class ObservableMessageWithPriority {
```
boceckts: Please add a brief description as java doc comment.

src/main/java/net/sf/jabref/logic/error/ObservableMessageWithPriority.java
```diff
@@ -0,0 +1,48 @@
...
+
+    private String message;
+
+    private MessagePriority priority;
+
+    private BooleanProperty isFiltered = new SimpleBooleanProperty();
```
boceckts: Remove empty lines between fields.

build.gradle
```diff
@@ -117,6 +117,8 @@ dependencies {
     compile 'org.apache.logging.log4j:log4j-api:2.6.2'
     compile 'org.apache.logging.log4j:log4j-core:2.6.2'
 
+    compile 'de.saxsys:mvvmfx:1.3.0'
+
```
boceckts: I can't see any usage for this..

src/main/java/net/sf/jabref/logic/logging/Cache.java
```diff
@@ -38,9 +41,8 @@ public Cache(int capacity) {
         this.capacity = capacity;
     }
 
-    public synchronized String get() {
-        ensureCacheIsFresh();
```
boceckts: Is this really not needed anymore (also the synchronized)? If so, you should delete the method `ensureCacheIsFresh()` as it is now unused.

src/main/java/net/sf/jabref/logic/logging/ObservableMessages.java
```diff
@@ -0,0 +1,25 @@
+package net.sf.jabref.logic.logging;
+
+import javafx.beans.Observable;
+import javafx.beans.property.ListProperty;
+import javafx.beans.property.SimpleListProperty;
+import javafx.collections.FXCollections;
+
+import net.sf.jabref.logic.error.ObservableMessageWithPriority;
+
+public enum ObservableMessages {
```
boceckts: Please also add a java doc comment.

src/main/resources/l10n/JabRef_en.properties
```diff
@@ -2238,10 +2238,24 @@ Copied_version_to_clipboard=Copied_version_to_clipboard
...
+Create_Issue_successful!=Create_Issue_successful!
+Your_issue_was_created_in_your_browser!=Your_issue_was_created_in_your_browser!
+The_log_and_exception_information_was_copied_to_your_clipboard.=The_log_and_exception_information_was_copied_to_your_clipboard.
+Please_paste_this_information_(with_Ctrl+V)_in_the_issue_description!=Please_paste_this_information_(with_Ctrl+V)_in_the_issue_description!
+
+
+Developer_information=Developer_information
+Create_Issue=Create_Issue
+*incl._log_and_error_details=*incl._log_and_error_details
+Details_infomation=Details_infomation:
```
boceckts: Please rework some of the language strings keeping in mind that:
- in english you usually write in lower case except at the start of an sentence and proper names such as GitHub. 
- exclamation marks ("!") are rather intusive and shouldn't be used too often, maybe replace some dot (".")
- use the past form if you display if you want to use it in error/info/warning dialogs as something has already happened
- can `Detail_information\:` and `Details_information` be merged?
- the last key and translation don't match

src/main/resources/l10n/Menu_en.properties
```diff
@@ -31,6 +31,7 @@ Disable_highlight_groups_matching_entries=Disable_highlight_groups_matching_entr
 
 # Help
 Online_help=Online_help
+Show_developer_information=Show_developer information
```
boceckts: Underscore ("_") is missing.

boceckts:
Please fix this bug:
- When you activate the "developer information" button and then click "Copy Log" the exceptions will get hidden even though the "developer information" toggle button is still active.
- Pressing the "developer information" button twice will show the exceptions again.

boceckts:
Please update your screenshots.

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleViewModel.java
```diff
@@ -0,0 +1,147 @@
...
+                            Localization.lang("Please paste this information (with Ctrl+V) in the issue description!"));
+            JabRefDesktop.openBrowser(CREATEISSUEURL + issueTitle + issueBody);
+            //get contents of listview
+            ObservableList<ObservableMessageWithPriority> masterData = ObservableMessages.INSTANCE.messagesPropety();
+            String listViewContent = "";
+            for (ObservableMessageWithPriority message : masterData) {
+                listViewContent += message.getMessage() + System.lineSeparator();
+            }
+            // format the contents of listview in Issue Description
+            String issueDetails = ("<details>\n" + "<summary>" + Localization.lang("Detail information:") + "</summary>\n```\n" + listViewContent + "\n```\n</details>");
```
boceckts: You don't need localization here as this gets pasted on github.

boceckts:
Please fix bug:
- Open the developer information console
- Click show "Developer information" toggle button to show exceptions
- Click "Ok" button
- Open the developer information console
"Developer information" toggle button is now deactivated again but the list view still shows the exception and the colour of the exception is black instead of red.

src/main/resources/net/sf/jabref/gui/errorconsole/ErrorConsole.fxml
```diff
@@ -0,0 +1,31 @@
...
+            <bottom>
+                <ButtonBar styleClass="custom-buttons" BorderPane.alignment="CENTER">
+                    <BorderPane.margin>
+                        <Insets top="10.0" />
+                    </BorderPane.margin>
+                    <buttons>
+                        <Button fx:id="closeButton" cancelButton="true" defaultButton="true" mnemonicParsing="false" onAction="#closeErrorDialog" text="Ok" textAlignment="CENTER" />
+                        <Button fx:id="copyLogButton" mnemonicParsing="false" onAction="#copyLogButton" text="Copy Log" />
+                        <ToggleButton fx:id="developerButton" mnemonicParsing="false" text="Developer information" />
+                        <Button fx:id="createIssueButton" mnemonicParsing="false" onAction="#createIssueButton" text="Create Issue" />
```
boceckts: Please use localized strings with by adding a percent sign ("%") in front of the strings. E.g. text="%Ok"

oscargus:
Any idea for a better icon? As pointed out in #1742 there is much more information than just errors here. It might even be better to skip the icon. (What about the information icon?)

tobiasdiez:
Some suggestions:
- I would use the [Console image](https://mdn.mozillademos.org/files/5597/web-console-commandline-highlighted.png) instead of the red error symbol. Also move it above the list and add some description left to it. In spirit "we now give you an insight into the inner workings of JabRef's brain. This might help to diagonize the root of problems. Please use the button below to inform the developers about an issue.
- Rename "create issue" to "report issue". Move "copy log" button to the left to group it with the other buttons there. Rename "ok" to "close".
- Do we really need a way to hide errors and warnings? I think every log entry should be shown to the user (i.e. no "developer information button"). I know, you spent a lot of time to implement it but I just don't see any advantages in having such a button.
- Add a small non-obtrusive icon in front of every line which indicates if it is an error / warning / info. Similar to https://mdn.mozillademos.org/files/5597/web-console-commandline-highlighted.png.

src/main/resources/l10n/JabRef_en.properties
```diff
@@ -2239,6 +2239,14 @@ Copied_version_to_clipboard=Copied_version_to_clipboard
 BibTeX_key=BibTeX_key
 Message=Message
 
+Developer_information=Developer_information
+We_now_give_you_an_insight_into_the_inner_workings_of_JabRef\'s_brain._=We_now_give_you_an_insight_into_the_inner_workings_of_JabRef\'s_brain._
+This_might_help_to_diagonalize_the_root_of_problems._=This_might_help_to_diagonalize_the_root_of_problems._
```
boceckts: Please fix typo, "diagonalize" should be "diagnose"

src/main/resources/l10n/JabRef_en.properties
```diff
@@ -2239,6 +2239,14 @@ Copied_version_to_clipboard=Copied_version_to_clipboard
 BibTeX_key=BibTeX_key
 Message=Message
 
+Developer_information=Developer_information
+We_now_give_you_an_insight_into_the_inner_workings_of_JabRef\'s_brain._=We_now_give_you_an_insight_into_the_inner_workings_of_JabRef\'s_brain._
```
boceckts: Why is there an underscore ("_") at the end? Please remove it.

src/main/resources/l10n/JabRef_en.properties
boceckts: Also remove underscore.

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleView.java
```diff
@@ -80,14 +90,91 @@ private void copyLogButton() {
...
+        ColumnConstraints graphicColumn = new ColumnConstraints();
+        graphicColumn.setFillWidth(false);
+        graphicColumn.setHgrow(Priority.NEVER);
+        ColumnConstraints textColumn = new ColumnConstraints();
+        textColumn.setFillWidth(true);
+        textColumn.setHgrow(Priority.ALWAYS);
+        headerGrid.getColumnConstraints().setAll(graphicColumn, textColumn);
+        headerGrid.setPadding(new Insets(10));
+
+        Image image = new Image("https://cdn2.iconfinder.com/data/icons/windows-8-metro-style/512/console.png");
```
boceckts: Don't use graphics from external resources such as urls. Have a look at #1390 to see how to use the material design icons in JavaFX.

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleView.java
```diff
@@ -80,14 +90,91 @@ private void copyLogButton() {
...
+                            setText(omp.getMessage());
+                            getStyleClass().clear();
+                            if (omp.getPriority() == MessagePriority.HIGH) {
+                                setGraphic(listImageResize("http://www.iconsdb.com/icons/preview/red/info-xxl.png"));
+                                getStyleClass().add("exception");
+                            } else if (omp.getPriority() == MessagePriority.MEDIUM) {
+                                setGraphic(listImageResize("http://www.iconsdb.com/icons/preview/black/info-xxl.png"));
+                                getStyleClass().add("output");
+                            } else {
+                                setGraphic(listImageResize("http://www.iconsdb.com/icons/preview/royal-blue/info-xxl.png"));
```
boceckts: Same as with the Console image.

src/main/resources/l10n/JabRef_en.properties
```diff
@@ -2239,6 +2239,14 @@ Copied_version_to_clipboard=Copied_version_to_clipboard
...
+This_might_help_to_diagonalize_the_root_of_problems._=This_might_help_to_diagonalize_the_root_of_problems._
+Please_use_the_button_below_to_inform_the_developers_about_an_issue.=Please_use_the_button_below_to_inform_the_developers_about_an_issue.
+Log_is_copied=Log_is_copied
+Copy_Log=Copy_Log
+Report_Issue=Report_Issue
+Report_Issue_on_GitHub_is_successful=Report_Issue_on_GitHub_is_successful
```
boceckts: Please reword to "Issue on GitHub successfully reported".

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleView.java
```diff
@@ -0,0 +1,180 @@
...
+import net.sf.jabref.logic.error.ObservableMessageWithPriority;
+import net.sf.jabref.logic.l10n.Localization;
+import net.sf.jabref.logic.logging.ObservableMessages;
+
+import com.airhacks.afterburner.views.FXMLView;
+
+public class ErrorConsoleView extends FXMLView {
+
+    private final ErrorConsoleViewModel errorViewModel = new ErrorConsoleViewModel();
+    private BooleanProperty isDeveloperButtonEnable = new SimpleBooleanProperty();
```
boceckts: I guess this isn't needed anymore, please remove it.

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleView.java
boceckts: Since you already set style classes to the rows and specify their style in the stylesheet, maybe you can also set the icon there?

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Wed Aug 03 11:49:15 CEST 2016
+#Thu Aug 18 17:00:33 CEST 2016
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.14.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-2.14.1-all.zip
```
boceckts: Why did this change?

boceckts:
Some minor remarks and a bigger one with the icons, once they are fixed I'm fine with this.
All in all I really like how the console looks now. Good job 👍

src/main/java/net/sf/jabref/gui/FXAlert.java
```diff
@@ -71,26 +72,29 @@ public void windowGainedFocus(WindowEvent e) {
         }
     };
 
-
-    public FXAlert(AlertType type, String title, Image image) {
```
tobiasdiez: I would leave all the old constructors and call the new overloads with `isModal=true` since this looks like the default behavior. For example:
`public FXAlert(AlertType type, String title) { this(type, title, true); }`.

src/main/java/net/sf/jabref/gui/actions/ErrorConsoleAction.java
```diff
@@ -39,57 +36,14 @@
...
-    private final JFrame frame;
-    private final StreamEavesdropper streamEavesdropper;
-    private final Cache cache;
-
-    public ErrorConsoleAction(JFrame frame, StreamEavesdropper streamEavesdropper, Cache cache) {
-        super(Localization.menuTitle("Show error console"));
-        this.streamEavesdropper = streamEavesdropper;
-        this.cache = cache;
+    public ErrorConsoleAction(StreamEavesdropper streamEavesdropper, Cache cache) {
+        super(Localization.menuTitle("Show developer information"));
```
tobiasdiez: "Developer information" is better then "error console", but maybe one would expect to get information about becoming a developer for JabRef (instead of information for us developer). What about a more general `Troubleshoot problems` or `View event log` or `Problem reporting`?

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleView.java
```diff
@@ -0,0 +1,141 @@
...
+    @FXML
+    private Label descriptionLabel;
+
+    public ErrorConsoleView() {
+        super();
+        bundle = Localization.getMessages();
+    }
+
+    public void show() {
+        FXAlert errorConsole = new FXAlert(AlertType.ERROR, Localization.lang("Developer information"), false);
```
tobiasdiez: See above regarding "developer info"

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleView.java
```diff
@@ -0,0 +1,141 @@
...
+    @FXML
+    private void initialize() {
+        ButtonBar.setButtonData(copyLogButton, ButtonBar.ButtonData.LEFT);
+        ButtonBar.setButtonData(createIssueButton, ButtonBar.ButtonData.LEFT);
+
+        ObservableList<ObservableMessageWithPriority> masterData = ObservableMessages.INSTANCE.messagesPropety();
+        listViewStyle();
+        allMessage.setItems(masterData);
+
+        //set console icon instead the error icon and insert a description
```
tobiasdiez: I think the convention for comments is `// With space and first capital letter`

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleView.java
```diff
@@ -0,0 +1,141 @@
...
+        Text graphic = new Text(IconTheme.JabRefIcon.CONSOLE.getCode());
```
tobiasdiez: In one of the recent FX-PRs (I hope it is already merged), a new method in `JabRefIcon` was introduced which gives you directly the `Text` control with the right style class.

boceckts: Unfortunately not yet merged.
https://github.com/JabRef/jabref/pull/1526/files#diff-714bc6a7e9d4aca70ed8b733a7c75ed8R204

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleView.java
```diff
@@ -0,0 +1,141 @@
...
+        graphic.getStyleClass().add("icon");
+        String headerLabel = Localization.lang("We now give you an insight into the inner workings of JabRef's brain.") + " "
```
tobiasdiez: "give you insight" (without "an")

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleView.java
```diff
@@ -0,0 +1,141 @@
...
+                + Localization.lang("This might help to diagonalize the root of problems.") + " " + System.lineSeparator()
```
tobiasdiez: These information might be helpful to diagnose the root cause of a problem.

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleView.java
```diff
@@ -0,0 +1,141 @@
...
+                + Localization.lang("Please use the button below to inform the developers about an issue.");
```
tobiasdiez: "Please feel free to inform ...".
Maybe make the part "inform the developers" a hyperlink-like button, which also invokes `errorViewModel.reportIssue();`

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleView.java
```diff
@@ -0,0 +1,141 @@
...
+        descriptionLabel.setText(headerLabel);
```
tobiasdiez: Is there a reason why you set the text here in the view instead of in the FXML file?

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleView.java
```diff
@@ -0,0 +1,141 @@
...
+                    public ListCell<ObservableMessageWithPriority> call(
+                            ListView<ObservableMessageWithPriority> listView) {
+                        return new ListCell<ObservableMessageWithPriority>() {
+
+                            @Override
+                            public void updateItem(ObservableMessageWithPriority omp, boolean empty) {
+                                super.updateItem(omp, empty);
+                                if (omp != null) {
+                                    setText(omp.getMessage());
+                                    getStyleClass().clear();
```
tobiasdiez: Why do you need to clear the style class?

src/main/resources/net/sf/jabref/gui/errorconsole/ErrorConsole.css
```diff
@@ -0,0 +1,54 @@
...
+.info-section {
+	-fx-padding: 10.0px;
+}
+
+.info-section .icon {
+	-fx-font-size: 48.0;
+}
+
+.list-cell {
+ -fx-background-color: white;
```
tobiasdiez: indention (here and in some other places)

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleView.java
```diff
@@ -0,0 +1,141 @@
...
+                                    Text graphic = new Text();
+                                    graphic.getStyleClass().add("icon");
+                                    if (omp.getPriority() == MessagePriority.HIGH) {
```
tobiasdiez: Use switch statement.

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleView.java
```diff
@@ -0,0 +1,141 @@
...
+        // handler for listCell appearance (example for exception Cell)
+        allMessage.setCellFactory(
+                new Callback<ListView<ObservableMessageWithPriority>, ListCell<ObservableMessageWithPriority>>() {
+                    @Override
+                    public ListCell<ObservableMessageWithPriority> call(
+                            ListView<ObservableMessageWithPriority> listView) {
+                        return new ListCell<ObservableMessageWithPriority>() {
+
+                            @Override
+                            public void updateItem(ObservableMessageWithPriority omp, boolean empty) {
```
tobiasdiez: rename omp -> message

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleView.java
```diff
@@ -0,0 +1,141 @@
...
+                                super.updateItem(omp, empty);
+                                if (omp != null) {
+                                    setText(omp.getMessage());
+                                    getStyleClass().clear();
+                                    Text graphic = new Text();
+                                    graphic.getStyleClass().add("icon");
+                                    if (omp.getPriority() == MessagePriority.HIGH) {
+                                        getStyleClass().add("exception");
+                                        graphic.setText(IconTheme.JabRefIcon.INTEGRITY_FAIL.getCode());
+                                    } else if (omp.getPriority() == MessagePriority.MEDIUM) {
```
tobiasdiez: Refactor, by moving some of the code to ObservableMessageWithPriority.
It should be as simple as `getStyleClass.add(omp.getStyleClass())` and `setGraphic(omp.getIcon())`.

src/main/java/net/sf/jabref/logic/error/MessagePriority.java
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * This enum will be define for three different priority of message
+ * <ul>
+ *     <li>LOW is for log entry
+ *     <li>MEDIUM is for output entry
+ *     <li>HIGH is for exception entry
+ * </ul>
+ */
+public enum MessagePriority {
+    LOW, MEDIUM, HIGH;
```
tobiasdiez: Since they represent not really a priority, but more a type of message, I would rename:
priority -> type and low -> log, medium -> output, high -> exception.

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleView.java
tobiasdiez: Now it is merged 😄

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleViewModel.java
```diff
@@ -0,0 +1,89 @@
...
+
+    private static final Log LOGGER = LogFactory.getLog(ErrorConsoleViewModel.class);
+    private final String REPORTISSUEURL = "https://github.com/JabRef/jabref/issues/new";
+    private final DateFormat dateFormat = new SimpleDateFormat("yyyyMMddHHmmss");
+    private final Date date = new Date();
+
+    // handler for copy of Log Entry in the List by click of Copy Log Button
+    public void copyLog() {
+        ObservableList<ObservableMessageWithPriority> masterData = ObservableMessages.INSTANCE.messagesPropety();
+        masterData.forEach(message -> message.setIsFiltered(message.getPriority() != MessagePriority.LOW));
```
tobiasdiez: I think there is no need to filter the before copying it to the clipboard.

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleViewModel.java
```diff
@@ -0,0 +1,89 @@
...
+        FilteredList<ObservableMessageWithPriority> filteredLowPriorityList = new FilteredList<>(masterData, t -> !t.isFilteredProperty().get());
+        String logContentCopy = "";
```
tobiasdiez: Use StringBuilder

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleViewModel.java
```diff
@@ -0,0 +1,89 @@
...
+
+        for (ObservableMessageWithPriority message : filteredLowPriorityList) {
+            logContentCopy += message.getMessage() + System.lineSeparator();
+        }
+        new ClipBoardManager().setClipboardContents(logContentCopy);
+        JabRefGUI.getMainFrame().output(Localization.lang("Log is copied"));
```
tobiasdiez: "Log copied to clipboard."

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleViewModel.java
```diff
@@ -0,0 +1,89 @@
...
+import org.apache.commons.logging.LogFactory;
+
+public class ErrorConsoleViewModel {
+
+    private static final Log LOGGER = LogFactory.getLog(ErrorConsoleViewModel.class);
+    private final String REPORTISSUEURL = "https://github.com/JabRef/jabref/issues/new";
+    private final DateFormat dateFormat = new SimpleDateFormat("yyyyMMddHHmmss");
+    private final Date date = new Date();
+
+    // handler for copy of Log Entry in the List by click of Copy Log Button
```
tobiasdiez: Use JavaDoc-style comments for method descriptions.

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleView.java
boceckts: great :smile: 
@motokito please also adjust the changelog according to the new pattern

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleViewModel.java
```diff
@@ -0,0 +1,89 @@
...
+            String info = String.format("JabRef %s%n%s %s %s %nJava %s\n\n", Globals.BUILD_INFO.getVersion(), BuildInfo.OS,
+                    BuildInfo.OS_VERSION, BuildInfo.OS_ARCH, BuildInfo.JAVA_VERSION);
+            String issueTitle = "?title=" + URLEncoder.encode("Automatic Bug Report-" + dateFormat.format(date), "UTF-8");
+            String issueBody = "&body=" + URLEncoder.encode(info, "UTF-8");
+            JabRefGUI.getMainFrame().output(Localization.lang("Issue on GitHub successfully reported."));
+            FXDialogs.showInformationDialogAndWait("Report Issue successful",
+                    "Your issue was reported in your browser." + "\n\n" +
+                            "The log and exception information was copied to your clipboard." + "\n\n" +
+                            "Please paste this information (with Ctrl+V) in the issue description.");
+            JabRefDesktop.openBrowser(REPORTISSUEURL + issueTitle + issueBody);
```
tobiasdiez: Use apache's URIBuilder to create the URL.

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleViewModel.java
```diff
@@ -0,0 +1,89 @@
...
+
+    // handler for report Issues on GitHub by click of Report Issue Button
+    public void reportIssue() {
+        try {
+            String info = String.format("JabRef %s%n%s %s %s %nJava %s\n\n", Globals.BUILD_INFO.getVersion(), BuildInfo.OS,
+                    BuildInfo.OS_VERSION, BuildInfo.OS_ARCH, BuildInfo.JAVA_VERSION);
+            String issueTitle = "?title=" + URLEncoder.encode("Automatic Bug Report-" + dateFormat.format(date), "UTF-8");
+            String issueBody = "&body=" + URLEncoder.encode(info, "UTF-8");
+            JabRefGUI.getMainFrame().output(Localization.lang("Issue on GitHub successfully reported."));
+            FXDialogs.showInformationDialogAndWait("Report Issue successful",
```
tobiasdiez: "Issue report successful"

src/main/java/net/sf/jabref/logic/error/ObservableMessageWithPriority.java
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * This class will be get the message and handle it with a priority, which will be define in MessagePriority.java
+ * <ul>
+ * <li>MessagePriority.LOW is define for log entries
+ * <li>MessagePriority.MEDIUM is define for output entries
+ * <li>MessagePriority.MEDIUM is define for exception entries
+ * </ul>
+ */
+public class ObservableMessageWithPriority {
```
tobiasdiez: Rename to LogMessage?

src/main/java/net/sf/jabref/logic/error/ObservableMessageWithPriority.java
```diff
@@ -0,0 +1,69 @@
...
+    with this program; if not, write to the Free Software Foundation, Inc.,
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+package net.sf.jabref.logic.error;
+
+import javafx.beans.property.BooleanProperty;
+import javafx.beans.property.SimpleBooleanProperty;
+
+/**
+ * This class will be get the message and handle it with a priority, which will be define in MessagePriority.java
```
tobiasdiez: you can reference a different class using `@link` in javadoc.

src/main/java/net/sf/jabref/logic/error/ObservableMessageWithPriority.java
```diff
@@ -0,0 +1,69 @@
...
+
+    public String getMessage() {
+        return message;
+    }
+
+    public void setMessage(String message) {
+        this.message = message;
+    }
+
+    public boolean getIsFiltered() {
```
tobiasdiez: Again, I think the filtering business is no longer needed, right?

src/main/java/net/sf/jabref/logic/logging/ObservableMessages.java
```diff
@@ -0,0 +1,30 @@
...
+import javafx.collections.FXCollections;
+
+import net.sf.jabref.logic.error.ObservableMessageWithPriority;
+
+/**
+ * This ObservableMessages enum will be save all message entries in a ListProperty.
+ * <p></p>
+ * This list will be use late in ErrorConsoleViewModel class to filter message after their priority
+ */
+public enum ObservableMessages {
```
tobiasdiez: why is this an enum and not a normal class?

src/main/java/net/sf/jabref/logic/logging/ObservableMessages.java
```diff
@@ -0,0 +1,30 @@
...
+
+    INSTANCE;
+
+    private final ListProperty<ObservableMessageWithPriority> messages = new SimpleListProperty<>(FXCollections.observableArrayList((item ->
+            new Observable[]{item.isFilteredProperty()})));
+
+    public ListProperty<ObservableMessageWithPriority> messagesPropety() {
```
tobiasdiez: "property"

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleView.java
motokito: I don't think that's needed, because of the sentence already stating where you can report the issue. Maybe change the sentence to: "Please feel free to inform the developers with the below button "Report Issue"." this would clearify where to click.

tobiasdiez: Yes, but I don't like the "click the button below"-part. It feels a bit outdated similar to the old-school "if you want something to happen, then click >here< " text in websites. 
But if there no easy solution to turn the text into a hyperlink, then its also no big deal (but in this case please also remove the "button below" part)

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleView.java
motokito: The reason it is 3 sentence and it will be throw a IllegalStateException, if i try to insert this sentences in FXML.
Like (%We now give you insight into the inner workings of JabRef's brain. These information might be helpful to diagnose the root cause of a problem. Please use the button below to inform the developers about an issue.)

tobiasdiez: Did you tried the text without the fullstops / points and apostrophe? What was the error message? I can't see why a longer text might lead to problems.

motokito: Now i have tried this text in FXML file to implement without error. But this text will display on one line. So i can not format this sentence in two line in FXML. If you know how to do it, then please give me a tip :smirk:

* take a look to this screenshot:

2016-09-07_14h06_54

tobiasdiez: I have not tried it but accordingly to http://stackoverflow.com/questions/15977295/control-for-displaying-multiline-text you can specify line breaks by `\n`. But probably the best way is to set a maxWidth and wrapText =true (http://stackoverflow.com/a/38622468/873661 but in the fxml of course).

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleViewModel.java
motokito: i should get only the log entry or ? i must get this about a filter. Or how i should implement it ? 😄

tobiasdiez: Just combine all messages in `masterData` to a string? I don't see the need to filter some messages.

motokito: i will copy all message incl. exception and output or ? because masterdata is a list of all message with different priority 😄

tobiasdiez: yes, especially the exceptions and output are important.

motokito:
@tobiasdiez: I have merged and done feedback

src/main/java/net/sf/jabref/gui/FXDialogs.java
```diff
@@ -119,14 +119,14 @@ public static void showErrorDialogAndWait(String title, String content) {
      */
     public static Optional<ButtonType> showCustomDialogAndWait(String title, DialogPane contentPane,
             ButtonType... buttonTypes) {
-        FXAlert alert = new FXAlert(AlertType.NONE, title);
+        FXAlert alert = new FXAlert(AlertType.NONE, title, true);
```
tobiasdiez: `true` is default, right? If so, remove it.

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleView.java
```diff
@@ -0,0 +1,142 @@
...
+        FXAlert errorConsole = new FXAlert(AlertType.ERROR, Localization.lang("Event log"), false);
+        DialogPane pane = (DialogPane) this.getView();
+        errorConsole.setDialogPane(pane);
+        errorConsole.setResizable(true);
+        errorConsole.show();
+    }
+
+    @FXML
+    private void initialize() {
+        ButtonBar.setButtonData(copyLogButton, ButtonBar.ButtonData.LEFT);
```
tobiasdiez: Please move this to FXML: http://stackoverflow.com/a/32427404/873661

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleView.java
```diff
@@ -0,0 +1,142 @@
...
+                                if (logMessageWithPriority != null) {
+                                    setText(logMessageWithPriority.getMessage());
+                                    Text graphic = new Text();
+                                    graphic.getStyleClass().add("icon");
+
+                                    MessagePriority prio = logMessageWithPriority.getPriority();
+                                    switch (prio) {
+                                        case HIGH:
+                                            getStyleClass().add("exception");
+                                            graphic.setText(IconTheme.JabRefIcon.INTEGRITY_FAIL.getCode());
```
tobiasdiez: `setGraphic(IconTheme....getGraphicNode())`

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleView.java
```diff
@@ -0,0 +1,142 @@
...
+        errorConsole.setResizable(true);
+        errorConsole.show();
+    }
+
+    @FXML
+    private void initialize() {
+        ButtonBar.setButtonData(copyLogButton, ButtonBar.ButtonData.LEFT);
+        ButtonBar.setButtonData(createIssueButton, ButtonBar.ButtonData.LEFT);
+
+        ObservableList<LogMessageWithPriority> masterData = LogMessage.getInstance().messagesProperty();
```
tobiasdiez: The view shouldn't know about the datasource (in this case `LogMessage`) but ask the viewmodel. Could you please add a pass-through getter in errorViewModel.
Also inline `masterData`, i.e. `allMessages.setItems(errorViewModel.getMessages())`

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleView.java
```diff
@@ -0,0 +1,142 @@
...
+    private final ErrorConsoleViewModel errorViewModel = new ErrorConsoleViewModel();
+
+    @FXML
+    private Button closeButton;
+    @FXML
+    private Button copyLogButton;
+    @FXML
+    private Button createIssueButton;
+    @FXML
+    private ListView<LogMessageWithPriority> allMessage;
```
tobiasdiez: add plural `s`

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleViewModel.java
```diff
@@ -0,0 +1,92 @@
...
+    /**
+     * Handler for report Issues on GitHub by click of Report Issue Button
+     */
+    public void reportIssue() {
+        try {
+            String issueTitle = "Automatic Bug Report-" + dateFormat.format(date);
+            String issueBody = String.format("JabRef %s%n%s %s %s %nJava %s\n\n", Globals.BUILD_INFO.getVersion(), BuildInfo.OS,
+                    BuildInfo.OS_VERSION, BuildInfo.OS_ARCH, BuildInfo.JAVA_VERSION);
+            JabRefGUI.getMainFrame().output(Localization.lang("Issue on GitHub successfully reported."));
+            FXDialogs.showInformationDialogAndWait("Issue report successful",
```
tobiasdiez: this should be localized

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleViewModel.java
```diff
@@ -0,0 +1,92 @@
...
+                    "Your issue was reported in your browser." + "\n\n" +
+                            "The log and exception information was copied to your clipboard." + "\n\n" +
+                            "Please paste this information (with Ctrl+V) in the issue description.");
+            URIBuilder uriBuilder = new URIBuilder().setScheme("https").setHost("github.com").setPath("/JabRef/jabref/issues/new")
+                    .setParameter("title", issueTitle).setParameter("body", issueBody);
+            JabRefDesktop.openBrowser(uriBuilder.build().toString());
+            // Get contents of listview
+            ObservableList<LogMessageWithPriority> masterData = LogMessage.getInstance().messagesProperty();
```
tobiasdiez: I think you use exactly the same code in `copyLog`. Please extract the code in a new private method `getMessagesAsString`

src/main/java/net/sf/jabref/gui/help/AboutDialogView.java
```diff
@@ -31,7 +31,7 @@ public AboutDialogView() {
     }
 
     public void show() {
-        FXAlert aboutDialog = new FXAlert(AlertType.INFORMATION, Localization.lang("About JabRef"));
+        FXAlert aboutDialog = new FXAlert(AlertType.INFORMATION, Localization.lang("About JabRef"), true);
```
tobiasdiez: Remove `true`

src/main/java/net/sf/jabref/logic/error/LogMessageWithPriority.java
```diff
@@ -0,0 +1,65 @@
...
+
+    public String getMessage() {
+        return message;
+    }
+
+    public void setMessage(String message) {
+        this.message = message;
+    }
+
+    public void setIsFiltered(boolean isFiltered) {
```
tobiasdiez: Is this still used?

src/main/java/net/sf/jabref/logic/error/MessagePriority.java
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * This enum will be define for three different type of message
+ * <ul>
+ * <li>log
+ * <li>output
+ * <li>exception
+ * </ul>
+ */
+public enum MessagePriority {
+    LOW, MEDIUM, HIGH;
```
tobiasdiez: As I said before, please rename
LOW -> LOG
MEDIUM -> OUTPUT
HIGH -> EXCEPTION

and MessagePriority -> MessageType as well as LogMessageWithPriority -> LogMessage

src/main/java/net/sf/jabref/logic/error/TeeStream.java
```diff
@@ -10,17 +12,24 @@
...
         this.outStream = out2;
+        this.priority = priority;
     }
 
     @Override
     public void write(byte[] buf, int off, int len) {
         try {
             super.write(buf, off, len);
             outStream.write(buf, off, len);
+            String s = new String(buf, off, len);
```
tobiasdiez: I think `TeeStream` is the wrong place to add this code. There should be a central error logger somewhere. Sorry for being a bit unspecific about this, but I right now I have no time for digging into the details.

src/main/java/net/sf/jabref/logic/logging/LogMessage.java
```diff
@@ -0,0 +1,37 @@
...
+import javafx.collections.FXCollections;
+
+import net.sf.jabref.logic.error.LogMessageWithPriority;
+
+/**
+ * This LogMessage will be save all message entries in a ListProperty.
+ * <p></p>
+ * This list will be use late in ErrorConsoleViewModel to filter message after their priority
+ */
+public class LogMessage {
```
tobiasdiez: Rename LogMessage_s_

tobiasdiez:
Thanks for the update. The code looks better now. I have just a few more remarks, then this can be merged. @koppor pls do the final review

src/main/java/net/sf/jabref/logic/logging/LogMessage.java
```diff
@@ -0,0 +1,37 @@
...
+import javafx.beans.property.ListProperty;
+import javafx.beans.property.SimpleListProperty;
+import javafx.collections.FXCollections;
+
+import net.sf.jabref.logic.error.LogMessageWithPriority;
+
+/**
+ * This LogMessage will be save all message entries in a ListProperty.
+ * <p></p>
+ * This list will be use late in ErrorConsoleViewModel to filter message after their priority
```
koppor: Please fix English. Proposal:

This call saves all messages in a list. The list is used in ErrorConsoleViewModel to filter messages according to their priority.

src/main/resources/net/sf/jabref/gui/errorconsole/ErrorConsole.fxml
```diff
@@ -0,0 +1,33 @@
...
+                    </buttons>
+                </ButtonBar>
+            </bottom>
+            <center>
+                <ListView fx:id="allMessage" styleClass="list-content" BorderPane.alignment="CENTER" />
+            </center>
+        </BorderPane>
+    </content>
+   <header>
+      <Label fx:id="descriptionLabel" styleClass="info-section" text="%We now give you insight into the inner workings of JabRef's brain. These information might be helpful to diagnose the root cause of a problem. Please feel free to inform the developers about an issue." wrapText="true" />
```
koppor: Do not use "JabRef's brain". Use "JabRef's internals".

Do not use "These information" use "This information".

motokito:
@tobiasdiez @koppor: feedbacks are done. Only this teestream thing, i am not sure how to fix it 😄

koppor:
Maybe someone else from @JabRef/stupro can help. @bartsch-dev maybe? It's a kind of architectural decision.

tobiasdiez:
Try to use StreamEavesdropper instead of TeeStream

boceckts:
@tobiasdiez Is this what you meant with moving the code to the streamEavesdropper?

src/main/java/net/sf/jabref/logic/error/StreamEavesdropper.java
```diff
@@ -30,13 +34,35 @@ public static StreamEavesdropper eavesdropOnSystem() {
...
     public TeeStream getErrStream() {
-        PrintStream consoleErr = new PrintStream(errByteStream);
-        return new TeeStream(consoleErr, systemErr, MessageType.EXCEPTION);
+        PrintStream consoleErr = new PrintStream(errByteStream){
+            @Override
+            public void write(byte[] buf, int off, int len) {
+                super.write(buf, off, len);
+                String s = new String(buf, off, len);
+                addToLog(s, MessageType.EXCEPTION
+                );
```
tobiasdiez: Formatting.

src/main/java/net/sf/jabref/logic/error/StreamEavesdropper.java
```diff
@@ -30,13 +34,35 @@ public static StreamEavesdropper eavesdropOnSystem() {
...
     }
 
     public TeeStream getErrStream() {
-        PrintStream consoleErr = new PrintStream(errByteStream);
-        return new TeeStream(consoleErr, systemErr, MessageType.EXCEPTION);
+        PrintStream consoleErr = new PrintStream(errByteStream){
+            @Override
+            public void write(byte[] buf, int off, int len) {
+                super.write(buf, off, len);
+                String s = new String(buf, off, len);
```
tobiasdiez: rename s -> message (also above)

src/main/java/net/sf/jabref/logic/error/StreamEavesdropper.java
```diff
@@ -30,13 +34,35 @@ public static StreamEavesdropper eavesdropOnSystem() {
...
+                addToLog(s, MessageType.EXCEPTION
+                );
+            }
+        };
+        return new TeeStream(consoleErr, systemErr);
```
tobiasdiez: If I understand it correctly, then everything from systemErr is also written to consoleErr. With your implementation all messages to consoleErr are posted to the cache (`addToLog`).
So far so good. But they also get posted as well as to the underlying `errByteStream`, which is no longer required (is it?). Thus could you remove the byte stream part?

tobiasdiez:
Yes this looks better.
So now I had a look at the code. If I understand it correctly, then there already exist two ways of listen to logger messages:
- GuiAppender: listens to everything which goes through the Log4j API (i.e. `LOGGER.log(...)`) and writes it to Cache
- StreamEavesdropper: listens to everything which goes through the System streams (`System.out/err`) and writes it to some streams

And you added:
- LogMessages: listens to everything and writes it some some internal list

Could you refactor the code so that everything is only written ̀to the internal list of  `LogMessages`, i.e. get rid of the Cache class and the printstreams in `StreamEavesdropper`. (And please add code comments to the classes, it was hard to understand what they actually do)

Appendum for myself:
Since there is essentially no code which still uses `System.out/err.printLn`, I would actually try to get rid of the `StreamEavesdropper` completely. Only problem: uncaught exceptions are not written via the Log4j interface. But this could be solved by setting the [default handler](http://stackoverflow.com/a/2344830/873661).

src/main/java/net/sf/jabref/logic/error/LogMessage.java
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * This class will be get the message and handle it with a typ, which will be define in {@link MessageType}
+ * <ul>
+ * <li>MessageType.LOG is define for log entries
+ * <li>MessageType.OUTPUT is define for output entries
+ * <li>MessageType.EXCEPTION is define for exception entries
+ * </ul>
+ */
+public class LogMessage {
```
tobiasdiez: Please replace this class by `org.apache.logging.log4j.core.LogEvent` and `Log4jLogEvent.newBuilder().setMessage(message).build()` if you need to convert a string to the `LogEvent`.

src/main/java/net/sf/jabref/logic/error/MessageType.java
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * This enum will be define for three different type of message
+ * <ul>
+ * <li>log
+ * <li>output
+ * <li>exception
+ * </ul>
+ */
+public enum MessageType {
```
tobiasdiez: Please use `org.apache.logging.log4j.Level` instead.

motokito:
@tobiasdiez I have refactored and have done feedback. Hope it will be merged this time 😄

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleView.java
```diff
@@ -59,12 +66,14 @@ public void show() {
         errorConsole.setDialogPane(pane);
         errorConsole.setResizable(true);
         errorConsole.show();
+        Log log = LogFactory.getLog(this.getClass());
```
tobiasdiez: I think this was for test purposes, right? Please remove it.

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleView.java
```diff
@@ -90,38 +99,32 @@ private void closeErrorDialog() {
...
-                                        case LOG:
-                                            getStyleClass().add("log");
-                                            setGraphic(IconTheme.JabRefIcon.INTEGRITY_INFO.getGraphicNode());
-                                            break;
-                                        default:
-                                            break;
+                                    setText(logMessage.getMessage().toString());
+
+                                    Level prio = logMessage.getLevel();
+                                    if (prio.equals(ERROR)) {
```
tobiasdiez: pls use a switch statement here and show an info by default (instead of nothing with `null/null`)

src/main/java/net/sf/jabref/logic/error/StreamEavesdropper.java
```diff
@@ -33,40 +34,47 @@ public static StreamEavesdropper eavesdropOnSystem() {
...
 
-    private void addToLog(String s, MessageType priority) {
-        if (!s.equals(System.lineSeparator())) {
-            LogMessage messageWithPriority = new LogMessage(s.replaceAll(System.lineSeparator(), ""), priority);
+    /**
+     * Creates a new log event with the given parameters and forwards it to the {@link LogMessages} archive.
+     * @param message message of log event
+     * @param priority level of log event
+     */
+    private void addToLog(String message, Level priority) {
```
tobiasdiez: Please go through the code that all `priority` and `prio` are properly renamed to `level`.

src/main/java/net/sf/jabref/logic/logging/GuiAppender.java
```diff
@@ -37,10 +39,25 @@ public static GuiAppender createAppender(@PluginAttribute("name") String name,
...
+     */
     @Override
     public void append(LogEvent event) {
-        CACHE.add(new String(this.getLayout().toByteArray(event)));
+        String message = new String(this.getLayout().toByteArray(event));
+        message = message.replace("\n\r", "\n").replace("\r", "\n");
+        // stack traces logged by 'Log.error("message"), e' will be split by new lines so we can create a new log event for each line as 'e.printStackTrace()' would do.
+        if (event.getLevel() == Level.ERROR) {
+            Arrays.asList(message.split("\n")).stream().filter(s -> !s.isEmpty()).forEach(log -> {
+                LogEvent messageWithPriority = Log4jLogEvent.newBuilder().setMessage(new SimpleMessage(log.replace("\n", ""))).setLevel(event.getLevel()).build();
```
tobiasdiez: Is this kind of message-rework really necessary? 
I had the hope that here a simple `LogMessages.getInstance().add(event)` would satisfy and then maybe use `event.getMessage().getFormattedMessage()` in the view (https://github.com/JabRef/jabref/pull/1383/commits/d6c8eb16d9dd0f964d0840c29a615bd19ece3819#diff-d08231c1b1e93f3d52d35f72f3b55821R112).

If this is not possible, please move the code to the view, i.e. determine there how a LogEvent should be displayed.

motokito: @tobiasdiez if i don't split this it will be show me not so nice in list view. It looks like this, because in append-method of GuiAppender I get only one LogEvent entry with all information by insert a log entry like Log.error (e). It is not like System.err:

2016-09-19_07h12_28
And if i don't replace System.lineseparator with "" then i will get 2 line each list cell.

That is the reason, why i have reformat this log event entry. That look better. 😄

tobiasdiez: I see. What happens if you use setText(logMessage.getMessage().getFormattedMessage()) to show the message instead of setText(logMessage.getMessage().toString()) ?

motokito: I have tried it, but i get the same effect like that screenshot 😄

tobiasdiez: Ok, then keep your code but move it to the view if this is possible.

motokito: Sorry I have tried it, but it doesn't let me extract to view. 😢. Other feedback I have done it 😄

tobiasdiez: Ok, then we leave it like that for the moment and maybe fix it later.

PR #1384, opened 2016-05-11
Fix #1334: Display correct number of entries in static groups

### LINKED ISSUE #1334: Bug in Jabref 3.4dev: "Number of elements contained in each group" always shown as "0"
JabRef version 3.4dev--snapshot--2016-05-01--master--212201a on Windows 10

Bug in current master branch of Jabref (this bug is not found in Jabref 3.3):

Steps to reproduce:

1. Create a new .bibtex file with a few entries and groups
2. Turn on: "Show number of elements contained in each group"
3. Every group in your database is shown to have 0 entries

tobiasdiez:
Fix of #1334. Nice bonus: total number of entries is displayed after "All Entries" group.

- [ ] Change in CHANGELOG.md described (bug not present in 3.3)

### DIFF
--- a/src/main/java/net/sf/jabref/logic/groups/ExplicitGroup.java
+++ b/src/main/java/net/sf/jabref/logic/groups/ExplicitGroup.java
@@ -168,10 +168,6 @@ public class ExplicitGroup extends KeywordGroup {
         return ExplicitGroup.ID;
     }
 
-    public int getNumEntries() {
-        return legacyEntryKeys.size();
-    }
-
     @Override
     public int hashCode() {
         return super.hashCode();
--- a/src/test/java/net/sf/jabref/logic/groups/AbstractGroupTest.java
+++ b/src/test/java/net/sf/jabref/logic/groups/AbstractGroupTest.java
@@ -1,0 +1,68 @@
+/*
+ * Copyright (C) 2003-2016 JabRef contributors.
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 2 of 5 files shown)


src/test/java/net/sf/jabref/logic/groups/AbstractGroupTest.java
```diff
@@ -0,0 +1,68 @@
...
+                invocation.getArgumentAt(0, BibEntry.class).getField("author").contains("author2"));
+        assertEquals(1, group.numberOfHits(entries));
+    }
+
+    @Test
+    public void numberOfHitsCallsContainsToDetermineMultipleHits() throws Exception {
+        when(group.contains(any())).then(invocation ->
+                invocation.getArgumentAt(0, BibEntry.class).getField("author").contains("author1"));
+        assertEquals(2, group.numberOfHits(entries));
+    }
```
simonharrer: I would have made it simpler and using the four different group types instead of mocking the abstract class to test more stuff. But I think this is OK as well.

simonharrer:
LGTM

src/main/java/net/sf/jabref/logic/groups/ExplicitGroup.java
```diff
@@ -168,10 +168,6 @@ public String getTypeId() {
         return ExplicitGroup.ID;
     }
 
-    public int getNumEntries() {
```
matthiasgeiger: Does the removal of this method introduce some performance issues? Or is iterating over all entries still fast enough?

tobiasdiez: For my database of around 500 entries this is no problem.

tobiasdiez: I tried it now with a bigger database and although the population of the group tree was not "instantaneously", the performance was acceptable. So I merge this in.

PR #1385, opened 2016-05-11
Fix a few group related bugs

tobiasdiez:
See commit messages.
I also removed the refresh button since I think this functionality is no longer required. So there was room to move the settings button to the top. Now also remove the border and you arrive at:
untitled


simonharrer:
We should think about the effects on performance here. I really do like the stay in sync feature, but how much does this slow the UI down? If I have 300 groups and 6000 entries which are evaluated after each change in any field, it could really go down performance wise.

tobiasdiez:
Yes you are right, there might be performance issues. For my database with 500 entries and a handful of groups this wasn't a problem. Do you have a db with more entries and groups so that I can test the performance?

stefan-kolb:
Amazing that you fixed so much issues :+1: !

stefan-kolb:
Can you please evaluate on the remark of Simon and then rebase so we can merge this in for v3.4 :smile: ?

tobiasdiez:
There are some performance related issues, however they are not really intrinsic to this PR. The problem is that some database open post actions modify every entry and thus trigger many repaints of the group tree. I will address this in a new PR.

simonharrer:
Think about the save actions as well - they also trigger a lot of entry changes. Hence, to me, it is still an issue of the group tree. Maybe only update the tree if no change happened for 500ms?

PR #1386, opened 2016-05-12
Entries retain their groupmembership when undoing their cut/deletion

bartsch-dev:
When undoing an entry which was deleted/cut while a group was selected of which the entry wasn't a member of it was after the undo.

I fixed that.

### Steps to reproduce
- delete an entry
- select a group of which the entry wasn't a member of
- undo the deletion
- now the entry is a member of the group

### DIFF
--- a/src/main/java/net/sf/jabref/gui/BasePanel.java
+++ b/src/main/java/net/sf/jabref/gui/BasePanel.java
@@ -1269,6 +1269,12 @@ public class BasePanel extends JPanel implements ClipboardOwner, FileUpdateListe
 
         @Subscribe
         public void listen(EntryAddedEvent addedEntryEvent) {
+            // if the added entry is an undo don't add it to the current group
+            if (addedEntryEvent.isUndo()){
+                scheduleUpdate();
+                return;
+            }
+
             // Automatically add new entry to the selected group (or set of groups)
             if (Globals.prefs.getBoolean(JabRefPreferences.AUTO_ASSIGN_GROUP) && frame.groupToggle.isSelected()) {
                 final List<BibEntry> entries = Collections.singletonList(addedEntryEvent.getBibEntry());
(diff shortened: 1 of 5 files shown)


simonharrer:
Why would they loose their group membership if they get deleted? Is this really required? @tobiasdiez what do you think as our group expert? :)

tobiasdiez:
I can confirm the issue. The problem seems to be connected to the functionality "Automatically assign new entry to selected group", although disabling this option doesn't solve it. 
If the user undos the deletion of an entry, then apparently it is treated as a new entry and added to the currently selected group.
Note that the problem also occurs with dynamic groups so that the proposed fix in this PR does not really solve the issue. I would prefer a fix which really addresses the problem at its root, i.e. how undos are treated.

bartsch-dev:
I added an undo-flag to the `EntryAddedEvent` and told the `GroupListener` to don't add the entry to the current marked group if the flag is set.

simonharrer:
What is status here? Is the solution of @bartsch-dev acceptable to you @tobiasdiez ?

src/main/java/net/sf/jabref/gui/BasePanel.java
```diff
@@ -1269,6 +1269,12 @@ public void run() {
 
         @Subscribe
         public void listen(EntryAddedEvent addedEntryEvent) {
+            // if the added entry is an undo don't add it to the current group
+            if (addedEntryEvent.isUndo()){
+                frame.getGroupSelector().revalidateGroups();
```
tobiasdiez: Replace by scheduleUpdate();

tobiasdiez:
Yeah, it is probably the best way to solve the problem. Good work :+1:

src/main/java/net/sf/jabref/gui/BasePanel.java
bartsch-dev: done

PR #1387, opened 2016-05-12
displayed entry count of explicit groups are now correct

bartsch-dev:
Since the change of how JabRef handles groups (specifically ExplicitGroup) the displayed number of how many entries in the group are (within the group panel) was wrong.
(`ExplicitGroup#getNumEntries()` would return the legacy count)

I fixed that.


matthiasgeiger:
oh - there is already #1384 in the pipeline which fixes this :grimacing:

bartsch-dev:
Oh damn, I found it a few days ago but didn't had time to fix it until today.
I'll close mine, his looks far more sophisticated.

PR #1390, opened 2016-05-12
Ported the KeyBindings dialog to JavaFX

boceckts:
The KeyBindings are now organized in categories referring to the JabRef menu titles.
The button graKey has been removed and once a entry is selected the table listens for valid key combinations. The FXAlert class has also been adjusted to listen to the closeDialog shortcut.

keybindingsfx
keybindingsfxdefault
keybindingssuccess


Siedlerchr:
UI Looks good :+1:  Just one thing you should change regarding the UI: Don't ask questions in dialogs. Because that is too much "cognitive expenditure" for the user. Label the buttons with a clear description: (e.g. Reset Keybinding, Back to Dialog).
Maybe offer a button "Restart JabRef now", if this is technically possible.


See here for a discussion on Why Questions with Yes/No are a bad idea
http://ux.stackexchange.com/questions/9946/should-i-use-yes-no-or-ok-cancel-on-my-message-box

src/main/java/net/sf/jabref/gui/FXAlert.java
```diff
@@ -89,6 +94,45 @@ public FXAlert(AlertType type) {
...
+        if (keyEvent.isAltDown()) {
+            modifiers = (modifiers.isEmpty()) ? "alt" : modifiers + " alt";
+        }
+        String newShortcut = (modifiers.isEmpty()) ? code.toString() : modifiers + " " + code;
+        KeyCombination pressedCombination = KeyCombination.valueOf(newShortcut);
+        if (combination.equals(pressedCombination)) {
+            return true;
+        } else {
+            return false;
+        }
```
tschechlovdev: if statement is not necessary, this could be done in one line ``return combination.equals(pressedCombination)``

src/main/resources/net/sf/jabref/gui/keyboard/KeyBindingsDialog.css
```diff
@@ -0,0 +1,12 @@
...
+}
+
+.custom-buttons {
+	-fx-padding: 5px;
+}
+
+.dialog-pane *.button-bar {
+	-fx-max-height: 0;
+	-fx-pref-height: 0;
+}
```
tschechlovdev: add a new line at the end of the file.

src/main/java/net/sf/jabref/gui/FXAlert.java
```diff
@@ -89,6 +94,45 @@ public FXAlert(AlertType type) {
...
+        if (code == KeyCode.UNDEFINED) {
+            return false;
+        }
+        // gather the pressed modifer keys
+        String modifiers = "";
+        if (keyEvent.isControlDown()) {
+            modifiers = "ctrl";
+        }
+        if (keyEvent.isShiftDown()) {
+            modifiers = (modifiers.isEmpty()) ? "shift" : modifiers + " shift";
```
tschechlovdev: why do you ``need modifiers.isEmpty() ?``
wouldn't ``modifiers + "shift"`` be enough?

tobiasdiez:
A few comments regarding the UI:
- Unify capitalization of headings (`File` vs `EDIT`), preferably via a global css property
- Better formatting of shortcuts, i.e. `Ctrl + C` instead of `ctrl C`
- Move `Default` button more to the left and rename it to something like `Reset all`
- Rename `OK` button to `Save`
- Remove light gray column separation
- Increase space between list and the buttons at the bottom (are there standards for such things? research and document them)
- The error/information dialogs look a bit bloated since they have two headings (in the dialog window header and in the first row next to the icon). If you can style these dialogs, I would prefer a flat layout without the window border (similar to https://developers.google.com/apps-script/images/style-dialogs-do.svg or https://debajitcrm.files.wordpress.com/2015/01/image_thumb7.png?w=757&h=396) or without the first row (https://i-msdn.sec.s-msft.com/dynimg/IC725781.png)

boceckts:
@tobiasdiez I have now adressed most of your comments. I called the default button `Reset binding` since it resets the selected binding only, except there is none selected. Also I wasn't really sure what you meant with "remove light gray column separation".
Also the issue with blank symbols in localized strings in fxml files as described in #968 still exists. For this reason you will see a '_' in the screenshots below in the `Reset binding`  button, otherwise the fxml file can't be loaded.

Here are some new screenshots

screen shot 2016-05-25 at 16 54 19
screen shot 2016-05-25 at 16 53 44
screen shot 2016-05-25 at 16 52 53

src/main/java/net/sf/jabref/gui/FXAlert.java
```diff
@@ -89,6 +96,42 @@ public FXAlert(AlertType type) {
...
+    }
+
+    /**
+     * Check if the given keyCombination equals the given keyEvent
+     *
+     * @param combination as KeyCombination
+     * @param keyEvent as KeEvent
+     * @return true if matching, else false
+     */
+    private boolean checkKeyCombinationEquality(KeyCombination combination, KeyEvent keyEvent) {
```
tobiasdiez: I would prefer if you move this code to the `KeyBindingPreferences` class. Than you can write something like `if(Globals.getKeyPrefs().checkKeyCombinationEquality(KeyBinding.CLOSE_DIALOG, event)`

src/main/java/net/sf/jabref/gui/FXAlert.java
```diff
@@ -89,6 +96,42 @@ public FXAlert(AlertType type) {
...
+            modifiers = "ctrl";
+        }
+        if (keyEvent.isShiftDown()) {
+            modifiers += " shift";
+        }
+        if (keyEvent.isAltDown()) {
+            modifiers += " alt";
+        }
+        modifiers = modifiers.trim();
+        String newShortcut = (modifiers.isEmpty()) ? code.toString() : modifiers + " " + code;
```
tobiasdiez: write the above with space at the end, i.e `modifiers = "ctrl "` and `"shift "`. Then a simple `newShortcut = modifiers.trim() + code` should work.

src/main/java/net/sf/jabref/gui/JabRefFrame.java
```diff
@@ -1672,11 +1672,10 @@ public SelectKeysAction() {
 
         @Override
         public void actionPerformed(ActionEvent e) {
-            KeyBindingsDialog d = new KeyBindingsDialog(new KeyBindingRepository(Globals.getKeyPrefs().getKeyBindings()));
-            d.setDefaultCloseOperation(WindowConstants.DISPOSE_ON_CLOSE);
-            d.pack(); //setSize(300,500);
-            d.setLocationRelativeTo(JabRefFrame.this);
-            d.setVisible(true);
+            Platform.runLater(() -> {
```
tobiasdiez: Reuse the new KeybindingAction defined below?

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingPreferences.java
```diff
@@ -48,6 +49,10 @@ public KeyStroke getKey(KeyBinding bindName) {
         }
     }
 
+    public KeyCombination getKeyCombinationFX(KeyBinding bindName) {
```
tobiasdiez: FX suffix required? 
`bindName`-> `binding`

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingsDialogViewModel.java
```diff
@@ -0,0 +1,210 @@
...
+    private Button resetButton;
+
+
+    @FXML
+    private void initialize() {
+        keyBindingsTable.getSelectionModel().setSelectionMode(SelectionMode.SINGLE);
+        ButtonBar.setButtonData(resetButton, ButtonData.LEFT);
+    }
+
+    public void initializeView() {
```
tobiasdiez: Moving the methods to the initialize() method does not work?

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingWrapper.java
```diff
@@ -0,0 +1,50 @@
+package net.sf.jabref.gui.keyboard;
+
+import javafx.beans.property.SimpleStringProperty;
+import javafx.beans.property.StringProperty;
+import org.apache.commons.lang.WordUtils;
+
+public class KeyBindingWrapper {
```
tobiasdiez: Rename `Wrapper` -> `ViewModel` (wrapping a datamodel for display is exactly the purpose of a view model)

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingWrapper.java
```diff
@@ -0,0 +1,50 @@
...
+
+    private KeyBinding keyBinding;
+    private final SimpleStringProperty bind;
```
tobiasdiez: Correct naming of `bind` in the whole class.

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingWrapper.java
```diff
@@ -0,0 +1,50 @@
...
+    private final KeyBindingCategory category;
+
+
+    public KeyBindingWrapper(KeyBinding keyBinding, String bind, KeyBindingCategory category) {
```
tobiasdiez: In the best case you just need the `KeyBinding` parameter and get/set the rest from it.

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingWrapper.java
```diff
@@ -0,0 +1,50 @@
...
+    public String getBind() {
+        return bind.get();
+    }
+
+    public StringProperty bindProperty() {
+        return this.bind;
+    }
+
+    public void setBind(String bind) {
+        bind = WordUtils.capitalizeFully(bind);
```
tobiasdiez: If I understand the code correctly, then something as `Crtl + C` is written to the `bind` property. But previously the code (the KeyBindingPreference class) expected something as `ctrl c`, correct? Does this lead to problems?

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingsDialogViewModel.java
```diff
@@ -0,0 +1,210 @@
...
+        populateTable();
+        bindColumnValues();
+    }
+
+    /**
+     * Read all keybindings from the keybinding repository and create table keybinding
+     * models for them
+     */
+    private void populateTable() {
+        TreeItem<KeyBindingWrapper> root = new TreeItem<>(new KeyBindingWrapper(null, "", KeyBindingCategory.FILE));
```
tobiasdiez: Add constructor to KeyBindingWrapper which only accepts a category.

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingsDialogViewModel.java
```diff
@@ -0,0 +1,210 @@
...
+            categoryItem.setExpanded(true);
+            root.getChildren().add(categoryItem);
+        }
+        root.setExpanded(true);
+        keyBindingsTable.setRoot(root);
+    }
+
+    private void bindColumnValues() {
+        actionColumn
+                .setCellValueFactory(cellData -> new SimpleStringProperty(cellData.getValue().getValue().getName()));
```
tobiasdiez: Also make `Name` a property in the KeyBindingWrapper class.

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingWrapper.java
tobiasdiez: Correct naming of `bind` in the whole class.

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingsDialogViewModel.java
```diff
@@ -0,0 +1,210 @@
...
+        shortcutColumn.setCellValueFactory(cellData -> cellData.getValue().getValue().bindProperty());
+    }
+
+    private void registerKeyEvents() {
+        keyBindingsTable.setOnKeyPressed(evt -> grabKey(evt));
+    }
+
+    private Stage getDialogStage() {
```
tobiasdiez: Can be replaced by https://github.com/boceckts/jabref/blob/a559e058c065fe58cb5ced1380dda4f7ed889f51/src/main/java/net/sf/jabref/gui/FXAlert.java#L146 ?
In this case it may be worthwhile to add a closeDialog method to FXAlert.

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingsDialogViewModel.java
```diff
@@ -0,0 +1,210 @@
...
+
+    @FXML
+    private void closeDialog() {
+        Stage stage = getDialogStage();
+        stage.close();
+    }
+
+    @FXML
+    private void saveKeyBindingsAndCloseDialog() {
+        Globals.getKeyPrefs().setNewKeyBindings(keyBindingRepository.getKeyBindings());
```
tobiasdiez: Pass the KeyPreference class to this viewmodel instead of the `keyBindingRepository` so that you don't need to access the Globals object here (we want to get rid of it in the end).

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingsDialogViewModel.java
```diff
@@ -0,0 +1,210 @@
...
+
+        // show success message
+        FXAlert alert = new FXAlert(AlertType.INFORMATION, Localization.lang("Key bindings changed"));
```
tobiasdiez: Extract these statements to new class `Dialogs` (or something like that). So you can simply call `Dialogs.showSuccess( title , text )`.

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingsDialogViewModel.java
```diff
@@ -0,0 +1,210 @@
...
+            }
+        });
+    }
+
+    @FXML
+    private void setDefaultBindings() {
+        ObservableList<TreeItem<KeyBindingWrapper>> selectedKeyBindingWrapperItems = keyBindingsTable
+                .getSelectionModel().getSelectedItems();
+        if (selectedKeyBindingWrapperItems.isEmpty()) {
+            FXAlert alert = new FXAlert(AlertType.INFORMATION, Localization.lang("Resetting all key bindings"));
```
tobiasdiez: Similar to above, move to `boolean Dialogs.showConfirmation( title, text, buttonText)`

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingsDialogViewModel.java
```diff
@@ -0,0 +1,210 @@
...
+            ButtonType resetButtonType = new ButtonType("Reset", ButtonData.OK_DONE);
+            alert.getDialogPane().getButtonTypes().addAll(resetButtonType, ButtonType.CANCEL);
+            alert.showAndWait().ifPresent(response -> {
+                if (response == resetButtonType) {
+                    keyBindingRepository.resetToDefault();
+                    populateTable();
+                }
+            });
+        } else {
+            selectedKeyBindingWrapperItems.forEach(keyBindWrapperItem -> {
```
tobiasdiez: I find it counterintuitive that the same button sometimes resets only a few bindings and sometimes all.
 Proposal: add a third column to the table which contains a small reset button (only icon, light gray by default, dark gray on hovering the row) which resets the corresponding binding. Then the button at the button is a `Reset all bindings` button.

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingsDialogViewModel.java
```diff
@@ -0,0 +1,210 @@
...
+        }
+    }
+
+    private void grabKey(KeyEvent evt) {
+        // first check if a valid entry is selected
+        if (keyBindingsTable.getSelectionModel().getSelectedIndex() < 0) {
+            return;
+        }
+        KeyBindingWrapper selectedEntry = keyBindingsTable.getSelectionModel().getSelectedItem().getValue();
+        if ((selectedEntry == null) || (selectedEntry.getKeyBinding() == null)) {
```
tobiasdiez: rewrite second check as `selectedEntry.isCategory()`

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingsDialogViewModel.java
```diff
@@ -0,0 +1,210 @@
...
+            return;
+        }
+
+        // validate the shortcut is no modifier key
```
tobiasdiez: Move the rest of this method to `selectedEntry.setNewBinding(event)`.

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingsDialogViewModel.java
```diff
@@ -0,0 +1,210 @@
...
+        KeyCode code = evt.getCode();
+        if (code.isModifierKey() || (code == KeyCode.BACK_SPACE) || (code == KeyCode.SPACE) || (code == KeyCode.TAB)
+                || (code == KeyCode.ENTER) || (code == KeyCode.UNDEFINED)) {
+            return;
+        }
+
+        // gather the pressed modifer keys
```
tobiasdiez: I have seen this code somewhere else too. Try to reuse it.

src/main/resources/net/sf/jabref/gui/keyboard/KeyBindingsDialog.css
```diff
@@ -0,0 +1,12 @@
+.keybinding-table {
+	-fx-padding: 5px;
```
tobiasdiez: Try if disabling the grid line (the separator between the columns) with ` -fx-table-cell-border-color: transparent;` looks good.

tobiasdiez:
Looks pretty good! Nice job :+1:
Overall the code is also fine, just some minor refactorings needed.
The only thing I'm missing are tests, especially for KeyBindingsDialogViewModel.

Spaces in the language keys should work as soon as https://github.com/JabRef/jabref/pull/1426 is merged.

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingWrapper.java
boceckts: All items in a `TreeTableView` are fom the same type, also the expandable ones. Since those are no real KeyBindings but rather only their category, I created the this wrapper class as view model to include the categories as well as the keyBindings.

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingWrapper.java
boceckts: No the problem is that we need the keycode combination to be in a blank separated string to ensure compatibility with AWT's `KeyStroke` as well as JavaFX's `KeyCombination`.

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingsDialogViewModel.java
boceckts: unfortuntely not

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingsDialogViewModel.java
boceckts: I don't see how as I don't have a hold to the dialog object but only its content.

boceckts:
I addressed most of the comments. Unfortunately I wasn't able to figure out how to include a icon from the webfont in the `TreeTableCell` just yet.

matthiasgeiger:
Which Icon do you want to include? 
The icons are organized in an enum here: https://github.com/JabRef/jabref/blob/master/src/main/java/net/sf/jabref/gui/IconTheme.java

Just search for the css code which is shown at https://materialdesignicons.com 
If the icon is not already used you have to add it to the `JabRefIcon` enum.
You can find the cheatsheet showing the needed codepoints here: https://materialdesignicons.com/cheatsheet

To actually get an `Icon` you'll have to call `getIcon()` or `getSmallIcon()` on an enum element. 
E.g., `IconTheme.JabRefIcon.NEW.getIcon()`

tobiasdiez:
The new changes are hard to review because many unchanged lines are marked as changed (probably line ending problems). But I really like the new Dialogs class (in particular :+1: for taking the time to write a meaningful documentation).

Regarding the icon, you just set the font via ` -fx-font-family: 'MaterialDesignIcons';` and set the unicode identifier as the text, see https://github.com/JabRef/jabref/pull/1438. For some reason the icon codepoints differ from the ones used by JabRef (e.g. `IconTheme.JabRefIcon.MAKE_KEY` results in a folder icon) @matthiasgeiger do you know why?

matthiasgeiger:
No idea why the codepoints differ... 

Is it possible to directly use the .css version of the font in JavaFX?

boceckts:
Any idea on why the icons are still not loading correctly?
missingicons

tobiasdiez:
For me (java 1.8.0_91, Win10) both versions display icons. However, only for `'Material Design Icons'` the codepoints are correct. So use the version with spaces.

src/main/java/net/sf/jabref/gui/FXAlert.java
```diff
@@ -89,6 +95,14 @@ public FXAlert(AlertType type) {
         fxDialogWindow.setOnHiding(evt -> setSwingWindowsEnabledAndFocusable(true));
 
         fxDialogWindow.setOnCloseRequest(evt -> this.close());
+
+        fxDialogWindow.getScene().setOnKeyPressed(evt -> {
+            KeyBindingPreferences keyPreferences = Globals.getKeyPrefs();
+            KeyCombination combination = keyPreferences.getKeyCombination(KeyBinding.CLOSE_DIALOG);
+            if (keyPreferences.checkKeyCombinationEquality(combination, evt)) {
```
tobiasdiez: Can you please also introduce an overload so that you can simply write `if keyPreferences.checkKeyCombinationEquality(KeyBinding.CLOSE_DIALOG, evt)`.

src/main/java/net/sf/jabref/gui/FXDialogs.java
```diff
@@ -0,0 +1,135 @@
...
+import javafx.scene.control.DialogPane;
+
+/**
+ * This class provides static methods to create default
+ * JavaFX dialogs which will also work on top of swing
+ * windows. The created dialogs are instances of the
+ * {@link FXAlert} class. The available dialogs in this class
+ * are useful for displaying small information graphic dialogs
+ * rather than complex windows. For more complex dialogs it is
+ * advised to rather a new sub class of {@link FXAlert}.
```
tobiasdiez: + create

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingViewModel.java
```diff
@@ -0,0 +1,134 @@
...
+
+    public StringProperty shownBindingProperty() {
+        return this.shownBinding;
+    }
+
+    public String getBinding() {
+        return realBinding;
+    }
+
+    public void setBinding(String bind) {
```
tobiasdiez: Make private as well as setKeyBinding? 
I suspect the proper way to change the keybinding is via `setNewBinding`

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingViewModel.java
```diff
@@ -0,0 +1,134 @@
...
+        bind = WordUtils.capitalizeFully(bind);
+        this.shownBinding.set(bind.replace(" ", " + "));
+    }
+
+    public KeyBindingCategory getCategory() {
+        return category;
+    }
+
+    private void setName() {
+        this.name.set((keyBinding == null) ? getCategory().getName() : keyBinding.getLocalization());
```
tobiasdiez: just set this in the constructor. Also rename to `displayName`

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingViewModel.java
```diff
@@ -0,0 +1,134 @@
...
+public class KeyBindingViewModel {
+
+    private KeyBinding keyBinding = null;
+    private String realBinding = "";
+    private final SimpleStringProperty name = new SimpleStringProperty();
+    private final SimpleStringProperty shownBinding = new SimpleStringProperty();
+    private final KeyBindingCategory category;
+
+
+    public KeyBindingViewModel(KeyBinding keyBinding, String binding, KeyBindingCategory category) {
```
tobiasdiez: The third parameter is not required as far as I can see (if it's a binding, then its category is stored in it)

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingsDialogViewModel.java
tobiasdiez: This is because the keyBindingRepository is not yet initialized, right?
In this case, just call the initializeView in setKeyBindingPreferences (and make it private)

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingsDialogViewModel.java
tobiasdiez: Ok.

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingsDialogViewModel.java
```diff
@@ -0,0 +1,196 @@
...
+                                }
+                                keyBindingsTable.refresh();
+                            }
+                        };
+                    }
+                });
+    }
+
+    private void resetKeyBinding(KeyBindingViewModel keyBindingViewModel) {
+        if (keyBindingViewModel.getKeyBinding() != null) {
```
tobiasdiez: Move this code to KeyBindingViewModel, so that you simply can write keyBindingViewModel.resetToDefault();

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingsDialogViewModel.java
```diff
@@ -0,0 +1,196 @@
...
+                    if (response == resetButtonType) {
+                        keyBindingRepository.resetToDefault();
+                        populateTable();
+                    }
+                });
+    }
+
+    private void grabKey(KeyEvent evt) {
+        // first check if a valid entry is selected
+        if (keyBindingsTable.getSelectionModel().getSelectedIndex() < 0) {
```
tobiasdiez: Just create a new field `selectedKeyBinding` and bind it to keyBindingsTable.getSelectionModel().selectedItemProperty().

tobiasdiez:
Good work!
I just have a few remarks about the general usage of view models and databinding.

The idea of this ViewModels is the following:
- Expose fields (as properties) so that the view can easily bind to them
- Listen to changes to the underlying data and update accordingly. These changes are then propagated to the view via the binding.
- Expose methods which modify the underlying data, the view never calls the logic methods themselves.
- In the best case scenario, **you never access the view or its controls directly but for setting up the binding (so only in initialize()).**

So in your case you have two ViewModels:
For the whole dialog:
- Expose bindingsTree and selectedItem
- Listen to external changes (which is setKeyBindingsRepository) and update bindingsTree then.
- Expose methods for resetting all keys, save changes, closing the dialog and relaying key pressed events to the selected item

For the item in the table: (KeyBindingViewModel)
- Expose displayName and binding
- No need to listen for changes since keybindings cannot be modified except in this dialog
- Expose methods for changing the keybinding

The reason for this "binding over everything else" philosophy is that you can easily write tests. For example:
````java
KeyBindingDialogViewModel dialog = new blabla();
dialog.setRepository(some mocked rep);
AssertEqual(expectedTree, dialog.bindingsTree);
% If this test passes then you know that the correct things are displayed in the dialog (except if you forgot to bind the vew to `bindingsTree`)

KeyBindingDialogViewModel dialog = new blabla();
dialog.setRepository(some mocked rep);
AssertCalls(mockedRep.resetBindinings(), dialog.resetBinding());
AssertEquals(defaultTree, dialog.bindingsTree);
% Check that the dialog relays the click to the correct method and updates view (only thing that can fail is again that the binding of the buttonclick is not bound to  dialog.resetBinding()

KeyBindingDialogViewModel dialog = new blabla();
dialog.setRepository(some mocked rep);
dialog.selectedIBinding = some mocked binding view model
AssertCalls(mockedBinding.setNewBinding, dialog.keyPressed(event))

KeyBindingViewModel binding = new blabla();
AssertCalls(repository.change(), binding.setNewBinding(event))
AssertEquals("Ctrl + Alt + f4", binding.KeyBinding);
````
Can you please try to add these tests?

File: src/main/java/net/sf/jabref/gui/keyboard/KeyBindingViewModel.java (hunk 28 above)
Comment (boceckts) [inline src/main/java/net/sf/jabref/gui/keyboard/KeyBindingViewModel.java]: Yes you are right, I forgot to remove it when adding the category constructor.

File: src/main/java/net/sf/jabref/gui/keyboard/KeyBindingViewModel.java (hunk 27 above)
Comment (boceckts) [inline src/main/java/net/sf/jabref/gui/keyboard/KeyBindingViewModel.java]: It only gets called in the constructors that's why it's `private`.

File: src/main/java/net/sf/jabref/gui/keyboard/KeyBindingsDialogView.java [hunk 31]
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    public void show(KeyBindingPreferences keyBindingPreferences) {
+        FXAlert keyBindingsDialog = new FXAlert(AlertType.INFORMATION, Localization.lang("Key bindings"));
+        keyBindingsDialog.setDialogPane((DialogPane) this.getView());
+        KeyBindingsDialogViewModel controller = (KeyBindingsDialogViewModel) fxmlLoader.getController();
+        controller.setKeyBindingPreferences(keyBindingPreferences);
+        controller.initializeView();
+        keyBindingsDialog.setResizable(true);
+        ((Stage) keyBindingsDialog.getDialogPane().getScene().getWindow()).setMinHeight(475);
```
tobiasdiez: Can this be set in the fxml file?

boceckts: Well the minimum size of the dialog pane is already set to this dimension in the fxml file. However if you omit these lines you are able to resize the dialog window to whatever you like but not the dialog content. So all properties set in the fxml file only affect the dialog content but not the dialog window itself. With these lines omitted you could resize the dialog to a point where the content doesn't resize and it may look bugged.

screen shot 2016-06-07 at 15 04 46

boceckts:
I added some tests for the dialog view model

src/test/java/net/sf/jabref/gui/keyboard/KeyBindingsDialogViewModelTest.java
```diff
@@ -0,0 +1,175 @@
...
+
+    KeyBindingsDialogViewModel model = null;
+    KeyBindingPreferences keyBindingsPreferences = null;
+    KeyBindingRepository keyBindingRepository = null;
+
+
+    @Before
+    public void setUp() {
+        model = new KeyBindingsDialogViewModel();
+        JabRefPreferences preferences = JabRefPreferences.getInstance();
```
tobiasdiez: Try to pass a mockito mock of the preferences to the keybindingpreferences (see wiki)

src/test/java/net/sf/jabref/gui/keyboard/KeyBindingsDialogViewModelTest.java
```diff
@@ -0,0 +1,175 @@
...
+        } catch (BackingStoreException e) {
+            e.printStackTrace();
+        }
+        keyBindingsPreferences = new KeyBindingPreferences(preferences);
+        model.setKeyBindingPreferences(keyBindingsPreferences);
+        keyBindingRepository = model.getKeyBindingRepository();
+    }
+
+    @Test
+    public void testNoModifierInvalidKeyBinding() {
```
tobiasdiez: Please change the name so that it reflects what is tested and what is the expected output (here `invalidKeyBindingIsNotSaved`), see wiki on conventions.

src/test/java/net/sf/jabref/gui/keyboard/KeyBindingsDialogViewModelTest.java
```diff
@@ -0,0 +1,175 @@
...
+
+        viewModel.resetToDefault(keyBindingRepository);
+        model.saveKeyBindings();
+
+        assertFalse(keyBindingsPreferences.checkKeyCombinationEquality(KeyBinding.ABBREVIATE, shortcutKeyEvent));
+    }
+
+    private KeyEvent prepareNewKeyBind(KeyBinding binding, String text, KeyCode code, boolean shiftDown,
+            boolean controlDown, boolean altDown, boolean metaDown) {
+        KeyBindingViewModel bindViewModel = new KeyBindingViewModel(binding, binding.getDefaultBinding());
```
tobiasdiez: I would split this method, so only create a shortcut for setting the selected keybinding.

src/test/java/net/sf/jabref/gui/keyboard/KeyBindingsDialogViewModelTest.java
```diff
@@ -0,0 +1,175 @@
...
+        model.grabKey(shortcutKeyEvent);
+
+        KeyEvent shortcutKeyEvent1 = prepareNewKeyBind(KeyBinding.FORWARD, "L", KeyCode.L, true, false, false, false);
+        assertFalse(keyBindingsPreferences.checkKeyCombinationEquality(KeyBinding.FORWARD, shortcutKeyEvent1));
+        model.grabKey(shortcutKeyEvent1);
+
+        KeyEvent shortcutKeyEvent2 = prepareNewKeyBind(KeyBinding.OPEN_FOLDER, "R", KeyCode.R, true, false, true,
+                false);
+        assertFalse(keyBindingsPreferences.checkKeyCombinationEquality(KeyBinding.OPEN_FOLDER, shortcutKeyEvent2));
+        model.grabKey(shortcutKeyEvent2);
```
tobiasdiez: Split into 3 tests. In general, test only one thing in a test (so at best you have only one assert per test)

src/test/java/net/sf/jabref/gui/keyboard/KeyBindingsDialogViewModelTest.java
```diff
@@ -0,0 +1,175 @@
...
+        assertFalse(keyBindingsPreferences.checkKeyCombinationEquality(KeyBinding.OPEN_FOLDER, shortcutKeyEvent2));
+    }
+
+    @Test
+    public void testSaveNewKeyBindingsToPreferences() {
+        KeyEvent shortcutKeyEvent = prepareNewKeyBind(KeyBinding.ABBREVIATE, "J", KeyCode.J, true, true, true, false);
+        assertFalse(keyBindingsPreferences.checkKeyCombinationEquality(KeyBinding.ABBREVIATE, shortcutKeyEvent));
+        model.grabKey(shortcutKeyEvent);
+
+        KeyEvent shortcutKeyEvent1 = prepareNewKeyBind(KeyBinding.UNMARK_ENTRIES, "X", KeyCode.X, true, false, false,
```
tobiasdiez: Also split

tobiasdiez:
Nice! I only have some minor remarks about the naming of the tests and their general structure (only one assert per test). If this is fixed, this can be merged from my point of view!

boceckts:
I refactored the test class.
Also I changed the FXDialogs to always show blocking dialogs and changed the way to create camel case bind names as there was a problem with the import of WordUtils.

src/test/java/net/sf/jabref/gui/keyboard/KeyBindingsDialogViewModelTest.java
```diff
@@ -30,20 +30,23 @@
     @Before
     public void setUp() {
         model = new KeyBindingsDialogViewModel();
-        JabRefPreferences preferences = JabRefPreferences.getInstance();
+
+        JabRefPreferences mockedPreferences = mock(JabRefPreferences.class);
         try {
-            preferences.clear();
+            mockedPreferences.clear();
```
tobiasdiez: No need to clear it.

tobiasdiez:
Ok, ready to go from my point of view! @JabRef/developers what do you think?

tobiasdiez:
Please rebase and remove the `BibTeX` localization key. Then I'll merge this in. (except if @JabRef/developers has more remarks)

boceckts:
@tobiasdiez The BibTeX localization key is used in the `KeyBindingCategory` class.

boceckts:
I found out travis ci uses different import orders for javafx than what is specified in eclipse.settings. I quickly changed it in #1518 so using "organize imports" in eclipse will work for javafx.

koppor:
Please be aware that menu translations are done using `Localization.menuTitle` and not by `Localization.lang`. See cb3729628cff8daba86546abea52fc4c9f46863d.

Issue #1063, opened 2016-03-30
Importer for Pubmed nbib format


stefan-kolb:
An importer for Pubmed nbib format would be a nice addition to the importers.
As an example, see http://www.ncbi.nlm.nih.gov/pubmed/23633646.
Also see #510.

Siedlerchr:
At least for their XML fomat there is a simple XSL-Script:
https://github.com/gui11aume/starcode/blob/master/doc/pubmed2bibtex.xsl

And I found a kind of Reference-Field-List for the nbib-Medline Format and their meanings:
https://www.nlm.nih.gov/bsd/mms/medlineelements.html
Together with this table:
http://www.ncbi.nlm.nih.gov/books/NBK3827/table/pubmedhelp.T.medline_display/?report=objectonly

stefan-kolb:
Hm, we got a Medline fetcher. Whats the difference between our Medline/Medline Plain and nbib?

tschechlovdev:
Medline Plain is a subset of nbib. So every field of Medline Plain is also a field of nbib. But not every nbib field is in the Medline Plain format. I will prepare a PR, where I add the fields to the Medline Plain Importer.
After that I will take care of Medline and nbib.

tschechlovdev:
I've added the nbib fields to Medline and MedlinePlain. So I think this issue can be closed now.

PR #1397, opened 2016-05-13
add fields for nbibImporter and add testfile

tschechlovdev:
Regarding: [#1063] (https://github.com/JabRef/jabref/issues/1063)
The medlineplain importer doesn't have all fields, that the nbib format has. So I added the fields to the medlineplain format and added a test+ testfile for the nbib format.
A list of the elements that should be imported now, can be found [here] (https://www.nlm.nih.gov/bsd/mms/medlineelements.html).

- [X ] Tests created for changes


src/main/java/net/sf/jabref/importer/fileformat/MedlinePlainImporter.java
```diff
@@ -141,20 +143,17 @@ public boolean isRecognizedFormat(InputStream stream) throws IOException {
...
                         type = "book";
-                    } else if ("journal article".equals(val)
-                            || "classical article".equals(val)
-                            || "corrected and republished article".equals(val)
-                            || "historical article".equals(val)
-                            || "introductory journal article".equals(val)
-                            || "newspaper article".equals(val)) {
+                    } else if ("journal article".equals(val) || "classical article".equals(val)
+                            || "corrected and republished article".equals(val) || "historical article".equals(value)
+                            || "introductory journal article".equals(val) || "newspaper article".equals(val)) {
```
boceckts: Why did you change these lines? IMO the old layout was easier to read.

src/main/java/net/sf/jabref/importer/fileformat/MedlinePlainImporter.java
```diff
@@ -141,20 +143,17 @@ public boolean isRecognizedFormat(InputStream stream) throws IOException {
                 String entry = current.toString();
 
                 String lab = entry.substring(0, entry.indexOf('-')).trim();
-                String val = entry.substring(entry.indexOf('-') + 1).trim();
+                String value = entry.substring(entry.indexOf('-') + 1).trim();
+
                 if ("PT".equals(lab)) {
-                    val = val.toLowerCase();
+                    String val = value.toLowerCase(Locale.ENGLISH);
```
boceckts: You should do a general renaming refactoring, it is hard to follow statements like this and with all the abbreviations, hm, lab, val,...

boceckts:
You should also record your changes in the readme

src/test/resources/net/sf/jabref/importer/fileformat/MedlinePlainImporterTestDOI.bib
```diff
@@ -1,3 +1,3 @@
 @TechReport{,
-  DOI                      = {10.1016/j.cpr.2005.02.002}
+  article-id                      = {doi:10.1016/j.cpr.2005.02.002}
```
stefan-kolb: Why article-id and not doi here?

tschechlovdev: I wasn't sure if it is right if there stands ``doi:``. [Here] (https://www.nlm.nih.gov/bsd/mms/medlineelements.html#aid) it says, that doi has to come at the end in brackets. That's why I thought it would be wrong to add this as doi. Should I keep both possibilities ?

stefan-kolb: Sorry, but I don't really get what you're trying to say :smile: .
Check out `DOI.java` it includes the regex for a valid DOI. Typically all our DOIs are just the DOI starting with the vendor code, i.e., 10.1016/j.cpr.2005.02.002.
From my understanding the removed line is the right result, except that the key should be lowercase but correct me if I'm wrong and don't understand the use case here!

tschechlovdev: Oh ok. The problem I thought is, that in the [medline description] (https://www.nlm.nih.gov/bsd/mms/medlineelements.html#aid)  for the AID element there come an example like this: ``AID - 10.1016/j.cpr.2005.02.002 [doi]``.
 So that ``[doi]`` has to come at the end.
That's why I thought that something like ``doi:10.1016/j.cpr.2005.02.002`` would be wrong. My question now is, whether both incidents of the doi should be possible or just one of them ?
Was this a little bit more understandable ? :)

stefan-kolb: Actually the DOI should be extracted by the importer to a correct format like I pointed out.

tschechlovdev: Ok, I've changed it to doi in lowercase.

src/main/java/net/sf/jabref/importer/fileformat/MedlinePlainImporter.java
```diff
@@ -45,6 +46,8 @@
     private static final Pattern PMID_PATTERN = Pattern.compile("PMID.*-.*");
     private static final Pattern PMC_PATTERN = Pattern.compile("PMC.*-.*");
     private static final Pattern PMCR_PATTERN = Pattern.compile("PMCR.*-.*");
+    private static final Pattern CREATE_DATE_PATTERN = Pattern.compile("\\d\\d\\d\\d/[0123]?\\d/\\s?[012]\\d:[0-5]\\d");
+    private static final Pattern COMPLETE_DATE_PATTERN = Pattern.compile("\\d\\d\\d\\d\\d\\d\\d\\d");
```
simonharrer: You can specify in regex how often a specific expression should match, instead of repeating it over and over.

simonharrer:
LGTM

src/main/java/net/sf/jabref/importer/fileformat/MedlinePlainImporter.java
stefan-kolb: \\d{8}

tschechlovdev: done

tschechlovdev:
For somehow the travis-ci is failing because an integration test, scrollingThroughEntryList, fails.

src/main/java/net/sf/jabref/importer/fileformat/MedlinePlainImporter.java
```diff
@@ -123,145 +125,110 @@ public ParserResult importDatabase(BufferedReader reader) throws IOException {
...
-                    hm.put("address", val);
-                } else if ("IS".equals(lab)) {
-                    hm.put("issn", val);
-                } else if ("VI".equals(lab)) {
-                    hm.put("volume", val);
-                } else if ("AB".equals(lab)) {
-                    String oldAb = hm.get("abstract");
-                    if (oldAb == null) {
-                        hm.put("abstract", val);
+                } else if ("PG".equals(label)) {
```
tobiasdiez: Could you please store the conversation as a map<MedlineKey, Bibtexkey> and then iterate over this map. Should be more readable.

src/main/java/net/sf/jabref/importer/fileformat/MedlinePlainImporter.java
```diff
@@ -272,4 +239,170 @@ else if ("AID".equals(lab)) {
         return new ParserResult(bibitems);
 
     }
+
+    private String addSourceType(String value, String type) {
+        String val = value.toLowerCase(Locale.ENGLISH);
+        String theType = type;
+        if ("book".equals(val)) {
```
tobiasdiez: Again, a switch or map should be more readable.

tobiasdiez:
LGTM, too. Just some minor remarks for readability.

src/main/java/net/sf/jabref/importer/fileformat/MedlinePlainImporter.java
tschechlovdev: done.

src/main/java/net/sf/jabref/importer/fileformat/MedlinePlainImporter.java
tschechlovdev: done (hope that I unterstood it right).

PR #1401, opened 2016-05-15
Fix #1399 export to html: html code is not well-formed

### LINKED ISSUE #1399: export to html: html code is not well-formed
JabRef 3.4dev--snapshot--2016-05-13--master--f28466e
windows 7 6.1 amd64 

Steps to reproduce:

1. Take any bib file, and press export to html
2. Open the html in a texteditor, there will be no < / body > or < / htm l>

Also a custom export I have been using seems to suffer (as if the end.layout is not used).

Thanks in advance for looking into it!

stefan-kolb:
Always write the footer, no matter if encoding is given or not.

### DIFF
--- a/src/main/java/net/sf/jabref/exporter/ExportFormat.java
+++ b/src/main/java/net/sf/jabref/exporter/ExportFormat.java
@@ -293,7 +293,7 @@ public class ExportFormat implements IExportFormat {
             }
 
             // Write footer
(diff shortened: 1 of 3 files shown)


stefan-kolb:
@tobiasdiez WDYT? You changed the code before.

simonharrer:
LGTM

PR #1403, opened 2016-05-16
Fix small event bug in PreviewPanel.java

obraliar:
This bug caused a "missing event subscriber" exception, when clicking on different bib entries in preview mode.

### DIFF
--- a/src/main/java/net/sf/jabref/gui/PreviewPanel.java
+++ b/src/main/java/net/sf/jabref/gui/PreviewPanel.java
@@ -275,13 +275,9 @@ public class PreviewPanel extends JPanel
 
     public void setEntry(BibEntry newEntry) {
 
-        if (entry.isPresent() && (entry.get() != newEntry)) {
-            entry.ifPresent(e -> e.unregisterListener(this));
-        }
-        if(newEntry != null) {
-            newEntry.registerListener(this);
-        }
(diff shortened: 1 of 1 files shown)


src/main/java/net/sf/jabref/gui/PreviewPanel.java
```diff
@@ -275,13 +275,9 @@ public void setLayout(Layout layout) {
...
 
-        if (entry.isPresent() && (entry.get() != newEntry)) {
-            entry.ifPresent(e -> e.unregisterListener(this));
-        }
-        if(newEntry != null) {
-            newEntry.registerListener(this);
-        }
+        entry.filter(e -> e != newEntry).ifPresent(e -> e.unregisterListener(this));
         entry = Optional.ofNullable(newEntry);
+        entry.ifPresent(e -> e.registerListener(this));
```
simonharrer: Can you please explain why this solves this issue? From a quick look, it essentially seems to be only the conversion to Optional code and no semantic changes.

tobiasdiez:
I think this bug was already fixed in #1374.

simonharrer:
Ah ok. Then this is only a small code improvement.

obraliar:
Yes, of course. I missed that while merging and replaced the code without paying attantion that it was already fixed.

Issue #1408, opened 2016-05-18
Positioning of variables in GUIGlobals


lenhard:
I am currently doing some architecture analysis of JabRef and stumbled into some part of the code that I'd like to discuss. 

We have the infamous class `GUIGlobals` in which several settings are hard-coded as public static instance variables. I noticed that some of these variables are only used in a single other class, thereby creating a dependency of that class to the gui package (which is very ugly). Example: The class `LatexFieldFormatter` is the only user of `GUIGlobals.LINE_LENGTH`, which prohibits the moval of `LatexFieldFormatter` into the `bibtex` package and makes that package transitively dependent of the `gui` package. Ultimately, if we fix this class and this dependency we can make the `bibtex` package completely self-contained and move it into logic. 

Several other variables in `GUIGlobals` are just used in a single other class. Therefore, I propose to move the variables closer to the scope of their usage where possible. This destroys `GUIGlobals` as a single point of configuration, but also improves our package dependencies. @JabRef/developers Are there objections to this?

tobiasdiez:
:+1: for removing GUIGlobals.

Siedlerchr:
Yes, I fully agree with removing Globals. Maybe you can also check for obsolete  Settings/Variables.

simonharrer:
Also think about whether they still make sense. Why do we need a line_length in that case anyway?

lenhard:
Ok, will do. @simonharrer: Honestly, I have no clue. But it is sometimes hard to decide if something still makes sense or not.

koppor:
I would not refactor just because ofa metrics value getting better.

Half of the values (such as `INDENT`, `NUMBER_COL_LENGTH`, `EXPORT_DIALOG_COL_0_WIDTH`) look like configuration variables. We had the feedback that one main value of JabRef is the plethora of configuration options. I don't know whether these values also fall in that category or whether we should reduce the configuration options.

I'm unsure whether we should switch to [Apache Commons Configuration 2.0](https://commons.apache.org/proper/commons-configuration/), where I really liked the 1.x series and the 2.x series seems even better.

Back to topic: For each value, it should be decided if it's a configuration option or if it's just a constant to gain code readability.

For the former, we should move it to our configuration thing, for the latter, we should move it to the appropriate places.

Beware: A configuration thing could also be used at one place only, but it really makes sense to have it stored at a central place to make users and programmers aware that this can be configured. `GUIGlobals.LINE_LENGTH` seems to be such a candidate.

[...]

PR #1409, opened 2016-05-19
Cleanup guiglobals

lenhard:
Implements #1408

Cleans up the `GUIGlobals` class and move `LatexFieldFormatter` to the `bibtex` package, which is now independent of the `gui` package.

As a final step, I would move the class `net.sf.jabref.importer.fileformat.FieldContentParser` to the `bibtex` package, which would make the package self-contained (apart from imports to the preferences and related classes) and would move the whole `bibtex` package into `logic`. Are there objections to this?


lenhard:
I restructured variable positioning as discussed in #1408 and turn MAX_BACK_HISTORY_SIZE, LINE_LENGTH, and INDENT into proper preferences.

@JabRef/developers What about turning the `bibtex` package into a subpackage of `logic`? Or should that rather stay a top-level package?

simonharrer:
The more stuff is in the logic package, the better.

lenhard:
Ok, from side this is good to go. I even was able to move some more classes into the model package.

matthiasgeiger:
LGTM

src/main/java/net/sf/jabref/JabRefPreferences.java
```diff
@@ -75,12 +73,16 @@
...
+import net.sf.jabref.model.entry.InternalBibtexFields;
 import net.sf.jabref.specialfields.SpecialFieldsUtils;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
 public class JabRefPreferences {
+    // Character separating field names that are to be used in sequence as
+    // fallbacks for a single column (e.g. "author/editor" to use editor where
+    // author is not set):
```
simonharrer: This comment is placed wrongly, I think.

src/test/java/net/sf/jabref/exporter/BibDatabaseWriterTest.java
```diff
@@ -269,6 +269,7 @@ public void writeEntryWithCustomizedTypeAlsoWritesTypeDeclaration() throws IOExc
                         + Globals.NEWLINE + Globals.NEWLINE
                         + "@Comment{jabref-entrytype: Customizedtype: req[required] opt[optional]}" + Globals.NEWLINE,
                 stringWriter.toString());
+        EntryTypes.removeAllCustomEntryTypes();
```
simonharrer: Hm, maybe use a try-finally here instead?

src/main/java/net/sf/jabref/gui/maintable/MainTableColumn.java
```diff
@@ -63,7 +64,7 @@ public String getDisplayName() {
             return null;
         }
 
-        StringJoiner joiner = new StringJoiner(MainTableFormat.COL_DEFINITION_FIELD_SEPARATOR);
+        StringJoiner joiner = new StringJoiner(JabRefPreferences.COL_DEFINITION_FIELD_SEPARATOR);
```
simonharrer: I do not understand this. Why move a constant to JabRefPreferences which is not a preference?

simonharrer: Another comment: the method getDisplayName is called quite often. If we would query a preference for each field in each row having 10k rows, this would lead to performance issues.

lenhard: The code is wrong anyway and somehow slipped my attention, thanks for noticing! It should have been `Globals.prefs.getValue(JabRefPreferences.COL_DEFINITION_FIELD_SEPARATOR)` of course. 

Regarding the migration of constants into preferences: that is what *you* asked me to do in #1408 ;-) However, it is no hassle to move this particular constant back into `GUIGlobals`. 

Will do, as with the rest of your comments

src/main/java/net/sf/jabref/exporter/ExportCustomizationDialog.java
```diff
@@ -61,16 +60,21 @@
 
 public class ExportCustomizationDialog extends JDialog {
 
+    // Column widths for export customization dialog table:
+    private static final int EXPORT_DIALOG_COL_1_WIDTH = 200;
```
tobiasdiez: Remove prefix `Export dialog` in the names?

src/main/java/net/sf/jabref/gui/SidePane.java
```diff
@@ -33,7 +33,9 @@
  */
 public class SidePane extends JPanel {
 
-    private final Dimension PREFERRED_SIZE = new Dimension(GUIGlobals.SPLIT_PANE_DIVIDER_LOCATION, 100);
+    private static final int SPLIT_PANE_DIVIDER_LOCATION = 175;
```
tobiasdiez: Not necessary as a separate variable and can be merged with next line, right?

src/test/java/net/sf/jabref/exporter/BibDatabaseWriterTest.java
tobiasdiez: or in teardown method

tobiasdiez:
LGTM, just some minor remarks.

What I do not understand is the purpose of the bibtex package. Some of the classes seem like they do not have anything to do with writing or parsing bibtex and a rather general, for example the comparators. Maybe it would make sense to completely dissolve the bibtex package by moving the parsing logic to the import package and writing to export...in the end the bibtex format is just a special import-export format.

lenhard:
@tobiasdiez Thanks for your remarks, I will integrate them as well.

As for the bibtex package: Its existence is a historical thing and I do not know the exact motivation behind. It seems like "bundle parsing and writing of pure bibtex", but, with the addition of new importers and exporters, responsibilities shifted and mixed over time. The big advantage we have now is that the package moved to logic, which means that there is a guarantee that it is independent of GUI code and it can be re-used for new GUI stuff with relative safety. If we move it out of logic (to importers or exporters) we loose that advantage. Ultimately, the goal should be to get the GUI independent code from the import/export package into logic as well.

lenhard:
I implemented the above comments and the PR is once again ready to go from my side.

One remark: `COL_DEFINITION_FIELD_SEPARATOR` cannot go to `GUIGlobals`, since this would create dependencies from logic to gui. Therefore, I moved it into `COL_DEFINITION_FIELD_SEPARATOR` to `Globals`. This is fair, I think, since the variables relates to our BibTeX sytnax and field content, which is independent of the GUI.

I still don`t get how the coverage can have increased that much.

PR #1411, opened 2016-05-19
Move Encoding from BasePanel to MetaData

tobiasdiez:

### DIFF
--- a/src/main/java/net/sf/jabref/MetaData.java
+++ b/src/main/java/net/sf/jabref/MetaData.java
@@ -18,6 +18,7 @@ package net.sf.jabref;
 import java.io.IOException;
 import java.io.Reader;
 import java.io.StringReader;
+import java.nio.charset.Charset;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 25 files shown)


lenhard:
Looks good overall.

I guess it makes sense to move the encoding, since it is nothing specific to `BasePanel`. However, it increases method call chains in quite a few places, which get the encoding via the panel. Maybe it is worth to consider leaving the `getEncoding`, `setEncoding` methods in `BasePanel` and have them delegate to `MetaData`, so you can write `panel.getEncoding()` instead of `panel.getBibDatabaseContext().getMetaData().getEncoding()`. WDYT?

tobiasdiez:
Yes you are right, the call chain gets length with this change. In my opinion, this is not really a "bug" but more a feature in the sense that you now see that some other methods should be changed as well. For example the `export` or `saveDatabase` method can now simply accept a BibDatabaseContext and then get the encoding themselves. Moreover, the BasePanel has absolutely no business to know about the file encoding, since it is a GUI class. Thus I would prefer not to reintroduce the shortcut methods.

simonharrer:
Compromise: add a getEncoding method to the BibDatabaseContext.

simonharrer:
There seems to be a lot of NPE according to the travis report. When this is fixed, we can merge this in.

PR #1422, opened 2016-05-19
Resolve #1417 Rename buttons in file list editor

### LINKED ISSUE #1417: Rename Auto button to Get Fulltext
It's hard to understand what `auto`and `download` does.
I think it doesn't really matter for a user that something is performed automatically, but that the outcome is to get the fulltext and save it to the local storage.
Also applies to all other automatically set file links actions that also perform a download.
Additionally all actions must be unified (some don't download)!

stefan-kolb:
from auto->get fulltext, download->download from URL

image


simonharrer:
LGTM

matthiasgeiger:
:+1:

PR #1423, opened 2016-05-19
Refactor Gradle script

stefan-kolb:
We can now
- [ ] split the script into smaller files
- [ ] group similar tasks into self-describing groups so they can be easily found.

I started this attempt with:
- Include localization tasks and package into a group

image


simonharrer:
LGTM

Please update the guide for the translations and the contribution.md as well

stefan-kolb:
@simonharrer Do we want to restructure more in this PR or just include the localization for now?

stefan-kolb:
@simonharrer Please just do a follow up PR with more restructuring, e.g., release groups etc. I need the localization stuff to be in there now.

PR #1426, opened 2016-05-19
Support language keys with spaces for JavaFX

tobiasdiez:
As discussed in https://github.com/JabRef/jabref/pull/968, JavaFX just calls "ResoucreBundle.getString" instead of using our methods in `Localization`. Thus entries like `Copy Version` are not found since they are stored as `Copy_Version`. This PR provides a wrapper which converts keys and translation properly.

### DIFF
--- a/src/main/java/net/sf/jabref/logic/l10n/LocalizationBundle.java
+++ b/src/main/java/net/sf/jabref/logic/l10n/LocalizationBundle.java
@@ -25,6 +25,11 @@ import java.util.Objects;
 import java.util.ResourceBundle;
 import java.util.stream.Collectors;
 
+/**
+ * A bundle containing localized strings.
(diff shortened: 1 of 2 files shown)


src/main/java/net/sf/jabref/logic/l10n/LocalizationBundle.java
```diff
@@ -0,0 +1,48 @@
...
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Enumeration;
+import java.util.List;
+import java.util.Objects;
+import java.util.ResourceBundle;
+import java.util.stream.Collectors;
+
+public class LocalizationBundle extends ResourceBundle {
```
boceckts: I would add a small comment as to why this class is needed.

boceckts:
Looks good and works as expected :+1:

tobiasdiez:
Thanks for the review. I'll merge this in now...

PR #1438, opened 2016-05-22
[WIP] Groups JavaFX Rework

tobiasdiez:
I played around a bit with JavaFX and tried to redesign the group pane. 
The result:
untitled

Question: so far it it is just the ui with dummy test data and without any interaction. If I now start to port everything to JavaFX then the JavaFX branch will diverge immensely from the master. Should I stop here and leave it as a prototype for now or should I finish it? 

### DIFF
--- a/src/main/java/net/sf/jabref/gui/GroupTreeViewModel.java
+++ b/src/main/java/net/sf/jabref/gui/GroupTreeViewModel.java
@@ -20,7 +20,6 @@ import javafx.beans.value.ObservableValue;
 import javafx.css.PseudoClass;
 import javafx.fxml.FXML;
 import javafx.scene.Node;
-import javafx.scene.control.Skin;
 import javafx.scene.control.Tooltip;
 import javafx.scene.control.TreeItem;
 import javafx.scene.control.TreeTableCell;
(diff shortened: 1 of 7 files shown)


simonharrer:
I vote for leaving it for now and focusing on the current system instead.

We try to wait for the feedback on using JavaFX from the stupro to decide how much effort a change would be.

boceckts:
In my opinion there is no problem in partly switching to the new JavaFX technology. Either you embedd JavaFX elements in swing using `JFXPanel` as shown here or you port whole dialogs as shown in #1390 . Looking at the code of both pull requests it is quite easy to separate the JavaFX code from the core swing code.
I also think it looks much better than the more out dated swing ui and is much more customizable. Although I wouldn't change the style of the new elements too much for the beginning as there could be a very inconsitent look and feel against the swing components.

src/main/java/net/sf/jabref/gui/GroupTreeViewModel.java
```diff
@@ -0,0 +1,316 @@
...
+            // simply setting to null is not enough since it would be replaced by the default node in this case
+            row.setDisclosureNode(null);
+            row.disclosureNodeProperty().addListener((observable, oldValue, newValue) -> row.setDisclosureNode(null));
+            return row;
+        });
+
+        groupTree.setRoot(root);
+    }
+
+    public static class CheckBoxTreeTableRowSkin<S> extends TreeTableRowSkin<S> {
```
boceckts: New class should be in a new file

src/main/java/net/sf/jabref/gui/GroupTreeViewModel.java
```diff
@@ -0,0 +1,316 @@
...
+                    disclosureNode.toFront();
+                }
+
+                x += tableCell.getWidth();
+            }
+        }
+        */
+    }
+
+    public static class NewGroupNodeViewModel {
```
boceckts: New class should be in a new file

src/main/resources/net/sf/jabref/gui/GroupTree.css
```diff
@@ -0,0 +1,94 @@
...
+    -fx-font-size: 12px;
+}
+
+.tree-table-row-cell > .tree-table-cell > .icon {
+    -fx-font-family: 'MaterialDesignIcons';
+    -fx-font-size: 16;
+    -fx-fill: #5e5f5e;
+    -fx-translate-x: -5px;
+}
+
```
boceckts: I think it would be better to define a global stylesheet when changing the style of JavaFX predefined style classes such as `.tree-table-view` and use the `@import "/path/to/global.css";`. This way all JavaFX components will have a consistent look and feel.

koppor:
How much work is it to get this merged into javafx? Does the current work on grouping influence this PR?

tobiasdiez:
Around 1-2 weeks for a working minimal version (display of the groups tree and adding/editing of groups) but a few more until all current features are included.

koppor:
I just close this PR as we currently focus on other issues. We will reopen as soon as we continue working on it.

AEgit:
Any news on this? I would still be quite interested in the JavaFX rework, especially since this was supposed to provide the functionality proposed in https://github.com/JabRef/jabref/issues/1904 
The latter would be quite helpful when dealing with many (~1000) groups.

matthiasgeiger:
We will now continue with our efforts to move on to JavaFX (there are already some dialogs implemented).

However, I don't know if and when @tobiasdiez will be able to continue on this.

AEgit:
Thanks - hmm, I see. Do you think it would be necessary to reopen this issue (just to raise awareness)? Or is it all right the way it's now?

PR #1440, opened 2016-05-23
Fix #1420 Auto downloader should respect file pattern and propose correct filename

### LINKED ISSUE #1420: Auto downloader should respect file pattern and propose correct filename
Currently it just renames to `\bibtexkey`

stefan-kolb:
image

### DIFF
--- a/src/main/java/net/sf/jabref/external/DownloadExternalFile.java
+++ b/src/main/java/net/sf/jabref/external/DownloadExternalFile.java
@@ -26,6 +26,7 @@ import javax.swing.JOptionPane;
 import javax.swing.SwingUtilities;
 
 import net.sf.jabref.BibDatabaseContext;
+import net.sf.jabref.Globals;
 import net.sf.jabref.JabRefExecutorService;
 import net.sf.jabref.gui.FileListEntry;
 import net.sf.jabref.gui.FileListEntryEditor;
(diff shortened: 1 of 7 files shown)


simonharrer:
LGTM

src/main/java/net/sf/jabref/external/DownloadExternalFile.java
```diff
@@ -267,7 +268,8 @@ private void downloadFinished() {
     }
     // FIXME: will break download if no bibtexkey is present!
     private String getSuggestedFileName(String suffix) {
-        String plannedName = bibtexKey == null ? "set-filename" : bibtexKey;
+        String plannedName = FileUtil.createFileNameFromPattern(databaseContext.getDatabase(), frame.getCurrentBasePanel().getSelectedEntries().get(0), Globals.journalAbbreviationLoader.getRepository());
```
tobiasdiez: So the field `bibtexKey` is no longer required, right?

tobiasdiez:
LGTM :+1:

src/main/java/net/sf/jabref/external/DownloadExternalFile.java
stefan-kolb: bibtexkey is used inside createFileNameFromPattern

PR #1453, opened 2016-05-26
Refactor placement of parsing and writing logic

### LINKED ISSUE #888: [Code] Placement of parsing and writing logic
Right now parsing and serialization is not handled in a uniform way. (see e.g #875)

Possible designs which come to my mind:
- The BibParser and BibWriter contain all the logic as private methods.
 - e.g.: BibParser.parseMetaData(String stringyMetadata) and BibWriter.writeMetaData(Metadata)
 -  pro: encapsulation of responsibilities
 - contra: rather big classes BibParser and BibWriter, hard to test
2. There are transformer classes which just convert the string to the object and vice-versa.
  - e.g.: MetadataTransformer with static parse(String stringyMetadata) and static write(Metadata) [suggestions for a better name then transformer? ReadWriter?]
   - pro: encapsulation of responsibilities (but weaker as 1), can be reused somewhere else, better testability
   - contra: exposes writing and parsing to the outside world, many new classes
3. The corresponding objects have methods which allow parsing and converting to string.
 - e.g.: Metadata has static parse(String stringyMetadata) [or constructor?] and serialize()
  - pro: encapsulation of responsibilities (but weaker as 1 and 2), can be reused somewhere else, better testability as 1
   - contra: exposes writing and parsing to the outside world, scatters parsing and writing logic 

I have no strong opinion about which option we should use (slightly preferring 2), but we should decide for *one* to handle parsing and writing uniformly.

tobiasdiez:
According to the following conventions:
- All objects which can be read from a string have `public static Type parse(Stringy input)` method.
- All objects which can be written to a string have a `public Stringy getAsString()` method.

Should fix #888.

--

### DIFF
--- a/src/main/java/net/sf/jabref/MetaData.java
+++ b/src/main/java/net/sf/jabref/MetaData.java
@@ -24,7 +24,6 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
-import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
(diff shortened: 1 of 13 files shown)


simonharrer:
I am still an advocate of separate read/write classes which handle the serialization in both ways because:

- Using an Interface with fromString/toString methods would allow to make this more explicit in the code and ensure that these methods are not renamed by accident
- Using separate classes for reading/writing would make the classes themselfs smaller and allow a better separation of concern. 
- Easier to test as one only needs to look at a single cread/write class with its own test.

tobiasdiez:
Yes, you are right that separate read/write classes are probably even better. Nonetheless, I find the changes made in this PR are still an improvement and I'm quite satisfied with the result. (Satisfied enough to have no motivation for introducing new read/write classes)

src/main/java/net/sf/jabref/MetaData.java
```diff
@@ -352,7 +350,7 @@ public boolean isProtected() {
     /**
      * Writes all data in the format <key, serialized data>.
      */
-    public Map<String, String> serialize() {
+    public Map<String, String> getAsString() {
```
lenhard: I am not so happy with naming convention here (and below). After all, this returns no String as the name indicates. Could we make the convention a little more flexible to better describe the result, say: `getAsStringMap`, `getAsStringList`, etc.?

simonharrer: I think that makes sense. Can you change that please @tobiasdiez and merge it in? :trophy:

tobiasdiez:
I changed the names. Thanks for the feedback!

Issue #661, opened 2016-01-10
Add auto-updater for development builds


tobiasdiez:
In view of the more regular release schedule, it would be nice if JabRef can perform an auto-update.

stefan-kolb:
Definitely :+1:

koppor:
install4j offers that functionality and also generates an [updates.xml](https://circle-artifacts.com/gh/JabRef/jabref/2337/artifacts/0/home/ubuntu/jabref/build/releases/updates.xml). Have to investigate where to put that file and how to automate that process.

Currently, I think, it should be put at https://www.jabref.org/updates.xml. Since that page is served via github, we currently do not get a valid SSL certificate. [It doesn't seem possible to offer SSL at custom domains](https://konklone.com/post/github-pages-now-sorta-supports-https-so-use-it). Thus, I have to provide https://updates.jabref.org/updates.xml or https://builds.jabref.org/updates.xml or https://files.jabref.org/updates.xml. Think, builds.jabref.org is the right choice, isn't it?

simonharrer:
:+1: but needs to be automated so this is not too much a hassle.

tobiasdiez:
[Apperently](http://stackoverflow.com/questions/8226486/how-to-configure-install4j-versioning-to-allow-both-stable-and-beta-alpha-upgrad), it is relatively easy to support a special channel for dev-builds.

obraliar:
Stupro features team would like to take care about this issue for the 3rd sprint which just started.

stefan-kolb:
This would free users from downloading from our affiliate sites. How much income do we potentially lose with this?

koppor:
The income numbers are confidential, because of company regulations.

simonharrer:
Ask github API whether newer version exists
Menu option: Help - Check for update
During startup: check and if newer display this somehow

matthiasgeiger:
Important for usability: Checkbox: "Do not notify for this update again" or sth like this.

matthiasgeiger:
Example from TexStudio:
![auto-update](https://cloud.githubusercontent.com/assets/676652/15392834/ea729a98-1dc9-11e6-8c1e-99fb76a53f22.png)

@obraliar Any news on this topic?

obraliar:
Due to decision problems we switched to other topic in this sprint. Full steam ahead now we are working on shared-database support. Should we implement only a checker (see https://github.com/JabRef/jabref/issues/984) without updating the software?
[...]

PR #1459, opened 2016-05-30
create update checker for the enduser

bartsch-dev:
Implements #661.

The updater regularly checks if there is a new stable version of JabRef available [via github API](https://api.github.com/repos/JabRef/jabref/releases/latest).
If there is, a dialog will popup:

newversion
- `Download update` points directly to [fosshub](http://www.fosshub.com/JabRef.html)
- `Ignore this update` will ignore this specific version on automatic version checks
- `Remind me later` will inform the user on the next version check (automatic or manual) of this version
- `View changelog` points directly to the changelog specific to this version (eg. v3.3 would point to https://github.com/JabRef/jabref/blob/v3.3/CHANGELOG.md)

Under `Help` there is another button to check for updates.

updatetab
under preferences you can see:
- the current version
- when the version was last checked and a button to check it now
- if there's a new version available and a download button
- if a version was ignored and a button to unignore it
- you can also define how often the automatic check should occur

### DIFF
--- a/src/main/java/net/sf/jabref/JabRefGUI.java
+++ b/src/main/java/net/sf/jabref/JabRefGUI.java
@@ -36,12 +36,15 @@ import net.sf.jabref.gui.GUIGlobals;
 import net.sf.jabref.gui.JabRefFrame;
 import net.sf.jabref.gui.ParserResultWarningDialog;
 import net.sf.jabref.gui.util.FocusRequester;
+import net.sf.jabref.gui.worker.VersionWorker;
 import net.sf.jabref.importer.AutosaveStartupPrompter;
 import net.sf.jabref.importer.OpenDatabaseAction;
 import net.sf.jabref.importer.ParserResult;
 import net.sf.jabref.logic.l10n.Localization;
 import net.sf.jabref.logic.preferences.LastFocusedTabPreferences;
 import net.sf.jabref.logic.util.OS;
+import net.sf.jabref.logic.util.Version;
+import net.sf.jabref.logic.util.VersionPreferences;
 import net.sf.jabref.migrations.PreferencesMigrations;
 
 import com.jgoodies.looks.plastic.Plastic3DLookAndFeel;
@@ -65,6 +68,13 @@ public class JabRefGUI {
         this.loaded = loaded;
         this.isBlank = isBlank;
         openWindow();
+        JabRefGUI.checkForNewVersion(false);
+    }
+
+    public static void checkForNewVersion(boolean manualExecution) {
+        Version toBeIgnored = new VersionPreferences(Globals.prefs).getIgnoredVersion();
+        Version currentVersion = Globals.BUILD_INFO.getVersion();
+        new VersionWorker(JabRefGUI.getMainFrame(), manualExecution, currentVersion, toBeIgnored).execute();
     }
 
     private void openWindow() {
--- a/src/main/java/net/sf/jabref/JabRefPreferences.java
+++ b/src/main/java/net/sf/jabref/JabRefPreferences.java
@@ -1,4 +1,4 @@
-/*  Copyright (C) 2003-2015 JabRef contributors.
+/*  Copyright (C) 2003-2016 JabRef contributors.
  This program is free software; you can redistribute it and/or modify
  it under the terms of the GNU General Public License as published by
  the Free Software Foundation; either version 2 of the License, or
@@ -70,6 +70,7 @@ import net.sf.jabref.logic.openoffice.OpenOfficePreferences;
 import net.sf.jabref.logic.openoffice.StyleLoader;
 import net.sf.jabref.logic.remote.RemotePreferences;
 import net.sf.jabref.logic.util.OS;
+import net.sf.jabref.logic.util.VersionPreferences;
 import net.sf.jabref.logic.util.strings.StringUtil;
 import net.sf.jabref.model.database.BibDatabaseMode;
 import net.sf.jabref.model.entry.CustomEntryType;
@@ -419,7 +420,7 @@ public class JabRefPreferences {
     private static final String USER_HOME = System.getProperty("user.home");
 
     /**
-     * Set with all custom {@link ImportFormat}s
+     * Set with all custom {@link net.sf.jabref.importer.fileformat.ImportFormat}s
      */
     public final CustomImportList customImports;
 
@@ -861,6 +862,9 @@ public class JabRefPreferences {
         defaults.put(MAX_BACK_HISTORY_SIZE, 10);
         defaults.put(LINE_LENGTH, 65);
         defaults.put(INDENT, 4);
+
+        //versioncheck defaults
+        defaults.put(VersionPreferences.VERSION_IGNORED_UPDATE, "");
     }
 
     public String getUser() {
--- a/src/main/java/net/sf/jabref/cli/JabRefCLI.java
+++ b/src/main/java/net/sf/jabref/cli/JabRefCLI.java
@@ -1,3 +1,18 @@
+/*  Copyright (C) 2003-2016 JabRef contributors.
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License along
+    with this program; if not, write to the Free Software Foundation, Inc.,
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
 package net.sf.jabref.cli;
 
 import net.sf.jabref.Globals;
--- a/src/main/java/net/sf/jabref/gui/JabRefFrame.java
+++ b/src/main/java/net/sf/jabref/gui/JabRefFrame.java
@@ -92,6 +92,7 @@ import net.sf.jabref.gui.actions.MnemonicAwareAction;
 import net.sf.jabref.gui.actions.NewDatabaseAction;
 import net.sf.jabref.gui.actions.NewEntryAction;
 import net.sf.jabref.gui.actions.NewSubDatabaseAction;
+import net.sf.jabref.gui.actions.SearchForUpdateAction;
 import net.sf.jabref.gui.actions.SortTabsAction;
 import net.sf.jabref.gui.dbproperties.DatabasePropertiesDialog;
 import net.sf.jabref.gui.groups.EntryTableTransferHandler;
@@ -715,7 +716,7 @@ public class JabRefFrame extends JFrame implements OutputPrinter {
 
     // General preferences dialog.  The MacAdapter calls this method when "Preferences..."
     // is selected from the application menu.
-    public void preferences() {
+    public void showPreferencesDialog() {
         output(Localization.lang("Opening preferences..."));
         if (prefsDialog == null) {
             prefsDialog = new PreferencesDialog(JabRefFrame.this);
@@ -1354,6 +1355,7 @@ public class JabRefFrame extends JFrame implements OutputPrinter {
         helpMenu.add(forkMeOnGitHubAction);
         helpMenu.add(donationAction);
         helpMenu.addSeparator();
+        helpMenu.add(new SearchForUpdateAction());
         helpMenu.add(about);
         mb.add(helpMenu);
 
@@ -1476,6 +1478,9 @@ public class JabRefFrame extends JFrame implements OutputPrinter {
         tlb.add(donationAction);
     }
 
+    /**
+     * displays the String on the Status Line visible on the bottom of the JabRef mainframe
+     */
     public void output(final String s) {
         SwingUtilities.invokeLater(() -> {
             statusLine.setText(s);
@@ -1703,7 +1708,7 @@ public class JabRefFrame extends JFrame implements OutputPrinter {
 
         @Override
         public void actionPerformed(ActionEvent e) {
-            preferences();
+            showPreferencesDialog();
         }
     }
 
--- a/src/main/java/net/sf/jabref/gui/desktop/JabRefDesktop.java
+++ b/src/main/java/net/sf/jabref/gui/desktop/JabRefDesktop.java
@@ -1,3 +1,18 @@
+/*  Copyright (C) 2003-2016 JabRef contributors.
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License along
+    with this program; if not, write to the Free Software Foundation, Inc.,
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
 package net.sf.jabref.gui.desktop;
 
 import java.io.File;
@@ -12,10 +27,12 @@ import javax.swing.JOptionPane;
 
 import net.sf.jabref.BibDatabaseContext;
 import net.sf.jabref.Globals;
+import net.sf.jabref.JabRefGUI;
 import net.sf.jabref.external.ExternalFileType;
 import net.sf.jabref.external.ExternalFileTypeEntryEditor;
 import net.sf.jabref.external.ExternalFileTypes;
 import net.sf.jabref.external.UnknownExternalFileType;
+import net.sf.jabref.gui.ClipBoardManager;
 import net.sf.jabref.gui.FileListEntry;
 import net.sf.jabref.gui.FileListEntryEditor;
 import net.sf.jabref.gui.FileListTableModel;
@@ -263,6 +280,27 @@ public class JabRefDesktop {
     public static void openBrowser(String url) throws IOException {
         Optional<ExternalFileType> fileType = ExternalFileTypes.getInstance().getExternalFileTypeByExt("html");
         openExternalFilePlatformIndependent(fileType, url);
+    }
+
+    /**
+     * Opens the url with the users standard Browser.
+     * If that fails a popup will be shown to instruct the user to open the link manually
+     * and the link gets copied to the clipboard
+     * @param url
+     */
+    public static void openBrowserShowPopup(String url) {
+        try {
+            openBrowser(url);
+        } catch (IOException exception) {
+            new ClipBoardManager().setClipboardContents(url);
+            LOGGER.error("Could not open browser", exception);
+            String couldNotOpenBrowser = Localization.lang("Could not open browser.");
+            String openManually = Localization.lang("Please open %0 manually.", url);
+            String copiedToClipboard = Localization.lang("The_link_has_been_copied_to_the_clipboard.");
+            JabRefGUI.getMainFrame().output(couldNotOpenBrowser);
+            JOptionPane.showMessageDialog(JabRefGUI.getMainFrame(), couldNotOpenBrowser + "\n" + openManually +"\n"+
+                    copiedToClipboard, couldNotOpenBrowser, JOptionPane.ERROR_MESSAGE);
+        }
     }
 
     public static void openConsole(File file) throws IOException {
--- a/src/main/java/net/sf/jabref/gui/help/NewVersionDialog.java
+++ b/src/main/java/net/sf/jabref/gui/help/NewVersionDialog.java
@@ -1,0 +1,110 @@
+/*  Copyright (C) 2016 JabRef contributors.
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License along
+    with this program; if not, write to the Free Software Foundation, Inc.,
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+package net.sf.jabref.gui.help;
+
+import java.awt.Cursor;
+import java.awt.GridBagConstraints;
+import java.awt.GridBagLayout;
+import java.awt.Insets;
+import java.awt.event.MouseEvent;
+
+import javax.swing.JButton;
+import javax.swing.JDialog;
+import javax.swing.JFrame;
+import javax.swing.JLabel;
+import javax.swing.JPanel;
+import javax.swing.event.MouseInputAdapter;
+
+import net.sf.jabref.Globals;
+import net.sf.jabref.gui.desktop.JabRefDesktop;
+import net.sf.jabref.logic.l10n.Localization;
+import net.sf.jabref.logic.util.Version;
+import net.sf.jabref.logic.util.VersionPreferences;
+
+public class NewVersionDialog extends JDialog {
+
+    public NewVersionDialog(JFrame frame, Version currentVersion, Version latestVersion, Version toBeIgnored) {
+        super(frame);
+        setTitle(Localization.lang("New version available"));
+
+        JLabel lblTitle = new JLabel(Localization.lang("A new version of JabRef has been released."));
+        JLabel lblCurrentVersion = new JLabel(Localization.lang("Installed version") + ": " + currentVersion.getFullVersion());
+        JLabel lblLatestVersion = new JLabel(Localization.lang("Latest version") + ": " + latestVersion.getFullVersion());
+
+        String localization = Localization.lang("To see what's new view the changelog.");
+        JLabel lblMoreInformation = new JLabel("<HTML><a href=" + latestVersion.getChangelogUrl() + "'>" + localization + "</a></HTML>");
+        lblMoreInformation.setCursor(Cursor.getPredefinedCursor(Cursor.HAND_CURSOR));
+        lblMoreInformation.addMouseListener(new MouseInputAdapter() {
+            @Override
+            public void mouseClicked(MouseEvent e) {
+                JabRefDesktop.openBrowserShowPopup(latestVersion.getChangelogUrl());
+            }
+        });
+
+        JButton btnIgnoreUpdate = new JButton(Localization.lang("Ignore this update"));
+        btnIgnoreUpdate.addActionListener(e -> {
+            new VersionPreferences(Globals.prefs).setAsIgnoredVersion(toBeIgnored);
+            dispose();
+        });
+
+        JButton btnDownloadUpdate = new JButton(Localization.lang("Download update"));
+        btnDownloadUpdate.addActionListener(e -> {
+            JabRefDesktop.openBrowserShowPopup(Version.JABREF_DOWNLOAD_URL);
+            dispose();
+        });
+
+        JButton btnRemindMeLater = new JButton(Localization.lang("Remind me later"));
+        btnRemindMeLater.addActionListener(e -> dispose());
+
+        JPanel panel = new JPanel();
+        panel.setLayout(new GridBagLayout());
+        GridBagConstraints c = new GridBagConstraints();
+        c.gridheight = 1;
+        c.fill = GridBagConstraints.BOTH;
+        c.insets = new Insets(2, 5, 5, 2);
+
+        c.gridx = c.gridy = 0;
+        c.gridwidth = 3;
+        panel.add(lblTitle, c);
+
+        c.gridy = 1;
+        panel.add(lblCurrentVersion, c);
+
+        c.gridy = 2;
+        panel.add(lblLatestVersion, c);
+
+        c.gridy = 3;
+        panel.add(lblMoreInformation, c);
+
+        c.gridy = 4;
+        c.gridx = 0;
+        c.gridwidth = 1;
+        panel.add(btnDownloadUpdate, c);
+
+        c.gridx = 1;
+        panel.add(btnIgnoreUpdate, c);
+
+        c.gridx = 2;
+        panel.add(btnRemindMeLater, c);
+
+        add(panel);
+        pack();
+        setLocationRelativeTo(frame);
+        setModalityType(ModalityType.APPLICATION_MODAL);
+        setVisible(true);
+    }
+
+}
--- a/src/main/java/net/sf/jabref/gui/worker/VersionWorker.java
+++ b/src/main/java/net/sf/jabref/gui/worker/VersionWorker.java
@@ -1,0 +1,101 @@
+/*  Copyright (C) 2003-2016 JabRef contributors.
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License along
+    with this program; if not, write to the Free Software Foundation, Inc.,
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+package net.sf.jabref.gui.worker;
+
+import java.io.IOException;
+import java.util.Objects;
+import java.util.concurrent.ExecutionException;
+
+import javax.swing.JOptionPane;
+import javax.swing.SwingWorker;
+
+import net.sf.jabref.gui.JabRefFrame;
+import net.sf.jabref.gui.help.NewVersionDialog;
+import net.sf.jabref.logic.l10n.Localization;
+import net.sf.jabref.logic.util.Version;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+public class VersionWorker extends SwingWorker<Version, Void> {
+
+    private static final Log LOGGER = LogFactory.getLog(VersionWorker.class);
+
+    private final JabRefFrame mainFrame;
+    private final boolean manualExecution;
+    private final Version installedVersion;
+    private final Version toBeIgnored;
+
+    public VersionWorker(JabRefFrame mainFrame, boolean manualExecution, Version installedVersion, Version toBeIgnored) {
+        this.mainFrame = Objects.requireNonNull(mainFrame);
+        this.manualExecution = manualExecution;
+        this.installedVersion = Objects.requireNonNull(installedVersion);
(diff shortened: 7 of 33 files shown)


src/main/java/net/sf/jabref/logic/util/version/Version.java
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * Represents the Application Version with the major and minor number, the full Version String and if it's a developer version
+ */
+public class Version {
+
+    private String fullVersion;
+    private int major;
+    private int minor;
+    private boolean devVersion;
```
simonharrer: please make all the fields final

src/main/java/net/sf/jabref/logic/util/version/Version.java
```diff
@@ -0,0 +1,98 @@
...
+    public Version(String version) {
+        if (version.equals("")) {
+            this.fullVersion = BuildInfo.UNKOWN_VERSION;
+            this.major = this.minor = 0;
+            this.devVersion = false;
+        } else {
+            this.fullVersion = version;
+            String[] versionParts = version.split(Pattern.quote("."));
+            major = Integer.parseInt(versionParts[0]);
+            minor = Integer.parseInt(versionParts[1].split("dev")[0]);
```
simonharrer: what happens when I use a version which does not have a "."? Can you add a test for this class?

src/main/java/net/sf/jabref/logic/util/version/Version.java
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    /**
+     * @return true if this version is newer than the passed one
+     */
+    public boolean isNewerThan(Version otherVersion) {
+        if (this.getFullVersion().equals(BuildInfo.UNKOWN_VERSION)) {
+            return false;
+        }
+        if (otherVersion.getFullVersion().equals(BuildInfo.UNKOWN_VERSION)) {
```
simonharrer: Try to use Objects.equals instead

src/main/java/net/sf/jabref/logic/util/version/VersionInformation.java
```diff
@@ -0,0 +1,184 @@
...
+                }
+                else if (manualExecution){
+                    JOptionPane.showMessageDialog(JabRefGUI.getMainFrame(),
+                            Localization.lang("uptodate"),
+                            Localization.lang("uptodate"),
+                            JOptionPane.INFORMATION_MESSAGE);
+                }
+                JabRefGUI.getMainFrame().updatePreferenceDialog();
+            }
+        }.start();
```
simonharrer: Threads should be started via the JabRefExecutorService. But if you want to update the UI upon completion, you need a SwingWorker which ensures that one part is executed outside the EDT, and the second part is executed on the EDT. Another issue here is that this class in the logic package, but makes use of GUI classes - this is forbidden.

src/main/java/net/sf/jabref/logic/util/version/VersionInformation.java
```diff
@@ -0,0 +1,184 @@
...
+
+    /**
+     * Use this method for the background version Check
+     * checks if an update is available if enough time has passed since the last check
+     */
+    public void checkIfNecessary() {
+        LocalDateTime lastCheckedDate = LocalDateTime.parse(Globals.prefs.get(JabRefPreferences.VERSION_LAST_CHECKED_DATE));
+        LocalDateTime currentDate = LocalDateTime.now();
+
+        switch (VersionCheckMode.valueOf(Globals.prefs.get(JabRefPreferences.VERSION_CHECK_EVERY))) {
```
simonharrer: The logic package has no notion of the preferences. This is only known in the GUI package. You should subdivide this class in a way that the GUI relevant code stays in the code package (using preferences and GUI classes) whereas the remaining stuff should be put in the logic package.

src/main/java/net/sf/jabref/logic/util/version/VersionInformation.java
```diff
@@ -0,0 +1,184 @@
...
+     *
+     * @param manualExecution if true a Dialog will popup if the update check fails
+     */
+    private void grabLatestVersion(boolean manualExecution) {
+        try {
+            URLConnection connection = new URL(JABREF_GITHUB_URL).openConnection();
+            connection.setRequestProperty("Accept-Charset", "UTF-8");
+            BufferedReader rd = new BufferedReader(new InputStreamReader(connection.getInputStream()));
+            JSONObject obj = new JSONObject(rd.readLine());
+            latestVersion = new Version(obj.getString("tag_name").replaceFirst("v", ""));
```
simonharrer: I would extract a class which simply gets the latest version, e.g., a VersionChecker, with a method Optional<Version> getLatestVersion() as the request can fail returning no version at all.

src/main/resources/l10n/JabRef_da.properties
```diff
@@ -1687,3 +1687,23 @@ Get_fulltext=
...
+update=
+new_version_available=
+last_checked=
+ignore_following_update=
+current_version=
+remind_me_later=
+ignore_update=
+weekly=
+monthly=
+couldnt_connect_to_the_updateserver=
```
simonharrer: A ' sign seems to be missing here. Also, an underscore/space is missing between update and server

src/main/java/net/sf/jabref/logic/util/version/VersionInformation.java
bartsch-dev: Would not this be a violation of the MVC pattern? When the GUI is accessing the date from the preferences, I only see two ways to check if I need to look for a new version. One would be to put the logic behind it into the GUI. Te other would be to access the GUI from within the logic. Am I getting it wrong?

simonharrer: In the GUI you do the following:
- get current version
- use swing worker
  - (outside EDT) use a logic class to get the latest version, you can use another class which knows whether the latest version is the current version as the current version can be passed to the swing worker
  - (inside EDT) update UI by showing the user something and updating the preferences

src/main/java/net/sf/jabref/logic/util/version/Version.java
bartsch-dev: I added some test cases for Version.
The version now only has to own the major part, the minor one is optional.

src/main/java/net/sf/jabref/gui/actions/version/IgnoreUpdateAction.java
```diff
@@ -0,0 +1,40 @@
...
+import net.sf.jabref.JabRefPreferences;
+import net.sf.jabref.logic.l10n.Localization;
+import net.sf.jabref.logic.util.version.Version;
+
+public class IgnoreUpdateAction extends AbstractAction {
+
+    private Version toBeIgnored;
+
+    public IgnoreUpdateAction(Version toBeIgnored){
+        super(Localization.lang("Ignore_this_update"));
```
boceckts: Use the string without spaces. This applies for all localization strings you used. For more information see [using localization correctly](https://github.com/JabRef/jabref/wiki/Code-Howtos#using-localization-correctly).

src/main/java/net/sf/jabref/gui/worker/VersionWorker.java
```diff
@@ -0,0 +1,92 @@
...
+
+    public VersionWorker(JabRefFrame mainFrame, boolean manualExecution, Version currentVersion, Version toBeIgnored) {
+        this.mainFrame = Objects.requireNonNull(mainFrame);
+        this.manualExecution = manualExecution;
+        this.currentVersion = Objects.requireNonNull(currentVersion);
+        this.toBeIgnored = Objects.requireNonNull(toBeIgnored);
+    }
+
+    @Override
+    protected Object doInBackground() throws Exception {
```
boceckts: SwingWorker should infer generic type arguments. Since this method only returns `null` objects you should add the type arguments `Void` and change this methods signature to `protected Void doInBackground() throws Exception`

src/main/java/net/sf/jabref/gui/worker/VersionWorker.java
```diff
@@ -0,0 +1,92 @@
...
+            }
+            this.mainFrame.output(Localization.lang("couldn't_connect_to_the_update_server") + " "
+                    + Localization.lang("Please_try_again_later_and/or_check_your_network_connection"));
+            LOGGER.warn("Couldn't connect to the updateserver.", ioException);
+            return null;
+        }
+
+        /*
+        only respect the ignored version on automated version checks
+         */
```
boceckts: Single line comments should use double slashes `//`

src/main/resources/l10n/JabRef_en.properties
```diff
@@ -2259,4 +2259,17 @@ Decryption_not_supported.=Decryption_not_supported.
...
+Download_update=Download_update
+New_version_available=New_version_available
+Current_version=Current_version
+Remind_me_later=Remind_me_later
+Ignore_this_update=Ignore_this_update
+couldn't_connect_to_the_update_server=Couldn't connect to the update server.
+Please_try_again_later_and/or_check_your_network_connection=Please try again later and/or check your network connection
+To_see_what's_new_view_the_changelog=To see what's new view the changelog
+A_new_version_of_JabRef_has_been_released_and_can_be_updated_now=A new version of JabRef has been released and can be updated now
+JabRef_is_uptodate=JabRef is uptodate
```
boceckts: Underscores are required here. Also "uptodate" looks weird, does this actually exist? I would replace it with "up to date" or "up-to-date".

src/main/java/net/sf/jabref/gui/actions/version/VisitDownloadPageAction.java
```diff
@@ -0,0 +1,37 @@
...
+
+public class VisitDownloadPageAction extends AbstractAction {
+
+    public VisitDownloadPageAction() {
+        super(Localization.lang("Download_update"));
+    }
+
+    @Override
+    public void actionPerformed(ActionEvent e) {
+        JabRefDesktop.openBrowserCatchException(VersionChecker.JABREF_DOWNLOAD_URL);
```
boceckts: In my opinion you should create a new method `VersionChecker.goToLatestVersionDownload()` and put the `JabrefDesktop.openBrowserCatchException` method in there. This way `JABREF_DOWNLOAD_URL` can be private.

src/main/java/net/sf/jabref/gui/desktop/JabRefDesktop.java
```diff
@@ -265,6 +280,14 @@ public static void openBrowser(String url) throws IOException {
...
     }
 
+    public static void openBrowserCatchException(String url) {
+        try {
+            openBrowser(url);
+        } catch (IOException exception) {
+            LOGGER.error("Error while opening following ling with a browser: "+ url, exception);
+        }
+    }
+
```
boceckts: This is unnecessary as you can use the `openBrowser(String url)` method. This way you can react with a more specific error message or further processing.

src/main/java/net/sf/jabref/gui/actions/version/RemindMeLaterAction.java
```diff
@@ -0,0 +1,34 @@
...
+
+    public RemindMeLaterAction(){
+        super(Localization.lang("Remind me later"));
+    }
+
+    @Override
+    public void actionPerformed(ActionEvent e) {
+        //do nothing, when the version is checked the next time the newVersionDialog will popup automatically
+    }
+}
```
boceckts: Why does does this action exist if it does nothing? Or am I missing something?

src/main/java/net/sf/jabref/gui/desktop/JabRefDesktop.java
```diff
@@ -265,6 +280,14 @@ public static void openBrowser(String url) throws IOException {
         openExternalFilePlatformIndependent(fileType, url);
     }
 
+    public static void openBrowserCatchException(String url) {
+        try {
+            openBrowser(url);
+        } catch (IOException exception) {
+            LOGGER.error("Error while opening following ling with a browser: "+ url, exception);
```
boceckts: > "Error while opening following ling with a browser: "

Typo?

boceckts:
Your tests mostly cover the `Version` class, did you also test (at least manually) if the new version dialog shows up as expected (on the start if newer version is available or not if you already pressed ignore this update)? I guess this is not easy to test with real data as you are already on the latest version when running this branch.

src/main/java/net/sf/jabref/JabRefGUI.java
```diff
@@ -65,6 +67,13 @@ public JabRefGUI(List<ParserResult> loaded, boolean isBlank) {
         this.loaded = loaded;
         this.isBlank = isBlank;
         openWindow();
+        JabRefGUI.checkVersion(false);
+    }
+
+    public static void checkVersion(boolean manualExecution){
+        Version toBeIgnored = new Version(Globals.prefs.get(JabRefPreferences.VERSION_IGNORED_UPDATE));
+        Version currentVersion = Globals.BUILD_INFO.getVersion();
+        new VersionWorker(JabRefGUI.getMainFrame(), manualExecution, currentVersion, toBeIgnored).execute();
```
boceckts: This means a new `VersionWorker` is created and started each time the method is executed. Here it might not be as impact full because the user probably won't check the version a hundred times manually but you should always cancel a background task before creating the same one again. You can cancel a SwingWorker by creating an object and calling `cancel(true)` on it.

boceckts:
Good work 👍
I have made a few minor comments but when they are addressed it's fine by my side.

src/main/java/net/sf/jabref/JabRefGUI.java
```diff
@@ -65,6 +67,13 @@ public JabRefGUI(List<ParserResult> loaded, boolean isBlank) {
         this.loaded = loaded;
         this.isBlank = isBlank;
         openWindow();
+        JabRefGUI.checkVersion(false);
+    }
+
+    public static void checkVersion(boolean manualExecution){
```
tobiasdiez: Please move this to VersionWorker (or Version?) and rename to checkForNewVersion.

src/main/java/net/sf/jabref/JabRefPreferences.java
```diff
@@ -54,6 +54,7 @@
 import net.sf.jabref.gui.maintable.PersistenceTableColumnListener;
 import net.sf.jabref.gui.preftabs.ImportSettingsTab;
 import net.sf.jabref.importer.CustomImportList;
+import net.sf.jabref.importer.fileformat.ImportFormat;
```
tobiasdiez: Why did this changed?

src/main/java/net/sf/jabref/logic/util/version/Version.java
```diff
@@ -0,0 +1,129 @@
...
+     * @return true if this version is newer than the passed one
+     */
+    /* Ignoring the SimplifyBooleanReturns request from PMD, you could write it all in one line but then it's unreadable */
+    @SuppressWarnings("PMD.SimplifyBooleanReturns")
+    public boolean isNewerThan(Version otherVersion) {
+        Objects.requireNonNull(otherVersion);
+        if (Objects.equals(this, otherVersion)) {
+            return true;
+        }
+        else if (this.getFullVersion().equals(BuildInfo.UNKOWN_VERSION)) {
```
tobiasdiez: Else on the same line as the closing bracket, i.e. `} else if (...)`  (I think)

src/main/java/net/sf/jabref/cli/JabRefCLI.java
```diff
@@ -244,7 +259,7 @@ public void printUsage() {
     }
 
     private String getVersionInfo() {
-        return String.format("JabRef %s", Globals.BUILD_INFO.getVersion());
+        return String.format("JabRef %s", Globals.BUILD_INFO.getVersion().getFullVersion());
```
tobiasdiez: implement Version.toString() and use it here.

src/main/java/net/sf/jabref/gui/JabRefFrame.java
```diff
@@ -1476,6 +1488,9 @@ private void createToolBar() {
         tlb.add(donationAction);
     }
 
+    /**
+     * displays the String on the Status Line visible on the bottom of the JabRef mainframe
```
tobiasdiez: I like that you add documenting comments!

src/main/java/net/sf/jabref/gui/actions/CopyVersionToClipboardAction.java
```diff
@@ -29,8 +29,8 @@
 
     @Override
     public void actionPerformed(ActionEvent e) {
-        String info = String.format("JabRef %s%n%s %s %s %nJava %s", Globals.BUILD_INFO.getVersion(), BuildInfo.OS,
-                BuildInfo.OS_VERSION, BuildInfo.OS_ARCH, BuildInfo.JAVA_VERSION);
+        String info = String.format("JabRef %s%n%s %s %s %nJava %s", Globals.BUILD_INFO.getVersion().getFullVersion(),
```
tobiasdiez: toString as above.

src/main/java/net/sf/jabref/gui/JabRefFrame.java
```diff
@@ -728,6 +729,16 @@ public void preferences() {
         output("");
     }
 
+    /**
+     * updates the values within the Preferences Dialog
+     */
+    public void updatePreferenceDialog() {
```
tobiasdiez: As far as I can see, this method is not used?

src/main/java/net/sf/jabref/gui/help/NewVersionDialog.java
```diff
@@ -0,0 +1,105 @@
...
+        JLabel lblMoreInformation = new JLabel("<HTML><a href=" + latestVersion.getChangelogUrl() + "'>" + localization + "</a></HTML>");
+        lblMoreInformation.setCursor(Cursor.getPredefinedCursor(Cursor.HAND_CURSOR));
+        lblMoreInformation.addMouseListener(new MouseAdapter() {
+            @Override
+            public void mouseClicked(MouseEvent e) {
+                JabRefDesktop.openBrowserCatchException(latestVersion.getChangelogUrl());
+            }
+        });
+
+        JButton btnIgnoreUpdate = new JButton(new IgnoreUpdateAction(toBeIgnored));
```
tobiasdiez: No need to have a bunch of classes for each action, just use lambda expressions.

src/main/java/net/sf/jabref/gui/actions/version/IgnoreUpdateAction.java
```diff
@@ -0,0 +1,40 @@
...
+    private Version toBeIgnored;
+
+    public IgnoreUpdateAction(Version toBeIgnored){
+        super(Localization.lang("Ignore this update"));
+        this.toBeIgnored = toBeIgnored;
+    }
+
+    @Override
+    public void actionPerformed(ActionEvent e) {
+        Globals.prefs.put(JabRefPreferences.VERSION_IGNORED_UPDATE, this.toBeIgnored.getFullVersion());
```
tobiasdiez: Move to Version class so that you can call `Version.setAsLastIgnoredVersion(Globals.prefs)`.

src/main/java/net/sf/jabref/gui/actions/version/RemindMeLaterAction.java
tobiasdiez: Close the dialog?

src/main/java/net/sf/jabref/gui/help/AboutDialog.java
```diff
@@ -78,7 +78,7 @@ public AboutDialog(JabRefFrame bf) {
         headerPanel.add(versionLabel, BorderLayout.CENTER);
         headerPanel.add(copyVersionButton, BorderLayout.EAST);
 
-        String version = String.format("JabRef %s", Globals.BUILD_INFO.getVersion());
+        String version = String.format("JabRef %s", Globals.BUILD_INFO.getVersion().getFullVersion());
```
tobiasdiez: toString as above.

src/main/java/net/sf/jabref/gui/help/NewVersionDialog.java
```diff
@@ -0,0 +1,105 @@
...
+import net.sf.jabref.logic.util.version.Version;
+
+public class NewVersionDialog extends JDialog {
+
+    public NewVersionDialog(JFrame frame, Version currentVersion, Version latestVersion, Version toBeIgnored) {
+        super(frame);
+        setTitle(Localization.lang("New version available"));
+
+        JLabel lblTitle = new JLabel(Localization.lang("A new version of JabRef has been released and can be updated now"));
+        JLabel lblCurrentVersion = new JLabel(Localization.lang("Current version") + ": " + currentVersion.getFullVersion());
```
tobiasdiez: toString

src/main/java/net/sf/jabref/gui/help/NewVersionDialog.java
```diff
@@ -0,0 +1,105 @@
...
+import net.sf.jabref.logic.l10n.Localization;
+import net.sf.jabref.logic.util.version.Version;
+
+public class NewVersionDialog extends JDialog {
+
+    public NewVersionDialog(JFrame frame, Version currentVersion, Version latestVersion, Version toBeIgnored) {
+        super(frame);
+        setTitle(Localization.lang("New version available"));
+
+        JLabel lblTitle = new JLabel(Localization.lang("A new version of JabRef has been released and can be updated now"));
```
tobiasdiez: Remove "and can be updated now..."

src/main/java/net/sf/jabref/gui/help/NewVersionDialog.java
tobiasdiez: Maybe `Installed version` instead of current version (which may be mistaken for "current" = "latest")

src/main/java/net/sf/jabref/gui/desktop/JabRefDesktop.java
koppor: Shouldn't there already exist similar methods and translation strings
somewhere in JabRef? JabRef can open links since ages!

src/main/java/net/sf/jabref/gui/worker/VersionWorker.java
```diff
@@ -0,0 +1,90 @@
...
+        this.toBeIgnored = Objects.requireNonNull(toBeIgnored);
+    }
+
+    @Override
+    protected Void doInBackground() throws Exception {
+        Version latestVersion;
+        try {
+            latestVersion = VersionChecker.grabLatestVersion();
+        } catch (IOException ioException) {
+            String couldntConnect = Localization.lang("Couldn't connect to the update server");
```
tobiasdiez: Add fullstop and the ends.

koppor:
Can't this be done by a) mocking the current version (mocking prefs?!) and
b) mocking http? We already did both somewhere else in the code.

src/main/java/net/sf/jabref/logic/util/version/Version.java
koppor: Yes, covered by our Eclipse formatter settings. Maybe, they can be imported
in IntelliJ. No clue about the current status though...

src/main/java/net/sf/jabref/gui/worker/VersionWorker.java
```diff
@@ -0,0 +1,90 @@
...
+                        couldntConnect,
+                        JOptionPane.ERROR_MESSAGE);
+            }
+            this.mainFrame.output(couldntConnect + " " + tryLater);
+            LOGGER.warn("Couldn't connect to the updateserver", ioException);
+            return null;
+        }
+
+        // only respect the ignored version on automated version checks
+        if (latestVersion.equals(toBeIgnored) && !manualExecution) {
```
tobiasdiez: maybe even latestVersion <= toBeIgnored (which shouldn't be necessary in most of the cases).

src/main/java/net/sf/jabref/logic/util/version/Version.java
```diff
@@ -0,0 +1,129 @@
...
+        this.minor = versionNumbers.length >= 2 ? Integer.parseInt(versionNumbers[1]) : 0;
+        this.patch = versionNumbers.length >= 3 ? Integer.parseInt(versionNumbers[2]) : 0;
+    }
+
+    /**
+     * @return true if this version is newer than the passed one
+     */
+    /* Ignoring the SimplifyBooleanReturns request from PMD, you could write it all in one line but then it's unreadable */
+    @SuppressWarnings("PMD.SimplifyBooleanReturns")
+    public boolean isNewerThan(Version otherVersion) {
```
tobiasdiez: Can one override `<=` for custom classes in java? If yes then I would prefer that.

src/main/java/net/sf/jabref/gui/worker/VersionWorker.java
```diff
@@ -0,0 +1,90 @@
...
+            return null;
+        }
+
+        boolean newer = latestVersion.isNewerThan(currentVersion);
```
tobiasdiez: newer -> updateAvailable.

src/main/java/net/sf/jabref/logic/util/version/Version.java
```diff
@@ -0,0 +1,129 @@
...
+/**
+ * Represents the Application Version with the major and minor number, the full Version String and if it's a developer version
+ */
+public class Version {
+
+    private final String fullVersion;
+    private final int major;
+    private final int minor;
+    private final int patch;
+    private final boolean devVersion;
```
tobiasdiez: isDevelopmentVersion

src/main/java/net/sf/jabref/logic/util/BuildInfo.java
```diff
@@ -19,16 +19,18 @@
 import java.io.InputStream;
 import java.util.Properties;
 
+import net.sf.jabref.logic.util.version.Version;
+
 public class BuildInfo {
 
-    private static final String UNKOWN_VERSION = "*unknown*";
+    public static final String UNKOWN_VERSION = "*unknown*";
```
tobiasdiez: UNK_n_OWN

src/main/java/net/sf/jabref/logic/util/version/Version.java
```diff
@@ -0,0 +1,129 @@
...
+
+    /**
+     * @return true if this version is newer than the passed one
+     */
+    /* Ignoring the SimplifyBooleanReturns request from PMD, you could write it all in one line but then it's unreadable */
+    @SuppressWarnings("PMD.SimplifyBooleanReturns")
+    public boolean isNewerThan(Version otherVersion) {
+        Objects.requireNonNull(otherVersion);
+        if (Objects.equals(this, otherVersion)) {
+            return true;
```
tobiasdiez: false in this case?

src/main/java/net/sf/jabref/logic/util/version/VersionChecker.java
```diff
@@ -0,0 +1,44 @@
...
+
+    private final static String JABREF_GITHUB_URL = "https://api.github.com/repos/JabRef/jabref/releases/latest";
+    public final static String JABREF_DOWNLOAD_URL = "http://www.fosshub.com/JabRef.html";
+
+    /**
+     * Grabs the latest release version from the JabRef GitHub repository
+     * @return
+     * @throws IOException
+     */
+    public static Version grabLatestVersion() throws IOException {
```
tobiasdiez: I think this can be moved to the Version class (and grab -> get)

src/test/java/net/sf/jabref/logic/util/version/VersionTest.java
```diff
@@ -0,0 +1,114 @@
...
+    @Test
+    public void versionNewerThan() {
+        Version olderVersion = new Version("2.4");
+        Version newerVersion = new Version("4.2");
+        assertTrue(newerVersion.isNewerThan(olderVersion));
+    }
+
+    @Test
+    public void versionEqualVersionsWithDev() {
+        Version latestVersion = new Version("4.2");
```
tobiasdiez: Also use `older` and `newer` here and in the other tests.

src/test/java/net/sf/jabref/logic/util/version/VersionTest.java
```diff
@@ -0,0 +1,114 @@
...
+        Version currentVersion = new Version("4.3dev");
+        assertFalse(currentVersion.isNewerThan(latestVersion));
```
tobiasdiez: True here?

I would say (please add tests for this):
3.0 dev < 3.0
3.1 dev < 3.1
3.1.1 dev <= 3.1.1
but
3.9 < 4.0 dev
4.0 < 4.1 dev
4.1 < 4.1.1 dev

Or do I miss something?

src/test/java/net/sf/jabref/logic/util/version/VersionTest.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    @Test
+    public void versionNewerThanDev() {
+        Version latestVersion = new Version("4.2.1");
+        Version currentVersion = new Version("4.3dev");
+        assertTrue(latestVersion.isNewerThan(currentVersion));
```
tobiasdiez: False?
If I'm on the 4.3dev then I don't want an update message about 4.2.1

tobiasdiez:
Looks good to me too. Just some minor remarks and a bigger one for the handling of the dev version in isNewerThan.

For testing: UnitTests are not really possible since it's a GUI thing (and mocking is not possible for UI tests, right?). But you can change the version information in the gradle file, build it and then run it (i.e. pretend that you have an older version). Or for the sake of testing, you can also modify the BuildInfo class to return a hard-coded version.

src/test/java/net/sf/jabref/logic/util/version/VersionTest.java
bartsch-dev: Currently the latest release is 3.4, and the version in the gradle file says im working on 3.5dev.
Does the minor version gets incremented if its a dev version or exactly how do you do it?

src/main/java/net/sf/jabref/logic/util/version/Version.java
bartsch-dev: sadly you cannot override Operators in Java (nor overload).

src/main/java/net/sf/jabref/gui/worker/VersionWorker.java
bartsch-dev: Shouldn't be necessary.
The only way this could happen is if we revoke a Version, which shouldn't be done at all.

src/main/java/net/sf/jabref/gui/desktop/JabRefDesktop.java
bartsch-dev: Every line of code which uses `openBrowser(String url)` just reacts with a status update in the lower update bar which can be easily overseen by the user.
I extracted it all to this new method which now additionally has an error popup window.

bartsch-dev:
I should have considered all mentioned issues.
Some methods can't be extract to another class (e.g.  `JabRefGUI#checkForNewVersion()` to `Version` b/c then they would violate the strict separation of the logic from the gui).

src/main/java/net/sf/jabref/gui/worker/VersionWorker.java
```diff
@@ -0,0 +1,84 @@
...
+    @Override
+    protected Void doInBackground() throws Exception {
+        Version latestVersion;
+        try {
+            latestVersion = Version.getLatestVersion();
+        } catch (IOException ioException) {
+            String couldntConnect = Localization.lang("Couldn't connect to the update server.");
+            String tryLater = Localization.lang("Please try again later and/or check your network connection.");
+            if (manualExecution) {
+                JOptionPane.showMessageDialog(this.mainFrame, couldntConnect + "\n" + tryLater,
```
simonharrer: You are not allowed to access the EDT in the doInBackground method. There is a method which is called after the worker is completed, there, you can access the EDT. Instead of Void, use a class which contains all the info which is required to show the result to the user.

src/main/java/net/sf/jabref/logic/util/Version.java
```diff
@@ -0,0 +1,163 @@
...
+    public static Version getLatestVersion() throws IOException {
+        URLConnection connection = new URL(JABREF_GITHUB_URL).openConnection();
+        connection.setRequestProperty("Accept-Charset", "UTF-8");
+        BufferedReader rd = new BufferedReader(new InputStreamReader(connection.getInputStream()));
+        JSONObject obj = new JSONObject(rd.readLine());
+        return new Version(obj.getString("tag_name").replaceFirst("v", ""));
+    }
+
+    public static Version getIgnoredVersion(JabRefPreferences preferences) {
+        return new Version(preferences.get(JabRefPreferences.VERSION_IGNORED_UPDATE));
```
simonharrer: The logic is not allowed to know the preferences. Please move this to the GUI.

src/main/java/net/sf/jabref/logic/util/Version.java
```diff
@@ -0,0 +1,163 @@
...
+    /**
+     * @return The link to the changelog on github to this specific version
+     * (https://github.com/JabRef/jabref/blob/vX.X/CHANGELOG.md)
+     */
+    public String getChangelogUrl() {
+        return "https://github.com/JabRef/jabref/blob/v" + this.getMajor() + "." + this.getMinor() + "/CHANGELOG.md";
+    }
+
+    public void setAsIgnoredVersion(JabRefPreferences prefs) {
+        prefs.put(JabRefPreferences.VERSION_IGNORED_UPDATE, this.toString());
```
simonharrer: You can move this and another method to a spearate class which handles just versions stored in the preferences. Have a look at the remote preferences.

src/test/java/net/sf/jabref/logic/util/version/VersionTest.java
```diff
@@ -0,0 +1,115 @@
...
+import net.sf.jabref.logic.util.BuildInfo;
+import net.sf.jabref.logic.util.Version;
+
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+public class VersionTest {
```
simonharrer: Cool tests. :)

src/main/java/net/sf/jabref/logic/util/Version.java
```diff
@@ -0,0 +1,163 @@
...
+    public boolean isDevelopmentVersion() {
+        return isDevelopmentVersion;
+    }
+
+    /**
+     * @return The link to the changelog on github to this specific version
+     * (https://github.com/JabRef/jabref/blob/vX.X/CHANGELOG.md)
+     */
+    public String getChangelogUrl() {
+        return "https://github.com/JabRef/jabref/blob/v" + this.getMajor() + "." + this.getMinor() + "/CHANGELOG.md";
```
koppor: What about patch version? It might happen, that we use that.

bartsch-dev: The patch number is in the version number only when it's non-zero, right?
Or do you now implement it all the time? e.g. 4.4.0

koppor: Only when necessary.

src/main/java/net/sf/jabref/logic/util/Version.java
bartsch-dev: moved to its own class: `VersionPreferences`

src/main/java/net/sf/jabref/logic/util/Version.java
bartsch-dev: moved to its own class: `VersionPreferences`

src/main/java/net/sf/jabref/logic/util/Version.java
bartsch-dev: fixed

simonharrer:
Hm, I am missing a CHANGELOG. :(

bartsch-dev:
Hm, I could swear I wrote one.
Should I write it quickly and open another PR, or do you push it directly onto the master?

matthiasgeiger:
I'll add it directly.

PR #1461, opened 2016-05-30
Fix RTF formatter bug

### LINKED ISSUE #1457: Bug in the conversion of \emph{...} in RTF export?
JabRef version 3.3 on Windows 7 64bit

Steps to reproduce:

% BEGIN OF BIBLATEX file:
% Encoding: UTF-8

@Article{article:0000,
  author       = {A. Author},
  title        = {An article title with \emph{a book title} emphasized},
  journaltitle = {Journal},
  volume       = {1},
  date         = {2016},
  pages        = {1-10},
}

@Comment{jabref-meta: databaseType:biblatex;}
% END OF BIBLATEX file

1. Open the file, select the only entry and click on File, Export selected entries
2. Select any of the RTF formats (DIN 1505, Harvard, MIS Quarterly) and save
3. In the resulting RTF file only "a" is in italics, not "book title".
The same problem does not seem to appear in the HTML exports.

Extract from the RTF file for the Harvard RTF conversion:
{\scaps Author, A.}
: An article title with {\i a} book title emphasized}
. In: {\i Journal} {1} (2016), S. 1-10 \par

Attachment: Wordpad screen shot
[harvard-test](https://cloud.githubusercontent.com/assets/19624893/15628199/7f6f207a-24fb-11e6-82ae-a3b98a9bc771.jpg)

stefan-kolb:
Fix #1457 Whole formatter needs a refactoring...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,7 +59,8 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - Fixed [#1412](https://github.com/JabRef/jabref/issues/1412): Save action *protect terms* protects terms within words unecessarily
 - Fixed [#1420](https://github.com/JabRef/jabref/issues/1420): Auto downloader should respect file pattern and propose correct filename
 - Fixed [#651](https://github.com/JabRef/jabref/issues/651): Improve parsing of author names containing braces
-- Fixed [#1421](https://github.com/JabRef/jabref/issues/1421):Auto downloader should try to retrieve DOI if not present and fetch afterwards
+- Fixed [#1421](https://github.com/JabRef/jabref/issues/1421): Auto downloader should try to retrieve DOI if not present and fetch afterwards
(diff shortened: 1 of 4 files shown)


simonharrer:
LGTM

tobiasdiez:
LGTM :+1:

bvdalen:
Thank you very much. I tested the result and it appears to work well.
I know the following may be too much to ask, but I nevertheless wanted to pose one more related question: If the \emph appears within a book title, which will usually be typeset in italics already, \i0 needs to be inserted in the RTF rather than \i in order to obtain the desired result (namely that the emphasized text appears in roman instead of in italics). Would this be possible to implement within the current formatter? Please let me know if you need a detailed example.

stefan-kolb:
Can you please open a new issue for that? Also, if you or your colleague still has interest in contributing, we would love to see a refactoring of the formatter!

PR #1463, opened 2016-05-31
Fix for #1450

### LINKED ISSUE #1450: Entry editor missing after changing autocompletion settings
Using 3.3 release (64 bit) under Windows 10 Pro x64 Java 8 91.
Disabled autocompletion and entry editor disappeared immediately (tried C-e, double click and menu icon to no avail). Resetting settings from dialogue did not solve the problem. Neither uninstalling and reinstalling Jabref.

`java -jar JabRef-3.3.jar -d all` at the command prompt solved the problem.

importing the saved settings from just after the problem started (attached file) reproduces the problem. (My settings have a very long history as I have been using Jabref for ages...).

The bad preferences are in the file below.

[jabref.prefs.txt](https://github.com/JabRef/jabref/files/281949/jabref.prefs.txt)

matthiasgeiger:
This fixes #1450 

After changing the preferences the mainTable is using a new `SplitPane` object to divide table and `EntryEditor` - in the course of creating the actual table existing entryEditors are closed which causes a saving of the current `EntryEditor` height in the prefs. As this is calculated based on the (already new) `SplitPane` a wrong value "1" is stored.

To prevent this the previous state of the divider for the SplitPane is restored based on the prefs before the mainTable is recreated.


stefan-kolb:
LGTM :+1:

Issue #1264, opened 2016-04-20
S with caron does not render correctly


pirlite2:
JabRef version 3.3 on Mint 17.3

The escaped character  `\v{S}' ('S' with caron) does not render properly in the main panel. On the other hand, '\v{Z}' does render correctly. (BTW: Having just upgraded to v3.x, the correct rendering of characters with diacritics as a Unicode character is a massive step forward!)

oscargus:
This is (most likely) a limitation of Java/your system. I've also noted that some characters just do not render properly without knowing why (probably font issue, I cannot really think of anything else because I am quite sure that the actual conversion is correct). In general, the preview panel (which use HTML) will be more successful in rendering non-standard characters, even Unicode ones (meaning that if you for fun convert your field to Unicode by right clicking in the field editor and select LaTeX to Unicode, it will not necessarily display correctly in the field editor, nor in the table, but will show it correctly in the preview).

pirlite2:
In fact, the preview panel also makes a mess of things. S-with-caron 
renders correctly but the following characters in the name are trashed. 
Similarly, Z-with-caron displays properly in the table but renders as 
"Z\caron..." in the preview pane. :-(

As to the experiment you suggest with converting to Unicode in the entry 
editor (nice new feature!), everything renders properly apart from 
S-with-caron (U 0160).

Does the same problem exist on Windows?

Peter

pirlite2:
FWIW: v3.3 displays identical behaviour on Win7! So I would suggest this 
narrows it down to a Java problem at best. Same behaviour on two 
radically different OSs!

Out of interest, I am using openjdk-8. Is this behaviour apparent using 
the Oracle JVM? Does "\v{S}iroky" display correctly as an uppercase 'S' 
with a caron?

P.

mlep:
JabRef version 3.3 on Linux Mint LMDE with Java(TM) SE Runtime Environment (build 1.8.0_66-b17):
- `\v{S}' is displayed as a little rectangle
- '\v{Z}' is ok

pirlite2:
I have now been through my database. Basically, 'S' or 's' (upper or lowercase) with caron does not render properly. It always produces a little rectangle. Lots of other characters (Z,z,n, etc.) with carons do render perectly. It seems to be only the letter 'S' (upper and lowercase) that is wrong.

[...]

PR #1464, opened 2016-05-31
LateXToUnicode Formatter does not handle commands correctly

stefan-kolb:
that are a substring of a special LaTex command like \c \~ \'...

A big problem is to support all kinds of special circumstances in LaTeX code.
The only chance to get the real output is to use the value of a real LaTeX run for stripping LaTeX commands and stuff. Can we somehow write a wrapper for this? Then we could transfer all the hard work to the LaTeX interpreter.

```tex  
\documentclass{article}
\begin{document}
$\chi$ some other italic \textit{text}
\end{document}
```

https://github.com/mcmtroffaes/latexcodec/

### DIFF
--- a/src/test/java/net/sf/jabref/logic/layout/format/LatexToUnicodeFormatterTest.java
+++ b/src/test/java/net/sf/jabref/logic/layout/format/LatexToUnicodeFormatterTest.java
@@ -46,8 +46,15 @@ public class LatexToUnicodeFormatterTest {
     }
 
     @Test
+    public void handleLaTeXCommandsWhichAreSubstrings() {
+        assertEquals("ç", formatter.format("$\\c{c}$"));
+        assertEquals("χ", formatter.format("$\\chi$"));
+    }
+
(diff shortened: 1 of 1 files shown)


simonharrer:
@oscargus how much effort would this be to implement a solution? You are our expert for encodings. :)

oscargus:
Regarding the encodings in general there is a quite substantial list. The weird thing, which I do not get, is that sometimes the Unicode characters ends up wrong. In the particular case `\chi` is not converted correctly (Unicode 967) while both `\phi` (966) and `\psi` (968) works without any problems... There is a related issue #1264 with another character `\v{S}` (352) behaving the same way. No obvious reason why exactly those characters should misbehave... In the latter case `\v{S}` is converted to a square, which can be converted back to `\v{S}`...

oscargus:
Implemented in #1581

matthiasgeiger:
Can delete the branch? @stefan-kolb @oscargus

Issue #421, opened 2015-11-30
Word exporter encoding of umlauts


wielanms:
The word exporter is very nice but following issue remains:

Conversion of Latex umlauts to UTF8 umlauts.

e.g. Oliver Kopp and Tobias Binz and Uwe Breitenb{\"u}cher ....
should get: Kopp, Oliver, Tobias Binz, Uwe Breitenbücher, ....

stefan-kolb:
Confirmed, we need a formatter to remove all LaTex-specific characters from the database.

koppor:
Is it [FormatChars](http://jabref.sourceforge.net/help/CustomExports.php)?

"This formatter converts LaTeX character sequences their equivalent unicode characters and removes other LaTeX commands without handling them."

PR #1465, opened 2016-05-31
Remove umlauts from word export title #421

stefan-kolb:
- From all fields.
- Rough refactoring of MSBib classes as a starting point for further improvements/rewrites.


tobiasdiez:
LGTM :+1:

PR #1466, opened 2016-05-31
Add missing status translations for various specialfields

matthiasgeiger:
The status messages for some specialfields have not been translated correctly (missing localization has not been detected automatically as a variable has been used.)

- [x] Refactor Java code to directly use `Localization.lang` consistently for all SpecialFields
- [x] Lookup missing translations in previous releases 
- [x] fix NPE when clearing multivalue fields (Ranking, Priority, Read Status)

### DIFF
--- a/src/main/java/net/sf/jabref/specialfields/SpecialField.java
+++ b/src/main/java/net/sf/jabref/specialfields/SpecialField.java
@@ -18,12 +18,18 @@ package net.sf.jabref.specialfields;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Objects;
 
 import javax.swing.Icon;
 
 import net.sf.jabref.logic.l10n.Localization;
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
 public abstract class SpecialField {
+
+    private static final Log LOGGER = LogFactory.getLog(SpecialField.class);
 
     // currently, menuString is used for undo string
     // public static String TEXT_UNDO;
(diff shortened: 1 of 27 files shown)


Siedlerchr:
LGTM :+1: 
Just one question:
> (missing localization has not been detected automatically as a variable has been used.

Is this a problem of the python? scripts we use and could it be implemented?

matthiasgeiger:
Not a problem of the Python scripts as they are more or less obsolete by now. 
I think this is not implemented in `LocalizationParser.java` which is used in the tests that check the localizations.

src/main/java/net/sf/jabref/specialfields/SpecialField.java
```diff
@@ -58,11 +57,35 @@ public SpecialFieldValue parse(String s) {
...
+                return Localization.lang("Toggled '%0' for %1 entries", getLocalizedFieldName(), params[0]);
+            }
+        } else {
+            if (params.length == 2 && params[0] != null) {
+                String[] allParams = {getLocalizedFieldName(), params[0], params[1]};
+                return Localization.lang("Set '%0' to '%1' for %2 entries", allParams);
+            } else if (params.length == 1) {
+                return Localization.lang("Cleared '%0' for %1 entries", getLocalizedFieldName(), params[0]);
+            }
+        }
```
simonharrer: Hard to read. Can you make it more ovious which three results are to be returned? Maybe only one level of indent would help here. Also what happens if I pass in null only and it is not a single value field?

stefan-kolb:
LGTM :+1:

PR #1470, opened 2016-06-02
Fixes #1181: Improved "Normalize to BibTeX name format"

### LINKED ISSUE #1181: "Normalize to BibTeX name format" should help when manually importing entries.
JabRef as of 2016-04-12, 39f7ed7

Steps:
1. Use ISBN to BibTeX fetcher with 978-3-642-02373-6
2. Go to required fields
3. Editor is empty
4. Go to http://www.springer.com/us/book/9783642023736
5. Copy the editors string `Ali Babar, M., Dingsøyr, T., Lago, P., van der Vliet, H.`
6. Right click in the editors field and select "Normalize to BibTeX name format"
7. Result: `Ali Babar, M., Dingsøyr T. Lago P. van der Vliet H.`

Expected: `Babar, Ali and M. and Dingsøyr, T. and Lago, P. and van der Vliet, H.`

JabRef 3.2: `Babar, Ali and M. and Dingsøyr and T. and Lago and P. and van der Vliet and H.`

Refs #318 and  #950

bruehldev:
Fixed #1181: Improved "Normalize to BibTeX name format": Separated names with colons only work now


koppor:
It shows `The command "./gradlew check integrationTest" exited with 1`, which should not be caused by this PR. LGTM.

stefan-kolb:
The problem are always the GUI tests. It's annoying...

src/main/java/net/sf/jabref/logic/formatter/bibtexfields/NormalizeNamesFormatter.java
```diff
@@ -36,6 +36,28 @@ public String getKey() {
 
     @Override
     public String format(String value) {
+        // Fixes https://github.com/JabRef/jabref/issues/1181
```
stefan-kolb: Please remove this comment. It's not relevant that this solved some issue.

src/main/java/net/sf/jabref/logic/formatter/bibtexfields/NormalizeNamesFormatter.java
```diff
@@ -36,6 +36,28 @@ public String getKey() {
...
+                    sb.append(',');
+                } else {
+                    // Hit separation between full names (e.g., Ali Babar, M. and Dingsøyr, T.) --> semicolon has to be used
+                    // Will be treated correctly by AuthorList.parse(value);
+                    sb.append(';');
+                }
+            }
+            value = sb.toString();
+        } else {
+            // No special treatment needed, proceed normally
```
stefan-kolb: Why is there a else branch if it does nothing?

src/test/java/net/sf/jabref/logic/formatter/bibtexfields/NormalizeNamesFormatterTest.java
```diff
@@ -92,6 +92,21 @@ public void lastThenJuniorThenFirst() {
         expectCorrect("Name, della, first", "Name, della, first");
     }
 
+    @Test
+    public void testSeperatedCommaNames() {
+        // Testing the issue "https://github.com/JabRef/jabref/issues/1181".
```
stefan-kolb: Remove this comment. The test name should describe the use case.

src/test/java/net/sf/jabref/logic/formatter/bibtexfields/NormalizeNamesFormatterTest.java
```diff
@@ -92,6 +92,21 @@ public void lastThenJuniorThenFirst() {
...
+        // Every second comma should become a semicolon
+        expectCorrect("Ali Babar, M., Dingsøyr, T., Lago, P., van der Vliet, H.",
+                "Ali Babar, M. and Dingsøyr, T. and Lago, P. and van der Vliet, H.");
+        expectCorrect("Ali Babar, M.", "Ali Babar, M.");
+    }
+
+    @Test
+    public void testSeperatedCommaNamesNonTreatedCases() {
+        // Odd count of commas --> Not treated by fix for 1181
```
stefan-kolb: Same here: Remove hint to issue please.

CHANGELOG.md
```diff
@@ -63,6 +63,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - Fixed [#1457](https://github.com/JabRef/jabref/issues/1457): Support multiple words inside LaTeX commands to RTF export
 - Entries retain their groupmembership when undoing their cut/deletion
 - Fixed [#1450](https://github.com/JabRef/jabref/issues/1450): EntryEditor is restored in the correct size after preference changes
+- Fixed [#1181](https://github.com/JabRef/jabref/issues/1181): Improved "Normalize to BibTeX name format": Separated names with colons only work now
```
stefan-kolb: "Support separated names with colons" or something alike?

src/test/java/net/sf/jabref/logic/formatter/bibtexfields/NormalizeNamesFormatterTest.java
```diff
@@ -92,6 +92,21 @@ public void lastThenJuniorThenFirst() {
         expectCorrect("Name, della, first", "Name, della, first");
     }
 
+    @Test
+    public void testSeperatedCommaNames() {
```
stefan-kolb: separated

src/test/java/net/sf/jabref/logic/formatter/bibtexfields/NormalizeNamesFormatterTest.java
```diff
@@ -92,6 +92,21 @@ public void lastThenJuniorThenFirst() {
...
+        // Testing the issue "https://github.com/JabRef/jabref/issues/1181".
+        // Every second comma should become a semicolon
+        expectCorrect("Ali Babar, M., Dingsøyr, T., Lago, P., van der Vliet, H.",
+                "Ali Babar, M. and Dingsøyr, T. and Lago, P. and van der Vliet, H.");
+        expectCorrect("Ali Babar, M.", "Ali Babar, M.");
+    }
+
+    @Test
+    public void testSeperatedCommaNamesNonTreatedCases() {
```
stefan-kolb: separated

src/main/java/net/sf/jabref/logic/formatter/bibtexfields/NormalizeNamesFormatter.java
koppor: To have a special comment, why only one branch is required and no else branch.

stefan-kolb: This makes no sense to me. Then you would have to comment each if without an else branch?!

koppor: OK, I agree, the comment is too general :innocent:

tobiasdiez:
As I said in #1181, there should be some additional handling of author names with Jr part like `von Last, Jr, First`. Moreover, the code probably has to go into the AuthorList parser.

src/test/java/net/sf/jabref/logic/formatter/bibtexfields/NormalizeNamesFormatterTest.java
```diff
@@ -93,17 +93,14 @@ public void lastThenJuniorThenFirst() {
     }
 
     @Test
-    public void testSeperatedCommaNames() {
-        // Testing the issue "https://github.com/JabRef/jabref/issues/1181".
-        // Every second comma should become a semicolon
+    public void testSeperatedCommaNames_EverySecondCommaBecomesAnAnd() {
```
stefan-kolb: why not secribe the use case here? `allowConcatenationOfAuthorsWithCommas`.

koppor:
I think (as outlined in https://github.com/JabRef/jabref/issues/1181#issuecomment-209390160), that `Jr` are seldom and that I need a thing working for 80% of the cases and not covering 100%. We just looked at `AuthorListParse.getAuthor()` and it seems to be *very* difficult to include the special treatment there. We would have to include hard coded strings for `jrPart` to be able to distinguish them from other strings.

The fix works for me and really improves my workflow.

stefan-kolb:
Do you often have authors separated by colons? I would have guessed this is also a very seldom case.

tobiasdiez:
In my opinion, the cleanup operations / save actions should respect the bib(la)tex standard before anything else. Otherwise they keep reformatting valid bibtex and thus become useless. 

Sorry, but I just can't see an automatic way to correctly determine the authors if they are only separated by comma. Consider strings like
`Surname, Jr, First, Surname2, First2`
or `Surname, First, First2 Surname2`.

stefan-kolb:
Is a comma even allowed for separating authors in Bib(la)tex?

koppor:
IMHO we discussed the separator at some other place that biblatex can be configured to use something else than `and`, but I do not find the source.

In bibtex, `and` really has to be used.

This tweak happens only if **only commans** and **2 or more commas** and **even number of commas**

Thus, it will currently. destroy both of your examples. - We can include a checker if your special cases ( `Jr`) and parts with one word only (`Surname`, `First` at the second example) and not handle these.

I still think that you have these special cases, but that these are not the majority of the cases.

tobiasdiez:
I'm fine with every change as long as standard bibtex names don't get destroyed. So in particular:
- Last, First and Last2, First2 and Last3, First3
- Last, First and Last2, First2 and Last3, First3 and First4 Last4 (this should be converted to Last4, First4)
- Last, Jr, First and Last2, First2

Moreover, semi-correct names using semicolon should still be converted to the correct format (i.e. replace every `and` in the above examples with a semicolon in the input). Please add corresponding tests.

src/main/java/net/sf/jabref/logic/formatter/bibtexfields/NormalizeNamesFormatter.java
```diff
@@ -36,6 +36,25 @@ public String getKey() {
 
     @Override
     public String format(String value) {
+        // Handle case names in order lastname, firstname and separated by ","
```
koppor: Sourround the new block with

    if (!value.containts(" and ")) {

koppor:
:open_mouth: - completely overseen that case.

@bruehldev Please include that test and fix the issue. I indicated the necessary change (check for `value.containts(" and ")`) at the appropriate place.

src/main/java/net/sf/jabref/logic/formatter/bibtexfields/NormalizeNamesFormatter.java
tobiasdiez: test also `{` as well as `;`

Siedlerchr: For checking for  Curly Braces there is a String Util Method which also has
a Unit Test. Maybe you can reuse that.

2016-06-03 14:13 GMT+02:00 Tobias Diez <notifications@github.com>:

src/main/java/net/sf/jabref/logic/formatter/bibtexfields/NormalizeNamesFormatter.java
```diff
@@ -69,4 +105,15 @@ public String getExampleInput() {
         return "Albert Einstein and Alan Turing";
     }
 
+    private static boolean contains(final String[] array, final String[] searchTerms) {
```
koppor: Do not use arrays. Use Collections. [Collections.contains](https://docs.oracle.com/javase/7/docs/api/java/util/Collection.html#contains(java.lang.Object)) is your friend.

matthiasgeiger:
Please check whether your improvement also solves #1504 by adding the example as an additional test case.

Thanks!

Braunch:
What is the status of this?

bruehldev:
I added the test with "von", "zu", "van".  I also covered the related issue (https://github.com/JabRef/jabref/issues/1504). 
What can i do additionally to close this pull request?

koppor:
At least: Fix [codacy issues](https://www.codacy.com/app/simonharrer/jabref/pullRequest?prid=249695) and resolve conflicts.

src/main/java/net/sf/jabref/logic/formatter/bibtexfields/NormalizeNamesFormatter.java
koppor: Remove this unused method.

src/main/java/net/sf/jabref/logic/formatter/bibtexfields/NormalizeNamesFormatter.java
```diff
@@ -36,6 +45,73 @@ public String getKey() {
...
+                    if (avoidTermsInLowerCase.contains(valueParts[i].toLowerCase())) {
+                        avoidIndex.add(i);
+                        valuePartsCount--;
+                    }
+                }
+
+                if ((valuePartsCount % 2) == 0) {
+                    // We hit the described special case with name affix like Jr
+                    StringBuilder stringBuilder = new StringBuilder();
+                    // avoidedTimes need to increase the count of avoided terms for correct module calculation
```
koppor: replace by "avoidedTimes needs to be increased by the count of avoided terms for correct odd/even calculation"

src/main/java/net/sf/jabref/logic/formatter/bibtexfields/NormalizeNamesFormatter.java
```diff
@@ -36,6 +45,73 @@ public String getKey() {
...
+            // We hit the comma name separator case
+            // Usually the getAsLastFirstNamesWithAnd method would separate them if pre- and lastname are separated with "and"
+            // If not, we check if spaces separate pre- and lastname
+            if (spaceInAllParts) {
+                value = value.replaceAll(",", " and");
+            } else {
+                // Looking for name affixes to avoid
+                // partCount need to reduce by the count off avoiding terms
+                int valuePartsCount = valueParts.length;
+                // Holding the index of every term, which need to avoid
```
koppor: "Holds the index of each term which needs to be avoided"

src/main/java/net/sf/jabref/logic/formatter/bibtexfields/NormalizeNamesFormatter.java
```diff
@@ -36,6 +45,73 @@ public String getKey() {
...
+            }
+
+            // We hit the comma name separator case
+            // Usually the getAsLastFirstNamesWithAnd method would separate them if pre- and lastname are separated with "and"
+            // If not, we check if spaces separate pre- and lastname
+            if (spaceInAllParts) {
+                value = value.replaceAll(",", " and");
+            } else {
+                // Looking for name affixes to avoid
+                // partCount need to reduce by the count off avoiding terms
```
koppor: The variable is now called "valuePartsCount".
"valuePartsCount holds the count of name parts without the avoided terms"

src/main/java/net/sf/jabref/logic/formatter/bibtexfields/NormalizeNamesFormatter.java
```diff
@@ -36,6 +45,73 @@ public String getKey() {
 
     @Override
     public String format(String value) {
+        Objects.requireNonNull(value);
+        // Handle case names in order lastname, firstname and separated by ","
+        // E.g., Ali Babar, M., Dingsøyr, T., Lago, P., van der Vliet, H.
+        if (!value.contains(" and ") && !value.contains("{") && !value.contains(";")) {
+            String[] valueParts = value.split(",");
+            // Delete spaces for correct case identification
+            for(int i=0; i < valueParts.length; i++) {
```
koppor: Space after `for` is missing

src/main/java/net/sf/jabref/logic/formatter/bibtexfields/NormalizeNamesFormatter.java
```diff
@@ -24,6 +30,9 @@
  */
 public class NormalizeNamesFormatter implements Formatter {
 
+    // Avoid partition where these values are contained
+    private final Collection<String> avoidTermsInLowerCase = Arrays.asList("jr", "sr", "jnr", "snr", "von", "zu", "van", "der");
```
koppor: Why isn't this called "affixesInLowerCase"? - Why is "von", "zu", "van" and "der" an affix? I think, only jr, sr, jnr, and snr are affixes. - As far as I can guess, "van" are not affixes

When reading https://en.wikipedia.org/wiki/Suffix_(name), it seems that jr is a name suffix, isn't it?

src/test/java/net/sf/jabref/logic/formatter/bibtexfields/NormalizeNamesFormatterTest.java
```diff
@@ -159,4 +155,8 @@ public void testPreposition() {
     public void testAvoidNameAffixes() {
         expectCorrect("Canon der Barbar, Alexander der Große","der Barbar, Canon and der Große, Alexander");
     }
+
+    private void expectCorrect(String input, String expected) {
```
koppor: Could you remove this method to make the tests being consistent with other tests?

Please be aware that the order of parameters has to be swapped!

src/main/java/net/sf/jabref/logic/formatter/bibtexfields/NormalizeNamesFormatter.java
```diff
@@ -36,6 +44,75 @@ public String getKey() {
 
     @Override
     public String format(String value) {
+        Objects.requireNonNull(value);
```
tobiasdiez: Can you please move all this code to `AuthorList.parse`. Thanks. Also split it into smaller methods, which are easier to understand and maintain.

CHANGELOG.md
```diff
@@ -63,6 +63,14 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - Fixed NullPointerException when trying to set a special field or mark an entry through the menu without having an open database
 - Fixed [#1257](https://github.com/JabRef/jabref/issues/1324): Preferences for the BibTeX key generator set in a version prior to 3.2 are now migrated automatically to the new version
 - Fixed [#1716](https://github.com/JabRef/jabref/issues/1716): `@`-Symbols stored in BibTeX fields no longer break the database
+- Fixed [#405](https://github.com/JabRef/jabref/issues/405): Added more {} around capital letters in Unicode/HTML to LaTeX conversion to preserve them
```
tobiasdiez: Probably some merge problem.

oscargus: Union merge problems...

bruehldev:
@tobiasdiez 
i moved the entire code to AuthorList.parse and solved the merge conflict with a hard reset to upstream/master.

src/main/java/net/sf/jabref/logic/formatter/bibtexfields/NormalizeNamesFormatter.java
oscargus: Does it really make sense to use this in `AuthorList`? Isn't `NormalizeNameFormatter` quite a bit about guessing? I'd prefer if `AuthorList` didn't guess.

bruehldev: @oscargus It doesen't matter in which class we put the new code. It will first be executed anyway. I think it belongs to AuthorList parse, because the other methods in authorlist will always be executed afterwards. Like this we have a better reading flow.

src/main/java/net/sf/jabref/logic/formatter/bibtexfields/NormalizeNamesFormatter.java
```diff
@@ -24,6 +29,9 @@
  */
 public class NormalizeNamesFormatter implements Formatter {
```
boceckts: Please add a java doc comment and describe what this formatter does.

src/main/java/net/sf/jabref/logic/formatter/bibtexfields/NormalizeNamesFormatter.java
boceckts: Remove the comment about `partCount`

src/main/java/net/sf/jabref/logic/formatter/bibtexfields/NormalizeNamesFormatter.java
```diff
@@ -36,6 +44,75 @@ public String getKey() {
 
     @Override
     public String format(String value) {
+        Objects.requireNonNull(value);
+        // Handle case names in order lastname, firstname and separated by ","
+        // E.g., Ali Babar, M., Dingsøyr, T., Lago, P., van der Vliet, H.
+        if (!value.contains(" and ") && !value.contains("{") && !value.contains(";")) {
+            String[] valueParts = value.split(",");
```
boceckts: Please use more appropriate variable names than `value` or `valueParts` since it makes the code really hard to read.

src/main/java/net/sf/jabref/logic/formatter/bibtexfields/NormalizeNamesFormatter.java
```diff
@@ -36,6 +44,75 @@ public String getKey() {
...
+            // Delete spaces for correct case identification
+            for (int i=0; i < valueParts.length; i++) {
+                valueParts[i] = valueParts[i].trim();
+            }
```
boceckts: If you only need each value in a collection, please use a foreach loop as this is more readable in my opinion.
e.g.:
```
for (String name : allNames) {
    name = name.trim();
}
```

CHANGELOG.md
```diff
@@ -75,6 +75,14 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
...
  
+- Fixed [#405](https://github.com/JabRef/jabref/issues/405): Added more {} around capital letters in Unicode/HTML to LaTeX conversion to preserve them
+- Alleviate multiuser concurrency issue when near simultaneous saves occur to a shared database file
+- Fixed [#1476](https://github.com/JabRef/jabref/issues/1476): NPE when importing from SQL DB because of missing DatabaseMode
+- Fixed [#1481](https://github.com/JabRef/jabref/issues/1481): Mac OS X binary seems broken for JabRef 3.4 release
+- Fixed [#1430](https://github.com/JabRef/jabref/issues/1430): "review changes" did misinterpret changes
+- Fixed [#1434](https://github.com/JabRef/jabref/issues/1434): Static groups are now longer displayed as dynamic ones
+- Fixed [#1482](https://github.com/JabRef/jabref/issues/1482): Correct number of matched entries is displayed for refining subgroups
+- Fixed [#1181](https://github.com/JabRef/jabref/issues/1181) and [#1504](https://github.com/JabRef/jabref/issues/1504): Improved "Normalize to BibTeX name format": Support separated names with commas and colons. Considered name affixes such as "Jr".
```
boceckts: Seems like a merge conflict to me.

bruehldev:
Did someone deleted commit where i moved the code to authorlist.parse? Right now i'm not even quite sure if i even pushed it :smile:. Should i push it again or should we leave the code in normalizenameformatter?

oscargus:
I think I saw the code there yesterday anyway. And I didn't delete it... ;-)

src/main/java/net/sf/jabref/logic/formatter/bibtexfields/NormalizeNamesFormatter.java
```diff
@@ -43,20 +43,23 @@ public String getKey() {
...
     @Override
-    public String format(String value) {
-        Objects.requireNonNull(value);
+    public String format(String nameList) {
+        Objects.requireNonNull(nameList);
         // Handle case names in order lastname, firstname and separated by ","
         // E.g., Ali Babar, M., Dingsøyr, T., Lago, P., van der Vliet, H.
-        if (!value.contains(" and ") && !value.contains("{") && !value.contains(";")) {
-            String[] valueParts = value.split(",");
+        if (!nameList.contains(" and ") && !nameList.contains("{") && !nameList.contains(";")) {
```
Siedlerchr: I recently ran into the same issue (after a user noted it) in the office xml export, make sure to check for and case independent : `nameList.toUpperCase(Locale.ENGLISH).contains(" AND "))`

And I would suggest adding a test for it, to make sure the behaviour stays the same in future ;)

src/main/java/net/sf/jabref/logic/formatter/bibtexfields/NormalizeNamesFormatter.java
```diff
@@ -43,20 +43,23 @@ public String getKey() {
...
+            String[] arrayNameList = nameList.split(",");
```
Siedlerchr: Make it a list with Arrays.asList(..split() )

src/main/java/net/sf/jabref/logic/formatter/bibtexfields/NormalizeNamesFormatter.java
```diff
@@ -43,20 +43,23 @@ public String getKey() {
...
             // Delete spaces for correct case identification
-            for (int i=0; i < valueParts.length; i++) {
-                valueParts[i] = valueParts[i].trim();
+            for (int i=0; i < arrayNameList.length; i++) {
```
Siedlerchr: When you have list you can do a fancy one liner `originalStrings.replaceAll(String::trim);`
https://docs.oracle.com/javase/8/docs/api/java/util/List.html#replaceAll-java.util.function.UnaryOperator-

bruehldev:
I added the new test and changed your remarks @Siedlerchr .
Thank you very much.
Theres still the question left if we move the code to authorlist.parse. 
I think it doesn't really matter so i left it in NormalizeNameFormatter.

Siedlerchr:
@JabRef/developers  Any strong opinions on having the code In NormalizeNameFormatter or Authorlist.parse?

tobiasdiez:
Yes, `Authorlist.parse` mainly because the test coverage for this method is way higher as for the name formatter (and for consistency reasons).

bruehldev:
Okay, i will move the code to `Authorlist.parse`.  Since i already did that on my computer at home i'll do that this evening and use the existing local commits.

bruehldev:
Rebased to upstream master. Should be ready to merge now

tobiasdiez:
As far as I can see the new code is still in the NormalizeNamesFormatter. Please also add the cases mentioned in https://github.com/JabRef/jabref/pull/1470#issuecomment-223537954 as tests for the AuthorList.parse method, i.e. as `assertEqual(expectedAuthor, Authorlist.parse(...))`. Thanks. 
Ready-for-review removed until this is done.

bruehldev:
@tobiasdiez sorry for the wrong statement. I thought it was pushed, From now on i will double check my uploaded code. 

I added the test cases for semicolons. The standart format is already tested by threeAuthorsSeperatedByAnd, testMultipleNameAffixes, testNormalizeAuthorList. I hope it's okay that i'm using last2,3,4 and first2,3,4 as names in the semicolon test.

Issue #1026, opened 2016-03-23
Jabref removes comments inside the Bibtex code


stefan-kolb:
@lenhard  Do we intend to do this? i thought we keep the source as is if possible?

``` tex
%% author = {Ahronovitz, Miha and Amrhein, Dustin and Anderson, Patrick and De, Andrew}
 @misc{Ahronovitz2010a,
 author = {{Cloud Computing Use Case Discussion Group}},
 title = {Cloud {Computing} {Use} {Cases} {White} {Paper} - {Version} 4.0},
 date = {2010-07}
 }
``` 

Will get

``` tex
 @misc{Ahronovitz2010a,
 author = {{Cloud Computing Use Case Discussion Group}},
 title = {Cloud {Computing} {Use} {Cases} {White} {Paper} - {Version} 4.0},
 date = {2010-07}
 }
``` 

This will only be done when changing the entry right?

oscargus:
Everything outside of an entry is a comment and should, if possible, be kept. This is not an answer to your question though, just a statement.

stefan-kolb:
Actually it is not at the moment at least for me.

lenhard:
I assume that the comment will be overwritten if the entry is reformatted. Otherwise, it is kept as part of the parsed serialization of the entry. 

As far as I know, JabRef has always overwritten such comments, but we might consider to change this functionalty (this has strong implications on parsing/writing)

tobiasdiez:
Right now everything not contained in an entry (or @comment tag) is completely ignored by the parser. In particular, it is not saved anywhere and gets lost on rewrite. The problem with this floating-around text is that is not connected to any entry and thus the new position of it can not be determined on save. By the way, even user-defined @comment tags get lost.

oscargus:
I can understand that it is hard to keep everything for sure and I think that it is OK to some extent. If you use JabRef you do not need to do manual edits...

koppor:
The fact that JabRef destroys content of the the `bib` file often prevents users to switch from plain text editing to JabRef. Therefore, I would propose to add a data structure to `BibEntry` which keeps the text lines above (somehow). Maybe, we offer a separate tab offering "Comment" showing the lines above the bib entry.

lenhard:
Adding a datastructure to `BibEntry` that keeps the lines above wouldn't actually be too difficult. The parser reads it like that already. But then, what happens to 
 - all the comments below the last entry
 - comments that are mixed between meta data entries
[...]

PR #1471, opened 2016-06-03
Preserve user comments in bib file

lenhard:
So I am finally having my take at #1026. The current solution was surprisingly easy so far (despite what I had written before). It looks explicitly for our encoding prefix in the file ("Encoding: ") and kills lines that contain it (but only lines preceding an entry, it will not delete something that is inside an entry). Other user comments are left untouched and are *always* written out again, even if the entry is reformatted.

User comments that are above meta data, bibtex strings, or the preamble will still be removed, though. Changing that would require large scale changes to these items in our model, since we would also need to store the parsed serialization for them, which we do not do so far.

This should receive significant automated, but also manual, testing, since it modifies a quite critical part and has the potential to destroy bib files. I will add a few more tests for this PR.


lenhard:
Ok, seems like rebasing my stuff here on master was not the smartest thing to do...

lenhard:
Ok, I'd like to put this up for discussion.

User comments are now kept under the following circumstances:
 - Above any BibEntry
 - Above any BibtexString
 - Comments at the end of the file

This is independent of whether an entry is changed/reformatted, but not if it is deleted. If it is reformated, user comment text is located exactly one blank line above the reformated entry.

User comments are still not kept when:
 - Above meta data comments
 - Above the Preamble
 - Contains JabRef's ENCODING_PREFIX
 - Above an entry that has been deleted

Is this good enough? It should be sufficient in most scenarios and work if someone opens his non-JabRef bib file with JabRef. If someone starts adding arbitrary text between JabRef's Metadata, though... The latter is hard to achieve since we do not store the serialization of MetaData. If we want to, I'd have to make some additions to our MetaData objects. I could also store the comments between meta data and add them as a bunch at the end of the file, but then the relative position gets lost. Before I do more work, I'd like to clarify the following points:
 - Does the storage of arbitrary text above MetaData really matter?
 - Is it ok to delete comments if the entry below them is deleted?

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 ## [Unreleased]
 
 ### Changed
+- JabRef does no longer delete user comments outside of BibTeX entries [#1026]
```
tobiasdiez: Change format of changelog entry to match the other ones.

src/main/java/net/sf/jabref/exporter/BibDatabaseWriter.java
```diff
@@ -354,6 +356,27 @@ private void writeString(Writer fw, BibtexString bs, Map<String, BibtexString> r
...
     }
 
+    private void writeUserCommentsForString(BibtexString string, Writer out) throws IOException {
+        String parsedSerialization = string.getParsedSerialization();
+
+        if(parsedSerialization != null) {
+
+            try {
+                // get the text before the string
+                String prolog = string.getParsedSerialization().substring(0, string.getParsedSerialization().indexOf('@'));
```
tobiasdiez: Idea: Move most of the code which determines the comment to `BibEntry.getUserComments` since in my opinion it is the responsibility of the BibEntry to store all the data in an accessible format.

src/main/java/net/sf/jabref/logic/bibtex/BibEntryWriter.java
```diff
@@ -39,22 +39,46 @@ public void write(BibEntry entry, Writer out, BibDatabaseMode bibDatabaseMode) t
...
     }
 
+    private void writeUserComments(BibEntry entry, Writer out) throws IOException {
+        String parsedSerialization = entry.getParsedSerialization();
+
+        if(parsedSerialization != null) {
+
+            try {
+                // get the text before the entry
+                String prolog = entry.getParsedSerialization().substring(0, entry.getParsedSerialization().indexOf('@'));
```
tobiasdiez: Same as above: move to `BibEntry`

src/test/java/net/sf/jabref/importer/fileformat/BibtexParserTest.java
```diff
@@ -1253,10 +1254,10 @@ public void parseSavesNewlinesBeforeEntryInParsedSerialization() throws IOExcept
     }
 
     @Test
-    public void parseSavesOnlyRealNewlinesBeforeEntryInParsedSerialization() throws IOException {
+    public void parseRemovesEncodingLineInParsedSerialization() throws IOException {
         String testEntry = "@article{test,author={Ed von Test}}";
         ParserResult result = BibtexParser.parse(
-                new StringReader("%Encoding: no" + Globals.NEWLINE + Globals.NEWLINE + Globals.NEWLINE + testEntry));
+                new StringReader(Globals.ENCODING_PREFIX + Globals.NEWLINE + Globals.NEWLINE + Globals.NEWLINE + testEntry));
```
tobiasdiez: the encoding "no" is missing in the new version

src/test/java/net/sf/jabref/importer/fileformat/BibtexParserTest.java
```diff
@@ -1486,4 +1487,43 @@ public void parseReturnsEntriesInSameOrder() throws IOException {
...
+                "  Note                     = {some note}," + Globals.NEWLINE +
+                "  Number                   = {1}" + Globals.NEWLINE +
+                "}";
+        // @formatter:on
+
+        // read in bibtex string
+        ParserResult result = BibtexParser.parse(new StringReader(bibtexEntry));
+
+        Collection<BibEntry> entries = result.getDatabase().getEntries();
+        assertEquals(1, entries.size());
```
tobiasdiez: Create expected bibentry and compare against it.

src/test/java/net/sf/jabref/logic/bibtex/BibEntryWriterTest.java
```diff
@@ -403,4 +403,83 @@ public void trimFieldContents() throws IOException {
...
+        // @formatter:on
+
+        // read in bibtex string
+        ParserResult result = BibtexParser.parse(new StringReader(bibtexEntry));
+
+        Collection<BibEntry> entries = result.getDatabase().getEntries();
+        assertEquals(1, entries.size());
+
+        BibEntry entry = entries.iterator().next();
+        assertEquals("test", entry.getCiteKey());
```
tobiasdiez: Remove these asserts

src/test/java/net/sf/jabref/importer/fileformat/BibtexParserTest.java
```diff
@@ -1486,4 +1487,43 @@ public void parseReturnsEntriesInSameOrder() throws IOException {
 
         assertEquals(expected, result.getDatabase().getEntries());
     }
+
+    @Test
+    public void parsePrecedingComment() throws IOException {
+        // @formatter:off
+        String bibtexEntry = "% Some random comment that should stay here" + Globals.NEWLINE +
```
tobiasdiez: What happens to some random test in front of the entry? Bib(la)tex does not know the percentage symbol as a comment symbol...

src/test/java/net/sf/jabref/importer/fileformat/BibtexParserTest.java
```diff
@@ -1486,4 +1487,43 @@ public void parseReturnsEntriesInSameOrder() throws IOException {
...
+                "@Article{test," + Globals.NEWLINE +
```
tobiasdiez: Add test for entries starting on the same line as the comment, i.e. `some text containing % and now the @article {...}`

src/test/java/net/sf/jabref/logic/bibtex/BibEntryWriterTest.java
```diff
@@ -403,4 +403,83 @@ public void trimFieldContents() throws IOException {
...
+                "}";
+        // @formatter:on
+
+        // read in bibtex string
+        ParserResult result = BibtexParser.parse(new StringReader(bibtexEntry));
+
+        Collection<BibEntry> entries = result.getDatabase().getEntries();
+        assertEquals(1, entries.size());
+
+        BibEntry entry = entries.iterator().next();
```
tobiasdiez: Also remove asserts here

src/test/resources/testbib/bibWithEpilog.bib
```diff
@@ -0,0 +1,43 @@
...
+  address = {New York, NY, USA},
+  publisher = {ACM},
+  bdsk-url-1 = {http://doi.acm.org/10.1145/1137627.1137631},
+  doi = {http://doi.acm.org/10.1145/1137627.1137631},
+  file = {:/Volumes/iDisk/Freie Universität Berlin/Semester 9/Softwareprozesse/p11-bostrom.pdf:PDF},
+  isbn = {1-59593-411-1},
+  location = {Shanghai, China}
+}
+
+@Comment{jabref-meta: databaseType:bibtex;}
```
tobiasdiez: I don't like these files for round-trips. In my opinion it would suffice to add comments to `complex.bib` which already has round-trip tests (as integration test) and concentrate on unit test for the rest.

tobiasdiez:
Looks good to me. Just had some more or less minor remarks (mostly about the testing code). 

I think we don't need support for comments between JabRef metadata.

src/main/java/net/sf/jabref/exporter/BibDatabaseWriter.java
lenhard: Good idea, will do.

However, I am a little bit opposed to simplifying the roundtrip test. This is the discussion on whether we prefer unit over integration tests. We have been there before :)

src/test/java/net/sf/jabref/importer/fileformat/BibtexParserTest.java
lenhard: That does not matter for the test (so in a sense, I minimized the test ;-) ) Only the `ENCODING_PREFIX` is relevant. A line with it will be purged regardless of what encoding is set.

src/test/java/net/sf/jabref/importer/fileformat/BibtexParserTest.java
lenhard: I do not really understand your comment. The last line in the test method compares the expected serialization with the actual one: `        assertEquals(bibtexEntry, entry.getParsedSerialization());`

src/test/java/net/sf/jabref/logic/bibtex/BibEntryWriterTest.java
lenhard: Why?

src/test/java/net/sf/jabref/importer/fileformat/BibtexParserTest.java
lenhard: This does not make a difference. Any non-whitespace text (except for the `ENCODING_PREFIX`) is treated as a comment. In a sense, the percentage sign here is a complete coincidence.

src/test/java/net/sf/jabref/logic/bibtex/BibEntryWriterTest.java
lenhard: As above: why?

src/test/resources/testbib/bibWithEpilog.bib
lenhard: I tend to disagree. If I point all kinds of test to `complex.bib`, these tests are indirectly coupled to each other. I would rather like to keep them as separate as possible and that means separate files.

lenhard:
@tobiasdiez If not mentioned otherwise, I addressed your comments above.

We tend to differ when it comes to testing in general, though. I really prefer integration tests when doing anything with reading/writing. In JabRef, these things are inherently coupled. To implement this PR (write out user comments correctly), I also had to do significant changes to reading. So, as long as no overwhelming majority of @JabRef/developers pressures me to write unit tests, I will rather write integration tests.

src/test/java/net/sf/jabref/importer/fileformat/BibtexParserTest.java
tobiasdiez: I meant:
````
BibEntry expected = new BibEntry().withField("author", "Foo Bar).withField(.....)
assertEquals(Collections.singletonList(expected), entries)
````

File: src/test/java/net/sf/jabref/logic/bibtex/BibEntryWriterTest.java (hunk 6 above)
Comment (tobiasdiez) [inline src/test/java/net/sf/jabref/logic/bibtex/BibEntryWriterTest.java]: Since a test should only fail because of one reason (i.e. in the best case only one assert). This test should check if parsing -> writing yields the same string which was originally passed to it. Thus only the last assert is what you want. https://github.com/JabRef/jabref/pull/1471/files#diff-b4b7ca399d396eb3e62fc3416156c92aR437

File: src/test/resources/testbib/bibWithEpilog.bib (hunk 10 above)
Comment (tobiasdiez) [inline src/test/resources/testbib/bibWithEpilog.bib]: I think you missunderstood me. For complex.bib there are already round-trip tests in place. Thus you just need to modify `complex.bib` instead of writing a new round-trip test with a new bib file.

File: src/test/java/net/sf/jabref/importer/fileformat/BibtexParserTest.java (hunk 5 above)
Comment (lenhard) [inline src/test/java/net/sf/jabref/importer/fileformat/BibtexParserTest.java]: The `withField` syntax is something new for me.

Anyway, this will not work here, unfortunately. The crucial part is the comment before the entry and this cannot be set as a field.

File: src/test/resources/testbib/bibWithEpilog.bib (hunk 10 above)
Comment (lenhard) [inline src/test/resources/testbib/bibWithEpilog.bib]: I see. And `complex.bib` already has an epilog. That means the above test can simply be removed.

File: src/test/java/net/sf/jabref/logic/bibtex/BibEntryWriterTest.java (hunk 6 above)
Comment (lenhard) [inline src/test/java/net/sf/jabref/logic/bibtex/BibEntryWriterTest.java]: Ok, I went through the class and removed all asserts I found redundant.

**Comment by simonharrer** [comment]
LGTM. I think integration tests are more valuable. So this would suffice for me.

**Comment by lenhard** [comment]
Great! Let us talk briefly about this during the next devcall and decide if we take it into 3.5 or 3.6.

**Comment by tobiasdiez** [comment]
For 3.6

File: jabref.install4j [hunk 11]
```diff
@@ -369,7 +369,7 @@ return true;</string>
                 </serializedBean>
                 <condition>context.getBooleanVariable("addToDockAction")</condition>
               </action>
-              <group name="Registry" id="239" customizedId="" beanClass="com.install4j.runtime.beans.groups.ActionGroup" enabled="true" commentSet="false" comment="" actionElevationType="inherit">
+              <group name="Registry" id="239" customizedId="" beanClass="com.install4j.runtime.beans.groups.ActionGroup" enabled="true" commentSet="false" comment="" actionElevationType="elevated">
```
koppor: Why is an install4j change part of this PR?

stefan-kolb: I revert it later, sorry...

src/main/java/net/sf/jabref/model/entry/BibtexString.java
```diff
@@ -152,4 +152,28 @@ public String getParsedSerialization() {
     public boolean hasChanged(){
         return hasChanged;
     }
+
+    /*
+    * Returns user comments (arbitrary text before the string) if there are any. If not returns the empty string
+     */
+    public String getUserComments() {
```
tobiasdiez: This is now an exact copy of BibEntry.getUserComments right? Maybe add a superclass BibItem containing this method and then BibEntry and BibString derive from this superclass (maybe there is even more common code which could be extracted to the super class)

CHANGELOG.md
```diff
@@ -13,12 +13,6 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 
 ### Changed
 - [#1026](https://github.com/JabRef/jabref/issues/1026) JabRef does no longer delete user comments outside of BibTeX entries and strings
-- Implemented [#661](https://github.com/JabRef/jabref/issues/661): Introducing a "check for updates" mechnism (manually/automatic at startup)
```
koppor: Why are these CHANGELOG.md entries deleted?

src/test/java/net/sf/jabref/exporter/BibDatabaseWriterTest.java
```diff
@@ -292,6 +292,82 @@ public void roundtrip() throws IOException {
...
+
+        databaseWriter.writePartOfDatabase(stringWriter, context, result.getDatabase().getEntries(), preferences);
+
+        try (Scanner scanner = new Scanner(Paths.get("src/test/resources/testbib/bibWithUserCommentAndEntryChange.bib"),encoding.name())) {
+            assertEquals(scanner.useDelimiter("\\A").next(), stringWriter.toString());
+        }
+    }
+
+    @Test
+    public void roundtripWithUserCommentBeforeString() throws IOException {
```
tobiasdiez: This looks like an exact copy of https://github.com/JabRef/jabref/blob/304d2802162a7e75edda5fbc893666cc73efe669/src/test/java/net/sf/jabref/exporter/BibDatabaseWriterTest.java#L279

stefan-kolb:
@lenhard Please squash the commits via Github when merging this :smile:

src/test/java/net/sf/jabref/importer/fileformat/BibtexParserTest.java
tobiasdiez: So what you actually want to test is
assertEquals("% Some random comment that should stay here", entry.getUserComment());
right?

src/main/java/net/sf/jabref/model/entry/BibtexString.java
simonharrer: Sorry to always slip into the role of your antagonist, but I am against creating a class hierarchy. In most cases, it makes the code harder to understand, and the code duplication savings are just not worth the effort. In addition, one has to adhere to the liskov substitution principle to create a good class hierarchy, but this is hard to do right.

tobiasdiez: No problem, having an experienced antagonist is probably the best way to learn something :smile: 

So now ignoring everything about the actual implementation and only speaking about "business objects": There are different items which can be contained in a bib file. For example `@Comments`, `@Strings` and normal bib entries. They all have a similar structure (e.g begin with `@`, followed by some identifier and then braces which surround the actual content) and all of them could have some text comments in front of them. Thus there is also a common behavior when it comes to parsing and writing.  
So how would my deer antagonist reflect this common structure and behavior in java code?

My idea would be: BibString, BibEntry, BibComment all derive from BibItem. BibItem has methods to parse and write at least the form `@something { abstract method to parse / write value }`. Then the writer just gets a list of BibItems and invokes the write method on them. Similarly the parser returns a list of BibItems while the parsing of every item was done in the subclass. But again...this idea somehow corresponds from a `service` implementation to a full-blown domain model....so yeah, maybe we should discuss this in the next dev call ;-)

src/test/java/net/sf/jabref/logic/bibtex/BibEntryWriterTest.java
```diff
@@ -117,14 +117,8 @@ public void roundTripWithPrependingNewlines() throws IOException {
         ParserResult result = BibtexParser.parse(new StringReader(bibtexEntry));
 
         Collection<BibEntry> entries = result.getDatabase().getEntries();
-        assertEquals(1, entries.size());
 
         BibEntry entry = entries.iterator().next();
-        assertEquals("test", entry.getCiteKey());
```
tobiasdiez: Why did you removed this (suboptimal) assertions?

src/test/java/net/sf/jabref/logic/bibtex/BibEntryWriterTest.java
```diff
@@ -149,18 +143,15 @@ public void roundTripWithModification() throws IOException {
...
-        assertEquals(5, entry.getFieldNames().size());
 
         // Modify entry
         entry.setField("author", "BlaBla");
 
         Set<String> fields = entry.getFieldNames();
         assertTrue(fields.contains("author"));
-        assertEquals("BlaBla", entry.getField("author"));
+        assertEquals("BlaBla", entry.getFieldOptional("author").get());
```
tobiasdiez: assertEquals(Optional.of(BlaBla), entry.getFieldOptional("author"))

src/test/java/net/sf/jabref/logic/bibtex/BibEntryWriterTest.java
lenhard: Are you being serious? In this very same PR you asked me to reduce the number asserts as far as possible, aiming for one assertion per test. This is the reason.

src/test/java/net/sf/jabref/logic/bibtex/BibEntryWriterTest.java
lenhard: Implemented

PR #1451, opened 2016-05-26
[WIP] Implementation of shared database support

obraliar:
Implementation of **shared database support** for MySQL, PostgreSQL and Oracle database systems.

Made changes:

- Old `sql`package removed. _(This request aims to replace the old system completely)_
- _Automatic_ change push to the server via EventBus. _(Support for all kinds of EntryEvents, see https://github.com/JabRef/jabref/pull/1028)_
- Full synchronization support for MySQL, PostgreSQL and Oracle
_(manual and semi-automated pull)_
- Conflict avoidance by version control and [_Optimistic Offline Lock_](http://martinfowler.com/eaaCatalog/optimisticOfflineLock.html)
- Efficient _pull_ using version numbers
- Synchronization of metadata with conflicts avoiding mechanisms
- Metadata appliance
- Groups synchronization
- Mode synchronization _(mode switch on the fly possible)_
- Adjusted tab and window titling
- New localization entries
- UI dialog for opening new or existing shared database
- UI dialogs for resolving conflicts between BibEntries
- Key bindings
- Connection loss handling
- Switch to offline working mode
- Driver availibilty check _(e.g. for the dropdown list)_
- Tests for all classes localized in shared package
- Synchronization tests

_Screenshots:_

open_rem_db

shareddb_sc1

shareddb_sc2

shareddb_sc3

conn_lost

- [x] Change in CHANGELOG.md described
- [x] Tests created for changes
- [x] Screenshots added (for bigger UI changes)

PR #1472, opened 2016-06-04
Refactor writing of bib files

tobiasdiez:
The main aim of this PR is to simplify the implementation of https://github.com/JabRef/jabref/pull/1451 (sql export) and other exporters.

- Abstract class BibDatabaseWriter, which controls the generic things of
writing a database (i.e. sorting entries, applying save actions, some
basic conversations, ...)
- Derived class BibtexDatabaseWriter, which only contains logic how to
actually write to in BibTeX format (should be simple to write similar
classes for other export formats like ris)
- Make SaveSession abstract and introduce two implementations which
write to a temporary file (FileSaveSession) or two a string/buffer
(StringSaveSession)
- Move code related to lock files to FileBasedLock

--

### DIFF
--- a/src/main/java/net/sf/jabref/exporter/BibDatabaseWriter.java
+++ b/src/main/java/net/sf/jabref/exporter/BibDatabaseWriter.java
@@ -69,12 +69,12 @@ public abstract class BibDatabaseWriter<E extends SaveSession> {
         List<FieldChange> changes = new ArrayList<>();
 
         Optional<FieldFormatterCleanups> saveActions = metaData.getSaveActions();
-        if (saveActions.isPresent()) {
+        saveActions.ifPresent(actions -> {
             // save actions defined -> apply for every entry
             for (BibEntry entry : toChange) {
-                changes.addAll(saveActions.get().applySaveActions(entry));
+                changes.addAll(actions.applySaveActions(entry));
             }
-        }
+        });
 
         return changes;
     }
@@ -104,11 +104,11 @@ public abstract class BibDatabaseWriter<E extends SaveSession> {
     }
 
     /*
-         * We have begun to use getSortedEntries() for both database save operations
-         * and non-database save operations.  In a non-database save operation
-         * (such as the exportDatabase call), we do not wish to use the
-         * global preference of saving in standard order.
-         */
+     * We have begun to use getSortedEntries() for both database save operations
+     * and non-database save operations.  In a non-database save operation
+     * (such as the exportDatabase call), we do not wish to use the
+     * global preference of saving in standard order.
+    */
     public static List<BibEntry> getSortedEntries(BibDatabaseContext bibDatabaseContext, List<BibEntry> entriesToSort,
             SavePreferences preferences) {
         Objects.requireNonNull(bibDatabaseContext);
@@ -241,14 +241,6 @@ public abstract class BibDatabaseWriter<E extends SaveSession> {
     protected abstract void writeMetaDataItem(Map.Entry<String, String> metaItem) throws SaveException;
 
     protected abstract void writePreamble(String preamble) throws SaveException;
-
-    private void writeUserCommentsForString(BibtexString string, Writer out) throws IOException {
-        String userComments = string.getUserComments();
-
-        if(!userComments.isEmpty()) {
-            out.write(userComments + Globals.NEWLINE);
-        }
-    }
 
     /**
      * Write all strings in alphabetical order, modified to produce a safe (for
--- a/src/main/java/net/sf/jabref/exporter/BibDatabaseWriter.java
+++ b/src/main/java/net/sf/jabref/exporter/BibDatabaseWriter.java
@@ -229,9 +229,7 @@ public abstract class BibDatabaseWriter<E extends SaveSession> {
      * Writes all data to the specified writer, using each object's toString() method.
      */
     protected void writeMetaData(MetaData metaData) throws SaveException {
-        if (metaData == null) {
-            return;
-        }
+        Objects.requireNonNull(metaData);
 
         Map<String, String> serializedMetaData = metaData.getAsStringMap();
 
@@ -296,10 +294,12 @@ public abstract class BibDatabaseWriter<E extends SaveSession> {
             String foundLabel = m.group(1);
             int restIndex = content.indexOf(foundLabel) + foundLabel.length();
             content = content.substring(restIndex);
-            Object referred = remaining.get(foundLabel.substring(1, foundLabel.length() - 1));
+            String label = foundLabel.substring(1, foundLabel.length() - 1);
+
             // If the label we found exists as a key in the "remaining" Map, we go on and write it now:
-            if (referred != null) {
-                writeString((BibtexString) referred, isFirstString, remaining, maxKeyLength, reformatFile);
+            if (remaining.containsKey(label)) {
+                BibtexString referred = remaining.get(label);
+                writeString(referred, isFirstString, remaining, maxKeyLength, reformatFile);
             }
         }
 
--- a/src/main/java/net/sf/jabref/exporter/BibDatabaseWriter.java
+++ b/src/main/java/net/sf/jabref/exporter/BibDatabaseWriter.java
@@ -1,22 +1,23 @@
-/*  Copyright (C) 2003-2016 JabRef contributors.
- This program is free software; you can redistribute it and/or modify
- it under the terms of the GNU General Public License as published by
- the Free Software Foundation; either version 2 of the License, or
- (at your option) any later version.
-
- This program is distributed in the hope that it will be useful,
- but WITHOUT ANY WARRANTY; without even the implied warranty of
- MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
- GNU General Public License for more details.
-
- You should have received a copy of the GNU General Public License along
- with this program; if not, write to the Free Software Foundation, Inc.,
- 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+/*
+ * Copyright (C) 2003-2016 JabRef contributors.
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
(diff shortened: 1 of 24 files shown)


tobiasdiez:
Any comments? Can I merge this in (of course after fixing compile errors)?

src/main/java/net/sf/jabref/exporter/BibDatabaseWriter.java
```diff
@@ -33,37 +34,50 @@
...
+
+    public interface SaveSessionFactory<E extends SaveSession> {
+        E createSaveSession(Charset encoding, Boolean makeBackup) throws SaveException;
+    }
+
+    private static List<FieldChange> applySaveActions(List<BibEntry> toChange, MetaData metaData) {
+        List<FieldChange> changes = new ArrayList<>();
+
+        Optional<FieldFormatterCleanups> saveActions = metaData.getSaveActions();
+        if (saveActions.isPresent()) {
```
Siedlerchr: Mabye use the lambda version: ifPresent

src/main/java/net/sf/jabref/exporter/BibDatabaseWriter.java
```diff
@@ -208,162 +197,60 @@ public SaveSession savePartOfDatabase(BibDatabaseContext bibDatabaseContext, Lis
                         entryType -> typesToWrite.put(entryType.getName(), entryType));
```
Siedlerchr: I think this is double code: First you check ! present and then you check the same with ifPresent..

src/main/java/net/sf/jabref/exporter/BibDatabaseWriter.java
```diff
@@ -208,162 +197,60 @@ public SaveSession savePartOfDatabase(BibDatabaseContext bibDatabaseContext, Lis
...
-        }
-    }
+    protected abstract void writeEpilogue(String epilogue) throws SaveException;
 
     /**
      * Writes all data to the specified writer, using each object's toString() method.
      */
-    private void writeMetaData(Writer out, MetaData metaData) throws IOException {
+    protected void writeMetaData(MetaData metaData) throws SaveException {
         if (metaData == null) {
```
Siedlerchr: Maybe use optional?

src/main/java/net/sf/jabref/exporter/BibDatabaseWriter.java
```diff
@@ -376,32 +263,55 @@ private void writeStrings(Writer fw, BibDatabase database, Boolean reformatFile)
...
+        // Then we go through the string looking for references to other strings. If we find references
+        // to strings that we will write, but still haven't, we write those before proceeding. This ensures
+        // that the string order will be acceptable for BibTeX.
+        String content = bibtexString.getContent();
+        Matcher m;
+        while ((m = REFERENCE_PATTERN.matcher(content)).find()) {
+            String foundLabel = m.group(1);
+            int restIndex = content.indexOf(foundLabel) + foundLabel.length();
+            content = content.substring(restIndex);
+            Object referred = remaining.get(foundLabel.substring(1, foundLabel.length() - 1));
```
Siedlerchr: First: Why Object? And second: Better use getOrDefault, avoids the null check

src/main/java/net/sf/jabref/exporter/BibtexDatabaseWriter.java
```diff
@@ -0,0 +1,167 @@
...
+    protected void writeEntry(BibEntry entry, BibDatabaseMode mode, Boolean isReformatFile) throws SaveException {
+        BibEntryWriter bibtexEntryWriter = new BibEntryWriter(new LatexFieldFormatter(), true);
+        try {
+            bibtexEntryWriter.write(entry, getWriter(), mode, isReformatFile);
+        } catch (IOException e) {
+            throw new SaveException(e, entry);
+        }
+    }
+
+    private Writer getWriter() {
```
Siedlerchr: I would suggest making that a field, e.g. Writer writer and use that everywhere instead of getWriter()
Improves readability

src/main/java/net/sf/jabref/exporter/ExportFormat.java
```diff
@@ -190,15 +190,15 @@ public void performExport(final BibDatabaseContext databaseContext, final String
         SaveSession ss = null;
```
Siedlerchr: Use nio Path method above.

src/main/java/net/sf/jabref/exporter/FileSaveSession.java
```diff
@@ -0,0 +1,144 @@
...
+    private static Path createTemporaryFile() throws SaveException {
+        try {
+            return Files.createTempFile(FileSaveSession.TEMP_PREFIX, FileSaveSession.TEMP_SUFFIX);
+        } catch (IOException e) {
+            throw new SaveException(e);
+        }
+    }
+
+    @Override
+    public void commit(File file) throws SaveException {
```
Siedlerchr: Use nio methods 
e.g. File.exist
and  sth like Path.get(path, name, backup Extension)

src/main/java/net/sf/jabref/exporter/FileSaveSession.java
```diff
@@ -0,0 +1,144 @@
...
+                        if (!FileBasedLock.waitForFileLock(file, 10)) {
+                            throw SaveException.FILE_LOCKED;
+                        }
+                    }
+                } catch (IOException ex) {
+                    LOGGER.error("Error when creating lock file.", ex);
+                }
+            }
+
+            FileUtil.copyFile(temporaryFile.toFile(), file, true);
```
Siedlerchr: Use Files.copy...

src/main/java/net/sf/jabref/logic/util/io/FileBasedLock.java
```diff
@@ -54,16 +59,17 @@ public static boolean hasLockFile(File file) {
      * @return the last modified time if lock file exists, -1 otherwise.
      */
     public static long getLockFileTimeStamp(File file) {
-        File lock = new File(file.getPath() + SaveSession.LOCKFILE_SUFFIX);
+        File lock = new File(file.getPath() + LOCKFILE_SUFFIX);
```
Siedlerchr: Paths.get

src/main/java/net/sf/jabref/logic/util/io/FileBasedLock.java
```diff
@@ -72,4 +78,25 @@ public static boolean deleteLockFile(File file) {
...
     }
+
+    /**
+     * Check if a lock file exists, and create it if it doesn't.
+     *
+     * @return true if the lock file already existed
+     * @throws IOException if something happens during creation.
+     */
+    public static boolean createLockFile(File file) throws IOException {
+        File lock = new File(file.getPath() + LOCKFILE_SUFFIX);
```
Siedlerchr: same as above, Paths.,get

src/main/java/net/sf/jabref/logic/util/io/FileBasedLock.java
```diff
@@ -72,4 +78,25 @@ public static boolean deleteLockFile(File file) {
...
+        if (lock.exists()) {
+            return true;
+        }
+        try (FileOutputStream out = new FileOutputStream(lock)) {
+            out.write(0);
+            out.close();
```
Siedlerchr: Is the close necessary? When you use a try with resource (Autocloseable), normally all Streams are closed and freed

src/main/java/net/sf/jabref/logic/util/strings/StringUtil.java
```diff
@@ -653,4 +653,7 @@ public static String repeat(int n, char c) {
 
     }
 
+    public static boolean isNullOrEmpty(String toTest) {
+        return (toTest == null || toTest.isEmpty());
```
Siedlerchr: You can use the apache commons StringUtil method for this

src/test/java/net/sf/jabref/exporter/BibtexDatabaseWriterTest.java
```diff
@@ -40,186 +37,175 @@
...
     @Test(expected = NullPointerException.class)
-    public void writeWithNullPreferencesThrowsException() throws IOException {
-        databaseWriter.writePartOfDatabase(mock(Writer.class), bibtexContext, Collections.emptyList(), null);
+    public void writeWithNullPreferencesThrowsException() throws Exception {
+        databaseWriter.savePartOfDatabase(bibtexContext, Collections.emptyList(), null);
     }
 
     @Test
-    public void writeEncoding() throws IOException {
+    public void writeEncoding() throws Exception {
```
Siedlerchr: Why did you remove all specific Exceptions?
Better let it throw the most specific exception. Otherwise stacktrace can be hard to find.

Siedlerchr:
Overall looks good, just some small remarks from my side :+1:

src/main/java/net/sf/jabref/exporter/BibDatabaseWriter.java
tobiasdiez: Yeah, but I find the old-style version more readable.

src/main/java/net/sf/jabref/exporter/BibDatabaseWriter.java
tobiasdiez: With a small difference: the first check is for `StandardType` (i.e. if it's a JabRef default type) and the second one a `KnownType` (i.e. standard or user-specfied).

src/main/java/net/sf/jabref/exporter/BibDatabaseWriter.java
tobiasdiez: `requireNonNull` is even more appropriate, I think.

src/test/java/net/sf/jabref/exporter/BibtexDatabaseWriterTest.java
tobiasdiez: Makes the tests more invariant under changes of the code-under-test. (and is default template for IntellliJ)

tobiasdiez:
@Siedlerchr Thanks for you feedback. Incorporated it and rebased.

src/main/java/net/sf/jabref/exporter/BibDatabaseWriter.java
```diff
@@ -90,11 +104,11 @@
     }
 
     /*
-     * We have begun to use getSortedEntries() for both database save operations
-     * and non-database save operations.  In a non-database save operation
-     * (such as the exportDatabase call), we do not wish to use the
-     * global preference of saving in standard order.
-     */
+         * We have begun to use getSortedEntries() for both database save operations
```
koppor: indent?

src/main/java/net/sf/jabref/exporter/BibDatabaseWriter.java
simonharrer: Optional only make sense if one uses ifPresent or map/flatMap. Otherwise, it does not really help that much, from my perspective, as one still requires an if check and a get, making it even more complicated to use instead of a null check and just using it.

simonharrer:
@tobiasdiez after rebase and successful tests, this can be merged as well.

PR #1474, opened 2016-06-05
Fix #1434: Static groups are now longer displayed as dynamic ones

### LINKED ISSUE #1434: Static groups from 3.3 converted to dynamic ones in 3.4dev
JabRef version 3.4dev--snapshot--2016-05-18--master--f4270a6 on Windows 10
(same issue is also found in JabRef version 3.4dev--snapshot--2016-05-18--master--2bd4c4a)

It seems that static groups of the current stable branch of JabRef 3.3 are converted to dynamic ones, in the new developer version 3.4

Steps to reproduce:

1. Open a Jabref database from the current stable release (3.3)
2. Previous static groups are displayed in italic (like dynamic groups)
3. Rightclick --> Edit group --> Shows the group as "Dynamically group entries by searching a field for a keyword"
4. The "Keyword" is the group name

tobiasdiez:
Reason for this bug: ExplicteGroup now derives from KeywordGroup and thus was sometimes displayed as a keyword group.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 ### Fixed
 - Fixed [#405](https://github.com/JabRef/jabref/issues/405): Added more {} around capital letters in Unicode/HTML to LaTeX conversion to preserve them
 - Alleviate multiuser concurrency issue when near simultaneous saves occur to a shared database file
-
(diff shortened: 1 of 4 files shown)


simonharrer:
Why does it make sense that the explicit group extends the keyword group? I do not see any advantages from that, despite making it more complex. Is really a lot of code saved? Maybe duplicating a few lines of code (in explicit and keyword group) creates a simpler structure overall?

simonharrer:
Thanks for the detailed explanation - makes sense. :)

Instead of relying on the class, I would recommend to introduce an enum that differentiates the group types, and which is returned through a method getType which each class overwrites if necessary.

PR #1479, opened 2016-06-06
Rewrite MedlineImporter with JAXB and add nbib fields

tschechlovdev:
Regarding:[#1063] (https://github.com/JabRef/jabref/issues/1063)
I've rewritten the MedlineImporter (to be more specific: the `importDatabase` method), so that it uses a JAXB parser now, instead of a SAX parser. I've also added the nbib fields to the importer. Tests for the changes will be coming.


CHANGELOG.md
```diff
@@ -73,7 +73,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - Integrity check table can be sorted by clicking on column headings
 - Added \SOFTWARE\Jabref 'Path' registry entry for installation path inside the installer
 - Added an additional icon to distinguish DOI and URL links ([feature request #696](https://github.com/JabRef/jabref/issues/696))
-- Added nbib fields to Medlineplain importer
+- Added nbib fields to Medlineplain importer and to MedlineImpporter
```
boceckts: Typo

src/main/java/net/sf/jabref/importer/fileformat/MedlineImporter.java
```diff
@@ -83,50 +118,349 @@ public boolean isRecognizedFormat(BufferedReader reader) throws IOException {
...
+                    PubmedArticle currentArticle = (PubmedArticle) articleSet.getPubmedArticleOrPubmedBookArticle()
+                            .get(i);
+                    parseArticle(currentArticle, bibItems);
+                }
+            } else {
+                PubmedArticle article = (PubmedArticle) o;
+                parseArticle(article, bibItems);
+            }
+        } catch (JAXBException e1) {
+            LOGGER.warn("Could not parse document");
```
boceckts: Using `LOGGER.debug("Message", e)` writes the stacktrace to the error console with its stacktrace which could be useful to the user/developer to obtain more information. I would also directly notify the user by using `JabRefGUI.getMainFrame().output(Localization.lang("Error") + ": " + e.getLocalizedMessage());`

src/main/java/net/sf/jabref/importer/fileformat/MedlineImporter.java
```diff
@@ -83,50 +118,349 @@ public boolean isRecognizedFormat(BufferedReader reader) throws IOException {
...
+            return pageRange;
+        }
+        String first = pageRange.substring(0, minusPos).trim();
+        String last = pageRange.substring(minusPos + 1).trim();
+        int llast = last.length();
+        int lfirst = first.length();
+        if (llast < lfirst) {
+            last = first.substring(0, lfirst - llast) + last;
+        }
+        return first + "--" + last;
```
boceckts: Please rename

src/test/resources/net/sf/jabref/importer/fileformat/MedlineImporterTestSingleEntry.xml
```diff
@@ -1,4 +1,3 @@
-<PubmedArticleSet>
 <PubmedArticle>
 <MedlineCitation Owner="NLM" Status="MEDLINE">
 <PMID Version="1">1</PMID>
```
boceckts: Also cover multiple PMIDs in a single entry

src/main/java/net/sf/jabref/logic/fileformats/medline/AccessionNumberList.java
```diff
@@ -0,0 +1,78 @@
...
+ */
+@XmlAccessorType(XmlAccessType.FIELD)
+@XmlType(name = "", propOrder = {
+    "accessionNumber"
+})
+@XmlRootElement(name = "AccessionNumberList")
+public class AccessionNumberList {
+
+    @XmlElement(name = "AccessionNumber", required = true)
+    protected List<String> accessionNumber;
```
boceckts: I would say rename it to `accessionNumbers` or `accessionNumberList` or `accessionNumbersList`. But since this applies to all auto generated files I don't know if it makes much sense to change it in all of the files.

src/main/java/net/sf/jabref/importer/fileformat/MedlineImporter.java
```diff
@@ -83,50 +118,349 @@ public boolean isRecognizedFormat(BufferedReader reader) throws IOException {
...
+            if (chemical != null) {
+                chemicalNames.add(chemical.getNameOfSubstance().getContent());
+            }
+        }
+        fields.put("chemicals", StringUtils.join(chemicalNames, ", "));
+    }
+
+    private void addArticleInformation(HashMap<String, String> fields, List<Object> content) {
+        //since the content is a list of Object, we go through the list
+        //and look of which class the object is an instance
```
boceckts: I think this is unecessary

src/main/java/net/sf/jabref/importer/fileformat/MedlineImporter.java
```diff
@@ -83,50 +118,349 @@ public boolean isRecognizedFormat(BufferedReader reader) throws IOException {
...
-            LOGGER.error("Error during file import", e);
-            return ParserResult.fromErrorMessage(e.getLocalizedMessage());
+    private void addKeyWords(HashMap<String, String> fields, List<KeywordList> keywordList) {
+        List<String> keywordStrings = new ArrayList<>();
+        //add keywords to the list
+        for (KeywordList keywords : keywordList) {
+            for (Keyword keyword : keywords.getKeyword()) {
+                for (Serializable content : keyword.getContent()) {
+                    if (content instanceof String) {
+                        keywordStrings.add((String) content);
```
boceckts: Can the names for the two lists be more specified?

src/main/java/net/sf/jabref/importer/fileformat/MedlineImporter.java
```diff
@@ -83,50 +118,349 @@ public boolean isRecognizedFormat(BufferedReader reader) throws IOException {
...
+                chemicalNames.add(chemical.getNameOfSubstance().getContent());
+            }
+        }
+        fields.put("chemicals", StringUtils.join(chemicalNames, ", "));
+    }
+
+    private void addArticleInformation(HashMap<String, String> fields, List<Object> content) {
+        //since the content is a list of Object, we go through the list
+        //and look of which class the object is an instance
+        for (Object o : content) {
```
boceckts: Please rename the Object

src/main/java/net/sf/jabref/importer/fileformat/MedlineImporter.java
```diff
@@ -83,50 +118,349 @@ public boolean isRecognizedFormat(BufferedReader reader) throws IOException {
...
-            // Switch this to true if you want to make a local copy for testing.
-            if (false) {
-                reader.reset();
-                try (FileOutputStream out = new FileOutputStream(new File("/home/alver/ut.txt"))) {
-                    int c;
-                    while ((c = reader.read()) != -1) {
-                        out.write((char) c);
+            JAXBContext context = JAXBContext.newInstance("net.sf.jabref.logic.fileformats.medline");
+            Unmarshaller unmarshaller = context.createUnmarshaller();
+            Object o = unmarshaller.unmarshal(reader);
```
boceckts: Please rename the object.

boceckts:
Very good work :+1:, only a few remarks from my side.

src/test/resources/net/sf/jabref/importer/fileformat/MedlineImporterTestSingleEntry.xml
```diff
@@ -1,7 +1,7 @@
-<PubmedArticleSet>
```
koppor: Why was this removed?

src/test/resources/net/sf/jabref/importer/fileformat/MedlineImporterTestSingleEntry.xml
koppor: Why wasn't this handled?

src/test/resources/net/sf/jabref/importer/fileformat/MedlineImporterTestSingleEntry.bib
```diff
@@ -1,15 +1,25 @@
...
+  keywords = {Aldehyde Oxidoreductases, metabolism; Animals; Body Fluids, analysis; Carbon Dioxide, blood; Formates, blood, poisoning; Haplorhini; Humans; Hydrogen-Ion Concentration; Kinetics; Methanol, blood; Methods; Pseudomonas, enzymology},
+  nlm-id = {0151424},
+  owner = {NLM},
+  pages = {117--126},
+  pmid = {10097079},
+  pubmodel = {Print},
+  pubstatus = {ppublish},
+  revised = {21/11/2013},
+  status = {MEDLINE},
+  title = {[Formate assay in body fluids: application in methanol poisoning.]},
```
tobiasdiez: Remove surrounding `[]`

src/test/resources/net/sf/jabref/importer/fileformat/MedlineImporterTestSingleEntry.bib
```diff
@@ -1,15 +1,25 @@
...
+  volume = {13},
+  year = {1999 Spring}
```
tobiasdiez: Without Springer

koppor:
Please do not include the generated classes. Try to use https://github.com/jacobono/gradle-jaxb-plugin to generate them. Does that work?

src/test/resources/net/sf/jabref/importer/fileformat/MedlineImporterTestSingleEntry.xml
tschechlovdev: Just to test the case, that PubMedArticleSet doesn't appear in a file.

src/test/resources/net/sf/jabref/importer/fileformat/MedlineImporterTestSingleEntry.xml
tschechlovdev: I weren't sure how to handle this. I thought that the PMID with the highest version should be added, since it is the newest version. Should the PMID with Version = 1 also be added? For example as pmid-version1?

matthiasgeiger:
Can you please document the source of the XSD you used to create the JAXB classes. Thanks!

src/main/java/net/sf/jabref/importer/fileformat/MedlineImporter.java
koppor: Don't get the importer an OutputPrinter passed? That should be used instead
of directly communicating with the UI. Reason: Usage of JabRef in the
command line without UI

src/main/java/net/sf/jabref/logic/fileformats/medline/AccessionNumberList.java
koppor: The whole package is auto generated (isn't it?) -- the generation (based on
the XSD) should be triggered during the build and the source reside in
...gen... (similar to the ANTLR grammar)

tschechlovdev:
@koppor I didn't get the plugin work. But I've wrote an ant task, that I've found [here](https://joerglenhard.wordpress.com/2012/01/10/xjc-and-schemagen-with-gradle/). This works local, but for some reason travis ci says, that src/main/gen is a non existend directory.

@matthiasgeiger I wrote the source as a comment in the xsd file(s), is that ok ?

tschechlovdev:
@koppor I didn't get the plugin work. But I've wrote an ant task, that I've found [here](https://joerglenhard.wordpress.com/2012/01/10/xjc-and-schemagen-with-gradle/). This works local, but for some reason travis ci says, that src/main/gen is a non existend directory.

@matthiasgeiger I wrote the source as a comment in the xsd file(s), is that ok ?

src/test/resources/net/sf/jabref/importer/fileformat/MedlineImporterTestSingleEntry.bib
tschechlovdev: done.

src/test/resources/net/sf/jabref/importer/fileformat/MedlineImporterTestSingleEntry.bib
tschechlovdev: done.

build.gradle
```diff
@@ -373,3 +373,29 @@ jmh {
...
+ schemaTargetDir.mkdirs()
+}
+ 
+task xjc () { 
+ ant.taskdef(name: 'xjc', classname: 'com.sun.tools.xjc.XJCTask', classpath: configurations.xjc.asPath)
+ 
+ ant.xjc(
+  destdir: schemaTargetDir,
+  schema: 'src/main/resources/xjc/medline.xsd'
+ )
```
simonharrer: This needs to be executed at runtime, not during configuration time. Please have a look at the gradle documentation.

build.gradle
```diff
@@ -373,3 +373,29 @@ jmh {
     iterations = 10
     fork = 2
 }
+
+configurations {
```
simonharrer: could all this be extracted into a separate gradle file which we import? Similar to the eclipse stuff?

src/main/java/net/sf/jabref/importer/fileformat/MedlineImporter.java
```diff
@@ -84,50 +121,371 @@ public boolean isRecognizedFormat(BufferedReader reader) throws IOException {
...
-            parser.parse(new InputSource(reader), handler);
-
-            // Switch this to true if you want to make a local copy for testing.
-            if (false) {
-                reader.reset();
-                try (FileOutputStream out = new FileOutputStream(new File("/home/alver/ut.txt"))) {
-                    int c;
-                    while ((c = reader.read()) != -1) {
-                        out.write((char) c);
+            JAXBContext context = JAXBContext.newInstance("generated");
```
simonharrer: Can you use a typed JAXBContext by passing the XXX.class, with XXX being the one with the `@XmlRootElement`?

simonharrer:
Made some comments, but in general LGTM

src/main/java/net/sf/jabref/importer/fileformat/MedlineImporter.java
```diff
@@ -16,29 +16,66 @@
...
+import javax.xml.bind.JAXBElement;
+import javax.xml.bind.JAXBException;
+import javax.xml.bind.Unmarshaller;
 
 import net.sf.jabref.importer.ParserResult;
 import net.sf.jabref.model.entry.BibEntry;
+import net.sf.jabref.model.entry.IdGenerator;
 
+import com.google.common.base.Joiner;
+import generated.Abstract;
```
koppor: I would use `net.sf.jabref.importer.fileformat.medline`. 
Reasons:
  -  Consistency with our ANTLR generation, which generates `src/main/gen/net/sf/jabref/bst/`. 
  - The common prefix `generated` does not indicate anyhting about medline. What if other XSDs will be generated?

Alternatively, use a combinded solution: `net.sf.jabref.importer.fileformat.medline.generated`.

src/main/java/net/sf/jabref/importer/fileformat/MedlineImporter.java
```diff
@@ -84,50 +121,371 @@ public boolean isRecognizedFormat(BufferedReader reader) throws IOException {
...
-                        out.write((char) c);
+            JAXBContext context = JAXBContext.newInstance("generated");
+            Unmarshaller unmarshaller = context.createUnmarshaller();
+            Object unmarshalledObject = unmarshaller.unmarshal(reader);
+
+            //check whether we have an article set or just an article
+            if (unmarshalledObject instanceof PubmedArticleSet) {
+                PubmedArticleSet articleSet = (PubmedArticleSet) unmarshalledObject;
+                for (Object article : articleSet.getPubmedArticleOrPubmedBookArticle()) {
+                    if (article instanceof PubmedArticle) {
```
koppor: It seems that PubmedBookArticle is another possibilty. This seems to be untested. Why?

src/main/java/net/sf/jabref/importer/fileformat/MedlineImporter.java
```diff
@@ -84,50 +121,371 @@ public boolean isRecognizedFormat(BufferedReader reader) throws IOException {
...
+                }
+            }
+        }
+        //Check whether MeshHeadingList exist or not
+        if (fields.get("keywords") == null) {
+            fields.put("keywords", join(keywordStrings, "; "));
+        } else {
+            if (keywordStrings.size() > 0) {
+                //if it exists, combine the MeshHeading with the keywords
+                String result = join(keywordStrings, "; ");
```
koppor: Keywords should be joined with the configured at `net.sf.jabref.JabRefPreferences.KEYWORD_SEPARATOR`.

src/main/java/net/sf/jabref/importer/fileformat/MedlineImporter.java
```diff
@@ -84,50 +121,371 @@ public boolean isRecognizedFormat(BufferedReader reader) throws IOException {
...
+    private String extractYear(String medlineDate) {
+        //The year of the medlineDate should be the first 4 digits
+        return medlineDate.substring(0, 4);
+    }
+
+    private String removeSurroundingBrackets(String title) {
+        if (title.startsWith("[") && title.endsWith("]")) {
+            return title.replace("[", "").replace("]", "");
+        }
+        return title;
```
koppor: Why is this untested?

src/test/resources/net/sf/jabref/importer/fileformat/MedlineImporterTestArticleID.xml
```diff
@@ -30,7 +27,7 @@
             </Journal>
             <ArticleTitle>Diagnostic performance of a multiplex PCR assay for meningitis in an HIV-infected population in Uganda.</ArticleTitle>
             <Pagination>
-                <MedlinePgn/>
+            <MedlinePgn>1234</MedlinePgn>
```
koppor: Why did you change the test case?

Where does `MedlineImporterTestArticleID.xml` come from?

I think, you should NEVER change a REAL file in a test.

src/main/java/net/sf/jabref/importer/fileformat/MedlineImporter.java
simonharrer: Remove surrounding brackes looks like a util method, and should maybe go into the StringUtil class?

Siedlerchr: In the StringUtil class there is also a similar method for the curly braces.

tobiasdiez: There is also RemoveBracesFormatter which does a similar job (and is a bit more intelligent, I think). For example your code turns `[te[st]` into `test` and not `te[st`.

src/main/java/net/sf/jabref/importer/fileformat/MedlineImporter.java
tschechlovdev: Shouldn't that be the same?

src/test/resources/net/sf/jabref/importer/fileformat/MedlineImporterTestArticleID.xml
tschechlovdev: I'm not sure from where the xml file comes. This file was made in an earlier pull request by @mairdl .
But like it was before, the document wasn't parsed, because it wasn't even well formed (closing tag, but no opening tag). That's why I changed it.

simonharrer:
LGTM :+1:

tschechlovdev:
So can this be merged in ? Or are there any other remarks?

matthiasgeiger:
Sorry, but this does not work at all for me.

Steps to reproduce: 

- go to https://www.ncbi.nlm.nih.gov/pubmed/?term=flu
- select one or more entries
- click on "Send to:" -> "File" -> "XML"
- Open JabRef and try to "Import into new database" the created XML file

What happens? There are two dialogs showing nothing: 
medline-import1

medline-import2

Followed by an error message:
medline-import3

Am I doing sth wrong?

tschechlovdev:
Yeah you're right. I'm getting the same problem. But for some reason, the files that I've tested are working. I'm looking for the problem.

tschechlovdev:
@matthiasgeiger now it should work. I'm not sure why, but somehow the first two lines
``<?xml version="1.0"?>
<!DOCTYPE PubmedArticleSet PUBLIC "-//NLM//DTD PubMedArticle, 1st January 2016//EN" "http://www.ncbi.nlm.nih.gov/corehtml/query/DTD/pubmed_160101.dtd">``
 were causing trouble. My solution is to skip the first lines, till we reach the first line, that contains <pubmedarticle>, <pubmedarticleset>, <pubmedbookarticle> or <pubmedbookarticleset>.

src/main/java/net/sf/jabref/importer/fileformat/MedlineImporter.java
```diff
@@ -172,6 +183,33 @@ public ParserResult importDatabase(BufferedReader reader) throws IOException {
         return new ParserResult(bibItems);
     }
 
+    private int SkipReaderToFirstElement(BufferedReader reader) {
```
koppor: XML parsers should do that work. Do not walk through XML by hand.

Please remove this method and try the solution presented at http://stackoverflow.com/a/22204172/873282

`xsr.next();` from XMLStreamReader seems to be your friend.

koppor:
XML parsers should do that work. Do not walk through XML by hand.

Please remove this method and try the solution presented at http://stackoverflow.com/a/22204172/873282

`xsr.next();` from XMLStreamReader seems to be your friend.

src/test/resources/net/sf/jabref/importer/fileformat/MedlineImporterTestSingleEntry.xml
koppor: Why not adding a second test case? - Modified test files always cause a strange feeling

src/test/resources/net/sf/jabref/importer/fileformat/MedlineImporterTestSingleEntry.xml
koppor: The latest one is OK. I just forgot that agreement.

src/main/java/net/sf/jabref/importer/fileformat/MedlineImporter.java
```diff
@@ -84,50 +140,532 @@ public boolean isRecognizedFormat(BufferedReader reader) throws IOException {
...
+                fields.put("author", join(personalNames, " and "));
+            }
+        }
+    }
+
+    private void addMeashHeading(HashMap<String, String> fields, MeshHeadingList meshHeadingList) {
+        ArrayList<String> keywords = new ArrayList<>();
+        String result = "";
+        for (MeshHeading keyword : meshHeadingList.getMeshHeading()) {
+            result = keyword.getDescriptorName().getContent();
```
koppor: Declare result variable here, not out of the loop

src/test/resources/net/sf/jabref/importer/fileformat/MedlineImporterTestSingleEntry.bib
```diff
@@ -1,15 +1,25 @@
-% Encoding: UTF-8
-
-@Article{,
```
koppor: Why was the complete test case changed? Can't the old test case be kept and an additional test case be added?

src/main/java/net/sf/jabref/importer/fileformat/MedlineImporter.java
```diff
@@ -84,50 +140,532 @@ public boolean isRecognizedFormat(BufferedReader reader) throws IOException {
...
+        }
+        if (!"".equals(result)) {
+            fields.put(medlineKey, result);
+        }
+    }
+
+    private void addContributionDate(HashMap<String, String> fields, ContributionDate contributionDate) {
+        if ((contributionDate.getDay() != null) && (contributionDate.getMonth() != null)
+                && (contributionDate.getYear() != null)) {
+            String result = contributionDate.getYear() + "/" + contributionDate.getMonth() + "/"
```
koppor: Why isn't `-` used as separator? See also https://xkcd.com/1179/ and #1575

koppor: Maybe String.format can be used to construct the string?

src/main/java/net/sf/jabref/importer/fileformat/MedlineImporter.java
```diff
@@ -84,50 +140,532 @@ public boolean isRecognizedFormat(BufferedReader reader) throws IOException {
...
+            if (medlineCitation.getInvestigatorList() != null) {
+                addInvestigators(fields, medlineCitation.getInvestigatorList());
+            }
+            if (medlineCitation.getGeneralNote() != null) {
+                addNotes(fields, medlineCitation.getGeneralNote());
+            }
+
+        }
+
+        BibEntry entry = new BibEntry(IdGenerator.next(), "article"); // id assumes an existing database so don't create one here
```
koppor: No need for the comment here. I think it's more confusing than helping.

src/main/java/net/sf/jabref/importer/fileformat/MedlineImporter.java
```diff
@@ -84,50 +140,532 @@ public boolean isRecognizedFormat(BufferedReader reader) throws IOException {
...
+                if (author.getForeName() != null) {
+                    authorName += ", " + author.getForeName();
+                }
+                authorNames.add(authorName);
+            }
+        }
+        fields.put("author", join(authorNames, " and "));
+    }
+
+    private String join(List<String> list, String string) {
```
koppor: make this static. Not sure, whether `Joiner.on(string).join(list)` should be just inlined as this is a oneliner.

src/main/java/net/sf/jabref/importer/fileformat/MedlineImporter.java
```diff
@@ -84,50 +140,532 @@ public boolean isRecognizedFormat(BufferedReader reader) throws IOException {
...
+            }
+        }
+        if (book.getIsbn() != null) {
+            fields.put("isbn", join(book.getIsbn(), ", "));
+        }
+    }
+
+    private void putStringFromSerializableList(HashMap<String, String> fields, String medlineKey,
+            List<Serializable> contentList) {
+        String result = "";
```
koppor: Not sure whether the "pattern" from the other methods should be used. `List` and `Joiner.join(list)`

koppor:
Other than my small comments: looks good (codewise). Haven't tried the UI.

src/main/java/net/sf/jabref/importer/fileformat/MedlineImporter.java
```diff
@@ -84,50 +140,532 @@ public boolean isRecognizedFormat(BufferedReader reader) throws IOException {
...
-            return ParserResult.fromErrorMessage(e.getLocalizedMessage());
-        } catch (org.xml.sax.SAXException e) {
-            LOGGER.error("Error during XML parsing", e);
-            return ParserResult.fromErrorMessage(e.getLocalizedMessage());
-        } catch (IOException e) {
-            LOGGER.error("Error during file import", e);
+        } catch (JAXBException e1) {
+            LOGGER.debug("could not parse document", e1);
+            return ParserResult.fromErrorMessage(e1.getLocalizedMessage());
+        } catch (XMLStreamException e) {
```
koppor: combine these two exeptions by `JAXBException | XMLStreamExeption e`

src/main/java/net/sf/jabref/importer/fileformat/MedlineImporter.java
Siedlerchr: I rather would discard assembling a String here Use some kind of Date Object and then use a toString method. Have a look at #1575, there were  introduced some new methods regarding Dates,

src/main/java/net/sf/jabref/importer/fileformat/MedlineImporter.java
```diff
@@ -84,50 +140,532 @@ public boolean isRecognizedFormat(BufferedReader reader) throws IOException {
...
+        return Joiner.on(string).join(list);
+    }
+
+    private void addDateRevised(HashMap<String, String> fields, DateRevised dateRevised) {
+        if ((dateRevised.getDay() != null) && (dateRevised.getMonth() != null) && (dateRevised.getYear() != null)) {
+            fields.put("revised", dateRevised.getDay() + "/" + dateRevised.getMonth() + "/" + dateRevised.getYear());
+        }
+    }
+
+    private void putIfNotNull(HashMap<String, String> fields, String medlineKey, String value) {
```
Siedlerchr: I would suggest renaming it, to make it more clear:_
putIfValueNotNull
Otherwise it could be confused with the jjava standard method putIfAbsent

src/main/java/net/sf/jabref/importer/fileformat/MedlineImporter.java
```diff
@@ -84,50 +140,532 @@ public boolean isRecognizedFormat(BufferedReader reader) throws IOException {
...
+            List<AuthorList> authorLists = book.getAuthorList();
+            //authorLists size should be one
+            if (authorLists.size() == 1) {
+                for (AuthorList authorList : authorLists) {
+                    handleAuthors(fields, authorList);
+                }
+            }
+        }
+
+        putIfNotNull(fields, "volume", book.getVolume());
```
Siedlerchr: I am not sure, but isn't there somewhere an Enum with all kind of BibTex/BibLatex fields?

src/main/java/net/sf/jabref/importer/fileformat/MedlineImporter.java
tschechlovdev: Hm, I could also use the  [SimpleDateFormat] (http://stackoverflow.com/questions/882420/parse-string-to-date-with-different-format-in-java) in combinition with the EasyDateFormat. Then I can construct a date object with SimpleDateFormat and then get the date as String with the EasyDateFormat.getDateAt(...) method. Did you mean something like that? 

Also should I use this for the other dates as well ?

src/main/java/net/sf/jabref/importer/fileformat/MedlineImporter.java
tschechlovdev: I'm not sure how I could use this here. I have a  list of Serializable and have to check if some of them is an instance of String and then I can add them to the result String.  
I had to do something like this ``fields.put(medlineKey, join("", (List<String>) contentList);``. But I cannot cast a Serializiablle list to a String list.

src/main/java/net/sf/jabref/importer/fileformat/MedlineImporter.java
Siedlerchr: Kind of. But don't use SimpleDateFormat, as it was not threadsafe and got eplaced wiht a new Date and Time api in Java8. See Example 18 from this link:
http://javarevisited.blogspot.de/2015/03/20-examples-of-date-and-time-api-from-Java8.html

What exactly do you store in the ContributionDate and the other Date-Objects? The pattern?

oscargus: Since the parts are in separate Strings(?) already I think it is OK to concatenate. If they where ints it might have been a different thing. But I also think that - is a better separator.

tschechlovdev: I just have the value of the year, month and the day as Strings in the Date-objects. In the xml it looks like this: 
``<year>2012<year>
<month>12<month>
<day>05<day>``
and I can use getYear(), getMonth(), getDay() to get them as separate Strings and I want to put them together as one date (as String).

Siedlerchr: Okay, then do what @koppor  suggested. Use a String.Format. Simplest solution. ;)

src/main/java/net/sf/jabref/importer/fileformat/MedlineImporter.java
koppor: We only have net.sf.jabref.model.entry.BibtexEntryTypes and `net.sf.jabref.Globals.FILE_FIELD`, but we want to get rid of Globals. Thus, there is none and it is OK to use the Strings as literals even though typos might occur.

simonharrer: In #1579 we discussed this, and we aim to create a class with all the field names as constants in the model.

tschechlovdev:
I've adressed the comments. Are there any other remarks?

koppor:
LGTM

tschechlovdev:
So can this be merged?

PR #1483, opened 2016-06-06
Fixes #1482: Correct number of matched entries is displayed for refining subgroups

### LINKED ISSUE #1482: Wrong number in groups
JabRef version 2.4 on WindowsXP (I know, I know..., but same thing happen in Ubuntu)

Some groups (not all) display a wrong number of entries. In my example, it's written 577 but it should be 38 (and 38 entries are well displayed). 
It also have error in "summer student", "International", "congrès-chris", "international"...

They are both dynamic groups, but many other dynamic groups are fine. I think is related to a regular expression with != 
(in this case, it is Note!=. )

jabref

I could send the file, but in private, since it contains a lots of confidential information.

tobiasdiez:
The issue was that, for the number of hits, JabRef only checked if the
group taken separately matched the entry. That is, it was completely
ignored that the group sits in a tree and might be set up to refine the
search of the parent. Taking the example given in #1482, 577 entries
satisfy the condition `articlé original` but only 38 are also matched by
`articlé`.

### DIFF
--- a/src/main/java/net/sf/jabref/logic/groups/GroupTreeNode.java
+++ b/src/main/java/net/sf/jabref/logic/groups/GroupTreeNode.java
@@ -213,4 +213,20 @@ public class GroupTreeNode extends TreeNode<GroupTreeNode> {
     public GroupTreeNode copyNode() {
         return new GroupTreeNode(group);
     }
+
+    /**
+     * Determines the number of entries in the specified list which are matched by this group.
+     * @param entries list of entries to be searched
+     * @return number of hits
+     */
+    public int numberOfHits(List<BibEntry> entries) {
+        int hits = 0;
+        SearchMatcher matcher = getSearchRule();
+        for (BibEntry entry : entries) {
+            if (matcher.isMatch(entry)) {
+                hits++;
+            }
+        }
+        return hits;
+    }
 }
--- a/src/test/java/net/sf/jabref/logic/groups/GroupTreeNodeTest.java
+++ b/src/test/java/net/sf/jabref/logic/groups/GroupTreeNodeTest.java
@@ -1,17 +1,30 @@
 package net.sf.jabref.logic.groups;
 
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 
 import net.sf.jabref.importer.fileformat.ParseException;
 import net.sf.jabref.logic.search.matchers.AndMatcher;
 import net.sf.jabref.logic.search.matchers.OrMatcher;
+import net.sf.jabref.model.entry.BibEntry;
 
+import org.junit.Before;
 import org.junit.Test;
 
 import static org.junit.Assert.assertEquals;
 
 public class GroupTreeNodeTest {
+
+    private List<BibEntry> entries = new ArrayList<>();
+
+    @Before
+    public void setUp() throws Exception {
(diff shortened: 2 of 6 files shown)


koppor:
LGTM :+1:

src/test/java/net/sf/jabref/logic/groups/GroupTreeNodeTest.java
```diff
@@ -174,4 +187,47 @@ public void getSearchRuleForIncludingGroupReturnsGroupOrSubgroupAsMatcher() thro
...
+
+    @Test
+    public void numberOfHitsMatchesMultipleEntries() throws Exception {
+        GroupTreeNode parent = getNodeInSimpleTree();
+        GroupTreeNode node = parent.addSubgroup(
+                new KeywordGroup("node", "author", "author1", true, false, GroupHierarchyType.INDEPENDENT));
+        assertEquals(2, node.numberOfHits(entries));
+    }
+
+    // Test for #1482
```
stefan-kolb: Unnecessary comment in my opinion. Doesn't matter for which issue this test is. The test covers the old problem.

src/test/java/net/sf/jabref/logic/groups/GroupTreeNodeTest.java
```diff
@@ -174,4 +187,47 @@ public void getSearchRuleForIncludingGroupReturnsGroupOrSubgroupAsMatcher() thro
...
+    public void numberOfHitsWorksForRefiningGroups() throws Exception {
+        GroupTreeNode grandParent = getNodeInSimpleTree();
+        GroupTreeNode parent = grandParent.addSubgroup(
+                new KeywordGroup("node", "author", "author2", true, false, GroupHierarchyType.INDEPENDENT));
+        GroupTreeNode node = parent.addSubgroup(
+                new KeywordGroup("node", "author", "author1", true, false, GroupHierarchyType.REFINING));
+        assertEquals(1, node.numberOfHits(entries));
+    }
+
+    // Test for #1482
```
stefan-kolb: See above.

stefan-kolb:
LGTM :+1:

src/main/java/net/sf/jabref/logic/groups/GroupTreeNode.java
```diff
@@ -213,4 +213,20 @@ public String toString() {
...
+    public int numberOfHits(List<BibEntry> entries) {
+        int hits = 0;
+        SearchMatcher matcher = getSearchRule();
+        for (BibEntry entry : entries) {
+            if (matcher.isMatch(entry)) {
+                hits++;
+            }
+        }
+        return hits;
+    }
```
simonharrer: Why is this moved? I think it fits more into the group itself which carries around the search rule.

tobiasdiez: The (Abstract)Group does not know where it is sitting in the group tree but the GroupTreeNode does. But the position in the tree might influence the search result (since we support refining / independent / ... search modes)

simonharrer: Ah  ok, the getSearchRule is automatically adjusting to the position in the group tree? Then this make sense.

simonharrer:
LGTM

stefan-kolb:
Ok, let's merge this in :smile: :+1:

PR #1491, opened 2016-06-10
Clean registry on uninstall

stefan-kolb:
Will delete the registry keys of JabRef on uninstall.
I'm not sure if we save anything that should permanently stay in the registry, custom entry types etc? Currently this reg key only hosts the Path word for me, which displays the installation location of JabRef.

### DIFF
--- a/jabref.install4j
+++ b/jabref.install4j
@@ -117,14 +117,10 @@
             <preActivation />
             <postActivation />
             <actions>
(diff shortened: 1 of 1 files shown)


koppor:
Is this deletion optional? Some tools have "clean up thoroughly". Maybe, some users don't want to have their settings to be destroyed. OK, they can export their settings. Maybe, we should display a warning that also their settings are lost when uninstalling.

PR #1511, opened 2016-06-22
Fixes #1507: Keywords are now separated by the delimiter specified in the preferences

### LINKED ISSUE #1507: Keywords management works only with the comma as a delimiter
I am running V3.3 under windows 8 on an I5 computer.
For convenience and practical reasons I am using the semicolon (;) as a delimiter in the keywords list.
I specified it in the options/preferences/groups menu.
When I use the Keywords Management Tools, keywords are modified as instructed but the resulting keywords lists are re-compiled with a comma (,) instead of the semicolon (;).
Additional query : I feel strange to be prompted to specify the keywords delimiter every time I generate a group and to find the comma as a default value there. As it has been input in the options/etc., it should be "remembered" by the software...
Regards

tobiasdiez:
Also did some small refactorings / renamings.


src/main/java/net/sf/jabref/model/entry/BibEntry.java
```diff
@@ -539,26 +553,23 @@ public void putKeywords(List<String> keywords) {
...
+                // Duplicate so don't add it
+                return;
             }
         }
 
-        if (!duplicate) {
-            keywords.add(keyword);
-            this.putKeywords(keywords);
-        }
+        keywords.add(keyword);
```
Siedlerchr: Why don't you put that in the loop and simply do  
! keyword.equals.... ?

Siedlerchr:
LGTM :+1:

matthiasgeiger:
Some tests are failing and it would be great if you could push your branch to the jabref repo as this facilitates testing by simply running a build from http://builds.jabref.org/

src/main/java/net/sf/jabref/model/entry/BibEntry.java
Siedlerchr: If you don't want keyword multiple times, then simply use a Set, which by default does not allow double values.
Then you could remove the check.

tobiasdiez: Thanks for the tip. Now moved all the keyword-related stuff to sets instead of lists.

tobiasdiez:
OK fixed the failing tests and the build is available under http://builds.jabref.org/fix1507/. (Will fix the checkstyle issues after green light for merge)

src/main/java/net/sf/jabref/model/entry/BibEntry.java
koppor: Using a list was intentional to *preserve the order given by the user*. Is
that also possible by using a set? I fear that the ordering of the keywords
change when using another machine, JRE, ...

tobiasdiez: LinkedHashSet also preserves the order.

src/main/java/net/sf/jabref/importer/fileformat/JSONEntryParser.java
```diff
@@ -104,7 +105,7 @@ public BibEntry parseBibJSONtoBibtex(JSONObject bibJsonEntry) {
         // Keywords
         if (bibJsonEntry.has("keywords")) {
             JSONArray keywords = bibJsonEntry.getJSONArray("keywords");
-            List<String> keywordList = new ArrayList<>();
+            LinkedHashSet<String> keywordList = new LinkedHashSet<>();
```
simonharrer: Set instead

PR #1518, opened 2016-06-27
Added import order for javafx packages to the eclipse settings

boceckts:
The eclipse import order settings don't cover javafx packages but the file checkstyle.xml does. Therefore using "organize import" in eclipse gave wrong import order warnings at travis CI.


koppor:
LGTM, the failing checks seem to be caused by other updates.

tobiasdiez:
LGTM :+1:

PR #1524, opened 2016-06-28
Reenable integration tests

koppor:
The integration tests stopped, because 
 - a) the JRE ran out of memory. Already fixed in master
 - b) there have been exceptions within `net.sf.jabref.gui.BasePanel.GroupTreeListener` that a timer was already canceled.

The solution for b) is presented here: The timer is started in daemon mode, which ensures that it is shut down at the end of the execution.

Finally, I treated something else, the `testCreateBibtexEntry`. This test missed closeDatabase and exitJabRef, but with them in place, the test does not work. Therefore, I ignored it -- I think, ignoring is better than having the test being wrong.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,7 +48,7 @@ script:
   - sudo service resolvconf stop
   - sudo service sshguard stop
   - sudo service ssh stop
-  - ./gradlew integrationTest --info || true
+  - ./gradlew integrationTest --info
 
 after_script:
   # enable codecov report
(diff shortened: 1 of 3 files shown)


simonharrer:
I am not sure whether we can rescue these integration tests at all. Maybe we have to skip this, and add them for JavaFX, which is maybe more stable.

I do not like that the Timer is a deamon, as this complicates things unnecessarily.

koppor:
Maybe we should have an online discussion about the daemonized timer. I would like to have UI tests to discover issues like #1551 as soon as possible and not when an interested user downloads the latest build. :innocent:

koppor:
Discussed with @simonharrer 

 - Side effects of daemonized timer are production JabRef is not clear
 - The UI tests cause much trouble
 - Our GUI code is of low quality
 - We think that we will migrate to JavaFX and better UI code

Decision: Postpone until we have a better UI code

Issue #1396, opened 2016-05-13
Journal abbreviations dialog mock up for JavaFX port


boceckts:
I want to port the journal abbreviations dialog to JavaFX. When playing with the dialog I first I didn't understand what everything meant. It was quite confusing why there were 3 sections - JabRef internal list, personal list, external list. Why can I only have one personal list? Why can't I edit the external lists? Why do I need different windows to view external lists or JabRef internal list? What does the download button do since there is no hint in the dialog itself and is it really necessary to have it? Therefore I thought when porting the dialog I could also redesign it. The following dialogs mock ups concentrate all the different lists in one place and treat them as the same.

The first mock up treats all lists the same, a user can set up multiple personal journal lists or open some from this hard drive. The user can also edit any journal list (except the JabRef internal journal list). Also entries in the lists should be able to be edited instead of deleted and then added as new.

![journalabbreviationsnewnew](https://cloud.githubusercontent.com/assets/15340757/15244225/5f562998-1900-11e6-87df-c87663e69141.png)

The second one treats the JabRef internal journal list as non editable and has to be viewed through a different window. Other lists will be shown in the top list.

![journalabbreviationsoldnew](https://cloud.githubusercontent.com/assets/15340757/15244226/5f59951a-1900-11e6-95b0-a142717b4ceb.png)

boceckts:
@JabRef/developers some feedback would be appreciated.

tobiasdiez:
I really like the idea to unify the ui for all abbreviation lists, so I prefer the first mockup.
What I don't like are the buttons on the right side. In my opinion, these should be integrated into the list. 

Some suggestions:
- I would replace the first list with a drop-down list so that it is clear which journal list the user is currently editing
- Switch order of Save and Cancel button, because the default one should be "Save"
- Make the help button a bit less dominant.

So I would propose:
![Image](https://jabref.mybalsamiq.com/mockups/4555701.png)
https://jabref.mybalsamiq.com/projects/ui/Journal%20Abbreviations

Siedlerchr:
The mockup of @tobiasdiez  looks nice. One thing we should think about is wether we allow editing the Abbrevation direct in the Table or open it in a new window.

koppor:
I also like the mockup of @tobiasdiez. The "view" button of @boceckts means "show source"? Maybe we could link the GitHub repo for the internal list. The others are just plain text imported files.

[...]

PR #1526, opened 2016-06-29
Journal abbreviations dialog ported to JavaFX

boceckts:
I tried to extract the logic in a test driven approach to completely separate ui and ui-model and to have a high code test coverage. The design of the dialog was made after the ui-model was finisehd and is close to what was specified in #1396.

listview
listvieweditname
listviewbuiltinlist


src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
```diff
@@ -0,0 +1,315 @@
...
+        journalTableEditColumn
+                .setCellFactory(new Callback<TableColumn<Abbreviation, String>, TableCell<Abbreviation, String>>() {
+
+                    @Override
+                    public TableCell<Abbreviation, String> call(TableColumn<Abbreviation, String> stringListView) {
+                        return new TableCell<Abbreviation, String>() {
+
+                            @Override
+                            protected void updateItem(String s, boolean b) {
+                                super.updateItem(s, b);
```
Siedlerchr: Try using  a lambda for the callback, makes the code more readable:
http://stackoverflow.com/questions/23978676/convert-callback-into-java-8-lambda-expression

Please use more speaking  variable names for s and b, for example like here:
http://code.makery.ch/blog/javafx-8-tableview-cell-renderer/

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsViewModel.java
```diff
@@ -0,0 +1,282 @@
...
+    }
+
+    /**
+     * Read all saved file paths and read their abbreviations
+     */
+    public void createFileObjects() {
+        List<String> externalFiles = Globals.prefs.getStringList(JabRefPreferences.EXTERNAL_JOURNAL_LISTS);
+        externalFiles.forEach(name -> {
+            try {
+                openFile(new File(name));
```
Siedlerchr: Please use the nio Paths.get method

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsViewModel.java
```diff
@@ -0,0 +1,282 @@
...
+    /**
+     * Checks whether the file exists or if a new file should be opened.
+     * The file will be added to the set of journal abbreviation files.
+     * If the file also exists its abbreviations will be read and written
+     * to the abbreviations property.
+     *
+     * @param file to be opened
+     * @throws JabRefException if journal file with same path is already open
+     */
+    public void openFile(File file) throws JabRefException {
```
Siedlerchr: As mentioned above, please use nio methods

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsViewModel.java
```diff
@@ -0,0 +1,282 @@
...
+        AbbreviationsFile abbreviationsFile = new AbbreviationsFile(file);
+        if (journalFiles.contains(abbreviationsFile)) {
```
Siedlerchr: Why not simply use a Set?
No need for checking duplicates, as a Set implicitly ignores duplicates

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsViewModel.java
```diff
@@ -0,0 +1,282 @@
...
+
+    /**
+     * Method to add a new abbreviation to the abbreviations property.
+     * The name and the actual abbreviation will be taken from the abbreviationsName
+     * and abbreviationsAbbreviation properties. It also sets the currentAbbreviation
+     * property to the new abbreviation.
+     */
+    public void addAbbreviation() throws JabRefException {
+        Abbreviation abbreviation = new Abbreviation(abbreviationsName.get(), abbreviationsAbbreviation.get());
+        if (abbreviations.contains(abbreviation)) {
```
Siedlerchr: see above, try using a set

Siedlerchr:
Looks good :+1:  ! 
Instead of the "+" Button  would suggest simply the text "Add".
What is the purpose of the "-" Icon? Deletion? If yes, a kind of trash can symbol would be more suitable

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsViewModel.java
boceckts: I used the list as I wanted to have the elements in an order and I can't find a `SortedSetProperty`

tobiasdiez: https://docs.oracle.com/javase/7/docs/api/java/util/LinkedHashSet.html ?

boceckts: But there exists no Property for it which is needed in order to use the bindings.

Siedlerchr: You could try the `SortedList `
Here is a quick example I found:
http://code.makery.ch/blog/javafx-8-tableview-sorting-filtering/

src/main/java/net/sf/jabref/gui/journals/AbbreviationsFile.java
```diff
@@ -1,39 +1,43 @@
...
 
     private final SimpleListProperty<Abbreviation> abbreviations = new SimpleListProperty<>(
             FXCollections.observableArrayList());
+    private final Path path;
 
 
-    public AbbreviationsFile(File file) {
-        super(file.getAbsolutePath());
+    public AbbreviationsFile(String filePath) {
+        path = FileSystems.getDefault().getPath(filePath.replaceAll("\\\\", "/"));
```
Siedlerchr: You could simply use `Paths.get(filePath)` this is a shorthand for the FileSystems.getDefault...

What are you trying to do here? I do not understand the Replace stuff...  In case you are trying to escape sth like Slashes, that is absolutely not necessary, as the Path methods are able to handle both relative and absolute.
Have a look here for some examples
https://docs.oracle.com/javase/tutorial/essential/io/pathOps.html

src/main/java/net/sf/jabref/gui/journals/AbbreviationsFile.java
```diff
@@ -43,4 +47,35 @@ public void WriteOrCreate() throws FileNotFoundException, IOException {
...
+
+    @Override
+    public int hashCode() {
+        return new HashCodeBuilder().append(path).toHashCode();
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (obj instanceof AbbreviationsFile) {
+            return this.path.equals(((AbbreviationsFile) obj).getPath());
```
Siedlerchr: Simpler: `Objects.equals(Object,Object) `

src/main/java/net/sf/jabref/gui/journals/AbbreviationsFile.java
```diff
@@ -43,4 +47,35 @@ public void WriteOrCreate() throws FileNotFoundException, IOException {
...
+    }
+
+    @Override
+    public String toString() {
+        return path.toString();
+    }
+
+    @Override
+    public int hashCode() {
+        return new HashCodeBuilder().append(path).toHashCode();
```
Siedlerchr: Simpler:
`return Objects.hash(...)`

src/main/java/net/sf/jabref/gui/journals/AbbreviationsFile.java
boceckts: The idea was indeed to escape the backslashes for windows systems...but thanks for the tip

boceckts:
I updated the description and added screenshots for the latest commit. Is it really the best idea to add the "+" to the end of the list of journal abbrevaitions? This would mean the user always has to scroll to the bottom to add more journal abbreviations to the list. In my opinion this is less ideal than having a button that is always visible and in the same place. Any thoughts?

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsViewModel.java
```diff
@@ -0,0 +1,308 @@
...
+import net.sf.jabref.Globals;
+import net.sf.jabref.JabRefException;
+import net.sf.jabref.JabRefGUI;
+import net.sf.jabref.JabRefPreferences;
+import net.sf.jabref.gui.BasePanel;
+import net.sf.jabref.logic.journals.Abbreviation;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
```
tschechlovdev: Could you add for all classes, that you've added, a short comment?

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsViewModel.java
```diff
@@ -0,0 +1,308 @@
...
+        List<Abbreviation> ieeeList;
+        if (Globals.prefs.getBoolean(JabRefPreferences.USE_IEEE_ABRV)) {
+            ieeeList = Globals.journalAbbreviationLoader.getOfficialIEEEAbbreviations();
+        } else {
+            ieeeList = Globals.journalAbbreviationLoader.getStandardIEEEAbbreviations();
+        }
+        AbbreviationsFile ieeeFile = new AbbreviationsFile(ieeeList, "IEEE built in list");
+        journalFiles.add(ieeeFile);
+    }
+
```
tschechlovdev: please add short comment

src/test/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsTest.java
```diff
@@ -0,0 +1,554 @@
...
+
+    @Test
+    public void testAddFile() {
+        int size = viewModel.journalFilesProperty().size();
+
+        // simulate add new file button
+        savelyAddNewFileToViewModel("newTestFile.txt");
+
+        // size of the list of journal files should be incremented by one
+        Assert.assertEquals(size + 1, viewModel.journalFilesProperty().size());
```
tschechlovdev: could you just use a explicity value instead of size +1?

src/test/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsTest.java
```diff
@@ -0,0 +1,554 @@
...
+        Assert.assertEquals(size, viewModel.abbreviationsProperty().size());
+        Assert.assertTrue(viewModel.abbreviationsProperty().contains(testAbbreviation1));
+
+        viewModel.saveJournalAbbreviationFiles();
+
+        String expected = "YetAnotherEntry = YAE" + Globals.NEWLINE + "Test Entry = TE" + Globals.NEWLINE
+                + "MoreEntries = ME" + Globals.NEWLINE + "Entry = E" + Globals.NEWLINE + "";
+        String actual = Files.contentOf(new File(testFile4Entries), StandardCharsets.UTF_8);
+        Assert.assertEquals(expected, actual);
+
```
tschechlovdev: could you in general try to split up the test methods into more test methods?

tobiasdiez:
Personally, I prefer inline editing for lists/tables with only a handful of columns. See for example http://designingwebinterfaces.com/ultimate-guide-to-table-ui-patterns. The very first row could be an "insert row", i.e. it shows (in italic) "Create new abbreviation" and when you click you just have an empty textfield to insert the new abbreviation. 

Other from that, I really like how the dialog looks. Maybe change the close/remove icon? Btw what is the purpose of this "JabRef Lists" button at the bottom?

boceckts:
@tobiasdiez so when clicking on the edit icon on the right the column should be editable?
The "JabRef Lists" button shows the jabref built in journal abbreviations which are not modifyable. In the swing version there is a "view" button to show these abbreviations in a separate window.

tobiasdiez:
@boceckts Yes exactly, clicking edit (and I would also support double clicking) makes the row editable. Mhh on second thought, one might not even need the edit button but show it directly in the cell upon mouse hover and a single click then makes the cell editable. Decide based on what is easier to implement. 

For the "JabRef List" button, you also have a JabRef built in list in the drop-down. What is the difference here?

boceckts:
@tobiasdiez The built in lists in the drop down will only appear when the toggle button is activated

tobiasdiez:
Is there a need to hide it from the drop-down list? If not, then I would just remove the toggle button.

boceckts:
@tobiasdiez I just thought it's nice to give the user  the option to only display their custom lists

koppor:
@boceckts Please remove "JabRef lists" button. :sunflower:

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
```diff
@@ -0,0 +1,546 @@
...
+        journalTableAbbreviationColumn.setCellValueFactory(cellData -> cellData.getValue().abbreviationProperty());
+        journalTableAbbreviationColumn.setCellFactory(cell -> new JournalAbbreviationsAbbreviationTableEditingCell());
+        journalTableEditColumn.setCellValueFactory(cellData -> cellData.getValue().isPseudoAbbreviationProperty());
+        journalTableDeleteColumn.setCellValueFactory(cellData -> cellData.getValue().isPseudoAbbreviationProperty());
+        journalTableEditColumn.setCellFactory(column -> new TableCell<AbbreviationViewModel, Boolean>() {
+
+            @Override
+            protected void updateItem(Boolean isPseudoAbbreviation, boolean isEmpty) {
+                super.updateItem(isPseudoAbbreviation, isEmpty);
+                if (isPseudoAbbreviation != null) {
```
tschechlovdev: is this really needed ? When can this be null?

src/main/java/net/sf/jabref/gui/journals/AbbreviationsFile.java
```diff
@@ -0,0 +1,118 @@
...
+
+import net.sf.jabref.Globals;
+import net.sf.jabref.logic.journals.Abbreviation;
+import net.sf.jabref.logic.journals.JournalAbbreviationLoader;
+
+/**
+ * This class provides a model for abbreviation files.
+ * It actually doesn't save the files as objects but rather saves
+ * their paths. This also allows to specify pseudo files as placeholder objects.
+ *
```
tschechlovdev: remove this unnecessary empty lines in the comments.

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
```diff
@@ -0,0 +1,546 @@
...
+                }
+            }
+        });
+        journalTableDeleteColumn.setCellFactory(column -> new TableCell<AbbreviationViewModel, Boolean>() {
+
+            @Override
+            protected void updateItem(Boolean isPseudoAbbreviation, boolean isEmpty) {
+                super.updateItem(isPseudoAbbreviation, isEmpty);
+                if (isPseudoAbbreviation != null) {
+
```
tschechlovdev: remove empty line

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
```diff
@@ -0,0 +1,546 @@
...
+                    if (!isEmpty) {
+                        if (isEditableAndRemovable.get()) {
+                            if (!isPseudoAbbreviation) {
```
tschechlovdev: why don't put the if-statements together with and? Then you have only one time ``else setGraphic(null)``

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsViewModel.java
```diff
@@ -0,0 +1,342 @@
...
+        } else {
+            abbreviations.add(abbreviationViewModel);
+            currentAbbreviation.set(abbreviationViewModel);
+        }
+    }
+
+    /**
+     * Method to change the currentAbbrevaition property to a new abbreviation.
+     * The name and the actual abbreviation will be taken from the abbreviationsName
+     * and abbreviationsAbbreviation properties.
```
tschechlovdev: add an empty line here.

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsViewModel.java
```diff
@@ -0,0 +1,342 @@
...
+     * @throws JabRefException
+     */
+    private void setCurrentAbbreviationNameAndAbbreviationIfValid() throws JabRefException {
+        String name = abbreviationsName.get();
+        String abbreviation = abbreviationsAbbreviation.get();
+        if (name.trim().isEmpty()) {
+            throw new JabRefException("Name can not be empty");
+        } else if (abbreviation.trim().isEmpty()) {
+            throw new JabRefException("Abbrevaition can not be empty");
+        } else {
```
tschechlovdev: else shouldn't be necessary here

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsViewModel.java
```diff
@@ -0,0 +1,342 @@
...
+     * the content of the abbreviations property of the AbbriviationsFile. Non
+     * existing files will be created.
+     */
+    public void saveJournalAbbreviationFiles() {
+        journalFiles.forEach(file -> {
+            try {
+                file.WriteOrCreate();
+            } catch (FileNotFoundException e) {
+                logger.debug(e.getLocalizedMessage());
+            } catch (IOException ioe) {
```
tschechlovdev: you can put the catch blocks together ``FileNotFoundException | IOException e``

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
boceckts: But I only need to call `setGraphic(null)` on two of the four if statements. However I can join the two if statements with no else block.

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsViewModel.java
boceckts: I found out `FileNotFoundException` is even redundant if I catch the `IOException` first.

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
boceckts: Yes it is necessary because we have to call `setGraphic(null)` if the item to update is null. Otherwise we get some weird graphic bugs when switching from a custom list to a built in list since the graphics don't get reset.

journalabb

src/test/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsTest.java
```diff
@@ -0,0 +1,614 @@
...
+     * This is useful for testing purposes so you don't have to always handle the exceptions.
+     *
+     * @param testAbbreviation which will be edited
+     */
+    private void savelyEditAbbreviation(Abbreviation testAbbreviation) {
+        try {
+            editAbbreviation(testAbbreviation);
+        } catch (JabRefException e1) {
+            Assert.fail("The same abbreviation already exists in the current list.");
+        }
```
simonharrer: do not catch exception in tests, just throw them. In this case, you hide the exception from JUnit, and the user does not get the stack trace, only your short text.

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ to [sourceforge feature requests](https://sourceforge.net/p/jabref/features/) by
 - Integrity check table can be sorted by clicking on column headings
 - Add tab which shows the MathSciNet review website if the `MRNumber` field is present.
 - Partly switched to new UI technology (JavaFX). Redesigned key bindings dialog.
+- Partly switched to new UI technology (JavaFX). Redesigned journal abbreviations dialog.
```
tobiasdiez: Make the different redesigned dialogs a subbulled of "Partly switched to ..."

src/main/java/net/sf/jabref/gui/journals/AbbreviationViewModel.java
```diff
@@ -0,0 +1,100 @@
...
+
+    public Abbreviation getAbbreviationObject() {
+        return this.abbreviationObject;
+    }
+
+    public void setName(String name) {
+        this.name.set(name);
+    }
+
+    public void setAbbreviation(String name) {
```
tobiasdiez: rename parameter "name"

src/main/java/net/sf/jabref/gui/journals/AbbreviationsFile.java
```diff
@@ -0,0 +1,132 @@
...
+import net.sf.jabref.Globals;
+import net.sf.jabref.logic.journals.Abbreviation;
+import net.sf.jabref.logic.journals.JournalAbbreviationLoader;
+
+/**
+ * This class provides a model for abbreviation files.
+ * It actually doesn't save the files as objects but rather saves
+ * their paths. This also allows to specify pseudo files as placeholder objects.
+ */
+public class AbbreviationsFile {
```
tobiasdiez: add suffix "ViewModel"

src/main/java/net/sf/jabref/gui/journals/AbbreviationsFile.java
```diff
@@ -0,0 +1,132 @@
...
+
+    /**
+     * This constructor should only be called to create a pseudo abbreviation file for built in lists.
+     * This means it is a placeholder and it's path will be null meaning it has no place on the filesystem.
+     * It's isPseudoFile property will therefore be set to true.
+     */
+    public AbbreviationsFile(List<AbbreviationViewModel> abbreviations, String name) {
+        this.abbreviations.addAll(abbreviations);
+        this.name = name;
+        this.path = null;
```
tobiasdiez: Make path an Optional<Path>

src/main/java/net/sf/jabref/gui/journals/AbbreviationsFile.java
```diff
@@ -0,0 +1,132 @@
...
+        this.isPseudoFile = new SimpleBooleanProperty(true);
+    }
+
+    public void readAbbreviations() throws FileNotFoundException {
+        List<Abbreviation> abbreviationList = JournalAbbreviationLoader.readJournalListFromFile(path.toFile());
```
tobiasdiez: A NPE is thrown if this file is a built-in list.

src/main/java/net/sf/jabref/gui/journals/AbbreviationsFile.java
```diff
@@ -0,0 +1,132 @@
...
+/**
+ * This class provides a model for abbreviation files.
+ * It actually doesn't save the files as objects but rather saves
+ * their paths. This also allows to specify pseudo files as placeholder objects.
+ */
+public class AbbreviationsFile {
+
+    private final SimpleListProperty<AbbreviationViewModel> abbreviations = new SimpleListProperty<>(
+            FXCollections.observableArrayList());
+    private final ReadOnlyBooleanProperty isPseudoFile;
```
tobiasdiez: Maybe rename isPseudoFile -> isBuiltinList to better reflect what the actual purpose of this field is.

src/main/java/net/sf/jabref/gui/journals/AbbreviationsFile.java
```diff
@@ -0,0 +1,132 @@
...
+        abbreviationList.forEach(abbreviation -> abbreviations.addAll(new AbbreviationViewModel(abbreviation)));
+    }
+
+    /**
+     * This method will write all abbreviations of this abbreviation file to the file on the file system.
+     * If the file exists its content will be overridden, otherwise a new file will be created.
+     *
+     * @throws IOException
+     */
+    public void WriteOrCreate() throws IOException {
```
tobiasdiez: This logic contains code which I would declare as "logic", namely how abbreviations are serialized. Please extract the code to a new method `writeAbbreviations(path, encoding?, List<Abbreviation>). This new method may life in `Abbrevation` or a new class `AbbrevationWriter`.

src/main/java/net/sf/jabref/gui/journals/AbbreviationsFile.java
```diff
@@ -0,0 +1,132 @@
...
+        if (!isPseudoFile.get()) {
+            try (OutputStream outStream = Files.newOutputStream(path);
+                    OutputStreamWriter writer = new OutputStreamWriter(outStream, Globals.prefs.getDefaultEncoding())) {
```
tobiasdiez: Do we really want the user decide about the encoding? I would propose to always use UTF-8 if this works.

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
```diff
@@ -0,0 +1,543 @@
...
+                journalFilesBox.getSelectionModel().selectLast();
+                viewModel.addBuiltInLists();
+                loadingLabel.setVisible(false);
+                progressIndicator.setVisible(false);
+            }
+        }).start();
+    }
+
+    private void setButtonStyles() {
+        journalFilesBox.setPromptText(Localization.lang("No abbreviation files loaded"));
```
tobiasdiez: This styling can't be done in FXML?

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
```diff
@@ -0,0 +1,543 @@
...
+        super();
+        bundle = Localization.getMessages();
+    }
+
+    @FXML
+    private void initialize() {
+        setUpTable();
+        setBindings();
+        setButtonStyles();
+        new Thread(new Runnable() {
```
tobiasdiez: JavaFX seems to have a new interface for background tasks: docs.oracle.com/javafx/2/threads/jfxpub-threads.htm
To be honest, I have no idea about its advantages/disadvantages. Could you pls do some research and maybe switch the code to `Task<...>`

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
```diff
@@ -0,0 +1,543 @@
...
+
+            @Override
+            public void run() {
+                loadingLabel.setVisible(true);
+                progressIndicator.setVisible(true);
+                viewModel.createFileObjects();
+                journalFilesBox.getSelectionModel().selectLast();
```
tobiasdiez: In the best case the viewModel also takes care of which item is selected (as usual through binding). Can you change this if it is not too hard.

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
```diff
@@ -0,0 +1,543 @@
...
+        ButtonBar.setButtonData(progressIndicator, ButtonData.LEFT);
+        ButtonBar.setButtonData(loadingLabel, ButtonData.LEFT);
+        ButtonBar.setButtonUniformSize(progressIndicator, false);
+        ButtonBar.setButtonUniformSize(loadingLabel, false);
+    }
+
+    private void setUpTable() {
+        journalAbbreviationsTable.setOnKeyPressed(event -> {
+            if ((event.getCode() == KeyCode.DELETE) && isEditableAndRemovable.get()) {
+                if ((viewModel.currentAbbreviationProperty().get() != null)
```
tobiasdiez: combine ifs

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
```diff
@@ -0,0 +1,543 @@
...
+import com.airhacks.afterburner.views.FXMLView;
+
+/**
+ * This class controls the user interface of the journal abbreviations dialog.
+ * The ui elements and their layout are defined in the fxml file in the resource folder.
+ */
+public class ManageJournalAbbreviationsView extends FXMLView {
+
+    private final ManageJournalAbbreviationsViewModel viewModel = new ManageJournalAbbreviationsViewModel();
+    private final BooleanProperty isEditableAndRemovable = new SimpleBooleanProperty(true);
```
tobiasdiez: Why is `isEditableAndRemovable` not part of the viewModel?

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
```diff
@@ -0,0 +1,543 @@
...
+        removeJournalAbbreviationsButton.setGraphic(removeJournalAbbreviationsButtonGraphic);
+        ButtonBar.setButtonData(progressIndicator, ButtonData.LEFT);
+        ButtonBar.setButtonData(loadingLabel, ButtonData.LEFT);
+        ButtonBar.setButtonUniformSize(progressIndicator, false);
+        ButtonBar.setButtonUniformSize(loadingLabel, false);
+    }
+
+    private void setUpTable() {
+        journalAbbreviationsTable.setOnKeyPressed(event -> {
+            if ((event.getCode() == KeyCode.DELETE) && isEditableAndRemovable.get()) {
```
tobiasdiez: Just notify the viewModel about the pressed key and let it decide what to do, i.e.
`if (event.getCode == KeyCode.DELETE) viewModel.deleteCurrentAbbreviation()` and the rest of the logic is in the viewModel.

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
```diff
@@ -0,0 +1,543 @@
...
+                }
+            }
+        });
+        journalTableNameColumn.setCellValueFactory(cellData -> cellData.getValue().nameProperty());
+        journalTableNameColumn.setCellFactory(cell -> new JournalAbbreviationsNameTableEditingCell());
+        journalTableAbbreviationColumn.setCellValueFactory(cellData -> cellData.getValue().abbreviationProperty());
+        journalTableAbbreviationColumn.setCellFactory(cell -> new JournalAbbreviationsAbbreviationTableEditingCell());
+        journalTableEditColumn.setCellValueFactory(cellData -> cellData.getValue().isPseudoAbbreviationProperty());
+        journalTableDeleteColumn.setCellValueFactory(cellData -> cellData.getValue().isPseudoAbbreviationProperty());
+        journalTableEditColumn.setCellFactory(column -> new TableCell<AbbreviationViewModel, Boolean>() {
```
tobiasdiez: I really don't like these complicated cell factories...stupid javafx (not your fault and nothing to do 😺 )

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
```diff
@@ -0,0 +1,543 @@
...
+
+            @Override
+            protected void updateItem(Boolean isPseudoAbbreviation, boolean isEmpty) {
+                super.updateItem(isPseudoAbbreviation, isEmpty);
+                if (isPseudoAbbreviation != null) {
+                    if (!isEmpty) {
+                        if (isEditableAndRemovable.get()) {
+                            if (!isPseudoAbbreviation) {
```
tobiasdiez: Remove negation from if statement and switch if and else blocks

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
```diff
@@ -0,0 +1,543 @@
...
+                                Text graphic = new Text(IconTheme.JabRefIcon.EDIT.getCode());
+                                graphic.getStyleClass().add("icon");
+                                setGraphic(graphic);
+                                setOnMouseClicked(evt -> {
```
tobiasdiez: the braces are not necessary for one-line statements, i.e. `setOnMouseClicked(evt -> editAbbrevation)` is sufficient.

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
```diff
@@ -0,0 +1,543 @@
...
+                                    editAbbreviation();
+                                });
+                            } else {
+                                Text graphic = new Text(IconTheme.JabRefIcon.ADD.getCode());
```
tobiasdiez: Is it easy to change the icon class a bit so that you can write
`setGraphic(IconTheme.JabRefIcon.ADD.getGraphic())`?

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
```diff
@@ -0,0 +1,543 @@
...
+            journalFilesBox.getSelectionModel().select(newvalue);
+            if (newvalue != null) {
+                isEditableAndRemovable.set(!newvalue.isPseudoFileProperty().get());
+            }
+            removeJournalAbbreviationsButton.setDisable((newvalue == null) || !isEditableAndRemovable.get());
+        });
+        viewModel.currentAbbreviationProperty().addListener((observable, oldvalue, newvalue) -> {
+            journalAbbreviationsTable.getSelectionModel().select(newvalue);
+        });
+        journalAbbreviationsTable.getSelectionModel().selectedItemProperty()
```
tobiasdiez: a bidirectional binding between `journalAbbreviationsTable.getSelectionModel().selectedItemProperty()` and `viewModel.currentFileProperty()` does not work?

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
```diff
@@ -0,0 +1,543 @@
...
+        });
+    }
+
+    private void setBindings() {
+        viewModel.currentFileProperty().addListener((observable, oldvalue, newvalue) -> {
+            journalFilesBox.getSelectionModel().select(newvalue);
+            if (newvalue != null) {
+                isEditableAndRemovable.set(!newvalue.isPseudoFileProperty().get());
+            }
+            removeJournalAbbreviationsButton.setDisable((newvalue == null) || !isEditableAndRemovable.get());
```
tobiasdiez: I would prefer if the disable property is bound to something in the viewmodel

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
```diff
@@ -0,0 +1,543 @@
...
+        journalAbbreviationsDialog.setDialogPane((DialogPane) this.getView());
+        ((Stage) this.getView().getScene().getWindow()).setMinHeight(400);
+        ((Stage) this.getView().getScene().getWindow()).setMinWidth(600);
+        journalAbbreviationsDialog.showAndWait();
+    }
+
+    @FXML
+    private void addNewFile() {
+        FileChooser chooser = new FileChooser();
+        chooser.getExtensionFilters().add(new ExtensionFilter("TXT files (*.txt)", "*.txt"));
```
tobiasdiez: Localize "files", i.e. Localization.lang("%0 file", "TXT") + "(*.txt)"

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
```diff
@@ -0,0 +1,543 @@
...
+
+    @FXML
+    private void editAbbreviation() {
+        journalAbbreviationsTable.edit(journalAbbreviationsTable.getSelectionModel().getSelectedIndex(),
+                journalTableNameColumn);
+    }
+
+    private void selectNewAbbreviation() {
+        int lastRow = viewModel.abbreviationsCountProperty().get() - 1;
+        journalAbbreviationsTable.scrollTo(lastRow);
```
tobiasdiez: Again, in the best case all of this is handled by binding to the viewModel. (if it is not too difficult)

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
```diff
@@ -0,0 +1,543 @@
...
+    }
+
+    @FXML
+    private void fileChanged() {
+        viewModel.changeActiveFile(journalFilesBox.getSelectionModel().getSelectedItem());
+    }
+
+    @FXML
+    private void addAbbreviation() {
+        viewModel.abbreviationsNameProperty().set("Name");
```
tobiasdiez: Why do you set name and abbreviation here?

tobiasdiez: Can these two set-lines also be moved to `viewModel.addAbbreviation`? Also localize the strings.

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
```diff
@@ -0,0 +1,543 @@
...
+    @FXML
+    private void openFile() {
+        FileChooser chooser = new FileChooser();
+        chooser.getExtensionFilters().add(new ExtensionFilter("TXT files (*.txt)", "*.txt"));
+        File file = chooser.showOpenDialog(null);
+        if (file != null) {
+            try {
+                viewModel.openFile(file.getAbsolutePath());
+                journalFilesBox.getSelectionModel().selectLast();
+            } catch (JabRefException e) {
```
tobiasdiez: I don't like that a generic `JabRefException` is thrown.

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
```diff
@@ -0,0 +1,543 @@
...
+
+    @FXML
+    private void closeDialog() {
+        Stage stage = (Stage) cancelButton.getScene().getWindow();
+        stage.close();
+    }
+
+    @FXML
+    private void saveAbbreviations() {
+        viewModel.saveExternalFilesList();
```
tobiasdiez: combine this into viewModel.saveEverthing() (or something like that) ?

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
```diff
@@ -0,0 +1,543 @@
...
+            }
+        }
+
+        private void createTextField() {
+            textField = new TextField(getString());
+            textField.setMinWidth(this.getWidth() - (this.getGraphicTextGap() * 2));
+            textField.focusedProperty().addListener(new ChangeListener<Boolean>() {
+
+                @Override
+                public void changed(ObservableValue<? extends Boolean> arg0, Boolean arg1, Boolean arg2) {
```
tobiasdiez: better names for arguments

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
```diff
@@ -0,0 +1,543 @@
...
```
tobiasdiez: here too

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsViewModel.java
```diff
@@ -0,0 +1,339 @@
...
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+/**
+ * This class provides a model for managing journal abbreviation lists.
+ * It provides all necessary methods to create, modify or delete journal
+ * abbreviations and files. To visualize the model one can bind the properties
+ * to ui elements.
+ */
+public class ManageJournalAbbreviationsViewModel {
```
tobiasdiez: I stop here for now... (just a reminder for myself)

CHANGELOG.md
boceckts: I now moved them to subbullets and to ## JavaFX Changes

src/main/java/net/sf/jabref/gui/journals/AbbreviationViewModel.java
boceckts: done

src/main/java/net/sf/jabref/gui/journals/AbbreviationsFile.java
boceckts: done

src/main/java/net/sf/jabref/gui/journals/AbbreviationsFile.java
boceckts: done

src/main/java/net/sf/jabref/gui/journals/AbbreviationsFile.java
boceckts: Now caught by `if (path.isPresent())` and if false, an appropriate `FileNotFoundException` is thrown

src/main/java/net/sf/jabref/gui/journals/AbbreviationsFile.java
boceckts: done

src/main/java/net/sf/jabref/gui/journals/AbbreviationsFile.java
boceckts: Extracted to new class AbbreviationWriter in logic package

src/main/java/net/sf/jabref/gui/journals/AbbreviationsFile.java
boceckts: Changed to `StandardCharset.UTF_8`

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
boceckts: Hm I must have missed that somehow, yes it can (and now will) be done by fxml

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
boceckts: I extracted the methode to the viewmodel, the currently selected file is already bound.

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
boceckts: `Task` seems like a similar class as `Runnable`, while `Service` is the JavaFX equivalent to a `Thread` it seems. The JavaFX classes support continuous feedback of the state of the background task and work with data binding to show a progress bar for the current status. `Service` can only be run from the JavaFX Application `Thread`. Unlike Threads they can also be restarted multiple times and you can add event handlers if the execution was successfull/stopped/unsuccessfull/etc.

In this particular scenario it shouldn't matter though. For the moment I changed it to a `Task` running in a `Thread` but `Thread` and `Service` are easily interchangable if needed.

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
boceckts: done

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
boceckts: Good point I will try to move it

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
boceckts: done

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
boceckts: done

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
boceckts: done in all cases

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
boceckts: Created a new method in `JabRefIcon` class

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
boceckts: I will look into this

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
boceckts: I think the disable property is readonly for ui elements, therefore it can not be bound to other properties.

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
boceckts: Changed it, but the console says:
>  could not get message translation for "%0 file" for locale en

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
boceckts: The selection is already bound to the view model, unfortunately the table view doesn't automatically scroll to the selected row...maybe the `scrollTo()` call can be moved to the property listener of the current abbreviation property to always scroll if the current abbreviation changes.

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
boceckts: Yes I will change it tomorrow, the test also have to be adjusted then.

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
boceckts: I introduced three new Exception extending JabRefException in net.sf.jabref.logic.journals:
`DuplicatedJournalAbbreviationException`, `DuplicatedJournalFileException` and `EmptyFieldException`

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
boceckts: changed it

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
boceckts: done

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
boceckts: also done

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
tobiasdiez: Thanks for doing the research. Can you please add the first paragraph to the Wiki?

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
tobiasdiez: Should work after master is merged into JavaFX.

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
```diff
@@ -0,0 +1,527 @@
...
+        journalAbbreviationsDialog.setDialogPane((DialogPane) this.getView());
+        ((Stage) this.getView().getScene().getWindow()).setMinHeight(400);
+        ((Stage) this.getView().getScene().getWindow()).setMinWidth(600);
+        journalAbbreviationsDialog.showAndWait();
+    }
+
+    @FXML
+    private void addNewFile() {
+        FileChooser chooser = new FileChooser();
+        chooser.getExtensionFilters().add(new ExtensionFilter(Localization.lang("%0 file", "TXT"), "*.txt"));
```
Siedlerchr: Have a look at the new FileExtensions Enum I created in #1336. Already inlcudes the Loalization.

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
boceckts: Just realised there are two properties, disbaledProperty is read only the other one can be bound

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
boceckts: Moving the call of `scrollTo()` to the current abbreviation property listener means the view will always scroll so the selected row is the first row in the visible view of rows. This means if if I select a row in the middle of the view it will still scroll so the selected row is the first row. Therefore I would rather leave it like this.

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
boceckts: `journalAbbreviationsTable.getSelectionModel().selectedItemProperty()` is read only and can therefore not be bound. The `journalAbbreviationsTable.itemsProperty()` is a `ObjectProperty<ObservableList<AbbreviationViewModel>>` which unfortunately can not be bound to my `SimpleListProperty<AbbreviationsFileViewModel>.

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
boceckts: Should I add it as a new chapter in the Code HowTos as something like "Working with Threads"?

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
tobiasdiez: ok that is fine with me

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.java
tobiasdiez: According to http://stackoverflow.com/questions/32349660/how-to-bidirectionally-bind-to-a-property-of-an-object-set-to-the-selected-item there should be a way to bind it using https://github.com/TomasMikula/EasyBind. This library seems really nice and thus I'm in favor of integrating it (if the stackoverflow answer is correct and it indeed works).

boceckts: Ignore my comment above, the lists can actually be bound without an extra library as I just found out when playing around with the EasyBind Library... :expressionless:
But I don't think `journalAbbreviationsTable.getSelectionModel().selectedItemProperty().addListener(...`  and similar can be replaced.

src/main/java/net/sf/jabref/gui/journals/AbbreviationsFileViewModel.java
```diff
@@ -0,0 +1,136 @@
...
+     * This method will write all abbreviations of this abbreviation file to the file on the file system.
+     * It essentially will check if the current file is a built in list and if not it will call
+     * {@link AbbreviationWriter#writeOrCreate()}.
+     *
+     * @throws IOException
+     */
+    public void writeOrCreate() throws IOException {
+        if (!isBuiltInList.get()) {
+            List<Abbreviation> actualAbbreviations = Lists.newArrayList();
+            abbreviations.forEach(abb -> {
```
tobiasdiez: I think the stream interface is more readable here, i.e. something along the lines `abbreviations.stream().filter( !isPseudoAbbreviation).map( getAbbreviationObject).collect()`.

src/main/java/net/sf/jabref/gui/journals/AbbreviationsFileViewModel.java
```diff
@@ -0,0 +1,136 @@
...
+    public SimpleListProperty<AbbreviationViewModel> abbreviationsProperty() {
+        return this.abbreviations;
+    }
+
+    public boolean exists() {
+        return path.isPresent() && Files.exists(path.get());
+    }
+
+    public String getAbsolutePath() {
+        return path.get().toAbsolutePath().toString();
```
tobiasdiez: Be careful. The path may be an empty optional. I would propose to change the return value to Optional<String> and write `return path.map(toAbsolutePath().toString());`

src/main/resources/l10n/JabRef_da.properties
```diff
@@ -1667,6 +1667,7 @@ Message=
 
 MathSciNet_Review=
 Reset_Bindings=
+<<<<<<< acdb36f664504aed9b5db5a173094ba075ee6c18
```
tobiasdiez: Remove this merge marker

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsViewModel.java
```diff
@@ -0,0 +1,356 @@
...
+     * to the global JabRef preferences. Pseudo abbreviation files will not be stored.
+     */
+    public void saveExternalFilesList() {
+        List<String> extFiles = new ArrayList<>();
+        journalFiles.forEach(file -> {
+            if (!file.isBuiltInListProperty().get()) {
+                extFiles.add(file.getAbsolutePath());
+            }
+        });
+        Globals.prefs.putStringList(JabRefPreferences.EXTERNAL_JOURNAL_LISTS, extFiles);
```
tobiasdiez: We are trying to reduce the occurrences of `Globals.prefs`. Thus I would like to see that the view model is independent on `Globals.prefs`. At least inject a JabRefPrefences class as a dependency and work on this class. Probably better it is to just work on an injected `JournalAbbreviationPreferences`. @oscargus does it make sense to introduce a field in `JournalAbbreviationPreferences` corresponding to `EXTERNAL_JOURNAL_LISTS`?

src/main/java/net/sf/jabref/logic/journals/Abbreviation.java
```diff
@@ -1,29 +1,66 @@
...
 
-    private final String name;
-    private final String abbreviation;
+    private final SimpleStringProperty name = new SimpleStringProperty("");
+    private final SimpleStringProperty abbreviation = new SimpleStringProperty("");
 
     public Abbreviation(String name, String abbreviation) {
-        this.name = java.util.Objects.requireNonNull(name).trim();
-        this.abbreviation = java.util.Objects.requireNonNull(abbreviation).trim();
+        this.name.set(java.util.Objects.requireNonNull(name).trim());
```
tobiasdiez: Remove the `java.util` part by using the proper import.

src/test/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsTest.java
```diff
@@ -0,0 +1,560 @@
...
+    }
+
+    private String createTemporaryTestFile(String name, String content) {
+        File testFile = null;
+        try {
+            testFile = tempFolder.newFile(name);
+            try (OutputStream outputStream = new FileOutputStream(testFile)) {
+                outputStream.write(content.getBytes(StandardCharsets.UTF_8));
+            }
+        } catch (IOException e) {
```
tobiasdiez: Don't catch this exception here. If something goes wrong with the initialization, the tests should better fail.

src/test/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsTest.java
```diff
@@ -0,0 +1,560 @@
...
+        viewModel.addAbbreviation();
+    }
+
+    private void editAbbreviation(Abbreviation testAbbreviation) throws JabRefException {
+        viewModel.abbreviationsNameProperty().set(testAbbreviation.getName());
+        viewModel.abbreviationsAbbreviationProperty().set(testAbbreviation.getAbbreviation());
+        viewModel.editAbbreviation();
+    }
+
+    private void saveFilesToPreferences() throws DuplicatedJournalFileException {
```
tobiasdiez: probably a better name: `addFilesToViewModel`

src/test/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsTest.java
```diff
@@ -0,0 +1,560 @@
...
+        // check whether all external files from the preferences will
+        // be read when creating the view model
+        Assert.assertEquals(4, viewModel.journalFilesProperty().size());
+        viewModel.changeActiveFile(viewModel.journalFilesProperty().get(3));
+        // remove duplicated entry
+        Assert.assertEquals(5, viewModel.abbreviationsProperty().size());
+    }
+
+    @Test
+    public void testAddFile() throws DuplicatedJournalFileException {
```
tobiasdiez: please rename the tests so that they reflect what you are testing. here `addFileIncreasesCounterOfOpenFiles`

src/test/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsTest.java
```diff
@@ -0,0 +1,560 @@
...
+        int size = viewModel.journalFilesProperty().size();
```
tobiasdiez: You are starting with an empty viewModel and so size == 0, right? In this case just remove the size variable.

src/test/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsTest.java
```diff
@@ -0,0 +1,560 @@
...
+        int size = 0;
+        Abbreviation testAbbreviation = new Abbreviation("Test Entry", "TE");
+
+        // simulate open file button
+        viewModel.addNewFile(testFile3Entries);
+        size++;
+        selectLastJournalFile();
+
+        // size of the list of journal files should be incremented by one
+        Assert.assertEquals(size, viewModel.journalFilesProperty().size());
```
tobiasdiez: size is always 1, so use hard-coded value.

src/test/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsTest.java
```diff
@@ -0,0 +1,560 @@
...
+        // our test file has 3 abbreviations
```
tobiasdiez: 4?

src/test/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsTest.java
```diff
@@ -0,0 +1,560 @@
...
+
+        // simulate open file button twice
+        viewModel.addNewFile(testFile3Entries);
+        size++;
+        viewModel.addNewFile(testFile4Entries);
+        size++;
+        viewModel.changeActiveFile(viewModel.journalFilesProperty().get(1));
+
+        // size of the list of journal files should be incremented by two
+        Assert.assertEquals(size, viewModel.journalFilesProperty().size());
```
tobiasdiez: again remove size variable

src/test/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsTest.java
```diff
@@ -0,0 +1,560 @@
...
+    @Test
+    public void testBuiltInLists() {
+        viewModel.addBuiltInLists();
+        Assert.assertEquals(2, viewModel.journalFilesProperty().getSize());
+        viewModel.changeActiveFile(viewModel.journalFilesProperty().get(0));
+        ObservableList<Abbreviation> expected = FXCollections
+                .observableArrayList(Globals.journalAbbreviationLoader.getBuiltInAbbreviations());
+        ObservableList<AbbreviationViewModel> actual = viewModel.abbreviationsProperty().get();
+        ObservableList<Abbreviation> actualAbbreviations = FXCollections.observableArrayList();
+        actual.forEach(abbViewModel -> actualAbbreviations.add(abbViewModel.getAbbreviationObject()));
```
tobiasdiez: use map function, `viewModel.getAbbrevations().stream().map(getAbbrevaitonObject).collect()`

src/test/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsTest.java
```diff
@@ -0,0 +1,560 @@
...
+    public void testBuiltInListsStandardIEEE() {
+        viewModel.addBuiltInLists();
+        selectLastJournalFile();
+        Globals.prefs.put(JabRefPreferences.USE_IEEE_ABRV, "true");
+        Assert.assertEquals(2, viewModel.journalFilesProperty().getSize());
+        ObservableList<Abbreviation> expected = FXCollections
+                .observableArrayList(Globals.journalAbbreviationLoader.getOfficialIEEEAbbreviations());
+        ObservableList<AbbreviationViewModel> actual = viewModel.abbreviationsProperty().get();
+        ObservableList<Abbreviation> actualAbbreviations = FXCollections.observableArrayList();
+        actual.forEach(abbViewModel -> actualAbbreviations.add(abbViewModel.getAbbreviationObject()));
```
tobiasdiez: see above

src/test/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsTest.java
```diff
@@ -0,0 +1,560 @@
...
+        Assert.assertEquals(expected, actualAbbreviations);
+    }
+
+    @Test
+    public void testOpenEmptyFile() throws DuplicatedJournalFileException {
+        int size = viewModel.journalFilesProperty().size();
+
+        // simulate open file button
+        viewModel.addNewFile(emptyTestFile);
+        size++;
```
tobiasdiez: again size variable, remove it everywhere pls

src/test/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsTest.java
```diff
@@ -0,0 +1,560 @@
...
+        viewModel.addAbbreviation();
+    }
+
+    @Test
+    public void testEditSameAbbreviationWithNoChange() throws JabRefException {
+        viewModel.addNewFile(emptyTestFile);
+        selectLastJournalFile();
+        Abbreviation testAbbreviation = new Abbreviation("YetAnotherEntry", "YAE");
+        addAbbrevaition(testAbbreviation);
+        editAbbreviation(testAbbreviation);
```
tobiasdiez: no assertion?

tobiasdiez:
I just made some final remarks. After these are fixed, this PR can be merged in my opinion. @koppor pls do the final review & merge, thanks.

@boceckts (and the all others who worked on this PR): good job! I really like your code, especially that you wrote tests to implement the logic! One can see that you learned a lot in the last few weeks.

src/test/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsTest.java
boceckts: But the important thing here is that the files are getting saved to the preferences..maybe I change it to `addFourTestFileToViewModelAndPreferences`.

boceckts:
I think I now addressed all of your comments.

tobiasdiez:
I'll merge this now, since it looks good to me and all the javafx code will be finally reviewed anyway.

koppor:
Does not work in `javafx`. @boceckts Can you please check?

```
Exception in thread "JavaFX Application Thread" java.lang.IllegalStateException: Cannot load net.sf.jabref.gui.journals.managejournalabbreviations
        at com.airhacks.afterburner.views.FXMLView.loadSynchronously(FXMLView.java:94)
        at com.airhacks.afterburner.views.FXMLView.initializeFXMLLoader(FXMLView.java:101)
        at com.airhacks.afterburner.views.FXMLView.getView(FXMLView.java:113)
        at net.sf.jabref.gui.journals.ManageJournalAbbreviationsView.showAndWait(ManageJournalAbbreviationsView.java:201)
        at net.sf.jabref.gui.journals.ManageJournalsAction.lambda$actionPerformed$0(ManageJournalsAction.java:21)
        at com.sun.javafx.application.PlatformImpl.lambda$null$173(PlatformImpl.java:295)
        at java.security.AccessController.doPrivileged(Native Method)
        at com.sun.javafx.application.PlatformImpl.lambda$runLater$174(PlatformImpl.java:294)
        at com.sun.glass.ui.InvokeLaterDispatcher$Future.run(InvokeLaterDispatcher.java:95)
        at com.sun.glass.ui.win.WinApplication._runLoop(Native Method)
        at com.sun.glass.ui.win.WinApplication.lambda$null$148(WinApplication.java:191)
        at java.lang.Thread.run(Thread.java:745)
Caused by: javafx.fxml.LoadException: No resources specified.
/C:/git-repositories/jabref/jabref/build/resources/main/net/sf/jabref/gui/journals/managejournalabbreviations.fxml:22

        at javafx.fxml.FXMLLoader.constructLoadException(FXMLLoader.java:2597)
```

boceckts:
@koppor I will look into it, did this happen after the javafx branch has been updated?

boceckts:
@koppor Ok so I tracked the issue and came to the conclusion that the this pr #1918 causes the exception. As described in the pr's description @tobiasdiez had to out comment the line `bundle = Localization.getMessages();` since otherwise the localization tests would fail. After adding it back it works for me again.

tobiasdiez:
@boceckts do the localization tests also work after readding the `bundle = ...` line?

boceckts:
No but it gives me the same errors when the line is out commented, I'm not sure if this statement is causing the failing tests..

koppor:
Yes, this happens with the current javafx branch (merge of master). I did not check the state before the merge. My gut feeling is that the merge broke it. But I'm not sure.

I is **not** about failing tests, it is about that "Options -> Manage Journal Abbreviations" is a noop.

boceckts:
Yes it's about opening the dialog. But still, re-adding the line as described by me two comments ago fixes the issue for me.

tobiasdiez:
Reenabled in #1935.

PR #1529, opened 2016-06-29
French localization: translation of empty strings; correction of a few glitches

mlep:
French localization: translation of empty strings; correction of a few glitches


matthiasgeiger:
Thanks @mlep!

PR #1532, opened 2016-06-29
Fix update notification for developer builds

tobiasdiez:
Just running the compiled version without using gradle resulted in a NumberFormatException since the version update was based on the wrong version string `${version}`. This is fixed with this PR (also in this case no update notification is shown).

### DIFF
--- a/src/main/java/net/sf/jabref/logic/util/Version.java
+++ b/src/main/java/net/sf/jabref/logic/util/Version.java
@@ -90,10 +90,8 @@ public class Version {
         } else if (this.getMajor() == otherVersion.getMajor()) {
             if (this.getMinor() > otherVersion.getMinor()) {
                 return true;
-            } else if ((this.getMinor() == otherVersion.getMinor()) && (this.getPatch() > otherVersion.getPatch())) {
-                return true;
             } else {
-                return false;
+                return (this.getMinor() == otherVersion.getMinor()) && (this.getPatch() > otherVersion.getPatch());
             }
         } else {
             return false;
--- a/src/main/java/net/sf/jabref/logic/util/Version.java
+++ b/src/main/java/net/sf/jabref/logic/util/Version.java
@@ -23,6 +23,8 @@ import java.net.URLConnection;
 import java.util.Objects;
 import java.util.regex.Pattern;
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 import org.json.JSONObject;
 
 /**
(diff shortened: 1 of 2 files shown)


src/main/java/net/sf/jabref/logic/util/Version.java
```diff
@@ -86,7 +89,7 @@ public boolean isNewerThan(Version otherVersion) {
             return false;
         }
         if (otherVersion.getFullVersion().equals(BuildInfo.UNKNOWN_VERSION)) {
-            return true;
+            return false;
```
koppor: I'm not sure if this is correct because of symmetry reasons?

koppor:
Besides the minor comment: LGTM

src/test/java/net/sf/jabref/logic/util/version/VersionTest.java
```diff
@@ -13,8 +13,29 @@
 public class VersionTest {
 
     @Test
-    public void unknownVersion() {
-        Version version = new Version(BuildInfo.UNKNOWN_VERSION);
+    public void unknownVersionAsString() {
+        String versionText = BuildInfo.UNKNOWN_VERSION;
```
tobiasdiez: @koppor No need to extract the parameter into a variable `versionText` in these tests, right?

src/test/java/net/sf/jabref/logic/util/version/VersionTest.java
```diff
@@ -124,4 +188,25 @@ public void changelogWithThreeDigits(){
         assertEquals("https://github.com/JabRef/jabref/blob/v3.4.1/CHANGELOG.md", version.getChangelogUrl());
     }
 
+    @Test
+    public void versionNotReplaced() {
```
tobiasdiez: @koppor Duplicates test above.

src/test/java/net/sf/jabref/logic/util/version/VersionTest.java
```diff
@@ -85,13 +106,42 @@ public void versionThreeDigitDevVersion() {
...
+        Version version1 = new Version("4.2");
+        Version version2 = new Version("4.2");
+        assertFalse(version1.isNewerThan(version2));
+    }
+
+    @Test
+    public void concreteVersionNewerThanUnknownVersion() {
+        Version concreteVersion = new Version("4.2");
+        Version unknownVersion = new Version(BuildInfo.UNKNOWN_VERSION);
+        assertTrue(concreteVersion.isNewerThan(unknownVersion));
```
tobiasdiez: @koppor duplicates test above but with opposite assertion. As outlined above, I think it is better to have unknown > valid versions since otherwise we developers constantly get "new update" notifications on startup.

tobiasdiez:
@koppor Thanks for your additions. Can you please fix my small comments (just push directly to the branch).

src/test/java/net/sf/jabref/logic/util/version/VersionTest.java
koppor: Just wanted to add some consistency with the other tests

src/test/java/net/sf/jabref/logic/util/version/VersionTest.java
koppor: Didn't we say that we want to have a single assert statement per test case and avoid multiple assert statements? Therefore, I duplicated the tests.

src/test/java/net/sf/jabref/logic/util/version/VersionTest.java
koppor: Leftover from my parallel work :innocent:. Removed.

koppor:
@tobiasdiez Your opinion regarding the [codacy issues regarding the initialization with already-set default values](https://www.codacy.com/app/simonharrer/jabref/pullRequest?prid=271317)?

tobiasdiez:
@kopper: I don' care 😸 Complaining about unnecessary code in the most verbose programming language ever feels a bit sarcastic...

src/main/java/net/sf/jabref/logic/util/Version.java
```diff
@@ -31,40 +33,38 @@
...
     /**
-     * @param version must be in form of X.X (eg 3.3; 3.4dev)
+     * @param version must be in form of X.X (e.g., 3.3; 3.4dev)
      */
     public Version(String version) {
-        if (version == null || "".equals(version) || version.equals(BuildInfo.UNKNOWN_VERSION)) {
-            this.fullVersion = BuildInfo.UNKNOWN_VERSION;
-            this.major = this.minor = this.patch = 0;
-            this.isDevelopmentVersion = false;
+        if ((version == null) || "".equals(version) || version.equals(BuildInfo.UNKNOWN_VERSION)) {
```
bartsch-dev: Can you add `"${version}".equals(version)` as another disjunction?

When you compile the code without gradle (like directly in eclipse or IntelliJ IDEA) the versionnumber is not set (as it is set within the gradle processResources task) and `${version}` will be read in which will currently lead to an error.

koppor: That was the quick fix I applied in parallel to the work of @tobiasdiez. He used exception handling at parsing to treat all kinds of errors. Please also note that testing for `${version}` has been added to the test cases: https://github.com/JabRef/jabref/pull/1532/files/8b14b0bb0c4b9c2069e4c81611ef9c33fccc8904#diff-6d5e38b382ea230b7f2b6184d9bcac76R35

koppor:
This is an improvement of #1459.

PR #1539, opened 2016-07-03
Fix #1500: Renaming of explicit groups changes entries accordingly

### LINKED ISSUE #1500: Rename a static group: Group entries are not updated accordingly
JabRef-3.4dev--snapshot--2016-04-29--master--694840a
on Windows 10

If you rename a static group, the group entries are not updated accordingly. Their "group" field will still contain the old group name (and NOT the new one).

Steps to reproduce:

1. Rename a static group
2. Articles/group entries won't be updated accordingly and are still assigned to the old (obsolete) group name

(might be related to this one: https://github.com/JabRef/jabref/issues/1434#event-691424400 ?)

tobiasdiez:
Also moved the code responsible for carrying over previous assignments from the gui to logic; add tests for it.

### DIFF
--- a/src/main/java/net/sf/jabref/gui/groups/GroupDialog.java
+++ b/src/main/java/net/sf/jabref/gui/groups/GroupDialog.java
@@ -88,14 +88,14 @@ class GroupDialog extends JDialog {
     private final JRadioButton unionButton = new JRadioButton(
             Localization.lang("Include subgroups: When selected, view entries contained in this group or its subgroups"));
     // for KeywordGroup
-    private final JTextField kgSearchField = new JTextField(GroupDialog.TEXTFIELD_LENGTH);
-    private final TextField kgSearchTerm = new TextField("keywords", "", false);
-    private final JCheckBox kgCaseSensitive = new JCheckBox(Localization.lang("Case sensitive"));
-    private final JCheckBox kgRegExp = new JCheckBox(Localization.lang("regular expression"));
+    private final JTextField keywordGroupSearchField = new JTextField(GroupDialog.TEXTFIELD_LENGTH);
+    private final TextField keywordGroupSearchTerm = new TextField("keywords", "", false);
+    private final JCheckBox keywordGroupCaseSensitive = new JCheckBox(Localization.lang("Case sensitive"));
+    private final JCheckBox keywordGroupRegExp = new JCheckBox(Localization.lang("regular expression"));
     // for SearchGroup
-    private final JTextField sgSearchExpression = new JTextField(GroupDialog.TEXTFIELD_LENGTH);
-    private final JCheckBox sgCaseSensitive = new JCheckBox(Localization.lang("Case sensitive"));
-    private final JCheckBox sgRegExp = new JCheckBox(Localization.lang("regular expression"));
+    private final JTextField searchGroupSearchExpression = new JTextField(GroupDialog.TEXTFIELD_LENGTH);
+    private final JCheckBox searchGroupCaseSensitive = new JCheckBox(Localization.lang("Case sensitive"));
+    private final JCheckBox searchGroupRegExp = new JCheckBox(Localization.lang("regular expression"));
     // for all types
     private final JButton okButton = new JButton(Localization.lang("OK"));
     private final JPanel optionsPanel = new JPanel();
@@ -130,7 +130,7 @@ class GroupDialog extends JDialog {
         super(jabrefFrame, Localization.lang("Edit group"), true);
 
         // set default values (overwritten if editedGroup != null)
-        kgSearchField.setText(jabrefFrame.prefs().get(JabRefPreferences.GROUPS_DEFAULT_FIELD));
+        keywordGroupSearchField.setText(jabrefFrame.prefs().get(JabRefPreferences.GROUPS_DEFAULT_FIELD));
 
         // configure elements
         ButtonGroup groupType = new ButtonGroup();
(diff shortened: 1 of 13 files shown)


src/main/java/net/sf/jabref/gui/groups/GroupDialog.java
```diff
@@ -296,27 +280,13 @@ public void actionPerformed(ActionEvent e) {
...
-                        mResultingGroup = m_editedGroup.deepCopy();
-                        mResultingGroup.setName(m_name.getText().trim());
-                        mResultingGroup.setHierarchicalContext(getContext());
-                    } else {
-                        mResultingGroup = new ExplicitGroup(m_name.getText().trim(), getContext());
-                        if (m_editedGroup != null) {
-                            addPreviousEntries();
-                        }
-                    }
+                    mResultingGroup = new ExplicitGroup(m_name.getText().trim(), getContext());
```
lenhard: Since you are touching that code anyway: How about refactoring the names here according to Java code conventions? `m_name` -> `mName`. What does the `m` stand for, btw?

src/main/java/net/sf/jabref/logic/groups/GroupTreeNode.java
```diff
@@ -55,10 +57,37 @@ public AbstractGroup getGroup() {
...
+     * entries so that they are now matched by the new group.
+     *
+     * @param newGroup the new group (has to be non-null)
+     * @param shouldKeepPreviousAssignments specifies whether previous matched entries should be carried over
+     * @param entriesInDatabase list of entries in the database
+     */
+    public Optional<EntriesGroupChange> setGroup(AbstractGroup newGroup, boolean shouldKeepPreviousAssignments,
+            List<BibEntry> entriesInDatabase) {
+        AbstractGroup oldGroup = getGroup();
+        setGroup(newGroup);
```
lenhard: You marked this method as deprecated above. Why are you still using it in new code?

src/test/java/net/sf/jabref/logic/groups/GroupTreeNodeTest.java
```diff
@@ -14,15 +16,23 @@
...
 
     @Before
     public void setUp() throws Exception {
+        entries.clear();
+        entry = new BibEntry();
+        entries.add(entry);
         entries.add(new BibEntry().withField("author", "author1 and author2"));
         entries.add(new BibEntry().withField("author", "author1"));
+
+        Globals.prefs = JabRefPreferences.getInstance();
```
lenhard: Shouldn't the preferences setting go into a `@BeforeClass`-annotated method? Or do you need this for every test?

lenhard:
Only minor remarks. LGTM.

src/main/java/net/sf/jabref/logic/groups/GroupTreeNode.java
tobiasdiez: Once the other method is no longer used it can be inlined (or at least made private).

src/test/java/net/sf/jabref/logic/groups/GroupTreeNodeTest.java
tobiasdiez: It would also work with @BeforClass but I like to have only one setup method.

src/main/java/net/sf/jabref/gui/groups/GroupDialog.java
tobiasdiez: Done.

src/test/java/net/sf/jabref/logic/groups/GroupTreeNodeTest.java
simonharrer: I would use a separate `@BeforeClass` method to distinguish code that is required only once, from code that is required to be executed for every test.

src/main/java/net/sf/jabref/gui/groups/GroupDialog.java
```diff
@@ -77,37 +69,37 @@
...
+    private final TextField kgSearchTerm = new TextField("keywords", "", false);
+    private final JCheckBox kgCaseSensitive = new JCheckBox(Localization.lang("Case sensitive"));
+    private final JCheckBox kgRegExp = new JCheckBox(Localization.lang("regular expression"));
     // for SearchGroup
-    private final JTextField m_sgSearchExpression = new JTextField(GroupDialog.TEXTFIELD_LENGTH);
-    private final JCheckBox m_sgCaseSensitive = new JCheckBox(Localization.lang("Case sensitive"));
-    private final JCheckBox m_sgRegExp = new JCheckBox(Localization.lang("regular expression"));
+    private final JTextField sgSearchExpression = new JTextField(GroupDialog.TEXTFIELD_LENGTH);
+    private final JCheckBox sgCaseSensitive = new JCheckBox(Localization.lang("Case sensitive"));
+    private final JCheckBox sgRegExp = new JCheckBox(Localization.lang("regular expression"));
```
simonharrer: what does sg stand for?

simonharrer:
LGTM, just merge after you addressed the minor comments.

src/main/java/net/sf/jabref/gui/groups/GroupDialog.java
tobiasdiez: searchGroup, I would make an exception here for the general rule of no abbreviations since `searchGroupRegualarExpressionCheckBox` is a bit to big ;)

simonharrer: Not convinced. As `sg` is not really that common. Why not searchGroupRegExCheckBox?

stefan-kolb:
LGTM :+1:

PR #1541, opened 2016-07-04
The customize entry dialogs field list gets the focus as soon as it is focused

Braunch:
When using tab to navigate through the entry customization dialog, the lists now selects the first entry to give visual feedback when they are selected and to enable the up and down buttons functionality.

image

### DIFF
--- a/src/main/java/net/sf/jabref/gui/FieldSetComponent.java
+++ b/src/main/java/net/sf/jabref/gui/FieldSetComponent.java
@@ -22,6 +22,8 @@ import java.awt.Insets;
 import java.awt.Rectangle;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
+import java.awt.event.FocusEvent;
+import java.awt.event.FocusListener;
 import java.util.ArrayList;
 import java.util.Enumeration;
 import java.util.HashSet;
@@ -48,7 +50,6 @@ import net.sf.jabref.logic.l10n.Localization;
 import net.sf.jabref.logic.labelpattern.LabelPatternUtil;
 
 /**
- *
  * @author alver
  */
 class FieldSetComponent extends JPanel implements ActionListener {
@@ -176,6 +177,8 @@ class FieldSetComponent extends JPanel implements ActionListener {
         gbl.setConstraints(add, con);
         add(add);
 
+        FieldListFocusListener fieldListFocusListener = new FieldListFocusListener(list);
+        list.addFocusListener(fieldListFocusListener);
     }
 
     public void setListSelectionMode(int mode) {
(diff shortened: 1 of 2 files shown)


src/main/java/net/sf/jabref/gui/FieldSetComponent.java
```diff
@@ -90,6 +92,7 @@ public FieldSetComponent(String title, List<String> fields, List<String> preset,
 
     private FieldSetComponent(String title, List<String> fields, List<String> preset, String addText, String removeText,
                               boolean arrows, boolean forceLowerCase) {
+
```
boceckts: Why did this change?

src/main/java/net/sf/jabref/gui/FieldSetComponent.java
```diff
@@ -364,4 +369,29 @@ public void actionPerformed(ActionEvent e) {
         }
     }
 
+    /**
+     * FocusListener to select the first entry in the list of fields when they are focused
+     */
+    protected class FieldListFocusListener implements FocusListener {
+
+        private JList list;
```
simonharrer: Can you parameterize this with generics?

src/main/java/net/sf/jabref/gui/FieldSetComponent.java
```diff
@@ -176,6 +178,9 @@ private FieldSetComponent(String title, List<String> fields, List<String> preset
         gbl.setConstraints(add, con);
         add(add);
 
+        FieldListFocusListener fieldListFocusListener = new FieldListFocusListener(list);
+        list.addFocusListener(fieldListFocusListener);
+
```
tobiasdiez: Remove empty line

src/main/java/net/sf/jabref/gui/FieldSetComponent.java
```diff
@@ -364,4 +369,29 @@ public void actionPerformed(ActionEvent e) {
...
+        private JList list;
+
+        public FieldListFocusListener(JList list) {
+            this.list = list;
+        }
+
+        @Override
+        public void focusGained(FocusEvent e) {
+            if (list.getSelectedValue() == null) {
+                list.setSelectedIndex(0);
```
tobiasdiez: If I click on an item in the list (without giving the focus to the list before), is then the clicked item selected or the first one?

tobiasdiez:
LGTM :+1: in general. Just some small remarks. Can be merged after these are fixed.

src/main/java/net/sf/jabref/gui/FieldSetComponent.java
Braunch: The clicked item is selected since the selection value is no longer null. I tested it to be sure.

Braunch:
Can this be merged?

simonharrer:
why did the gradle script change? Otherwise, LGTM.

Braunch:
Gradle script is back to normal.

PR #1545, opened 2016-07-04
Added content-type heading for HTML export fixing #1484

### LINKED ISSUE #1484: HTML export handles some UTF characters wrong
JabRef 3.4 on Ubuntu 16.04

Steps to reproduce:

1. Copy and paste the following test entry into a new database in JabRef.
2. Make sure the database is stored as UTF-8 encoded (should be default).
3. Export the database and select the file type 'HTML (*.html)'
4. In the resulting HTML file, the character 'Ø' is displayed wrong in the Firefox browser.
jabref
firefox

This does not happen when exporting to 'Simple HTML', 'HTML list', 'HTML table', or 'HTML table (with Abstract and BibTeX)'.

oscargus:
Added a content type tag (and fixed tag name, head instead of header) to fix #1484.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - Fixed [#1482](https://github.com/JabRef/jabref/issues/1482): Correct number of matched entries is displayed for refining subgroups
 - Fixed [#1444](https://github.com/JabRef/jabref/issues/1444): Implement getExtension and getDescription for importers.
 - Fixed [#1507](https://github.com/JabRef/jabref/issues/1507): Keywords are now separated by the delimiter specified in the preferences
+- Fixed [#1484](https://github.com/JabRef/jabref/issues/1484): HTML export handles some UTF characters wrong
 
 ### Removed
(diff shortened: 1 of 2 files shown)


Siedlerchr:
Yeah, but there is one problem! It works only for UTF8 databases.
If I switch my db to ASCII (right click on the bib tab) or whatever Encoding, my Unicode would be screwed again...

oscargus:
Correct. However, in that case all the other HTML exports will also screw up as they state UTF-8 as well.

(Btw, ASCII and ISO 8859-1/Latin 1 will work...)

oscargus:
A possible argument (apart from being lazy and providing a quick fix) is that we cannot convert from any other encoding apart from UTF-8 anyway. So even if we go Unicode -> LaTeX -> HTML it will only work for UTF-8.

A better option might be to add a layout formatter that outputs the database encoding string.

Siedlerchr:
Yeah. I think we should stay with your quick fix first. I doubt that people use other weird encoding than UTF-8 together with the HTML export. And if really, they can still do a Quick Search + Replace in the HTML code.

simonharrer:
LGTM

PR #1546, opened 2016-07-04
Some code cleanups

oscargus:
While I was at it: some code cleanups...

* entrySet instead of keySet
* correct order of static and final
* used equalsIgnoreCase instead of equals() and toLowerCase()
* added default switch case
* "correct" form of lambda (not sure what it is called)
* isEmpty() instead of size() > 0
* removed unused private variable

### DIFF
--- a/src/main/java/net/sf/jabref/gui/ContentSelectorDialog2.java
+++ b/src/main/java/net/sf/jabref/gui/ContentSelectorDialog2.java
@@ -268,12 +268,12 @@ class ContentSelectorDialog2 extends JDialog {
         }
 
         // Cycle through all fields that we have created listmodels for:
-        for (String fieldName : wordListModels.keySet()) {
+        for (Map.Entry<String, DefaultListModel<String>> entry : wordListModels.entrySet()) {
             // For each field name, store the values:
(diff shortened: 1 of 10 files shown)


src/main/java/net/sf/jabref/gui/ContentSelectorDialog2.java
```diff
@@ -268,12 +268,12 @@ private void applyChanges() {
         }
 
         // Cycle through all fields that we have created listmodels for:
-        for (String fieldName : wordListModels.keySet()) {
+        for (Map.Entry<String, DefaultListModel<String>> entry : wordListModels.entrySet()) {
```
Siedlerchr: Just an idea: Maps etc support Streams with Lambdas, too.
http://www.mkyong.com/java8/java-8-filter-a-map-examples/

Siedlerchr:
LGTM :+1:

src/main/java/net/sf/jabref/gui/ContentSelectorDialog2.java
oscargus: Out of my league at the moment, but most likely a good idea! :-)

stefan-kolb:
LGTM :+1:

tobiasdiez:
LGTM :+1:

PR #1548, opened 2016-07-04
A few conversions of getField to getFieldOptional

oscargus:
Only a few, but always something...

### DIFF
--- a/src/main/java/net/sf/jabref/external/SynchronizeFileField.java
+++ b/src/main/java/net/sf/jabref/external/SynchronizeFileField.java
@@ -77,11 +77,8 @@ public class SynchronizeFileField extends AbstractWorker {
 
     private int entriesChangedCount;
 
-    private final Object[] brokenLinkOptions = {
-            Localization.lang("Ignore"),
-            Localization.lang("Assign new file"),
-            Localization.lang("Remove link"),
-            Localization.lang("Remove all broken links"),
+    private final Object[] brokenLinkOptions = {Localization.lang("Ignore"), Localization.lang("Assign new file"),
+            Localization.lang("Remove link"), Localization.lang("Remove all broken links"),
             Localization.lang("Quit synchronization")};
 
     private boolean goOn = true;
@@ -301,14 +298,10 @@ public class SynchronizeFileField extends AbstractWorker {
         private final JButton cancel = new JButton(Localization.lang("Cancel"));
         private boolean canceled = true;
         private final BibDatabaseContext databaseContext;
-        private final JRadioButton autoSetUnset = new JRadioButton(
-                Localization.lang("Automatically set file links") + ". "
-                        + Localization.lang("Do not overwrite existing links."),
-                true);
-        private final JRadioButton autoSetAll = new JRadioButton(
-                Localization.lang("Automatically set file links") + ". "
-                        + Localization.lang("Allow overwriting existing links."),
-                false);
+        private final JRadioButton autoSetUnset = new JRadioButton(Localization.lang("Automatically set file links")
+                + ". " + Localization.lang("Do not overwrite existing links."), true);
+        private final JRadioButton autoSetAll = new JRadioButton(Localization.lang("Automatically set file links")
+                + ". " + Localization.lang("Allow overwriting existing links."), false);
         private final JRadioButton autoSetNone = new JRadioButton(Localization.lang("Do not automatically set"), false);
         private final JCheckBox checkLinks = new JCheckBox(Localization.lang("Check existing file links"), true);
 
@@ -341,11 +334,13 @@ public class SynchronizeFileField extends AbstractWorker {
             bg.add(autoSetNone);
             bg.add(autoSetAll);
 
-            FormLayout layout = new FormLayout("fill:pref", "pref, 2dlu, pref, 2dlu, pref, pref, pref, 2dlu, pref, 2dlu, pref, 2dlu, pref, 2dlu, pref");
+            FormLayout layout = new FormLayout("fill:pref",
+                    "pref, 2dlu, pref, 2dlu, pref, pref, pref, 2dlu, pref, 2dlu, pref, 2dlu, pref, 2dlu, pref");
             FormBuilder builder = FormBuilder.create().layout(layout);
-            JLabel description = new JLabel("<HTML>" + Localization.lang(
-                    "Attempt to automatically set file links for your entries. Automatically setting works if "
-                            + "a file in your file directory<BR>or a subdirectory is named identically to an entry's BibTeX key, plus extension.")
+            JLabel description = new JLabel("<HTML>"
+                    + Localization
+                            .lang("Attempt to automatically set file links for your entries. Automatically setting works if "
+                                    + "a file in your file directory<BR>or a subdirectory is named identically to an entry's BibTeX key, plus extension.")
                     + "</HTML>");
 
             builder.addSeparator(Localization.lang("Automatically set file links")).xy(1, 1);
--- a/src/main/java/net/sf/jabref/external/SynchronizeFileField.java
+++ b/src/main/java/net/sf/jabref/external/SynchronizeFileField.java
@@ -147,11 +147,11 @@ public class SynchronizeFileField extends AbstractWorker {
             boolean removeAllBroken = false;
             mainLoop: for (BibEntry aSel : sel) {
                 panel.frame().setProgressBarValue(progress++);
-                final String old = aSel.getField(Globals.FILE_FIELD);
+                final Optional<String> old = aSel.getFieldOptional(Globals.FILE_FIELD);
                 // Check if a extension is set:
-                if ((old != null) && !(old.isEmpty())) {
+                if (old.isPresent() && !(old.get().isEmpty())) {
                     FileListTableModel tableModel = new FileListTableModel();
-                    tableModel.setContentDontGuessTypes(old);
+                    tableModel.setContentDontGuessTypes(old.get());
 
                     // We need to specify which directories to search in for Util.expandFilename:
                     List<String> dirsS = panel.getBibDatabaseContext().getFileDirectory();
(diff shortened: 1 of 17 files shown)


src/main/java/net/sf/jabref/logic/util/DOI.java
```diff
@@ -107,6 +107,23 @@ public DOI(String doi) {
...
+     *
+     * Useful for suppressing the <c>IllegalArgumentException</c> of the Constructor
+     * and checking for Optional.isPresent() instead.
+     *
+     * @param doi the DOI string
+     * @return an Optional containing the DOI or an empty Optional
+     */
+    public static Optional<DOI> build(Optional<String> doi) {
+        if (doi.isPresent()) {
+            return DOI.build(doi.get());
```
Siedlerchr: One liner: `doi.isPresent ? DOO.build(doi.get()) : Optional.empty() `

src/main/java/net/sf/jabref/model/DuplicateCheck.java
```diff
@@ -133,17 +133,21 @@ public static boolean isDuplicate(BibEntry one, BibEntry two, BibDatabaseMode bi
...
 
     private static int compareSingleField(String field, BibEntry one, BibEntry two) {
-        String s1 = one.getField(field);
-        String s2 = two.getField(field);
-        if (s1 == null) {
-            if (s2 == null) {
+        Optional<String> os1 = one.getFieldOptional(field);
+        Optional<String> os2 = two.getFieldOptional(field);
+        if (!os1.isPresent()) {
+            if (!os2.isPresent()) {
```
Siedlerchr: What about using an `&&` here? `!os1.isPresent() && ! os2.isPresnet()`

src/main/java/net/sf/jabref/logic/util/DOI.java
stefan-kolb: I like it the way it is.

src/main/java/net/sf/jabref/logic/fulltext/GoogleScholar.java
```diff
@@ -35,7 +35,7 @@
         }
 
         String url = String.format(SEARCH_URL,
-                URLEncoder.encode(entry.getField("title"), StandardCharsets.UTF_8.name()));
+                URLEncoder.encode(entry.getFieldOptional("title").get(), StandardCharsets.UTF_8.name()));
```
tobiasdiez: what happens if no title is present?

src/main/java/net/sf/jabref/logic/fulltext/IEEE.java
```diff
@@ -47,16 +48,19 @@
...
-                // Download the HTML page from IEEE
-                String resolvedDOIPage = new URLDownload(doi.get().getURI().get().toURL())
-                        .downloadToString(StandardCharsets.UTF_8);
-                // Try to find the link
-                Matcher matcher = STAMP_PATTERN.matcher(resolvedDOIPage);
-                if (matcher.find()) {
-                    // Found it
-                    stampString = matcher.group(1);
+            Optional<String> doiString = entry.getFieldOptional("doi");
+            if (doiString.isPresent()) {
```
tobiasdiez: Use the new DOI.build(Optional) method here?

src/main/java/net/sf/jabref/model/DuplicateCheck.java
```diff
@@ -133,17 +133,21 @@ public static boolean isDuplicate(BibEntry one, BibEntry two, BibDatabaseMode bi
     }
 
     private static int compareSingleField(String field, BibEntry one, BibEntry two) {
-        String s1 = one.getField(field);
-        String s2 = two.getField(field);
-        if (s1 == null) {
-            if (s2 == null) {
+        Optional<String> os1 = one.getFieldOptional(field);
```
tobiasdiez: Better names pls

tobiasdiez: Better names pls

src/main/java/net/sf/jabref/model/database/BibDatabase.java
```diff
@@ -352,6 +353,15 @@ public String resolveForStrings(String content) {
     }
 
     /**
+     * Resolves any references to strings contained in this field content,
+     * if possible.
+     */
+    public String resolveForStrings(Optional<String> content) {
```
tobiasdiez: I don't like the optional parameter here.

tobiasdiez:
LGTM :+1:

src/main/java/net/sf/jabref/external/SynchronizeFileField.java
```diff
@@ -147,11 +147,11 @@ public void run() {
             boolean removeAllBroken = false;
             mainLoop: for (BibEntry aSel : sel) {
                 panel.frame().setProgressBarValue(progress++);
-                final String old = aSel.getField(Globals.FILE_FIELD);
+                final Optional<String> old = aSel.getFieldOptional(Globals.FILE_FIELD);
                 // Check if a extension is set:
-                if ((old != null) && !(old.isEmpty())) {
+                if (old.isPresent() && !(old.get().isEmpty())) {
```
simonharrer: try to use ifPresent instead of isPresent and get if possible.

src/main/java/net/sf/jabref/logic/util/DOI.java
```diff
@@ -107,6 +107,23 @@ public DOI(String doi) {
...
     /**
+     * Creates an Optional<DOI> from various schemes including URL, URN, and plain DOIs.
+     *
+     * Useful for suppressing the <c>IllegalArgumentException</c> of the Constructor
+     * and checking for Optional.isPresent() instead.
+     *
+     * @param doi the DOI string
+     * @return an Optional containing the DOI or an empty Optional
+     */
+    public static Optional<DOI> build(Optional<String> doi) {
```
simonharrer: Optionals shall only be used as return values, never as parameters.

tobiasdiez: I kind of like it here since it leads to a more fluid API. See also the discussion at http://stackoverflow.com/questions/31922866/why-should-java-8s-optional-not-be-used-in-arguments

simonharrer: Better would be if you have an optional string to use `optionalString.flatMap(string -> DOI.build(string))` as this would then return an Optional<DOI>, and the code would only fire if there is a string.

src/main/java/net/sf/jabref/logic/util/DOI.java
oscargus: I'm in general not that fond of the '?:' syntax.

src/main/java/net/sf/jabref/model/DuplicateCheck.java
oscargus: One would still need three if-clauses:

'!os1.isPresent() && !os2.isPresent()' => 'EMPTY_IN_BOTH'
'!os1.isPresent()'  => 'EMPTY_IN_ONE'
'!os2.isPresent()'  => 'EMPTY_IN_TWO'

src/main/java/net/sf/jabref/model/database/BibDatabase.java
oscargus: Me neither, but at the moment it looked like the better option (compared to blindly doing `get()` or explicitly adding `ifPresent()`). I realized another one now though.

src/main/java/net/sf/jabref/external/SynchronizeFileField.java
oscargus: Yes, you are correct. A bit lazy there...

oscargus: Here it is not actually not possible.

src/main/java/net/sf/jabref/logic/util/DOI.java
oscargus: Even better:
`optionalString.flatMap(DOI::build)`
;-)

src/main/java/net/sf/jabref/specialfields/SpecialFieldsUtils.java
```diff
@@ -85,14 +85,15 @@ public static void updateField(SpecialField e, String value, BibEntry be, NamedC
...
+        SpecialFieldsUtils.exportFieldToKeywords(e, be.getFieldOptional(e.getFieldName()), be, ce);
     }
 
     private static void exportFieldToKeywords(SpecialField e, BibEntry be, NamedCompound ce) {
-        SpecialFieldsUtils.exportFieldToKeywords(e, be.getField(e.getFieldName()), be, ce);
+        SpecialFieldsUtils.exportFieldToKeywords(e, be.getFieldOptional(e.getFieldName()), be, ce);
     }
 
-    private static void exportFieldToKeywords(SpecialField e, String newValue, BibEntry entry, NamedCompound ce) {
+    private static void exportFieldToKeywords(SpecialField e, Optional<String> newValue, BibEntry entry,
```
oscargus: Here I think the API simplification is so useful that I still use the Optional.

stefan-kolb: I think the name of the parameter already states that one shouldn't use an optional here? Why should a newValue be optional?

oscargus: Well, I can of course add an `.orElse(null)`, but somehow it seems to remove the idea of `Optional` if every `getField()` is simply replaced with `getFieldOptional().orElse(null)`...

src/main/java/net/sf/jabref/specialfields/SpecialFieldsUtils.java
```diff
@@ -162,8 +163,7 @@ public static void syncSpecialFieldsFromKeywords(BibEntry be, NamedCompound ce)
         if (!be.hasField(KEYWORDS_FIELD)) {
             return;
         }
-        Set<String> keywordList = net.sf.jabref.model.entry.EntryUtil
-                .getSeparatedKeywords(be.getField(KEYWORDS_FIELD));
+        Set<String> keywordList = net.sf.jabref.model.entry.EntryUtil.getSeparatedKeywords(be);
```
tobiasdiez: import model namespace?

oscargus: I also prefer that, but I am quite sure that someone else did indeed change it to look like this...

simonharrer: Just use an import.

simonharrer:
LGTM

PR #1552, opened 2016-07-07
Performance improvements for groups

tobiasdiez:
The BibEntry now stores a list of words which are contained in the field. in this way the KeywordGroup.isMatch method just has to compare two arrays instead of analyzing strings over and over again. For a big file (> 7000 entries, > 100 groups) the load time decreased by 1 sec from 4.8 seconds to 3.7. Half of the time needed for showing the groups tree is spent in calculating the number of hits. Moreover, subsequent rerendings of the groups tree (for example after changing an entry) are also quicker.

Note that the match algorithm changed slightly. For example, previously looking for `test` in `some$test` resulted in a match; this is no longer the case. However, normal word separators (like whitespace, comma and semicolon are recognized properly), i.e. searching for `test` in `some test` returns true.

Also fixed a NPE.

### DIFF
--- a/src/main/java/net/sf/jabref/gui/groups/GroupSelector.java
+++ b/src/main/java/net/sf/jabref/gui/groups/GroupSelector.java
@@ -723,9 +723,8 @@ public class GroupSelector extends SidePaneComponent implements TreeSelectionLis
 
     private void setGroups(GroupTreeNode groupsRoot) {
         this.groupsRoot = new GroupTreeNodeViewModel(groupsRoot);
-        // refactor notice: groupsTreeModel::nodeStructureChanged cannot be used, because an NPE will be risen if no groupsTreeModel exists
-        this.groupsRoot.subscribeToDescendantChanged(source -> groupsTreeModel.nodeStructureChanged(source));
         groupsTreeModel = new DefaultTreeModel(this.groupsRoot);
+        this.groupsRoot.subscribeToDescendantChanged(groupsTreeModel::nodeStructureChanged);
         groupsTree.setModel(groupsTreeModel);
         if (Globals.prefs.getBoolean(JabRefPreferences.GROUP_EXPAND_TREE)) {
             this.groupsRoot.expandSubtree(groupsTree);
--- a/src/main/java/net/sf/jabref/logic/groups/KeywordGroup.java
+++ b/src/main/java/net/sf/jabref/logic/groups/KeywordGroup.java
@@ -18,6 +18,7 @@ package net.sf.jabref.logic.groups;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
+import java.util.Set;
 import java.util.regex.Pattern;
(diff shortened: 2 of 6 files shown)


src/main/java/net/sf/jabref/logic/groups/KeywordGroup.java
```diff
@@ -197,17 +200,41 @@ public boolean equals(Object o) {
...
         if (caseSensitive) {
-            return KeywordGroup.containsWord(searchExpression, content);
+            return words.containsAll(searchWords);
+        }
+        return containsCaseInsensitive(searchWords, words);
+    }
+
+    private boolean containsCaseInsensitive(List<String> searchText, Set<String> words) {
+        for (String searchWord : searchText) {
+            if (!containsCaseInsensitive(searchWord, words)) {
```
Siedlerchr: Could be optimized with Streams and anyMatch()
Or even better, you could directly compare the two collections: 
`ListA.stream().anyMatch(setB::contains)`

http://www.leveluplunch.com/java/examples/java-util-stream-find-match-example/

tobiasdiez: Since this a rather performance-critical method, I decided to use the good old collections interface instead of having the stream overhead.

koppor: Please document that decision in a comment 😇

src/main/java/net/sf/jabref/gui/groups/GroupSelector.java
```diff
@@ -722,8 +722,8 @@ public void componentClosing() {
 
     private void setGroups(GroupTreeNode groupsRoot) {
         this.groupsRoot = new GroupTreeNodeViewModel(groupsRoot);
-        this.groupsRoot.subscribeToDescendantChanged(groupsTreeModel::nodeStructureChanged);
         groupsTreeModel = new DefaultTreeModel(this.groupsRoot);
+        this.groupsRoot.subscribeToDescendantChanged(groupsTreeModel::nodeStructureChanged);
```
oscargus: Ahh, reading the code makes the NPE a bit more understandable... Good to initialize the object... Still interesting that the earlier version worked though!

src/main/java/net/sf/jabref/logic/groups/KeywordGroup.java
```diff
@@ -197,17 +200,41 @@ public boolean equals(Object o) {
 
     @Override
     public boolean contains(BibEntry entry) {
-        if (!entry.hasField(searchField)) {
-            return false;
-        }
-        String content = entry.getField(searchField);
         if (regExp) {
+            String content = entry.getField(searchField);
```
oscargus: Better to use the non-deprecated method getFieldOptional... ;-)

oscargus:
LGTM!

stefan-kolb:
LGTM :+1:

Issue #1521, opened 2016-06-28
Promote discourse.jabref.org


simonharrer:
- [x] README ~~with a badge~~
  - badges for discourse seem not to be available
- CONTRIBUTING.md as we would like users to participate there
- [x] Alter issue-template so that features and enhancements can be discussed in "To propose new features and enhancements please use http://discourse.jabref.org/c/features"
- Add label for issues which have a discourse thread attached?

mlep:
I have just noticed that the mailing lists are still mentioned in http://www.flosscoach.com/index.php/communication-jabref#mailing-lists

matthiasgeiger:
@koppor can you edit this information?

koppor:
@igorsteinmacher Can you take care?

matthiasgeiger:
Should we integrate a link "Online forum" in the "Help" menu?

simonharrer:
With the icon https://materialdesignicons.com/icon/forum?

matthiasgeiger:
Potentially... but I don't think this is really needed in the menu if it is not used anywhere else. But using the icon doesn't hurt, though ;-)

stefan-kolb:
Any more work to do here or are we good for now?

PR #1555, opened 2016-07-08
Link to forum

matthiasgeiger:
Adds another entry to the "Help" menu: "Online help forum" (see https://github.com/JabRef/jabref/issues/1521#issuecomment-230994579)


src/main/java/net/sf/jabref/gui/actions/OpenBrowserAction.java
```diff
@@ -0,0 +1,49 @@
...
+
+import javax.swing.AbstractAction;
+import javax.swing.Action;
+import javax.swing.Icon;
+
+import net.sf.jabref.gui.desktop.JabRefDesktop;
+
+public class OpenBrowserAction extends AbstractAction {
+
+    private String urlToOpen;
```
oscargus: final

oscargus:
LGTM!

src/main/java/net/sf/jabref/gui/actions/OpenBrowserAction.java
```diff
@@ -0,0 +1,49 @@
...
+
+import java.awt.event.ActionEvent;
+
+import javax.swing.AbstractAction;
+import javax.swing.Action;
+import javax.swing.Icon;
+
+import net.sf.jabref.gui.desktop.JabRefDesktop;
+
+public class OpenBrowserAction extends AbstractAction {
```
tobiasdiez: I think the UpdateDialog also has some actions which might be replaced by this one.

matthiasgeiger: No. Only the `JabRefDesktop.openBrowserShowPopup(...)` method is used there - but not by an `AbstractAction` (which is used for menus and icon bars only).

stefan-kolb:
LGTM :+1:

PR #1557, opened 2016-07-08
Some internationalization improvements

oscargus:
* Better casing
* Removed duplicate strings
* More consistent naming


lenhard:
More quality control regarding internationalization is really a good idea. LGTM!

PR #1559, opened 2016-07-08
Fixed #1534

### LINKED ISSUE #1534: "Mark entries imported into database" does not work correctly
JabRef version 3.4, 3.5dev on Windows 10

Steps to reproduce:

1. Preference setting General -> "Mark entries imported into an existing database" is activated
2. Import an entry, e.g., using JabFox
3. The entry is not marked (and therefore not floating at top)

oscargus:
Marking of imported entries should work again (#1534). Not really sure why I introduced that bug. Looking back at the code https://github.com/JabRef/jabref/commit/8d9c806eb41b809822e3acadc9299315850bd10c it doesn't really make sense to do as I did...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - Fixed [#1444](https://github.com/JabRef/jabref/issues/1444): Implement getExtension and getDescription for importers.
 - Fixed [#1507](https://github.com/JabRef/jabref/issues/1507): Keywords are now separated by the delimiter specified in the preferences
 - Fixed [#1484](https://github.com/JabRef/jabref/issues/1484): HTML export handles some UTF characters wrong
+- Fixed [#1534](https://github.com/JabRef/jabref/issues/1534): "Mark entries imported into database" does not work correctly
 
 ### Removed
 
--- a/src/main/java/net/sf/jabref/gui/EntryMarker.java
+++ b/src/main/java/net/sf/jabref/gui/EntryMarker.java
@@ -196,8 +196,6 @@ public class EntryMarker {
     }
 
     public static boolean shouldMarkEntries() {
-        return (Globals.prefs.getBoolean(JabRefPreferences.MARK_IMPORTED_ENTRIES)
-                && (Globals.prefs.getBoolean(JabRefPreferences.USE_OWNER)
-                        || Globals.prefs.getBoolean(JabRefPreferences.USE_TIME_STAMP)));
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - Fixed [#1444](https://github.com/JabRef/jabref/issues/1444): Implement getExtension and getDescription for importers.
 - Fixed [#1507](https://github.com/JabRef/jabref/issues/1507): Keywords are now separated by the delimiter specified in the preferences
 - Fixed [#1484](https://github.com/JabRef/jabref/issues/1484): HTML export handles some UTF characters wrong
+- Fixed [#1534](https://github.com/JabRef/jabref/issues/1534): "Mark entries imported into database" does not work correctly
```
koppor: I don't think, we need a changelog entry here, because the issue was introduced in the dev version.

src/main/java/net/sf/jabref/gui/EntryMarker.java
```diff
@@ -196,8 +196,6 @@ public static int isMarked(BibEntry be) {
     }
 
     public static boolean shouldMarkEntries() {
-        return (Globals.prefs.getBoolean(JabRefPreferences.MARK_IMPORTED_ENTRIES)
-                && (Globals.prefs.getBoolean(JabRefPreferences.USE_OWNER)
-                        || Globals.prefs.getBoolean(JabRefPreferences.USE_TIME_STAMP)));
+        return Globals.prefs.getBoolean(JabRefPreferences.MARK_IMPORTED_ENTRIES);
```
koppor: how much effort would it be to create a test case? :innocent: 
- set marking in the prefs to true
- Import an entry
- check if the entry is marked

CHANGELOG.md
oscargus: I do actually not think so. It was introduced March 25, so it is (should be) in both 3.3 and 3.4...

src/main/java/net/sf/jabref/gui/EntryMarker.java
oscargus: I considered that, but I'm not really sure how to import something from code...

tobiasdiez:
LGTM :+1:

src/main/java/net/sf/jabref/gui/EntryMarker.java
lenhard: @oscargus: Have a look at `test.java.net.sf.jabref.importer.*`. Thanks to @JabRef/stupro, we have a lot of tests for the importers. Ideally, you should just be able to extend one of the existing importer tests.

oscargus:
I merge this in and try to add a test later. Yes, really.

PR #1560, opened 2016-07-08
Moved final methods from Util

oscargus:
Final methods moved from Util and the class finally removed. I'm sure the DbImportAction can be improved by using `Thread.start()`, but this seems like a generic problem for JabRef.

### DIFF
--- a/src/main/java/net/sf/jabref/gui/groups/WarnAssignmentSideEffects.java
+++ b/src/main/java/net/sf/jabref/gui/groups/WarnAssignmentSideEffects.java
@@ -1,24 +1,4 @@
-/*  Copyright (C) 2003-2016 JabRef contributors.
-    Copyright (C) 2015 Oliver Kopp
-
-    This program is free software; you can redistribute it and/or modify
-    it under the terms of the GNU General Public License as published by
-    the Free Software Foundation; either version 2 of the License, or
-    (at your option) any later version.
-
-    This program is distributed in the hope that it will be useful,
-    but WITHOUT ANY WARRANTY; without even the implied warranty of
-    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-    GNU General Public License for more details.
-
-    You should have received a copy of the GNU General Public License along
-    with this program; if not, write to the Free Software Foundation, Inc.,
-    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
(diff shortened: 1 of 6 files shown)


tobiasdiez:
LGTM :+1:

src/main/java/net/sf/jabref/gui/groups/WarnAssignmentSideEffects.java
```diff
@@ -102,7 +45,7 @@ public static boolean warnAssignmentSideEffects(List<AbstractGroup> groups, Comp
         if (affectedFields.isEmpty()) {
             return true; // no side effects
         }
-
+
```
koppor: Could you change your commit that these added spaces go away? :innocent:

src/main/java/net/sf/jabref/gui/groups/WarnAssignmentSideEffects.java
```diff
@@ -116,7 +59,7 @@ public static boolean warnAssignmentSideEffects(List<AbstractGroup> groups, Comp
         int choice = JOptionPane.showConfirmDialog(parent, message, Localization.lang("Warning"),
                 JOptionPane.YES_NO_OPTION, JOptionPane.WARNING_MESSAGE);
         return choice != JOptionPane.NO_OPTION;
-
+
```
koppor: see above :innocent:

src/main/java/net/sf/jabref/sql/importer/DbImportAction.java
```diff
@@ -196,4 +197,29 @@ public void update() {
         frame.output(Localization.lang("Imported %0 databases successfully", Integer.toString(databases.size())));
     }
 
+    /**
+     * Run using Spin features to put each method on the correct thread.
+     *
+     * @throws Throwable
```
koppor: This can be removed as it does not provide any additional information.

OK for me to move such a util class to here. As The DbImportAction will be gone soon (see #1451), this dependency on spin will go, too. :tada:

src/main/java/net/sf/jabref/gui/groups/WarnAssignmentSideEffects.java
oscargus: I do not understand why they show up... If you see some of my other PRs
these spaces are actually removed...

I'll fix these things and merge.

PR #1562, opened 2016-07-08
Fixed number of entries fetched from Springer

oscargus:
Earlier one less entry than requested was fetched when searching Springer. Added ChangeLog entry in case someone else detects this...

### DIFF
--- a/src/main/java/net/sf/jabref/importer/fileformat/JSONEntryParser.java
+++ b/src/main/java/net/sf/jabref/importer/fileformat/JSONEntryParser.java
@@ -233,10 +233,11 @@ public class JSONEntryParser {
         }
 
         // Clean up abstract (often starting with Abstract)
-        String abstr = entry.getField("abstract");
-        if ((abstr != null) && abstr.startsWith("Abstract")) {
-            entry.setField("abstract", abstr.substring(8));
-        }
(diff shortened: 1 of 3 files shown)


src/main/java/net/sf/jabref/importer/fileformat/JSONEntryParser.java
```diff
@@ -233,10 +233,11 @@ public static BibEntry parseSpringerJSONtoBibtex(JSONObject springerJsonEntry) {
         }
 
         // Clean up abstract (often starting with Abstract)
-        String abstr = entry.getField("abstract");
-        if ((abstr != null) && abstr.startsWith("Abstract")) {
-            entry.setField("abstract", abstr.substring(8));
-        }
+        entry.getFieldOptional("abstract").ifPresent(abstr -> {
```
tobiasdiez: rename abstr -> abstract

tobiasdiez:
LGTM :+1: just some very small remark. You can merge it in afterwards.

src/main/java/net/sf/jabref/importer/fileformat/JSONEntryParser.java
oscargus: abstract is a Java keyword. ;-)

Issue #1338, opened 2016-05-02
Feature request: Clickable crossref field.


domhardt:
Click on BibTeX-Key in crossref field should lead to parent. Parent should indicate existing children (BibTeX-Keys, also clickable).

oscargus:
There is now a PR in #1563 which adds a button in the entry editor for the crossref field and that the crossref column is clickable. Regarding linking between entries in general this is discussed in https://github.com/koppor/jabref/issues/14 .

PR #1563, opened 2016-07-08
Added button to crossref entry editor field to select parent entry

oscargus:
Based on the idea in #1338 a button is added in the crossref entry editor field to select the parent entry in the main table. It is not exactly the same thing as suggested in #1338 though.

### DIFF
--- a/src/main/java/net/sf/jabref/gui/entryeditor/FieldExtraComponents.java
+++ b/src/main/java/net/sf/jabref/gui/entryeditor/FieldExtraComponents.java
@@ -507,7 +507,7 @@ public class FieldExtraComponents {
             }
 
             private void checkValidKey() {
-                if (panel.getDatabase().containsEntryWithKey(crossref.getText())) {
+                if (panel.getDatabase().getEntryByKey(crossref.getText()) != null) {
                     button.setEnabled(true);
                 } else {
                     button.setEnabled(false);
--- a/src/main/java/net/sf/jabref/gui/entryeditor/FieldExtraComponents.java
+++ b/src/main/java/net/sf/jabref/gui/entryeditor/FieldExtraComponents.java
@@ -134,7 +134,7 @@ public class FieldExtraComponents {
         // enable/disable button
         JTextComponent url = (JTextComponent) fieldEditor;
 
-        DocumentListener documentListener = new DocumentListener() {
+        url.getDocument().addDocumentListener(new DocumentListener() {
             @Override
             public void changedUpdate(DocumentEvent documentEvent) {
                 checkUrl();
@@ -157,8 +157,7 @@ public class FieldExtraComponents {
                     button.setEnabled(false);
                 }
             }
-        };
-        url.getDocument().addDocumentListener(documentListener);
+        });
 
         return Optional.of(controls);
     }
@@ -205,7 +204,7 @@ public class FieldExtraComponents {
         // enable/disable button
         JTextComponent doi = (JTextComponent) fieldEditor;
 
-        DocumentListener documentListener = new DocumentListener() {
+        doi.getDocument().addDocumentListener(new DocumentListener() {
             @Override
             public void changedUpdate(DocumentEvent documentEvent) {
                 checkDoi();
@@ -231,8 +230,7 @@ public class FieldExtraComponents {
                     fetchButton.setEnabled(false);
                 }
             }
-        };
-        doi.getDocument().addDocumentListener(documentListener);
+        });
 
         return Optional.of(controls);
     }
@@ -476,9 +474,9 @@ public class FieldExtraComponents {
     }
 
     /**
-     * Return a button which sets the owner if the field for fields with EXTRA_SET_OWNER
-     * @param fieldEditor
-     * @param storeFieldAction
+     * Return a button which allows to go to the parent entry of the crossref field
+     * @param fieldEditor The FieldEditor component to get the entry key from
+     * @param panel The current BasePanel
      * @return
      */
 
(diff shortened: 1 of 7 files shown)


oscargus:
Well, now it is more like the actual feature request. The reverse linking is not obvious how to implement graphically and relates to https://github.com/koppor/jabref/issues/14 as well.

src/main/java/net/sf/jabref/gui/entryeditor/FieldExtraComponents.java
```diff
@@ -474,4 +474,55 @@ public void mouseClicked(MouseEvent e) {
         return Optional.of(gender);
 
     }
+
+    /**
+     * Return a button which sets the owner if the field for fields with EXTRA_SET_OWNER
```
tobiasdiez: I think the comment is wrong here.

src/main/java/net/sf/jabref/gui/entryeditor/FieldExtraComponents.java
```diff
@@ -474,4 +474,55 @@ public void mouseClicked(MouseEvent e) {
...
+     * @param fieldEditor
+     * @param storeFieldAction
+     * @return
+     */
+
+    public static Optional<JComponent> getCrossrefExtraComponent(FieldEditor fieldEditor,
+            BasePanel panel) {
+        JButton button = new JButton(Localization.lang("Select"));
```
tobiasdiez: Rename to "Go to parent"?

src/main/java/net/sf/jabref/gui/entryeditor/FieldExtraComponents.java
```diff
@@ -474,4 +474,55 @@ public void mouseClicked(MouseEvent e) {
...
+            panel.highlightEntry(panel.getDatabase().getEntryByKey(crossref.getText()));
+        });
+
+        // enable/disable button
+
+        DocumentListener documentListener = new DocumentListener() {
+
+            @Override
+            public void changedUpdate(DocumentEvent documentEvent) {
+                checkUrl();
```
tobiasdiez: rename checkUrl

src/main/java/net/sf/jabref/gui/entryeditor/FieldExtraComponents.java
```diff
@@ -474,4 +474,55 @@ public void mouseClicked(MouseEvent e) {
...
+            private void checkUrl() {
+                if (panel.getDatabase().containsEntryWithKey(crossref.getText())) {
+                    button.setEnabled(true);
+                } else {
+                    button.setEnabled(false);
+                }
+            }
+        };
+
+        crossref.getDocument().addDocumentListener(documentListener);
```
tobiasdiez: Inline documentListener?

src/main/java/net/sf/jabref/model/database/BibDatabase.java
```diff
@@ -120,6 +120,20 @@ public boolean containsEntryWithId(String id) {
         return internalIDs.contains(id);
     }
 
+    /**
+     * Returns whether an entry with the given bibtex key exists.
+     */
+    public synchronized boolean containsEntryWithKey(String key) {
+        int keyHash = key.hashCode(); // key hash for better performance
```
tobiasdiez: just use getEntryByKey(key) != null.
By the way, I would welcome if you could change the return type of getEntryByKey to optional.

src/main/java/net/sf/jabref/gui/entryeditor/FieldExtraComponents.java
oscargus: Yes, a bit of copy paste coding, like checkURL...

I'll sort it out. "Go to parent" may be a better name, but I wasn't sure
how well defined "parent" actually is here and somehow what happens is that
the parent entry is selected. And couldn't figure out a good translation to
Swedish...

Another issue is that the button is enabled when the entry editor is
enabled. At least for new entries... Once something is typed it works
though. Guess one should try to trigger the documentListener once, but
didn't figure out how...

src/main/java/net/sf/jabref/gui/entryeditor/FieldExtraComponents.java
oscargus: My code checking tool already complains that the anonymous class is too
long. :-)

But, yes, could indeed be done.

src/main/java/net/sf/jabref/model/database/BibDatabase.java
oscargus: After checking that code I decided to write this method. I think it will be
noticeable on large databases. (getEntryByName always goes through the
whole database and returns the last entry with that key).

I was thinking about the optional thing though. :-)

src/main/java/net/sf/jabref/gui/entryeditor/FieldExtraComponents.java
Siedlerchr: In that step you could make a lamdba from it...

oscargus: I'm not sure if I understand how. As I recall, lambda are only usable if there is a single method that is called? (Which there is, but through three different overridden methods.)

oscargus:
I kept the button name and didn't understand the lambda thing. Other than that, the comments are honored.

Siedlerchr:
@oscargus  Forget about the lambda thing. I oversaw that it contains more than one method :confused:

src/main/java/net/sf/jabref/model/database/BibDatabase.java
tobiasdiez: Well, I think it is safe to change getEntryByKey to return the first entry with the key (instead of the last one). And if you at changing this method, then replace the comparison of hashcodes with real equals (like in getEntriesByKey)

oscargus: MIssed this comment, but getEntryByKeyOptional was introduced in #1568. It is an intermediate solution before figuring out the final conversion problem.

Makes sense to return the first, so I'll implement that later on tonight.

oscargus: Two-stage solution. Removed `containsEntryWithKey` here and will do the other changes in #1568.

PR #1564, opened 2016-07-08
Corrected month value when using drop-down in entry editor

oscargus:
Using the drop down resulted in strings like `##jan##`. Now, the correct `#jan#` is obtained.

### DIFF
--- a/src/main/java/net/sf/jabref/gui/entryeditor/FieldExtraComponents.java
+++ b/src/main/java/net/sf/jabref/gui/entryeditor/FieldExtraComponents.java
@@ -276,7 +276,7 @@ public class FieldExtraComponents {
                 if (type == BibDatabaseMode.BIBLATEX) {
                     fieldEditor.setText(String.valueOf(monthnumber));
                 } else {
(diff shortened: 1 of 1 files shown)


Siedlerchr:
It would be nice, if you could take a look at the date format which gets entered, too:
For example, when I select a date from the calendar, it gets entered as :  `2016.07.09`
OS lang is German, but JabRef is english. 
The ideal solution would be that the date is entered initially in ISO format, yyyy-MM-dd

tobiasdiez:
LGTM :+1: but it would be nice if you could also fix the problem @Siedlerchr mentioned.

oscargus:
I'll see what I can do @Siedlerchr Turned out I didn't have the date field
enabled in my version and I didn't prioritize getting it there.

What should the correct way be? Based on system preferences? Isn't there a
preference setting for it? (Writing on my phone...)

Will be a separate PR though.

tobiasdiez:
@oscargus always in iso-format yyyy-mm-dd since this as what biblatex expects.

oscargus:
Ah, that's right. I just remembered the part about system language. :-)

Siedlerchr:
@oscargus  Just change your db to biblatex and it will the there ;)

PR #1565, opened 2016-07-08
Added information in the tool tip for undo and redo and disabled when…

oscargus:
A tool tip text is shown describing the undo/redo action. Also, the undo/redo buttons are enabled/disabled when there are things to undo/redo.

capture3 2

Current problem is that it doesn't work on startup (both undo and redo is enabled as before), but as soon as one edit is done it works just fine.

… no undo/redo is available


simonharrer:
Shouldn't this be queriable from the undo/redo action? So each action that can be made undo/redo should have an identifier describing its action.

oscargus:
You mean basically move the `updateText` logic to a method for the UndoAction and RedoAction classes and have a single call in `JabRefFrame` like `updateUndoAndRedoTexts` which calls these to methods?

Yes, that might be a better idea. However, I'm not really sure if one also needs to change the corresponding classes in the preamble editor and string dialog. (Nor, actually, how it works there now...)

oscargus:
It can indeed be queried. However, the actual undo/redo classes are in BasePanel but the actions associated to menu and toolbar are defined in JabRefFrame, providing the undo/redo class instance. So, the undo/redo class have no idea about its tooltip.

When I tried it out now, it appeared much more flaky, completely ignoring "change field" edits. No idea why this happens as it worked well earlier. This happens also in master so trying to figure out what might have changed... Most (all?) other types of edits can be undone, but "change field"...

oscargus:
@JabRef/developers : I'm not opening a new issue, but can you please check to see if you can undo field changes in the current master? It seems like it stopped working sometime between creating this PR and now. From debugging I can tell that `addEdit()` is indeed called as it should, but somehow the edits are not stored. This appears to only happen for "change field" edits as the other ones I've tried are OK.

I've looked through the recent commits multiple times to try to figure out what have changed, but at no luck...

oscargus:
@JabRef/developers: Just after posting, I think I found the issue... No emergency...

oscargus:
The problem was actually discovered because of this and just having worked with #1575 (which I will fix later) and fixed in #1576. Totally unrelated to any recent commit.

tobiasdiez:
I don't like that the UndoManager has to know about the frame. Thus I would propose to use "inversion of control" here. That is, the CountingUndoManager throws an event (using the new EventBus) whenever the last undoable action changes. In the frame you can then subscribe to this event and change the display accordingly. In the event you can store the undoable event (of type `UndoableEdit`) and then simply call `event.getUndoPresentationName()` to get the description.

Using events in a similar manner, you can also move the `panel.updateEntryEditorIfShowing();` callings from the undo manager to the frame.

oscargus:
OK, EventBus is introduced and I think the structure is clearly better. It would now be nice if there are more "advanced" `getPresentationName()` for the Undoable actions with content from the actual action. Not in this PR though.

oscargus:
There is a minor flaw which may not be so easy to solve (and is not that serious): if one makes an edit in the entry editor and directly selects a different tab, the undo/redo state is from the previous tab. Also, when trying to undo in the new tab, the state is not updated as an Exception is thrown and the code never gets to the place where the update is done.

oscargus:
Rebased. Anything else here?

src/main/java/net/sf/jabref/logic/undo/UndoRedoEvent.java
```diff
@@ -0,0 +1,10 @@
+package net.sf.jabref.logic.undo;
+
+
+public class UndoRedoEvent extends AddUndoEvent {
```
simonharrer: My head is spinning by thinking about UndoRedoEvents, as undo and redo should cancel each other out, leaving only a event? Can you please document this? And the AddUndoEvent als well? What should be described in these events? When are they fired?

src/main/java/net/sf/jabref/gui/JabRefFrame.java
```diff
@@ -641,6 +644,7 @@ public void windowClosing(WindowEvent e) {
             bp.updateSearchManager();
             // Set correct enabled state for Back and Forward actions:
             bp.setBackAndForwardEnabledState();
+            bp.getUndoManager().postUndoRedoEvent();
```
tobiasdiez: Why is this necessary?

src/main/java/net/sf/jabref/gui/JabRefFrame.java
```diff
@@ -2253,4 +2261,28 @@ public void setPreviewToggle(boolean enabled) {
...
         return pushApplications;
     }
+
+
+    private class UndoRedoEventManager {
+
+        @Subscribe
+        public void listen(UndoRedoEvent event) {
+            updateTexts(event);
+            JabRefFrame.this.getCurrentBasePanel().updateEntryEditorIfShowing();
```
tobiasdiez: Is this really necessary? I think the entry editor listens for changes to the displayed entry, right?

src/main/java/net/sf/jabref/gui/undo/CountingUndoManager.java
```diff
@@ -60,4 +66,28 @@ public synchronized void markUnchanged() {
     public synchronized boolean hasChanged() {
         return (current != unchangedPoint);
     }
+
+
+    public void registerListener(Object object) {
+        this.eventBus.register(object);
+        postUndoRedoEvent(); // Send event to trigger changess
```
tobiasdiez: only one `s`

src/main/java/net/sf/jabref/logic/undo/AddUndoEvent.java
```diff
@@ -0,0 +1,18 @@
+package net.sf.jabref.logic.undo;
+
+
+public class AddUndoEvent {
+
+    public final boolean canUndo;
```
tobiasdiez: Hide these fields behind getters (no setters). And rename `undoText` to `undoDescription`

src/main/java/net/sf/jabref/logic/undo/UndoRedoEvent.java
tobiasdiez: In addition, I also don't get why UndoRedo derives from AddUndo.

src/main/java/net/sf/jabref/gui/JabRefFrame.java
oscargus: To set the correct enable/disable state on the panel. Maybe a bit hacky,
but the alternative is to define another event type.

src/main/java/net/sf/jabref/gui/JabRefFrame.java
oscargus: Didn't try without it. Earlier it resided in the undo manager.

src/main/java/net/sf/jabref/logic/undo/UndoRedoEvent.java
oscargus: I'll add comments. Quickly: AddUndoEvent when something is added to the
undo manager (update actions), UndoRedoEvent when something is
undone/redone (update actions and possibly entry editor).

oscargus: Basically to avoid copying code as they have exactly the same content but
two different types are needed. An alternative might be to add the type in
the event but that seemed worse.

matthiasgeiger: Just duplicate the code and do not introduce inheritance if it is semantically inadequate.

Maybe the meaning of "AddUndoEvent" gets clearer if it is named "AddUndoableActionEvent"?

simonharrer: :+1: for code duplication in that case.
:+1: for "AddUndoableActionEvent", but still, is this just not something that is happening automatically because of other events? So do we really need this event as such? To add an undoable item to the undo manager seems a little bit too low from a granularity point of view. We should add the event that a save was triggered, and this would then be also the "AddUndoableActionEvent" as it was a save event. Does this make sense?
:+1: for "UndoOrRedoEvent"

oscargus: Another reason for inheriting was that there could be a shared function for
updating the text. I have copied that now though.

The reason that such an event is needed is that it is triggering the tool
tip changes. That is all it does.

Will push tomorrow. :-)

tobiasdiez: You can create a common base event like `Undo(Manager?)ChangedEvent` and let both events derive from this new class.

src/main/java/net/sf/jabref/gui/JabRefFrame.java
tobiasdiez: Sorry, I still don't get it. 
So I close the window, why should the undo manager care about it??

src/main/java/net/sf/jabref/gui/undo/UndoableChangeType.java
```diff
@@ -35,12 +36,19 @@ public UndoableChangeType(BibEntry entry, String oldType, String newType) {
 
     @Override
     public String getUndoPresentationName() {
-        return "Undo: change type";
+        return Localization.lang("Undo") + ": " + getPresentationName();
```
tobiasdiez: It looks to me that a lot of events now have the same `getUndo/RedoPresentationName`. Can this be pushed to the parent class AbstractUndoableEdit ?

oscargus: Yes, that should indeed be possible. Didn't think of it. Earlier only some had a `getPresentationName` but based on the structure of the code, it made more sense to only change any significant text in one position.

oscargus: Except that `AbstractUndoableEdit` is in `javax.swing.undo`.

src/main/java/net/sf/jabref/gui/JabRefFrame.java
oscargus: The undoManager has the event bus. Any other suggestions for the location
of the undo event bus?

oscargus:
I improved the information a bit:

capture13

src/main/java/net/sf/jabref/gui/JabRefFrame.java
oscargus: Oh, it is actually not in `windowClosing()`. It is in the `tabbedPane`  changeListener in `init()`.

PR #1568, opened 2016-07-11
Added getEntryByKeyOptional and used it. Changed result of getFieldOr…

oscargus:
* Added `getEntryByKeyOptional` and almost managed to completely remove the use of `getEntryByKey`.
* Changed the return type of `getFieldOrAlias` to be `Optional<String>`
* Some more conversions of `getField` to `getFieldOptional`

### DIFF
--- a/src/main/java/net/sf/jabref/gui/maintable/MainTableColumn.java
+++ b/src/main/java/net/sf/jabref/gui/maintable/MainTableColumn.java
@@ -112,9 +112,13 @@ public class MainTableColumn {
             if (field.equals(BibEntry.TYPE_HEADER)) {
                 content = EntryUtil.capitalizeFirst(entry.getType());
             } else {
-                content = entry.getFieldOrAlias(field);
-                if (database.isPresent() && "Author".equalsIgnoreCase(columnName) && (content != null)) {
-                    content = database.get().resolveForStrings(content);
+                Optional<String> newContent = entry.getFieldOrAlias(field);
+                if (newContent.isPresent()) {
+                    if (database.isPresent() && "Author".equalsIgnoreCase(columnName)) {
+                        content = database.get().resolveForStrings(newContent.get());
+                    } else {
+                        content = newContent.get();
+                    }
                 }
             }
             if (content != null) {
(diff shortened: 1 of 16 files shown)


oscargus:
Since I managed to convert all calls I renamed `getEntryByKeyOptional` to `getEntryByKey` and removed the original `getEntryByKey`.

src/test/java/net/sf/jabref/model/entry/BibEntryTests.java
```diff
@@ -43,7 +43,7 @@ public void testDefaultConstructor() {
         // we have to use `getType("misc")` in the case of biblatex mode
         Assert.assertEquals("misc", entry.getType());
         Assert.assertNotNull(entry.getId());
-        Assert.assertNull(entry.getField("author"));
+        Assert.assertFalse(entry.getFieldOptional("author").isPresent());
```
tobiasdiez: Better `assertEquals(Optional.empty(), entry.getField(....))` in this way one gets a better failure message. Same applies to the other tests, i.e. `assertEquals(Optional.of("2003-03"), emptyEntry.getFieldOrAlias("date"))` below.

oscargus: Makes sense.

src/main/java/net/sf/jabref/gui/maintable/MainTableColumn.java
```diff
@@ -112,9 +112,13 @@ public Object getColumnValue(BibEntry entry) {
             if (field.equals(BibEntry.TYPE_HEADER)) {
                 content = EntryUtil.capitalizeFirst(entry.getType());
             } else {
-                content = entry.getFieldOrAlias(field);
-                if (database.isPresent() && "Author".equalsIgnoreCase(columnName) && (content != null)) {
-                    content = database.get().resolveForStrings(content);
+                Optional<String> newContent = entry.getFieldOrAlias(field);
+                if (newContent.isPresent()) {
```
simonharrer: use ifPresent

simonharrer:
LGTM

PR #1569, opened 2016-07-11
Removed support for optinally following crossrefs since it wasn't used

oscargus:
Initialized to true and set to true...


oscargus:
Merged this in so that there would be no later merge conflicts with #1568.

PR #1570, opened 2016-07-11
Some more testing of BibDatabase, primarily related to DuplicationChe…

oscargus:
A bit higher coverage of `BibDatabase` and `DuplicationChecker`.

### DIFF
--- a/src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
+++ b/src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
@@ -273,4 +273,48 @@ public class BibDatabaseTest {
         assertEquals(database.getNumberOfKeyOccurrences("AAA"), 1);
     }
 
+    @Test
+    public void circularStringResolving() {
+        BibtexString string = new BibtexString(IdGenerator.next(), "AAA", "#BBB#");
+        database.addString(string);
+        string = new BibtexString(IdGenerator.next(), "BBB", "#AAA#");
+        database.addString(string);
+        assertEquals(database.resolveForStrings("#AAA#"), "AAA");
+        assertEquals(database.resolveForStrings("#BBB#"), "BBB");
+    }
+
+    @Test
+    public void circularStringResolvingLongerCycle() {
+        BibtexString string = new BibtexString(IdGenerator.next(), "AAA", "#BBB#");
+        database.addString(string);
+        string = new BibtexString(IdGenerator.next(), "BBB", "#CCC#");
+        database.addString(string);
+        string = new BibtexString(IdGenerator.next(), "CCC", "#DDD#");
+        database.addString(string);
+        string = new BibtexString(IdGenerator.next(), "DDD", "#AAA#");
+        database.addString(string);
+        assertEquals(database.resolveForStrings("#AAA#"), "AAA");
+        assertEquals(database.resolveForStrings("#BBB#"), "BBB");
+        assertEquals(database.resolveForStrings("#CCC#"), "CCC");
+        assertEquals(database.resolveForStrings("#DDD#"), "DDD");
+    }
+
+    @Test
+    public void resolveForStringsMonth() {
+        assertEquals(database.resolveForStrings("#jan#"), "January");
+    }
+
+    @Test
+    public void resolveForStringsSurroundingContent() {
+        BibtexString string = new BibtexString(IdGenerator.next(), "AAA", "aaa");
+        database.addString(string);
+        assertEquals(database.resolveForStrings("aa#AAA#AAA"), "aaaaaAAA");
+    }
+
+    @Test
+    public void resolveForStringsOddHashMarkAtTheEnd() {
+        BibtexString string = new BibtexString(IdGenerator.next(), "AAA", "aaa");
+        database.addString(string);
+        assertEquals(database.resolveForStrings("AAA#AAA#AAA#"), "AAAaaaAAA#");
+    }
 }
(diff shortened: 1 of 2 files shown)


Siedlerchr:
LGTM! You could add one more test for checking duplication for empty bibtex keys, e.g. two entries with no bibtex keys.

src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
```diff
@@ -197,4 +197,124 @@ public void changingEntryPostsChangeEntryEvent() {
...
+        database.insertEntry(entry);
+        assertEquals(database.getNumberOfKeyOccurrences("AAA"), 1);
+    }
+
+    @Test
+    public void correctKeyCountTwo() {
+        BibEntry entry = new BibEntry();
+        entry.setCiteKey("AAA");
+        database.insertEntry(entry);
+        entry = new BibEntry();
```
tobiasdiez: Create a new `BibEntry secondEntry` instead of reusing the first one.

src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
```diff
@@ -197,4 +197,124 @@ public void changingEntryPostsChangeEntryEvent() {
...
+        assertEquals(database.getNumberOfKeyOccurrences("AAA"), 2);
+    }
+
+    @Test
+    public void setCiteKeySameKeySameEntry() {
+        BibEntry entry = new BibEntry();
+        entry.setCiteKey("AAA");
+        database.insertEntry(entry);
+        assertFalse(database.setCiteKeyForEntry(entry, "AAA"));
+        assertEquals(database.getNumberOfKeyOccurrences("AAA"), 1);
```
tobiasdiez: Remove second assert

src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
```diff
@@ -197,4 +197,124 @@ public void changingEntryPostsChangeEntryEvent() {
...
+        database.insertEntry(entry);
+        assertEquals(database.getNumberOfKeyOccurrences("AAA"), 2);
+    }
+
+    @Test
+    public void setCiteKeySameKeySameEntry() {
+        BibEntry entry = new BibEntry();
+        entry.setCiteKey("AAA");
+        database.insertEntry(entry);
+        assertFalse(database.setCiteKeyForEntry(entry, "AAA"));
```
tobiasdiez: I would remove the setCiteKeyForEntry method. Half of its users does not need the return value and it definitely not clear from the name of the method that the return value gives you some information about duplicates.

src/test/java/net/sf/jabref/model/entry/MonthUtilTest.java
```diff
@@ -75,5 +75,6 @@ public void testToMonthNumber() {
         Assert.assertEquals(-1, MonthUtil.getMonth("#test#").index);
         Assert.assertEquals(-1, MonthUtil.getMonth("").index);
         Assert.assertFalse(MonthUtil.getMonth("8,").isValid());
+        Assert.assertTrue(MonthUtil.getMonth("jan").isValid());
```
tobiasdiez: New test method pls.

src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
```diff
@@ -197,4 +197,124 @@ public void changingEntryPostsChangeEntryEvent() {
 
         assertEquals(entry, tel.getBibEntry());
     }
+
+    @Test
+    public void correctKeyCountOne() {
+        BibEntry entry = new BibEntry();
```
tobiasdiez: Init the entry with key `AAA` in the `@Before` method so that you can reuse it in every test.

src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
```diff
@@ -197,4 +197,124 @@ public void changingEntryPostsChangeEntryEvent() {
...
+        assertFalse(database.setCiteKeyForEntry(entry, "AAA"));
+        assertEquals(database.getNumberOfKeyOccurrences("AAA"), 1);
+    }
+
+    @Test
+    public void setCiteKeyRemoveKey() {
+        BibEntry entry = new BibEntry();
+        entry.setCiteKey("AAA");
+        database.insertEntry(entry);
+        assertFalse(database.setCiteKeyForEntry(entry, null));
```
tobiasdiez: Please test only one aspect of the method per test, i.e. checking the return value and that the cite key was changed are two independent tests. This applies also to some of the test methods below.

src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
```diff
@@ -197,4 +197,124 @@ public void changingEntryPostsChangeEntryEvent() {
...
+        assertEquals(database.getNumberOfKeyOccurrences("BBB"), 1);
+    }
+
+
+    @Test
+    public void setCiteKeySameKeyDifferentEntries() {
+        BibEntry entry = new BibEntry();
+        entry.setCiteKey("AAA");
+        database.insertEntry(entry);
+        entry = new BibEntry();
```
tobiasdiez: Don't reuse entry.

tobiasdiez:
I have some small remarks. It would be nice if you could fix them although this was already merged in. Thanks.

Issue #1356, opened 2016-05-06
Add HTML to Unicode formatter


tobiasdiez:
Add HTML to Unicode formatter in cleanup / save actions.

@oscargus Something for you?

lenhard:
Implemented by #1572

PR #1572, opened 2016-07-11
Implement 1356

lenhard:
Implements  #1356 and provides a formatter that converts HTML to Unicode (converting characters and eliminating tags)


simonharrer:
LGTM

src/main/resources/l10n/JabRef_de.properties
```diff
@@ -2279,6 +2279,7 @@ Changes_the_first_letter_of_all_words_to_capital_case_and_the_remaining_letters_
 Cleans_up_LaTeX_code.=Räumt_LaTeX-Code_auf.
 
 Converts_HTML_code_to_LaTeX_code.=Konvertiere_HTML-Code_in_LaTeX-Code.
+Converts_HTML_code_to_Unicode.= Konvertiere_HTML-Code_in_Unicode.
```
Siedlerchr: Is there a space after the = sign?

stefan-kolb:
LGTM :+1:

src/test/java/net/sf/jabref/logic/formatter/bibtexfields/HtmlToUnicodeFormatterTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+import static org.junit.Assert.assertEquals;
+
+public class HtmlToUnicodeFormatterTest {
+
+    private final HtmlToUnicodeFormatter formatter = new HtmlToUnicodeFormatter();
+
+    @Before
+    public void setUp() throws Exception {
+        Globals.prefs = JabRefPreferences.getInstance();
```
tobiasdiez: It is best practice to init the class under test in the `@Before` method so that every test method gets a fresh class, i.e move the `= new HtmlToUnicodeFormatter` part to this method.

src/test/java/net/sf/jabref/logic/formatter/bibtexfields/HtmlToUnicodeFormatterTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+import net.sf.jabref.Globals;
+import net.sf.jabref.JabRefPreferences;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+public class HtmlToUnicodeFormatterTest {
```
tobiasdiez: Add this class to the parametrized FormatterTest class (not sure about the exact name)

src/test/java/net/sf/jabref/logic/formatter/bibtexfields/HtmlToUnicodeFormatterTest.java
```diff
@@ -0,0 +1,59 @@
...
+        assertEquals("í", formatter.format("i&#x301;"));
+    }
+
+    @Test
+    public void testBasic() {
+        assertEquals("aaa", formatter.format("aaa"));
+    }
+
+    @Test
+    public void testHTML() {
```
tobiasdiez: I know I'm pedantic, but please create 4 tests for this.

src/main/java/net/sf/jabref/logic/formatter/bibtexfields/HtmlToUnicodeFormatter.java
```diff
@@ -0,0 +1,36 @@
+package net.sf.jabref.logic.formatter.bibtexfields;
+
+import net.sf.jabref.logic.formatter.Formatter;
+import net.sf.jabref.logic.l10n.Localization;
+import net.sf.jabref.logic.layout.LayoutFormatter;
+
+import org.apache.commons.lang3.StringEscapeUtils;
+
+public class HtmlToUnicodeFormatter implements LayoutFormatter, Formatter {
```
tobiasdiez: We should also add a corresponding entry in the help file (for LayoutFormatter)

src/test/java/net/sf/jabref/logic/formatter/bibtexfields/HtmlToUnicodeFormatterTest.java
```diff
@@ -0,0 +1,59 @@
...
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+public class HtmlToUnicodeFormatterTest {
+
+    private final HtmlToUnicodeFormatter formatter = new HtmlToUnicodeFormatter();
+
+    @Before
+    public void setUp() throws Exception {
```
tobiasdiez: Is the prefs statement really necessary?

tobiasdiez:
I have only a few small remarks. :+1: for merge

src/main/resources/l10n/JabRef_de.properties
lenhard: Indeed! Thanks for spotting!

src/test/java/net/sf/jabref/logic/formatter/bibtexfields/HtmlToUnicodeFormatterTest.java
lenhard: You are correct. That comes from a copy paste from the other formatter tests, which all share this problem. I will fix all of them in this PR.

src/test/java/net/sf/jabref/logic/formatter/bibtexfields/HtmlToUnicodeFormatterTest.java
lenhard: Will do. And `FormatterTest` is indeed the correct class name.

src/test/java/net/sf/jabref/logic/formatter/bibtexfields/HtmlToUnicodeFormatterTest.java
lenhard: I'll separate it in two tests. One that tests for the different variants of `ä` (I think it makes sense to bundle those) and one that tests for the greek letter.

src/test/java/net/sf/jabref/logic/formatter/bibtexfields/HtmlToUnicodeFormatterTest.java
lenhard: No, I remove it.

src/main/java/net/sf/jabref/logic/formatter/bibtexfields/HtmlToUnicodeFormatter.java
lenhard: Refs: #8 None of the current formatters are documented and I won't build documentation for all as part of this PR. This has to go separately.

tobiasdiez: I meant http://help.jabref.org/en/CustomExports under `Built-in export formatters` since the class is implementing LayoutFormatter and thus can be used for the exports too, right?

lenhard: Yes, but not a single class of the package `net.sf.jabref.logic.formatter.*` is listed in this section. I don't think a patchwork solution, where I document just one of those, would be a good idea. The whole page desperately needs updating.

PR #1573, opened 2016-07-11
Introduced Optional return values

oscargus:
A few less `return null`. Also removed the unused method `purgeCustomEntryTypes()` from `JabRefPreferences`.


simonharrer:
LGTM

tobiasdiez:
To me too.

PR #1574, opened 2016-07-11
Refactored exporter and split it into gui and logic

oscargus:
Should be a no-brainer but better wait until after 3.5 with merging.


oscargus:
And probably after #1472 as well...

src/main/java/net/sf/jabref/gui/exporter/ExportFileFilter.java
```diff
@@ -13,26 +13,28 @@
     with this program; if not, write to the Free Software Foundation, Inc.,
     51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 */
-package net.sf.jabref.exporter;
+package net.sf.jabref.gui.exporter;
```
lenhard: This class is really independent of GUI code, so it should be in logic and not in gui.

lenhard:
Nice work! I am looking into architectural structure at the moment as well. I would only move one class, as mentioned above.

I think it would be cool if we got the `AutoSaveManager` to work independent of the GUI, but that would mean a rewrite. Thus, it is out of the scope of this PR.

src/main/java/net/sf/jabref/gui/exporter/ExportFileFilter.java
oscargus: `import javax.swing.filechooser.FileFilter;`

Had to restructure ExportFormat a bit because of that darn `FileFilter`...

oscargus:
Thanks! The main purpose was really to see which classes are "easily" testable (=logic) at the moment.

src/main/java/net/sf/jabref/gui/exporter/ExportFileFilter.java
lenhard: I see. You are correct! That means the PR is fine from my point of view.

stefan-kolb:
LGTM :+1:

PR #1581, opened 2016-07-13
Fixed #1264

### LINKED ISSUE #1264: S with caron does not render correctly

oscargus:
I finally got around to understand the reasons for #1264 (and parts of #1464). It turned out that commands starting with c was handled in a special way and that `{\v{S}}` was for unknown reasons used as "line tabulation set".

As a result, not only does JabRef now render `\v{S}` and `\chi` correctly, it also renders combining accents, see screen shot.
capture4

### DIFF
--- a/src/main/java/net/sf/jabref/Globals.java
+++ b/src/main/java/net/sf/jabref/Globals.java
@@ -53,8 +53,8 @@ public class Globals {
     public static final ImportFormatReader IMPORT_FORMAT_READER = new ImportFormatReader();
 
 
-
-    public static final String SPECIAL_COMMAND_CHARS = "\"`^~'=.|"; // "\"`^~'c="
+    // Non-letters which are used to denote accents in LaTeX-commands, e.g., in {\"{a}}
+    public static final String SPECIAL_COMMAND_CHARS = "\"`^~'=.|";
 
     // In the main program, this field is initialized in JabRef.java
     // Each test case initializes this field if required
--- a/src/main/java/net/sf/jabref/Globals.java
+++ b/src/main/java/net/sf/jabref/Globals.java
@@ -54,7 +54,7 @@ public class Globals {
 
 
 
-    public static final String SPECIAL_COMMAND_CHARS = "\"`^~'c=";
+    public static final String SPECIAL_COMMAND_CHARS = "\"`^~'=.|"; // "\"`^~'c="
 
     // In the main program, this field is initialized in JabRef.java
     // Each test case initializes this field if required
--- a/src/main/java/net/sf/jabref/logic/formatter/bibtexfields/UnicodeToLatexFormatter.java
+++ b/src/main/java/net/sf/jabref/logic/formatter/bibtexfields/UnicodeToLatexFormatter.java
@@ -15,8 +15,8 @@
  */
 package net.sf.jabref.logic.formatter.bibtexfields;
 
+import java.util.Map;
 import java.util.Objects;
-import java.util.Set;
 
 import net.sf.jabref.logic.formatter.Formatter;
 import net.sf.jabref.logic.l10n.Localization;
@@ -39,9 +39,9 @@ public class UnicodeToLatexFormatter implements LayoutFormatter, Formatter {
         }
 
         // Standard symbols
-        Set<String> strings = HTMLUnicodeConversionMaps.UNICODE_LATEX_CONVERSION_MAP.keySet();
-        for (String string : strings) {
-            result = result.replace(string, HTMLUnicodeConversionMaps.UNICODE_LATEX_CONVERSION_MAP.get(string));
+        for (Map.Entry<String, String> unicodeLatexPair : HTMLUnicodeConversionMaps.UNICODE_LATEX_CONVERSION_MAP
+                .entrySet()) {
+            result = result.replace(unicodeLatexPair.getKey(), unicodeLatexPair.getValue());
         }
 
         // Combining accents
--- a/src/main/java/net/sf/jabref/logic/formatter/bibtexfields/UnicodeToLatexFormatter.java
+++ b/src/main/java/net/sf/jabref/logic/formatter/bibtexfields/UnicodeToLatexFormatter.java
@@ -39,10 +39,9 @@ public class UnicodeToLatexFormatter implements LayoutFormatter, Formatter {
         }
 
         // Standard symbols
-        Set<Character> chars = HTMLUnicodeConversionMaps.UNICODE_LATEX_CONVERSION_MAP.keySet();
-        for (Character character : chars) {
-            result = result.replace(character.toString(),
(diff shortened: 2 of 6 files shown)


src/main/java/net/sf/jabref/logic/formatter/bibtexfields/UnicodeToLatexFormatter.java
```diff
@@ -39,10 +39,9 @@ public String format(String text) {
         }
 
         // Standard symbols
-        Set<Character> chars = HTMLUnicodeConversionMaps.UNICODE_LATEX_CONVERSION_MAP.keySet();
-        for (Character character : chars) {
-            result = result.replace(character.toString(),
-                    HTMLUnicodeConversionMaps.UNICODE_LATEX_CONVERSION_MAP.get(character));
+        Set<String> strings = HTMLUnicodeConversionMaps.UNICODE_LATEX_CONVERSION_MAP.keySet();
```
Siedlerchr: Please find a more speaking variable name  as "strings" here and below in the loop
e.g unicode_strings or whatever..

stefan-kolb:
@oscargus Do the tests from #1464 pass in your branch?

oscargus:
@stefan-kolb as it was only `\chi` that was failing, I would expect that, yes. The other two did already work. (Right?)

stefan-kolb:
@oscargus It was only the \chi, yes because \c is handled as special character. If my test from the PR pass we can merge my test PR into your branch and then merge your fix :smile:

tobiasdiez:
In general, I would like tests for this PR.

stefan-kolb:
My PR #1464  includes tests for the bug.

oscargus:
@stefan-kolb There is a `\chi` first at the second row of the screen shot. Turned out all commands starting with c didn't work as c was hardcoded into Globals as an accent, so `\chi` was interpreted similar to if it had said `{\c{h}}i` and since `\c{h}` wasn't listed it was simply removed. Now `\c{h}` works as well, giving ḩ .

@tobiasdiez: I was considering that. However, the main feature is adding support for accented unicode characters and they are a hassle to type properly... Nor will they make much sense except lots of comments. For example, spot the difference between ä and ä...

stefan-kolb:
@oscargus Can you just cherry-pick my commit from the PR and you have your first test :smile: 
And then just close my PR afterwards :+1:

oscargus:
I added some tests now.

@stefan-kolb I'm afraid that is a bit out of my comfort/knowledge zone... My solution would be to almost simultaneously merge both PRs. :-)

stefan-kolb:
@oscargus Then just copy my tests and close the PR :smile:

oscargus:
@stefan-kolb That should be within grasp. :-)

src/main/java/net/sf/jabref/Globals.java
```diff
@@ -54,7 +54,7 @@
 
 
 
-    public static final String SPECIAL_COMMAND_CHARS = "\"`^~'c=";
+    public static final String SPECIAL_COMMAND_CHARS = "\"`^~'=.|"; // "\"`^~'c="
```
simonharrer: Can you add a comment describing this string?

simonharrer:
LGTM thanks for the quick solution! :)

stefan-kolb:
LGTM :+1:

src/main/java/net/sf/jabref/logic/layout/format/LatexToUnicodeFormatter.java
```diff
@@ -31,6 +31,7 @@
 public class LatexToUnicodeFormatter implements LayoutFormatter, Formatter {
 
     private static final Map<String, String> CHARS = HTMLUnicodeConversionMaps.LATEX_UNICODE_CONVERSION_MAP;
+    private static final Map<String, String> ACCENTS = HTMLUnicodeConversionMaps.UNICODE_ESCAPED_ACCENTS;
```
tobiasdiez: @JabRef/developers Does java copies here the whole map or just point to the original one? If the map is copied, then one should inline `CHARS` and `ACCENTS`

src/main/java/net/sf/jabref/logic/layout/format/LatexToUnicodeFormatter.java
```diff
@@ -109,7 +111,15 @@ public String format(String inField) {
                         }
                         Object result = LatexToUnicodeFormatter.CHARS.get(command + combody);
 
-                        if (result != null) {
+                        if (result == null) {
+                            // Use combining accents if argument is single character or empty
+                            if (combody.length() <= 1) {
```
tobiasdiez: combody -> commandBody

src/main/java/net/sf/jabref/logic/util/strings/HTMLUnicodeConversionMaps.java
```diff
@@ -765,17 +766,19 @@
                 if (!(aConversionList[0].isEmpty())) {
                     NUMERICAL_LATEX_CONVERSION_MAP.put(Integer.decode(aConversionList[0]), aConversionList[2]);
                     if (Integer.decode(aConversionList[0]) > 128) {
-                        Character c = (char) Integer.decode(aConversionList[0]).intValue();
-                        UNICODE_LATEX_CONVERSION_MAP.put(c, aConversionList[2]);
+                        String s = String.valueOf(Character.toChars(Integer.decode(aConversionList[0])));
```
tobiasdiez: s -> unicodeText or something similar

src/test/java/net/sf/jabref/logic/layout/format/LatexToUnicodeFormatterTest.java
```diff
@@ -61,4 +61,16 @@ public void testEquations() {
     public void formatExample() {
         assertEquals("Mönch", formatter.format(formatter.getExampleInput()));
     }
+
+    @Test
+    public void testEarlierProblems() {
```
tobiasdiez: Please split tests so that the contain only one assertEquals, also `convertChi` is probably a better name (you can also reference the issue to give some context)

tobiasdiez:
LGTM :+1: just some minor remarks about naming variables and tests

src/main/java/net/sf/jabref/logic/layout/format/LatexToUnicodeFormatter.java
Siedlerchr: It doesn't matter, as this is a kind of constant which gets replaced by the containing value on compile time, directly in the byte code.
In general, java is  by value,  it copies just the object reference (pointer).
**http://stackoverflow.com/questions/40480/is-java-pass-by-reference-or-pass-by-value**

PR #1583, opened 2016-07-15
Fixed #1527 and some refactoring/cleanups

### LINKED ISSUE #1527: 'Get BibTeX data from DOI' Removes Marking
If I right-click on a 'marked' entry  in the main table and successfully update the BibTeX data from DOI, the entry is no longer marked.

Version: 3.5dev, commit 87389811baa50f750853e2b1fd1de07e10bf13e3

oscargus:
Fixed #1527 and did some refactoring and cleanups.

### DIFF
--- a/src/main/java/net/sf/jabref/gui/mergeentries/MergeEntryDOIDialog.java
+++ b/src/main/java/net/sf/jabref/gui/mergeentries/MergeEntryDOIDialog.java
@@ -15,6 +15,7 @@
  */
 package net.sf.jabref.gui.mergeentries;
 
+import java.util.Optional;
 import java.util.Set;
 import java.util.TreeSet;
 
@@ -62,7 +63,8 @@ public class MergeEntryDOIDialog extends JDialog {
         this.panel = panel;
 
         if (panel.getSelectedEntries().size() != 1) {
-            JOptionPane.showMessageDialog(panel.frame(), Localization.lang("Select one entry."),
+            JOptionPane.showMessageDialog(panel.frame(),
+                    Localization.lang("This operation requires exactly one item to be selected."),
                     Localization.lang("Merge entry with DOI information"), JOptionPane.INFORMATION_MESSAGE);
             this.dispose();
             return;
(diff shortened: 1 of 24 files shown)


matthiasgeiger:
changes LGTM :+1: 

However, the usage of "magic numbers" for the diff modes should be avoided - but this can also be solved in another PR :wink:

src/main/java/net/sf/jabref/gui/mergeentries/MergeEntryDOIDialog.java
```diff
@@ -162,21 +169,22 @@ private void buttonPressed(String button) {
...
                     edited = true;
                 }
             }
 
-            // Remove fields which are not in the merged entry
+            // Remove fields which are not in the merged entry, unless they are internal fields
             for (String field : originalFields) {
-                if (!jointFields.contains(field)) {
-                    String originalString = originalEntry.getField(field);
+                if (!jointFields.contains(field) && !field.startsWith("__")) {
```
lenhard: Maybe it would make sense to add a method `public static boolean isInternalField(String field)` to `EntryUtil` and use it for the second part of the condition statement?

Just so that the knowledge on the structure of internal fields moves into the model package.

lenhard:
Just a minor comment, otherwise looks good.

Have a thought on my suggestion and decide if you want to implement it. Afterwards rebase and merge.

oscargus:
@matthiasgeiger I agree. However, I do not really get how to map a `JComboBox` and an `enum`...

src/main/java/net/sf/jabref/gui/mergeentries/MergeEntryDOIDialog.java
oscargus: Makes sense.

oscargus: It ended up in `InternalBibtexFields` as that class had similar methods.

matthiasgeiger:
`new JComboBox(EnumType.values())` should work

PR #1586, opened 2016-07-15
Added ISBN integrity checker

oscargus:
ISBN integrity checker

### DIFF
--- a/src/main/java/net/sf/jabref/logic/integrity/ISBNChecker.java
+++ b/src/main/java/net/sf/jabref/logic/integrity/ISBNChecker.java
@@ -1,0 +1,69 @@
+package net.sf.jabref.logic.integrity;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import net.sf.jabref.logic.integrity.IntegrityCheck.Checker;
+import net.sf.jabref.logic.l10n.Localization;
+import net.sf.jabref.model.entry.BibEntry;
+
+
+public class ISBNChecker implements Checker {
+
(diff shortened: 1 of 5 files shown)


src/main/java/net/sf/jabref/logic/integrity/ISBNChecker.java
```diff
@@ -0,0 +1,69 @@
...
+
+    @Override
+    public List<IntegrityMessage> check(BibEntry entry) {
+        if (!entry.hasField("isbn")) {
+            return Collections.emptyList();
+        }
+
+        // Check that the ISSN is on the correct form
+        String isbn = entry.getFieldOptional("isbn").get().trim().replace("-", "");
+        Matcher isbnMatcher = ISBN_PATTERN.matcher(isbn);
```
tobiasdiez: See https://github.com/JabRef/jabref/pull/1590#discussion_r70992658

oscargus: Again, how to handle invalid data? Here it is slightly more obvious as the dashes are ignored, but should one store both version? Some people may prefer the dashes in particular positions...

I do not see this happening in this PR.

simonharrer: I would simply create an ISBN class in the model which receives a String and then can verify whether this String is a valid ISBN or not. Hence, isolate the check logic. I would not care about storing here.

simonharrer:
LGTM just extract the validation logic, as it could be useful elsewhere, and then merge this in.

oscargus:
OK! Without the `toString()` it makes sense (and is easier to cope with than ISSN, as there isn't an "almost correct" ISBN format...).

Issue #1579, opened 2016-07-13
Desired Architecture


lenhard:
Following the discussion from today's devcall, here comes an outline of our desired high-level architecture for JabRef and some data on the status quo. This is a refinement from [our high-level documentation](https://github.com/JabRef/jabref/wiki/High-Level-Documentation).

@JabRef/developers Please have a look, mention anything that you find worth remarking or discussing, and eventually I will update the high level documentation to reflect our desired architecture. Here is a screenshot from my architecture modeling tool for JabRef 3.4 (apologies for overlapping arrows and numbers, there's no possibility to avoid this tool-wise):
![jabref-high-level-architecture](https://cloud.githubusercontent.com/assets/1515701/16802473/144674b8-4901-11e6-967c-9e5aa65c7e9e.jpg)
Arrows mark dependencies and the number of an arrow reflects the number of dependencies. Dashed arrows mark forbidden dependencies, whereas complete arrows mark allowed dependencies. The model depicts the desired architecture, *not* the actual architecture. I manually sorted all classes of jabref into these high-level packages, for instance the `MetaData` class should be in `jabref.model` and that would currently lead to forbidden dependencies from `jabref.model` to `jabref.gui`. Based on my sorting and the actual current source code, the architecture tool computes the number of dependencies between the packages.

The content of `jabref.model`, `jabref.logic`, and `jabref.gui` is described in [our high-level documentation](https://github.com/JabRef/jabref/wiki/High-Level-Documentation). On top of that `jabref.globals` should contain bootstrapping classes and global information (`JabRefMain`, `Defaults`, `Globals`, etc.). `jabref.preferences` should bundle the classes `JabRefPreferences` and `JabRefPreferencesFilter`. `jabref.cli` should bundle the CLI classes (already a top level package). According to the model, the only allowed dependencies in jabref should be:

`jabref.logic` -> `jabref.model`

`jabref.globals` -> `everywhere`

`jabref.gui` -> `everywhere`

`jabref.cli` -> `jabref.preferences`
`jabref.cli` -> `jabref.globals`
`jabref.cli` -> `jabref.logic`
`jabref.cli` -> `jabref.model`

This is an open discussion, any comments are welcome!

oscargus:
Some comments/questions, primarily based on the experience from #1593:

* Where will the `String`constants from `Globals` end up? These are:
``` java
    public static final String FILE_FIELD = "file";
    public static final String FOLDER_FIELD = "folder";
[...]

PR #1587, opened 2016-07-15
Move bst package to logic

lenhard:
Moves the `bst` package from root into `logic`. This operation is possible without any functional source code change. Moving the package does not break any of our existing dependency tests.

However, it requires re-generating the antlr sources (`gradlew generateSource`).

Refs:  #1579


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 ## [Unreleased]
 
 ### Changed
+- Moved the bst package into logic. For developers: This requires the regeneration of antlr sources, execute: gradlew generateSource
```
oscargus: One could argue if this should be in the change log as it is mainly a code issue. However, as there are some earlier entries like that and there is the need of antlr source regeneration it might just as well stay there.

oscargus:
LGTM! :+1:

CHANGELOG.md
matthiasgeiger: Perhaps just start the whole entry with "For developers: ..."

lenhard: @matthiasgeiger: Done. And I will of course squash the commits when merging

matthiasgeiger:
LGTM - @lenhard And don't forget to squash :wink:

Issue #1516, opened 2016-06-26
typo Isbn -> ISBN


jonasstein:
JabRef version 3.4

Isbn should be ISBN (https://en.wikipedia.org/wiki/International_Standard_Book_Number)
Doi should be DOI (https://en.wikipedia.org/wiki/Digital_object_identifier)

see screenshot
![2016-06-26_jabref3 4_isbn](https://cloud.githubusercontent.com/assets/298682/16362770/a670ac98-3bb7-11e6-96d9-03abe133ba93.png)

koppor:
URL, URI, ISSN are also candidates for other casing.

Maybe we should display the fields as we currently write them out:
everything in lower case.

Would that also be fine with you?

isbn, doi, url

jonasstein:
Why not write it in their correct spelling? I am not in the development team, so I do not have the right to decide about spellings, but to my understanding it would be very logical to write things as they are written. Please correct me, if I miss some obvious point here. Best regards, Jonas

lenhard:
Implemented and merged with #1589

PR #1589, opened 2016-07-15
Implement 1516

lenhard:
Implements #1516. Selected fields (`ISBN`, `URL`, `URI`, `ISSN`, `DOI`) are displayed in upper case in the entry editor.


oscargus:
LGTM! :+1:

matthiasgeiger:
one minor remark: "ISRN" which is used in Biblatex TechReports could - or should - be added 

Apart from that: LGTM

Issue #1345, opened 2016-05-03
Cleanup ISSN


tobiasdiez:
The new ISSN integrity check found some ISSN missing the dash in my db, for example `0003486X` insead of `0003-486X`. It would be nice if we could provide a cleanup operation which automatically fixes such ISSNs.

@oscargus something for you?

oscargus:
Implemented in #1590.

PR #1590, opened 2016-07-15
Implemented #1345 - cleanup ISSN

oscargus:
Implemented cleanup that adds a missing dash in the ISSN field if it looks like an ISSN (checksum is not controlled, but eight digits, where the last may be an x).

### DIFF
--- a/src/main/java/net/sf/jabref/gui/cleanup/CleanupPresetPanel.java
+++ b/src/main/java/net/sf/jabref/gui/cleanup/CleanupPresetPanel.java
@@ -22,6 +22,7 @@ public class CleanupPresetPanel {
 
     private final BibDatabaseContext databaseContext;
     private JCheckBox cleanUpDOI;
+    private JCheckBox cleanUpISSN;
     private JCheckBox cleanUpMovePDF;
     private JCheckBox cleanUpMakePathsRelative;
     private JCheckBox cleanUpRenamePDF;
@@ -44,6 +45,7 @@ public class CleanupPresetPanel {
     private void init() {
         cleanUpDOI = new JCheckBox(
                 Localization.lang("Move DOIs from note and URL field to DOI field and remove http prefix"));
+        cleanUpISSN = new JCheckBox(Localization.lang("Add possibly missing dash to ISSN"));
         if (databaseContext.getMetaData().getDefaultFileDirectory().isPresent()) {
             cleanUpMovePDF = new JCheckBox(Localization.lang("Move linked files to default file directory %0",
                     databaseContext.getMetaData().getDefaultFileDirectory().get()));
@@ -71,7 +73,7 @@ public class CleanupPresetPanel {
         updateDisplay(cleanupPreset);
 
         FormLayout layout = new FormLayout("left:15dlu, pref:grow",
-                "pref, pref, pref, pref, pref, pref, pref,pref, 190dlu, fill:pref:grow,");
+                "pref, pref, pref, pref, pref, pref, pref,pref, pref,190dlu, fill:pref:grow,");
 
         FormBuilder builder = FormBuilder.create().layout(layout);
         builder.add(cleanUpDOI).xyw(1, 1, 2);
@@ -84,7 +86,8 @@ public class CleanupPresetPanel {
         builder.add(new JLabel(currentPattern)).xy(2, 6);
         builder.add(cleanUpRenamePDFonlyRelativePaths).xy(2, 7);
         builder.add(cleanUpBibLatex).xyw(1, 8, 2);
-        builder.add(cleanUpFormatters).xyw(1, 9, 2);
+        builder.add(cleanUpISSN).xyw(1, 9, 2);
+        builder.add(cleanUpFormatters).xyw(1, 10, 2);
         panel = builder.build();
     }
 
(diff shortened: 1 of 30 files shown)


src/main/java/net/sf/jabref/logic/cleanup/ISSNCleanup.java
```diff
@@ -0,0 +1,33 @@
...
+
+    private static final Pattern ISSN_PATTERN_NODASH = Pattern.compile("^(\\d{4})(\\d{3}[\\dxX])$");
+    @Override
+    public List<FieldChange> cleanup(BibEntry entry) {
+        Optional<String> issn = entry.getFieldOptional("issn");
+        if (!issn.isPresent()) {
+            return Collections.emptyList();
+        }
+
+        Matcher matcher = ISSN_PATTERN_NODASH.matcher(issn.get());
```
tobiasdiez: I would move this functionality to a separate `ISSN` class which does the parsing. For example, the code then might look as follows:
````
ISSN issn = new ISSN(issnText);
if (issn.isValid()) {
   entry.setField("issn", issn.ToString())
}
````
Then you could reuse the same logic to add a ISSN integrity check. The same remark applies to your other ISBN PR.

File: src/main/java/net/sf/jabref/gui/cleanup/CleanupPresetPanel.java [hunk 2]
```diff
@@ -43,6 +44,7 @@ public CleanupPresetPanel(BibDatabaseContext databaseContext, CleanupPreset clea
     private void init() {
         cleanUpDOI = new JCheckBox(
                 Localization.lang("Move DOIs from note and URL field to DOI field and remove http prefix"));
+        cleanUpISSN = new JCheckBox(Localization.lang("Add possibly missing dash to ISSN"));
```
tobiasdiez: I would be more vague and say `Reformat ISSN`.

src/test/java/net/sf/jabref/logic/cleanup/CleanupWorkerTest.java
```diff
@@ -167,6 +168,36 @@ public void cleanupDoiReturnsChangeWhenDoiInURLField() {
     }
 
     @Test
+    public void cleanupISSNReturnsCorrectISSN() {
```
tobiasdiez: I would move these tests to a new ISSNCleanupTests class. Don't know why nearly all cleanup tests are bundled here in this class.

src/main/java/net/sf/jabref/logic/cleanup/ISSNCleanup.java
oscargus: How would you handle a string "3457345X"? Not set it? Add a dash? What about an arbitrary string?

Either way, I won't implement this at the moment.

src/main/java/net/sf/jabref/gui/cleanup/CleanupPresetPanel.java
oscargus: Could do that.

src/test/java/net/sf/jabref/logic/cleanup/CleanupWorkerTest.java
oscargus: Doable.

oscargus: I think the reason is all the dependency injections. Now it wasn't that hard, just mock both arguments, but it is clearly quicker to just add them to existing tests...

src/main/java/net/sf/jabref/logic/cleanup/ISSNCleanup.java
tobiasdiez: I don't get your question, sorry. If the passed string is not matched by the regex, then `issn.isValid` returns false. So it is essentially just moving your code to three different methods.

oscargus: So, what if it is a valid ISSN, but the dash is missing?

Siedlerchr: An ISSN would then not be valid!
It must contain an dash:
https://en.wikipedia.org/wiki/International_Standard_Serial_Number

oscargus: Sure, but it can be cleaned up as opposed to all other cases.

Maybe I was focusing too much on your example, but to not set the field
based on a missing dash seemed quite stupid to me. Better then to add it.
Hence the questions how to handle that.

Still, I do not really see it happening at the moment though.

tobiasdiez: So your problem is that for this PR you want to make sense of strings which look like ISSNs but are not quite in the standard format, right? What about creating `static Optional<ISSN> fromString(text)` which returns an ISSN object if JabRef could make sense of the input. Then the ISSN class has a constructor accepting the different parts of the ISSN (first part, second part and checkdigit), `isValid` returns true if the checksum is correct and toString formats it in the standard way using a separating dash. Would this be a solution?

oscargus:
The whole purpose of this PR was indeed to add the missing dash if it otherwise looks like an ISSN, yes. That works at the moment.

Would we actually need a clean up action then? If properly used, the ISSN class would provide automatic clean up, right? Which is useful, but may be annoying?

Btw, I doubt using the different parts explicitly is very useful though. But if there is a valid use case where the string is not enough, so sure. Should the parts then be `String` or `int` (x is 10 and only appears in the checksum)?

It really looks like over-engineering to me, I must say though...

src/main/java/net/sf/jabref/logic/cleanup/ISSNCleanup.java
stefan-kolb: Something similar should be in `DOI.java` if I remember correctly.

oscargus:
Extracted an `ISSN` class for checking validity and cleaning up.

simonharrer:
I am missing tests for the ISSN class. Otherwise, LGTM.

PR #1591, opened 2016-07-15
Replace getField with getFieldOptional in all of the tests and in som…

oscargus:
132 to go.

There may be cases where the Optional can be better used...


simonharrer:
LGTM

PR #1593, opened 2016-07-16
Some Globals.prefs injection in logic and model

oscargus:
Based on the discussion that Globals should not be accessed from logic and model.


src/main/java/net/sf/jabref/JabRefPreferences.java
```diff
@@ -360,7 +360,7 @@
         activeFormatterCleanups.add(new FieldFormatterCleanup("title", new ProtectTermsFormatter()));
         activeFormatterCleanups.add(new FieldFormatterCleanup("title", new UnitsToLatexFormatter()));
         activeFormatterCleanups.add(new FieldFormatterCleanup("title", new LatexCleanupFormatter()));
-        activeFormatterCleanups.add(new FieldFormatterCleanup("title", new HtmlToLatexFormatter()));
+        activeFormatterCleanups.add(new FieldFormatterCleanup("title", new HtmlToLatexFormatter(false))); // TODO: Get the actual value USE_CONVERT_TO_EQUATION
```
lenhard: Why is it not possible to use the actual value for `USE_CONVERT_TO_EQUATION` here? After all, this is the preferences class?

src/main/java/net/sf/jabref/logic/bibtex/LatexFieldFormatter.java
```diff
@@ -49,24 +49,24 @@
...
 
 
-    public LatexFieldFormatter() {
-        this(true);
+    public LatexFieldFormatter(JabRefPreferences prefs) {
+        this(true, prefs);
     }
 
-    private LatexFieldFormatter(boolean neverFailOnHashes) {
+    private LatexFieldFormatter(boolean neverFailOnHashes, JabRefPreferences prefs) {
```
lenhard: In case you need a set of configuration options from the preferences (as it is the case here), you could introduce a configuration object that takes all these configurations as parameters. By passing this object, you can avoid long parameter lists and get rid of the preferences dependency.

src/main/java/net/sf/jabref/JabRefPreferences.java
oscargus: It cannot be accessed. `prefs` is non-static and `Globals.prefs` is not yet set (and this is executed before it can be set).

src/main/java/net/sf/jabref/logic/formatter/Formatters.java
```diff
@@ -45,7 +45,7 @@
 public class Formatters {
 
     public static final List<Formatter> CONVERTERS = Arrays.asList(
-            new HtmlToLatexFormatter(),
+            new HtmlToLatexFormatter(false), // TODO: Get the correct value from USE_CONVERT_TO_EQUATION
```
lenhard: Maybe we should keep the Globals dependency here for now (`new HtmlToLatexFormatter(Globals.prefs.USE_CONVERT_TO_EQUATION)`), because this changes the behavior otherwise.

src/main/java/net/sf/jabref/logic/bibtex/LatexFieldFormatter.java
oscargus: Conceptually, yes. There are two problems (in general): 

1. that configuration object is in many cases a running target. You pass something that may or may not be used five calls from here. Layout is a good example, where some formatters use some preferences and you have to inject that in lots of methods. Adding a new formatter may change the object and so on.

2. That configuration object will depend on the preferences, so I do not really see where this disconnection should happen. At least not if the configuration of object should configure itself based on the preferences.

lenhard:
Overall, one step in the right direction! However, there is one immediate issue that we need to consider: If we decouple a class completely from `Globals` and `JabRefPreferences`, for instance by just passing String values read from the preferences as parameters to the constructor, we might run into refreshment bugs. This means that if someone changes the preferences via the GUI, the change might have no effect on JabRef, since the value is not re-read every time it is needed. The obvious solution that comes to my mind to avoid this is the Windows approach: Always restart JabRef for preferences changes. What do @JabRef/developers think?

matthiasgeiger:
> The obvious solution that comes to my mind to avoid this is the Windows approach: Always restart JabRef for preferences changes.

:weary: I don't like this. It already unnerves me that in some occassions a restart is needed. 

I don't have the time (and I don't want to :wink:) check this whole PR... but using params with the current settings, or recreating the object instances on pref changes is not possible?

lenhard:
@matthiasgeiger At least I do not really know where every preference is needed. And I also do not know if it is possible to just recreate certain parts of JabRef on demand, without a restart of the program (because JabRef is just so tangled and that is what we are trying to improve here). @oscargus: Can you perhaps look at all occasions where you removed `Globals` and `JabRefPreferences` in this PR and try to guess if there might be problems when preferences are changed?

src/main/java/net/sf/jabref/logic/bibtex/LatexFieldFormatter.java
tobiasdiez: 1. This is actual no problem but a feature. If you add a new formatter, then you just need to add a variable to the configuration object and that's it. No need to modify lots of methods to just pass the dependency.

2. That's a good point. For now think about the configuration object as a place to collect the access to the preference class. Later we might rework these classes so that they no longer dependent on the preferences. Take for example `SavePreferences`. This class encapsulates (nearly) all configuration options for saving a bib-file. Right now it depends on the preference class via `SavePreferences.fromPreferences(JabRefPreferences)` but this is the only place. Now it is easy to invert this method and change it to `SavePreferences JabRefPreferences.getSavePreferences()`.

src/main/java/net/sf/jabref/logic/formatter/Formatters.java
tobiasdiez: I think we need a general strategy for how preferences in these formatters are handled. The idea of save actions was that these are specified in the bib-file and can be invoked on every save without destroying the bibitems. For this it is really necessary that every user runs the same save actions for the same file. But this is might be no longer the case if some formatters depend on user preferences. For the moment I would not change them but document which formatter depends on which preferences.

src/main/java/net/sf/jabref/logic/bibtex/LatexFieldFormatter.java
oscargus: 1. Makes sense. Passing an object may be a good idea, I agree.

2. Ahh, that's correct. Didn't think of that option. More the `SavePreferences.fromPreferences(JabRefPreferences)` version.

src/main/java/net/sf/jabref/logic/formatter/Formatters.java
oscargus: Maybe this isn't needed and we just can decide on one way to convert subscripts and superscripts from HTML to LaTeX. It is there from historical reasons, where I first did it using equations and then someone wanted the `\textsuperscript` version. I would almost suggest always using `\textsuperscript` now that the table view can render it properly (earlier it was more compact with `$^2$` compard to `\textsuperscript{2}`.

This would also reduce the amount of settings.

lenhard: +1 for always using `\textsuperscript`

oscargus:
It is definitely a problem for `HtmlToLatexFormatter`, but maybe we should remove that support anyway. It may be a problem for the AutoCompleter (not sure where the preferences are used there, if they are just passed on to LayOut, in which case it is used for formatting the File-formatters, so probably not an issue).

There are some way to alleviate this though:
* In `JournalAbbreviationLoader.getRepository()` the `JournalAbbreviationRepository` is only generated once. Here one can probably check if the preferences (configuration object) have changed, and in that case generate a new `JournalAbbreviationRepository`. This would improve the behavior compared to the current one (before this PR).

* In `CustomExportList` the `JabRefPreferences` are passed to every call. This is something I was considering fixing and only use the one from the constructor, but it is actually a feature here.

As it is said that this is an evolving process, what remains to be done? I think dropping the preferences from `HtmlToLatexFormatter` if nothing else. There are some clear cases where a specific preferences class and the generic one are both passed. This is something for later. Anything else?

src/main/java/net/sf/jabref/logic/formatter/Formatters.java
oscargus: The original reason here was actually to avoid an NPE rather than avoiding `Globals`. (That comment went missing it seems like.)

lenhard:
Though it is an evolving process, we should not introduce bugs in the meantime :-) In `HTMLToLatexFormatter`, you can remove the preferences configuration (which you already did). 

As for the `AutoCompleter`, it should really still work and update as before the PR. So, if it doesn't at the moment you should implement an update of the preferences for the `AutoCompleter`, e.g. by recreating and resetting the object if changes in the preferences occur. 

The whole rationale is: JabRef should behave in the same fashion after this PR as it did before (with the exception of `HTMLToLatexFormatter`). Architecture refactorings should not change functionality. It is quite hard for me to guarantee this, so it really depends on your intuition and testing.

If you find it too hard to implement a preferences update at a certain position, it would be better if you just reverted the dependency refactoring for that position. Then we can have a closer look at the specific dependency in another PR.

oscargus:
Regarding `AutoCompleter`: it is currently (in the `master` branch) not updating when the preferences are changed. This PR actually makes it easier to update (although it is indeed behaving the same as before at the moment).

lenhard:
Ok, then LGTM!

oscargus:
To be more precise: `JournalAbbreviationLoader.getRepository()` used to get `Globals.prefs` on the first call. Now it is passed as an argument. Hence, any relevant changes in the preferences between starting (where e.g. `AutoCompleter` is initialized with the current `Globals.prefs` and later call `getRepository()`) and the first actual call to `getRepository()` (not sure when that is, it might be at start up as well and then there is no change) will be ignored now. However, the fundamental issue that any subsequent changes to the preferences after `getRepository()` is called first are ignored is still the same.

I think that the idea of a `PreferenceChanged` event is really the way to go. Although not obvious how to easily implement it at a "higher level", i.e. at the classes passing the preferences on rather than in the classes using the preferences.

src/main/java/net/sf/jabref/logic/bibtex/LatexFieldFormatter.java
oscargus: OK, I added `LatexFieldFormatterPreferences` and `FieldContentParserPreferences`. More are clearly needed to cover the complete `logic`, but at least something.

src/main/java/net/sf/jabref/logic/exporter/BibtexDatabaseWriter.java
```diff
@@ -109,7 +113,10 @@ protected void writeString(BibtexString bibtexString, boolean isFirstString, int
                 getWriter().write("{}");
             } else {
                 try {
-                    String formatted = new LatexFieldFormatter().format(bibtexString.getContent(), LatexFieldFormatter.BIBTEX_STRING);
+                    String formatted = new LatexFieldFormatter(
+                            LatexFieldFormatterPreferences.fromPreferences(jabRefPreferences))
```
tobiasdiez: Could you include the `LatexFieldFormatterPreferences` in `SavePreferences` (and remove the explicit reference on `JabRefPreferences` from `BibtexDatabaseWriter`)

oscargus: Yes, that would make sense. I just did something similar to AutoCompleterPreferences and JournalAbbreviationPreferences.

src/main/java/net/sf/jabref/logic/groups/KeywordGroup.java
```diff
@@ -370,7 +371,7 @@ public static String getDescriptionForPreview(String field, String expr, boolean
     public String getShortDescription() {
         StringBuilder sb = new StringBuilder();
         sb.append("<b>");
-        if (Globals.prefs.getBoolean(JabRefPreferences.GROUP_SHOW_DYNAMIC)) {
+        if (jabRefPreferences.getBoolean(JabRefPreferences.GROUP_SHOW_DYNAMIC)) {
```
tobiasdiez: is it possible to pass `jabRefPreferences.getBoolean(JabRefPreferences.GROUP_SHOW_DYNAMIC)` as a boolean parameter to getShortDescription?

src/test/java/net/sf/jabref/model/entry/BibEntryTests.java
```diff
@@ -212,16 +212,21 @@ public void setNullField() {
 
     @Test(expected = NullPointerException.class)
     public void addNullKeywordThrowsNPE() {
-        keywordEntry.addKeyword(null);
+        keywordEntry.addKeyword(null, Globals.prefs.get(JabRefPreferences.KEYWORD_SEPARATOR));
```
tobiasdiez: introduce a variable `separator = ', '` on class level and reuse it in every test.

src/test/java/net/sf/jabref/logic/search/SearchQueryTest.java
```diff
@@ -33,7 +35,7 @@ public void testIsGrammarBasedSearch() {
     @Test
     public void testGrammarSearch() {
         BibEntry entry = new BibEntry();
-        entry.addKeyword("one two");
+        entry.addKeyword("one two", Globals.prefs.get(JabRefPreferences.KEYWORD_SEPARATOR));
```
tobiasdiez: Just use constant ` , ` as second parameter.

src/test/java/net/sf/jabref/logic/groups/SearchGroupTest.java
```diff
@@ -21,27 +21,27 @@ public static void setUp() {
...
         SearchGroup group = new SearchGroup("myExplicitGroup", "review",
-                true, true, GroupHierarchyType.INDEPENDENT);
+                true, true, GroupHierarchyType.INDEPENDENT, Globals.prefs);
         assertEquals("SearchGroup:myExplicitGroup;0;review;1;1;", group.toString());
 
         BibEntry entry = new BibEntry();
         assertFalse(group.contains(entry));
 
-        entry.addKeyword("review");
+        entry.addKeyword("review", Globals.prefs.get(JabRefPreferences.KEYWORD_SEPARATOR));
```
tobiasdiez: see below

tobiasdiez:
LGTM :+1: after having a quick look

src/test/java/net/sf/jabref/model/entry/BibEntryTests.java
oscargus: Does that really make sense? I just had to add some stuff in `LatexFieldFormatterPreferences` as some other behaviour really relies on the preferences but it is not handled in the tests.

src/main/java/net/sf/jabref/logic/groups/KeywordGroup.java
oscargus: How would you then run `fromString` if the result is a KeyWordGroup?

getShortDescription is not a problem. I'll fix that.

oscargus:
I should just note that I was incorrect about `AutoCompleter`/`JournalAbbreviationLoader`. It seems quite clever and I missed the point of having an `update` method. Everytime anything(?) related to the `JournalAbbreviationLoader` is changed in the preferences it is indeed updated.

simonharrer:
Please rebase and merge.

src/test/java/net/sf/jabref/model/entry/BibEntryTests.java
tobiasdiez: `Globals.prefs` is not really initialized in tests (on your system, the current JabRef preferences are used while on Travis the default preferences are taken). Thus one should in general avoid it and use hard-coded input values.

oscargus: Well, there are many tests that rely on default preferences without
specifying it. For example, the begin and end symbols for field `{` and `}`
are from the preferences and using an empty `SavePreferences` will
indirectly cause a requirement of specifying these (and, hence, it would
have been better to initialize those from the preferences in all tests).
I'm not really sure about all the cases here but somehow it may be better
to avoid later surprises...

Either way: not very hard to search and replace later. I'm travelling so
most likely nothing will be changed until tomorrow evening at the latest.

tobiasdiez: Well, if the KeywordGroup has problems with different separators, then one should write new tests for this separator. But then it would be easier to pass them directly to the method instead of setting a preference. Anyway, logic should be independent on the preferences and so should be their tests.

oscargus: That basically means that there shouldn't be any logic tests which initializes `Globals.prefs`. The question is just how long we should enforce that... I can remove the ones here, though.

tobiasdiez: Right! Tests should try to never initialize `Globals.prefs`. Actually, your PR is one step in this direction since so far it was difficult to tell which preferences are accessed deep in the call chain.

Issue #550, opened 2015-12-21
"New entry from Id" supporting just adding with an id (and not the Id type)


koppor:
As user, I want to generate an entry based on an Id. JabRef should know the id type by itself. Thus, the drop down "ID type" should be optional. Just reading the id and checking whether it is a valid doi, valid isbn, ... If yes, fetch the entry.

Additon: If an Id is pasted into the table, the entry should be created by the Id.

# Old request

In the web search, we have two types of fetchers: The first types are crawlers (like IEEE or Springer) which return multiple entires. The second type are fetchers (like DOIToBibTeX, ISBNToBibTex, or Fetch ArXiv.org) which return a single entry.

Proposal move all single-entry-fetchers to "BibTeX -> New entry". One possibility is to offer a dropdown where the fetcher can be chosen. Next to the dropdown, the id of the entry can be input. Next to it, the button "Fetch" triggers the action.

## Current UI

![grafik](https://user-images.githubusercontent.com/1366654/69181806-12bd4e80-0b10-11ea-800f-5e1b10c905fe.png)

simonharrer:
Or something similar to "New entry ..." which has a sub-menu with all possibilities. This would help the user find this functionality more easily. Otherwise, it can be so well hidden that no one uses it.

koppor:
I meant adding at the end of the "Select entry type" dialog. Marked below. I think, there, the users really find the possibility.

![grabbed_20151221-103329](https://cloud.githubusercontent.com/assets/1366654/11927642/4ebe6ae4-a7ce-11e5-892d-bcb6a3e4c8df.png)

simonharrer:
Hm, I would not integrate this there but rather create a new action in the menu.

koppor:
Is "there" above the "Cancel" button like

    <Fetcher dropdown>  <input field for id>  [Fetch Button]

Example

    DOI   10.1007/978-3-642-45005-1_64    Fetch

Then it's the same as I have in mind.

stefan-kolb:
@JabRef/developers This was done, right? Can be closed then...

tobiasdiez:
Kind of....it is in the "New entry..." dialog but in https://github.com/JabRef/jabref/pull/1925#issuecomment-244870799 we said, that we also want to add a new "New entry from ID" dialog. I mark it as after-javafx.

Siedlerchr:
Thank you for reporting this issue. We think, that is already fixed in our development version and consequently the change will be included in the next release.

[...]

Issue #383, opened 2015-11-23
Distinguish simple fetchers and web searches


koppor:
JabRef offers a Web Search ( `F5`), where on can choose between things like DBLP or `DOI to BibTeX`. The DBLP search is a real search returning multiple entires, whereas the latter is a "simple" fetcher for a single entry identified by a DOI.

I propose to move the simple fetchers returning a single entry into "BibTeX -> new entry" as fourth group: "Generate based on id"

    Fetcher                  Identifier
    [dropdown for fetchers]  [empty field to key in id]

Example

    Fetcher                          Identifier
    DOI                              10.1145/2578903.2579140

- [ ] Move the single-id fetchers away from web search

simonharrer:
:+1:

koppor:
Implemented in https://github.com/JabRef/jabref/pull/1925

PR #1594, opened 2016-07-16
[WIP] Create new fetcher infrastructure

tobiasdiez:
The aim of this PR is to refactor the fetcher interface, in particular separate logic and GUI code as well as implement https://github.com/JabRef/jabref/issues/383 and https://github.com/JabRef/jabref/issues/550. 

I propose the following class hierarchy for fetchers:

- WebFetcher: search web resources for matching BibEntries
	- SearchBasedFetcher: based on a free-text query (may return multiple results)
	- IdBasedFetcher: based on an ID (single item as result)
- IdFetcher: looks for an ID (arXiv, DOI,...) based on an existing BibEntry
- FulltextFetcher: find fulltext of an existing BibEntry (i.e. rebranding of FullTextFinder)

--


oscargus:
Are these classes logic or gui? What is the complete work flow? ImportInspector? Merging? (Adding empty fields may not be enough.)

tobiasdiez:
They are logic and try to extract the non-gui part from the current `EntryFetcher`. Thus the workflow is not changed (at least not with this PR).

matthiasgeiger:
Is generally fine for me. 

One remark: There are currently two different workflows for the "SearchBasedFetchers". 
Normally one hits the "Search" button after inserting some search query and a list of found entries is presented, after selecting the desired entries they are imported. 
However, at least for the GoogleScholar fetcher there is another dialog presenting more details to "preselect" the entries which should be shown in the list of found entries (I think the intention is to reduce the amount of queries).

Is the proposed "shallow" search method used in this case? Or is this difference in the workflow currently not really covered, yet?

koppor:
I think the workflow is supported by the two methods `performShallowSearch` and `performDeepSearch`. I assume that the dialog is presented if `performDeepSearch` is present. I think that one needs an additional class having only `performShallowSearch` and a second one containing both methods.

My concern is that the intermediate results might not be available as bibtex entries, but as HTML only. I think, the `ImportInspectionDialog` was used for that, wasn't it.

What about error messages? I assume the aim is to use `LOGGER` only and to use the error console? I used to like the GUI dialogs showing errors, but OK for me to remove them to improve the code. (Until now, this was done by the `OutputPrinter status`, wasn't 

@zellerdev will try to port GVKFetcher and the ISBNToBibTeX fetcher to the new infrastructure. He will do a PR on your repo, is that OK, @tobiasdiez?

Siedlerchr:
For the FulltextFetcher it would be awesome if the FullText finding could be run as part of a Cleanup operation, e.g. globally finding fulltext for all entries.
This should be kept in mind

tobiasdiez:
Thanks for the feedback. 
@matthiasgeiger, @koppor  yes I thought about covering the two-stage search by the deep and shallow methods. Maybe this is not the best idea and one needs a different approach for the Google search. Thus for now I would concentrate on the simple fetchers which get all the metadata in one query.

@koppor The result from the server is of course HTML (or XML, JSON,...). It is the task of the fetcher to translate the response to something meaningful, i.e. to `BibEntries`. I expanded the arXiv fetcher to outline how this works (in fact the arXiv fetcher now understands free text queries and not just ids). I also included a way to notify the caller about errors...by, well, throwing exceptions (with meaningful error messages). Maybe the fetchers also need to post some events to let the GUI know about the progress. 

@Siedlerchr Good idea. However a bit out of the scope of this PR. Could you file a new issue for it? Thanks!

tobiasdiez:
So PR is ready for review. 
The new fetcher can now be included via a wrapper. In this way the fetcher can be transformed to the new interface step-by-step without the need to refactor everything at once.

If I get the ok for merge, then I (or @zellerdev) will move `HelpFiles` to logic and thereby fix the failing architecture test.

oscargus:
Many of the current fetchers will only fetch a limited number of entries. Typically based on user input. How is that going to be handled here?

src/main/java/net/sf/jabref/logic/fulltext/SearchBasedEntryFetcher.java
```diff
@@ -0,0 +1,85 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+package net.sf.jabref.logic.fulltext;
```
oscargus: gui?

src/main/java/net/sf/jabref/logic/fulltext/SearchBasedEntryFetcher.java
```diff
@@ -0,0 +1,85 @@
...
+
+import java.util.List;
+import java.util.Objects;
+
+import javax.swing.JPanel;
```
oscargus: No really enough to move HelpFile.

src/main/java/net/sf/jabref/logic/util/strings/StringUtil.java
```diff
@@ -656,4 +658,12 @@ public static String repeat(int n, char c) {
     public static boolean isNullOrEmpty(String toTest) {
         return (toTest == null || toTest.isEmpty());
     }
+
+    public static boolean isNotBlank(Optional<String> string) {
```
oscargus: Useful method!

src/main/java/net/sf/jabref/logic/util/strings/StringUtil.java
```diff
@@ -656,4 +658,12 @@ public static String repeat(int n, char c) {
...
+        return string.isPresent() && isNotBlank(string.get());
+    }
+
+    public static boolean isNotBlank(String string) {
```
oscargus: Not so useful method. (As it doesn't work...)

stefan-kolb:
Whereas I think the interface redesign is a good thing, I'd like to throw a few things for discussion.

1. The interface hierarchy is too complex in my opinion why not just use:
- WebFetcher: Has two methods for SearchBasedFetcher & IdBasedFetcher
- IdFetcher
- FulltextFetcher

The web fetcher hierarchy seems artificial to me. Why do we ultimately need one method interfaces?

2. I really don't like the effect that we merge all fetchers into one class, e.g. your ArxiV.java class.
Some fetcher are complex enough on their own so why create new 1000 lines classes being capable of searching, id fetching and full text retrieval. Let us just use 3 classes for the three types of fetching.

stefan-kolb:
On another note: I don't think that `HelpFile.java` belongs into the logic package.

tobiasdiez:
@oscargus Right now every parameter has to be passed to the fetcher via the constructor. Maybe it would make sense to include the number of results as a method parameter. Right now I don't have a feeling on how many fetcher actually support this (or if a common fixed value of say 30 would also work). 

@stefan-kolb The idea to distinct between ID-based and text-based fetchers came from #383 and #550. Right now the interfaces only have one method but this might change in the future (for example, a method to check validity of a given ID would make sense for the ID-based fetcher). 
If a fetcher is complex enough, then one of course still decompose the functionality in different classes. For the arXiv fetcher it felt like all the functionality is essentially given by one method (callApi) and the other methods are just different wrappers around it. 
Do you have a good idea how to handle for the help files in fetchers (logic) if HelpFiles stays in gui?

stefan-kolb:
I still don't get why these two types of fetcher methods cannot live in one interface.
I haven't checked the usage of the HelpFiles yet, tho I think this has to be managed by the GUI somehow.

Siedlerchr:
The helpfiles could be under the logic,, as they do not contain any gui related methods or any other related thing.
They just serve as a kind of Container class, they simply store the name of the HelpPage-Markdown file in an Enum.
The assembling of the url and opening is in the gui in HelpAction.

oscargus:
I think a two-stage search is preferred in many situations. All is not
really an option and sometimes one would really like to go for more than
100 (which is a more reasonable number). Personally, I often use JabRef as
a search engine to find interesting papers (or rather, I search for a
specific paper and then find related ones so it makes sense to not have a
too narrow query, also some publishers has a logic or on the search terms
so you get more hits the more terms you add).

Just try searching Springer or DOAJ for some common term, "efficient" or
"Smith" or something.

I think this can be solved though by adding another call so query -> number
of hits -> number to fetch -> list of entries.

matthiasgeiger:
Thanks Oscar for bringing up this aspect. 

I think the main reason for introducing this step was to reduce the number of requests fired to the different providers as some of them are rather quick with blocking IPs crawling to much information in a short time.
Thus the number of matches to be shown is not only dependent on the intention of the user (find a single paper vs. search for all possible papers matching the search) but also might have technical reasons... 
However, personally I don't like the workflow "Enter term" > "Start search" > "Determine number of hits and ask user how much hits should be displayed" > "List chosen number of hits" 
I think "Enter term" > "Start search" >"List first X hits" + "Label: Showing first X of Y hits" + Button "get next X hits" would be better... but this would require some more methods in the Interfaces to start fetching from/to a given offset...

stefan-kolb:
@simonharrer and @matthiasgeiger have persuaded me that I'm dumb, please merge as intended. Maybe the naming can be improved a little bit but this is NP-hard.

oscargus:
OK, so how should blocking be avoided? I agree that @matthiasgeiger's approach is preferred, but there must be some support for fetching an arbitrary number of items.

PR #1596, opened 2016-07-16
Cleanup FindFile and asssociated tests

Siedlerchr:
Cleaned up the findFiles method
Reworked tests, added some new Tests
Tests should now work in Linux, too 


Siedlerchr:
I have no idea why the test fails, however, it seems to be a Linux related problem.
Could anyone please test my code on Linux?

src/main/java/net/sf/jabref/logic/util/io/FileFinder.java
```diff
@@ -1,43 +1,42 @@
...
 
-    private static Collection<? extends File> findFiles(Collection<String> extensions, File directory) {
-        Set<File> result = new HashSet<>();
+    public static Set<File> findFiles(Collection<String> extensions, Collection<File> directories) {
 
-        File[] children = directory.listFiles();
-        if (children == null) {
-            return result; // No permission?
-        }
+        Objects.requireNonNull(directories, "Directories must not null!");
```
oscargus: "not be"

src/main/java/net/sf/jabref/logic/util/io/FileFinder.java
```diff
@@ -1,43 +1,42 @@
...
-            if (child.isDirectory()) {
-                result.addAll(FileFinder.findFiles(extensions, child));
-            } else {
-                FileUtil.getFileExtension(child).ifPresent(extension -> {
-                    if (extensions.contains(extension)) {
-                        result.add(child);
-                    }
-                });
+        for (File directory : directories) {
+            Set<File> files;
```
oscargus: Why is this declared outside of the try? (Just curious if there is a reason.)

src/test/java/net/sf/jabref/logic/util/io/UtilFindFileTest.java
```diff
@@ -21,24 +23,41 @@
...
+    public void testFindAssociatedFilesException() {
+        Collection<String> extensions = Arrays.asList("jpg", "pdf");
+        Collection<File> dirs = Arrays.asList(new File("asdfasdf/asdfasdf"));
+        Set<File> results = FileFinder.findFiles(extensions, dirs);
+
+        Assert.assertEquals(0, results.size());
     }
 
+    @Test(expected = NullPointerException.class)
+    public void testFindFilesNUllPointerException() {
```
oscargus: spelling: Null

src/main/java/net/sf/jabref/logic/util/io/FileFinder.java
```diff
@@ -1,43 +1,42 @@
...
-        return result;
-    }
 
-    private static Collection<? extends File> findFiles(Collection<String> extensions, File directory) {
-        Set<File> result = new HashSet<>();
+    public static Set<File> findFiles(Collection<String> extensions, Collection<File> directories) {
 
-        File[] children = directory.listFiles();
-        if (children == null) {
-            return result; // No permission?
```
oscargus: Are there cases where the result is explicitly checked for an empty `Set` and some warning related to permissions is shown?

oscargus: Sorry, I misread the code... It should be OK.

oscargus:
No idea why the tests fail. Interesting that the NPE test is failing as well...

src/main/java/net/sf/jabref/logic/util/io/FileFinder.java
Siedlerchr: No reason. Was from my first experiment without try-catch but throws.

Siedlerchr:
@oscargus  Yeah, that bothered me, too.  I tested on a Linux VM and it only fails there.
I will try to debug further.

src/test/java/net/sf/jabref/logic/util/io/UtilFindFileTest.java
```diff
@@ -21,24 +23,36 @@
...
+        Collection<File> dirs = Arrays.asList(rootPath.resolve("graphicsdir").toFile(),
+                rootPath.resolve("pdfs").toFile());
 
         Map<BibEntry, List<File>> results = FileUtil.findAssociatedFiles(entries, extensions, dirs);
 
         Assert.assertEquals(2, results.get(entry).size());
+        Assert.assertTrue(results.get(entry)
+                .contains(rootPath.resolve(Paths.get("graphicsDir", "subDir", "HipKro03test.jpg")).toFile()));
+        Assert.assertFalse(results.get(entry)
+                .contains(rootPath.resolve(Paths.get("graphicsDir", "subDir", "HipKro03test.png")).toFile()));
```
oscargus: Could it be that it says png here but pdf in extensions? (Not sure why it would work on Windows though...)

oscargus: Oh, is said `assertFalse`. That explains things. :-)

Siedlerchr: The problem is that under linux the first AssertEquals fails. because it says size is 1 but excepted 2.

Siedlerchr:
I will debug it tomorrow at university. I have absolutely no idea why it fails. All stuff is done in the BaseClass in the Before Method.

src/test/java/net/sf/jabref/logic/util/io/UtilFindFileTest.java
Siedlerchr: @oscargus  I think I found the error! Case Sensitivy. Windows is case insenstive, but Linux not!
`graphicsDir `vs g`raphicsdir`

oscargus: Seems reasonable! Well hunted!

Siedlerchr:
Okay, only the Exeption test did not work...

src/test/java/net/sf/jabref/FileBasedTestCase.java
```diff
@@ -38,78 +39,55 @@ public void setUp() throws IOException {
...
+        assertTrue(Files.exists(Files.createFile(graphicsSubDir.resolve("HipKro03test.png"))));
 
     }
 
     @After
     public void tearDown() {
-        FileBasedTestHelper.deleteRecursive(root);
+
+        try {
+            for (Path p : Files.walk(rootDir).sorted((a, b) -> b.compareTo(a)). // reverse; files before dirs
```
simonharrer: Files.walk creates a stream, which must be closed using try-with-resources, otherwise, this may affect another test as it will not free the directory resource.

src/main/java/net/sf/jabref/logic/util/io/FileFinder.java
```diff
@@ -1,43 +1,42 @@
...
         Set<File> result = new HashSet<>();
+        for (File directory : directories) {
 
-        File[] children = directory.listFiles();
-        if (children == null) {
-            return result; // No permission?
-        }
+            try {
+                Set<File> files = Files
+                        .find(directory.toPath(), Integer.MAX_VALUE,
```
simonharrer: also Files.find must be closed

src/test/java/net/sf/jabref/logic/util/io/UtilFindFileTest.java
```diff
@@ -21,24 +24,37 @@
...
-                results.get(entry).contains(new File(root.getAbsoluteFile() + "/graphicsDir/subDir/HipKro03test.png")));
-        Assert.assertTrue(results.get(entry).contains(new File(root.getAbsoluteFile() + "/pdfs/sub/HipKro03-sub.pdf")));
+                results.get(entry).contains(rootDir.resolve(Paths.get("pdfs", "sub", "HipKro03-sub.pdf")).toFile()));
     }
 
+    @Test
+    public void testFindAssociatedFilesException() {
+        Collection<String> extensions = Arrays.asList("jpg", "pdf");
+        Collection<File> dirs = Arrays.asList(new File("asdfasdf/asdfasdf"));
+        Set<File> results = FileFinder.findFiles(extensions, dirs);
```
simonharrer: close

simonharrer:
Do you need any help here?

Siedlerchr:
@tobiasdiez  Yes, I still have no idea why the test is failing only on the CI Server.  Both under win and linux all tests are passing. Do you have a clue what the reason can be?

simonharrer:
You could try to get more information from Gradle about the failing test. Maybe run it with --stacktrace in the travis configuration, and have a look at the gradle test configuration so that the full stack trace is shown. This could help to find the cause.

src/main/java/net/sf/jabref/logic/util/io/FileFinder.java
```diff
@@ -1,43 +1,42 @@
...
-        Set<File> result = new HashSet<>();
+    private static final Log LOGGER = LogFactory.getLog(FileFinder.class);
 
-        for (File directory : directories) {
-            result.addAll(FileFinder.findFiles(extensions, directory));
-        }
 
-        return result;
-    }
+    public static Set<File> findFiles(Collection<String> extensions, Collection<File> directories) {
```
stefan-kolb: Maybe just use List here instead of Collection?

src/main/java/net/sf/jabref/logic/util/io/FileFinder.java
```diff
@@ -1,43 +1,42 @@
...
         Set<File> result = new HashSet<>();
+        for (File directory : directories) {
 
-        File[] children = directory.listFiles();
-        if (children == null) {
-            return result; // No permission?
-        }
+            try (Stream<File> files = Files
+                    .find(directory.toPath(), Integer.MAX_VALUE,
+                            (path, attr) -> !Files.isDirectory(path)
```
stefan-kolb: Wow, this is complex here/ugly...Maybe split this up a little bit? Even consider not using streams here?

src/test/java/net/sf/jabref/logic/util/io/UtilFindFileTest.java
```diff
@@ -1,45 +1,64 @@
...
 
-        Assert.assertEquals(2, results.get(entry).size());
-        Assert.assertTrue(
-                results.get(entry).contains(new File(root.getAbsoluteFile() + "/graphicsDir/subDir/HipKro03test.jpg")));
-        Assert.assertFalse(
-                results.get(entry).contains(new File(root.getAbsoluteFile() + "/graphicsDir/subDir/HipKro03test.png")));
-        Assert.assertTrue(results.get(entry).contains(new File(root.getAbsoluteFile() + "/pdfs/sub/HipKro03-sub.pdf")));
+        assertEquals(2, results.get(entry).size());
+        assertTrue(results.get(entry)
+                .contains(rootDir.resolve(Paths.get("graphicsDir", "subDir", "HipKro03test.jpg")).toFile()));
```
stefan-kolb: Maybe the Windows/Linux thing is a file name casing thing? Just guessing here...!

src/test/java/net/sf/jabref/logic/util/io/UtilFindFileTest.java
```diff
@@ -1,45 +1,64 @@
...
+        Collection<String> extensions = Arrays.asList("jpg", "pdf");
+        Collection<File> dirs = Arrays.asList(rootDir.resolve("asdfasdf/asdfasdf").toFile());
+        Set<File> results = FileFinder.findFiles(extensions, dirs);
 
+        assertEquals(0, results.size());
+    }
+
+    @Ignore("Fails on CI Server")
+    @Test(expected = NullPointerException.class)
+    public void testFindFilesNUllPointerException() {
```
stefan-kolb: NUll -> Null

src/main/java/net/sf/jabref/logic/util/io/FileFinder.java
Siedlerchr: I can maybe split the Bi-Predicate lambda in a new function..
The reason for streams is that is lazily walking the file system and increases performance for huge directories.

src/test/java/net/sf/jabref/logic/util/io/UtilFindFileTest.java
Siedlerchr: Yeah, that was previous  Error. Now the tests work on linux, too. But not on the CI Server.

tobiasdiez:
Just a quick guess: maybe the `@Before` method from the base class is not called and thus the root directory is null, which gives the NPE. You might try to call the initialization method explicitly, see http://stackoverflow.com/a/2132852/873661

Siedlerchr:
@tobiasdiez  Thanks . I already thought that it may be related to the sublassing.
I will investigate this option

Siedlerchr:
Okay, not it only fails at the tearDown method, when restoring the prefs.
@oscargus  Could you please have a look, as you recently changed the Globals.pref stuff?

oscargus:
Well, I didn't actually change anything there. Only encapsulated the use of it in a number of places, so shouldn't affect this.

Siedlerchr:
Finally all tests green :cake:

oscargus:
:+1:

simonharrer:
Hm, what about the still marked `@Ignore` test?

src/test/java/net/sf/jabref/logic/util/io/FileBasedTestCase.java
```diff
@@ -0,0 +1,124 @@
...
+
+
+    @Before
+    public void setUp() throws IOException {
+        Globals.prefs = mock(JabRefPreferences.class);
+
+        BibDatabase database = BibtexTestData.getBibtexDatabase();
+        entry = database.getEntries().iterator().next();
+
+        rootDir = Files.createTempDirectory("UtilFindFileTest");
```
tobiasdiez: I think the proper way to handle temporary folders in tests is via 
`@Rule public TemporaryFolder temporaryFolder = new TemporaryFolder();`,
which also takes care of deleting the folder after every test.

src/test/java/net/sf/jabref/logic/util/io/FileBasedTestCase.java
```diff
@@ -0,0 +1,124 @@
...
+
+
+
+        Path subDir = Files.createDirectory(rootDir.resolve("Organization Science"));
+        Path pdfSubDir = Files.createDirectory(rootDir.resolve("pdfs"));
+
+        assertTrue(Files.exists(Files.createFile(subDir.resolve("HipKro03 - Hello.pdf"))));
```
tobiasdiez: I would prefer a smiple `Files.createFile` without the assertion for readability (and since I suspect that creating the file never fails...)

src/test/java/net/sf/jabref/logic/util/io/FileBasedTestCase.java
```diff
@@ -0,0 +1,124 @@
...
+
+        List<BibEntry> entries = Collections.singletonList(entry);
+        List<String> extensions = Arrays.asList("jpg", "pdf");
+        List<File> dirs = Arrays.asList(rootDir.resolve("graphicsDir").toFile(), rootDir.resolve("pdfs").toFile());
+
+        when(Globals.prefs.getBoolean(JabRefPreferences.AUTOLINK_EXACT_KEY_ONLY)).thenReturn(false);
+
+        Map<BibEntry, List<File>> results = FileUtil.findAssociatedFiles(entries, extensions, dirs, Globals.prefs);
+
+        assertEquals(2, results.get(entry).size());
```
tobiasdiez: Better create a `Map<BibEntry, List<File>> expected` and compare that it equals `results`

src/test/java/net/sf/jabref/logic/util/io/FileBasedTestCase.java
```diff
@@ -0,0 +1,124 @@
...
+        assertTrue(results.get(entry).contains(rootDir.resolve(Paths.get("pdfs", "sub", "HipKro03-sub.pdf")).toFile()));
+    }
+
+    @Test
+    public void testFindFilesException() {
+        List<String> extensions = Arrays.asList("jpg", "pdf");
+        List<File> dirs = Arrays.asList(rootDir.resolve("asdfasdf/asdfasdf").toFile());
+        Set<File> results = FileFinder.findFiles(extensions, dirs);
+
+        assertEquals(0, results.size());
```
tobiasdiez: compare to emptyset

src/test/java/net/sf/jabref/logic/util/io/FileBasedTestCase.java
```diff
@@ -0,0 +1,124 @@
...
+    public void testFindFilesNullPointerException() {
+
+        assumeFalse(DevEnvironment.isCIServer());
+        FileFinder.findFiles(null, null);
+    }
+
+    @After
+    public void tearDown() throws IOException {
+
+        Globals.prefs = null;
```
tobiasdiez: the `Globals.prefs = null` is  not required.

src/test/java/net/sf/jabref/logic/util/io/FileBasedTestCase.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    @Test
+    public void testFindAssociatedFiles() {
+
+        List<BibEntry> entries = Collections.singletonList(entry);
+        List<String> extensions = Arrays.asList("jpg", "pdf");
+        List<File> dirs = Arrays.asList(rootDir.resolve("graphicsDir").toFile(), rootDir.resolve("pdfs").toFile());
+
+        when(Globals.prefs.getBoolean(JabRefPreferences.AUTOLINK_EXACT_KEY_ONLY)).thenReturn(false);
```
tobiasdiez: As far as I can see, there is no need to mock `Globals.prefs`. A simple `preferences = mock(...)` and `findAssociatedFiles(..., preferences)` should work.

Is it much work to refactor `findAssociatedFiles` to accept a boolean `AUTOLINK_EXACT_KEY_ONLY` instead of the preference class?

src/test/java/net/sf/jabref/logic/util/io/FileBasedTestCase.java
Siedlerchr: In this case this is not the best idea, as the pahts to the file are os dependent resolved and differ on the location of the temp file etc.

Siedlerchr:
@simonharrer  The formerly ignored NPE test works now, too. 
If no more things, then I would squash and merge.

simonharrer:
Go ahead. Nice job! :) Maybe we can reduce the number of ignored tests over time.

PR #1597, opened 2016-07-16
Add fileEncoding property to utf8, to fix failing tests on windows

Siedlerchr:
Add fileEncoding property to utf8, to fix failing tests on windows

When I run ./gradlew check on Windows, all kind of tests fail due to wrong encoding.
This fixes the problem by explicitly setting the property to utf8

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -217,6 +217,7 @@ javadoc {
 }
 
 test {
+	systemProperty "file.encoding", "utf-8"
     testLogging {
         events "failed"
         exceptionFormat "full"


koppor:
We voted against it at #514. At that time, the test worked locally, too. It seems, we are relying on the tests on Travis only. Can the tests themselves be fixed?

[Simon said](https://github.com/JabRef/jabref/pull/514#issuecomment-164757267):

> I would vote for always reading only with UTF_8 set explicitly instead of relying on the default encoding.

oscargus:
Should we even support any other encoding than UTF-8? There are quite a few cases where the encoding must be specified in the source and at the moment it is a bit tricky to really know what to set. Get the encoding from preferences or go with UTF-8?

Siedlerchr:
The problem are mostly the tests. Some already set explicit UTF8, some don't,
As we therotically support other encodings in the bib files/exports/imports, it should be always used the Setting, in my humble opinion

simonharrer:
In general, you always have to specify the encoding for every I/O operation you do. Setting it through gradle would make the tests pass, but if a user uses a different platform dependent encoding, the code may brake. To detect these breakages, it is best to make everything explicit, and not change the environment.

simonharrer:
Another argument: I want to execute the tests within IDEs as well, without gradle. And they should pass there as well. Hence, just switching gradle to UTF-8 is no solution.

koppor:
Current status: tests run in IntelliJ, but not in gradle.

The tests should be self-contained. We should aim to fix the tests.

PR #1601, opened 2016-07-18
Move event

lenhard:
Moves the `event` package into `model`, as discussed in #1579

No need for changelog, since this has no externally visible effect. Commits will be squashed on merge.


oscargus:
LGTM! :+1:

PR #1604, opened 2016-07-19
Move preferences

lenhard:
Moves preferences-related classes into a new top-level package as outlined in #1579. This is a pure refactoring, no functional changes. It doesn't fix the dependencies on preferences, but separates them from the remaining classes in the root package.

Additionally, the class `JabRefPreferencesFilterDialog` is moved into the gui package, where it belongs. This also allows to rename the class and reduce its visibility.

Commits can be squashed on merge.


simonharrer:
LGTM

PR #1606, opened 2016-07-20
Updated preview entries

oscargus:
While working on #1602 I realized that the preview entries used deprecated fields, so I updated them and improved the casing a bit.

### DIFF
--- a/src/main/java/net/sf/jabref/logic/util/TestEntry.java
+++ b/src/main/java/net/sf/jabref/logic/util/TestEntry.java
@@ -6,28 +6,24 @@ import net.sf.jabref.model.entry.IdGenerator;
 
 public class TestEntry {
 
-    private static BibEntry entry;
-
-
     public static BibEntry getTestEntry() {
-        if (entry == null) {
-            entry = new BibEntry(IdGenerator.next(), "article");
-            entry.setField(FieldName.AUTHOR, "Smith, Bill and Jones, Bob and Williams, Jeff");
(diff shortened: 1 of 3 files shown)


simonharrer:
I would prefer a single class which contains a preview entry, which is then used for both style and preview.  What do you think?

oscargus:
Shouldn't be a problem and makes sense. I guess there are two for
historical reasons (plugin). I'm even wondering if there isn't a third one
although it is somehow hidden in the code and never seen... (Only used as
some sort of initialization.) Didn't find it yesterday though.

oscargus:
Now there is a single preview entry.

simonharrer:
I would always create a new bib entry instead of reusing it. Returning a copy is more robust, as you never know who modifies this entry, resulting in strange preview behavior. With this in place, just go ahead and merge.

src/main/java/net/sf/jabref/logic/util/TestEntry.java
```diff
@@ -0,0 +1,33 @@
+package net.sf.jabref.logic.util;
```
tobiasdiez: Not sure that it should life in logic.

oscargus: I'm open for suggestions on location. No dependency issues either way.

simonharrer: I merged it. We can discuss moving the class somewhere else somewhere else. ;)

PR #1607, opened 2016-07-20
Added LabelPatternPreferences

oscargus:
The `static` initializer is quite ugly, but I do not really know what the best way to change it is. Maybe call `updateDefaultPattern` in `JabRefMain` after setting up the preferences?


oscargus:
I added the initializer to JabRefMain. Again, a non-related test fails.

lenhard:
This might be a good occasion for some more improvements in `LabelPatternUtilTest`. Some test methods do not have descriptive names (e.g. `testCheckLegalKey2`) and @tobiasdiez would probably complain that there are two many asserts in the test methods. This is ok from my point of view in most of the methods, but perhaps you could split `testCheckLegalKey` into tests for the different accent types?

Apart from these things: LGTM

oscargus:
I think this test splitting better wait until the new checkLegalKey is implemented... I merge this for now, so the move can happen.

PR #1608, opened 2016-07-20
Added LayoutFormatterPreferences (and related files)

oscargus:

### DIFF
--- a/src/main/java/net/sf/jabref/gui/BasePanel.java
+++ b/src/main/java/net/sf/jabref/gui/BasePanel.java
@@ -117,6 +117,7 @@ import net.sf.jabref.logic.l10n.Encodings;
 import net.sf.jabref.logic.l10n.Localization;
 import net.sf.jabref.logic.labelpattern.LabelPatternUtil;
 import net.sf.jabref.logic.layout.Layout;
+import net.sf.jabref.logic.layout.LayoutFormatterPreferences;
 import net.sf.jabref.logic.layout.LayoutHelper;
 import net.sf.jabref.logic.util.UpdateField;
 import net.sf.jabref.logic.util.io.FileBasedLock;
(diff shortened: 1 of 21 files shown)


src/main/java/net/sf/jabref/gui/BasePanel.java
```diff
@@ -990,7 +991,8 @@ private void copyKeyAndTitle() {
                     "\\bibtexkey - \\begin{title}\\format[RemoveBrackets]{\\title}\\end{title}\n");
             Layout layout;
             try {
-                layout = new LayoutHelper(sr, Globals.prefs, Globals.journalAbbreviationLoader).getLayoutFromText();
+                layout = new LayoutHelper(sr, LayoutFormatterPreferences.fromPreferences(Globals.prefs),
```
tobiasdiez: I would also include the `journalAbbreviationLoader` in `LayoutFormatterPreferences`.

src/main/java/net/sf/jabref/gui/openoffice/OpenOfficePanel.java
```diff
@@ -126,8 +127,8 @@ private OpenOfficePanel() {
         update = new JButton(IconTheme.JabRefIcon.REFRESH.getSmallIcon());
         update.setToolTipText(Localization.lang("Sync OpenOffice/LibreOffice bibliography"));
         preferences = new OpenOfficePreferences(Globals.prefs);
-        loader = new StyleLoader(preferences, Globals.prefs, Globals.journalAbbreviationLoader,
-                Globals.prefs.getDefaultEncoding());
+        loader = new StyleLoader(preferences, LayoutFormatterPreferences.fromPreferences(Globals.prefs),
```
tobiasdiez: So if the OOPanel is open and I make some changes to the preferences, then `StyleLoader` is not changed, i.e. every action from the OOPanel still uses the old preferences, right?

tobiasdiez:
LGTM :+1:

src/main/java/net/sf/jabref/gui/openoffice/OpenOfficePanel.java
oscargus: Currently, yes.

src/main/java/net/sf/jabref/gui/BasePanel.java
oscargus: Yes, that may make sense. I made an attempt including in JournalAbbreviationPreferences, but that wasn't the correct place it turned out. Will try again.

oscargus:
I included `JournalAbbreviationLoader` in `LayoutPreferences`. Tests failing because of non-related connection issues.

PR #1611, opened 2016-07-21
Fix OutOfBoundsException when importing multiple entries in medline format

tschechlovdev:
@zellerdev found out a bug when importing more than one entry in the medline format into JabRef. 
JabRef version
JabRef 3.5
windows 8.1 6.3 amd64 
Java 1.8.0_25

Steps to reproduce:
1. Select more than entry from [here] (http://www.ncbi.nlm.nih.gov/pubmed) .
2. Select "Send to File" and then choose "Medline" Format.
3. Open Jabref and select "Import into new database".
4.  I get an IndexOutOfBoundException shown on the console and no entries are imported.

Alternatively use the testfiles I added in this PR.

Quick fix: Add an if clause around the substring method, that causes the Exception.


boceckts:
I can confirm the bug but next time please create an issue for this and assign it to yourself if you want to fix it. Also please use the "copy version to clipboard" functionallity in the About dialog in the help menu to easily paste your JabRef-, OS- and Java-version.

src/main/java/net/sf/jabref/importer/fileformat/MedlinePlainImporter.java
```diff
@@ -127,91 +126,93 @@ public ParserResult importDatabase(BufferedReader reader) throws IOException {
                     }
                 }
                 String entry = current.toString();
+                if (entry.contains("-")) {
```
boceckts: I think the while-loop above this is already trying to check the current line for validity and tries to skip all the invalid ones. I would propose to extract a new method called `private boolean checkLineValidity(String line)` and do all the checks in there such as: empty line, line doesn't contain "-", "-" symbol not at 4th position, etc...
then the while loop above can also be replaced and you don't need to modify the counter variable "j". You can then simply skip any invalid lines with:
```
if (!checkLineValidity(lines[j])) {
    continue;
}
```

boceckts: This would also mean you don't need to modify as much code as you did now because then you don't have the if-clause around all the code.

src/main/java/net/sf/jabref/importer/fileformat/MedlinePlainImporter.java
```diff
@@ -26,7 +26,6 @@
 import java.util.Locale;
 import java.util.Map;
 import java.util.regex.Pattern;
-
```
boceckts: Add the empty line again to fix Travis build.

src/main/java/net/sf/jabref/importer/fileformat/MedlinePlainImporter.java
tschechlovdev: But I wouldn't extract the while loop above, because in the while loop are also correcctions, like adding space if '-' is at third position. And putting the while loop into a method that returns a boolean could have side effects, since you don't expect, that this method is changing something.

boceckts: You could also extract to multiple methods, `private boolean first checkLineValidity(String line)`, if false try to correct it by calling another method that tries to insert spaces or similar things.

src/main/java/net/sf/jabref/importer/fileformat/MedlinePlainImporter.java
```diff
@@ -227,13 +231,20 @@ public ParserResult importDatabase(BufferedReader reader) throws IOException {
...
         }
 
         return new ParserResult(bibitems);
 
     }
 
+    private boolean checkLineValidity(String line) {
+        if (line.length() >= 4) {
+            return line.contains("-") || !(line.charAt(4) == '-') ? true : false;
```
boceckts: `!(line.charAt(4) == '-')` this will throw an OutOfBoundsException if the length of the line is exactly 4. Please cover this case.

boceckts: Also please check the if statement again, I can't find a case where this will return `false`.

boceckts: `? true : false` is unnecessary.

src/main/java/net/sf/jabref/importer/fileformat/MedlinePlainImporter.java
```diff
@@ -212,6 +215,7 @@ public ParserResult importDatabase(BufferedReader reader) throws IOException {
                     }
                     comment = comment + value;
                 }
+
```
boceckts: Why did you add this line?

src/main/java/net/sf/jabref/importer/fileformat/MedlinePlainImporter.java
```diff
@@ -227,13 +231,20 @@ public ParserResult importDatabase(BufferedReader reader) throws IOException {
             // create one here
             b.setField(fields);
             bibitems.add(b);
-
```
boceckts: Why did you remove this line?

tschechlovdev: because it was an unnecessary line.

src/main/java/net/sf/jabref/importer/fileformat/MedlinePlainImporter.java
```diff
@@ -238,11 +237,7 @@ public ParserResult importDatabase(BufferedReader reader) throws IOException {
     }
 
     private boolean checkLineValidity(String line) {
-        if (line.length() >= 4) {
-            return line.contains("-") || (line.charAt(4) != '-') ? true : false;
-        } else {
-            return false;
-        }
+        return (line.length() >= 5) ? line.contains("-") && (line.charAt(4) == '-') : false;
```
boceckts: I think `line.contains("-")` is unnecessary when you also check for it's position.

src/main/java/net/sf/jabref/importer/fileformat/MedlinePlainImporter.java
```diff
@@ -237,7 +237,7 @@ public ParserResult importDatabase(BufferedReader reader) throws IOException {
     }
 
     private boolean checkLineValidity(String line) {
-        return (line.length() >= 5) ? line.contains("-") && (line.charAt(4) == '-') : false;
+        return (line.length() >= 5) ? (line.charAt(4) == '-') : false;
```
matthiasgeiger: `return ((line.length() >= 5) && (line.charAt(4) == '-'))`  - why not as simple as that? :wink:

tschechlovdev:
No remarks?

src/test/java/net/sf/jabref/importer/fileformat/MedlinePlainImporterTest.java
```diff
@@ -198,6 +198,14 @@ public void testWithNbibFile() throws IOException, URISyntaxException {
     }
 
     @Test
+    public void testWithMultipleEntries() throws IOException, URISyntaxException {
+        Path file = Paths
+                .get(MedlinePlainImporter.class.getResource("MedlinePlainImporterStringOutOfBounds.txt").toURI());
+        List<BibEntry> entries = importer.importDatabase(file, Charset.defaultCharset()).getDatabase().getEntries();
```
Siedlerchr: When the bib file is in UTF8, then please set the Charset explicit to UTF8, too. `DefaultCharsets.UTF8` (or similar), otherwise test will fail on windows in gradle.

Siedlerchr:
In general good. :+1:  But I would recommend adding some test with invalid formats, e.g. if an exceptions is thrown (expected) or whatever.

src/test/java/net/sf/jabref/importer/fileformat/MedlinePlainImporterTest.java
tschechlovdev: done.

tschechlovdev:
Hm an Excpetion should be not thrown, except if the reader would be null. If the format is invalid then the entries list will be empty, but no exception will be thrown. I'll added tests for both.

tschechlovdev:
Any other remarks?

stefan-kolb:
LGTM, just one remark: can you keep the test files as minimal as needed inside the tests? No need for 300 attributes when you can recreate the test issue with one or two.

tschechlovdev:
Ok, I've minimized the testfile for the issue.

stefan-kolb:
Maybe fix the two Codacy warnings (https://www.codacy.com/app/simonharrer/jabref/pullRequest?prid=289709) and add a key to the @misc entries. Otherwise im :+1: We need a second review from a @JabRef/developers tho.

src/test/java/net/sf/jabref/importer/fileformat/MedlinePlainImporterTest.java
```diff
@@ -201,11 +204,27 @@ public void testWithNbibFile() throws IOException, URISyntaxException {
...
+        List<BibEntry> entries = importer.importDatabase(file, Charsets.UTF_8).getDatabase().getEntries();
         BibEntryAssert.assertEquals(MedlinePlainImporter.class, "MedlinePlainImporterStringOutOfBounds.bib", entries);
     }
 
     @Test
+    public void testInvalidFormat() throws URISyntaxException, IOException {
+        Path file = Paths
+                .get(MedlinePlainImporter.class.getResource("MedlinePlainImporterTestInvalidFormat.xml").toURI());
+        List<BibEntry> entries = importer.importDatabase(file, Charsets.UTF_8).getDatabase().getEntries();
+        assertEquals(Collections.EMPTY_LIST, entries);
```
Siedlerchr: Use the method Collections,emptyList() rather than the constant, because the constant does not return a generic list and is therefore not typesafe.
(In this case not so important, but could possinly produce an unchecked warning)

src/test/java/net/sf/jabref/importer/fileformat/MedlinePlainImporterTest.java
```diff
@@ -201,11 +204,27 @@ public void testWithNbibFile() throws IOException, URISyntaxException {
...
+    }
+
+    @Test(expected = NullPointerException.class)
+    public void testNullReader() throws IOException {
+        try (BufferedReader reader = null) {
+            ParserResult entries = importer.importDatabase(reader);
```
Siedlerchr: For the codacy issue just remove the assignment

Siedlerchr:
Just the collections,emptyList and the codacy and it's fine :+1:  ( I already reviewd the main parts before)

src/test/java/net/sf/jabref/importer/fileformat/MedlinePlainImporterTest.java
tschechlovdev: done.

src/test/java/net/sf/jabref/importer/fileformat/MedlinePlainImporterTest.java
tschechlovdev: done.

PR #1615, opened 2016-07-22
Always use https for help files

stefan-kolb:


oscargus:
Do we have a properly signed certificate? Adding exceptions are quite complicated otherwise (I noted after reinstalling as DOAJ apparently uses a self-signed certificate). I assume we do and then :+1:.

stefan-kolb:
We have a Cloudflare certificate.

simonharrer:
LGTM just merge

PR #1618, opened 2016-07-22
Moved, removed, and used String constants

oscargus:
Moved some String constants from `Globals` to better places. Maybe we should introduce a class `logic.util.string.StringConstant`? Not sure if there is a better long term plan for `Globals.NEWLINE`? 

Unified used of the constant `"/"` as field separator.

Removed unused preference key and default value in `PersistenceTableColumnListener`.

### DIFF
--- a/src/main/java/net/sf/jabref/importer/EntryFromPDFCreator.java
+++ b/src/main/java/net/sf/jabref/importer/EntryFromPDFCreator.java
@@ -14,9 +14,9 @@ import net.sf.jabref.external.ExternalFileTypes;
 import net.sf.jabref.gui.IconTheme;
 import net.sf.jabref.logic.xmp.XMPUtil;
 import net.sf.jabref.model.entry.BibEntry;
-import net.sf.jabref.model.entry.InternalBibtexFields;
 import net.sf.jabref.pdfimport.PdfImporter;
 import net.sf.jabref.pdfimport.PdfImporter.ImportPdfFilesResult;
+import net.sf.jabref.preferences.JabRefPreferences;
 
 import org.apache.pdfbox.pdmodel.PDDocument;
 import org.apache.pdfbox.pdmodel.PDDocumentInformation;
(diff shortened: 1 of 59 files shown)


simonharrer:
LGTM in general, as it improves the code. I am uncertain about the StringUtil.NEWLINE constant. Maybe IO.NEWLINE would be better? Also, we should try to encapsulate these constants, e.g., DIRECTORY, within classes that provide methods which automatically handle the constant usage internally. But this does not have to be implemented right now.

oscargus:
I moved NEWLINE (and FILE_SEPARATOR) to OS. This is probably a better location than StringUtil anyway and both actually are OS dependent...

oscargus:
There are two more Strings to potentially move from Globals:

```
    // Signature written at the top of the .bib file.
    public static final String SIGNATURE = "This file was created with JabRef";
    public static final String ENCODING_PREFIX = "Encoding: ";
```

`SIGNATURE` is not written anymore and is only used in `BibtexImporter` so makes sense to move it there and make it `private`.

`ENCODING_PREFIX` is used for both import and export. `SavePreferences`?

Moving these would remove quite a few of the remaining dependencies from `logic` to `Globals`.

src/main/java/net/sf/jabref/importer/EntryFromPDFCreator.java
```diff
@@ -97,7 +97,7 @@ private void addEntryDataFromPDDocumentInformation(File pdfFile, BibEntry entry)
                         // default time stamp follows ISO-8601. Reason: https://xkcd.com/1179/
                         String date = new SimpleDateFormat("yyyy-MM-dd")
                                 .format(creationDate.getTime());
-                        appendToField(entry, InternalBibtexFields.TIMESTAMP, date);
+                        appendToField(entry, Globals.prefs.get(JabRefPreferences.TIME_STAMP_FIELD), date);
```
lenhard: Why is this field name moved from model to `JabRefPreferences`?

oscargus: Because the field name is configurable in the preferences, so bad idea to
hard code it in a file. ;-)

lenhard: Ok, I see. We might consider to drop this configuration option at some point in time. I don't really see why you might want to configure this.

lenhard:
Very good PR in my point of view! This really kills a lot of ugly dependencies to `Globals` and moves constants to better places. I had been thinking for a while about `NEWLINE`, the new location is perfect!

src/main/java/net/sf/jabref/importer/EntryFromPDFCreator.java
oscargus: I agree. Moving ahead to date-added, date-modified (#1575, comment and https://github.com/koppor/jabref/issues/130) will sort this out.

PR #1620, opened 2016-07-23
Protected terms management dialog

oscargus:
The idea is that the lists with protected terms should be user configurable. This PR adds the mechanism for this as well as a dialog to manage the lists.

Eventually the internal lists should be managed as resources to simplify adding more.


Siedlerchr:
Good idea! Please provide a screenshot.

oscargus:
Preliminary screen shot:
capture6

"Enabled" is change by double-clicking or right-click menu. I may add buttons as well (but may prefer a checkbox column, not sure how to implement that easily though)
.
Problem: `ProtectTermsLoader` should be provided to the formatters, but the static initialization in `JabRefPreferences` and `Formatters` makes it troublesome. This is really a quite good case where the formatter indeed needs some information from the preferences...

src/main/resources/l10n/JabRef_da.properties
```diff
@@ -1714,3 +1714,18 @@ Executing_command_\"%0\"...=
...
+Add_protected_term_file=
+Add_term_file=
+Are_you_sure_you_want_to_remove_the_term_list?=
+Countries_and_territories_in_English=
+Electrical_engineering_terms=
+Enabled=
+Internal_list=
+Manage_term_files=
+Months_and_weekdays_in_English=
+No=
```
koppor: Aren't there any other translated strings for "yes" or "no"? Could we be more specific somehow? See our discussion on the database-changed-what-do-you-want-to-do-dialog.

koppor: Maybe we just can use a checkbox for enabled/disabled?

oscargus:
New screen shot: 
capture7

Width of the "Enabled" column is fixed.

This is sort of done now. Remaining issues (from my side):

[ ] Tests
[ ] The formatter doesn't work because of the static initialization in `Formatter` (no idea how to solve this really...)
[ ] Maybe one should remove possible duplicates from the final list

src/main/resources/l10n/JabRef_da.properties
oscargus: Checkboxes are now used.

oscargus:
I pushed to upstream now, so you can try it out at http://builds.jabref.org/casekeeper in a while. Note that the actual formatting is not working...

oscargus:
Looks like there are no builds if the PR is pushed to upstream at a later point...

jkulesza:
Thank you for doing this!   The field I work in (like all fields, I'm sure) has specific terms that JabRef doesn't recognize that I have to constantly repair if I want to take advantage of some of its automated case management.

oscargus:
The idea was that this should have been included much earlier. :-)

Updated version with right-click menu support for easier management.

capture11

Also possible to create new files from the management dialog

capture10

I'm quite happy at the moment, so I will rebase and set "Ready for review".

oscargus:
Conflicts in translations. Anything else missing here?

src/main/java/net/sf/jabref/logic/protectedterms/ProtectedTermsList.java
```diff
@@ -0,0 +1,117 @@
...
+        return true;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (!(o instanceof ProtectedTermsList)) {
+            return false;
+        }
+        ProtectedTermsList otherList = (ProtectedTermsList) o;
+        return (this.location == otherList.location) && (this.description == otherList.description);
```
Siedlerchr: You are comparing strings with == here!
Use Objects.equals()

src/main/java/net/sf/jabref/logic/protectedterms/ProtectedTermsParser.java
```diff
@@ -0,0 +1,119 @@
...
+    /**
+     * Read the given file, which should contain a list of journal names and their
+     * abbreviations. Each line should be formatted as: "Full Journal Name=Abbr. Journal Name"
+     *
+     * @param in
+     */
+    private void readTermsList(Reader in) {
+        try (BufferedReader reader = new BufferedReader(in)) {
+            String line;
+            while ((line = reader.readLine()) != null) {
```
Siedlerchr: You could use a Stream with Files.lines and then apply a filter to filter out the comment line: Either directly or when the list is complete get the first/last occurrence of a comment.

http://stackoverflow.com/questions/19605151/using-java-8-what-is-the-most-preferred-and-concise-way-of-printing-all-the-lin
https://www.mkyong.com/java8/java-8-streams-filter-examples/

Siedlerchr:
Apart from the equals, LGTM

PR #1621, opened 2016-07-24
Lookup BibEntry from ISBN and merge information

oscargus:
Should fix the translations and just add ISBN where applicable.


src/main/java/net/sf/jabref/gui/mergeentries/MergeEntryDOIDialog.java
```diff
@@ -50,46 +51,75 @@
     private final BasePanel panel;
```
tobiasdiez: Rename class to MergeEntryDialog?

src/main/java/net/sf/jabref/gui/mergeentries/MergeEntryDOIDialog.java
```diff
@@ -50,46 +51,75 @@
...
                     Localization.lang("This operation requires exactly one item to be selected."),
-                    Localization.lang("Merge entry with DOI information"), JOptionPane.INFORMATION_MESSAGE);
+                    Localization.lang("Merge entry with DOI/ISBN information"), JOptionPane.INFORMATION_MESSAGE);
             this.dispose();
             return;
         }
 
         this.originalEntry = panel.getSelectedEntries().get(0);
-        panel.output(Localization.lang("Fetching info based on DOI"));
         Optional<String> doi = this.originalEntry.getFieldOptional(FieldName.DOI);
```
tobiasdiez: I would restructure the flow a bit: The dialog just gets two BibEntries which should be compared/merged. So it's the callers responsibility to provide the second entry (i.e. fetch from ISBN / DOI /...)

src/main/java/net/sf/jabref/importer/fetcher/ISBNtoBibTeXFetcher.java
```diff
@@ -61,14 +62,26 @@ public void stopFetching() {
...
+        Optional<BibEntry> entry = getEntryFromISBN(query, status);
+        if (entry.isPresent()) {
+            inspector.addEntry(entry.get());
+            return true;
+        }
+        return false;
+
+    }
+
+    public Optional<BibEntry> getEntryFromISBN(String query, OutputPrinter status) {
```
tobiasdiez: Could you please implement the new `IdBasedFetcher` interface (i.e. throw a `FetcherException` whenever the output printer was used). I think this makes in particular sense since you pass `null` as the second argument above (which would result in a NPE in case of an error)

src/main/resources/l10n/JabRef_de.properties
```diff
@@ -2047,14 +2047,14 @@ Fetching_info_based_on_DOI=Rufe_Informationen_basierend_auf_DOI_ab
...
-Merge_entry_with_DOI_information=Eintrag_mit_DOI-Informationen_zusammenführen
+Merge_entry_with_DOI/ISBN_information=Eintrag_mit_DOI-Informationen_zusammenführen
 No_information_added=Keine_Informationen_hinzugefügt
 Opening_dialog=Öffne_Dialog
 Original_entry=Originaleintrag
 Replace_original_entry=Ersetze_Originaleintrag
 Right_entry=Rechter_Eintrag
 Select_at_least_one_entry_to_manage_keywords.=Es_muss_mindestens_ein_Eintrag_ausgewählt_sein_um_Stichworte_verwalten_zu_können.
-Updated_entry_with_info_from_DOI=Eintrag_wurde_mit_DOI-Information_aktualisiert.
+Updated_entry_with_info_from_DOI/ISBN=Eintrag_wurde_mit_DOI-Information_aktualisiert.
```
tobiasdiez: Now all the translations with changed key but unchanged value are invalid.

oscargus: Yes, that was my main reason for not putting "Ready for review". I'll sort it out eventually.

src/main/java/net/sf/jabref/gui/mergeentries/MergeEntryDOIDialog.java
oscargus: There is already one with that name. :-)

src/main/java/net/sf/jabref/gui/mergeentries/MergeEntryDOIDialog.java
oscargus: The main difference between MergeEntryDialog and this is the number of options one has at the end. However, one may of course call the MergeEntryDOIDialog something like `MergeFetchedEntryDialog` to make it more generic and, as you say, use an entry from a database and a fetch, temporary, entry.

tobiasdiez: I think one should merge the two merge entries dialogs (pun intended :smile: ). They actually share a lot of common code and the differences are quite small (and can be controlled by passing a boolean as an option to the constructor)

oscargus:
I pass on rewriting to the new interface at the moment, but there is a rewrite of the code to make it very simple to add merging based on other fields.

I'll make one small fix (add ISBN as a FieldExtra) and then it is ready for review.

oscargus:
With the current code structure it would be simple to allow marking of several entries and loop through them. Main issue I would guess is that there is no way to effectively canceling this operation at the moment, one will just have to go through everything and it is not straightforward to solve that part.

Should I still implement it?

simonharrer:
I think the looping through multiple entries should be done in another PR. And we have to think about how to handle this in general. The user should be able to abort such actions if they require user interaction per entry.

oscargus:
I agree that there should be some better mechanism for canceling multiple entries.

The conflicts are all for translations. I wait with rebasing until this gets an OK to be merged. 

Btw, I added merging from ArXiv as well to this PR.

simonharrer:
To me, it looks good. But I think the IdBasedFetcher would be really helpful to implement here. However, this can be done in another PR.

src/main/java/net/sf/jabref/gui/BasePanel.java
```diff
@@ -652,7 +653,18 @@ public void update() {
...
-        actions.put(Actions.MERGE_DOI, (BaseAction) () -> new MergeEntryDOIDialog(BasePanel.this));
+        actions.put(Actions.MERGE_WITH_FETCHED_ENTRY, (BaseAction) () -> {
+            if (mainTable.getSelectedEntries().size() == 1) {
+                BibEntry originalEntry = mainTable.getSelectedEntries().get(0);
+                new FetchAndMergeEntry(originalEntry, this,
+                        Arrays.asList(FieldName.DOI, FieldName.ISBN, FieldName.EPRINT));
+            } else {
+                JOptionPane.showMessageDialog(frame(),
+                        Localization.lang("This operation requires exactly one item to be selected."),
+                        Localization.lang("Merge entry with %0 information", "DOI/ISBN"),
```
tobiasdiez: "with fetched information" ? since DOI, ISBN and now eprint is supported.

src/main/java/net/sf/jabref/gui/mergeentries/FetchAndMergeEntry.java
```diff
@@ -0,0 +1,74 @@
...
+
+    public FetchAndMergeEntry(BibEntry entry, BasePanel panel, List<String> fields) {
+        for (String field : fields) {
+            Optional<String> fieldContent = entry.getFieldOptional(field);
+
+            // Get better looking name for status messages
+            String type;
+            switch(field) {
+            case FieldName.DOI:
+                type = "DOI";
```
tobiasdiez: maybe move this constant to FieldName.DOI_DISPLAY_NAME?

src/main/java/net/sf/jabref/gui/mergeentries/FetchAndMergeEntry.java
```diff
@@ -0,0 +1,74 @@
...
+                break;
+            case FieldName.ISBN:
+                type = "ISBN";
+                break;
+            case FieldName.EPRINT:
+                type = "EPrint";
```
tobiasdiez: "E-Print"?

src/main/java/net/sf/jabref/gui/mergeentries/FetchAndMergeEntry.java
```diff
@@ -0,0 +1,74 @@
...
+                // Get entry based on field
+                if (FieldName.DOI.equals(field)) {
+                    fetchedEntry = new DOItoBibTeXFetcher().getEntryFromDOI(fieldContent.get());
+                } else if (FieldName.ISBN.equals(field)) {
+                    fetchedEntry = new ISBNtoBibTeXFetcher().getEntryFromISBN(fieldContent.get(), null);
+                } else if (FieldName.EPRINT.equals(field)) {
+                    try {
+                        fetchedEntry = new ArXiv().performSearchById(fieldContent.get());
+                    } catch (FetcherException e) {
+                        // Ignore for now
```
tobiasdiez: better print out error message as below

src/main/java/net/sf/jabref/gui/entryeditor/FieldExtraComponents.java
```diff
@@ -239,6 +243,56 @@ private void checkDoi() {
     }
 
     /**
+     * Add button for fetching by ISBN
+     *
+     * @param fieldEditor
+     * @param panel
+     * @return
+     */
+    public static Optional<JComponent> getIsbnExtraComponent(BasePanel panel, EntryEditor entryEditor,
```
tobiasdiez: Is there a similar button for the ArXiv?

tobiasdiez:
In general, this PR looks good to me 👍 Just a few minor remarks.

By the way, I still don't get why there are two different dialogs for comparing/merging entries (MergeEntryDialog vs the new MergeFetchedEntryDialog)

oscargus:
I changed things based on your comments, but will not push until Thursday.

I wasn't completely sure if ArXiv was the only thing that used eprint so
didn't want to add the button before clarifying that. Now I introduced a
more general approach so will be easier to do that now.

Historical reasons. There are some differences in the handling of entries
at the end though which may complicate merging them. Most of the code is in
MergeEntries which generates the table though.

src/main/java/net/sf/jabref/gui/menus/RightClickMenu.java
```diff
@@ -278,6 +279,29 @@ private boolean isFieldSetForSelectedEntry(String fieldname) {
...
+            return entry.getFieldNames().containsAll(fieldnames);
+        } else {
+            return false;
+        }
+    }
+
+    private boolean isAnyFieldSetForSelectedEntry(List<String> fieldnames) {
+        if (panel.getMainTable().getSelectedRowCount() == 1) {
+            BibEntry entry = panel.getMainTable().getSelected().get(0);
+            for (String fieldname : fieldnames) {
```
Siedlerchr: Simple one liner: `! Collections.disjoint(list1, list2); `
From the apidoc:  disjoint (a, b) Returns true if the two specified collections have no elements in common.

oscargus: Great! There should have been an easier way...

PR #1623, opened 2016-07-24
Table row height adjusts on Windows as the font scales with the menu

oscargus:
When increasing the menu font size on Windows (at least WIndows 10), the tables also scale but not the row width. This has been fixed in some dialogs earlier, but this is a general fix.

I have no idea how it works on OSX, but on Linux (my CentOS office machine) it seems like the table font is not the same as the menu font.

This is relevant for high resolution displays (although I cannot find the issue that discussed it...)

Before:
capture8

After:
capture9

### DIFF
--- a/src/main/java/net/sf/jabref/collab/ChangeDisplayDialog.java
+++ b/src/main/java/net/sf/jabref/collab/ChangeDisplayDialog.java
@@ -35,10 +35,13 @@ import javax.swing.event.TreeSelectionEvent;
 import javax.swing.event.TreeSelectionListener;
 import javax.swing.tree.DefaultMutableTreeNode;
 
+import net.sf.jabref.Globals;
 import net.sf.jabref.gui.BasePanel;
 import net.sf.jabref.gui.undo.NamedCompound;
 import net.sf.jabref.logic.l10n.Localization;
+import net.sf.jabref.logic.util.OS;
 import net.sf.jabref.model.database.BibDatabase;
+import net.sf.jabref.preferences.JabRefPreferences;
 
 class ChangeDisplayDialog extends JDialog implements TreeSelectionListener {
 
@@ -64,6 +67,9 @@ class ChangeDisplayDialog extends JDialog implements TreeSelectionListener {
         }
         tree = new JTree(root);
         tree.addTreeSelectionListener(this);
+        if (OS.WINDOWS) { // Arbitrary trees scale with menu font size on Windows
+            tree.setRowHeight(Globals.prefs.getInt(JabRefPreferences.MENU_FONT_SIZE) + 2);
+        }
         JSplitPane pane = new JSplitPane();
         pane.setLeftComponent(new JScrollPane(tree));
         JPanel infoBorder = new JPanel();
--- a/src/main/java/net/sf/jabref/collab/ChangeDisplayDialog.java
+++ b/src/main/java/net/sf/jabref/collab/ChangeDisplayDialog.java
@@ -16,6 +16,7 @@
 package net.sf.jabref.collab;
 
 import java.awt.BorderLayout;
+import java.awt.FontMetrics;
 import java.awt.Insets;
 import java.util.Collections;
 import java.util.Enumeration;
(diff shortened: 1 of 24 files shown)


simonharrer:
:+1: Really nice. What I am concerned about is the number of places we have to put this check. We should really think of a way where configuration changes can flow from the preferences to the GUI elements. Maybe use a event bus for the preferences as well, especially regarding the classes in the GUI package? This could maybe reduce the number of restarts that are required upon changing preferences. What do the other developers think? @JabRef/developers

oscargus:
It is a bit weird that the rows are not automatically resized on Windows. It seems like this happens for the heading but not the content rows. Maybe easier to send in a bug report to Oracle/Microsoft?

Siedlerchr:
We are not the first who stumble across this issue. And I doubt that it will be fixed for Swing. 
Fpr high dpi things, a  bit searching on SO found that source, which compares different sizes and dpi settings, including JavaFX:
http://kynosarges.org/GuiDpiScaling.html

oscargus:
This isn't really the same thing as in the link, although it is somehow
related as it tries to solve the same problem. Still, bad rendering when
changing the font size is a fundamental issue. (Or is there some automatic
size parameter that must be enabled?)

I also doubt that they will fix it, though. Although it is probably quite
easy as it works for the heading row, so just copy whatever different row
height calculation is there.

oscargus:
There is a better solution though (maybe better Googling before coding):

http://stackoverflow.com/questions/15568054/how-to-change-jtable-row-height-globally

I'll try this out. Guess it is look-and-feel dependent but worth a try.

matthiasgeiger:
As there was recently a issue or forum discussion regarding this topic I just want to make sure that you are aware of two different configurations: There is the option to increase/decrease the table font size using the "View" menu and there is also the "Menu and label font size" in Preferences -> Appearance. The latter setting not only affects the "Menu and label font size" but also sets the font size in the entry editor and potentially at other places...

And I just saw that there is also a "Table row height padding" setting which might be reason for the strange height behavior...?

oscargus:
The current behavior is affected by "Menu and label font size".

As far as I am concerned "Table row height padding" is only for the main entry table, as well as the explicit font setting and increase/decrease font size.

In the link there are arguments that this is a feature, so they will indeed not change it. (It appears as if the "ideal" default table row height is different on different OS to start with, so they decided to just use the same constant for all systems...)

oscargus:
Found a consistent way to do this. It also applied for JTree. I think that the "global" approach is Look-and-feel-dependent so this seems to be the best way to do it.

src/main/java/net/sf/jabref/collab/ChangeDisplayDialog.java
```diff
@@ -64,6 +65,10 @@ public ChangeDisplayDialog(JFrame owner, final BasePanel panel,
         }
         tree = new JTree(root);
         tree.addTreeSelectionListener(this);
+        // Fix tree row height
+        FontMetrics metrics = tree.getFontMetrics(tree.getFont());
```
koppor: Can these two lines be separated into a utility method, which gets `tree` passed? (At the other places, the appropriate other object; `table`, ...)

koppor:
LGTM. Please extract the two lines of code to a separate method and then it's good to go (from my side :innocent:).

src/main/java/net/sf/jabref/collab/ChangeDisplayDialog.java
oscargus: I realized that as well after changing everywhere... ;-)

src/main/java/net/sf/jabref/gui/util/GUIUtil.java
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    /**
+     * Update table row height to the best possible considering font size changes on the current platform
+     *
+     * http://bugs.sun.com/bugdatabase/view_bug.do?bug_id=4760081
+     *
+     * @param table
+     */
+    public static void correctRowHeight(JTable table) {
```
koppor: Is it also possible to pass [JComponent](https://docs.oracle.com/javase/8/docs/api/javax/swing/JComponent.html) and merge this method with `correctRowHeight(JTree tree)`?

oscargus: No. JComponent doesn't have a setRowHeight method (I checked it before writing the code) and I do not think it is possible to cast from `JTree` to `JTable` or the opposite.

Issue #1272, opened 2016-04-21
bibtex key entry rejects ( and )


wbmacleod:
JabRef version 3.3 Linux
Hint: If you use a [development version](http://builds.jabref.org/master/), ensure that you use the latest one.

Steps to reproduce:

1. Try to enter a bibtex key with parenthesis in it - eg: Kaur(2014)
2. Invalid bibtex key is returned

Also this implies that the automatic bibtex key generator with parenthesis no longer works (I have 4000 keys with parenthesis...)

Work around - enter key manually in bibtex tex window.

I checked and ( and ) ARE valid in a bibtex key.

This bug was introduced between 3.2 and 3.3.

wbmacleod:
With some checking it seems that biber does not allow for parentheses. See:

http://tex.stackexchange.com/questions/96454/using-bibtex-keys-containing-parentheses-with-biber

A solution would be to allow parentheses in bibtex mode, and then in bibalatex mode add an IDS field that would move offending keys to IDS field and remove offending characters.  Many of us have thousands of references with parenthesis in the keys, so changing our bibtex files would be lots of of work....

oscargus:
Introdced in d505452d04b421818dd966a85121468138de8ff3 

@lenhard : comments?

lenhard:
The reason for this was #631 

Hard to say what the right path of action is here. If brackets are allowed in bibtex keys, we should support such keys in JabRef. I guess we could use different key checking functionality depending on the mode, but I am strongly against workarounds such as the one described by @wbmacleod If one uses brackets in keys and switches to biblatex, it should be sufficient to display an error. But I have no strong opinion here. @koppor?

The most simple solution would be to just revert the commit.

As a side note: Removing brackets in thousands of bibtex keys is not too hard with some automation.

wbmacleod:
I would agree that the workaround is not a good idea - a better solution might be to use the flag of bibtex vs bibalatex/biber in jabref to also affect the syntax checking - in bibtex mode any bibtex allowed citation could  be permitted.

With regards to changing the cite key, this is very non-trivial - I have dozens of co-authors and RAs that use a common bibtex file, so this requires changing dozens of bibtex files and keys in latex papers - it would cause several years of problems as the changes propagate through the teams all over the place.

Also, I have been trying to get Columbia students to use jabref, and I think killing backwards compatibility in this way is not helpful to getting jabref more widely used.  A better way would be to allows full compatibility with bibtex (which is a very stable system...).

[...]

PR #1626, opened 2016-07-25
Move labelpattern

lenhard:
Minor refactoring and moving of classes from logic into model. The `XXXLabelPattern` classes really describe how we serialize and store label patterns (part of the structure of meta data). Hence, they belong into model and have already been almost independent of other classes. The util class with helpful functions regarding label patterns can stay in util.


src/main/java/net/sf/jabref/model/labelpattern/GlobalLabelPattern.java
```diff
@@ -13,15 +13,31 @@
...
 public class GlobalLabelPattern extends AbstractLabelPattern {
 
+    private static List<String> defaultLabelPattern;
+
+    static {
+        updateDefaultPattern();
+    }
+
+    public static void updateDefaultPattern() {
+        defaultLabelPattern = AbstractLabelPattern.split(JabRefPreferences.getInstance().get(JabRefPreferences.DEFAULT_LABEL_PATTERN));
```
oscargus: See #1607. :-) Not sure what is the best merging order...

lenhard: @simonharrer Good idea
@oscargus Oh no, now I have already rewritten `GlobalLabelPattern` :-(

However, it seems you did not really touch the class in #1607? From what I have seen and tested with the GUI, you can completely drop the static part of `GlobalLabelPattern`. It is just initialized in two places (the gui and the preferences) and both of these places have access to the gui, so you can provide the correct parameters right away.

Regarding merging order: It prefer mine to be merged first :P Anyway, let Simon decide!

simonharrer:
LGTM

@oscargus how does this affect #1607 ?

oscargus:
No idea. Probably a bit of resolving independent of who merges first. The feeling is that it is easier to merge the PR without moving of files first, but I'm not sure if that is true or if I just want less work. ;-)

oscargus:
By the way, there is a preference setting "Enforce legal characters in BibTeX keys". Is that really needed?

The difference is the characters in `"{}(),\\\"#~^'"` but not in `"{}(),\\\""` so `#~^'`. I'm wondering if we upset many people by always enforcing legal characters?

lenhard:
@oscargus: My changes were quite small and would be relatively easy to redo, so it's ok for me if you go first.

Regarding the BibTex keys: We have already upset at least one person in #1272. We talked about this issue on the devcall right now and arrived at the following decision:
 - The preference setting "Enforce legal characters in BibTeX keys" can be removed, hence we will only enforce `"{}(),\\\"#~^'"`.
 - Key validation should be based on the mode (BibTeX, BibLaTeX). Especially BibLaTeX is more strict and there should be an additional validation that checks for the symbols that are not allowed in BibLaTeX

oscargus:
OK! If anyone could have a look at the PR I'll merge it.

So basically, the (in the PR introduced) boolean argument to `checkLegalKey` should be replaced with isBibLatex? Is it clear what the set of symbols are for the two cases?

tobiasdiez:
And we considered moving the check to the integrity dialog.

lenhard:
Should be good to go. 

Just as a side note: What are the "default" buttons in Preferences-> BibTeX key generator supposed to do? For me, only the button for the default pattern works and actually resets something. The other default buttons basically just delete what I have entered into the corresponding text field. This behavior is independent of this PR.

oscargus:
I think the default is that every type has the same key pattern, so it makes sense that the pattern is cleared when clicking Default.

oscargus:
LGTM! :+1:

matthiasgeiger:
I'm not sure whether we discussed it in this context at the last devcall, but moving those classes will lead to another node in the registry for storing those preferences - if the storing mechanism is not adjusted.

lenhard:
@matthiasgeiger: Are you sure? Didn't that already happen through #1635 by introducing the new `LabelPatternPreferences`? As far as I understand, the old label preferences get lost through this change, but will not so easily get lost in the future, since they are now normal preferences.

matthiasgeiger:
Just tested it: v3.5 and current master reads and writes from/to  `HKEY_CURRENT_USER\SOFTWARE\JavaSoft\Prefs\net\sf\jabref\logic\labelpattern` - your branch reads and writes from/to `HKEY_CURRENT_USER\SOFTWARE\JavaSoft\Prefs\net\sf\jabref\model\labelpattern`

matthiasgeiger:
Caused by: https://github.com/JabRef/jabref/blob/master/src/main/java/net/sf/jabref/preferences/JabRefPreferences.java#L1151

Maybe we should discuss again how to implement this better and provide migrations this time ;-)

lenhard:
Migrations are overrated :-)

We should at least change that to 

`Preferences pre = Preferences.userNodeForPackage(JabRefMain.class);`

to avoid future issues.

Although not part of this PR, let us keep this open and discuss at the next devcall on how to proceed.

matthiasgeiger:
Another topic for the devcall: Can we rename this "LabelPattern..." stuff to "BibtexKeyPattern..." ? - I never understood why this is called "label pattern" :wink:

lenhard:
TODO in new PR after this one is merged:
 - Integrate label pattern preferences tree into global preferences tree
 - Rename label pattern to bibtex key pattern.
 - Possible migration of old key patterns

PR #1627, opened 2016-07-25
More field names and a method

oscargus:
More field names in `FieldName` and a method `FieldName.orFields(FieldName.YEAR, FieldName.DATE)` to generate `"year/date"`.

### DIFF
--- a/src/main/java/net/sf/jabref/model/entry/FieldName.java
+++ b/src/main/java/net/sf/jabref/model/entry/FieldName.java
@@ -1,6 +1,5 @@
 package net.sf.jabref.model.entry;
 
-
 public class FieldName {
 
     // Character separating field names that are to be used in sequence as
@@ -22,6 +21,7 @@ public class FieldName {
     public static final String DOI = "doi";
     public static final String EDITION = "edition";
     public static final String EDITOR = "editor";
+    public static final String EID = "eid";
     public static final String EPRINT = "eprint";
     public static final String EPRINTCLASS = "eprintclass";
     public static final String EPRINTTYPE = "eprinttype";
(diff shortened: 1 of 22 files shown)


src/main/java/net/sf/jabref/model/entry/FieldName.java
```diff
@@ -42,16 +43,35 @@
...
     public static final String URLDATE = "urldate";
+    public static final String VERSION = "version";
     public static final String VOLUME = "volume";
     public static final String YEAR = "year";
+    public static final String YEARFILED = "yearfiled";
+
+    // JabRef internal field names
+    public static final String OWNER = "owner";
+    public static final String TIMESTAMP = "timestamp"; // Not the actual field name, but the default value
+    public static final String NUMBER_COL = "#";
```
matthiasgeiger: Is this a real "field" with a field name and used as such?

matthiasgeiger: Okay it is... https://github.com/JabRef/jabref/pull/1627/files#diff-59eb0b0c47b8c7b57710bde78ec9f349L215

matthiasgeiger:
LGTM!

src/main/java/net/sf/jabref/model/entry/FieldName.java
oscargus: I think so, yes.

From InternalBibtexFields:

``` java 
         dummy = new BibtexSingleField(FieldName.NUMBER_COL, false, 32);
         dummy.setPrivate();
         dummy.setWriteable(false);
         dummy.setDisplayable(false);
         add(dummy);
```

PR #1630, opened 2016-07-26
Export OO/LO citations to new database

oscargus:
Based on a forum discussion, http://discourse.jabref.org/t/export-a-bibtex-file-with-all-papers-cited-in-a-doc-via-ooplugin/87 . 40 minutes later.

### DIFF
--- a/src/main/java/net/sf/jabref/gui/openoffice/OOBibBase.java
+++ b/src/main/java/net/sf/jabref/gui/openoffice/OOBibBase.java
@@ -17,6 +17,7 @@ package net.sf.jabref.gui.openoffice;
 
 import java.io.File;
 import java.io.IOException;
+import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.net.MalformedURLException;
 import java.net.URL;
@@ -68,6 +69,7 @@ import com.sun.star.beans.UnknownPropertyException;
 import com.sun.star.beans.XPropertyContainer;
 import com.sun.star.beans.XPropertySet;
 import com.sun.star.comp.helper.Bootstrap;
+import com.sun.star.comp.helper.BootstrapException;
 import com.sun.star.container.NoSuchElementException;
 import com.sun.star.container.XEnumeration;
 import com.sun.star.container.XEnumerationAccess;
@@ -150,7 +152,9 @@ class OOBibBase {
     private static final Log LOGGER = LogFactory.getLog(OOBibBase.class);
 
 
-    public OOBibBase(String pathToOO, boolean atEnd) throws Exception {
+    public OOBibBase(String pathToOO, boolean atEnd) throws IOException, IllegalAccessException,
+            InvocationTargetException, BootstrapException, CreationException, UnknownPropertyException,
+            WrappedTargetException, IndexOutOfBoundsException, NoSuchElementException, NoDocumentException {
         authorYearTitleList.add(authComp);
         authorYearTitleList.add(yearComp);
         authorYearTitleList.add(titleComp);
@@ -224,7 +228,9 @@ class OOBibBase {
 
     }
 
-    private XDesktop simpleBootstrap(String pathToExecutable) throws Exception {
+    private XDesktop simpleBootstrap(String pathToExecutable)
+            throws IOException, IllegalAccessException, InvocationTargetException, BootstrapException,
+            CreationException {
 
         ClassLoader loader = ClassLoader.getSystemClassLoader();
         if (loader instanceof URLClassLoader) {
@@ -250,13 +256,17 @@ class OOBibBase {
         XMultiComponentFactory xServiceManager = xContext.getServiceManager();
         //Create the desktop, which is the root frame of the
         //hierarchy of frames that contain viewable components:
-        Object desktop = xServiceManager.createInstanceWithContext("com.sun.star.frame.Desktop", xContext);
+        Object desktop;
+        try {
+            desktop = xServiceManager.createInstanceWithContext("com.sun.star.frame.Desktop", xContext);
+        } catch (Exception e) {
+            throw new CreationException(e.getMessage());
+        }
         XDesktop xD = UnoRuntime.queryInterface(XDesktop.class, desktop);
 
         UnoRuntime.queryInterface(XComponentLoader.class, desktop);
 
         return xD;
-
     }
 
     private List<XTextDocument> getTextDocuments() throws NoSuchElementException, WrappedTargetException {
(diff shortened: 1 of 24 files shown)


src/main/java/net/sf/jabref/gui/openoffice/OOBibBase.java
```diff
@@ -1296,4 +1297,31 @@ public int hashCode() {
...
+                Optional<BibEntry> entry = loopDatabase.getEntryByKey(key);
+                if (entry.isPresent()) {
+                    BibEntry clonedEntry = (BibEntry) entry.get().clone();
+                    clonedEntry.setId(IdGenerator.next());
+                    resultDatabase.insertEntry(clonedEntry);
+                    entry.get().getFieldOptional(FieldName.CROSSREF).ifPresent(crossref -> {
+                        if (!resultDatabase.getEntryByKey(crossref).isPresent()) {
+                            Optional<BibEntry> refEntry = loopDatabase.getEntryByKey(crossref);
+                            if (refEntry.isPresent()) {
+                                resultDatabase.insertEntry(refEntry.get());
```
Siedlerchr: Here I would use ifPresent(resultDatabase.insertEntry)

oscargus:
OK, this is not flawless. Generating the database a second time in the same JabRef session leads to duplicate entries. Third time, tripled entries and so on...

src/main/java/net/sf/jabref/gui/openoffice/OOBibBase.java
oscargus: Makes sense. Copied the code from the AUX parser where more things were done in the corresponding place and didn't reflect on it.

src/main/java/net/sf/jabref/gui/openoffice/OOBibBase.java
```diff
@@ -1296,4 +1297,31 @@ public int hashCode() {
...
+        BibDatabase resultDatabase = new BibDatabase();
+        List<String> cited = findCitedKeys();
+        for (String key : cited) {
+            for (BibDatabase loopDatabase : databases) {
+                Optional<BibEntry> entry = loopDatabase.getEntryByKey(key);
+                if (entry.isPresent()) {
+                    BibEntry clonedEntry = (BibEntry) entry.get().clone();
+                    clonedEntry.setId(IdGenerator.next());
+                    resultDatabase.insertEntry(clonedEntry);
+                    entry.get().getFieldOptional(FieldName.CROSSREF).ifPresent(crossref -> {
```
Siedlerchr: 'Avoid checking every optional for isPresent, use a flatMap for chained optionals:
(Optional FlatMap example)
http://codingjunkie.net/working-with-java8-optionals/

oscargus: I had some plans to break the loop once the key was found. In that case one is stuck with `isPresent` as far as I know.

oscargus: The break solved the problem with the multiple entries. I was really confused for a while, but since the function looks through all tabs, it found the first generated database and included those entries once more and so on... :-)

Siedlerchr: Okay. This whole loop with all the presents looked very complex and confusing at the first moment. :confused:

src/main/java/net/sf/jabref/gui/openoffice/OOBibBase.java
```diff
@@ -1296,4 +1297,39 @@ public int hashCode() {
...
 
+
+    public BibDatabase generateDatabase(List<BibDatabase> databases)
+            throws NoSuchElementException, WrappedTargetException {
+        BibDatabase resultDatabase = new BibDatabase();
+        List<String> cited = findCitedKeys();
+
+        // For each cited key
+        for (String key : cited) {
+            // Loop through the avialables databases
```
oscargus: Bad spelling. Fixed it, but will not push just for that.

src/main/java/net/sf/jabref/gui/openoffice/OOBibBase.java
oscargus: I added some comments. Most of the time I like Optional (even though I'm at the best is using `ifPresent`), but the lack of `if ... else ...` and the fact that the lambda becomes its own function (which is very natural, but still prohibits writing some things, like return or break) is a bit annoying.

src/main/java/net/sf/jabref/gui/openoffice/OpenOfficePanel.java
```diff
@@ -322,6 +332,39 @@ public void actionPerformed(ActionEvent e) {
...
+                        Localization.lang("Unable to generate new database"), JOptionPane.ERROR_MESSAGE);
+            }
+
+            Defaults defaults = new Defaults(
+                    BibDatabaseMode.fromPreference(Globals.prefs.getBoolean(JabRefPreferences.BIBLATEX_DEFAULT_MODE)));
+
+            BibDatabaseContext databaseContext = new BibDatabaseContext(newDatabase, defaults);
+            this.frame.addTab(databaseContext, true);
+
+        } catch (Exception e) {
```
lenhard: Can you specify a more concrete type of exception here. Catching `Exception` itself is not really good practice as you will end up catching programming errors such as `NullPointerException` as well.

lenhard:
Apart from the things mentioned above, LGTM code-wise. But since I do not use open office, I have not tested the feature itself.

src/main/java/net/sf/jabref/gui/openoffice/OpenOfficePanel.java
oscargus: The problem is that `refreshCiteMarkers` throws `Exception` because `refreshCiteMarkersInternal` throws `Exception` because `insertReferenceMark` throws `Exception` because `com.sun.star.lang.XMultiServiceFactory.createInstance` throws `Exception`. 

I have earlier tried to specify the exceptions in OO/LO where possible, but there are still some methods that only throw `Exception`. It would be possible to borrow the code from the update action:

``` java
                } catch (UndefinedCharacterFormatException ex) {
                    reportUndefinedCharacterFormat(ex);
                } catch (UndefinedParagraphFormatException ex) {
                    reportUndefinedParagraphFormat(ex);
                } catch (ConnectionLostException ex) {
                    showConnectionLostErrorMessage();
                } catch (IOException ex) {
                    JOptionPane.showMessageDialog(frame,
                            Localization
                                    .lang("You must select either a valid style file, or use one of the default styles."),
                            Localization.lang("No valid style file defined"), JOptionPane.ERROR_MESSAGE);
                    LOGGER.warn("Problem with style file", ex);
                } catch (BibEntryNotFoundException ex) {
                    JOptionPane.showMessageDialog(frame,
                            Localization.lang(
                                    "Your OpenOffice/LibreOffice document references the BibTeX key '%0', which could not be found in your current database.",
                                    ex.getBibtexKey()),
                            Localization.lang("Unable to synchronize bibliography"), JOptionPane.ERROR_MESSAGE);
                    LOGGER.debug("BibEntry not found", ex);
                } catch (Exception ex) {
                    LOGGER.warn("Could not update bibliography", ex);
                }
```

which probably is slightly better.

The main "problem" is that there is very limited exception handling locally in the OO/LO part. Everything is propagated to the calling function. Which is good from some perspectives but makes it hard to change some things as you do not know what is actually happening and how it affect things. (Unless you really understand the code...)

lenhard: Another option would be to convert the `Exception` thrown by `com.sun.star.lang.XMultiServiceFactory.createInstance` into a custom `JabRefException` (the class does already exist) and check for that. This  preserves stack trace and message of the original exception and is as simple as:
```
catch(Exception e) {
  throw new JabRefException(e);
}
```
Then you can only deal with `JabRefException` from that point on and avoid accidentally catching another one. It also doesn't clutter the code with handling code. You can do the handling as you did before, just on a more limited type of exception. Would this be an option for the OO/LO part?

oscargus: Well, there are quite a few Exception types already defined in OO/LO so defining one more (in case we want JabRefException to be used later for something else) shouldn't be a problem. It may be a problem naming it since I'm not really sure what the errors are, maybe `ThingsWentWrongException`? ;-)

lenhard: I see, there are indeed quite some customn exceptions. But I don't understand why you want to define yet another exception? Why not just use `net.sf.jabref.JabRefException`

oscargus: I can of course use that, but I'm still tempted to try to figure out
something more clever... ;-)

oscargus: I introduced `CreationException`... One should probably "group" the exceptions slightly earlier and not just propagate them, but at least there are no `catch (Exception ex)` in the OO/LO part of the code.

lenhard: Ok, then this is good enough! I'll merge now.

PR #1634, opened 2016-07-27
Fixed a minor issue and refactored MergeEntries

oscargus:
In earlier versions, the column headings were wrong when merging two entries from the database after information had been merged from DOI. This fixes that (removed `static` for the heading strings).

Also refactored the code a bit and extracted `DiffHighlightingTextPane` which may be useful elsewhere (such as in the external change viewer).


simonharrer:
Add changelog entry for the bug fix, then merge this in.

oscargus:
Did that. I initially ignored it since it was a quite harmless bug that probably very few people would even notice.

matthiasgeiger:
Documenting improvements of JabRef does not harm :wink:

PR #1635, opened 2016-07-27
More tests

oscargus:
A few more tests of Cookie, StringUtil.intValueOfWithNull, and reenabling some tests for Layout.

### DIFF
--- a/src/test/java/net/sf/jabref/logic/layout/LayoutTest.java
+++ b/src/test/java/net/sf/jabref/logic/layout/LayoutTest.java
@@ -68,7 +68,6 @@ public class LayoutTest {
     }
 
     @Test
-    @Ignore
     public void testHTMLChar() throws IOException {
         String layoutText = layout("\\begin{author}\\format[HTMLChars]{\\author}\\end{author} ",
                 "@other{bla, author={This\nis\na\ntext}}");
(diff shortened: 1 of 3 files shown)


src/test/java/net/sf/jabref/logic/layout/LayoutTest.java
```diff
@@ -80,10 +78,10 @@ public void testHTMLChar() throws IOException {
 
         Assert.assertEquals("This is a text", layoutText);
 
-        layoutText = layout("\\begin{author}\\format[HTMLChars]{\\author}\\end{author} ",
+        /*  layoutText = layout("\\begin{author}\\format[HTMLChars]{\\author}\\end{author} ",
                 "@other{bla, author={This\nis\na\n\ntext}}");
 
-        Assert.assertEquals("This is a<br>text ", layoutText);
+        Assert.assertEquals("This is a<br>text ", layoutText); */
```
simonharrer: Why is this commented out? I would split this test up in multiple tests, and then mark the last as ignored so that we can work on that later.

lenhard:
LGTM code-wise. One remark though: We decided in the last devcall that we want to replace the custom cookie implementation of JabRef with some external library. We haven't decided which one to use yet, so some more tests do certainly not hurt.

PR #1636, opened 2016-07-28
Updated Errorprone to 2.0.11

oscargus:
Version 2.0.11 should have solved the issue making 2.0.10 unusable. As I did not face the problem, I have no idea if it actually is working though...


Siedlerchr:
I recently tested the version on Windows in a branch locally,  worked good. The only thing which failed was the test you already corrected.

PR #1637, opened 2016-07-28
[WIP] Special field editor for fields with entry keys

oscargus:
Related to https://github.com/koppor/jabref/issues/14 among others. 

capture12

There are some fundamental issues with drag and drop, such as the entry in the entry editor keeps changing when selecting other entries to drag. Also, I do not really get the table to update very nicely, even though the code is to a large extent a copy of the file link editor.

One can see this as a proof of concept and maybe something to build further upon. If it wasn't for the weird behaviour of the table, I would be more positive.

Ideally we should specify one or more fields that can be used to relate entries, such as "cites" and "similar" (or something).


oscargus:
Some comments: 
* There will be a more intelligent text in the right column. Now I just took the title. The idea is to use some layout formatter to get a nice string.
* It is possible to drag entries from the main table, but hard to know for sure which entry is in the editor when dropping.
* The key column can be edited and the right column updates appropriately.
* No support for entries changing keys (yet, this is already an issue for e.g. the crossref field and should be solver independently).
* The buttons for the crossref field are wrong. I just removed two to make it look like one cannot move entries up and down (since there can only be one entry there).
* It looks like it is possible to right-click on an entry an select it in the main table, but it doesn't work (yet).
* When moving entries up and down as well as adding them, there are some weird things going on. Maybe it is needed to synchronize absolutely every access to the underlying list?

oscargus:
@JabRef/developers : This can be tested at http://builds.jabref.org/entryfield/ now as it works quite well. Make sure that you show the `crossref`, `related`, and/or `entryset` field(s) in some tab. With this it would also make sense to start thinking about `citedby`/`references` fields as it will be easily implementable now.

oscargus:
I'm having some problems getting a bibtex key auto completer to work in the table. Has anyone worked with the auto completer and knows how it works?

koppor:
Can we first do something without autocompleter?

I worked with the auto completer ~5 years ago, but I have nearly forgotten anything. And it was a mess.

koppor:
Is the current state that we "only" have to test and then this is good to be reviewed?

oscargus:
Yes, I think that it is good to try out. Apart from the auto completer and drag-and-drop (which seems quite hard to get right since the entry editor is updated when selecting fields) this should be OK.

src/test/java/net/sf/jabref/logic/layout/format/AuthorsTest.java
```diff
@@ -129,4 +129,20 @@ public void testNoPeriod() {
         Assert.assertEquals("B C Bruce, C K von Manson and J Jumper",
                 a.format("Bruce, Bob Croydon and Charles Kermit von Manson and Jumper, Jolly"));
     }
+
+    @Test
```
koppor: This seems to come from another issue, but we can merge it nevertheless.

src/main/java/net/sf/jabref/gui/groups/TransferableEntrySelection.java
```diff
@@ -25,7 +25,7 @@
 
 import net.sf.jabref.model.entry.BibEntry;
 
-class TransferableEntrySelection implements Transferable {
+public class TransferableEntrySelection implements Transferable {
```
koppor: Seems to come from another issue, but is also OK for me.

koppor:
Do you have time to work on the functionality a bit?

- [ ] After clicking `+` and double clicking on `Unknown entry`, the editor of the bibtex key has to be focused
- [ ] When double clicking the second column, nothing happens. "Select" should be executed
- [ ] When pressing `+` and then the "go down" button, an exception is raised

```
Exception in thread "AWT-EventQueue-0" java.lang.IllegalArgumentException: Row index out of range
        at javax.swing.JTable.boundRow(JTable.java:2143)
        at javax.swing.JTable.setRowSelectionInterval(JTable.java:2166)
        at net.sf.jabref.gui.fieldeditors.EntryLinkListEditor.moveEntry(EntryLinkListEditor.java:343)
        at net.sf.jabref.gui.fieldeditors.EntryLinkListEditor.lambda$new$3(EntryLinkListEditor.java:124)
```

- [ ] I have to restart JabRef to get "Other fields" shown. Steps: New database, add entry with ` surelynotknown = {value}`, add another entry, switch back to first entry. No tab "Other fields shown". This might really be out of scope.

koppor:
Autocompeltion: @stefan-kolb might know something (because of #1136), but this might be too far away. @stefan-kolb We would "just" need to have an AutoCompleter for all existing bibtex keys. If this is easy possible, that would be fine.

tobiasdiez:
[BibtexKeyAutoCompleter](https://github.com/JabRef/jabref/blob/ee08252a495c8951206201829651a2c320893735/src/main/java/net/sf/jabref/logic/autocompleter/BibtexKeyAutoCompleter.java) ?

oscargus:
Yes, the auto completer is there. I just cannot get it to work in a table cell...

src/main/java/net/sf/jabref/gui/groups/TransferableEntrySelection.java
```diff
@@ -11,7 +11,7 @@
...
```
oscargus: This is a heritage from when I tried to implement drag and drop. currently it is not needed.

koppor: OK for me to move forward here. I wish the nasty UI issues of https://github.com/JabRef/jabref/pull/1637#issuecomment-246793569 can be solved.

oscargus:
I removed more unused code. Added double click (and a place holder for focusing on the bibtex key field, which I do not know how to do). I couldn't trigger the error you mentioned, but noted that the last edit is not always saved properly. For example, if one selects a new entry. I do not think this happened earlier, but not sure.

PR #1640, opened 2016-07-29
Unified some equals

oscargus:
I might regret this, but I tried to unify some of the equals considering the discussion in #1631. The style is not completely identical, but the ideas are there at least.

Looking at the `equals` in `ProxyPreferences` and `FieldChange` one can conclude that hopefully that code is automatically generated, although it might not be the best way to generate that code...

In some classes implementing `Formatter`, `equals` and `hashCode` are overridden by `defaultEquals` and `defaultHashCode` but in most not. Not really clear how this really should be.


lenhard:
No need to regret, I won't force you to format every equals implementation equally ;-)

No harm can come from the intial check you implemented for the equals methods (it's a performance optimization after all). 

I'll merge this right away.

PR #1641, opened 2016-07-29
Fixed #1639

### LINKED ISSUE #1639: Jabref websearch: Google Scholar shows no results on second pop-up menu
JabRef 3.5
windows 10 10.0 amd64 
Java 1.8.0_101

The Jabref websearch with Google Scholar does not show any results on the second pop-up menu
Steps to reproduce:

1. Set Web search in Jabref to "Google Scholar"
2. Copy "On the relationship between tooth shape and masticatory efficiency: a finite element study" into the Web search field
3. Fetch
4. One article is found
5. Tick the "Keep" box and click on "OK"
6. The second menu turns up: No entry to select is shown. The "Select all", "Deselect all", "Delete", "Generate now", "OK" and "Cancel" buttons are all shown as activable.

I guess this is the exact same problem as described in:
https://sourceforge.net/p/jabref/mailman/message/32485880/

I haven't had this problem with an earlier development version of Jabref 3.5 (JabRef-3.5dev--snapshot--2016-06-28--master--a24b544), but since I've upgraded I've encountered it now a couple of times. Can you confirm this issue?

oscargus:
Fixed #1639 as Google now moved the link generation to http://scholar.googleusercontent.com the full link is now provided.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - Fixed [#1592](https://github.com/JabRef/jabref/issues/1592): LibreOffice: wrong numbers in citation labels
 - The merge entry dialog showed wrong heading after merging from DOI
 - Fixed [#1321](https://github.com/JabRef/jabref/issues/1321): LaTeX commands in fields not displayed in the list of references
+- Fixed [#1639](https://github.com/JabRef/jabref/issues/1639): Google Scholar fetching works again.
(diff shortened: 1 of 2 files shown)


Siedlerchr:
Remove the old code. It can be restored at every point. That's why version control exists ;)

oscargus:
I removed it now. Although I rarely go back in the repository to see how things were solved earlier... (Btw, there are some more example of commented out code in that file, but I left it for the moment. Things like almost getting next page fetching to work.)

PR #1642, opened 2016-07-29
Moved the main part of XMPUtil to jabref.XMPUtilMain and injected a b…

oscargus:
Not sure about the future of the stand-alone tool, but from a dependency point of view it made sense to move it.

Note that the tests are not moved since there were some strange problems I couldn't really (find worthwhile at the moment to) solve, related to common test data for XMPUtil and XMPUtilMain.

### DIFF
--- a/src/main/java/net/sf/jabref/XMPUtilMain.java
+++ b/src/main/java/net/sf/jabref/XMPUtilMain.java
@@ -16,6 +16,7 @@ import net.sf.jabref.importer.fileformat.BibtexParser;
 import net.sf.jabref.logic.bibtex.BibEntryWriter;
 import net.sf.jabref.logic.bibtex.LatexFieldFormatter;
 import net.sf.jabref.logic.bibtex.LatexFieldFormatterPreferences;
+import net.sf.jabref.logic.xmp.XMPPreferences;
 import net.sf.jabref.logic.xmp.XMPUtil;
 import net.sf.jabref.model.database.BibDatabaseMode;
 import net.sf.jabref.model.entry.BibEntry;
@@ -57,6 +58,8 @@ public class XMPUtilMain {
             Globals.prefs = JabRefPreferences.getInstance();
         }
 
+        XMPPreferences xmpPreferences = XMPPreferences.fromPreferences(Globals.prefs);
+
         switch (args.length) {
         case 0:
             usage();
@@ -65,7 +68,7 @@ public class XMPUtilMain {
 
             if (args[0].endsWith(".pdf")) {
                 // Read from pdf and write as BibTex
-                List<BibEntry> l = XMPUtil.readXMP(new File(args[0]), Globals.prefs);
+                List<BibEntry> l = XMPUtil.readXMP(new File(args[0]), xmpPreferences);
 
                 BibEntryWriter bibtexEntryWriter = new BibEntryWriter(
                         new LatexFieldFormatter(LatexFieldFormatterPreferences.fromPreferences(Globals.prefs)), false);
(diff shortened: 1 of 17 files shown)


oscargus:
As a bonus, I cleaned up in the constructors of `TypedBibEntry` and only kept two public:
```
    public TypedBibEntry(BibEntry entry, BibDatabaseMode mode)
    public TypedBibEntry(BibEntry entry, BibDatabaseContext databaseContext)
```

skipping the latter one, this could reside in `model` but I'm sure there's some other good argument not to keep it there.

oscargus:
And extracted the preferences to XMPPreferences.

lenhard:
Honestly, I am not sure if this is worth the effort. The current XMP functionality builds upon an oudated library (apache pdfbox). We only keep to this, since newer versions of said library dropped the functionality we need. Hopefully, they re-introduce it and once they do, all our current XMP code will land in the bin.

I'll have a look at your changes nonetheless.

src/main/java/net/sf/jabref/XMPUtilMain.java
```diff
@@ -0,0 +1,174 @@
...
+import net.sf.jabref.logic.xmp.XMPPreferences;
+import net.sf.jabref.logic.xmp.XMPUtil;
+import net.sf.jabref.model.database.BibDatabaseMode;
+import net.sf.jabref.model.entry.BibEntry;
+import net.sf.jabref.preferences.JabRefPreferences;
+
+import org.apache.jempbox.impl.XMLUtil;
+import org.apache.jempbox.xmp.XMPMetadata;
+
+public class XMPUtilMain {
```
lenhard: This essentially is a cli class. Thus, it should reside in `jabref.cli` and not in the root package.

oscargus: Yes, but it is also (primarily) a stand-alone program. Hence, my (initial) choice of location. I would imagine that the command-line only JabRef version that is discussed would end up in the same location.

oscargus:
I have never used the XMP feature... ;-) But all 28 tests still works.

src/main/java/net/sf/jabref/XMPUtilMain.java
lenhard: Since most of the code is inside a main method, you are right. I would still put it in the cli package, since it otherwise clutters up the root package, though. What is the opinion of others? @JabRef/developers

Siedlerchr: Standalone --> cli

2016-07-29 18:13 GMT+02:00 Jörg Lenhard <notifications@github.com>:

obraliar:
`./gradlew check` is failing on my machine and on travis.
```
	at net.sf.jabref.logic.xmp.XMPSchemaBibtex.setBibtexEntry(XMPSchemaBibtex.java:291)
	at net.sf.jabref.logic.xmp.XMPSchemaBibtex.setBibtexEntry(XMPSchemaBibtex.java:279)
	at net.sf.jabref.logic.xmp.XMPSchemaBibtexTest.testSetBibtexEntry(XMPSchemaBibtexTest.java:250)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
...
```

stefan-kolb:
I fixed the test.

oscargus:
Thanks! For some reason the tests were not executed on Travis after the
final commit, but as I just moved one file I thought it shouldn't affect
anything.  (I have problems running anything from gradlew locally, but the
actual test passed...)

PR #1644, opened 2016-07-29
Resolve crossrefs and strings in main table

oscargus:
Based on a discussion in the forum, this is a quick hack to show the possibilities. Can be downloaded and tested from http://builds.jabref.org/resolvedfieldsinmaintable 

It would be especially interesting for people with large databases with many strings to test it for performance. A better implementation can be done if this will make it to the released version. At least removing the commented out lines.


ilippert:
Am not sure what counts as a large database. My database has currently 9883 entries, out of which 591 use crossref information. I do not perceive any slowdown in daily use.

ilippert:
Am not sure what counts as a large database. My largest bib file has >9800 entries, out of which 591 use the crossref field. In daily use I do not perceive performance issues compared to the days before the "hack".

oscargus:
Thanks @ilippert! I'd say that it counts as a large database here, especially considering the amount of crossref entries.

I've updated the code a bit and added a CHANGELOG entry describing the change.

oscargus:
OK, I thought it would make sense if one could see which fields are resolved (if nothing else to not confuse some users), so I added that as well. This is a different setting since I think there are different things to want to see required and optional fields and which field that are resolved. At the moment resolved field have precedence if both are enabled.

capture15

Good thing is that the check for optional/required/resolved is now only done if any of the flags are enabled, and, hence, the rendering should be slightly faster compared to before when not enabled.

src/main/resources/l10n/JabRef_tr.properties
```diff
@@ -1684,3 +1684,52 @@ You_must_enter_at_least_one_field_name=En_az_bir_alan_adı_girmelisiniz
...
+Style_file=
+
+Open_OpenOffice/LibreOffice_connection=
+
+You_must_enter_at_least_one_field_name=
+
+
+Non-ASCII_encoded_character_found=
+
+Background_color_for_resolved_fields=
```
koppor: I would just add the lines of the language strings you added. The other lines should come in in different ways (other PR, via master branch). Otherwise, we will have duplicate lines (refs #1719)

src/main/java/net/sf/jabref/model/database/BibDatabase.java
```diff
@@ -523,7 +523,7 @@ private String resolveContent(String result, Set<String> usedIds) {
      */
     public static Optional<String> getResolvedField(String field, BibEntry entry, BibDatabase database) {
         Objects.requireNonNull(entry, "entry cannot be null");
-        if ("bibtextype".equals(field)) {
+        if (BibEntry.TYPE_HEADER.equals(field) || "bibtextype".equals(field)) {
```
koppor: Why do we need two different strings for that? (`entryType` and `bibtextype`)?

When searching the code, only `BibEntry.TYPE_HEADER` should be sufficient. However, it might be the case that my search was too straight-forward.

grabbed_20160815-081924

src/main/java/net/sf/jabref/gui/maintable/MainTableColumn.java
```diff
@@ -144,4 +132,26 @@ public JLabel getHeaderLabel() {
             return new JLabel(getDisplayName());
         }
     }
+
+    public boolean isResolved(BibEntry entry) {
```
koppor: Please add a JavaDoc comment indicating the intention of this method. Currently, it seems if the **first** field of entry contains something, it returns true.

src/main/java/net/sf/jabref/gui/maintable/MainTableColumn.java
```diff
@@ -144,4 +132,26 @@ public JLabel getHeaderLabel() {
...
+        }
+
+        Optional<String> content = Optional.empty();
+        Optional<String> entryContent = Optional.empty();
+        for (String field : bibtexFields) {
+            if (BibEntry.TYPE_HEADER.equals(field) || "bibtextype".equals(field)) {
+                return false;
+            } else {
+                entryContent = entry.getFieldOptional(field);
+                content = BibDatabase.getResolvedField(field, entry, database.orElse(null));
```
koppor: When checking the code, it seems that this method either returns the content of the field or, if it can be resolved, the content of the resolved field. Correct me, if I'm wrong.

koppor:
Could you add an example bib file (maybe artificially generated) to show the effects of your change. I see your screenshot, but I want to try it for myself. -- Up to now, I did not use resolved fields at all as it seems that bibtex/biblatex resolves them differently and I don't know, how far JabRef follows this. Moreover, copy and paste does not work intuitively, but this is addressed at #1677.

src/main/resources/l10n/JabRef_tr.properties
oscargus: Yes, this is probably the reason of not having union merge. However, normally I check these things and unless it takes very long time to merge the PR it doesn't happen that much.

src/main/java/net/sf/jabref/gui/maintable/MainTableColumn.java
oscargus: Yes, that is how I understand it as well, where resolved means
cross-reference and/or strings.

src/main/java/net/sf/jabref/model/database/BibDatabase.java
koppor: Refs #1651

src/main/java/net/sf/jabref/gui/maintable/MainTableColumn.java
oscargus: Yes, that is correct. As the class is called MainRefColumn, however, it is
the first of the column fields (`journal/booktitle`).

src/main/java/net/sf/jabref/gui/maintable/MainTableColumn.java
koppor: Oh. OK. Maybe just rename `entryContent` -> `plainFieldContent` and `content` to `resolvedFieldContent`.

src/main/java/net/sf/jabref/gui/maintable/MainTableColumn.java
koppor: I'm trying to understand why one needs a flag if the entry is resolved at all. Don't I need to in a field-bases?

oscargus: Just to be able to mark if the content is resolved.

koppor: I think, that one needs the information on a per-field, not on a per-entry basis. As far as I understood, I get it for both cases: resolved fields are highlighted and there is an indication on the complete entry. - This is OK :innocent:

oscargus: Well, there is nothing saying if it is a crossref, string, or alias
resolving at the moment. Hence, the month field will mark gray with content
"Jan" independent of:
* the entry not having a month/date, but the crossref does
* it says #jan# in the field
* the entry has a date xxxx-01-xx

Indeed it may (initially) be a bit confusing for the user when the entry
editor says something different, that is why I wanted a marker. If tooltips
would work in the table one could (in the long run) add a tooltip saying
where the information origins. Also, the idea of having a "resolved"
grayed/italic default text in the entry editor for empty fields may be
worthwhile.

src/main/java/net/sf/jabref/model/database/BibDatabase.java
oscargus: I would guess that in the current master there will be different layout if you have entrytype and bintextype columns. We should probably add a migration that changes any bibtextype column to entrytype. This seems to have been neglected when changing the field name.

src/main/java/net/sf/jabref/gui/maintable/MainTableColumn.java
oscargus: I didn't really understand what I commented on. You suggested naming still makes sense, but the method returns false if the content of the field is the same as the resolved content of the field, i.e., no string, crossrefs, or alias fields are used.

src/main/java/net/sf/jabref/model/database/BibDatabase.java
oscargus: This has got a bit more clear now. At some stage the name of the entry type was changed from `bibtextype` to `entrytype`, but not completely, only in parts of the code, which is evident from the old code and all layout. However, when I tested to use `bibtextype` in a column nothing was shown, not clear why though. Anyway, the changes in #1651 are unified in this PR and maybe we should simply remove "bibtextype", but better wait for a while. No need for any migration as "bibtextype" didn't work in a column anyway and we do not want to modify peoples layout files...

oscargus:
Here is a bib-file to show a bit (must use extension .txt for GitHub). Make sure you show year, date, and month (this also shows that we should try to normalize the month display based on database type, I think). 
[resolvedatabase.txt](https://github.com/JabRef/jabref/files/418847/resolvedatabase.txt)

koppor:
I checked it with normal JabRef config. The "background color for resolved fields" doesn't seem to have any effect, but it should.

### normal operation

sp160817_081811

### resolved

resolved

oscargus:
You mean it should be enabled by default or that enabling it doesn't lead to any changes? There should be a light-gray background...

oscargus:
OK, I think I understand what you meant. Either way, I fixed a problem related to that now.

Which leads to another question: how should we deal with marking mandatory fields if there are, say, "author/editor" and exactly one of them is mandatory, both are mandatory, or it is mandatory to have one of them? Now, the column name is simply compared with the list of mandatory fields, so it only works for single field columns.

koppor:
The "Background color for resolved fields" seems to have no effect here:

grabbed_20160817-213158

Regarding "author/editor" -> Why not splitting the name using "/" and then checking each of them during the column name comparison? I know, we should use some cool data structures. Since we are aiming for JavaFX, this can be that dirty IMHO.

oscargus:
You need to tick "Color codes for resolved fields".

The splitting can be easily done. The question is just what the logic should be? Say that a column shows "author/doi", "author" is a mandatory field, but only the "doi" is present. Color?

CHANGELOG.md
```diff
@@ -28,6 +28,8 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - Added ISBN integrity checker
 - Added filter to not show selected integrity checks
 - The contents of `crossref` and `related` will be automatically updated if a linked entry changes key
+- The information shown in the main table now resolves crossrefs and strings
```
koppor: Remove this line - the next line contains the same information (and more)

oscargus: Not to mention the grammatical error in the next line... I would like to blame union merge. ;-)

src/main/resources/l10n/JabRef_zh.properties
```diff
@@ -1680,3 +1680,5 @@ You_must_enter_at_least_one_field_name=
 Non-ASCII_encoded_character_found=
 
 Toggle_web_search_interface=
+Background_color_for_resolved_fields=
+Color_codes_for_resolved_fields=
```
koppor: Shouldn't it be singular? "Color code for resolved fields"?

oscargus: You are probably right. I thought about it when I wrote it, think of consistency and "fields", but you are correct, there is only one color. I'll fix it tomorrow.

koppor:
Is it possible to enable "Color codes for resolved fields" as soon as one changes the color for resolved fields?

Color: If (at least) one mandatory field is missing: Color it red. Otherwise, normal color. In your example, the field would be red: Some content is there, but not the mandatory one.

oscargus:
Currently, only the number is colored red, so I do not think that is actually the answer. There may, for example, be another column with "editor" which is present and the entry is a book. I know it sounds complicated (and made up examples), but I thought about it and I do not really see what logic should be used for marking required and optional fields when there are multiple fields in one column.

Maybe the easiest way is to just enable the resolved entry marking by default? While it is possible, with the current code structure it will be quite hacky (since the buttons are initialized in one loop, I will have to single it out, etc, now, since it is the last one, at the moment, it is quite reasonable, but not really a nice solution).

oscargus:
For consistency one should in that case also do the same thing for the other color codes (required/optional) and grid lines. Maybe one can pass a JCheckButton to the ColorButton as an extra argument, passing null for all other cases?

oscargus:
I added automatic enabling for those three cases...

Dellu:
Here is one suggestion about the Crossref: 
It would be nice if the **BookTitle** could be displayed, rather than the **Title**

That, is when the parent entry has only BookTitle (not the title--due to the problems of Bibtex; we cannot use the Title unless we migrate to BibLatex). 

Here is a Proceeding with the Booktitle: [Collection](https://snag.gy/ySOs67.jpg)

Look at the following link: the Booktitle is not visible: 
[crossrefered](https://snag.gy/JEk4dC.jpg)

ilippert:
@Dellu 
it might be the case that you can achieve the desired effect by setting in the jabref preferences|Entry table columns| a field name to e.g. title/booktitle/chapter.
The effect is that if available the first, and if not available the second field name (and so on) content for the field is shown.

Dellu:
I am not talking about the column; I am talking about inside the entry table. Look at the second snapshot. I don't think the Preference has effect on what to display in the entry row.

ilippert:
sorry, yes, now I see more clearly. My fault for commenting too swiftly.

PR #1645, opened 2016-07-29
Added change listener for crossref field

oscargus:
Another WIP just to show what is coming (as I may be unavailable for a week or so now).

Change listener that updates the crossref field when the key changes. In the longer run this will be extended to all field with keys, see #1637.

Will continue with tests and CHANGELOG entries eventually.

### DIFF
--- a/src/main/java/net/sf/jabref/logic/autocompleter/AutoCompleterFactory.java
+++ b/src/main/java/net/sf/jabref/logic/autocompleter/AutoCompleterFactory.java
@@ -1,4 +1,4 @@
-/*  Copyright (C) 2003-2015 JabRef contributors.
+/*  Copyright (C) 2003-2016 JabRef contributors.
     This program is free software; you can redistribute it and/or modify
     it under the terms of the GNU General Public License as published by
     the Free Software Foundation; either version 2 of the License, or
@@ -15,7 +15,6 @@
 */
 package net.sf.jabref.logic.autocompleter;
 
-import java.util.Arrays;
 import java.util.Objects;
 
 import net.sf.jabref.logic.journals.JournalAbbreviationLoader;
@@ -44,9 +43,10 @@ public class AutoCompleterFactory {
 
         if (InternalBibtexFields.getFieldExtras(fieldName).contains(FieldProperties.PERSON_NAMES)) {
             return new NameFieldAutoCompleter(fieldName, preferences);
-        } else if (FieldName.CROSSREF.equals(fieldName)) {
+        } else if (InternalBibtexFields.getFieldExtras(fieldName).contains(FieldProperties.SINGLE_ENTRY_LINK)) {
             return new BibtexKeyAutoCompleter(preferences);
-        } else if (FieldName.JOURNAL.equals(fieldName) || FieldName.PUBLISHER.equals(fieldName)) {
+        } else if (InternalBibtexFields.getFieldExtras(fieldName).contains(FieldProperties.JOURNAL_NAME)
+                || FieldName.PUBLISHER.equals(fieldName)) {
             return new JournalAutoCompleter(fieldName, preferences, abbreviationLoader);
         } else {
             return new DefaultAutoCompleter(fieldName, preferences);
@@ -54,7 +54,7 @@ public class AutoCompleterFactory {
     }
 
     public AutoCompleter<String> getPersonAutoCompleter() {
-        return new NameFieldAutoCompleter(Arrays.asList(FieldName.AUTHOR, FieldName.EDITOR), true, preferences);
+        return new NameFieldAutoCompleter(InternalBibtexFields.getPersonNameFields(), true, preferences);
     }
 
 }
--- a/src/main/java/net/sf/jabref/model/database/KeyChangeListener.java
+++ b/src/main/java/net/sf/jabref/model/database/KeyChangeListener.java
@@ -1,0 +1,88 @@
+package net.sf.jabref.model.database;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+import net.sf.jabref.model.entry.BibEntry;
+import net.sf.jabref.model.entry.FieldProperties;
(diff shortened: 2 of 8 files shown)


oscargus:
Potential issue: when deleting an entry all references to it in `crossref`and `related` is removed, which is good. However, when undo:ing, that information is not restored. Similarly, undoing after removing a BibTeX key doesn't restore `crossref`and `related`. Undoing works for plainly changing the key, as this will trigger new events.

Not sure what the best solution is. Either not removing the keys in the fields, leading to an inconsistent database, or keeping it as it is, that the information is lost. I cannot really see how to actually undo it. It is of course possible to add undo information from the listener, but then it will be a separate undo compound compared to the remove, so one will need to undo twice...

Apart from that, this PR should be ready for review.

src/test/java/net/sf/jabref/model/database/KeyChangeListenerTest.java
```diff
@@ -0,0 +1,98 @@
...
+
+
+    @Before
+    public void setUp() throws IOException {
+        Globals.prefs = JabRefPreferences.getInstance();
+        try (FileInputStream stream = new FileInputStream("src/test/resources/testbib/keychangelistener.bib");
+                InputStreamReader fr = new InputStreamReader(stream, StandardCharsets.UTF_8)) {
+            ParserResult result = BibtexParser.parse(fr);
+
+            db = result.getDatabase();
```
tobiasdiez: Please create the database by hand instead of reading it from the file. It also has the advantage that you can save the reference to the entries in private fields.

oscargus: OK! I'll do that.

oscargus:
Anything else here?

src/main/java/net/sf/jabref/model/database/KeyChangeListener.java
```diff
@@ -0,0 +1,78 @@
...
+        String oldKey = event.getBibEntry().getCiteKey();
+        updateEntryLinks(null, oldKey);
+    }
+
+    private void updateEntryLinks(String newKey, String oldKey) {
+        for (BibEntry entry : database.getEntries()) {
+            for (String field : keyFields) {
+                entry.getFieldOptional(field).ifPresent(fieldContent -> {
+                    if (InternalBibtexFields.getFieldExtras(field).contains(FieldProperties.SINGLE_ENTRY_LINK)) {
+                        if (fieldContent.equals(oldKey)) {
```
tobiasdiez: Maybe move the contents of the if and else bodies to seperate methods to increase readability.

src/test/java/net/sf/jabref/model/database/KeyChangeListenerTest.java
```diff
@@ -0,0 +1,98 @@
...
+    private BibDatabase db;
+    private BibEntry entry1;
+    private BibEntry entry2;
+    private BibEntry entry3;
+    private BibEntry entry4;
+
+
+    @Before
+    public void setUp() throws IOException {
+        Globals.prefs = JabRefPreferences.getInstance();
```
tobiasdiez: prefs not needed here, right?

tobiasdiez:
Looks good to me.

src/test/java/net/sf/jabref/model/database/KeyChangeListenerTest.java
oscargus: Needed for InternalBibtexFields (unfortunately).

oscargus:
I changed so that each BibDatabase creates and adds a KeyChangeListener, so
it will always be present.

I must have downloaded some updates on my mobile broadband, so I have very
limited data left for the rest of the month. Hence, all updates will come
Thursday evening (when I'm back with WiFi).

oscargus:
The tests will pass again once #1697 is merged.

src/main/java/net/sf/jabref/logic/autocompleter/AutoCompleterFactory.java
```diff
@@ -44,17 +44,20 @@ public AutoCompleterFactory(AutoCompletePreferences preferences, JournalAbbrevia
...
+                || FieldName.PUBLISHER.equals(fieldName)) {
             return new JournalAutoCompleter(fieldName, preferences, abbreviationLoader);
         } else {
             return new DefaultAutoCompleter(fieldName, preferences);
         }
     }
 
     public AutoCompleter<String> getPersonAutoCompleter() {
-        return new NameFieldAutoCompleter(Arrays.asList(FieldName.AUTHOR, FieldName.EDITOR), true, preferences);
+        return new NameFieldAutoCompleter(InternalBibtexFields.getAllPublicFieldNames().stream().filter(
```
tobiasdiez: I propose to move some of the logic to InternalBibtexFields so that you can write `InternalBibtexFields.getPersonFieldNames()`.

oscargus: I was also considering that. Maybe even introduce a constant list as there are for some other properties. Advantage: no need for two long lines of code. Disadvantages: not user configurable (if we ever want that, easy to change though) and, if we add a method, there may be quite a few special methods. Which on the other hand may be better than quite a few two line statements...

I might go for the constant list as there are probably more fields that should count as journal names and then we might include them in one step as well.

oscargus: I introduced it in #1720, so easiest thing is to merge that first and then rebase this one.

Dellu:
This is dangerious. 
I removed  a large number of book entries to only update with a new version of them exported from Bookends. All the _incollection_ entries crossrefed with them have lost their crossref field. 
I thought the **update** would not delete the fields: only update when there is change on the key. That is what BibDesk does. It never delets the keys. 

Unless you can implement it like the BibDesk, I think it is better to put this feature into the Cleanup tools. The user can call for it only when she wants to update changes. It is dangerous otherwise. I was shocked when I see all of them have lost their link. Thanks for Timemachine.

PR #1648, opened 2016-07-30
Refactoring of importer

oscargus:
Moved classes to gui and logic and created a bit more structure.

(Eclipse just hang and I have other things to do at the moment, but there are mainly some code cleanups:
* getting rid of a GUI-dependency in ArXiv by moving correctLinebreaks
* import grouping in a few files
)


src/main/java/net/sf/jabref/cli/ArgumentProcessor.java
```diff
@@ -19,12 +19,9 @@
 import net.sf.jabref.JabRefGUI;
 import net.sf.jabref.MetaData;
 import net.sf.jabref.external.AutoSetLinks;
-import net.sf.jabref.importer.ImportFormatReader;
-import net.sf.jabref.importer.ImportInspectionCommandLine;
-import net.sf.jabref.importer.OpenDatabaseAction;
-import net.sf.jabref.importer.ParserResult;
-import net.sf.jabref.importer.fetcher.EntryFetcher;
-import net.sf.jabref.importer.fetcher.EntryFetchers;
+import net.sf.jabref.gui.importer.actions.OpenDatabaseAction;
```
oscargus: Hmm, here are some more things to sort out...

src/main/java/net/sf/jabref/cli/ImportInspectionCommandLine.java
```diff
@@ -13,14 +13,16 @@
...
-package net.sf.jabref.importer;
+package net.sf.jabref.cli;
 
 import java.util.Collection;
 import java.util.Collections;
 import java.util.LinkedList;
 import java.util.List;
 
-import net.sf.jabref.importer.fetcher.EntryFetcher;
+import net.sf.jabref.gui.importer.fetcher.EntryFetcher;
```
oscargus: And here. Getting rid of EntryFetcher would be good...

src/main/java/net/sf/jabref/logic/importer/ImportFormatReader.java
```diff
@@ -24,28 +24,30 @@
...
-import net.sf.jabref.importer.fileformat.MedlineImporter;
-import net.sf.jabref.importer.fileformat.MedlinePlainImporter;
-import net.sf.jabref.importer.fileformat.MsBibImporter;
-import net.sf.jabref.importer.fileformat.OvidImporter;
-import net.sf.jabref.importer.fileformat.PdfContentImporter;
-import net.sf.jabref.importer.fileformat.PdfXmpImporter;
-import net.sf.jabref.importer.fileformat.RepecNepImporter;
-import net.sf.jabref.importer.fileformat.RisImporter;
-import net.sf.jabref.importer.fileformat.SilverPlatterImporter;
+import net.sf.jabref.gui.undo.NamedCompound;
```
oscargus: This is also a bit annoying...

oscargus:
OK. If only the build passes, I think I'm done here. 

Issues remaining (some which are not easily fixed, others are just too time consuming this weekend):
* dependency `cli` -> `gui` (because of `EntryFetcher`)
* some dependencies `logic.importer` -> `Globals` (fixed by #1642 and with new `LabelPatternUtil.checkLegalKey` default)
* `DOItoBibTeX` is a quite ugly class. Might just as well be a proper class implementing the new Fetcher interfaces and not just static methods. On the other hand, right now there are less issues with preference changes.
* The static methods in `logic.util.io.AutoSaveUtil` can maybe find a better home.

Let me just say that if you find this OK, please merge it. I'm not looking forward to rebasing to many times... ;-)

src/main/java/net/sf/jabref/cli/SystemOutputPrinter.java
```diff
@@ -0,0 +1,21 @@
...
+
+public class SystemOutputPrinter implements OutputPrinter {
+
+    @Override
+    public void setStatus(String s) {
+        System.out.println(s);
+    }
+
+    @Override
+    public void showMessage(Object message, String title, int msgType) {
```
Siedlerchr: Can this be a more explicit type than Object?

oscargus: The implementation in FetcherPreviewDialog shows the Object as:

``` java
    @Override
    public void showMessage(Object message, String title, int msgType) {
        JOptionPane.showMessageDialog(this, message, title, msgType);
    }
```
so I guess that is the reason for having an Object.

oscargus:
There are two "new" tests as well: ParsedBibEntryTests.java and BibDatabaseTestsWithFiles.java containing tests which used to reside in model, but since they also depended on importer, the architecture test now complains. Since they are supposed to test BibDatabase and BibEntry, it should be OK to hard code the entries rather than parsing them.

src/main/java/net/sf/jabref/cli/SystemOutputPrinter.java
Siedlerchr: It can be safely replaced with String overall. While the showMessageDialog in JOption accepts an object as message param, the implementations in JabRef are always called with Strings. I did a quick check.

oscargus: OK, I've changed it.

src/test/java/net/sf/jabref/logic/importer/BibDatabaseTestsWithFiles.java
```diff
@@ -0,0 +1,46 @@
...
+    @Before
+    public void setUp() {
+        Globals.prefs = JabRefPreferences.getInstance(); // set preferences for this test
+
+        importFormatPreferences = ImportFormatPreferences.fromPreferences(Globals.prefs);
+    }
+
+    @Test
+    public void resolveStrings() throws IOException {
+        try (FileInputStream stream = new FileInputStream("src/test/resources/net/sf/jabref/util/twente.bib");
```
tobiasdiez: I would prefer if you just create the database with the strings by hand (and split the test into 4)

oscargus: For sure. At the moment I was happy with moving the tests though. ;-)

(I've summarized this and some other things that can/should be done, but considering the effort in just refactoring and moving things to make the architecture tests pass, I would prefer merging this before there are lots of other, smaller, PRs, merged.)

oscargus:
@JabRef/developers anything else at the moment here? There are some things remaining, but it would be nice to merge this and not risk complicated rebasing. In total there's been a significant step towards the wanted architecture (and, if nothing else, any remaining issues is made clearer with this PR).

src/test/java/net/sf/jabref/logic/importer/fileformat/CustomImporterTest.java
```diff
@@ -107,14 +109,17 @@ public void testToString() {
 
     @Test
     public void testClassicConstructor() {
-        CustomImporter customImporter = new CustomImporter("Copac", "cpc", "net.sf.jabref.importer.fileformat.CopacImporter", "src/main/java/net/sf/jabref/importer/fileformat/CopacImporter.java");
+        CustomImporter customImporter = new CustomImporter("Copac", "cpc",
```
Siedlerchr: Remove empty linebreak

src/test/java/net/sf/jabref/logic/importer/fileformat/PdfContentImporterTestFiles.java
```diff
@@ -49,7 +44,8 @@ public static void setUp() {
     public void correctContent() throws IOException, URISyntaxException {
         String pdfFileName = fileName + ".pdf";
         String bibFileName = fileName + ".bib";
-        PdfContentImporter importer = new PdfContentImporter();
+        PdfContentImporter importer = new PdfContentImporter(
+                ImportFormatPreferences.fromPreferences(JabRefPreferences.getInstance()));
         Path pdfFile = Paths.get(PdfContentImporter.class.getResource(pdfFileName).toURI());
         List<BibEntry> result = importer.importDatabase(pdfFile, Charset.defaultCharset()).getDatabase().getEntries();
```
Siedlerchr: Please use explicit charset of the file/database

src/test/java/net/sf/jabref/logic/importer/util/JSONEntryParserTest.java
```diff
@@ -39,7 +39,7 @@ public void testBibJSONConverter() {
                         + "\"id\": \"1563-5171\"},{\"type\": \"doi\",\"id\": \"10.1155/2014/217495\""
                 + "}],\"created_date\":\"2014-05-09T19:38:31Z\"}\"";
         JSONObject jo = new JSONObject(jsonString);
```
Siedlerchr: Better variable names here, please. "jo" reminds me of the German slang word for ok;)

Siedlerchr:
LGTM :+1:  Please have a look at the Codacy Issues.

oscargus:
OK, I fixed the comments (and note that only one of the comments and Codacy issues were generated by this PR ;-)).

src/main/java/net/sf/jabref/gui/importer/fetcher/ISBNtoBibTeXFetcher.java
```diff
@@ -101,7 +102,8 @@ public boolean processQuery(String query, ImportInspector inspector, OutputPrint
                 bibtexString = scan.useDelimiter("\\A").next();
             }
 
-            BibEntry entry = BibtexParser.singleFromString(bibtexString);
```
tobiasdiez: I'm not so sure that it is always a good idea to use `ImportFormatPreferences.fromPreferences` for the fetchers. For example the encoding should be specified by the fetcher and not by the user preferences.

src/main/java/net/sf/jabref/logic/importer/ImportFormatPreferences.java
```diff
@@ -0,0 +1,81 @@
...
+
+    public boolean isConvertUnitsOnSearch() {
+        return convertUnitsOnSearch;
+    }
+
+    public boolean isUseCaseKeeperOnSearch() {
+        return useCaseKeeperOnSearch;
+    }
+
+    public ImportFormatPreferences setEncoding(Charset newEncoding) {
```
tobiasdiez: rename to withEncoding

src/main/java/net/sf/jabref/logic/importer/ImportFormatPreferences.java
```diff
@@ -0,0 +1,81 @@
...
+import net.sf.jabref.logic.bibtex.FieldContentParserPreferences;
+import net.sf.jabref.logic.importer.fileformat.CustomImporter;
+import net.sf.jabref.logic.labelpattern.LabelPatternPreferences;
+import net.sf.jabref.preferences.JabRefPreferences;
+
+public class ImportFormatPreferences {
+
+    private final Set<CustomImporter> customImportList;
+
+    private Charset encoding;
```
tobiasdiez: I'm fine with this (and the "setter" below) but @simonharrer usually immutable preferences classes.

src/main/java/net/sf/jabref/logic/importer/fileformat/BibtexParser.java
```diff
@@ -74,16 +73,16 @@
...
-            FieldContentParserPreferences.fromPreferences(Globals.prefs));
+    private final FieldContentParser fieldContentParser;
     private ParserResult parserResult;
     private static final Integer LOOKAHEAD = 64;
     private final Deque<Character> pureTextFromFile = new LinkedList<>();
+    private final ImportFormatPreferences importFormatPreferences;
 
-
-    public BibtexParser(Reader in) {
+    public BibtexParser(Reader in, ImportFormatPreferences importFormatPreferences) {
```
tobiasdiez: I think the whole `ImportFormatPreferences` contains to much information for the parser. In the end only the following information is needed:
- keyword separator
- list of wrapped fields
Thus I would prefer if only these two are passed as parameters.
(One can argue that both data should actually be stored in the bib-metadata. )

src/main/java/net/sf/jabref/gui/importer/fetcher/ISBNtoBibTeXFetcher.java
oscargus: When that happens, it is easy to remove it from the  `ImportFormatPreferences`. At the moment it is the most feasible way to get rid of `Globals` dependencies.

src/main/java/net/sf/jabref/logic/importer/fetcher/ArXiv.java
```diff
@@ -261,7 +261,8 @@ public ArXivEntry(Node item) {
             publishedDate = XMLUtil.getNodeContent(item, "published");
 
             // Abstract of the article
-            abstractText = XMLUtil.getNodeContent(item, "summary").map(OAI2Fetcher::correctLineBreaks).map(
+            abstractText = XMLUtil.getNodeContent(item, "summary").map(OAI2Handler::correctLineBreaks)
+                    .map(
```
oscargus: I'll reformat and see. :-)

src/main/java/net/sf/jabref/logic/importer/ImportFormatPreferences.java
oscargus: OK! I think that the setEncoding was "required" or I wouldn't have introduced it.

oscargus: I checked and it is only used in `OpenDatabaseTest` for convenience. Is it worth copying and returning a new `ImportFormatPreferences`?

src/main/java/net/sf/jabref/logic/importer/ImportFormatPreferences.java
oscargus: Done.

src/main/java/net/sf/jabref/logic/importer/ImportFormatPreferences.java
oscargus: I returned a new instance now.

src/main/java/net/sf/jabref/logic/importer/fileformat/BibtexParser.java
oscargus: This is a granularity issue. As the BibtexParser is an import format it got that now.

With `FieldContentParserPreferences` it is the opposite. Yes, it can be accessed directly from `ImportFormatPreferences`, but, for historical reasons, a finer granularity was chosen there. This change would then have to be done in `LatexFieldFormatterPreferences` as well (which is not required here), so somehow it makes sense to have a separate `FieldContentParserPreferences` since it is accessed in two completely different ways.

oscargus:
@JabRef/developers I merge this now. Nothing is really changed (except for moving things around), and there has at least been some reviewing.

PR #1649, opened 2016-07-30
Fixed one more NPE which should have been fixed in #1631

oscargus:
Apparently I missed one place in #1631... Same CHANGELOG though.


Siedlerchr:
LGTM :+1: , However, will be sooner or later replaced by the work in #1336. So I would vote for merging in

PR #1650, opened 2016-07-30
Replaced output of getResolvedField to Optional<String>

oscargus:
A bit weird that there are no explicit tests at all for `getResolvedField`...

### DIFF
--- a/src/main/java/net/sf/jabref/logic/layout/LayoutEntry.java
+++ b/src/main/java/net/sf/jabref/logic/layout/LayoutEntry.java
@@ -207,11 +207,8 @@ class LayoutEntry {
         case LayoutHelper.IS_LAYOUT_TEXT:
             return text;
         case LayoutHelper.IS_SIMPLE_FIELD:
-            String value = BibDatabase.getResolvedField(text, bibtex, database);
+            String value = BibDatabase.getResolvedField(text, bibtex, database).orElse("");
 
-            if (value == null) {
-                value = "";
-            }
             // If a post formatter has been set, call it:
             if (postFormatter != null) {
                 value = postFormatter.format(value);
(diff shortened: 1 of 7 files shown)


src/main/java/net/sf/jabref/logic/layout/LayoutEntry.java
```diff
@@ -268,38 +261,37 @@ private String handleOptionField(BibEntry bibtex, BibDatabase database) {
     }
 
     private String handleFieldOrGroupStart(BibEntry bibtex, BibDatabase database, Optional<Pattern> highlightPattern) {
-        String field;
+        Optional<String> field;
         if (type == LayoutHelper.IS_GROUP_START) {
             field = BibDatabase.getResolvedField(text, bibtex, database);
         } else if (text.matches(".*(;|(\\&+)).*")) {
```
Siedlerchr: A great improvement would be to compile the regexes to pattern and then use simply
split()  on that pattern object , otherwise java creates each regex pattern everytime again.
(I did not know that until I just found it in javadoc  :sunglasses: )
https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html#split-java.lang.CharSequence-

oscargus: Sure. Create an issue in koppor/JabRef. ;-)

oscargus: 28 hits of `matches("`

simonharrer:
LGTM

src/main/java/net/sf/jabref/logic/layout/LayoutEntry.java
simonharrer: Maybe that is time for another PR.

PR #1651, opened 2016-07-30
Changed name of entry type in layout files

oscargus:
For consistency with the recent versions of the code. Still backwards compatible though.


oscargus:
I didn't see any information about this at the help pages. Maybe it is worthwhile pointing out the field name to use?

Siedlerchr:
Not sure, but the sample code on writing CustomImporters should be then corrected, too?
https://help.jabref.org/en/CustomImports

oscargus:
Not because of this PR, but clearly because of the importer refactoring.
Hadn't thought about that...

It has been `entrytype` for quite some time internally, so may make sense
to use the same. However, it struck me that there could be other formats
using `bibtextype` such as BibXML or whatever it is called. Not that it is
a problem but maybe it is good to be consistent...

stefan-kolb:
An automatic migration for this would be good?!

oscargus:
I'm not sure if we want to change peoples layout files... This is sort of similar as when FormatChars was replaced. At the moment there is a still support for the old name.

It would be nice to make people aware of it one way or the other though...

oscargus:
I should change the default previews as well.

src/main/java/net/sf/jabref/gui/preftabs/FileTab.java
```diff
@@ -285,10 +285,6 @@ public void storeSettings() {
             prefs.put(JabRefPreferences.NON_WRAPPABLE_FIELDS, nonWrappableFields.getText());
             updateSpecialFields = true;
         }
-        // If either of the two last entries were changed, run the update for special field handling:
```
koppor: Why was this code removed? OK, I don't even get the deleted comment, but maybe the code was useful.

oscargus: See `updateSpecialFieldHandling()` :-) 

It updates `private final Set<String> nonWrappableFields = new HashSet<>(5)` which is never used somewhere else...

oscargus:
I updated the code to print a warning if `bibtextype` is used in a layout, with an encouragement to change it.

src/main/java/net/sf/jabref/logic/xmp/XMPUtil.java
```diff
@@ -832,7 +832,7 @@ private static void writeToDCSchema(XMPSchemaDublinCore dcSchema,
              *
              * Description: Relationships to other documents.
              *
-             * Bibtex-Fields used: bibtextype
+             * Bibtex-Fields used: entrytype
              */
```
koppor: Can the comments be merged?

koppor:
Please adress my minor comment and merge.

I would have liked to see test cases for the layouts, but I think, this is too complicated for now.

Minor thing: We should render the article type in a similar way as in the `bib` file: `InProceedings` vs. `Inproceedings`:

grabbed_20160815-091644

grabbed_20160815-091659

src/main/java/net/sf/jabref/logic/xmp/XMPUtil.java
oscargus: Probably. But to be honest I didn't understand the code more than I'm quite sure that I should change the comment.

oscargus:
The rendering in line 113 of MainTableColumn, but I have no idea where the camel cased entry types reside (should be easy to find though).

oscargus:
I fixed rendering in the main table and added a formatter, `EntryTypeFormatter` for camel casing entry type names.

src/main/java/net/sf/jabref/logic/xmp/XMPUtil.java
oscargus: That comment was quite misplaced. I correct the actual type comment.

src/main/java/net/sf/jabref/logic/layout/format/EntryTypeFormatter.java
```diff
@@ -0,0 +1,22 @@
...
+public class EntryTypeFormatter implements LayoutFormatter {
+
+    @Override
+    public String format(String fieldText) {
+        // Fancy typesetting
+        BibEntry entry = new BibEntry();
+        entry.setType(fieldText);
+        TypedBibEntry typedEntry = new TypedBibEntry(entry, BibDatabaseMode.BIBLATEX);
+        return typedEntry.getTypeForDisplay();
+
```
koppor: no empty line :innocent:

src/main/java/net/sf/jabref/logic/layout/LayoutEntry.java
```diff
@@ -235,9 +236,15 @@ public String doLayout(BibEntry bibtex, BibDatabase database, Optional<Pattern>
...
-        if ("bibtextype".equals(text)) {
+        if (BibEntry.TYPE_HEADER.equals(text)) {
+            fieldEntry = bibtex.getType();
+        } else if (BibEntry.OBSOLETE_TYPE_HEADER.equals(text)) {
+            LOGGER.warn("'" + BibEntry.OBSOLETE_TYPE_HEADER
+                    + "' is an obsolete name for the entry type. Please update your layout to use '"
+                    + BibEntry.TYPE_HEADER + "' instead.");
             fieldEntry = bibtex.getType();
         } else {
+
```
koppor: Why the empty line?

src/main/java/net/sf/jabref/logic/layout/format/EntryTypeFormatter.java
```diff
@@ -0,0 +1,22 @@
...
+import net.sf.jabref.logic.layout.LayoutFormatter;
+import net.sf.jabref.model.database.BibDatabaseMode;
+import net.sf.jabref.model.entry.BibEntry;
+
+
+public class EntryTypeFormatter implements LayoutFormatter {
+
+    @Override
+    public String format(String fieldText) {
+        // Fancy typesetting
```
koppor: Could you remove that line and add a class comment with an example (above line 9)?

src/main/java/net/sf/jabref/logic/layout/format/EntryTypeFormatter.java
```diff
@@ -0,0 +1,22 @@
...
+import net.sf.jabref.logic.TypedBibEntry;
+import net.sf.jabref.logic.layout.LayoutFormatter;
+import net.sf.jabref.model.database.BibDatabaseMode;
+import net.sf.jabref.model.entry.BibEntry;
+
+
+public class EntryTypeFormatter implements LayoutFormatter {
+
+    @Override
+    public String format(String fieldText) {
```
koppor: Could you state that here the entryType as string is expected? Maybe rename the parameter form `fieldText` to `entryType`?

src/test/java/net/sf/jabref/logic/layout/format/EntryTypeFormatterTest.java
```diff
@@ -0,0 +1,33 @@
+package net.sf.jabref.logic.layout.format;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+public class EntryTypeFormatterTest {
```
koppor: :+1:

src/main/java/net/sf/jabref/logic/layout/format/EntryTypeFormatter.java
oscargus: Ahh, missed your comment. Will do that in next PR.

src/main/java/net/sf/jabref/logic/layout/format/EntryTypeFormatter.java
oscargus: Same here.

src/main/java/net/sf/jabref/logic/layout/LayoutEntry.java
oscargus: Hmm, no idea...

PR #1652, opened 2016-07-31
Made possible to make the OO/LO panel a bit more narrow

oscargus:
By using a shorter text for the button introduced in #1630 , it is possible to make it more narrow. Now, "OpenOffice/LibreOffice" is limiting the width.

Some small refactorings.

### DIFF
--- a/src/main/java/net/sf/jabref/gui/openoffice/OOBibBase.java
+++ b/src/main/java/net/sf/jabref/gui/openoffice/OOBibBase.java
@@ -192,17 +192,17 @@ class OOBibBase {
 
     public void selectDocument() throws UnknownPropertyException, WrappedTargetException, IndexOutOfBoundsException,
             NoSuchElementException, NoDocumentException {
-        List<XTextDocument> ls = getTextDocuments();
+        List<XTextDocument> textDocumentList = getTextDocuments();
         XTextDocument selected;
-        if (ls.isEmpty()) {
+        if (textDocumentList.isEmpty()) {
             // No text documents found.
             throw new NoDocumentException("No Writer documents found");
-        } else if (ls.size() == 1) {
+        } else if (textDocumentList.size() == 1) {
             // Get the only one
-            selected = ls.get(0);
+            selected = textDocumentList.get(0);
         } else {
             // Bring up a dialog
-            selected = selectComponent(ls);
+            selected = selectComponent(textDocumentList);
         }
 
         if (selected == null) {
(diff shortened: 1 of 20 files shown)


src/main/java/net/sf/jabref/gui/openoffice/OOBibBase.java
```diff
@@ -1188,9 +1185,9 @@ private void italicizeOrBold(XTextCursor position, boolean italicize, int start,
     }
 
     private void removeReferenceMark(String name) throws NoSuchElementException, WrappedTargetException {
-        XReferenceMarksSupplier xSupplier = UnoRuntime.queryInterface(XReferenceMarksSupplier.class, xCurrentComponent);
-        if (xSupplier.getReferenceMarks().hasByName(name)) {
-            Object o = xSupplier.getReferenceMarks().getByName(name);
+        XNameAccess nameAccess = getReferenceMarks();
+        if (nameAccess.hasByName(name)) {
+            Object o = nameAccess.getByName(name);
```
Siedlerchr: Possible to have a more speaking name for Object o?

Siedlerchr: A more speaking name for the nameAcces, maybe xNamedRefMarks, or similar as in the getBookmarkRange method, would be nice, too.

Siedlerchr:
Just some minor remarks, rest looks good :+1:

src/main/java/net/sf/jabref/gui/openoffice/OOBibBase.java
oscargus: For sure it is possible. I just left it as is (and have no real idea what it does since I didn't consider that part of the code). While it looks like added code, it is just a consequence of the diff algorithm not really working.

Siedlerchr: Ah okay. yeah, it looked like you changed it. If it is too much effort then leave it as is.

oscargus: Maybe not too much effort, but there are many of these small details here
and there that are quite tedious to fix when you touch the code for a
completely different reason...

oscargus: There are 111 hits for `Object o). Some are probably OK as they are in equals, but it may be better to make a major overhaul for that issue.

Siedlerchr: Okay, then I would suggest making this in a new PR.

oscargus: I updated most variable names that were a bit short in OOBibBase.

PR #1654, opened 2016-08-01
ISBN Fetcher using the new fetcher infrastructure

zellerdev:
Implement an ISBN fetcher using the new fetcher infrastructure #1594 


src/test/java/net/sf/jabref/logic/importer/fetcher/IsbnFetcherTest.java
```diff
@@ -0,0 +1,68 @@
...
+    public void testHelpPage() {
+        assertEquals("ISBNtoBibTeXHelp", fetcher.getHelpPage().getPageName());
+    }
+
+    @Ignore("Create reproducible offline test")
+    @Test
+    public void testFetcher10() throws FetcherException, IOException {
+        Optional<BibEntry> isbn = fetcher.performSearchById("0321356683");
+        Assert.assertNotNull(isbn);
+        try (InputStream bibStream = IsbnFetcherTest.class.getResourceAsStream("IsbnFetcherTest.bib")) {
```
tobiasdiez: Please create the BibEntry by hand instead of using the BIB file (see ArxivTests)

src/main/java/net/sf/jabref/gui/mergeentries/FetchAndMergeEntry.java
```diff
@@ -55,7 +55,13 @@ public FetchAndMergeEntry(BibEntry entry, BasePanel panel, List<String> fields)
                 if (FieldName.DOI.equals(field)) {
                     fetchedEntry = new DOItoBibTeXFetcher().getEntryFromDOI(fieldContent.get());
                 } else if (FieldName.ISBN.equals(field)) {
-                    fetchedEntry = new ISBNtoBibTeXFetcher().getEntryFromISBN(fieldContent.get(), null);
+
+                    try {
+                        fetchedEntry = new IsbnFetcher().performSearchById(fieldContent.get());
+                    } catch (FetcherException fe) {
+                        fe.printStackTrace();
```
tobiasdiez: see how exceptions are handled below for the ArXiv fetcher

src/main/java/net/sf/jabref/importer/fetcher/GeneralFetcher.java
```diff
@@ -320,4 +295,4 @@ public int compare(EntryFetcher entryFetcher, EntryFetcher entryFetcher1) {
             return entryFetcher.getTitle().compareTo(entryFetcher1.getTitle());
         }
     }
-}
```
tobiasdiez: Keep new lines at end of file (probably a editor configuration)

src/main/java/net/sf/jabref/importer/fetcher/IdBasedEntryFetcher.java
```diff
@@ -0,0 +1,81 @@
...
+        this.fetcher = Objects.requireNonNull(fetcher);
+    }
+
+    @Override
+    public boolean processQuery(String query, ImportInspector inspector, OutputPrinter status) {
+
+        status.setStatus(Localization.lang("Processing %0", query));
+        try {
+            Optional<BibEntry> match = fetcher.performSearchById(query);
+            inspector.addEntry(match.get());
```
tobiasdiez: match.ifPresent(inspector::addEntry) otherwise a NPE is thrown

src/main/java/net/sf/jabref/logic/importer/fetcher/IsbnFetcher.java
```diff
@@ -0,0 +1,101 @@
...
+                    bibtexString = scan.useDelimiter("\\A").next();
+                }
+
+                entry = BibtexParser.singleFromString(bibtexString);
+                if (entry != null) {
+                    // Optionally add curly brackets around key words to keep the case
+                    String title = entry.getField("title");
+                    if (title != null) {
+                        // Unit formatting
+                        if (Globals.prefs.getBoolean(JabRefPreferences.USE_UNIT_FORMATTER_ON_SEARCH)) {
```
tobiasdiez: Could you please move all these formatter stuff to IdBased (and also SearchBased) EntryFetcher.processQuery() since it applies to every fetcher. I think the simplest solution is to create a new class which applies the formatters after fetching.

src/main/java/net/sf/jabref/logic/importer/fetcher/IsbnFetcher.java
```diff
@@ -0,0 +1,101 @@
...
+        BibEntry entry = null;
+
+        if (isbn.isValidChecksum() && isbn.isValidFormat()) {
+            try {
+                q = URLEncoder.encode(identifier, StandardCharsets.UTF_8.name());
+            } catch (UnsupportedEncodingException e) {
+                LOGGER.warn("Shouldn't happen...", e);
+            }
+
+            String urlString = String.format(URL_PATTERN, q);
```
tobiasdiez: Use UrlBuilder to create the url.

src/test/java/net/sf/jabref/logic/importer/fetcher/IsbnFetcherTest.java
```diff
@@ -0,0 +1,74 @@
...
+        bibEntry.setField("bibtexkey","9780321356680");
+        bibEntry.setField("title", "Effective Java");
+        bibEntry.setField("publisher", "Addison Wesley");
+        bibEntry.setField("year", "2008");
+        bibEntry.setField("author", "Joshua Bloch");
+        bibEntry.setField("date", "2008-05-08");
+        bibEntry.setField("ean", "9780321356680");
+        bibEntry.setField("isbn", "0321356683");
+        bibEntry.setField("pagetotal", "384 Seiten");
+        bibEntry.setField("url", "http://www.ebook.de/de/product/6441328/joshua_bloch_effective_java.html");
```
tobiasdiez: I think we don't need to advertise `ebook.de` so clear the url field in the fetcher.

src/test/java/net/sf/jabref/logic/importer/fetcher/IsbnFetcherTest.java
```diff
@@ -0,0 +1,74 @@
...
+        bibEntry.setType(BibLatexEntryTypes.BOOK);
+        bibEntry.setField("bibtexkey","9780321356680");
+        bibEntry.setField("title", "Effective Java");
+        bibEntry.setField("publisher", "Addison Wesley");
+        bibEntry.setField("year", "2008");
+        bibEntry.setField("author", "Joshua Bloch");
+        bibEntry.setField("date", "2008-05-08");
+        bibEntry.setField("ean", "9780321356680");
+        bibEntry.setField("isbn", "0321356683");
+        bibEntry.setField("pagetotal", "384 Seiten");
```
tobiasdiez: Remove "Seiten"

src/test/java/net/sf/jabref/logic/importer/fetcher/IsbnFetcherTest.java
```diff
@@ -0,0 +1,74 @@
...
+    public void testHelpPage() {
+        assertEquals("ISBNtoBibTeXHelp", fetcher.getHelpPage().getPageName());
+    }
+
+    @Test
+    public void testFetcher10() throws FetcherException, IOException {
+        Optional<BibEntry> isbn = fetcher.performSearchById("0321356683");
+        Assert.assertNotNull(isbn);
+        assertEquals(bibEntry, isbn.get());
+
```
tobiasdiez: Remove empty lines

src/test/java/net/sf/jabref/logic/importer/fetcher/IsbnFetcherTest.java
```diff
@@ -0,0 +1,74 @@
...
+    @Test
+    public void testHelpPage() {
+        assertEquals("ISBNtoBibTeXHelp", fetcher.getHelpPage().getPageName());
+    }
+
+    @Test
+    public void testFetcher10() throws FetcherException, IOException {
+        Optional<BibEntry> isbn = fetcher.performSearchById("0321356683");
+        Assert.assertNotNull(isbn);
+        assertEquals(bibEntry, isbn.get());
```
tobiasdiez: Compare to Optional.of(bibEntry))

src/main/java/net/sf/jabref/logic/importer/fetcher/IsbnFetcher.java
```diff
@@ -0,0 +1,101 @@
...
+                        if (Globals.prefs.getBoolean(JabRefPreferences.USE_CASE_KEEPER_ON_SEARCH)) {
+                            title = protectTermsFormatter.format(title);
+                        }
+                        entry.setField("title", title);
+                    }
+                    result = Optional.ofNullable(entry);
+                    return result;
+                }
+            } catch (IOException e) {
+                LOGGER.warn("Error opening URL");
```
tobiasdiez: throw FetcherException here

src/test/java/net/sf/jabref/logic/importer/fetcher/IsbnFetcherTest.java
```diff
@@ -0,0 +1,74 @@
...
+
+    @Test
+    public void testFetcher13() throws FetcherException, IOException {
+        Optional<BibEntry> isbn = fetcher.performSearchById("978-0321356680");
+        Assert.assertNotNull(isbn);
+        assertEquals(bibEntry, isbn.get());
+
+    }
+
+    @Test(expected = NoSuchElementException.class)
```
tobiasdiez: Why NoSuchElementException? The fetcher should only throw FetcherExceptions (could you please add this fact to the interface description, thx)

src/test/resources/net/sf/jabref/logic/importer/fetcher/IsbnFetcherTest.bib
```diff
@@ -0,0 +1,11 @@
+@Book{9780321356680,
```
tobiasdiez: Not required I think

tobiasdiez:
I know, WIP but it look so good that I already reviewed it. After fixing my small remarks, this should be good to go

src/main/java/net/sf/jabref/gui/mergeentries/FetchAndMergeEntry.java
```diff
@@ -58,8 +58,9 @@ public FetchAndMergeEntry(BibEntry entry, BasePanel panel, List<String> fields)
 
                     try {
                         fetchedEntry = new IsbnFetcher().performSearchById(fieldContent.get());
-                    } catch (FetcherException fe) {
-                        fe.printStackTrace();
+                    } catch (FetcherException e) {
+                        panel.frame().setStatus(
+                                Localization.lang("Cannot get info based on given %0:_%1", type, fieldContent.get()));
```
Siedlerchr: Please log the exception, too, e.g. Logger.error/debug (just look at other code pieces :smile:  )
And I think the underscore in the Localization is wrong, should be just a space.

src/test/java/net/sf/jabref/logic/importer/fetcher/IsbnFetcherTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Before
+    public void setUp() {
+        Globals.prefs = JabRefPreferences.getInstance();
+        fetcher = new IsbnFetcher();
+
+        bibEntry = new BibEntry();
+        bibEntry.setType(BibLatexEntryTypes.BOOK);
+        bibEntry.setField("bibtexkey","9780321356680");
+        bibEntry.setField("title", "Effective Java");
```
tschechlovdev: use ``FieldName.TITLE``. For the other fields as well.

src/main/java/net/sf/jabref/logic/importer/fetcher/IsbnFetcher.java
```diff
@@ -0,0 +1,94 @@
...
+            try {
+                //Build the URL. In this case: http://www.ebook.de/de/tools/isbn2bibtex?isbn=identifier
+                URIBuilder uriBuilder = new URIBuilder(URL_PATTERN);
+                uriBuilder.addParameter("isbn", identifier);
+                URL url = uriBuilder.build().toURL();
+
+                //Downloads the source code of the site and then creates a .bib file out of the String
+                URLDownload urlDownload = new URLDownload(url);
+                String bibtexString = urlDownload.downloadToString(StandardCharsets.UTF_8);
+                BibEntry entry = BibtexParser.singleFromString(bibtexString, ImportFormatPreferences.fromPreferences(Globals.prefs));
```
tschechlovdev: If I see right, then the entry could be null here. Maybe you should add then a check if the entry is null, otherwise you could get a NPE below.

src/test/java/net/sf/jabref/logic/importer/fetcher/IsbnFetcherTest.java
tobiasdiez: not in the tests, these should be as literal as possible.

src/main/java/net/sf/jabref/logic/importer/fetcher/IsbnFetcher.java
tobiasdiez: If this is the case, then add `Optional<BibEntry> BibtexParser.singleFromStringOptional` and use it (and mark the other method as deprecated).

tschechlovdev:
LGTM :+1:

src/main/java/net/sf/jabref/logic/importer/fetcher/IsbnFetcher.java
```diff
@@ -43,6 +49,9 @@
 
     private static final String URL_PATTERN = "http://www.ebook.de/de/tools/isbn2bibtex?";
 
+    private final ProtectTermsFormatter protectTermsFormatter = new ProtectTermsFormatter(new ProtectedTermsLoader(new ProtectedTermsPreferences(ProtectedTermsLoader.getInternalLists(), Collections.emptyList(), Collections.emptyList(), Collections.emptyList())));
```
tobiasdiez: What was the reason to reintroduce these formatters? I really don't like them in the fetchers.

zellerdev: I wasn't sure about that. They caused problems during the tests. After I found a way that the tests worked again I thought I add them. But I don't like them in the fetchers, too.

src/main/java/net/sf/jabref/logic/importer/fetcher/IsbnFetcher.java
```diff
@@ -72,20 +68,23 @@ public HelpFile getHelpPage() {
...
+                if (entry.isPresent()) {
+                    BibEntry bibEntry = entry.get();
+                    if (bibEntry.hasField(FieldName.URL)) {
+                        bibEntry.clearField(FieldName.URL);
+                    }
 
-                result = Optional.ofNullable(entry);
+                    //Removes every non-digit character in the PAGETOTAL field.
+                    Optional<String> pagetotal = bibEntry.getFieldOptional(FieldName.PAGETOTAL);
+                    if (pagetotal.isPresent()) {
```
Siedlerchr: Just a small thing, but here the pagetotal.ifPresent(Consumer) would make sense as you only have on single statement.

src/main/java/net/sf/jabref/logic/importer/fetcher/IsbnFetcher.java
tobiasdiez: Ok, in this case just remove them. They can be latter added as configurable cleanup operations if this is needed.

src/main/java/net/sf/jabref/logic/importer/fetcher/IsbnFetcher.java
```diff
@@ -0,0 +1,93 @@
...
+
+        if (isbn.isValidChecksum() && isbn.isValidFormat()) {
+            try {
+                //Build the URL. In this case: http://www.ebook.de/de/tools/isbn2bibtex?isbn=identifier
+                URIBuilder uriBuilder = new URIBuilder(URL_PATTERN);
+                uriBuilder.addParameter("isbn", identifier);
+                URL url = uriBuilder.build().toURL();
+
+                //Downloads the source code of the site and then creates a .bib file out of the String
+                URLDownload urlDownload = new URLDownload(url);
```
stefan-kolb: Maybe we should use Unirest like in the other parsers here? personally I want to switch from our own URLDownload implementation.

src/test/java/net/sf/jabref/logic/importer/fetcher/IsbnFetcherTest.java
```diff
@@ -0,0 +1,61 @@
...
+        bibEntry.setField("ean", "9780321356680");
+        bibEntry.setField("isbn", "0321356683");
+        bibEntry.setField("pagetotal", "384");
+    }
+
+    @Test
+    public void testName() {
+        assertEquals("ISBN", fetcher.getName());
+    }
+
```
stefan-kolb: Please add some more test for invalid input and error scenarios. Also see coverage info for that.

src/main/java/net/sf/jabref/logic/importer/fetcher/IsbnFetcher.java
```diff
@@ -0,0 +1,93 @@
...
+                        bibEntry.clearField(FieldName.URL);
+                    }
+
+                    //Removes every non-digit character in the PAGETOTAL field.
+                    Optional<String> pagetotal = bibEntry.getFieldOptional(FieldName.PAGETOTAL);
+                    if (pagetotal.isPresent()) {
+                        bibEntry.setField(FieldName.PAGETOTAL, pagetotal.get().replaceAll("[\\D]", ""));
+                    }
+
+                    result = Optional.of(bibEntry);
```
tobiasdiez: use `return Optional.of(...)` instead of using a `result` variable.

src/main/java/net/sf/jabref/logic/importer/fetcher/IsbnFetcher.java
```diff
@@ -0,0 +1,93 @@
...
+                uriBuilder.addParameter("isbn", identifier);
+                URL url = uriBuilder.build().toURL();
+
+                //Downloads the source code of the site and then creates a .bib file out of the String
+                URLDownload urlDownload = new URLDownload(url);
+                String bibtexString = urlDownload.downloadToString(StandardCharsets.UTF_8);
+                Optional<BibEntry> entry = BibtexParser.singleFromStringOptional(bibtexString, ImportFormatPreferences.fromPreferences(Globals.prefs));
+
+                if (entry.isPresent()) {
+                    BibEntry bibEntry = entry.get();
```
tobiasdiez: move the code of the if statement to new private method `BibEntry postProcessEntry(BibEntry)`. Then you can simply say `return entry.map(postProcessEntry);` and the code gets more readable.

tobiasdiez:
LGTM :+1: 
Can be merged after the remarks are fixed.

src/main/java/net/sf/jabref/logic/importer/fetcher/IsbnFetcher.java
```diff
@@ -66,28 +66,31 @@ public HelpFile getHelpPage() {
                 URL url = uriBuilder.build().toURL();
 
                 //Downloads the source code of the site and then creates a .bib file out of the String
-                URLDownload urlDownload = new URLDownload(url);
-                String bibtexString = urlDownload.downloadToString(StandardCharsets.UTF_8);
+                String bibtexString = Unirest.get(url.toString()).asString().getBody();
                 Optional<BibEntry> entry = BibtexParser.singleFromStringOptional(bibtexString, ImportFormatPreferences.fromPreferences(Globals.prefs));
```
oscargus: Please inject the `ImportFormatPreferences` in the constructor. I've worked hard recently to get rid of all `Globals` references in `logic`. ;-)

src/main/java/net/sf/jabref/importer/fetcher/IdBasedEntryFetcher.java
```diff
@@ -0,0 +1,82 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+package net.sf.jabref.importer.fetcher;
```
oscargus: Please move to `gui.importer.fetcher` as the current location is obsolete.

koppor: I think, the idea is to move out of `gui` to have gui independent fetchers, so the location should be `net.sf.jabref.logic.importer`, shouldn't it? I am unsure, but `gui` feels wrong.

Refs #1594 @tobiasdiez

oscargus:
Apart from the obsolete location and the `Globals`: LGTM! :+1:

src/main/java/net/sf/jabref/logic/importer/fetcher/IsbnFetcher.java
```diff
@@ -66,28 +66,31 @@ public HelpFile getHelpPage() {
...
-            } catch (IOException | URISyntaxException e) {
+
+            } catch (UnirestException | IOException | URISyntaxException e) {
                 throw new FetcherException("Bad URL when fetching ISBN info", e);
             }
         }
         return result;
     }
+
+    private Optional<BibEntry> postProcessEntry(BibEntry entry) {
```
koppor: I would return a BibEntry here as a BibEntry comes in and this is post processed. Always. The Optional is always present. Thus, the Optional conversion should be done by the caller.

src/main/java/net/sf/jabref/importer/fetcher/IdBasedEntryFetcher.java
oscargus: This class is the current wrapper between the new `logic`-based fetchers and the old `gui`-based. Compare with https://github.com/JabRef/jabref/blob/master/src/main/java/net/sf/jabref/gui/importer/fetcher/SearchBasedEntryFetcher.java

oscargus: And there is nothing else in `net.sf.jabref.logic.importer` after #1648 :-)

stefan-kolb:
@oscargus  @tobiasdiez  @koppor Are we ready to merge this?

oscargus:
I would prefer that my comments are taken care of. On the other hand, it is quite quickly done, so not a show stopper if I have to do it myself after merging.

stefan-kolb:
Let's wait for the comments to be implemented!

zellerdev:
I will try to implement the comments by Thursdsay afternoon due to an exam on that midday. I am sorry for the waiting time.

zellerdev:
Are the changes as expected?

stefan-kolb:
@oscargus

src/main/java/net/sf/jabref/logic/importer/fetcher/IsbnFetcher.java
```diff
@@ -0,0 +1,101 @@
...
+
+/**
+ * Fetcher for ISBN.
+ */
+public class IsbnFetcher implements IdBasedFetcher {
+
+    private static final String URL_PATTERN = "http://www.ebook.de/de/tools/isbn2bibtex?";
+    private ImportFormatPreferences prefs;
+
+    public IsbnFetcher(){
```
oscargus: I'd prefer that the constructor gets the ImportFormatPreferences object here (and, hence, move the `ImportFormatPreferences.fromPreferences(Globals.prefs)` 
to `EntryFetchers` as

`entryFetchers.add(new IdBasedEntryFetcher(new IsbnFetcher(ImportFormatPreferences.fromPreferences(Globals.prefs))));`

In that way there won't be any import of Globals in logic. But a minor thing at the moment. Still things to fix which are much harder compared to this before being able to enforce it through tests.

oscargus:
It is OK to merge for me. I commented on what I really would have liked to see, but it is easy to change later and the change provided makes it even easier. The location of `IdBasedFetcher` is perfect! :+1:

src/main/java/net/sf/jabref/logic/importer/fetcher/IsbnFetcher.java
oscargus: Now it is: Globals.prefs.getImportFormatPreferences ()

oscargus:
LGTM! :+1: (The copyright comments should be removed as we changed license, but, again, minor detail and nothing that would stop me from merging.)

src/main/java/net/sf/jabref/gui/importer/fetcher/IdBasedEntryFetcher.java
```diff
@@ -0,0 +1,80 @@
+/*
+ * Copyright (C) 2003-2016 JabRef contributors.
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
```
stefan-kolb: Please remove the header.

src/main/java/net/sf/jabref/logic/importer/fetcher/IsbnFetcher.java
```diff
@@ -0,0 +1,100 @@
+/*
+ * Copyright (C) 2003-2016 JabRef contributors.
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
```
stefan-kolb: Please remove the header.

koppor:
Is it possible to update the fetcher to fix https://github.com/koppor/jabref/issues/125? I think, just the normalization functionality has to be called. Maybe a cleanup job? Maybe after #1929 is merged to use the `postProcess` method?

PR #1659, opened 2016-08-01
PushToApplication cleanup and refactoring

oscargus:
Moved PushToApplication functionality from external.push to gui.push and cleaned up a bit, like getting rid of the static list.


simonharrer:
LGTM

PR #1661, opened 2016-08-02
Replaced some array return types

oscargus:

### DIFF
--- a/src/main/java/net/sf/jabref/gui/groups/EntryTableTransferHandler.java
+++ b/src/main/java/net/sf/jabref/gui/groups/EntryTableTransferHandler.java
@@ -315,11 +315,9 @@ public class EntryTableTransferHandler extends TransferHandler {
      * @return success status for the operation
      */
     private boolean handleDraggedFiles(List<File> files, final int dropRow) {
-        final String[] fileNames = new String[files.size()];
-        int i = 0;
+        final List<String> fileNames = new ArrayList<>();
         for (File file : files) {
-            fileNames[i] = file.getAbsolutePath();
-            i++;
+            fileNames.add(file.getAbsolutePath());
         }
         // Try to load BIB files normally, and import the rest into the current
         // database.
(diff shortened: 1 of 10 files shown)


src/main/java/net/sf/jabref/gui/groups/EntryTableTransferHandler.java
```diff
@@ -315,11 +315,9 @@ private boolean handleDraggedFilenames(String s, final int dropRow) {
      * @return success status for the operation
      */
     private boolean handleDraggedFiles(List<File> files, final int dropRow) {
-        final String[] fileNames = new String[files.size()];
-        int i = 0;
+        final List<String> fileNames = new ArrayList<>();
```
Siedlerchr: The loop could be replaced with a nice stream/lamda expression:
`fileNames = new ArrayList<>(files.stream().map(f->f.getAbsolutePath()).collect(Collectors.toList());`

Siedlerchr:
Except for the remark LGTM :+1:

src/main/java/net/sf/jabref/gui/groups/EntryTableTransferHandler.java
oscargus: Yes, but I know that some developers not always appreciate these as they are harder to read. Have not really understood when they are preferred.

Siedlerchr: I do not think they are harder to read, it is just some novel style, once you get used to it, it is plain simple and makes programming for fluent. In general I would prefer them if it makes sense.
The functional style makes it easier to crate a kind of  Collection-pipelline, e.g. introduce a filter, another transformation or limit the results or just replace the collector.  And could be easy parallelized, in cases where needed.

oscargus: I do not have any strong opinions but I know I made a PR a number of months
ago where I used these a lot and there were some objections.

simonharrer: I think we do not have a full consensus on this. But I would agree to @Siedlerchr in that case. :)

PR #1662, opened 2016-08-02
Fixed most of the ErrorProne warnings

oscargus:
introduced when updating to the latest ErrorProne version.


stefan-kolb:
LGTM :+1:

Issue #1665, opened 2016-08-02
Importing an entry, that were exported with BibTeXML format failed. 


tschechlovdev:
JabRef version
JabRef 3.5
windows 8.1 6.3 amd64
Java 1.8.0_25

Steps to reproduce:

1. Select an entry in JabRef
2. Go to "File" and then select "Export"
3. Use "BibTeXML" as export format.
4. Import the exported entry into a new database

I get hte following warning:
"1. The character reference must end with the ';' delimiter."

And after that no entries were imported.

tschechlovdev:
I think the reason for this is, that by the authors, the "," gets exported as "and &#44 ". So for example ``von Hippel, Eric`` gets exported as ``von Hippel and &#44 Eric ``. And then a SAXParser Exception is thrown, because after ``&#44`` is a ';' missing.

boceckts:
I think the authors aren't escaped at all because with my test file I also found out that &#45, &#34 and &#39 are not escaped in the authors field. After adding a";" after each of these codes I was able to import my xml file.

matthiasgeiger:
I think the field gets generally escaped correctly, but the `;` which marks the end of an HTML entity such as `&#44;` is stripped or interpreted as an delimiter between different authors.

stefan-kolb:
Related to #1337?!

matthiasgeiger:
Not only related but a duplicate ;-)

Issue #898, opened 2016-03-01
BibteXML syntax


zellerdev:
The BibteXMLImporter only accepts xml files where tags have a syntax like `<bibtex:author>`. Otherwise JabRef does not read the data or even the file. Shouldn't it be possible to read files with normal syntax?
I created a few tests for this Importer. Two tests of them are ignored, because otherwise circleci would return an error.

Refs #511

koppor:
The BibTeXML parser is implemented in a wrong way. [BibTeXML](http://bibtexml.sourceforge.net/) is a DTD document. XML is namespace aware. Therefore, the document should be parsed namespace aware.

Proposal: Rewrite the whole BibTeXML parser using `javax.xml.parsers.DocumentBuilderFactory`.

koppor:
Update: Schemas are available at https://github.com/Zearin/BibTeXML/tree/master/bibtexml/schema.I would propose to build on the XML Schema

tschechlovdev:
I would like to take care of this issue. But  in my opinion a JAXB parser would be easier than just using a DOM parser. Would this be also ok?

Siedlerchr:
JAXB sounds like a good idea. Mabye you can reuse some of your code.

tschechlovdev:
#1666 is merged, so I think this can be closed.

PR #1666, opened 2016-08-02
Rewrite bibtexml importer with JAXB parser

tschechlovdev:
Regarding: #898 
I've rewritten the bibtexml importer: It was written with a SAX parser and I've used a JAXB parser. I will be addings some more tests for the changes I've made.

Note: Exporting in bibtexml format and then trying to import in bibtexml format is not working because of #1665.


src/main/java/net/sf/jabref/importer/fileformat/BibTeXMLImporter.java
```diff
@@ -78,33 +102,539 @@ public ParserResult importDatabase(BufferedReader reader) throws IOException {
...
+                }
+                if (entry.getTechreport() != null) {
+                    bibEntry.setType("techreport");
+                    parseTechReport(entry.getTechreport(), fields);
+                }
+                if (entry.getUnpublished() != null) {
+                    bibEntry.setType("unpublished");
+                    parseUnpublished(entry.getUnpublished(), fields);
+                }
+                if (entry.getId() != null) {
```
boceckts: This if condition shouldn't be necessary as the xsd schema file specifies the attribute as `use="required"`

src/main/java/net/sf/jabref/importer/fileformat/BibTeXMLImporter.java
```diff
@@ -78,33 +102,539 @@ public ParserResult importDatabase(BufferedReader reader) throws IOException {
...
+        putIfValueNotNull(fields, FieldName.KEY, unpublished.getKey());
+        putIfValueNotNull(fields, FieldName.KEYWORDS, unpublished.getKeywords());
+        putIfValueNotNull(fields, FieldName.LANGUAGE, unpublished.getLanguage());
+        putIfValueNotNull(fields, "lccn", unpublished.getLccn());
+        putIfValueNotNull(fields, FieldName.LOCATION, unpublished.getLocation());
+        putIfValueNotNull(fields, FieldName.MONTH, unpublished.getMonth());
+        putIfValueNotNull(fields, "mrn-number", unpublished.getMrnumber());
+        putIfValueNotNull(fields, FieldName.NOTE, unpublished.getNote());
+        putIfValueNotNull(fields, FieldName.TITLE, unpublished.getTitle());
+        putIfValueNotNull(fields, FieldName.URL, unpublished.getUrl());
```
boceckts: There is quite a lot of redundant code, please extract the code snippets with the same code to a new method. You can create your methods after the same schema as the xsd:group elements are build in the file bibtexml.xsd.

For example extract all fields that are specified in the "user" group starting at line 77 in bibtexml.xsd and create a method `private void putIfUserNotNull()`.

src/main/java/net/sf/jabref/importer/fileformat/BibTeXMLImporter.java
```diff
@@ -78,33 +102,539 @@ public ParserResult importDatabase(BufferedReader reader) throws IOException {
...
 
+    private void parseUnpublished(Unpublished unpublished, HashMap<String, String> fields) {
+        putIfValueNotNull(fields, FieldName.ABSTRACT, unpublished.getAbstract());
+        putIfValueNotNull(fields, "affiliation", unpublished.getAffiliation());
+        putIfValueNotNull(fields, FieldName.AUTHOR, unpublished.getAuthor());
+        putIfValueNotNull(fields, "category", unpublished.getCategory());
+        putIfValueNotNull(fields, "copyright", unpublished.getCopyright());
+        putIfValueNotNull(fields, FieldName.CROSSREF, unpublished.getCrossref());
+        putIfValueNotNull(fields, FieldName.ISBN, unpublished.getIsbn());
+        putIfValueNotNull(fields, FieldName.ISSN, unpublished.getIssn());
```
boceckts: According to the bibtexml.xsd file an entry can have either an isbn or an issn number but not both.

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - It is now possible to generate a new BIB database from the citations in an OpenOffice/LibreOffice document
 - The arXiv fetcher now also supports free-text search queries
 - [#1345](https://github.com/JabRef/jabref/issues/1345) Cleanup ISSN
+- [#898](https://github.com/JabRef/jabref/issues/898) Rewrite bibtexml importer with JAXB parser
```
boceckts: Make a less technical description in the changelog and rather explain what has changed for the user.

src/main/java/net/sf/jabref/importer/fileformat/BibTeXMLImporter.java
```diff
@@ -189,254 +183,46 @@ public ParserResult importDatabase(BufferedReader reader) throws IOException {
...
+        for (Method method : declaredMethods) {
+            try {
+                if (method.getName().equals("getYear")) {
+                    putYear(fields, (XMLGregorianCalendar) method.invoke(t));
+                    continue;
+                }
+                if (method.getName().equals("getNumber")) {
+                    putNumber(fields, (BigInteger) method.invoke(t));
+                    continue;
+                }
```
boceckts: Use `else if`

src/main/java/net/sf/jabref/importer/fileformat/BibTeXMLImporter.java
```diff
@@ -189,254 +183,46 @@ public ParserResult importDatabase(BufferedReader reader) throws IOException {
...
+                if (isFieldToSkip(method.getName())) {
+                    continue;
+                }
+                if (method.getName().contains("get")) {
+                    putIfValueNotNull(fields, method.getName().replace("get", ""), (String) method.invoke(t));
+                }
```
boceckts: Please append this to the first if statement with an `else if`

src/main/java/net/sf/jabref/importer/fileformat/BibTeXMLImporter.java
```diff
@@ -106,7 +110,7 @@ public ParserResult importDatabase(BufferedReader reader) throws IOException {
             File file = (File) unmarshaller.unmarshal(xmlReader);
 
             List<Entry> entries = file.getEntry();
-            HashMap<String, String> fields = new HashMap<>();
+            Map<String, String> fields = new HashMap<String, String>();
```
boceckts: The second type arguments are redundant, please rewrite as 
```
Map<String, String> fields = new HashMap<>();
```

src/main/java/net/sf/jabref/importer/fileformat/BibTeXMLImporter.java
```diff
@@ -78,33 +97,177 @@ public ParserResult importDatabase(BufferedReader reader) throws IOException {
...
         return new ParserResult(bibItems);
     }
 
+    /**
+     * In this method, all <Code>get</Code> methods that t has will be used and their value will be put to fields,
+     * if it is not null. So for example if t has the method <Code>getAbstract</Code>, then
+     * "abstract" will be put as key to fields and the value of <Code>getAbstract</Code> will be put as value to fields.
+     * Some <Code>get</Code> methods shouldn't be mapped to fields, so <Code>getClass</Code> for example will be skipped.
+     *
+     * @param t
```
koppor: Either write a parameter description or leave the tag out.

boceckts: Also rename, maybe `entryType` is fitting?

src/main/java/net/sf/jabref/importer/fileformat/BibTeXMLImporter.java
```diff
@@ -78,33 +97,216 @@ public ParserResult importDatabase(BufferedReader reader) throws IOException {
...
+            } else if (elementValue instanceof XMLGregorianCalendar) {
+                XMLGregorianCalendar value = (XMLGregorianCalendar) elementValue;
+                if (FieldName.YEAR.equals(localName)) {
+                    putYear(fields, value);
+                }
+            }
+        }
+    }
+
+    private void putStandardFields(String localName, String value, Map<String, String> fields) {
```
boceckts: Add a comment explaining the method and where you took the "standard" fields from; e.g. from some group definition of the xsd file?

src/main/java/net/sf/jabref/importer/fileformat/BibTeXMLImporter.java
tschechlovdev: I renamed it to entryType and added a description.

src/main/java/net/sf/jabref/importer/fileformat/BibTeXMLImporter.java
```diff
@@ -78,33 +98,177 @@ public ParserResult importDatabase(BufferedReader reader) throws IOException {
...
-            return ParserResult.fromErrorMessage(e.getLocalizedMessage());
-        } catch (IOException e) {
-            LOGGER.error("Error during file import", e);
-            return ParserResult.fromErrorMessage(e.getLocalizedMessage());
         }
         return new ParserResult(bibItems);
     }
 
+    /**
+     * In this method, all <Code>get</Code> methods that t has will be used and their value will be put to fields,
```
koppor: `t` -> `entryType`

src/main/java/net/sf/jabref/importer/fileformat/BibTeXMLImporter.java
```diff
@@ -78,33 +98,177 @@ public ParserResult importDatabase(BufferedReader reader) throws IOException {
...
+     * if it is not null. So for example if entryType has the method <Code>getAbstract</Code>, then
+     * "abstract" will be put as key to fields and the value of <Code>getAbstract</Code> will be put as value to fields.
+     * Some <Code>get</Code> methods shouldn't be mapped to fields, so <Code>getClass</Code> for example will be skipped.
+     *
+     * @param entryType This can be all possible BibTeX types. It contains all fields of the entry and their values.
+     * @param fields A map where the name and the value of all fields, that the entry contains, will be put.
```
koppor: Remove commas

src/main/java/net/sf/jabref/importer/fileformat/BibTeXMLImporter.java
```diff
@@ -78,33 +98,177 @@ public ParserResult importDatabase(BufferedReader reader) throws IOException {
...
-            LOGGER.error("Error during XML parsing", e);
-            return ParserResult.fromErrorMessage(e.getLocalizedMessage());
-        } catch (IOException e) {
-            LOGGER.error("Error during file import", e);
-            return ParserResult.fromErrorMessage(e.getLocalizedMessage());
         }
         return new ParserResult(bibItems);
     }
 
+    /**
```
koppor: Add some more explanation why you choose that implementation method and not working on the real classes.

src/main/java/net/sf/jabref/importer/fileformat/BibTeXMLImporter.java
```diff
@@ -78,33 +98,177 @@ public ParserResult importDatabase(BufferedReader reader) throws IOException {
...
+    }
+
+    /**
+     * Returns whether the value of the given method name should be mapped or whether the field can be skipped.
+     *
+     * @param name of a method
+     * @return true if the field can be skipped, else false
+     */
+    private boolean isFieldToSkip(String name) {
+        return FIELDS_TO_SKIP.contains(name);
```
koppor: Either remove the `get` prefix and check for ignored case or name this "ignoredMethods" etc. --> `getSize` is not a bibtex field.

src/main/java/net/sf/jabref/importer/fileformat/BibTeXMLImporter.java
```diff
@@ -78,33 +98,177 @@ public ParserResult importDatabase(BufferedReader reader) throws IOException {
...
+    }
+
+    private void parseInbook(Inbook inbook, Map<String, String> fields) {
```
koppor: Why do you need a special Treatment for Inbook, but not for the others?

src/main/java/net/sf/jabref/importer/fileformat/BibTeXMLImporter.java
tschechlovdev: done.

src/main/java/net/sf/jabref/importer/fileformat/BibTeXMLImporter.java
tschechlovdev: done.

src/main/java/net/sf/jabref/importer/fileformat/BibTeXMLImporter.java
tschechlovdev: done.

src/main/java/net/sf/jabref/importer/fileformat/BibTeXMLImporter.java
tschechlovdev: done.

src/main/java/net/sf/jabref/importer/fileformat/BibTeXMLImporter.java
tschechlovdev: done.

tschechlovdev:
Are there any other remarks?

src/main/java/net/sf/jabref/logic/importer/fileformat/BibTeXMLImporter.java
```diff
@@ -79,33 +98,184 @@ public ParserResult importDatabase(BufferedReader reader) throws IOException {
...
-            BibTeXMLHandler handler = new BibTeXMLHandler();
-            // Start the parser. It reads the file and calls methods of the handler.
-            parser.parse(new InputSource(reader), handler);
-            // When you're done, report the results stored by your handler object
-            bibItems.addAll(handler.getItems());
-
-        } catch (javax.xml.parsers.ParserConfigurationException e) {
+            JAXBContext context = JAXBContext.newInstance("net.sf.jabref.importer.fileformat.bibtexml");
+            XMLInputFactory xmlInputFactory = XMLInputFactory.newFactory();
+            XMLStreamReader xmlReader = xmlInputFactory.createXMLStreamReader(reader);
```
koppor: Why can't directly the reader used as input for the unmarshaller?

The JavaDocs of the unmarshaller say so: https://jaxb.java.net/nonav/2.2.4/docs/api/javax/xml/bind/Unmarshaller.html

Reading that, the code would be much shorter and especially lines 106 to 109 would be obsolete

src/main/java/net/sf/jabref/logic/importer/fileformat/BibTeXMLImporter.java
```diff
@@ -79,33 +98,184 @@ public ParserResult importDatabase(BufferedReader reader) throws IOException {
...
+            List<Entry> entries = file.getEntry();
+            Map<String, String> fields = new HashMap<>();
+
+            for (Entry entry : entries) {
+                BibEntry bibEntry = new BibEntry(DEFAULT_BIBTEXENTRY_ID);
+                if (entry.getArticle() != null) {
+                    bibEntry.setType("article");
+                    parse(entry.getArticle(), fields);
+                }
+                if (entry.getBook() != null) {
```
koppor: Why not `} else if (`? Isn't it the case, that is is either article, or book, or booklet, etc?

src/main/java/net/sf/jabref/logic/importer/fileformat/BibTeXMLImporter.java
```diff
@@ -79,33 +98,184 @@ public ParserResult importDatabase(BufferedReader reader) throws IOException {
...
+            }
+
+            Unmarshaller unmarshaller = context.createUnmarshaller();
+            File file = (File) unmarshaller.unmarshal(xmlReader);
+
+            List<Entry> entries = file.getEntry();
+            Map<String, String> fields = new HashMap<>();
+
+            for (Entry entry : entries) {
+                BibEntry bibEntry = new BibEntry(DEFAULT_BIBTEXENTRY_ID);
```
koppor: I think, the variable `DEFAULT_BIBTEXENTRY_ID` should be removed and the default constructor without parameter be used.

src/main/java/net/sf/jabref/logic/importer/fileformat/BibTeXMLImporter.java
```diff
@@ -79,33 +98,184 @@ public ParserResult importDatabase(BufferedReader reader) throws IOException {
...
+                    parse(entry.getConference(), fields);
+                }
+                if (entry.getInbook() != null) {
+                    bibEntry.setType("inbook");
+                    parseInbook(entry.getInbook(), fields);
+                }
+                if (entry.getIncollection() != null) {
+                    bibEntry.setType("incollection");
+                    Incollection incollection = entry.getIncollection();
+                    if (incollection.getChapter() != null) {
```
koppor: Please test the other branch (a collection without a chapter), too.

src/main/java/net/sf/jabref/logic/importer/fileformat/BibTeXMLImporter.java
```diff
@@ -79,33 +98,184 @@ public ParserResult importDatabase(BufferedReader reader) throws IOException {
...
+                if (entry.getBook() != null) {
+                    bibEntry.setType("book");
+                    parse(entry.getBook(), fields);
+                }
+                if (entry.getBooklet() != null) {
+                    bibEntry.setType("booklet");
+                    parse(entry.getBooklet(), fields);
+                }
+                if (entry.getConference() != null) {
+                    bibEntry.setType("conference");
```
koppor: Please also cover this case

src/main/java/net/sf/jabref/logic/importer/fileformat/BibTeXMLImporter.java
```diff
@@ -79,33 +98,184 @@ public ParserResult importDatabase(BufferedReader reader) throws IOException {
...
+            try {
+                if (method.getName().equals("getYear")) {
+                    putYear(fields, (XMLGregorianCalendar) method.invoke(entryType));
+                    continue;
+                } else if (method.getName().equals("getNumber")) {
+                    putNumber(fields, (BigInteger) method.invoke(entryType));
+                    continue;
+                } else if (isMethodToIgnore(method.getName())) {
+                    continue;
+                } else if (method.getName().contains("get")) {
```
koppor: Why not `startsWith`?

src/main/java/net/sf/jabref/logic/importer/fileformat/BibTeXMLImporter.java
```diff
@@ -79,33 +98,184 @@ public ParserResult importDatabase(BufferedReader reader) throws IOException {
...
+        for (JAXBElement<?> element : content) {
+            String localName = element.getName().getLocalPart();
+            Object elementValue = element.getValue();
+            if (elementValue instanceof String) {
+                String value = (String) elementValue;
+                putIfValueNotNull(fields, localName, value);
+            } else if (elementValue instanceof BigInteger) {
+                BigInteger value = (BigInteger) elementValue;
+                if (FieldName.NUMBER.equals(localName)) {
+                    putNumber(fields, value);
```
koppor: Why is that line not covered?

src/main/java/net/sf/jabref/logic/importer/fileformat/BibTeXMLImporter.java
```diff
@@ -79,33 +98,184 @@ public ParserResult importDatabase(BufferedReader reader) throws IOException {
...
+                }
+                if (FieldName.CHAPTER.equals(localName) && (value != null)) {
```
koppor: Shouldn't be that an `else if`?

src/main/java/net/sf/jabref/logic/importer/fileformat/BibTeXMLImporter.java
```diff
@@ -79,33 +98,184 @@ public ParserResult importDatabase(BufferedReader reader) throws IOException {
...
+        List<JAXBElement<?>> content = inbook.getContent();
+        for (JAXBElement<?> element : content) {
+            String localName = element.getName().getLocalPart();
+            Object elementValue = element.getValue();
+            if (elementValue instanceof String) {
+                String value = (String) elementValue;
+                putIfValueNotNull(fields, localName, value);
+            } else if (elementValue instanceof BigInteger) {
+                BigInteger value = (BigInteger) elementValue;
+                if (FieldName.NUMBER.equals(localName)) {
```
koppor: Should the `null` check not come directly before this if?

src/main/java/net/sf/jabref/logic/importer/fileformat/BibTeXMLImporter.java
```diff
@@ -79,33 +98,184 @@ public ParserResult importDatabase(BufferedReader reader) throws IOException {
...
+                    putNumber(fields, value);
+                }
+                if (FieldName.CHAPTER.equals(localName) && (value != null)) {
+                    fields.put(FieldName.CHAPTER, String.valueOf(value));
+                }
+            } else if (elementValue instanceof XMLGregorianCalendar) {
+                XMLGregorianCalendar value = (XMLGregorianCalendar) elementValue;
+                if (FieldName.YEAR.equals(localName)) {
```
koppor: What happens if there is another localName used? Maybe, just log it to indicate that is is an unexpected file?

src/main/java/net/sf/jabref/logic/importer/fileformat/BibTeXMLImporter.java
```diff
@@ -79,33 +98,184 @@ public ParserResult importDatabase(BufferedReader reader) throws IOException {
...
+                    putYear(fields, value);
+                }
+            }
```
koppor: Shouldn't there be an `else` branch logging that an unexpected field was found?

tschechlovdev: I don't think that this is necessary. If there is an unexpected field, it will not be mapped by JAXB. So if I add here an else, it would be unreachable in my opinion. I've also tested this by adding a field in the ``BibTeXMLImporterTestInbookLessFields`` file by adding a field, that should not appear in the Inbook type.

src/main/java/net/sf/jabref/logic/importer/fileformat/BibTeXMLImporter.java
tschechlovdev: See my comment below. I don't think that this can happen here, because it should be skipped JAXB.

src/main/java/net/sf/jabref/logic/importer/fileformat/BibTeXMLImporter.java
tschechlovdev: Done. I've used directly the reader as input.

src/main/java/net/sf/jabref/logic/importer/fileformat/BibTeXMLImporter.java
tschechlovdev: Yeah you're right. Changed it to ``} else if {``

src/main/java/net/sf/jabref/logic/importer/fileformat/BibTeXMLImporter.java
tschechlovdev: Done.

src/main/java/net/sf/jabref/logic/importer/fileformat/BibTeXMLImporter.java
tschechlovdev: Done.

src/main/java/net/sf/jabref/logic/importer/fileformat/BibTeXMLImporter.java
tschechlovdev: Done.

src/main/java/net/sf/jabref/logic/importer/fileformat/BibTeXMLImporter.java
tschechlovdev: Changed to ``startsWith``.

src/main/java/net/sf/jabref/logic/importer/fileformat/BibTeXMLImporter.java
tschechlovdev: Now it should be covered.

src/main/java/net/sf/jabref/logic/importer/fileformat/BibTeXMLImporter.java
tschechlovdev: Yes. Changed it to ``else if``.

src/main/java/net/sf/jabref/logic/importer/fileformat/BibTeXMLImporter.java
tschechlovdev: Done.

tschechlovdev:
Travis is failing because of UI tests...

src/main/java/net/sf/jabref/logic/importer/fileformat/BibTeXMLImporter.java
koppor: Please add the log nevertheless. In case the XSD is updated, the code might not get updated and the import fail. Having a log (INFO), it is more easy to trace where the code has to be adapted.

Please remove this `school` line again. Create another `xml` file and name it `BibTeXMLImporterInvalidInbook.xml` . Put "should not be mapped, because not valid according to XSD" as field content.

koppor:
Please not that the UI tests are done but **ignored**. Please read the travis output and check for **red lines**

Checkstyle failed. Your IDE settings seem to be non-formant. Please double check: https://github.com/JabRef/jabref/wiki/Guidelines-for-setting-up-a-local-workspace#intellij

```
:checkstyleMain[ant:checkstyle] [ERROR] /home/travis/build/JabRef/jabref/src/main/java/net/sf/jabref/logic/importer/fileformat/BibTeXMLImporter.java:36: 'net.sf.jabref.importer.fileformat.bibtexml.Entry' should be separated from previous imports. [ImportOrder]

 FAILED

* What went wrong:

Execution failed for task ':checkstyleMain'.

> Checkstyle rule violations were found. See the report at: file:///home/travis/build/JabRef/jabref/build/reports/checkstyle/main.html

* Try:

Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

BUILD FAILED

Total time: 1 mins 16.912 secs

```

src/main/java/net/sf/jabref/logic/importer/fileformat/BibTeXMLImporter.java
tschechlovdev: Done.

PR #1671, opened 2016-08-03
Fix #1288 Newly opened bib-file is not focused

### LINKED ISSUE #1288: Newly opened bib-file is not focused
JabRef version 3.3

Steps to reproduce:
Prerequisite: JabRef is registered as the default application for *.bib files
1. Close all JabRef instances
2. Open a bib file from the file system
3. JabRef is opened, the file used least recently is selected - and not the newly opened bib file

(Side note: If remot ops are activated and JabRef is already running, the newly opened file is focused correctly)

stefan-kolb:
- [x] Open a NEW file and it is correctly focused afterwards
- [x] Open a file that was already OPENED before is correctly focused afterwards
- [x] Open a file TWO TIMES and it is only opened once


stefan-kolb:
@matthiasgeiger Should work now :smile:

matthiasgeiger:
LGTM (apart from the minor comments via Skype :wink: )

PR #1672, opened 2016-08-03
Got rid of unused preferences

oscargus:
And marked some which can not be changed from within JabRef.


oscargus:
Should we try to collect the preference strings we remove and provide a function to erase those from the preferences? The string can then be kept for a few years and finally retired for good.

lenhard:
How did you find out which preferences are unused? Did the IDE mark them as unused?

There is no need to manually cleanup the registry from old preferences. Hardly any software actually does that and we are just talking about a few key value pairs.

src/main/java/net/sf/jabref/JabRefGUI.java
```diff
@@ -88,7 +88,6 @@ public static void checkForNewVersion(boolean manualExecution) {
 
     private void openWindow() {
         // Perform checks and changes for users with a preference set from an older JabRef version.
-        PreferencesMigrations.replaceAbstractField();
```
stefan-kolb: Why did you remove this? Because it's too old?

CHANGELOG.md
```diff
@@ -51,6 +51,10 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 ### Removed
 - It is not longer possible to choose to convert HTML sub- and superscripts to equations
 - [#1610](https://github.com/JabRef/jabref/issues/1610) Removed the possibility to auto show or hide the groups interface
+- Removed option to open right-click menu with ctrl + left-click as it was not working
```
stefan-kolb: it would be nice if you had created several PRs for each change. it would be so much easier to review the PR.

oscargus:
I searched (almost) every string constant in `JabRefPreferences` and checked how they were used (with some thinking as clearly the count is not always a good indication). For most of the removed ones I also search for the actual string, just to confirm that someone had hardcoded the string instead. All four "features" are also quite straightforward to test and see if they are working and if the preference makes any difference (hint: no). Typically these were only used in the preferences and the preference dialog.

The ones marked as not changable are simply used in (typically) two places, once to set the default value and then they are used in one line of code (and that is not the preference dialog).

src/main/java/net/sf/jabref/JabRefGUI.java
oscargus: Because the code checks if GENERAL_FIELDS contains `abstract` and in that case removes if from `GENERAL_FIELDS`.

The thing is that the only place `GENERAL_FIELDS` is used in the current master is for setting a default value...

``` Java
// The general fields stuff is made obsolete by the CUSTOM_TAB_... entries.
        defaults.put(GENERAL_FIELDS, "crossref;keywords;file;doi;url;urldate;"
                + "pdf;comment;owner");
```

lenhard:
Ok, since this is somewhat hard to review by looking at the code, we'll trust you here. Please merge and rebase!

CHANGELOG.md
oscargus: I could have split it is two, as have separate ones would be a conflict heaven, especially considering that I had to manually renumber the layout of the preference tab where four entries were removed.

What I could have done though was to not group the preference keys.

The best way to review is to check the non-JabRefPreferences code, especially the preference tabs. You will notice that the key removed there does not exist in any of the other files (except JabRefPreferences). This covers most changes (GENERAL_FIELDS) is the other, as I recall it.

oscargus:
I can summarize:

* `JabRefPreferences.GROUPS_VISIBLE_ROWS` not needed since group pane now always maximizes the height.
* `JabRefPreferences.GENERAL_FIELDS` not used (so the migration is not needed)
* `JabRefPreferences.DISABLE_ON_MULTIPLE_SELECTION` never used in the code (so entry editor is never disabled on multiple selection)
* `JabRefPreferences.VALUE_DELIMITERS` (2) JabRef always use `{}` for storing fields (the comboBox was not in the layout, but in the code)
* `JabRefPreferences.DEFAULT_AUTO_SORT` No default auto sort any more (as this preference is not used except in the preference dialog)
* `JabRefPreferences.CTRL_CLICK` Ctrl + left click will never ever bring up the right-click menu (as this preference is not used except in the preference dialog)
* `JabRefPreferences.DIALOG_WARNING_FOR_DUPLICATE_KEY` and `JabRefPreferences.DIALOG_WARNING_FOR_EMPTY_KEY` there will never be any automatic warnings when entering an empty or duplicate key

stefan-kolb:
Just go ahead and merge this :smile: :+1:

PR #1673, opened 2016-08-03
Removed thrown Exception declarations

oscargus:
When they were not thrown...

### DIFF
--- a/src/test/java/net/sf/jabref/importer/fileformat/PdfXmpImporterTest.java
+++ b/src/test/java/net/sf/jabref/importer/fileformat/PdfXmpImporterTest.java
@@ -49,14 +49,14 @@ public class PdfXmpImporterTest {
     }
 
     @Test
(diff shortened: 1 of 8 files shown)


src/test/java/net/sf/jabref/importer/fileformat/PdfXmpImporterTest.java
```diff
@@ -49,14 +49,14 @@ public void testGetDescription() {
     }
 
     @Test
-    public void importEncryptedFileReturnsError() throws IOException, URISyntaxException {
+    public void importEncryptedFileReturnsError() throws URISyntaxException {
```
tobiasdiez: In tests it is preferable to mention only the generic `Exception` (since then test code is more stable when the code-under-test throws a different kind of exception).

tobiasdiez:
LGTM :+1:

PR #1677, opened 2016-08-04
[WIP] Proper copying of entries

oscargus:
I'm making a method to properly copy entries from one database to another, including crossref entries and used strings. This is useful in generating a database from an aux file and for generating a database based on an OO/LO document. It could also be useful for generating a subdatabase from selected entries (or copying relevant strings when copying entries).

There are some problems though related to the preamble. In OO/LO the entries may come from several databases and there it is not obvious how to deal with the preamble. However, a possibly solvable problem is to detect which strings are used in the preamble.

In AuxParserTest @stefan-kolb have generated a test file with a string in the preamble. JabRef doesn't parse this in a way such that the string can be resolved automatically. Earlier, the preamble and all strings were copied in AuxParser, but I was hoping to only copy the relevant strings. Any input on how the preamble issue should be solved? One way would simply be to parse the preamble for strings, I guess. Any other suggestions?

### DIFF
--- a/src/main/java/net/sf/jabref/logic/auxparser/AuxParserResult.java
+++ b/src/main/java/net/sf/jabref/logic/auxparser/AuxParserResult.java
@@ -1,12 +1,14 @@
 package net.sf.jabref.logic.auxparser;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 
 import net.sf.jabref.logic.l10n.Localization;
 import net.sf.jabref.model.database.BibDatabase;
+import net.sf.jabref.model.entry.BibtexString;
 
 public class AuxParserResult {
 
@@ -17,6 +19,7 @@ public class AuxParserResult {
     private final BibDatabase auxDatabase = new BibDatabase();
     private int nestedAuxCount;
     private int crossRefEntriesCount;
+    private int insertedStrings;
 
     public AuxParserResult(BibDatabase masterDatabase) {
         this.masterDatabase = masterDatabase;
@@ -42,6 +45,10 @@ public class AuxParserResult {
         return unresolvedKeys.size();
     }
 
+    public int getInsertedStringsCount() {
+        return insertedStrings;
+    }
+
     /**
      * Query the number of extra entries pulled in due to crossrefs from other entries.
      *
@@ -59,6 +66,13 @@ public class AuxParserResult {
         nestedAuxCount++;
     }
 
+    public void insertStrings(Collection<BibtexString> usedStrings) {
+        for (BibtexString string : usedStrings) {
+            auxDatabase.addString(string);
+            insertedStrings++;
+        }
+    }
+
     /**
      * Prints parsing statistics
      *
@@ -73,7 +87,8 @@ public class AuxParserResult {
                 .append(Localization.lang("resolved")).append(' ').append(getResolvedKeysCount()).append('\n')
                 .append(Localization.lang("not_found")).append(' ').append(getUnresolvedKeysCount()).append('\n')
                 .append(Localization.lang("crossreferenced entries included")).append(' ')
-                .append(crossRefEntriesCount).append('\n');
+                .append(crossRefEntriesCount).append(Localization.lang("strings included")).append(' ')
+                .append(insertedStrings).append('\n');
 
         if (includeMissingEntries && (getUnresolvedKeysCount() > 0)) {
             for (String entry : unresolvedKeys) {
--- a/src/main/java/net/sf/jabref/model/database/BibDatabase.java
+++ b/src/main/java/net/sf/jabref/model/database/BibDatabase.java
@@ -342,7 +342,33 @@ public class BibDatabase {
      */
     public String resolveForStrings(String content) {
         Objects.requireNonNull(content, "Content for resolveForStrings must not be null.");
-        return resolveContent(content, new HashSet<>());
+        return resolveContent(content, new HashSet<>(), null);
+    }
+
+    /**
+     * Get all strings used in the entries.
+     */
+    public Collection<BibtexString> getUsedStrings(Collection<BibEntry> entries) {
+        List<BibtexString> result = new ArrayList<>();
+        Set<String> allUsedIds = new HashSet<>();
+
+        // All entries
+        for (BibEntry entry : entries) {
+            for (String fieldContent : entry.getFieldValues()) {
+                resolveContent(fieldContent, new HashSet<>(), allUsedIds);
+            }
+        }
+
+        // Preamble
+        if (preamble != null) {
+            resolveContent(preamble, new HashSet<>(), allUsedIds);
+        }
+
+        for (String stringId : allUsedIds) {
+            result.add((BibtexString) bibtexStrings.get(stringId).clone());
+        }
+
+        return result;
     }
 
     /**
@@ -400,7 +426,7 @@ public class BibDatabase {
      * care not to follow a circular reference pattern.
      * If the string is undefined, returns null.
      */
-    private String resolveString(String label, Set<String> usedIds) {
+    private String resolveString(String label, Set<String> usedIds, Set<String> allUsedIds) {
         for (BibtexString string : bibtexStrings.values()) {
             if (string.getName().equalsIgnoreCase(label)) {
                 // First check if this string label has been resolved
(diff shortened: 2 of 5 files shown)


koppor:
Is it possible to also copy the files if the absolute directory of the file differs? The relative path of the source should be resolved against the file path of the current bib file. Including directory creation.

oscargus:
That is indeed another thing which I haven't considered. I was primarily thinking of generating a bib without files (for OO/LO and AUX), but clearly, for proper copying files should be included. Even better if everything can be added to a Zip archive as well.

koppor:
@oscargus If the only thing is that the files aren't copied, this should be ready for review, isn't it?

oscargus:
Maybe it is OK to add. It is not really what I aimed for (yet), but the string copying works for AuxParser (I think), so always something.

koppor:
Could you point to the concrete file "with a string in the preamble". Are these comment strings? I think, you reference #900?

oscargus:
https://github.com/JabRef/jabref/blob/master/src/test/resources/net/sf/jabref/logic/auxparser/config.bib

koppor:
@lenhard You recently worked on reading and writing bib databases. Can you comment on parsing `@preamble`. As far as I understood @oscargus, he wants to know which strings are used at `@preamble` and thus only copy these fields.

I don't know `@preamble`. Any pointers to documentation? :innocent: 

@oscargus: Leaving ` @preamble` and file copying aside, this should be ready for review, isn't it? I would like to get this merged to minimize the LOC for PRs. This enables us to quickly review PRs in our spare time. Longer PRs cannot be quickly reviewed.

lenhard:
The preamble is parsed as-is with formatting and everything and set as a String member of `BibDataBase`. I am unaware of any formatting or syntax restrictions within the preamble.

Honestly, I do not really get what it should be good for, it seems like a relic. Apparently it is part of Bibtex, and, therefore, we parse it, have an editor for it and serialize it. It seems that it is more relevant for programs that compile the bib file. Here are some, not very conclusive references on the preamble: [first](http://maverick.inria.fr/~Xavier.Decoret/resources/xdkbibtex/bibtex_summary.html#preamble), [second](http://www.bibtex.org/Format/), [third](https://www.cs.arizona.edu/~collberg/Teaching/07.231/BibTeX/bibtex.html). 

If you want to decide, which strings are used in a preamble, it would be best to parse the preamble string and look for occurrences of `#`. Best to keep this simple. If we miss edge cases of preamble usage, it is probably irrelevant, as hardly anybody uses it.

src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
```diff
@@ -317,4 +320,52 @@ public void resolveForStringsOddHashMarkAtTheEnd() {
...
+        BibtexString string = new BibtexString(IdGenerator.next(), "AAA", "Some other #BBB#");
+        BibtexString string2 = new BibtexString(IdGenerator.next(), "BBB", "Some more text");
+        BibtexString string3 = new BibtexString(IdGenerator.next(), "CCC", "Even more text");
+        database.addString(string);
+        database.addString(string2);
+        database.addString(string3);
+        database.insertEntry(entry);
+        List<BibtexString> usedStrings = (List<BibtexString>) database.getUsedStrings(Arrays.asList(entry));
+        assertEquals(2, usedStrings.size());
+        assertTrue((string.getName() == usedStrings.get(0).getName())
```
koppor: Can't we put all strings in a set and just compare two sets?

oscargus: That might indeed be a better solution. The problem at the moment is really that I do not know which order they are added in (or rather, I want the test to be independent of that).

oscargus:
@koppor yes, the introduced functionality works, although it is not enough to obtain the complete purpose, so it can be reviewed and merged for now.

Personally, I do not really see the need for strings in the preamble, but since we have a test for that it felt wrong to break/remove that test just because I couldn't bother to handle it...

src/main/java/net/sf/jabref/model/database/BibDatabase.java
```diff
@@ -342,7 +342,33 @@ public synchronized boolean hasStringLabel(String label) {
      */
     public String resolveForStrings(String content) {
         Objects.requireNonNull(content, "Content for resolveForStrings must not be null.");
-        return resolveContent(content, new HashSet<>());
+        return resolveContent(content, new HashSet<>(), null);
```
koppor: Please don't pass `null`. Pass an empty set.

src/main/java/net/sf/jabref/model/database/BibDatabase.java
```diff
@@ -400,7 +426,7 @@ public BibEntry resolveForStrings(BibEntry entry, boolean inPlace) {
      * care not to follow a circular reference pattern.
      * If the string is undefined, returns null.
      */
-    private String resolveString(String label, Set<String> usedIds) {
+    private String resolveString(String label, Set<String> usedIds, Set<String> allUsedIds) {
```
koppor: Please add `Objects.requireNonNull(label);`, `Objects.requireNonNull(usedIds);`, `Objects.requireNonNull(allUsedIds);`

Please add `@paramter allUsedIds` containing a short description of the parameter

src/main/java/net/sf/jabref/model/database/BibDatabase.java
koppor: Please don't pass `null`. Pass an empty set.

src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
koppor: Yeah, therefore I asked for a Set. The Set is independent of the order - or did I get your comment wrong?

src/main/java/net/sf/jabref/logic/auxparser/AuxParserResult.java
```diff
@@ -42,6 +45,10 @@ public int getUnresolvedKeysCount() {
         return unresolvedKeys.size();
     }
 
+    public int getInsertedStringsCount() {
+        return insertedStrings;
```
koppor: Is it possible to add a test case covering that?

src/main/java/net/sf/jabref/model/database/BibDatabase.java
tobiasdiez: What is the difference between `usedIds` and `allUsedIds`? They seem to have the same purpose.

src/main/java/net/sf/jabref/model/database/BibDatabase.java
```diff
@@ -342,7 +342,33 @@ public synchronized boolean hasStringLabel(String label) {
...
+            }
+        }
+
+        // Preamble
+        if (preamble != null) {
+            resolveContent(preamble, new HashSet<>(), allUsedIds);
+        }
+
+        for (String stringId : allUsedIds) {
+            result.add((BibtexString) bibtexStrings.get(stringId).clone());
```
tobiasdiez: why do you use clone here?

src/main/java/net/sf/jabref/model/database/BibDatabase.java
```diff
@@ -413,11 +439,14 @@ private String resolveString(String label, Set<String> usedIds) {
                 }
                 // If not, log this string's ID now.
                 usedIds.add(string.getId());
```
tobiasdiez: Here you get the id of the string and above in getUsedStrings you get the string from the id.
Thus I would propose to change `Set<String> usedIds` to `Set<BibString> usedStrings`.

src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
```diff
@@ -317,4 +320,52 @@ public void resolveForStringsOddHashMarkAtTheEnd() {
         database.addString(string);
         assertEquals(database.resolveForStrings("AAA#AAA#AAA#"), "AAAaaaAAA#");
     }
+
+    @Test
+    public void getUsedStringsSingleStringWithString() {
+        BibEntry entry = new BibEntry(IdGenerator.next());
+        entry.setField("author", "#AAA#");
+        BibtexString string = new BibtexString(IdGenerator.next(), "AAA", "Some other #BBB#");
```
tobiasdiez: Please move the string initalization and adding them to the db to the @setup method.

src/test/java/net/sf/jabref/model/database/BibDatabaseTest.java
```diff
@@ -317,4 +320,52 @@ public void resolveForStringsOddHashMarkAtTheEnd() {
...
+        BibtexString string2 = new BibtexString(IdGenerator.next(), "BBB", "Some more text");
+        BibtexString string3 = new BibtexString(IdGenerator.next(), "CCC", "Even more text");
+        database.addString(string);
+        database.addString(string2);
+        database.addString(string3);
+        database.insertEntry(entry);
+        List<BibtexString> usedStrings = (List<BibtexString>) database.getUsedStrings(Arrays.asList(entry));
```
tobiasdiez: create overload of getUsedStrings for one entry and use Collections.singletonList.

lenhard:
As of 1684cdccc205e4e836b2379d3c97b7e14bcc5186 this is integrated into master now. I had to do it on a separate branch since I have no commit access to the branch here. However, all commits are preserved. 

I addressed the issues mentioned above that I agreed with, which were all rather minor quality things.

So this branch can be closed now.

Issue #1669, opened 2016-08-03
JabRef 3.5 cannot connect ot LibreOffice on ubuntu


pgarciabaez:
I have problems to connect JabRef and LibreOffice. I use:

* JabRef 3.5 (same problem with JabRef-3.6dev--snapshot--2016-08-03--master--f77daac.jar)
* LibreOffice 4.2.6.3
* Java 1.8.0_20
* Ubuntu 14.04.1
Here you can see the error output of my console:

Manual connection:
````
Exception in thread "AWT-EventQueue-0" java.lang.IndexOutOfBoundsException: The row index 3 must be less than or equal to 1.
	at com.jgoodies.forms.layout.CellConstraints.ensureValidGridBounds(CellConstraints.java:967)
	at com.jgoodies.forms.layout.FormLayout.setConstraints(FormLayout.java:805)
	at com.jgoodies.forms.layout.FormLayout.addLayoutComponent(FormLayout.java:1165)
	at java.awt.Container.addImpl(Container.java:1125)
	at java.awt.Container.add(Container.java:971)
	at com.jgoodies.forms.builder.FormBuilder.addImpl(FormBuilder.java:1896)
	at com.jgoodies.forms.builder.FormBuilder$ComponentAdder.add(FormBuilder.java:2550)
	at com.jgoodies.forms.builder.FormBuilder$ComponentAdder.at(FormBuilder.java:2140)
	at com.jgoodies.forms.builder.FormBuilder$ComponentAdder.xy(FormBuilder.java:2160)
	at net.sf.jabref.gui.openoffice.OpenOfficePanel.showConnectDialog(OpenOfficePanel.java:502)
	at net.sf.jabref.gui.openoffice.OpenOfficePanel.connect(OpenOfficePanel.java:357)
	at net.sf.jabref.gui.openoffice.OpenOfficePanel.lambda$initPanel$2(OpenOfficePanel.java:164)
	at net.sf.jabref.gui.openoffice.OpenOfficePanel$$Lambda$61/1488891022.actionPerformed(Unknown Source)
	at javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:2022)
	at javax.swing.AbstractButton$Handler.actionPerformed(AbstractButton.java:2346)
	at javax.swing.DefaultButtonModel.fireActionPerformed(DefaultButtonModel.java:402)
	at javax.swing.DefaultButtonModel.setPressed(DefaultButtonModel.java:259)
[...]

PR #1678, opened 2016-08-04
Fixed OO/LO manual connection dialog on Linux

oscargus:
Doesn't fix the whole issue #1669, but at least the bug in the UI code so that the manual connection dialog can be drawn on Linux.


matthiasgeiger:
:+1:

Issue #1682, opened 2016-08-05
ooplugin: it shouldn't be permitted to cite a publication that doesn't yet have a bibtex key


jdpipe:
When using Jabref with LibreOffice, I can insert a reference to an entry before assign a bibtexkey to the entry. When I do so, I get "[??]" in my document. If I then click the refresh (circular arrow) icon in the ooplugin sidebar, then I get a popup message in Jabref saying "your OpenOffice/LibreOffice document references the BibTexKey 'null', which could not be found in the current database".

I think it would be more helpful if such citations were barred from being inserted in the first place, since that is the best time for the user to fix the problem. Later they can only wonder pointlessly what the [??] was referring to.

Better still, a BibTexKey could auto-generated (and again every following time, if the user desired it)

"You are citing a publication with an empty BibTexKey. A key will automatically generated [OK] [Cancel]  [x] always do this in future"

Version:
JabRef 3.6dev--snapshot--2016-07-29--extractoobibdatabase--fba22d9
linux 4.4.0-31-generic amd64 
Java 1.8.0_91

oscargus:
When working on another (code) enhancement related to the keys (#1676), I realized that in quite a few cases JabRef simply assumed that the key is set. This is clearly one of the cases causing actual issues. I'll have a look at it.

oscargus:
You can try a test build in five minutes at http://builds.jabref.org/fix1682 (I'm usually not this quick, but now I was looking for a small thing to fix and this fit well.)

oscargus:
Although the check box is a really good idea, it was a bit too much of a hassle to fix it right now.

jdpipe:
I tried it, worked fine. Thanks, Oscar!

oscargus:
Fixed in #1683

PR #1683, opened 2016-08-05
Do not cite entries without a key in OpenOffice/LibreOffice (#1682)

oscargus:
Check that all entries to be cited have non-empty keys, show a popup if there are empty keys, and generate keys if instructed to do so.

References #1682 


src/main/java/net/sf/jabref/gui/openoffice/OpenOfficePanel.java
```diff
@@ -669,6 +674,57 @@ private void pushEntries(boolean inParenthesisIn, boolean withText, boolean addP
...
+
+        // If no empty keys, return true
+        if (!emptyKeys) {
+            return true;
+        }
+
+        // Ask if keys should be generated
+        int answer = JOptionPane.showConfirmDialog(this.frame,
+                Localization.lang("Cannot cite entries without BibTeX keys. Generate keys now?"),
+                Localization.lang("Cite"), JOptionPane.YES_NO_OPTION);
```
Siedlerchr: Dialogs asking a question and then having yes/no  actions are generally not a good idea, because the "thinking overhead" for the user in the moment is too much.  http://ux.stackexchange.com/questions/756/what-are-some-alternatives-to-the-phrase-are-you-sure-you-want-to-xyz-in-confi

Better add a button like  "Generate keys" and a Cancel button.

oscargus:
capture14

stefan-kolb:
Please do not use "Implemented #1682" as title. One cannot follow the link to the root issue :smile: A small description makes it much easier.

stefan-kolb:
Otherwise, please rebase and merge :+1:

Issue #1681, opened 2016-08-05
Latex code in groups


grimes2:
JabRef version 3.5 on Windows 10

Steps to reproduce:
1. (Main menu) Groups
2. Toggle groups interface
3. Automatically create groups for database (Magic wand symbol)
4. Generate groups for author last names
5. (Some) Group entries are created in Latex code (example: B{"{o}}hmer)
6. Selecting entries fails (no result)

oscargus:
I tried this now and to me it works (current development version, but I do not see that anything should have been fixed since 3.5). I'm wondering if it related to how the names are written? To me, the following works: `Palom\"aki`, `Curtic\u{a}pean`, `P\'{a}pay`, `K\"{a}llstr\"{o}m`. After doing Latex -> Unicode -> Latex it also works for `Curtic{\u{a}}pean`, etc.

Do you have the number of entries showing? Are there any entries in the group you selected? (Groups settings button.)

grimes2:
Show number of elements is off. Toggling Show number of elements and then automatically generate groups generates also Latex code groups, but these Latex code groups are now filled with elements. Strange!

oscargus:
OK, I think I found the reason. The groups are not loaded correctly. So while the generate group contains ` B{\"{o}}hmer`, the loaded group contains `B{"{o}}hmer` (despite being saved as `B{\\\\"{o}}hmer`). I mark this as a bug and hope that someone with more insight can sort it out.

koppor:
Thank you for reporting this issue. We think, that is already fixed in our development version and consequently the change will be included in the next release.

We would like to ask you to use a development build from https://builds.jabref.org/master and report back if it works for you.

PR #1684, opened 2016-08-05
Group names can now be rendered using LaTeXs commands

oscargus:
Inspired by #1681 

### DIFF
--- a/src/main/java/net/sf/jabref/gui/groups/AutoGroupDialog.java
+++ b/src/main/java/net/sf/jabref/gui/groups/AutoGroupDialog.java
@@ -49,6 +49,7 @@ import net.sf.jabref.logic.groups.GroupTreeNode;
 import net.sf.jabref.logic.groups.GroupsUtil;
 import net.sf.jabref.logic.groups.KeywordGroup;
 import net.sf.jabref.logic.l10n.Localization;
+import net.sf.jabref.logic.layout.format.LatexToUnicodeFormatter;
 import net.sf.jabref.model.entry.FieldName;
 
 import com.jgoodies.forms.builder.ButtonBarBuilder;
@@ -72,6 +73,8 @@ class AutoGroupDialog extends JDialog implements CaretListener {
     private final GroupTreeNodeViewModel m_groupsRoot;
     private final JabRefFrame frame;
     private final BasePanel panel;
+
+    private static final LatexToUnicodeFormatter FORMATTER = new LatexToUnicodeFormatter();
 
 
     /**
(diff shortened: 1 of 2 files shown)


tobiasdiez:
Mhh, I think this is the wrong place to add this functionality. If I name my group `\relax` (or `\beta`) then I probably want to keep this name.
But I think it is a good idea to add the Latex-to-Unicode-formatter when groups are automatically created.

oscargus:
Really!? If you name a group \beta, in a program dealing with LaTeX, would
you be upset if it shows the beta properly? (The \relax use case is, ehh,
far  fetched?)

stefan-kolb:
Do we really need to allow latex code for group names? Why is normal text not enough? I'd rather not have this option and keep it simple.

oscargus:
See #1681. Names containing accents is the main reason. Already now, the names can contain LaTeX code (except that the reading is flawed, again see #1681). It's just that they are properly rendered now. I  do not really see the problem.

tobiasdiez:
For me the problem is that the Latex -> Unicode converter is not good enough to handle arbitrary text and I think we may run into problems where names of groups are not properly rendered (i.e. "A or B" in the form "A\B", or similar things with backslash). As a user, I kind of expect that JabRef takes what I give it as a name (especially since I can provide unicode strings, I don't see the reason to support latex accents). 

The idea is nonetheless good. Just put the conversion code in the automatic generation of groups (since there the user has no control over the names of the groups).

oscargus:
I see your point, although I'm not fully agreeing with it (I'd rather write $\Delta\Sigma$ than look up the Unicode key and writing \ in anything related to LaTeX expecting it to stay a backslash seems questionable... ;-)), but you solution is fair enough and leads to more or less the same end result. Only question is if BibTeX is bothered by Unicode characters in fields that are ignored anyway? (Yes, there are good reasons to use BibTeX for many people.)

Could you have a look at why the backslashes are removed when reading back the group definitions? (Since you have better insight into the group package.)

oscargus:
I changed it to do the conversion upon generation of automatic groups (and correspondingly changed the CHANGELOG entry).

src/main/java/net/sf/jabref/gui/groups/AutoGroupDialog.java
```diff
@@ -73,6 +74,8 @@
     private final JabRefFrame frame;
     private final BasePanel panel;
 
+    private static final LatexToUnicodeFormatter FORMATTER = new LatexToUnicodeFormatter();
```
tobiasdiez: I would make the formatter a normal variable (initialized directly before the for loop below) instead of static final.

tobiasdiez:
Ok, just make the fomatter a normal variable and then this can be merged. I will have a look at the serialization / parsing problem as soon as I'm back in Germany (i.e. in a month)

src/main/java/net/sf/jabref/gui/groups/AutoGroupDialog.java
oscargus: I had it like that first. ;-)

PR #1685, opened 2016-08-06
Added DiVA fetcher based on the new interface

oscargus:

### DIFF
--- a/src/main/java/net/sf/jabref/logic/importer/fetcher/DiVA.java
+++ b/src/main/java/net/sf/jabref/logic/importer/fetcher/DiVA.java
@@ -1,0 +1,80 @@
+package net.sf.jabref.logic.importer.fetcher;
+
+import java.io.IOException;
+import java.io.UnsupportedEncodingException;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
+import java.util.Optional;
+
+import net.sf.jabref.importer.fileformat.BibtexParser;
+import net.sf.jabref.logic.formatter.bibtexfields.UnicodeToLatexFormatter;
+import net.sf.jabref.logic.help.HelpFile;
+import net.sf.jabref.logic.importer.FetcherException;
+import net.sf.jabref.logic.importer.IdBasedFetcher;
+import net.sf.jabref.logic.net.URLDownload;
+import net.sf.jabref.model.entry.BibEntry;
+import net.sf.jabref.model.entry.FieldName;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+public class DiVA implements IdBasedFetcher {
+
+    private static final Log LOGGER = LogFactory.getLog(DiVA.class);
+
+    private static final String URL_PATTERN = "http://www.diva-portal.org/smash/getreferences?referenceFormat=BibTex&pids=%s";
+
+
+    @Override
+    public String getName() {
+        return "DiVA";
+    }
+
+    @Override
+    public HelpFile getHelpPage() {
+        return HelpFile.FETCHER_DIVA_TO_BIBTEX;
+    }
+
+    @Override
+    public Optional<BibEntry> performSearchById(String identifier) throws FetcherException {
+        String q;
+        try {
+            q = URLEncoder.encode(identifier, StandardCharsets.UTF_8.name());
+        } catch (UnsupportedEncodingException e) {
+            // this should never happen
+            LOGGER.warn("Encoding issues", e);
+            return Optional.empty();
+        }
+
+        String urlString = String.format(DiVA.URL_PATTERN, q);
+
+        // Send the request
+        URL url;
+        try {
+            url = new URL(urlString);
+        } catch (MalformedURLException e) {
+            LOGGER.warn("Bad URL", e);
+            return Optional.empty();
+        }
+
+        String bibtexString;
+        try {
+            URLDownload dl = new URLDownload(url);
+
+            bibtexString = dl.downloadToString(StandardCharsets.UTF_8);
+        } catch (IOException e) {
+            throw new FetcherException("Communication problems", e);
+        }
+
+        Optional<BibEntry> result = Optional.ofNullable(BibtexParser.singleFromString(bibtexString));
+
+        result.ifPresent(entry -> entry.getFieldOptional(FieldName.INSTITUTION).ifPresent(institution -> entry
+                .setField(FieldName.INSTITUTION, new UnicodeToLatexFormatter().format(institution))));
+
+        return result;
+
+    }
+
+}
--- a/src/main/java/net/sf/jabref/logic/importer/fetcher/DiVA.java
+++ b/src/main/java/net/sf/jabref/logic/importer/fetcher/DiVA.java
@@ -1,30 +1,26 @@
 package net.sf.jabref.logic.importer.fetcher;
 
 import java.io.IOException;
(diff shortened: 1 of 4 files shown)


src/main/java/net/sf/jabref/logic/importer/fetcher/DiVA.java
```diff
@@ -0,0 +1,80 @@
...
+        return HelpFile.FETCHER_DIVA_TO_BIBTEX;
+    }
+
+    @Override
+    public Optional<BibEntry> performSearchById(String identifier) throws FetcherException {
+        String q;
+        try {
+            q = URLEncoder.encode(identifier, StandardCharsets.UTF_8.name());
+        } catch (UnsupportedEncodingException e) {
+            // this should never happen
```
tobiasdiez: Just throw a FetcherException as you do below.

src/main/java/net/sf/jabref/logic/importer/fetcher/DiVA.java
```diff
@@ -0,0 +1,80 @@
...
+import net.sf.jabref.logic.importer.FetcherException;
+import net.sf.jabref.logic.importer.IdBasedFetcher;
+import net.sf.jabref.logic.net.URLDownload;
+import net.sf.jabref.model.entry.BibEntry;
+import net.sf.jabref.model.entry.FieldName;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+public class DiVA implements IdBasedFetcher {
```
tobiasdiez: Can shortly explain what DiVA is in a comment and add a link to its API.

src/main/java/net/sf/jabref/logic/importer/fetcher/DiVA.java
```diff
@@ -0,0 +1,80 @@
...
+        String q;
+        try {
+            q = URLEncoder.encode(identifier, StandardCharsets.UTF_8.name());
+        } catch (UnsupportedEncodingException e) {
+            // this should never happen
+            LOGGER.warn("Encoding issues", e);
+            return Optional.empty();
+        }
+
+        String urlString = String.format(DiVA.URL_PATTERN, q);
```
tobiasdiez: Use Apache`s URIBuilder to create the URL.

src/main/java/net/sf/jabref/logic/importer/fetcher/DiVA.java
```diff
@@ -0,0 +1,80 @@
...
+    @Override
+    public String getName() {
+        return "DiVA";
+    }
+
+    @Override
+    public HelpFile getHelpPage() {
+        return HelpFile.FETCHER_DIVA_TO_BIBTEX;
+    }
+
```
tobiasdiez: Could you please also add a method `isValidId(String identifier)` which checks that a given ID is of the expected format. I plan to add some functionality which depends on such a method but have no insights what DiVA identifiers look like. Thanks

src/main/java/net/sf/jabref/logic/importer/fetcher/DiVA.java
```diff
@@ -0,0 +1,80 @@
...
+            URLDownload dl = new URLDownload(url);
+
+            bibtexString = dl.downloadToString(StandardCharsets.UTF_8);
+        } catch (IOException e) {
+            throw new FetcherException("Communication problems", e);
+        }
+
+        Optional<BibEntry> result = Optional.ofNullable(BibtexParser.singleFromString(bibtexString));
+
+        result.ifPresent(entry -> entry.getFieldOptional(FieldName.INSTITUTION).ifPresent(institution -> entry
```
tobiasdiez: I would leave it to the users cleanup actions to format unicode to latex.

src/test/java/net/sf/jabref/logic/importer/fetcher/DiVATest.java
```diff
@@ -0,0 +1,56 @@
...
+
+
+public class DiVATest {
+
+    private DiVA fetcher;
+
+
+    @Before
+    public void setUp() {
+        Globals.prefs = JabRefPreferences.getInstance();
```
tobiasdiez: Prefs are not required, right?

src/main/java/net/sf/jabref/logic/importer/fetcher/DiVA.java
tobiasdiez: I suggest to add this fetcher to the list of web searches and also remove the old one (or is there reason to keep the old code?)

src/test/java/net/sf/jabref/logic/importer/fetcher/DiVATest.java
oscargus: They are required to parse the BibTeX entry (wrapped fields).

src/main/java/net/sf/jabref/logic/importer/fetcher/DiVA.java
oscargus: As there is no easy way to wrap an IdBased searcher at the moment, it is
not possible. There was a class in the ISBN stupro PR so once that is merge
it makes sense.

src/main/java/net/sf/jabref/logic/importer/fetcher/DiVA.java
oscargus: I have no idea what the requirements are. Most I've seen are on the format `diva2:XXXXX` though, where X is a digit.

tobiasdiez: I tried to play around with the API and it appears that the identifier definitely has to start with `diva2` so just take this as a test for valid identifier.

oscargus:
@JabRef/developers: I rebased this and introduced a change in `MetaData`, making constructors without provided `Charset` deprecated. This is an attempt to get rid of the `Globals` dependency. Didn't change anything else related to that in this PR, but once merged there are many things to be changed.

src/main/java/net/sf/jabref/logic/importer/fetcher/DiVA.java
```diff
@@ -0,0 +1,69 @@
...
+    @Override
+    public Optional<BibEntry> performSearchById(String identifier) throws FetcherException {
+        Optional<BibEntry> result = Optional.empty();
+        try {
+            URIBuilder uriBuilder = new URIBuilder(URL);
+
+            uriBuilder.addParameter("referenceFormat", "BibTex");
+            uriBuilder.addParameter("pids", identifier);
+
+            String bibtexString;
```
tobiasdiez: Combine this line with line 55.

src/main/java/net/sf/jabref/logic/importer/fetcher/DiVA.java
```diff
@@ -0,0 +1,69 @@
...
+            URLDownload dl = new URLDownload(uriBuilder.build().toURL());
+
+            bibtexString = dl.downloadToString(StandardCharsets.UTF_8);
+            result = Optional.ofNullable(BibtexParser.singleFromString(bibtexString, importFormatPreferences));
```
tobiasdiez: Just `return Optional.of(...)` and remove result variable.

tobiasdiez:
Some small remarks. After fixing these and the codacy issue, this can be merged.

src/main/java/net/sf/jabref/logic/importer/fetcher/DiVA.java
oscargus: Yeah, weird code... :-)

PR #1693, opened 2016-08-08
Removed (false) NPE issue reported by FindBugs

oscargus:
FindBugs reported a potential NPE here, which never occurred since it would only happen if `model.getSearchState() == MainTableDataModel.DisplayOption.DISABLED`. As this is checked for in the first part of the if-clause, I removed the second check., so this shouldn't affect any perfomance since `matches` will only be triggered under the same conditions as before.

Nothing in CHANGELOG since the NPE couldn't actually happen.


matthiasgeiger:
LGTM

PR #1695, opened 2016-08-08
Injected LayoutFormatterPreferences in ExportFormat

oscargus:
And removed the need for the static variable `customExportNameFormatters` in `JabRefPreferences`.

~~Still `SavePreferences` to inject and remove the reference for resource loading to~~ completely get rid of `Globals`.

### DIFF
--- a/src/main/java/net/sf/jabref/gui/exporter/ExportAction.java
+++ b/src/main/java/net/sf/jabref/gui/exporter/ExportAction.java
@@ -19,6 +19,7 @@ import net.sf.jabref.gui.actions.MnemonicAwareAction;
 import net.sf.jabref.gui.worker.AbstractWorker;
 import net.sf.jabref.logic.exporter.ExportFormats;
 import net.sf.jabref.logic.exporter.IExportFormat;
+import net.sf.jabref.logic.exporter.SavePreferences;
 import net.sf.jabref.logic.l10n.Localization;
 import net.sf.jabref.logic.layout.LayoutFormatterPreferences;
 import net.sf.jabref.model.entry.BibEntry;
@@ -62,7 +63,8 @@ public class ExportAction {
             public void actionPerformed(ActionEvent e) {
                 ExportFormats.initAllExports(Globals.prefs.customExports.getCustomExportFormats(Globals.prefs,
                         Globals.journalAbbreviationLoader),
-                        LayoutFormatterPreferences.fromPreferences(Globals.prefs, Globals.journalAbbreviationLoader));
+                        LayoutFormatterPreferences.fromPreferences(Globals.prefs, Globals.journalAbbreviationLoader),
+                        SavePreferences.loadForExportFromPreferences(Globals.prefs));
                 JFileChooser fc = ExportAction
                         .createExportFileChooser(Globals.prefs.get(JabRefPreferences.EXPORT_WORKING_DIRECTORY));
                 fc.showSaveDialog(frame);
--- a/src/main/java/net/sf/jabref/gui/exporter/ExportAction.java
+++ b/src/main/java/net/sf/jabref/gui/exporter/ExportAction.java
@@ -17,6 +17,7 @@ import net.sf.jabref.Globals;
 import net.sf.jabref.gui.JabRefFrame;
 import net.sf.jabref.gui.actions.MnemonicAwareAction;
 import net.sf.jabref.gui.worker.AbstractWorker;
+import net.sf.jabref.logic.exporter.ExportFormat;
 import net.sf.jabref.logic.exporter.ExportFormats;
 import net.sf.jabref.logic.exporter.IExportFormat;
 import net.sf.jabref.logic.exporter.SavePreferences;
@@ -61,10 +62,11 @@ public class ExportAction {
 
             @Override
             public void actionPerformed(ActionEvent e) {
-                ExportFormats.initAllExports(Globals.prefs.customExports.getCustomExportFormats(Globals.prefs,
-                        Globals.journalAbbreviationLoader),
-                        LayoutFormatterPreferences.fromPreferences(Globals.prefs, Globals.journalAbbreviationLoader),
-                        SavePreferences.loadForExportFromPreferences(Globals.prefs));
+                Map<String, ExportFormat> customFormats = Globals.prefs.customExports.getCustomExportFormats(Globals.prefs,
+                        Globals.journalAbbreviationLoader);
+                LayoutFormatterPreferences layoutPreferences = LayoutFormatterPreferences.fromPreferences(Globals.prefs, Globals.journalAbbreviationLoader);
+                SavePreferences savePreferences = SavePreferences.loadForExportFromPreferences(Globals.prefs);
+                ExportFormats.initAllExports(customFormats, layoutPreferences, savePreferences);
                 JFileChooser fc = ExportAction
                         .createExportFileChooser(Globals.prefs.get(JabRefPreferences.EXPORT_WORKING_DIRECTORY));
                 fc.showSaveDialog(frame);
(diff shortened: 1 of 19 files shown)


oscargus:
Messed up different versions, so will check on my other computer...

oscargus:
Found the most recent version, so good to check now.

src/main/java/net/sf/jabref/gui/exporter/ExportAction.java
```diff
@@ -59,7 +61,10 @@ public InternalExportAction(JabRefFrame frame, boolean selectedOnly) {
 
             @Override
             public void actionPerformed(ActionEvent e) {
-                ExportFormats.initAllExports(Globals.prefs.customExports.getCustomExportFormats(Globals.prefs));
+                ExportFormats.initAllExports(Globals.prefs.customExports.getCustomExportFormats(Globals.prefs,
```
oscargus: This call looks really weird, but one need to check what is actually done and then there is nothing strange. Having three temporary variables will "hide" it, but not worthwhile IMHO.

koppor: Maybe introduce the temporary variables nevertheless :innocent:

src/main/java/net/sf/jabref/logic/exporter/CustomExportList.java
```diff
@@ -93,20 +99,22 @@ private void readPrefs(JabRefPreferences prefs) {
...
+    public void addFormat(List<String> s, LayoutFormatterPreferences layoutPreferences, SavePreferences savePreferences) {
+        createFormat(s, layoutPreferences, savePreferences).ifPresent(format -> {
             formats.put(format.getConsoleName(), format);
             list.add(s);
         });
     }
 
-    public void remove(List<String> toRemove) {
-        createFormat(toRemove).ifPresent(format -> {
+    public void remove(List<String> toRemove, LayoutFormatterPreferences layoutPreferences,
```
koppor: Is it really necessry to pass layoutPreferences and savePreferences? Might there be two custom exporters with the same names, but different preferences? Does that make sense?

src/main/java/net/sf/jabref/logic/exporter/ExportFormat.java
```diff
@@ -78,6 +79,24 @@ public ExportFormat(String displayName, String consoleName, String lfFileName, S
...
+     * Initialize another export format based on templates stored in dir with
+     * layoutFile lfFilename.
+     *
+     * @param displayName Name to display to the user.
+     * @param consoleName Name to call this format in the console.
+     * @param lfFileName  Name of the main layout file.
+     * @param directory   Directory in which to find the layout file.
+     * @param extension   Should contain the . (for instance .txt).
+     * @param layoutPreferences Preferences for layout
+     */
```
koppor: savePreferences should be described and stated for what they are used in the formatter.

src/main/java/net/sf/jabref/logic/exporter/ExportFormat.java
```diff
@@ -78,6 +79,24 @@ public ExportFormat(String displayName, String consoleName, String lfFileName, S
...
+            LayoutFormatterPreferences layoutPreferences, SavePreferences savePreferences) {
+        this(displayName, consoleName, lfFileName, directory, extension);
+        this.layoutPreferences = layoutPreferences;
+        this.savePreferences = savePreferences;
+    }
+
+    /**
      * Empty default constructor for subclasses
      */
     ExportFormat() {
```
koppor: `protected`?

src/main/java/net/sf/jabref/logic/exporter/ExportFormat.java
```diff
@@ -310,7 +324,7 @@ public void performExport(final BibDatabaseContext databaseContext, final String
             }
 
             // Clear custom name formatters:
-            Globals.prefs.customExportNameFormatters = null;
+            layoutPreferences.getCustomExportNameFormatters().clear();
```
koppor: There should be a method offered at layoutPreferences to clear the list

Typically, a copy of the internal data structure should be returned and thus `clear()` has no effect. layoutPreferences.getCustomExportNameFormatters should be wrapped (there!) in [Collections.unmodifiableList](http://docs.oracle.com/javase/7/docs/api/java/util/Collections.html#unmodifiableList%28java.util.List%29).

src/main/java/net/sf/jabref/logic/exporter/ExportFormat.java
```diff
@@ -357,7 +369,7 @@ public void performExport(final BibDatabaseContext databaseContext, Path file, f
                     if ((index > 0) && ((index + 1) < line.length())) {
                         String formatterName = line.substring(0, index);
                         String contents = line.substring(index + 1);
-                        formatters.put(formatterName, contents);
+                        layoutPreferences.getCustomExportNameFormatters().put(formatterName, contents);
```
koppor: See above - if possible.

koppor:
Besides my minor comments: LGTM

src/main/java/net/sf/jabref/logic/exporter/CustomExportList.java
tobiasdiez: I would include the Layout preferences as a part of the Save preference class.

koppor: I think, the granularity is fine here. :innocent:

src/main/java/net/sf/jabref/logic/exporter/ExportFormat.java
oscargus: This list is local to the preferences object, so it should work. Earlier on it was a static variable in JabRefPreferences, but now it is created with the LayoutFormatPreferences object. I still think it should be cleared since the same LayoutFormatPreferences object may be passed to several classes at different times.

src/main/java/net/sf/jabref/logic/exporter/ExportFormat.java
oscargus: Yes.

src/main/java/net/sf/jabref/logic/exporter/CustomExportList.java
oscargus: There are indeed many injections of preferences in different ways. I'm not really sure that this is the best approach (or rather sure that there is probably a better). In addition, the only instance of CustomExportList exists in JabRefPreferences... However, at the moment I was more focused on getting rid of Globals.prefs. Regarding granularity, it is always tricky.

matthiasgeiger:
Uhhh... I can't run JabRef at the moment:

```
$ java -jar JabRef-3.6dev--snapshot--2016-08-18--master--0c16669.jar
Exception in thread "AWT-EventQueue-0" java.lang.NullPointerException
        at java.util.Objects.requireNonNull(Unknown Source)
        at net.sf.jabref.logic.layout.LayoutFormatterPreferences.fromPreferenc
(LayoutFormatterPreferences.java:33)
        at net.sf.jabref.logic.exporter.CustomExportList.readPrefs(CustomExpor
ist.java:76)
        at net.sf.jabref.logic.exporter.CustomExportList.getCustomExportFormat
CustomExportList.java:59)
        at net.sf.jabref.JabRefMain.start(JabRefMain.java:85)
        at net.sf.jabref.JabRefMain.lambda$main$0(JabRefMain.java:56)
        at net.sf.jabref.JabRefMain$$Lambda$1/1557989809.run(Unknown Source)
        at java.awt.event.InvocationEvent.dispatch(Unknown Source)
        at java.awt.EventQueue.dispatchEventImpl(Unknown Source)
        at java.awt.EventQueue.access$500(Unknown Source)
        at java.awt.EventQueue$3.run(Unknown Source)
        at java.awt.EventQueue$3.run(Unknown Source)
        at java.security.AccessController.doPrivileged(Native Method)
        at java.security.ProtectionDomain$1.doIntersectionPrivilege(Unknown So
ce)
        at java.awt.EventQueue.dispatchEvent(Unknown Source)
        at java.awt.EventDispatchThread.pumpOneEventForFilters(Unknown Source)
        at java.awt.EventDispatchThread.pumpEventsForFilter(Unknown Source)
        at java.awt.EventDispatchThread.pumpEventsForHierarchy(Unknown Source)
        at java.awt.EventDispatchThread.pumpEvents(Unknown Source)
        at java.awt.EventDispatchThread.pumpEvents(Unknown Source)
        at java.awt.EventDispatchThread.run(Unknown Source)```

Related to this PR?

oscargus:
Yes, I do not really understand exactly why it worked earlier, but I fixed
it now.

PR #1697, opened 2016-08-08
Removed Globals dependency for InternalBibtexFields

oscargus:

### DIFF
--- a/src/main/java/net/sf/jabref/model/entry/InternalBibtexFields.java
+++ b/src/main/java/net/sf/jabref/model/entry/InternalBibtexFields.java
@@ -42,8 +42,6 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 
-import net.sf.jabref.specialfields.SpecialFieldsUtils;
-
 public class InternalBibtexFields {
 
     // contains all bibtex-field objects (BibtexSingleField)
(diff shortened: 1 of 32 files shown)


src/main/java/net/sf/jabref/model/entry/InternalBibtexFields.java
```diff
@@ -305,17 +273,29 @@ private InternalBibtexFields(boolean serializeSpecialFields) {
...
-                // or export the complete BibtexSingleField ?
-                // BibtexSingleField.toString() { return fieldname ; }
-            }
-        }
+    }
 
-        // sort the entries
-        Collections.sort(PUBLIC_FIELDS);
+
+    public static void updateTimeStampField(String timeStampFieldName) {
```
tobiasdiez: I would propose to switch to a "normal class" without using the singleton pattern. The constructor takes some preferences as input and then initializes the fields accordingly. In this way you don't need these static update methods.

src/main/java/net/sf/jabref/model/entry/SpecialFields.java
```diff
@@ -0,0 +1,43 @@
...
+public class SpecialFields {
+
+    public static final String FIELDNAME_PRINTED = "printed";
+    public static final String FIELDNAME_PRIORITY = "priority";
+    public static final String FIELDNAME_QUALITY = "qualityassured";
+    public static final String FIELDNAME_RANKING = "ranking";
+    public static final String FIELDNAME_READ = "readstatus";
+    public static final String FIELDNAME_RELEVANCE = "relevance";
+
+    public static final String PREF_SPECIALFIELDSENABLED = "specialFieldsEnabled";
```
tobiasdiez: If these are values for preference keys, then these should be stored inside the JabRefPreference class

oscargus: That seems to be an arguable point in general.

src/main/java/net/sf/jabref/model/entry/InternalBibtexFields.java
oscargus: I think we need a thorough discussion on statics vs Globals vs injection. There are 25 imports of InternalBibtexFields in the current source (excluding model.entry), so we would have to inject in, say, half of them if we go for the Globals vs injection, plus passing through.

I do not really see it as work for this PR, although I do not deem it unlikely that it will happen eventually. (The problem is that as soon as someone starts touching these "bad" designs, one is expected to go all the way and make a perfect redesign, which rarely happens, better to iteratively improve it. With this PR it is most likely slightly simpler to modify it later and we can enforce more architecture constraints.)

src/main/java/net/sf/jabref/model/entry/SpecialFields.java
tobiasdiez: Yes and I argued a lot with Simon a while ago about this since I was the opinion, that all the smaller preference classes should contain their keys and default values. In the end Simon convinced me that for now it is better if everything stays in one place and we will later reconsider moving it to different places.

src/main/java/net/sf/jabref/model/entry/InternalBibtexFields.java
tobiasdiez: Yes you are right, your PR goes into the right direction. I just saw a new "static" method and thought that it was a bad idea (in the end it is, but the best possible with the current design). Sorry for that.

src/main/java/net/sf/jabref/model/entry/SpecialFields.java
oscargus: OK, fair enough.

src/main/java/net/sf/jabref/model/entry/InternalBibtexFields.java
oscargus: No problems. I've just noted this trend and there is a risk that people get discouraged touching the really troublesome code.

What we should try to do though, since clearly most of the things pointed out are valid concerns, is to try to write them down somewhere, like https://github.com/koppor/jabref/issues for those rainy days... But I can admit that I'm not very good at it, even though I try from time to time.

PR #1699, opened 2016-08-08
Removed bst from architecture tests and prepared for Globals

oscargus:
No need for testing bst since it moved to logic.


koppor:
Formatting issues with Eclipse - I will change Eclipse config.

@stefan-kolb will remove the lines

We didn't like to do premature optimization as we don't like disabled code. We should enable the check as soon as Globals is really independent of Logic.

Issue #1257, opened 2016-04-20
Upgrade from 3.2 to 3.3: BibtexKeyGenerator preferences are lost


jkulesza:
JabRef version 3.3 on OSX 10.11.4

Steps to reproduce:

1. Install 3.3 on OSX, selecting to replace 3.2.
2. After installation, launch Preferences, and note that there are differences (most notably in hand-specified BibTeX key generator formats.
3. Attempt to reimport old preferences file.  Choose preferences file, click OK.  A dialog shows "You must restart JabRef for this to come into effect."  Click OK. Click OK.  ⌘-Q to Quit fully.
4. Relaunch and note that preferences are not updated.

matthiasgeiger:
Manually changing the preferences and saving does work?

jkulesza:
Yes, manually updating a key generator entry, saving, quitting, and restarting preserves the change.

matthiasgeiger:
I can confirm, this at least partially... The bug only affects the exported/imported "BibTeX key generator formats": Those settings are not loaded automatically and also import of exported 3.2 settings does not work. 

Other settings are imported correctly. 

This is caused by some changes in the casing of the stored information: In JabRef 3.2 the exported XML contains the following:
```xml
            <node name="labelPattern">
              <map>
                <entry key="article" value="[author][year]"/>
              </map>
            </node>
```
In JabRef 3.3 this is slightly changed: 
```xml
            <node name="labelpattern">
              <map>
                <entry key="article" value="[author][year]"/>
              </map>
            </node>
```
--> `labelPattern` vs. `labelpattern` (small vs. capital P)

**Quick-Fix for users: Open the exported preferences in a text editor and replace the `labelPattern` with `labelpattern`, save the file and import again in JabRef 3.3.**

simonharrer:
I think we should bring the JabRef preferences and the labelPattern preferences together, and on the way solve this issue as well.

stefan-kolb:
Lots of old artifacts...
```xml
          <node name="labelPattern">
            <map/>
          </node>
          <node name="logic">
            <map/>
            <node name="labelPattern">
              <map/>
            </node>
            <node name="labelpattern">
              <map/>
            </node>
          </node>
```

matthiasgeiger:
Starting from version 3.6 key pattern  preferences will bemigrated upon first start of the new JabRef version.

PR #1704, opened 2016-08-09
Rename labelpattern to bibtexkeypattern and fix pref issue

matthiasgeiger:
As mentioned by @lenhard in https://github.com/JabRef/jabref/pull/1626#issuecomment-238533144 - this is a follow up PR:

- [x] Rename label pattern to bibtex key pattern.
  - [x] rename Help file reference (at least until 3.6 is released)
  - [x] rename prefs key 
- [x] Integrate label pattern preferences tree into global preferences tree
- [x] Possible migration of old key patterns (The first time JabRef is run it is checked whether old patterns are there - if yes, they are migrated. Subsequently migration is skipped.)
- ~~Change in CHANGELOG.md described~~ internal cleanup
- ~~Tests created for changes~~ old tests still pass
- ~~Screenshots added (for bigger UI changes)~~ no UI changes
  - [x] duplicate old Page under new name (at least until 3.6 is released) https://github.com/JabRef/help.jabref.org/issues/45


matthiasgeiger:
@JabRef/developers Two related questions regarding the migration of LabelPatterns: 
- What happens if old prefs after migration? Delete them?
- What happens if old and new prefs are there? Keep the new ones?

koppor:
Keep the old prefs. They are not consuming much space in the registry. If someone starts an old JabRef version, the old patterns should still be there. It might be that users update from <= 3.5 to a far newer version (4.0?) and want to go back to 3.5, because something does not work on 4.0.

Keep the new ones, if old prefs are there.

I foresee that some users will complain that the patterns change when having used an old version and changed the patterns there. But I think, those guys will be seldom :)

stefan-kolb:
Probably just keeping them is the easiest alternative, as long as this does not create any side effects.

matthiasgeiger:
Chosen approach is: The first time JabRef is run it is checked whether old patterns are there - if yes, they are migrated. Subsequently migration is skipped. Old prefs are not deleted.

Apart from the changes in the help.jabref.org repo this is good to go from my side.

src/main/java/net/sf/jabref/migrations/PreferencesMigrations.java
```diff
@@ -79,4 +88,43 @@ public static void upgradeSortOrder() {
         }
     }
 
+    /**
+     * Migrate LabelPattern configuration from version 3.3-3.5 to new BibtexKeyPatterns
```
koppor: Minor comment: Could we also upgrade from <3.3? We still have users coming from JabRef 2.9.2.

koppor:
Besides the minor comment: LGTM.

As far as I can see, import/export of preferences doesn't include the bibtex key patterns, does it?

src/main/java/net/sf/jabref/migrations/PreferencesMigrations.java
```diff
@@ -79,4 +88,43 @@ public static void upgradeSortOrder() {
...
+                // Check whether old defaultLabelPattern is set
+                String oldDefault = mainPrefsNode.get("defaultLabelPattern", null);
+                if(oldDefault!=null) {
+                    prefs.put(JabRefPreferences.DEFAULT_BIBTEX_KEY_PATTERN, oldDefault);
+                    LOGGER.info("Upgraded old default key generator pattern '"+oldDefault+"' to new version.");
+                }
+
+            }
+            if (mainPrefsNode.nodeExists("bibtexkeypatterns")) {
+                return; //Pref node already exists do not migrate from previous version
```
koppor: Could you move the comment above the statement?

src/main/java/net/sf/jabref/migrations/PreferencesMigrations.java
matthiasgeiger: Should be possible. I'll implement the logic that the newest available prefs are migrated to 3.6

matthiasgeiger:
Prefs were already reported correctly, regarding this point I was wrong at the Dev call. But I should check whether importing is now working fine.

stefan-kolb:
LGTM :+1:

matthiasgeiger:
Added further migrations fors 3.0-3.2 and <3.0.

Exporting and importing prefs is working with KeyPatterns.

oscargus:
Does this fix #1257 ? (Except the 3.2 to 3.3 part...) In that case add a ChangeLog entry and remove the wontfix tag. :-)

matthiasgeiger:
Added the changelog entry.

... can be merged as soon as https://github.com/JabRef/help.jabref.org/pull/46 is merged - which fixes the failing test.

PR #1705, opened 2016-08-09
Improve db tests

koppor:
The new database tests cannot be run locally as it requires postgres and MySQL to be installed. Therefore, I reused the idea of the integration tests to separate the tests as `databaseTest`.

Alterantive solution: AssumeTrue (http://eclipsesource.com/blogs/2009/10/07/using-junits-assume-for-faster-tests/) in `@BeforeClass` and check whether the database is there. I opted for the separate grdle task to make it obvious when the database test is done and when not.

<s>Minor thing: I added `AssertUtil.assertDoesNotThrow` (inspired by http://stackoverflow.com/a/37203455/873282) to improve the output of unexpected exceptions.</s>


src/databaseTest/java/net/sf/jabref/shared/DBMSConnectorTest.java
```diff
@@ -26,12 +27,9 @@
     public void testGetNewConnection() throws ClassNotFoundException, SQLException {
         DBMSConnectionProperties properties = TestConnector.getConnectionProperties(dbmsType);
 
-        try {
+        AssertUtil.assertDoesNotThrow(() -> {
```
tobiasdiez: No idea why this is needed. If an exception is thrown than the test fails automatically?!

koppor: ItelliJ shows orange icon when unexpected Exception is thrown, but red icon when `assertDoesNotThrow` is used. I really like this triple state: failed, not failed, something unexpected happened (unexpected in the sense that the test does not forsee the exception). 

The old code did this with try/catch and Assert.fail(). I don't like that, because I don't see the exception clearly. With the code above, the Assertion fails and the exception is shown clearly.

matthiasgeiger: I don't think that this is necessary. Either an exception is provoked by a test - or it is "something unexpected".
Thus: Annotate the expected excpetion for tests which should throw exceptions and do not catch exceptions in the tests...

Issue #1609, opened 2016-07-21
"New file link" opens wrong folder in FileDialog


matthiasgeiger:
JabRef version 3.5 and current 3.6dev on Windows 10

Steps to reproduce:

1. Define a "main file directory" for a DB or globally
2. Try to create a "new file link" by hitting the "+" in next to the file field
3. The file dialog does not open the "main file directory" but its parent folder

[Reported by @bkleine in the discourse forum](http://discourse.jabref.org/t/external-file-links-not-properly-set/76/1)

Siedlerchr:
I will try to work on this in the next days. See #1336

bernhard-kleine:
On a first glance it does not open the intended directory but requires a further click on the "browse" button to open the link. This is a different behaviour as before. Actually I prefer when the link is opened. I have installed: JabRef_windows-x64_3_6dev--snapshot--2016-08-10--master--8342956.exe

Siedlerchr:
I rechecked the code and you are partly right.
1) When you click the "+" sign on an entry, then the open dialog **should** be directly shown (no further click necessary.  (will fix this)  
2.) However, when you do Quality-> Lookup full text document, then the open dialog is only shown after you click on Browse.

Siedlerchr:
Please test again with the version from here:
http://builds.jabref.org/fix1609/

bernhard-kleine:
I tested it with that version. It works. However I experienced repeatingly file save issues: The first time an entry was cut before I was fully saved, interestingly the same entry was completely saved just after that faulty saved one. The next time, there were three entries not completely saved, however the completely entries were just behind. (Detected by editing the file with Notepad++). The error is in the save process and is not detected by Jabref itself, only by running biber on this file.

Siedlerchr:
@bernhard-kleine Thanks for your feedback. Would you please open a new issues for this and provide some steps for reproducing? As many details as possible. That helps us to track the error/faulty save process.

PR #1712, opened 2016-08-10
Fix  auto open dialog on adding new file manually

Siedlerchr:
Fix for #1609

Manually tested
Is just a simple restoring of the orig feature

### DIFF
--- a/src/main/java/net/sf/jabref/gui/FileListEntryEditor.java
+++ b/src/main/java/net/sf/jabref/gui/FileListEntryEditor.java
@@ -41,6 +41,7 @@ import javax.swing.JLabel;
 import javax.swing.JProgressBar;
 import javax.swing.JTextField;
 import javax.swing.SwingConstants;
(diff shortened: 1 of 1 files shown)


src/main/java/net/sf/jabref/gui/FileListEntryEditor.java
```diff
@@ -345,6 +346,7 @@ private void storeSettings(FileListEntry entry) {
 
     public boolean okPressed() {
         return okPressed;
+
```
oscargus: Feel free to remove this line. :-) But should be OK to merge either way.

oscargus:
LGTM! :+1:

src/main/java/net/sf/jabref/gui/FileListEntryEditor.java
Siedlerchr: Will be removed the next time I touch it ;)

PR #1717, opened 2016-08-11
Fixes broken serialization of bibentries containing an @

### LINKED ISSUE #1716: Parsing/Writing of Entry fields containing an @ broken
JabRef version 3.6dev

Steps to reproduce:

1. Use the character "@" in a field of an entry
2. Save the file; Entry looks like this:

```
@Article{asda,
  author = {böser Tit@l},
  title  = {change...},
}
```
3. Change the entry
4. Save again - file is broken:

```
@Article{asda,
  author = {böser Tit

@Article{asda,
  author = {böser Tit@l},
  title  = {change...},
}
```

I don't have any clue why this is happening. But this is clearly a blocker for a 3.6 release. @lenhard You want to waste some of your valuable vacation time? :wink:

matthiasgeiger:
Fixes #1716.

- ~~Screenshots added (for bigger UI changes)~~
- ~~Check documentation status (Issue created for outdated help page at [help.jabref.org](https://github.com/JabRef/help.jabref.org/issues)?)~~

As this is rather crucial it would be great if someone else checks what I have done ;-)

To explain: 
```
-                // get the text before the entry
-                String prolog = parsedSerialization.substring(0, parsedSerialization.lastIndexOf('@'));
``` 
... did not work properly if an `@` is used within an entry.

However, determining from the complete parsed serialization where the entry begins and what are comments before the entry is not so trivial. Thus, I added another attribute in the bibentry `commentsBeforeEntry` in which the serialized form of the comments stored in front of the entry is directly saved during parsing the entry.

### DIFF
--- a/src/main/java/net/sf/jabref/importer/fileformat/BibtexParser.java
+++ b/src/main/java/net/sf/jabref/importer/fileformat/BibtexParser.java
@@ -224,10 +224,14 @@ public class BibtexParser {
          * for the user.
          */
         try {
+            String commentsAndEntryTypeDefinitionString = dumpTextReadSoFarToString();
+
             BibEntry entry = parseEntry(type);
 
+            entry.setParsedSerialization(commentsAndEntryTypeDefinitionString+dumpTextReadSoFarToString());
+            entry.setCommentsBeforeEntry(commentsAndEntryTypeDefinitionString.substring(0,commentsAndEntryTypeDefinitionString.lastIndexOf('@')));
+
             boolean duplicateKey = database.insertEntry(entry);
-            entry.setParsedSerialization(dumpTextReadSoFarToString());
             if (duplicateKey) {
                 parserResult.addDuplicateKey(entry.getCiteKey());
             } else if ((entry.getCiteKey() == null) || entry.getCiteKey().isEmpty()) {
--- a/src/main/java/net/sf/jabref/importer/fileformat/BibtexParser.java
+++ b/src/main/java/net/sf/jabref/importer/fileformat/BibtexParser.java
@@ -224,12 +224,15 @@ public class BibtexParser {
          * for the user.
          */
         try {
-            String commentsAndEntryTypeDefinitionString = dumpTextReadSoFarToString();
+            // collect all comments and the entry type definition in front of the actual entry
+            // this is at least `@Type`
+            String commentsAndEntryTypeDefinition = dumpTextReadSoFarToString();
 
             BibEntry entry = parseEntry(type);
-
-            entry.setParsedSerialization(commentsAndEntryTypeDefinitionString+dumpTextReadSoFarToString());
-            entry.setCommentsBeforeEntry(commentsAndEntryTypeDefinitionString.substring(0,commentsAndEntryTypeDefinitionString.lastIndexOf('@')));
+            // store comments collected without type definition
+            entry.setCommentsBeforeEntry(commentsAndEntryTypeDefinition.substring(0,commentsAndEntryTypeDefinition.lastIndexOf('@')));
+            // store complete parsed serialization (comments, type definition + type contents)
+            entry.setParsedSerialization(commentsAndEntryTypeDefinition+dumpTextReadSoFarToString());
 
             boolean duplicateKey = database.insertEntry(entry);
             if (duplicateKey) {
--- a/src/main/java/net/sf/jabref/model/entry/BibEntry.java
+++ b/src/main/java/net/sf/jabref/model/entry/BibEntry.java
@@ -69,6 +69,8 @@ public class BibEntry implements Cloneable {
 
     private String parsedSerialization;
 
+    private String commentsBeforeEntry = "";
+
     /*
      * Marks whether the complete serialization, which was read from file, should be used.
      *
@@ -573,6 +575,11 @@ public class BibEntry implements Cloneable {
         return parsedSerialization;
     }
 
+    public void setCommentsBeforeEntry(String parsedComments) {
+        this.commentsBeforeEntry = parsedComments;
+    }
+
+
     public boolean hasChanged() {
         return changed;
     }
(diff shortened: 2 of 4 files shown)


src/main/java/net/sf/jabref/model/entry/BibEntry.java
```diff
@@ -573,6 +575,11 @@ public String getParsedSerialization() {
         return parsedSerialization;
     }
 
+    public void setCommentsBeforeEntry(String parsedComments) {
+        this.commentsBeforeEntry = parsedComments;
+    }
+
+
```
stefan-kolb: Please remove this unecessary space.

src/test/java/net/sf/jabref/importer/fileformat/BibtexParserTest.java
```diff
@@ -1631,4 +1631,25 @@ public void parseCommentContainingEntries() throws IOException {
...
+                "  Number                   = {1}" + OS.NEWLINE +
+                "}";
+        // @formatter:on
+
+        ParserResult result = BibtexParser.parse(new StringReader(bibtexEntry));
+        Collection<BibEntry> entries = result.getDatabase().getEntries();
+        BibEntry entry = entries.iterator().next();
+
+        assertEquals(bibtexEntry, entry.getParsedSerialization());
+    }
```
tobiasdiez: Besides checking the parsed serialization, could you please also add another test which checks that an entry containing an at symbol is parsed correctly, i.e. has the correct field values.

tobiasdiez:
LGTM :+1:

src/main/java/net/sf/jabref/importer/fileformat/BibtexParser.java
```diff
@@ -224,10 +224,14 @@ private void parseAndAddEntry(String type) {
          * for the user.
          */
         try {
+            String commentsAndEntryTypeDefinitionString = dumpTextReadSoFarToString();
+
             BibEntry entry = parseEntry(type);
 
+            entry.setParsedSerialization(commentsAndEntryTypeDefinitionString+dumpTextReadSoFarToString());
+            entry.setCommentsBeforeEntry(commentsAndEntryTypeDefinitionString.substring(0,commentsAndEntryTypeDefinitionString.lastIndexOf('@')));
```
Siedlerchr: I would extract the `@` symbol as a constant, to make it clear. e.g.  private final `COMMENT_SYMBOL = '@'`

src/main/java/net/sf/jabref/importer/fileformat/BibtexParser.java
```diff
@@ -224,10 +224,14 @@ private void parseAndAddEntry(String type) {
          * for the user.
          */
         try {
+            String commentsAndEntryTypeDefinitionString = dumpTextReadSoFarToString();
```
Siedlerchr: Remove the "string" suffix from the variable.

src/main/java/net/sf/jabref/model/entry/BibEntry.java
```diff
@@ -679,25 +686,8 @@ public BibEntry withField(String field, String value) {
...
-                if (prolog.length() > 0) {
-                    return prolog;
-                }
-            } catch (StringIndexOutOfBoundsException ignore) {
-                // if this occurs a broken parsed serialization has been set, so just do nothing
-            }
-        }
-        return "";
+        // delete trailing whitespaces (between entry and text) from stored serialization
+        return commentsBeforeEntry.replaceFirst("\\s+$", "");
```
Siedlerchr: Would a simple trim() work, too?
https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#trim--

Siedlerchr:
Just some minor remarks, LGTM :+1:

src/main/java/net/sf/jabref/model/entry/BibEntry.java
matthiasgeiger: Nope. Whitespace in front of the comment should not be removed.

Siedlerchr: Okay. Then I would suggest extracting the Pattern with Pattern.compile and use Matcher.replaceF'irst...
Discussion about it on SO: http://stackoverflow.com/questions/1466959/string-replaceall-vs-matcher-replaceall-performance-differences
https://docs.oracle.com/javase/7/docs/api/java/util/regex/Matcher.html#replaceFirst(java.lang.String)

src/main/java/net/sf/jabref/importer/fileformat/BibtexParser.java
matthiasgeiger: It is not the `COMMENT_SYMBOL` but the general "start of bibtex entry/string/preamble/comment" symbol.
`@` as `char` is used in various places in the parser. However, a) I can't find a good name for the final field and b) I prefer touching this code as little as possible -> I stick to the `'@'` here :)

src/test/java/net/sf/jabref/importer/fileformat/BibtexParserTest.java
```diff
@@ -310,6 +310,40 @@ public void parseRecognizesEntryWhereLastFieldIsFinishedWithComma() throws IOExc
     }
 
     @Test
+    public void parseRecognizesEntryWithAtInField() throws IOException {
+        ParserResult result = BibtexParser.parse(new StringReader("@article{test,author={Ed von T@st}}"));
+
+        List<BibEntry> parsed = result.getDatabase().getEntries();
+        assertEquals(1, parsed.size());
```
tobiasdiez: It is a bit better to compare the parsed list to an expected entry list, i.e.
````
BibEntry expected = new BibEntry("article", "test").withField("author", "bla");
assertEquals(expected as list, parsed);
````

matthiasgeiger: ... which leads to the question whether comments prior the entry are significant to determine "equality" of two entries ;-)

I tend to say "no", thus not modifying the `equals()` method. Other opinions?

matthiasgeiger: > (How that effects @tobiasdiez comment I have no clue about though.)

If `equals()` does not include the pre-entry comments, it is not sufficient in this test to state `assertEquals(expectedList, parsedList)` - because the purpose of the test is to check for the correct parsing of these comments.

oscargus:
We are discussing the failing/unstable GUI tests at #1700

lenhard:
Just my two cents regarding this PR (by looking at the code): Should be fine! I was thinking about implementing something similar (separate comment field in BibEntry) before, but had avoided it so far. There are a lot of tests regarding parsing, so I am confident that we should not meet unexpected NullPointerExceptions and the like.

Good to see that we have a new person who is responsible for implementing the parsing logic ;-)

matthiasgeiger:
Forget it... This parsing stuff is still some black magic for me ;)

koppor:
Refs #123

PR #1718, opened 2016-08-11
Ascii integrity checker

stefan-kolb:
To avoid problems using BibTeX and non-ASCII chars.

Quick tests show that it also marks `é` as an error. Does anyone know what chars BibTeX can really handle? Umlauts are a problem but `é`  should work?! So i cannot include all of the extended ascii table http://www.theasciicode.com.ar/extended-ascii-code/letter-e-acute-accent-e-acute-lowercase-ascii-code-130.html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - It is now possible to add your own lists of protected terms, see Options -> Manage protected terms
 - Automatically generated group names are now converted from LaTeX to Unicode
 - Unified dialogs for opening/saving files
+- Add integrity check to avoid non-ASCII characters in BibTeX files
 
 ### Fixed
 - Fixed [#1632](https://github.com/JabRef/jabref/issues/1632): User comments (@Comment) with or without brackets are now kept
--- a/src/main/java/net/sf/jabref/logic/integrity/IntegrityCheck.java
+++ b/src/main/java/net/sf/jabref/logic/integrity/IntegrityCheck.java
@@ -23,6 +23,8 @@ import net.sf.jabref.model.entry.FileField;
 import net.sf.jabref.model.entry.InternalBibtexFields;
 import net.sf.jabref.model.entry.ParsedFileField;
 
+import com.google.common.base.CharMatcher;
+
 public class IntegrityCheck {
 
     private final BibDatabaseContext bibDatabaseContext;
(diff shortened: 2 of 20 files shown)


tobiasdiez:
LGTM :+1: but `\usepackage[utf8]{inputenc}` also saves the day 😄

oscargus:
It's not just umlauts. Anything outside of 7-bit ASCII will typically break LaTeX unless one uses `inputenc`. Break as in quite likely produce the wrong character in the document.

stefan-kolb:
So this checker seems reasonable for you?!

oscargus:
Yes, without more insight on the exact details (that LaTeX may use some 8-bit mapping that makes sense sometimes, some people may always use inputenc, etc) I would say that this is a good start. Then, ideally I would like the integrity checks to be disabled (or at least the filter settings remembered), but I'm not going to ask you to implement that now. :-)

oscargus:
(But please merge another PR with translation strings first, so you change your standpoint on union merge... ;-) I do not know how many times I've rebased in the last week where I had to do a manual edit that could be solve with union merge of the translation files...)

stefan-kolb:
I'm not 100 % sure what was the reason to change union merge but this messed up our changelogs and translations as far as i can remember, adding duplicates and so on?!

Siedlerchr:
@stefan-kolb  Locally it works, but there seemed to be problems with github not supporting this:
isaacs/github/issues/487

tobiasdiez:
@Siedlerchr @stefan-kolb so when union merge is enabled, then github don't automatically resolves the merge conflicts and a manual rebase / push is needed, right? (which is still better than the current system since the local rebase would be easier with union merge). 
But these problems with github dosn't explain why union merge was actually disabled for the changelog https://github.com/JabRef/jabref/commit/ac08efbf42638ade9a93136999ff61b71510c70f

Siedlerchr:
@tobiasdiez  Yeah. I think we should enable it back again. For the changelog and the lang files.
Conflicting lang files are hell!

oscargus:
Yes, GitHub did not support it (no idea how it is now), so that wasn't the problem. It did indeed lead to duplicates, but since the alternative is conflict markers and in quite a few cases it completely avoids conflict markers I still do not see the problem with duplicates. I rather remove duplicates half of the time than conflict markers twice as often.

As I recall it some developers got annoyed with the duplicates and removed the union merge.

oscargus:
A bit late, but wasn't the idea that all new integrity checks should come
in separate files?

stefan-kolb:
I though about this too when looking at the current implementation. We can just extract them all in a separate PR.

PR #1722, opened 2016-08-12
Better layout of general preferences tab

oscargus:
I know I have strange font settings, but I couldn't see some preferences on my portrait screen:

before

Hence, I changed the layout a bit to scale:

after

It is not perfect since all the stuff to the right keeps moving, even though all text is shown but still more practical than the current layout.

### DIFF
--- a/src/main/java/net/sf/jabref/gui/preftabs/GeneralTab.java
+++ b/src/main/java/net/sf/jabref/gui/preftabs/GeneralTab.java
@@ -111,8 +111,9 @@ class GeneralTab extends JPanel implements PrefsTab {
         encodings = new JComboBox<>();
         encodings.setModel(new DefaultComboBoxModel<>(Encodings.ENCODINGS));
 
-        FormLayout layout = new FormLayout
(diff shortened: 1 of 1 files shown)


matthiasgeiger:
Can you please push to the jabref repo, too. 

Testing whether your changes affect (the already working) prefs on my windows machine is easier if there is a pre-built version at builds.jabref.org.

oscargus:
I should probably have done that, yes. I'll try although experience says that pushing afterwards does not result in any built binaries...

oscargus:
And that is something recent, earlier it worked without any problems.

matthiasgeiger:
http://builds.jabref.org/bettergeneralprefs/ here it is - thanks! ;-)

oscargus:
And just because of all that hesitance it is working: http://builds.jabref.org/bettergeneralprefs/ :smile:

oscargus:
This is the consequence (3610 pixels wide), but there are other tabs that also could look better then.

wide

Still, this should be compared to the usability of the first screen shot.

PR #1725, opened 2016-08-12
Fixed some EDT issues

oscargus:
I looked into EDT bugs and found a few. Quite often they related to `AbstractWorker` it turned out.

If someone wants to find out more, add the following first in `main` or `start` in `JabRefMain`:
``` Java
RepaintManager.setCurrentManager(new CheckingRepaintManager());
```
then it is just to work with JabRef until an exception happens.

The most prominent issue solved was on saving where the UI is frozen during save, but not from the EDT. It is possible that some of the `AbstractWorker` issues are better to solve by running the complete `run` in the EDT thread, but somehow I guess one wants to avoid arbitrary processing there and this at least solves the problems.


src/main/java/net/sf/jabref/external/FindFullTextAction.java
```diff
@@ -59,6 +58,7 @@ public void init() throws Throwable {
     public void run() {
         if (basePanel.getSelectedEntries().size() != 1) {
             basePanel.output(Localization.lang("This operation requires exactly one item to be selected."));
+            result = Optional.empty();
```
oscargus: This is not related to EDT, but still solves a bug.

src/main/java/net/sf/jabref/gui/PreviewPanel.java
```diff
@@ -306,9 +307,19 @@ public void update() {
         );
         String newValue = sb.toString();
 
-        previewPane.setText(newValue);
-        previewPane.revalidate();
-
+        if (SwingUtilities.isEventDispatchThread()) {
```
koppor: Could this kind of code (checking for isEventDispatch etc.) extracted to some util method getting passed a function?

oscargus: Probably, yes. However, I'm not sure if that is a good idea in general. (Except for the fact that I do not really know how to pass it... As `Runnable`? Probably...

src/main/java/net/sf/jabref/importer/EntryFromPDFCreator.java
```diff
@@ -64,7 +64,7 @@ public boolean accept(File f) {
         PdfImporter pi = new PdfImporter(JabRefGUI.getMainFrame(), JabRefGUI.getMainFrame().getCurrentBasePanel(), JabRefGUI.getMainFrame().getCurrentBasePanel().getMainTable(), -1);
         ImportPdfFilesResult res = pi.importPdfFiles(Collections.singletonList(pdfFile.toString()));
         if (res.getEntries().size() == 1) {
-            return Optional.of(res.getEntries().get(0));
+            return Optional.ofNullable(res.getEntries().get(0));
```
oscargus: Bug fix so that one can click "Cancel" instead of selecting an entry type and still not get an NPE.

src/main/java/net/sf/jabref/external/SynchronizeFileField.java
```diff
@@ -179,7 +179,8 @@ public void run() {
                             } else {
                                 answer = JOptionPane.showOptionDialog(panel.frame(),
                                         Localization.lang("<HTML>Could not find file '%0'<BR>linked from entry '%1'</HTML>",
-                                                flEntry.link, aSel.getCiteKey()),
+                                                flEntry.link,
+                                                aSel.getCiteKeyOptional().orElse(Localization.lang("undefined"))),
```
oscargus: NPE when no cite key is set...

src/main/java/net/sf/jabref/importer/EntryFromPDFCreator.java
```diff
@@ -64,7 +64,7 @@ public boolean accept(File f) {
         PdfImporter pi = new PdfImporter(JabRefGUI.getMainFrame(), JabRefGUI.getMainFrame().getCurrentBasePanel(), JabRefGUI.getMainFrame().getCurrentBasePanel().getMainTable(), -1);
```
Siedlerchr: It would be nice, if you could at least assign the getMainFrame value to a local variable. Would improve the readability a bit.

oscargus: I'm sorry, but I have a hard time understanding comments like this. ~10 EDT bugs, 3 NPEs, 6 forgotten translations and the comment relates to the readability of a white line? 

Again, I think people hesitate touching bad code in the long run if the reviews focus more on what could be done as well (just because there are lines on the screen that are not perfect) instead of what has been done. No offense though and none taken.

(I think the readability is OK and if anything one should maybe just skip many of the arguments and get it from JabRefGUI in PdfImporter instead, but I haven't checked how it is called in other places.)

src/main/java/net/sf/jabref/external/FindFullTextAction.java
tobiasdiez: Probably better to initialize in the constructor.

src/main/java/net/sf/jabref/gui/FindUnlinkedFilesDialog.java
```diff
@@ -532,10 +533,12 @@ private void startSearch() {
                         @Override
                         public void stateChanged(ChangeEvent e) {
                             counter++;
-                            progressBarSearching.setString(counter + " files found");
+                            SwingUtilities.invokeLater(() -> progressBarSearching
```
tobiasdiez: This is hard to read. Maybe determine the localized string before calling `SwingUtil.invokeLater()`

src/main/java/net/sf/jabref/gui/FindUnlinkedFilesDialog.java
```diff
@@ -608,11 +613,11 @@ public void stateChanged(ChangeEvent e) {
     private void importFinishedHandler(List<String> errors) {
 
         if ((errors != null) && !errors.isEmpty()) {
-
             JOptionPane.showMessageDialog(this,
-                    "The import finished with warnings:\n" + "There " + (errors.size() > 1 ? "were " : "was ")
-                            + errors.size() + (errors.size() > 1 ? " files" : " file")
-                            + (errors.size() > 1 ? " which" : " that") + " could not be imported.",
+                    Localization.lang("The import finished with warnings:") + "\n"
+                            + (errors.size() > 1 ? Localization.lang("There were %0 files which could not be imported.",
```
tobiasdiez: This is again hard to read. Extract the second part to a separate variable.

src/main/java/net/sf/jabref/gui/JabRefFrame.java
```diff
@@ -1788,15 +1789,32 @@ public void removeCachedEntryEditors() {
...
      */
     public void block() {
-        getGlassPane().setVisible(true);
+        if (SwingUtilities.isEventDispatchThread()) {
+            getGlassPane().setVisible(true);
+        } else {
+            try {
+                SwingUtilities.invokeAndWait(() -> getGlassPane().setVisible(true));
+            } catch (InvocationTargetException | InterruptedException e) {
+                LOGGER.info("Problem blocking UI", e);
```
tobiasdiez: Logger.error? same below

src/main/java/net/sf/jabref/gui/PreviewPanel.java
tobiasdiez: Is this if statement even necessary? What happens if I'm on the event dispatcher thread and invoke 
````
SwingUtilities.invokeAndWait(() -> {
                    previewPane.setText(newValue);
                    previewPane.revalidate();
                });
````

File: src/main/java/net/sf/jabref/gui/importer/EntryFromPDFCreator.java [hunk 9]
```diff
@@ -64,7 +64,7 @@ public boolean accept(File f) {
         PdfImporter pi = new PdfImporter(JabRefGUI.getMainFrame(), JabRefGUI.getMainFrame().getCurrentBasePanel(), JabRefGUI.getMainFrame().getCurrentBasePanel().getMainTable(), -1);
         ImportPdfFilesResult res = pi.importPdfFiles(Collections.singletonList(pdfFile.toString()));
         if (res.getEntries().size() == 1) {
-            return Optional.of(res.getEntries().get(0));
+            return Optional.ofNullable(res.getEntries().get(0));
```
tobiasdiez: So `res.getEntries().get(0)` could be `null`? 
In this case I would prefer if `pi.importPdfFiles` is fixed. A null value in a list is very very counterintuitive.

tobiasdiez:
A few remarks, but in general looks good to me. But I have no experience with this swing threading stuff, so it would be better if somebody else also has a look at the code.

src/main/java/net/sf/jabref/gui/importer/EntryFromPDFCreator.java
Siedlerchr: Maybe there can be used a simple `Collections.emptyList()`

src/main/java/net/sf/jabref/gui/entryeditor/EntryEditor.java
```diff
@@ -947,7 +947,11 @@ public void updateAllContentSelectors() {
     @Subscribe
     public void listen(FieldChangedEvent fieldChangedEvent) {
         String newValue = fieldChangedEvent.getNewValue() == null ? "" : fieldChangedEvent.getNewValue();
-        setField(fieldChangedEvent.getFieldName(), newValue);
+        if (SwingUtilities.isEventDispatchThread()) {
```
koppor: I assume that this `if` statement is necessary and one cannot just call `invokeLater()`?

src/main/java/net/sf/jabref/external/FindFullTextAction.java
oscargus: I think it needs to be reset. If not, if you run it once and then try to run it with multiple entries selected, update will still be called and then result is not empty.

src/main/java/net/sf/jabref/gui/PreviewPanel.java
oscargus: Then you get an error. `invokeLater` works, but not `invokeAndWait`.

simonharrer:
For a hack, this is ok. I would prefer that a SwingWorker is used which can send updates of its progress. The AbstractWorker with its init/run/update methods is somewhat legacy which will not work with JavaFX at all. But the SwingWorker can be easily ported to the JavaFX equivalent, the stuff behind the AbstractWorker which is based on the spin library cannot. So I would recommend to invest the effort in slowly migrating to SwingWorker and then to JavaFx.

oscargus:
Thanks for looking at it @simonharrer. I take it as merge (subject to the comments) and look at SwingWorker eventually. Much more work to restructure all AbstractWorker but still worth getting rid of the issues fixed here.

I'll have a look at the other comments later tonight.

src/main/java/net/sf/jabref/gui/importer/EntryFromPDFCreator.java
oscargus: I got an NPE here at least. I do not remember the details (I think pressing cancel when asked about entry type) and I think I tested it after changing it. although I'm hindsight it looks more likely that the list is null than the list entry and that fixing it at the source may be better. emptyList() will break it as a null is better than index out of bounds. I'll have a new look later tonight.

src/main/java/net/sf/jabref/gui/JabRefFrame.java
oscargus: Makes sense. It is a severe problem if it happens. Still quite unlikely (which was why I selected warn), but clearly no harm in having error. (I'm thinking a lot in risk = severity*probability, but not useful here.)

src/main/java/net/sf/jabref/gui/entryeditor/EntryEditor.java
oscargus: You can, but I thought it made sense to change it directly rather than put it in a queue with possibly strange, despite unlikely, consequences. Edits -> do them (if allowed), gui updates -> put them in the queue.

src/main/java/net/sf/jabref/gui/JabRefFrame.java
oscargus: Fixed.

src/main/java/net/sf/jabref/gui/FindUnlinkedFilesDialog.java
oscargus: Still easier than before. ;-) I'll fix it.

src/main/java/net/sf/jabref/gui/importer/EntryFromPDFCreator.java
oscargus: I fixed it now. The null came from four-five methods down, but now that part is better with `Optional`.

oscargus:
OK, I changed some things based on the comments and improved the code in a few places.

oscargus:
I merge this in based on the previous reviews, so we have some time to find any unexpected consequences before release.

PR #1728, opened 2016-08-13
Added makeLabel(BibDatabaseContext, ...) method

oscargus:
Quite often `makelLabel` is called with MetaData and BibDatabase from the same BibDatabaseContext, so I created a convenience method for this.


oscargus:
I think it is unlikely that `BibDatabaseContext` can ever get rid of `Globals` so maybe this was a waste of time...

tobiasdiez:
I like it if methods operate on the database context instead of database + metadata separately. Thus I would like to merge this PR.

oscargus:
I agree, but since we will never (?) manage to move `BibDatabaseContext` to `logic` (even less `model`) because of all the `Globals` I'm not sure if it is wise to introduce this dependency.

PR #1729, opened 2016-08-13
Updated jabref_tr.properties

hakova:
Several entries translated. However, I believe there are typos in the original English entries in lines 1647 and 1654 as follows:
Line 1647 reads: "Connection_to_%0_server_stablished.", whereas it should read "Connection_to_%0_server_established."
Line 1654 reads: "The_connection_to_the_server_has_been_determinated.", whereas it should read "The_connection_to_the_server_has_been_terminated."
Turkish translations were done based on these assumptions.


oscargus:
Great! Thanks. I agree with you regarding the strings.

PR #1731, opened 2016-08-13
Always provide encoding in MetaData constructor

oscargus:
Unfortunately there was still a `Globals` dependency recently introduced...

### DIFF
--- a/src/main/java/net/sf/jabref/MetaData.java
+++ b/src/main/java/net/sf/jabref/MetaData.java
@@ -83,24 +83,15 @@ public class MetaData implements Iterable<String> {
     private MetaData(Map<String, String> inData, Charset encoding) throws ParseException {
         Objects.requireNonNull(inData);
         setData(inData);
-        this.encoding = encoding;
+        this.encoding = Objects.requireNonNull(encoding);
     }
 
     /**
-     * The MetaData object can be constructed with no data in it.
+     * The MetaData object can be constructed with no data in it, but needs encoding information
      */
-    @Deprecated
-    public MetaData() {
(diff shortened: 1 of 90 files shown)


src/main/java/net/sf/jabref/MetaData.java
```diff
@@ -83,26 +83,17 @@
     private MetaData(Map<String, String> inData, Charset encoding) throws ParseException {
         Objects.requireNonNull(inData);
         setData(inData);
-        this.encoding = encoding;
+        this.encoding = Objects.requireNonNull(encoding);
     }
 
     /**
-     * The MetaData object can be constructed with no data in it.
+     * The MetaData object can be constructed with no data in it, but needs encoding information
```
tobiasdiez: Why is the encoding special and always has to be provided while other metadata is optional? I would still provide an empty constructor and let `getEncoding` return an `Optional<Encoding>`. Then it is the caller business to decide which encoding he wants if no encoding is set in the metadata.

oscargus:
Clearly a better idea. :+1: There are quite often cases where we need to provide "unnecessary" data unfortunately, but then that's no reason to increase them.

oscargus:
OK, I changed that now. I also changed `ImportFormat` to always take an `ImportFormatPreferences` as an argument to the constructor. This was really required to be able to test `CustomImports` and I think it makes sense for custom import format developers to have some access to preferences. Not sure if I should provide an argument-less constructor as well, creating an `ImportFormatPreferences` from `Globals`, just to ease the transition (also, people may have old custom import formats that they do not have source for and so on...). WDYT?

oscargus:
I'll try to redo this in smaller steps.

PR #1736, opened 2016-08-14
Fixed some Codacy issues

oscargus:
Fixed a number of Codacy issues.


Siedlerchr:
In General good, but I am not sure if this rule: "Declarations at top",  should also apply to Lambdas...

oscargus:
I see your point, but it was still a reasonably short lambda.

PR #1737, opened 2016-08-14
Importer now use FileExtensions enum, too, FileDialogs display extension

Siedlerchr:
Grouped Importer in FileExtension
Created test for FileExtensions
File Dialogs now display description and extension 

(internal change)

filefilter


src/main/java/net/sf/jabref/gui/FileExtensions.java
```diff
@@ -12,36 +14,34 @@
...
+    REPEC("REPEC New Economic Papers (NEP)", "txt"),
+    RIS(Localization.lang("%0 file", "RIS"), "ris"),
+    SILVER_PLATTER(Localization.lang("%0 file", "SilverPlatter"), "dat", "txt"),
+    XMP(Localization.lang("XMP-annotated PDF"), "pdf"),
+
+    AUX(Localization.lang("%0 file", "AUX"), "aux"),
+    JSTYLE(Localization.lang("Style file"), "jstyle"),
+    LAYOUT(Localization.lang("Custom layout file"), "layout"),
+    TERMS(Localization.lang("Protected terms file"), "terms"),
+    TXT(Localization.lang("%0 file", Localization.lang("plain text")), "txt"),
```
oscargus: There is already an old translation string for "plain text", right? If not, "Plain text" would be preferred.

oscargus: Or maybe even go for "Plain text file" now that I read the whole line of code. :-)

src/main/java/net/sf/jabref/gui/importer/ImportFormats.java
```diff
@@ -26,6 +26,7 @@
 import javax.swing.Action;
 import javax.swing.JFileChooser;
 import javax.swing.JOptionPane;
+
```
oscargus: Checkstyle will complain (I guess).

src/main/java/net/sf/jabref/gui/FileExtensions.java
```diff
@@ -12,36 +14,34 @@
 public enum FileExtensions {
```
oscargus: Slightly unrelated, but didn't @tobiasdiez have some good argumentation that this type of file should be in singular? So `FileExtension`. Not in any way required for the PR though.

oscargus:
Looks good to me! Some minor comments, but nothing crucial (and I guess Checkstyle will sort one of them out :-)).

src/main/java/net/sf/jabref/gui/FileExtensions.java
Siedlerchr: Ah I just noticed, we have already both cases of Plain text in the translations.

src/main/java/net/sf/jabref/gui/FileExtensions.java
Siedlerchr: I already thought about it when doing it first in  #1336 and would generally agree, but in this case one entry can have multiple extensions (for example EndNote) and I therefore thought the plural form would reflect it better

Siedlerchr:
Anything else, otherwise I would merge it in. (Codacy probably failed because I rebased/squashed some things)

PR #1738, opened 2016-08-14
Added method singleFromStringOptional

oscargus:
Seemed convenient, even though the source does not fully benefit from it.

### DIFF
--- a/src/main/java/net/sf/jabref/logic/importer/fileformat/BibtexParser.java
+++ b/src/main/java/net/sf/jabref/logic/importer/fileformat/BibtexParser.java
@@ -118,23 +118,19 @@ public class BibtexParser {
         }
     }
 
+    @Deprecated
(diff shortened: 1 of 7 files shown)


src/main/java/net/sf/jabref/logic/importer/fileformat/BibtexParser.java
```diff
@@ -135,6 +135,15 @@ public static BibEntry singleFromString(String bibtexString,
         return entries.iterator().next();
```
tobiasdiez: make singleFromString deprecated and rewrite it as `return singleFromStringOptional(...).orElse(null)`;
Then this can be merged.

oscargus: Makes sense.

PR #1752, opened 2016-08-16
Enhance error message of ensureNoDuplicates test

Braunch:
I added the information in which bundle the test failed when the assertion fails.

### DIFF
--- a/src/test/java/net/sf/jabref/logic/l10n/LocalizationConsistencyTest.java
+++ b/src/test/java/net/sf/jabref/logic/l10n/LocalizationConsistencyTest.java
@@ -91,7 +91,7 @@ public class LocalizationConsistencyTest {
 
                 List<String> duplicates = properties.getDuplicates();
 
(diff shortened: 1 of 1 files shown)


src/test/java/net/sf/jabref/logic/l10n/LocalizationConsistencyTest.java
```diff
@@ -91,7 +91,7 @@ public void ensureNoDuplicates() {
 
                 List<String> duplicates = properties.getDuplicates();
 
-                assertEquals("Duplicate keys of " + lang, Collections.emptyList(), duplicates);
+                assertEquals("Duplicate keys of " + lang + " within bundle " + bundle, Collections.emptyList(), duplicates);
```
koppor: `in`  instead of `within`?

Otherwise: LGTM.

bartsch-dev:
LGTM

stefan-kolb:
Directly commited to `master`. Thanks for pointing this out!

PR #1753, opened 2016-08-16
[WIP] Enable failing unit tests for JavaFX

bruehldev:
Fixed failing tests 
(localization, duplicate or obsolete or missing keys)


src/main/resources/l10n/Menu_de.properties
```diff
@@ -211,3 +211,5 @@ Disable_highlight_groups_matching_entries=Keine_Gruppen_hervorheben
 Online_help_forum=Online-Hilfeforum
 
 Manage_protected_terms=
+
+BibTeX=
```
koppor: Is it really required to translate "BibTeX"? It should not need to be

Siedlerchr: All format names like BibTeX or BibLaTeX do not need to be translated, same applies to things like the fetcher names (e.g. Springer, Gvk ), as this are usually proper nouns.

tobiasdiez: Can we add tests, that certain words are never translated?

Siedlerchr: Maybe a blacklist would be a good idea. If the word is contained in the blacklist, then an error.

Braunch: I think adding these tests should be discussed in a separate Issue. @bruehldev s task here is to reenable the tests so that the JavaFX branch can be merged with the master.

bruehldev:
Curiors failure of the checks, it works on my system.  Should i merge the other javafx branches and hope the failure will dissapear?

Braunch:
@bruehldev  you could merge one in and then see if the travis job passes when you push. If it still fails, you can be pretty sure it's not just a travis instability problem.

bruehldev:
Sorry for the huge amount of commits. The current version of this branch can't display the Customize key bindings dialog. I wanted to share the code with @boceckts to debug the issue.

bruehldev:
I started over and pushed it to get a new clean version. Unfourtnatly the test `LocalizationConsistencyTest. findObsoleteLocalizationKeys` shows wrong obsolete keys and i deleted necessary keys for the managekey dialog. 
So theres one failure left from the unit tests.

koppor:
>. Unfourtnatly the test LocalizationConsistencyTest.
findObsoleteLocalizationKeys shows wrong obsolete keys

Which ones? Please provide examples.

> and i deleted necessary keys for the managekey dialog.

What does that mean? Keys requiring a translation are deleted?

tobiasdiez:
The problem is that the localization tests don't recognize the strings in the fxml files (since they are of the form `<label text = "%localized description">`. I will take care of it.

bruehldev:
I rebased my branch to upstream/javafx and forced pushed. Thus, GitHub closed the PR automatically. Will open a new PR soon.

koppor:
I reviewed the merge (2eafa31) and the two commits before together with @bruehldev and it is OK. The other commits appearing here are OK (!), because we merge master into javafx and master contains these commits. A review of the merge commit using GitHub is too hard and therefore, we did it on screen.

PR #1756, opened 2016-08-16
Moved a few more initialization to GUIGlobals.init()

oscargus:
No point in having a static initializer when there is an init method that is called at startup or a single line just after the init method call.

### DIFF
--- a/src/main/java/net/sf/jabref/JabRefGUI.java
+++ b/src/main/java/net/sf/jabref/JabRefGUI.java
@@ -15,7 +15,6 @@
 */
 package net.sf.jabref;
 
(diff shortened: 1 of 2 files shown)


stefan-kolb:
The difference is that the static initializer is called every time the class is first used. If you put it into the init method you must be sure that this method is called explicitely every time. Also, if you move it to the init method better put it at the very beginning of the method to prevent any side effects of calling it later than on class startup.

oscargus:
Since `init()` is called before any GUI is initialized at all (as far as I can tell), and definitely before any entry editors are initialized (the static method updated the colors for them), it should be safe.

PR #1766, opened 2016-08-17
Added program to generate a table of all characters and fixed some ch…

oscargus:
I made a small program that generates a LaTeX table with all characters in the converter. It prints the number, the character as Unicode `\symbol{nnn}`, the resulting LaTeX symbol, and the LaTeX code, see the attached file. In this way, we can make sure that all symbols makes sense (some of the current one were incorrect). Also relates to #1761 which is not currently included.

I note two thing: 
1) using `\usepackage[utf8]{inputenc}` did only work for accented characters. I had a version where I printed the Unicode character in the table, but that only rarely worked, although they appeared nicely in the .tex-file
2) Packages and fonts are a bit of a mystery. It may be that with a better font and maybe a few more packages, more symbols are shown.
[CharGen5.pdf](https://github.com/JabRef/jabref/files/423135/CharGen5.pdf)

### DIFF
--- a/src/main/java/net/sf/jabref/cli/GenerateCharacterTable.java
+++ b/src/main/java/net/sf/jabref/cli/GenerateCharacterTable.java
@@ -1,0 +1,45 @@
+package net.sf.jabref.cli;
+
+import java.util.Map;
+import java.util.TreeMap;
+
+import net.sf.jabref.logic.util.strings.HTMLUnicodeConversionMaps;
+
+public class GenerateCharacterTable {
+
+    public static void main(String[] args) {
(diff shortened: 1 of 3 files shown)


oscargus:
Updated version with Unicode characters as well. Clearly it is some sort of font issue.

[CharGen5.pdf](https://github.com/JabRef/jabref/files/423428/CharGen5.pdf)

oscargus:
Would it be feasible to run this program and compile the resulting tex file as part of the tests? If no compilation errors, all symbols are at least syntactically correct.

tobiasdiez:
You can probably add it to the build script. The following links might be helpful:
https://gist.github.com/LindsayBradford/11274456
https://github.com/mgmax/pdflatex-travis-example

koppor:
The tex file needs too much packages and might take a long time to get texlive running at **each** test.

PR #1772, opened 2016-08-18
Removed/moved preferences which are constants

oscargus:
Remove constant preference values.

Relates to: https://github.com/koppor/jabref/issues/155


stefan-kolb:
:+1:

PR #1773, opened 2016-08-18
Added UpdateFieldPreferences

oscargus:
Less dependency on JabRefPreferences in a small part of logic...


tobiasdiez:
LGTM :+1:

stefan-kolb:
Hm what is the big difference here? The only thing that changed is that some prefs are now stored in another class that still need JabRefPreferences?!

oscargus:
Correct. But it is very simple to move the `fromPreferences` method to
`JabRefPreferences` and eventually the whole logic will not access
`JabRefPreferences` at all. I think we are soon reaching the point where
that can happen as there are no Globals calls left (once one of my other
PRs is merged) and most classes and methods have this type of specialised
preferences. (Simon and Tobias convinced me of this as I had similar
doubts.)

oscargus:
I thought I'd add a method to `JabRefPreferences` as we actually are closing in on this now.

PR #1774, opened 2016-08-18
Fix #1771 Show all supported import types as default file filter

### LINKED ISSUE #1771: Open dialog: All supported types as default
For open dialogs with multiple extensions is would be nice to have "All supported types (*.bibx, *.bib, *.txt, ...)" as default. (File > Import into new database)

stefan-kolb:
Little bit tricky to get this done. Would have preferred a cleaner option but I still need access to some JFileChooser methods...


stefan-kolb:
@Siedlerchr This is one for you to review :smile:

src/main/java/net/sf/jabref/gui/importer/ImportFormats.java
```diff
@@ -84,64 +59,56 @@ public static AbstractAction getImportAction(JabRefFrame frame, boolean openInNe
...
-
-                try {
-                    if (!file.exists()) {
-                        // Warn that the file doesn't exists:
-                        JOptionPane.showMessageDialog(frame,
-                                Localization.lang("File not found") + ": '" + file.getName() + "'.",
-                                Localization.lang("Import"), JOptionPane.ERROR_MESSAGE);
-                        return;
+                SortedSet<ImportFormat> importers = Globals.IMPORT_FORMAT_READER.getImportFormats();
+                List<FileExtensions> extensions = importers.stream().map(p -> p.getExtensions()).collect(Collectors.toList());
```
Siedlerchr: I would rather use the ImportFileFilter and there add a new constructor overload which accepts a List of Import Formats, e.g. just moving the Stream stuff in the ImportfileFilter, because it internally uses already the 
Would make it here much cleaner. And in the dlg class you just call SetFileFilter(importFF)... 

Maybe in future we can recycle that for the Export formats, too...

src/main/java/net/sf/jabref/gui/importer/ImportFormats.java
```diff
@@ -84,64 +59,56 @@ public static AbstractAction getImportAction(JabRefFrame frame, boolean openInNe
...
+                // Add filters for extensions
+                dialog.withExtensions(extensions);
+
+                Optional<Path> selectedFile = dialog.showDialogAndGetSelectedFile();
+
+                selectedFile.ifPresent(sel -> {
+                    try {
+                        File file = sel.toFile();
+
+                        if (!file.exists()) {
```
Siedlerchr: Isn't this redundant`? If the path is giving an Optional.Empty and you check with ifPresent then you already checked if the file/path exists.

Siedlerchr:
Generally a good solution, but I would move the logic to the ImportFileFilter, makes it a bit more clear.

src/main/java/net/sf/jabref/gui/importer/ImportFormats.java
stefan-kolb: I checked if there is a Path inside the Optional but not if the file exists. However, this will probably never happen as long as the file is not removed between the selection and the import. Was a artifact from the old code i was also thinking about...

src/main/java/net/sf/jabref/gui/importer/ImportFormats.java
stefan-kolb: Hm, I don't really get the idea with this text. Adds way more complexity in my opinion. I need a new Import format then and matching FileExtensions etc pp?!

Siedlerchr: Well, the ImportFileFilter is an custom already exiting FileFilter, which maps the extension to the import format.
It internally calls the accept method of the method of an ExternalFileFilter. See #1775 which you already merged in...

If you stay with your solution, then you could remove it completely.

src/main/java/net/sf/jabref/preferences/JabRefPreferences.java
```diff
@@ -303,7 +303,6 @@
     public static final String SHOW_FILE_LINKS_UPGRADE_WARNING = "showFileLinksUpgradeWarning";
     public static final String SIDE_PANE_WIDTH = "sidePaneWidth";
     public static final String LAST_USED_EXPORT = "lastUsedExport";
-    public static final String LAST_USED_IMPORT = "lastUsedImport";
```
Siedlerchr: Why did you remove the Preference? It is a really nice feature to preselect the last used Filefilter for exporting/importing.

stefan-kolb: Because all supported filters are now default so you can definitely see possible files. If the importer has problems with auto recognition one can select the import format by hand (although this should not happen). Just wanted to remove complexity here. Maybe it's still needed - we have to discuss this...

Siedlerchr: Yeah, you are right, forgot that new feature. Then it's ok to be removed.

2016-08-19 11:58 GMT+02:00 Stefan Kolb <notifications@github.com>:

src/main/java/net/sf/jabref/gui/importer/ImportFormats.java
stefan-kolb: I know that but I need 2 new Objects for getting that to work, a FileExtension enum with all import endings and a new Import Format class for all Import Files?! This is a little bit complex to implement as the number of import formats can change. or am I on the wrong track here?

tobiasdiez: Not sure if I get the point of your discussion, but in general I support the idea that FileDialog should support a way to set the extensions and file filter based on import formats (because this should happen quite often).
What about adding a `static FileDialog createDialogForImport(frame, workingDir, Collection<ImportFormat>)` which contains most of the initialization code here.

Siedlerchr: @stefan-kolb  You are thinking a bit too complex ;)
We already have those two thing in the ImportFormat class.
The importFormat  has a method getExtensions wich returns the asscoatiacted enum value from the FileExtensions class.  
Currently in the importFileFilter we have support for exaclty one importFormat.

stefan-kolb: A ok, so youre thinking about `public ImportFileFilter(String name, List<ImportFormat> formats)`. I'm still note quite sure if we really need the `ImportFileFilter`. Internally all it really does is `new FileNameExtensionFilter(extensions.getDescription(), extensions.getExtensions());` and then overriding all methods introducing potential problems as in #1775.

stefan-kolb: But ok, let's try this. Maybe we can reuse this then somehow.

stefan-kolb: @Siedlerchr Moved the logic into the ImportFileFilter class.

Siedlerchr: Yeah. That looks better. Keeps the dialog more clear. :+1: 
I have no further remarks. You can merge it,.

src/main/java/net/sf/jabref/gui/importer/ImportFormats.java
Siedlerchr: Yeah. That can always happen regardless of files or not.
I think In the dialog itself is already checked if the file/path exists.

tobiasdiez: Maybe use the nio equivalent? Files.exist(path) or whatever the name of the method was.

PR #1775, opened 2016-08-18
Some OS didn't show directories when a file type was selected in the import dialog

bartsch-dev:
When importing a file (`Import into new | current Database`) and a file type was selected sometimes the directories weren't shown anymore (we experienced it under win10 and linux).

 Furthermore BibTeXml can also be in .xml format.

### DIFF
--- a/src/main/java/net/sf/jabref/gui/importer/ImportFileFilter.java
+++ b/src/main/java/net/sf/jabref/gui/importer/ImportFileFilter.java
@@ -27,14 +27,14 @@ class ImportFileFilter extends FileFilter implements Comparable<ImportFileFilter
 
     private final ImportFormat format;
     private final String name;
-    private final FileNameExtensionFilter filextFilter;
+    private final FileNameExtensionFilter extensionFilter;
 
 
     public ImportFileFilter(ImportFormat format) {
         this.format = format;
         FileExtensions extensions = format.getExtensions();
         this.name = extensions.getDescription();
-        filextFilter = new FileNameExtensionFilter(extensions.getDescription(), extensions.getExtensions());
+        this.extensionFilter = new FileNameExtensionFilter(extensions.getDescription(), extensions.getExtensions());
     }
 
     public ImportFormat getImportFormat() {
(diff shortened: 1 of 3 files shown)


src/main/java/net/sf/jabref/gui/importer/ImportFileFilter.java
```diff
@@ -43,7 +43,7 @@ public ImportFormat getImportFormat() {
 
     @Override
     public boolean accept(File file) {
-        return filextFilter.accept(file);
+        return file.isDirectory() || filextFilter.accept(file);
```
stefan-kolb: Maybe add if (file == null) return false;

stefan-kolb:
LGTM :+1:

src/main/java/net/sf/jabref/gui/importer/ImportFileFilter.java
bartsch-dev: done

CHANGELOG.md
```diff
@@ -72,7 +72,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - Fixed [#1716](https://github.com/JabRef/jabref/issues/1716): `@`-Symbols stored in BibTeX fields no longer break the database
 - Fixed [#1499](https://github.com/JabRef/jabref/issues/1499): {} braces are now treated correctly in in author/editor
 - Fixed [#1531](https://github.com/JabRef/jabref/issues/1531): `\relax` can be used for abbreviation of author names
-
```
oscargus: Did this show up in the development version (like a few hours ago ;-))? In that case I think we should skip the ChangeLog entry. Great job though!

Siedlerchr: Must be dev version only, related to the restructuring of the Dialogues which came with #1336

oscargus: I thought so. Well, it is not entirely consistent (there are similar
entries in the current development section of the ChangeLog), so more for
next time.

(I'm thinking that maybe someone should traverse the ChangeLog and sort
issues at some stage. Then one can rephrase some entries, which are just
copies of the issue, I've done that, and possibly remove "doubtful" entries
like this. Hard to always know and keep track of though.)

bartsch-dev: Removed the ChangeLog entry.

Issue #1748, opened 2016-08-16
Check import .bib file


koppor:
Setting connecting to database written by JabRef 2.11.

 * User uses JabRef 2.11 to work with an external database
 * User uses JabRef 3.6dev and connects to external database
 * User sees no entries
 * User imports `bib` file with entries "Import into current database"
 * JabRef shows "Aktualisierung verweigert. Den Eintrag, den Sie zur Zeit bearbeiten, ist auf der geminsam genutzen Datenbank nicht mehr vorhanden." for each entry.
 * After the import, all entries appear twice.

@mtecfred

Refs #1747

obraliar:
This issue is solved and can be closed.

PR #1777, opened 2016-08-18
Quickfix for import into current [shared] database

obraliar:
Quickfix for the comments in https://github.com/JabRef/jabref/issues/1748:

> JabRef shows "Aktualisierung verweigert. Den Eintrag, den Sie zur Zeit bearbeiten, ist auf der geminsam genutzen Datenbank nicht mehr vorhanden." for each entry.

and

> After the import, all entries appear twice.

Bug occured due to change of a `BibEntry` before adding it into local database.


koppor:
Quick fix is OK. A discussion with @obraliar revealed that the reason of the wrong event is `entry.setId(IdGenerator.next());` in `gui/importer/ImportInspectionDialog.java` line 474. We have to investigate further if the event system has some huger issue. An event for an entry not being present in the database should not be posted.

PR #1778, opened 2016-08-18
[WIP] Add test for ModsExportFormat

tschechlovdev:
In this PR the ModsExportFormat, that can be found in the exporter package, will be tested. The logic of the ModsExportFormat is located in the logic/mods package. 
More tests for the class PageNumbers and PersonName will  be added soon.


src/main/java/net/sf/jabref/logic/mods/MODSEntry.java
```diff
@@ -116,11 +116,11 @@ private void populateFromBibtex(BibEntry bibtex) {
         if (bibtex.hasField(BibEntry.KEY_FIELD)) {
             id = bibtex.getCiteKey();
```
Siedlerchr: Use the Optional methods getFiedOptional 
Then you can use the ifPresent or isPresent to check if an Entry is there

src/main/java/net/sf/jabref/logic/mods/MODSEntry.java
```diff
@@ -116,11 +116,11 @@ private void populateFromBibtex(BibEntry bibtex) {
...
         }
-        if (bibtex.hasField("place")) { // TODO: "place" is the MODS version, in BibTeX: "address", BibLaTeX: "location"?
+        if (bibtex.hasField("address")) { // TODO: "place" is the MODS version, in BibTeX: "address", BibLaTeX: "location"?
```
Siedlerchr: Have a look at the getFieldAlias (just don't renenber the correct name).  It allows you to get the value .e.g adress or journal, independent of the actual field. For example when year not present than the value is extracted from the date field

src/test/java/net/sf/jabref/logic/exporter/ModsExportFormatTest.java
```diff
@@ -0,0 +1,87 @@
...
+            return stream.map(n -> n.getFileName().toString()).filter(n -> n.endsWith(".bib"))
+                    .filter(n -> n.startsWith("Mods")).collect(Collectors.toList());
+        }
+    }
+
+    @Before
+    public void setUp() throws Exception {
+        Globals.prefs = JabRefPreferences.getInstance();
+        databaseContext = new BibDatabaseContext();
+        charset = Charsets.UTF_8;
```
Siedlerchr: StandardCharsets.UTF8 is the correct java method, the other is an external lib.
All "default" enums from the Java Api have the prefix Standard (e.g. StandardFileOptions)

src/test/resources/net/sf/jabref/logic/exporter/ModsExportFormatTestMultipleEntries.bib
```diff
@@ -0,0 +1,20 @@
+% Encoding: UTF-8
+
+@Article{,
+  author   = {Shruthi, Basavaradhya Sahukar and Vinodhkumar, Palani and Selvamani},
+  title    = {Proteomics: A new perspective for cancer.},
+  journal  = {Advanced biomedical research},
```
Siedlerchr: I am not aware of the mods fornat, but does it have an issue field`?
Then this would be worth testing, too. Some journals have both volume and issue

src/main/java/net/sf/jabref/logic/mods/MODSEntry.java
tschechlovdev: Done.

src/main/java/net/sf/jabref/logic/mods/MODSEntry.java
tschechlovdev: Done.

src/test/java/net/sf/jabref/logic/exporter/ModsExportFormatTest.java
tschechlovdev: Done.

src/test/resources/net/sf/jabref/logic/exporter/ModsExportFormatTestMultipleEntries.bib
tschechlovdev: I don't think so. The issue field will not get exported and I couldn't find anywhere that the mods format should  have an issue field.

oscargus: https://sourceforge.net/p/bibutils/home/Working%20with%20MODS/

"The part tag is the only new tag defined in the <relatedItem> that isn't defined in the main mods level. For a <relatedItem type="host">, it describes where the object is in terms of volume, issue, pages, and the like. From the above example:

    <part>
       <date>1984</date>
       <detail type="volume"><number>38</number></detail>
       <extent unit="page">
           <start>879</start>
           <end>887</end>
       </extent>
    </part>

The detail is used to describe single item locators, such as volume, issue, number, section, and session."

Siedlerchr: Btw, the shemas are now here on github, too: according to their release annoucncement:
http://www.loc.gov/standards/mods/mods-3-6-announcement.html

tschechlovdev: Thanks! I've added the issue field.

src/test/java/net/sf/jabref/logic/exporter/ModsExportFormatTestFiles.java
```diff
@@ -0,0 +1,87 @@
...
+    @Parameter
+    public String filename;
+
+    @Rule
+    public TemporaryFolder testFolder = new TemporaryFolder();
+
+
+    @Parameters(name = "{0}")
+    public static Collection<String> fileNames() throws IOException {
+        try (Stream<Path> stream = Files.list(Paths.get(PATH_TO_FILE))) {
```
Braunch: Can you use class.getResource to get the resource folder? This makes the test more robust to changes in the project structure.

src/test/java/net/sf/jabref/logic/exporter/ModsExportFormatTestFiles.java
```diff
@@ -65,7 +62,7 @@ public void setUp() throws Exception {
         charset = StandardCharsets.UTF_8;
         modsExportFormat = new ModsExportFormat();
         tempFile = testFolder.newFile();
-        testImporter = new BibtexImporter(ImportFormatPreferences.fromPreferences(Globals.prefs));
+        testImporter = new BibtexImporter(Globals.prefs.getImportFormatPreferences());
     }
 
     @Test
```
Siedlerchr: Just saw it: Please use Standard.Charsets.UTF-8, otherwise gradle check on windows  will be complaining about the encoding again ;)

zellerdev: I can not use  `Standard.Charsets.UTF-8` on this branch. It does not exist. My master does not have it, too.

oscargus: I think @Siedlerchr meant StandardCharsets.UTF_8, which it looks like you are using, at least assigning it to charset.

Siedlerchr: Yes ;) Don't take my code suggestions too literally, as I sometimes recall them only from my memory when reading a review

src/test/java/net/sf/jabref/logic/exporter/ModsExportFormatTestFiles.java
```diff
@@ -0,0 +1,82 @@
...
+
+    @Test
+    public final void testPerformExport() throws Exception {
+        String xmlFileName = filename.replace(".bib", ".xml");
+        Path importFile = Paths.get(ModsExportFormatTestFiles.class.getResource(filename).toURI());
+        String tempFilename = tempFile.getCanonicalPath();
+        List<BibEntry> entries = testImporter.importDatabase(importFile, Charset.defaultCharset()).getDatabase()
+                .getEntries();
+
+        modsExportFormat.performExport(databaseContext, tempFile.getPath(), charset, entries);
```
boceckts: Why don't you use `Charset.defaultCharset()` as you used for the import.

src/test/java/net/sf/jabref/logic/exporter/ModsExportFormatTestFiles.java
```diff
@@ -0,0 +1,82 @@
...
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameter;
+import org.junit.runners.Parameterized.Parameters;
+
+import static org.junit.Assert.assertEquals;
+
+@RunWith(Parameterized.class)
+public class ModsExportFormatTestFiles {
+
+    public Charset charset;
```
boceckts: This variable is only used once, I think it can be replaced by `Charset.defaultCharset()`.

src/test/java/net/sf/jabref/logic/exporter/ModsExportFormatTestFiles.java
oscargus: Better to explicitly use a well defined charset. `Charset.defaultCharset()` is system dependent, so it might be that the test will not pass on certain systems then.

src/test/resources/net/sf/jabref/logic/exporter/ModsExportFormatTestMultipleEntries.xml
```diff
@@ -0,0 +1,105 @@
...
+<roleTerm type="text">author</roleTerm>
+</role>
+</name>
+<originInfo>
+<dateIssued>2016-</dateIssued>
+<issuance>monographic</issuance>
+</originInfo>
+<identifier>ShruthiVinodhkumarSelvamani2016</identifier>
+<typeOfResource>text</typeOfResource>
+<genre authority="marc">article</genre>
```
boceckts: I think this entry corresponds to the second entry in the ModsExportFormatTestMultipleEntries.bib file. Here the genre is defined as `article` but in the .bib file the second entry is an `Inproceedings`.

src/test/resources/net/sf/jabref/logic/exporter/ModsExportFormatTestMultipleEntries.xml
```diff
@@ -0,0 +1,105 @@
...
+<role>
+<roleTerm type="text">author</roleTerm>
+</role>
+</name>
+<originInfo>
+<dateIssued>2016-</dateIssued>
+<issuance>monographic</issuance>
+</originInfo>
+<typeOfResource>text</typeOfResource>
+<genre authority="marc">inproceedings</genre>
```
boceckts: Same as the last comment, here the genre is defined as `inproceedings` but the matching entry in the .bib file is an `Article`, is this intentional?

src/test/java/net/sf/jabref/logic/exporter/ModsExportFormatTestFiles.java
Siedlerchr: Please not! When you test sth with export/import, please always use StandardCharsets.UTF-8.
Otherwise gradle build on win will complain (because the console charset is different from utf-8 on win)

boceckts:
When trying to execute the `ModsExportFormatTestFiles` class I get an `java.nio.file.InvalidPathExceptoion`.
Does this happen to anyone else, as the checks are passing I guess linux users don't have this issue?

JabRef *unknown*
windows 8.1 6.3 amd64 
Java 1.8.0_25

src/test/java/net/sf/jabref/logic/exporter/ModsExportFormatTestFiles.java
Siedlerchr: Yes, please always use StandardCharsets.UTF-8 everywhere for import and export

src/test/java/net/sf/jabref/logic/exporter/ModsExportFormatTest.java
```diff
@@ -0,0 +1,50 @@
...
+        charset = StandardCharsets.UTF_8;
+        modsExportFormat = new ModsExportFormat();
+        testImporter = new BibtexImporter(Globals.prefs.getImportFormatPreferences());
+    }
+
+    @Test(expected = SaveException.class)
+    public final void testPerformExportTrowsSaveException() throws Exception {
+        String filename = "ModsExportFormatTestAllFields.bib";
+        Path importFile = Paths.get(ModsExportFormatTestFiles.class.getResource(filename).toURI());
+        List<BibEntry> entries = testImporter.importDatabase(importFile, Charset.defaultCharset()).getDatabase()
```
Siedlerchr: see below, please use utf-8 charset

src/test/resources/net/sf/jabref/logic/exporter/ModsExportFormatTestMultipleEntries.xml
tschechlovdev: The order how the entries get exported are not the same as the order in the bib file. So that's why they are switched.

src/test/resources/net/sf/jabref/logic/exporter/ModsExportFormatTestMultipleEntries.xml
tschechlovdev: See above

tschechlovdev:
I've used now StandardCharsets.UTF-8

@boceckts The exception should be fixed now.

src/test/resources/net/sf/jabref/logic/exporter/ModsExportFormatTestMultipleEntries.xml
boceckts: Yes they are switched but what I mean is that the second entry's genre is "Inproceedings" according to ModsExportFormatTestMultipleEntries.bib but here the matching entry's genre is defined as "article".

tschechlovdev: Now I know what you mean... Yeah you're right. I'll look into it.

tschechlovdev: should be in the right order now.

boceckts: Do the tests need to be adjusted as they didn't fail when there actually was an error?

tschechlovdev:
This PR will be suspended, untill the issue #1920 is solved. Because the exported files could be very different to how they are now. In my opinion it makes the most sense, to first rewrite the exporter and then add a test for it (or adapt the existing test).
I will close this PR and then open a PR for the rewrite of the exporter and the test of it at once.

PR #1780, opened 2016-08-18
Fix arxiv fetcher not working correctly

### LINKED ISSUE #1776: ArXiv fetcher seems to be broken
Check current builds. https://travis-ci.org/JabRef/jabref/builds/153277476

Siedlerchr:
Fix #1776 remove Cookie Handler implementation
Our UrlDownload works without or own CookieHandler cookies, too.

Edit// 3.5 worked partly for me with arxivID
All unit test and manual tests work again.

I think the problem was that due to the inheritance, our cookie handler was called in every HTTP connection. No idea why it failed with arxiv only.
@stefan-kolb


src/main/java/net/sf/jabref/logic/importer/fetcher/ArXiv.java
```diff
@@ -74,6 +74,7 @@
 
     private static final String API_URL = "http://export.arxiv.org/api/query";
 
+
```
stefan-kolb: Why do you introduce these additional lines?

stefan-kolb:
I wanted to remove this Cookie Handler anyway :+1: Not sure what it did before. If everything works without it, it would be great!

src/main/java/net/sf/jabref/logic/importer/fetcher/ArXiv.java
Siedlerchr: Hm, seems like a formatting thing from Eclipse, when I saved the file. They were not intented by me.

Siedlerchr:
Yes, everything works fine without it. I checked some urlDownloads (e.g. Lookup Fulltex) and some other manual fetching things. I really have no idea, why it was added before. Maybe because as a workaround for some things in the past.

stefan-kolb:
Please remove the empty lines. other than that I'm happy with removing the Cookie-Impl. Let's see what the other think :+1:

PR #1781, opened 2016-08-18
Moved some logic preference instantiation to JabRefPreferences

oscargus:
Less passing of JabRefPreferences for better long term architecture.


oscargus:
The failing tests will be fixed once #1731 is merged as there is some relations there.

PR #1782, opened 2016-08-19
Removed one Globals call from MetaData

oscargus:
One further step towards a `Globals`-free `MetaData`.

### DIFF
--- a/src/main/java/net/sf/jabref/logic/bibtexkeypattern/BibtexKeyPatternPreferences.java
+++ b/src/main/java/net/sf/jabref/logic/bibtexkeypattern/BibtexKeyPatternPreferences.java
@@ -1,5 +1,6 @@
 package net.sf.jabref.logic.bibtexkeypattern;
 
+import net.sf.jabref.model.bibtexkeypattern.GlobalBibtexKeyPattern;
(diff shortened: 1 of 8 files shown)


src/main/java/net/sf/jabref/logic/bibtexkeypattern/BibtexKeyPatternPreferences.java
```diff
@@ -51,4 +56,9 @@ public boolean isEnforceLegalKey() {
     }
 
     public String getDefaultBibtexKeyPattern() { return defaultBibtexKeyPattern;}
+
+
```
stefan-kolb: Duplicate empty line. @koppor is this caused by our Eclipse Settings? I'm seeing that all the time.

stefan-kolb:
Minor Comments otherwise LGTM :+1:

src/main/java/net/sf/jabref/logic/bibtexkeypattern/BibtexKeyPatternPreferences.java
oscargus: This one is probably my bad, although I know sometimes Eclipse add double empty lines.

PR #1783, opened 2016-08-19
Fix NPE when closing the last tab

oscargus:
The (new?) focus feature lead to an NPE when closing the last tab since there was no table to focus...

- [ ] Change in CHANGELOG.md described - introduced in dev version


matthiasgeiger:
LGTM

PR #1787, opened 2016-08-19
Removed entry preview toolbar

oscargus:
Has been disabled since long so might as well remove the code.

Also cleaned up the preferences a bit.


stefan-kolb:
LGTM :+1:

PR #1789, opened 2016-08-20
Replace all file dialogs with new FileDialog

stefan-kolb:
and set default FileFilter or aggregate FileFilter for all supported types.

- [x] Database from AUX
- [x] Remove BrowseAction


Siedlerchr:
Why do you kill the BrowseActio? It already uses the new FileDialogs and is used in many places (because eseentially the dialogs are everywhere the same)! It should be necessary to only call the setDefaultExtension method there. Or to inlcude it.

stefan-kolb:
I just want to keep it simple and consistent. Either use Browse action everywhere or filedialog. I think for saving one or two lines its not worth keeping different classes here and there, but I'm open for suggestions.

stefan-kolb:
Additionally, the BrowseAction does not make it obvious or easy to understand what it really does without inspecting the BrowseAction class. I'm open for suggestions and a different decision here. Maybe we should discuss this in a Devcall or other Devs can quickly take a look at the code. In my opinion the small code duplication makes it easier to understand whats happening here and therefore might be preferable over the encapsulation inside BrowseAction.

Siedlerchr:
I gave the BroweAction a second thought and yes, it could be removed.

src/main/java/net/sf/jabref/gui/exporter/CustomExportDialog.java
```diff
@@ -104,7 +104,12 @@ public CustomExportDialog(final JabRefFrame parent) {
         cancel.addActionListener(e -> dispose());
 
         JButton browse = new JButton(Localization.lang("Browse"));
-        browse.addActionListener(BrowseAction.buildForFile(layoutFile, FileExtensions.LAYOUT));
+        FileDialog dialog = new FileDialog(null).withExtension(FileExtensions.LAYOUT);
```
Siedlerchr: I think you should pass the jFrame parent here instead if null

src/main/java/net/sf/jabref/gui/protectedterms/NewProtectedTermsFileDialog.java
```diff
@@ -49,7 +51,12 @@ public NewProtectedTermsFileDialog(JabRefFrame mainFrame, ProtectedTermsLoader l
 
     private void setupDialog() {
         JButton browse = new JButton(Localization.lang("Browse"));
-        browse.addActionListener(BrowseAction.buildForFile(newFile, FileExtensions.TERMS));
+        FileDialog dialog = new FileDialog(null).withExtension(FileExtensions.TERMS);
```
Siedlerchr: same as above?

stefan-kolb:
What we should discuss here is the state model of the FileDialog class. All methods alter the state of the internal JFileChooser dialog. But if we show a Directory selection dialog, we cannot revert back to a File selection dialog. Also if we use multiple files selection this is used in former single file selection dialogs...

Siedlerchr:
Regarding the state: I do not think that this causes a big problem here, as we create a new Object of the FileDialog including the underlying internal jFileChooser for each dialog where it is used.

The other point: Directories only is clear. And MultipleFiles vs Single Files should be clear, too. However, even in a muliple file dialog I am able to select only file.

stefan-kolb:
@Siedlerchr Yeah, but if you don't know this - it creates potential bugs. because not all state transitions are allowed as I explained.

Siedlerchr:
I know what you mean. You could introduce a` private boolean called = false` and check for it and throw an exception if it's true on entering any other method...

grimes2:
This state problem is also present in current master. So for example in database properties the encoding is set to the first entry Big5.

Siedlerchr:
@stefan-kolb  Take a look at #1803,  seems like the full path (including filename) is stored in the prefs.

stefan-kolb:
Is this a new One caused by this pr or was this the case before?

stefan-kolb:
@koppor @grimes2 
> This state problem is also present in current master. So for example in database properties the encoding is set to the first entry Big5.

I don't get this one. What doo you mean?

grimes2:
Sorry, my fault. I've no evidence, that the bug is caused also by this pull request. I've opened a new issue for this: #1806

PR #1793, opened 2016-08-20
Some improvements to LayoutFormatterPreferences

oscargus:
Some suggestions from by @koppor in #1695 which I didn't fully understand then. Also final Globals call in `logic`removed.

### DIFF
--- a/src/main/java/net/sf/jabref/logic/layout/LayoutEntry.java
+++ b/src/main/java/net/sf/jabref/logic/layout/LayoutEntry.java
@@ -585,14 +585,13 @@ class LayoutEntry {
             String className = strings.get(0).trim();
 
             // Check if this is a name formatter defined by this export filter:
-            if (!prefs.getCustomExportNameFormatters().isEmpty()) {
-                String contents = prefs.getCustomExportNameFormatters().get(className);
-                if (contents != null) {
-                    NameFormatter nf = new NameFormatter();
-                    nf.setParameter(contents);
-                    results.add(nf);
-                    continue;
-                }
+
+            String contents = prefs.getCustomExportNameFormatter(className);
+            if (contents != null) {
+                NameFormatter nf = new NameFormatter();
+                nf.setParameter(contents);
+                results.add(nf);
+                continue;
             }
 
             // Try to load from formatters in formatter folder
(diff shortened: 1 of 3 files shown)


src/main/java/net/sf/jabref/logic/layout/LayoutFormatterPreferences.java
```diff
@@ -45,15 +45,23 @@ public JournalAbbreviationPreferences getJournalAbbreviationPreferences() {
...
+
+    public void clearCustomExportNameFormatters() {
+        customExportNameFormatters.clear();
+    }
+
+    public void putCustomExportNameFormatter(String formatterName, String contents) {
+        customExportNameFormatters.put(formatterName, contents);
+    }
+
+    public String getCustomExportNameFormatter(String className) {
```
koppor: Should this parameter be called `formatterName`?

src/main/java/net/sf/jabref/logic/layout/LayoutEntry.java
```diff
@@ -585,14 +585,13 @@ private LayoutFormatter getLayoutFormatterByName(String name) throws Exception {
...
-                String contents = prefs.getCustomExportNameFormatters().get(className);
-                if (contents != null) {
-                    NameFormatter nf = new NameFormatter();
-                    nf.setParameter(contents);
-                    results.add(nf);
-                    continue;
-                }
+
+            String contents = prefs.getCustomExportNameFormatter(className);
+            if (contents != null) {
```
koppor: How much effort would it be to switch to `Optional` at `prefs.getCustomExportNameFormatter`?

oscargus: Not much. However, the `continue` prohibits the use of `ifPresent` so there will still be a `if (contents.isPresent())`. Still that is probably better.

src/main/java/net/sf/jabref/logic/layout/LayoutFormatterPreferences.java
oscargus: Yes. Will change in LayoutEntry as well (I just used the name from refactoring).

oscargus:
Fixed the comments.

PR #1794, opened 2016-08-20
Removed some warnings and findbugs issues

oscargus:
* Renaming variables
* Closing resources
* Checking for null
* Removed deprecated method
* Added annotations
* Removed unused code

### DIFF
--- a/src/main/java/net/sf/jabref/cli/CrossrefFetcherEvaluator.java
+++ b/src/main/java/net/sf/jabref/cli/CrossrefFetcherEvaluator.java
@@ -24,67 +24,67 @@ import net.sf.jabref.preferences.JabRefPreferences;
  * Useful for checking out new algorithm improvements and thresholds. Not used inside the JabRef code itself.
  */
 public class CrossrefFetcherEvaluator {
+
     public static void main(String[] args) throws IOException, InterruptedException {
         Globals.prefs = JabRefPreferences.getInstance();
+        try (FileReader reader = new FileReader(args[0])) {
+            BibtexParser parser = new BibtexParser(reader, ImportFormatPreferences.fromPreferences(Globals.prefs));
+            ParserResult result = parser.parse();
+            BibDatabase db = result.getDatabase();
 
-        BibtexParser parser = new BibtexParser(new FileReader(args[0]),
-                ImportFormatPreferences.fromPreferences(Globals.prefs));
-        ParserResult result = parser.parse();
-        BibDatabase db = result.getDatabase();
+            int total = result.getDatabase().getEntryCount();
 
-        int total = result.getDatabase().getEntryCount();
+            AtomicInteger dois = new AtomicInteger();
+            AtomicInteger doiFound = new AtomicInteger();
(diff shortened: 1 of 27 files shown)


src/main/java/net/sf/jabref/cli/CrossrefFetcherEvaluator.java
```diff
@@ -24,67 +24,67 @@
  * Useful for checking out new algorithm improvements and thresholds. Not used inside the JabRef code itself.
  */
 public class CrossrefFetcherEvaluator {
+
     public static void main(String[] args) throws IOException, InterruptedException {
         Globals.prefs = JabRefPreferences.getInstance();
+        try (FileReader reader = new FileReader(args[0])) {
```
oscargus: This is the actual change here, the rest is formatting.

src/main/java/net/sf/jabref/gui/JabRefFrame.java
```diff
@@ -1569,10 +1569,8 @@ public void updateEnabledState() {
 
         if (tabCount > 0) {
             BasePanel current = getCurrentBasePanel();
-            if(current != null) {
```
oscargus: never null since tabCount > 0

src/main/java/net/sf/jabref/gui/desktop/JabRefDesktop.java
```diff
@@ -316,26 +316,20 @@ public static void openConsole(File file) throws IOException {
...
-                    commandLoggingText.append(subcommands[i]);
-                    if (i < (subcommands.length - 1)) {
-                        commandLoggingText.append(" ");
-                    }
-                }
 
-                JabRefGUI.getMainFrame().output(Localization.lang("Executing command \"%0\"...", commandLoggingText.toString()));
-                LOGGER.info("Executing command \"" + commandLoggingText.toString() + "\"...");
+                // replace the placeholder if used
+                String commandLoggingText = command.replace("%DIR", absolutePath);
```
oscargus: Should be enough to replace on string level rather than each space-separated sub-string.

src/main/java/net/sf/jabref/gui/maintable/MainTableDataModel.java
```diff
@@ -193,6 +193,8 @@ public void updateGroupFilter() {
 
         private final List<Comparator<BibEntry>> comparators;
 
+
+        @SafeVarargs
```
Siedlerchr: Is this attribute really needed (I guess it complained about unsafe operation, but I do not see any unsafe operation here...) It may go away if you use Arrays.stream() instead of wrapping it in a new list before.

Siedlerchr:
Good work! :+1:  Just one little thing.

src/main/java/net/sf/jabref/gui/maintable/MainTableDataModel.java
oscargus: As I understand it, the attribute is a way of the developer saying that there are no unsafe operations, rather than the checker claiming that there is. So it doesn't really matter what the code is. But I may be wrong.

PR #1800, opened 2016-08-20
Converted a number of deprecated method calls

oscargus:

### DIFF
--- a/src/main/java/net/sf/jabref/gui/groups/TransferableEntrySelection.java
+++ b/src/main/java/net/sf/jabref/gui/groups/TransferableEntrySelection.java
@@ -6,6 +6,7 @@ import java.awt.datatransfer.UnsupportedFlavorException;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.util.List;
+import java.util.Optional;
 import java.util.stream.Collectors;
 
 import net.sf.jabref.model.entry.BibEntry;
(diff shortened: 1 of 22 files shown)


src/main/java/net/sf/jabref/gui/groups/TransferableEntrySelection.java
```diff
@@ -39,7 +39,8 @@
     public TransferableEntrySelection(List<BibEntry> list) {
         this.selectedEntries = list;
         selectedEntriesCiteKeys = String.join(",",
-                this.selectedEntries.stream().map(BibEntry::getCiteKey).collect(Collectors.toList()));
+                this.selectedEntries.stream().map(BibEntry::getCiteKeyOptional).filter(key -> key.isPresent())
```
Siedlerchr: The filter and second map is not necessary, it should workt with   map(BibEntry::getCiteKeyOptional).orElse().collect...

Siedlerchr:
LGTM :+1:

src/main/java/net/sf/jabref/gui/groups/TransferableEntrySelection.java
oscargus: Obviously! I remember thinking that this was a weird call... One shouldn't write code tired...

oscargus: I agree my code looks weird, but I cannot figure it out. orElse won't work since we are operating on a stream (which happens to be a stream of optional). Based on the example here one can write it slightly more compact though. http://stackoverflow.com/questions/24254209/using-optional-with-streams-in-java

Siedlerchr: I was not fully sure either. But I think that's one of the current limits of the optional stream and ist listed for jdk9...
http://stackoverflow.com/questions/22725537/using-java-8s-optional-with-streamflatmap

Issue #1664, opened 2016-08-02
Auto-create groups from AUX file


stefan-kolb:
Create groups of papers cited in a paper via an aux-file

Mixopteryx:
The same as "Tools --> New Subdatabase based on AUX file", but output a group instead of a new .bib file.

tobiasdiez:
@stefan-kolb @Mixopteryx @oscargus @koppor do you guys actually prefer the aux file? I would like to work on this but my approach would be to directly parse the tex-file and look for `\cite` commands and its relatives. Reason: the tex file is always up-to-date while I need to recompile to get a correct aux file (also the location of the aux file may change when compiled with the `-output-directory` option or be deleted by a "cleanup of temporary files" action of a text editor).

stefan-kolb:
Just a few quik thoughts:
1. The aux file is smaller and already has the necessary cites aggregated
2. You need to search all folders for multiple tex files
3. Tex files might not be included in the main document that is actually compiled
4. Cites might be commented out or something in the code

koppor:
citet, citep (natbib), citeN (elsart) -- no additional latex parser in
JabRef. There is a ten year old existing implementation in Java, but too
slow - maybe 😂😂

Am 07.11.2017 13:53 schrieb "Stefan Kolb" <notifications@github.com>:

Mixopteryx:
In principle using the .tex file would be better, but using the .aux file is much more robust and simpler to make.

I'd rather use the aux file and recompile than ensuring that my tex file is supported by a parser. In addition to what @stefan-kolb mention I sometimes define \newcommands that include \cite commands, and a I expect a parser that would catch that need to be awfully complicated compared to the aux parser.

tobiasdiez:
This is now implemented in the latest development version. Could you please check the build from http://builds.jabref.org/master/. Thanks!
The user interface is a bit spartanic (since we plan to rewrite the groups dialog) but it should work if you paste the full path to the aux-file in the textbox.

PR #1801, opened 2016-08-21
[WIP] Auto-create groups from AUX file #1664

stefan-kolb:
- [ ] Aux File listener?
- [ ] introduce new paper folder?
- [ ] Look for all aux files in paper folder
- [ ] create icon inside groups menu/groups sidepane or under tools?


src/main/java/net/sf/jabref/gui/auximport/FromAuxGroups.java
```diff
@@ -0,0 +1,291 @@
...
+            return group;
+        } catch  (ParseException e) {
+            return null;
+        }
+    }
+
+    /**
+     * Action used to produce a "Browse" button for one of the text fields.
+     */
+    static class BrowseAction extends AbstractAction {
```
Siedlerchr: I think you can get rid of the Action Class and just use the actionPerformed method.

stefan-kolb: Yeah this is just an old artifact from the master branch. WIP right now!

PR #1809, opened 2016-08-22
Fix preview update when selecting single entry after selecting multiple entries

### LINKED ISSUE #1760: Preview not updated properly
JabRef 3.6dev--snapshot--2016-07-29--fix1639--e838541
windows 10 10.0 amd64 
Java 1.8.0_101

Steps to reproduce:

1. Select an entry. The entry preview (correctly) appears at the bottom of Jabref.
2. Now select several entries, e. g. the first two entries below the entry you had intially selected. I've done this by holding SHIFT and clicking onto the last entry I want to select.
3. Three entries are now selected. The entry preview of the initially selected entry is still visible. So far, so good.
4. Now select another entry among the three entries that you selected beforehand. This is now the only entry that is selected (indicated in "blue" by Jabref). The entry preview, however, is not updated properly. It still shows the entry preview of the entry that you had selected at the start.
5. If you select now another entry and then go back to the entry selected in 4), the entry preview will be updated correctly.

bartsch-dev:
Fixes #1760.

Preview is now updated correctly when selecting a single entry after selecting multiple entries.

### DIFF
--- a/src/main/java/net/sf/jabref/gui/maintable/MainTableSelectionListener.java
+++ b/src/main/java/net/sf/jabref/gui/maintable/MainTableSelectionListener.java
@@ -105,24 +105,14 @@ public class MainTableSelectionListener implements ListEventListener<BibEntry>, 
         if (!enabled) {
             return;
         }
+
         EventList<BibEntry> selected = e.getSourceList();
-        BibEntry newSelected = null;
-        while (e.next()) {
-            if (e.getType() == ListEvent.INSERT) {
-                if (newSelected == null) {
-                    if (e.getIndex() < selected.size()) {
-                        newSelected = selected.get(e.getIndex());
-                    }
-                } else {
-                    return; // More than one new selected. Do nothing.
-                }
-            }
+        if (selected.isEmpty()){
+            return;
         }
 
+        final BibEntry newSelected = selected.get(0);
         if (newSelected != null) {
-
-            // Ok, we have a single new entry that has been selected. Now decide what to do with it:
-            final BibEntry toShow = newSelected;
             final BasePanelMode mode = panel.getMode(); // What is the panel already showing?
             if ((mode == BasePanelMode.WILL_SHOW_EDITOR) || (mode == BasePanelMode.SHOWING_EDITOR)) {
                 // An entry is currently being edited.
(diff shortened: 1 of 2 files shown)


src/main/java/net/sf/jabref/gui/maintable/MainTableSelectionListener.java
```diff
@@ -105,25 +105,16 @@ public void listChanged(ListEvent<BibEntry> e) {
...
+            return; //None or more than one new Selected. Do Nothing.
         }
 
+        final BibEntry newSelected = selected.get(0);
         if (newSelected != null) {
-
-            // Ok, we have a single new entry that has been selected. Now decide what to do with it:
-            final BibEntry toShow = newSelected;
             final BasePanelMode mode = panel.getMode(); // What is the panel already showing?
+
```
Braunch: Why did this change?

bartsch-dev: To increase the readability.

src/main/java/net/sf/jabref/gui/maintable/MainTableSelectionListener.java
```diff
@@ -105,25 +105,16 @@ public void listChanged(ListEvent<BibEntry> e) {
         if (!enabled) {
             return;
         }
+
         EventList<BibEntry> selected = e.getSourceList();
-        BibEntry newSelected = null;
-        while (e.next()) {
-            if (e.getType() == ListEvent.INSERT) {
```
oscargus: What happens if an entry is inserted? I have no idea how the code works, but it seems like this is some sort of handling for inserted entries?

bartsch-dev: `ListEvent.INSERT` means that the entry was inserted into the selection (thus selected by the user), and there lies the problem.
When multiple entries were selected and you select an single entry inside that selection there are only `ListEvent.DELETE`-events, thus JabRef doesn't recognize the change.

oscargus: OK. Thanks for clarifying! :+1:

oscargus:
LGTM! :+1:

PR #1812, opened 2016-08-22
Fix #1803 Working directory is set to full path including filename

### LINKED ISSUE #1803: Working directory is set to full path including filename
JabRef 3.5
windows 10 10.0 amd64 
Java 1.8.0_101

1. File > Open database, open a bib file
2. Show preferences, string working directory
3. string is a full path (with bib file) not a directory

"Save database as" works correct.

stefan-kolb:


oscargus:
LGTM! :+1:

PR #1814, opened 2016-08-22
Made output of BibDatabase.getPreamble() Optional

oscargus:
The code was/is a bit confusing since both null and "" preamble are dealt with in the same way (not saved).

### DIFF
--- a/src/main/java/net/sf/jabref/collab/ChangeScanner.java
+++ b/src/main/java/net/sf/jabref/collab/ChangeScanner.java
@@ -51,13 +51,14 @@ public class ChangeScanner implements Runnable {
 
     private static final String[] SORT_BY = new String[] {FieldName.YEAR, FieldName.AUTHOR, FieldName.TITLE};
 
-    private final File f;
-    private final BibDatabase inMem;
-    private final MetaData mdInMem;
+    private final File file;
+    private final BibDatabase databaseInMemory;
+    private final MetaData metadataInMemory;
     private final BasePanel panel;
 
     private final JabRefFrame frame;
-    private BibDatabase inTemp;
+    private BibDatabase databaseInTemp;
+    private MetaData metadataInTemp;
 
     private MetaData mdInTemp;
 
@@ -75,9 +76,9 @@ public class ChangeScanner implements Runnable {
     public ChangeScanner(JabRefFrame frame, BasePanel bp, File file) {
         this.panel = bp;
         this.frame = frame;
-        this.inMem = bp.getDatabase();
-        this.mdInMem = bp.getBibDatabaseContext().getMetaData();
-        this.f = file;
+        this.databaseInMemory = bp.getDatabase();
+        this.metadataInMemory = bp.getBibDatabaseContext().getMetaData();
+        this.file = file;
     }
 
     @Override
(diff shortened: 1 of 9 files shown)


stefan-kolb:
Can you please add a test with missing preamble?

oscargus:
Sure.

src/main/java/net/sf/jabref/collab/ChangeScanner.java
```diff
@@ -344,16 +344,14 @@ private static BibEntry bestFit(EntrySorter old, EntrySorter neu, int index) {
...
-        String tmp = onTmp.getPreamble();
-        String disk = onDisk.getPreamble();
-        if (tmp == null) {
-            if ((disk != null) && !disk.isEmpty()) {
-                changes.add(new PreambleChange(mem, disk));
-            }
+        String mem = inMem1.getPreamble().orElse(null);
+        Optional<String> tmp = onTmp.getPreamble();
+        Optional<String> disk = onDisk.getPreamble();
+        if (!tmp.isPresent()) {
```
oscargus: If someone please could double check this rewrite. It is quite confusing sometimes to change the null vs isPresent...

src/main/java/net/sf/jabref/collab/ChangeScanner.java
```diff
@@ -344,16 +344,14 @@ private static BibEntry bestFit(EntrySorter old, EntrySorter neu, int index) {
...
+            disk.filter(diskContent -> !diskContent.isEmpty()).ifPresent(diskContent -> changes.add(new PreambleChange(mem, diskContent)));
```
Siedlerchr: Is it okay that mem can be null?

src/main/java/net/sf/jabref/gui/PreambleEditor.java
```diff
@@ -153,15 +151,14 @@ public void actionPerformed(ActionEvent e) {
             // We check if the field has changed, since we don't want to mark the
```
oscargus: This logic I will rewrite.

src/main/java/net/sf/jabref/collab/ChangeScanner.java
Siedlerchr: LGTM :+1:  I see no wrong things. 
You could just rename the mem to make it a bit clearer what it is purpose/what is stored in it,
I can only guess: Mem=Memory?

oscargus: Yeah, I can only guess as well. `disk` is quite clear, but the difference between `mem` and `tmp` is not obvious...

src/main/java/net/sf/jabref/collab/ChangeScanner.java
oscargus: Yes. It is only used for the undo to indicate that there was no content earlier.

oscargus:
Test added and some general renaming of variables in `ChangeScanner`.

src/test/java/net/sf/jabref/logic/importer/fileformat/BibtexParserTest.java
```diff
@@ -935,39 +935,39 @@ public void parseRecognizesPreamble() throws IOException {
...
         ParserResult result = BibtexParser.parse(new StringReader("@preamble(some text and \\latex)"),
                 importFormatPreferences);
-        assertEquals("some text and \\latex", result.getDatabase().getPreamble());
+        assertEquals(Optional.of("some text and \\latex"), result.getDatabase().getPreamble());
     }
 
     @Test
     public void parseRecognizesPreambleWithConcatenation() throws IOException {
 
         ParserResult result = BibtexParser.parse(new StringReader("@preamble{\"some text\" # \"and \\latex\"}"),
```
stefan-kolb: What happens for, e.g., `new StringReader("@preamble{}` and `new StringReader(""}`

oscargus: That may indeed be more relevant things to test. :-)

oscargus: I added those tests and they returned as expected. However, neither now or before will the saved file contain the empty preamble. Hence, it does not really matter if the preamble is "" or null as it will be "" when shown in the dialog and null when saved (i.e. not saved).

stefan-kolb:
Can you please rebase this and merge it in afterwards? :+1:

Issue #1235, opened 2016-04-19
Modified Key bindings do not work correctly


matthiasgeiger:
JabRef version 3.2, 3.3 on Windows 10 with German Keyboard

Steps to reproduce:
  1. Default Key-Binding for "Delete Entry" is `DELETE`
  2. Change key-binding for "Delete Entry" using the "grab" functionality and hit `Entf` key on a german keyboard. Save changes and restart jabref.
  3. Select and entry in the main table and hit "Entf" key
  4. nothing happens...

simonharrer:
@matthiasgeiger will you have a look here? Otherwise, move it to the milestrone 3.5

matthiasgeiger:
I don't think I'll have to time to investigate this week... so: moved to 3.5

stefan-kolb:
@koppor Did StuPro look at this when they reimplemented the hotkeys?

koppor:
@mairdl and @Braunch did not look at it during #1525 and do not know the reason why this causes issues. There, however, seems to be https://github.com/mairdl/jabref/pull/2, but I think, this code is lost somewhere in the git universe. @Braunch Can you have a look?

Braunch:
Fixed by #1815

PR #1815, opened 2016-08-22
Fix delete entry keybindings bug

Braunch:
When using a different localization then English, setting the keybindings does not work correctly (see here #1235 ). I figured out that the problem is that the `KeyEvent` returns localized key texts so you can not use this to set the `KeyBindings`(the JabRef `KeyBindings` use English key texts). At the moment I only see two possibilities that do not end up in comparing tons of Strings to manually get the english localization. 

~~First would be to change the locale to English, so that the `KeyEvent` returns english key texts, but I don't know what I might break with that and if the rest of the localization will still work.~~

 ~~Second solution would be to change the `KeyBindings` so that they use `KeyCodes` instead of Strings. This would IMHO improve the `KeyBindings` system but it would mean a lot of changes in all the `KeyBinding` related code.~~
See the comment below for my approach https://github.com/JabRef/jabref/pull/1815#issuecomment-242346564.


src/test/resources/testbib/testjabref_292.bib
```diff
@@ -299,3 +299,4 @@ @siumed.edu
   url = {http://dx.doi.org/10.1016/j.jneumeth.2008.01.009}
 }
 
+@Comment{jabref-meta: databaseType:bibtex;}
```
koppor: This is JabRef 3.x - do not change bib files created with old versions of JabRef

Braunch:
I now found a way to fix it. I am building a String with the English modifier key text using the triggered KeyEvents  `is-ModifierKey-Down` methods

CHANGELOG.md
```diff
@@ -77,6 +77,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - Fixed a number of issues related to accessing the GUI from outside the EDT
 - Added a few missing translation strings
 - Fixed NullPointerException when opening Customize entry type dialog without an open database
+- Fixed localization error when changing key bindings while not using the english localization
```
boceckts: Instead of "Fixed localization error when changing key bindings ..." I would rather write "Fixed modified keybindings not working ..."

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingsListener.java
```diff
@@ -2,65 +2,111 @@
...
     }
+
+    /**
+     * Checks if the event has pressed modifiers and adds their English key text to an ArrayList
+     * from which a String is build
+     *
+     * @param evt the KeyEvent that was triggered to set the KeyBindings
+     * @return a String containing the modifier keys text
+     */
+    private String getModifierTextAsList(KeyEvent evt) {
```
boceckts: Rename without "AsList" or "AsListing" as it doesn't return a list.

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingsListener.java
```diff
@@ -2,65 +2,111 @@
...
+        }
+        //Now build the String with all the modifier texts
+        String modifiersAsString = "";
+        for (String s : modifiersList) {
+            //add a spacing if its not the first modifier to be added
+            if (!"".equals(modifiersAsString)) {
+                modifiersAsString += " ";
+            }
+            modifiersAsString += s;
+        }
```
boceckts: This could be replaced by one line by using `modifiersList.stream.collect`

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingsListener.java
```diff
@@ -2,65 +2,111 @@
...
+     * Checks if the event has pressed modifiers and adds their English key text to an ArrayList
+     * from which a String is build
+     *
+     * @param evt the KeyEvent that was triggered to set the KeyBindings
+     * @return a String containing the modifier keys text
+     */
+    private String getModifierTextAsList(KeyEvent evt) {
+        ArrayList<String> modifiersList = new ArrayList<>();
+
+        if (evt.getModifiers() != 0) {
```
boceckts: Why do you need this? I think `evt.isControlDown()` will return `false` anyway.

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingsListener.java
```diff
@@ -2,65 +2,111 @@
...
+            if (evt.isAltDown()) {
+                modifiersList.add("alt");
+            }
+            if (evt.isShiftDown()) {
+                modifiersList.add("shift");
+            }
+            if (evt.isAltGraphDown()) {
+                modifiersList.add("alt gr");
+            }
+        }
```
boceckts: I'm missing the meta key, which is especially for mac users an important modifier key. The default shortcuts of JabRef also include this modifier, so please add it.

boceckts:
Please resolve conflicts and minor comments, then it's good to me 👍

Braunch:
@boceckts can you test it on win and mac for me please? The meta key does not work on my linux system since there are OS functions on it.

boceckts:
@Braunch Windows uses the windows key for its own shortcuts, same as with some linux distributions. On mac it now does work to make shortcuts directly with the command(meta) key....but I just realized that if you create a shortcut with the control key, e.g. "ctrl + S" it will automatically be replaced by "command(meta) + S".
So I guess the meta key doesn't have to be listened to..sry for that.

Braunch:
Is this intended? Why should the grabbed keys differ from the keys actually used by the user? I changed the behavior so that you can set the 'meta + key' shortcut with the meta. If this is alright I am probably done here.

Braunch:
Any opinion on my solution with the meta key from the Developers?

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingsListener.java
```diff
@@ -2,65 +2,107 @@
...
+import java.util.Locale;
+import java.util.stream.Collectors;
 
 /**
  * respond to grabKey and display the key binding
  */
 public class KeyBindingsListener extends KeyAdapter {
 
     private final KeyBindingTable table;
+    private boolean isDeleteKey;
```
koppor: You don't need to make them class variables, do you?

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingsListener.java
```diff
@@ -2,65 +2,107 @@
...
+            } else if (isEscapeKey) {
+                newKey = "ESCAPE";
+            } else if (isDeleteKey) {
+                newKey = "DELETE";
+            } else {
+                return;
+            }
         } else {
-            newKey = modifier.toLowerCase().replace("+", " ") + " " + code;
+            newKey = modifier.toLowerCase(Locale.ENGLISH) + " " + KeyEvent.getKeyText(code);
```
koppor: Why is the replacement of "+" to " " no longer needed?

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingsListener.java
```diff
@@ -2,65 +2,107 @@
...
+            modifiersList.add("ctrl");
+        }
+        if (evt.isAltDown()) {
+            modifiersList.add("alt");
+        }
+        if (evt.isShiftDown()) {
+            modifiersList.add("shift");
+        }
+        if (evt.isAltGraphDown()) {
+            modifiersList.add("alt gr");
```
koppor: Is that true? altgr without space?

Braunch: On linux 'alt gr' works just fine. On my windows machine at home, 'ctrl alt' is set instead of 'alt gr' but no matter if it's written with or without space in between. I tested it on the master and 'alt gr' is always parsed as 'ctrl alt' so it works as intended I guess.

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingsListener.java
koppor: Still unanswered. :innocent:

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingsListener.java
```diff
@@ -2,65 +2,110 @@
...
+        final String modifier = getModifierText(evt);
 
         // VALIDATE code and modifier
         // all key bindings must have a modifier: ctrl alt etc
         if ("".equals(modifier)) {
             int kc = evt.getKeyCode();
-            boolean isFunctionKey = (kc >= KeyEvent.VK_F1) && (kc <= KeyEvent.VK_F12);
-            boolean isEscapeKey = kc == KeyEvent.VK_ESCAPE;
-            boolean isDeleteKey = kc == KeyEvent.VK_DELETE;
+            isFunctionKey = (kc >= KeyEvent.VK_F1) && (kc <= KeyEvent.VK_F12);
```
koppor: why not keeping the delcaration locally? I don't see any usage outside the if statement.

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingsListener.java
```diff
@@ -2,65 +2,110 @@
...
+            isEscapeKey = kc == KeyEvent.VK_ESCAPE;
+            isDeleteKey = kc == KeyEvent.VK_DELETE;
             if (!(isFunctionKey || isEscapeKey || isDeleteKey)) {
                 return; // need a modifier except for function, escape and delete keys
```
koppor: Move comment form after the return to the line above

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingsListener.java
```diff
@@ -2,65 +2,110 @@
...
-                || "Ctrl".equals(code)
-                || "Shift".equals(code)
-                || "Alt".equals(code)) {
+        int code = evt.getKeyCode();
+        String newKey;
+        //skip the event triggered only by a modifier
+        if (code == KeyEvent.VK_ALT ||
+                code == KeyEvent.VK_TAB ||
+                code == KeyEvent.VK_BACK_SPACE ||
+                code == KeyEvent.VK_ENTER ||
```
koppor: Could you comment why `ENTER` is treated as modifier? Also applies to "TAB" and "BACK_SPACE". Maybe the reason is that these keys should not be used at shortcuts at all? Maybe a new `if` statement should be introduced with an appropriate comment.

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingsListener.java
```diff
@@ -2,65 +2,110 @@
...
-        // SHOW new key binding
+        //SHOW new key binding
         //find which key is selected and set its value
         table.setValueAt(newKey, selRow, 1);
         table.revalidate();
         table.repaint();
     }
+
+    /**
+     * Checks if the event has pressed modifiers and adds their English key text to an ArrayList
```
koppor: The internal implementation should not exposed. Formulate it in an abstract way:

> Collects the English translations of all modifiers and returns a space separated list

koppor:
Please merge `upstream/mater` and fix minor comments. Then, it should be good to go.

Braunch:
I addressed the small comments and resolved the conflicts.

koppor:
You commited with both your st university address and your normal development address. Please double check your git config.

Please merge upstream/master again. Can you make one commit as described at https://github.com/JabRef/jabref/wiki/Tools#rebase-everything-as-one-commit-on-master? Then, I can resolve the conflicts by myself.

Braunch:
I resolved the conflicts and squashed my commits. Should be good to go.

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingsListener.java
koppor: BUMP

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingsListener.java
koppor: BUMP

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingsListener.java
koppor: BUMP

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingsListener.java
koppor: BUMP

koppor:
Please do not squash, this makes reviewing very hard as we cannot see, what has been adressed. I added a `BUMP` on all my unaswered comments. Please fix/answer

Braunch:
I addressed your comments and merged the upstream/master. I squashed following your comment https://github.com/JabRef/jabref/pull/1815#issuecomment-246499706 Did I get that wrong?

CHANGELOG.md
```diff
@@ -45,12 +45,13 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - Selecting an entry in the search result Window will now select the correct entry in the bib file
 - Entries in the SearchResultDialog are now converted to Unicode
 - Suggestions in the autocomplete (search) are now in Unicode
-- Fixed NullPointerException when opening search result window for an untitled database 
+- Fixed NullPointerException when opening search result window for an untitled database
```
Braunch: I encountered an unnecessary space so i deleted it

PR #1816, opened 2016-08-22
When inserting a duplicate the right entry will be selected

bartsch-dev:
When you insert a duplicate in your database the editor has the wrong entry linked.

### Steps to reproduce
- select an entry
- copy and paste it (`ctrl+c` & `ctrl+v`; leave the old entry selected)
- now the entry editor is open, edit any value and you'll see that the lower duplicate has changed although the upper one is selected

### DIFF
--- a/src/main/java/net/sf/jabref/gui/BasePanel.java
+++ b/src/main/java/net/sf/jabref/gui/BasePanel.java
@@ -845,7 +845,9 @@ public class BasePanel extends JPanel implements ClipboardOwner, FileUpdateListe
             if (Globals.prefs.getBoolean(JabRefPreferences.AUTO_OPEN_FORM)) {
                 selectionListener.editSignalled(firstBE);
             }
-            highlightEntry(firstBE);
+
+//            If we inserted a duplicate we want to select the duplicate
+            highlightLastEntry(firstBE);
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/main/java/net/sf/jabref/gui/BasePanel.java
```diff
@@ -1840,17 +1842,6 @@ private synchronized void markChangedOrUnChanged() {
         frame.setWindowTitle();
     }
 
-    /**
-     * Selects a single entry, and scrolls the table to center it.
-     *
-     * @param pos Current position of entry to select.
-     */
-    public void selectSingleEntry(int pos) {
```
Braunch: Did I get it right that this was just an unsafe version of `public void highlightEntry(final BibEntry be)` ?

src/main/java/net/sf/jabref/gui/BasePanel.java
```diff
@@ -1721,6 +1721,8 @@ public void hideBottomComponent() {
     public void highlightEntry(final BibEntry be) {
```
Braunch: Maybe rename` be` to `bibEntry`. I dont see why you should not call it what it is.

src/main/java/net/sf/jabref/gui/BasePanel.java
bartsch-dev: nope, this was an unused method which did almost the same as `public void highlightEntry(final BibEntry be)` (accepting a position instead of a `BibEntry`

src/main/java/net/sf/jabref/gui/BasePanel.java
bartsch-dev: fixed

src/main/java/net/sf/jabref/gui/BasePanel.java
```diff
@@ -845,7 +845,9 @@ private void paste() {
             if (Globals.prefs.getBoolean(JabRefPreferences.AUTO_OPEN_FORM)) {
                 selectionListener.editSignalled(firstBE);
             }
-            highlightEntry(firstBE);
+
+//            If we inserted a duplicate we want to select the duplicate
```
stefan-kolb: Please use correct indent.

bartsch-dev: fixed

koppor:
I tried it out, but I would expect that after <kbd>CTRL</kbd>+<kbd>V</kbd>, the pasted entry would be selected, not the copied one.

bartsch-dev:
Did your run include 01a5a3e?
After pasting an entry I'm searching from the back, thus I should select the pasted duplicate.

PR #1817, opened 2016-08-22
Resolves #1798 Remove unused DragAndDrop Handlers and URL extra field…

### LINKED ISSUE #1798: Reference to old "pdfDirectory" in the code
In `UrlDragDrop` there is the following piece of code:

``` java 
                        //auto filename:
                        File file = new File(new File(Globals.prefs.get("pdfDirectory")),
                                editor.getEntry().getCiteKey() + ".pdf");
                        frame.output(Localization.lang("Downloading..."));
```

This should be changed to use, if nothing else, the correct directory and preferably the correct file name.

stefan-kolb:
@JabRef/developers Seems to be unused and can be removed?! if we want such functionality we have to recode it in a proper fashion anyway?!


src/main/java/net/sf/jabref/model/entry/FieldProperties.java
```diff
@@ -18,7 +16,6 @@
     INTEGER,
     GENDER,
     LANGUAGE,
-    LANG_ID,
```
oscargus: This will return, but not hard to add. :-)

oscargus:
I think it should be OK to remove it. It would be simple to just add FieldProperties.URL to the URL field, but I tried it and do not really see the benefit. I guess the idea is to only accept urls, but there are other drop listeners accepting the url as text anyway (if I understand things, makes no obvious difference anyway).

:+1:

stefan-kolb:
@koppor Any thoughts about this legacy functionality?

matthiasgeiger:
As it is not working - and the workflow for downloading files is now somehow different I also vote for removing this...

PR #1821, opened 2016-08-23
Added links to internet resources

oscargus:
Added links to home page, blog, and Facebook.
capture20

Not obvious where to put it in the menu nor if the "JabRef on the internet" is a good name (probably not, but cannot really find anything better).

I didn't bother adding the strings so far as I expect at least a few of them to change. :-)


matthiasgeiger:
How about adding those links to the "About JabRef" dialog?

We already have to much menu entries, imho. :wink:

mlep:
I would say the same as @matthiasgeiger about:
- "Online Help Forum"  (since this is a secondary mean of finding help, as mentioned in http://help.jabref.org/en/ )
- "Fork me on GitHub" (if you are a decent programmer, you do not need this item to start contributing code)

oscargus:
I was also thinking about it (main reason to use a sub menu). What about moving at least the GitHub link to the same sub menu? The forum link may be worth to keep.

oscargus:
Or rename it "JabRef resources" and/or add the forum link there as well.

oscargus:
Considering the translation effort in #1824 it may not be worth trying to get this into 3.6 (avoid adding more new strings at the moment), so maybe better to wait and discuss it a bit further.

matthiasgeiger:
I still think that the other links are more important/relevant and could remain in the menu:
It is the "Help" menu - so linking to the online forum to get help is reasonable.
Showing possibilities to contribute/donate is also clearly important for us.

The web page, the blog, facebook are all rather unimportant links for the user - and for us. But if we decide to add a "JabRef resources" sub menu I would at least add the github and donate link there (icons are already shown in the toolbar) and place the submenu between "Check for updates" and "About JabRef".

And if we should reach some concensus it should be also possible to include this in 3.6 as @mlep, you and me could provide translations for the three language with the best translation status ;-)

mlep:
I am fine with the proposal of @matthiasgeiger

oscargus:
OK! I'll have a go with that then. Just thought that 13 or whatever followers on Facebook didn't seem enough. :-)

oscargus:
New menu:

capture21

Siedlerchr:
Looks cool! I would move Donate more upwards ;)
And I would rename Web page to Website.  Apart from that :heart:

oscargus:
I changed to website. Regarding the ordering I had an idea to introduce a few more items and also separators:

Website
Blog
Facebook
---
Github
Developer version (-> builds.jabref.org/master)
Change log (-> Github master version)
---
Donate

In this way the donate link will get more attention, but I didn't want to add the other two without discussion.

matthiasgeiger:
Okay - as we already introduced the submenu we can also add the other links ;-)

Instead of "Developer version", which sounds like "a version specific for developers" I propose to use "Development version"; GitHub instead of "Github" :wink:

German translations:
```
Website=Webseite
Blog=Blog
Development_version=Entwicklungsversion
JabRef_resources=Mehr_zu_JabRef
```

The last one is rather hard to translate, imho... Perhaps another native speaker can come up with a better translation...

oscargus:
I was actually not planning on changing the "Fork me..." (although I would like to), so do not worry about "Github". ;-)

I agree with you regarding Developer/Development. Just couldn't figure out the better wording.

oscargus:
capture23

oscargus:
A bit annoying that the change log is shown with the development information as well. Is it possible to show only the most recent change log in some way?

matthiasgeiger:
I'm already trying to figur this out ;-)

oscargus:
I think "Mehr zu JabRef" makes sense as we might just as well have called it "More on JabRef".

matthiasgeiger:
I have not found a way to link the changelog of the latest release.. There is only the shortcut https://github.com/JabRef/jabref/releases/latest 

So we either link to the `master` CHANGELOG or we have to put some more effort in the link creation logic by determining the current JabRef version and adding it to `https://github.com/JabRef/jabref/blob/v**versionNumberHere**/CHANGELOG.md`

oscargus:
I'd say go with the current solution and eventually script a copy of the
ChangeLog to jabref.org/ChangeLog.md (or something) on merging to master.

mlep:
Maybe we need to be able to link to the sections of the master changelog.md file
Related to https://github.com/JabRef/www.jabref.org/issues/24 ?

Siedlerchr:
@matthiasgeiger  I would propose "Entwicklerversion" for Development Version as translation

matthiasgeiger:
No :-) See the discussion above: It is not a "version for developers" (==Entwicklerversion) but a version which is currently under development (==Entwicklungsversion) :stuck_out_tongue_winking_eye:

oscargus:
@mlep My plan was really just to link to the current CHANGELOG.md, so maybe the formulation is misleading (although correct as you can see all changes in all versions there).

mlep:
My misunderstanding, sorry.

oscargus:
@mlep No problems. I fully understand it (and was thinking of this possible misunderstanding when I wrote it, seems like I was correct...).

src/main/java/net/sf/jabref/gui/JabRefFrame.java
```diff
@@ -206,6 +206,21 @@
...
+            "Facebook", Localization.lang("Opens JabRef's Facebook page"),
+            IconTheme.JabRefIcon.FACEBOOK.getSmallIcon(), IconTheme.JabRefIcon.FACEBOOK.getIcon());
+    private final AbstractAction jabrefBlogAction = new OpenBrowserAction("https://blog.jabref.org/",
+            Localization.menuTitle("Blog"), Localization.lang("Opens JabRef's blog"),
+            IconTheme.JabRefIcon.BLOG.getSmallIcon(), IconTheme.JabRefIcon.BLOG.getIcon());
+    private final AbstractAction developmentVersionAction = new OpenBrowserAction("https://builds.jabref.org/master/",
+            Localization.menuTitle("Development version"),
+            Localization.lang("Opens a link where the current development version can be downloaded"));
+    private final AbstractAction changeLogAction = new OpenBrowserAction(
+            "https://github.com/JabRef/jabref/blob/master/CHANGELOG.md", Localization.menuTitle("View change log"),
```
koppor: Isn't it `changelog`? (See for instance https://en.wikipedia.org/wiki/Changelog)

koppor:
One can directly link to the release in CHANGELOG.md, when one uses the GitHub repository: One could also link the concrete release. E.g., https://github.com/JabRef/jabref/blob/master/CHANGELOG.md#35---2016-07-13 

However, in the URL, the release date has also encoded. Maybe we directly change the link before each release?

matthiasgeiger:
Yeah with link changes it would be no problem - but we wanted to avoid these. 

Shall we add this last minute to 3.6?

mlep:
Since it may help in attracting attention to JabRef supporting tools, and hence foster its development, I tend to answer yes.

oscargus:
I'll resolve the merge conflicts if nothing else and let us see.

I think it is fine to link to the complete CHANGELOG.md, not just for the actual release as it is quite clear what the different versions are. I tend to think that the current solution is good enough right now, but a better looking version (without the version management stuff) can be found for 3.7.

matthiasgeiger:
@mlep Can you provide us the translations for French? :grinning:

stefan-kolb:
I'd betetr liked the Jabref resources under help forum. 3 links for web resources, then error console check for updates and about dialog.

matthiasgeiger:
I like the current position :wink:

oscargus:
OK, I've rebased. Still in the position of the last screen shot...

src/main/resources/l10n/JabRef_ru.properties
```diff
@@ -2424,3 +2424,9 @@ However,_a_new_database_was_created_alongside_the_pre-3.6_one.=
 Click_here_to_learn_about_the_migration_of_pre-3.6_databases.=
 
 Connecting...=
+Opens_JabRef's_Facebook_page=
```
mlep: Translation: Ouvre_la_page_Facebook_de_JabRef

src/main/resources/l10n/JabRef_ru.properties
```diff
@@ -2424,3 +2424,9 @@ However,_a_new_database_was_created_alongside_the_pre-3.6_one.=
...
+Opens_JabRef's_blog=
```
mlep: Translation: Ouvre_le_blog_de_JabRef

src/main/resources/l10n/JabRef_ru.properties
```diff
@@ -2424,3 +2424,9 @@ However,_a_new_database_was_created_alongside_the_pre-3.6_one.=
...
+Opens_JabRef's_website=
```
mlep: Translation: Ouvre_le_site_Internet_de_JabRef

src/main/resources/l10n/JabRef_ru.properties
```diff
@@ -2424,3 +2424,9 @@ However,_a_new_database_was_created_alongside_the_pre-3.6_one.=
...
+
+Opens_a_link_where_the_current_development_version_can_be_downloaded=
```
mlep: Translation: Ouvre_un_lien_d'où_la_version_de_développement_actuelle_peut_être_téléchargée

src/main/resources/l10n/JabRef_ru.properties
```diff
@@ -2424,3 +2424,9 @@ However,_a_new_database_was_created_alongside_the_pre-3.6_one.=
...
+See_what_has_been_changed_in_the_JabRef_versions=
```
mlep: Translation: Afficher_les_changements_dans_les_versions_de_JabRef

src/main/resources/l10n/Menu_fr.properties
```diff
@@ -125,3 +125,10 @@ Disable_highlight_groups_matching_entries=Désactiver_le_surlignement_des_groupe
 Online_help_forum=Forum_d'aide_en_ligne
 
 Manage_protected_terms=Gérer_les_termes_protégés
+
+Website=
```
mlep: Translation: Site Internet

src/main/resources/l10n/Menu_fr.properties
```diff
@@ -125,3 +125,10 @@ Disable_highlight_groups_matching_entries=Désactiver_le_surlignement_des_groupe
...
+Blog=
```
mlep: Translation: Blog

src/main/resources/l10n/Menu_fr.properties
```diff
@@ -125,3 +125,10 @@ Disable_highlight_groups_matching_entries=Désactiver_le_surlignement_des_groupe
...
+JabRef_resources=
```
mlep: Translation: Plus_sur_JabRef

src/main/resources/l10n/Menu_fr.properties
```diff
@@ -125,3 +125,10 @@ Disable_highlight_groups_matching_entries=Désactiver_le_surlignement_des_groupe
...
+
+Development_version=
```
mlep: Translation: Version_en_développement

src/main/resources/l10n/Menu_fr.properties
```diff
@@ -125,3 +125,10 @@ Disable_highlight_groups_matching_entries=Désactiver_le_surlignement_des_groupe
...
+View_change_log=
```
mlep: Translation: Afficher_le_fichier_des_changements

matthiasgeiger:
Okay. Thanks @mlep! I'll merge the PR as is and add the French translations directly at `master`.

src/main/resources/l10n/JabRef_ru.properties
matthiasgeiger: this should be `d'où-la ... peut-être` -right?

matthiasgeiger: or if the internet is right `d'où_la` :wink:

PR #1822, opened 2016-08-23
Integrity checker for fields with keys

oscargus:
Checks field containing keys so that an entry with the key exists.


matthiasgeiger:
LGTM

Minor comment: Maybe the localization string should be "*Referenced*_BibTeX_key_does_not_exist"?

You can also add the german translation right away ;-)

> BibTeX-Key_existiert_nicht

or

> Referenzierter_BibTeX-Key_existiert_nicht

matthiasgeiger:
... and of course tests would be nice :stuck_out_tongue_winking_eye:

oscargus:
Might indeed be a better wording. I was happy with the current one as I did not have to write it with a starting lower case letter, as most of the other ones. Consistency is really annoying. ;-)

(I skipped the `%0` version of the string as they in most cases would be very similar.)

Tests are planned and the main reason for WIP. :-) As I have to pass the database they require a bit more effort than the standard checkers.

PR #1828, opened 2016-08-23
Made MetaData.getGroups return Optional

oscargus:
- [x] Tests created for changes (sort of, at least that it returns an Optional.empty() when nothing is set)


stefan-kolb:
LGTM :+1:

PR #1830, opened 2016-08-23
Made BibDatabaseContext.getDatabaseFile() return Optional

oscargus:
Plus one (or two, I couldn't confirm the second but I think there will be an NPE if starting JabRef with an already existing file as an argument and an unsaved (untitled) database open) bug fixes.

This also includes #1782 as I messed up a bit...

### DIFF
--- a/src/main/java/net/sf/jabref/BibDatabaseContext.java
+++ b/src/main/java/net/sf/jabref/BibDatabaseContext.java
@@ -98,10 +98,10 @@ public class BibDatabaseContext {
     /**
      * Get the file where this database was last saved to or loaded from, if any.
      *
-     * @return The relevant File, or null if none is defined.
+     * @return Optional of the relevant File, or Optional.empty() if none is defined.
      */
-    public File getDatabaseFile() {
-        return file;
+    public Optional<File> getDatabaseFile() {
+        return Optional.ofNullable(file);
     }
 
     public void setDatabaseFile(File file) {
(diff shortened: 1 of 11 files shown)


src/main/java/net/sf/jabref/BibDatabaseContext.java
```diff
@@ -175,13 +175,14 @@ private String getFileDirectoryPath(String directoryName) {
         String dir = directoryName;
         // If this directory is relative, we try to interpret it as relative to
         // the file path of this BIB file:
-        if (!new File(dir).isAbsolute() && (getDatabaseFile() != null)) {
+        Optional<File> databaseFile = getDatabaseFile();
+        if (!new File(dir).isAbsolute() && databaseFile.isPresent()) {
```
Siedlerchr: I am not quite sure about this one, but if you use the nio Paths methods, you could omit the check for absolute/relative directory.

oscargus: OK. I have very little (basically no) idea about nio.Paths though, so I prefer to avoid it at this time... If you could spare some time and provide code I can just paste it would be appreciated though.

src/main/java/net/sf/jabref/gui/JabRefFrame.java
```diff
@@ -1624,21 +1620,21 @@ public void updateAllTabTitles() {
...
+            Optional<File> file = getBasePanelAt(i).getBibDatabaseContext().getDatabaseFile();
 
-            if (file == null) {
-                tabbedPane.setTitleAt(i, getBasePanelAt(i).getTabTitle());
-            } else {
-                if (!uniqPath.equals(file.getName())) {
+            if (file.isPresent()) {
+                if (!uniqPath.equals(file.get().getName())) {
                     // remove filename
                     uniqPath = uniqPath.substring(0, uniqPath.lastIndexOf(File.separator));
```
Siedlerchr: Just Paths.get().getParent()

Siedlerchr:
Just some minor nio related things you could change, too. Apart from that, the optional stuff LGTM :+1:

oscargus:
I'm not confident enough regarding nio.Paths to edit blindly (many other APIs I do that though), nor do I really know how to trigger the code correctly, so I would very much prefer to leave it as is. Even though I conceptually understand at least the last comment and see that makes sense to use it there.

oscargus:
I merge this in and hope @Siedlerchr are helpful at some stage regarding the nio-stuff...

PR #1837, opened 2016-08-24
Fixes #1687 "month" field ascending/descending sorting swapped

### LINKED ISSUE #1687: "Month" field ascending/descending sorting swapped
JabRef version 3.5 on Windows 10 64-bit

When sorting according to "month", "ascending"/"descending" are switched.

For "year", for example, descending order results in later years at the top, as expected.
For "month", however, descending order results in January at the top, and December at the bottom, which is inconsistent.

This affects both the main view and the order in which the entries are saved (if a custom save order based on the "month" filed is specified in the database properties).

stefan-kolb:
When sorting according to "month", "ascending"/"descending" are switched.


oscargus:
LGTM! Great with the comparator test! :+1:

PR #1811, opened 2016-08-22
Preview panel height is now saved immediately

bartsch-dev:
The Preview panel height is not stored when changed (only when JabRef is closed)

### Steps
 - selecting an entry (thus opening the preview panel)
 - changing the preview panel height
 - closing the preview panel
 - selecting an entry (thus opening the preview panel)
 - the preview panel has the old height, not the new one we just changed

PR #1840, opened 2016-08-24
Preview panel height is not saved after the PreferencesDialog has closed

bartsch-dev:
Follow up to #1811.

I didn't consider that the splitpane is reinitialized each time the preferencesDialog is closed (then the listener would get lost and the divider position wouldn't get saved anymore).

This fixes that.


boceckts:
LGTM 👍

oscargus:
LGTM! :+1:

PR #1842, opened 2016-08-24
Some more Optional

oscargus:
* ParserResult.getFile()
* Author.getFirst(), getVon(), getJR(), and getLast()

Some minor clean ups in HTMLChars and LatexToUnicodeFormatter

Also fixed an NPE when the pattern `authFirstFull` is used and the author does not have a von-part.


Siedlerchr:
LGTM :+1:

PR #1844, opened 2016-08-24
Cleanup of some import tests

Siedlerchr:
Fixes some encoding problems when running gradle check on win


src/test/java/net/sf/jabref/logic/importer/fileformat/BibTeXMLImporterTest.java
```diff
@@ -34,11 +34,12 @@
      * @throws IOException
      */
     public List<Path> getTestFiles() throws IOException {
-        List<Path> files = new ArrayList<>();
-        try (DirectoryStream<Path> stream = Files.newDirectoryStream(Paths.get(FILEFORMAT_PATH))) {
-            stream.forEach(files::add);
+        try (Stream<Path> stream = Files.list(Paths.get(BibTeXMLImporterTest.class.getResource("").toURI()))) {
+            return stream.filter(p -> !Files.isDirectory(p)).collect(Collectors.toList());
+        } catch (URISyntaxException e) {
```
oscargus: Would it make sense to just throw this Exception from the method instead? I guess the test will/should fail anyway.

oscargus:
Looks good! :+1:

There reformatting of strings is a bit annoying but I do not know how to get around it.

Throwing the exception would get rid of the printed stack trace and it would show up slightly better in Eclipse  (I think).

Issue #1703, opened 2016-08-09
Improve shared database feature


koppor:
Follow up of #1451

- [x] Database properties dialog:
  - [x] Do not show "Save sort order" - database sorts arbitrarily :innocent: 
  - [x] Do not show "Database protection" - makes no sense in a shared database setting
- [x] Session restore: JabRef should open all `[shared]` tabs after startup
- [x] Store last used connection in the "open remote database dialog"
- [x] "open remote database dialog": "Connect" button should be disabled an label change to "Connecting...". Reason: Connection may take longer and pressing the button multiple times leads to error message dialogs. (You know, http://www.freemysqlhosting.net/ latency)

Siedlerchr:
- [ ] Use `AssumeTrue(DevEnvironment.isCIServer())`  switch to disable database tests locally. Prevents failing tests, if no db is running locally.

koppor:
Alternative solution: #1705 to make it explicit when to call the database tests. I know that one can also run the tests using the ide, but I would like to keep the consistency with our integration tests. Please follow up at #1705.

obraliar:
This issue could be closed.

PR #1846, opened 2016-08-25
Add possibility to remember password for shared databases.

obraliar:
Issue: https://github.com/JabRef/jabref/issues/1703. _(completition)_

>  Store last used connection in the "open remote database dialog"

_Screenshot:_
openshareddatabase


src/main/java/net/sf/jabref/shared/prefs/SharedDatabasePreferences.java
```diff
@@ -0,0 +1,90 @@
...
+    private static final String SHARED_DATABASE_TYPE = "sharedDatabaseType";
+    private static final String SHARED_DATABASE_HOST = "sharedDatabaseHost";
+    private static final String SHARED_DATABASE_PORT = "sharedDatabasePort";
+    private static final String SHARED_DATABASE_NAME = "sharedDatabaseName";
+    private static final String SHARED_DATABASE_USER = "sharedDatabaseUser";
+    private static final String SHARED_DATABASE_PASSWORD = "sharedDatabasePassword";
+    private static final String SHARED_DATABASE_REMEMBER_PASSWORD = "sharedDatabaseRememberPassword";
+
+    // This {@link Preferences} is used only for things which should not appear in real JabRefPreferences due to security reasons.
+    private final Preferences internalPrefs = Preferences.userNodeForPackage(OpenSharedDatabaseDialog.class);
```
koppor: Is it possible to use another node for storing? The whole preference tree gets exported. Therefore, this should be located somewhere else @matthiasgeiger

koppor:
Could you use some unique system properties such as the login name, serial number of Windows or something as additional encryption key?

matthiasgeiger:
Maybe this `Preferences userPrefs = Preferences.userRoot().node( "/com/tutego/insel" );` as used here http://openbook.rheinwerk-verlag.de/javainsel/javainsel_11_009.html could be used to create another registry node just for the password which is then not exported by JabRef.

src/main/java/net/sf/jabref/gui/shared/OpenSharedDatabaseDialog.java
```diff
@@ -194,6 +197,12 @@ private void applyGlobalPrefs() {
         if (sharedDatabaseUser.isPresent()) {
             userField.setText(sharedDatabaseUser.get());
         }
+
+        if (sharedDatabasePassword.isPresent()) {
```
boceckts: The password should only be set when there is a password available **and** the checkbox to remember the password is checked so please add `&& sharedDatabaseRememberPassword`.

src/main/java/net/sf/jabref/shared/security/Password.java
```diff
@@ -0,0 +1,59 @@
...
+
+import javax.crypto.Cipher;
+import javax.crypto.spec.SecretKeySpec;
+
+import net.sf.jabref.shared.DBMSConnector;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+public class Password {
```
boceckts: Please add a javadoc as this class might be usefull to others in the future.

src/main/java/net/sf/jabref/shared/security/Password.java
```diff
@@ -0,0 +1,59 @@
...
+    private static final Log LOGGER = LogFactory.getLog(DBMSConnector.class);
+
+    private static final String ALGORITHM = "AES";
+    private static final String KEY = "ThisIsA128bitKey";
+
+    private final String phrase;
+
+
+    public Password(String phrase) {
+        this.phrase = phrase;
```
boceckts: Does the parameter need to be a class field?
Why not make the class a utility class that doesn't need to be instantiated by making all methods static?
e.g.
`Password.encrypt("PasswordToEncrypt");`

boceckts:
Only minor coments from me. Once they are done and the conflicts are reseolved, it looks good to me :+1:

Siedlerchr:
Password Hashing is the solution! (It should have come earlier to my mind)
https://paragonie.com/blog/2016/02/how-safely-store-password-in-2016#java

src/main/java/net/sf/jabref/gui/shared/OpenSharedDatabaseDialog.java
obraliar: See: https://github.com/JabRef/jabref/pull/1846/files#diff-bf287bbdc594ec6ec66a15154e141110R316. If the checkbox is not select the pref is going to be cleared.

obraliar:
@matthiasgeiger Unfortunately it does not function. I used for testing `private final Preferences internalPrefs = Preferences.userRoot().node("/bla/bla1/security");` If I export the preferences, the following nodes appear again:
```
  1 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
  2 <!DOCTYPE preferences SYSTEM "http://java.sun.com/dtd/preferences.dtd">
  3 <preferences EXTERNAL_XML_VERSION="1.0">
  4   <root type="user">
  5     <map/>
  6     <node name="net">
  7       <map/>
  8       <node name="sf">
  9         <map/>
 10         <node name="jabref">
 11           <map>
 12           <!-- ... -->
 13           </map>
 14           <node name="bibtexKeyPattern">
 15             <map/>
 16           </node>
 17           <node name="bibtexkeypatterns">
 18             <map/>
 19           </node>
 20           <node name="gui">
 21             <map/>
 22             <node name="shared">
 23               <map>
 24                 <entry key="sharedDatabasePassword" value="kJicIWFbSjTjw5QYiENFLA=="/>
 25               </map>
 26             </node>
 27           </node>
 28           <node name="labelPattern">
 29             <map/>
 30           </node>
 31           <node name="logic">
 32             <map/>
 33             <node name="labelpattern">
 34               <map/>
 35             </node>
 36           </node>
 37         </node>
 38       </node>
 39     </node>
 40   </root>
 41 </preferences>
```

As another solution we could add `new SharedDatabasePreferences().clear();` into `JabRefPreferences.exportPreferences(String filename)`.

src/main/java/net/sf/jabref/shared/security/Password.java
obraliar: Done.

src/main/java/net/sf/jabref/shared/security/Password.java
obraliar: Could You suggest a new class name for this case, cause `Password` is not adoptable any more. Sth. which symantically means like `PasswordEncrypterAndDecrypter`.

matthiasgeiger:
@obraliar I think that the exported key is a remnant of previously saved settings.

If you are using windows use `regedit` to delete the keys in the node `HKEY_CURRENT_USER\SOFTWARE\JavaSoft\Prefs\net\sf\jabref`

Your addition should be placed in `HKEY_CURRENT_USER\SOFTWARE\JavaSoft\Prefs\bla\bla1\security`

obraliar:
@Siedlerchr There should be a possibility to decrypt the password again as this is going to be sent to DBMS. Could you explain what to do with hashing?

obraliar:
@matthiasgeiger :+1:  I'm using Linux (Debian) and there you have to remove `~/.java/.userPrefs/net/sf` from file system. Now it works very well!
Now using: `private final Preferences internalPrefs = Preferences.userRoot().node("/net/sf/security/jabref");`

obraliar:
@koppor What about using the MAC address?

src/main/java/net/sf/jabref/shared/prefs/SharedDatabasePreferences.java
obraliar: Done.

matthiasgeiger:
I think the suggestion is now obsolote as exporting the prefs without the password is now possible. Maybe we should add a comment in the help file that the password is saved on the local hard disk - then a user can decide whether he wants to save it or not...

Regarding the prefs node: As we want to move the package from "net.sf" to "org" eventually in the future it would perhaps better to use `Preferences.getUserNodeForPackage(JabRefMain.class).parent().node("jabref-pwdstorage")` to use a relative approach.

obraliar:
@matthiasgeiger I agree cause the safest way it not to use the function "remember password" as this is generally a security gap due to symetric encryption.

Siedlerchr:
Forget about the Hashing. I didn't think about that you need it in plaintex to send to the db..

koppor:
@matthiasgeiger I would nevertheless keep a simple encryption to make it a bit harder to access the password.

obraliar:
@JabRef/developers Could someone look at this PR and possibly merge it in, please?
Reason: @koppor is offline for a week.

stefan-kolb:
LGTM :+1:

src/main/java/net/sf/jabref/gui/shared/OpenSharedDatabaseDialog.java
```diff
@@ -39,9 +41,16 @@
...
 import net.sf.jabref.shared.exception.DatabaseNotSupportedException;
+import net.sf.jabref.shared.prefs.SharedDatabasePreferences;
+import net.sf.jabref.shared.security.Password;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 
 public class OpenSharedDatabaseDialog extends JDialog {
 
+    private static final Log LOGGER = LogFactory.getLog(Password.class);
```
Siedlerchr: Change the class in getLog to the actual class name (OpenSharedDatabaseDialog)
Otherwise logs shows wrong class.

src/main/java/net/sf/jabref/gui/shared/OpenSharedDatabaseDialog.java
```diff
@@ -198,6 +207,16 @@ private void applyGlobalPrefs() {
         if (sharedDatabaseUser.isPresent()) {
             userField.setText(sharedDatabaseUser.get());
         }
+
+        if (sharedDatabasePassword.isPresent() && sharedDatabaseUser.isPresent()) {
+            try {
+                passwordField.setText(new Password(sharedDatabasePassword.get(), sharedDatabaseUser.get()).decrypt());
+            } catch (GeneralSecurityException e) {
+                LOGGER.error("Could not read the password due to decryption problems.");
```
Siedlerchr: LOGGER.errror has a second paramter for the exception, so just pass, `e` as 2nd parameter

src/main/java/net/sf/jabref/gui/shared/OpenSharedDatabaseDialog.java
```diff
@@ -300,12 +322,24 @@ private void initLayout() {
...
+        prefs.setHost(hostField.getText());
+        prefs.setPort(portField.getText());
+        prefs.setName(databaseField.getText());
+        prefs.setUser(userField.getText());
+
+        if (rememberPassword.isSelected()) {
+            try {
+                prefs.setPassword(new Password(new String(passwordField.getPassword()), userField.getText()).encrypt());
+            } catch (GeneralSecurityException e) {
+                LOGGER.error("Could not store the password due to encryption problems.");
```
Siedlerchr: same as above

src/main/java/net/sf/jabref/gui/shared/OpenSharedDatabaseDialog.java
```diff
@@ -300,12 +322,24 @@ private void initLayout() {
...
+    public void setPreferences() {
+        prefs.setType(((DBMSType) dbmsTypeDropDown.getSelectedItem()).toString());
+        prefs.setHost(hostField.getText());
+        prefs.setPort(portField.getText());
+        prefs.setName(databaseField.getText());
+        prefs.setUser(userField.getText());
+
+        if (rememberPassword.isSelected()) {
+            try {
+                prefs.setPassword(new Password(new String(passwordField.getPassword()), userField.getText()).encrypt());
```
Siedlerchr: With this line your are probably (breaking/reducing the security):
And a char array "protects" you from accidentally writing the password in plaintext in a log...
http://stackoverflow.com/questions/8881291/why-is-char-preferred-over-string-for-passwords-in-java

sso I would recommend using a char[] in your Password class

src/main/java/net/sf/jabref/shared/security/Password.java
```diff
@@ -0,0 +1,68 @@
...
+
+        byte[] operand = STATIC_KEY.getBytes();
+
+        // increase key complexity using XOR
+        for (int i = 0; i < convertedKey.length; i++) {
+            operand[i] = (byte) (operand[i] ^ convertedKey[i]);
+        }
+        return operand;
+    }
+}
```
Siedlerchr: Add new line

src/main/java/net/sf/jabref/shared/security/Password.java
```diff
@@ -0,0 +1,68 @@
...
+     * @param key Secret key which is used to improve symmetric encryption
+     */
+    public Password(String phrase, String key) {
+        this.phrase = phrase;
+        this.key = convertToAESKey(key);
+    }
+
+    /**
+     *  Encrypts the set phrase/password with strong symmetric encryption algorithms.
+     *
```
Siedlerchr: Well, AES 128 bit is a joke and not really secure, I personally would prefer at least 256biz, but using stronger encryption is only possible with the Cryptographic Extensions Policy and unfortunately  those fall under the US import/export regulation laws. 

It would be possible to check if they are installed and use a greater key size. Not sure it if is really necessary here.

src/main/java/net/sf/jabref/shared/security/Password.java
```diff
@@ -0,0 +1,68 @@
...
+
+import javax.crypto.Cipher;
+import javax.crypto.spec.SecretKeySpec;
+
+/**
+ *   {@link Password} contains methods which are useful to encrypt and decrypt passwords using symetric algorithms.
+ */
+public class Password {
+
+    private static final String ALGORITHM = "AES";
```
Siedlerchr: Please specify transformation/padding/mode : 

> It is recommended to use transformations that fully specify the algorithm, mode, and padding instead of relying on the defaults.

As noted here: https://docs.oracle.com/javase/8/docs/technotes/guides/security/SunProviders.html

Siedlerchr:
Generally good work, if you address the comments, then we can merge it in :+1:

src/main/java/net/sf/jabref/gui/shared/OpenSharedDatabaseDialog.java
obraliar: Done.

src/main/java/net/sf/jabref/gui/shared/OpenSharedDatabaseDialog.java
obraliar: Done.

src/main/java/net/sf/jabref/gui/shared/OpenSharedDatabaseDialog.java
obraliar: Done.

src/main/java/net/sf/jabref/gui/shared/OpenSharedDatabaseDialog.java
obraliar: Done.

src/main/java/net/sf/jabref/shared/security/Password.java
obraliar: This functionality is unsecure anyway. Aim: https://github.com/JabRef/jabref/pull/1846#issuecomment-242583251.

src/main/java/net/sf/jabref/shared/security/Password.java
obraliar: Done.

src/main/java/net/sf/jabref/shared/security/Password.java
obraliar: Done.

src/main/java/net/sf/jabref/gui/shared/OpenSharedDatabaseDialog.java
```diff
@@ -40,9 +43,16 @@
...
 import net.sf.jabref.shared.exception.DatabaseNotSupportedException;
+import net.sf.jabref.shared.prefs.SharedDatabasePreferences;
+import net.sf.jabref.shared.security.Password;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 
 public class OpenSharedDatabaseDialog extends JDialog {
 
+    protected static final Log LOGGER = LogFactory.getLog(OpenSharedDatabaseDialog.class);
```
stefan-kolb: Why is the logger protected here?

obraliar: Done. Copy & paste from `DBMSProcessor` :see_no_evil:.

obraliar:
All issues have been fixed.

Issue #1813, opened 2016-08-22
Request: Import/Export preferences dialog with working directory as default


grimes2:
The current default directory is user-dir. I think, the correct place for the xml file is the working directory (or a subdirectory of it). The working directory variable should work now.

Talking about this dialog:
![image](https://cloud.githubusercontent.com/assets/2141507/17858466/264fed2a-6886-11e6-9dd0-dbec9591ae3c.png)

stefan-kolb:
@JabRef/developers We got a few different working dirs. Maybe we should discuss if we really need them or just use one working dir for all dialogs:
```
    public static final String FILE_WORKING_DIRECTORY = "fileWorkingDirectory";
    public static final String IMPORT_WORKING_DIRECTORY = "importWorkingDirectory";
    public static final String EXPORT_WORKING_DIRECTORY = "exportWorkingDirectory";
    public static final String WORKING_DIRECTORY = "workingDirectory";
```

stefan-kolb:
Especially `FILE_WORKING_DIRECTORY` is debatable. used for external file type application selection and in the file dialog of the general tab of an entry.

Siedlerchr:
Working Directory is clear, it is always the location of the bib file (if not otherwise set).
Yeah, I  think File_Working_Directory is confusing, especially for adding a file to an entry. There I would expect  to open the dlg in the Working_Directory.  However, some may prefer having two different dirs for that. No idea. Ask the users.

stefan-kolb:
It's not really about the user right now it's about the concept because it's inconsistent at the moment.

grimes2:
My user opinion regarding `FILE_WORKING_DIRECTORY`: You can remove the variable, text field, browse button in the preferenes. And also remove the checkbox below and set `WORKING_DIRECTORY` as default for the entry editor dialog.

grimes2:
I want to work on this issue. But I need your ok, that this change from user-dir to working-dir for the preferences xml is useful.

Siedlerchr:
@JabRef/developers  Any comments on this? From my point of view it's okay to change it to working-dir

stefan-kolb:
Issue not closed yet, see my comment on file prefs.

koppor:
@grimes2 Where is the text field for `FILE_WORKING_DIRECTORY`? And the checkbox? I see only the screenshot with Export Preferenes.

In the preferences, I see:

[...]

PR #1856, opened 2016-08-25
Removed Globals from BibDatabaseContext

oscargus:
This almost worked. The penalty was that I had to (temporarily) introduce it in `FileUtil.expandFilename` where the arbitrary extension line makes it extremely cumbersome to get around it.

### DIFF
--- a/src/main/java/net/sf/jabref/external/WriteXMPAction.java
+++ b/src/main/java/net/sf/jabref/external/WriteXMPAction.java
@@ -121,11 +121,16 @@ public class WriteXMPAction extends AbstractWorker {
             List<File> files = new ArrayList<>();
 
             // First check the (legacy) "pdf" field:
-            entry.getFieldOptional(FieldName.PDF).ifPresent(pdf ->
-                FileUtil.expandFilename(pdf, panel.getBibDatabaseContext().getFileDirectory("pdf"))
-                    .ifPresent(files::add));
+            entry.getFieldOptional(FieldName.PDF)
+                    .ifPresent(
+                            pdf -> FileUtil
+                                    .expandFilename(pdf,
+                                            panel.getBibDatabaseContext().getFileDirectory(
+                                                    Globals.prefs.getFileDirectoryPreferences(FieldName.PDF)))
+                            .ifPresent(files::add));
             // Then check the "file" field:
-            List<String> dirs = panel.getBibDatabaseContext().getFileDirectory();
+            List<String> dirs = panel.getBibDatabaseContext()
+                    .getFileDirectory(Globals.prefs.getFileDirectoryPreferences());
             if (entry.hasField(FieldName.FILE)) {
                 FileListTableModel tm = new FileListTableModel();
                 entry.getFieldOptional(FieldName.FILE).ifPresent(tm::setContent);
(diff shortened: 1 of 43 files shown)


src/main/java/net/sf/jabref/external/WriteXMPAction.java
```diff
@@ -121,11 +121,16 @@ public void run() {
             List<File> files = new ArrayList<>();
 
             // First check the (legacy) "pdf" field:
-            entry.getFieldOptional(FieldName.PDF).ifPresent(pdf ->
-                FileUtil.expandFilename(pdf, panel.getBibDatabaseContext().getFileDirectory("pdf"))
-                    .ifPresent(files::add));
+            entry.getFieldOptional(FieldName.PDF)
+                    .ifPresent(
+                            pdf -> FileUtil
```
Siedlerchr: The formatting here and below looks a bit weird...

src/main/java/net/sf/jabref/external/WriteXMPEntryEditorAction.java
```diff
@@ -55,11 +55,16 @@ public void actionPerformed(ActionEvent actionEvent) {
         List<File> files = new ArrayList<>();
 
         // First check the (legacy) "pdf" field:
-        entry.getFieldOptional(FieldName.PDF).ifPresent(pdf -> FileUtil
-                .expandFilename(pdf, panel.getBibDatabaseContext().getFileDirectory("pdf")).ifPresent(files::add));
+        entry.getFieldOptional(FieldName.PDF)
```
Siedlerchr: see above

oscargus: It does indeed, but that is what Eclipse gives me when reformatting the whole expression...

oscargus:
I think I know how to get around this now. It seems like there are three possible directories "fileDirectory", "pdfDirectory", and "psDirectory", where only the first one is normally used and the other comes from the legacy pdf and ps fields. Now, it might be possible to hard-code in the preferences support for more extensions, but do we really need to support that or would it be OK to include a limitation to only those three?

Siedlerchr:
LGTM, jut two weird formatting things.
Regardng the FileUtil class. I will rework this., because it contains a lot of outdated code which can be replaced with Standard Java implementation. Maybe we could get rid of some prefs there.

Regarding the directories, there is also a discussion about the import/export dir preferences #1813 
I don't think we need the legacy pdf/ps fields. Isn't there a cleanup job to move them?

oscargus:
There was an old issue where someone used them for something completely different, so maybe we shouldn't remove them. As long as it is only those three (or at least a well known subset of all possible `extDirectory` it should be OK. Now the problem is that one can use any extension and then the code tries to find a dedicated directory for that extension, which cannot be generated in any other way than editing the preference XML...

oscargus:
#496 I think. Although that doesn't discuss arbitrary extensions (from a quick browse).

lenhard:
I would also say that it is ok to remove support for the legacy file directories. And I do not really see why we should support arbitrary extension directories.

oscargus:
OK, dropping arbitrary extensions I'm fine with, but I will not remove support for the pdf and ps directories right now.

src/main/java/net/sf/jabref/external/WriteXMPAction.java
oscargus: One of them was sorted out with the new code, but not the other...

oscargus:
Now there are no more dependencies on Globals or logic in BibDatabaseContext. Still a dependency on shared though, but that I haven't looked into.

Issue #1841, opened 2016-08-24
Impossible to set EtAlString empty


ph0boss:
JabRef version 3.5

Steps to reproduce:

1. Use a style with Authors formatting as follows:
article=\format[Authors(LastFirst,Initials,FullPunc,1,1,EtAl=)]{\author}
In this example EtAlString is set to be empty.
2. Use it to cite something.
3. Author is formatted as follows:
Surname, N. et al.

[example_style.jstyle.zip](https://github.com/JabRef/jabref/files/435439/example_style.jstyle.zip)

oscargus:
Thanks for the report. I can confirm that there is code prohibiting setting the string empty. However, as there seems to (at least) one use case and I do not really see why it should be prohibited, there is a fixed version in https://builds.jabref.org/fix1841/ in five minutes. Please try it out.

oscargus:
I close this as we merged the fix. Feel free to reopen it if it doesn't work. (3.6 will be released tomorrow and should include it.)

ph0boss:
Maybe it works.. I couldn't try it because select style button doesn't work. Nothing happens when I press it. I tried both the link you provided and the latest master. Fedora 25 64bit. 
Running  as $ java -jar ./JabRef-3.6dev--snapshot--2016-08-25--master--40ae061.jar
Should I file another bug report about this issue?

oscargus:
Strange... I'll have to look into that. As it is (should be) unrelated to this, it would be great if you can file a new bug report, although I'll still hopefully remember it for tonight.

oscargus:
OK, I think I found the problem. If you add a "default" style you should be able to test it (unless you have it already, but when I tried you example I got an error from not having a style for default, only article). I'll try to fix that problem in the source so that it behaves slightly better, at least shows a dialog instead of throwing errors.

ph0boss:
Yeah, now both issues are fixed. Thanx for such quick help!

PR #1857, opened 2016-08-25
Fixed NPE when loading a style file that has no default style

oscargus:
Found when trying #1841 although non-related.


Siedlerchr:
Maybe we could get that into 3.6, too?

oscargus:
Yes, I think it would make sense/be safe. A better solution is discussed in #1858 .

Issue #1757, opened 2016-08-17
Empty Error setting field


grimes2:
JabRef version 3.5 on Windows 10

Steps to reproduce:

1. Insert [{a in article's title field
2. Save database
3. Error setting field empty
4. forced: Kill JabRef process

![jabref - c__users_lange_downloads_test bib_ bibtex mode _2016-08-17_01-54-20](https://cloud.githubusercontent.com/assets/15456836/17719978/bf292780-641e-11e6-8084-b4e528022000.png)
(covering another message box: Error: Braces don't match)

mlep:
Thank you for your report.
I have tried to reproduce it with the current development version (on my linux computer): It shows the message box "Error:Braces don't match", but I could click on "OK" and continue editing the file. So, it seems to work.
Please, could you try with the dev version on your windows 10?
You can find the development version here: http://builds.jabref.org/master/

grimes2:
Same bug with the development version. Please try to save database repeatedly, because the bug is not always reproducible.. [{a is the only entry in title field. (I think the problem is, that two braces don't match, so two exeptions are thrown)

oscargus:
I can reproduce it. You (or at least I) have to click on the save icon, pressing Ctrl+s doesn't trigger the crash.

oscargus:
Interesting:
* Ctrl + s => two error dialogs, continue working
* File -> Save database => one error dialog, continue working
* Save button in toolbar => crash with blank (after a very short time) error dialog

My guess is that this may be related to some old code where we catch more or less any exception and ignore it...

oscargus:
I can also add that it doesn't seem related to any EDT issue (although there is one solved in #1725 that will be triggered later in the save process, but JabRef freezes before getting there).

oscargus:
I looked a bit more intro this and what happens is that for ctrl+s the first dialog shows up and then is the save action not performed until "OK" is clicked. Then during the actual save, the next dialog shows.

When pressing the toolbar button, the save action starts despite the first dialog not being confirmed. Then, after a while, a second dialog shows up and then it freeze. It seems related to the spin library somehow. Looking at BasePanel.runCommand() it is the following line that freezes the program:

``` java
                wrk.run(); // Runs the potentially time-consuming action
                // without freezing the GUI. The magic is that THIS line
                // of execution will not continue until run() is finished.
```

probably because both dialogs are up at the same time (the first one cannot be clicked during debug).

oscargus:
[...]

PR #1863, opened 2016-08-26
Error setting field freeze

grimes2:


lenhard:
Refs #1757

It is always nice to see a new contributor! :) You took a rather challenging task for your first contribution.

When you consider your PR ready, please indicate this to us by setting the label "ready-for-review" at the right hand side. Then, we will have a look at your code, comment on things to improve, and try to verify and test that your solution works as desired. 

Probably @oscargus should have a look at this.

grimes2:
My pull request is ready. Please set the label "ready-for-review".

oscargus:
Thanks! It clearly solves the problem! However, it also removes the warning when changing field, but I am not sure if this is an improvement or an issue. ;-) A maybe more serious issue is that when selecting another entry, no warning is shown either and the edit is dropped. However, the dropping appears without the fix as well, so again, the question is what is the worst.

One thing is clear, that even though this fix solves the crash problem, which is clearly significant, we really need to look into this automatic field storage/warning mechanism... Also, the red color only flashes by in some situations, although the dialog shows up, so many events happening more or less at the same time. I guess I'm not missing many brackets as I haven't really notices this earlier...

@JabRef/developers : crash on rare occasions or missing warnings slightly more often? I'd say missing warnings.
@grimes2 : feel free to add a change log entry, if not, I'll do it later.

grimes2:
I can add a else branch with the red background color only, for the EDT Events. A error message is impossible, because JOptionPane is not thread-safe.

oscargus:
Yes, either that or set the background outside of the if. Another good thing with the fix is that only one dialog is shown for the save toolbar button. :+1:

lenhard:
@oscargus Missed warnings are certainly preferable to application crashes. Moreover, I can confirm the change losses when selecting a new entry. I will open a new issue for that.

I have tested this PR locally and can confirm that it fixes #1757. The coloring also seems to be consistent after the last commit.

All that is missing for merging is the Changelog entry. This is a little complicated since we had the intermediate release of 3.6. @grimes2: It would be best if you merge master into your branch before editing the Changelog. This should carry over the release data.

grimes2:
I don't know, how to merge master into branch.

Siedlerchr:
If you have setup your local git repository  (look here for details on how to https://help.github.com/articles/fork-a-repo/)
git fetch upstream 
git merge upstream/master
Resolve any conflicts
git push (best done over the git gui)

https://github.com/JabRef/jabref/wiki/Guidelines-for-setting-up-a-local-workspace

oscargus:
Great, thanks! I merge this in.

oscargus:
@grimes Feel free to open a PR with your name added in https://github.com/JabRef/jabref/blob/master/AUTHORS 

We forgot about that. Not so used with new contributors... :+1:

stefan-kolb:
This is autogenerated from the git history @oscargus

oscargus:
Ah, OK! Thanks @stefan-kolb .

PR #1865, opened 2016-08-26
Fixed #1663 Better multi-monitor support

### LINKED ISSUE #1663: Opening the Merge BibTeX dialog on multi-monitor setup
When running JabRef in a multi-monitor setup, e.g. on the second screen, the merge BibTeX dialog always appears on the first screen and not on the screen where the current JabRef instance is running on.

Also applies to other dialogs!

stefan-kolb:
- [x] Location of Windows is keep also for multi-monitor setups
- [x] Location of Window is only reverted when it is not displaybale (no multi-monitor anymore/wrong preferences)
- [x] Removed duplicate maximized state logic
- [x] Dialogs are centered as overlay over the main window not always on the main screen

@JabRef/developers Please try on your multi-monitor setup @ work and or a @ home. 
- The location of the windows should stay the same or should be adapted to a visible alternate location. 
- Also the maximized state should be remembered correctly on all monitors.
- Do you see any dialogs that are not correctly centered in multi monitor scenario (JabRef on 2nd screen, dialog on main screen)


oscargus:
I run a single X screen two monitor setup and I cannot tell the difference compared to before: if any dialog cover both monitors it is moved and resized to stay inside one of the monitors.

stefan-kolb:
Is it really inside the boundaries of the screen rectangle? Only then it can work.
image

oscargus:
When closing the merge entries window:
screenshot-4
When opening it:
screenshot-5

(It may be that the size is kept better now, that I didn't think of.)

Same with the main window (which is adapted to the left monitor on both images).

oscargus:
With that said, I'm sure that this PR solves some problems and I'm not sure that the behaviour I report really is a problem as such.

stefan-kolb:
No your example should work. For me and my horizontal/horizontal setup it works. It looks like for your horizontal/vertical it does not... We have to test more settings to smooth this functionality.

stefan-kolb:
@oscargus I tried your use case on a similar setup on Windows and it worked, maybe a Linux problem?
image

oscargus:
Could indeed be. Especially with the option to have more than one X-screen (which I don't but some people do). Are you interested in any numbers from my setup (obtained from the code)?

stefan-kolb:
Sure, maybe you can also take a look. I simplified the code very much, so it should be easy to debug if Java/the OS supports our case or if we cannot do anything about the problem.

matthiasgeiger:
LGTM

oscargus:
@stefan-kolb I will try to see why things go wrong on my office setup "soon"...

Issue #1496, opened 2016-06-13
FullText Download Dialog only shows doc for latest selected entry if multiple times triggered


Siedlerchr:
JabRef 3.4
windows 7 6.1 amd64 
Java 1.8.0_74

Steps to reproduce:
1. Database with 2 different entries
2. Hit Get Fulltext on entry 1 (e.g. ABC)  -> Searching for Fulltext starts
3. Hit Get Fulltext on entry 2  (e.g. DEF) -> Searching for Fulltext starts 
4. Download Dialog appears: PDF is named after the second entry
5. Download Dialog appears a second time: PDF is named after the second entry
6. If you rename the second pdf, e.g. _2.pdf and look at its content: Different files

stefan-kolb:
Hm, yeah this is expected. Is a thread problem...

oscargus:
I happened to solve this (I think), when trying to enable full text search for multiple files. Please see #1869

tobiasdiez:
@stefan-kolb, @oscargus what is the status of this bug? Is it indeed fixed with #1869?

stefan-kolb:
I don't think so. If so only for the multi downloader action and not in general?!

koppor:
The CHNAGELOG of 3.6 says it is fixed: https://github.com/JabRef/jabref/blob/master/CHANGELOG.md#changed-1

Maybe the changelog entry is too optimistic?

Siedlerchr:
I think this is fixed, it now kinda queues the dls

Am 23.11.2016 1:49 vorm. schrieb "Oliver Kopp" <notifications@github.com>:

Siedlerchr:
Solved. Works.

PR #1869, opened 2016-08-26
Download many files in one go

oscargus:
Implements #1496 and also allows selecting multiple documents to find full text documents for. I think that it doesn't always show a dialog for every failed files when having multiple ongoing downloads.

Please try it out at http://builds.jabref.org/downloadmanyfiles


Siedlerchr:
If it does what you promise, I love you! :heart_eyes:  :laughing:  Will try this tomorrow. 
>think that it doesn't always show a dialog for every failed files when having multiple ongoing downloads.

Not sure if I understand you correctly: So I download e.g. 20 files, and 5 fails. Is then the dialog "Failed DL" displayed 5 times?
Or after all downloads are finished it shows a dialog with e.g. 15/20 files downloaded succesfully?

src/main/java/net/sf/jabref/external/FindFullTextAction.java
```diff
@@ -41,52 +44,60 @@ public void init() throws Throwable {
...
+                List<String> dirs = basePanel.getBibDatabaseContext()
+                        .getFileDirectory(Globals.prefs.getFileDirectoryPreferences());
+                if (dirs.isEmpty()) {
+                    JOptionPane.showMessageDialog(basePanel.frame(),
+                            Localization.lang("Main file directory not set!") + " " + Localization.lang("Preferences")
+                                    + " -> " + Localization.lang("External programs"),
+                            Localization.lang("Directory not found"), JOptionPane.ERROR_MESSAGE);
+                    return;
+                }
+                // TODO: this needs its own thread as it blocks the UI!
```
Siedlerchr: Is this blocking still a problem?

Siedlerchr:
Works technically fine, but I have some remarks, as it's currently not effective usable:
I ran it over Quality->Look up fulltext
When the process is finished, it shows the file save/ Download failed dialog for every entry which I previously selected.
So if I have 200 entries select, after the process it opens 200 dialogs...
I would suggest having just a final download result dialog or directly a process window, perhaps sth like this from Down Them All:
dta


There is apparently no option to cancel the lookup process.

src/main/java/net/sf/jabref/external/FindFullTextAction.java
oscargus: Not sure if it was before either...

oscargus:
Well, what you select is what you get. :-)

I agree that your approach would be even better, but that is out of the scope for me at the moment. One could still reenable one selected entry at a time, with the added benefit of being able to start more than one, but I do not really know if that is a better way to do it. Maybe it is better to pop up a dialog asking if the user is sure they want to try to find > 10(?) full text documents as this can be quite a hassle?

Siedlerchr:
Agree with you. Add this dialog warning, for more than x? entries .
For the rest we should create a new issue/enhancement. Mabye the other devs have some other 

Regarding the blocking: Does not block the UI atm.

oscargus:
I updated and set it to 10. Not sure how much one should actually write in the dialog...

src/main/java/net/sf/jabref/external/FindFullTextAction.java
```diff
@@ -44,6 +44,20 @@ public void init() throws Throwable {
 
     @Override
     public void run() {
+        if (basePanel.getSelectedEntries().size() >= 10) {
```
Siedlerchr: Please extract that "magic number" to a constant.
https://en.wikipedia.org/wiki/Magic_number_(programming)

Siedlerchr:
Code LGTM, just move that magic number to a constant.

oscargus:
Yeah, was thinking about the constant thing as well. Will fix it. I guess it is a good idea to merge now so people have time to find possible issues.

oscargus:
Improved the file name generation as well. If the formatter does return an empty string, the BibTeX key is used. If no BibTeX key, "default" is used.

PR #1874, opened 2016-08-27
Moved BibDatabaseContext and MetaData to model

oscargus:
Not working because of GroupTreeNode being in logic. Any ideas how to solve this? (Not obvious where to set the cut between logic.groups and model here.)

The parsing and serialization is moved out from MetaData, which looks a bit ugly. The alternative is to move quite a bit of classes to model, so I went for the other approach.


oscargus:
In the second commit groups, much of search, and a few methods from StringUtil is moved. The only logic dependency now is Localization. Is there a clever way to solve this? Localization is needed to generate the description for searches and groups (and some exceptions, although those can be removed I think).

oscargus:
@JabRef/developers I managed to move everything to model without any logic imports. Some things are not so nice. The main problem at the end was Localization. Hence, the name of AllEntriesGroup will have to be propagated. Also, as some error messages were derived from a localized exception message, two of these are currently not localized.

Some observations:
* `getDescription` was never used for the groups
* Some strings missed spaces and translation arguments
* `UnsupportedVersionException` was never used

I have not moved the tests as this was more or less an experiment to see if it was feasible at to what cost. If you think it is a good idea, I will move them as well.

src/main/java/net/sf/jabref/logic/exporter/MetaDataSerializer.java
```diff
@@ -0,0 +1,57 @@
...
+    public static Map<String, String> getSerializedStringMap(MetaData metaData) {
+    
+        Map<String, String> serializedMetaData = new TreeMap<>();
+    
+        // first write all meta data except groups
+        for (Map.Entry<String, List<String>> metaItem : metaData.getMetaData().entrySet()) {
+    
+            StringBuilder stringBuilder = new StringBuilder();
+            for (String dataItem : metaItem.getValue()) {
+                stringBuilder.append(ModelStringUtil.quote(dataItem, ";", '\\')).append(";");
```
Siedlerchr: What does this  ";" and \\ mean?
Maybe move it to a constant?`e.g. Delimiter?

src/main/java/net/sf/jabref/logic/groups/GroupDescriptions.java
```diff
@@ -0,0 +1,120 @@
...
+import net.sf.jabref.model.groups.ExplicitGroup;
+import net.sf.jabref.model.groups.KeywordGroup;
+import net.sf.jabref.model.groups.SearchGroup;
+
+public class GroupDescriptions {
+
+    public static String getDescriptionForPreview(String field, String expr, boolean caseSensitive, boolean regExp) {
+        String header = regExp ? Localization.lang("This group contains entries whose <b>%0</b> field contains the regular expression <b>%1</b>",
+                field, StringUtil.quoteForHTML(expr))
+                : Localization.lang("This group contains entries whose <b>%0</b> field contains the keyword <b>%1</b>",
```
Siedlerchr: I think it would be more readable to use if/else instead of the ternary operators

src/main/java/net/sf/jabref/logic/importer/util/MetaDataParser.java
```diff
@@ -0,0 +1,90 @@
...
+            }
+        }
+        return newMetaData;
+    }
+
+    /**
+     * Reads the next unit. Units are delimited by ';'.
+     */
+    private static Optional<String> getNextUnit(Reader reader) throws IOException {
+        int c;
```
Siedlerchr: Why not declare it as char? Char ist auto represented as int, so no need to convert it back and forth

src/main/java/net/sf/jabref/logic/util/strings/StringUtil.java
```diff
@@ -673,4 +593,16 @@ public static String boldHTML(String input, String alternative) {
...
     }
 
+    /**
+     * Quotes each and every character, e.g. '!' as &#33;. Used for verbatim
+     * display of arbitrary strings that may contain HTML entities.
+     */
+    public static String quoteForHTML(String toQuote) {
+        StringBuilder result = new StringBuilder();
+        for (int i = 0; i < toQuote.length(); ++i) {
+            result.append("&#").append((int) toQuote.charAt(i)).append(';');
```
Siedlerchr: The explicit int conversion can be removed

Siedlerchr:
Looks good! I know that you only copied the methods to new places, but it would be nice to fix those  unnecessary int/char conversions,

src/main/java/net/sf/jabref/logic/util/strings/StringUtil.java
oscargus: No, because then you get `.append(char)` which will not output a number. `Integer.valueOf(char)` may be a more clear alternative?

 (I should have moved it back to the original place ;-)).

src/main/java/net/sf/jabref/logic/importer/util/MetaDataParser.java
oscargus: No idea.

src/main/java/net/sf/jabref/logic/groups/GroupDescriptions.java
oscargus: Makes sense. Not a big fan of ternary either (except for very short ones on rare occasions).

src/main/java/net/sf/jabref/logic/exporter/MetaDataSerializer.java
oscargus: I *think* `";"` is (all) the characters that should be escaped (quoted) and `\` is the character to escape with (which will also be escaped). Constant makes sense.

src/main/java/net/sf/jabref/logic/util/strings/StringUtil.java
Siedlerchr: Maybe better: https://docs.oracle.com/javase/8/docs/api/java/lang/Character.html#getNumericValue-char-

Siedlerchr: I just looked a bit more and saw that the methods purpose is to encode chars into html and fount out that there is a simpler method from apache commons lang (StingEscapeUtils):
http://stackoverflow.com/questions/1265282/recommended-method-for-escaping-html-in-java

lenhard:
I checked out the PR locally and reviewed the code. I think this PR is a good thing that more or less completes the moving to the model package. I was actually surprised to see that moving `MetaData` was (relatively) feasible. I suggest you go ahead and adjust the structure of the tests, too.

Some remarks / things to improve: 
 - Somebody has recreated the event package and added a few events. These should go into the event package in model. Could you perhaps move them as part of this PR and delete the re-introduced top-level event package?
 - I think it would be worthwhile to bundle `MetaData` and its related classes (`SaveOrderConfig`) in a subpackage of model, e.g. `model.metadata`.
 - My IDE marks `MetaData.initializeNewDatabase` as unused. So why not remove it?
 - `BibDatabaseContext` should be moved into `model.database`.
 - I think it would be better to keep `Defaults` in the root package, since it a rather global aspect (a simple one, also). What about changing the type of the `defaults` variable in `BibDatabaseContext` to `BibDatabaseMode`? You would have to modify the constructors, but from what I see it would disentagle `BibDatabaseContext` and `Defaults`.
 - The tie between `BibDatabaseContext` and `DBMSSynchronizer` is not good. We will have to change it sooner or later. However, this need not necessarily be part of this PR.
 - I don't quite like `ModelStringUtil` (and vaguely remember extended discussions with @tobiasdiez  and @simonharrer about such utility classes). How about renaming it to `StringUtil` again? Sure that might result in naming conflicts, but the fact that it is included in `model` already adds that information to that class. And I assume that no class needs to import two `StringUtil` classes anyway, since the methods are just duplicated.
 - I find the propagation of localization not so problematic (especially since no better solution comes to my mind at the moment). It is not in too many positions right now and there will not be many future problems, since the model package is practically complete with this PR.
 - Also the moving of group related aspects (`GroupTreeNode`) is fine by me.
 - Moving parsing/serialization out of the `MetaData` is the right thing to do. After all `BibDatabase` also does not care about this, `BibDatabaseWriter` does. You made the correct decision in my point of view.

I hope this feedback answers your questions :-)

oscargus:
Thanks for the elaborate feedback.

Just some quick partial comments from the phone:

There are methods classes that use both StringUtil and ModelStringUtil,
hence the naming to avoid fully specified paths. Maybe better to duplicate
the code then (now they are moved)?

One out of parsing and serialisation can now easily be moved back
(serialisation I think, since it is only OS.NEWLINE that must be injected
and maybe an odd StringUtil method copied/moved). Not sure what is the best
way. There has also been discussions about self-contained serialisation.

Your package suggestions are fine with me and makes sense. Will implement.

Let's just see how many conflicts I get. ;-)

src/main/java/net/sf/jabref/logic/groups/GroupDescriptions.java
oscargus: I leave the ternaries as they jointly have some structure, despite each being quite hard to read.

oscargus:
Regardind `Defaults` vs `BibDatabaseMode`: I do not really see the point of `Defaults` to start with, but I guess that the idea may be to add more defaults to it. On the other hand,  found this quite interesting piece of code:

``` Java
        Defaults defaults = new Defaults(BibDatabaseMode.fromPreference(Globals.prefs.getBoolean(JabRefPreferences.BIBLATEX_DEFAULT_MODE)));
        BibDatabaseMode mode = new BibDatabaseContext(pr.getDatabase(), pr.getMetaData(), defaults).getMode();
``` 
So either these lines have just propagated through a number of refactorings without someone really looking at them or there is something more to it...

stefan-kolb:
The idea of `Defaults` was to encapsulate more than just the mode inside a Defaults class. Was done by @simonharrer and me on the way back home after JabCon as far as I remember.

oscargus:
OK, then I leave `Defaults` as is for now.

src/main/java/net/sf/jabref/model/groups/SearchGroup.java
```diff
@@ -21,18 +21,26 @@
...
+
     /**
      * Creates a SearchGroup with the specified properties.
      */
     public SearchGroup(String name, String searchExpression, boolean caseSensitive, boolean regExp,
             GroupHierarchyType context) {
         super(name, context);
 
-        this.query = new SearchQuery(searchExpression, caseSensitive, regExp);
+        this.searchExpression = searchExpression;
```
tobiasdiez: I think it is enough that these information are stored in the query. What was your idea behind these changes?

src/main/java/net/sf/jabref/model/groups/SearchGroup.java
```diff
@@ -151,42 +159,13 @@ public boolean isDynamic() {
     }
 
     @Override
-    public String getDescription() {
-        return this.query.getDescription();
-    }
-
```
tobiasdiez: Could you please leave these methods. I think they are used in the `Edit Groups Dialog` and if they are not used, then this is probably a bug.

src/main/java/net/sf/jabref/model/util/RemoveLatexCommands.java
```diff
@@ -1,12 +1,7 @@
-package net.sf.jabref.logic.layout.format;
+package net.sf.jabref.model.util;
 
-import net.sf.jabref.logic.layout.LayoutFormatter;
-import net.sf.jabref.logic.util.strings.StringUtil;
+public class RemoveLatexCommands {
 
-public class RemoveLatexCommands implements LayoutFormatter {
```
tobiasdiez: Why did you removed this as a LayoutFormatter?

src/main/java/net/sf/jabref/model/groups/GroupsParser.java
```diff
@@ -1,16 +1,15 @@
-package net.sf.jabref.logic.groups;
+package net.sf.jabref.model.groups;
```
tobiasdiez: This should be logic, I think (as also the bib database parser/writer are logic)

src/main/java/net/sf/jabref/model/event/EntryChangedEvent.java
```diff
@@ -1,6 +1,5 @@
 package net.sf.jabref.model.event;
```
tobiasdiez: I would actually sort the events in the corresponding package (i.e. model.entry or model.database, model.groups)

src/main/java/net/sf/jabref/model/util/RemoveLatexCommands.java
oscargus: As I in that case had to move the interface to model as well, which didn't really make sense just to have that single function in model. The actual layout formatter is still there in logic though, calling the class in model.

src/main/java/net/sf/jabref/model/groups/SearchGroup.java
oscargus: They were clearly not used. :-) May be a bug, but also a problem when moving them as they would introduce a dependency on logic in model. Better to reintroduce the functionality properly later. (And if you look at the translation strings only the one for "All entries" was actually removed, so everything else is still there (and the full description of "All entries" is never shown as it is only shown when you edit the group and you can't edit that group.)

src/main/java/net/sf/jabref/model/groups/GroupsParser.java
oscargus: I'm not sure if it is possible to move it to logic, but I might have a look when rebasing. The main reason (shorttime) to move things like this to logic is that they introduce many dependencies to logic, so it is easier to move one method to logic than many to model. This one seems OK from that perspective if nothing else.

src/main/java/net/sf/jabref/model/event/EntryChangedEvent.java
oscargus: This probably needs to be discussed, but is a two minute operation once decided. I can see advantages with both approaches.

src/main/java/net/sf/jabref/model/groups/SearchGroup.java
oscargus: To reduce the dependencies and move fewer classes to model. Not sure if it actually did that at the end of the day though or if it was just the result of some temporary refactoring. I'll have a look.

oscargus: Checking again, it was to not having to provide get-methods to the new, minimal, `GroupSearchQuery` as that is only used here and the information might just as well stay on this level. (The reason for introducing `GroupSearchQuery` was to not move even more search stuff to model.)

src/main/java/net/sf/jabref/model/groups/GroupsParser.java
lenhard: I aggree with @tobiasdiez. This should be placed in logic, if possible.

src/main/java/net/sf/jabref/model/event/EntryChangedEvent.java
lenhard: Same for me. There are good reasons for both ways of structuring. @tobiasdiez Please provide a compelling argument, and we will go with your suggestion ;-)

From what I see: Sorting the events into different packages increases cohesion, but also "pollutes" the packages with more classes. What about further structuring, e.g. `model.entry.event`, `model.database.event`, `model.groups.event`?

tobiasdiez:
I'll mark this PR as a discussion point for the next devcall, since I think there are some...well..points for discussion 😄

lenhard:
Our decision at the devcall: @oscargus Please introduce the packages `model.entry.event`, `model.database.event`, `model.groups.event` and sort the relevant events into these packages. Thereafter, you can update the branch and we can merge it into master.

oscargus:
Rebased and implemented changes.

With the recent changes `RemoveLatex` was moved back to `logic` and instead `LatexToUnicode` ended up in `model`.

PR #1875, opened 2016-08-28
InvalidBackgroundColor flickering

grimes2:
In entry editor saving an invalid argument in a field causes a InvalidArgumentException. Using Ctrl-s or File>Save database the field flickers red only for a short moment. This is fixed and the three possibilities to save database are now consistent.


grimes2:
Please label "ready-to-review". Thanks.

src/main/java/net/sf/jabref/gui/fieldeditors/FieldEditorFocusListener.java
```diff
@@ -23,7 +23,7 @@ public void focusGained(FocusEvent event) {
     @Override
     public void focusLost(FocusEvent event) {
         if (event.getSource() instanceof FieldEditor) {
-            ((FieldEditor) event.getSource()).setValidBackgroundColor();
+            //Do nothing
```
stefan-kolb: I cannot comment if this is the correct way to implement the problem right now, but if it is please rephrase the condition to 
```java
if (!(event.getSource() instanceof FieldEditor)) {
...
}
```

stefan-kolb:
@grimes2 Can you please address the comments?

tobiasdiez:
LGTM :+1: (without actually having tested the code)

Could you please rebase so that we can merge it? Thanks

grimes2:
Sorry I missed that. Please merge.

Issue #1633, opened 2016-07-27
Searching for ( in specific fields


oscargus:
JabRef version: Latest developer

Search for e.g. `booktitle=(` without regex or `booktitle=\(` with regex. It seems like the parser does not find the field as it says `any field that contains booktitle=(`. Maybe a weird case, but I actually wanted to do that right now.

Searching for `(` works well though.

simonharrer:
The problem is that this is not detected as a grammar based search, but as a contains-based or simple search. You can circumvent the problem by searching for booktitle="(" which should work as expected. We could improve the search grammar, but this just takes too much time right now. Hence, I close this.

oscargus:
OK! Didn't think of using "", but clearly that is a good (enough) approach!

ajbelle:
Although this is closed I post here as a user report as I think my experience is related to the “uncertain” search nomenclature that is required.  Also related is issue #1505 regarding the apparent heterogeneous regex-regular expression implementation. IMHO being able to search reliably IS ESSENTIAL and of high priority for large database situation.
I was attempting to do a general search with a specific exclusion (word1 and not word2) and conclude that at best the current help files http://help.jabref.org/en/SearchHelp and http://jabref.sourceforge.net/help/SearchHelp.php for `Search` are misleading and incomplete. The poor search behaviour has been present within JabRef and the recent changes have unfortunately not corrected these issues it seems.

Hover-over the search entry shows the full logic in words. This is a great feature, but doesn't seem to reflect what the help file suggest is should, not what Search seems to do.

""word1 and not word2""  selects all entries and highlights all those words including 'and' and 'not'.
""word1 not word2"" selects all entries and highlights all those words including 'not'.

If I enter "word1 and not field2=word2""  eg: "flow and not memo=PhD" search totally fails with hover-over showing "... flowand and  not and memo=PhD"  It appears to not recognise the and, instead ignoring the space and adding it to the word, then inserting extra ands!

I concluded the algorithm has be rewritten to look for `Field2=word1...Field2=word2... ` but how do you specify to search all fields for one word? From the helpfile the following should work
 title|keywords = "image processing" 
 title--keywords = "image processing"  but neither works.
 eg: (author = miller or title|keywords = "image processing") and not author = brown cut and paste from the help file does work (neither with a hyphen or a double hypen).

Interestingly `Search` automatically decides content in a non-regex entry is regex, but it doesn't seem to work.
[...]

PR #1876, opened 2016-08-28
Added anykeyword and anyfield pseudo search fields

oscargus:
See https://github.com/JabRef/jabref/issues/1633#issuecomment-236909025

Will add tests etc if it is a good idea and works.


src/main/java/net/sf/jabref/logic/search/rules/GrammarBasedSearchRule.java
```diff
@@ -146,10 +148,28 @@ public boolean compare(BibEntry entry) {
                 return matchFieldValue(entry.getType());
             }
 
+            // special case for searching a single keyword
+            if (fieldPattern.matcher("keyword").matches()) {
+                if (entry.hasField(FieldName.KEYWORDS)) {
+                    List<String> keywords = new ArrayList<>(entry.getKeywords());
```
Siedlerchr: Why not a simple stream on the list with anyMatch (returns a boolean)?

Siedlerchr: And you could even use that together with the prev line if you use your getFieldOptional method

oscargus: The getFieldOptional is no more. ;-)

You mean:
`return entry.getKeywords().stream().anyMatch(this::matchFieldValue);`
?

Could work. In the most recent version I swapped order of hasField and matcher as I expected the first one to be faster. Still, I think getKeywords() returns an empty list if the field is not set, so doesn't really need it.

oscargus:
I'm thinking that the distinction between keyword and keywords is not enough. What do you suggest? anykeyword? singlekeyword?

oscargus:
Also, when writing a bit more search queries for the tests: is `allfields` a good name? Seems like `anyfield` would be better. (And then `anykeyword`)

`anyfield contains fruit and anykeyword matches apple`

Siedlerchr:
Difficult question... To me `anykeyword`  would be okay in this context and it's shorter than singlekeyword.
E.g thinking of  `anykeyword == "test"` it has clear semantics

Siedlerchr:
Be aware of backwards compatibility. People may be used to "allFields" already..

oscargus:
`allfields` has been around for 8 hours so I doubt it has been that well established. ;-)

src/main/java/net/sf/jabref/logic/search/rules/GrammarBasedSearchRule.java
```diff
@@ -142,16 +141,25 @@ public Comparator(String field, String value, ComparisonOperator operator, boole
...
 
-            List<String> matchedFieldKeys = fieldsKeys.stream().filter(matchFieldKey()).collect(Collectors.toList());
+            // special case for searching allfields=cat and title=dog
+            if (!fieldPattern.matcher("anyfield").matches()) {
+                // Filter out the requested fields
+                fieldsKeys = fieldsKeys.stream().filter(matchFieldKey()).collect(Collectors.toSet());
+            }
 
-            for (String field : matchedFieldKeys) {
+            for (String field : fieldsKeys) {
```
Siedlerchr: You could here do the same as above, just with `return fieldsKeys.stream().filter(this::matchFieldValue).isPresent()`
The nice thing is that all operations like map., filter etc  are only performed if the value inside the optional is present, so absolutely type safe

src/test/java/net/sf/jabref/logic/search/SearchQueryTest.java
```diff
@@ -66,4 +66,51 @@ public void testSearchingForOpenBraketInBooktitle() {
         assertTrue(searchQuery.isMatch(e));
     }
 
+    @Test
+    public void testSearchMatchesSingleKeywordNotPart() {
+        BibEntry e = new BibEntry(IdGenerator.next(), BibtexEntryTypes.INPROCEEDINGS.getName());
+        e.setField("keywords", "banana, pineapple, orange");
```
Siedlerchr: Maybe useful to add some umlaut tests?

Siedlerchr:
Overall lgtm, you could replace that one case with a stream too and maybe add some test regarding umlauts...(not sure if they are actually handled correctly)

tobiasdiez:
I find the `anykeywords` field a bit counter-intuitive. I would change the search syntax to search for the exact expression vs "contains":
- `keywords == test`: matches only when the content is equal to `test`.
- `keywords <> test`: matches if it contains `test` as a keyword. (or maybe a different symbol in place of `<>`)

oscargus:
@JabRef/developers Opinions?

Personally, I think that adding a search operator which only can be used for a specific field is a bit counter-intuitive.

Siedlerchr:
I like ` anykeywords` better than a new operator.  Especially the search query can be read as a whole sentence and is for the user directly recognizable and has a clear semantic, e.g.` anykeyword contains test `

For an operator I would expect it to work everywhere, not containing any special semantics for a specific field.

tobiasdiez:
But one could also have different fields with values separated by some character ("keyword fields"). For example, it would make sense to search for entries in specific groups or some user defined fields. But one probably doesn't want to define `any*`-pseudofields for all fields.

oscargus:
You do have a point, but:
* there are quite few such fields at the moment
* the separator discussion is ongoing and far from concluded, less
implemented
* group hierarchies/restoring need to be improved
* and, also important in this context, I have no idea how to introduce a
new operator

Hence, this has to be postponed and done by someone else. If it is likely
to happen we can just put this on hold and see. If it doesn't happen before
3.7 is released, I suggest we merge this in.

Clearly, it would also be much better if `anyfield` didn't have to exist.
but any term without a field specified does search in every field, even if
fields are specified elsewhere. Again, a hypothetical better solution
compared to an OK existing.

(Btw, I think that one can already search for groups, although it will be
in the "contains" way to make sense.)

src/test/java/net/sf/jabref/logic/search/SearchQueryTest.java
oscargus: It will work for what you write. Hence, `ö` will match `ö`, but not `\"{o}`, and vice versa. Once the LaTeX free BibEntry is functioning it should work both ways.

koppor:
Alternative: Use [Apache Lucense's Syntax](https://lucene.apache.org/core/2_9_4/queryparsersyntax.html)

lenhard:
We discussed the following at the devcall: The search parser should be extended to support an `any*`  syntax, e.g. it checks if a search term starts with `any`, if so it cuts of the `any` part and tries to use the remaining part as a field name. In this fashion, we would have support for the `any`-functionality for arbitrary fields, which sounds quite cool. Effortwise, it should just require some String manipulation in the query parser. @oscargus Would you be willing to implement this here?

The long term goal would be to move towards a sophisticated query parser, such as Lucene. However, the effort for this and the required time is quite unsure.

oscargus:
I do not get it. First, `anyfield` does not apply to the logic described and still needs to be treated separately (and this is really the main thing of this PR). Second, the whole point of `anykeyword` is that one can *split* the keywords field into smaller parts (and `anykeywords` sounds, ehh, strange...) and match one of these parts (exactly, as contains is more or less the same thing as before). Yes, it might work for e.g. `authors` but I cannot really see the point or how. In that case it may make more sense to be able to match the last name or first name or so, rather than the whole name. The key thing here is how to split the fields and that is field dependent (I can see a few fields where it might work: `related`, `authors`, `file` and there are probably some more. Most it won't make sense for.)

So, no, as I do not understand how it should be done (in a way that makes sense).

koppor:
Dumb question: What currently happens when searching for a keyword? Isn't
substring matching enough?

oscargus:
Say you have two keywords `aa` and `aaa`. Currently you cannot search for `aa` without also finding `aaa` (assuming there is more than one keyword for each entry, so you need to search for `keywords=aa`). With this PR you can do `anykeyword==aa` and only get entries where one keyword matches `aa` or `anykeyword!=aa` to get entries without that keyword. (`anykeyword=aa` will get both `aa` and `aaa`, so in that case there is not difference to before.)

src/main/java/net/sf/jabref/logic/search/rules/GrammarBasedSearchRule.java
koppor: Not quite. This has to be wrapped in an `if` which executes `return true;`

koppor:
OK for me. Just resolve https://github.com/JabRef/jabref/pull/1876/files#r78694966 and then we can merge this in.

tobiasdiez:
@oscargus based on your description, the problem is not specific to the keywords fields. For example, I can't find the author `aa` without also finding `aaa` and the same for  `aa` in the abstract field. This is why I suggested to use a new comparison operator which checks that the search string is exactly a "word" and not just a substring (and for the keywords field we maybe can be so intelligent and only take the keyword separator as word separator). 

I think the standard solution is: only match whole words (i.e. `aa` only matches `aa` but not `aaa`) and use wildcard queries like `aa*` to match `aa` as well as `aaa`.

Siedlerchr:
:+1:  for the wildcard query option. e.g. *  and ?

oscargus:
As I wrote there are a few more fields, yes, put providing a generic split
is more or less impossible (` and `, `, `, `,`). In addition there are more
fields that can't be split than the opposite. `anyyear`? `anyissue`?
`anyabstract`? For author, you more likely want to search for
`lastname==aa` than `anyauthors=="a. aa" or anyauthors=="aa, a."` (not to
mention you actually search for a specific author, not a specific
lastname), so adding something similar for names would make sense, but I do
not see that as `anyauthors`.

My suggestion is that we add `any`-support for the fields where it makes
sense and as this needs to be done specifically for each (class of) field
it can be done incrementally.

koppor:
:+1: - @oscargus please solve https://github.com/JabRef/jabref/pull/1876/files#r78694966, then we can merge and then we can work on other fields in another PR. I'm trying to keep the PRs small and want to merge them fast.

tobiasdiez:
The new github review is a bit strange... no idea why there are now two comments and one approval. It should be actually "one comment" :smile:

oscargus:
Correct, `fruity and keywords=apple`does not work, that was more or less the issue to start with. Feel free to propose a PR solving it in the proper way (I quote myself: "The allfields approach should be quite feasible to implement, although somehow it would be even better if not specifying a field leads to that behaviour (that I do not know how to do though).").

Incomplete, yes. But not really possible to generalize. Hence the need for more PRs improving the search, which, I guess, can always be improved.

Issue #662, opened 2016-01-11
Import automatically pdf annotations into the note field


Olifair:
This is something that is done by Mendeley internally, but I think I have seen this in one of the other programmes (don't remember if Zotero or Docear) on standard pdf annotations : 
standard pdf comments or highlighted text (as done with and compatible between Adobe, Foxit, X-Change viewer etc) is taken automatically from linked pdf and shown in the note field. With Mendeley, which uses an internal pdf reader, this is click-able, so that then the pdf will then be opened at the position of the comment /highlight.

arurke:
+1

koppor:
This is a modern version of [sf feature request 835](https://sourceforge.net/p/jabref/feature-requests/835/). I will close that request and point to here.

stefan-kolb:
We wanted to have an option to show PDF comments inside JabRef (implemented by Stupro), but not in that case that it gets written to the DB?!

oscargus:
I'm wondering if the note field is the best choice since BibTeX always adds this to the reference. I would suggest comment or review.

Braunch:
I see some troubles using already existing fields. If the user wants to use the review function, the pdf comments will be in the way. Also the review field is editable and we don't want to have editable pdf comments, so that the user won't accidentally change the comments when he doesn't want to. The pure text field of the notes and the review field can be very confusing because you can not show the comments structured in different GUI components. My alternative suggestion would be create a special tab in the entry editor and only load the comments from the file, when the tab is opened up and not storing them into the db at all.

tobiasdiez:
Sounds like a good solution, @Braunch. JabRef could internally store the extracted PDF comments if the extraction is time-consuming. But I also wouldn't write it into a BibTex field (because then you run into all kinds of synchronization problems).

Olifair:
If the extracted pdf comments are stored internally, this has also the advantage that the inverse functionality could be provided : if I open up a reference, and then see a comment I'm interested in, Jabref could open the pdf at the correct page in an external viewer ? This would be great for people who work rather with books than with articles.

lenhard:
Tracked in #2545

The option to open a PDF at a specific comment is not part of this though, since this is quite hard to support for arbitrary combinations of operating systems and pdf readers.

lenhard:
The pull request is merged into the master branch now and the entry editor displays an additional tab with pdf annotations (if present).

Hence, I am closing this issue.

Issue #139, opened 2015-09-02
Synchronize PDF highlighted text, comments, and bookmarks


koppor:
[Docear Desktop](https://github.com/Docear/Desktop) has the feature to synchronize highlighted text, comments, and bookmarks from a PDF with Docear. The same should be done at JabRef.

Highlighted text:
![pdfviewer](https://cloud.githubusercontent.com/assets/1366654/9620492/f71985e4-511a-11e5-99db-1607363dae14.png)

Rendering in Docear:
![docear](https://cloud.githubusercontent.com/assets/1366654/9620497/fb3b09e0-511a-11e5-924c-4285fd3da899.png)

(via @Joeran, @Docear)

koppor:
Refs #662

Joeran:
the public docear version of cannot extract "real" highlighted text but only text that is stored in "pop-up" boxes
![image](https://cloud.githubusercontent.com/assets/2410381/15660017/4ae65654-2710-11e6-94d5-87b4773f379a.png)
we recently paid a student to enhance docear, so docear could also extract the "real" highlighted text. unfortunately, the student never 100% completed the job (the new function has some bugs, but they relate mostly to Docear not the extraction of highlighted text itself). 

if you are interested in using that add-on for JabRef, let me know. Otherwise, you are welcome to just copy the functionality from Docear's public repository on GitHub. https://github.com/Docear

tobiasdiez:
This might be helpful: http://stackoverflow.com/questions/33253757/java-apache-pdfbox-extract-highlighted-text

lenhard:
Now tracked in  #2545.

lenhard:
#2545 is merged now, so we can display pdf comments and highlighting.

Hence, I am merging this issue.

PR #1096, opened 2016-04-04
Update pdfbox to 2.0.0 and migrate from jempbox to xmpbox

lenhard:
This PR addresses #1004

There are significant changes in the APIs from jempbox and xmpbox. The current state of this PR is just a plain translation from jempbox to xmpbox to get the code to compile. The tests are not working yet, so there are probably some errors in the translation that need to be fixed. Also, travis seems to have problems with xmpbox.

Comments from anyone who is familiar with XMP handling are very welcome.

- [ ] Change in CHANGELOG.md described

PR #1883, opened 2016-08-29
[WIP] Show PDF-Comments in JabRef

ayanai1:
Extract Comments from linked PDF-Files and show them in JabRef.

See here for related issues:
https://github.com/JabRef/jabref/issues/662
https://github.com/JabRef/jabref/issues/139
- [ ] Check documentation status (Issue created for outdated help page at [help.jabref.org])
- [ ] internal qs

Features TODO:
- [x] Show comments and highlight of multiple attached pdf-files
  - [x] Meta info inclusive
- [ ] Open File at annotation position
  - [x] Windows
  - [x] Linux
  - [ ] MacOS
- [x] Cache loaded annotatons
- [x] Copy Meta and Content info to clipboard
- [x] Import annotations when tab is opened only (releasable proof-of-concept)


src/main/java/net/sf/jabref/gui/entryeditor/PdfCommentsTab.java
```diff
@@ -66,8 +66,9 @@ private void setUpPdfCommentsTab() {
             DefaultListModel<String> listModel = new DefaultListModel<>();
             commentList.setModel(listModel);
             PdfCommentImporter commentImporter = new PdfCommentImporter();
-            HashMap<String, String> importedNotes = commentImporter.importNotes(field.get());
-            for (String note : importedNotes.values()) listModel.addElement(note);
+            String pdfPath = field.get().replaceFirst(".*?:", System.getProperty("file.separator")).replaceAll(":PDF", "");
```
Siedlerchr: Just a hint, there is a convenience method in `FileUtil `which does the job for you and because the file field could contain more than one file entry.
getListOfLinkedFiles(...) just look at the call hierarchy of that method to see how to use it...

Braunch: Thank you for that! Is it a valid use case that more than one pdf file is linked? If so, I would have to take care about parsing multiple files and we need a UI that is capable of representing that.

Siedlerchr: I am not sure about that, well at least it is possible to add/have multiple pdf files ;) I think the typical workflow is to attach one pdf and some additional resources (e.g. maybe source code archive or some other data).

Braunch: But still I think we should take care about that. The simplest way would be to create a new pdf tab for every attached pdf file or a drop down to switch files. I will have a discussion with @ayanai1  and @mairdl  on that.

tobiasdiez: Or have a list of all comments but grouped by the containing pdf file. Probably 90% of the entries have just one linked pdf file and thus the UI should primarily support this.

src/main/java/net/sf/jabref/gui/entryeditor/PdfCommentsTab.java
```diff
@@ -0,0 +1,113 @@
...
+import net.sf.jabref.gui.BasePanel;
+import net.sf.jabref.gui.JabRefFrame;
+import net.sf.jabref.logic.pdf.PdfCommentImporter;
+import net.sf.jabref.model.entry.BibEntry;
+import net.sf.jabref.model.entry.FieldName;
+
+public class PdfCommentsTab extends JPanel {
+
+    private JPanel informationPanel = new JPanel();
+
```
Siedlerchr: Please remove the empty lines between the variable declarations or better group them (e.g. all GUI/panel related stuff first) then a new line and next group.

koppor:
Please also try to consider the wishes at http://discourse.jabref.org/t/seamless-crossreferencing/100/6 - if possible.

mairdl:
This is a screenshot of our current version for the PDF-Tab-UI:
pdf_tab_ui

The left panel shows the comments and highlighted text of the PDF file. On the right side we have an information panel. There you can select from the attached PDF files. Beneath the file-name you can find the author, date, page and content from the selected comment. Also, you can open the PDF via the button (`Öffne PDF`) or you can copy the content of the comment into the clipboard (button `In die Zwischenablage kopieren`).

koppor:
Please make an English screenshot and please add pseudo data

You can create PDFs using the templates available at https://github.com/latextemplates/

If you compile https://github.com/latextemplates/uni-stuttgart-dissertation-template, you should get a PDF with comments, too.

This approach ensures that you use PDFs without copyrighted text. The PDFs you used for the screenshots should also be used at test cases.

koppor:
Is there an updated screenshot available?

mairdl:
@koppor yes, now it's ready:

This is a screenshot of our current version for the PDF-Tab-UI:
pdf_tab_ui

The left panel shows the comments and highlighted text of the PDF file. On the right side we have an information panel.

On the left panel you can choose from a dropdown menu which attached PDF-file you want to open.
dropdown_filename

If the selected annotation is linked with a highlighted text or the other way around, the corresponding entry will be colored in light blue. Also, the content box will combine the information of both.

Braunch:
Are there any comments on the concept? Did someone try it out? It is the last week of our project and it would be great if we at least get a working concept that can easily be integrated into JabRef afterwards.

tobiasdiez:
I didn't tried it out, but the screenshot above looks good. Do you need feedback to a specific feature/implementation detail?

Braunch:
No the general impression and if you believe the implementation is useful for the user. The only specific thing I can think of is if showing both highlighted text and the annotation linked to it in the comments list on the left makes sense or if it's redundant. Maybe one entry for both and showing the linked content in the content box on the right is enough?

tobiasdiez:
Of course it is useful if the user has PDFs with comments / highlights. I think it is enough to have on entry for the highlighted text and its annotation in the left list, details then right.

tobiasdiez:
I just had a quick look over the code and in general it looks good. 
The only bigger thing is, that you often coded against an instance instead of an interface (for example PdfCommentImporter vs a general CommentImporter interface). This makes it hard to support different file formats. For example, djvu also has annotations (which might be a better word instead of "comments" ?) and it would be nice if we could (later) add support for these, see https://github.com/gthurm/javadjvu/blob/920f36db24f70b7dd9038073ead9c5ad476d8d7a/src/com/lizardtech/djvu/anno/DjVuAnno.java#L361.

Braunch:
I now introduced an Interface for comment importers and renamed the pdfComment class to fileAnnotation. For consistency I also renamed the tab to FileAnnotationTab and changed the Strings referring to PDF to something with File instead. I had athought about creating an AbstractFileAnnotation class, but I think the meta info you want to get from the annotation will probably be the same no matter the format, so you just need to map the annotation to the model classes parameters.

Braunch:
I have necountered a problem with the travis-ci check. It fails with: `is net.sf.jabref.logic independent of java.awt?] FAILED` 
The only awt class used is the awt.geom.Rectangle2D which is needed to extract the highlighted text from a pdf. Can I just ignore the failing check?

Siedlerchr:
@Braunch  The tests are there to ensure a separation of logic and gui. And as normally the awt classes are associated with gui stuff, they should be not used in logic. 
Mabye @lenhard  has an idea for the architecture

lenhard:
Sorry, but no GUI technology in logic. The check cannot be ignored. I know this is not your fault and you get the dependency from pdfbox. This is rather bad design on side of pdfbox/awt, but we will not bend our architectural constraints for that. Is it possible to implement the text extraction without the awt dependency? If not, than the code needs to go into the gui package.

What version of pdfbox is that? We have quite some problems with the library, because the XMP support depends on it and we cannot upgrade it without losing XMP support. This is where our "dependencys outdated"-badge comes from.

koppor:
Quick hack: net.sf.pdfboxbridge package? 😇

@braunch you find @lenhard s PR for updating PDFBox on the on-hold labelled
closed PRs

Braunch:
The problem is that pdfBox can not extract highlighted text, but text within an area. This area has to be represented by a Rectangle2D. Putting the code into the GUI would be a solution, but it bends the architectural constraints anyway, only the other way round.

Braunch:
I will have a look at the PR as soon as I am finished with the caching.

lenhard:
@Braunch You are entirely correct in the view that the gui package is a mess. It is the infamous part of JabRef, where "anything goes" and where we quarantine the toxic stuff. That is why putting the code into gui is less of a problem than compromising the logic package.

Here is my PR on upgrading pdfbox: https://github.com/JabRef/jabref/pull/1096

@koppor Wasn't there an attempt in stupro to encode all XMP stuff into dublin core? That would solve our pdfbox XMP problems.

koppor:
@StuPro could not work on that due to time constraints and the task was
also to think about migrations, current use cases etc...

Braunch:
Would it be possible to create an exception for awt.geom package? This package seems to contain only geometry and logic related classes. There is no real GUI stuff inside and if awt geometry should be needed in the future you would be forced to do all the geometry logic in the jabref.gui packages. I am implementing @koppor s quick hack but to me it looks wrong to have a package with the same architectural hierarchy than net.sf.jabref that only contains one method.

lenhard:
Why would we need geometry in JabRef? In my honest opinion, the use case here is already quite weird.

As always in JabRef, things will be decided by consensus of @JabRef/developers I'll vote for making no exception, though.

Braunch:
Ok then. I pushed the quick hack with the bridge to awt.

tobiasdiez:
I support @Braunch here and would vote for an exception for `java.awt.geom`.
Reasoning:
- The `PdfAnnotationImporterImpl` class analyzes pdf files and thus is correctly placed in the logic package (its the same as `BibParser` which analyzes bib files)
- `awt.geom` has no gui code but is a model class in our terminology (the only reference to gui is that the represented objects have a geometric interpretation)
- If the same code would be placed in a package with a different name, then there wouldn't be any problem... but names are no more than smoke and mirrors.
- The hack/workaround https://github.com/JabRef/jabref/pull/1883/commits/31d975b5e81f04937a10a0f9ec547fdd55b447b1 is quite ugly. In the end, the code still depends on `awt.geom`, just that the dependency is indirect and thus our tests are not able to detect it.
- We will have the same problem with the JavaFX `ObservableLists` which provide a convenient implementation of the list interface (firing events upon update). But they are placed in `
javafx.collections` and thus would not be allowed in logic (if we decide to ban `javafx` as we did with `swing/awt`.)

Siedlerchr:
I would follow the argumentation by @tobiasdiez. As he says names are just hollow words. Add an exception for awt.geom. I took a quick look at the package api dog and it seems like it contains not any logic for drawing, but just proving a model

src/main/resources/l10n/JabRef_zh.properties
```diff
@@ -2278,6 +2284,13 @@ Full_text_document_download_failed_for_entry_%0.=
 Look_up_full_text_documents=
 You_are_about_to_look_up_full_text_documents_for_%0_entries.=
 last_four_nonpunctuation_characters_should_be_numerals=
+
+Author=
+Date=
+File_annotations=
+Show_file_annotations=
+Adobe_Acrobat_Reader=
+Sumatra_Reader=
```
koppor: The name is Sumatra PDF (https://www.sumatrapdfreader.org/free-pdf-reader-de.html) and does not need to be translated, does it? - See asian Sumatra PDF page: https://www.sumatrapdfreader.org/free-pdf-reader-ja.html

lenhard:
So we decided in the devcall that we will create an exception for the `awt.geom` package. So please modify the code of `ArchitectureTests` to exclude the package and I will have a look to see if I like the way in which you implemented it ;-)

Braunch:
Ok I will change that back and add the exception.

Braunch:
I added an 'exceptionStrings' map that uses the first package as a key and a list of exceptions as value. When checking the imported packages I then filter for packages, that have the first package as key and are imported there. If they are exceptions, they are not added to the list of files. Adding more exceptions is easy. Just add them to the list in the map that belongs to the corresponding key.

lenhard:
Ok, the test works. In my honest opinion that is a good example that lambdas are hard to understand when they get big. So, could you perhaps just refactor the test a little bit to make it more easily understandable? 

What I am thinking about is extracting the boolean parts of the lambda into predicates that you define at the start of the method and then just reusing that, i.e.:
```
Predicate<String> isExceptionPackage = (s) -> (exceptionStrings.get(firstPackage).stream()
                                .filter(exception -> s.startsWith("import " + exception)).findAny().isPresent()
                        );
```
and then you can call
```
...
 !(isExceptionPackage.test(s))
...
```

Please add and use this predicate and build at least one more nicely named one for another boolean condition. Then the lambda will be much more readable and I promise that I will be happy :-)

P.S.: The localization tests fail, so this needs to be fixed before merging as well.

Braunch:
That is a good idea. The filters are getting unreadable. But I think I have to rewrite the predicate a bit different, because the predicates are the parameters for the filter and the one you wrote is just one condition for the predicate of another filter so a boolean is expected. I put the whole predicate in it (still a big filter but already way better to read) and extract the isPackage predicate from the filter above too.

lenhard:
I just had a look at the `ArchitectureTests` and am happy with them. So, from my side this is good to go. Someone else should double check the main functionality, though.

tobiasdiez:
Is this PR ready for review? (I saw a few "TODO" statements in the code, so I suspect it is not yet finished.)

Braunch:
No its not ready. I dint try the open pdf on an other setup than linux/acroread
So this is TODO.

koppor:
@Braunch Any news? Any timeline?

lenhard:
The code is moved over to #2545 and this can be closed.

PR #1884, opened 2016-08-29
Improve duplicate detection

stefan-kolb:
- [ ] strict comparison does not care about entry type. Is this intended?

``` java
    @Test
    public void noStrictDuplicateForDifferentTypes() {
        BibEntry e1 = new BibEntry("1", "article");
        BibEntry e2 = new BibEntry("2", "journal");
        assertEquals(0, DuplicateCheck.compareEntriesStrictly(e1, e2), 0.01);
    }
```
- [ ] Correlate by words has a strange algorithm. Only works good for words appended at the end of the String.

``` java
     * Compare two strings on the basis of word-by-word correlation analysis.
     * TODO: strange algorithm as when there are only words inserted this gives a bad value, e.g.,
     * a test -> this a test (0.0)
     * characterization -> characterization of me (1.0)
```
- [ ] We need a small benchmark for the duplicate testing, i.e., a DB that has all kinds of expected duplicates.

### DIFF
--- a/src/main/java/net/sf/jabref/model/DuplicateCheck.java
+++ b/src/main/java/net/sf/jabref/model/DuplicateCheck.java
@@ -24,6 +24,8 @@ import org.apache.commons.logging.LogFactory;
 
 /**
  * This class contains utility method for duplicate checking of entries.
+ *
+ * TODO: http://hpi.de/naumann/projects/data-quality-and-cleansing/dude-duplicate-detection.html
  */
 public class DuplicateCheck {
     private static final Log LOGGER = LogFactory.getLog(DuplicateCheck.class);
--- a/src/main/java/net/sf/jabref/model/DuplicateCheck.java
+++ b/src/main/java/net/sf/jabref/model/DuplicateCheck.java
@@ -3,6 +3,7 @@ package net.sf.jabref.model;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
@@ -16,6 +17,7 @@ import net.sf.jabref.model.entry.FieldName;
 import net.sf.jabref.model.entry.FieldProperty;
 import net.sf.jabref.model.entry.InternalBibtexFields;
 
+import info.debatty.java.stringsimilarity.Levenshtein;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
@@ -23,18 +25,18 @@ import org.apache.commons.logging.LogFactory;
  * This class contains utility method for duplicate checking of entries.
  */
 public class DuplicateCheck {
-
     private static final Log LOGGER = LogFactory.getLog(DuplicateCheck.class);
 
-    /*
-     * Integer values for indicating result of duplicate check (for entries):
-     *
+    /**
+     * Enumeration for indicating the result of a duplicate check
      */
-    private static final int NOT_EQUAL = 0;
-    private static final int EQUAL = 1;
-    private static final int EMPTY_IN_ONE = 2;
-    private static final int EMPTY_IN_TWO = 3;
-    private static final int EMPTY_IN_BOTH = 4;
+    private enum CheckResult {
+        NOT_EQUAL,
+        EQUAL,
+        EMPTY_IN_ONE,
+        EMPTY_IN_TWO,
+        EMPTY_IN_BOTH
+    }
 
     public static double duplicateThreshold = 0.75; // The overall threshold to signal a duplicate pair
     // Non-required fields are investigated only if the required fields give a value within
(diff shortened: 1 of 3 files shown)


src/main/java/net/sf/jabref/model/DuplicateCheck.java
```diff
@@ -16,25 +18,28 @@
 import net.sf.jabref.model.entry.FieldProperty;
 import net.sf.jabref.model.entry.InternalBibtexFields;
 
+import info.debatty.java.stringsimilarity.Levenshtein;
```
oscargus: Is it worth having this build dependency as it basically does what the removed code below does and adds quite a bit of other unused methods? (I know @koppor was also skeptic.)

oscargus:
Good that you give it a go!

src/main/java/net/sf/jabref/model/DuplicateCheck.java
stefan-kolb: We already have this dependency inside the code that's why I reused it. For the general dependecy question, I'm not 100% sure as we might use more distance measures and the library is probably not very large. So why should I reinvent the wheel here.

oscargus: Yes, from that perspective it is OK for sure and we can of course revive the old code if required later. Oh, here I found it: https://github.com/koppor/jabref/issues/131

koppor: Just go ahead. No worries. We migrate to another library if the functionality is the same. If not, we just keep the dependency. Reason for migrating is only https://github.com/koppor/jabref/issues/135

src/main/java/net/sf/jabref/model/DuplicateCheck.java
```diff
@@ -63,28 +66,28 @@
...
-        EntryType type = EntryTypes.getTypeOrDefault(one.getType(), bibDatabaseMode);
 
-        // The check if they have the same required fields:
-        List<String> var = type.getRequiredFieldsFlat();
+        // check for equal required fields
+        EntryType entryType = EntryTypes.getTypeOrDefault(one.getType(), bibDatabaseMode);
+        List<String> requiredFields = entryType.getRequiredFieldsFlat();
         double[] req;
-        if (var == null) {
+        if (requiredFields == null) {
```
koppor: Are we really returning null? I think, we ommitted null, thus also line 79 is not covered by tests at all.

src/test/java/net/sf/jabref/model/DuplicateCheckTest.java
```diff
@@ -77,14 +129,31 @@ public void testDuplicateDetection() {
     }
 
     @Test
-    public void testWordCorrelation() {
-        String d1 = "Characterization of Calanus finmarchicus habitat in the North Sea";
```
koppor: Why is this test removed? I don't find "North Sea" in the new tests. The method `DuplicateCheck.correlateByWords` still exists.

koppor:
Strict comparison: We should also consider the type.

Correlate by words: I think, you did not find any other algorithm? Can't an existing algorithm be used? Something like counting the similar and the different words and dividing by max(len1, len2)?

Small benchmark: Can we do that as a separate issue? Maybe in the koppor-repo? :innocent:

src/test/resources/net/sf/jabref/model/duplicates.bib
```diff
@@ -0,0 +1,17 @@
...
+
+@InProceedings{dupJournalTechreport,
+  author    = {Kolb, Stefan and Wirtz, Guido},
+  title     = {This was published at a Workshop and a Journal with the same Title},
+  booktitle = {Proc. Workshop Cloud},
+  year      = {2016},
+}
+
+@Article{dupJournalTechreport,
+  author  = {Kolb, Stefan and Wirtz, Guido},
```
koppor: Can we also add a test where the authors are written in a different serialization? Firstname Lastname?

tobiasdiez:
What is the state of this PR?

Another improvement to the duplicate algorithm concerns how DOI, ISBN and edition fields should be weighted:
- Entries with the same DOI or ISBN are almost surely duplicates although they might have different titles or authors (Citavi [is doing something similar](https://www.citavi.com/sub/manual4/en/eliminating_duplicates.html))
- Entries which are almost identical but for edition or volume are probably no duplicates (but just a different version of the same book/article)

src/main/java/net/sf/jabref/model/DuplicateCheck.java
koppor: Since JabRef 3.6 is close to be available in Debian/unstable, there is no need for a library migration.

lenhard:
And again: @stefan-kolb Is there an advance here? I think at some point, we should either try to proceed with the old PRs or scrap them.

stefan-kolb:
No time right now. There are some point that we should fix however. You can close if you want tho.

PR #1894, opened 2016-08-30
Implemented title and camel key modifiers

oscargus:
Implemented #1506 


CHANGELOG.md
```diff
@@ -13,13 +13,14 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
...
 - Added integrity check for fields with BibTeX keys, e.g., `crossref` and `related`, to check that the key exists
-- [#1496](https://github.com/JabRef/jabref/issues/1496) Keep track of which entry a downloaded file belongs to
 - Made it possible to download multiple entries in one action
+- [#1496](https://github.com/JabRef/jabref/issues/1496) Keep track of which entry a downloaded file belongs to
+- [#1506](https://github.com/JabRef/jabref/issues/1506) It is possible to apply two new key modifier `title` for Title Case and `camel` for Camel Case (difference is that title case will leave prepositions etc in lower case) 
 
 ### Fixed
 - Fixed NullPointerException when opening search result window for an untitled database 
-- Fixed #1757: Crash after saving illegal argument in entry editor
 - Fixed[#1663](https://github.com/JabRef/jabref/issues/1663): Better multi-monitor support
```
Siedlerchr: Space after fixed

src/main/java/net/sf/jabref/logic/bibtexkeypattern/BibtexKeyPatternUtil.java
```diff
@@ -519,6 +521,13 @@ public static String applyModifiers(String label, List<String> parts, int offset
                     }
                     resultingLabel = abbreviateSB.toString();
 
+                } else if ("camel".equals(modifier)) {
+                    // CamelCasingOfEveryWord
+                    Title title = new Title(resultingLabel);
+                    title.getWords().stream().forEach(Word::toUpperFirst);
```
Siedlerchr: If getWord returns a list you can also use the new fancy replaceAll overload instead of the forEach.  But in this case it doesn't really matter ;)

https://docs.oracle.com/javase/8/docs/api/java/util/List.html#replaceAll-java.util.function.UnaryOperator-

Siedlerchr:
LGTM,  just some misisng space in the changelog  :+1:

src/test/java/net/sf/jabref/logic/bibtexkeypattern/MakeLabelWithDatabaseTest.java
```diff
@@ -285,4 +285,52 @@ public void generateKeyTitleAbbr() {
         BibtexKeyPatternUtil.makeLabel(metadata, database, entry, preferences);
         assertEquals(Optional.of("AapoJ"), entry.getCiteKeyOptional());
     }
+
+    @Test
+    public void generateKeyTitleTitle() {
+        preferences = new BibtexKeyPatternPreferences("", "", false, true, false, pattern);
+        DatabaseBibtexKeyPattern bibtexKeyPattern = new DatabaseBibtexKeyPattern(pattern);
```
tobiasdiez: I think it makes sense to move `preferences = ...; DatabaseBibtexKeyPattern bibtexKeyPattern = ...` and `metadata.setBibtexKeyPattern(bibtexKeyPattern);
entry.setField("title",...)`
to the setup method.

src/main/java/net/sf/jabref/logic/bibtexkeypattern/BibtexKeyPatternUtil.java
```diff
@@ -519,6 +521,13 @@ public static String applyModifiers(String label, List<String> parts, int offset
                     }
                     resultingLabel = abbreviateSB.toString();
 
+                } else if ("camel".equals(modifier)) {
```
tobiasdiez: Would it make sense to support all Formatters as modifiers? If this is the case, I would add a method `Optional<Formatter> gefFormatterByTitle(String title)` to `Formatters` and use this here.
I think this also results in a more cleaner design.

src/main/java/net/sf/jabref/logic/bibtexkeypattern/BibtexKeyPatternUtil.java
tobiasdiez: If this code works, then it relies on a bad designed `Title` class. A collection exposed by a `get` method shouldn't give access to the internal state of the object. I think the best solution would be to add methods to the `Title` class which convert the title to a lower/upper/whatever case (i.e. `Title toLower()`). Since this probably goes beyond what you want to do in this PR, could you at least add a method `getWordsReadonly` (returning `Collections.unmodifiableList(words)`) and mark `getWords` deprecated. Thanks.

CHANGELOG.md
oscargus: Do you hear that @stefan-kolb ;-)

src/main/java/net/sf/jabref/logic/bibtexkeypattern/BibtexKeyPatternUtil.java
oscargus: I also reacted on that. There is quite a bit of more code than this relying on the current design though, so I add an issue and implement the `getWordsReadonly` thing.

src/test/java/net/sf/jabref/logic/bibtexkeypattern/MakeLabelWithDatabaseTest.java
oscargus: The preferences are different for different tests. There is a standard preferences in `setUp`, but in this case it is not the standard preferences. Adding set methods which are only used in the tests doesn't seem worthwhile to me.

Regarding entries, yes, one could have one in the preferences with some standard content, and edit as required in the tests.

(This also holds for your similar comment in the other PR, although I noted I used the standard preferences in some test and still redefined them...)

src/main/java/net/sf/jabref/logic/bibtexkeypattern/BibtexKeyPatternUtil.java
oscargus: Maybe. I still wonder how many of those are really useful and how many of them may lead to unexpected problems... We would in that case still need to have some explicit name mapping to not loose compatibility, introduce the (rather useless in other cases) CamelCaseFormatter, and come up with a new syntax. I did in fact consider it earlier, but didn't really see the benefit of it.

CHANGELOG.md
stefan-kolb: What?

oscargus: I think the missing space came from you. I'll change it though. ;-)

src/main/java/net/sf/jabref/logic/formatter/casechanger/Title.java
```diff
@@ -16,6 +17,11 @@ public Title(String title) {
         this.words.addAll(new TitleParser().parse(title));
     }
 
+    public List<Word> getWordsReadOnly() {
```
lenhard: Why not immediately replace `getWords` with `getWordsReadOnly`, i.e. just have `getWords` return `Collections.unmodifiableList(words);` and skip the introduction of `getWordsReadOnly`? From what I see this should not break the existing implementations.

oscargus: I think it will break it... See @tobiasdiez comment above. The thing is that `forEach(Word::toUpperFirst)` actually modifies the `Title` object, which it can't (I think) if `Collections.unmodifiableList(words);` is returned. On the other hand, I do not really see the use case of `Title` unless one is allowed to modify the content.

lenhard: Ah, I get it. The real problem is that the `Word` objects are mutable, less that the list is.

However, that means that you can *still* return `Collections.unmodifiableList(words)` and execute `forEach(Word::toUpperFirst)`. The `unmodifiableList` only ensures that the structure of the list will not change (and your lambda does not change it), the objects contained therein can still be modified (which is ugly, but anyway...). Therefore, I would ask you to implement my suggestion and skip the deprecated getter.

tobiasdiez: And please add at least a comment in `Word` that this class should be rewritten to be not mutable.

oscargus: Makes sense. I'll add immutable `getWords` as default and add a comment. On the other hand it is quite convenient that `Word` mutable. Somehow the whole `Title` class depends on it as the `toString()` method really wouldn't make sense if you cannot change any `Word` in `Title`. You would just get the same string that you entered and there is no way to convert a stream to a new `Title` at the moment.

src/main/java/net/sf/jabref/logic/bibtexkeypattern/BibtexKeyPatternUtil.java
oscargus: Turns out that the (useless) camel case formatter is actually already available as `CapitalizeFormatter`, then there is the `SentenceFormatter` left and that I can add for completeness (`ProtectedTermsFormatter` really doesn't make sense here...).

oscargus:
Added `lower_case`, `upper_case`, `sentence_case`, `capitalize`, and `title_case` modifiers (the two first are just additional names).

Unmodifiable return from `Title.getWords()` but the function of the modifier rely heavily on mutable `Word`.

CHANGELOG.md
```diff
@@ -46,9 +47,8 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - Entries in the SearchResultDialog are now converted to Unicode
 - Suggestions in the autocomplete (search) are now in Unicode
 - Fixed NullPointerException when opening search result window for an untitled database 
-- Fixed entry table traversal with Tab (no column traversal thus no double jump)
```
oscargus: Whoops...

src/main/java/net/sf/jabref/logic/bibtexkeypattern/BibtexKeyPatternUtil.java
```diff
@@ -493,10 +504,10 @@ public static String applyModifiers(String label, List<String> parts, int offset
             for (int j = offset; j < parts.size(); j++) {
                 String modifier = parts.get(j);
 
-                if ("lower".equals(modifier)) {
-                    resultingLabel = resultingLabel.toLowerCase(Locale.ENGLISH);
-                } else if ("upper".equals(modifier)) {
-                    resultingLabel = resultingLabel.toUpperCase(Locale.ENGLISH);
+                if ("lower".equals(modifier) || lowerCaseFormatter.getKey().equals(modifier)) {
```
tobiasdiez: Please refactor the code a bit: first convert the modifier to the corresponding formatter `Formatter getFormatterForModifier(string modifier)` and then apply it to the label.

src/test/java/net/sf/jabref/logic/bibtexkeypattern/MakeLabelWithDatabaseTest.java
```diff
@@ -321,4 +321,57 @@ public void generateKeyAuthIniMany() {
...
+        bibtexKeyPattern.setDefaultValue("[title:title_case]");
+        metadata.setBibtexKeyPattern(bibtexKeyPattern);
+        BibtexKeyPatternUtil.makeLabel(metadata, database, entry, preferences);
+        assertEquals(Optional.of("AnAwesomePaperonJabref"), entry.getCiteKeyOptional());
+    }
+
+    @Test
+    public void generateKeyTitleCapitalize() {
+        DatabaseBibtexKeyPattern bibtexKeyPattern = new DatabaseBibtexKeyPattern(pattern);
```
tobiasdiez: The lines `DatabaseBibtexKeyPattern bibtexKeyPattern = ...` and `metadata.setBibtexKeyPattern(bibtexKeyPattern);` can be extracted to the setup method.

koppor:
@oscargus Could you follow up so that we can merge?

tobiasdiez:
@koppor pls also move `metadata.setBibtexKeyPattern(bibtexKeyPattern);` to the setup method

koppor:
@tobiasdiez Sorry for overlooking your second move comment. I also created `Formatters.getFormatterForModifier`. Hope, this is good to go now.

tobiasdiez:
Yes, looks good now. Thanks @koppor for finishing this PR 😄

PR #1896, opened 2016-08-30
More tests and some cleanups of BibtexKeyPatternUtil

oscargus:
Removed some unused/unnecessary code and added a number of tests.

### DIFF
--- a/src/main/java/net/sf/jabref/logic/bibtexkeypattern/BibtexKeyPatternUtil.java
+++ b/src/main/java/net/sf/jabref/logic/bibtexkeypattern/BibtexKeyPatternUtil.java
@@ -662,15 +662,16 @@ public class BibtexKeyPatternUtil {
             } else if ("veryshorttitle".equals(val)) {
                 return getTitleWords(1, entry.getField(FieldName.TITLE).orElse(""));
             } else if ("shortyear".equals(val)) {
-                String ss = entry.getFieldOrAlias(FieldName.YEAR).orElse("");
-                if (ss.isEmpty()) {
-                    return ss;
-                } else if (ss.startsWith("in") || ss.startsWith("sub")) { // TODO: Year starting with in or sub? Why?
+                String yearString = entry.getFieldOrAlias(FieldName.YEAR).orElse("");
+                if (yearString.isEmpty()) {
+                    return yearString;
+                    // In press/in preparation/submitted
+                } else if (yearString.startsWith("in") || yearString.startsWith("sub")) {
                     return "IP";
-                } else if (ss.length() > 2) {
-                    return ss.substring(ss.length() - 2);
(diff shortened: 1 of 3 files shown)


src/main/java/net/sf/jabref/logic/bibtexkeypattern/BibtexKeyPatternUtil.java
```diff
@@ -692,7 +665,7 @@ else if (val.matches("edtr\\d+")) {
                 String ss = entry.getFieldOrAlias(FieldName.YEAR).orElse("");
                 if (ss.isEmpty()) {
                     return ss;
-                } else if (ss.startsWith("in") || ss.startsWith("sub")) {
+                } else if (ss.startsWith("in") || ss.startsWith("sub")) { // TODO: Year starting with in or sub? Why?
```
Siedlerchr: No idea why sub or in either. Any idea why the var is called "ss"? I would change it to a more expressive variable name

Siedlerchr:
If you change the one variable name to a more expresive one, I'm fine :+1:

src/test/java/net/sf/jabref/logic/bibtexkeypattern/MakeLabelWithDatabaseTest.java
```diff
@@ -285,4 +285,131 @@ public void generateKeyTitleAbbr() {
...
+        metadata.setBibtexKeyPattern(bibtexKeyPattern);
+        BibEntry entry = new BibEntry();
+        entry.setField("title", "An awesome paper on JabRef");
+        BibtexKeyPatternUtil.makeLabel(metadata, database, entry, preferences);
+        assertEquals(Optional.of("awesomepaperJabRef"), entry.getCiteKeyOptional());
+    }
+
+    @Test
+    public void generateKeyVeryshorttitle() {
+        preferences = new BibtexKeyPatternPreferences("", "", false, true, false, pattern);
```
tobiasdiez: Move som of this initialization code to the setup method.

tobiasdiez:
I second @Siedlerchr's opinion: LGTM. Just move some of the test code to `@Setup` and change the variable name, then this can be merged. (I thus remove the ready-for-review tag)

src/main/java/net/sf/jabref/logic/bibtexkeypattern/BibtexKeyPatternUtil.java
oscargus: Ahh, I think I understand it now: `IP` = in press/preparation, `in` = in press/preparation, and `sub` = submitted. Better to use the note field for that in general I guess, but maybe worthwhile keeping for compatibility. (One may even add support `under review` leading to either `IP` or `UR`.)

PR #1898, opened 2016-08-31
[WIP] Fix entry table traversal with Tab

bartsch-dev:
When traversing the entry table with Tab not only changed the selected row but the column did too.
When the most right column was selected and the user would press Tab one entry would be skipped,

Extracted some logic too.


boceckts:
Works like a charm and the code looks great as well 👍 
Please resolve conflicts and it's good from my side.

bartsch-dev:
Rebased.

Siedlerchr:
LGTM from my side, too :+1:  I would merge it in then.

PR #1861, opened 2016-08-25
[WIP] Cleanup FileUtil class and related things

Siedlerchr:
Rework/Cleanup FileUtil
- [X] Replace copy and rename
- [ ] expandFileName 
- [ ] shortenFileName
- [ ] uniquePathSubstrings

Issue #1092, opened 2016-04-03
Feature: move files to subfolders (Article,Book) of file directory


davydden:
Is there a workaround not only to rename a file (via `Cleanup entries` or `Move to file directory`) but also to move it to a subfolder (Article,Book,etc) where subfolder name is the same as entry type?

Siedlerchr:
There is a workaround:
You have to open the entry in the Entry Editor -> General -> File -> RightClick -> Move/Rename File-> Choose your directory...

But I will look if you can specify a global pattern.

davydden:
perhaps along the `filename format pattern` there could be a `directory format pattern`

davydden:
@Siedlerchr btw, the `Move/Rename File` is not really a workaround as one has to type the new file name which defeats the purpose of automatic renaming using `filename format pattern`.

tobiasdiez:
Maybe one can reuse the functionality of the old plug-in https://github.com/korv/Jabref-plugins/tree/master/renamefile/src/renamefilin which supported subfolders in the `filename format pattern` (at least according to their [about file](https://github.com/korv/Jabref-plugins/blob/master/renamefile/src/about.txt)).

davydden:
@tobiasdiez that's correct, i used that plugin quite a lot and subfolders worked there without an issue.

Siedlerchr:
I will look into that plugin from @korv and see how it could be integrated in JabRef.

davydden:
@Siedlerchr any updates to this issue?

Siedlerchr:
@davydden  Unfortunately not. Atm I have not that much time due do university. But I hope that I soon have more time in a few weeks.

oscargus:
I tried a workaround: `\entrytype\\\bibtexkey` hoping that that should lead to something good. However, I got `Article_Brisigotti2006.pdf` so something is replacing the `\\` (escaped `\`) with `_`. Might be one approach to enable it. (The same thing happens for `/` btw, I'm on Windows, so it seems that it makes sense. Maybe add a formatter that outputs the file system separator?

Siedlerchr:
I wanted to take care of this in the next days.  @oscargus :
My idea was to intdroduce a) either a second field similar to the existing one for filenames or b) to have one field for both.
I like option b), but I think it could have side effects, as the layout stuff is also used for displaying entry previews  and for html output


When we internally use the Paths methods from nio, the file system separator is implicit taken care of (btw, Windows supports `/` as file separator, too)

[...]

PR #1899, opened 2016-08-31
Move files to subfolders in file directory

Siedlerchr:
Implementation for #1092 
ATM hardcoded to entrytype
ATM only Drag and Drop respects fileDirPattern

Already implemented Copy and Move methods from #1861 

jabreffiledirpattern


Siedlerchr:
TODO: Pass global as param
TODO: Check relative path

Siedlerchr:
@JabRef/developers  I need some help with the GUI: I desperately tried to add the new field below the other, but I can't get it work, no matter what I tried :rage: 

The original layout code is this: 
`FormLayout layout = new FormLayout("1dlu, 8dlu, left:pref, 4dlu, fill:3dlu")` in:
https://github.com/JabRef/jabref/blob/fileDirPattern/src/main/java/net/sf/jabref/gui/preftabs/ImportSettingsTab.java

The new field is there, but kinda hidden. No idea how to change it back:

settingsimportpanel

Siedlerchr:
Yeah, finally the layout works :sunglasses:

dierkes:
Thanks for the work! I'm really missing this feature in JabRef 3.

However, the current behavior is a little unexpected. First, the file seems to be moved relative to its current directory. For example directory pattern '\EntryType' moves the file with every execution of cleanup to a new subdirectory. For me, expected behavior would be, that the new path is relative to the directory of the bib-file or some otherwise specified root directory. 

Second, the file field in the bib entry seems to be adjusted only, if the option 'Use the BIB file location as primary file directory' is set. Otherwise, the file is moved, but the file path is not adjusted.

Siedlerchr:
@dierkes  Thanks for your feedback. I would have asked you to test it otherwise, too ;)
The moving files is a bit complicated as it has to be kept backwards compatible.

The file directory is the first found directory: a) global file dire from setting b) bib file relative c) bib specific setting (atm not sure about the order b and c)

Regarding the file field, I will check it

dierkes:
Regarding the new file directory: That's not the behavior I observed.

My test case was:
original file: phdthesis/xyz_2015.pdf (all paths relative to the bib-file)
file name pattern: '\bibtexkey'
file dir pattern: '\EntryType'

cleanup -> phdthesis/PhdThesis/xyz_215.pdf
cleanup -> phdthesis/PhdThesis/PhdThesis/xyz_215.pdf
cleanup -> phdthesis/PhdThesis/PhdThesis/PhdThesis/xyz_215.pdf

It didn't matter if 'Use the BIB file location as primary file directory' was set or if the default directory for files was set. (I tried with setting it to the directory of the bib-file.) I didn't try to set the default file directory for the specific bib-file, however. (Can do that later.)

From my perspective, a backward compatible solution could be: If the file dir pattern is not set, rename the file in its current location. If it is set, move the file to the directory specified by file dir pattern (relative to bib-file directory, if it's a relative path) and rename it there. However, an empty file name pattern could mean a) rename the file in place or b) move file to bib-file directory, so an additional option is probably necessary, that specifies whether moving files is active or not.

Interpreting the file dir pattern as a path relative to the current directory of the file doesn't make sense to me, since executing cleanup would always change the file path. Which is probably the opposite effect one expects from a cleanup ;)

Siedlerchr:
Ah I see you are right.

Will look around.

2016-10-14 12:48 GMT+02:00 dierkes <notifications@github.com>:

Siedlerchr:
@dierkes  I fixed it, when all tests are green you could try it out.

dierkes:
Now it works mostly as expected for me.

Of course, as described above, with the current state it's not possible to rename a file only (if it's current location doesn't match the file dir pattern). So we might need the option to activate/deactivate moving the files.

And a very strange thing I noticed: If 'FileDir pattern' is empty and 'Main file directory' is empty and 'Use the BIB file location as primary file directory' is not set, then the file is not moved and a directory relative to the JabRef directory (or maybe to the current working directory)  is created with the name of the desired file path. Staying in the example from above, instead of renaming the file to
(literature/) PhdThesis/xyz_2015.pdf
a directory named
(jabref/) PhdThesis/xyz_2015.pdf/
is created.

Siedlerchr:
Thanks for the feedback, I will check it
Normally renaming file only should work if fileDirPattern is empty in the
settings


Am 15.10.2016 11:52 nachm. schrieb "dierkes" <notifications@github.com>:

Siedlerchr:
@dierkes  I fixed the strange behaviour. There was some directory path which was just "" and not null, so it was resolved to the current dir of the jar.  Now it only renames the file.
And the old behaviour now works, too, when fileDirPattern is Empty, the file is moved to the file directory, 
The file directories are checked in this order:
 The settings are prioritized in the following order and the first defined setting is used

1. metadata user-specific directory
2. metadata general directory
3. preferences directory
4. BIB file directory

dierkes:
This works, almost. ;-)

I encountered a similar problem when dragging a file and dropping it to an entry. When I select to move the file, again instead of a file a directory is created (same path as the file should have).

dierkes:
And another issue: If multiple files are linked to an entry, the first file is renamed and the other file links get lost. In JabRef 3.6 the first file is renamed and the other files are moved to the same directory.

Siedlerchr:
Thanks for your intenstive testing.
The issue with the drag and drop is now fixed.  There was a problem with the layout which would have been default to bibtexkey if the pattern param is empty.

I am aware of the thing with the multi files. As far as I discovered the first successful rename/move exits the function. Will try to find out why. Not sure if I find time for it next week.

Addtionally I will create some automatic tests, too if it works finally..

dierkes:
I think, I fixed the issue with the multiple files, see #2248.

The behavior now, as far as I understand it, is: The function will try to move/rename all attached files. If a file with the new file path already exist (that happens if two attached files have the same file extension), it doesn't move/rename the file and keeps the old file path in the bib entry.

For example, let's consider we have three attached files:
OldDir1/OldName1.pdf
OldDir2/OldName2.pdf
OldDir3/OldName3.txt

The new file paths will be:
NewDir/NewName.pdf
OldDir2/OldName2.pdf
NewDir/NewName.txt

I think, there is some room for improvement. For example, to keep OldName2 but move the file to NewDir, if possible. However, before doing this, I think it's necessary to define the desired behavior in detail.

Siedlerchr:
Thanks for your work! Just one thing, it would be nice if you could check that locally, too.
There is a junit test I set to ignore (due to the bug) which should now work, too. Just remove the ignore-Tag
and  execute it (either directly  or run gradlew check)

` cleanupRenamePdfRenamesWithMultipleFiles()` in  `src/test/java/net/sf/jabref/logic/cleanup/CleanupWorkerTest.java `

dierkes:
I removed the tag.

When I run `gradlew check`, the test `cleanupRenamePdfRenamesRelativeFile()` in `CleanupWorkerTest.java` fails. However, I think the problem lies within the design of the test. The expected path of the file is `Toot.tmp`, but the actual path is `Misc/Toot.tmp`. I guess, there is some default value set for fileDirPattern.

Siedlerchr:
Hm I think the value has to be set to an empty string in the test because
otherwise it will take the value of the settings.

Am 10.11.2016 12:57 vorm. schrieb "dierkes" <notifications@github.com>:
>
> I removed the tag.
>
> When I run gradlew check, the test cleanupRenamePdfRenamesRelativeFile()
in CleanupWorkerTest.java fails. However, I think the problem lies within
the design of the test. The expected path of the file is Toot.tmp, but the
actual path is Misc/Toot.tmp. I guess, there is some default value set for
fileDirPattern.

Siedlerchr:
@dierkes  I just fixed the failing test. As I expected it, it was due to the setting which got read

dierkes:
Great. Is there a chance to integrate this branch into the next release?

Siedlerchr:
Technically, this branch is based on the current master, but not included
with the 3.7 release which will be probably released today

So if you stick for the moment to the dev version of this branch, you have
all stuff from 3.7 release plus this feature.

Of course it will be in the next release then. There are only some tests
which now have to be added to make it ready for a merge

Am 13.11.2016 10:16 nachm. schrieb "dierkes" <notifications@github.com>:

src/main/java/net/sf/jabref/gui/externalfiles/DroppedFileHandler.java
```diff
@@ -358,13 +351,33 @@ private boolean showLinkMoveCopyRenameDialog(String linkFileName, ExternalFileTy
...
 
-        renameToTextBox.setText(targetName.concat(".").concat(fileType.getExtension()));
+        String fileDirPattern = Globals.prefs.get(JabRefPreferences.IMPORT_FILEDIRPATTERN);
 
+        String targetDirName = "";
+        if (!fileDirPattern.isEmpty()) {
+            targetDirName = FileUtil.createFileNameFromPattern(database, entry, fileDirPattern, layoutPrefs);
+        }
+
+        System.out.println("TARGET Filename " + targetName); //fileNam pattern, e.g bibtexkey
```
stefan-kolb: Remove Sysouts, please.

koppor:
Documentation: Could you please add an issue at help.jabref.org or even create an initial page with `skeleton: true` in the header?

You can link https://help.jabref.org/en/CustomExports#built-in-export-formatters for a description of the formatters which can be used.

src/main/java/net/sf/jabref/gui/externalfiles/DroppedFileHandler.java
```diff
@@ -471,45 +483,42 @@ private void doLink(BibEntry entry, ExternalFileType fileType, String filename,
...
-            if (answer == JOptionPane.NO_OPTION) {
+
+            Path fromFile = Paths.get(fileName);
+            try {
+                if (!Files.exists(destFile)) {
+                    Files.createDirectories(destFile);
+                }
+            } catch (IOException e) {
+                // TODO Auto-generated catch block
+                LOGGER.error("Problem creating target direcotires", e);
```
stefan-kolb: directories

src/main/java/net/sf/jabref/gui/externalfiles/DroppedFileHandler.java
```diff
@@ -471,45 +483,42 @@ private void doLink(BibEntry entry, ExternalFileType fileType, String filename,
...
+            }
+            if (FileUtil.renameFile(fromFile, destFile, true)) {
+                return true;
+            } else {
```
stefan-kolb: else can be omitted?!

src/main/java/net/sf/jabref/gui/externalfiles/DroppedFileHandler.java
```diff
@@ -538,23 +548,32 @@ private boolean doCopy(String fileName, String toFile, NamedCompound edits) {
             LOGGER.warn("Cannot determine destination directory or destination directory does not exist");
             return false;
         }
-        String destinationFileName = new File(toFile).getName();
 
-        File destFile = new File(dirs.get(found) + OS.FILE_SEPARATOR + destinationFileName);
-        if (destFile.equals(new File(fileName))) {
+        Path destFile = Paths.get(dirs.get(found)).resolve(toFile);
+        System.out.println("COpy Handler dest " + destFile);
```
stefan-kolb: Remove Sysout

src/main/java/net/sf/jabref/gui/externalfiles/DroppedFileHandler.java
```diff
@@ -358,13 +351,33 @@ private boolean showLinkMoveCopyRenameDialog(String linkFileName, ExternalFileTy
         moveRadioButton.setText(Localization.lang("Move file to file directory"));
         renameCheckBox.setText(Localization.lang("Rename file to").concat(": "));
 
+        //TODO: here to check for
```
koppor: What is the TODO here? Should it be removed?

src/main/java/net/sf/jabref/logic/cleanup/RenamePdfCleanup.java
```diff
@@ -54,59 +64,88 @@ public RenamePdfCleanup(boolean onlyRelativePaths, BibDatabaseContext databaseCo
...
-                    newFileEntryFileName = parent.toString().concat(OS.FILE_SEPARATOR).concat(newFilename.toString());
+                    unsuccessfulRenames++;
                 }
-                newFileList.add(new ParsedFileField(description, newFileEntryFileName, type));
-            } else {
-                unsuccessfulRenames++;
             }
         }
 
+        //TODO: Checken: Der macht das return wenn change = true, aber wenn noch ein nicht change dahinter kommt?`
```
stefan-kolb: There are some TODOs left?!

src/main/java/net/sf/jabref/logic/cleanup/RenamePdfCleanup.java
```diff
@@ -120,6 +159,7 @@ public RenamePdfCleanup(boolean onlyRelativePaths, BibDatabaseContext databaseCo
         }
 
         return Collections.emptyList();
+
```
stefan-kolb: Some empty lines could be removed in this PR.

src/main/java/net/sf/jabref/gui/externalfiles/DroppedFileHandler.java
Siedlerchr: Oops. Seems like I forgot that to delete. I will remove it

src/main/java/net/sf/jabref/logic/util/io/FileUtil.java
```diff
@@ -171,11 +170,29 @@ public static boolean copyFile(Path pathToSourceFile, Path pathToDestinationFile
      * @param toFile   The target fileName
      * @return True if the rename was successful, false if an exception occurred
      */
-    public static boolean renameFile(String fromFile, String toFile) {
+    public static boolean renameFile(Path fromFile, Path toFile) {
+
```
stefan-kolb: Empty line.

src/main/java/net/sf/jabref/logic/layout/format/RTFChars.java
```diff
@@ -206,46 +206,46 @@ private StringInt getPart(String text, int i, boolean commandNestedInBraces) {
      * @return returns the basic character of the given unicode
      */
     private String transformSpecialCharacter(long c) {
-        if((192 <= c && c <= 197) || c == 256 || c == 258 || c == 260) {
+        if(((192 <= c) && (c <= 197)) || (c == 256) || (c == 258) || (c == 260)) {
```
koppor: Could you please run the formatter on the whole method (after using `gradlew eclipse` from master branch). Then, the necessary space after `if` should be added, too.

src/main/java/net/sf/jabref/model/database/BibDatabaseContext.java
```diff
@@ -28,6 +31,7 @@
     private DBMSSynchronizer dbmsSynchronizer;
     private DatabaseLocation location;
 
+
```
stefan-kolb: Empty line.

src/main/java/net/sf/jabref/model/database/BibDatabaseContext.java
```diff
@@ -154,18 +170,18 @@ public boolean isBiblatexMode() {
     * @param fieldName The field type
     * @return The default directory for this field type.
     */
-    public List<String> getFileDirectory(String fieldName, FileDirectoryPreferences preferences) {
-        List<String> fileDirs = new ArrayList<>();
+    public List<String> getFileDirectories(String fieldName, FileDirectoryPreferences preferences) {
+        List<String> fileDirs = new ArrayList<>(4);
```
stefan-kolb: Why 4?

src/main/java/net/sf/jabref/model/database/BibDatabaseContext.java
```diff
@@ -182,8 +198,7 @@ public boolean isBiblatexMode() {
                 fileDirs.add(parentDir);
             }
         });
-
-        return fileDirs;
+        return  fileDirs;
```
stefan-kolb: Obsolete 2nd space?

src/test/java/net/sf/jabref/logic/cleanup/CleanupWorkerTest.java
```diff
@@ -323,4 +327,10 @@ public void cleanupWithDisabledFieldFormatterChangesNothing() {
         worker.cleanup(preset, entry);
         Assert.assertEquals(Optional.of("01"), entry.getField("month"));
     }
+
+    @After
+    public void tearDown() {
+        JabRefPreferences.getInstance().putBoolean(JabRefPreferences.BIB_LOC_AS_PRIMARY_DIR, false);
+
```
stefan-kolb: Would be cool if you can remove all thos obsolete empty lines 😄 I know its not important but 😸

src/main/java/net/sf/jabref/gui/externalfiles/DroppedFileHandler.java
Siedlerchr: Hm, that is weird. I already deleted all that stuff I thougt.

Siedlerchr:
I fixed the things from the review

src/test/java/net/sf/jabref/logic/cleanup/CleanupWorkerTest.java
```diff
@@ -54,10 +55,17 @@ public void setUp() throws IOException {
         MetaData metaData = new MetaData();
         metaData.setDefaultFileDirectory(pdfFolder.getAbsolutePath());
         BibDatabaseContext context = new BibDatabaseContext(new BibDatabase(), metaData, bibFolder.newFile("test.bib"));
+
+        JabRefPreferences prefs = JabRefPreferences.getInstance();
+        prefs.putBoolean(JabRefPreferences.BIB_LOC_AS_PRIMARY_DIR, true);
```
tobiasdiez: This really changes the preferences of the developer. Can you try to mock/init by hand the layout formatter and file directory preferences.

src/test/java/net/sf/jabref/logic/cleanup/RenamePdfCleanupTest.java
```diff
@@ -32,13 +35,16 @@
...
 
+
     @Before
     public void setUp() throws Exception {
         prefs = JabRefPreferences.getInstance();
         MetaData metaData = new MetaData();
         context = new BibDatabaseContext(new BibDatabase(), metaData, new Defaults());
         context.setDatabaseFile(testFolder.newFile("test.bib"));
 
+        prefs.putBoolean(JabRefPreferences.BIB_LOC_AS_PRIMARY_DIR, true);
```
tobiasdiez: as above

Siedlerchr:
included feedbacjk

stefan-kolb:
One or two of my comments are still open when I look at the Github UI above. Actually I'm only interested why you initialized the Array with a length of 4, I'm fine with the rest as thes are only cosmetic changes.

src/main/java/net/sf/jabref/model/database/BibDatabaseContext.java
Siedlerchr: Seems like it was an artifact from my previous implementation approach.. It is no longer necessary.

Siedlerchr:
Fixed the rest and the one test which @koppor commented out, in c9726d94686e6687af66476d26b0a1d08c61b41c works fine on travis now again.

stefan-kolb:
@JabRef/developers we can merge right?

lenhard:
@stefan-kolb @Siedlerchr Yes we can. That is, once the merge conflict in the italian localization gets fixed.

Siedlerchr:
@JabRef/developers  all Conflicts fixed.

dierkes:
Hi, on my system two tests are failing, cleanupRenamePdfRenamesWithMultipleFiles and cleanUpRenamePdfRenameFileDirectoryPatternSubDirectory. However, the problem seems to be that these and other tests do not remove the temporary files properly after execution. The behavior of the cleanup function itself is as expected, but since a file with the target path exists already (left from another test), the attached file is not renamed and the assertion fails. I'm running the tests on Fedora.

Siedlerchr:
Thanks  for  the feedback, I  will  check  that.

Am 06.12.2016 11:45 nachm. schrieb "dierkes" <notifications@github.com>:

Siedlerchr:
Should now be fixed. Was some Preferences problem as I already thought. Mocked it away.

PR #1900, opened 2016-08-31
Initial tests for WrapFileLinks

oscargus:

### DIFF
--- a/src/test/java/net/sf/jabref/logic/layout/format/WrapFileLinksTest.java
+++ b/src/test/java/net/sf/jabref/logic/layout/format/WrapFileLinksTest.java
@@ -1,0 +1,92 @@
+package net.sf.jabref.logic.layout.format;
+
+import java.util.Collections;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+
+
+public class WrapFileLinksTest {
+
+    private WrapFileLinks formatter;
+
+
+    @Before
+    public void setUp() {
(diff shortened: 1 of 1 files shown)


src/test/java/net/sf/jabref/logic/layout/format/WrapFileLinksTest.java
```diff
@@ -0,0 +1,92 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+
+
+public class WrapFileLinksTest {
+
+    private WrapFileLinks formatter;
```
stefan-kolb: What does this class do? The class name is not intuitive to me, maybe a candidate for a rename?!

oscargus: It is a quite flexible formatter for the file field. The things I have not tested relates to the actual search path.

See http://help.jabref.org/en/CustomExports

Regarding renaming, it is always tricky for layout formatters as people may have used them. On the other hand that is disconnected now, but still a point in consistency. Name is not really clear though.

Siedlerchr:
Good! 
It would be nice, if you could add tests for the examples used in the help file, to assure that they are correctly documented?
And maybe you could also check/take a look at this (problem) here:
http://discourse.jabref.org/t/the-wrapfilelinks-formatter/172

oscargus:
It was indeed triggered by that issue, but I didn't come that far.

oscargus:
I merge this in to be able to easily continue working on it (as, even though it is not complete, it adds a bit of coverage). Name can easily be changed later.

PR #1905, opened 2016-09-01
Fixed problem where closing brackets could not be used as texts in la…

oscargus:
The first closing brace in a layout argument was interpreted as a closing brace, so it could not be used as text, despite the help file doing so.

See: 
- http://discourse.jabref.org/t/the-wrapfilelinks-formatter/172
- http://help.jabref.org/en/CustomExports#the-wrapfilelinks-formatter


oscargus:
I do not know if this is the best solution as this relies on balanced brackets and I imagine there may be use cases where it is not necessarily balanced. I think the alternative is to escape all brackets that should be used as text.

src/test/java/net/sf/jabref/logic/layout/format/WrapFileLinksTest.java
```diff
@@ -89,4 +91,46 @@ public void testIteratorTwoItems() {
...
+        assertEquals("(Test file)", formatter.format("Test file:test.pdf:PDF"));
+    }
+
+    @Test
+    public void testPath() throws IOException {
+        FileLinkPreferences preferences = new FileLinkPreferences(Collections.emptyList(),
+                Collections.singletonList("src/test/resources/pdfs/"));
+        formatter = new WrapFileLinks(preferences);
+        formatter.setArgument("\\p");
+        assertEquals(new File("src/test/resources/pdfs/encrypted.pdf").getCanonicalPath(),
```
Siedlerchr: You should better use the nio equivalent Paths.get( ...).toString() (or if really needed the canonincal path op, than use toRealPath
https://docs.oracle.com/javase/tutorial/essential/io/legacy.html#mapping

lenhard:
I think you are right: An escaping character for use with braces in a formatter is probably the most clean solution.

However, since you have already implemented the PR in this fashion, we might just as well merge it and wait until someone comes up with the case of needing non-balanced braces in a formatter statement. @oscargus  I would say it is up to you: Do you want to try a rewrite of this or do you think this can be merged (after addressing the comment by @Siedlerchr , of course)?

oscargus:
Right now, I'd say merge as I do not really think I know when I will get around to it. On the other hand, that change may come before the `nio`-stuff. ;-)

For use case, I'd say:

1) filename.pdf
2) anotherfile.txt

src/main/java/net/sf/jabref/logic/layout/LayoutEntry.java
```diff
@@ -644,9 +646,14 @@ private LayoutFormatter getLayoutFormatterByName(String name) throws Exception {
                             int startParam = i;
                             i++;
                             boolean escaped = false;
-                            while (((i + 1) < c.length) && !(!escaped && (c[i] == '"') && (c[i + 1] == ')'))) {
+                            while (((i + 1) < c.length)
```
tobiasdiez: I vaguely remember that I have already seen some code which took care of balanced braces. Probably in the BibtexParser. Could you please check there and maybe extract some common code (to StringUtil?).

lenhard: Yes, there is brace checking and handling code in `BibtexParser`, but that is absolutely not extractable. It depends on the internal usage of the `PushBackReader` in the class and interferes with parsing in general.

Siedlerchr: When I remeber it correclty, there is also a method in StringUtil which does the same and checks for curly(?) braces...

tobiasdiez: Even if there is no suitable method which can be reused, it may be worthwhile to extract some of the code here to StringUtils.

src/test/java/net/sf/jabref/logic/layout/format/WrapFileLinksTest.java
```diff
@@ -89,4 +91,46 @@ public void testIteratorTwoItems() {
...
+                Collections.singletonList("src/test/resources/pdfs/"));
+        formatter = new WrapFileLinks(preferences);
+        formatter.setArgument("\\p");
+        assertEquals("test.pdf", formatter.format("Preferences:test.pdf:PDF"));
+    }
+
+    @Test
+    public void testRelativePath() {
+        formatter.setArgument("\\r");
+        assertEquals("test.pdf",
```
tobiasdiez: code fomatting: on same line?

tobiasdiez:
LGTM :+1:
Things to do before merge: consider moving/reusing some code to/from StringUtils and the nio-path stuff

stefan-kolb:
@oscargus Can you please implement the changes and rebase so we can merge this for 3.7?

src/test/java/net/sf/jabref/logic/layout/format/WrapFileLinksTest.java
stefan-kolb: Can you push the NIO stuff on this branch so we can merge this in even if Oscar is away at the moment?

Issue #1897, opened 2016-08-31
Integrity check should report year not consisting of 4 numerals


grimes2:
Field year should consist of four numerals (year = {1986}). Other formats are allowed (year = {about 1986}), but should be reported by Check integrity.
EDIT: Because these formats are allowed, it should be not reported by Check integrity.

Siedlerchr:
Yeah, but the official bibtex spec defines it a bit more detailed bibtex spec: http://ftp.fernuni-hagen.de/ftp-dir/pub/mirrors/www.ctan.org/biblio/bibtex/base/btxdoc.pdf

> Generally it should consist of four numerals, such as 1984, although the stan-
dard styles can handle any year whose last four nonpunctuation characters are
numerals, such as ‘(about 1984)’.

grimes2:
Interesting, but then this is not allowed: '(1984 but I am not sure)' or '(1984a)'

grimes2:
My interpretation is: The Year Checker should check for "last nonpunctuation character should not be an alphabetic character". I would like to implement this.

Siedlerchr:
You're welcome :) should be relatively simple to check if the last 4 chars of the field string are numeric.
Go for it :+1:

PR #1906, opened 2016-09-01
Check Integrity year check added

grimes2:
#1897 Added last four nonpunctuation characters should be numerals in Check Integrity year


oscargus:
Thanks! The tests fails because of missing translation strings.

If you run the logic/i10/LocalizationConsistencyTest.java you will get a list of strings which are not in the translation files. Add those to the English one and run the script that the output tells you to. Then, all languages are updated and the test should pass (you may have to refresh the files in your IDE in case it fails again). Also, see: https://github.com/JabRef/jabref/wiki/Code-Howtos#using-localization-correctly

If you can add a test to IntegrityCheckTest (or whatever it is called) that really tests for this new feature it would be great.

grimes2:
I hope, it is correct now. Please set label "ready-for-review". Thanks.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - [#1496](https://github.com/JabRef/jabref/issues/1496) Keep track of which entry a downloaded file belongs to
 - Made it possible to download multiple entries in one action
 - [#1813](https://github.com/JabRef/jabref/issues/1813) Import/Export preferences dialog default directory set to working directory
+- [#1897](https://github.com/JabRef/jabref/issues/1897) Added last four nonpunctuation characters should be numerals in Check Integrity year
```
Siedlerchr: I would reword it a bit to make it a a bit more clear: 
>  Implemented integrity check for year field: Last four nonpunctuation characters should be numerals 

(Put the word "year" in backticks (the french accent grave), just look at other examples in the changelog)

src/main/java/net/sf/jabref/logic/integrity/IntegrityCheck.java
```diff
@@ -306,9 +306,11 @@ private BracketChecker(String field) {
     private static class YearChecker implements Checker {
 
         private static final Predicate<String> CONTAINS_FOUR_DIGIT = Pattern.compile("([^0-9]|^)[0-9]{4}([^0-9]|$)").asPredicate();
+        private static final Predicate<String> ENDS_WITH_FOUR_DIGIT = Pattern.compile("[0-9][0-9][0-9][0-9]$")
```
Siedlerchr: You could use` [0-9]{4}$` but it's essentially the same.

Siedlerchr:
Thanks for your work :+1: LGTM! 
If you reword the changelog, we can merge it in.

src/main/java/net/sf/jabref/logic/integrity/IntegrityCheck.java
```diff
@@ -321,6 +323,12 @@ private BracketChecker(String field) {
                 return Collections.singletonList(new IntegrityMessage(Localization.lang("should contain a four digit number"), entry, FieldName.YEAR));
             }
```
stefan-kolb: We should add a comment to the BinteX doc here, because this decision on year structure is not intuitive.

src/main/java/net/sf/jabref/logic/integrity/IntegrityCheck.java
```diff
@@ -321,6 +323,12 @@ private BracketChecker(String field) {
...
 
+            if (!ENDS_WITH_FOUR_DIGIT.test(value.get().replaceAll("[](){},.;!?<>%&§]", ""))) {
```
stefan-kolb: Where comes the decision on these chars?

stefan-kolb: Please define them as punctuation chars as a final variable.

tobiasdiez:
Just a probably stupid question: why not just check that the year field consists exactly of 4 digits?

src/main/java/net/sf/jabref/logic/integrity/IntegrityCheck.java
grimes2: Some regex chars have a special function, they are ommited. `String.replaceAll(String regex, String replacement)` I failed to define a variable.

Siedlerchr:
@tobiasdiez  Because the offical bibtex spec says so:
See the issue for details:
https://github.com/JabRef/jabref/issues/1897#issuecomment-243756785

src/main/java/net/sf/jabref/logic/integrity/IntegrityCheck.java
Siedlerchr: And I tink you mixed one up  must be `$ instead of §`

src/main/java/net/sf/jabref/logic/integrity/IntegrityCheck.java
```diff
@@ -306,8 +306,9 @@ private BracketChecker(String field) {
     private static class YearChecker implements Checker {
 
         private static final Predicate<String> CONTAINS_FOUR_DIGIT = Pattern.compile("([^0-9]|^)[0-9]{4}([^0-9]|$)").asPredicate();
-        private static final Predicate<String> ENDS_WITH_FOUR_DIGIT = Pattern.compile("[0-9][0-9][0-9][0-9]$")
+        private static final Predicate<String> ENDS_WITH_FOUR_DIGIT = Pattern.compile("[0-9]{4}$")
                 .asPredicate();
+        private static final String regex = "[](){},.;!?<>%&$]";
```
stefan-kolb: Please uppercase constant name and call it something like `PUNCTUATION_MARKS`. Just calling it regex doesn't really help in understanding what it's for.

src/main/java/net/sf/jabref/logic/integrity/IntegrityCheck.java
```diff
@@ -306,9 +306,12 @@ private BracketChecker(String field) {
     private static class YearChecker implements Checker {
 
         private static final Predicate<String> CONTAINS_FOUR_DIGIT = Pattern.compile("([^0-9]|^)[0-9]{4}([^0-9]|$)").asPredicate();
+        private static final Predicate<String> ENDS_WITH_FOUR_DIGIT = Pattern.compile("[0-9]{4}$")
+                .asPredicate();
+        private static final String regex = "[](){},.;!?<>%&$]";
 
         /**
-         * Checks, if the number String contains a four digit year
+         * Checks, if the number String contains a four digit year and ends with it
```
stefan-kolb: Please add a comment pointing to the BiBTeX manual where this decision to accept around 1997 and so on is explained.

src/main/java/net/sf/jabref/logic/integrity/IntegrityCheck.java
```diff
@@ -321,6 +324,12 @@ private BracketChecker(String field) {
                 return Collections.singletonList(new IntegrityMessage(Localization.lang("should contain a four digit number"), entry, FieldName.YEAR));
             }
 
+            if (!ENDS_WITH_FOUR_DIGIT.test(value.get().replaceAll(regex, ""))) {
+                return Collections.singletonList(new IntegrityMessage(
+                        Localization.lang("last four nonpunctuation characters should be numerals"), entry,
```
stefan-kolb: Localization Message should start with Uppercase character?!

stefan-kolb:
Got 3 minor comments and then we can merge :+1:

src/test/java/net/sf/jabref/logic/integrity/IntegrityCheckTest.java
```diff
@@ -45,9 +45,13 @@ public void testUrlChecks() {
     public void testYearChecks() {
         assertCorrect(createContext("year", "2014"));
         assertCorrect(createContext("year", "1986"));
+        assertCorrect(createContext("year", "around 1986"));
+        assertCorrect(createContext("year", "(around 1986)"));
         assertWrong(createContext("year", "abc"));
         assertWrong(createContext("year", "86"));
         assertWrong(createContext("year", "204"));
+        assertWrong(createContext("year", "1986a"));
```
Siedlerchr: Please add also a test case with one of the regex symbols

src/main/java/net/sf/jabref/logic/integrity/IntegrityCheck.java
grimes2: Most of the Check Integrity messages start with lower-case. It is an inconsistency.

src/main/java/net/sf/jabref/logic/integrity/IntegrityCheck.java
```diff
@@ -321,6 +324,12 @@ private BracketChecker(String field) {
                 return Collections.singletonList(new IntegrityMessage(Localization.lang("should contain a four digit number"), entry, FieldName.YEAR));
             }
 
+            if (!ENDS_WITH_FOUR_DIGIT.test(value.get().replaceAll(regex, ""))) {
```
Siedlerchr: Something bothered me with the replaceall and the list of regex chars...
After thinking about it for a while I found why:
The first argument of replaceAll is a regex, so your regex char string would be interpreted as actual pattern.
To prevent this, you have to escape the chars, or better use http://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html#quote-java.lang.String-

grimes2: The pattern [abc] is needed: Set definition, can match the letter a or b or c.

src/main/java/net/sf/jabref/logic/integrity/IntegrityCheck.java
oscargus: Most messages do actually start with a lower case character right now so
for consistency it is currently correct. Although I think that conceptually
it should be upper case.

stefan-kolb:
Thanks :smile:  LGTM :+1:

Siedlerchr:
LGTM, too.

PR #1913, opened 2016-09-03
Mark some methods as deprecated in BibEntry and BibDatabase

tobiasdiez:
Changes are only in BibEntry and BibDatabase (the rest are propagated renames / method signature changes).

### DIFF
--- a/src/main/java/net/sf/jabref/model/database/BibDatabase.java
+++ b/src/main/java/net/sf/jabref/model/database/BibDatabase.java
@@ -18,6 +18,7 @@ import java.util.stream.Collectors;
 import net.sf.jabref.event.source.EntryEventSource;
 import net.sf.jabref.model.entry.BibEntry;
 import net.sf.jabref.model.entry.BibtexString;
+import net.sf.jabref.model.entry.FieldName;
 import net.sf.jabref.model.entry.InternalBibtexFields;
 import net.sf.jabref.model.entry.MonthUtil;
 import net.sf.jabref.model.event.EntryAddedEvent;
@@ -535,5 +536,9 @@ public class BibDatabase {
     @Subscribe
     private void relayEntryChangeEvent(FieldChangedEvent event) {
         eventBus.post(event);
+    }
+
+    public Optional<BibEntry> getReferencedEntry(BibEntry entry) {
+        return entry.getField(FieldName.CROSSREF).flatMap(this::getEntryByKey);
     }
 }
--- a/src/main/java/net/sf/jabref/model/database/BibDatabase.java
+++ b/src/main/java/net/sf/jabref/model/database/BibDatabase.java
@@ -16,12 +16,9 @@ import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import net.sf.jabref.event.source.EntryEventSource;
-import net.sf.jabref.model.EntryTypes;
 import net.sf.jabref.model.entry.BibEntry;
 import net.sf.jabref.model.entry.BibtexString;
-import net.sf.jabref.model.entry.EntryType;
-import net.sf.jabref.model.entry.EntryUtil;
-import net.sf.jabref.model.entry.FieldName;
+import net.sf.jabref.model.entry.InternalBibtexFields;
 import net.sf.jabref.model.entry.MonthUtil;
 import net.sf.jabref.model.event.EntryAddedEvent;
 import net.sf.jabref.model.event.EntryChangedEvent;
@@ -110,11 +107,7 @@ public class BibDatabase {
         for (BibEntry e : getEntries()) {
             allFields.addAll(e.getFieldNames());
         }
-        return allFields.stream().filter(field -> !isInternalField(field)).collect(Collectors.toSet());
-    }
-
-    public static boolean isInternalField(String field) {
-        return field.startsWith("__");
+        return allFields.stream().filter(field -> !InternalBibtexFields.isInternalField(field)).collect(Collectors.toSet());
     }
 
     /**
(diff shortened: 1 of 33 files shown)


src/main/java/net/sf/jabref/model/entry/BibEntry.java
```diff
@@ -104,6 +111,58 @@ public BibEntry(String id, String type) {
...
+
+        // Changed this to also consider alias fields, which is the expected
+        // behavior for the preview layout and for the check whatever all fields are present.
+        // TODO: But there might be unwanted side-effects?!
+        Optional<String> result = getFieldOrAlias(field);
+
+        // If this field is not set, and the entry has a crossref, try to look up the
+        // field in the referred entry: Do not do this for the bibtex key.
+        if (!result.isPresent() && (database != null)) {
+            Optional<String> crossrefKey = getField(FieldName.CROSSREF);
```
Siedlerchr: Better use a flatmap, it would only get evaluated if a value is present in all fields and addtionally you can add a filter condition for checking !isEmpty 
http://www.nurkiewicz.com/2013/08/optional-in-java-8-cheat-sheet.html (scroll down to end to see the "bigger example")

Siedlerchr:
What is the reason behind separating the insert and the duplication check?

tobiasdiez:
Most of the callers of `insertEntry` don't use the returned value of the duplication check. Moreover, I find it more readable if the caller first checks if a duplicate already exists (and handles this case appropriately) and then inserts the entry in the db.

src/main/java/net/sf/jabref/model/entry/BibEntry.java
tobiasdiez: Good point, changed the code accordingly.

Siedlerchr:
Thanks for the explaination. Now code LGTM :+1:

lenhard:
Ok, this sounds reasonable and the code that I looked at (in BibEntry and BibDatabase, as you suggested) looks good. I am merging it in.

oscargus:
As commented in the source, the deprecation of constants misses the target. By far.

lenhard:
My bad for merging! The changes in methods still have some values, don't they? So, instead of reverting, how about just removing the deprecation marking for the constants again?

@tobiasdiez What do you think?

oscargus:
Yeah, it was mainly the `BibEntry`-constants I reacted on, the rest I have
not looked carefully at, but I guess those are fine. The idea from
@tobiasdiez to use get/set/clear is of course 100% correct, but in almost
all cases the constants are not used for that purpose and it is not easily
solved. Unless we want to add methods like `isBibtexKeyField(String
fieldName)` and `isEntryTypeField(String fieldName)` which would solve
maybe half of the issues. Then there is sorting and to some extent tables
left.

oscargus:
Another comment: the fact that the duplication checker is not called when inserting the entry means that the keys do not add additional letters after it.

Yes, the callers did more than often not check the return value but now one will have to perform two operations for properly inserting an entry into a database. Both insert it into the database and into the duplicationchecker.

oscargus:
Not to mention the fact that the duplication checker will not find duplicates inserted the new way (as far as I can tell).

oscargus:
Good practice when marking things as deprecated: make some simple changes of the deprecated function to the non-deprecated function. Here, it would be obvious that the key-tests failed, despite no actual need to check for any duplicates before inserting them. Also, it would be obvious that most of the constants were not used in a bad way etc.

lenhard:
Good thing that you bring this back up! It seems it slipped under during this weeks devcall, but we should talk about it again during the next one (propably next week).

Siedlerchr:
I recenlty discoverd two problems:
The problem with this is deprecated is, that some of the constants are heavily used in a static manner in importers/exporters/fetchers to just check if the current processed field is the bibtexkeyfield.

Second problem: I recently tried to replace getText with getResolvedField in a method and suddenly all kind of test failed....(Unfortunately I don't remember the PR).
So these are not equal/have the same behaviour

src/main/java/net/sf/jabref/model/database/BibDatabase.java
```diff
@@ -157,8 +151,14 @@ public static boolean isInternalField(String field) {
      * @return false if the insert was done without a duplicate warning
      * @throws KeyCollisionException thrown if the entry id ({@link BibEntry#getId()}) is already  present in the database
      */
-    public synchronized boolean insertEntry(BibEntry entry) throws KeyCollisionException {
-        return insertEntry(entry, EntryEventSource.LOCAL);
+    @Deprecated // use insertEntry and DuplicationChecker separately
```
koppor: Please move the comment to `@deprecated` javadoc

src/main/java/net/sf/jabref/model/database/BibDatabase.java
```diff
@@ -222,10 +220,11 @@ public int getNumberOfKeyOccurrences(String key) {
      *
      * @return true, if the entry contains the key, false if not
      */
+    @Deprecated // use BibEntry.setCiteKey (and DuplicationChecker)
```
koppor: `@deprecated` JavaDoc?

src/main/java/net/sf/jabref/model/database/BibDatabase.java
```diff
@@ -490,65 +489,13 @@ private String resolveContent(String result, Set<String> usedIds) {
...
-    }
-
-    /**
      * Returns a text with references resolved according to an optionally given database.
      *
      * @param toResolve maybenull The text to resolve.
      * @param database  maybenull The database to use for resolving the text.
      * @return The resolved text or the original text if either the text or the database are null
      */
+    @Deprecated // use BibDatabase.resolveForStrings
```
koppor: `@deprecated` JavaDoc?

koppor:
I reverted it at #2023. Please comment, directly improve, or merge. :octocat:

Issue #1912, opened 2016-09-03
Integrity Check: check for field edition


grimes2:
Official bibtex spec:
The edition of a book - for example, “Second”. This should be an ordinal, and should have the first letter capitalized.
A check for "should have the first letter capitalized" in the field edition should be added. I would like to implement this.

Siedlerchr:
You could also check if the biblatex spec says something related forthe  fields:
https://www.ctan.org/pkg/biblatex?lang=de

Siedlerchr:
You have to treat bibtex and biblatex mode different. (there should be a flag or whatever to indicate this somewhere related to the bib database code)
Especially for the edition field, biblatex says  in section   2.2.2

> field (integer or literal)
The edition of a printed publication. This must be an integer, not an ordinal. Don’t say edition={First} or edition={1st} but edition={1}. The bibliography style converts this to a language dependent ordinal. It is also possible to give theedition as a literal string, for example “Third, revised and expanded edition”.

grimes2:
I think, this issue can be closed.

PR #1914, opened 2016-09-03
Check integrity edition check implemented

grimes2:
#1912 Implemented a check for the field edition. The check differentiates between BibTeX and BibLaTeX mode of the database.


grimes2:
Please set label "ready-for-review". Thank you.

src/main/java/net/sf/jabref/logic/integrity/IntegrityCheck.java
```diff
@@ -339,6 +340,44 @@ private BracketChecker(String field) {
...
+         * It is also possible to give the edition as a literal string, for example "Third, revised and expanded edition".
+         */
+        @Override
+        public List<IntegrityMessage> check(BibEntry entry) {
+            Optional<String> value = entry.getField(FieldName.EDITION);
+            if (!value.isPresent()) {
+                return Collections.emptyList();
+            }
+            //BibTeX
+            if (!FIRST_LETTER_CAPITALIZED.test(value.get().trim()) && (!bibDatabaseContext.isBiblatexMode())) {
```
Siedlerchr: I would have changed the order of the statements in the conditions, checking for bibllatex mode first:

Siedlerchr:
LGTM +1, Just let someone other look at it, too.

Another thing I just got in mind is that the whole checker code could be refactored to use lamdbas, because we have a nice Functional Interface. Will create a new issue for discussing that.

src/main/java/net/sf/jabref/logic/integrity/IntegrityCheck.java
```diff
@@ -28,7 +28,7 @@
 
 public class IntegrityCheck {
 
-    private final BibDatabaseContext bibDatabaseContext;
+    private static BibDatabaseContext bibDatabaseContext;
```
simonharrer: why do you want this to be static?

grimes2: Because I want to use a method of this class in class EditionChecker. An alternative is, to define a `private final BibDatabaseContext bibDatabaseContext;` in class EditionChecker, but this is redundant.

tobiasdiez:
LGTM :+1:, too. Just change the order in the if statement as @Siedlerchr suggest and then this could be merged.

src/main/java/net/sf/jabref/logic/integrity/IntegrityCheck.java
koppor: >. An alternative is, to define a private final BibDatabaseContext
bibDatabaseContext; in class EditionChecker,

Seems to be the better solution as static variables sound like global
variables...

grimes2: Many FieldChecker need access to the isBiblatexMode() method. It is bad to create for every FieldChecker class a new bibDatabaseContext object and inizialize it (in the FieldChecker constructor?).

Siedlerchr: What about passing the bibDatabaseContext in the constructor of the Field Checker?

oscargus:
And please extract the checker to its own file.

Siedlerchr:
LGTM from my side now.

lenhard:
@grimes2 Please rebase this on master and we will merge it in.

src/main/java/net/sf/jabref/logic/integrity/EditionChecker.java
```diff
@@ -0,0 +1,60 @@
...
+     */
+    @Override
+    public List<IntegrityMessage> check(BibEntry entry) {
+        Optional<String> value = entry.getField(FieldName.EDITION);
+        if (!value.isPresent()) {
+            return Collections.emptyList();
+        }
+
+        //BibLaTeX
+        if (!ONLY_NUMERALS_OR_LITERALS.test(value.get().trim()) && (bibDatabaseContextEdition.isBiblatexMode())) {
```
oscargus: Just a minor thing: if you change the order of the test and mode it will be slightly faster (since checking mode is faster and the test will not happen is the mode is wrong).

Not required for merge though.

src/main/java/net/sf/jabref/logic/integrity/EditionChecker.java
```diff
@@ -0,0 +1,60 @@
...
+            return Collections.singletonList(new IntegrityMessage(
+                    Localization.lang("should contain an integer or a literal"), entry, FieldName.EDITION));
+        }
+
+        //BibTeX
+        if (!FIRST_LETTER_CAPITALIZED.test(value.get().trim()) && (!bibDatabaseContextEdition.isBiblatexMode())) {
```
oscargus: And the same thing here then. One may even think of a `if mode ... else ...`

oscargus:
:+1:

tobiasdiez:
Thanks for your contribution!

PR #1915, opened 2016-09-03
improve find entry

bartsch-dev:
I imroved the `findEntry`-method some more. Now it always finds the correct Entry (by searching it with the id and not with the canonical representation.

### Steps to reproduce a situation in which the problem occurs
- have multiple duplicates in a database
- search them and open the results in the extra window
- select one of the duplicates in the extra window but the first
- in the main window always the first of the duplicates is selected

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - [#1897](https://github.com/JabRef/jabref/issues/1897) Implemented integrity check for `year` field: Last four nonpunctuation characters should be numerals
 
 ### Fixed
+- Fixed selecting an entry out of multiple duplicates
 - Fixed NullPointerException when opening search result window for an untitled database 
 - Fixed entry table traversal with Tab (no column traversal thus no double jump)
 - Fixed [#1757](https://github.com/JabRef/jabref/issues/1757): Crash after saving illegal argument in entry editor
--- a/src/main/java/net/sf/jabref/gui/maintable/MainTable.java
+++ b/src/main/java/net/sf/jabref/gui/maintable/MainTable.java
@@ -502,7 +502,7 @@ public class MainTable extends JTable {
         EventList<BibEntry> tableRows = model.getTableRows();
         for (int row = 0; row < tableRows.size(); row++) {
             BibEntry bibEntry = tableRows.get(row);
-            if (entry == bibEntry) {
+            if (entry == bibEntry) { // NOPMD (equals doesn't recognise duplicates)
                 return row;
             }
         }
--- a/src/main/java/net/sf/jabref/gui/maintable/MainTable.java
+++ b/src/main/java/net/sf/jabref/gui/maintable/MainTable.java
@@ -499,11 +499,14 @@ public class MainTable extends JTable {
     }
 
     public int findEntry(BibEntry entry) {
-        return model.getTableRows().indexOf(entry);
-    }
-
-    public int findLastEntry(BibEntry entry) {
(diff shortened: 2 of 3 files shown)


src/main/java/net/sf/jabref/gui/maintable/MainTable.java
```diff
@@ -499,11 +499,14 @@ public void setSelected(int row) {
     }
 
     public int findEntry(BibEntry entry) {
-        return model.getTableRows().indexOf(entry);
-    }
-
-    public int findLastEntry(BibEntry entry) {
-        return model.getTableRows().lastIndexOf(entry);
+        EventList<BibEntry> tableRows = model.getTableRows();
```
Siedlerchr: I am not sure, but I think this could be replaced a combination of stream,filter  and findFirst 
sth like : t`ableRows.stream().filter(bibEntry->bibEntry.getId().equals(entry.getId()).findFirst()

And it has the advantage to return an optional which is **way** better to return a -1 value. Returning 1 comes from the old times of programming where everything >0  was indicating succes and negative things like -1 indicated errors.  (Unfortunately I just can't find a source explaining it, but it is not considered good practice returning -1 nowadays if no element is found or error happend)

bartsch-dev: I need to find the position of the given entry in the model.
With the statement you described you would get an Optional of an Entry.

Siedlerchr: Ah yes, you are right. Currently there is no real good solution to return the index. :(

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - [#1897](https://github.com/JabRef/jabref/issues/1897) Implemented integrity check for `year` field: Last four nonpunctuation characters should be numerals
 
 ### Fixed
+- Fixed selecting an entry out of multiple duplicates
```
boceckts: Please add it at the end of the section.

src/main/java/net/sf/jabref/gui/maintable/MainTable.java
boceckts: I think in this particular case it is fine to use an "old school" for loop. But please rename `i` to `row`.

boceckts:
Very small comments by my side, otherwise it looks good to me 👍

CHANGELOG.md
bartsch-dev: I try not to do that b/c then I have to rebase after nearly each merged PR.

src/main/java/net/sf/jabref/gui/maintable/MainTable.java
bartsch-dev: done

src/main/java/net/sf/jabref/gui/maintable/MainTable.java
```diff
@@ -499,11 +499,14 @@ public void setSelected(int row) {
...
+        for (int row = 0; row < tableRows.size(); row++) {
+            BibEntry bibEntry = tableRows.get(row);
+            if (entry.getId().equals(bibEntry.getId())) {
```
tobiasdiez: I don't like that the ID is used to compare the entries (because I hope that we can get rid of the ID completely at some point). Does `if (entry == bibentry)` work?

bartsch-dev: done

koppor:
[LGTM](http://lgtm.in/i/KxPiQYVnq)

PR #1918, opened 2016-09-04
Improve LocalizionConsistencyTest to include JavaFX files

tobiasdiez:
The FXML files may contain localized strings in the form `<label text = "%localized description">` (and not `Localization.lang("localized description"`)). I adapted the LocalizationParser to also look for these kind of strings. Together with #1753, the JavaFX branch should now build without errors.

For some reason the ManageJournalAbbreviationsView has some problems with localized strings, although the same code works for the other dialogs. This needs further investigations (not part of this PR).


boceckts:
For me there are quite some errors in the localization test files, are they intended?

tobiasdiez:
What do you mean by errors? Some of the localization tests still fail, but I suspect all of them pass with the fixes of #1753.

boceckts:
No I mean, there is an error in LocalizationParserTest that it can't find the `LocalizationParser.JavaLocalizationEntryParser` due to its visibility. Setting it to public gives an error in the LocalizationBundle class.
> The type LocalizationBundle is already defined

tobiasdiez:
Thanks, fixed the visibility issue.

boceckts:
Now I get the following error
> The type LocalizationBundle is already defined

in the class `LocalizationBundle.java` in src/test/java/ in package net.sf.jabref.logic.l10n

PR #1923, opened 2016-09-05
ADS Fetcher using the new fetcher infrastructure 

zellerdev:
Recreate the ADS Fetcher from ADS using the new fetcher infrastructure #1594

- [x] Tests created for changes
- [x] Manually tested changed features in running JabRef

PR #1885, opened 2016-08-29
DOI Fetcher using the new fetcher infrastructure

zellerdev:
(Please do not squash at merge to keep author ships - @koppor)

Implement a new DOI Fetcher using the new fetcher infrastructure #1594 
In this case, the two classes DOItoBibTeX & DOItoBibTeXFetcher has been merged into the new Fetcher.

- [x] Tests created for changes
- [x] Manually tested changed features in running JabRef

PR #1925, opened 2016-09-05
Add ID Fetcher in Entrytypedialog

zellerdev:
Add an extra field in the EntryTypeDialog to fetch the ID Based fetcher. 
The field is in on the bottom of the dialog and is named `ID based generator`.

- [ ] DoiFetcher merged #1885 
- [ ] AdsFetcher merged #1923 

Normal
dialog30-f

Searching
dialog31-f

No internet connection
dialog34-f

Error message
fetcher35-f


Siedlerchr:
To make it a bit clearer for the user, I would rename it "From ID based fetcher".
So the semantic workflow could be  "create new entry from isbn 12346567"

oscargus:
Or just "From ID".

I think the idea is really good, but I'm thinking that there may be a better location, like a separate menu item. However, I assume that this is the dialog shown when pressing "new entry" in the toolbar (which I rarely use). No really good idea, but later it may be worthwhile trying to e.g. have a side panel with that functionality (or rather separating the fetcher types in search and ID, I realize).

src/main/resources/l10n/JabRef_sv.properties
```diff
@@ -1640,3 +1640,8 @@ last_four_nonpunctuation_characters_should_be_numerals=
 Remember_password?=Kom_ihåg_lösenord?
 
 shared=delad
+Remember_password?=
```
oscargus: Duplicate. :-)

tobiasdiez:
I'm also in favor of a new menu-item + dialog. This also provides you more room to include a bit more details. My proposal:
Image

Has the advantage that the user need not select the kind of ID he has (most of them are quite distinct and JabRef should be able to tell them apart) and he gets a preview of the entry (which tells him that we put the correct ID in)

koppor:
@tobiasdiez I like the idea, but do it *in addition*. I like the current dialog by @zellerdev. I vote for doing both!

I like the current "ID-based entry generator" (@zellerdev please rename acoordingly). Rename "Search" to "Generate". Is it possible to place "Cancel" to the right?

src/main/java/net/sf/jabref/gui/EntryTypeDialog.java
```diff
@@ -148,6 +174,95 @@ private JPanel createEntryGroupPanel(String groupTitle, Collection<EntryType> en
...
+                        JOptionPane.showMessageDialog(null, Localization.lang("Error_while_fetching_from_%0", fetcher.getName()), Localization.lang("Error_while_fetching"), JOptionPane.ERROR_MESSAGE);
+                    }
+                }
+                return bibEntry;
+            }
+
+            @Override
+            protected void done() {
+                if (isCancelled()) {
+                    JOptionPane.showMessageDialog(null, "Fetching was canceled", "Canceled fetching", JOptionPane.INFORMATION_MESSAGE);
```
koppor: Why is the mesasge not localized?

src/main/java/net/sf/jabref/gui/importer/fetcher/EntryFetchers.java
```diff
@@ -30,13 +31,19 @@ public EntryFetchers(JournalAbbreviationLoader abbreviationLoader) {
...
         entryFetchers.add(new SearchBasedEntryFetcher(new ArXiv()));
         entryFetchers.add(new SearchBasedEntryFetcher(new GvkFetcher()));
     }
 
     public List<EntryFetcher> getEntryFetchers() {
         return Collections.unmodifiableList(this.entryFetchers);
     }
+
+    public static List<IdBasedFetcher> getIdFetchers() {
+        List<IdBasedFetcher> list = new LinkedList<>();
```
koppor: Why not `ArrayList`? It is [faster at sorting](http://stackoverflow.com/a/26013151/873282).

src/main/resources/l10n/JabRef_da.properties
```diff
@@ -1781,3 +1780,8 @@ should_have_the_first_letter_capitalized=
 Remember_password?=
 
 shared=
+Finished_downloading_full_text_document_for_entry_%0.=
+
+No_entry_with_id_'%0'_for_fetcher_'%1'_was_found.=
+
+Error_while_fetching=Fejl_under_hentning
```
koppor: Where do you know Dansk from? :innocent: - Are we really sure that there is no equivalent translation somewhere? What do the web-based fetchers output if something goes wrong?

src/main/java/net/sf/jabref/gui/EntryTypeDialog.java
```diff
@@ -148,6 +174,95 @@ private JPanel createEntryGroupPanel(String groupTitle, Collection<EntryType> en
         return panel;
     }
 
+    private JPanel createIdFetcher(String groupTitle) {
```
koppor: Rename to "createIdFetcherPanel" and remove parameter. Just pass it directly to the method below and use Localization.lang to have the title translated into other languages.

koppor:
Please add updated screenshot, also the German screenshot to be sure that everything is translated. Ater that, it should be good to go.

src/main/java/net/sf/jabref/gui/EntryTypeDialog.java
```diff
@@ -9,33 +9,52 @@
...
 import net.sf.jabref.model.entry.BibtexEntryTypes;
 import net.sf.jabref.model.entry.EntryType;
 import net.sf.jabref.model.entry.IEEETranEntryTypes;
 
 import com.jgoodies.forms.builder.ButtonBarBuilder;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 import org.jdesktop.swingx.VerticalLayout;
 
+import static net.sf.jabref.gui.importer.fetcher.EntryFetchers.getIdFetchers;
```
tobiasdiez: I would prefer if you don't import this as static.

src/main/java/net/sf/jabref/gui/EntryTypeDialog.java
```diff
@@ -91,12 +112,17 @@ private JPanel createEntryGroupsPanel() {
...
 
-        if(biblatexMode) {
+        if (biblatexMode) {
             panel.add(createEntryGroupPanel("BibLateX", EntryTypes.getAllValues(bibDatabaseContext.getMode())));
         } else {
             panel.add(createEntryGroupPanel("BibTeX", BibtexEntryTypes.ALL));
             panel.add(createEntryGroupPanel("IEEETran", IEEETranEntryTypes.ALL));
-            panel.add(createEntryGroupPanel(Localization.lang("Custom"), CustomEntryTypesManager.ALL));
+
+            if (!CustomEntryTypesManager.ALL.isEmpty()) {
```
tobiasdiez: The custom entry types and the id based entry are not displayed in BibLatexMode. Is this by design? @JabRef/developers

src/main/java/net/sf/jabref/gui/EntryTypeDialog.java
```diff
@@ -91,12 +112,17 @@ private JPanel createEntryGroupsPanel() {
...
+                panel.add(createEntryGroupPanel(Localization.lang("Custom"), CustomEntryTypesManager.ALL));
+            }
+
+            panel.add(createIdFetcher("ID-based entry generator"));
```
tobiasdiez: -> not displayed in BibLatexMode?

src/main/java/net/sf/jabref/gui/EntryTypeDialog.java
```diff
@@ -148,6 +174,95 @@ private JPanel createEntryGroupPanel(String groupTitle, Collection<EntryType> en
         return panel;
     }
 
+    private JPanel createIdFetcher(String groupTitle) {
+        JButton generateButton = new JButton(Localization.lang("Generate"));
+        JButton cancelButton = new JButton(Localization.lang("Cancel"));
+        JTextField idTextField = new JTextField("");
+        JComboBox<String> comboBox = new JComboBox<>();
+        getIdFetchers().forEach(n -> comboBox.addItem(n.getName()));
```
tobiasdiez: rename n -> fetcher

src/main/java/net/sf/jabref/gui/EntryTypeDialog.java
```diff
@@ -148,6 +174,95 @@ private JPanel createEntryGroupPanel(String groupTitle, Collection<EntryType> en
...
+            protected Optional<BibEntry> doInBackground() throws Exception {
+                searchID = idTextField.getText();
+                if (!searchID.isEmpty()) {
+                    fetcher = EntryFetchers.getIdFetchers().get(comboBox.getSelectedIndex());
+                    try {
+                        bibEntry = fetcher.performSearchById(searchID);
+                        dispose();
+                    } catch (FetcherException e) {
+                        LOGGER.error("Error fetching from " + fetcher.getName(), e);
+                        JOptionPane.showMessageDialog(null, Localization.lang("Error_while_fetching_from_%0", fetcher.getName()), Localization.lang("Error_while_fetching"), JOptionPane.ERROR_MESSAGE);
```
tobiasdiez: without underscore

src/main/java/net/sf/jabref/gui/EntryTypeDialog.java
```diff
@@ -148,6 +174,95 @@ private JPanel createEntryGroupPanel(String groupTitle, Collection<EntryType> en
...
+                    JOptionPane.showMessageDialog(null, "Fetching was canceled", "Canceled fetching", JOptionPane.INFORMATION_MESSAGE);
+                    return;
+                }
+
+                try {
+                    Optional<BibEntry> result = get();
+                    if (result.isPresent()) {
+                        frame.getCurrentBasePanel().insertEntry(result.get());
+                    } else {
+                        JOptionPane.showMessageDialog(null, Localization.lang("No_entry_with_id_'%0'_for_fetcher_'%1'_was_found.", searchID, fetcher.getName()), Localization.lang("No_files_found."), JOptionPane.WARNING_MESSAGE);
```
tobiasdiez: no underscore

src/main/java/net/sf/jabref/gui/EntryTypeDialog.java
```diff
@@ -148,6 +174,95 @@ private JPanel createEntryGroupPanel(String groupTitle, Collection<EntryType> en
...
+                    }
+                } catch (ExecutionException | InterruptedException e) {
+                    e.printStackTrace();
```
tobiasdiez: Proper logging

src/main/java/net/sf/jabref/gui/EntryTypeDialog.java
```diff
@@ -148,6 +174,95 @@ private JPanel createEntryGroupPanel(String groupTitle, Collection<EntryType> en
...
+                }
+            }
+        };
+
+        generateButton.addActionListener(n -> fetcherWorker.execute());
```
tobiasdiez: rename n

src/main/java/net/sf/jabref/gui/EntryTypeDialog.java
tobiasdiez: fetcher could be null I think (select a fetcher in the dialog but let the id empty -> click search)

tobiasdiez:
Just some small remarks, then this could be merged.

src/main/java/net/sf/jabref/gui/EntryTypeDialog.java
Siedlerchr: No, they are already displayed (even in biblatex mode)  but they are mixed with the others:

jabrefmycustomentrytype

src/main/resources/l10n/JabRef_en.properties
```diff
@@ -2177,7 +2177,7 @@ Copied_version_to_clipboard=Copied_version_to_clipboard
 
 BibTeX_key=BibTeX_key
 Message=Message
-
+ID-based_entry_generator=ID-based_entry_generator
```
koppor: The order of entries in `JabRef_en.properties` is important. Please move this together with `No_entry_with_id_'%0'_for_fetche...`. Maybe at the end of the file (as it is common for new entries)

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - [#1897](https://github.com/JabRef/jabref/issues/1897) Implemented integrity check for `year` field: Last four nonpunctuation characters should be numerals
 - Address in MS-Office 2007 xml format is now imported as `location`
 - [#1912](https://github.com/JabRef/jabref/issues/1912) Implemented integrity check for `edition` field: Should have the first letter capitalized (BibTeX), Should contain an integer or a literal (BibLaTeX)
+- Add ID Fetcher in EntryTypeDialog
```
koppor: We decided for "ID-based_entry_generator". Thus, this term should be used here. Further, I don't believe it's called "EntryTypeDialog` on the user's side (see http://help.jabref.org/en/BaseFrameHelp). Maybe phrase it "Support entry generation based on IDs in the dialog for choosing the type for a new entry" or "The dialog for choosing new entries additionally supports ID-based entry generation"

src/main/java/net/sf/jabref/gui/EntryTypeDialog.java
```diff
@@ -148,6 +172,95 @@ private JPanel createEntryGroupPanel(String groupTitle, Collection<EntryType> en
...
+                    }
+                }
+                dispose();
+                return bibEntry;
+            }
+
+            @Override
+            protected void done() {
+                if (isCancelled()) {
+                    JOptionPane.showMessageDialog(null, Localization.lang("%0 import canceled", fetcher.getName()),Localization.lang("Import canceled by user"), JOptionPane.INFORMATION_MESSAGE);
```
koppor: According to http://stackoverflow.com/a/6204600/873282 this is not thread safe as it is not called in the EDT. Maybe, it is also not called at all, according to the last comment by [searchengine27](http://stackoverflow.com/users/3560873/searchengine27) at http://stackoverflow.com/a/6204600/873282. Did you test whether this dialog appears?

src/main/java/net/sf/jabref/gui/EntryTypeDialog.java
```diff
@@ -148,6 +172,95 @@ private JPanel createEntryGroupPanel(String groupTitle, Collection<EntryType> en
...
+
+                try {
+                    Optional<BibEntry> result = get();
+                    if (result.isPresent()) {
+                        frame.getCurrentBasePanel().insertEntry(result.get());
+                    } else {
+                        JOptionPane.showMessageDialog(null, Localization.lang("No entry with id '%0' for fetcher '%1' was found.", searchID, fetcher.getName()), Localization.lang("No files found."), JOptionPane.WARNING_MESSAGE);
+                    }
+                } catch (ExecutionException | InterruptedException e) {
+                    LOGGER.error("No entry with id '" + searchID + " for fetcher '" + fetcher.getName() + "' was found.");
```
koppor: We changed that when we set together. Why did you change that back? Is it because of some rebase stuff? In general, to not rebase, but do merge.

src/main/java/net/sf/jabref/gui/EntryTypeDialog.java
```diff
@@ -148,6 +172,95 @@ private JPanel createEntryGroupPanel(String groupTitle, Collection<EntryType> en
...
+        constraints.gridy = 2;
+        JPanel buttons = new JPanel();
+        ButtonBarBuilder bb = new ButtonBarBuilder(buttons);
+        bb.addButton(generateButton);
+        bb.addButton(cancelButton);
+
+        jPanel.add(buttons, constraints);
+        jPanel.setBorder(BorderFactory.createTitledBorder(BorderFactory.createEtchedBorder(), Localization.lang("ID-based_entry_generator")));
+
+
```
koppor: Please remove line, so that **one** empty line is remaining.

src/main/resources/l10n/JabRef_de.properties
```diff
@@ -2490,6 +2490,12 @@ last_four_nonpunctuation_characters_should_be_numerals=Die letzten vier Nichtint
 Remember_password?=Passwort_merken?
 
 shared=geteilt
+No_entry_with_id_'%0'_for_fetcher_'%1'_was_found.=Kein_Eintrag_mit_der_ID_'%0'_für_fetcher_'%1'_wurde_gefunden.
```
koppor: We also updated this when we met. Why is our new String lost?

koppor:
Can you rewrite the code to depend on #1885 only. As #1923 depends on #1929, it will take more than this week to get it merged. I don't want to wait that long.

PR #1927, opened 2016-09-06
Some OO/LO cleanups

oscargus:
Removed deprecated calls, renamed a few variables, and restructured the code a bit.

### DIFF
--- a/src/main/java/net/sf/jabref/gui/openoffice/OOBibBase.java
+++ b/src/main/java/net/sf/jabref/gui/openoffice/OOBibBase.java
@@ -248,11 +248,11 @@ class OOBibBase {
         } catch (Exception e) {
             throw new CreationException(e.getMessage());
         }
-        XDesktop xDesktop = UnoRuntime.queryInterface(XDesktop.class, desktop);
+        XDesktop resultDesktop = UnoRuntime.queryInterface(XDesktop.class, desktop);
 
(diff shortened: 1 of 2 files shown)


src/main/java/net/sf/jabref/gui/openoffice/OOBibBase.java
```diff
@@ -1366,13 +1366,13 @@ public BibDatabase generateDatabase(List<BibDatabase> databases)
                     BibEntry clonedEntry = (BibEntry) entry.get().clone();
                     clonedEntry.setId(IdGenerator.next());
                     // Insert a copy of the entry
-                    resultDatabase.insertEntryWithDuplicationCheck(clonedEntry);
+                    resultDatabase.insertEntry(clonedEntry);
```
lenhard: Just wondering: Is it save to remove the duplication check here? The documentation of `insertEntryWithDuplicationCheck` says `use insertEntry and DuplicationChecker separately`.

The comment also applies to the replacement below.

lenhard:
Apart from the clarification above: LGTM!

oscargus:
Well the duplication check just returns a boolean and as that is not checked here it shouldn't make any functional difference. If it was a good idea in the first place is another question though...

tobiasdiez:
LGTM :+1:

PR #1928, opened 2016-09-06
Support CitationStyles

bartsch-dev:
Implements #119.

The User can choose multiple CitationStyles which he can cycle threw in the preview panel (forward with <kbd>F9</kbd>, backwards <kbd>shift</kbd> + <kbd>F9</kbd>, both are in the context menu and the menu bar) .
As discussed I removed the 2nd Preview.

### Preview before:
preview_before
### Preview after: 
preview_after

### Preferences before:
preferences_before
 
### Preferences after: 
preferences_after

I didn't include the localization (other the English) on purpose because they create unnecessary conflicts when rebasing quite often (I push them when everything else is fine).

The Help page has to be updated too. I do that as soon as this PR is merged.

Warning: currently I read the available CitationStyles at runtime. I had problems reading them from the gradle dependency thus currently they are only found when JabRef is started from the JAR.


boceckts:
The text of the preferences tab is not completely visible, either reword or adjust the width of the side panel.

src/main/java/net/sf/jabref/gui/JabRefFrame.java
```diff
@@ -638,8 +642,12 @@ public void windowClosing(WindowEvent e) {
                 return;
             }
 
+            if (bp.getPreviewPanel() != null) {
+                bp.getPreviewPanel().updateLayout();
```
boceckts: Please rename

src/main/java/net/sf/jabref/gui/preftabs/PreviewPrefsTab.java
```diff
@@ -1,160 +1,119 @@
...
-        defaultButton2.addActionListener(e -> {
-            String tmp = layout2.getText().replace("\n", "__NEWLINE__");
-            PreviewPrefsTab.this.prefs.remove(JabRefPreferences.PREVIEW_1);
-            layout2.setText(PreviewPrefsTab.this.prefs.get(JabRefPreferences.PREVIEW_1).replace("__NEWLINE__", "\n"));
-            PreviewPrefsTab.this.prefs.put(JabRefPreferences.PREVIEW_1, tmp);
+        available.getSelectionModel()
+                .addListSelectionListener(e -> btnRight.setEnabled(!((ListSelectionModel) e.getSource()).isSelectionEmpty()));
+
+        btnRight.addActionListener(e -> {
+            for (Object s : available.getSelectedValuesList()) {
```
boceckts: Please rename

src/main/java/net/sf/jabref/gui/preftabs/PreviewPrefsTab.java
```diff
@@ -1,160 +1,119 @@
...
-                testPanel.setPreferredSize(new Dimension(800, 350));
-                JOptionPane.showMessageDialog(null, testPanel, Localization.lang("Preview"), JOptionPane.PLAIN_MESSAGE);
-            } catch (StringIndexOutOfBoundsException ex) {
-                LOGGER.warn("Parsing error.", ex);
-                JOptionPane.showMessageDialog(null,
-                        Localization.lang("Parsing error") + ": " + Localization.lang("illegal backslash expression")
-                                + ".\n" + ex.getMessage(),
-                        Localization.lang("Parsing error"), JOptionPane.ERROR_MESSAGE);
+        btnLeft.addActionListener(e -> {
+            for (Object s : chosen.getSelectedValuesList()) {
```
boceckts: Please rename, this applies to all the variables that are only one or two characters long.

boceckts:
When pressing the up or down button the selected entry becomes deselected and I have to select it again, this can be quite frustrating if I want to move a citation style to the top. Can you please change it so the selected entry stays selected?

src/main/java/net/sf/jabref/gui/preftabs/PreviewPrefsTab.java
```diff
@@ -165,27 +124,133 @@ public PreviewPrefsTab(JabRefPreferences prefs) {
...
-        layout2.setText(prefs.get(JabRefPreferences.PREVIEW_1).replace("__NEWLINE__", "\n"));
+        PreviewPreferences previewPreferences = new PreviewPreferences(Globals.prefs);
+
+        chosenModel.clear();
+        boolean isPreviewChosen = false;
+        for (String style : previewPreferences.getPreviewCycle()) {
+            if (CitationStyle.isCitationStyleFile(style)){
+                chosenModel.addElement(CitationStyle.createCitationStyleFromFile(style));
+            }
+            else {
```
boceckts: Weird formatting, move this to the end of the line above.

boceckts:
The preview panel is not being updated after changing the citation style but rather displays always the first selected entry.

Steps to reproduce:
1. go to preferences and add a citation style
2. In the main table select an entry and open the preview panel
3. press F9 to see the newly added citation style
4. select a different entry and the preview panel will not update but use the information of the first selected entry

boceckts:
The preview panel is not being updated after changing the citation style when working with multiple databases.

Steps to reproduce:
1. go to preferences and add a citation style
2. open at least two databases with at least one entry
3. in the main table select an entry and open the preview panel
4. select the second database and an entry
6. press F9 to change to the newly added citation style
5. the preview panel will not update at all but the status says the citation style has changed
When switching to the first database the preview panel has updated and responds to citation style changes.

boceckts:
I think it would be great to also add the selected citation styles to the "Export to clipboard" dialog.

src/main/java/net/sf/jabref/gui/worker/CitationStyleWorker.java
```diff
@@ -0,0 +1,67 @@
...
+import net.sf.jabref.gui.PreviewPanel;
+import net.sf.jabref.logic.l10n.Localization;
+import net.sf.jabref.logic.search.MatchesHighlighter;
+import net.sf.jabref.model.entry.BibEntry;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+
+public class CitationStyleWorker extends SwingWorker<String, Void> {
```
boceckts: Please add a javadoc comment.

src/main/java/net/sf/jabref/logic/citationstyle/CitationStyle.java
```diff
@@ -0,0 +1,148 @@
...
+import org.apache.commons.logging.LogFactory;
+import org.w3c.dom.CharacterData;
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
+import org.w3c.dom.NodeList;
+import org.xml.sax.InputSource;
+import org.xml.sax.SAXException;
+
+
+public class CitationStyle {
```
boceckts: Please add some java doc coments to the class and its methods.

src/main/java/net/sf/jabref/logic/citationstyle/CitationStyleCache.java
```diff
@@ -0,0 +1,80 @@
...
+
+import net.sf.jabref.BibDatabaseContext;
+import net.sf.jabref.model.entry.BibEntry;
+import net.sf.jabref.model.event.EntryChangedEvent;
+import net.sf.jabref.model.event.EntryRemovedEvent;
+
+import com.google.common.eventbus.Subscribe;
+
+
+public class CitationStyleCache {
```
boceckts: Also java doc comment.

src/main/java/net/sf/jabref/logic/citationstyle/CitationStyleGenerator.java
```diff
@@ -0,0 +1,63 @@
...
+import de.undercouch.citeproc.bibtex.BibTeXItemDataProvider;
+import de.undercouch.citeproc.output.Bibliography;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.jbibtex.BibTeXDatabase;
+import org.jbibtex.ParseException;
+import org.jbibtex.TokenMgrError;
+
+
+public class CitationStyleGenerator {
```
boceckts: Please add a java doc comment at least for the class.

src/main/resources/l10n/Menu_de.properties
```diff
@@ -216,4 +215,6 @@ View_change_log=Changelog_öffnen
 Website=Webseite
 Blog=Blog
 Development_version=Entwicklungsversion
-JabRef_resources=Mehr_zu_JabRef
\ No newline at end of file
+JabRef_resources=Mehr_zu_JabRef
```
boceckts: This shouldn't be changed.

boceckts:
When the bugs are fixed and the other minor comments have been addressed then it's fine by me.

src/main/resources/l10n/Menu_de.properties
bartsch-dev: Previously the new line command was missing in this line.

src/main/java/net/sf/jabref/gui/JabRefFrame.java
bartsch-dev: fixed

src/main/java/net/sf/jabref/gui/preftabs/PreviewPrefsTab.java
bartsch-dev: fixed

src/main/java/net/sf/jabref/gui/preftabs/PreviewPrefsTab.java
bartsch-dev: fixed

src/main/java/net/sf/jabref/gui/preftabs/PreviewPrefsTab.java
bartsch-dev: fixed

src/main/java/net/sf/jabref/gui/worker/CitationStyleWorker.java
bartsch-dev: fixed

src/main/java/net/sf/jabref/logic/citationstyle/CitationStyle.java
bartsch-dev: fixed

src/main/java/net/sf/jabref/logic/citationstyle/CitationStyleCache.java
bartsch-dev: fixed

src/main/java/net/sf/jabref/logic/citationstyle/CitationStyleGenerator.java
bartsch-dev: fixed

build.gradle
```diff
@@ -118,6 +121,12 @@ dependencies {
     compile 'org.apache.logging.log4j:log4j-api:2.6.2'
     compile 'org.apache.logging.log4j:log4j-core:2.6.2'
 
+    // need to use snapshots as the stable version is from 2013 and doesn't support v1.0.1 CitationStyles
+    compile 'org.citationstyles:styles:1.0.1-SNAPSHOT'
```
koppor: Please also follow https://github.com/JabRef/jabref/blob/master/CONTRIBUTING.md#when-adding-a-library

Is there an issue asking for a release of the library? - If yes, please link it here and open an issue at jabref/issues linking that issue. If no, please either contact the maintainers directly or open an issue at their repo. Also open an issue at jabref/issues.

bartsch-dev: done

bartsch-dev:
Fixed the Bug (both were caused by the same thing)
Is the Export to clipboard important? You can already copy the current preview via context menu on the preview panel.

koppor:
The context menu exports text only. Is it possible to export the HTML? The "export to clipboard" is not important. If some interested user requests it, work on it can be done.

CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 ## [Unreleased]
 
 ### Changed
+- Added support for over 1000 (self updating) 1.0.1 Citation Styles
+- You can set and cycle different preview Styles (including CitationStyles)
```
koppor: "styles" with lower case s

CHANGELOG.md
```diff
@@ -53,6 +55,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - Fixed InvalidBackgroundColor flickering with Ctrl-s and File > Save database
 
 ### Removed
+- Removed 2nd Preview Style
```
koppor: "preview style" (lower case letters)

src/main/java/net/sf/jabref/gui/BasePanel.java
```diff
@@ -692,8 +697,9 @@ public void update() {
         }
 
         actions.put(Actions.TOGGLE_PREVIEW, (BaseAction) () -> {
-            boolean enabled = !Globals.prefs.getBoolean(JabRefPreferences.PREVIEW_ENABLED);
-            Globals.prefs.putBoolean(JabRefPreferences.PREVIEW_ENABLED, enabled);
+            PreviewPreferences previewPreferences = new PreviewPreferences(Globals.prefs);
+            boolean enabled = !previewPreferences.isPreviewEnabled();
```
Siedlerchr: I find that a bit confusing: Is it actually enabled or disabled?

bartsch-dev: The boolean stores if the preview is enabled.
This line simply inverses what is saved in the preferences.

tobiasdiez: so better rename it to `toggled/invertedIsEnabled`

koppor: Without context, it is not clear, why this is inversed. If the user chooses
to have a preview, he gets none (!) and vice versa? Why?

src/main/java/net/sf/jabref/logic/citationstyle/CitationStyleCache.java
```diff
@@ -0,0 +1,86 @@
...
+import net.sf.jabref.model.event.EntryChangedEvent;
+import net.sf.jabref.model.event.EntryRemovedEvent;
+
+import com.google.common.eventbus.Subscribe;
+
+
+/**
+ * Caches the generated Citations for quicker access
+ */
+public class CitationStyleCache {
```
tobiasdiez: Is generating the preview so slow that a cache is needed? You are investing quite a bit of memory to save what looks to me like a few stringy operations.

src/main/java/net/sf/jabref/logic/citationstyle/CitationStyleGenerator.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    /**
+     * Generates a Citation based on the given entry, style, and output format
+     * WARNING: this may take some time, better call it in outside the main Thread
+     */
+    public static String generateCitation(BibEntry entry, String style, CitationStyleOutputFormat outputFormat) {
+        try {
+            String parsedEntry = new UnicodeToLatexFormatter().format(entry.toString());
+            InputStream stream = new ByteArrayInputStream(parsedEntry.getBytes(StandardCharsets.UTF_8));
```
tobiasdiez: I don't like that the BibEntry is converted to a string and then to a BibTeXEntry, which is then used in the `BibTeXItemDataProvider`. 
I think it shouldn't be a big problem to create a `BibTeXEntry` via the [addAllFields](https://github.com/jbibtex/jbibtex/blob/master/src/main/java/org/jbibtex/BibTeXEntry.java#L52) directly from our `BibEntry`.

src/main/java/net/sf/jabref/logic/citationstyle/CitationStyleCache.java
bartsch-dev: The library generates the citation with JavaScript. In the beginning it will take about a second, after the JavaScript engine warmup it is much quicker but still noticeable (100ms-200ms), especially if you scroll with the arrow keys over multiple entries.

src/main/java/net/sf/jabref/logic/citationstyle/CitationStyleGenerator.java
bartsch-dev: I rewrote that section.
It looks a bit cleaner but I have to parse each field as the `addAllFields`-method expects `Map<org.jbibtex.Key, org.jbibtex.Value>` instead of our `Map<String, String>`.

src/main/java/net/sf/jabref/logic/citationstyle/CitationStyleGenerator.java
```diff
@@ -42,21 +42,17 @@ public static String generateCitation(BibEntry entry, String style) {
...
-            BibTeXDatabase db = new BibTeXConverter().loadDatabase(stream);
-            BibTeXItemDataProvider provider = new BibTeXItemDataProvider();
-            provider.addDatabase(db);
-
-            CSL citeproc = new CSL(provider, style);
-            citeproc.setOutputFormat(outputFormat.format);
-            provider.registerCitationItems(citeproc);
-
-            Bibliography bibliography = citeproc.makeBibliography();
+            CSLItemData cslItemData = new BibTeXConverter().toItemData(bibTeXEntry);
```
koppor: Is it really required to instantiate BibTeXConverter each time? - The `UNICODE_TO_LATEX_FORMATTER` is also instantiated only once.

bartsch-dev: extracted

src/main/java/net/sf/jabref/logic/citationstyle/CitationStyleGenerator.java
tobiasdiez: Looks good to me now. Thanks!

src/main/java/net/sf/jabref/logic/citationstyle/CitationStyleCache.java
tobiasdiez: ok that's a good reason to cache it, could you add your explanation to the cache class. Thanks.

src/main/java/net/sf/jabref/logic/citationstyle/CitationStyleCache.java
```diff
@@ -0,0 +1,86 @@
...
+
+    private CitationStyle citationStyle;
+    private Map<String, Map<BibEntry, String>> citationStylesCache = new HashMap<>();
```
tobiasdiez: I would only cache for one citation style and reset the whole cache if another style is used (in `setCitationStyle`). That makes the code more understandable and I suspect most people actually use just one citation style or don't constantly switch between them (at least I never used the second preview). 
Or are there good reasons to store them for different styles?

src/main/java/net/sf/jabref/logic/citationstyle/CitationStyleCache.java
```diff
@@ -0,0 +1,86 @@
...
+
+
+    private class BibDatabaseEntryListener {
+        /**
+         * removes the outdated citation of the changed entry
+         */
+        @Subscribe
+        public void listen(EntryChangedEvent entryChangedEvent) {
+            if (entryChangedEvent != null && entryChangedEvent.getBibEntry() != null){
+                citationStylesCache.get(citationStyle.getSource()).remove(entryChangedEvent.getBibEntry());
```
tobiasdiez: If you keep using the current system (cache for every citation style), then you should also remove the cached result for every style here (and below).

src/main/java/net/sf/jabref/logic/citationstyle/CitationStyleGenerator.java
```diff
@@ -0,0 +1,74 @@
...
+import org.apache.commons.logging.LogFactory;
+import org.jbibtex.BibTeXDatabase;
+import org.jbibtex.ParseException;
+import org.jbibtex.TokenMgrError;
+
+
+/**
+ * WARNING: The generation of a citation may take some time, better call it in outside the main Thread
+ */
+public class CitationStyleGenerator {
```
tobiasdiez: You can also mark the class as protected so that nobody outside of the package can use it (i.e. everybody is forced to use the cache).

src/main/java/net/sf/jabref/preferences/PreviewPreferences.java
```diff
@@ -0,0 +1,108 @@
...
+    private static final String CYCLE_PREVIEW_POS = "cyclePreviewPos";
+    private static final String CYCLE_PREVIEW = "cyclePreview";
+    private static final String PREVIEW_PANEL_HEIGHT = "previewPanelHeight";
+    private static final String PREVIEW_STYLE = "previewStyle";
+    private static final String PREVIEW_ENABLED = "previewEnabled";
+
+    private final JabRefPreferences preferences;
+
+
+    public PreviewPreferences(JabRefPreferences preferences) {
```
tobiasdiez: See @oscargus recent PR's (for example https://github.com/JabRef/jabref/pull/1969) for how preference classes are structured (i.e. constructor takes a bunch of arguments describing the settings and `JabRefPreferences` handles the initialization and storage from/to the preferences.

bartsch-dev: The way @oscargus did it you can only save all preferences at the same time. What if I want just to save the preview panel height and nothing else?

src/main/java/net/sf/jabref/logic/citationstyle/CitationStyleOutputFormat.java
```diff
@@ -0,0 +1,23 @@
...
+
+public enum CitationStyleOutputFormat {
+
+    HTML("html"),
+    TEXT("text"),
+    ASCII_DOC("asciidoc"),
+    FO("fo"),
+    RTF("rtf");
+
+    public final String format;
```
bartsch-dev: fixed

src/main/java/net/sf/jabref/logic/citationstyle/CitationStyleGenerator.java
bartsch-dev: ^done

src/main/java/net/sf/jabref/logic/citationstyle/CitationStyleCache.java
bartsch-dev: reduced the cache to one citationStyle

src/main/java/net/sf/jabref/logic/citationstyle/CitationStyleCache.java
bartsch-dev: reduced the cache to one citationStyle

src/main/java/net/sf/jabref/logic/citationstyle/CitationStyleGenerator.java
bartsch-dev: why not package-private?

src/main/java/net/sf/jabref/preferences/PreviewPreferences.java
tobiasdiez: Mhhh... I would suggest to always read/write all the preference values. With this you have a performance overhead but the code looks better.

src/main/java/net/sf/jabref/logic/citationstyle/CitationStyleGenerator.java
tobiasdiez: I don't like package-private (because writing nothing looks like somebody just forgot to specify the correct visibility)

CHANGELOG.md
```diff
@@ -20,6 +20,8 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - The search result Window won't stay on top anymore if the main Window is focused and will be present in the taskbar
 - The user can jump from the searchbar to the maintable  with `ctrl+enter`
 - Implemented [#573 (comment)](https://github.com/JabRef/jabref/issues/573#issuecomment-232284156): Added shortcut: closing the search result window with `ctrl+w`
+- Added support for over 1000 1.0.1 CitationStyles
```
tobiasdiez: Maybe reformulate to 1000 citation styles (using CitationStyles 1.0.1) and add link.

src/main/java/net/sf/jabref/gui/PreviewPanel.java
```diff
@@ -203,12 +187,43 @@ public void setDatabaseContext(BibDatabaseContext databaseContext) {
...
+        if (fixedLayout) {
+            LOGGER.debug("cannot change the layout because the layout is fixed");
+            return;
+        }
+
+        PreviewPreferences previewPreferences = new PreviewPreferences(Globals.prefs);
+        String style = previewPreferences.getPreviewCycle().get(previewPreferences.getCyclePreviewPosition());
+
+        if (CitationStyle.isCitationStyleFile(style)) {
+            if (getBasePanel() != null) {
```
tobiasdiez: basepanel.ifPresent(...) and maybe remove `getBasePanel`

src/main/java/net/sf/jabref/gui/PreviewPanel.java
```diff
@@ -203,12 +187,43 @@ public void setDatabaseContext(BibDatabaseContext databaseContext) {
...
+                layout = Optional.empty();
+                CitationStyle citationStyle = CitationStyle.createCitationStyleFromFile(style);
+                if (citationStyle != null) {
+                    getBasePanel().getCitationStyleCache().setCitationStyle(citationStyle);
+                    getBasePanel().output(Localization.lang("Preview style changed to: %0", citationStyle.getTitle()));
+                }
+            }
+        } else {
+            updatePreviewLayout(previewPreferences.getPreviewStyle());
+            if (getBasePanel() != null) {
```
tobiasdiez: Same as above

src/main/java/net/sf/jabref/gui/PreviewPanel.java
```diff
@@ -250,30 +264,40 @@ public BibEntry getEntry() {
...
-        StringBuilder sb = new StringBuilder();
         ExportFormats.entryNumber = 1; // Set entry number in case that is included in the preview layout.
-        bibEntry.ifPresent(entry ->
-                layout.ifPresent(acutalLayout -> sb.append(acutalLayout
-                        .doLayout(entry, databaseContext.map(BibDatabaseContext::getDatabase).orElse(null),
-                                highlightPattern)))
-        );
-        String newValue = sb.toString();
 
+        if (citationStyleWorker != null){
```
tobiasdiez: Make citationStyleWorker also an Optional.

src/main/java/net/sf/jabref/gui/maintable/MainTableSelectionListener.java
```diff
@@ -79,28 +77,18 @@ public MainTableSelectionListener(BasePanel panel, MainTable table) {
...
-                        Globals.prefs.get(JabRefPreferences.PREVIEW_0)),
-                new PreviewPanel(panel.getBibDatabaseContext(), null, panel,
-                        Globals.prefs.get(JabRefPreferences.PREVIEW_1))};
-
-        panel.frame().getGlobalSearchBar().getSearchQueryHighlightObservable()
-                .addSearchListener(previewPanel[0])
-                .addSearchListener(previewPanel[1]);
-
-        this.preview = previewPanel[activePreview];
+        if (panel.getPreviewPanel() != null){
```
tobiasdiez: Make getPreviewPanel return an Optional

src/main/java/net/sf/jabref/gui/worker/CitationStyleWorker.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    @Override
+    protected String doInBackground() throws Exception {
+        BasePanel basePanel = this.previewPanel.getBasePanel();
+        BibEntry entry = this.previewPanel.getEntry();
+        Optional<Pattern> highlightPattern = this.previewPanel.getHighlightPattern();
+
+        String fieldText = "";
+        if (entry != null && basePanel != null) {
```
tobiasdiez: Optionals again

src/main/java/net/sf/jabref/logic/citationstyle/CitationStyleCache.java
```diff
@@ -0,0 +1,82 @@
...
+        String citation = citationStylesCache.get(entry);
+        if (citation == null) {
+            citation = CitationStyleGenerator.generateCitation(entry, this.citationStyle.getSource());
+            citationStylesCache.put(entry, citation);
+        }
+        return citation;
+    }
+
+    public void setCitationStyle(CitationStyle citationStyle) {
+        if (citationStyle == null || this.citationStyle == citationStyle) {
```
tobiasdiez: I would guard `citationStyle` from being null (Objects.requireNonNull).

src/main/java/net/sf/jabref/logic/citationstyle/CitationStyleCache.java
```diff
@@ -0,0 +1,82 @@
...
+            return;
+        }
+        if (this.citationStyle == null || !this.citationStyle.getSource().equals(citationStyle.getSource())) {
```
tobiasdiez: Overwrite equals in CitationStyle and reuse here.

src/main/java/net/sf/jabref/logic/citationstyle/CitationStyleCache.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+
+    private class BibDatabaseEntryListener {
+        /**
+         * removes the outdated citation of the changed entry
+         */
+        @Subscribe
+        public void listen(EntryChangedEvent entryChangedEvent) {
+            if (entryChangedEvent != null && entryChangedEvent.getBibEntry() != null){
```
tobiasdiez: Can the event and/or the BibEntry ever be null?

src/main/java/net/sf/jabref/logic/citationstyle/CitationStyleGenerator.java
```diff
@@ -0,0 +1,71 @@
...
+            Bibliography bibliography = CSL.makeAdhocBibliography(style, outputFormat.getFormat(), cslItemData);
+            return bibliography.getEntries()[0];
+
+        } catch (IOException  e) {
+            LOGGER.error("Could not generate BibEntry Citation", e);
+        } catch (TokenMgrError e) {
+            LOGGER.error("Bad character inside BibEntry", e);
+            // sadly one can not easily retrieve the bad char from the TokenMgrError
+            return  new StringBuilder()
+                    .append(Localization.lang("Bad character inside BibEntry"))
```
tobiasdiez: Use a proper word instead of `BibEntry` (no idea what we usually use...probably `entry`)

src/test/java/net/sf/jabref/logic/citationstyle/CitationStyleTest.java
```diff
@@ -0,0 +1,14 @@
...
+import org.junit.Test;
+
+
+public class CitationStyleTest {
+
+    @Test
+    public void getDefault() throws Exception {
+        Assert.assertNotNull(CitationStyle.getDefault());
+    }
+
```
tobiasdiez: Please add some tests. At least one test should verify that a BibEntry is properly converted into the correct string.

src/main/java/net/sf/jabref/preferences/PreviewPreferences.java
bartsch-dev: Reading isn't a Problem, but writing looks quite ugly in my opinion, or how else could I do it?

```Java
PreviewPreferences previewPreferences = preferences.getPreviewPreferences();
PreviewPreferences newPreviewPreferences = new PreviewPreferences(previewheight, previewPreferences.isEnabled(), previewPreferences.getpreviewStyle(), previewPreferences.getPreviewCycle());
```

CHANGELOG.md
bartsch-dev: done

src/main/java/net/sf/jabref/gui/PreviewPanel.java
bartsch-dev: done

src/main/java/net/sf/jabref/gui/PreviewPanel.java
bartsch-dev: done

src/main/java/net/sf/jabref/gui/maintable/MainTableSelectionListener.java
bartsch-dev: done

src/main/java/net/sf/jabref/gui/worker/CitationStyleWorker.java
bartsch-dev: panel.getPreviewPanel()

src/main/java/net/sf/jabref/logic/citationstyle/CitationStyleCache.java
bartsch-dev: panel.getPreviewPanel()

src/main/java/net/sf/jabref/logic/citationstyle/CitationStyleCache.java
bartsch-dev: done

src/main/java/net/sf/jabref/logic/citationstyle/CitationStyleGenerator.java
bartsch-dev: `entry`was used in other translations

src/main/java/net/sf/jabref/logic/citationstyle/CitationStyleCache.java
bartsch-dev: The event itself shouldn't be null, for its properties there are no null checks.

src/test/java/net/sf/jabref/logic/citationstyle/CitationStyleTest.java
bartsch-dev: I added a Test to check the generated citation with the default citation style on the test bibEntry.

src/main/java/net/sf/jabref/gui/PreviewPanel.java
```diff
@@ -113,23 +105,15 @@ public PreviewPanel(BibDatabaseContext databaseContext, BibEntry entry,
...
-        this.closeAction = new CloseAction();
-        this.printAction = new PrintAction();
-        this.copyPreviewAction = new CopyPreviewAction();
-
         this.basePanel = Optional.ofNullable(panel);
 
         createPreviewPane();
+        updateLayout();
 
         if (panel != null) {
```
oscargus: Use `this.basePanel` instead here?

src/main/java/net/sf/jabref/preferences/PreviewPreferences.java
oscargus: Either like that or add methods like `previewPreferences.withPreviewHeight(previewHeight)`.

Depending on where the writing actually occurs, you may write the new data directly as well.

src/main/java/net/sf/jabref/preferences/PreviewPreferences.java
```diff
@@ -0,0 +1,108 @@
...
+        this.preferences = preferences;
+    }
+
+    public static void putDefaults(Map<String, Object> defaults) {
```
oscargus: Although there are some classes that use this, as far as I know, the decision is that all constants and all defaults setting should be in `JabRefPreferences`.

bartsch-dev: They are back in `JabRefPreferences`.

src/main/java/net/sf/jabref/gui/PreviewPanel.java
bartsch-dev: done

src/main/java/net/sf/jabref/preferences/PreviewPreferences.java
bartsch-dev: I rewrote it using the builder pattern.

src/main/java/net/sf/jabref/logic/citationstyle/CitationStyleGenerator.java
koppor: Just look at other localizations.

[Please write cannot](https://en.oxforddictionaries.com/usage/cannot-or-can-not)

src/main/java/net/sf/jabref/logic/citationstyle/CitationStyleCache.java
koppor: Then just add `Objects.requireNonNull(entryChangedEvent)` and remove the `!= null` check.

koppor: Further please investigate the "call hierarchy" whether `getBibEntry()` may ever be null. I don't think so. Thus, please add `Objects.requireNonNull(entryChangedEvent.getBibEntry();`.

src/main/java/net/sf/jabref/gui/PreviewPanel.java
bartsch-dev: done

src/main/java/net/sf/jabref/logic/citationstyle/CitationStyleCache.java
bartsch-dev: Is it not more useful to add the `Objects.requireNonNull` to the constructor of the `EntryEvent`?

src/main/java/net/sf/jabref/logic/citationstyle/CitationStyleGenerator.java
bartsch-dev: done

src/main/java/net/sf/jabref/logic/citationstyle/CitationStyleCache.java
koppor: The more places the better

Olifair:
Is the use of Citation Styles only for the preview pane ?  I have downloaded the latest developer version and connected to a LibreOffice document to see if I could use the CSL styles to format my bibliography, but when I open the "Select style" window, only the jstyle format seems to be supported.

georgd:
I think, this would be the next logical step: to use Citation Styles in Libre Office.

koppor:
Please open a new issue to offer tracking the status. Maybe @oscargus has some time to work on that.

wujastyk:
If I use Biblatex style biblio entries, then I have fields like "journaltitle," "location," and "date" instead of "journal," "address," and "year."  These are not picked up by the new Preview citation styles.  

Further, as reported by boceckts above, one can't get the citation style displayed, so as to edit it.

The whole new Preview citation feature is wonderfully helpful. I use it a lot!  Thank you.

PR #1929, opened 2016-09-07
New fetcher

tobiasdiez:
A few new fetcher are introduced:

- ADS search based
- MathSciNet 
- Zentralblatt

Also a bit of code was refactored (mainly to facilitate the development of fetcher). 


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 ## [Unreleased]
 
 ### Changed
+- Added fetcher for MathSciNet, zbMATH and Astrophysics Data System
```
koppor: Is it possible to add links to these databases?

src/test/java/net/sf/jabref/logic/util/io/RegExpFileSearchTests.java
```diff
@@ -30,25 +25,23 @@
...
-
-        BibtexParser parser = new BibtexParser(reader, JabRefPreferences.getInstance().getImportFormatPreferences());
-        ParserResult result = null;
-
-        result = parser.parse();
-
-        database = result.getDatabase();
-        entry = database.getEntryByKey("HipKro03").get();
-
-        Assert.assertNotNull(database);
```
koppor: Assert statements removed because of performance reasons?

src/main/java/net/sf/jabref/logic/importer/fileformat/BibtexParser.java
```diff
@@ -79,9 +84,9 @@ public BibtexParser(Reader in, ImportFormatPreferences importFormatPreferences)
      * @param in the Reader to read from
      * @throws IOException
      */
+    @Deprecated // inline this method
```
koppor: Please add `@deprecated` JavaDoc comment, too.

src/main/java/net/sf/jabref/logic/importer/fileformat/BibtexParser.java
```diff
@@ -119,6 +124,16 @@ public static ParserResult parse(Reader in, ImportFormatPreferences importFormat
         return Optional.of(entries.iterator().next());
     }
 
+    @Override
+    public List<BibEntry> parseEntries(InputStream inputStream) throws ParserException {
+        BufferedReader reader = new BufferedReader(new InputStreamReader(inputStream, StandardCharsets.UTF_8));
```
koppor: On the one hand, I think, it's good to have `UTF-8` as default encoding. On the other hand, there are preferences where the user sets the default encoding.

Maybe we should remove the global preference and always set UTF-8 as default encoding?

src/test/java/net/sf/jabref/logic/importer/fileformat/BibtexParserTest.java
```diff
@@ -127,15 +127,6 @@ public void singleFromStringReturnsNullIfNoEntryRecognized() {
     }
 
     @Test
-    public void parseTwoTimesReturnsSameResult() throws IOException {
```
koppor: Why can't an equivalent test be written using the new structure? Two times calling the parser with the string thing and then compare the resulting entries?

koppor:
Is it intentional that 4fc8551cbe2b32e708eeeea20d86b51115b459cb and 443f50fc4e05a56986f9c0dff6790a23f24f9a5e went into this commit? Seems to be unrelated.

src/main/java/net/sf/jabref/logic/importer/Parser.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+package net.sf.jabref.logic.importer;
+
+import java.io.InputStream;
+import java.util.List;
+
+import net.sf.jabref.model.entry.BibEntry;
+
+public interface Parser {
```
koppor: Could you add a JavaDoc comment here?

src/test/java/net/sf/jabref/logic/importer/fetcher/GvkParserTest.java
```diff
@@ -0,0 +1,71 @@
...
+            for (String resourceName : resourceNames) {
+                BibEntryAssert.assertEquals(GvkParser.class, resourceName, entries.get(i));
+                i++;
+            }
+        }
+    }
+
+    @Test
+    public void emptyResult() throws Exception {
+        doTest("gvk_empty_result_becaue_of_bad_query.xml", 0, Collections.emptyList());
```
koppor: Could you rename the file to `gvk_empty_result_becaue_of_bad_query.xml`? `s` missing in `becaue`

src/main/java/net/sf/jabref/logic/importer/SearchBasedParserFetcher.java
```diff
@@ -48,14 +50,36 @@
...
+     * (i.e. if information is consistently contained in the wrong field or the wrong format)
+     * but not cosmetic issues which may depend on the user's taste (for example, LateX code vs HTML in the abstract).
+     *
+     * Try to reuse existing {@link Formatter} for the cleanup. For example,
+     * {@code new FieldFormatterCleanup(FieldName.TITLE, new RemoveBracesFormatter()).cleanup(entry);}
+     *
+     * By default, no cleanup is done.
+     * @param entry the entry to be cleaned-up
+     */
+    default void doPostCleanup(BibEntry entry) {
```
koppor: @zellerdev This is interesting for you as we discussed this about two weeks ago.

src/main/java/net/sf/jabref/logic/importer/SearchBasedParserFetcher.java
```diff
@@ -48,14 +50,36 @@
...
         if (StringUtil.isBlank(query)) {
             return Collections.emptyList();
         }
 
         try (InputStream stream = new BufferedInputStream(getQueryURL(query).openStream())) {
-            return getParser().parseEntries(stream);
+            List<BibEntry> fetchedEntries = getParser().parseEntries(stream);
+
+            // Post-cleanup
+            fetchedEntries.forEach(this::doPostCleanup);
```
koppor: Why is the implementor of `SearchBasedParserFetcher` required to do the cleanup. Why isn't there a fetcher framework which calls `performSearch` and after successfull fetching `doPostCleanup`?

src/main/java/net/sf/jabref/logic/importer/fetcher/AstrophysicsDataSystem.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    @Override
+    public String getName() {
+        return "SAO/NASA Astrophysics Data System";
+    }
+
+    @Override
+    public HelpFile getHelpPage() {
+        return null;
```
koppor: Don't we have a default implementation? Maybe, we want to force the developers to think about a help page? Nevertheless, I think a default implementation returning `null` would reduce the barrier to implement new fetchers, wouldn't it?

src/main/java/net/sf/jabref/logic/importer/fetcher/AstrophysicsDataSystem.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    @Override
+    public Parser getParser() {
+        return new BibtexParser(preferences);
+    }
+
+    @Override
+    public void doPostCleanup(BibEntry entry) {
+        new FieldFormatterCleanup(FieldName.ABSTRACT, new RemoveBracesFormatter()).cleanup(entry);
```
koppor: Comment for future work: Can we rework the formatter code to follow some sort of builder pattern? I once stumbled upon [EqualsBuilder](https://commons.apache.org/proper/commons-lang/javadocs/api-3.4/org/apache/commons/lang3/builder/EqualsBuilder.html) and found it quite nice.

src/main/java/net/sf/jabref/logic/importer/fetcher/AstrophysicsDataSystem.java
```diff
@@ -0,0 +1,88 @@
...
+        new FieldFormatterCleanup(FieldName.TITLE, new RemoveBracesFormatter()).cleanup(entry);
+        new FieldFormatterCleanup(FieldName.AUTHOR, new NormalizeNamesFormatter()).cleanup(entry);
+        new FieldFormatterCleanup("adsnote", new ClearFormatter()).cleanup(entry);
```
koppor: Could you provide example content in the comment and reason why it is not included? Personally, I keep as much information in the bib files as the styles ignore them nevertheless.

src/test/java/net/sf/jabref/logic/importer/fetcher/AstrophysicsDataSystemTest.java
```diff
@@ -0,0 +1,83 @@
...
+        diezSliceTheoremEntry.setField("archiveprefix", "arXiv");
+        diezSliceTheoremEntry.setField("eprint", "1405.2249");
+        diezSliceTheoremEntry.setField("journal", "ArXiv e-prints");
+        diezSliceTheoremEntry.setField("keywords", "Mathematical Physics, Mathematics - Differential Geometry, Mathematics - Symplectic Geometry, 58B99, 58Z05, 58B25, 22E65, 58D19, 53D20, 53D42");
+        diezSliceTheoremEntry.setField("month", "#may#");
+        diezSliceTheoremEntry.setField("primaryclass", "math-ph");
+        diezSliceTheoremEntry.setField("abstract",
+                "A general slice theorem for the action of a Fr$\\backslash$'echet Lie group on a" + NEWLINE
+                        + "Fr$\\backslash$'echet manifolds is established. The Nash-Moser theorem provides the"
+                        + NEWLINE + "fundamental tool to generalize the result of Palais to this" + NEWLINE
```
koppor: Could you  move the `NEWLINE` one line up to be consistent with the other lines?

src/main/java/net/sf/jabref/logic/importer/EntryBasedParserFetcher.java
```diff
@@ -0,0 +1,87 @@
...
+        // Do nothing by default
+    }
+
+    @Override
+    default List<BibEntry> performSearch(BibEntry entry) throws FetcherException {
+        try (InputStream stream = new BufferedInputStream(getURLForEntry(entry).openStream())) {
+            List<BibEntry> fetchedEntries = getParser().parseEntries(stream);
+
+            // Post-cleanup
+            fetchedEntries.forEach(this::doPostCleanup);
```
koppor: Same comment as at the other fetcher interface: Shouldn't we have a framework so that the not each implementer of a fetcher has to take care about the right calling, but just offer the cleanup method? -- Maybe, this is too much work and we have to stick with the current solution. Then, the method `doPostCleanup` is just an implementation recommendation without any outside use, isn't it?

tobiasdiez:
Thanks @koppor for reviewing. https://github.com/JabRef/jabref/commit/4fc8551cbe2b32e708eeeea20d86b51115b459cb is needed because it refactors the BibtexParser to confirm to the new `Parser` interface. But https://github.com/JabRef/jabref/commit/443f50fc4e05a56986f9c0dff6790a23f24f9a5e is not really needed (I tried to implement the `doPostCleanup` method via `CleanupWorker` but it turned out to be to messy and unneeded).

Will incorporate your feedback.

src/main/java/net/sf/jabref/logic/importer/fetcher/AstrophysicsDataSystem.java
```diff
@@ -42,10 +25,12 @@
  * Fetches data from the SAO/NASA Astrophysics Data System (http://www.adsabs.harvard.edu/)
  *
  * Search query-based: http://adsabs.harvard.edu/basic_search.html
+ * Entry -based: http://adsabs.harvard.edu/abstract_service.html
  */
-public class AstrophysicsDataSystem implements SearchBasedParserFetcher {
+public class AstrophysicsDataSystem implements SearchBasedParserFetcher, EntryBasedParserFetcher {
 
-    private static String API_URL = "http://adsabs.harvard.edu/cgi-bin/nph-basic_connect";
+    private static String API_QUERY_URL = "http://adsabs.harvard.edu/cgi-bin/nph-basic_connect";
```
koppor: make both constants final?

src/main/java/net/sf/jabref/logic/importer/fileformat/BibtexParser.java
```diff
@@ -95,12 +95,12 @@ public static ParserResult parse(Reader in, ImportFormatPreferences importFormat
      * @param bibtexString
      * @return Returns returns an empty collection if no entries where found or if an error occurred.
      */
+    @Deprecated // use parseEntries
```
koppor: Could you add JavaDocs `@deprecated` instead of the comment?

src/main/java/net/sf/jabref/logic/util/OptionalUtil.java
```diff
@@ -0,0 +1,26 @@
+package net.sf.jabref.logic.util;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import static org.bouncycastle.asn1.ua.DSTU4145NamedCurves.params;
```
koppor: Do we really need bouncycastle here?

src/main/java/net/sf/jabref/logic/util/OptionalUtil.java
```diff
@@ -0,0 +1,26 @@
...
+
+public class OptionalUtil {
```
koppor: Looks good. I wonder, why there is no other similar class already existing in the OSS universe. Nevertheless, let's go forward!

src/main/java/net/sf/jabref/logic/importer/fetcher/zbMATH.java
```diff
@@ -0,0 +1,138 @@
...
+        uriBuilder.addParameter("start", "0"); // start index
+        uriBuilder.addParameter("count", "200"); // should return up to 200 items (instead of default 100)
+
+        fixSSLVerification();
+
+        return uriBuilder.build().toURL();
+    }
+
+    /**
+     * Older java VMs does not automatically trust the zbMATH certificate. In this case the following exception is thrown:
```
koppor: Can we install this hack only if JRE is 8 and < 101? I feel uncomfortable to install that fix globally.

Alternatively, can this handler be uninstalled after using this fetcher? Maybe when calling `getParser`?

src/test/java/net/sf/jabref/logic/importer/fetcher/zbMATHTest.java
```diff
@@ -0,0 +1,50 @@
...
+
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class zbMATHTest {
+    zbMATH fetcher;
```
koppor: Is package private really necessary? Shouldn't it be private?

koppor:
Went through everything. Besides https://github.com/JabRef/jabref/pull/1929/#discussion_r78320105 everything else are minor comments.

src/main/java/net/sf/jabref/logic/importer/fileformat/BibtexParser.java
Siedlerchr: At least in the tests it's useful to have it fixed to UTF8. But we should be able to support other encodings for import.

tobiasdiez: This is just to parse a bunch of BibEntries which may come from anywhere. Importing / opening a database uses the other parse method which uses the proper encoding specified in the file or preferences.

src/test/java/net/sf/jabref/logic/importer/fileformat/BibtexParserTest.java
tobiasdiez: No not really. The code no longer stores the previous result but parses the reader again and again.

src/test/java/net/sf/jabref/logic/util/io/RegExpFileSearchTests.java
tobiasdiez: Because they are no longer needed (entry and db are always non-null if they are created by hand)

PR #1930, opened 2016-09-07
Integrity Check for note and howpublished field

grimes2:
In BibTeX mode, the first word should be capitalized for the fields note and howpublished. In BibLaTeX mode, there are no restrictions for these fields.

See also #1914


tobiasdiez:
LGTM :+1:

src/main/java/net/sf/jabref/logic/integrity/HowpublishedChecker.java
```diff
@@ -0,0 +1,48 @@
...
+
+    private final BibDatabaseContext bibDatabaseContextEdition;
+
+
+    public HowpublishedChecker(BibDatabaseContext bibDatabaseContext) {
+        this.bibDatabaseContextEdition = Objects.requireNonNull(bibDatabaseContext);
+    }
+
+    /**
+     * BibLaTeX package documentation p.222:
```
Siedlerchr: Better add the section reference, as page numbers may change

src/main/java/net/sf/jabref/logic/integrity/EditionChecker.java
```diff
@@ -16,7 +16,7 @@
 public class EditionChecker implements Checker {
 
     private static final Predicate<String> FIRST_LETTER_CAPITALIZED = Pattern.compile("^[A-Z]").asPredicate();
-    private static final Predicate<String> ONLY_NUMERALS_OR_LITERALS = Pattern.compile("^([0-9]+|[^0-9]+)$")
+    private static final Predicate<String> ONLY_NUMERALS_OR_LITERALS = Pattern.compile("^([0-9]+|[^0-9].+)$")
```
Siedlerchr: Why did this change?

grimes2: Literals are numbers, letters and strings. [^0-9]+ excludes numbers. This is an unintentinal restriction. The fullstop indicates, that all literals are allowed, except for the first character, that should not be a number (to allow "Edition 2000" and avoid something like "2nd").

stefan-kolb:
Hm, do we really need two checker for this very trivial functionality? I tend to either not restrict this or make one checker for fields that should start with uppercase letters.

Siedlerchr:
@grimes2  Please check if their are other fields which have the same requirements with uppercase letters. It would then be ideal to have one checker (e.g. StartsWitchUpperCaseChecker:..) which checks for mulitple fields.

grimes2:
According to the bibtex specification `edition`, `note` and `howpublished` are the only fields with requirement upper case. The advantage of the field checker (HowpublishedChecker) over "function" checker (StartsWitchUpperCaseChecker) is, that each field checker can be expanded easily, if there are new requirements for that field. Furthermore it should be consistent: FieldChecker or FunctionChecker.

tobiasdiez:
I would also prefer a `StartsWitchUpperCaseChecker`. This is probably the only requirement for these fields and even if not then one could still create a new `HowpublishedChecker`. 
We are already mixing field and function checkers, for example `AbbreviationChecker` works exactly like a `StartsWitchUpperCaseChecker`. 

Could you please change the code accordingly, @grimes2. Thanks. ("ready-for-review" removed for the moment)

grimes2:
This request exceeds my skills. So I withdraw my pull request.

Siedlerchr:
Okay, then I would say that we first merge this and create a follow up issue for this.

Siedlerchr:
@grimes2  Thanks for your contribution!

PR #1933, opened 2016-09-07
New Gradle Task: sort Localization Keys

bartsch-dev:
As soon as you have a PR which adds new localization propertires and another PR with new localization gets merged you have to solve these conflicts which is just annoying.

With this gradle task the keys in the Localization files get rearranged according to the English file. 
The top 3 comment lines of each file stay the same (some files have different encoding) but everything else (comments, blank lines) are getting synced with the English file.

The localization file are only sorted if they have no missing, obsolete, or duplicate keys.
A colored output is printed to inform the developer.

When this PR has passed I plan to refactor `syncLang.py`, code and output wise.


mlep:
Just to make sure: Will this PR keep empty entries at the bottom of the file? (that makes localization much easier)

bartsch-dev:
No, it will set each entry to the exact same line it is in the English file.

mlep:
Thanks for your quick reply.

Siedlerchr:
@mlep  I think for new entries (not yet present in English), they will be still stay and  added at the bottom.

src/main/resources/l10n/JabRef_no.properties
```diff
@@ -968,53 +856,45 @@ no_database_generated=ingen_database_generert
...
 
-
 Note_that_you_must_specify_the_fully_qualified_class_name_for_the_look_and_feel,=Merk_at_du_m\u00e5_spesifisere_det_fullstendige_klassenavnet_for_utseendet,
 
 Nothing_to_redo=Ingenting_\u00e5_gjenta
 
 Nothing_to_undo=Ingenting_\u00e5_angre
 
-# The next is used like in "References found\: 1  Number of references to fetch?"
+#Thenextisusedlikein"Referencesfound\:1Numberofreferencestofetch?"
```
koppor: Why are the spaces removed?

src/main/resources/l10n/JabRef_no.properties
```diff
@@ -576,28 +511,24 @@ External_programs=Eksterne_programmer
...
 
-
 Fetch=Hent
 
 Field=Felt
 
 field=felt
 
-# Integrity check is a process that checks for indications of wrongly filled out BibTeX fields. "Scan" is the button that starts the check.
+#IntegritycheckisaprocessthatchecksforindicationsofwronglyfilledoutBibTeXfields."Scan"isthebuttonthatstartsthecheck.
```
koppor: See below - spaces removed?

src/main/resources/l10n/JabRef_no.properties
bartsch-dev: The spaces are already missing in the current English localization.
Comments in languages other than English overridden by the English one.

https://github.com/JabRef/jabref/blob/master/src/main/resources/l10n/JabRef_en.properties#L520

src/main/resources/l10n/JabRef_no.properties
bartsch-dev: see answer in the other comment

koppor: Then, please fix the spaces in the English version!

PR #1934, opened 2016-09-07
Removed external dependency in logic

oscargus:
An alternative to #1924 : hard coded `PS` and `PDF` as file type names.

Still some methods that can be removed in `ExternalFileType` (the original thing done in #1924 .

I created three `gui` packages:
 * `externalfiletypes` for things related to `ExternalFileType`
 * `externalfiles` for things (primarily) related to the actual files
 * `filelist` for things related to `FileList`


src/main/java/net/sf/jabref/gui/importer/ImportInspectionDialog.java
```diff
@@ -1224,7 +1224,7 @@ public void actionPerformed(ActionEvent actionEvent) {
             // links:
             JDialog diag = new JDialog(ImportInspectionDialog.this, true);
             JabRefExecutorService.INSTANCE.execute(
-                    net.sf.jabref.external.AutoSetLinks.autoSetLinks(entry, localModel, bibDatabaseContext, e -> {
+                    net.sf.jabref.gui.externalfiles.AutoSetLinks.autoSetLinks(entry, localModel, bibDatabaseContext, e -> {
```
tobiasdiez: use package import

tobiasdiez:
LGTM :+1:
I like this solution. Personally, I wouldn't separate externalfiletypes and externalfile, but this is probably a matter of taste.

lenhard:
LGTM as well! And I am in favor of the separation.

Just for clarification before merging: Do you still intend to move `ExternalFileType` and `ExternalFileTypes` into one of the other packages? Or, since the external dependencies on the classes are removed now, will you keep them in gui?

Both ways would be fine in a sense. If those classes are exclusively used in gui, it is fine to keep them there, even if they are could be refactored in a gui-independent fashion. 

Before merging the build has to succeed of course.

src/main/java/net/sf/jabref/gui/importer/ImportInspectionDialog.java
oscargus: Indeed.

oscargus:
The plan is to keep it as is as it solves the problem.

One may indeed think of different solutions regarding the amount of packages and there is no clear border, especially considering `filelist`.

Regarding the builds: clearly I picked the wrong field in `ExternalFileType`... Will fix it.

src/main/java/net/sf/jabref/gui/importer/ImportInspectionDialog.java
oscargus: There was a local class with the same name. I renamed that.

oscargus:
Correct field. Bad at reading...

oscargus:
Not sure why the Circle CI-tests are not executed, but the things are fixed now anyway.

tobiasdiez:
I closed and reopened the PR to give travis another chance to build the branch. 
So can we merge this?

lenhard:
Good idea and it worked! The build succeeded, so we can merge it.

PR #1940, opened 2016-09-08
Added default Icon to all JavaFX dialogs that inherit from FXAlert

boceckts:
This will add a small JabRef icon to the upper left of every dialog that is a subclass of FXAlert.
Also the keybindings dialog has used the localization strings from the menu files which displayed an and ("&") sign in front of the actual tranlsation. I changed them to use the "normal" localization files and updated the localization files as well. (There were quite some keys missing and/or not available)

This is the current state of the dialog (applies to all JavaFX dialogs)
keybindingsnoicon

and after the change
keybinding
keybiningssave
journalduplicate
about


src/main/java/net/sf/jabref/gui/keyboard/KeyBindingCategory.java
```diff
@@ -5,19 +5,19 @@
 public enum KeyBindingCategory {
 
     FILE(
-            Localization.menuTitle("File")),
+            Localization.lang("File")),
```
koppor: Why this change? Isn't the category mirroring the menu structure? Menu strings are translated in the menu translation files, not the normal language files.

src/main/java/net/sf/jabref/logic/l10n/Localization.java
```diff
@@ -15,7 +15,6 @@
 
     public static final String RESOURCE_PREFIX = "l10n/JabRef";
     public static final String MENU_RESOURCE_PREFIX = "l10n/Menu";
-    public static final String BIBTEX = "BibTeX";
```
koppor: Why did you remove that?

src/main/resources/l10n/JabRef_da.properties
```diff
@@ -1800,3 +1791,6 @@ should_have_the_first_letter_capitalized=
 Remember_password?=
 
 shared=
+
+Tools=
+BibTex=
```
koppor: BibTeX is always spelled BibTeX - see https://github.com/koppor/jabref/issues/73

BibTeX is a name like "OS/2" or "Linux" or "LaTeX" and must not be translated.

src/main/resources/l10n/JabRef_da.properties
```diff
@@ -1800,3 +1791,6 @@ should_have_the_first_letter_capitalized=
 Remember_password?=
 
 shared=
+
+Tools=
```
koppor: This is a menu title? This, it does not beed to be translated.

koppor:
The screenshots look good, but I wonder why you touched the language files. Please do one thing at one PR and do not mix issues in a PR.

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingCategory.java
boceckts: Yes it is but the translation strings have a leading and ("&") sign and since most of the strings are also in the non menu localization files I changed it.

On Thu, Sep 08, 2016 at 10:30 PM, Oliver Kopp < notifications@github.com [notifications@github.com] > wrote:
In src/main/java/net/sf/jabref/gui/keyboard/KeyBindingCategory.java [https://github.com/JabRef/jabref/pull/1940#discussion_r78082976] :

> @@ -5,19 +5,19 @@
>  public enum KeyBindingCategory {
>  
>      FILE(
> -            Localization.menuTitle("File")),
> +            Localization.lang("File")),


Why this change? Isn't the category mirroring the menu structure? Menu strings are translated in the menu translation files, not the normal language files.

src/main/java/net/sf/jabref/logic/l10n/Localization.java
boceckts: It was only used in the viewModel where I changed it to Localization.lang("BibTex").

On Thu, Sep 08, 2016 at 10:30 PM, Oliver Kopp < notifications@github.com [notifications@github.com] > wrote:
In src/main/java/net/sf/jabref/logic/l10n/Localization.java [https://github.com/JabRef/jabref/pull/1940#discussion_r78083018] :

> @@ -15,7 +15,6 @@
>  
>      public static final String RESOURCE_PREFIX = "l10n/JabRef";
>      public static final String MENU_RESOURCE_PREFIX = "l10n/Menu";
> -    public static final String BIBTEX = "BibTeX";


Why did you remove that?

koppor: BibTeX is always spelled BibTeX - see https://github.com/koppor/jabref/issues/73 -- BibTeX is a name like "OS/2" or "Linux" or "LaTeX" and must not be translated.

Thus `Localiatzion.lang("BibTex")` is unfortunately wrong in two aspects: 1) casing, 2) requiring translation

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingCategory.java
koppor: I would not require the translators to replace terms twice. Why not just removing the `&` after getting the translation from `menuTitle`?

Reason: You are displaying menu strings. And menu strings are contained in `Menu_X.properties`. It might happen that a menu string is translated differently in JAbRef_*.properties. Using `Menu_*.properties` avoids that issue.

boceckts:
Since I added something to the non menu translation files I updated them via gradle, that's why there is so many changes. I will change it back to use the menu files.

boceckts:
Comments have been addressed

tobiasdiez:
LGTM :+1:

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingCategory.java
```diff
@@ -23,7 +23,7 @@
 
 
     private KeyBindingCategory(String name) {
-        this.name = name;
+        this.name = name.replace("&", "");
```
tobiasdiez: I know Olly made you change this, but actually your previous fix was better.
The KeyBindingCategory is used as the header in the keybinding dialog and thus has absolutely nothing to do with the menu. That the strings exists in the menu is pure (kind of) coincidence. Thus it should be indeed `FILE(Localization.lang("File"))` in `KeyBindingCategory` instead of using `menuTitle` there.

boceckts: Well it's not really coincidence that the categories match the menu titles as I used the menu titles where the shortkey's action is located as the its category.

tobiasdiez: Yes I know, but one may add keyboard shortcuts for actions which are not present in the menu or reorder the categories a bit.

boceckts: Ok then I change it back to use the normal localization files.

boceckts:
Comments are addressed.

src/main/java/net/sf/jabref/gui/keyboard/KeyBindingCategory.java
koppor: > The KeyBindingCategory is used as the header in the keybinding dialog and
thus has absolutely nothing to do with the menu.

Sure? I thought, this was intentional to mirror the menu structure here...
And thus reuse the menu translations?

tobiasdiez:
Thanks

Issue #1920, opened 2016-09-05
Importer/Exporter for MODS format


tschechlovdev:
Since JabRef has an exporter for the MODS format, it would be nice to have also an importer for this format.
There is a XML schema available [here] (http://www.loc.gov/standards/mods/mods-3-6-announcement.html), so I think it would be the best to write the importer with a JAXB parser.
But then it could happen, that exporting in MODS format and importing in MODS format are not working correctly. So I think the exporter should also be rewritten with a JAXB parser.

If there are no arguments against it, I would take care of this.

koppor:
+1

boceckts:
I will close this as everything seems to be done.

Issue #1877, opened 2016-08-28
Keyword separator behaviour (again, from a groups perspective)


oscargus:
Let me rephrase this: the use of the keyword separator is ambiguous in JabRef. 

It is by default set to `, `. This leads to:
1. StringTokenizer splits on both `,` and ` `
2. when adding keywords `, ` is used as a separator

Clearly these two behaviors are just by luck sometimes compatible. Based on 1. it would be better to have `,` as the default.

In addition, the EntryUtil method for splitting on keywords used by BibEntry totally ignores this.

Relates to #705 

( @Siedlerchr 's comment was for the previous wording of this, although it is still relevant.)

I can in this context note that quite a bit of the refactoring I've made recently has been to pass this separator down to the correct methods. Except for the fact that I know that it is bad, I would almost suggest having the separator as a global static variable in model...

oscargus:
I might not have been in that PR, but it is not working as expected anyway.

oscargus:
OK, new info. It wasn't in that commit, but it seems like StringTokenizer is not handling spaces in the separator well...

Siedlerchr:
String Tokenizer ist really really legacy code.  Should either be String.split or scanner
http://stackoverflow.com/questions/691184/scanner-vs-stringtokenizer-vs-string-split

oscargus:
And in MedlineImporter the author decided to use `; ` as separator and ignore the preferences (the separator from the preferences is available in ImportFormatPreferences).

koppor:
Refs #1950

PR #1942, opened 2016-09-09
Mods importer

tschechlovdev:
Regarding: #1920 

Adds an importer for the MODS (xml) format. The importer is written with a JAXB parser.
Details on the format and the newest xml schema can be found [here] (http://www.loc.gov/standards/mods/).
I've used the newest schema (version 3.6).


Siedlerchr:
I just merged my XMLunit changes n #1944.  I think it can useful for you too. Allows checking/validating XML against a schema, too

src/main/java/net/sf/jabref/logic/importer/fileformat/ModsImporter.java
```diff
@@ -0,0 +1,427 @@
...
+        bibItems.add(entry);
+    }
+
+    private void parseModsGroup(Map<String, String> fields, List<Object> modsGroup, BibEntry entry) {
+        List<String> keywords = new ArrayList<>();
+        List<String> authors = new ArrayList<>();
+        List<String> notes = new ArrayList<>();
+
+        for (Object groupElement : modsGroup) {
+            if (groupElement instanceof AbstractDefinition) {
```
Siedlerchr: This whole if/else instanceof looks really ugly..
A possible solution would be to create a Generic method like:

`   public static <T> T unmarshal( String xsdSchema, String xmlDatei, Class<T> clss )`
Have a look at the link here, there is an example implementation
http://www.torsten-horn.de/techdocs/java-xml-jaxb.htm#JAXB-Unmarshaller

src/main/java/net/sf/jabref/logic/importer/fileformat/ModsImporter.java
```diff
@@ -0,0 +1,427 @@
...
+                StringPlusLanguage country = element.getValue();
+                putIfValueNotNull(fields, "country", country.getValue());
+            }
+        }
+    }
+
+    private void parseLocationAndUrl(Map<String, String> fields, LocationDefinition location) {
+        List<String> locations = new ArrayList<>();
+        List<PhysicalLocationDefinition> physicalLocations = location.getPhysicalLocation();
+        for (PhysicalLocationDefinition physicalLocation : physicalLocations) {
```
boceckts: Can't this be replaced by `location.getPhysicalLocation().stream().map(loc -> loc.getValue()).forEach(locations::add);`

src/main/java/net/sf/jabref/logic/importer/fileformat/ModsImporter.java
```diff
@@ -0,0 +1,427 @@
...
+            locations.add(physicalLocation.getValue());
+        }
+        putIfListIsNotEmpty(fields, locations, FieldName.LOCATION, ", ");
+
+        List<UrlDefinition> urlDefinitions = location.getUrl();
+        List<String> urls = new ArrayList<>();
+        for (UrlDefinition url : urlDefinitions) {
```
boceckts: Again try to use streams and lambdas

src/main/java/net/sf/jabref/logic/importer/fileformat/ModsImporter.java
```diff
@@ -0,0 +1,427 @@
...
+                    } else if (object instanceof ExtentDefinition) {
+                        ExtentDefinition extentDefinition = (ExtentDefinition) object;
+                        putPageInformation(extentDefinition, fields);
+                    }
+                }
+            } else if (groupElement instanceof TitleInfoDefinition) {
+                TitleInfoDefinition titleInfo = (TitleInfoDefinition) groupElement;
+                List<Object> titleOrSubTitleOrPartNumber = titleInfo.getTitleOrSubTitleOrPartNumber();
+                for (Object object : titleOrSubTitleOrPartNumber) {
+                    if (object instanceof JAXBElement) {
```
boceckts: A for loop and only one if statement on the elements seems a lot like filtering, have a look at `stream.filter()`.

src/main/java/net/sf/jabref/logic/importer/fileformat/ModsImporter.java
```diff
@@ -0,0 +1,427 @@
...
+            for (PlaceTermDefinition placeTerm : placeTerms) {
+                if ((placeTerm.getValue() != null) && !placeTerm.getType().equals(CodeOrText.CODE)) {
+                    places.add(placeTerm.getValue());
+                }
+            }
+            putIfListIsNotEmpty(fields, places, "place", ", ");
+        } else if (value instanceof DateDefinition) {
+            DateDefinition date = (DateDefinition) value;
+            switch (elementName) {
+            case "dateIssued":
```
boceckts: I think this should be indented.

src/main/java/net/sf/jabref/logic/importer/fileformat/ModsImporter.java
tschechlovdev: Sorry, but I don't really get how this can help here. I have to know explicit which class groupElement is an instance of, otherwise I'm not able to use the specific methods.
 
For unmarshalling I have the problem, that there can be two elements be the root element (ModsDefinition and ModsCollectionDefinition).

boceckts:
I think your code looks good but generally I would like to see more comments and try using more lambdas. :+1:

src/main/java/net/sf/jabref/logic/importer/fileformat/ModsImporter.java
tschechlovdev: Maybe you should looked at an old commit. Now there are two if's and I think using a stream here would make it much more unreadable.

src/main/java/net/sf/jabref/logic/importer/fileformat/ModsImporter.java
Siedlerchr: Yeah, I know. The thing is, my idea/thought was to create a single method which does the instanceof checking in a generic way, like  `static <T> getElement(Obj groupElement, Class<T> clazz) `
Have a look at the second anwer here:
http://stackoverflow.com/questions/14228637/why-instanceof-does-not-work-with-generic?noredirect=1&lq=1

Then you could simply do a call: like  `GenreDefinition def = getElement(modsElement, GenreDefintion.class)`
And of course you would need to check for isAssignable

src/main/java/net/sf/jabref/logic/importer/fileformat/ModsImporter.java
```diff
@@ -126,41 +128,49 @@ private void parseModsGroup(Map<String, String> fields, List<Object> modsGroup,
...
+            Optional<NameDefinition> name = getElement(groupElement, NameDefinition.class);
+            Optional<OriginInfoDefinition> originInfo = getElement(groupElement, OriginInfoDefinition.class);
+            Optional<RecordInfoDefinition> recordInfo = getElement(groupElement, RecordInfoDefinition.class);
+            Optional<NoteDefinition> note = getElement(groupElement, NoteDefinition.class);
+            Optional<RelatedItemDefinition> relatedItem = getElement(groupElement, RelatedItemDefinition.class);
+            Optional<SubjectDefinition> subject = getElement(groupElement, SubjectDefinition.class);
+            Optional<IdentifierDefinition> identifierDefinition = getElement(groupElement, IdentifierDefinition.class);
+            Optional<TitleInfoDefinition> titleInfo = getElement(groupElement, TitleInfoDefinition.class);
+
+            if (abstractDefinition.isPresent()) {
```
Siedlerchr: Like the idea with the Optionals :+1:  Keep on the good work!
You could even simplify that more: if you just use the `ifPresent `of Optional (accepts a lambda) you can omit those if/else if complety and the check of isPresent, because as the name already says it gets only executed if the value in the optional is present.

tschechlovdev: Done. Thanks for your feedback !

src/main/java/net/sf/jabref/logic/importer/fileformat/ModsImporter.java
```diff
@@ -220,31 +245,16 @@ private void parseGeographicInformation(Map<String, String> fields,
...
+        List<String> locations = locationDefinition.getPhysicalLocation().stream().map(location -> location.getValue())
+                .collect(Collectors.toList());
         putIfListIsNotEmpty(fields, locations, FieldName.LOCATION, ", ");
 
-        List<UrlDefinition> urlDefinitions = location.getUrl();
-        List<String> urls = new ArrayList<>();
-        for (UrlDefinition url : urlDefinitions) {
-            urls.add(url.getValue());
-        }
+        List<String> urls = locationDefinition.getUrl().stream().map(url -> url.getValue())
```
boceckts: Replace `url -> url.getValue()` by `UrlDefinition::getValue`

src/main/java/net/sf/jabref/logic/importer/fileformat/ModsImporter.java
```diff
@@ -254,16 +264,19 @@ private void parseRecordInfo(Map<String, String> fields, RecordInfoDefinition re
                 putIfValueNotNull(fields, "source", source.getValue());
             } else if (value instanceof LanguageDefinition) {
                 LanguageDefinition language = (LanguageDefinition) value;
-                List<String> languages = new ArrayList<>();
                 List<LanguageTermDefinition> languageTerms = language.getLanguageTerm();
-                for (LanguageTermDefinition languageTerm : languageTerms) {
-                    languages.add(languageTerm.getValue());
-                }
+                List<String> languages = languageTerms.stream().map(languageTerm -> languageTerm.getValue())
```
boceckts: Same as above.

src/main/java/net/sf/jabref/logic/importer/fileformat/ModsImporter.java
```diff
@@ -324,38 +338,35 @@ private void putPlaceOrPublisherOrDate(Map<String, String> fields, String elemen
...
-            List<String> places = new ArrayList<>();
             List<PlaceTermDefinition> placeTerms = place.getPlaceTerm();
-            for (PlaceTermDefinition placeTerm : placeTerms) {
-                if ((placeTerm.getValue() != null) && !placeTerm.getType().equals(CodeOrText.CODE)) {
-                    places.add(placeTerm.getValue());
-                }
-            }
-            putIfListIsNotEmpty(fields, places, "place", ", ");
+            List<String> places = placeTerms.stream().filter(placeTerm -> placeTerm.getValue() != null)
+                    .map(element -> element.getValue()).collect(Collectors.toList());
```
boceckts: `element -> element.getValue()` can also be replaced.

src/main/java/net/sf/jabref/logic/importer/fileformat/ModsImporter.java
```diff
@@ -324,38 +338,35 @@ private void putPlaceOrPublisherOrDate(Map<String, String> fields, String elemen
...
+            putIfListIsNotEmpty(fields, places, FieldName.ADDRESS, ", ");
         } else if (value instanceof DateDefinition) {
             DateDefinition date = (DateDefinition) value;
-            switch (elementName) {
-            case "dateIssued":
-                if ("yes".equals(date.getKeyDate()) && (date.getValue() != null)) {
+            if (date.getValue() != null) {
+                switch (elementName) {
+
+                case "dateIssued":
```
boceckts: Switch statements should be indented.

tschechlovdev:
I don't really get why travis is failing...

Siedlerchr:
The problem seems to with the integration tests, not related to your changes

CHANGELOG.md
```diff
@@ -35,6 +35,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - Implemented integrity check for `note` and `howpublished` field: Should have the first letter capitalized (BibTeX)
 - <kbd>Pos1</kbd> / <kbd>HOME</kbd> now select the first/last entry in the main table and the search result frame.
 - <kbd>UP</kbd> / <kbd>Down</kbd> / <kbd>Tab</kbd> / <kbd>shift+Tab</kbd> in the search result frame have now the same functionality as in the main  table.
+- importer for MODS format added
```
Siedlerchr: Uppercase `i`

src/main/java/net/sf/jabref/logic/importer/fileformat/ModsImporter.java
```diff
@@ -0,0 +1,463 @@
...
+ *
+ * Importer for the MODS format.<br>
+ * More details about the format can be found here <a href="http://www.loc.gov/standards/mods/">http://www.loc.gov/standards/mods/</a>. <br>
+ * The newest xml schema can also be found here <a href="www.loc.gov/standards/mods/mods-schemas.html.">www.loc.gov/standards/mods/mods-schemas.html.</a>.
+ *
+ */
+public class ModsImporter extends ImportFormat {
+
+    private static final Log LOGGER = LogFactory.getLog(ModsImporter.class);
+    private static final String KEYWORD_SEPARATOR = "; ";
```
Siedlerchr: @tobiasdiez Did we want to get the Keyword Separator from Preferences now? Refs #1877

src/main/java/net/sf/jabref/logic/importer/fileformat/ModsImporter.java
Siedlerchr: It's okay. No need for streams here.

src/main/java/net/sf/jabref/logic/importer/fileformat/ModsImporter.java
```diff
@@ -0,0 +1,463 @@
...
+                    foreName.add(namePart.getValue());
+                } else if ("family".equals(type) && (namePart.getValue() != null)) {
+                    familyName = namePart.getValue();
+                }
+            } else if ((value instanceof StringPlusLanguage) && "affiliation".equals(elementName)) {
+                StringPlusLanguage affiliation = (StringPlusLanguage) value;
+                putIfValueNotNull(fields, "affiliation", affiliation.getValue());
+            }
+        }
+        if (!foreName.isEmpty() && !"".equals(familyName)) {
```
Siedlerchr: Better use isEmpty() for familyname, too

src/main/java/net/sf/jabref/logic/importer/fileformat/ModsImporter.java
```diff
@@ -0,0 +1,463 @@
...
+            String author = familyName + ", " + Joiner.on(" ").join(foreName);
+            authors.add(author);
+        } else if (foreName.isEmpty() && !"".equals(familyName)) {
```
Siedlerchr: see above

src/main/java/net/sf/jabref/logic/importer/fileformat/ModsImporter.java
tschechlovdev: If I use the keyword separator from the preferences and I have keyword1 and keyword2, I've got ``keyword1groupKeywordSeparatorkeyword2". Is this how it is wanted?

tschechlovdev: I've used the keywordSeparator from the preferences now.

src/main/java/net/sf/jabref/logic/importer/fileformat/ModsImporter.java
tschechlovdev: Done.

src/main/java/net/sf/jabref/logic/importer/fileformat/ModsImporter.java
tschechlovdev: Done.

tschechlovdev:
Hm, I'm also wondering why codecov is not running the tests. The test coverage should be much higher...

Siedlerchr:
I dunno about CodeCo, from my point of view it's good! :+1: 
@boceckts  Please have a final look

src/main/java/net/sf/jabref/logic/importer/fileformat/ModsImporter.java
```diff
@@ -0,0 +1,465 @@
...
+
+    @Override
+    public ParserResult importDatabase(BufferedReader input) throws IOException {
+        Objects.requireNonNull(input);
+
+        List<BibEntry> bibItems = new ArrayList<>();
+
+        JAXBContext context;
+        try {
+            context = JAXBContext.newInstance("net.sf.jabref.logic.importer.fileformat.mods");
```
lenhard: Just a non-so-critical remark: `JAXBContext` is known to be an object that is costly to construct and doing this often might result in performance issues. Hence, you typically build it once and keep it. Could you extract `context` as an instance variable? Then, check on execution of `importDatabase` if it is null and if so initialize it.

tschechlovdev: Done.

Siedlerchr:
Okay, I merge it in now as all things are resolved :+1:  Regarding the failing tests we should investigate that in a new issue.

Issue #1178, opened 2016-04-12
Unicode Names not displayed correctly


Siedlerchr:
JabRef_windows-x64_3_3dev--snapshot--2016-04-12--master--1c5dcf1  on Win10

![aboutjabref](https://cloud.githubusercontent.com/assets/320228/14460208/89f54618-00bc-11e6-84af-4feb3e37ef17.png)

koppor:
Duplicate of  #682.

stefan-kolb:
Gradle issue is merged but we cannot see that it is working for us. Reinvestigate this for v3.7.

grimes2:
File AUTHORS is encoded in UTF-8 without BOM. All other files are encoded in ANSI. What is the reason for the different encoding?

koppor:
What are "all other files"? Language files are encoded in UTF-8. Source
code also in UTF-8. I'm currently not in ANSI encoding. Is it ISO ...-15?
Windows CP 8..? Or 7bit only?

grimes2:
"all other files" is for example CHANGELOG.md. Here are the screenshots in notepad++, that demonstrates the different encoding of the files:
![authors_in_notepad](https://cloud.githubusercontent.com/assets/15456836/18403735/e1a46858-76e6-11e6-909a-36f014da2ec8.png)
![changelog_in_notepad](https://cloud.githubusercontent.com/assets/15456836/18403741/ea0a4e54-76e6-11e6-9d04-d3ed7116c552.png)
The file were loaded from my local master.

Siedlerchr:
My changelog is in UTF8...like all other files


2016-09-09 23:43 GMT+02:00 Jürgen Lange <notifications@github.com>:

grimes2:
I've now cloned the upstream master again, to exclude that it is an problem introduced by me. Same result: AUTHORS still has another encoding than all the other files.
System: windows 10, Notepad++

Siedlerchr:
Fixed with #1954

koppor:
[...]

PR #1945, opened 2016-09-09
Add filteringCharset = 'UTF-8' to gradle

Siedlerchr:
When I read it correct in the doc, this should help fixing the Unicode problems in #1178 
I tested locally and it seemed working for me. 

>It's a good practice to specify the charset when reading and writing the file, using the filteringCharset property. If not specified, the JVM default charset is used, which might not match with the actual charset of the files to filter, and might be different from one machine to another. 

https://docs.gradle.org/current/userguide/working_with_files.html#sec:filtering_files

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -47,7 +47,7 @@ install4j {
 }
 
 task wrapper(type: Wrapper) {
-    gradleVersion = '2.14.1'
+    gradleVersion = '3.0'
 }
 
 
@@ -62,8 +62,8 @@ configurations {
     databaseTestCompile.extendsFrom testCompile
     databaseTestRuntime.extendsFrom testRuntime
 
-	integrationTestCompile.extendsFrom testCompile
-	integrationTestRuntime.extendsFrom testRuntime
+    integrationTestCompile.extendsFrom testCompile
+    integrationTestRuntime.extendsFrom testRuntime
 }
 
 dependencies {
@@ -136,7 +136,7 @@ sourceSets {
             runtimeClasspath += main.output + test.output
         }
     }
-	integrationTest {
+    integrationTest {
         java {
             compileClasspath += main.output + test.output
             runtimeClasspath += main.output + test.output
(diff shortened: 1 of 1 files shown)


grimes2:
On my system the encoding is now working.

encoding

I changed
filteringCharset = 'UTF-8'
to
filteringCharset = 'ISO-8859-1'

JabRef 3.7dev
windows 10 10.0 amd64 
Java 1.8.0_101

koppor:
Checking @Siedlerchr's changes, it works. @grimes2 I don't get, why ISO-8859-1 works at your side and UTF-8 doesn't. I just merge this in and we can continue evaluating afterwards.

build.gradle
```diff
@@ -47,7 +47,7 @@ install4j {
 }
 
 task wrapper(type: Wrapper) {
-    gradleVersion = '2.14.1'
+    gradleVersion = '3.0'
```
koppor: This is not necessary anymore. See https://gradle.org/blog/gradle-3-0-m1-unleash-the-daemon/

`./gradlew wrapper --gradle-version 3.0` has the same effect as `./gradlew wrapper` with that config.

I'll just change it directly in the master branch.

grimes2:
It seems that this bug is not fixed.
wrong_encoding

JabRef 3.7dev--snapshot--2016-09-11--master--f5ed5ac
windows 10 10.0 amd64 
Java 1.8.0_101

Siedlerchr:
Yes, you are right. Windows 7 x64 the same.

Siedlerchr:
I think I found the reason for it.  Property.load assumes automatically ISO... charset.
I try with wrapping it in an `InputStreamReader `
http://stackoverflow.com/a/17852323

koppor:
Strange, that it was not fixed at your side, here on Win 10, 64bit, the umlauts were displayed correctly.

Siedlerchr:
@koppor  Nope, if you look closely in the middle you have someone named `Krzysztof A. Kościuszkiewicz`
 where the s -Char is not displayed correctly, only a question mark => Latin 1 (supports only a subset of chars). (I recently read an article about a system in the Swiss Government which had the same problems...)

PR #1947, opened 2016-09-10
Redesigned about dialog in JavaFX

boceckts:
As discussed in #968 the about dialog is still not satisfying and therefore I redesigned it. The new design is a more flat like design with all important information in one place. The version still clickable and copies the version information to the clipboard and the logo is a hyperlink to www.jabref.org.
I also changed the copied version string to include JabRef, Java and OS version so it is the same as in the master branch.

screen shot 2016-09-10 at 14 59 58


src/main/java/net/sf/jabref/gui/help/AboutDialogViewModel.java
```diff
@@ -21,74 +6,54 @@
...
 import org.apache.commons.logging.LogFactory;
 
 public class AboutDialogViewModel {
 
+    private final String HOMEPAGE = "http://www.jabref.org";
+    private final String DONATION = "http://www.jabref.org/#donations";
+    private final String LIBRARIES = "https://github.com/JabRef/jabref/blob/master/external-libraries.txt";
+    private final String GITHUB = "https://github.com/JabRef/jabref";
+    private final String CHANGELOG = "https://github.com/JabRef/jabref/blob/master/CHANGELOG.md";
+    private final String LICENSE = "https://opensource.org/licenses/mit-license.php";
```
tobiasdiez: Link to https://github.com/JabRef/jabref/blob/master/LICENSE.md

tobiasdiez:
LGTM, just change the link to the license and I think this can be merged.
@koppor please check that the new library is included in the correct way. Thanks

koppor:
Please merge `javafx` and add new screenshot.

external-libraries.txt
```diff
@@ -164,4 +164,9 @@ Project: afterburner.fx
 URL:     https://github.com/AdamBien/afterburner.fx
 License: Apache 2.0
 
+Id:      de.codecentric.centerdevice
+Project: javafxsvg
+URL:     https://github.com/codecentric/javafxsvg
+License: BSD 3-Clause
```
koppor: :+1:

boceckts:
@koppor Merged javafx, why do you want a new screenshot? The layout hasn't changed, only difference is that there is a small JabRef icon in the title bar of the dialog.

src/main/java/net/sf/jabref/gui/help/AboutDialogViewModel.java
```diff
@@ -21,74 +6,54 @@
...
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
 public class AboutDialogViewModel {
 
+    private final String HOMEPAGE = "http://www.jabref.org";
+    private final String DONATION = "http://www.jabref.org/#donations";
+    private final String LIBRARIES = "https://github.com/JabRef/jabref/blob/master/external-libraries.txt";
+    private final String GITHUB = "https://github.com/JabRef/jabref";
+    private final String CHANGELOG = "https://github.com/JabRef/jabref/blob/master/CHANGELOG.md";
```
koppor: Can we add a deep link here?

I would suggest https://github.com/koppor/keep-a-changelog/blob/master/CHANGELOG.md#unreleased for the development versions and add a release TODO to replace this by something like `https://github.com/JabRef/jabref/blob/master/CHANGELOG.md#36---2016-08-26`

koppor:
@boceckts I want to double check that the fix of https://github.com/JabRef/jabref/pull/1954 also works here.

koppor:
The about dialog is not centered on the screen. Does anyone else experience such an issue?

boceckts:
I think it is centered on the main frame, not the screen.

PR #1948, opened 2016-09-10
Fixed some warnings in Eclipse

Siedlerchr:
Some code cleanups


Siedlerchr:
Failing tests are due to ACS timeouts.

src/main/java/net/sf/jabref/gui/ClipBoardManager.java
```diff
@@ -66,6 +66,7 @@ public String getClipboardContents() {
         return result;
     }
 
+    @SuppressWarnings("unchecked")
```
koppor: Can this be moved to a place where the unchecked conversion really occurs? If not, please add a comment why the suppression is needed at the whole method.

src/main/java/net/sf/jabref/gui/fieldeditors/FileListEditorTransferHandler.java
```diff
@@ -70,6 +70,7 @@ public void exportToClipboard(JComponent comp, Clipboard clip, int action) {
         }
     }
 
+    @SuppressWarnings("unchecked")
```
koppor: See above

src/main/java/net/sf/jabref/gui/importer/fetcher/EntryFetcher.java
```diff
@@ -8,6 +8,7 @@
 import net.sf.jabref.logic.importer.OutputPrinter;
 
 /**
+ * <b>use SearchBasedFetcher instead</b> <br>
```
koppor: Please use the [@deprecated](http://stackoverflow.com/a/24605999/873282) annotation.

src/main/java/net/sf/jabref/gui/importer/fetcher/EntryFetcher.java
```diff
@@ -18,8 +19,11 @@
  *
  * A Fetcher should not execute any GUI Operations, because it might be run in
  * headless mode, but rather use the OutputPrinter for talking to the user.
+ *
```
koppor: Why these new empty lines?

src/main/java/net/sf/jabref/model/entry/BibEntry.java
```diff
@@ -38,14 +38,25 @@
...
+    /**
+     *  use dedicated methods like get/set/clearCiteKey
+     */
+    @Deprecated
     public static final String KEY_FIELD = "bibtexkey";
     protected static final String ID_FIELD = "id";
-    @Deprecated // use constructor without type
+
+    /**
+     * use constructor without type
```
koppor: Please add `@deprecated` Javaoc

src/main/java/net/sf/jabref/model/entry/BibEntry.java
```diff
@@ -38,14 +38,25 @@
     private static final Log LOGGER = LogFactory.getLog(BibEntry.class);
 
     // All these fields should be private or protected
-    @Deprecated // use get/setType
+    /**
+     * use get/setType
```
koppor: Please add `@deprecated` Javaoc

src/main/java/net/sf/jabref/model/entry/BibEntry.java
```diff
@@ -38,14 +38,25 @@
...
+     */
+    @Deprecated
     public static final String TYPE_HEADER = "entrytype";
     @Deprecated
```
koppor: Please add empty line inbetween.

src/main/java/net/sf/jabref/model/EntryTypes.java
```diff
@@ -95,8 +98,11 @@ public void removeType(String name) {
 
     }
 
-    public static final InternalEntryTypes BIBTEX = new InternalEntryTypes(BibtexEntryTypes.MISC, BibtexEntryTypes.ALL, IEEETranEntryTypes.ALL);
-    public static final InternalEntryTypes BIBLATEX = new InternalEntryTypes(BibLatexEntryTypes.MISC, BibLatexEntryTypes.ALL);
+
+    public static final InternalEntryTypes BIBTEX = new InternalEntryTypes(BibtexEntryTypes.MISC,
+            Arrays.asList(BibtexEntryTypes.ALL, IEEETranEntryTypes.ALL));
```
koppor: Which warning does "asList" remove? I might be, that `List<EntryTypes>...` has to be changed to be changed to `EntryTypes...` Or the `...` simply be removed?

src/main/java/net/sf/jabref/gui/importer/fetcher/EntryFetcher.java
tobiasdiez: Do we really need to write proper javadoc documentation for deprecated fields or methods? I would rather invest my time in removing the old code. For me the previous syntax with method annotation plus explaining comment was satisfying enough.

Siedlerchr: The comment as  javadoc is better, because when you see a method call only marked as deprecated you usually don't know what to use as replacement. And when you do openDecleration and hover over the type, the javadoc shows directly the alternative

src/main/java/net/sf/jabref/model/EntryTypes.java
Siedlerchr: @koppor Some possible Heap pollution with the usage of the vargars parameter 
http://stackoverflow.com/questions/12462079/potential-heap-pollution-via-varargs-parameter

I tried removing the ... first but then I noticed that you have a List<List<EntryType>>
Because, Types.ALL returns already a List. so you have a list containig a List of EntryType

PR #1950, opened 2016-09-11
Towards hierarchical keywords

tobiasdiez:
This PR is the first step towards supporting hierarchical keywords #628. 

- [x] Refactor the keyword-related code to use new classes `KeywordList` and `Keyword` 
- [x] Unified keyword delimiter to be a single character and in this way fixed the issues mentioned in #705 and #1877
- [ ] Move keyword delimiter from preferences to metadata (database properties) -> new PR
- [ ] Add hierarchical delimiter and change parsing logic in `KeywordList` accordingly -> new PR
- [ ] Change UI to support hierarchical keywords -> new PR

Note: I also moved the StringUtil class to model (in some sense it is JabRef's own String class) and removed EntryUtil (only had methods related to strings (-> StringUtil) or keywords (-> KeywordList) ).


src/test/java/net/sf/jabref/logic/importer/fetcher/GvkParserTest.java
```diff
@@ -0,0 +1,75 @@
...
+import javax.xml.parsers.ParserConfigurationException;
+
+import net.sf.jabref.logic.bibtex.BibEntryAssert;
+import net.sf.jabref.model.entry.BibEntry;
+
+import org.junit.Assert;
+import org.junit.Test;
+import org.xml.sax.SAXException;
+
+public class GvkParserTest {
```
lenhard: I don't quite get why this class is introduced in this PR.

src/test/java/net/sf/jabref/model/entry/KeywordListTest.java
```diff
@@ -0,0 +1,66 @@
...
+
+import java.util.Arrays;
+import java.util.Collections;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+public class KeywordListTest {
```
lenhard: Please add tests that document the behavior in case of duplicate keywords.

src/main/java/net/sf/jabref/BibDatabaseContext.java
```diff
@@ -65,7 +65,7 @@ public BibDatabaseContext(BibDatabase database, MetaData metaData, File file) {
         this(database, metaData, file, new Defaults());
     }
 
-    public BibDatabaseContext(Defaults defaults, DatabaseLocation location, String keywordSeparator) {
+    public BibDatabaseContext(Defaults defaults, DatabaseLocation location, Character keywordSeparator) {
```
tobiasdiez: There were some issues related to strings as keyword separators. For example, the default is `, ` (with space) but some places tried to split strings only at the comma (i.e. recognize `apple,orange`). The space was only used to reformat the list as `apple, orange`. So I decided to force that only one character can be used as a separator and the final space is always adapted upon writing.

src/integrationTest/java/net/sf/jabref/gui/ParameterizedMenuNewEntryTest.java
```diff
@@ -3,7 +3,7 @@
 import java.util.Arrays;
 import java.util.Collection;
 
-import net.sf.jabref.model.entry.EntryUtil;
+import net.sf.jabref.model.strings.StringUtil;
```
tobiasdiez: Yes we now have a ModelStringUtil class, which already moved some string-related methods to model. I would propose to merge everything together: logic.StringUtil, model.ModelStringUtil and model.EntryUtil, they all deal with the same thing. I will create a separate PR with this change (because it makes merging easier).

src/main/java/net/sf/jabref/model/entry/KeywordList.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.StringTokenizer;
+import java.util.stream.Collectors;
+
+import net.sf.jabref.model.strings.StringUtil;
+
+/**
+ * Represents a list of keyword chains.
+ * For example, "Type > A, Type > B, Something else".
+ */
+public class KeywordList extends ArrayList<Keyword> {
```
tobiasdiez: Composition over inheritance, noted!

I wasn't sure how to handle duplicates. There was some code which tried to replace a keyword exactly at the same position: `a, b, c` -> `a, replace, c`. This was not easily possible with Set, but just one line with List. Not sure that this was the right decision through. What is your opinion?

src/main/java/net/sf/jabref/BibDatabaseContext.java
lenhard: Ok, sounds reasonable. Let's hope that no one ever wants to use some weird multi-keyword separator.

src/integrationTest/java/net/sf/jabref/gui/ParameterizedMenuNewEntryTest.java
lenhard: Ok, I understand. Do this in a separate PR, but please do not forget it :)

src/main/java/net/sf/jabref/model/entry/KeywordList.java
lenhard: Tough question, really. Personally, I cannot come up with a reason to have duplicate keywords, but users are weird, as we all know :) It is probably best if we just keep things as they were for the moment. Since the keywords were prior stored as a String, it was certainly possible to add duplicates. So let us keep this for now. My fear is that it may result in random bugs, where processing stops, because a keyword is found and another occurrence of the keyword is still in the list.

As a side note: You do not have to use a Set to avoid duplicates. You can also use a List internally and purge duplicates at the level of `KeywordList`.

PR #1960, opened 2016-09-12
Hotfix: Fix loop when pulling changes when current selected field has changed

bartsch-dev:
@obraliar and I fixed a critical Bug in the shared Database (broken in v3.6):

### Steps to Reproduce:
- open 2 JabRef instances with shared database
- edit a field of an entry in the first instance and push the change (eg. selecting another field)
- select the same entry/field which was edited int the second instance
- pull changes in the second instance
- second JabRef loops indefinitely


Siedlerchr:
Code LGTM :+1:

PR #1961, opened 2016-09-12
Fix when editing a bibEntry field the searchbar would be focused

bartsch-dev:
Fix for #1549

### Steps to reproduce:
- Select a bibentry
- Edit a field
- save the change (eg. click outside the field)
- the searchbar would be focused

### DIFF
--- a/src/main/java/net/sf/jabref/gui/search/GlobalSearchBar.java
+++ b/src/main/java/net/sf/jabref/gui/search/GlobalSearchBar.java
@@ -65,7 +65,11 @@ public class GlobalSearchBar extends JPanel {
     private SearchResultFrame searchResultFrame;
 
     private SearchDisplayMode searchDisplayMode;
-    private boolean switchedDatabase;
+
+    /**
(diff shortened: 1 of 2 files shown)


src/main/java/net/sf/jabref/gui/search/GlobalSearchBar.java
```diff
@@ -370,15 +374,19 @@ public void setSearchResultFrame(SearchResultFrame searchResultFrame) {
...
         }
 
-        this.switchedDatabase = switchedDatabase;
+        dontSelectSearchBar(dontSelectSearchBar);
         searchField.setText(searchTerm);
         // to hinder the autocomplete window to popup
         autoCompleteSupport.setVisible(false);
     }
 
+    public void dontSelectSearchBar(boolean dontSelectSearchBar) {
```
boceckts: Please rename either the method or the parameter.

bartsch-dev: already fixed

boceckts:
LGTM :+1:

Issue #1959, opened 2016-09-12
Improve JavaFX about dialog


koppor:
Follow up of #1947

- [x] clicked link should stay blue (see screenshot below)
- [x] is it possible to link https://github.com/koppor/keep-a-changelog/blob/master/CHANGELOG.md#unreleased instead of CHANGELOG.md. During a release, we will replace that by something like `https://github.com/JabRef/jabref/blob/master/CHANGELOG.md#36---2016-08-26`
- [x] Spell "Used libraries" instead of "Used Libraries".
- [x] Prefix "MIT License" by "License:" and remove the year or change to " - Copyright (c) 2003 - 2016"
- [x] Add a dot (`.`) after "get involved" as this is a complete sentence and the question before ends with a question mark.

![grabbed_20160912-133251](https://cloud.githubusercontent.com/assets/1366654/18434450/7779b01a-78ed-11e6-825c-eddd28d46b7b.png)

I know, I might be picky, but the about dialog is seen by many users and should be good.

grimes2:
For authors and developers it is important to know the actual snapshot version (latest commit hash) of JabRef.
![about](https://cloud.githubusercontent.com/assets/15456836/18435241/50a98226-78f2-11e6-862c-a8310b796d82.png)
Here the version is truncated. An idea is to add the complete version as a tooltip.

koppor:
Why not just simply in the next line with normal font size?
"dev info: snapshot--2017..."

The tool tip is too hidden for me.

boceckts:
@koppor discussion moved to #1962 with new screenshots and new way to find the link to the changelog of the used version.

PR #1962, opened 2016-09-12
Minor improvements to the JavaFX about dialog

boceckts:
Follow up of #1959.
@bartsch-dev had a great idea to use his `Version` class to get the changelog url since it already provides a method for it. I only adjusted it slightly so dev versions will be redirected to https://github.com/JabRef/jabref/blob/master/CHANGELOG.md#unreleased. Links will also stay gighlighted after they have been visited.

normal version

for snapshot versions a new label will be added for the complete version string.
snapshot


tobiasdiez:
I think the "License: " prefix is not really necessary. "MIT license" is already self-explanatory.  @koppor do you really want this?

boceckts:
@tobiasdiez I get the version string from `Globals.BUILD_INFO.getVersion();` do you only want to remove "snapshot" in the about dialog or in general? For example copy version also calls the same method.

Siedlerchr:
Like the idea with the VersioInfo solution! :+1: 
Is commit included?

grimes2:
"snapshot" is part of the builds delivered by builds.jabref.org.
For example:

> JabRef 3.7dev--snapshot--2016-09-12--javafx--bd3dbe8
> windows 10 10.0 amd64 
> Java 1.8.0_101

The complete version string should be shown somewhere in the About dialog.  Does the text wrap work also for this long string without whitespace?

boceckts:
@grimes2  yes it will automatically be broken down to fit multiple lines.

longversion

tobiasdiez:
Ok, leave it like that (with snapshot)...it is only for the dev version, so no reason to be pixel perfect.

grimes2:
That's fine. The normal versions JabRef 3.7 or JabRef 3.7dev will anyway fit in one line. And for the authors all informations are now delivered. Thank you!

tobiasdiez:
I still have the opinion that "License:" should be removed, but apart from that 👍 for merge. @koppor your call.

boceckts:
After talking to @koppor I changed minor things and how the snapshot versions will be displayed, I updated the screenshots in the description of this pr accordingly.

koppor:
The about dialog is not centered on the JabRef window. See attached screenshot.

JabRef 3.7dev--snapshot--2016-10-06--javafx--5a4e79a
windows 10 10.0 amd64 
Java 1.8.0_101

grabbed_20161006-181713

boceckts:
Hmm I just tried it with the jar available at http://builds.jabref.org/javafx/ and it worked perfectly for me :thinking:

aboutdialogcenter

JabRef 3.7dev--snapshot--2016-09-26--javafx--6134490
windows 8.1 6.3 amd64 
Java 1.8.0_102

koppor:
Works on a single screen. Have to connect to an external screen and check again. I have the feeling that JavaFX does not work well with two screens having a different magnification.

boceckts:
This was actually tested on a second screen, that's why there are no icons. Your build is also a newer one, maybe this is the reason.

PR #1964, opened 2016-09-12
First half of refactoring special fields

oscargus:
`SpecialField` is now free of gui-code and moved to logic. Problem is that `SpecialFieldValue` still relies on gui-code and to get rid of it, the code will be quite ugly... I do not really have any good idea how to do that.

It was discussed moving it to model. However, there will be localization issues then, so I moved it to logic.


oscargus:
`SpecialFieldDatabaseChangeListener` should be easy to move if it wasn't for the `Globals`. This is one of the situations where I do not really know how to handle it. Using a static variable to say if keyword synchronization is on may be one alternative, but it doesn't really seem like a good way to do it.

tobiasdiez:
idea: use an enum `SpecialFields` (with priorty, printed, ... and accepting the fieldname and a list of possible field values [prio1, prio2,...]) and use this in the logic code (in the end this is all you need in logic, right?). Then all the other code could be transferred to gui (especially SpecialFieldValue and thus also SpecialField has so many gui-related things like icons, text and menu descriptions).

koppor:
This refs https://github.com/koppor/jabref/issues/84. @oscargus Now that you are familiar with the special fields code, could you comment on that?

koppor:
Does undo/redo still work? :)

oscargus:
It works exactly as before, yes. Unless I missed something. The code I removed (with comments) wasn't really used. It just filled the compound but didn't add it.

koppor:
As far as I remember, this should be then good to go and in a second PR, https://github.com/JabRef/jabref/pull/1964#issuecomment-246444967 should be thought of?

oscargus:
Maybe, some parts are good to go, while others depends on the final solution. I do not fully follow the `enum` suggestion, but I'm not very good at more advanced use of `enum`. More specifically how and where to map `enum` to `Icon` in a way that makes sense.

src/main/java/net/sf/jabref/gui/actions/ManageKeywordsAction.java
```diff
@@ -313,7 +314,9 @@ private NamedCompound updateKeywords(List<BibEntry> entries, Set<String> keyword
             }
 
             if (Globals.prefs.isKeywordSyncEnabled()) {
-                SpecialFieldsUtils.syncSpecialFieldsFromKeywords(entry, ce);
+                List<FieldChange> changes = new ArrayList<>();
```
lenhard: Could you maybe do a little refactoring here? The `changes` list is really a return argument. So instead of having the caller of `syncSpecialFieldsFromKeywords` create the list and pass it to the method, it would be better if the method created the list itself and just returns it to the caller. Return arguments are common in C, but considered really bad style in Java. 

This change would propagate to the other `synchSpecialFields` methods and require some rewriting. If I did not miss anything, there is are only internal calls in `SpecialFieldsUtil` that might have a list that has actual content inserted into these methods. Even in this case, it would be better to create a new list as return object internally and copy the contents of the list inserted as argument is needed.

This is really not the fault of this PR, but the bad design of special fields which should be fixed.

src/main/java/net/sf/jabref/gui/specialfields/SpecialFieldAction.java
```diff
@@ -46,17 +53,20 @@ public void action() {
                 return;
             }
             NamedCompound ce = new NamedCompound(undoText);
+            List<FieldChange> changes = new ArrayList<>();
             for (BibEntry be : bes) {
                 // if (value==null) and then call nullField has been omitted as updatefield also handles value==null
-                SpecialFieldsUtils.updateField(specialField, value, be, ce, nullFieldIfValueIsTheSame);
+                SpecialFieldsUtils.updateField(specialField, value, be, changes, nullFieldIfValueIsTheSame,
```
lenhard: As above, the return structure of the `updateField` method should be changed.

src/main/java/net/sf/jabref/logic/specialfields/SpecialFieldsUtils.java
```diff
@@ -1,20 +1,22 @@
...
-import net.sf.jabref.preferences.JabRefPreferences;
+import net.sf.jabref.specialfields.Printed;
+import net.sf.jabref.specialfields.Priority;
+import net.sf.jabref.specialfields.Quality;
+import net.sf.jabref.specialfields.Rank;
+import net.sf.jabref.specialfields.ReadStatus;
+import net.sf.jabref.specialfields.Relevance;
 
 
 public class SpecialFieldsUtils {
```
lenhard: Many changes to the methods here, if you implement the refactoring I suggested above.

src/main/java/net/sf/jabref/logic/specialfields/SpecialField.java
```diff
@@ -51,8 +50,6 @@ protected void setValues(List<SpecialFieldValue> values) {
```
lenhard: What @tobiasdiez means with the enum solution is to turn this class into an enum and merge all of the different special field classes (`Priority`, `Printed`, etc.) into this one. This seems reasonable, since all special fields are singletons anyway (you always access them via `getInstance`). You loose the inheritance relation, but apart from that enums can also have arbitrary methods, constructors, just as normal classes. Here is some explanation of how this works: https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html

The problem that this raises, however, is that all special field classes would move into `SpecialField` and, thereby, also `SpecialFieldValue` would be needed there. To enable this, we have to move the gui stuff out of `SpecialFieldValue` and that is at least the `Icon` in there. In the end, this class really needs to be broken up into the parts that are gui independent and the parts that are gui dependent. I understand that this is not easy.

lenhard:
@oscargus: If you feel uncomfortable with implementing the enum solution, I might be prepared to take over. Or we merge this and I submit a new PR. 

I like the idea of improving the special fields, but I am afraid of the miseries that hide in there that I have overlooked so far...

lenhard:
@oscargus How is the status here? Would it be possible to merge master into this branch, so we can progress? Most of the things should be easy to merge, but I fear that `SpecialFieldUtils` will not be entirely easy, because of the recent changes to keywords.

oscargus:
Ahh, sorry for the lack of attention. Been busy with other things.

Not sure what the aim of the long term solution of the special fields are anymore. Just noticed that a lot of things got made deprecated without an obvious way to a solution.

@lenhard Feel free to take over.

lenhard:
@oscargus No problem! I don't know about the deprecation, but I guess special fields will stay, so refactoring them still makes sense.

I'll take over and try to complete the refactoring. I will do that in a new branch, as I cannot write on yours. I just can't promise as to when this will happen, so we leave the PR open for now.

As a side note: How about creating future PRs directly as branches of jabref instead of on your personal fork? That way it would be easier for members of @JabRef/developers to take over.

oscargus:
I sometimes do that (when I want a build that can be tested), but can
switch to it more continuously.

Thanks!

lenhard:
After trying to merge master into this for quite some time, I think it might better to just restart from current master and redo the whole thing. With recent changes in master regarding the keyword-stuff, merging `SpecialFieldUtils` is practically a rewrite and that is way less complicated if you do not do it as part of a merge.

lenhard:
This PR can be closed, the follow up is #2162.

PR #1965, opened 2016-09-12
Removed last Globals in logic tests and introduced architecture test

oscargus:

### DIFF
--- a/src/test/java/net/sf/jabref/logic/importer/fetcher/IsbnFetcherTest.java
+++ b/src/test/java/net/sf/jabref/logic/importer/fetcher/IsbnFetcherTest.java
@@ -3,10 +3,10 @@ package net.sf.jabref.logic.importer.fetcher;
 import java.io.IOException;
 import java.util.Optional;
 
-import net.sf.jabref.Globals;
 import net.sf.jabref.logic.importer.FetcherException;
 import net.sf.jabref.model.entry.BibEntry;
 import net.sf.jabref.model.entry.BibLatexEntryTypes;
+import net.sf.jabref.preferences.JabRefPreferences;
 
 import org.junit.Before;
 import org.junit.Test;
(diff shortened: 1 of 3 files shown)


Siedlerchr:
Good work! What does the architecture test enforce? That no more globals are used in logic?

src/test/java/net/sf/jabref/logic/importer/fetcher/IsbnFetcherTest.java
```diff
@@ -3,10 +3,10 @@
 import java.io.IOException;
 import java.util.Optional;
 
-import net.sf.jabref.Globals;
 import net.sf.jabref.logic.importer.FetcherException;
 import net.sf.jabref.model.entry.BibEntry;
 import net.sf.jabref.model.entry.BibLatexEntryTypes;
+import net.sf.jabref.preferences.JabRefPreferences;
```
koppor: According to https://github.com/JabRef/jabref/wiki/High-Level-Documentation#package-structure, the logic should not reference the preferences.

At least, this PR is a huge step in the direction of enforcing the idea of https://github.com/JabRef/jabref/wiki/High-Level-Documentation#package-structure.

koppor:
Even though another nasty dependency is introduced, this PR is a huge step towards a good architecture.

oscargus:
@Siedlerchr yes, no globals in test or model.

@koppor there are still some dependencies on JabRefPreferences in logic (not just here). Primarily in test, but also in the actual code. As I recall it, Globals.prefs is even injected in some methods, so having this dependency in test is quite OK at the moment...

oscargus:
There are currently nine imports in `logic` (most related to preferences classes, but not all) and 42 in `test/logic` of `JabRefPreferences`.

oscargus:
Related (to solve parts of it): #1968 #1969

PR #1972, opened 2016-09-13
Removed duplicate translation keys causing failing checks

boceckts:
This should fix the failing ensureNoDuplicates test method in LocalizationConsistencyTest class.

[ ] fix failing integration tests

### DIFF
--- a/src/main/resources/l10n/JabRef_es.properties
+++ b/src/main/resources/l10n/JabRef_es.properties
@@ -206,8 +206,6 @@ Do_not_wrap_the_following_fields_when_saving=No_envolver_los_siguientes_campos_a
 Do_not_write_the_following_fields_to_XMP_Metadata\:=No_escribir_los_campos_a_continuación_a_los_metadatos_XMP
 Do_you_want_JabRef_to_do_the_following_operations?=¿Quiere_que_JabRef_efectúe_las_siguientes_operaciones?
 Down=Abajo
-Download=Descargar
-Download_completed=Descarga_completa
 Download_file=Descargar_archivo
 Downloading...=Descargando...
 
@@ -253,7 +251,6 @@ Error_setting_field=Error_estableciendo_el_campo
 
 Error_while_writing=Error_al_escribir
 Error_writing_to_%0_file(s).=Error_escribiendo_%0_archivo(s).
-Exceptions=Excepciones
 '%0'_exists._Overwrite_file?='%0'_existe._¿Sobreescribir?
 Overwrite_file?=¿Sobreescribir?
 Expand_subtree=Expandir_subárbol
@@ -449,7 +446,6 @@ no_database_generated=No_se_generó_base_de_datos
 No_entries_found._Please_make_sure_you_are_using_the_correct_import_filter.=No_se_han_encontrado_entradas._Asegúrese_de_que_está_usando_el_filtro_de_importación_correcto.
 No_entries_found_for_the_search_string_'%0'=No_se_han_encontrado_entdas_para_la_cadena_de_búsqueda_'%0'
 No_entries_imported.=No_se_han_importado_entradas.
-No_exceptions_have_occurred.=No_han_ocurrido_excepciones.
 No_files_found.=No_se_han_encontrado_archivos.
 No_GUI._Only_process_command_line_options.=Sin_GUI._Procesar_sólo_opciones_de_la_línea_de_comandos.
 No_journal_names_could_be_abbreviated.=No_se_pudieron_abreviar_nombres_de_revistas.
@@ -485,7 +481,6 @@ Operation_not_supported=Operación_no_soportada
 Optional_fields=Campos_opcionales
 Options=Opciones
 or=o
-Output=Salida
 Output_or_export_file=Sacar_o_exportar_archivo,
 Override=Ignorar
 Override_default_file_directories=Ignorar_carpetas_de_archivo_por_defecto
@@ -522,7 +517,6 @@ Primary_sort_criterion=Criterio_de_ordenación_primario
 Problem_with_parsing_entry=Problemas_analizando_entradas
 
 Processing_%0=Procesando_%0
-Program_output=Salida_del_programa
 Pull_changes_from_shared_database=
 Pushed_citations_to_%0=Se_han_enviado_las_citas_a_%0
 Quit_JabRef=Salir_de_JabRef
@@ -982,9 +976,6 @@ Doing_a_cleanup_for_%0_entries...=Haciendo_la_limpieza_de_%0_entradas
 No_entry_needed_a_clean_up=Ninguna_entrada_necesitó_de_limpieza
 One_entry_needed_a_clean_up=Sólo_una_entrada_necesitó_limpieza
 %0_entries_needed_a_clean_up=%0_entradas_necesitarion_limpieza
-Error_downloading_file_'%0'=Error_descargando_el_archivo_'%0'
-Download_failed=Falló_la_descarga
-%0_databases_will_be_imported=%0_bases_de_datos_serán_importadas
 Remove_selected=Eliminar_seleccionados
 Group_tree_could_not_be_parsed._If_you_save_the_BibTeX_database,_all_groups_will_be_lost.=El_arbol_de_grupo_no_puede_ser_examinado._Si_guarda_la_base_de_datos_BibTeX,_todos_los_grupos_se_perderán.
 Attach_file=Adjuntar_archivo
@@ -1087,7 +1078,6 @@ Please_enter_a_search_string=Introduzca_una_cadena_de_búsqueda,_por_favor
 Please_open_or_start_a_new_database_before_searching=Abra_o_cree_una_nueva_base_de_datos_antes_de_buscar,_por_favor
 An_error_occurred_while_fetching_from_ADS_(%0)\:=Ocurrió_un_error_al_recuperar_desde_ADS_(%0)\:
 An_error_occurred_while_parsing_abstract=Ocurrió_un_error_mientras_se_analizaba_el_resumen
-Log=Registro
 Canceled_merging_entries=Cancelar_fusionado_de_entradas
 Format_units_by_adding_non-breaking_separators_and_keeping_the_correct_case_on_search=Formatear_unidades_añadiendo_separadores_no_divisores_y_manteniendo_mayúsculas/minúsculas_correctamente_en_la_búsqueda
 Merge_entries=Fusionar_entradas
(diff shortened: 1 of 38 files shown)


src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsController.java
```diff
@@ -0,0 +1,7 @@
+package net.sf.jabref.gui.journals;
+
+/**
+ * Created by Tobias on 13.09.2016.
+ */
+public class ManageJournalAbbreviationsController {
```
koppor: Why does this class come in in this PR?

src/main/resources/l10n/JabRef_es.properties
```diff
@@ -982,9 +980,6 @@ Doing_a_cleanup_for_%0_entries...=Haciendo_la_limpieza_de_%0_entradas
 No_entry_needed_a_clean_up=Ninguna_entrada_necesitó_de_limpieza
 One_entry_needed_a_clean_up=Sólo_una_entrada_necesitó_limpieza
 %0_entries_needed_a_clean_up=%0_entradas_necesitarion_limpieza
-Error_downloading_file_'%0'=Error_descargando_el_archivo_'%0'
```
koppor: I hope, you did not delete existing translations

koppor:
I am not sure whether this is the right approach. Why not merging `master`, take all language files from there and add missing localizations afterwards?

Assumption: `master` has no duplicate entries and nothing has been translated in the `javafx` branch yet.

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsController.java
boceckts: As described in #1918 somehow the manageJournalAbbreviations dialog made some trouble for the tests. running the tests locally gave me stacktrace below. So after some trial and error and investigation I came to the conclusion that the only differenec for the journal abbreviations dialog is that the controller specified in the fxml file is also the View class. After creating this class and setting it as the controller the error message is gone.
<details>
<summary>Stacktrace for failed test</summary>

```
net.sf.jabref.logic.l10n.LocalizationConsistencyTest > findObsoleteLocalizationKeys FAILED
    java.lang.VerifyError: Bad type on operand stack
    Exception Details:
      Location:
        net/sf/jabref/gui/journals/ManageJournalAbbreviationsView.<init>()V @31: putfield
      Reason:
        Type 'net/sf/jabref/logic/l10n/LocalizationBundle' (current frame, stack[1]) is not assignable to 'java/util/ResourceBundle'
      Current Frame:
        bci: @31
        flags: { }
        locals: { 'net/sf/jabref/gui/journals/ManageJournalAbbreviationsView', '[Z' }
        stack: { 'net/sf/jabref/gui/journals/ManageJournalAbbreviationsView', 'net/sf/jabref/logic/l10n/LocalizationBundle' }
      Bytecode:
        0x0000000: b802 cb4c 2ab7 0006 2b03 0454 2abb 0007
        0x0000010: 59b7 0008 b500 032b 0404 542a b800 09b5
        0x0000020: 000a 2b05 0454 b1
        at java.lang.Class.getDeclaredConstructors0(Native Method)
        at java.lang.Class.privateGetDeclaredConstructors(Class.java:2671)
        at java.lang.Class.getConstructor0(Class.java:3075)
        at java.lang.Class.newInstance(Class.java:412)
        at sun.reflect.misc.ReflectUtil.newInstance(ReflectUtil.java:51)
        at javafx.fxml.FXMLLoader$ValueElement.processAttribute(FXMLLoader.java:927)
        at javafx.fxml.FXMLLoader$InstanceDeclarationElement.processAttribute(FXMLLoader.java:971)
        at javafx.fxml.FXMLLoader$Element.processStartElement(FXMLLoader.java:220)
        at javafx.fxml.FXMLLoader$ValueElement.processStartElement(FXMLLoader.java:744)
        at javafx.fxml.FXMLLoader.processStartElement(FXMLLoader.java:2707)
        at javafx.fxml.FXMLLoader.loadImpl(FXMLLoader.java:2527)
        at javafx.fxml.FXMLLoader.loadImpl(FXMLLoader.java:2441)
        at javafx.fxml.FXMLLoader.load(FXMLLoader.java:2409)
        at net.sf.jabref.logic.l10n.LocalizationParser.getLanguageKeysInFxmlFile(LocalizationParser.java:173)
        at net.sf.jabref.logic.l10n.LocalizationParser.lambda$findLocalizationEntriesInFxmlFiles$2(LocalizationParser.java:90)
        at java.util.stream.ReferencePipeline$7$1.accept(ReferencePipeline.java:267)
        at java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:175)
        at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193)
        at java.util.Iterator.forEachRemaining(Iterator.java:116)
        at java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1801)
        at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481)
        at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:471)
        at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708)
        at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
        at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:499)
        at net.sf.jabref.logic.l10n.LocalizationParser.findLocalizationEntriesInFxmlFiles(LocalizationParser.java:91)
        at net.sf.jabref.logic.l10n.LocalizationParser.findLocalizationEntriesInFiles(LocalizationParser.java:73)
        at net.sf.jabref.logic.l10n.LocalizationParser.findObsolete(LocalizationParser.java:52)
        at net.sf.jabref.logic.l10n.LocalizationConsistencyTest.findObsoleteLocalizationKeys(LocalizationConsistencyTest.java:151)
```

</details>

src/main/resources/l10n/JabRef_es.properties
boceckts: I ran `./gradlew test` locally and it gave me obsolete and missing translation keys which I removed/added to specific files. This translation in particular should not be in use anymore.

boceckts:
Unfortunately one test still fails and it seems to be JavaFX related as the error message says that it has a problem starting the javafx PlatformToolkit. Locally the error message doesn't appear.

stefan-kolb:
Be aware that a message from the localization test that a key is not used anymore might be a false positive! See this commit: https://github.com/JabRef/jabref/commit/05e7f0def59b7c34707f9e0a1f3dc17db678d1f0
This test is not 100% reliable...

koppor:
Why does the " javafx PlatformToolkit" need to be started? We require the start at the integration test and try to avoid UI tests in the master branch. - Is it really required for the localization tests?

src/main/resources/l10n/JabRef_es.properties
koppor: Did you do an [ack](http://beyondgrep.com/) on this string? Did you check with <kbd>CTRL</kbd>+<kbd>SHIFT</kbd>+<kbd>F</kbd> in IntelliJ that this string is not used anywhere?

src/main/java/net/sf/jabref/gui/journals/ManageJournalAbbreviationsController.java
tobiasdiez: That's interesting. Thanks for investigating!

tobiasdiez:
@koppor the JavaFX FXMLLoader is used to parse the FXML files and find the used language keys there.
@boceckts so now the unit tests pass but the integration tests make problems?

boceckts:
Yes but only for thr travis ci, running the tests with gradle on my local machine I have no trouble.

On Thu, Sep 15, 2016 at 9:53am, Tobias Diez < notifications@github.com [notifications@github.com] > wrote:
@koppor [https://github.com/koppor] the JavaFX FXMLLoader is used to parse the FXML files and find the used language keys there.
@boceckts [https://github.com/boceckts] so now the unit tests pass but the integration tests make problems?

boceckts:
@tobiasdiez please ignore my last comment! On my local machine the same tests are passing as on the master branch (4 fetcher tests fail which are not related to javafx). For travis ci the unit test LocalizationConsistencyTest doesn't fail but errors as it can not initialize the javafx PlatformToolkit which is required to get the keys from the fxml files as you already mentioned. I read that travis supports ui tests by adding xvfb so I gave it a shot and the error message changed but unfortunately it still errors.
@stefan-kolb I wasn't aware of that, I will check them by searching them manually.

src/main/resources/l10n/JabRef_es.properties
boceckts: No I thought the gradle test task was is reliable. I checked this particular string and it turned out it was used in the ManageJournalsPanel class which has been removed on the javafx branch.

tobiasdiez:
Ok, could you please create a new PR only with the View -> ViewModel commit https://github.com/JabRef/jabref/pull/1972/commits/76f176e4701707742416172ab829f841987ebc8b. This can be merged immediately. Then I can also investigate the issue. We will probably need a custom FXML Loader which does not draw the controls but only parses the fxml file.

boceckts:
@tobiasdiez I'm not sure about a custom FXML Loader, the integration tests don't just fail but rather timeout which leads to the assumption that something doesn't get terminated properly. Maybe JavaFX ApplicationThread?

tobiasdiez:
So with https://github.com/JabRef/jabref/pull/1986 the tests are now passing but the integration tests still hang. Did you find a way to solve it?

boceckts:
Unfortunately not, I tried changing the travis file with different configurations and changing the closing mechanism but so far I had no luck.

tobiasdiez:
Ok, but you found a solution so that the unit tests pass? (probably https://github.com/JabRef/jabref/pull/1972/files#diff-354f30a63fb0907d4ad57269548329e3R24, right?). 
If this is the case, please rebase and fix the "unused import"-message. Then we can merge this PR in and think about the integration tests later (or think about how to convince @koppor that we disable them completely :smile_cat: )

tobiasdiez:
Ok, then please rebase and we will merge it in.

boceckts:
The travis file still needed a small change to enable the LocalizationConsistencyTest with fxml files, otherwise a warning like "unable to open DISPLAY" comes up when testing with travis. The translation files had to be reworked as the console error dialog has been merged in the meantime. And the controller of the error console fxml file had to be extracted from the view class, same as it was the case with the manage journal abbreviations dialog. Now it seems like the unit test are excecuted and some integration test times out.

tobiasdiez:
Thanks. Looks good to me. @koppor if the travis changes are ok for you, then we take this approach instead of #1986. Your call.

PR #1973, opened 2016-09-13
Update Japanese translation

kojiyokota:
Update
- JabRef_ja.properties
- Menu_ja.properties

Sorry that I couldn’t make it in time for 3.6…

PR #1977, opened 2016-09-13
Update japanse trranlsation (removed duplicates)

Siedlerchr:
@kojiyokota I was so kind and fixed the language files for you. You had some duplicates in there, probably resulted in conflicts from merging

@koppor  We could merge this and then @kojiyokota can update the missing strings again as new pr.

Follow up from #1973


koppor:
[Kudos](http://blog.openhub.net/kudos/) to both of you!

PR #1983, opened 2016-09-14
Master

kojiyokota:
Removing duplicates and added some new.

Currently, following keys seems missing:
- Gray_out_non-hits
- Hide_non-hits


stefan-kolb:
Thank you for your contribution :smile: !

PR #1987, opened 2016-09-15
Fix Entrytypedialog

zellerdev:
Fixing a problem when closing the EntryTypeDialog. The problem is that it is not checked if the fetcherWorker already started. So if we closed the Dialog with `ESC` or `X` a CancelException was thrown. To fix this we check the status of the SwingWorker when we try to close the window.


Siedlerchr:
LGTM :+1:  When tests are green we can merge it.

PR #1997, opened 2016-09-16
Move StringUtil to model

tobiasdiez:
In #1874 part of the StringUtil was already moved to the model package. With this PR the whole class is moved. I think it does not make sense to have two classes with the same purpose but different locations. 

The StringUtil class is rather big and we should consider splitting it or moving some parts of it to different classes. 


stefan-kolb:
How do we really define model? In which way is a string util a model?

tobiasdiez:
For me, these are methods which actually should be in `String.java` but since we have no access to the String-model class (since it is in java.model :smile: ) we put them in a separate file.
 
Of course, not all methods in StringUtil fit this interpretation (for example there is a method related to file names which is probably better moved to `File.getFormattedName` or something like that).

lenhard:
For the sake of moving forward, I would say go ahead. There is no point in stalling eternally and since some parts of the model need these utility methods, (or rather since people are refusing to re-implement them at the respective positions), the class is also needed in model.

Just please, please make sure that we do not add endless methods to it and it turns into yet another good class (which it in a sense already is). How about a compromise: You add a test to `ArchitectureTests` that checks that this class doesn't grow endlessly? Say, you check that its size does not grow higher than it is at the moment. That way, we are forced to discuss every time somebody makes a change.

Furthermore, I think it is not ok to have a `strings` and `util` package in `model`. These packages overlap very much, please merge `util` into `strings`.

Finally, your build currently fails, you need to fix this before merging of course.

tobiasdiez:
Ok, feedback taken into account.

PR #1998, opened 2016-09-16
Introduce convienent interface for ID-based fetcher

tobiasdiez:
Follow-up from #1929. Since the changes in this PR are completely parallel to #1929, I will merge it as soon as the build passes.

### DIFF
--- a/src/main/resources/l10n/JabRef_da.properties
+++ b/src/main/resources/l10n/JabRef_da.properties
@@ -2302,3 +2302,5 @@ Select_first_entry=
 Select_last_entry=
 
 Search_in_all_open_databases=
+
+Invalid_ISBN\:_'%0'.=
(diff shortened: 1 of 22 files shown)


src/main/resources/l10n/JabRef_da.properties
```diff
@@ -2302,3 +2302,5 @@ Select_first_entry=
 Select_last_entry=
 
 Search_in_all_open_databases=
+
+Invalid_ISBN\:_'%0'.=
```
Siedlerchr: Sure we don't have a related error message already?

tobiasdiez: Didn't find anything related to ISBN

Siedlerchr: Ah, yes, we have Invalid_DOI already. It would make sense to have sth like Invalid_%0 which we can reuse for all id fetchers. (e.g. doi/isbn) so we

Issue #1996, opened 2016-09-16
Unnecessary other field tab


grimes2:
JabRef 3.7dev--snapshot--2016-09-16--master--629c3f0
windows 10 10.0 amd64 
Java 1.8.0_101

Steps to reproduce:

1. File > New Bibtex database
2. Bibtex > New entry > Article
3. Entry editor optional field volume: a
4. Save database
5. File > Switch to Biblatex mode
6. File > Switch to Bibtex mode
7. Unnecessary "other field" tab (with doublet field volume)

Siedlerchr:
Fixed with #2002

PR #2002, opened 2016-09-17
Unnecessary other field tab

grimes2:
Switching Bibtex/Biblatex mode causes an unnecessary "other field" tab #1996, filled with optional fields, in entry editor. Other fields are defined as fields that are not required fields + optional fields. The bug is, that on switching database mode the wrong optional fields list is filled. This is fixed.


lenhard:
I have tested your PR with the help of the description in #1996 and everything works as desired. There is only one thing missing in this PR: Please add a Changelog entry and reference the issue that you fixed :)

It seems somewhat counter-intuitive that explicitly adding the optional fields makes the "other fields" tab disappear, but well... it's JabRef.

Siedlerchr:
Thanks for your contribution! 
Fix for #2002

PR #2014, opened 2016-09-20
Add restoration functionality for shared database

obraliar:
Related to https://github.com/JabRef/jabref/issues/1703:

> Session restore: JabRef should open all `[shared]` tabs after startup.

-> Decision: Open from last connection.


src/main/java/net/sf/jabref/JabRefGUI.java
```diff
@@ -192,6 +196,19 @@ private void openWindow() {
...
+
+        boolean isSharedDatabaseEdited = Globals.prefs.getBoolean(JabRefPreferences.SHARED_DATABASE_LAST_EDITED);
+        if (isSharedDatabaseEdited) {
+            boolean isFocused = Globals.prefs.getBoolean(JabRefPreferences.SHARED_DATABASE_LAST_FOCUSED);
+            String keywordSeparator = Globals.prefs.get(JabRefPreferences.KEYWORD_SEPARATOR);
+
+            try {
+                new SharedDatabaseUIManager(mainFrame, keywordSeparator).openLastSharedDatabaseTab(isFocused);
+            } catch (ClassNotFoundException | SQLException | DatabaseNotSupportedException e) {
+                LOGGER.error("Failed to restore shared database. Use connection dialog to connect.");
```
Siedlerchr: Please add the `exception e` as second parameter to the logger.error (there is already an overload for that=

obraliar: I intentionally didn't add it, cause those exceptions basically inform the user about the failure. It's not a real problem. But I think it's better to use `LOGGER.info(...)`.

Siedlerchr: Ah okay. It's just about that in case of a real exception, e.g. soemthing is wrong it helps the developers to debug the root cause ;)

src/main/java/net/sf/jabref/shared/DBMSConnectionProperties.java
```diff
@@ -75,4 +91,45 @@ public void setPassword(String password) {
...
+     */
+    private void setFromPreferences(SharedDatabasePreferences prefs) {
+        if (prefs.getType().isPresent()) {
+            Optional<DBMSType> dbmsType = DBMSType.fromString(prefs.getType().get());
+            if (dbmsType.isPresent()) {
+                this.type = dbmsType.get();
+            }
+        }
+
+        if (prefs.getHost().isPresent()) {
```
tschechlovdev: you could use ``prefs.getHost().ifPresent(prefsHost -> this.host = prefsHost)``.
This avoids the if statements

src/main/java/net/sf/jabref/shared/DBMSConnectionProperties.java
```diff
@@ -75,4 +91,45 @@ public void setPassword(String password) {
...
+
+        if (prefs.getUser().isPresent()) {
+            this.user = prefs.getUser().get();
+        }
+
+        if (prefs.getPassword().isPresent() && prefs.getUser().isPresent()) {
+            try {
+                this.password = new Password(prefs.getPassword().get().toCharArray(), prefs.getUser().get()).decrypt();
+            } catch (UnsupportedEncodingException | GeneralSecurityException e) {
+                LOGGER.error("Could not decrypt pasword" + e);
```
tschechlovdev: replace ``+`` with ``,``

src/main/java/net/sf/jabref/shared/DBMSConnectionProperties.java
```diff
@@ -75,4 +91,45 @@ public void setPassword(String password) {
...
+                this.type = dbmsType.get();
+            }
+        }
+
+        if (prefs.getHost().isPresent()) {
+            this.host = prefs.getHost().get();
+        }
+
+        if (prefs.getPort().isPresent()) {
+            this.port = Integer.parseInt(prefs.getPort().get());
```
tschechlovdev: Maybe you should catch here the NumberFormatException and log, if the port contains some values that cannot converted to an int. Or is this somewhere else done?

src/main/java/net/sf/jabref/shared/DBMSConnectionProperties.java
```diff
@@ -75,4 +91,45 @@ public void setPassword(String password) {
...
+        }
+
+        if (prefs.getName().isPresent()) {
+            this.database = prefs.getName().get();
+        }
+
+        if (prefs.getUser().isPresent()) {
+            this.user = prefs.getUser().get();
```
tschechlovdev: Why don't add here ``if(prefs.getPassword.isPresent())``? Then you don't have to check twice if user is present.

Siedlerchr: I think you mean ifPresent...

src/main/java/net/sf/jabref/shared/DBMSConnectionProperties.java
obraliar: A `NumberFormatException` can never occur, except hard changes in `prefs` (on the file system or registry) were made.

obraliar:
@tschechlovdev Thanks for suggestions :+1:. Appropriate changes have been applied.

src/main/java/net/sf/jabref/shared/DBMSConnectionProperties.java
obraliar: Done.

src/main/java/net/sf/jabref/shared/DBMSConnectionProperties.java
obraliar: Done.

src/main/java/net/sf/jabref/shared/DBMSConnectionProperties.java
obraliar: As well the user as the password is required for the class `Password`. It takes the Username as secret key. Therefore I have to check both. Or am I understanding sth. wrong?

tschechlovdev: No, here I meant ``isPresent``. You check here for user.isPresent and below you check whether password.isPresent and user.isPresent. So why don't combine both:
   ```
    if(prefs.getUser().isPresent() { 
      this.user = ...
       if(prefs.getPassword().isPresent() {
            this.password = ....
        }
      }```

Siedlerchr: When both is required you could just combine the two with and &&

tschechlovdev: Ok nevermind. That's ok how it is.

src/main/java/net/sf/jabref/shared/DBMSConnectionProperties.java
```diff
@@ -75,4 +91,33 @@ public void setPassword(String password) {
...
+
+        if (prefs.getPassword().isPresent() && prefs.getUser().isPresent()) {
+            try {
+                this.password = new Password(prefs.getPassword().get().toCharArray(), prefs.getUser().get()).decrypt();
+            } catch (UnsupportedEncodingException | GeneralSecurityException e) {
+                LOGGER.error("Could not decrypt password", e);
+            }
+        }
+
+        if (!prefs.getPassword().isPresent()) {
```
tschechlovdev: you could combine this if and the if statement above with an if / else if like the following:
```
if(!prefs.getPassword().isPresent()) {
    ...
} else if(prefs.getPassword().isPresent() && prefs.getUser().isPresent()){
  ....
}
```

src/main/java/net/sf/jabref/shared/DBMSConnectionProperties.java
obraliar: Ah you you mean sth. like:
```
if (A) {
   x = ...
}
if (A && B) {
   y = ...
}
```
=> 
```
if (A) {
   x = ...
      if (B) {
         y = ...
      }
}
```
Here you can save one check (A). :+1: 

Done.

src/main/java/net/sf/jabref/shared/DBMSConnectionProperties.java
obraliar: Oh, now I choosed the option from above.

tschechlovdev: Yeah that's ok :+1:

obraliar: Oh, yeah. Done.

koppor:
Went through with Admir. LGTM.

PR #2018, opened 2016-09-20
Add Italian translation for new strings

frithnanth:
Added Italian translation for new strings.
Modified some existing translations.

PR #2020, opened 2016-09-20
 Add Italian translation for new strings (Follow  up)

Siedlerchr:
Follow up from #2018


mlep:
Could/should @frithnanth be added to the team @JabRef/translators ?

koppor:
I merged, resolved the conflicts and sent out the invitation.

Issue #1779, opened 2016-08-18
Integrity check should report empty BibTexKey as problem, too


Siedlerchr:
JabRef version 3.6dev
Steps to reproduce:

1. Create an entry with no BibTeyKey but otherwise all fields set
2. run Integrity Check 
3. Integrity check reports no problem

Warning about empty BibtexKey occurs only at startup. However, the integrity checker should report the empty bibtex key, too.

```
@Patent{,
  author    = {Abraham Lincoln},
  title     = {The Vampire Hunter},
  number    = {12345677},
  year      = {1978},
}
```

grimes2:
Empty required fields are indicated with a red entry number field. So it might be more intuitive, that an empty bibtexkey field is also red.

Siedlerchr:
Refs: https://github.com/koppor/jabref/issues/152

grimes2:
I try to implement this. This is the result:

![emptybibtexkey](https://cloud.githubusercontent.com/assets/15456836/18685316/cc303e9a-7f77-11e6-85fb-1d0150996e68.png)

I've added AuthorTitleYear, because bibtexkey is missing. This is in agreement with the warning on startup.

Siedlerchr:
Looks good! That would be enough. 
If you create PR, we would be pleased :+1:

simonharrer:
Does the "click on item jumps to entry in main table" still work if there is no citekey?

grimes2:
Unfortunately no.

Siedlerchr:
@simonharrer  I have not looked into the code, but I think that the key is used as reference to located the entry...Maybe we should switch that do id?

tobiasdiez:
We should switch it to `BibEntry`, i.e. just keep a reference to the entry and the use/create a method `highlight(BibEntry)`

tobiasdiez:
The checker for empty keys is now in place with #2022. I'll keep this open nonetheless as a reminder that we should fix the highlight functionality at some point.

stefan-kolb:
Follow up in #2181

PR #2022, opened 2016-09-20
Implemented BibtexkeyChecker

grimes2:
Check for empty bibtexkey. #1779.

### DIFF
--- a/src/main/java/net/sf/jabref/logic/integrity/BibtexkeyChecker.java
+++ b/src/main/java/net/sf/jabref/logic/integrity/BibtexkeyChecker.java
@@ -13,6 +13,7 @@ public class BibtexkeyChecker implements Checker {
 
     @Override
     public List<IntegrityMessage> check(BibEntry entry) {
+        Optional<String> valuekey = entry.getCiteKeyOptional();
         Optional<String> valueauthor = entry.getField(FieldName.AUTHOR);
         Optional<String> valuetitle = entry.getField(FieldName.TITLE);
         Optional<String> valueyear = entry.getField(FieldName.YEAR);
@@ -22,7 +23,7 @@ public class BibtexkeyChecker implements Checker {
             return Collections.emptyList();
         }
 
-        if (!entry.getCiteKeyOptional().isPresent() || entry.getCiteKeyOptional().get().isEmpty()) {
+        if (!valuekey.isPresent() || valuekey.get().isEmpty()) {
             return Collections.singletonList(new IntegrityMessage(
                     Localization.lang("empty bibtexkey") + ": " + authortitleyear, entry, BibEntry.KEY_FIELD));
         }
--- a/src/main/java/net/sf/jabref/logic/integrity/BibtexkeyChecker.java
+++ b/src/main/java/net/sf/jabref/logic/integrity/BibtexkeyChecker.java
@@ -1,0 +1,33 @@
+package net.sf.jabref.logic.integrity;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
+
+import net.sf.jabref.logic.integrity.IntegrityCheck.Checker;
+import net.sf.jabref.logic.l10n.Localization;
+import net.sf.jabref.model.entry.BibEntry;
+import net.sf.jabref.model.entry.FieldName;
+
+public class BibtexkeyChecker implements Checker {
+
+    @Override
+    public List<IntegrityMessage> check(BibEntry entry) {
+        Optional<String> valuekey = entry.getField(BibEntry.KEY_FIELD);
+        Optional<String> valueauthor = entry.getField(FieldName.AUTHOR);
+        Optional<String> valuetitle = entry.getField(FieldName.TITLE);
+        Optional<String> valueyear = entry.getField(FieldName.YEAR);
+        String authortitleyear = entry.getAuthorTitleYear(100);
+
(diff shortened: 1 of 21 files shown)


koppor:
Could add tests and a CHANGELOG entry? Maybe in a follow up, could you address https://github.com/JabRef/jabref/issues/1779#issuecomment-248422973 so that the jumping to an entry works better. @bartsch-dev Maybe you have an idea for a quick implementation of that?

grimes2:
Tests: It is difficult to test for an unexisting variable in this test environment. I'll test only for "assertCorrect".

src/test/java/net/sf/jabref/logic/integrity/IntegrityCheckTest.java
```diff
@@ -112,6 +112,12 @@ public void testMonthChecks() {
     }
 
     @Test
+    public void testBibtexkeyChecks() {
+        assertCorrect(createContext("bibtexkey", "Knuth2014"));
```
koppor: Could you do one assertCorrect per Test please? :innocent:

koppor:
Tests: I view the changes using the [CodeCov plugin](https://github.com/codecov/browser-extension) (see also https://github.com/JabRef/jabref/wiki/Tools) and this is the result:

grabbed_20160921-094240

I just wanted to ensure that everything is at least orange if not green :roller_coaster:

src/main/java/net/sf/jabref/logic/integrity/BibtexkeyChecker.java
```diff
@@ -0,0 +1,33 @@
...
+import net.sf.jabref.logic.integrity.IntegrityCheck.Checker;
+import net.sf.jabref.logic.l10n.Localization;
+import net.sf.jabref.model.entry.BibEntry;
+import net.sf.jabref.model.entry.FieldName;
+
+public class BibtexkeyChecker implements Checker {
+
+    @Override
+    public List<IntegrityMessage> check(BibEntry entry) {
+        Optional<String> valuekey = entry.getField(BibEntry.KEY_FIELD);
```
tobiasdiez: Please use `entry.getCiteKey`.

grimes2:
I think, it is difficult to implement empty bibtexkey in a clean way. The problem is that the bibtexkey is not existing, but the constructor IntegrityMessage requires an existing entry and field. So this might be the reason, why codecov is complaining. A solution could be to overload the constructor IntegrityMessage for this special case.

src/main/java/net/sf/jabref/logic/integrity/BibtexkeyChecker.java
```diff
@@ -0,0 +1,33 @@
...
+        Optional<String> valueauthor = entry.getField(FieldName.AUTHOR);
+        Optional<String> valuetitle = entry.getField(FieldName.TITLE);
+        Optional<String> valueyear = entry.getField(FieldName.YEAR);
+        String authortitleyear = entry.getAuthorTitleYear(100);
+
+        if (!valueauthor.isPresent() || !valuetitle.isPresent() || !valueyear.isPresent()) {
```
tobiasdiez: I don't understand this if statement. So if an entry is missing an author (or title/year) and also has no bibtexkey, then no warning will be printed?

src/main/java/net/sf/jabref/logic/integrity/BibtexkeyChecker.java
```diff
@@ -0,0 +1,33 @@
...
+            return Collections.emptyList();
+        }
+
+        if (!valuekey.isPresent()) {
```
tobiasdiez: I would also check that the key is not empty (like in your second test below). 
There is a similar convenient method in StringUtil already (I think it was called `isNotBlank` not sure if `isBlank` already exists but it should be easy to add).

src/main/java/net/sf/jabref/logic/integrity/BibtexkeyChecker.java
grimes2: This is because getAuthorTitleYear(i) requires, that all three field exist. Otherwise some check tests are failing.

tobiasdiez: Judging based on the code, I think getAuthorTitleYear would return `N/A: N/A (N/A)` in the case where the fields are missing. I would suggest that you change the order of the if statements.
If the key is empty, then we definitely want to report a message. But if then all the author, title and year are empty, we use BibEntry.toString (it is a bit ugly) or otherwise we use getAuthorTitleYear.

grimes2: Changing the order of the if statements causes the fail of **all** checker tests:
java.lang.AssertionError: expected:<[]> but was:<[[] in bibtexkey: empty bibtexkey: N/A: "N/A" (N/A)]

src/test/java/net/sf/jabref/logic/integrity/IntegrityCheckTest.java
Siedlerchr: As we now have a class for each checker, we should also create a test class for each checker separately.
But I would think that is content for a new PR

grimes2: Removed my thumbs up. A separate class for every checker is too much (for evt. one or two entries).  I think, it is better to organize the checker tests by fields (in methods) and not by checker classes.

Issue #1647, opened 2016-07-30
RIS import


asnobli:
1. doi import does not work with the DO or M3 (RIS standard fields)
2. patent is imported as "other"

tschechlovdev:
Thanks for reporting. Could you please write the steps to reproduce and the file(s) you have used ? That would help me a lot.

boceckts:
I can import the doi using the M3 field. Make sure the "-" sign is at the fourth position and the doi has to be in the originally specified [doi](https://en.wikipedia.org/wiki/Digital_object_identifier#Display) format. An url can not be parsed (just yet).
For example I can read a .ris file with the field `M3  - doi:10.1000/182` and the doi displayed in jabref is 10.1000/182.
However looking at the code, the DO field seems to be not specified in the current ris importer and there is only a limited set of types for the TE field supported, excluding Patent. We should definitely rework this importer.

stefan-kolb:
@tschechlovdev  @Braunch  @boceckts This was fixed in #2028?!

tschechlovdev:
Yes this was fixed, so I think this can be closed now.

crystalfp:
Well, no.

In latest version (see below) JR ignores DO field. Try importing (after changing extension to .ris) the following RIS file: [PMC3113697.txt](https://github.com/JabRef/jabref/files/3004842/PMC3113697.txt)

JabRef 4.3.1
Windows 10 10.0 amd64 
Java 1.8.0_201

chapb:
Same problem as @crystalfp. The DO field is ignored during RIS import from Mendeley.

Issue #1074, opened 2016-04-01
RIS import not working for journal title


herrschaftsseitn:
JabRef version 3.2/2.10/ and earlier on Windows 7/Ubuntu 14.04

Steps to reproduce:

1. import a RIS file from Scopus into a *bib database
2. check for journal/booktitle
3. the title of the journal is imported as the book title

See attached two examples of RIS files from Scopus and Sciencedirect for the same article. In addition the bib-database with both imported. The import from ScienceDirect works fine, the one from Scopus needs to be edited for the journal. 

Hope it is repcoducible and one finds an explanation/solution. Since I'm using mainly Scopus, it would save tons of manual editting. 

[SC vs SD.zip](https://github.com/JabRef/jabref/files/199000/SC.vs.SD.zip)

koppor:
Refs #817

tschechlovdev:
When I see it right, then it's a problem from Scopus. Because in the RIS File form scopus the journal title is in the field T2, which is the field for the "second title". In Sciencedirect it is the field JO, which stands for the journal title. In JabRef the T2 field is the field "booktitle", which is misleading. So what we could do is change the field to second title, instead of booktitle. But this won't solve your problem since scopus is exporting the journal title as second title.

herrschaftsseitn:
Yes, that äs basically true. 
What I don't understand is the exakt definition of the RIS-file format, i.e. is the Scopus or the ScienceDirect mode of export correct. So does Wikipedia [1] state that J2 would be correct for the journal title; similar the documentation at Thomson Reuters [2]. And thet would say that Scopus is the correct way how to use it. 
Since it is obvious that there exist different definitions for export, would it be an option to create a mapping that can be edited by the jabref user? But I have no idea how easy/complicated it would be to implement this. 

[1] https://en.wikipedia.org/wiki/RIS_%28file_format%29
[2] http://refman.com/sites/rm/files/m/direct_export_ris.pdf

tschechlovdev:
This should be fixed in #2028. The journal title from a scopus file should now be imported as journal title.
So I think this can be closed now.

koppor:
Thank you for reporting this issue. We think, that is already fixed in our development version and consequently the change will be included in the next release.

We would like to ask you to use a development build from https://builds.jabref.org/master.

PR #2028, opened 2016-09-21
[WIP] Fix ris import

tschechlovdev:
Regarding: #1647 
Doi import should now work with the DO field. Also I've added some fields, that are available in the [RIS specification](http://referencemanager.com/sites/rm/files/m/direct_export_ris.pdf). I've also done some code cleanups.
#1074 should be also fixed now. I've added the scopus file as test file.


src/main/java/net/sf/jabref/logic/importer/fileformat/RisImporter.java
```diff
@@ -43,28 +45,17 @@ public String getDescription() {
...
 
-        String line;
-        while ((line = reader.readLine()) != null) {
-            sb.append(line);
-            sb.append('\n');
-        }
+        String linesAsString = reader.lines().reduce((line, nextline) -> line + "\n" + nextline).get();
 
-        String[] entries = sb.toString().replace("\u2013", "-").replace("\u2014", "--").replace("\u2015", "--")
+        String[] entries = linesAsString.replace("\u2013", "-").replace("\u2014", "--").replace("\u2015", "--")
```
Siedlerchr: Maybe you could extract that method to the StringUtil class, e.g. call it ReplaceUnicodeDashes..

src/test/resources/net/sf/jabref/logic/importer/fileformat/RisImporterTestDoiAndJournalTitle.bib
```diff
@@ -0,0 +1,10 @@
+@patent{,
+  author   = {some author},
+  month    = dec,
+  year     = {2014},
+  caption  = {254},
+  edition  = {5},
+  editor   = {some editor},
+  language = {eng},
+  number   = {25},
+}
```
boceckts: The name of the test files confuses me a lot...Isn't the purpose here to test the journal title and doi fields? So why are the fields not present in the file?

src/test/resources/net/sf/jabref/logic/importer/fileformat/RisImporterTest3.bib
```diff
@@ -1,14 +1,14 @@
 @incollection{,
   author = {Barata, Catarina and Celebi, MEmre and Marques, JorgeS},
-  booktitle = {Dermoscopy Image Analysis},
   comment = {doi:10.1201/b19107-2},
   doi = {10.1201/b19107-2},
```
boceckts: I have seen that RisImporterTest3.ris specifies the both doi fields (M3 and DO) with the same doi string. Please move one of them to a different or new test file. This way the tests would have already showed that the DO field didn't import the doi.

src/test/resources/net/sf/jabref/logic/importer/fileformat/RisImporterTestDoiAndJournalTitle.bib
tschechlovdev: Yeah you're right... Don't know how I could forget them :smile:

src/test/resources/net/sf/jabref/logic/importer/fileformat/RisImporterTest3.bib
tschechlovdev: I've used the DO field in the TestDoiAndJournalFile.

tschechlovdev:
Travis fails, because the Architecture tests say, that the StringUtil class should be kept as small as possible ...

Siedlerchr:
Speak with @lenhard if it's okay to increase it and how to fix the test

tschechlovdev:
I tried to contact @lenhard, but didn't get an answer. So for the moment I think it is the best to remove the method from the StringUtil class, so that this PR don't have to wait to long. In addition it would only be 2 classes who would use this method.

Travis is still failing because of a DOI test.

Siedlerchr:
Yes, it is the proper way for the moment. We can think later on for moving it.
I would merge it in then when you have done it

lenhard:
@tschechlovdev You contacted me? How? :-) This mention of myself is the first sign I see, but maybe I overlooked something.

Siedlerchr:
@lenhard  The question was about the "Failing travis test" regarding the StringUtil method.
The ideas was to add  a method for replacing unicode dashes:
` String[] entries = linesAsString.replace("\u2013", "-").replace("\u2014", "--").replace("\u2015", "--") `
which is also used in medline Plain.

tobiasdiez:
I know I'm late to the party, but one question: is this unicode dash needed for the page numbers? Because then there is a cleanup formatter which does the job (Page number normalizer, or some permutation of these words).

lenhard:
The reason for moving more functionality to `StringUtil` in model was, if I recall it correctly, that the group fixes needed these methods. If the dash replacement functionality is not needed by model classes, why should it go into that package? If you need that in multiple importer classes, why not introduce a util class for the importer package?

Issue #2029, opened 2016-09-22
Localization tests refer to nonexistant Gradle tasks


stefan-kolb:
e.g. to remove obsolete or missing language keys 

`gradlew -b localization.gradle generateMissingTranslationKeys`

Caused by #2010

PR #2031, opened 2016-09-22
Fix localization task hints

bartsch-dev:
fixes: #2029

Replaces hints to the old gradle localization task names with the new ones.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -85,7 +85,7 @@ because <additional rationale>.
 Add new Localization.lang("KEY") to Java file.
 Tests fail. In the test output a snippet is generated which must be added to the English translation file. There is also a snippet generated for the non-English files, but this is irrelevant.
 Add snippet to English translation file located at `src/main/resources/l10n/JabRef_en.properties`
-With `gradlew update` the "KEY" is added to the other translation files as well.
+With `gradlew localizationUpdate` the "KEY" is added to the other translation files as well.
 Tests are green again.
 
 You can also directly run the specific test in your IDE. The test "LocalizationConsistencyTest" is placed under `src/test/java/net.sf.jabref.logic.l10n/LocalizationConsistencyTest.java`
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -85,7 +85,7 @@ because <additional rationale>.
 Add new Localization.lang("KEY") to Java file.
 Tests fail. In the test output a snippet is generated which must be added to the English translation file. There is also a snippet generated for the non-English files, but this is irrelevant.
 Add snippet to English translation file located at `src/main/resources/l10n/JabRef_en.properties`
-With `gradlew generateMissingTranslationKeys` the "KEY" is added to the other translation files as well.
+With `gradlew update` the "KEY" is added to the other translation files as well.
 Tests are green again.
 
 You can also directly run the specific test in your IDE. The test "LocalizationConsistencyTest" is placed under `src/test/java/net.sf.jabref.logic.l10n/LocalizationConsistencyTest.java`
--- a/src/test/java/net/sf/jabref/logic/l10n/LocalizationConsistencyTest.java
+++ b/src/test/java/net/sf/jabref/logic/l10n/LocalizationConsistencyTest.java
@@ -156,7 +156,7 @@ public class LocalizationConsistencyTest {
             System.out.println();
             System.out.println("1. CAREFULLY CHECK IF THE KEY IS REALLY NOT USED ANYMORE");
             System.out.println("2. REMOVE THESE FROM THE ENGLISH LANGUAGE FILE");
-            System.out.println("3. EXECUTE gradlew update");
+            System.out.println("3. EXECUTE gradlew localizationUpdate");
             System.out.println("REMOVE THESE FROM THE NON-ENGLISH LANGUAGE FILES");
             fail("Obsolete keys " + obsoleteKeys + " found in properties file which should be removed");
         }
@@ -172,7 +172,7 @@ public class LocalizationConsistencyTest {
             System.out.println(obsoleteKeys.stream().map(Object::toString).collect(Collectors.joining("\n")));
             System.out.println();
             System.out.println("1. REMOVE THESE FROM THE ENGLISH LANGUAGE FILE");
-            System.out.println("3. EXECUTE gradlew update");
+            System.out.println("2. EXECUTE gradlew localizationUpdate");
             System.out.println("REMOVE THESE FROM THE NON-ENGLISH LANGUAGE FILES");
             fail("Obsolete keys " + obsoleteKeys + " found in menu properties file which should be removed");
         }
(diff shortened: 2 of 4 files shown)


src/test/java/net/sf/jabref/logic/l10n/LocalizationConsistencyTest.java
```diff
@@ -156,7 +156,7 @@ public void findObsoleteLocalizationKeys() throws IOException {
             System.out.println();
             System.out.println("1. CAREFULLY CHECK IF THE KEY IS REALLY NOT USED ANYMORE");
             System.out.println("2. REMOVE THESE FROM THE ENGLISH LANGUAGE FILE");
-            System.out.println("3. EXECUTE gradlew -b localization.gradle generateMissingTranslationKeys TO");
+            System.out.println("3. EXECUTE gradlew update");
```
matthiasgeiger: This should be `gradlew -b localization.gradle update`, right?

src/test/java/net/sf/jabref/logic/l10n/LocalizationConsistencyTest.java
```diff
@@ -172,7 +172,7 @@ public void findObsoleteMenuLocalizationKeys() throws IOException {
             System.out.println(obsoleteKeys.stream().map(Object::toString).collect(Collectors.joining("\n")));
             System.out.println();
             System.out.println("1. REMOVE THESE FROM THE ENGLISH LANGUAGE FILE");
-            System.out.println("2. EXECUTE gradlew -b localization.gradle generateMissingTranslationKeys" + " TO");
+            System.out.println("3. EXECUTE gradlew update");
```
matthiasgeiger: same as above

src/test/java/net/sf/jabref/logic/l10n/LocalizationConsistencyTest.java
```diff
@@ -197,7 +198,7 @@ private String convertToEnglishPropertiesFile(List<LocalizationEntry> missingKey
     }
 
     private String convertPropertiesFile(List<LocalizationEntry> missingKeys) {
-        System.out.println("EXECUTE gradlew -b localization.gradle generateMissingTranslationKeys TO");
+        System.out.println("EXECUTE gradlew update");
```
matthiasgeiger: same as above

src/test/java/net/sf/jabref/logic/l10n/LocalizationConsistencyTest.java
bartsch-dev: Nope, `localization.gradle` is imported in the `build.gradle` file, thus gradle knows the tasks form `localization.gradle` too.
Just try it out.

src/test/java/net/sf/jabref/logic/l10n/LocalizationConsistencyTest.java
bartsch-dev: see above, but fixed wrong item enumeration.

CONTRIBUTING.md
```diff
@@ -85,7 +85,7 @@ because <additional rationale>.
 Add new Localization.lang("KEY") to Java file.
 Tests fail. In the test output a snippet is generated which must be added to the English translation file. There is also a snippet generated for the non-English files, but this is irrelevant.
 Add snippet to English translation file located at `src/main/resources/l10n/JabRef_en.properties`
-With `gradlew generateMissingTranslationKeys` the "KEY" is added to the other translation files as well.
+With `gradlew update` the "KEY" is added to the other translation files as well.
```
stefan-kolb: Hm, I don't like the current task name. It should say what it does, like `updateLocalization`. Please also change the other Gradle tasks accordingly.

bartsch-dev: You're right. I use the gradle tasks out of IntelliJ IDEA which groups them by its category.
I added `localization` as a prefix to each of them.

stefan-kolb: I also use them that way, thats the reason why I only realized this problem when looking at the PR code.

src/test/java/net/sf/jabref/logic/l10n/LocalizationConsistencyTest.java
```diff
@@ -172,7 +172,7 @@ public void findObsoleteMenuLocalizationKeys() throws IOException {
             System.out.println(obsoleteKeys.stream().map(Object::toString).collect(Collectors.joining("\n")));
             System.out.println();
             System.out.println("1. REMOVE THESE FROM THE ENGLISH LANGUAGE FILE");
-            System.out.println("3. EXECUTE gradlew update");
+            System.out.println("2. EXECUTE gradlew localizationUpdate");
```
tobiasdiez: If you already touch the localization tests, could you also remove the `System.out.println` statements by moving everything to the error message.
So for example this here should be 
`assertEquals("Obsolte kyes found in menu properties file. \n 1. Remove these... \n 2...", emptyList, obsolteKeys)`

bartsch-dev: done

src/test/java/net/sf/jabref/logic/l10n/LocalizationConsistencyTest.java
```diff
@@ -124,86 +122,49 @@ public void findMissingLocalizationKeys() throws IOException {
...
-        assertEquals(
-                "source code contains language keys for the messages which are not in the corresponding properties file",
-                "", resultString);
+        assertEquals("DETECTED LANGUAGE KEYS WHICH ARE NOT IN THE ENGLISH LANGUAGE FILE\n" +
+                "1. PASTE THESE INTO THE ENGLISH LANGUAGE FILE\n" +
+                "2. EXECUTE: gradlew localizationUpdate\n" +
+                missingKeys.parallelStream()
+                        .map(key -> String.format("%s=%s", key.getKey(), key.getKey()))
+                        .collect(Collectors.toList()),
+                Collections.EMPTY_LIST, missingKeys);
```
tobiasdiez: Collections.emptyList() is the preferred way to access it (don't ask me why).

tobiasdiez: Collections.emptyList() is the preferred way to access it (don't ask me why).

bartsch-dev: The advantage of `Collections.EMPTY_LIST` is that this list is immutable.

Siedlerchr: Collections.emptyList() is better because it is a generic implementatin! The constant is not generic

Siedlerchr: And it's also immuntable: https://docs.oracle.com/javase/8/docs/api/java/util/Collections.html#emptyList-- 
(Unlike this method, the field does not provide type safety.)

bartsch-dev: Thanks, I changed the lists.

src/test/java/net/sf/jabref/logic/l10n/LocalizationConsistencyTest.java
```diff
@@ -128,7 +128,7 @@ public void findMissingLocalizationKeys() throws IOException {
                 missingKeys.parallelStream()
                         .map(key -> String.format("%s=%s", key.getKey(), key.getKey()))
                         .collect(Collectors.toList()),
-                Collections.EMPTY_LIST, missingKeys);
+                Collections.<LocalizationEntry>emptyList(), missingKeys);
```
Siedlerchr: Just for the future, you could have just taken the Collections.emptyList() without specifying the type explicit.

PR #2035, opened 2016-09-22
French localization: Jabref_fr: correction of a glitch (+untranslated strings)

mlep:
By the way, I believe that in English, their is no such thing as "large capital" (while there are "small capitals"...). "Large capitals" should probably be "Capital letters" in the key.


Siedlerchr:
Hm, yeah, this is confusing: https://en.wikipedia.org/wiki/Small_caps
And I think it is wrongly used and capital letters should be used indeed:
http://lawprofessors.typepad.com/legal_skills/2011/07/another-typing-trick-large-small-caps.html

Issue #549, opened 2015-12-21
shortcut for copying a url link in the name of bibtexkey


jhshinn:
This is a sort of feature request.
It would great if JabRef provides a shortcut key for copying the url in the name of bibtexkey.
During discussion with other people in email, I've encountered many occasions that I should direct them to some articles.
I think the url link is useful for this purpose, especially in the name of bibtexkey. e.g. [Zaritsky(1994)AJ_108_1619](http://adsabs.harvard.edu/abs/1994AJ....108.1619Z)

simonharrer:
> copying the url in the name of bibtexkey.

What do you mean by that? Can you please elaborate?

koppor:
I think, when issuing the new "copy key and link" command, a RTF/HTML should be copied.

    <a href="http://adsabs.harvard.edu/abs/1994AJ....108.1619Z">Zaritsky(1994)AJ_108_1619</a>

The content of `href` should be taken from the `URL` field, the content of the link text is the bibtex key.

simonharrer:
Ah ok. Should be straight forward then.

jhshinn:
@simonharrer I'm sorry not to clarify it enough. It is as @koppor described. Thanks much, @koppor

boceckts:
I will close this as everything seems to be done.

tobiasdiez:
This is now implemented in the latest development version. Note that not every applications support HTML content in the clipboard and thus paste does not always work (e.g. the link appears in Word but not in Notepad). Could you please check the build from http://builds.jabref.org/master/. Thanks! Please remember to make a backup of your library before trying-out this version.

jhshinn:
Thanks much for your excellent job! I've found that the "copy bibtex key and link" function of the recent build works properly on Windows 10, but not on Ubuntu 16.04 as before. I could not paste anything on Ubuntu. I've checked the function in the new message window of Gmail.

PR #2037, opened 2016-09-22
Added new menu and action to copy bibtex key as url

boceckts:
Added a new sub menu item to the menu and the right click menu to quickly copy the BibTeX key as a hyperlink to the entries' url as described in #549.
In case of an empty url the BibTeX key will be copied by its own.

rightclickmenuentry
menuentry

Copying BibTeX keys as hyperlink with their url: source
screen shot 2016-09-23 at 10 53 26

Copying BibTeX keys as hyperlink with their url: markdown formated
screen shot 2016-09-23 at 10 53 42


src/main/java/net/sf/jabref/gui/BasePanel.java
```diff
@@ -950,6 +955,40 @@ private void copyKeyAndTitle() {
...
 
+    private void copyKeyAndLink() {
+        List<BibEntry> selectedEntries = mainTable.getSelectedEntries();
+        if (!selectedEntries.isEmpty()) {
+            storeCurrentEdit();
+
+            StringBuilder sb = new StringBuilder();
+
+            long copied = 0;
+            for (BibEntry entry : selectedEntries.stream().filter(BibEntry::hasCiteKey).collect(Collectors.toList())) {
```
tschechlovdev: I think it would be better if you extract the filtered list to a local variable

src/main/java/net/sf/jabref/gui/BasePanel.java
```diff
@@ -950,6 +955,40 @@ private void copyKeyAndTitle() {
...
+                String key = entry.getCiteKeyOptional().get();
+                String url = entry.getField(FieldName.URL).isPresent() ? entry.getField(FieldName.URL).get() : "";
```
tschechlovdev: Hm, which sin does it have to have an empty url? 
Wouldn't it be better if you also filter if the url is not empty?

src/main/java/net/sf/jabref/gui/BasePanel.java
```diff
@@ -950,6 +955,40 @@ private void copyKeyAndTitle() {
...
+                sb.append(String.format("<a href=\"%s\">%s</a>", url, key));
+                sb.append(OS.NEWLINE);
+                copied++;
+            }
+
+            if (copied == 0) {
```
tschechlovdev: If you extract the list above, you could also check this before the for loop, by checking if the list is empty.

CHANGELOG.md
```diff
@@ -42,6 +42,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - Importer for MODS format added
 - [#2012](https://github.com/JabRef/jabref/issues/2012) Implemented integrity check for `month` field: Should be an integer or normalized (BibLaTeX), Should be normalized (BibTeX)
 - [#1779](https://github.com/JabRef/jabref/issues/1779) Implemented integrity check for `bibtexkey` field: Empty bibtexkey
+- [#549](https://github.com/JabRef/jabref/issues/549) Added new shortcut to copy the BibTeX key as a hyperlink to its url to the clipboard
```
tschechlovdev: I thought that the url should be copied? Don't really get this from the description.

src/main/java/net/sf/jabref/gui/BasePanel.java
```diff
@@ -950,6 +955,40 @@ private void copyKeyAndTitle() {
...
+                output(Localization.lang("None of the selected entries have BibTeX keys."));
+                return;
+            }
+
+            final StringSelection ss = new StringSelection(sb.toString());
```
tschechlovdev: please rename ss, for example to selection.

src/main/java/net/sf/jabref/gui/BasePanel.java
boceckts: Hm you are right there is not much sense in using an empty hyperlink but shouldn't at least the BibTeX key get copied?

CHANGELOG.md
boceckts: A hyperlink should be copied in the format <a href="url">key</a>, meaning the Bibtex key gets displayed but it's actually a hyperlink to the url of the entry.

src/main/java/net/sf/jabref/gui/BasePanel.java
boceckts: I just used this code snipped from the `copyKeyAndTitle` mthod but actually I think it can be replaced by the ClipboardManager's `setClipboardContents()` method.

boceckts:
@bartsch-dev I think you recently reworked the localization script so maybe you can help me here. When I add a new translation string to en property file, is there anything else to do than to execute `./gradlew update`? Because this gives me a syntax error in the python script...
```
Traceback (most recent call last):
  File "/Users/tobias/git/jabref/scripts/syncLang.py", line 423, in <module>
    update(extended=len(sys.argv) == 3 and sys.argv[2] == "--extended")
  File "/Users/tobias/git/jabref/scripts/syncLang.py", line 384, in update
    update_properties(main_property_file=get_main_menu_properties(), other_property_files=get_other_menu_properties())
  File "/Users/tobias/git/jabref/scripts/syncLang.py", line 334, in update_properties
    keys, not_fixed, fixed = fix_duplicates(lines=lines)
  File "/Users/tobias/git/jabref/scripts/syncLang.py", line 191, in fix_duplicates
    fixed.append("{key}={value}".format(key=key, value=value))
UnicodeEncodeError: 'ascii' codec can't encode characters in position 0-6: ordinal not in range(128)
```

src/main/java/net/sf/jabref/gui/BasePanel.java
tschechlovdev: Yeah you're right. The current version looks good :+1:

bartsch-dev:
The linked PR fixes that (I forgot to mark some strings as Unicode strings).

src/main/java/net/sf/jabref/gui/BasePanel.java
```diff
@@ -961,6 +966,41 @@ private void copyKeyAndTitle() {
...
+
+            List<BibEntry> entriesWithKey = selectedEntries.stream().filter(BibEntry::hasCiteKey).collect(Collectors.toList());
+            
+            if (entriesWithKey.isEmpty()) {
+                output(Localization.lang("None of the selected entries have BibTeX keys."));
+                return;
+            }
+            for (BibEntry entry : entriesWithKey) {
+                String key = entry.getCiteKeyOptional().get();
+                String url = entry.getField(FieldName.URL).isPresent() ? entry.getField(FieldName.URL).get() : "";
```
tobiasdiez: getField(URL).orElse("");

boceckts: Done

boceckts:
I extracted the method to a new action class.

src/main/java/net/sf/jabref/gui/actions/CopyBibTeXKeyAndLink.java
```diff
@@ -0,0 +1,61 @@
...
+import net.sf.jabref.logic.util.OS;
+import net.sf.jabref.model.entry.BibEntry;
+import net.sf.jabref.model.entry.FieldName;
+
+/**
+ * This class will copy each selected entry's BibTeX key as a hyperlink to it's url to the clipboard.
+ * In case an entry doesn't have a BibTeX key it will not be copied.
+ * In case an entry doesn't have an url this will only copy the BibTeX key.
+ */
+public class CopyBibTeXKeyAndLink implements BaseAction {
```
Braunch: Please add the suffix "Action" to the class name, since this is the convenient way to name the actions in the actions package.

tobiasdiez:
Thanks! 👍 Can be merged from my point of view.

PR #2039, opened 2016-09-22
[WIP] Hotfix for IndexOutOfBoundsException in MainTable

obraliar:
Hotfix `MainTable` using `Optional`.

Steps to reproduce the bug:

* Open two JabRef instances A and B
* Connect A and B to a shared database supporting LiveUpdate
* Create a new Entry in A
* Open the entry and change the value of a field like author
* Delete the new Entry from B and confirm the upcoming dialog in A

`MainTable` and `EntryMarker` contains a lot of ugly code which produces such errors and can be refactored. @bartsch-dev shows interest to do this probably in the next week. :+1:

### DIFF
--- a/src/main/java/net/sf/jabref/gui/maintable/MainTable.java
+++ b/src/main/java/net/sf/jabref/gui/maintable/MainTable.java
@@ -555,30 +555,39 @@ public class MainTable extends JTable {
     }
 
     private boolean matches(int row, Matcher<BibEntry> m) {
-        return m.matches(getBibEntry(row));
+        Optional<BibEntry> bibEntry = getBibEntry(row);
+
+        if (bibEntry.isPresent()) {
+            return m.matches(bibEntry.get());
+        }
+        return m.matches(null);
     }
 
     private boolean isComplete(int row) {
-        try {
-            BibEntry entry = getBibEntry(row);
-            TypedBibEntry typedEntry = new TypedBibEntry(entry, panel.getBibDatabaseContext());
+        Optional<BibEntry> bibEntry = getBibEntry(row);
+
+        if (bibEntry.isPresent()) {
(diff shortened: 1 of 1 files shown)


src/main/java/net/sf/jabref/gui/maintable/MainTable.java
```diff
@@ -555,30 +555,39 @@ public boolean isFileColumn(int modelIndex) {
     }
 
     private boolean matches(int row, Matcher<BibEntry> m) {
-        return m.matches(getBibEntry(row));
+        Optional<BibEntry> bibEntry = getBibEntry(row);
+
+        if (bibEntry.isPresent()) {
+            return m.matches(bibEntry.get());
+        }
+        return m.matches(null);
```
koppor: I would return `true` here.

src/main/java/net/sf/jabref/gui/maintable/MainTable.java
```diff
@@ -555,30 +555,39 @@ public boolean isFileColumn(int modelIndex) {
...
-            TypedBibEntry typedEntry = new TypedBibEntry(entry, panel.getBibDatabaseContext());
+        Optional<BibEntry> bibEntry = getBibEntry(row);
+
+        if (bibEntry.isPresent()) {
+            TypedBibEntry typedEntry = new TypedBibEntry(bibEntry.get(), panel.getBibDatabaseContext());
             return typedEntry.hasAllRequiredFields();
-        } catch (NullPointerException ex) {
-            return true;
         }
+        return true;
```
koppor: I would return `false` here.

koppor:
I could not reproduce on Windows, but I saw it on Linux. Seems to be an issue at timing with the UI and refreshes.

Issue #490, opened 2015-12-09
Main Table: Right click on DOI field: Allow copying DOI


koppor:
Currently, the popup of DOI allows for clicking the DOI only. It should also be possible to isse "copy DOI", where the complete link is copied.

Usecase: Quick sharing of URLs to references.

stefan-kolb:
On which field? In field editor? See #492

koppor:
In the main table. At the DOI column. See screenshot:

![grabbed_20151209-205427](https://cloud.githubusercontent.com/assets/1366654/11696901/0facf692-9eb7-11e5-9533-7b6bb0ff59f1.png)

matthiasgeiger:
I'll check whether I can integrate this in #452.

matthiasgeiger:
Status: not yet implemented

simonharrer:
Function could also be added to the entry editor.

boceckts:
I will close this as everything seems to be done.

PR #2040, opened 2016-09-22
Right click menus should allow copying doi url

boceckts:
Right clicking on the doi field in the main table or the entry editor shows a sub menu where the user can copy the url of the doi to the clipboard. Refs #490

Right click menu when doi field isn't present
rightclick

Right click menu in entry editor when doi field is not empty
entryrightclick

Right click menu in entry editor when doi field is empty
emptydoi

- [x] Add the menu to the entry editor


src/main/java/net/sf/jabref/gui/actions/CopyDoiUrlAction.java
```diff
@@ -0,0 +1,52 @@
...
+    public CopyDoiUrlAction(JTextComponent component) {
+        this(component.getText());
+        this.component = component;
+    }
+
+    @Override
+    public void actionPerformed(ActionEvent e) {
+        identifier = component == null ? identifier : component.getText();
+
+        Optional<DOI> doiOptional = DOI.build(identifier);
```
tobiasdiez: This can be simplified a bit:
`Optional<String> url = DOI.build(identifier).map(DOI::getURIAsASCIIString)`

boceckts: Done

tobiasdiez:
Thanks!

src/main/java/net/sf/jabref/gui/actions/CopyDoiUrlAction.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    public CopyDoiUrlAction(JTextComponent component) {
+        this(component.getText());
+        this.component = component;
+    }
+
+    @Override
+    public void actionPerformed(ActionEvent e) {
+        identifier = component == null ? identifier : component.getText();
```
Siedlerchr: Please split that in an if/else, because that ternary operator and the assignment in one statement make it hard to understand for non java exports, especially if you don't know about the operator precedence.

src/main/java/net/sf/jabref/gui/maintable/MainTableSelectionListener.java
```diff
@@ -407,6 +408,9 @@ private void showIconRightClickMenu(MouseEvent e, int row, MainTableColumn colum
                             }
                             menu.add(new ExternalFileMenuItem(panel.frame(), entry, content.get(), content.get(), icon,
                                     panel.getBibDatabaseContext(), field));
+                            if (field == FieldName.DOI) {
```
Siedlerchr: Beware! Although FieldName.XXX looks like an enum, is is String constant, so I think you should better use equals

Siedlerchr:
@tobiasdiez  A bit too fast. At least the equals thing should be fixed.

boceckts:
@Siedlerchr can I still push to this branch or should I create a new branch + pull request?

Siedlerchr:
Hm, not sure. Better create a new one based on master and then fix the things and we merge it in

src/main/java/net/sf/jabref/gui/maintable/MainTableSelectionListener.java
tobiasdiez: Always tricked by this equals vs == in Java....when do we finally port JabRef to the nice C# world 😄

Siedlerchr: Blame Java for no operator overloading...

src/main/java/net/sf/jabref/gui/fieldeditors/contextmenu/FieldTextMenu.java
```diff
@@ -87,13 +93,22 @@ private void maybeShowPopup(MouseEvent e) {
             if (protectedTermsMenu != null) {
                 protectedTermsMenu.updateFiles();
             }
+
+            boolean isDOIField = field.getFieldName().equals(FieldName.DOI);
+            doiMenuItem.setVisible(isDOIField);
+            boolean isDoiFieldEmpty = field.getText().isEmpty();
+            doiMenuItem.setEnabled(!isDoiFieldEmpty);
```
tobiasdiez: I find this version more readable 😸

tobiasdiez:
I fixed it in https://github.com/JabRef/jabref/commit/ffa3299d09616c8c779e0a080797c83f517a16fb.

boceckts:
@tobiasdiez thank you.

Issue #1993, opened 2016-09-16
Performance issue with new search


AEgit:
JabRef 3.7dev--snapshot--2016-09-16--master--629c3f0
windows 10 10.0 amd64 
Java 1.8.0_101

With the current master branch (master--629c3f0) there is a significant decrease in performance when using the "search" feature of Jabref compared to a previous master branch (master--4913665).

Before the update entering a search term led to an instantaneous search and nearly instantaneous results. Now I have to wait up to 3 sec every time I'm entering a search term or modifying it. As I'm using the search feature quite extensively this is quite annoying.

The database is quite large (>8000 entries), but this shouldn't be the problem as it worked before the update.

Unfortunately I forgot to save the previous installer - Could you provide me with the older master--4913665?

AEgit:
Ok, my current workaround is to use the latest release version (JabRef 3.6) which does not exhibit these performance problems.

lenhard:
I cannot confirm this on my ~2200 entry database. There search returns instant results. However, my database is also considerably smaller than yours. Also, I am not up to date with the recent changes in search. @koppor?

Also, @JabRef/developers What is the largest test file we have for testing performance?

koppor:
@bartsch-dev Can you have a look?

AEgit:
@lenhard 
If needed, I could send you an older (and smaller) copy of my database (still >6000 entries). It has the same issues as my current database, albeit Jabref seems a bit faster with the smaller database: Still, if I open this database with Jabref 3.6 and use the search feature I get instant results. If I open the same database with JabRef 3.7dev--snapshot--2016-09-16--master--629c3f0 I have to wait a couple of seconds for search results (and you also hear the CPU cooling unit spinning up, indicating that more processing power is needed (this is not the case with Jabref 3.6).

If I send you the database, however, I would ask you not to share it or make it publicly available, as it may contain sensible information (preferentially you would delete the file, once it is no longer needed).

lenhard:
@AEgit ok, thanks! You can send it to joerg.lenhard@kau.se Then I will see how search works with my machine and delete your file afterwards.

It would be cool if one day we had a huge file for JabRef for testing purposes. Maybe we could go for auto-generated entries, but then again randomized data will never be representative of real actual data.

tobiasdiez:
I can confirm the performance problems with a DB of 6000 entries. 
@lenhard there is one big file of mine in the admin rep.
@AEgit You don't need to send your file. But thanks for the offer.

[...]

PR #2043, opened 2016-09-22
Use parallel stream while searching

bartsch-dev:
With this the search should be much faster on platforms where multiple cores are available. (#1993)

### DIFF
--- a/src/jmh/java/net/sf/jabref/benchmarks/Benchmarks.java
+++ b/src/jmh/java/net/sf/jabref/benchmarks/Benchmarks.java
@@ -2,7 +2,6 @@ package net.sf.jabref.benchmarks;
 
 import java.io.IOException;
 import java.io.StringReader;
-import java.util.ArrayList;
 import java.util.List;
 import java.util.Random;
 import java.util.stream.Collectors;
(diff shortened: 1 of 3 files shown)


tobiasdiez:
Could you please create one benchmark which uses the normal stream and one for the parallel one, please also post the benchmark results so that we can see the improvement.

Siedlerchr:
The differences between parallel and normal stream will be only visible in
large database, there the parallel stream is more advanced, because the jvm
optimizes it

2016-09-22 23:00 GMT+02:00 Tobias Diez <notifications@github.com>:

bartsch-dev:
### stream 
```
Benchmark                             Mode  Cnt        Score        Error  Units
Benchmarks.htmlToLatexConversion     thrpt   20     3430,821 ±     12,570  ops/s
Benchmarks.inferBibDatabaseMode      thrpt   20    12196,397 ±    250,891  ops/s
Benchmarks.keywordGroupContains      thrpt   20     7367,579 ±    702,650  ops/s
Benchmarks.keywordGroupContainsWord  thrpt   20  8080932,143 ± 190154,353  ops/s
Benchmarks.latexToHTMLConversion     thrpt   20    62292,491 ±   1375,804  ops/s
Benchmarks.latexToUnicodeConversion  thrpt   20    59849,947 ±    298,067  ops/s
Benchmarks.parse                     thrpt   20       20,002 ±      0,809  ops/s
Benchmarks.search                    thrpt   20       19,986 ±      0,729  ops/s
Benchmarks.write                     thrpt   20       39,440 ±      0,232  ops/s
```

### parallel stream
``` 
Benchmark                             Mode  Cnt        Score       Error  Units
Benchmarks.htmlToLatexConversion     thrpt   20     3478,589 ±    58,752  ops/s
Benchmarks.inferBibDatabaseMode      thrpt   20    12098,187 ±   235,957  ops/s
Benchmarks.keywordGroupContains      thrpt   20     6694,197 ±    84,251  ops/s
Benchmarks.keywordGroupContainsWord  thrpt   20  7858681,820 ± 11459,928  ops/s
Benchmarks.latexToHTMLConversion     thrpt   20    60325,789 ±  3041,157  ops/s
Benchmarks.latexToUnicodeConversion  thrpt   20    59671,261 ±   535,001  ops/s
Benchmarks.parse                     thrpt   20       20,349 ±     0,154  ops/s
Benchmarks.search                    thrpt   20       63,016 ±     0,843  ops/s
Benchmarks.write
```

tobiasdiez:
Ok, that are good numbers. Perfect. Could you please keep the original `@Benchmark` with normal `stream` and add a new one for the parallel stream (just copy 1:1 the method). Then this can be merged.

bartsch-dev:
Created a separate Benchmark.

tobiasdiez:
Thanks :+1:

Issue #2060, opened 2016-09-25
No utf8 encoding while fetching from Medline


bernhard-kleine:
JabRef 3.6
windows 7 6.1 amd64 
Java 1.8.0_66

Steps to reproduce:

1. Ensure that you have utf8 as the preferred encoding 
2. import from pubmed/medline ID 27653033
3. notice that the import does not show the alphas and betas in the title and in the abstract.

I would expect utf8 encoding for the import since the medline has really a good way to deal with the alphas and betas, as well as any other greek letters or even japanese or other languages. While the utf8 encoding does well work in jabref, it is ?ommited? in the import. Please change. 
```

Siedlerchr:
Thanks for your report. I could not reproduce that issue in the latest development version. All Betas and Alphas etc are shown correctly. Database encoding is set to UTF8.
Please try again with the latest master from here: http://builds.jabref.org/master/

bernhard-kleine:
I just installed:
JabRef 3.7dev--snapshot--2016-09-25--master--693c705
windows 7 6.1 amd64 
Java 1.8.0_66
I have utf8 default encoding for the database: Importing PMID:27653033 still gives me in the title:
title = {Effects of 17Î²-estradiol and androgen on glucose metabolism in skeletal muscle.},
instead of
Effects of 17β-estradiol and androgen on glucose metabolism in skeletal muscle.
The entry 27653033 in firefox on the shows the β as well and also the alphas in the abstract. What do I have to do for proper import?

grimes2:
JabRef 3.7dev--snapshot--2016-09-25--master--693c705
windows 10 10.0 amd64 
Java 1.8.0_101

I can confirm: Î²

Siedlerchr:
Okay, I have an idea. And I will release a fix soon.

Siedlerchr:
When I am right it should be now correct:

![utf8medline](https://cloud.githubusercontent.com/assets/320228/18815448/f9e0dde6-8331-11e6-8a19-41bd4827237a.png)

You can try this snapshot:
http://builds.jabref.org/medlineutf8/

grimes2:
Works here too. 👍

bernhard-kleine:
works! Thank you for a fast correction!

bernhard-kleine:
BTW where is the line for the DOI search in this snapshot?

Siedlerchr:
[...]

PR #2061, opened 2016-09-25
Medlineutf8

Siedlerchr:
Seee #2060  for details


tobiasdiez:
Could you please add a test for the change. This is especially important for the medline importer since @zellerdev will probably rewrite it using the new fetcher interface and we don't want the bug to resurface.

Issue #2012, opened 2016-09-19
month fields in braces results in a warning


bernhard-kleine:
JabRef 3.6
windows 7 6.1 amd64 
Java 1.8.0_66

Steps to reproduce:
1. load the pubmed/medline entry ID 27623067
2. make sure that the % signs in the abstract are escaped to \%
3. the compilation with the actual biber gives a warning:
   Utils.pm:164> WARN - month field 'Sep' in entry 'PRR+16' is not an integer - this will probably not sort properly

This is due to (from texexchange):

Such month abbreviations are BibTeX macros and if you protect them in braces, they are not parsed as macros but as literal strings. You need, in your example:
MONTH = Sep
and not
MONTH = {Sep}

While all entries in Jabref are now in braces, this one should be not.

Removing the braces did the job. but how to do this in a 1800 entry bib-file

```

```

grimes2:
> Removing the braces did the job. but how to do this in a 1800 entry bib-file

With Quality > Cleanup entries
The field formatter is month: Normalize month.

Siedlerchr:
Thanks for your report! I can confirm the behavior. The cleanup as proposed by @grimes2  does the job.

As the cleanup works, I would suggest adding a case for the Integrity Checker, according to BibLaTeX manual, [section 2.3.9](http://ftp.fau.de/ctan/macros/latex/contrib/biblatex/doc/biblatex.pdf)

> The month feld is an integer feld. The bibliography style converts the month to
> a language-dependent string as required. For backwards compatibility, you may
> also use the following three-leter abbreviations in the month feld: jan, feb, mar,
> apr, may, jun, jul, aug, sep, oct, nov, dec. Note that these abbreviations are
> BibTeX strings which must be given without any braces or quotes

grimes2:
I would like to do the MonthChecker.

tobiasdiez:
Perfect!
And the month-cleanup should be performed automatically for the pubmed fetcher. @zellerdev are you willing to have a look at this and rewrite the MedlineFetcher using the new framework (no need to ask the user how much entries he wants to fetch, just take the default value from the API).

bernhard-kleine:
with the cleanup the warnings are gone.

lenhard:
If I understand all the comments here correctly, the issue is fixed with #2015 and we can close it.

tobiasdiez:
Not quite fixed yet; the medline fetcher still produces "wrong" month fields. @zellerdev will take care of it.

[...]

PR #2034, opened 2016-09-22
Fix bad web search error messages

bartsch-dev:
Fixes #1542

I rewrote the error messages when the fetchers fail (or added them).

PR #2066, opened 2016-09-25
Medline Fetcher using new infrastructure

zellerdev:
Rewrite the Medline fetcher.
Refs #2012 


Siedlerchr:
Please make sure that  creating the BibEntry is done in UTF-8 encoding  #2060 / #2061

src/main/java/net/sf/jabref/logic/importer/fetcher/MedlineFetcher.java
```diff
@@ -1,81 +1,74 @@
...
-        matcher = PART1_PATTERN.matcher(result);
-        result = matcher.replaceAll("\\+AND\\+");
-        matcher = PART2_PATTERN.matcher(result);
-        result = matcher.replaceAll("\\+AND\\+");
-        result = result.replace(" ", "+");
-        return result;
+    private static String toSearchTerm(String query) {
+        query = query.replaceAll(", ", " AND ");
+        query = query.replaceAll(",", " AND ");
+        return query;
```
tschechlovdev: Why don't just ``return query.replacAll(...).replaceAll(...)`` ?

src/main/java/net/sf/jabref/logic/importer/fetcher/MedlineFetcher.java
```diff
@@ -1,81 +1,74 @@
...
-        // This can probably be simplified using simple String.replace()...
-        String result = in;
-        Matcher matcher;
-        matcher = PART1_PATTERN.matcher(result);
-        result = matcher.replaceAll("\\+AND\\+");
-        matcher = PART2_PATTERN.matcher(result);
-        result = matcher.replaceAll("\\+AND\\+");
-        result = result.replace(" ", "+");
-        return result;
+    private static String toSearchTerm(String query) {
```
tschechlovdev: Maybe rename to replace comma

src/main/java/net/sf/jabref/logic/importer/fetcher/MedlineFetcher.java
```diff
@@ -119,134 +112,93 @@ public HelpFile getHelpPage() {
...
             return new ArrayList<>();
         }
     }
 
+    @Override
+    public Optional<BibEntry> performSearchById(String identifier) throws FetcherException {
+        String cleanQuery = identifier.trim().replace(';', ',');
+
+        if (cleanQuery.matches("\\d+[,\\d+]*")) {
+            List<BibEntry> bibs = fetchMedline(cleanQuery);
```
tschechlovdev: bibs? why not entries?

src/main/java/net/sf/jabref/logic/importer/fetcher/MedlineFetcher.java
```diff
@@ -119,134 +112,93 @@ public HelpFile getHelpPage() {
...
+            LOGGER.warn(e.getLocalizedMessage(), e);
             return new ArrayList<>();
         }
     }
 
+    @Override
+    public Optional<BibEntry> performSearchById(String identifier) throws FetcherException {
+        String cleanQuery = identifier.trim().replace(';', ',');
+
+        if (cleanQuery.matches("\\d+[,\\d+]*")) {
```
tschechlovdev: You could extract this as Pattern

src/main/java/net/sf/jabref/logic/importer/fetcher/MedlineFetcher.java
```diff
@@ -0,0 +1,201 @@
...
+    private static final String API_MEDLINE_FETCH = "http://eutils.ncbi.nlm.nih.gov/entrez/eutils/efetch.fcgi";
+    private static final String API_MEDLINE_SEARCH = "http://eutils.ncbi.nlm.nih.gov/entrez/eutils/esearch.fcgi";
+
+    private static final Pattern ID_PATTERN = Pattern.compile("<Id>(\\d+)</Id>");
+    private static final Pattern COUNT_PATTERN = Pattern.compile("<Count>(\\d+)<\\/Count>");
+    private static final Pattern RET_MAX_PATTERN = Pattern.compile("<RetMax>(\\d+)<\\/RetMax>");
+    private static final Pattern RET_START_PATTERN = Pattern.compile("<RetStart>(\\d+)<\\/RetStart>");
+
+
+    private static String replaceCommaWithAND(String query) {
```
boceckts: This seems to be a perfect method for the `StringUtil` class. Please move it if soemthing similar doesn't already exist.

src/main/java/net/sf/jabref/logic/importer/fetcher/MedlineFetcher.java
```diff
@@ -0,0 +1,201 @@
...
+
+    static class SearchResult {
+
+        int count;
+
+        int retmax;
+
+        int retstart;
+
+        String ids = "";
```
boceckts: Why are there spaces between the fields?

src/main/java/net/sf/jabref/logic/importer/fetcher/MedlineFetcher.java
```diff
@@ -0,0 +1,201 @@
...
+            }
+            LOGGER.info("Fetcher " + getName() + "found more than one result for identifier " + identifier
+                    + ". We will use the first entry.");
+            return Optional.of(entry.get(0));
+
+        }
+        return Optional.empty();
+    }
+
+    static class SearchResult {
```
boceckts: I don't really like this class, I would extract it to a new file and rename it to something like `MedlineSearchResult` or similar.

src/main/java/net/sf/jabref/logic/importer/fetcher/MedlineFetcher.java
boceckts: Fields should not have the default visibility.

src/main/java/net/sf/jabref/logic/importer/fetcher/MedlineFetcher.java
```diff
@@ -0,0 +1,199 @@
...
+
+    private static final Pattern ID_PATTERN = Pattern.compile("<Id>(\\d+)</Id>");
+    private static final Pattern COUNT_PATTERN = Pattern.compile("<Count>(\\d+)<\\/Count>");
+    private static final Pattern RET_MAX_PATTERN = Pattern.compile("<RetMax>(\\d+)<\\/RetMax>");
+    private static final Pattern RET_START_PATTERN = Pattern.compile("<RetStart>(\\d+)<\\/RetStart>");
+
+    /**
+     * Gets the initial list of ids
+     */
+    private SearchResult getIds(String term, int start, int pacing) {
```
tobiasdiez: As far as I can see, only the returned IDs are used and not the additional data in SearchResult. Could you change the result to List<Integer>. Please also rename term -> query

src/main/java/net/sf/jabref/logic/importer/fetcher/MedlineFetcher.java
```diff
@@ -0,0 +1,202 @@
...
+                    retstart = Integer.parseInt(retStartMatcher.group(1));
+                }
+                Matcher countMatcher = COUNT_PATTERN.matcher(inLine);
+                if (doCount && countMatcher.find()) {
+                    count = Integer.parseInt(countMatcher.group(1));
+                    doCount = false;
+                }
+            }
+        } catch (URISyntaxException | MalformedURLException e) { // new URL() failed
+            LOGGER.warn("Bad url", e);
```
tobiasdiez: Rethrow as FetcherException instead of failing silently.

src/main/java/net/sf/jabref/logic/importer/fetcher/MedlineFetcher.java
```diff
@@ -0,0 +1,202 @@
...
+     * @param query input to remove comma's
+     * @return input without comma's
+     */
+    private static String replaceCommaWithAND(String query) {
+        return query.replaceAll(", ", " AND ").replaceAll(",", " AND ");
+    }
+    /**
+     * Gets the initial list of ids
+     */
+    private void getIds(String term, int start, int pacing) {
```
tobiasdiez: Rename method to something more specific

src/main/java/net/sf/jabref/logic/importer/fetcher/MedlineFetcher.java
```diff
@@ -0,0 +1,202 @@
...
+
+            List<BibEntry> bibs = fetchMedline(ids);
+            entryList.addAll(bibs);
+
+            return entryList;
+        }
+        throw new FetcherException(Localization.lang("Input error") + ". " + Localization.lang("Please enter a comma separated list of Medline IDs (numbers) or search terms."));
+    }
+
+    static URL createSearchUrl(String term, int start, int pacing) throws URISyntaxException, MalformedURLException {
```
tobiasdiez: Make these methods private, nobody needs to know the details (not even a test 😄 )

src/main/java/net/sf/jabref/logic/importer/fetcher/MedlineFetcher.java
```diff
@@ -0,0 +1,202 @@
...
+    @Override
+    public List<BibEntry> performSearch(String query) throws FetcherException {
+        final int NUMBER_TO_FETCH = 50;
+        List<BibEntry> entryList = new LinkedList<>();
+
+        if (!query.isEmpty()) {
+            String searchTerm = replaceCommaWithAND(query);
+
+            // get the ids from entrez
+            getIds(searchTerm, 0, 1);
```
tobiasdiez: Why do you first get 1 entry and then 49 in a second request?

src/main/java/net/sf/jabref/logic/importer/fetcher/MedlineFetcher.java
```diff
@@ -0,0 +1,202 @@
...
+import org.apache.http.client.utils.URIBuilder;
+
+/**
+ * Fetch or search from Pubmed http://www.ncbi.nlm.nih.gov/sites/entrez/
+ */
+public class MedlineFetcher implements IdBasedFetcher, SearchBasedFetcher {
+
+    private static final Log LOGGER = LogFactory.getLog(MedlineFetcher.class);
+
+    private static final String API_MEDLINE_FETCH = "http://eutils.ncbi.nlm.nih.gov/entrez/eutils/efetch.fcgi";
```
tobiasdiez: If the URL and the pattern are just used in one place, then you can inline them in the corresponding method.

src/main/java/net/sf/jabref/logic/importer/fetcher/MedlineFetcher.java
```diff
@@ -0,0 +1,202 @@
...
+            }
+
+            getIds(searchTerm, 0, NUMBER_TO_FETCH-1);
+
+            List<BibEntry> bibs = fetchMedline(ids);
+            entryList.addAll(bibs);
+
+            return entryList;
+        }
+        throw new FetcherException(Localization.lang("Input error") + ". " + Localization.lang("Please enter a comma separated list of Medline IDs (numbers) or search terms."));
```
tobiasdiez: I wouldn't throw this exception if the query is empty. Just return an empty list in this case (also change the order: empty query -> empty list, otherwise fetch)

src/main/java/net/sf/jabref/logic/importer/fetcher/MedlineFetcher.java
```diff
@@ -0,0 +1,202 @@
...
+        return uriBuilder.build().toURL();
+    }
+
+    /**
+     * Fetch and parse an medline item from eutils.ncbi.nlm.nih.gov.
+     *
+     * @param id One or several ids, separated by ","
+     * @return Will return an empty list on error.
+     */
+    private static List<BibEntry> fetchMedline(String id) {
```
tobiasdiez: It appears to me that the id-based part can be refactored to use the `Parser`-helper (I think the class was called IdBasedParserFetcher). For this MedlineImporter needs to implement the Parser interface but this should be easy.

src/main/java/net/sf/jabref/logic/importer/fetcher/MedlineFetcher.java
```diff
@@ -0,0 +1,202 @@
...
+
+            // get the ids from entrez
+            getIds(searchTerm, 0, 1);
+
+            if (count == 0) {
+                LOGGER.warn(Localization.lang("No references found"));
+                return Collections.emptyList();
+            }
+
+            getIds(searchTerm, 0, NUMBER_TO_FETCH-1);
```
tobiasdiez: Please add a short comment about the general strategy. 
I.e. we call medline but only get a the pubmed ids as response -> fetch details for ids in a second request.

src/main/resources/l10n/Menu_vi.properties
```diff
@@ -2,133 +2,136 @@
...
 
+Abbreviate_journal_names_(ISO)=Viết_tắt_tên_các_tạp_chí_(ISO)
+Abbreviate_journal_names_(MEDLINE)=Viết_tắt_tên_các_tạp_chí_(MEDLINE)
+About_JabRef=Nói_về_JabRef
+Append_database=Nối_CSDL
+Autogenerate_BibTeX_keys=Tự_động_tạo_khóa_BibTeX
+Close_database=Đóng_CSDL
+Copy=Sao_chép
+Copy_\\cite{BibTeX_key}=Sao_chép\\trích_dẫn{khóa_BibTeX}
+Copy_BibTeX_key=Sao_chép_khóa_BibTeX
```
tobiasdiez: Why does this file has so many changes?

src/test/java/net/sf/jabref/logic/importer/fetcher/MedlineFetcherTest.java
```diff
@@ -0,0 +1,182 @@
...
+        entryEndharti.setField("volume", "16");
+        entryEndharti.setField("year", "2016");
+
+        bibEntryIchikawa = new BibEntry();
+        bibEntryIchikawa.setType(BibLatexEntryTypes.ARTICLE);
+        bibEntryIchikawa.setField("author", "Ichikawa-Seki, Madoka and Guswanto, Azirwan and Allamanda, Puttik and Mariamah, Euis Siti and Wibowo, Putut Eko and Igarashi, Ikuo and Nishikawa, Yoshifumi");
+        bibEntryIchikawa.setField("chemicals", "Antibodies, Protozoan, Antigens, Protozoan, GRA7 protein, Toxoplasma gondii, Protozoan Proteins");
+        bibEntryIchikawa.setField("citation-subset", "IM");
+        bibEntryIchikawa.setField("completed", "2016-07-26");
+        bibEntryIchikawa.setField("copyright", "Copyright © 2015 Elsevier Ireland Ltd. All rights reserved.");
```
tobiasdiez: Be so free and delete the "copyright" field :)

src/test/java/net/sf/jabref/logic/importer/fetcher/MedlineFetcherTest.java
```diff
@@ -0,0 +1,182 @@
...
+        bibEntryIchikawa.setField("country", "Netherlands");
+        bibEntryIchikawa.setField("created", "2015-09-26");
+        bibEntryIchikawa.setField("doi", "10.1016/j.parint.2015.07.004");
+        bibEntryIchikawa.setField("issn", "1873-0329");
+        bibEntryIchikawa.setField("issn-linking", "1383-5769");
+        bibEntryIchikawa.setField("issue", "6");
+        bibEntryIchikawa.setField("journal", "Parasitology international");
+        bibEntryIchikawa.setField("journal-abbreviation", "Parasitol Int");
```
tobiasdiez: Remove abbrevation field, JabRef has own functionality to toggle the abbreviation of journals.

src/test/java/net/sf/jabref/logic/importer/fetcher/MedlineFetcherTest.java
```diff
@@ -0,0 +1,182 @@
...
+        entryEndharti.setField("journal-abbreviation", "BMC Complement Altern Med");
+        entryEndharti.setField("keywords", "CAC; Dendrophtoe pentandra; IL-22; MPO; Proliferation; p53");
+        entryEndharti.setField("nlm-id", "101088661");
+        entryEndharti.setField("owner", "NLM");
+        entryEndharti.setField("pages", "374");
+        entryEndharti.setField("pmid", "27670445");
+        entryEndharti.setField("pubmodel", "Electronic");
+        entryEndharti.setField("pubstatus", "epublish");
+        entryEndharti.setField("revised", "2016-9-27");
+        entryEndharti.setField("status", "Publisher");
```
tobiasdiez: Do you know what this "status" field is? Looks to me like we can safely delete it.

src/main/resources/l10n/Menu_vi.properties
zellerdev: I don't really know it. I guess localizationUpdate?

tobiasdiez:
Please also make sure that the strange error message reported in https://github.com/JabRef/jabref/pull/2034#pullrequestreview-2529277 does not appear anymore. Thanks.

zellerdev:
I can't reproduce the Issue. But I changed the Exception to a more understandable one.

src/main/java/net/sf/jabref/logic/importer/fetcher/MedlineFetcher.java
```diff
@@ -116,45 +118,61 @@ public HelpFile getHelpPage() {
...
     @Override
-    public Optional<BibEntry> performSearchById(String identifier) throws FetcherException {
-        String cleanQuery = identifier.trim().replace(';', ',');
+    public URL getURLForID(String identifier) throws URISyntaxException, MalformedURLException, FetcherException {
+        return createFetchUrl(identifier);
+    }
+
+    @Override
+    public Parser getParser() {
+        Parser medlineParser = parserInputStream -> {
```
tobiasdiez: Could you please move the code to the `MedlineImporter` class i.e. implement the `Parser` interface there and here just return `new MedlineImporter()`

src/test/java/net/sf/jabref/logic/importer/fetcher/MedlineFetcherTest.java
```diff
@@ -195,9 +173,15 @@ public void testEmptyEntryList() throws Exception {
...
+    @Test
     public void testEmptyInput() throws Exception {
-        fetcher.performSearch("");
-        fail();
+        assertEquals(Collections.emptyList(),fetcher.performSearch(""));
+    }
+
+    private void cleanEntries(BibEntry entry){
+        entry.clearField(FieldName.ABSTRACT); //Remove abstract due to copyright
+        entry.clearField("copyright");
```
tobiasdiez: Why are these `clearFields` are required (except abstract, which is of course ok)? The test should actually check that the fields are correctly deleted.

src/main/java/net/sf/jabref/logic/importer/fetcher/MedlineFetcher.java
```diff
@@ -116,45 +118,61 @@ public HelpFile getHelpPage() {
     }
 
     @Override
-    public Optional<BibEntry> performSearchById(String identifier) throws FetcherException {
-        String cleanQuery = identifier.trim().replace(';', ',');
+    public URL getURLForID(String identifier) throws URISyntaxException, MalformedURLException, FetcherException {
+        return createFetchUrl(identifier);
```
tobiasdiez: inline `createFetchUrl` here and replace all calls to it with `getURLForID`?

src/main/java/net/sf/jabref/logic/importer/fetcher/MedlineFetcher.java
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    /**
+     * When using 'esearch.fcgi?db=<database>&term=<query>' we will get a list of ID's matching the query.
+     * Input: Any text query (&term)
+     * Output: List of UIDs matching the query
+     *
+     * @see <a href="https://www.ncbi.nlm.nih.gov/books/NBK25500/">www.ncbi.nlm.nih.gov/books/NBK25500/</a>
+     */
+    private void getPubMedIdsFromQuery(String query, int start, int pacing) throws FetcherException {
```
tobiasdiez: Please try to return here a list of IDs instead of calling `addId` (see my previous review)

zellerdev:
CI fails due to maintenance on ebook.de. So a few IsbnFetcher tests fail.

tobiasdiez:
👍 good job. can be merged from my point of view.

src/main/java/net/sf/jabref/logic/importer/fetcher/MedlineFetcher.java
```diff
@@ -0,0 +1,190 @@
...
+    private static final Pattern RET_START_PATTERN = Pattern.compile("<RetStart>(\\d+)<\\/RetStart>");
+
+    private int count;
+    private int retmax;
+    private int retstart;
+
+    /**
+     * Removes all comma's in a given String
+     *
+     * @param query input to remove comma's
```
koppor: Do not put `'` at plural-s. Plural of "comma" is "commas".

src/main/java/net/sf/jabref/logic/importer/fetcher/MedlineFetcher.java
```diff
@@ -0,0 +1,190 @@
...
+    private static final Pattern COUNT_PATTERN = Pattern.compile("<Count>(\\d+)<\\/Count>");
+    private static final Pattern RET_MAX_PATTERN = Pattern.compile("<RetMax>(\\d+)<\\/RetMax>");
+    private static final Pattern RET_START_PATTERN = Pattern.compile("<RetStart>(\\d+)<\\/RetStart>");
+
+    private int count;
+    private int retmax;
+    private int retstart;
+
+    /**
+     * Removes all comma's in a given String
```
koppor: `string` (with lower case s)

src/main/java/net/sf/jabref/logic/importer/fetcher/MedlineFetcher.java
koppor: Do it for all other things AND change comments to state that commas are replaced by " AND ".

src/main/java/net/sf/jabref/logic/importer/fileformat/MedlineImporter.java
```diff
@@ -80,7 +86,7 @@
  * check here for details on the format
  * https://www.nlm.nih.gov/bsd/licensee/elements_descriptions.html
  */
-public class MedlineImporter extends Importer {
+public class MedlineImporter extends Importer implements Parser {
```
koppor: I did not quite follow the new developments at the importers -> why does this Importer need to implement Parser, while the other Importers do not. Or do all importers also do that?

src/main/java/net/sf/jabref/logic/importer/fetcher/MedlineFetcher.java
```diff
@@ -0,0 +1,190 @@
...
+            String searchTerm = replaceCommaWithAND(query);
+
+            //searching for pubmed id's matching the query
+            List<String> idList = getPubMedIdsFromQuery(searchTerm, 0, NUMBER_TO_FETCH);
+
+            if (count == 0) {
+                LOGGER.info("No references found");
+            }
+
+            //pass the id list to fetchMedline to download them. like a id fetcher for mutliple id's
```
koppor: ... `ids` ... - Please check all code ;)

src/main/java/net/sf/jabref/logic/importer/fetcher/MedlineFetcher.java
```diff
@@ -0,0 +1,190 @@
...
+                Matcher retMaxMatcher = RET_MAX_PATTERN.matcher(inLine);
+                if (retMaxMatcher.find()) {
+                    retmax = Integer.parseInt(retMaxMatcher.group(1));
+                }
+                Matcher retStartMatcher = RET_START_PATTERN.matcher(inLine);
+                if (retStartMatcher.find()) {
+                    retstart = Integer.parseInt(retStartMatcher.group(1));
+                }
+                Matcher countMatcher = COUNT_PATTERN.matcher(inLine);
+                if (doCount && countMatcher.find()) {
```
koppor: Why not moving "doCount" one level up? Just do countMatcher if `doCount`. Further, you should comment that you only count at the first hit when count results are there, because ...can-you-give-a-reason...?

src/main/java/net/sf/jabref/logic/importer/fetcher/MedlineFetcher.java
```diff
@@ -0,0 +1,190 @@
...
+
+        if (query.isEmpty()) {
+            return Collections.emptyList();
+        } else {
+            String searchTerm = replaceCommaWithAND(query);
+
+            //searching for pubmed id's matching the query
+            List<String> idList = getPubMedIdsFromQuery(searchTerm, 0, NUMBER_TO_FETCH);
+
+            if (count == 0) {
```
koppor: Why doesn't one check for idList.isEmpty()? Why the global variable count?

koppor: Is the global variable "count" really needed anymore?

koppor:
Other than the minor comments: LGTM

zellerdev:
Now I use the counter to inform the user if more entries then the fetched ones are available. I completely removed retmax and retstart and use the default values. I also updated the comments. @koppor

tobiasdiez:
I preferred the version where 50 entries were returned 😄

src/main/java/net/sf/jabref/logic/importer/fetcher/MedlineFetcher.java
```diff
@@ -22,75 +22,67 @@
...
         List<String> idList = new ArrayList<>();
         try {
-            URL ncbi = createSearchUrl(query, start, pacing);
+            URL ncbi = createSearchUrl(query);
             BufferedReader in = new BufferedReader(new InputStreamReader(ncbi.openStream()));
             String inLine;
+            //Everything relevant is listed before the IdList. So we break the loop right after the IdList tag closes.
             while ((inLine = in.readLine()) != null) {
-
+                if (inLine.contains("</IdList>")) {
```
koppor: Uhoh, the service returns XML, but it is parsed using strings. @tschechlovdev Can you give us some indication about the effort for rewrite?

koppor:
@tobiasdiez Didn't want to change the semantics with my comments -> I just read the code. Should @zellerdev revert this part of his changes? For me, both is fine as long as the fetcher works 😇

Siedlerchr:
I always thought the limit of 50 was some server-side thing?

tobiasdiez:
I think the part witch changed the number of fetched entries from 20 to 50 can be reverted but removing the local variables was indeed a good observation. I think even the count variable is not necessary and maybe can be replaced by `list of fetched bib entries.size()`.

src/main/java/net/sf/jabref/logic/importer/fetcher/MedlineFetcher.java
tschechlovdev: If I see right, then using a StAX parser should be good to go here.

src/main/java/net/sf/jabref/logic/importer/fetcher/MedlineFetcher.java
```diff
@@ -0,0 +1,211 @@
...
+            ParserResult result = new MedlineImporter().importDatabase(
+                    new BufferedReader(new InputStreamReader(data.getInputStream(), StandardCharsets.UTF_8)));
+            if (result.hasWarnings()) {
+                LOGGER.warn(result.getErrorMessage());
+            }
+            List<BibEntry> resultList = result.getDatabase().getEntries();
+            resultList.forEach(this::doPostCleanup);
+            return resultList;
+        } catch (URISyntaxException | IOException e) {
+            throw new FetcherException(e.getLocalizedMessage(), e);
```
matthiasgeiger: The messages of (most) API exceptions are not localized. Thus, calling `e.getLocalizedMessage()` will always return English messages. Provide an own localized message, please.

src/main/java/net/sf/jabref/logic/importer/fetcher/MedlineFetcher.java
```diff
@@ -203,8 +203,10 @@ private URL createSearchUrl(String term) throws URISyntaxException, MalformedURL
             List<BibEntry> resultList = result.getDatabase().getEntries();
             resultList.forEach(this::doPostCleanup);
             return resultList;
-        } catch (URISyntaxException | IOException e) {
-            throw new FetcherException(e.getLocalizedMessage(), e);
+        } catch (URISyntaxException | MalformedURLException e) {
+            throw new FetcherException("Error while generating fetch URL", Localization.lang("Error while generating fetch URL"), e);
+        } catch (IOException e) {
+            throw new FetcherException("Error while fetching from Medline", Localization.lang("Error while fetching from Medline"), e);
```
matthiasgeiger: There's already "Error while fetching from %0" which you can reuse.

zellerdev: Thats right. I guess I missed that. :sweat_smile:

matthiasgeiger:
Can you please have a look whether #2172 is fixed with this PR?

zellerdev:
@tschechlovdev is preparing a fix. He wrote the MedlineImporter in the first place.

koppor:
Tests fail, but I could fix them locally to get the fix into the master branch (0e5e15d65fba4116a5672 and 966e1967046db466da0). Please consider rewriting the fetcher using xsd.

PR #2068, opened 2016-09-26
Remove content selectors

tobiasdiez:
This PR removes the "content selectors".
According to our help page, they are supposed "to store a selection of often-used words or phrases that you use often in your database". But exactly the same functionality is already there with the autocompletion. Thus I don't see any advantage of having the content selectors. 

I did some background check and was not able to find a single bib file in the web which uses the content selectors. Moreover, the most discussions about them were from around 2004. The only feature request related to content selectors is the still open https://sourceforge.net/p/jabref/feature-requests/669/ which requests to store certain phrases that are often used in reviews. 
I kind of see the use case there, but if this feature is really wished I would opt for a review template.
To be honest, I doubt that anybody even understood how to add phrases in the "Manage content selector" dialog, not to mention that probably nobody looked under "content selectors" for such a feature. 

@JabRef/developers do you know of a good use case for this feature??

(will fix the localization stuff before merge)


stefan-kolb:
Auto-completion is basically broken but you can still remove this feature anyway 👍

Siedlerchr:
Seems like a good idea. Never heard of them before

lenhard:
I, too, cannot add an additional sensible use case, so removing is fine from my point of view. 

Note that, next to the failing localization tests, you also seem to have another failed test regarding to parsing content selectores

Issue #1808, opened 2016-08-22
Import Preferences don't set font size


grimes2:
JabRef 3.6dev--snapshot--2016-08-22--master--74afbb7
windows 10 10.0 amd64 
Java 1.8.0_101

Steps to reproduce:

1. Change Preferences > Appearance > Set table font to font size 14
2. Export preferences (xml)
3. Change Preferences > Appearance > Set table font to font size 10
4. Import preferences (xml)
5. fontsize is still 10 (should be 14)

Remark:
font size is saved in xml:
`<entry` key="fontSize" value="14"/>

oscargus:
I do not know if it is related, but if one just does the following:

1. Press "Set table font" and change something. Press "OK".
2. Press "Set table font" and everything is back to before the change in 1.

I guess this is related to the preference change not actually happening before pressing "OK" in the Preferences dialog and the font dialog is initialized from `GUIGlobals.currentFont`.

grimes2:
Import Preferences also doesn't import "push to application Lyx/Kile", it is set to "push to Emacs" instead.

koppor:
Thank you for reporting this issue. We think, that is already fixed in our development version and consequently the change will be included in the next release.

We would like to ask you to use a development build from https://builds.jabref.org/master.

PR #2069, opened 2016-09-26
Fix changing the font size not working when importing preferences

Braunch:
I initialized the FontSelectorDialog with the variable that keeps changes made in the dialog instead of the current Globals. This seems to fix both issues named here #1808 


src/main/java/net/sf/jabref/gui/preftabs/AppearancePrefsTab.java
```diff
@@ -182,7 +182,7 @@ public AppearancePrefsTab(JabRefPreferences prefs) {
         overrideFonts.addActionListener(e -> fontSize.setEnabled(overrideFonts.isSelected()));
 
         fontButton.addActionListener(
-                e -> new FontSelectorDialog(null, GUIGlobals.currentFont).getSelectedFont().ifPresent(x -> usedFont = x));
+                e -> new FontSelectorDialog(null, usedFont).getSelectedFont().ifPresent(x -> usedFont = x));
```
simonharrer: The problematic part may be the `ifPresent`, as I am not sure if you can reassign a variable `usedFont` within a lambda `x -> usedFont = x` so that it `x` will be accessible from the outside.

Siedlerchr: It works, when the variable usedFont is a field: http://stackoverflow.com/questions/25055392/lambdas-local-variables-need-final-instance-variables-dont

Braunch:
I made the PreferenceDialog store the settings when preferences are exported so that the exported preferences are the same as the ones shown in the dialog. I thought about doing the same when the "show preferences" button is pressed. But I am not convinced that the user would expect that the changes made in the dialog are stored when he wants to show them. Maybe he just wants to compare stuff and then accidentally changes the settings.

Braunch:
The preference dialog now gets disposed when preferences has been imported. This prevents that the user overrides the imported preferences.

src/main/java/net/sf/jabref/gui/preftabs/PreferencesDialog.java
```diff
@@ -219,15 +208,38 @@ public PreferencesDialog(JabRefFrame parent) {
...
         frame.removeCachedEntryEditors();
-        Globals.prefs.updateEntryEditorTabList();
+        prefs.updateEntryEditorTabList();
+    }
+
+    private void storeAllSettings(){
+        // First check that all tabs are ready to close:
+        int count = main.getComponentCount();
+        Component[] comps = main.getComponents();
+        for (int i = 0; i < count; i++) {
```
boceckts: If you don't need the counter variable please use for each. `for (Component currentComponent : comps)`.

src/main/java/net/sf/jabref/gui/preftabs/PreferencesDialog.java
```diff
@@ -219,15 +208,38 @@ public PreferencesDialog(JabRefFrame parent) {
...
+            if (!((PrefsTab) comps[i]).validateSettings()) {
+                return; // If not, break off.
+            }
+        }
+        // Then store settings and close:
+        for (int i = 0; i < count; i++) {
```
boceckts: Same as last comment.

src/main/java/net/sf/jabref/gui/preftabs/PreferencesDialog.java
```diff
@@ -239,27 +251,32 @@ public OkAction() {
...
+            Optional<Path> path = dialog.saveNewFile();
+
+            path.ifPresent(exportFile -> {
+                try {
+                    storeAllSettings();
+                    prefs.exportPreferences(exportFile.toString());
+                } catch (JabRefException ex) {
+                    LOGGER.warn(ex.getMessage(), ex);
+                    JOptionPane.showMessageDialog(PreferencesDialog.this, ex.getLocalizedMessage(),
+                            Localization.lang("Export preferences"), JOptionPane.ERROR_MESSAGE);
```
boceckts: The Exception should be treated consistently. Either use `LOGGER.error` or `JOptionPane.WARNING_MESSAGE`.

src/main/java/net/sf/jabref/gui/preftabs/PreferencesDialog.java
```diff
@@ -219,15 +208,38 @@ public PreferencesDialog(JabRefFrame parent) {
...
 
         frame.removeCachedEntryEditors();
-        Globals.prefs.updateEntryEditorTabList();
+        prefs.updateEntryEditorTabList();
+    }
+
+    private void storeAllSettings(){
+        // First check that all tabs are ready to close:
+        int count = main.getComponentCount();
+        Component[] comps = main.getComponents();
```
boceckts: I thought this can only be tabs? If so please rename variable to `preferenceTabs`.

src/main/java/net/sf/jabref/gui/preftabs/PreferencesDialog.java
```diff
@@ -222,15 +222,15 @@ private void updateAfterPreferenceChanges() {
     private void storeAllSettings(){
         // First check that all tabs are ready to close:
         int count = main.getComponentCount();
```
boceckts: You don't need this anymore. Once removed it's fine by me.

src/main/java/net/sf/jabref/gui/preftabs/PreferencesDialog.java
```diff
@@ -42,6 +42,8 @@
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
+import static net.sf.jabref.Globals.prefs;
```
oscargus: I'm not sure if we want to import this statically. I see the point, but it will be quite confusing compared to all other files where it is accessed as `Globals.prefs`...

Braunch: Done

Braunch:
Can this be merged?

src/main/java/net/sf/jabref/gui/preftabs/PreferencesDialog.java
Braunch: Done

src/main/java/net/sf/jabref/gui/preftabs/PreferencesDialog.java
Braunch: Done

src/main/java/net/sf/jabref/gui/preftabs/PreferencesDialog.java
Braunch: Done

src/main/java/net/sf/jabref/gui/preftabs/PreferencesDialog.java
Braunch: Done

src/main/java/net/sf/jabref/gui/preftabs/PreferencesDialog.java
Braunch: Done

PR #2071, opened 2016-09-26
Tooltips now working in the main table

bartsch-dev:
Fixes: https://github.com/koppor/jabref/issues/160.

Tool tips in the main table now work as expected.

If the content of the cell is too big to be rendered in the cell itself it will be shown as a tool tip. What is more important, the original tool tip or the content?


tobiasdiez:
Good work.

Issue #2064, opened 2016-09-25
Regression with Unnecessary other field tab (#2002)


grimes2:
I've observed a regression with #2002. Under certain conditions other fields are not displayed. It's better to undo the commit. Should I add a new pull request to revert the changes?

Siedlerchr:
Okay, I would say: First create a new PR. Then you could investigate this issue in more detail. Maybe you find the conditions and can add a check then for it.

grimes2:
Steps to reproduce:

1. File > New Biblatex database
2. Bibtex > New entry > Article (First entry)
3. Bibtex > New entry > Article (Second entry)
4. (Second entry) language field: a
5. File > Switch to Bibtex mode
6. Restart Jabref
7. Doubleclick on First entry
8. Click on Second entry
9. No other field tab with language field

Siedlerchr:
Okay, I merged your undo PR in. Would you like to investigate if you could find a solution for the problem?

grimes2:
Investigating the issue shows, that this is a more general problem. Selecting another entry in main table doesn't update the Entry Editor properly. A similar problem as in Invalid field content lost when selecting different entry #1866.

grimes2:
I found no solution for the problem. But I have a proposal for the "other fields" tab in Bibtex mode. This tab could show all the pure Biblatex fields of a given type. So the "other fields" tab has many fields. A common case is, that some fetched entries contain Biblatex fields. In this case, these fields are reliably displayed also in Bibtex mode.
> An entry can also contain other fields, which are ignored by standard bibliography styles.

(Bibtexing - official spec)

grimes2:
This issue can be closed. The incorrect cache mechanism was removed in #2075. Other fields are now displayed correct.

PR #2075, opened 2016-09-27
Other fields fix and changes

grimes2:
These are the claimed improvements:
- Removed optional fields from `other fields` (BibTeX)
- Removed deprecated fields from `other fields` (BibLaTeX)

See: #2064


Siedlerchr:
I'm sorry, but I am not sure if I get/understand your changes, because what does "OtherFields" refer to?

I have the following tabs. in BibTex:
`Required, Optional, General, Abstract, Review, Source`
and in Biblatex:
`Required, Optional1, Optional2, Deprecated, General, Abstract, Review, Source`

Deprecated fields exist for bibtex compatibility, when you import an entry you often get in bibtex. 
So these fields are somehow still required.

grimes2:
Tab Other fields shows the remaining fields:
otherfields

grimes2:
I think I found the bug in #2064. It was not a regression. The new entry editor was not instantiated on changing an entry of the same type. So some "other fields" were not displayed.

grimes2:
Plese set label "ready-for-review". Thanks.

grimes2:
The idea was originally to have a better performance. So the entry editor is retained for the same type. This is not a good idea. But the entry editor needs to be build completetely new in all cases, because the new entry can have some additional (layout changing) other fields.

tobiasdiez:
So the whole `entryEditors` list can be deleted, right? (because its sole purpose was to cache the entry editor for each type). Moreover, there is a similar code in the showEntry method which differentiates between cached vs not-yet-cached.

grimes2:
```
//To indicate which entry is currently shown.
private final Map<String, EntryEditor> entryEditors = new HashMap<>();
```
I think, `entryEditors` stores all the old entry editors and the new entry editor. So I'm not sure, that the sole purpose is to store an entry editor of each type. The similar code in `showEntry` should also be adjusted, but there is currently no bug report for it.

grimes2:
I've checked the code. @tobiasdiez you are right. The whole entry editor caching (for each type) can be removed. I have no performance issues with my databases.

tobiasdiez:
I noticed a very small latency if you quickly switch from editing one entry to another (for example with the arrow keys). But as I don't think this is a big deal, I'll merge this PR now.

grimes2:
Yes, fast scrolling of the entries causes a small latency and moderate CPU usage. But I think, this is no problem. I have an idea for caching the entry editor: entry editor A -> B -> A. A can be cached. But this takes real effect only on heavy usage of the database.

koppor:
Or cache based on the "properties" of the entry. One could generate a
unique ID for each entry property. -- Property: entry type, other fields.
(future: hadPDFAttached) @braunch

tobiasdiez:
The problem is generating the textfields. So especially in biblatex mode there are around 80 fields which are generated every time. I would not cache the whole editor but just generate the textboxes when the tab is focused.

grimes2:
The other field tab is optional (No field -> No tab). I would prefer to create the other fields tab always (No field -> Empty tab). The advantage is, that every entry editor has always the same layout of the tabs. This preserves the active (other fields) tab on changing the entry.

Issue #1886, opened 2016-08-29
Google search issue in version 3.6 [Fixed in DevBuilds]


swamptromp:
Hello,

Although JabRef 3.6 resolved issues with Google Search, I am still having problems. Google search worked for me for a few hours (I am trying to set up my entire library, so I have been using the search function a lot), but suddenly stopped fetching any search results. In the meantime, Springer and other searches continue to work - it sounds just like the 3.5 issues described on the forum. The first time this happened, I reinstalled JabRef 3.6, and Google search started working again. The problem came up again, though, and this time reinstalling isn't solving it. Is anyone else still having Google search issues? Any suggestions?

Thanks for your help!

**// Edit by @matthiasgeiger: The problem with version 3.6 is fixed in the current development builds which are available at https://builds.jabref.org/master** (for details see discussion below)

Siedlerchr:
Hello @swamptromp, thanks for your report,   I guess you ran into the Google limit. Google blocks your IP for a while, if you do too many automated requests (spam/bot protection) (as noted here)
https://github.com/JabRef/jabref/issues/1694#issuecomment-238435978

A solution would be to use the Browser Addon JabFox to import the entries

There is currently nothing we can do and I am not sure if we could display a more specific dialog/error. However, I will create a new issue for that.

swamptromp:
Ah, thanks so much for this @Siedlerchr! Great to know, glad it isn't a longer-term issue.

swamptromp:
Hi, I have one more question. I verified my identity via the Google Scholar website, and although I can now get results through Google Scholar, JabRef still won't return google search requests, even after reinstalling it. Does this problem just go away after a set amount of time?

Siedlerchr:
@swamptromp  As it works for me, I can only try to give an explanation. I still think that your JabRef is blocked (uses a specific user agent)
Your Google Scholar settings in the browser (e.g. the fact that you are successfully authorized) are stored in a cookie.  However, JabRef does currently not support any form of authentication and therefore is not able to store your account info.

Maybe we can add this for the future which would resolve the problems a bit. In the meantime I would suggest using other fetchers or to manually import it. If you have a DOI or ISBN for a paper/book/..., try to use the DOI to Bibtex/ISBN fetchers, as they directly resolve the number to a bibtex entry.

tobiasdiez:
@JabRef/developers did we tried to get clearance/allowance from google for the userargent = JabRef? This might be worth a try.

[...]

PR #2082, opened 2016-09-28
Quick fix google scholar entry fetching

matthiasgeiger:
Google Scholar fetching was broken again (see #1886)

With this fix at least getting the first 10 search hits is possible again. 

Configuration is no longer possible in the current form and google generally limits the responses (per page) to 20 hits (however, even using this will cause a captcha challenge for JabRef).

As only 10 hits are allowed a rewrite to the new FetcherInfrastructure should now be possible (thus, omitting the 2-step approach).

  - [ ] Update of http://help.jabref.org/en/GoogleScholar might be useful to indicate that only 10 hits are (currently) shown

### DIFF
--- a/src/main/java/net/sf/jabref/gui/importer/fetcher/GoogleScholarFetcher.java
+++ b/src/main/java/net/sf/jabref/gui/importer/fetcher/GoogleScholarFetcher.java
@@ -2,7 +2,6 @@ package net.sf.jabref.gui.importer.fetcher;
 
 import java.io.IOException;
 import java.io.StringReader;
-import java.io.UnsupportedEncodingException;
 import java.net.MalformedURLException;
 import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
@@ -12,7 +11,6 @@ import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
-import java.util.stream.Collectors;
 
 import javax.swing.JLabel;
 import javax.swing.JPanel;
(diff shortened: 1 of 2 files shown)


src/main/java/net/sf/jabref/gui/importer/fetcher/GoogleScholarFetcher.java
```diff
@@ -193,13 +165,21 @@ private static void runConfig() throws IOException {
...
-        String cont = new URLDownload(urlQuery).downloadToString(StandardCharsets.UTF_8);
-        Matcher m = GoogleScholarFetcher.BIBTEX_LINK_PATTERN.matcher(cont);
+        String cont = URLDownload.createURLDownloadWithBrowserUserAgent(urlQuery).downloadToString(StandardCharsets.UTF_8);
+
+        Matcher m = GoogleScholarFetcher.SCHOLAR_ID_PATTERN.matcher(cont);
         int lastRegionStart = 0;
 
         while (m.find()) {
-            String link = m.group(1).replace("&amp;", "&");
-            link = link+"&oe=utf-8"; // append param 'oe=utf-8' to tell google to serve UTF-8 encoded results
```
tobiasdiez: The UTF-8 string is no longer required?

src/main/java/net/sf/jabref/gui/importer/fetcher/GoogleScholarFetcher.java
```diff
@@ -193,13 +165,21 @@ private static void runConfig() throws IOException {
...
+
+            String citationsPageURL = CITATIONS_PAGE_URL_BASE+m.group(1)+CITATIONS_PAGE_URL_SUFFIX;
```
tobiasdiez: Please use Apache's URIBuilder instead of String concatenation.

src/main/java/net/sf/jabref/gui/importer/fetcher/GoogleScholarFetcher.java
```diff
@@ -193,13 +165,21 @@ private static void runConfig() throws IOException {
...
+
+            String citationsPage = URLDownload.createURLDownloadWithBrowserUserAgent(citationsPageURL).downloadToString(StandardCharsets.UTF_8);
+
+            Matcher citationPageMatcher = GoogleScholarFetcher.BIBTEX_LINK_PATTERN.matcher(citationsPage);
```
tobiasdiez: It is not possible to get the BibTeX Url directly from the id?
For example, directly access https://scholar.googleusercontent.com/scholar.bib?q=info:b2pGeL14LLMJ:scholar.google.com/&output=citation&scisig=AAGBfm0AAAAAV-vITSVVNk7OSER8S_LFaMSElM9jnUcv&scisf=4&ct=citation&cd=-1&hl=en

matthiasgeiger: No, as this "scisig" is necessary and seems to be changing.

src/main/java/net/sf/jabref/gui/importer/fetcher/GoogleScholarFetcher.java
matthiasgeiger: Yes, as I am using a Firefox Useragent UTF-8 is automatically delivered.

src/main/java/net/sf/jabref/gui/importer/fetcher/GoogleScholarFetcher.java
matthiasgeiger: Mhmm. I don't think this is that much an improvement as still some string concatenation is needed for the param "q" with `"info:"+id+":scholar.google.com/"`

src/main/java/net/sf/jabref/gui/importer/fetcher/GoogleScholarFetcher.java
tobiasdiez: That's unfortunate. 
I did some research and I think I found something. 
Google stores the setting to display bib-links in a cookie. Sending the cookie `GSP=IN=d192d757fd09588a+7e6cc990821af64:CF=4` along the first query directly shows the Import into Bibtex link, 
````
<a href="https://scholar.googleusercontent.com/scholar.bib?q=info:b2pGeL14LLMJ:scholar.google.com/&amp;output=citation&amp;scisig=AAGBfm0AAAAAV-vafm6NWi20weGxxou9W2xi8GzZ8YCf&amp;scisf=4&amp;ct=citation&amp;cd=0&amp;hl=en" class="gs_nta gs_nph">Import into BibTeX</a>
````

Appearently, the part after GSP=IN=... is not that important since my real id had 63 at the end instead of 64.

PR #2085, opened 2016-09-28
Localization script: Simplify update output and fix printing of special characters

bartsch-dev:
I really don't like the missing type safety in python ...

Now the problematic translations get printed to the terminal correctly.
I simplified the output while running the update task; printing only the changes.
I also added a new checkbox to the PR-template to make it easier for new people.


Siedlerchr:
You should switch do Python 3.x, there is everything default in Unicode..

bartsch-dev:
Maybe I'll do that ...

PR #2087, opened 2016-09-28
Fix NPE in the PreviewPanel

bartsch-dev:
Fixed a NPE which rarely appears on Window, more often on linux (b/c of the `invokeLater` in line 299)


tobiasdiez:
LGTM :+1:

Issue #2078, opened 2016-09-27
Namespace in jaxb should not rely on internal java com.sun classes


Siedlerchr:
Follow up for the problem in #1989 
@tschechlovdev 

Atm the commit was reverted in #2077 

I found this discussion here: 

http://stackoverflow.com/questions/2326107/what-happened-to-jaxbs-namespaceprefixmapper-in-jdk6u18

tschechlovdev:
So if I get it right, I should not use the NamespacePrefixMapper any more. There is an interface for an [JAXBNamepsacePrefixMapper](http://mvnrepository.com/artifact/com.googlecode.jaxb-namespaceprefixmapper-interfaces/JAXBNamespacePrefixMapper/2.2.4). Can this be used?

Siedlerchr:
Hm, from what I see you are using a very old version of jaxb and when I see it correctly, since several versions it is already included in java8 
https://docs.oracle.com/javase/8/docs/api/javax/xml/bind/JAXB.html

And also have a look here,https://jaxb.java.net/2.2.11/docs/release-documentation.html#a-2-2-11 
There seems to be a sample somewhere which inlcudes some namespace stuff

Look also here, this is a newer gradle plugin

https://github.com/ewerk/gradle-plugins/tree/master/jaxb2-plugin

tschechlovdev:
Hm ok. If I see right then it's the easiest way is to add a ``@XmlNs`` to the package-info.java.
That would look like the following:
```    
// Package level annotation
    @javax.xml.bind.annotation.XmlSchema (
      xmlns = {
        @javax.xml.bind.annotation.XmlNs(prefix = "po",
                   namespaceURI="http://www.example.com/myPO1"),

        @javax.xml.bind.annotation.XmlNs(prefix="xs",
                   namespaceURI="http://www.w3.org/2001/XMLSchema")
      )
    )
```
This seems to work for me. 
I've also found a [maven-plugin](https://java.net/projects/jaxb2-commons/pages/Namespace-prefix) that seems to do this. But I coulnd't find one for gradle.

Siedlerchr:
Yeah, I would go with the Namespace annoation, if it works, its's good.

tschechlovdev:
@Siedlerchr This can be closed now, right?

PR #1989, opened 2016-09-15
Rewrite mods exporter and add test for it

tschechlovdev:
Regarding: #1920 
Rewrite the MODS exporter with a JAXB parser. 
Details on the format and the newest xml schema can be found [here] (http://www.loc.gov/standards/mods/).
I've used the newest schema (version 3.6).
This PR uses the same gradle script and schema as in #1942 . 

- [ ] internal QS
- [x] Tests created for changes
- [x] Manually tested changed features in running JabRef

PR #2088, opened 2016-09-28
[WIP] Rewrite mods exporter and add test

tschechlovdev:
Follow up of #1989 and regarding #2078.
Rewrites the mods exporter with a jaxb parser. 

In comparison to the last PR, I do not longer rely on the NamespacePrefixMapper. I added an annotation in the package-info.java which also does the namespace prefix mapping. Only problem is that this annotation cannot be automatically generated by a gradle or ant script, that's why added the package-info.java to src/main/java/net/...../mods.


Siedlerchr:
Okay, now all seems to be executed fine on travis and codecov is increased

tschechlovdev:
Yeah codecov was increased but the class ``modsExportFormatTestFiles`` was still not run, so this issue still remains.

Siedlerchr:
Codecov itself does not run any tests, it relies on the results from travis:
http://docs.codecov.io/docs/ci-service-relationship
Mabye it's a travis configuration thing. No idea. It would be nice if you could search for information about that

tschechlovdev:
Yeah, I will try to find a solution for this.

PR #2091, opened 2016-09-29
[WIP] Speed up search

bartsch-dev:
Refers to: #1993 

At the request of @koppor I speed up the search by using a second map of fields inside BibEntry which stores the field latex free.
I also internalize these fields and set the necessary JVM arguments in install4j (I hope those are right).

#### TODO: 
- Identifiy all the uses of the LatexToUnicodeFormatter and replace it.

### DIFF
--- a/jabref.install4j
+++ b/jabref.install4j
@@ -750,4 +750,9 @@ return true;</string>
   </mediaSets>
   <buildIds buildAll="true" />
   <buildOptions verbose="false" faster="false" disableSigning="false" disableJreBundling="false" debug="false" />
+  <jvmArguments>
+      <arg>-XX:+UseG1GC</arg>
+      <arg>-XX:+UseStringDeduplication</arg>
+      <arg>-XX:StringTableSize=1000003</arg>
+  </jvmArguments>
 </install4j>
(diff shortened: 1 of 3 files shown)


koppor:
This PR breaks our architecture. Do we need a more sophisticated solution?

```
net.sf.jabref.ArchitectureTests > fistPackageIsIndependentOfSecondPackage[7 -- is net.sf.jabref.model independent of net.sf.jabref.logic?] FAILED

    java.lang.AssertionError: expected:<[]> but was:<[src/main/java/net/sf/jabref/model/entry/BibEntry.java]>
```

lenhard:
@koppor "Could have told you so" ;-) The architecture issue was pretty obvious here. Let us talk this through before implementing a quick shot, I have a number of solutions on my mind to these architecture issues as well, as well as for the rest.

Let us also keep in mind that the VM args might not be quite be as portable or effective for non-Oracle JVMs.

oscargus:
Just use `LatexToUnicode` instead of `LatexToUnicodeConverter`...

lenhard:
Indeed. At the devcall, we decided to implement basically everything, i.e., cached LaTeX-free fields in `BibEntry`, String internalization, additional command-line args, and some pre-compilation of regular expression.

I'll hijack this PR and create a new one shortly, albeit not today.

lenhard:
I have moved over the ideas here into #2102 and build some more on top. Hence, I'll close this PR.

PR #2097, opened 2016-09-29
RTFchars fix

mairdl:
RTFcharsTest had some ignored test. So I looked into it and made the necessery changes. I added some of the tested but missing special characters. I also modified the `expected` and `actual` vaulues where needed (because to my understanding of this topic, some values were wrong). Since I am not 100% sure if everything is correct, comments are appreciated. 

Also I added "translator" which takes the unicode and transforms it back into its root letter.
E.g. `é - > \\u233e`

My final change is that I removed duplicate keys inside the RTFCharMap.

- [x] Tests work for changes
- [x] internal qs
- [x] double check mappings


mairdl:
looks like i need some more testscases for my changes

src/main/java/net/sf/jabref/logic/layout/format/RTFChars.java
```diff
@@ -199,4 +199,100 @@ private StringInt getPart(String text, int i, boolean commandNestedInBraces) {
...
+            return "TH";
+        if(223 == c)
+            return "ss";
+        if(161 ==c)
+            return "!";
+        return "?";
+    }
 }
+//if( <= c && c <= )
+//        return "";
```
boceckts: Remove out commented code.

src/main/java/net/sf/jabref/logic/layout/format/RTFChars.java
```diff
@@ -199,4 +199,100 @@ private StringInt getPart(String text, int i, boolean commandNestedInBraces) {
         // the wrong "}" at the end is removed by "format(res)"
         return new StringInt(format(res), part.length());
     }
+
+    private String transformSpecialCharacter(long c) {
```
boceckts: Please add a javadoc comment.

src/main/java/net/sf/jabref/logic/layout/format/RTFChars.java
```diff
@@ -199,4 +199,100 @@ private StringInt getPart(String text, int i, boolean commandNestedInBraces) {
...
+            return "OE";
+        if(339 == c)
+            return "oe";
+        if(222 == c)
+            return "TH";
+        if(223 == c)
+            return "ss";
+        if(161 ==c)
+            return "!";
+        return "?";
```
boceckts: I don't  like the multiple return statements and no brackets. Please use a variable which gets returned once in the end of the method and use `if else` with brackets.

Siedlerchr:
What about a switch Statement with case?
Am 01.10.2016 2:20 nachm. schrieb "Tobias Boceck" <notifications@github.com

src/main/java/net/sf/jabref/logic/layout/format/RTFChars.java
mairdl: yes sure

src/main/java/net/sf/jabref/logic/layout/format/RTFChars.java
mairdl: the way it is implemented now it has a better performance 👍

mairdl:
I also thought about switch case, but i came to the conclusion it gets way to big 🎅

src/main/java/net/sf/jabref/logic/util/strings/RtfCharMap.java
```diff
@@ -9,11 +9,11 @@ public RtfCharMap () {
         put("`i", "\\'ec");
         put("`o", "\\'f2");
         put("`u", "\\'f9");
-        put("?a", "\\'e1");
-        put("?e", "\\'e9");
-        put("?i", "\\'ed");
-        put("?o", "\\'f3");
-        put("?u", "\\'fa");
+        put("'a", "\\'e1");
```
tobiasdiez: Could you please explain the reason for this change (? -> ' )

src/main/java/net/sf/jabref/logic/util/strings/RtfCharMap.java
```diff
@@ -49,79 +49,58 @@ public RtfCharMap () {
         // Use UNICODE characters for RTF-Chars which can not be found in the
         // standard codepage
 
-        put("`A", "\\u192A"); // "Agrave"
```
tobiasdiez: Why did you removed some of these mappings?

src/test/java/net/sf/jabref/logic/layout/format/RTFCharsTest.java
```diff
@@ -49,51 +47,123 @@ public void testLaTeXHighlighting() {
...
-    @Ignore
     public void testComplicated() {
-        Assert.assertEquals("R\\u233eflexions sur le timing de la quantit\\u233e \\u230ae should be \\u230ae",
-                formatter.format("Réflexions sur le timing de la quantité \\ae should be æ"));
-
-        Assert.assertEquals("h\\u225all{\\uc2\\u339oe}", formatter.format("h\\'all\\oe "));
+        Assert.assertEquals("R\\u233eflexions sur le timing de la quantit\\u233e {\\u230ae} should be \\u230ae",
+                formatter.format("Réflexions sur le timing de la quantité {\\ae} should be æ"));
+
+        Assert.assertEquals("h\\'e1ll{\\u339oe}", formatter.format("h\\'all{\\oe}"));
```
tobiasdiez: Please only one assert per test method.

tobiasdiez: Please only one assert per test method.

src/main/java/net/sf/jabref/logic/util/strings/RtfCharMap.java
mairdl: Some keys got 2 values assigned, so I removed the double assignement. If a key got an RTF value I removed the unicode value.

src/main/java/net/sf/jabref/logic/util/strings/RtfCharMap.java
mairdl: Yes sure, later in the RTFCharMap you would find "'a" assigned with unicode, so I thought why not just map it to the RTF value if it already exists.

src/main/java/net/sf/jabref/logic/layout/format/RTFChars.java
```diff
@@ -199,4 +199,103 @@ private StringInt getPart(String text, int i, boolean commandNestedInBraces) {
...
+    }
+}
```
mairdl: I'll add the brackets and the newline.

src/test/java/net/sf/jabref/logic/layout/format/RTFCharsTest.java
mairdl: Some test methods inside this test class have lots of assertions for single characters, can I leave it like that? I will split the longer ones like you recommend.

lenhard:
Code-wise this looks fine for me. Though I'll admit that I have not looked up the mapping table and would trust you on this.

I have one more question regarding code style that is not your fault, but could you improve it anyway? I do not like that `RtfCharMap` extends `HashMap<String><String>`, which is also completely unecessary. Through this, a malicious user could compromise the structure of the table itself. It would be better if `RtfCharMap` just used an internal `HashMap`. All you would need to offer to the outside is a `get` methods, because nothing more is used atm. If you add an internal private `put` method, you can also avoid rewriting the code.

mairdl:
I looked the mappings up and the ones in JabRef seemed to be fine, so I used them. I will double check my code at the end to be 100% sure the mappings are fine. Also I will improve the code quality like @lenhard mentioned.

lenhard:
This looks done, or is there something left? @mairdl If so, please merge master into your branch and we are ready to go.

mairdl:
I'm on it.

mairdl:
So can it be merged now?

lenhard:
Yes, thanks for your contribution!

PR #2099, opened 2016-09-29
Clear searchterm with ESC from main table if neither preview nor editor is open

bartsch-dev:
At request of @koppor.

When hitting `ESC` in the main table and neither preview nor editor is open the search term will be cleared.

Since the action in the main table overshadows the actions in the preview-/editorpanel I have to call the `close()` method manually (same Keybinding).

### DIFF
--- a/src/main/java/net/sf/jabref/gui/BasePanel.java
+++ b/src/main/java/net/sf/jabref/gui/BasePanel.java
@@ -1390,9 +1390,7 @@ public class BasePanel extends JPanel implements ClipboardOwner, FileUpdateListe
                 // need to close these here, b/c this action overshadows the responsible actions when the main table is selected
                 switch (mode) {
                     case SHOWING_NOTHING:
-                        frame.getGlobalSearchBar().setSearchTerm("", true);
-                        mainTable.requestFocus();
-                        SwingUtilities.invokeLater(() -> mainTable.ensureVisible(mainTable.getSelectedRow()));
+                        frame.getGlobalSearchBar().endSearch();
                         break;
                     case SHOWING_PREVIEW:
                         getPreviewPanel().close();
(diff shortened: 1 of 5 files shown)


src/main/java/net/sf/jabref/gui/BasePanel.java
```diff
@@ -1381,6 +1382,28 @@ private void createMainTable() {
...
+                        frame.getGlobalSearchBar().setSearchTerm("", true);
+                        mainTable.requestFocus();
+                        break;
+                    case SHOWING_PREVIEW:
+                        getPreviewPanel().close();
+                        break;
+                    case SHOWING_EDITOR:
+                    case WILL_SHOW_EDITOR:
+                        getCurrentEditor().close();
+                        break;
```
tschechlovdev: please add default case. (Codacy would complain this)

bartsch-dev: done

lenhard:
I have tested the changes locally and can confirm it works.

It itches me a little that we are adding code to a class that is already more than 2000 lines long... Moving the code into `MainTable` would probably not improve things, though, as this class also has more than 600 lines.

I guess this is good to go. @bartsch-dev If you see potential for moving the new code to some place that is less bloated, please go ahead. Otherwise, I think we can live with it and merge.

koppor:
First, I thought that I would expect the search field to be focused. But I think, the current behavior is OK.

BUT: the selected entry changes when pressing "ESC". This should not happen: I expect the same entry to be focused if I continue pressing ESC.

esc-jumps-to-wrong-entry

koppor:
Special request: ESC when in the search field should close the auto completion and keep the inserted letters. Current behavior: clear search completely. ESC with the opened auto completer should just (i) close the auto completer and (ii) focus the main table.  - (ii) is debatable.

bartsch-dev:
1. Now it will scroll to the selected entry after dismissing the search.
2. When inside the the searchbar and the popup is open the first <kbd>ESC</kbd> will only close the popup
  - if you want to jump from the searchbar to the main table directly you can use <kbd>ctrl</kbd>+<kbd>enter</kbd>

Issue #1252, opened 2016-04-20
Reuse bibers data file for Latex to Unicode conversion


tobiasdiez:
Biber has a nice list mapping latex commands to their unicode equivalent: https://github.com/plk/biber/blob/dev/lib/Biber/LaTeX/recode_data.xml. 
After a quick look, this list seems to be more extensive than our map. Moreover, I like the xml format.

Thus I propose to switch to XML and reuse Biber's file (maybe with some additions which we can hand back to the biber team). The XML format has also the advantage that a help file listing all the conversations can be easily created .

@oscargus you worked a lot on this stuff. Are you willing to work on this issue too?

oscargus:
There's also the HTML and the capitalization thing. And the need for  ✎. So for me: no.

oscargus:
And Unicode to LaTeX (which basically includes the capitalization thing).

koppor:
This is

Siedlerchr:
The code how biber does the Transformation is there too https://github.com/plk/biber/blob/dev/lib/Biber/LaTeX/Recode.pm

Siedlerchr:
The code how biber does the Transformation is there too https://github.com/plk/biber/blob/dev/lib/Biber/LaTeX/Recode.pm

lenhard:
We decided to go for a conversion with latex2unicode instead, which is integrated into JabRef with #2532. So, there is no more need for reusing biber and this issue can be closed.

PR #2102, opened 2016-09-30
Store LaTeX-free fields in BibEntry

lenhard:
Based on the discussion in #1993 and the work of @bartsch-dev in #2091, this PR aims to improve the speed of search by storing LaTeX-free versions of field values in a cache. It aims at a balance between better performance and an acceptable memory footprint. Basically, all ideas from #1993 are implemented: a cached store of LaTeX-free fields which is computed on demand, String internalization, and regex performance improvements in `LatexToUnicode`.

Here are measurements with the new branch:
```
Benchmark                             Mode  Cnt         Score        Error  Unit
Benchmarks.latexToUnicodeConversion  thrpt   20    119823.078   2063.425  ops/s
Benchmarks.parallelSearch            thrpt   20       735.890    113.685  ops/s
Benchmarks.search                    thrpt   20       397.807     28.557  ops/s
```
The memory footprint of a database with 6500 entries is ~ 912 MB on my machine

And on current master:
```
Benchmark                             Mode  Cnt         Score        Error  Unit
Benchmarks.latexToUnicodeConversion  thrpt   20     94367.718   2570.784  ops/s
Benchmarks.parallelSearch            thrpt   20        58.204       3.285  ops/s
Benchmarks.search                    thrpt   20        30.531       0.952  ops/s
```
The memory footprint of the same 6500 entry database is ~ 888 MB.

All in all, this looks like a hefty performance improvement at little memory cost.

### DIFF
--- a/jabref.install4j
+++ b/jabref.install4j
@@ -750,4 +750,9 @@ return true;</string>
   </mediaSets>
   <buildIds buildAll="true" />
   <buildOptions verbose="false" faster="false" disableSigning="false" disableJreBundling="false" debug="false" />
+  <jvmArguments>
+      <arg>-XX:+UseG1GC</arg>
+      <arg>-XX:+UseStringDeduplication</arg>
+      <arg>-XX:StringTableSize=1000003</arg>
+  </jvmArguments>
 </install4j>
--- a/src/main/java/net/sf/jabref/model/entry/BibEntry.java
+++ b/src/main/java/net/sf/jabref/model/entry/BibEntry.java
@@ -53,17 +53,18 @@ public class BibEntry implements Cloneable {
 
     private String type;
     private Map<String, String> fields = new ConcurrentHashMap<>();
-    /*
+
+    /**
      * Map to store the words in every field
      */
     private final Map<String, Set<String>> fieldsAsWords = new HashMap<>();
 
-    /*
-     * Cache that stores latex free versions of fields. Is populated as a cash
+    /**
+     * Cache that stores latex free versions of fields.
      */
     private final Map<String, String> latexFreeFields = new ConcurrentHashMap<>();
 
-    /*
+    /**
      * Used to cleanse field values for internal LaTeX-free storage
      */
     private LatexToUnicode unicodeConverter = new LatexToUnicode();
@@ -76,7 +77,7 @@ public class BibEntry implements Cloneable {
 
     private String commentsBeforeEntry = "";
 
-    /*
+    /**
      * Marks whether the complete serialization, which was read from file, should be used.
      *
      * Is set to false, if parts of the entry change. This causes the entry to be serialized based on the internal state (and not based on the old serialization)
--- a/src/main/java/net/sf/jabref/model/entry/BibEntry.java
+++ b/src/main/java/net/sf/jabref/model/entry/BibEntry.java
@@ -59,7 +59,7 @@ public class BibEntry implements Cloneable {
     private final Map<String, Set<String>> fieldsAsWords = new HashMap<>();
 
     /*
-     * Map that stores latex free versions of fields. Is populated as a cash
+     * Cache that stores latex free versions of fields. Is populated as a cash
      */
     private final Map<String, String> latexFreeFields = new ConcurrentHashMap<>();
 
@@ -191,7 +191,6 @@ public class BibEntry implements Cloneable {
 
     /**
      * Sets the cite key AKA citation key AKA BibTeX key.
-     * <p>
      * Note: This is <emph>not</emph> the internal Id of this entry. The internal Id is always present, whereas the BibTeX key might not be present.
      *
      * @param newCiteKey The cite key to set. Must not be null; use {@link #clearCiteKey()} to remove the cite key.
--- a/src/main/java/net/sf/jabref/model/entry/BibEntry.java
+++ b/src/main/java/net/sf/jabref/model/entry/BibEntry.java
@@ -201,7 +201,6 @@ public class BibEntry implements Cloneable {
 
     /**
      * Returns the cite key AKA citation key AKA BibTeX key, or null if it is not set.
-     * <p>
      * Note: this is <emph>not</emph> the internal Id of this entry. The internal Id is always present, whereas the BibTeX key might not be present.
      */
     @Deprecated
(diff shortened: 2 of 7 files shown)


stefan-kolb:
LGTM. Although the converter is probably not perfect and we might run into new problems caused by that fact.

src/main/java/net/sf/jabref/model/entry/BibEntry.java
```diff
@@ -57,6 +58,16 @@
      */
     private final Map<String, Set<String>> fieldsAsWords = new HashMap<>();
 
+    /*
+     * Map that stores latex free versions of fields. Is populated as a cash
```
matthiasgeiger: cache

src/main/java/net/sf/jabref/model/entry/BibEntry.java
```diff
@@ -57,6 +58,16 @@
...
+     */
+    private final Map<String, String> latexFreeFields = new ConcurrentHashMap<>();
+
+    /*
+     * Used to cleanse field values for internal LaTeX-free storage
```
matthiasgeiger: clean

src/main/java/net/sf/jabref/model/entry/BibEntry.java
```diff
@@ -180,7 +191,7 @@ public String getId() {
 
     /**
      * Sets the cite key AKA citation key AKA BibTeX key.
-     *
+     * <p>
```
matthiasgeiger: remove?

src/main/java/net/sf/jabref/model/entry/BibEntry.java
```diff
@@ -191,7 +202,7 @@ public void setCiteKey(String newCiteKey) {
 
     /**
      * Returns the cite key AKA citation key AKA BibTeX key, or null if it is not set.
-     *
+     * <p>
```
matthiasgeiger: remove?

lenhard:
The functionality in the converter has not changed, it is just a rewording of `String.replace` calls into regular expressions that are compiled once instead of every time. Thus, there should be no new problems that weren't broken before anyway. You can never be sure, though...

src/main/java/net/sf/jabref/model/entry/BibEntry.java
lenhard: Ok, will change

src/main/java/net/sf/jabref/model/entry/BibEntry.java
lenhard: What is wrong with cleanse? My dictionary says this is a proper term :-)

src/main/java/net/sf/jabref/model/entry/BibEntry.java
```diff
@@ -57,6 +58,16 @@
      */
     private final Map<String, Set<String>> fieldsAsWords = new HashMap<>();
 
+    /*
+     * Cache that stores latex free versions of fields. Is populated as a cash
```
tobiasdiez: Second sentence doesn't make sense?!

lenhard:
Regarding `GrammarBasedSearchRule`: I am not 100% sure, but it looks like it on the one hand builds upon `ContainBasedSearchRule`. But it also has some self-searching code, which I modified to use the LaTeX-free version. This was probably a bug.

koppor:
Implementation alternatives for the LaTeX-to-Unicode conversions are discussed at https://github.com/JabRef/jabref/issues/1252

src/main/java/net/sf/jabref/model/entry/BibEntry.java
```diff
@@ -479,6 +491,8 @@ public void setField(Map<String, String> fields) {
 
         fields.remove(fieldName);
         fieldsAsWords.remove(fieldName);
```
tobiasdiez: maybe also move `fieldsAsWords.remove(fieldName);` to the invalidate cache method?

src/main/java/net/sf/jabref/model/entry/BibEntry.java
```diff
@@ -57,6 +58,16 @@
      */
     private final Map<String, Set<String>> fieldsAsWords = new HashMap<>();
 
+    /*
```
bartsch-dev: `/** `instead of `/*`

src/main/java/net/sf/jabref/model/entry/BibEntry.java
```diff
@@ -57,6 +58,16 @@
...
+     * Cache that stores latex free versions of fields. Is populated as a cash
+     */
+    private final Map<String, String> latexFreeFields = new ConcurrentHashMap<>();
+
+    /*
+     * Used to cleanse field values for internal LaTeX-free storage
+     */
```
bartsch-dev: `/** `instead of `/*`

src/main/java/net/sf/jabref/model/entry/BibEntry.java
lenhard: Good point. It's strange that Intellij does not distinguish between JavaDoc and Inline documentation in its color.

src/main/java/net/sf/jabref/model/entry/BibEntry.java
lenhard: right

src/main/java/net/sf/jabref/model/entry/BibEntry.java
lenhard: Good point! Done!

src/main/java/net/sf/jabref/model/entry/BibEntry.java
bartsch-dev: With the Darcula theme it does.

lenhard:
As a few people have looked at the code and only criticized minor aspects (which I fixed), I'll merge this now. The codacy build seems to be broken somehow, but then it is also not that critical.

jabref.install4j
```diff
@@ -750,4 +750,9 @@ return true;</string>
   </mediaSets>
   <buildIds buildAll="true" />
   <buildOptions verbose="false" faster="false" disableSigning="false" disableJreBundling="false" debug="false" />
+  <jvmArguments>
```
koppor: Note that if these settings go missing, it may caused by a plain save of the config using the Install4J GUI. I'm trying to add it as "VM options file" at https://github.com/JabRef/jabref/commit/53ffd02287a0cd4b83115243f631df6551d08086, which seems to work.

PR #1988, opened 2016-09-15
German translation fix

mairdl:
The German translation of JabRef has some issues, have a look at the following screenshot:

file-open

The function to open a database is called `Datei öffnen` the displayed hotkey is `F4`. In the English version of JabRef the file menu looks like this:

openfilemenu

As you can see the right hotkey for this function should be` CTRL + O`. Also, the word database gets translated as `Datei` instead of `Datenbank`.

This PR aims to fix this issue by rewording the translation. Every English occurrence of `database `is now translated as `Datenbank`. This is what it looks like now:

openfilemenu

- [x] Change in CHANGELOG.md described
- [x] Manually tested changed features in running JabRef
- [x] internal qs

PR #2103, opened 2016-09-30
Change "Recent Files" to "Recent Databases"

mairdl:
This is a followup of https://github.com/JabRef/jabref/pull/1988.

In the sense of the PR mentioned above, I changed "Recent files" to "Recent databases". Since it only displays recently opened databases. 

This is the file menu after the changes:
recentdatabase

As you can see it fits in with the rest of the menu. One could say change the file menu to the database menu, but I think keeping the file menu named as it is, is better because it stays consistent with other programms.

Furthermore the issue https://github.com/JabRef/jabref/issues/2095 seems to be related.

- [x] internal qs


src/main/java/net/sf/jabref/gui/menus/FileHistoryMenu.java
```diff
@@ -22,7 +22,7 @@
 
 
     public FileHistoryMenu(JabRefPreferences prefs, JabRefFrame frame) {
-        String name = Localization.menuTitle("Recent files");
+        String name = Localization.menuTitle("Recent_databases");
```
boceckts: Underscore is not needed in when using translations in code.

src/main/resources/l10n/Menu_en.properties
```diff
@@ -56,7 +56,7 @@ Paste=&Paste
 Preferences=&Preferences
 Previous_tab=&Previous_tab
 Quit=&Quit
-Recent_files=&Recent_files
+Recent_databases=Recent_databases
```
boceckts: Please add the `&` sign in front of the 'R' of the translated string as before. In the menu properties this indicates which letter should be used as mnemonic for the underlying action.

src/main/resources/l10n/Menu_de.properties
```diff
@@ -56,7 +56,7 @@ Paste=&Einfügen
 Preferences=&Einstellungen
 Previous_tab=&Vorheriger_Tab
 Quit=&Beenden
-Recent_files=&Zuletzt_geöffnete_Dateien
+Recent_databases=
```
boceckts: Please add a German translation as I think this was the whole idea of the linked pr.

mairdl: yes sure

src/main/java/net/sf/jabref/gui/menus/FileHistoryMenu.java
mairdl: ok

src/main/resources/l10n/Menu_en.properties
mairdl: I see i will correct it

src/main/resources/l10n/Menu_es.properties
```diff
@@ -56,7 +56,7 @@ Paste=&Pegar
 Preferences=&Preferencias
 Previous_tab=&Pestaña_anterior
 Quit=&Salir
-Recent_files=Bases_de_datos_&recientes
+Recent_databases=
```
koppor: Only the key should be changed, not the values. We have no clue, whether all the other translations use a translation of "File" or "Database". Even though, some language use "File", I lean towards a more complete translation instead of having an incomplete one. - In the concrete case here "Bases de datos" seems to be the translation of "databases" and has for sure to be kept.

So, please, just change the key and do not remove the translation.

mairdl: I think if we leave the old translations like they are now, no one feels like imediatly updating the translation. By removing them completly the need to change ist bigger.

mairdl: @boceckts what do you think?

boceckts:
As @koppor pointed out in some languages this is already translated as "recent databases". So it is probably better to leave the already translated strings. Also you don't know if in different languages the other database related strings are always translated as "database" and not file as it was the case in the german translation.

mairdl:
Ok, in that case i will "translate" them. 💃

tobiasdiez:
LGTM :+1:

PR #2106, opened 2016-09-30
Crash after saving BibTeX source with parsing error

### LINKED ISSUE #2104: Empty modal error box when saving database with parsing error
JabRef version 3.6 on Windows 10

Steps to reproduce:

1. Open Jabref
2. Create a database
3. Add an entry (I used article)
4. Save the database (if you skip this step, the problem does not occur. See later.)
5. Open the "{} Bibtex Source" editing tab
6. Insert an extra comma just after the last one (the one closing the last field). Other mistakes may work as well; I have not tried.
7. Click on the "save" button on the toolbar.

At this point, in addition to a "legitimate" error box signaling the inconsistency, another empty one appears that prevents any further action (see attached graphics). Jabref must be terminated forcefully.
Note that, as mentioned in step 4, if the database has not been saved before, the "save as" dialog box appears instead of the empty error box, and the user can proceed regularly by saving the database first, and responding to the error message then.
jabreferror

grimes2:
Fixes #2104.


lenhard:
I just tested your PR following the instructions stated in #2104. JabRef does not crash anymore. However, no error message appears and the additional faulty comma is just not stored silently. When I look into the error console, I see the error.

Imho, this is not quite the behavior we want. It is fine to avoid storing the broken BibTeX, but there should be an error message (that does not crash JabRef). Otherwise, a user might expect that the save worked and would be surprised to see that it did not. This is critical, since all changes made to the entry are lost, not just the comma. Therefore, we need the error message so that the user has a change to fix the mistakes.

grimes2:
I get an error message:
parsingerror
I don't know, why it does not work for you.
`Edit` keeps the BibTeX source (with parsing error) and gives the opportunity to correct the error.
`Revert to original source` does not store the changes.

lenhard:
You are correct, my bad! I was hitting CTRL+S and not pressing the button in the toolbar. Your PR can be merged, and apologies for not getting it.

Didn't we use to have a keyboard shortcut for saving?

grimes2:
Ohh, is this another bug? Ctrl+s should also trigger the error message.

lenhard:
@grimes2 For me it did not. Can you please double check? If you can reproduce it, would you be willing to submit a second pull request?

This could be a keybinding issue though, not directly related to the saving logic.

grimes2:
File > Save database works
Ctrl+s fails
Strange!
I'll open a new PR.

lenhard:
Thanks a lot!

PR #2107, opened 2016-10-02
Updated jabref_tr.properties

hakova:
Translated the few new entries to Turkish


lenhard:
The build seems to have failed for an unrelated reason. I restarted it and once it has passed, I will merge this as well.

lenhard:
And thanks again!

PR #2112, opened 2016-10-03
Refactoring: completey typed metadata

tobiasdiez:
The metadata class is completely rewritten and now only uses typed fields instead of a map of serialized strings. Serialization and parsing now happens in the dedicated MetadataSerializer/parser classes.
For this also some of the cleanup classes had to be moved to the model package. 

Final big refactoring PR from my side :smile:  


Siedlerchr:
I think your problem with the failing test couuld be due to the renaming of the DBMSSynchronizerTest.java
There was a case change of the test name. At least I had a similar problem a while ago on Win.

src/main/java/net/sf/jabref/gui/groups/GroupSelector.java
```diff
@@ -250,7 +250,6 @@ public void stateChanged(ChangeEvent e) {
                 groupsRoot.addNewGroup(newGroup, panel.getUndoManager());
                 panel.markBaseChanged();
                 frame.output(Localization.lang("Created group \"%0\".", newGroup.getName()));
-                panel.getBibDatabaseContext().getMetaData().postChange();
```
lenhard: Why are all the change notifications in this class no longer needed?

src/main/java/net/sf/jabref/logic/exporter/MetaDataSerializer.java
```diff
@@ -13,13 +20,42 @@
     /**
      * Writes all data in the format <key, serialized data>.
      */
-    public static Map<String, String> getSerializedStringMap(MetaData metaData) {
+    public static Map<String, String> getSerializedStringMap(MetaData metaData,
```
lenhard: How about refactoring this method a little bit? After all, it consists of various blocks, where different types of meta data contents are put into the final list. Each such block could be extract into a beautifully named private method :-)

src/main/java/net/sf/jabref/model/cleanup/FieldFormatterCleanups.java
```diff
@@ -0,0 +1,122 @@
+package net.sf.jabref.model.cleanup;
```
lenhard: I have a suggestion regarding this new package: How about moving it into `model.metadata`? After all, the classes in the package are about how the cleanup information is stored in the meta data. What do you think?

src/main/java/net/sf/jabref/model/metadata/MetaData.java
```diff
@@ -139,32 +70,20 @@ public void putData(String key, List<String> orderedData) {
...
     }
 
     /**
      * @return the stored label patterns
      */
-    public AbstractBibtexKeyPattern getBibtexKeyPattern(GlobalBibtexKeyPattern globalPattern) {
-        if (bibtexKeyPattern != null) {
-            return bibtexKeyPattern;
-        }
+    public AbstractBibtexKeyPattern getCiteKeyPattern(GlobalBibtexKeyPattern globalPattern) {
```
lenhard: Since you are doing this for the other methods: Do a null check of `globalPattern`.

src/main/java/net/sf/jabref/model/metadata/MetaData.java
```diff
@@ -173,139 +92,117 @@ public AbstractBibtexKeyPattern getBibtexKeyPattern(GlobalBibtexKeyPattern globa
...
-    public Optional<List<String>> getSaveActions() {
-        return Optional.ofNullable(getData(SAVE_ACTIONS));
+    public Optional<FieldFormatterCleanups> getSaveActions() {
+        return Optional.ofNullable(saveActions);
     }
 
     public Optional<BibDatabaseMode> getMode() {
-        List<String> data = getData(DATABASE_TYPE);
-        if ((data == null) || data.isEmpty()) {
-            return Optional.empty();
```
lenhard: Why is the `Optional.empty` no longer a desired return type here?

src/main/java/net/sf/jabref/model/metadata/MetaData.java
```diff
@@ -173,139 +92,117 @@ public AbstractBibtexKeyPattern getBibtexKeyPattern(GlobalBibtexKeyPattern globa
...
     public void setMode(BibDatabaseMode mode) {
-        putData(DATABASE_TYPE, Collections.singletonList(mode.getAsString()));
+        this.mode = Objects.requireNonNull(mode);
+        postChange();
     }
 
     public void markAsProtected() {
-        putData(PROTECTED_FLAG_META, Collections.singletonList("true"));
+        isProtected = true;
+        postChange();
```
lenhard: So the change notification I asked for earlier have been moved into the meta data. Am I right?

src/test/resources/testbib/complex.bib
```diff
@@ -296,8 +296,6 @@ @Comment{jabref-meta:
 1 KeywordGroup:DynamicGroup\;0\;author\;Churchill\;0\;0\;;
 }
 
-@Comment{jabref-meta: groupsversion:3;}
```
lenhard: Is it ok to just remove that? Wouldn't it be better to have it in and verify that it is just ignored by the new parser?

src/main/java/net/sf/jabref/collab/ChangeScanner.java
```diff
@@ -165,37 +165,16 @@ private void storeTempDatabase() {
...
-                metadataChange.insertMetaDataRemoval(key);
-            } else {
-                // Both exist. Check if they are different:
-                List<String> vit = inTemp1.getData(key);
-                if (!vod.equals(vit)) {
-                    metadataChange.insertMetaDataChange(key, vod);
-                }
-                // Remember that we've handled this one:
-                handledOnDisk.add(key);
+    private void scanMetaData(MetaData inMemory, MetaData onTmp, MetaData onDisk) {
```
tobiasdiez: Yes the detailed information about which key has changed is not contained in the event. Of course this information could be easily extracted from the two metadata by comparing each contained meta item separately. I didn't found this information to useful and thus just left a todo https://github.com/JabRef/jabref/pull/2112/files#diff-423409b75a40e426ad5123cabbe295e6R33.

src/main/java/net/sf/jabref/gui/groups/GroupSelector.java
tobiasdiez: The metadata class now directly subscribes to changes in the group tree and relays these change events https://github.com/JabRef/jabref/pull/2112/files#diff-20cd356a80ee8ffb10c848005c89b949R74

src/main/java/net/sf/jabref/model/cleanup/FieldFormatterCleanups.java
tobiasdiez: Mhhh.... none of the other metadata items are currently under the metadata namespace:
import net.sf.jabref.model.cleanup.FieldFormatterCleanups;
import net.sf.jabref.model.bibtexkeypattern.AbstractBibtexKeyPattern;
import net.sf.jabref.model.database.BibDatabaseMode;
import net.sf.jabref.model.groups.GroupTreeNode;

I think it is more consistent to have a model.cleanup or also move the rest to model.metadata(.groups/.bibtexkeypattern...)

src/main/java/net/sf/jabref/model/metadata/MetaData.java
tobiasdiez: Good observation :) Done.

src/main/java/net/sf/jabref/model/metadata/MetaData.java
tobiasdiez: If `mode` is null, then still an empty optional is returned (Optional.ofNullable(mode))

src/test/resources/testbib/complex.bib
tobiasdiez: We have a test `integrationTestGroupTree` https://github.com/JabRef/jabref/pull/2112/files#diff-837597ff5fafbe4e71d5733d490e3454R1479 which still have the ̀`groupsversion:3` string. But since we no longer write the the `groupsversion` information, the roundtrip tests for complex.bib fail. This is why I removed it.

tobiasdiez:
Thanks for the review. I addressed all your remarks and hopefully the build now runs through without problems.

Siedlerchr:
Regarding the failing tests: When you log in to Travis-CI and check the build, you can access the raw log.
In your case it showed:
```

Formatting ./build/test-results/databaseTest/TEST-net.sf.jabref.shared.DBMSSynchronizerTest.xml
=====================================================
Testsuite: net.sf.jabref.shared.DBMSSynchronizerTest
Tests run: 16, Failures: 2, Errors: 0, Time elapsed: 2.522 sec
--------- ----------- ---------
Testcase: testMetaDataChangedEventListener[Test with PostgreSQL database system] took 0.209 FAILURE java.lang.AssertionError: expected:<{databaseType=bibtex;}> but was:<{}> java.lang.AssertionError
java.lang.AssertionError: expected:<{databaseType=bibtex;}> but was:<{}>
```

Siedlerchr:
There is also an error which seems to be related to your changes:
```

Oct 04, 2016 6:04:34 PM com.google.common.eventbus.EventBus$LoggingHandler handleException
SEVERE: Exception thrown by subscriber method listen(net.sf.jabref.model.metadata.event.MetaDataChangedEvent) on subscriber net.sf.jabref.shared.DBMSSynchronizer@5d506d49 when dispatching event: net.sf.jabref.model.metadata.event.MetaDataChangedEvent@34a71a73
	at net.sf.jabref.shared.DBMSSynchronizer.listen(DBMSSynchronizer.java:121)
```

src/main/java/net/sf/jabref/model/cleanup/FieldFormatterCleanups.java
lenhard: Ok, at some point we might think of structuring the model package a little more, but for now it is ok, I think.

obraliar:
@tobiasdiez Since this merge the synchronization of metadata for shared databases is totally broken. Could you please investigate why this happened? There is also a new issue https://github.com/JabRef/jabref/issues/2219.

lenhard:
@obraliar: Looking at the code, there have been changes in ` src/main/java/net/sf/jabref/shared/DBMSSynchronizer.java `, so my best guess would be those changes. Could you have a look at the changes to this class? Maybe some meta data are not carried over properly due to the typing.

Lacking an actual database for testing, I cannot provide more help unfortunately. It seems that we need more tests for the database code, as travis did not find this problem.

tobiasdiez:
Sorry I have no db to test the sync (and no intellij at the moment either). I have to admit, I kind of trusted that our db-tests would catch possible problems.

src/main/java/net/sf/jabref/model/cleanup/FieldFormatterCleanups.java
koppor: @calixtus and @koppor currently wonder why this class was moved to `model`. All implementing classes are in the `logic` package: 

grafik

Issue #2110, opened 2016-10-03
Medline search throws Exception and hangs (Jabref 3.6)


cookpa:
JabRef version 3.6 on Mac OS X El Capitan

Steps to reproduce:

1. Open web search
2. Enter "Cook PA" (with or without quotes)
3. It says "Number of references 105. Number to fetch? [click OK]
4. The blue progress bar fills up, but the references never appear

Sometimes I can run a query and some references will appear, but then it hangs. For example, if I search "Tustison [au]" (no quotes), I get some results but then it hangs at about 25% progress. The hangups are accompanied by Exceptions in the console window:

```
Exception in thread "JabRef CachedThreadPool" java.lang.NullPointerException
	at net.sf.jabref.logic.importer.fileformat.MedlineImporter.addArticleInformation(MedlineImporter.java:516)
	at net.sf.jabref.logic.importer.fileformat.MedlineImporter.parseArticle(MedlineImporter.java:341)
	at net.sf.jabref.logic.importer.fileformat.MedlineImporter.importDatabase(MedlineImporter.java:151)
	at net.sf.jabref.gui.importer.fetcher.MedlineFetcher.fetchMedline(MedlineFetcher.java:226)
	at net.sf.jabref.gui.importer.fetcher.MedlineFetcher.processQuery(MedlineFetcher.java:199)
	at net.sf.jabref.gui.importer.fetcher.GeneralFetcher.lambda$actionPerformed$5(GeneralFetcher.java:223)
	at net.sf.jabref.gui.importer.fetcher.GeneralFetcher$$Lambda$191/1683863109.run(Unknown Source)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
```

IEEE and Springer search don't seem to have this problem. 

I can get specific things from Medline, for example if I enter a single PMID I can get the result. But larger searches sometimes fail with the above error. I had thought it was due to Medline refusing large requests, but I can enter some very general terms (eg searching my last name without initials) and get hundreds of results each time.

I tried the same thing on 3.0 and it works correctly, I can search "Cook PA [au]" and the results show up as expected.

tobiasdiez:
[...]

PR #2113, opened 2016-10-03
Fix NPE in Medline fetcher on missing ISSN

Siedlerchr:
Fix for #2110

When issn was null, calling a method on it resulted in NPE.

### DIFF
--- a/src/main/java/net/sf/jabref/logic/importer/fileformat/MedlineImporter.java
+++ b/src/main/java/net/sf/jabref/logic/importer/fileformat/MedlineImporter.java
@@ -513,7 +513,9 @@ public class MedlineImporter extends ImportFormat {
                 putIfValueNotNull(fields, FieldName.JOURNAL, journal.getTitle());
 
                 ISSN issn = journal.getISSN();
-                putIfValueNotNull(fields, FieldName.ISSN, issn.getContent());
+                if (issn != null) {
+                    putIfValueNotNull(fields, FieldName.ISSN, issn.getContent());
+                }
 
                 JournalIssue journalIssue = journal.getJournalIssue();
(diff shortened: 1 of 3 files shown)


tobiasdiez:
As a quick fix this seems to be ok. Two remarks:
- Are all other fields always not null (e.g. JournalIssue ?)
- Can we change the Journal class to return optionals instead of null?

Siedlerchr:
@tobiasdiez  I will check it if Journal issue is also possible  null from the annotation in the Journal class
Changing the code of Journal is not possible because it is autogenerated from the jaxb stuff on base of the xml class 

But I wanted to add a test, too

lenhard:
In theory, all fields can be null. JAXB will take whatever is there and will set null for everything that it cannot find. Ultimately, null values depend on how consistent the data from medline are, and I guess we should be rather conservative in our assumptions...

There is already a fair amount of null checking code in the importer and it is quite hard to see what could be missing. @Siedlerchr We will rely on your knowledge and intuition here, but of course it is easy to miss something and then we will have to fix the error when someone opens an issue.

Apart from that I second @tobiasdiez: Please add a test and this is ready to go!

tobiasdiez:
Apparently there is a way to let JAXB convert nulls to empty optionals http://stackoverflow.com/a/23113540/873661. Not sure how complicated this is versus just checking for null.

lenhard:
@tobiasdiez If I understand it correctly, this solution is not feasible for us. It requires changing the annotations in the Java class and in our case this Java class is generated from an XSD.

Apart from that I am always happy when we can avoid that darn Optional and do a good ol' null check ;-D

Siedlerchr:
Apart from codecov having problems with parameterizest test, all works

PR #2116, opened 2016-10-04
Permanent other fields tab

grimes2:
The other fields tab is now permanently shown.
Reason: This avoids major layout rearrangements of the entry editor tabs on entry change.
Advantage: Better usability and performance.
Disadvantage: A useless (empty) other fields tab is shown in databases without other fields.


Siedlerchr:
Just one thing about I have thought. Would it be possible to "create" the other fields tab everytime, but not showing (visible=false)  if it is empty ? 
Would that work, too?

grimes2:
Do you mean, that the other fields tab is grayed out (inactive)?

Siedlerchr:
Yeah kind of. But  when thinking again about it, graying out is a bad idea. Leave it empty, All fine. I just had the idea it could be complete hidden as it was before.

koppor:
My current feeling is that always shown and with the text "no other fields
found" would be the most intuitive.

Double check: If I add a field "realotherfield" at the bibtex source, the
content of the other tab should show this field.

grimes2:
Other fields tab is not updated after adding an other field to source. This is not surprising, because changing the tab in entry editor does not instanciate the EntryEditor object. You can choose another entry and go back to rebuild the Entry Editor.

grimes2:
I found an alternative solution. I have added an empty field "otherfield" to every other fields tab. So the tab is never empty. The advantage is, that it's possible to add a field to the source easily. This field can be renamed in the source panel. Please set label "ready-for-review". Thanks.

stefan-kolb:
Imho, I only want to see the other field tab if there are other fields.

koppor:
In the devcall, we discussed that this is not an ideal solution and we should really work on the other fields tab being handled correctly.

Issue #344, opened 2015-11-15
Feature request: autosave


vrossum:
Hi
Currently autosave is active every 5 minutes. Would it be possible to whenever an entry is added? 

This would help a lot when one is typing a latex document and needs to include an extra ref in jabref.
Currently one needs to remember to save jabref everytime before running latex.

stefan-kolb:
You are talking about an autosave so a Latex run can include the newly created entry inside the references?! If not please clarify your use case.

How should the algorithm detect when you are finished with the entry?
It may save when it has a Bibtex key but might still be incomplete.

I don't see a good algorithm to detect the autosave point. Imho the normal use case should be:
1. create your entry
2. save when your done (just CTRL-S for convenient saving)
3. Insert cite inside Latex document
4. compile Latex

tobiasdiez:
We could consider saving after every edit (maybe as an additional preference option).

vrossum:
Thanks for looking into this.

On 16/11/15 23:22, Stefan Kolb wrote:
> You are talking about an autosave so a Latex run can include the newly
> created entry inside the references?! If not please clarify your use case.

Indeed.

> How should the algorithm detect when you are finished with the entry?
> It may save when it has a Bibtex key but might still be incomplete.

I see the problem.
However, I typically search via google scholar or pubmed. In that case 
saving after getting the bibtex key would work.

-- 
Mark van Rossum, Reader, School of Informatics, U Edinburgh
Forum Rm 2.52, 44-131-6511211

The University of Edinburgh is a charitable body, registered in
Scotland, with registration number SC005336.

stefan-kolb:
@JabRef/developers Any comments if we want to change anything or keep the current behavior?

lenhard:
I guess unless @vrossum commits to creating a pull request and implementing the feature, nothing will happen here. At least I do not see any other volunteers.

@vrossum Will you implement this, or can we close the issue?

obraliar:
The JabRef StuPro feature team will take care of this issue.

simonharrer:
Hm, to make this a useful feature it should make saving obsolete. Every edit or change should be directly reflected in the persisted the database.

obraliar:
Now we are collectiong ideas about realizing this. Suggestions are welcome.

koppor:
Maybe, it should be possible to turn on and off this feature on demand. If I work on a dropbox shared folder, I don't like that each change is instantly saved in dropbox, do I?

[...]

PR #2118, opened 2016-10-04
Implementation of autosave & backup feature

obraliar:
Related to: https://github.com/JabRef/jabref/issues/344.

**Utility for local databases:**

This feature is event based and saves the database if activated on every user interaction. An intelligent `ExecutorService` and `workerQueue` are preventing a high load while saving. Furthermore all redundant save tasks will be rejected.

An event based `BackupManager` replaces the old time based backup system. This one is also using a 
`ExecutorService` to prevent high load issues.

**Utility for shared databases:**

For shared databases this feature actually synchronizes the local file which is associated with the shared database supporting LiveUpdate!

- [x] Internal QA
- [x] _MVC pattern_


matthiasgeiger:
I have not tested it, but what are the implacations of this? 

If I open a bib-File in JabRef, do some changes and then decide to discard all changes since opening the database. Is this still possible?

*(While this is a rather obscure use case for normal users, this is exactly my approach when trying to reproduce bugs or checking changes made in code. So I should know whether I have to change my workflow in order not to "destroy" some files...)*

obraliar:
@matthiasgeiger Thanks for the objection. For the case to discard all changes while _Autosave_ is enabled the _Undo_ operation could be used gradually as _Undo_ or _Redo_ is also supported by this feature.

An idea is to add an "_Undo all_" operation (possibly a new button next to the existing _Undo_ button) which could recover the state the file was opened with. Currently I can not judge whether it's worth to do this but I could open a new issue or PR for this if desired.

src/main/java/net/sf/jabref/shared/prefs/SharedDatabasePreferences.java
```diff
@@ -89,4 +108,25 @@ public void clear() throws BackingStoreException {
...
 
+    private Optional<String> getOptionalValue(String key) {
+        return Optional.ofNullable(internalPrefs.get(key, null));
+    }
+
+    public static void clearAll() throws BackingStoreException {
+        Preferences.userNodeForPackage(JabRefMain.class).parent().node(PARENT_NODE).clear();
+    }
+
+    public void putAllDBMSConnectionProperties(DBMSConnectionProperties properties) {
```
tschechlovdev: Why don't check here if the properties are valid? (Senn this method somewhere above...)

src/main/java/net/sf/jabref/autosave/event/AutosaveEvent.java
```diff
@@ -0,0 +1,10 @@
+package net.sf.jabref.autosave.event;
+
+import net.sf.jabref.autosave.Autosaver;
+
+/**
+ * This Event is fired from {@link Autosaver} in case that a save task is pending.
+ */
+public class AutosaveEvent {
```
obraliar: This is just an Object which is passed through the eventBus according to the context. It carries no special data as it's only used for notifying methods which are listening for `AutosaveEvent`. I'm following this strategy to preserve the MVC pattern. Otherwise I could call saving methods directly from `Autosaver` without using the eventBus.

src/main/java/net/sf/jabref/gui/exporter/SaveDatabaseAction.java
```diff
@@ -303,23 +311,46 @@ public void saveAs() throws Throwable {
...
+        boolean autosave = (((context.getLocation() == DatabaseLocation.SHARED) && Globals.prefs.getBoolean(JabRefPreferences.SHARED_AUTO_SAVE)) ||
+                ((context.getLocation() == DatabaseLocation.LOCAL) && Globals.prefs.getBoolean(JabRefPreferences.LOCAL_AUTO_SAVE))) &&
+                context.getDatabaseFile().isPresent();
+
+        // activate a new autosaver if enabled
+        if (autosave) {
+            new Autosaver(context).registerListener(new AutoSaveUIManager(panel));
+        }
+
+        frame.getFileHistory().newFile(context.getDatabaseFile().get().getPath());
```
obraliar: This line has been assumed. However I added a `ifPresent` check for this.

src/main/java/net/sf/jabref/gui/JabRefFrame.java
```diff
@@ -1700,6 +1689,16 @@ public void addTab(BasePanel bp, boolean raisePanel) {
 
         // Register undo/redo listener
         bp.getUndoManager().registerListener(new UndoRedoEventManager());
+
+        BibDatabaseContext context = bp.getBibDatabaseContext();
```
obraliar: Done.

src/main/java/net/sf/jabref/gui/exporter/SaveDatabaseAction.java
```diff
@@ -303,23 +311,46 @@ public void saveAs() throws Throwable {
...
+            // Generate an ID when saving a shared database
+            String databaseID = new BigInteger(128, new SecureRandom()).toString(32);
+            context.getDatabase().setDatabaseID(databaseID);
+            DBMSProcessor dbmsProcessor = context.getDBMSSynchronizer().getDBProcessor();
+            DBMSConnectionProperties properties = dbmsProcessor.getDBMSConnectionProperties();
+            // Save all properties dependent on the ID. This makes it possible to restore them.
+            new SharedDatabasePreferences(databaseID).putAllDBMSConnectionProperties(properties);
+        }
+
+        File oldFile = context.getDatabaseFile().orElse(null);
```
obraliar: Yes. :+1: Done.

src/main/java/net/sf/jabref/gui/exporter/SaveDatabaseAction.java
```diff
@@ -303,23 +311,46 @@ public void saveAs() throws Throwable {
...
-                    panel.getBibDatabaseContext().getDatabaseFile().orElse(null)));
+                    context.getDatabaseFile().orElse(null)));
         } catch (IOException ex) {
             LOGGER.error("Problem registering file change notifications", ex);
         }
-        frame.getFileHistory().newFile(panel.getBibDatabaseContext().getDatabaseFile().get().getPath());
+
+        boolean autosave = (((context.getLocation() == DatabaseLocation.SHARED) && Globals.prefs.getBoolean(JabRefPreferences.SHARED_AUTO_SAVE)) ||
+                ((context.getLocation() == DatabaseLocation.LOCAL) && Globals.prefs.getBoolean(JabRefPreferences.LOCAL_AUTO_SAVE))) &&
+                context.getDatabaseFile().isPresent();
```
obraliar: Done.

src/main/java/net/sf/jabref/gui/importer/actions/OpenDatabaseAction.java
```diff
@@ -171,111 +171,71 @@ private void openTheFile(File file, boolean raisePanel) {
...
-            }
 
-            boolean done = false;
-            while (!done) {
-                String fileName = file.getPath();
-                Globals.prefs.put(JabRefPreferences.WORKING_DIRECTORY, file.getParent());
-                // Should this be done _after_ we know it was successfully opened?
+            String fileName = file.getPath();
+            Globals.prefs.put(JabRefPreferences.WORKING_DIRECTORY, file.getParent());
+            // Should this be done _after_ we know it was successfully opened?
```
obraliar: Deleted.

src/main/java/net/sf/jabref/shared/prefs/SharedDatabasePreferences.java
obraliar: `DBMSConnectionProperties.isValid()` is performed by methods which open a shared database from the local file. This method is only a shortcut for all `put*` methods.

matthiasgeiger:
@obraliar Thanks for the explanation!
After a quick look at the code it seems that you have removed the old functionality to autosave after a given time interval. Right? 

I'm not sure whether we'll make our users happy with this changes. 

I know that this feature is inspired by the way IntelliJ is working, but using IntelliJ the files modified are a) mostly under version control, b) IntelliJ also has a good local history to quickly revert to a specific state and c) the undo/redo functionality is working in all cases. 

I don't think that we can assume a) for (the majority of) our  users, b) is not there in JabRef and I'm not sure whether c) is really given...

@JabRef/developers WDYT?

mlep:
My 2 cents (not from a developer), but from a user: As a User I do not like the software to change if there is no obvious gain. Here, the change could be detrimental since I am used to save after I am happy with my changes.
Instead of an "new autosave", I may have preferred this work to be an "autobackup" (so that I do not lose anything even in case of a crash, but I can still decide when the changes are worth saving).
Sorry for not taking part in the discussion of the related issue.

lenhard:
Well, looking at #344 I would expect that the autosave is disabled by default and can be enabled by the user. As long as this is the case, I do not see too much of a problem, since the workflow of most users would not be disturbed. Enabling it by default would be problematic and I would oppose this strongly. Speaking for myself, I want to be in control when the save happens (and I do not really like the IntellJ feature too much anyway).

Regarding dropping or keeping the time-based autosave, I am rather agnostic.

koppor:
Instead of removing the old "Autosave", could you rename it to "backup" and run it each 5 minutes. **Without** any user configuration. This is consistent to other programs such as Word.

koppor:
- [ ] The old preference should **not** be reused, just a new key. Reason: the old "Autosave" is more a "Backup" functionality.

koppor:
- [ ] Please rewrite https://help.jabref.org/en/Autosave. The "new" backup functionality should be mentioned elsewhere.

lenhard:
Basically, the old autosave was no autosave since it only created the backup file. Backup file creation can be done as is, but without the user having a choice about it. So the old preferences can be removed.

The real autosave implemented here is really new and should base on a new preference option, which is disabled by default.

mlep:
@koppor I think it will make sense for autosave and backup to be described in the same help file.

matthiasgeiger:
Yes,  @mlep. Especially as the old backup functionality was named "Autosave" ;-) So it must be made clear that the functionality is now enabled by default (and can not be disabled) and the new functionality is now a real autosave affecting the actual bib-file.

obraliar:
Thanks for the discussion! Here comes the clarification & short summary:

- The **old** autosave functionality has been fully removed due to conceptual errors and unclean code.
- A **new** autosave functionality (which is now **disabled** by default) has been implemented for _local_ and _shared_ databases (see the description on the top).
- A **new** backup functionality (which runs continuously without users influence) has also been implemented for _local_ databases. This one is also event based, while using a separate `.swp` (swap) file for the backup.
- The help file from https://github.com/JabRef/help.jabref.org/pull/108 now only contains helping contents related to the **new** autosave feature including a clarification (see https://github.com/JabRef/help.jabref.org/pull/108/commits/c8bdbe4e2944031835cb3879748bbebec1a1fc27).

For my part, this feature is ready.

src/main/java/net/sf/jabref/autosave/AutosaveManager.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    /**
+     * Shuts down the Autosaver which is associated with the given {@link BibDatabaseContext}.
+     *
+     * @param bibDatabaseContext Associated {@link BibDatabaseContext}
+     */
+    public static void shutdown(BibDatabaseContext bibDatabaseContext) {
+        for (AutosaveManager autosaver : runningInstances) {
+            if (autosaver.bibDatabaseContext == bibDatabaseContext) {
```
tschechlovdev: Why do you use here ``==`` and not ``equals()``?

src/main/java/net/sf/jabref/gui/importer/actions/OpenDatabaseAction.java
```diff
@@ -171,111 +173,74 @@ private void openTheFile(File file, boolean raisePanel) {
...
+            ParserResult result;
+            try {
+                result = OpenDatabase.loadDatabase(fileToLoad, Globals.prefs.getImportFormatPreferences());
+            } catch (IOException ex) {
+                LOGGER.error("Error loading database " + fileToLoad, ex);
+                result = ParserResult.getNullResult();
+            }
+
+            if (result.isNullResult()) {
+                JOptionPane.showMessageDialog(null, Localization.lang("Error opening file") + " '" + fileName + "'",
```
tschechlovdev: Can't this be done in the catch block above?

src/main/java/net/sf/jabref/logic/importer/ParserResult.java
```diff
@@ -49,6 +50,9 @@ public ParserResult(BibDatabase base, MetaData metaData, Map<String, EntryType>
         this.base = base;
         this.metaData = metaData;
         this.entryTypes = entryTypes;
+        if (Objects.nonNull(base) && Objects.nonNull(metaData)) {
```
tschechlovdev: why do you check base and metaData for nonNull but not file?

src/main/java/net/sf/jabref/model/metadata/MetaData.java
```diff
@@ -209,8 +209,15 @@ private void postChange() {
     }
 
     public void setEncoding(Charset encoding) {
+        setEncoding(encoding, true);
+    }
+
+    // The parameter postChanges has been introduced to avoid event loops while saving a database
```
tschechlovdev: Why not change it to a javadoc? Then one can see it, if he wants to call the method ;)

src/main/java/net/sf/jabref/shared/prefs/SharedDatabasePreferences.java
tschechlovdev: So you can guarantee that they are valid when calling this method?

src/main/java/net/sf/jabref/gui/importer/actions/OpenDatabaseAction.java
obraliar: This come out from defensive programming. But yes you are right. Done.

src/main/java/net/sf/jabref/shared/prefs/SharedDatabasePreferences.java
obraliar: It's only a shortcut. But in this context this method is always called with legal parameters.

src/main/java/net/sf/jabref/logic/importer/fileformat/BibtexParser.java
```diff
@@ -175,6 +178,28 @@ private void initializeParserResult() {
...
+            skipWhitespace();
+            char c = (char) read();
+
+            if (c == '%') {
+                skipWhitespace();
+                String label = parseTextToken().trim();
+
+                if (label.equals(BibtexDatabaseWriter.DATABASE_ID_PREFIX)) {
+                    skipWhitespace();
+                    database.setDatabaseID(parseTextToken().trim());
```
koppor: Is it possible to cover also this branch in a test case?

oscargus:
I'd say move `autosave` to `logic.autosave` and `BibDatabaseContextChangedEvent` to `model.database.event`.

src/databaseTest/java/net/sf/jabref/shared/DBMSConnectionTest.java
```diff
@@ -23,13 +25,13 @@
     }
 
     @Test
-    public void testGetConnection() throws SQLException {
+    public void testGetConnection() throws SQLException, InvalidDBMSConnectionPropertiesException {
```
tobiasdiez: (For the next time:) Just use the generic `throws Exception` in tests so that you don't have to edit tests if you decide to throw a different exception.

obraliar:
@tobiasdiez `databaseID` has to be there, as `BibtexParser` is only working on `BibDatabase` while not having a context.

src/main/java/net/sf/jabref/JabRefGUI.java
```diff
@@ -115,18 +114,26 @@ private void openWindow() {
...
                     parserResultIterator.remove();
-                } else if (!pr.isPostponedAutosaveFound()) {
-                    if (pr.toOpenTab()) {
-                        // things to be appended to an opened tab should be done after opening all tabs
-                        // add them to the list
-                        toOpenTab.add(pr);
-                    } else {
-                        JabRefGUI.getMainFrame().addParserResult(pr, first);
-                        first = false;
+                } else if (Objects.nonNull(pr.getDatabase().getDatabaseID())) {
```
tobiasdiez: What is the purpose of this check?

src/main/java/net/sf/jabref/gui/importer/actions/OpenDatabaseAction.java
```diff
@@ -171,111 +173,71 @@ private void openTheFile(File file, boolean raisePanel) {
...
 
-                // After adding the database, go through our list and see if
-                // any post open actions need to be done. For instance, checking
-                // if we found new entry types that can be imported, or checking
-                // if the database contents should be modified due to new features
-                // in this version of JabRef:
-                final ParserResult finalReferenceToResult = result;
-                SwingUtilities.invokeLater(
-                        () -> OpenDatabaseAction.performPostOpenActions(panel, finalReferenceToResult, true));
+            if (Objects.nonNull(result.getDatabase().getDatabaseID())) {
```
tobiasdiez: Same here, I don't understand the purpose of getDatabaseID. You load the db above and what are you now doing in this if statement? I'm probably just to stupid to understand what openSharedDatabaseFromParserResult does. Can you please enlighten me.

tobiasdiez: Ok, thank you very much. Your remarks were indeed helpful. I think I got it now 😄 .
In this case, I would propose to add the db id as a meta-data. Reasoning: all other metadata is also stored there and we had some problems with parsing the %Encoding string. Thus I would minimize all data which is stored in such comments. (As a key, I propose `remoteDatabaseID`. ) 
@obraliar what do you think?

obraliar: I already thought about this option, but this is unfortunately not possible as meta-data, preamble, etc. are also shared. This would end in disaster. This key/ID has to be stored in a local file but it also must not be in shared.

tobiasdiez: Ok, thats a good reason.
So we have two options: store the database id as something special (e.g. as a comment % DB: some id) or we exclude some specific metadata items from the sync (or we create a new metadata which is never synced).
I have no strong opinion about this. What do you think @JabRef/developers ?

obraliar: > So we have two options: store the database id as something special (e.g. as a comment % DB: some id) or (...)

This is exactly the method I'm using now and it works well. I dont see any reasons for changing it.

koppor:
Tested it locally and works good.

koppor:
Went through with @obraliar and we made the fixes in a pair-pgramming style. LGTM. An UI update on the autosave of shared database feature will coming to make it more visible to the users.

src/main/java/net/sf/jabref/gui/exporter/SaveDatabaseAction.java
```diff
@@ -233,7 +243,8 @@ private boolean saveDatabase(File file, boolean selectedOnly, Charset encoding)
         try {
             if (success) {
                 session.commit(file.toPath());
-                panel.getBibDatabaseContext().getMetaData().setEncoding(encoding); // Make sure to remember which encoding we used.
+                // Make sure to remember which encoding we used.
+                panel.getBibDatabaseContext().getMetaData().setEncoding(encoding, ChangePropagation.DO_NOT_POST_EVENT);
```
tobiasdiez: Why don't we want to send a change event here? Probably you don't want to trigger a new save of the file, but the problem with the current solution is that nobody else can listen to encoding changes in a reliable way (not that I have a particular case in mind where listeners to the encoding are important, but I think we shouldn't start with exempting something from the event system).

src/main/java/net/sf/jabref/gui/exporter/SaveDatabaseAction.java
```diff
@@ -304,26 +315,58 @@ public void saveAs() throws Throwable {
...
+            new SharedDatabasePreferences(databaseID).putAllDBMSConnectionProperties(properties);
+        }
+
+        context.setDatabaseFile(file);
         Globals.prefs.put(JabRefPreferences.WORKING_DIRECTORY, file.getParent());
         runCommand();
         // If the operation failed, revert the file field and return:
         if (!success) {
-            panel.getBibDatabaseContext().setDatabaseFile(oldFile);
+            context.setDatabaseFile(context.getDatabaseFile().orElse(null));
```
tobiasdiez: I don't understand this line (set/get from the same context), especially since you set the database file a few lines earlier.

src/main/java/net/sf/jabref/gui/exporter/SaveDatabaseAction.java
```diff
@@ -304,26 +315,58 @@ public void saveAs() throws Throwable {
             }
         }
 
-        File oldFile = panel.getBibDatabaseContext().getDatabaseFile().orElse(null);
-        panel.getBibDatabaseContext().setDatabaseFile(file);
+        BibDatabaseContext context = panel.getBibDatabaseContext();
+
+        if (context.getLocation() == DatabaseLocation.SHARED) {
```
tobiasdiez: This  additional code should be in some logic code (some exporter?) instead of gui.

src/main/java/net/sf/jabref/gui/exporter/SaveDatabaseAction.java
```diff
@@ -304,26 +315,58 @@ public void saveAs() throws Throwable {
...
+            // Generate an ID when saving a shared database
+            String databaseID = new BigInteger(128, new SecureRandom()).toString(32);
+            context.getDatabase().setDatabaseID(databaseID);
```
tobiasdiez: The database should take care of its ID and thus also how a new one is generated -> move this two lines to a new method `createNewID` in the database class. I would also rename everything there from databaseID to simply ID (since it is already a member of the database class so it should be clear from the context)

src/main/java/net/sf/jabref/gui/importer/actions/OpenDatabaseAction.java
```diff
@@ -171,111 +173,71 @@ private void openTheFile(File file, boolean raisePanel) {
...
+                try {
+                    new SharedDatabaseUIManager(frame).openSharedDatabaseFromParserResult(result);
+                } catch (SQLException | DatabaseNotSupportedException | InvalidDBMSConnectionPropertiesException e) {
+                    result.getDatabaseContext().setDatabaseFile(null); // do not open the original file
```
tobiasdiez: I think our convention for setters was that the never except a null value but there is a clear* method for this case.

src/main/java/net/sf/jabref/logic/autosaveandbackup/AutosaveManager.java
```diff
@@ -0,0 +1,94 @@
...
+
+    private AutosaveManager(BibDatabaseContext bibDatabaseContext) {
+        this.bibDatabaseContext = bibDatabaseContext;
+        this.workerQueue = new ArrayBlockingQueue<>(1);
+        this.executor = new ThreadPoolExecutor(1, 1, 0, TimeUnit.SECONDS, workerQueue);
+        this.eventBus = new EventBus();
+    }
+
+    @Subscribe
+    public synchronized void listen(@SuppressWarnings("unused") BibDatabaseContextChangedEvent event) {
```
tobiasdiez: What is the purpose of this class? It looks like it just reposts the event, right? Why can't one directly subscribe to the BibDatabaseContextChangedEvents of the BibDatabaseContext?

But this class might be a solution for the above issue with setEncoding and not firing the event. You pause the AutosaveManager once you begin with writing the file so that changes (SaveActions?) during the writing process don't trigger an additional writing.

src/main/java/net/sf/jabref/logic/autosaveandbackup/AutosaveManager.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    /**
+     * Starts the Autosaver which is associated with the given {@link BibDatabaseContext}.
+     *
+     * @param bibDatabaseContext Associated {@link BibDatabaseContext}
+     */
+    public static AutosaveManager start(BibDatabaseContext bibDatabaseContext) {
+        AutosaveManager autosaver = new AutosaveManager(bibDatabaseContext);
+        bibDatabaseContext.getDatabase().registerListener(autosaver);
```
tobiasdiez: I think it is not a good idea that the metadata and the database post BibContextChanged events...because they don't know the context, only the context knows about them. So I think the proper way to implement it is that the BibContext listens to events of its metadata and db and then reposts them as BibContextChanged events. In this way, the autosavemanager also always listens to the correct db/metadata (which is not guaranteed in the current solution if you change the metadata via setMetadata of the context)

src/main/java/net/sf/jabref/logic/autosaveandbackup/BackupManager.java
```diff
@@ -0,0 +1,159 @@
...
+
+import com.google.common.eventbus.Subscribe;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+/**
+ * Backups the given bib database file from {@link BibDatabaseContext} on every {@link BibDatabaseContextChangedEvent}.
+ * An intelligent {@link ExecutorService} with a {@link BlockingQueue} prevents a high load while making backups and rejects all redundant backup tasks.
+ */
+public class BackupManager {
```
tobiasdiez: Could you please also add a comment, that this backup is not about the bak-file which is also created upon saving.
Maybe we should distinguishes between Backup (*bak) and Live Backup (swp), or whatever we want to call these concepts.

src/main/java/net/sf/jabref/logic/exporter/BibDatabaseWriter.java
```diff
@@ -158,6 +158,8 @@ public E savePartOfDatabase(BibDatabaseContext bibDatabaseContext,
 
         session = saveSessionFactory.createSaveSession(preferences.getEncodingOrDefault(), preferences.getMakeBackup());
 
+        writeDatabaseID(bibDatabaseContext.getDatabase().getDatabaseID());
```
tobiasdiez: Only write it if we don't want to save in plan bibtex mode (see below, line 198)

src/main/java/net/sf/jabref/logic/exporter/BibtexDatabaseWriter.java
```diff
@@ -137,6 +139,25 @@ protected void writePrelogue(BibDatabaseContext bibDatabaseContext, Charset enco
...
+
+        } catch (IOException e) {
+            throw new SaveException(e);
+        }
+    }
+
+    @Override
+    protected void writeDatabaseID(String databaseID) throws SaveException {
+        try {
+            if (Objects.nonNull(databaseID)) {
```
tobiasdiez: Please don't use null values for databaseID. (also check for empty string, StringUtils.isBlank)

src/main/java/net/sf/jabref/logic/importer/ParserResult.java
```diff
@@ -167,7 +163,19 @@ public void setErrorMessage(String errorMessage) {
     }
 
     public BibDatabaseContext getDatabaseContext() {
-        return new BibDatabaseContext(base, metaData, file);
+        if (Objects.isNull(this.bibDatabaseContext)) {
```
tobiasdiez: I think the good old  `something == null` is more readable. According to the documentation, Objects.isNull just exists to be used as a predicate (i.e. ̀`filter(Objects::isNull)`)

src/main/java/net/sf/jabref/shared/prefs/SharedDatabasePreferences.java
```diff
@@ -1,18 +1,29 @@
 package net.sf.jabref.shared.prefs;
 
+import java.io.UnsupportedEncodingException;
+import java.security.GeneralSecurityException;
 import java.util.Optional;
 import java.util.prefs.BackingStoreException;
 import java.util.prefs.Preferences;
 
-import net.sf.jabref.Globals;
```
tobiasdiez: Now that this globals is removed, I think the shared namespace can be moved to logic.

src/test/java/net/sf/jabref/logic/importer/fileformat/BibtexImporterTest.java
```diff
@@ -121,4 +121,12 @@ public void testGetDescription() {
                         + "The metadata is not required to be read here, as this class is NOT called at --import.",
                 importer.getDescription());
     }
+
+    @Test
+    public void testRecognizesDatabaseID() throws Exception {
+        Path file = Paths.get(BibtexImporterTest.class.getResource("AutosavedSharedDatabase.bib").toURI());
```
tobiasdiez: Could you please add the database id also to complex.bib. This test file should illustrate every feature Jabref has. Then also an additional `AutosavedSharedDb.bib` is no longer necessary.

src/test/java/net/sf/jabref/logic/importer/fileformat/BibtexParserTest.java
```diff
@@ -1407,6 +1407,24 @@ public void parseRecognizesSaveActionsAfterEntry() throws IOException {
...
+                .append(OS.NEWLINE)
+                .append("@Article{a}");
+
+        ParserResult parserResult = parser.parse(new StringReader(sharedDatabaseFileContent.toString()));
+
+        String actualDatabaseID = parserResult.getDatabase().getDatabaseID();
+
+        assertEquals(expectedDatabaseID, actualDatabaseID);
+    }
+
```
tobiasdiez: Please also add a test that the database id comment is not recognized as the comment of the first entry.

koppor:
The more test cases, the better. I would not remove any test file to ensure
high test coverage.

A DBID is uncommon, therefore I would create a second complex.bib.

If we want, we can do a generation of all variants of compley.bib (with
encoding info, without, with special fields, without, with old groups, with
new grouos, ...) and check them all.

Currently, complex.bib is not written with a recent JabRef version (casing
at the entry types is wrong).

tobiasdiez:
The roundtrip tests only work for the complex.bib and thus I would just add everything in this file. And for this particular test, there is no reason to have a special test file instead of creating the input directly as a string in the test method.

src/main/java/net/sf/jabref/gui/exporter/SaveDatabaseAction.java
obraliar: > Probably you don't want to trigger a new save of the file

It's even worse, this would cause an event loop which would never end.

> the problem with the current solution is that nobody else can listen to encoding changes in a reliable way

There is still the old method `setEncoding(Charset encoding)` which is now a shortcut for `setEncoding(encoding, ChangePropagation.POST_EVENT)`. Every time you set the encoding the event is also going to be posted (which will also trigger the autosave). It's ok so far. But the event is not going to be posted while saving a file. I'm more skaptical why this is called here at all. The old comment `// Make sure to remember which encoding we used.` is not really justified, as the encoding should be set immediately when it changes and not "at the last second".

src/main/java/net/sf/jabref/gui/exporter/SaveDatabaseAction.java
obraliar: Now I moved the logic part as much as possible from here. But nevertheless there are multiple methods like [`run(...)`](https://github.com/JabRef/jabref/blob/master/src/main/java/net/sf/jabref/gui/exporter/SaveDatabaseAction.java#L114) or even worse like [`saveDatabase(...)`](https://github.com/JabRef/jabref/blob/master/src/main/java/net/sf/jabref/gui/exporter/SaveDatabaseAction.java#L166) which radically violate the _MVC pattern_. So this is a general problem in jabRef which requires a special attention.

src/main/java/net/sf/jabref/gui/exporter/SaveDatabaseAction.java
obraliar: Done.

src/main/java/net/sf/jabref/gui/importer/actions/OpenDatabaseAction.java
obraliar: Done.

src/main/java/net/sf/jabref/logic/autosaveandbackup/AutosaveManager.java
obraliar: > What is the purpose of this class?

Event classes `EntryEvent`, `MetaDataChangedEvent`, etc. extend  `BibDatabaseContextChangedEvent`. E.g., every time an `EntryEvent` is posted, the `BibDatabaseContextChangedEvent` will get posted too. So this class summarizes multiple eventclasses. Otherwise we would have to listen for each Event in separate methods.

> It looks like it just reposts the event, right? 

If you look closer, a new `AutosaveEvent` is posted within a submission to the executor. This is notifying a class within the `gui` package which actually starts `new SaveDatabaseAction(panel).runCommand();` (I'm not really happy with the logical code of `SaveDatabaseAction` as this is placed whithin a `gui` package, but we have to use it.)

Now the `executor` submits the tasks in asynchronous way while the eventBus is synchronous. Therefore a new Event is used. Of course I coud move `...runCommand();` here, but this would violate the _MVC pattern_. So semantically this does not repost the event.

src/main/java/net/sf/jabref/logic/autosaveandbackup/AutosaveManager.java
obraliar: > I think it is not a good idea that the metadata and the database post BibContextChanged events

metadata and database don't post the `BibContextChanged` at all. This is done by the hierarchical support of the event system. See the comment above.

src/main/java/net/sf/jabref/logic/autosaveandbackup/BackupManager.java
obraliar: Done.

src/main/java/net/sf/jabref/logic/exporter/BibtexDatabaseWriter.java
obraliar: This check is not needed any more. Done.

src/main/java/net/sf/jabref/logic/importer/ParserResult.java
obraliar: Good to know. Done.

src/main/java/net/sf/jabref/logic/autosaveandbackup/AutosaveManager.java
tobiasdiez: In the end they do post BibContextChanged, since this is now the base class for EntryEvent and MetadataChangedEvent (and this is something I don't like too much). Hence my proposal, that the BibContext throws new BibContextChanged events which wrap EntryEvent/MetadataChanged (how was this called...composition over inheritance?)

src/main/java/net/sf/jabref/logic/exporter/BibDatabaseWriter.java
tobiasdiez: I think plain bibtex is used when the user want's to export the bib file and send it to somebody not using JabRef (e.g. when submitting to a journal). Then all the JabRef things like metadata are removed.

PR #2120, opened 2016-10-05
Change title/tooltip of source panel in biblatex mode

lenhard:
This is a minor change to the source panel, which used to always display "BibTeX source" as title and a similar tooltip. Now, that is dependent on the mode of the database and reads "BibLaTeX source" in biblatex mode.

### DIFF
--- a/src/main/resources/l10n/JabRef_fa.properties
+++ b/src/main/resources/l10n/JabRef_fa.properties
@@ -153,8 +153,7 @@ Backup_old_file_when_saving=
 
 BibTeX_key_is_unique.=
 
-BibTeX_source=
-BibLaTeX_source=
+%0_source=Source_%0
 
 Broken_link=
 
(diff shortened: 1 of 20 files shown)


src/main/resources/l10n/JabRef_fa.properties
```diff
@@ -1149,7 +1149,7 @@ Settings=
 
 Shortcut=
 
-Show/edit_BibTeX_source=
+Show/edit_%0_source=Montrer/�diter_le_source_BibTeX
```
bartsch-dev: 1. this file is not in `utf-8` thus the `é` makes problems with the python script (it should make problems with JabRef too as they're loaded with `utf-8` in Java)
- isn't this french?

lenhard: Indeed, I copied this into the wrong language file... Thanks a lot for spotting!

I will fix this directly on master, but I am actually not sure what to do with the UTF-8 problem. I copied the string from the french localization file. Do encodings differ between the localization files?

matthiasgeiger: They all should be UTF-8.

matthiasgeiger: @lenhard Can you change the encoding of the farsi file? Should be no problem there are no translations in this file.

Siedlerchr: And make sure that it's UTF-8 without BOM..

PR #2124, opened 2016-10-05
Unify ParserException and ParseException

lenhard:
We have two very similar exceptions in the code that are used for several very similar purposes: `ParseException` and `ParserException`. This PR replaces the latter with former. Additionally, `ParseException` is moved from `model` to the exact same place of `ParserException`. Its prior position in `model` was architecturally misplaced and its usages in `model` have been replaced with `IllegalArgumentException`, which is more suitable for the actual purpose there.

No need for a changelog entry, this is purely internal.


Siedlerchr:
LGTM, too! Merge it in

Issue #2105, opened 2016-09-30
Entry preview not updated properly


AEgit:
JabRef 3.7dev--snapshot--2016-09-30--latex-free-fields--0ee1d17
windows 10 10.0 amd64 
Java 1.8.0_101

Select a group that contains articles
1. Perform a search
2. One of the articles within that group is found by the search and is selected (indicated in blue)
3. The entry preview of the selected article is displayed
4. If we select now another article in the same group that is also found by the search, the entry preview is not shown - it still shows the previous article. This behaviour doesn't seem to happen in a repoducible way. It could be that only newly created entries (indicated in green) are affected by this, i. e. if you switch over to a newly created entry, the entry preview of this entry won't be shown. Instead it will show the entry preview of the previous entry,
5. This does not apply to articles that are not part of the same group (indicated in grey)

This issue was first mentioned here:
https://github.com/JabRef/jabref/issues/1993#issue-177367050

Similar (but different) problems have been mentioned here:
https://github.com/JabRef/jabref/issues/1760
https://github.com/JabRef/jabref/issues/1903

koppor:
@bartsch-dev Maybe you have time to have a look at the end of next week?

AEgit:
Ok, so it appears that newly created entries and entries that are newly assigned to a group are affected by this behvaiour.

To reproduce it, the following steps need to be taken:

1. Select a group that contains articles/items
2. Create a new article/item in this group or assign an existing article from another group to this group
3. Select another article that is or is not part of the group: The entry preview of that article is correctly displayed.
4. Now select the newly created/assigned article: The entry preview is NOT updated properly. Instead of showing the entry preview of the selected article, the entry preview of the previously selected article is still shown.
If you select another article and then select again the newly created/assigned article, the problem persists: The entry preview of the selected newly created/assigned article is not shown. Instead, the entry preview of the article that you just selected previously is shown.
5. If you create/assign another new article to the group, it exhibits the same problems. The "old new article" (i. e. the one, that we crated at step 2) behaves now properly and shows the entry preview when selected.

AEgit:
This bug is also present in the current master build:
JabRef 3.7dev--snapshot--2016-10-03--master--565676c

bartsch-dev:
The issue can be reproduced even easier.

1. have at least 2 entries
-  open the entry editor and close it 
  - the preview will be rendered correctly
- select the other entry 
  - the preview will be rendered correctly
- select the first entry (the one you opened the editor with)
[...]

PR #2125, opened 2016-10-05
Fix entry preview not updated properly

bartsch-dev:
Fixes: #2105.

### Steps to reproduce:
1. have at least 2 entries
- open the entry editor and close it 
  - the preview will be rendered correctly
- select the other entry 
  - the preview will be rendered correctly
- select the first entry (the one you opened the editor with)
- the preview will not be updated

The problem is that the editor variable doesn't get cleared when the editor closes, thus the bottom component in the main table doesn't get updated if that entry is the last one which had the editor open.

### DIFF
--- a/src/main/java/net/sf/jabref/gui/maintable/MainTableSelectionListener.java
+++ b/src/main/java/net/sf/jabref/gui/maintable/MainTableSelectionListener.java
@@ -102,11 +102,6 @@ public class MainTableSelectionListener implements ListEventListener<BibEntry>, 
         }
 
         final BibEntry newSelected = selected.get(0);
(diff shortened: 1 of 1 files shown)


lenhard:
I have tested this locally and can reproduce that your changes fix the bug. Since it is only about five lines, I think it is save if I merge this even without a second review.

Thanks for your contribution!

obraliar:
Since this commit there were problems with the field focus.
Steps to reproduce:
- Create a new database
- Create a new entry
- Change one field like `author`
- Click on another field like `title`

--> The `title` field does **not** request the focus as before.

(discovered using `git bisect`)

bartsch-dev:
Bug mentioned by @obraliar is fixed by https://github.com/JabRef/jabref/pull/2138.

Issue #1938, opened 2016-09-08
Cursor jumps away from the search field


radeksimik:
JabRef version 3.6 on Mac OS X 10.11.6

Steps to reproduce:

1. Typing a string to be searched into the search field on the top left of the window.
2. At the same time, the entry editor (at the bottom) is open.
3. Problem: Sometimes (didn't find the pattern) the text-cursor jumps away from the search field into one of the slots of the entry editor. From that point on, typing happens not in the search field but in the slot of the entry editor. This can cause retyping (so: deletion and then typing) of some of the information within the entry editor. (Luckily, this can undone by the Undo command.)

Essentially, this problem makes me search only when the entry editor is closed/deactivated. Pretty annoying.

```
Put the excerpt of the log file here
```

AEgit:
I can confirm this problem - unfortunately I've also not been able to find the pattern behind it, so I did not report the bug.

cr333:
Same for me with JabRef 3.6 on Windows 10 (1511). The cursor jumps away sometimes (but not every time) when the partial search term is a valid text completion. For example, searching for "bilateral" may mean that the cursor jumps away after "bi" as this is the surname of some author (Bi). But this is somehow not reproducible.

lenhard:
I have been trying to reproduce this, but fail to do so. My cursor stays in the search field, also with search terms that are valid text completions.

Sorry, but as long as we haven't found a way of reproducing this, there is also no chance of fixing it.

radeksimik:
It's frustrating. I just sat down to try to find the pattern, but have failed to reproduce the effect even once. I want to mention two points, just in case they help others look for the pattern:

- The target of the jumping was often (though not exclusively) the field "Crossref". The field is always empty in my database, so it's not the case that the target of the jumping is the string being searched for.
-  My impression is that slow typing won't cause the jumping effect. It only seems to happen when I'm typing with all 10 fingers, in a significant speed.
- Don't know whether this matters, but my database is fairly large, containing 6000 entries.

I'll try to observe this further and let you know once I know more. The bug is pretty annoying mainly because it has a potential to mess up my carefully built database. I already started having bibtex-compilation problems because I've unintentionally added strings that caused trouble (e.g. into the "Crossref" field).

AEgit:
I've also tried to reproduce it, but as you say it kind of happens in a random way. One point that you mention is interesting: You have a large database and so have I (currently >8000 entries). It would be interesting to see, if everyone who is affected, has a large database.

[...]

PR #2129, opened 2016-10-06
Don't change the selected entry after search if you edit an entry

### LINKED ISSUE #2142: New entry is not focused when search is active
1. open `complex.bib`
2. sort by `author/editor`
3. search for `bard`
4. double click on "Boström" (which is not a search hit) -> entry editor opens
5. BibTeX -> New entry
6. Press "Article"

Expected: Entry editor for the new article opens
Behavior: First search hit is focused.

bartsch-dev:
Fixes: https://github.com/koppor/jabref/issues/172.
**Edit:** Fixes #2142.

When the search is active and you edit an entry until the entry doesn't match the search an longer the first match would be selected.

Now, if you edit an entry the selected entry won't change.

### DIFF
--- a/src/main/java/net/sf/jabref/gui/BasePanel.java
+++ b/src/main/java/net/sf/jabref/gui/BasePanel.java
@@ -1301,6 +1301,7 @@ public class BasePanel extends JPanel implements ClipboardOwner, FileUpdateListe
 
         @Subscribe
         public void listen(EntryRemovedEvent removedEntryEvent) {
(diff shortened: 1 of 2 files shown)


lenhard:
I tested this locally and LGTM!

bartsch-dev:
Will also fix https://github.com/JabRef/jabref/issues/2142.

matthiasgeiger:
This potentially also fixes https://github.com/JabRef/jabref/issues/1938, I think.

PR #2132, opened 2016-10-06
Remove some UnicodeToLatex uses

bartsch-dev:
I removed some uses of the `UnicodeToLatexFormatter` and used the latexFreeField directly.

### DIFF
--- a/src/main/java/net/sf/jabref/logic/importer/fetcher/CrossRef.java
+++ b/src/main/java/net/sf/jabref/logic/importer/fetcher/CrossRef.java
@@ -4,6 +4,7 @@ import java.util.Locale;
 import java.util.Objects;
 import java.util.Optional;
 
+import net.sf.jabref.logic.formatter.bibtexfields.RemoveBracesFormatter;
 import net.sf.jabref.logic.util.DOI;
 import net.sf.jabref.model.entry.BibEntry;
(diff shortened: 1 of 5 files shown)


src/main/java/net/sf/jabref/logic/importer/fetcher/CrossRef.java
```diff
@@ -114,16 +111,6 @@ private static boolean checkValidity(BibEntry entry, JSONArray result) {
         }
     }
 
-    private static String removeLaTeX(String text) {
-        String result;
-        // remove braces
-        result = new RemoveBracesFormatter().format(text);
```
lenhard: This also removes the `RemoveBracesFormatter`, which is not part of the computation of latex-free fields. So, if I understand the code correctly, this changes functionality.

bartsch-dev: I re-added it, but I don't know if its important. The latex-free field and the removeBraces are only used in the validation method, in the `findDOI` method the normal field is used.
Maybe someone has to look over it.

lenhard: Agreed! At least for this PR, this is now good enough.

Issue #2136, opened 2016-10-07
Cannot switch with tab between different fields in (manually) newly created entry


AEgit:
JabRef 3.7dev--snapshot--2016-10-07--master--03bc4a3
windows 10 10.0 amd64 
Java 1.8.0_101

1. Create a new entry in your database manually
2. Start entering the names of the authors
3. Try to switch to the "Title" field using the tabulator
4. This does not work.
5. If you close the editing mode and reopen it, you are now able to switch/hop between different fields using tab

This is a regression that was introduced between these two versions:
 JabRef 3.7dev--snapshot--2016-10-03--master--565676c
JabRef 3.7dev--snapshot--2016-10-07--master--03bc4a3

AEgit:
Another workaround (instead of leaving the 'edit' mode) is to save the new entry. After having saved it at least once, switching between differnt fields using tab is possible.

lenhard:
@bartsch-dev: Could you have a look?

bartsch-dev:
I'll look into it.

koppor:
Thank you for reporting this issue. We think, that is already fixed in our development version and consequently the change will be included in the next release.

We would like to ask you to use a development build from https://builds.jabref.org/master.

AEgit:
I can confirm that this problem has been solved in the following development build:

JabRef 3.7dev--snapshot--2016-10-11--master--3459ec3

Thank you very much for your great work!

PR #2138, opened 2016-10-08
Fix switching edited textfield in the entry editor with TAB

bartsch-dev:
Fixes https://github.com/JabRef/jabref/issues/2136

### Steps to reproduce
1. open any entry
- edit any field
- try switching to the next field with <kbd>TAB</kbd>
- focus is taken away from the next field


koppor:
LGTM! - Maybe, it's easy for you to fix https://github.com/koppor/jabref/issues/2, too.

PR #2139, opened 2016-10-08
Replaced a number of calls to BibtexParser.parse

oscargus:
Some search and replace to remove deprecated BibtexParser.parse calls. Far from all though...

### DIFF
--- a/src/main/java/net/sf/jabref/gui/importer/fetcher/ACMPortalFetcher.java
+++ b/src/main/java/net/sf/jabref/gui/importer/fetcher/ACMPortalFetcher.java
@@ -313,8 +313,7 @@ public class ACMPortalFetcher implements PreviewEntryFetcher {
             Collection<BibEntry> items = null;
             try (BufferedReader in = new BufferedReader(
                     new InputStreamReader(connection.getInputStream(), StandardCharsets.UTF_8))) {
-                items = BibtexParser.parse(in, Globals.prefs.getImportFormatPreferences()).getDatabase()
-                        .getEntries();
+                items = BibtexParser.parse(in, Globals.prefs.getImportFormatPreferences()).getDatabase().getEntries();
             } catch (IOException e) {
                 LOGGER.info("Download of BibTeX information from ACM Portal failed.", e);
             }
--- a/src/main/java/net/sf/jabref/gui/importer/fetcher/ACMPortalFetcher.java
+++ b/src/main/java/net/sf/jabref/gui/importer/fetcher/ACMPortalFetcher.java
@@ -313,7 +313,8 @@ public class ACMPortalFetcher implements PreviewEntryFetcher {
             Collection<BibEntry> items = null;
             try (BufferedReader in = new BufferedReader(
                     new InputStreamReader(connection.getInputStream(), StandardCharsets.UTF_8))) {
-                items = BibtexParser.parse(in, Globals.prefs.getImportFormatPreferences()).getDatabase().getEntries();
+                items = new BibtexParser(Globals.prefs.getImportFormatPreferences()).parse(in).getDatabase()
+                        .getEntries();
             } catch (IOException e) {
                 LOGGER.info("Download of BibTeX information from ACM Portal failed.", e);
             }
--- a/src/test/java/net/sf/jabref/logic/groups/GroupsUtilTest.java
+++ b/src/test/java/net/sf/jabref/logic/groups/GroupsUtilTest.java
@@ -28,7 +28,7 @@ public class GroupsUtilTest {
         try (BufferedReader fr = Files.newBufferedReader(Paths.get("src/test/resources/testbib/testjabref.bib"),
                 StandardCharsets.UTF_8)) {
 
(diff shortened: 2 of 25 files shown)


src/main/java/net/sf/jabref/gui/importer/fetcher/ACMPortalFetcher.java
```diff
@@ -313,8 +313,7 @@ private boolean getNextEntryURL(String allText, int entryNumber,
             Collection<BibEntry> items = null;
             try (BufferedReader in = new BufferedReader(
                     new InputStreamReader(connection.getInputStream(), StandardCharsets.UTF_8))) {
-                items = BibtexParser.parse(in, Globals.prefs.getImportFormatPreferences()).getDatabase()
-                        .getEntries();
+                items = BibtexParser.parse(in, Globals.prefs.getImportFormatPreferences()).getDatabase().getEntries();
```
stefan-kolb: new Bib..?

src/test/java/net/sf/jabref/logic/groups/GroupsUtilTest.java
```diff
@@ -28,8 +28,7 @@ public void test() throws IOException {
         try (BufferedReader fr = Files.newBufferedReader(Paths.get("src/test/resources/testbib/testjabref.bib"),
                 StandardCharsets.UTF_8)) {
 
-            ParserResult result = BibtexParser.parse(fr,
-                    JabRefPreferences.getInstance().getImportFormatPreferences());
+            ParserResult result = BibtexParser.parse(fr, JabRefPreferences.getInstance().getImportFormatPreferences());
```
stefan-kolb: new Bib?

src/test/java/net/sf/jabref/logic/layout/LayoutTest.java
```diff
@@ -44,7 +44,7 @@ public String t1BibtexString() {
 
     public static BibEntry bibtexString2BibtexEntry(String s) throws IOException {
         ParserResult result = BibtexParser.parse(new StringReader(s),
-JabRefPreferences.getInstance().getImportFormatPreferences());
+                JabRefPreferences.getInstance().getImportFormatPreferences());
```
stefan-kolb: Indent?

src/test/java/net/sf/jabref/logic/groups/GroupsUtilTest.java
oscargus: Same here (although I could have rerun the same, was traveling and then ran out of time).

src/test/java/net/sf/jabref/logic/layout/LayoutTest.java
oscargus: Better now. (Although not actively changed as far as I recall).

src/test/java/net/sf/jabref/logic/importer/fileformat/BibtexParserTest.java
```diff
@@ -132,7 +132,8 @@ public void singleFromStringReturnsEmptyIfNoEntryRecognized() throws IOException
 
     @Test
     public void parseRecognizesEntry() throws IOException {
-        ParserResult result = new BibtexParser(importFormatPreferences).parse(new StringReader("@article{test,author={Ed von Test}}"));
+        ParserResult result = new BibtexParser(importFormatPreferences)
```
tobiasdiez: Could you also please extract `new BibtexParser(prefs)` the the setup method.

oscargus: Well, I changed about 100 or so calls marked deprecated by someone else (as far as I can tell, very few of the similar calls were actually replaced by whoever marked it as deprecated, not even in the actual code). There are still things to clean up before the methods are removed and all tests there can make use of it.

tobiasdiez: We talked about the `deprecated` issue in the last devcall. Since you weren't there, I give you a short summary. 
The consensus was that we should concentrate more on new features / bug fixing instead of cleaning up the code since the code quality is rather high now (at least in logic + model). But of course there are still some problematic code, for example superfluous methods / strange implementation / things which look like duplicated code but there might be subtle differences. If one works one some PR and notice such a flaw in the code, then one can simple mark the corresponding method as deprecated (and document why one did this). Then the deprecated flag just means "try not to use this method anymore". The hope was that while modifying some code which still uses the deprecated method one has enough context to be able to decide if the corresponding call can be replaced by something else. Thus at some point the usage of the deprecated method dies over time and at some point it can be replaced. So the aim was to reduce the amount of pure refactoring/cleanup PRs and let the code evolve over time in the right direction by smaller changes (with context). This of course doesn't mean that cleanup PRs like this one here don't have value - in contrast. But they are are often harder to review.

Anyway, please clean the tests (ready-for-review removed until then)

lenhard:
I hijacked this PR now in #2414, just to move onwards and get the contribution here into master. All made here are appropriately preserved.

PR #2141, opened 2016-10-08
Code cleanups

oscargus:
Removed some unused arguments.
Used constants from top level definition.

I also added some comments to consider during the review.


src/main/java/net/sf/jabref/JabRefGUI.java
```diff
@@ -263,7 +264,7 @@ private void setLookAndFeel() {
                     && !System.getProperty("java.runtime.name").contains("OpenJDK")) {
                 // try to avoid ending up with the ugly Metal L&F
                 Plastic3DLookAndFeel lnf = new Plastic3DLookAndFeel();
-                Plastic3DLookAndFeel.setCurrentTheme(new SkyBluer());
+                MetalLookAndFeel.setCurrentTheme(new SkyBluer());
```
stefan-kolb: Why did you change that to Metal as we want to avoid it here?

oscargus: Apparently Plastic is inheriting the method from Metal.

src/main/java/net/sf/jabref/gui/EntryCustomizationDialog.java
```diff
@@ -376,10 +378,11 @@ public void actionPerformed(ActionEvent e) {
      * a valid type, that no obsolete entry editors are around, and that
      * the right-click menus' change type menu is up-to-date.
      */
-    private void updateTypesForEntries(String typeName) {
+    private void updateTypesForEntries() {
         for (BasePanel bp : frame.getBasePanelList()) {
 
             for (BibEntry entry : bp.getDatabase().getEntries()) {
+                // Please explain what this line, and therefore the whole method, is doing (if it is doing something)
                 EntryTypes.getType(entry.getType(), bibDatabaseMode).ifPresent(entry::setType);
```
koppor: I think, it is because the types are objects. If one switches from bibtex to biblatex, the types get different objects even though the string representation is the same. This loop ensures that the type objects match the type of the database.

koppor:
I just provided the comment on the unclear method - I think, @stefan-kolb can approve.

src/main/java/net/sf/jabref/model/ChainNode.java
```diff
@@ -126,6 +126,7 @@ protected void setParent(T parent) {
      *
      */
     public void removeChild() {
+        // NPE if this is ever called
         child.setParent(null);
```
tobiasdiez: Could you please fix it by wrapping everything in a `if( child != null)`

oscargus: Isn't it better that the author of the faulty code fixes it when the implementation is finished? I do not really see how this code was ever approved...

src/main/java/net/sf/jabref/gui/EntryCustomizationDialog.java
oscargus: OK! I'll remove the comment.

src/main/java/net/sf/jabref/model/ChainNode.java
oscargus: Oh, yes. The problem is actually passing `null` to a metod with an `Objects.requireNonNull` as the first call... So clearly that code was never executed.

tobiasdiez: > Isn't it better that the author of the faulty code fixes it

The nice thing about open source code is that there is no "your / mine code", everyone is responsible for all the code. I can't count how many times I already fixed some small bugs which somebody else created and similarly many others (including you) cleaned up my mess.

Please just add the if statement there and then merge this in directly (ready-for-review removed for this reason).

oscargus: The first comment was not really relevant here. Again, the thing is not that `child` becomes `null`, although that will also break the code, but that whenever `child.setParent(null)` is called there will be an NPE as the code for `setParent`reads (second comment):
```
     protected void setParent(T parent) {
         this.parent = Objects.requireNonNull(parent);
}
```
Hence, yes, I can add the if statement, but that is not really solving the problem I pointed out in the comment.

oscargus:
As I added the if-statement (kept the comment though, as described above), I merge this now.

Issue #1670, opened 2016-08-03
FossHub hacked: Is Jabref also affected?


AEgit:
FossHub has been hacked, see:

https://www.reddit.com/r/pcmasterrace/comments/4vw21h/massive_psa_do_not_download_classic_shell_read/

and

http://www.heise.de/newsticker/meldung/FossHub-kompromittiert-Software-Installer-mit-Malware-infiziert-3286347.html
(German only)

The software installer has been infected with a Malware that overwrites the boot loader.

As Jabref ist offered on FossHub as well, I was just wondering, whether the offered .exe files were affected as well or not?

matthiasgeiger:
Thanks for informing us about this issue.

The files that are the moment distributed by fosshub don't seem to be affected.

The Signatures shown at fosshub are the right ones and the files hosted there still have the same checksum:
```
 JabRef Windows Installer (64 bit) - 25.13 MB | version: 3.5
MD5: fdf4fa0e33019b2882c39e2066c65fb5
SHA1: 8709f73a4204b882fec7ef6b9877774708a13f8d
SHA256: 9438a6762db4e4504793d5a1dd05a18f7772fbe0d77c06eaeb83e4b80bd5922c

JabRef Platform independent runnable JAR - 23.83 MB | version: 3.5
MD5: c63caa885622c9057c6e926ffc0ed132
SHA1: e6892ad307947429311db8db452b607ce3e2b0af
SHA256: b5a9148008005f7ba081614693275d6b9226875286e99cda47d2648c583708a7 

JabRef Mac OS X - 24.46 MB | version: 3.5
MD5: ffd9845d1d3769324ec371e6cd515957
SHA1: da1fcdebcc70658fe0ecd7827e2a8f88855392b5
SHA256: fea9b90668c94c31e18c0a6593adacb84e1247aa4f937ddeb8719deb2a900e4c 

JabRef Windows Installer (32 bit) - 25.04 MB | version: 3.5
MD5: a3765d5568b88bdb9197725f04d7a15e
SHA1: 0d7593fbca6dbb8270ca9fc2674159a3477be9c0
SHA256: 1f891a7c539844594fb50aab073db1d8c5d55a9d52fbcf669546f82d82da1469 
```

So - as far as I can tell ATM the JabRef installers should do no harm. Nevertheless we'll link to other trustworthy download sources until the situation at fosshub.com has been investigated fixed.

AEgit:
Glad to hear I was of any help!

[...]

PR #2145, opened 2016-10-09
Change download URL to downloads.jabref.org

koppor:
We want to get back control of the download site we offer. Thus, we created https://downloads.jabref.org (https://github.com/JabRef/downloads.jabref.org) with a redirection to fosshub.

Refs #1670

### DIFF
--- a/src/main/java/net/sf/jabref/logic/util/Version.java
+++ b/src/main/java/net/sf/jabref/logic/util/Version.java
@@ -17,7 +17,7 @@ import org.json.JSONObject;
  */
 public class Version {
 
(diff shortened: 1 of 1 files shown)


koppor:
Pro:
 * more convenience for the users

Con
 * We have to update the deep links for each release
 * The users won't see the welcome page of FossHub any more. This is different to sourceforge - there, a "deep" link leads to a landing page, too: https://sourceforge.net/projects/jabref/files/v3.6/JabRef_windows_3_6.exe/download
 * We have to implement an operating system detection

PR #2147, opened 2016-10-10
Added gray background text to authors field to assist newcomers

boceckts:
I added a gray background text to the "Author" and "Editor" field that appears when the fields don't contain any text. See https://github.com/koppor/jabref/issues/61 for reference.

editorplaceholder

placeholderfocused

- [x] Interal quality assurance


src/main/java/net/sf/jabref/gui/fieldeditors/JTextAreaWithUnfocusedText.java
```diff
@@ -0,0 +1,69 @@
...
+        this.addFocusListener(this);
+    }
+
+    @Override
+    protected void paintComponent(Graphics g) {
+        super.paintComponent(g);
+
+        if (!this.hasFocus() && this.getText().isEmpty()) {
+            Font prev = g.getFont();
+            Color prevColor = g.getColor();
```
tschechlovdev: Maybe you could rename some variables here.
g- > graphic
h -> height

src/main/java/net/sf/jabref/gui/fieldeditors/JTextAreaWithHighlighting.java
```diff
@@ -35,13 +34,15 @@
     private UndoManager undo;
 
     public JTextAreaWithHighlighting() {
-        super();
-        setupUndoRedo();
-        setupPasteListener();
+        this(null);
```
koppor: Why not `""` instead of `null`?

src/main/java/net/sf/jabref/gui/entryeditor/EntryEditorTab.java
```diff
@@ -151,7 +153,8 @@ private void setupPanel(JabRefFrame frame, BasePanel bPanel, boolean addKeyField
                         false);
                 defaultHeight = 0;
             } else {
-                fieldEditor = new TextArea(field, null);
+                String prompt = (field.equals(FieldName.AUTHOR)) ? Localization.lang("Firstname Lastname and Firstname and Lastname and others") : "";
```
koppor: Wrong text

It should be `Firstname Lastname and Firstname Lastname`. ` and others` must not be translated as it is a fixed bibtex term used to write `et al.` or something else as configured in the bibtex style.

The rendering of your example is
Lastname, F.; Firstname; Lastname & others
but it should be
Lastname, F.; Lastname, F. & others

src/main/java/net/sf/jabref/gui/fieldeditors/JTextAreaWithHighlighting.java
```diff
@@ -35,13 +34,15 @@
...
+    }
+
+    public JTextAreaWithHighlighting(String text) {
+        this(text, "");
     }
 
-    JTextAreaWithHighlighting(String text) {
-        super(text);
+    public JTextAreaWithHighlighting(String text, String title) {
```
koppor: Please add javadoc comments at least describing `title`. Isn't title the `textWhenNotFocused`

Please rename `textWhenNotFocused` to `placeholder` to be in line with HTML5 --> http://www.w3schools.com/tags/att_input_placeholder.asp

koppor:
I think, it's difficult to achieve the same effect as html5's placeholders:

There, when clicked in an empty field, the placeholder text remains in light gray. As soon as a key is pressed, the placeholder disappears.

Fun fact: If the focus is put away from JabRef, the text field is displayed as I wish - besides the missing cursor: 
grabbed_20161011-042036

Should also be put for the field `editor`.

Other than that: LGTM, - I assume that before merge, all other required fields also have to get placeholders, This can possibly done by others more familiar with bib(la)tex if the developers agree on this concept.

Siedlerchr:
@koppor  This can be achieved with a kind of Custom Component:
http://stackoverflow.com/questions/1738966/java-jtextfield-with-input-hint

matthiasgeiger:
@Siedlerchr thats the way it's currently implemented: Showing/Hiding the hint on FocusGained/FocusLost.

But of course, it is also possible to wait for a "KeyPressedEvent" to delete the placeholder. 
However, I think using "isEmpty()" will no longer work, but a flag indicating whether there was an edit (i.e., a key has been pressed) will be necessary.

koppor:
The solution to use [WebLookAndFeel](https://github.com/mgarin/weblaf) (http://stackoverflow.com/a/30782629/873282) also looks promising ( `setInptPrompt`), but the license of that library is GPL 😟

koppor:
Can't just the simple solution using `setText()` and `getText().equals(promptText)` be used? http://stackoverflow.com/a/20200870/873282

src/main/java/net/sf/jabref/gui/entryeditor/EntryEditorTab.java
matthiasgeiger: I think, as our "normalize to bibtex name format" formatter will produce "Last, First and Last, First and other", the text should be in this form.

boceckts:
Using `setText` will save the text to the BibTeX source and  the text will apear in the fields of the maintable. I guess this is not desired.

matthiasgeiger:
@boceckts Are you sure that a `setText(...)` will already trigger an edit of the `BibEntry`? I thought this is performed upon changing the active entry editor textfield...

koppor:
Only when the text field gets focus. 😇 

Could you try at the JavaFX branch? The solution http://stackoverflow.com/a/29946224/873282 looks promising.

    `.text-field {
        -fx-prompt-text-fill: derive(-fx-control-inner-background,-30%);
    }
    .text-field:focused {
    	-fx-prompt-text-fill: derive(-fx-control-inner-background,-30%);
    }`

and add the 'promptText' entry in fxml file:

    <TextField fx:id="XY_Id" promptText="First name">

No need to recode the whole tabbing, just a small test and if it works, we can decide to postpone a full implementation until javafx is our main UI technology.

boceckts:
I will give it a try. And yes I think it trigerres when the field loses it's focus.

src/main/java/net/sf/jabref/gui/fieldeditors/JTextAreaWithHighlighting.java
boceckts: Done

src/main/java/net/sf/jabref/gui/fieldeditors/JTextAreaWithUnfocusedText.java
boceckts: Done.

boceckts:
I changed the behaviour so that the placeholder will always be displayed when the content of the text component is empty. This also means that the "search..." string will also be displayed when nothing is typed even when the search field is focused. I also renamed the classes to `JTextAreaWithPlaceholder` and `JTextFieldWithPlaceholder`. I also included the "Editor" field to display the same placeholder.

As @koppor mentioned, JavaFX TextField element has already a palceholder attribute which worked for me as you would expected. (Same as html5)

src/main/java/net/sf/jabref/gui/entryeditor/EntryEditorTab.java
```diff
@@ -151,7 +153,12 @@ private void setupPanel(JabRefFrame frame, BasePanel bPanel, boolean addKeyField
                         false);
                 defaultHeight = 0;
             } else {
-                fieldEditor = new TextArea(field, null);
+                String prompt = "";
+                if (field.equals(FieldName.AUTHOR) || field.equals(FieldName.EDITOR)) {
+                     prompt = Localization.lang("Firstname Lastname and Firstname Lastname") + " and others";
```
oscargus: I'd suggest removing the `and` from the localization as well as it is required that it stays `and`, therefore making it easier to translate.

boceckts: I changed it accordingly.

matthiasgeiger:
This is not really working on `master`, is it?

boceckts:
@matthiasgeiger what do you mean, what exactly does not work? Is there an exception? I tested it on the current masters branch it works as expected...

matthiasgeiger:
Okay... you are right. Sorry, I tried the last build from https://builds.jabref.org/master which are currently outdated 😞

PR #2150, opened 2016-10-10
[WIP] Found entries will be shown when dropping a pdf with xmp meta data

boceckts:
When dropping a pdf with xmp meta data to the "File" field of an entry the information dialog should diplay the found entries. See https://github.com/koppor/jabref/issues/5 for reference.

xmpmetaentries

- [ ] Internal quality assurance

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - Improve language quality of the German translation of shared database
 - Change "Recent files" to "Recent databases" to keep the file menu consistent
 - Customized importer files need to be slightly changed since the class `ImportFormat` was renamed to `Importer`
+- [koppor#5](https://github.com/koppor/jabref/issues/5) When entries are found while dropping a pdf with xmp meta data to the "File" field the found entries will be displayed in the information dialog
 
 ### Fixed
 - Fixed [#2092](https://github.com/JabRef/jabref/issues/2092): "None"-button in date picker clears the date field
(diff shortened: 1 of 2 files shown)


koppor:
Could you also treat the dialog which appears when dropping on an entry in the entry table:

grabbed_20161011-040244

koppor:
Apart from my question: LGTM :+1:

boceckts:
I also added the found entries to the gerneral pdf import dialog and updated the description of this pr.

koppor:
Please push and do not include 2f3c47fde08ded07a4c6ebc as this seems to be from another PR.

grabbed_20161011-182013

boceckts:
Sry I guess I forgot to push...
The commit history looks good in my opinion. Here I can only see the two commits I made and all the changes shown in this pr are from me.

oscargus:
Looks good!

Would it make sense to show the resulting entry independent of which option is selected in the "Import metadata from PDF" dialog? (The heading of the source box needs to be changed in that case.)

boceckts:
@oscargus what do you mean by that? The entry from the XMP metadata will always be shown if an entry is found, regardless of what is selected. The "XMP-metadata" section is only hidden if no entry has been found.

oscargus:
I'm thinking that it can always be shown and it shows whatever entry is generated. For a blank it may be:

```
@misc{,
file = {...},
}
```

and so on. In this way, the user knows what the resulting entry will be.

No idea how easy it is (i.e., when the actual entry is created and therefore if one can get hold of it to show) and if it should be in this PR. Just seemed like a nice extension to the general idea of this PR.

boceckts:
I get the entries from the `XMPUtil.readXMP(..)` which can also return an empty list if it didn't find any entry. So I guess I would have to adjust this method to get a "blank" entry. Does it really make sense to create an empty entry?

oscargus:
Not in that case. I'm more thinking if you select one of the other import
cases with the radio buttons (blank with link, content). Not sure how to
handle all possible outcomes though. What happens if you select XMP and
there is no XMP data? Anything gets added?

You can leave it for now, if you want to, as I do not know how realistic it
is to implement it.

boceckts:
XMP can not be selected if there is no xmp metadata. :)

oscargus:
OK! Then that part is solved. :-)

Issue #2089, opened 2016-09-28
Autogenerate Keys Finds/Produces Incorrect Duplicate Keys


jkulesza:
JabRef 3.7dev--snapshot--2016-09-28--master--555944f
mac os x 10.11.6 x86_64 
Java 1.8.0_102

Steps to reproduce:

1. Select all entries in DB.
2. Autogenerate BibTeX keys.
3. Cleanup entries below.

67 out of 110 of my entries have modified keys which append "a" as though there are duplicates (note that there are not duplicate entries in my .bib file confirmed with Quality -> Find Duplicates).  The associated PDFs are accordingly renamed.  

If I Save/Exit/Re-launch and repeat the process, the "a"s are removed and the database is returned to the original state.  

If I repeat the process two times before saving, those entries have "b" appended.  An additional time and I have "c" and so forth.  The only way to revert to have no letter is to Save/Exit/Re-key/Cleanup.

**Bottom line**: there appears to be a flaw in how entries are stored / uniqueness checked as part of the re-key process.

My key generator and cleanup settings are below.

<img width="727" alt="screen shot 2016-09-28 at 15 12 52" src="https://cloud.githubusercontent.com/assets/9939059/18932512/0ec208c6-858e-11e6-9209-66a64d0dd008.png">

<img width="624" alt="screen shot 2016-09-28 at 15 11 52" src="https://cloud.githubusercontent.com/assets/9939059/18932516/128db7b6-858e-11e6-83bf-2b31fd20dcdf.png">

jkulesza:
Has anyone been able to reproduce?

bartsch-dev:
It suffices if you select a single entry and generate the BibTeX key multiple times; the BibTeX key will count up (..a, ..b, ...).

I'll look into it in the next few days.

koppor:
Thank you for reporting this issue. We think, that is already fixed in our development version and consequently the change will be included in the next release.

We would like to ask you to use a development build from https://builds.jabref.org/master.

jkulesza:
Thank you for the quick fix!  I downloaded the development build and can confirm that the issue is fixed.

PR #1577, opened 2016-07-12
Remove teamscale findings in the database package

tschechlovdev:
Remove the teamscale finding in the database package, that is in the model package. This means refactoring of methods, that are too long, adding comments to public method/classes and used streams in some situations for better readability (and reducing code).

PR #2151, opened 2016-10-10
Fix DuplicationChecker and key generator

bartsch-dev:
Fixes https://github.com/JabRef/jabref/issues/2089.

The Duplication checker was faulty and didn't catch all the cite keys. There needed to be used a special method in the BasePanel which wasn't used at all.
I registered the DuplicationChecker with the Eventbus which now gets notified when an Entry changes, gets added, or removed.

This fixes the automatic Key generation which before had wrong information on what cite keys exist.

### DIFF
--- a/src/main/java/net/sf/jabref/gui/BasePanel.java
+++ b/src/main/java/net/sf/jabref/gui/BasePanel.java
@@ -112,6 +112,7 @@ import net.sf.jabref.logic.util.io.FileBasedLock;
 import net.sf.jabref.logic.util.io.FileUtil;
 import net.sf.jabref.logic.util.io.RegExpFileSearch;
 import net.sf.jabref.model.FieldChange;
+import net.sf.jabref.model.bibtexkeypattern.AbstractBibtexKeyPattern;
 import net.sf.jabref.model.database.BibDatabase;
 import net.sf.jabref.model.database.BibDatabaseContext;
 import net.sf.jabref.model.database.DatabaseLocation;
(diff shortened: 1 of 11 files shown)


src/main/java/net/sf/jabref/gui/BasePanel.java
```diff
@@ -434,29 +435,22 @@ public void init() {
...
-                // First check if any entries have keys set already. If so, possibly remove
-                // them from consideration, or warn about overwriting keys.
-                // This is a partial clone of net.sf.jabref.gui.entryeditor.EntryEditor.GenerateKeyAction.actionPerformed(ActionEvent)
-                for (final Iterator<BibEntry> i = entries.iterator(); i.hasNext();) {
-                    bes = i.next();
-                    if (bes.hasCiteKey()) {
-                        if (Globals.prefs.getBoolean(JabRefPreferences.AVOID_OVERWRITING_KEY)) {
-                            // Remove the entry, because its key is already set:
+                if (Globals.prefs.getBoolean(JabRefPreferences.AVOID_OVERWRITING_KEY)) {
+                    for (final Iterator<BibEntry> i = entries.iterator(); i.hasNext();) {
```
tschechlovdev: Please rename i to iterator.

tschechlovdev: And why not using stream with filter? Or even better entries.removeIf?

src/main/java/net/sf/jabref/model/database/DuplicationChecker.java
```diff
@@ -2,62 +2,43 @@
...
-class DuplicationChecker {
-
-    private static final Log LOGGER = LogFactory.getLog(DuplicationChecker.class);
+public class DuplicationChecker {
 
-    // use a map instead of a set since i need to know how many of each key is in there
+    /** use a map instead of a set since I need to know how many of each key is in there */
     private final Map<String, Integer> allKeys = new HashMap<>();
 
+
```
tschechlovdev: please delete empty line.

src/main/java/net/sf/jabref/model/database/DuplicationChecker.java
```diff
@@ -2,62 +2,43 @@
...
-            }
-        }
-        if (duplicate) {
-            LOGGER.warn("Warning there is a duplicate key: " + newKey);
-        }
-        return duplicate;
+    /**
+     * Checks if there is more than one occurrence of this key
+     */
+    public boolean isDuplicateExisting(Optional<String> citeKey) {
```
tschechlovdev: Please don't use optionals as params. They should only be used as return values. See https://github.com/JabRef/jabref/pull/1577#r75998843

src/main/java/net/sf/jabref/model/database/DuplicationChecker.java
bartsch-dev: I always like a little visual separation between the members from the methods, and seen this in some other classes too.

src/main/java/net/sf/jabref/gui/BasePanel.java
bartsch-dev: Lambdas still doesn't come that easy to my mind.
I rewrote this and the part just beneath with lambdas.

src/main/java/net/sf/jabref/model/database/DuplicationChecker.java
tschechlovdev: I know that I got many comments for empty lines like this.
But I think in this case it's okay to leave it.

src/main/java/net/sf/jabref/gui/BasePanel.java
```diff
@@ -434,65 +434,39 @@ public void init() {
...
-                            if (answer == JOptionPane.NO_OPTION) {
-                                // Ok, break off the operation.
-                                canceled = true;
-                                return;
-                            }
-                            // No need to check more entries, because the user has already confirmed
-                            // that it's ok to overwrite keys:
-                            break;
+                if (Globals.prefs.getBoolean(JabRefPreferences.AVOID_OVERWRITING_KEY)) {
+                    entries.removeIf(BibEntry::hasCiteKey);
```
koppor: Please readd the comments from the old code if possible. Here, the comment is `// Remove the entry, because its key is already set:`

src/main/java/net/sf/jabref/model/database/DuplicationChecker.java
```diff
@@ -96,17 +71,35 @@ protected boolean addKeyToSet(String key) {
...
+    public void listen(FieldChangedEvent fieldChangedEvent) {
+        if (fieldChangedEvent.getFieldName().equals(BibEntry.KEY_FIELD)) {
+            removeKeyFromSet(fieldChangedEvent.getOldValue());
+            addKeyToSet(fieldChangedEvent.getNewValue());
         }
     }
+
+    @Subscribe
+    public void listen(EntryRemovedEvent entryRemovedEvent) {
+        removeKeyFromSet(entryRemovedEvent.getBibEntry().getCiteKeyOptional().orElse(null));
```
koppor: I am not sure whether `null` should really be treated as real key. Why not excluding null from FieldChanged, EntryRemoved, and EntryAdded?

As a consequence, null always appears 0 times in the database.

koppor:
Tested it  manually: works. Besides the minor question: LGTM.

src/main/java/net/sf/jabref/logic/bibtexkeypattern/BibtexKeyPatternUtil.java
```diff
@@ -457,7 +448,7 @@ public static void makeLabel(AbstractBibtexKeyPattern citeKeyPattern, BibDatabas
             }
 
             String moddedKey = key + getAddition(number);
-            occurrences = database.getNumberOfKeyOccurrences(moddedKey);
+            occurrences = database.getDuplicationChecker().getNumberOfKeyOccurrences(moddedKey);
 
             if (Objects.equals(oldKey, moddedKey)) {
                 occurrences--;
```
bartsch-dev: Can only happen if the `BibtexKeyPatternUtil#getAddition` method is buggy.

src/main/java/net/sf/jabref/model/database/DuplicationChecker.java
bartsch-dev: if the key is `null` `removeKeyFromSet` will return immediately.
now I use the `ifPresent` method of the optional.

src/main/java/net/sf/jabref/gui/BasePanel.java
bartsch-dev: I added some comments

PR #2152, opened 2016-10-10
Add test case for other and unknown type

tschechlovdev:
Regarding: [#29] (https://github.com/koppor/jabref/issues/29) 
I've added a test case for reading and writing a file with other and unknwon type.

### DIFF
--- a/src/test/java/net/sf/jabref/logic/bibtex/BibEntryWriterTest.java
+++ b/src/test/java/net/sf/jabref/logic/bibtex/BibEntryWriterTest.java
@@ -3,6 +3,7 @@ package net.sf.jabref.logic.bibtex;
 import java.io.IOException;
 import java.io.StringReader;
 import java.io.StringWriter;
+import java.net.URISyntaxException;
 import java.util.Collection;
 import java.util.Set;
 
@@ -59,6 +60,50 @@ public class BibEntryWriterTest {
         // @formatter:on
 
         assertEquals(expected, actual);
+    }
+
+    @Test
+    public void roundTripTestOtherType() throws IOException, URISyntaxException {
+        String bibtexEntry = "@Other{test," + OS.NEWLINE +
+                "  Comment                  = {testentry}" + OS.NEWLINE +
+                "}";
+
+        // read in bibtex string
+        ParserResult result = BibtexParser.parse(new StringReader(bibtexEntry), importFormatPreferences);
+        Collection<BibEntry> entries = result.getDatabase().getEntries();
+        BibEntry entry = entries.iterator().next();
+
+        String resourceName = "/testbib/othertype.bib";
+        BibEntryAssert.assertEquals(BibEntryWriter.class, resourceName, entry);
+
+        //write out bibtex string
+        StringWriter stringWriter = new StringWriter();
+        writer.write(entry, stringWriter, BibDatabaseMode.BIBTEX);
+        String actual = stringWriter.toString();
+
+        assertEquals(bibtexEntry, actual);
+    }
+
+    @Test
+    public void roundTripTestReallyunknownType() throws IOException, URISyntaxException {
+        String bibtexEntry = "@ReallyUnknownType{test," + OS.NEWLINE +
+                "  Comment                  = {testentry}" + OS.NEWLINE +
+                "}";
+
+        // read in bibtex string
+        ParserResult result = BibtexParser.parse(new StringReader(bibtexEntry), importFormatPreferences);
+        Collection<BibEntry> entries = result.getDatabase().getEntries();
+        BibEntry entry = entries.iterator().next();
+
+        String resourceName = "/testbib/reallyunknowntype.bib";
+        BibEntryAssert.assertEquals(BibEntryWriter.class, resourceName, entry);
+
+        //write out bibtex string
+        StringWriter stringWriter = new StringWriter();
+        writer.write(entry, stringWriter, BibDatabaseMode.BIBTEX);
+        String actual = stringWriter.toString();
+
+        assertEquals(bibtexEntry, actual);
     }
 
     @Test
--- a/src/test/java/net/sf/jabref/logic/bibtex/BibEntryWriterTest.java
+++ b/src/test/java/net/sf/jabref/logic/bibtex/BibEntryWriterTest.java
@@ -3,7 +3,6 @@ package net.sf.jabref.logic.bibtex;
 import java.io.IOException;
 import java.io.StringReader;
 import java.io.StringWriter;
-import java.net.URISyntaxException;
 import java.util.Collection;
 import java.util.Set;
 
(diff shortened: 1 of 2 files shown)


src/test/java/net/sf/jabref/logic/bibtex/BibEntryWriterTest.java
```diff
@@ -62,6 +63,50 @@ public void testSerialization() throws IOException {
     }
 
     @Test
+    public void roundTripTestOtherType() throws IOException, URISyntaxException {
+        String bibtexEntry = "@Other{test," + OS.NEWLINE +
+                "  Comment                  = {testentry}" + OS.NEWLINE +
+                "}";
+
+        // read in bibtex string
+        ParserResult result = BibtexParser.parse(new StringReader(bibtexEntry), importFormatPreferences);
```
oscargus: Please use `new BibtexParser(importFormatPreferences).parse(new StringReader(bibtexEntry))` (or `Collection<BibEntry> entries = new BibtexParser(importFormatPreferences).parseEntries(bibtexEntry);`.

src/test/java/net/sf/jabref/logic/bibtex/BibEntryWriterTest.java
```diff
@@ -62,6 +63,50 @@ public void testSerialization() throws IOException {
...
+
+    @Test
+    public void roundTripTestReallyunknownType() throws IOException, URISyntaxException {
+        String bibtexEntry = "@ReallyUnknownType{test," + OS.NEWLINE +
+                "  Comment                  = {testentry}" + OS.NEWLINE +
+                "}";
+
+        // read in bibtex string
+        ParserResult result = BibtexParser.parse(new StringReader(bibtexEntry), importFormatPreferences);
+        Collection<BibEntry> entries = result.getDatabase().getEntries();
```
oscargus: Same here.

src/test/java/net/sf/jabref/logic/bibtex/BibEntryWriterTest.java
```diff
@@ -62,6 +63,50 @@ public void testSerialization() throws IOException {
...
+        String bibtexEntry = "@Other{test," + OS.NEWLINE +
+                "  Comment                  = {testentry}" + OS.NEWLINE +
+                "}";
+
+        // read in bibtex string
+        ParserResult result = BibtexParser.parse(new StringReader(bibtexEntry), importFormatPreferences);
+        Collection<BibEntry> entries = result.getDatabase().getEntries();
+        BibEntry entry = entries.iterator().next();
+
+        String resourceName = "/testbib/othertype.bib";
```
tobiasdiez: Please remove the assert statement here. Only one assert / failure reason per test. If you think that is valuable to test that the other type is written correctly, then extract this test to a new test method (I would actually prefer two tests for writing and reading separately instead of a roundtrip test)

src/test/java/net/sf/jabref/logic/bibtex/BibEntryWriterTest.java
```diff
@@ -62,6 +63,50 @@ public void testSerialization() throws IOException {
...
+        String bibtexEntry = "@ReallyUnknownType{test," + OS.NEWLINE +
+                "  Comment                  = {testentry}" + OS.NEWLINE +
+                "}";
+
+        // read in bibtex string
+        ParserResult result = BibtexParser.parse(new StringReader(bibtexEntry), importFormatPreferences);
+        Collection<BibEntry> entries = result.getDatabase().getEntries();
+        BibEntry entry = entries.iterator().next();
+
+        String resourceName = "/testbib/reallyunknowntype.bib";
```
tobiasdiez: same remark as above.

src/test/java/net/sf/jabref/logic/bibtex/BibEntryWriterTest.java
tschechlovdev: Done.

src/test/java/net/sf/jabref/logic/bibtex/BibEntryWriterTest.java
tschechlovdev: Done. I've moved the assert to a new test method.

src/test/java/net/sf/jabref/logic/bibtex/BibEntryWriterTest.java
```diff
@@ -62,6 +62,71 @@ public void testSerialization() throws IOException {
...
+        StringWriter stringWriter = new StringWriter();
+        writer.write(entry, stringWriter, BibDatabaseMode.BIBTEX);
+        String actual = stringWriter.toString();
+
+        assertEquals(bibtexEntry, actual);
+    }
+
+    @Test
+    public void readReallyUnknownTypeTest() throws Exception {
+        String bibtexEntry = "@ReallyUnknownType{test," + OS.NEWLINE +
```
tobiasdiez: If you just write it, then there is no need to read it from a string...thus please create the BibEntry by hand in these cases.

tobiasdiez: If you have a test for reading something, then you should place it in the ParserTest ;-)

tschechlovdev: Ok, removed it from here and placed it to the ParserTest 😄

src/test/java/net/sf/jabref/logic/importer/fileformat/BibtexParserTest.java
```diff
@@ -236,6 +237,32 @@ public void parseRecognizesEntryWithUnknownType() throws IOException {
...
     @Test
+    public void parseReallyUnknownType() throws Exception {
+        String bibtexEntry = "@ReallyUnknownType{test," + OS.NEWLINE +
+                " Comment                  = {testentry}" + OS.NEWLINE +
+                "}";
+
+        Collection<BibEntry> entries = new BibtexParser(importFormatPreferences).parseEntries(bibtexEntry);
+        BibEntry entry = entries.iterator().next();
+
+        String resourceName = "/testbib/reallyunknowntype.bib";
```
Siedlerchr: Why is the resourceName  necessary? The other tests don't use it

tschechlovdev: I've wanted to test explicit against the testfile "reallyunknowntype.bib". If I see right, the other tests don't use such test files.

Siedlerchr: I think this is not necessary in this case. If you really want to test it, then please use the Resources directory handling, like in the` BibTexImportetTest.class ` and do not hardcode the relative dir. 
` Paths.get(BibtexImporterTest.class.getResource("BibtexImporter.examples.bib").toURI());`

tobiasdiez: Yes better compare it directly against a hand-created bibEntry (copy from above). 
Then you can use `assertEquals(singletonlist of expected entry, entries)`.

Apart from this last remark (sorry for being pedantic) this looks good to me and can be merged.

tschechlovdev: Ok, I've created the bibentry now by hand.

tobiasdiez:
Thanks for the follow-up!

Issue #97, opened 2015-08-12
Remove old compatibility updates


stefan-kolb:
Do we still need this method to adapt JabRef preferences?
https://github.com/JabRef/jabref/blob/f69e93368bcbe076a9874b4c9b31899623b061df/src/main/java/net/sf/jabref/util/Util.java#L575
Seems like this is only needed for a very old JR version <= 1.55

simonharrer:
I think JR version <= 1.55 is too old. We can remove this upgrade method.

joberreiter:
Hm. You intentionally waive compatibility, just to achieve a code saving of a handful lines?

stefan-kolb:
The current master branch is intended to be a major rewrite. 
Version 1.55 dates back to 2004/11 (if this code is only for 1.55).
Quite a few things will probably break in the transition to JabRef v3.
We try to get rid of a lot of 'minor' legacy hacks to make the code base maintainable again to be able move faster again :sweat_smile:

joberreiter:
Ok, this is understandable.

stefan-kolb:
@simonharrer You found another one?!

simonharrer:
https://github.com/JabRef/jabref/blob/master/src/main/java/net/sf/jabref/groups/migrations/VersionHandling.java

It handles group versions 0, 1, 2 and 3. 3 is the current version of how groups are stored.

stefan-kolb:
Jörg moved all this stuff into `PreferencesMigrations.java`, which is a good place to tidy things up.
We can discuss a removal later on. 
Closing for now.

PR #2153, opened 2016-10-10
Remember last exported preferences

tschechlovdev:
fixes/implements [#97] (https://github.com/koppor/jabref/issues/97)
Now if the preferences are exported and later import preferences is choosen, then the explorer will start where the preferences are last exported.


oscargus:
Looks good! Would it make sense to store the location to the preferences so it is consistent between sessions?

tobiasdiez:
Yes, I think @oscargus is right and it make sense to store this in the preferences (we already have something similar in place for import/export of bib files. I think it is called `working directory`). Please also add a changelog entry.

tschechlovdev:
I've stored it now in the preferences and added a changelog entry.

src/main/java/net/sf/jabref/gui/preftabs/PreferencesDialog.java
```diff
@@ -63,6 +63,7 @@
     private final JButton showPreferences = new JButton(Localization.lang("Show preferences"));
 
     private final JButton resetPreferences = new JButton(Localization.lang("Reset preferences"));
+    private Path lastExportedPath;
```
tobiasdiez: not needed?

tschechlovdev: Removed. Travis is failing, but that doesn't seem to be related to this PR.

PR #2158, opened 2016-10-12
Hotfix: NPE in the main table

bartsch-dev:
#### Steps to Reproduce
1. Create new database  (or use an existing one, the important part is that the editor was never opened in this database before)
- create a new entry


Siedlerchr:
Can't reproduce that here (with an older dev version) But the code looks okay.

tobiasdiez:
LGTM

PR #2162, opened 2016-10-13
Specialfields refactoring

lenhard:
Rewrite of #1964 

This kept bugging me, so I invested some time in it. It integrates the changes from #1964 and turns the different special fields into enum values. The different classes composing special fields are now aligned to our architecture. The enum `SpecialField` has to go to `logic` as it involves quite some localization.

There are still some possible improvements based on this PR, in particular:
- Outsourcing localization from `SpecialField` so that this class can be moved into `model` (this is not necessarily needed, though)
- Rewriting `SpecialFieldsUtil`, which is already marked as deprecated.

To maintain a reasonable scope for this PR, these improvements should be done in separate new PRs. They should now also be much simpler based on the new structure.


src/main/java/net/sf/jabref/gui/specialfields/SpecialFieldIcon.java
```diff
@@ -0,0 +1,75 @@
...
+        } else if("setRank3".equals(value.getActionName())) {
+            return IconTheme.JabRefIcon.RANK3.getSmallIcon();
+        } else if("setRank4".equals(value.getActionName())) {
+            return IconTheme.JabRefIcon.RANK4.getSmallIcon();
+        } else if("setRank5".equals(value.getActionName())) {
+            return IconTheme.JabRefIcon.RANK5.getSmallIcon();
+        } else if("clearReadStatus".equals(value.getActionName())) {
+            return null;
+        } else if("setReadStatusToRead".equals(value.getActionName())) {
+            return IconTheme.JabRefIcon.READ_STATUS_SKIMMED.getSmallIcon();
```
matthiasgeiger: READ_STATUS_READ

matthiasgeiger:
Iconbar icons for multi value specialfields are too small. (sorry no time to find the cause in the code)

src/main/java/net/sf/jabref/gui/specialfields/SpecialFieldIcon.java
lenhard: Thanks! The wrong Iconbar icons (see below) is certainly also just some misassigned. I'll have a look.

Siedlerchr: Why not switch-case here?

src/main/java/net/sf/jabref/gui/specialfields/SpecialFieldIcon.java
```diff
@@ -0,0 +1,75 @@
...
+
+import net.sf.jabref.gui.IconTheme;
+import net.sf.jabref.logic.specialfields.SpecialField;
+import net.sf.jabref.model.entry.SpecialFieldValue;
+
+
+public class SpecialFieldIcon {
+
+    public static Icon getRepresentingIcon(SpecialField field) {
+        if (SpecialField.PRINTED.equals(field)) {
```
Siedlerchr: I think a switch-case would be more readable and since java 7 it is possible to directly a switch on Strings:
http://docs.oracle.com/javase/7/docs/technotes/guides/language/strings-switch.html
It uses the equals method

Siedlerchr: and:  The Java compiler generates generally more efficient bytecode from switch statements that use String objects than from chained if-then-else statements.

lenhard: Heh, I was expecting that. Why not turn the whole thing into a `HashMap` instead of an if/else or a switch? I have no clue if that is preferable over the other alternatives, though.

Side remark: This would not be switching on a String, but on an enum value.

Siedlerchr: Hm. When it is already an enum, why not setting the Icon(Name) directly as Enum parameter?
Then we neither need a switch nor a hashmap

lenhard: @Siedlerchr Could you explain some more? I don't really understand what you mean with setting the icon directly as enum parameter?

Siedlerchr: @lenhard  Yeah. Just add a fieldfrom the type Icon in the Special Field enum
then you could directly assign the Icon and we don't need a Hashmap or the equals stuff.

e.g `MY_SPEC_FIELD("asdf"...,....,IconTheme.JabRefIcon.READ_STATUS.getSmallIcon())`

lenhard:
After a night of almost enough sleep, I have some more ideas and will push up the localization upwards to get the essence into model/logic. I'll readd the ready-for-review when I am done.

src/main/java/net/sf/jabref/gui/specialfields/SpecialFieldIcon.java
matthiasgeiger: No. That's not possible/desired as the enum `SpecialField` is part of `logic` and should not contain UI elements (such as `JabRefIcon`)

@lenhard Icons for `Priority` and `Read Status` in the Icon bar are still too small :wink:

lenhard:
So, now there is a much better structure. `SpecialFieldValue` is also an enum now, the absolute most important information has moved into model, switches are done based on enums, the localization is in logic, and the remaining stuff in gui.

The next thing for a new PR (this time for real) is a rewriting of `SpecialFieldUtil` and moving that logic into keyword synchronization. 

When all checkstyle issues are taken care of, the PR would be ready to go from my side.

src/main/java/net/sf/jabref/gui/specialfields/SpecialFieldIcon.java
stefan-kolb: No switch case! 😄

src/main/java/net/sf/jabref/gui/specialfields/SpecialFieldIcon.java
```diff
@@ -0,0 +1,84 @@
...
+            case QUALITY:
+                return IconTheme.JabRefIcon.QUALITY.getSmallIcon();
+            case RANK:
+                return IconTheme.JabRefIcon.RANKING.getIcon();
+            case READ_STATUS:
+                return IconTheme.JabRefIcon.READ_STATUS.getSmallIcon();
+            case RELEVANCE:
+                return IconTheme.JabRefIcon.RELEVANCE.getSmallIcon();
+            default:
+                return null;
```
Siedlerchr: Is this a good idea to return here null?

src/main/java/net/sf/jabref/gui/specialfields/SpecialFieldIcon.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+
+    public static JLabel createSpecialFieldValueLabel(SpecialFieldValue value) {
+        JLabel label = new JLabel(getSpecialFieldValueIcon(value));
+        label.setToolTipText(SpecialFieldLocalization.getToolTipText(value));
+        return label;
+    }
+
+}
```
Siedlerchr: No newline at end of file

src/main/java/net/sf/jabref/logic/specialfields/SpecialFieldsUtils.java
```diff
@@ -0,0 +1,139 @@
...
+import net.sf.jabref.logic.util.UpdateField;
+import net.sf.jabref.model.FieldChange;
+import net.sf.jabref.model.entry.BibEntry;
+import net.sf.jabref.model.entry.FieldName;
+import net.sf.jabref.model.entry.Keyword;
+import net.sf.jabref.model.entry.KeywordList;
+import net.sf.jabref.model.entry.specialfields.SpecialField;
+import net.sf.jabref.model.entry.specialfields.SpecialFields;
+
+@Deprecated // the class should be refactored and partly integrated into BibEntry
```
Siedlerchr: Please add a deprecated javadoc comment, too.
http://stackoverflow.com/questions/5039723/should-i-use-javadoc-deprecation-or-the-annotation-in-java

src/main/java/net/sf/jabref/gui/specialfields/SpecialFieldIcon.java
lenhard: Well, the code is actually unreachable, since all enum values are cases. The compiler is just not smart enough to detect it, although it would be feasible. The default case would only be reached if someone added a new SpecialField and forgot to add an Icon mapping. So I would say it does not really matter.

src/main/java/net/sf/jabref/gui/specialfields/SpecialFieldIcon.java
lenhard: Will add!

src/main/java/net/sf/jabref/logic/specialfields/SpecialFieldsUtils.java
lenhard: Will do!

src/main/java/net/sf/jabref/logic/specialfields/SpecialFieldsUtils.java
```diff
@@ -0,0 +1,139 @@
...
+        fieldChanges.addAll(SpecialFieldsUtils.exportFieldToKeywords(SpecialField.RANK, be, isKeywordSyncEnabled, keywordDelimiter));
+        fieldChanges.addAll(SpecialFieldsUtils.exportFieldToKeywords(SpecialField.RELEVANCE, be, isKeywordSyncEnabled, keywordDelimiter));
+        fieldChanges.addAll(SpecialFieldsUtils.exportFieldToKeywords(SpecialField.QUALITY, be, isKeywordSyncEnabled, keywordDelimiter));
+        fieldChanges.addAll(SpecialFieldsUtils.exportFieldToKeywords(SpecialField.READ_STATUS, be, isKeywordSyncEnabled, keywordDelimiter));
+        fieldChanges.addAll(SpecialFieldsUtils.exportFieldToKeywords(SpecialField.PRINTED, be, isKeywordSyncEnabled, keywordDelimiter));
+
+        return fieldChanges;
+    }
+
+    private static List<FieldChange> importKeywordsForField(KeywordList keywordList, SpecialField c, BibEntry be) {
```
tobiasdiez: no abbreviated names please (c, val and further down s) 😈

src/main/java/net/sf/jabref/logic/specialfields/SpecialFieldsUtils.java
```diff
@@ -0,0 +1,139 @@
...
+     * updates field values according to keywords
+     */
+    public static List<FieldChange> syncSpecialFieldsFromKeywords(BibEntry be, Character keywordDelimiter) {
+        List<FieldChange> fieldChanges = new ArrayList<>();
+        if (!be.hasField(FieldName.KEYWORDS)) {
+            return fieldChanges;
+        }
+        KeywordList keywordList = be.getKeywords(keywordDelimiter);
+
+        fieldChanges.addAll(SpecialFieldsUtils.importKeywordsForField(keywordList, SpecialField.PRIORITY, be));
```
tobiasdiez: foreach?

src/main/java/net/sf/jabref/logic/specialfields/SpecialFieldsUtils.java
```diff
@@ -0,0 +1,139 @@
...
+        fieldChanges.addAll(SpecialFieldsUtils.importKeywordsForField(keywordList, SpecialField.PRINTED, be));
+
+        return fieldChanges;
+    }
+
+    /**
+     * @param fieldName the fieldName
+     * @return an instance of that field. The returned object is a singleton. null is returned if fieldName does not indicate a special field
+     */
+    public static Optional<SpecialField> getSpecialFieldInstanceFromFieldName(String fieldName) {
```
tobiasdiez: move to `SpecialField` class (since it is similar to "parse")

src/main/java/net/sf/jabref/model/entry/specialfields/SpecialField.java
```diff
@@ -0,0 +1,91 @@
...
+    private KeywordList keywords;
+    private HashMap<String, SpecialFieldValue> map;
+    private String fieldName;
+    private String localizationKey;
+    private boolean isSingleFieldValue;
+
+    SpecialField(String fieldName, String localizationKey, boolean isSingleFieldValue, SpecialFieldValue... values) {
+        this.fieldName = fieldName;
+        this.localizationKey = localizationKey;
+        this.isSingleFieldValue = isSingleFieldValue;
```
tobiasdiez: isSingleFieldValue is automatically determined by values.size() == 1, right?

src/main/java/net/sf/jabref/model/entry/specialfields/SpecialFieldValue.java
```diff
@@ -0,0 +1,56 @@
...
+    CLEAR_READ_STATUS(null, "clearReadStatus"),
+    READ("read", "setReadStatusToRead"),
+    SKIMMED("skimmed", "setReadStatusToSkimmed"),
+    RELEVANT("relevant", "toggleRelevance");
+
+    // keyword used at keyword field
+    private final Optional<Keyword> keyword;
+
+    // action belonging to this value
+    private final String actionName;
```
tobiasdiez: actionName sounds very much like gui.... I would suggest the same as for the `localizationKey`: converter `FieldValue -> ActionName` in the gui.

src/test/java/net/sf/jabref/logic/specialfields/SpecialFieldsUtilsTest.java
```diff
@@ -1,120 +1,95 @@
...
     public void testSyncKeywordsFromSpecialFieldsBibEntry() {
         BibEntry entry = new BibEntry();
         entry.setField("ranking", "rank2");
-        SpecialFieldsUtils.syncKeywordsFromSpecialFields(entry);
+        SpecialFieldsUtils.syncKeywordsFromSpecialFields(entry, true, ',');
         assertEquals(Optional.of("rank2"), entry.getField("keywords"));
     }
 
     @Test
     public void testSyncKeywordsFromSpecialFieldsBibEntryNamedCompoundHasEdits() {
```
tobiasdiez: Change test name.

src/test/java/net/sf/jabref/logic/specialfields/SpecialFieldsUtilsTest.java
```diff
@@ -1,120 +1,95 @@
...
         BibEntry entry = new BibEntry();
-        NamedCompound nc = new NamedCompound("Test");
         entry.setField("ranking", "rank2");
-        SpecialFieldsUtils.syncKeywordsFromSpecialFields(entry, nc);
+        SpecialFieldsUtils.syncKeywordsFromSpecialFields(entry, true, ',');
         assertEquals(Optional.of("rank2"), entry.getField("keywords"));
     }
 
     @Test
     public void testSyncKeywordsFromSpecialFieldsBibEntryNamedCompoundNoEdits() {
```
tobiasdiez: change test name (no named compound anymore)

src/test/java/net/sf/jabref/logic/specialfields/SpecialFieldsUtilsTest.java
```diff
@@ -1,120 +1,95 @@
...
     public void testSyncSpecialFieldsFromKeywordsBibEntry() {
         BibEntry entry = new BibEntry();
         entry.setField("keywords", "rank2");
-        SpecialFieldsUtils.syncSpecialFieldsFromKeywords(entry);
+        SpecialFieldsUtils.syncSpecialFieldsFromKeywords(entry, ',');
         assertEquals(Optional.of("rank2"), entry.getField("ranking"));
     }
 
     @Test
     public void testSyncSpecialFieldsFromKeywordsBibEntryNamedCompoundHasEdits() {
```
tobiasdiez: dito

src/test/java/net/sf/jabref/logic/specialfields/SpecialFieldsUtilsTest.java
```diff
@@ -1,120 +1,95 @@
...
-        NamedCompound nc = new NamedCompound("Test");
         entry.setField("keywords", "rank2");
-        SpecialFieldsUtils.syncSpecialFieldsFromKeywords(entry, nc);
-        assertTrue(nc.hasEdits());
+        List<FieldChange> changes = SpecialFieldsUtils.syncSpecialFieldsFromKeywords(entry, ',');
+        assertTrue(changes.size() > 0);
     }
 
     @Test
     public void testSyncSpecialFieldsFromKeywordsBibEntryNamedCompoundCorrectContent() {
```
tobiasdiez: well, I think you got the pattern: search and replace 😄

src/main/java/net/sf/jabref/logic/specialfields/SpecialFieldsUtils.java
lenhard: Done

src/main/java/net/sf/jabref/gui/actions/ManageKeywordsAction.java
```diff
@@ -332,44 +327,44 @@ private void synchronizeSpecialFields(KeywordList keywordsToAdd, KeywordList key
 
         // Priority
         clone = keywordsToAdd.createClone();
```
lenhard: I agree, done.

src/main/java/net/sf/jabref/gui/actions/ManageKeywordsAction.java
```diff
@@ -332,44 +327,44 @@ private void synchronizeSpecialFields(KeywordList keywordsToAdd, KeywordList key
...
-        clone.retainAll(Priority.getInstance().getKeyWords());
+        clone.retainAll(SpecialField.PRIORITY.getKeyWords());
```
lenhard: done

src/main/java/net/sf/jabref/logic/specialfields/SpecialFieldLocalization.java
```diff
@@ -0,0 +1,91 @@
+package net.sf.jabref.logic.specialfields;
+
+import java.util.Objects;
+
+import net.sf.jabref.logic.l10n.Localization;
+import net.sf.jabref.model.entry.specialfields.SpecialFieldValue;
+
+public class SpecialFieldLocalization {
+
+    // this is just to satisfy our own localization tests, since they will not detect if a variable is passed into Localization.lang()
```
lenhard: I admit that this was itching me. It is now moved out of SpecialField into the gui.

src/main/java/net/sf/jabref/logic/specialfields/SpecialFieldLocalization.java
```diff
@@ -0,0 +1,91 @@
...
+        Localization.lang("Quality");
+        Localization.lang("Rank");
+        Localization.lang("Read status");
+        Localization.lang("Relevance");
+    }
+
+    public static String getMenuString(SpecialFieldValue value) {
+        Objects.requireNonNull(value.getActionName());
+
+        switch(value.getActionName()){
```
lenhard: Not really. Fixed.

src/main/java/net/sf/jabref/logic/specialfields/SpecialFieldLocalization.java
```diff
@@ -0,0 +1,91 @@
...
+                return Localization.lang("Toggle relevance");
+            default:
+                return "";
+        }
+    }
+
+    public static String getToolTipText(SpecialFieldValue value) {
+        Objects.requireNonNull(value.getActionName());
+
+        switch(value.getActionName()){
```
lenhard: Same as above, fixed.

src/main/java/net/sf/jabref/gui/menus/RightClickMenu.java
```diff
@@ -112,32 +108,32 @@ public RightClickMenu(JabRefFrame frame, BasePanel panel) {
...
-                RightClickMenu.populateSpecialFieldMenu(rankingMenu, Rank.getInstance(), frame);
+                RightClickMenu.populateSpecialFieldMenu(rankingMenu, SpecialField.RANK, frame);
                 add(rankingMenu);
             }
 
             // TODO: multiple handling for relevance and quality-assurance
             // if multiple values are selected ("if (multiple)"), two options (set / clear) should be offered
             // if one value is selected either set or clear should be offered
             if (Globals.prefs.getBoolean(JabRefPreferences.SHOWCOLUMN_RELEVANCE)) {
-                add(Relevance.getInstance().getValues().get(0).getMenuAction(frame));
```
lenhard: There is just one value used here, so, as far as I see, no potential for a forEach.

src/main/java/net/sf/jabref/logic/specialfields/SpecialFieldsUtils.java
```diff
@@ -0,0 +1,139 @@
...
+        return fieldChanges;
+    }
+
+    /**
+     * Update keywords according to values of special fields
+     */
+    public static List<FieldChange> syncKeywordsFromSpecialFields(BibEntry be, boolean isKeywordSyncEnabled, Character keywordDelimiter) {
+        List<FieldChange> fieldChanges = new ArrayList<>();
+
+        fieldChanges.addAll(SpecialFieldsUtils.exportFieldToKeywords(SpecialField.PRIORITY, be, isKeywordSyncEnabled, keywordDelimiter));
```
lenhard: Done

src/main/java/net/sf/jabref/logic/specialfields/SpecialFieldsUtils.java
lenhard: Improved

src/main/java/net/sf/jabref/logic/specialfields/SpecialFieldsUtils.java
lenhard: Done

src/test/java/net/sf/jabref/logic/specialfields/SpecialFieldsUtilsTest.java
lenhard: I am not sure if I was using the right schema for you, but at least I tried to improve all of the test names.

src/main/java/net/sf/jabref/logic/specialfields/SpecialFieldsUtils.java
lenhard: Done

src/main/java/net/sf/jabref/model/entry/specialfields/SpecialField.java
lenhard: You are of course correct, done.

src/main/java/net/sf/jabref/gui/maintable/SpecialMainTableColumns.java
```diff
@@ -43,45 +39,45 @@ public String getDisplayName() {
         public Object getColumnValue(BibEntry entry) {
 
             return entry.getField(SpecialFields.FIELDNAME_RANKING)
-                    .flatMap(Rank.getInstance()::parse).map(rank -> rank.createLabel()).orElse(null);
+                    .flatMap(SpecialField.RANK::parse).map(rank -> SpecialFieldIcon.createSpecialFieldValueLabel(rank)).orElse(null);
         }
     };
 
     public static final MainTableColumn PRIORITY_COLUMN = new MainTableColumn(SpecialFields.FIELDNAME_PRIORITY,
```
lenhard: Well, the columns differ somewhat (multi-value, versus single-value), so not in an entirely beautiful way. But I turned the whole thing into a package-private non-static builder class, so it is much less ugly.

src/main/java/net/sf/jabref/gui/menus/RightClickMenu.java
tobiasdiez: I think I meant foreach over the special fields (ranking, relevance, ...).

lenhard:
Now that the army of fridges and web cams has stopped trying to bring down the internet and our build system is running again, I can finally certify that this works.

So, thanks to the excellent review by @tobiasdiez this has been a huge improvement of the special fields feature. `SpecialField` and `SpecialFieldValue` are now in model and are limited to the absolute essence. All gui-related aspects are moved into `ViewModel` classes in the gui.

The only thing that remains now is the refactoring of the deprecated `SpecialFieldUtils` class and the moving of its methods into `BibEntry`. This, however, cannot easily be achieved, since it depends on `UpdateField`, which is in logic and depends on preferences. If this class stays in logic, then so do the special field update methods. Alternatively, it would be possible to move the methods from `SpecialFieldUtils to `UpdateField`, but I do not really see the point in this. We might consider restructuring the `UpdateField` field class, but in the end this is not the most pressing issue we have.

The refactoring/rewriting of `SpecialFieldUtils` requires more discussion in the group and can be done separate from the changes in this PR. Hence, I consider this PR as finished and it is once again ready for review.

lenhard:
@JabRef/developers This PR is reviewable / mergeable. To avoid having to fix merge conflicts that build up over time, I would like to get this into master. 

I will merge this in, if no one objects within the next seven days.

stefan-kolb:
I trust your skills, go for it!

src/main/java/net/sf/jabref/gui/JabRefFrame.java
```diff
@@ -288,19 +285,19 @@
...
-            Quality.getInstance().getValues().get(0).getToolTipText(),
+            new SpecialFieldValueViewModel(SpecialField.QUALITY.getValues().get(0)).getActionName(),
+            new SpecialFieldValueViewModel(SpecialField.QUALITY.getValues().get(0)).getMenuString(),
+            new SpecialFieldValueViewModel(SpecialField.QUALITY.getValues().get(0)).getToolTipText(),
             IconTheme.JabRefIcon.QUALITY_ASSURED.getIcon());
     private final AbstractAction togglePrinted = new GeneralAction(
-            Printed.getInstance().getValues().get(0).getActionName(),
-            Printed.getInstance().getValues().get(0).getMenuString(),
-            Printed.getInstance().getValues().get(0).getToolTipText(),
+            new SpecialFieldValueViewModel(SpecialField.PRINTED.getValues().get(0)).getActionName(),
```
koppor: Why three times the instantiation? Think, it won't be a memory issue, will it?

src/main/java/net/sf/jabref/logic/specialfields/SpecialFieldsUtils.java
```diff
@@ -0,0 +1,125 @@
...
+        fieldChanges.addAll(SpecialFieldsUtils.exportFieldToKeywords(field, entry, isKeywordSyncEnabled, keywordDelimiter));
+
+        return fieldChanges;
+    }
+
+    private static List<FieldChange> exportFieldToKeywords(SpecialField specialField, BibEntry entry, boolean isKeywordSyncEnabled, Character keywordDelimiter) {
+        List<FieldChange> fieldChanges = new ArrayList<>();
+
+        if (!isKeywordSyncEnabled) {
+            return fieldChanges;
```
koppor: Could that be tested, too? - It is set by a preference. Default is to sync.

src/main/java/net/sf/jabref/logic/specialfields/SpecialFieldsUtils.java
```diff
@@ -0,0 +1,125 @@
...
+
+    public static void synchronizeSpecialFields(KeywordList keywordsToAdd, KeywordList keywordsToRemove) {
+        // we need to check whether a special field is added
+        // for each field:
+        //   check if something is added
+        //   if yes, add all keywords of that special fields to the keywords to be removed
+
+        KeywordList clone;
+
+        // Priority
```
koppor: This comment can be removed

src/main/java/net/sf/jabref/logic/specialfields/SpecialFieldsUtils.java
```diff
@@ -0,0 +1,125 @@
...
+
+        for(SpecialField field: SpecialField.values()) {
+            fieldChanges.addAll(SpecialFieldsUtils.importKeywordsForField(keywordList, field, entry));
+        }
+
+        return fieldChanges;
+    }
+
+    public static void synchronizeSpecialFields(KeywordList keywordsToAdd, KeywordList keywordsToRemove) {
+        // we need to check whether a special field is added
```
koppor: This comment should be moved somewhere else, shouldn't it?

src/main/java/net/sf/jabref/logic/specialfields/SpecialFieldsUtils.java
```diff
@@ -0,0 +1,125 @@
...
+        // for each field:
+        //   check if something is added
+        //   if yes, add all keywords of that special fields to the keywords to be removed
+
+        KeywordList clone;
+
+        // Priority
+        clone = keywordsToAdd.createClone();
+        for(SpecialField field: SpecialField.values()){
```
koppor: No test coverage. Is this intended?

src/main/java/net/sf/jabref/model/entry/specialfields/SpecialField.java
```diff
@@ -0,0 +1,111 @@
...
+    ),
+
+    READ_STATUS("readstatus",
+        SpecialFieldValue.CLEAR_READ_STATUS,
+        SpecialFieldValue.READ,
+        SpecialFieldValue.SKIMMED
+    ),
+
+    RELEVANCE("relevance",
+        SpecialFieldValue.RELEVANT
```
koppor: Why no CLEAR_RELEVANT? Maybe a comment should be added to the class itself how toggle values are distinguished from multi-values. Is it only because of the number of values given to the enum?

src/main/java/net/sf/jabref/model/entry/specialfields/SpecialField.java
```diff
@@ -0,0 +1,111 @@
...
+
+    public String getFieldName() {
+        return fieldName;
+    }
+
+    public boolean isSingleValueField() {
+        return this.values.size() == 1;
+    }
+
+    public static Optional<SpecialField> getSpecialFieldInstanceFromFieldName(String fieldName) {
```
koppor: Can we achieve 100% coerage here.

src/main/java/net/sf/jabref/model/entry/specialfields/SpecialField.java
```diff
@@ -0,0 +1,111 @@
...
+    public Optional<SpecialFieldValue> parse(String value) {
+        return Optional.ofNullable(map.get(value));
+    }
+
+    public String getFieldName() {
+        return fieldName;
+    }
+
+    public boolean isSingleValueField() {
+        return this.values.size() == 1;
```
koppor: Why not code covered?

koppor:
I tested it locally and it works with keyword sync and without keyword sync. I am OK for merge and would address my minor comments (mostly regarding test coverage) later.

koppor:
It would be nice if there was a howto to add a new special field.

PR #2165, opened 2016-10-15
Support \r when using CitationStyles

bartsch-dev:
Fixes: https://github.com/koppor/jabref/issues/174

The CitationStyle library cannot handle `\r` thus I parse to a `space`.
I also added another return (and localization) to always have a feedback for the user.


src/main/java/net/sf/jabref/logic/citationstyle/CitationStyleGenerator.java
```diff
@@ -56,17 +56,18 @@ protected static String generateCitation(BibEntry entry, String style, CitationS
 
         } catch (IOException | ArrayIndexOutOfBoundsException e) {
             LOGGER.error("Could not generate BibEntry Citation", e);
+            return Localization.lang("Cannot generate Citation");
```
koppor: I think, this generates the string displayed in the preview panel. Maybe, the message should be "Cannot generate preview based on selected citation style".

Further, "Citation" is not a fixed term, which should be uppercased. Put it lower case as most English words are lower case.

koppor:
Have you filed an upstream issue? You know, including minimal example. Similar to @obraliar did it for pgjdb-ng (https://github.com/obraliar/pgjdbc-ng-test#pgjdbc-ng-casing-test).

Maybe you also dived into the CSL library and could propose a fix for the lib (PR at their repo)

koppor:
Besides minor comment: LGTM.

src/main/java/net/sf/jabref/logic/citationstyle/CitationStyleGenerator.java
bartsch-dev: fixed

bartsch-dev:
*Sadly* I cannot reproduce another case where the `TokenMgrError` is thrown.

koppor:
OK for me. - Either we keep this fix or include `1.0.0-SNAPSHOT` as version of the library (see https://github.com/michel-kraemer/citeproc-java/issues/30#issuecomment-254053018).

stefan-kolb:
If I got this right, just use the SNAPSHOT version for now. Is a more long term solution using the fixed version instead of monkey-patching it?

bartsch-dev:
Used the Snapshot which gets rid of the Rhino dependency as well.

PR #2175, opened 2016-10-18
Fix memory leak in entry editor

### LINKED ISSUE #2166: Memory leak (?) when browsing bibtex file with editor open
JabRef development version  on Mac OS X 10.11.4.

Steps to reproduce:

1. Open a small bib database in JabRef. The one I was working with had 19 entries.
2. Open the entry editor by double-clicking on any entry in the table, or through the menu (BibTex > Edit Entry).
3. Move focus back on the main table by selecting an entry.
4. Keep pressing up/down arrow keys and watch the memory usage grow from ~50MB to 2 GB and the number of threads owned by JabRef grow from 27 to over 2000.

Screenshots from activity monitor
screen shot 2016-10-15 at 6 46 33 pm

and visualvm
screen shot 2016-10-15 at 6 28 38 pm

If you are looking at the second chart, you are probably wondering what happened at the end that made the memory/threads stop growing. That was because at this point JabRef had apparently given up on trying to follow which entry I was trying to edit and stopped updating the entry editor altogether.

shitikanth:
This is to fix #2166.

List of changes -
* Updated microba to latest version
* Unregister EntryEditor as a listener to Entry 
* Implement SearchQueryHighlightObservable using EntryBus
* Make EntryEditor unregister all fields that are listening on the global search bar.


koppor:
LGTM.

PR #2178, opened 2016-10-19
Improve TitleCaseFormatterTest

koppor:
I had issues with `[title]` until I found that I should use `[title:capitalize]`. During the search, I thought, `title` was functioning wrong, but it doesn't. Nevertheless, I added a test case and in parallel tried to change so that the one-assert-per-test-case rule holds.


src/test/java/net/sf/jabref/logic/formatter/casechanger/TitleCaseFormatterTest.java
```diff
@@ -17,20 +17,47 @@ public void setUp() {
...
         Assert.assertEquals("Upper Each First", formatter.format("upper each first"));
+    }
+
+    @Test
+    public void anUpperEachFirstAnd() {
         Assert.assertEquals("An Upper Each First And", formatter.format("an upper each first and"));
+    }
+
+    @Test
+    public void anUpperEachOfTheAndFirstAnd1(){
```
tobiasdiez: I don't like the tests names too much since they don't describe the behavior which is expected but the input.
What about `prepositionIsLeftUntouched` and `theAfterColonGetsCapitalized` ?

src/test/java/net/sf/jabref/logic/formatter/casechanger/TitleCaseFormatterTest.java
```diff
@@ -17,20 +17,47 @@ public void setUp() {
...
+    @Test
+    public void anUpperEachFirstWithAndWithoutCurlyBrackets2() {
         Assert.assertEquals("An Upper First with {A}nd without {C}urly {b}rackets",
                 formatter.format("AN UPPER FIRST WITH {A}ND WITHOUT {C}URLY {b}rackets"));
     }
 
     @Test
+    public void testTwoExperiencesTitle() {
+        Assert.assertEquals(
+        "Two Experiences Designing for Effective Security",
```
tobiasdiez: Formatting (indention) seems to be off.

src/test/java/net/sf/jabref/logic/formatter/casechanger/TitleCaseFormatterTest.java
```diff
@@ -17,20 +17,47 @@ public void setUp() {
     }
 
     @Test
-    public void test() {
```
tobiasdiez: Thanks for splitting the tests!

stefan-kolb:
I give my 👍 with @tobiasdiez changes. Go for the merge then.

PR #2187, opened 2016-10-23
Fixed close entry editor block

grimes2:
Something went wrong with the old PR dealing with this issue. So here a new PR. Fixes an issue with closing the entry editor. (I've discarded some additional save action, because this might not be the desired behaviour.)
Here again the bug:
Fixed close action of entry editor not working after parsing (and invalid field) error corrected.
1. generate parsing error in source tab
2. Save database
3. error message > Edit
4. correct parsing error
5. Close entry editor (x) blocked

### DIFF
--- a/src/main/java/net/sf/jabref/gui/entryeditor/EntryEditor.java
+++ b/src/main/java/net/sf/jabref/gui/entryeditor/EntryEditor.java
@@ -1092,7 +1092,7 @@ public class EntryEditor extends JPanel implements EntryContainer {
         }
     }
 
-    public class StoreFieldAction extends AbstractAction {
+    class StoreFieldAction extends AbstractAction {
 
         public StoreFieldAction() {
             super("Store field value");
@@ -1193,8 +1193,8 @@ public class EntryEditor extends JPanel implements EntryContainer {
                         // properly formatted. If that happens, the field
                         // is not stored and the textarea turns red.
                         if (toSet != null) {
-                            new LatexFieldFormatter(Globals.prefs.getLatexFieldFormatterPreferences())
-                                    .format(toSet, fieldEditor.getFieldName());
+                            new LatexFieldFormatter(Globals.prefs.getLatexFieldFormatterPreferences()).format(toSet,
+                                    fieldEditor.getFieldName());
                         }
 
                         String oldValue = entry.getField(fieldEditor.getFieldName()).orElse(null);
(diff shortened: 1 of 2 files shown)


koppor:
You don't need to create a new PR - you can do a force push to overwrite something. No need to create new pull requests. So, you could have force-pushed your new local changes to the existing "origin" branch and no one would have noted. 😇 We can have a private chat about that in case you want 😄

grimes2:
Please set label "ready-for-review". Thanks.

lenhard:
I have tested your PR locally and it works for me. I do not quite like the fact that we are adding code to this huge class, but I do not see a straight-forward alternative right now. If you have an idea how we can extract all the actions out of the `EntryEditor` class, a follow-up PR would be very welcome :)

Since you are modifying it, could you remove the `public` modifier of `StoreFieldAction`. This is not really needed, as far as I can see.

Finally, we should get the travis build to work before merging. The failing tests should be fixed if you merge the current master into your branch.

When these things are done, the PR should be good to go.

lenhard:
Ok, looks good for me. Let us wait for a second developer to double check and merge!

Siedlerchr:
LGTM, too!

2016-10-25 12:53 GMT+02:00 Jörg Lenhard <notifications@github.com>:

src/main/java/net/sf/jabref/gui/entryeditor/EntryEditor.java
```diff
@@ -166,15 +166,12 @@
 
     // text area from getting updated. This is used in cases where the source
     // couldn't be parsed, and the user is given the option to edit it.
-    private boolean lastSourceAccepted = true; // This indicates whether the last
-
-    // attempt
+    private boolean lastFieldAccepted = true;
+    private boolean lastSourceAccepted = true; // This indicates whether the last attempt
     // at parsing the source was successful. It is used to determine whether the
```
bartsch-dev: At first sight it's quite obscure to which field these comment belongs.
Can you please move them (and the other comments) before their corresponding field and use JavaDoc?

grimes2: I don't know to which field the comment belongs. It is not my comment. Parts of the comment seem to be fragmented. Can you help?

bartsch-dev: I'll fix them (and some others) in another PR; less trouble for you!

lenhard:
So, we have two developer who tested the code. Since @bartsch-dev wants to fix the comments in another PR, this one can be merged, which I will do right away.

@grimes2 Once again, thank you for your contribution!

bartsch-dev:
I refactored them (and some more) in https://github.com/JabRef/jabref/pull/2199.

PR #2188, opened 2016-10-23
Replace ".bib.swp" by ".bib.sav" for autosave file

koppor:
While updating `.gitignore` files to contain `.bib.swp`, I [saw that there is a WinEdt setting for `*.sav`](https://www.gitignore.io/api/latex). To avoid complaints by users, I think, we should also use `.sav` for the autosave extension.

I know, that this is the typical backup functionality. However, we use `.bak` during saving an thus, I opted for `.sav`.

WinEdt settings:

WinEdt Settings

Source: http://tex.stackexchange.com/a/240965/9075

### DIFF
--- a/src/main/java/net/sf/jabref/logic/autosaveandbackup/BackupManager.java
+++ b/src/main/java/net/sf/jabref/logic/autosaveandbackup/BackupManager.java
@@ -7,7 +7,6 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardCopyOption;
 import java.util.HashSet;
-import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ArrayBlockingQueue;
@@ -21,6 +20,7 @@ import net.sf.jabref.logic.exporter.BibtexDatabaseWriter;
 import net.sf.jabref.logic.exporter.FileSaveSession;
 import net.sf.jabref.logic.exporter.SaveException;
 import net.sf.jabref.logic.exporter.SavePreferences;
+import net.sf.jabref.logic.util.io.FileUtil;
 import net.sf.jabref.model.database.BibDatabaseContext;
 import net.sf.jabref.model.database.event.BibDatabaseContextChangedEvent;
 import net.sf.jabref.preferences.JabRefPreferences;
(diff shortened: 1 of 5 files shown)


koppor:
Regarding codacy: I need package-level private to enable testing. Should I put the method public accepting the downside that I make internal methods public?

src/main/java/net/sf/jabref/logic/autosaveandbackup/BackupManager.java
```diff
@@ -98,6 +96,13 @@ private void shutdown() {
         }
     }
 
+    static Path getBackupPath(Path originalPath) {
+        Objects.requireNonNull(originalPath.getParent());
+        Path oldFileName = originalPath.getFileName();
+        String newFileName = oldFileName.toString() + BACKUP_FILENAME_ENDING;
```
tobiasdiez: What about [Path.ChangeExtension](https://msdn.microsoft.com/en-us/library/system.io.path.changeextension(v=vs.110).aspx). Ahhh damn, we are in the Java world, so:
probably `return originalPath.resolveSibling(BackupEnding);` (or something similar, but there should be a way to change the extension without converting to string)
And we should probably extract it to a helper class (because its used not only for autosave but also backup).

src/test/java/net/sf/jabref/logic/autosaveandbackup/BackupManagerTest.java
```diff
@@ -0,0 +1,18 @@
...
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+public class BackupManagerTest {
+
+    @Test
+    public void testBackupFileNameGenerationWithLinuxPath() {
```
tobiasdiez: Please follow the naming convention for tests as outlined in the wiki 😸 
That is, no ̀ test` prefix and the test name should state the desired outcome (i.e. getBackupPathAppendsSavExtension)

src/main/java/net/sf/jabref/logic/autosaveandbackup/BackupManager.java
koppor: The code was in `FileSaveSession` - I moved it to `FileUtil`.

We don't change the extension, but merely just add a new one. If we changed it, we would have to rely to String magic - at least according to answers to a [similar question on code review](http://codereview.stackexchange.com/questions/110720/adding-old-to-a-file-name/110739#110739?newreg=7a30458926424b07a700b627007e6701).

src/test/java/net/sf/jabref/logic/autosaveandbackup/BackupManagerTest.java
koppor: Fixed 😇

lenhard:
I agree, looks good! I'll merge.

Siedlerchr:
We have a File util class with renaming method

Am 24.10.2016 4:33 vorm. schrieb "Tobias Diez" <notifications@github.com>:

koppor:
I added the method there. This feature does not renaming (with two given
names), but merely creates a new name.

Issue #2192, opened 2016-10-25
Backup manager error


grimes2:
On my local installation I get the following error after compilation:

```
ERROR net.sf.jabref.logic.autosaveandbackup.BackupManager - Error while deleting the backup file.
java.nio.file.NoSuchFileException path/database.bib.swp

```

This error is maybe not relevant for a "normal" installation, but shouldn't this be fixed?

matthiasgeiger:
Thanks for the report. The problem is already being investigated by @koppor and @obraliar.

PR #2194, opened 2016-10-25
Improve autosave and backup feature

obraliar:
Applies the remaining feedback from https://github.com/JabRef/jabref/pull/2118 and fixes https://github.com/JabRef/jabref/issues/2192.


src/main/java/net/sf/jabref/logic/autosaveandbackup/BackupManager.java
```diff
@@ -32,6 +32,8 @@
 /**
  * Backups the given bib database file from {@link BibDatabaseContext} on every {@link BibDatabaseContextChangedEvent}.
  * An intelligent {@link ExecutorService} with a {@link BlockingQueue} prevents a high load while making backups and rejects all redundant backup tasks.
+ *
```
tschechlovdev: Why is the space here?

src/main/java/net/sf/jabref/logic/autosaveandbackup/BackupManager.java
```diff
@@ -72,8 +74,13 @@ public void run() {
         }
     };
 
+
```
tschechlovdev: Why adding new empty line?

src/main/java/net/sf/jabref/logic/exporter/BibDatabaseWriter.java
```diff
@@ -158,7 +158,11 @@ public E savePartOfDatabase(BibDatabaseContext bibDatabaseContext,
 
         session = saveSessionFactory.createSaveSession(preferences.getEncodingOrDefault(), preferences.getMakeBackup());
 
-        writeDatabaseID(bibDatabaseContext.getDatabase().getDatabaseID());
+        Optional<String> sharedDatabaseIDOptional = bibDatabaseContext.getDatabase().getSharedDatabaseID();
+
+        if (sharedDatabaseIDOptional.isPresent()) {
```
tschechlovdev: why not use ifPresent?

src/main/java/net/sf/jabref/model/database/BibDatabase.java
```diff
@@ -586,15 +588,33 @@ private void relayEntryChangeEvent(FieldChangedEvent event) {
...
+        return Optional.ofNullable(this.sharedDatabaseID);
+    }
+
+    public boolean isShared() {
+        return getSharedDatabaseID().isPresent();
     }
 
     /**
-     * @param databaseID use null to unset the id
+     * @param sharedDatabaseID use null to unset the id
```
tschechlovdev: I thought null shouldn't be used? Shouldn't the clear method used in this case?

src/main/java/net/sf/jabref/logic/autosaveandbackup/BackupManager.java
obraliar: Done.

src/main/java/net/sf/jabref/logic/autosaveandbackup/BackupManager.java
obraliar: Done.

src/main/java/net/sf/jabref/logic/exporter/BibDatabaseWriter.java
obraliar: Not recommended due to Exception handling. `try`-`catch` would take more lines.

tschechlovdev: you could use orElseThrow() (http://stackoverflow.com/questions/28596790/throw-an-exception-if-an-optional-is-present). But from my side it's ok how it is.

koppor:
[LGTM](https://lgtm.in/i/cghBUT10q) (LGTM)

src/test/java/net/sf/jabref/logic/importer/fileformat/BibtexImporterTest.java
```diff
@@ -125,8 +125,7 @@ public void testGetDescription() {
     @Test
     public void testRecognizesDatabaseID() throws Exception {
         Path file = Paths.get(BibtexImporterTest.class.getResource("AutosavedSharedDatabase.bib").toURI());
-        String databaseID = importer.importDatabase(file, StandardCharsets.UTF_8).getDatabase().getDatabaseID();
-        assertEquals("13ceoc8dm42f5g1iitao3dj2ap", databaseID);
+        String sharedDatabaseID = importer.importDatabase(file, StandardCharsets.UTF_8).getDatabase().getSharedDatabaseID().get();
```
tobiasdiez: Compare to Optional.of(...) (also below)

obraliar: Done.

obraliar:
@tobiasdiez 
> Can you please also add a test that the comment at the beginning of the file with the shared database id is not taken as the comment of the first bibentry.

Done. But what's the sense of this test? `%` marks up a comment generally.

src/test/java/net/sf/jabref/logic/importer/fileformat/BibtexImporterTest.java
```diff
@@ -125,7 +126,17 @@ public void testGetDescription() {
...
+    }
+
+    @Test
+    public void testDatabaseIDisNotABibEntryComment() throws Exception {
+        Path file = Paths.get(BibtexImporterTest.class.getResource("AutosavedSharedDatabase.bib").toURI());
+        BibDatabase database = importer.importDatabase(file, StandardCharsets.UTF_8).getDatabase();
+
+        BibEntry entry = database.getEntries().get(0);
+
+        assertTrue(!entry.getField("comment").isPresent());
```
obraliar: Done.

tobiasdiez:
BibTeX does not know the concept of "%"-comments. Everything outside of a BibEntry is considered a comment and is simply ignored by bibtex. However, JabRef safes the text before an entry as "user comments" so that they are kept upon rewriting an entry. However, we don't want to store the shared-db-id as a usercomment, since its an internal JabRef thing.

Please change the test code to "assertEquals("", string)"

obraliar:
@tobiasdiez Done. Also thanks for the explantation.

tobiasdiez:
One last question before merge: does the test fails if you remove the encoding line? (I would prefer if you write the tests against strings as opposed to bib-files)

obraliar:
Yes, it fails...

koppor:
Does that form a second test case? 😇

obraliar:
No, the first one fails. The second one does not check whether a Shared ID is present. I could also add this assert to the second test but its already present in the previous one. Should I?

tobiasdiez:
I think the cleanest way is to add a second test, which makes sure that a single db-string in front of an entry is not captured as "user comment". And then fix the code so that the test passes. 
(please move the tests to the ParserTests and don't use bib files as input but strings)

koppor:
Other than my minor comment: LGTM. @tobiasdiez It seems, your test is now also included and working.

src/test/java/net/sf/jabref/logic/importer/fileformat/BibtexImporterTest.java
```diff
@@ -121,11 +121,4 @@ public void testGetDescription() {
                         + "The metadata is not required to be read here, as this class is NOT called at --import.",
                 importer.getDescription());
     }
-
-    @Test
-    public void testRecognizesDatabaseID() throws Exception {
```
tobiasdiez: and please re-add this test in the parser test class.

tobiasdiez:
Yes, also looks good to me. If the `testRecognizesDatabaseID` is added, the PR can be merged.

src/test/java/net/sf/jabref/logic/importer/fileformat/BibtexImporterTest.java
obraliar: @koppor A test file has to be used as `importer.importDatabase(...)` requires a `File`. @tobiasdiez There was already a testing method `parseRecognizesDatabaseID()`.

tobiasdiez: ok, in this case of course no special importer test is needed. sorry for the confusion 🍡

PR #2195, opened 2016-10-25
Enable version checker to recognize alpha/beta builds

bartsch-dev:
The Update checker is now able to handle alpha/beta builds.
The alpha/beta Tag has to be appended after the version number with a prepended `-` (eg. `3.7-alpha`, `3.7.1-beta`).
I also suggest that the `dev` tag gets prepended by a `-` to make it easier to read (especially if its a pre-stable version (eg. `3.7-alpha-dev--snapshot...`)). Currently this isn't necessary but it is supported.

The versionchecker only suggest non-stable builds if the currently installed version is also not stable (that's why I need to grab all the available versions now).


src/main/java/net/sf/jabref/gui/worker/VersionWorker.java
```diff
@@ -61,22 +63,32 @@ public void done(){
                 return;
             }
 
-            // only respect the ignored version on automated version checks
-            if (latestVersion.equals(toBeIgnored) && !manualExecution) {
-                return;
+            // the newest version, excluding any alpha or beta builds, except if the installed one is one too
+            Version newestVersion = null;
```
tschechlovdev: why don't assign here with installedVersion? then the newestVersiob==null chzeck below is not necessary

bartsch-dev: b/c the `newestVersion` is the new available version which the user gets prompted to download.
If I assign the `installedVersion` it just gets messy.

src/main/java/net/sf/jabref/logic/util/Version.java
```diff
@@ -5,63 +5,93 @@
...
+                this.major = Integer.parseInt(matcher.group("major"));
+
+                String minorString = matcher.group("minor");
+                this.minor = minorString == null ? 0 : Integer.parseInt(minorString);
+
+                String patchString = matcher.group("patch");
+                this.patch = patchString == null ? 0 : Integer.parseInt(patchString);
+
+                String versionStageString = matcher.group("stage");
+                this.developmentStage = DevelopmentStage.getStage(versionStageString == null ? DevelopmentStage.STABLE.stage : versionStageString);
```
tobiasdiez: Does it make sense to integrate `dev` into the development stage enum?

src/main/java/net/sf/jabref/logic/util/Version.java
```diff
@@ -5,63 +5,93 @@
...
-        String[] versionNumbers = versionParts[0].split(Pattern.quote("."));
-        try {
-            this.major = Integer.parseInt(versionNumbers[0]);
-            this.minor = versionNumbers.length >= 2 ? Integer.parseInt(versionNumbers[1]) : 0;
-            this.patch = versionNumbers.length >= 3 ? Integer.parseInt(versionNumbers[2]) : 0;
-            this.fullVersion = version;
-            this.isDevelopmentVersion = version.contains("dev");
-        } catch (NumberFormatException exception) {
-            LOGGER.warn("Invalid version string used: " + version, exception);
+        this.fullVersion = version;
```
tobiasdiez: Could you please extract the parsing logic to a new `Version parse(String version)` method. The constructor should only accept already-parsed values (i.e. major, minor etc as integers and the developmentStage). In this way the Version class adheres to our conventions for parsing from string values.

src/main/java/net/sf/jabref/logic/util/Version.java
```diff
@@ -71,17 +101,28 @@ public boolean isNewerThan(Version otherVersion) {
...
                 return true;
-            } else {
-                return (this.getMinor() == otherVersion.getMinor()) && (this.getPatch() > otherVersion.getPatch());
+            } else if (this.getMinor() == otherVersion.getMinor()) {
+                // if the minors are equal compare the patch numbers
+                if (this.getPatch() > otherVersion.getPatch()) {
+                    return true;
+                } else if (this.getPatch() == otherVersion.getPatch()) {
+                    // The higher the ordinal the stabler the version (this is the cause the enum has to be sorted)
+                    if (this.developmentStage.ordinal() > otherVersion.developmentStage.ordinal()) {
```
tobiasdiez: I would prefer to introduce an integer which specifies the "degree of stability". People tend to ignore comments even if they stare in their face :-)

src/main/java/net/sf/jabref/logic/util/Version.java
```diff
@@ -137,4 +196,35 @@ public String toString() {
...
+        STABLE("");
+
+
+        private final String stage;
+
+        DevelopmentStage(String stage) {
+            this.stage = stage;
+        }
+
+        public static DevelopmentStage getStage(String stage) {
```
tobiasdiez: rename method to `parse`

src/main/java/net/sf/jabref/logic/util/Version.java
```diff
@@ -137,4 +196,35 @@ public String toString() {
...
+                LOGGER.warn("Unknown development stage");
+                return UNKNOWN;
+            } else if (stage.equals(STABLE.stage)) {
+                return STABLE;
+            } else if (stage.equals(ALPHA.stage)) {
+                return ALPHA;
+            } else if (stage.equals(BETA.stage)) {
+                return BETA;
+            }
+            LOGGER.warn("Unknown development stage");
```
tobiasdiez: add the passed stage to the warn message to make debugging easier.

src/main/java/net/sf/jabref/gui/worker/VersionWorker.java
```diff
@@ -61,22 +79,33 @@ public void done(){
...
-                new NewVersionDialog(this.mainFrame, installedVersion, latestVersion);
+            // no new version could be found, only respect the ignored version on automated version checks
+            if (newerVersion == null || (newerVersion.equals(toBeIgnored) && !manualExecution)) {
+                String upToDate = Localization.lang("JabRef is up-to-date.");
+                if (manualExecution) {
+                    JOptionPane.showMessageDialog(this.mainFrame, upToDate, upToDate, JOptionPane.INFORMATION_MESSAGE);
+                }
+                this.mainFrame.output(upToDate);
                 return;
             }
```
bartsch-dev: I extracted that method to 2 more methods and removed the early return.

bartsch-dev: I extracted that method to 2 more methods and removed the early return.

src/main/java/net/sf/jabref/logic/util/Version.java
bartsch-dev: Done

src/main/java/net/sf/jabref/logic/util/Version.java
bartsch-dev: done

src/main/java/net/sf/jabref/logic/util/Version.java
bartsch-dev: done

src/main/java/net/sf/jabref/logic/util/Version.java
bartsch-dev: done

Issue #2184, opened 2016-10-21
Unexpected behaviour when assigning an article to a group using drag'n'drop


AEgit:
JabRef 3.7dev--snapshot--2016-10-14--master--fb0990d
windows 10 10.0 amd64 
Java 1.8.0_111

1. Start with a couple of groups and items/articles within them
2. Reassign one of the articles to a group it had not been assigned to previously. Do this by drag'n'dropping it onto the "new" group
3. Check the "Groups" field of the article: Previous/older group assignments are separated from each other with comma+space, but the newly assigned group is separated from the older ones by just the comma (no space)

When assigning an item/article to a group using drag'n'drop an unexpected behaviour is detected. Previously, when you reassigned an item/article to a new group, the format of the added lines in the "Groups" field would be as follows:

OldGroup_1, OldGroup_2, NewGroup_1, NewGroup_2

Currently, the new groups NewGroup_1 and NewGroup_2 are added as follows:

OldGroup_1, OldGroup_2,NewGroup_1,NewGroup2

i.e., there's no space between the commas separating the new entries.

This behaviour has been introduced in the following version:
 JabRef 3.7dev--snapshot--2016-10-14--master--fb0990d
(this was a fix of another bug, that prevented the use of the drag'n'drop feature for group assignments). See also here:
https://github.com/JabRef/jabref/issues/2098#issuecomment-255267055

ilippert:
i use JabRef 3.7dev--snapshot--2016-10-24--master--7d6ee2b
linux 4.7.9-200.fc24.x86_64 amd64 
Java 1.8.0_111
and I can assign entries via drag and drop.

AEgit:
Hmmm yes, but the problem I'm describing is different - the issue with being able to assign entries via drag'n'drop has indeed been fixed in one of the more recent snapshots.

ilippert:
ok, sorry, now I see how the problem you point to is indeed different. sorry!

AEgit:
@ilippert: No worries ;)

AEgit:
I can confirm that this problem has been solved in;
JabRef 3.7-dev--snapshot--2016-11-03--master--9928179

PR #2196, opened 2016-10-26
Fix for #2184: Add space after separator

matthiasgeiger:
... seems to be as simple as that.

Change has been introduced as the delimitor is no longer saved as a String ", " but as a `Character` which chops away the space after the comma.

- ~~[ ] Change in CHANGELOG.md described~~ introduced in dev Builds

### DIFF
--- a/src/test/java/net/sf/jabref/model/groups/ExplicitGroupTest.java
+++ b/src/test/java/net/sf/jabref/model/groups/ExplicitGroupTest.java
@@ -5,52 +5,58 @@ import java.util.Optional;
 import net.sf.jabref.model.entry.BibEntry;
 import net.sf.jabref.model.entry.FieldName;
 
+import org.junit.Before;
 import org.junit.Test;
 
 import static org.junit.Assert.assertEquals;
 
 public class ExplicitGroupTest {
 
+    private ExplicitGroup group;
+    private ExplicitGroup group2;
+
+    private BibEntry emptyEntry;
+
+    @Before
+    public void setUp() {
+        group = new ExplicitGroup("myExplicitGroup", GroupHierarchyType.INDEPENDENT, ',');
+        group2 = new ExplicitGroup("myExplicitGroup2", GroupHierarchyType.INCLUDING, ',');
+        emptyEntry = new BibEntry();
+    }
 
     @Test
      public void testToStringSimple() {
-        ExplicitGroup group = new ExplicitGroup("myExplicitGroup", GroupHierarchyType.INDEPENDENT, ',');
         assertEquals("ExplicitGroup:myExplicitGroup;0;", group.toString());
     }
 
     @Test
     public void toStringDoesNotWriteAssignedEntries() {
-        ExplicitGroup group = new ExplicitGroup("myExplicitGroup", GroupHierarchyType.INCLUDING, ',');
-        group.add(new BibEntry());
-        assertEquals("ExplicitGroup:myExplicitGroup;2;", group.toString());
+        group.add(emptyEntry);
+
+        assertEquals("ExplicitGroup:myExplicitGroup;0;", group.toString());
     }
 
     @Test
-    public void testEntryStorageSingleGroups() {
-        BibEntry entry = new BibEntry();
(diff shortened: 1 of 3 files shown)


lenhard:
That looks simple indeed. 

Since we cannot let you get away with a one-liner: Please add a test that detects the bug you fixed, so that the build would break if this happens again.

tobiasdiez:
Yes this fix works. I'm unsure if we should use the KeywordList class here (since in the end we assume that the field is a list of words and we are adding a new keyword).

matthiasgeiger:
@tobiasdiez Sorry I don't get it. I've put it in the `KeywordGroup` class as the `add(List<BibEntry> entriesToAdd)` method is not overriden in the subclass `ExplicitGroup`.

Apart from that: If talking only about keywords (without the group context) the keywordSeparator is there followed by a `+" "`, too. (Sorry, I'm currently on my way home and have just closed IntelliJ - so I cannot provide a pointer for now.)

tobiasdiez:
@matthiasgeiger sorry for being cryptic. What I meant was that the class [KeywordList](https://github.com/JabRef/jabref/blob/955b94d7221e5ee3c102731f885eb395e1c2ad64/src/main/java/net/sf/jabref/model/entry/KeywordList.java) probably should be used to modify the field (instead of being text-based).
The code then looks similar to:
````
String oldContent = ...
KeywordList wordlist = new KeywordList(oldContent);
wordlist.add(searchExpression);
String newContent = wordlist.getAsString(seperator);
````

**Comment by matthiasgeiger** [comment]
Ahhh, I see... and this is exactly the position where the `+ "  "` addition is made for keywords.

Will check whether using this is actually possible without sideeffects.

File: src/test/java/net/sf/jabref/model/groups/ExplicitGroupTest.java [hunk 1]
```diff
@@ -20,17 +21,36 @@ public void testToStringSimple() {
...
+    public void testEntryStorageSingleGroups() {
+        BibEntry entry = new BibEntry();
+        ExplicitGroup group = new ExplicitGroup("myExplicitGroup", GroupHierarchyType.INDEPENDENT, ',');
+        group.add(entry);
+        assertEquals(Optional.of("myExplicitGroup"), entry.getField(FieldName.GROUPS));
+    }
+
+    @Test
+    public void testEntryStorageTwoGroups() {
+        BibEntry entry = new BibEntry();
```
tobiasdiez: please extract all the initialization (entry, group, group2) to a setup method and reuse it in the tests.

src/test/java/net/sf/jabref/model/groups/ExplicitGroupTest.java
```diff
@@ -20,17 +21,36 @@ public void testToStringSimple() {
...
+        ExplicitGroup group = new ExplicitGroup("myExplicitGroup", GroupHierarchyType.INDEPENDENT, ',');
+        ExplicitGroup group2 = new ExplicitGroup("myExplicitGroup2", GroupHierarchyType.INDEPENDENT, ',');
+        group.add(entry);
+        group2.add(entry);
+        assertEquals(Optional.of("myExplicitGroup, myExplicitGroup2"), entry.getField(FieldName.GROUPS));
+    }
+
+    @Test
+    public void testGroupStorageAlreadyInGroup() throws Exception {
```
tobiasdiez: please remove `test` prefix from test names and add the expected outcome to the name (e.g. addToAlreadyPresentGroupDoesNotChangeEntry)

src/test/java/net/sf/jabref/model/groups/KeywordGroupTest.java
```diff
@@ -85,4 +88,37 @@ public void containsWordWithWhitespaceInSentence() throws Exception {
 
         assertTrue(group.isMatch(entry));
     }
+
+    @Test
+    public void testGroupStorageEmptyKeywords() throws Exception {
+        KeywordGroup group = new KeywordGroup("name", FieldName.KEYWORDS, "test", false, false, GroupHierarchyType.INDEPENDENT,
```
tobiasdiez: same remarks as above: setup() and test names.

matthiasgeiger:
Done. Will merge after Travis is green.

Issue #1908, opened 2016-09-02
Shortcut for Check Integrity


grimes2:
For my workflow "Check integrity" gets more and more important. I wish I could have a (default) shortcut (keybinding) for that action.

lenhard:
The keybindings have recently been reworked by @JabRef/stupro Maybe this is something that can easily be added?

Do you have a suggestion for which keys would be appropriate for such a binding?

stefan-kolb:
@koppor talks to @JabRef/stupro for a sensible hotkey.

mairdl:
I suggest a shortcut like `CTRL + F8`.

F8 is currently used to `Cleanup entries`. These 2 functions seem to be related so it would make sense to combine them. What do you think about my idea?

koppor:
:+1:

matthiasgeiger:
Fixed with #2198

PR #2198, opened 2016-10-26
Integrity hotkey

mairdl:
The "Check integrity"-action seems to be in need of a hotkey. (See #1908) I used `CTRL + F8` for this action, because F8 is currently used to Cleanup entries. These 2 functions seem to be related so it would make sense to group them.

- [x] internal qs


src/main/java/net/sf/jabref/gui/keyboard/KeyBinding.java
```diff
@@ -13,6 +13,7 @@
     BACK("Back", Localization.lang("Back"), "alt LEFT"),
     CLEANUP(
             "Cleanup", Localization.lang("Cleanup entries"), "F8"),
+    CHECK_INTEGRITY("Check integrity", Localization.lang("Check integrity"), "ctrl F8"),
```
koppor: What was our last consensus about showing menu entries in the hotkey dialog? I think, I opted for using `Localization.lang`, but we might have decided to re-translate these entries. @boceckts Do you remember? I think, we had the discussion. @tobiasdiez Is that documented in our CodeHowTo in the Wiki?

CHANGELOG.md
```diff
@@ -66,6 +66,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - [koppor#61](https://github.com/koppor/jabref/issues/61) Display gray background text in "Author" and "Editor" field to assist newcomers
 - Updated Vietnam translation
 - Added greyed-out suggestion for `year`/`date`/`url` fields
+- [#1908] (https://github.com/JabRef/jabref/issues/1908) Add a shortcut for check integrity
```
koppor: Why is the new shortcut not listed here?

src/main/java/net/sf/jabref/gui/keyboard/KeyBinding.java
boceckts: The thing I rember was about the keybinding **category** where we should use `Localization.menuTitle` as you wrote in https://github.com/JabRef/jabref/pull/1390#issuecomment-238339367. This affects only the javafx branch so far. Other than that I don't remeber a discussion about the menu entries in the hotkey dialog. :thinking:

mairdl: In the `KeyBinding.java` only `Localization.lang` gets used. So I guess it must be right.

src/main/resources/l10n/JabRef_fa.properties
```diff
@@ -2300,3 +2300,5 @@ Do_you_want_to_recover_the_database_from_the_backup_file?=Do_you_want_to_recover
 Firstname_Lastname=
 
 This_might_be_caused_by_reaching_the_traffic_limitation_of_Google_Scholar_(see_'Help'_for_details).=
+
+Check_integrity=?????_???????
```
stefan-kolb: This doesn't look right here?

koppor:
Thanx. So, please, no copy'n'paste solutions here and useage of .menuTitle

mairdl:
I changed it to menuTitle and removed the unused keys.

PR #2199, opened 2016-10-26
Refactor EntryEditor & EntryEditorTab

bartsch-dev:
I refactored many comments (some of them even referred to variables which do not exist anymore).
I removed some unused Methods, replaced some code with already existing methods and refactored a switch-case.

### DIFF
--- a/src/main/java/net/sf/jabref/gui/entryeditor/EntryEditor.java
+++ b/src/main/java/net/sf/jabref/gui/entryeditor/EntryEditor.java
@@ -236,19 +236,13 @@ public class EntryEditor extends JPanel implements EntryContainer {
         addRequiredTab(type);
 
         // optional fields
-        List<String> displayedOptionalFields = new ArrayList<>();
         Set<String> deprecatedFields = new HashSet<>(EntryConverter.FIELD_ALIASES_TEX_TO_LTX.keySet());
         Set<String> usedOptionalFieldsDeprecated = new HashSet<>(deprecatedFields);
 
         if ((type.getOptionalFields() != null) && !type.getOptionalFields().isEmpty()) {
             if (!frame.getCurrentBasePanel().getBibDatabaseContext().isBiblatexMode()) {
-                displayedOptionalFields.addAll(type.getOptionalFields());
-
                 addOptionalTab(type);
             } else {
-                displayedOptionalFields.addAll(type.getPrimaryOptionalFields());
-                displayedOptionalFields.addAll(type.getSecondaryOptionalFields());
-
                 addOptionalTab(type);
 
                 deprecatedFields.add(FieldName.YEAR);
--- a/src/main/java/net/sf/jabref/gui/entryeditor/EntryEditor.java
+++ b/src/main/java/net/sf/jabref/gui/entryeditor/EntryEditor.java
@@ -464,22 +464,6 @@ public class EntryEditor extends JPanel implements EntryContainer {
     }
 
     /**
-     * Rebuild the field tabs. This is called e.g. when a new content selector
-     * has been added.
-     */
-    public void rebuildPanels() {
-        // Remove change listener, because the rebuilding causes meaningless
-        // events and trouble:
-        tabbed.removeChangeListener(tabListener);
-
-        setupFieldPanels();
-        // Add the change listener again:
-        tabbed.addChangeListener(tabListener);
-        revalidate();
-        repaint();
-    }
-
-    /**
      * getExtra checks the field name against InternalBibtexFields.getFieldExtras(name).
      * If the name has an entry, the proper component to be shown is created and
      * returned. Otherwise, null is returned. In addition, e.g. listeners can be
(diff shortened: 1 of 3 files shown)


src/main/java/net/sf/jabref/gui/entryeditor/EntryEditor.java
```diff
@@ -328,7 +327,8 @@ private void addSourceTab() {
         srcPanel.setName(panelName);
         tabbed.addTab(panelName, IconTheme.JabRefIcon.SOURCE.getSmallIcon(), srcPanel, toolTip);
         tabs.add(srcPanel);
-        sourceIndex = tabs.size() - 1; // Set the sourceIndex variable.
+        // Set the sourceIndex variable
```
tschechlovdev: I think the comment can be deleted.

src/main/java/net/sf/jabref/gui/entryeditor/EntryEditorTab.java
```diff
@@ -300,9 +281,7 @@ public FieldEditor getActive() {
 
     public void activate() {
         if (activeField != null) {
-            /**
-             * Corrected to fix [ 1594169 ] Entry editor: navigation between panels
-             */
+            // Corrected to fix [ 1594169 ] Entry editor: navigation between panels
```
tschechlovdev: This can be also deleted I think. Or you could replace it to not point to the bug.

src/main/java/net/sf/jabref/gui/entryeditor/EntryEditor.java
bartsch-dev: done

src/main/java/net/sf/jabref/gui/entryeditor/EntryEditorTab.java
bartsch-dev: done

grimes2:
In the prompt Firstname Lastname... "and others" is not translated. My proposal is to replace it by "et al.". This is international, so no translation needed.
`prompt = String.format("%1$s and %1$s and others", Localization.lang("Firstname Lastname"))`

src/main/java/net/sf/jabref/gui/entryeditor/EntryEditor.java
```diff
@@ -231,19 +236,13 @@ private void setupFieldPanels() {
...
         Set<String> deprecatedFields = new HashSet<>(EntryConverter.FIELD_ALIASES_TEX_TO_LTX.keySet());
         Set<String> usedOptionalFieldsDeprecated = new HashSet<>(deprecatedFields);
 
         if ((type.getOptionalFields() != null) && !type.getOptionalFields().isEmpty()) {
             if (!frame.getCurrentBasePanel().getBibDatabaseContext().isBiblatexMode()) {
-                displayedOptionalFields.addAll(type.getOptionalFields());
-
                 addOptionalTab(type);
             } else {
-                displayedOptionalFields.addAll(type.getPrimaryOptionalFields());
```
stefan-kolb: Why is this removed here? If I remember this right this was needed to distinguish optional fields from fields that are displayed in the other fields tab?

koppor:
`and others` is the bibtex string to procude `et al.` using bibtex/biblatex Therefore, this *must not* be translated. It is contained in the help to educate users to use this string.

src/main/java/net/sf/jabref/gui/entryeditor/EntryEditor.java
bartsch-dev: Items of the list are never used, only added to it (`displayedOptionalFields` was only used in line 234, 240, 244, and 245; or am I missing something?)

bartsch-dev:
Changed the prompt for `Month`-field.
I don't know about the prompts for the other fields, maybe you can point me into the right direction?

grimes2:
From my side it looks ok now. I misinterpreted the author... prompt, because I didn't know the function of `and others` in BibTeX author/editor field. I found this documented in the official specification "Bibtexing".

grimes2:
Just one small addition. I found following person name fields for BibLaTeX:
```
private static final List<String> BIBLATEX_PERSON_NAME_FIELDS = Arrays.asList(FieldName.AUTHOR, FieldName.EDITOR,
            FieldName.EDITORA, FieldName.EDITORB, FieldName.EDITORC, FieldName.TRANSLATOR, FieldName.ANNOTATOR,
            FieldName.COMMENTATOR, FieldName.INTRODUCTION, FieldName.FOREWORD, FieldName.AFTERWORD,
            FieldName.BOOKAUTHOR, FieldName.HOLDER, FieldName.SHORTAUTHOR, FieldName.SHORTEDITOR, FieldName.SORTNAME,
            FieldName.NAMEADDON, FieldName.ASSIGNEE);
```
 I assume that the suggestion syntax
`return String.format("%1$s and %1$s and others", Localization.lang("Firstname Lastname"));`
 is valid for all the mentioned fields above. What do you think? Should we add the missing person name fields to the switch construction?

bartsch-dev:
No I reuse the already existing lists.

koppor:
This PR breaks <kbd>Ctrl</kbd> + <kbd>A</kbd>. See https://github.com/JabRef/jabref/issues/2615

Issue #2208, opened 2016-10-28
Move default focus from search to Main Table for better handling of pasting entries


Siedlerchr:
Very annoying in current master:

Create some entries
Delete one entry
Cursors moves to Seach field
Paste an entry from clipboard (bibtex code)
Entry code appears in search field

koppor:
@bartsch-dev Could you have a look? 😇

PR #2211, opened 2016-10-29
Prevent the search to get focus when deleting/pasting an entry

bartsch-dev:
Fixes: #2208

When you delete an entry the focus will no longer jump to the searchbar, instead it will select the next entry (also counts for pasting).

### DIFF
--- a/src/main/java/net/sf/jabref/gui/BasePanel.java
+++ b/src/main/java/net/sf/jabref/gui/BasePanel.java
@@ -772,6 +772,13 @@ public class BasePanel extends JPanel implements ClipboardOwner, FileUpdateListe
             return;
         }
 
+        // select the next entry to stay at the same place as before (or the previous if we're already at the end)
+        if (mainTable.getSelectedRow() != mainTable.getRowCount() -1){
+            selectNextEntry();
+        } else {
+            selectPreviousEntry();
+        }
+
         NamedCompound compound = new NamedCompound(
                 (entries.size() > 1 ? Localization.lang("delete entries") : Localization.lang("delete entry")));
         for (BibEntry entry : entries) {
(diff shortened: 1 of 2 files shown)


bartsch-dev:
Now it works for cutting entries too.
I removed the code duplicate, although I'm not sure about the the pattern (not able to use `runCommand(Actions.DELETE)`).

src/main/java/net/sf/jabref/gui/BasePanel.java
```diff
@@ -783,7 +785,10 @@ private void delete() {
         getUndoManager().addEdit(compound);
 
         markBaseChanged();
-        frame.output(formatOutputMessage(Localization.lang("Deleted"), entries.size()));
+        frame.output(formatOutputMessage(!cut ? Localization.lang("Deleted") : Localization.lang("Cut"), entries.size()));
+
```
Siedlerchr: Either you change the condition and put cut instead of not cut or you wrap it in if and else

src/main/java/net/sf/jabref/gui/maintable/MainTableSelectionListener.java
```diff
@@ -183,11 +183,10 @@ public void editSignalled() {
 
     public void editSignalled(BibEntry entry) {
         final BasePanelMode mode = panel.getMode();
-        EntryEditor editor = panel.getEntryEditor(entry);
```
Siedlerchr: Why did you change this?

bartsch-dev: The new editor was only needed if the condition was met, and previously the focus was possibly set to the wrong editor (if the condition wasn't met).

src/main/java/net/sf/jabref/gui/BasePanel.java
bartsch-dev: I changed it to cut (but didn't change the semantic').

Siedlerchr:
To me code looks good :+2:

Issue #2210, opened 2016-10-29
Improve error console


koppor:
JabRef 3.7dev--snapshot--2016-10-27--javafx--e489a53
windows 10 10.0 amd64 
Java 1.8.0_101

The error console (see #1383) displays as follows:

![grabbed_20161029-154956](https://cloud.githubusercontent.com/assets/1366654/19830016/b452b2ee-9def-11e6-8e37-7fae6c7add7b.png)

When scrolling down, it displays as follows:

![grabbed_20161029-155309](https://cloud.githubusercontent.com/assets/1366654/19830023/d1bc40c0-9def-11e6-93fd-ed605196807a.png)

I see following issues:
- [x]  When starting, the last line should be displayed, not the first one
- [ ] The colors should be consistent
- [ ] The dialog should allow copy and paste.

Especially, the last point is important to me as I don't want to copy the whole log if something goes wrong, but only the relevant parts.

koppor:
Right click on a single line and copy does not work, because there is no context menu.

When having many errors "Copy Log" does not help:

![grabbed_20161213-134010](https://cloud.githubusercontent.com/assets/1366654/21140669/ad6b4b5e-c139-11e6-875a-aefda21aaf7e.png)

Does someone disagree that the following is not easy to work with for a user?

13:19:10.027 [AWT-EventQueue-0] INFO  net.sf.jabref.logic.importer.OpenDatabase - Opening: C:\git-repositories\jabref\jabref\src\test\resources\testbib\complex.bib
13:19:10.120 [AWT-EventQueue-0] INFO  net.sf.jabref.logic.importer.OpenDatabase - Opening: C:\git-repositories\jabref\jabref\src\test\resources\testbib\crossref.bib
errorconsole.fxml not found, attempting with camel case
errorconsole.css not found, attempting with camel case
errorconsole.fxml not found, attempting with camel case
errorconsole.css not found, attempting with camel case
Dec 13, 2016 1:38:57 PM javafx.scene.CssStyleHelper calculateValueWARNING: Could not resolve '-fx-text-background-color' while resolving lookups for '-fx-text-fill' from rule '*.list-cell' in stylesheet jar:file:/C:/Program%20Files/Java/jre1.8.0_111/lib/ext/jfxrt.jar!/com/sun/javafx/scene/control/skin/modena/modena.bss
[...]

PR #1917, opened 2016-09-03
Introduce cell factory helper (JavaFX)

tobiasdiez:
It is relatively cumbersome to create custom (tree) table cells in JavaFX. The aim of this PR is to introduce helper classes which provide a simpler way to customize tables.

@JabRef/stupro @JabRef/stupro-qa please also review this PR, thanks.

- [ ] Change in CHANGELOG.md described
- [ ] Tests created for changes
- [ ] Screenshots added (for bigger UI changes)
- [x] Manually tested changed features in running JabRef
- [ ] Check documentation status (Issue created for outdated help page at [help.jabref.org](https://github.com/JabRef/help.jabref.org/issues)?)

PR #2213, opened 2016-10-29
Fix strange styling behaviour of the error console

boceckts:
This should fix the wrong styling behaviour of the different log messages when scrolling. It seems like the styling error came because we never resetted the style classes of the cells when adding new ones.
I also created a generic ListFactory class (similar to #1917) so this issue won't occur in the future. Refs #2210

### DIFF
--- a/src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleController.java
+++ b/src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleController.java
@@ -3,12 +3,11 @@ package net.sf.jabref.gui.errorconsole;
 import javafx.fxml.FXML;
 import javafx.scene.control.Button;
 import javafx.scene.control.Label;
-import javafx.scene.control.ListCell;
 import javafx.scene.control.ListView;
 import javafx.stage.Stage;
-import javafx.util.Callback;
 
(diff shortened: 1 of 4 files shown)


src/main/java/net/sf/jabref/gui/util/ViewModelListCellFactory.java
```diff
@@ -0,0 +1,93 @@
+/*
+ * Copyright (C) 2003-2016 JabRef contributors.
```
tobiasdiez: Remove copyright notice

src/main/java/net/sf/jabref/gui/errorconsole/ErrorConsoleController.java
```diff
@@ -59,43 +58,35 @@ private void closeErrorDialog() {
...
-                                        default:
-                                            setText(null);
-                                            setGraphic(null);
-                                            break;
-                                    }
-                                }
-                            }
-                        };
+        allMessages.setCellFactory(new ViewModelListCellFactory<LogEvent>().
+                withGraphic( viewModel -> {
```
tobiasdiez: If you feel like it, you can actually create a real ViewModel around `LogEvent`...

PR #2217, opened 2016-10-30
Autosave for shared databases

obraliar:
This _UI_ extension makes _shared databases_ (and also _autosave_ functionality) more popular as the combination of those two makes it possible to synchronize _bib_ files on the local file system, without saving them manually.

_Screenshot:_
shd_autosave

_Changes:_
Remove _autosave for shared databases_ option from _preftab_ and add an asking panel into connection dialog. 
Change the name of "Open shared database"-menu entries to "Connect to shared database".


src/main/java/net/sf/jabref/gui/actions/ConnectToSharedDatabaseAction.java
```diff
@@ -0,0 +1,32 @@
...
+    public ConnectToSharedDatabaseAction(JabRefFrame jabRefFrame) {
+        super();
+        this.jabRefFrame = jabRefFrame;
+        putValue(Action.NAME, Localization.menuTitle("Connect to shared database"));
+        putValue(Action.SHORT_DESCRIPTION, Localization.lang("Connect to shared database"));
+    }
+
+    @Override
+    public void actionPerformed(ActionEvent e) {
+
```
koppor: remove empty line

src/main/java/net/sf/jabref/gui/exporter/SaveDatabaseAction.java
```diff
@@ -72,6 +73,11 @@ public SaveDatabaseAction(BasePanel panel) {
         this.frame = panel.frame();
     }
 
+    public SaveDatabaseAction(BasePanel panel, Path filePath) {
```
koppor: Could you add a parameter description here?

src/main/java/net/sf/jabref/gui/exporter/SaveDatabaseAction.java
```diff
@@ -85,6 +91,8 @@ public void init() throws Throwable {
 
             panel.frame().output(Localization.lang("Saving database") + "...");
             panel.setSaving(true);
+        } else if (filePath.isPresent()) {
+            saveAs(filePath.get().toFile());
```
koppor: (i) Shouldn't the real working be done in `run`? - Maybe you just have to add some comment here why it is directly saved. Even though, the internal implementation seems to offload this using SPIN.
(ii) Has this something todo with the LOCAL/REMOTE state? Why not using this as checking condition?

src/main/java/net/sf/jabref/gui/shared/ConnectToSharedDatabaseDialog.java
```diff
@@ -100,21 +112,46 @@ public OpenSharedDatabaseDialog(JabRefFrame frame) {
...
 
         if (isSharedDatabaseAlreadyPresent()) {
-            JOptionPane.showMessageDialog(OpenSharedDatabaseDialog.this,
+            JOptionPane.showMessageDialog(ConnectToSharedDatabaseDialog.this,
                     Localization.lang("You are already connected to a database using entered connection details."),
                     Localization.lang("Warning"), JOptionPane.WARNING_MESSAGE);
             return;
         }
 
+        // TODO begin
```
koppor: Why marked as TODO -> This should be removed, shouldn't it?

src/main/java/net/sf/jabref/gui/shared/ConnectToSharedDatabaseDialog.java
```diff
@@ -100,21 +112,46 @@ public OpenSharedDatabaseDialog(JabRefFrame frame) {
...
+                        Localization.lang("'%0' exists. Overwrite file?", localFilePath.getFileName().toString()),
+                        Localization.lang("Existing file"), JOptionPane.YES_NO_OPTION);
+                if (answer == JOptionPane.NO_OPTION) {
+                    fileLocationField.requestFocus();
+                    return;
+                }
+            }
+        }
+
+        // TODO end
```
koppor: remove?

src/main/java/net/sf/jabref/gui/shared/ConnectToSharedDatabaseDialog.java
```diff
@@ -168,6 +205,9 @@ public void actionPerformed(ActionEvent e) {
                 "Enter_pressed");
         connectButton.getActionMap().put("Enter_pressed", openAction);
 
+        browseButton.addActionListener(e -> showFileChooser());
+
```
koppor: Remove empty line, also empty line 207 can be removed.

src/main/java/net/sf/jabref/gui/shared/ConnectToSharedDatabaseDialog.java
```diff
@@ -366,6 +432,10 @@ private void checkFields() throws JabRefException {
             userField.requestFocus();
             throw new JabRefException(Localization.lang("Required_field_\"%0\"_is_empty.", Localization.lang("User")));
```
koppor: Could you remove the `_` in the language strings? They are not required.

src/main/java/net/sf/jabref/gui/shared/SharedDatabaseUIManager.java
```diff
@@ -98,7 +98,7 @@ public void listen(SharedEntryNotPresentEvent event) {
      * @param dbmsConnectionProperties Connection data
      * @param raiseTab If <code>true</code> the new tab gets selected.
      */
-    public void openNewSharedDatabaseTab(DBMSConnectionProperties dbmsConnectionProperties)
+    BasePanel openNewSharedDatabaseTab(DBMSConnectionProperties dbmsConnectionProperties)
```
koppor: Please comment why this is package private? Testing is not the reason. Maybe this should be private or public?

koppor: Please comment why BasePanel is returned. -> Add `@return` JavaDoc.

src/main/resources/l10n/JabRef_da.properties
```diff
@@ -1460,7 +1460,8 @@ Looking_for_full_text_document...=Søger_efter_tekstdokument...
 Autosave=Automatisk_sikkerhedskopi
 A_local_copy_will_be_opened.=
 Autosave_local_databases=
-Autosave_shared_databases=
+Autosave_the_database_to=
+Please_enter_a_valid_file_path.=
```
koppor: Isn't there another string available when the user uses the normal save as or is this functionality missing at the other save operation, too?

src/main/java/net/sf/jabref/gui/shared/ConnectToSharedDatabaseDialog.java
```diff
@@ -100,21 +112,46 @@ public OpenSharedDatabaseDialog(JabRefFrame frame) {
...
+
         setLoadingConnectButtonText(true);
 
         try {
-            new SharedDatabaseUIManager(frame).openNewSharedDatabaseTab(connectionProperties);
+            BasePanel panel = new SharedDatabaseUIManager(frame).openNewSharedDatabaseTab(connectionProperties);
             setPreferences();
             dispose();
+            try {
+                new SaveDatabaseAction(panel, Paths.get("/home/admir/tmp3/demo.bib")).runCommand();
```
tobiasdiez: The path should be replaced with the textfield path....

src/main/resources/l10n/JabRef_de.properties
```diff
@@ -1460,7 +1460,8 @@ Looking_for_full_text_document...=Suche_Volltext-Dokument...
 Autosave=Automatische_Speicherung
 A_local_copy_will_be_opened.=Eine_lokale_Kopie_wird_geöffnet.
 Autosave_local_databases=Automatische_Abspeicherung_für_lokale_Datenbanken
-Autosave_shared_databases=Automatische_Abspeicherung_für_geteilte_Datenbanken
+Autosave_the_database_to=
```
tobiasdiez: change to "Automatically save the database to"

src/main/java/net/sf/jabref/gui/actions/ConnectToSharedDatabaseAction.java
obraliar: Done.

src/main/java/net/sf/jabref/gui/exporter/SaveDatabaseAction.java
obraliar: Done.

src/main/java/net/sf/jabref/gui/exporter/SaveDatabaseAction.java
obraliar: (i) This is just going the same way as the save as functionality
(ii) I wanted to keep that general. It's fully independent of the database location. If a target file is already given, it should only save it there without asking.

src/main/java/net/sf/jabref/gui/shared/ConnectToSharedDatabaseDialog.java
obraliar: Done.

src/main/java/net/sf/jabref/gui/shared/ConnectToSharedDatabaseDialog.java
obraliar: Done.

src/main/java/net/sf/jabref/gui/shared/ConnectToSharedDatabaseDialog.java
obraliar: Done.

src/main/java/net/sf/jabref/gui/shared/SharedDatabaseUIManager.java
obraliar: The method is private as it's only called from the same package. JavaDoc: Done.

src/main/resources/l10n/JabRef_da.properties
obraliar: Unfortunately not. The normal _save as_ dialog does nothing after clicking on `save` when no path is entered. This also seems to be OK.

src/main/java/net/sf/jabref/gui/shared/ConnectToSharedDatabaseDialog.java
obraliar: Oh yes. Done.

obraliar:
@tobiasdiez you can save the shared database locally wherever you want using _save as_ (`CTRL` + [`SHIFT`] + `S`) or using the connection dialog. This part of dialog was not intended to change the path. If you want to change the path, just move the shared database (the _bib_ file) anywhere others on your local file system.
As I described in header, this should only make the file synchroniazation using shared databases more popular.

PR #2223, opened 2016-11-02
Fixes #2222 URLDownload now resolves redirects by default

### LINKED ISSUE #2222: Look up full text ArXiv broken
The functionality added in https://github.com/JabRef/jabref/issues/1023 is broken in the current build. Namely it fails to fetch the pdf file. It generates an empty file without extension. 

Also by default when adding an entry from the ArXiv web search it includes the online link. I would like to deactivate that or a way to order that last with respect to the local PDF copy.

stefan-kolb:

### DIFF
--- a/src/main/java/net/sf/jabref/logic/net/URLDownload.java
+++ b/src/main/java/net/sf/jabref/logic/net/URLDownload.java
@@ -17,6 +17,7 @@ import java.net.CookieHandler;
 import java.net.CookieManager;
 import java.net.CookiePolicy;
 import java.net.HttpCookie;
+import java.net.HttpURLConnection;
 import java.net.MalformedURLException;
 import java.net.URISyntaxException;
 import java.net.URL;
@@ -45,11 +46,11 @@ import org.apache.commons.logging.LogFactory;
  * @author Simon Harrer
  */
 public class URLDownload {
-
-    private final URL source;
-
     private static final Log LOGGER = LogFactory.getLog(URLDownload.class);
 
+    private static final String USER_AGENT= "JabRef";
+
+    private final URL source;
     private final Map<String, String> parameters = new HashMap<>();
 
     private String postData = "";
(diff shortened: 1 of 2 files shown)


src/main/java/net/sf/jabref/logic/net/URLDownload.java
```diff
@@ -119,6 +118,20 @@ private URLConnection openConnection() throws IOException {
...
+        // normally, 3xx is redirect
+        int status = connection.getResponseCode();
+        if (status != HttpURLConnection.HTTP_OK) {
+            if (status == HttpURLConnection.HTTP_MOVED_TEMP
+                    || status == HttpURLConnection.HTTP_MOVED_PERM
+                    || status == HttpURLConnection.HTTP_SEE_OTHER) {
+                // get redirect url from "location" header field
+                String newUrl = connection.getHeaderField("Location");
+                // open the new connnection again
+                connection = (HttpURLConnection) new URLDownload(newUrl).openConnection();
```
koppor: Can this be tested with mockito?

Can't it happen that the new location also points to a new location?

The client library doesn't transparently redirects and there is no utility layer in the lib? Strange....

Siedlerchr: It sounds weird, but it is the truth!
Explaination from the sun java engineers: (Security Reasons)
http://bugs.java.com/bugdatabase/view_bug.do?bug_id=4620571

Siedlerchr: Addtionally we should nontheless add conn.setInstanceFollowRedirects(true);

tobiasdiez:
Could you please also add a test for the arxiv entry which failed to download before the fix.

src/main/java/net/sf/jabref/logic/net/URLDownload.java
stefan-kolb: I think this is the default!

stefan-kolb:
A test for the entry would be no different. It must be done for the URLDownloader, but I'm not entirely sure how right now.

Issue #2220, opened 2016-11-01
Entry editor freeze


grimes2:
JabRef 3.7-dev--snapshot--2016-11-01--master--458ee79
windows 10 10.0 amd64 
Java 1.8.0_111

Steps to reproduce:

1. open entry editor
2. change entry
3. Freeze. 50% CPU

koppor:
Can you do a git bisect run to determine which commit is the cause?

grimes2:
The last commit 458ee7999699b3489684178605c3f8f55294fd63 seems to be bad.

lenhard:
I can confirm the behavior on current master. This renders JabRef completely unusable and is a blocker for a release of 3.7.

koppor:
The commit points to https://github.com/JabRef/jabref/pull/2199. I contacted @bartsch-dev

stefan-kolb:
I cannot reproduce this. After entering something inside the entry editor it should automatically freeze the program? I can edit and change between entries with no problems. When I add a new entry afterwards I get kind of a freeze. But some things are still functional in Jabref. i just cannot enter things in the entry editor or select different entries.

koppor:
When I edit an exiting entry, I have following issues:

1. Open entry
2. Edit "title" field: Add three letters "abc"
3. Click textfield of "journaltitle".
4. Field does not get focus
3. Click textfield of "journaltitle".
4. Field gets focus

bartsch-dev:
I tried to fix it, can you please try it out?
I also have to say that with #2075 the entry editor has become noticeable slower.

grimes2:
Can you provide a build? I want to test it also.

@stefan-kolb: No. Open the entry editor by double-click on an entry in main table. Go to another entry in the main table. You don't need to edit or change a field in the entry editor at all.

koppor:
The PR is #2225. I (manually) mirrored the branch ``fixEditorFreeze` at the JabRef repo. It doesn't build as we have [issues with the Medline test on the master branch](https://travis-ci.org/JabRef/jabref/builds/172742443). The latter refs https://github.com/koppor/jabref/issues/181

You can just checkout the branch or add [bartsch's fork](https://github.com/bartsch-dev/jabref) as additional git remote for fetching.

grimes2:
JabRef 3.7-dev--snapshot--2016-11-02--fixEditorFreeze--d1c1801
windows 10 10.0 amd64 
Java 1.8.0_111

[...]

Issue #282, opened 2015-10-30
Ensure that entry editor has minimal height


grimes2:
Double-click on entry in table of references doesn't open entry editor. Win10, JabRef 2.11 beta4. With 2.10 entry editor opens.

koppor:
Could you try latest 2.11 development snapshot at http://ge.tt/9Idc446?

grimes2:
No success with 2.11dev - 2015-10-20 01.30.45

koppor:
Do you see some output in the "Error console"? (Help -> Show error console). Check all tabs.

If not, please export your preferences (Options -> Preferences -> Export preferences), reset them (`java -jar JabRef-....jar -d all` and try again. If everything works now, your preferences are broken somehow...

grimes2:
I've resetted the preferences. Everything works now. Preferences were broken. Thank you!

koppor:
Could you send us the preferences? My Email-Adress should be public on the GitHub profile. Think, you are not the only one having issues with broken preferences (see #174).

Mixopteryx:
I had the same problem. Fixed it by exporting the preferences and changing the line `<entry key="entryEditorHeight" value="4"/>` to `<entry key="entryEditorHeight" value="500"/>` and reimporting.

tobiasdiez:
I reopen the issue since I think JabRef should prevent heights which are too small.

matthiasgeiger:
There was a bug related to this that has been already fixed (#1450 and #1463) - but intential small sizes where not covered by this fix.

What would be a "height which is too small"? Use 10-50px minimal size to indicate that the editor is open?

Mixopteryx:
For me 30 px seems like a sensible minimum, at least at my system (windows 7). Then I see the top menu and would know that I could just drag it up. Its still not usable for anything so I wouldn't mind a bigger minimum size.
This wasn't intentional by me. I do not know how I managed to set the size of the entry editor to 4 px in the first place. The least height I can make by dragging seem to be 297 px, so it seems to be some kind of a limit in place already.

grimes2:
I can't minimize the entry editor in newest development version. Is this fixed already? Can't find the commit.

stefan-kolb:
Seems to be fixed in current `master`.

PR #2225, opened 2016-11-02
Fix editor freeze

bartsch-dev:
Possibly fixes: #2220.

Mainly I removed `SwingUtilities.invokeLater(() ->`. I also added some lines to remove the event listener from the old entry editors.

### DIFF
--- a/src/main/java/net/sf/jabref/gui/BasePanel.java
+++ b/src/main/java/net/sf/jabref/gui/BasePanel.java
@@ -1614,29 +1614,17 @@ public class BasePanel extends JPanel implements ClipboardOwner, FileUpdateListe
 
         }
 
-        EntryEditor entryEditor;
-        int divLoc = -1;
         String visName = null;
         if ((getShowing() != null) && isShowingEditor()) {
             visName = ((EntryEditor) splitPane.getBottomComponent()).getVisiblePanelName();
         }
-        if (getShowing() != null) {
-            divLoc = splitPane.getDividerLocation();
-        }
 
         // We must instantiate a new editor.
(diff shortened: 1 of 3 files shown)


src/main/java/net/sf/jabref/gui/BasePanel.java
```diff
@@ -1614,29 +1614,17 @@ public void showEntry(final BibEntry be) {
...
 
         // We must instantiate a new editor.
-        entryEditor = new EntryEditor(frame, BasePanel.this, be);
+        EntryEditor entryEditor = new EntryEditor(frame, BasePanel.this, be);
         if (visName != null) {
             entryEditor.setVisiblePanel(visName);
         }
-        splitPane.setBottomComponent(entryEditor);
-
-        if (divLoc > 0) {
```
koppor: Is this code appearing somewhere else? I remember, we had some issues with the minimum editor height (#1450 and #282). Does this code removal affect that, too?

koppor:
For @grimes2, [this fixed the issue](https://github.com/JabRef/jabref/issues/2220#issuecomment-258010791).

src/main/java/net/sf/jabref/gui/BasePanel.java
bartsch-dev: The height of the editor (or preview) is handled by the `BasePanel#adjustSplitter()` method which just sets the saved height.

The height of the preview panel is saved in `BasePanel#saveDividerLocation()` without any maximum/minimum size adjustment. Currently the minimum size of the preview panel is 5px.
 
The entry editor height is saved in 3 places without any adjustments, but the minimum height of the editor is limited via the editor toolbar and the layout of the tabs (which depends on the entry type) .

PR #2230, opened 2016-11-03
Fix maintable sorting

### LINKED ISSUE #2200: Sorting by title is broken
JabRef version 3.6 and current master

[As reported in the forum](http://discourse.jabref.org/t/sorting-by-any-column/282) sorting by title does not sort ALL entries correctly: 

In fact there seem to be two independent lists which are sorted correctly. Without a look at the good I could not find a pattern which entries are assigned to which list...

matthiasgeiger:
Fixes #2200 by using the latex free value of the field for sorting the main table.

However, just using `getLatexFreeField()` was not possible as the table is using `getFieldOrAlias`. Thus I introduced a new method `getFieldOrAliasLatexFree()` which is based on the previous implementation of `getFieldOrAlias()` using lambdas. This approach is used instead of formatting the `getFieldOrAlias()` result for performance reasons as the latex-free field values are cached in the `BibEntry`.

*Sidenote: If this approach is approved by you, the same should be applied to `BibDatabase.getResolvedFields()` which is used in `MainTableColumn.getColumnValue()` to show a latex-free version of the contents in the main table)*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -111,6 +111,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - Fixed close action of entry editor not working after parsing error corrected
 - Fixed [koppor/#128](https://github.com/koppor/jabref/issues/128): Sensible default settings for "Enable save actions" and "Cleanup"
 - Fixed [#2201](https://github.com/JabRef/jabref/issues/#2201) and [#1825](https://github.com/JabRef/jabref/issues/#1825): Status of the Group panel is saved and reused for next startup of JabRef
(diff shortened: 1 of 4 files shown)


lenhard:
I had a look at the code and tested it locally. The usage of functional interfaces and method references really touches the boundaries of my Java knowledge here. I thought a little over it, but came to the same conclusion: The only way to avoid duplicating the code of the `getAlias` method is to introduce an interface and delegate the invocation of `getField` to the proper implementation (with or without LaTeX-freeness). Your solution is much more concise than a traditional one would be. This will make `BibEntry` a little harder to understand for now, but the right way for us (me) is to master the new capabilities of Java and not keeping the code to an older level.

The failing tests in branch build come from failing fetcher, which are fixed on master, so I am merging right away.

Issue #2233, opened 2016-11-04
Month field not shown in BibLaTeX mode


grimes2:
JabRef 3.7-dev--snapshot--2016-11-04--master--949ca17
windows 10 10.0 amd64 
Java 1.8.0_111

In BibLaTeX mode a `month = {...}` field in the source is not shown up in the entry editor. This field should be in the `Deprecated fields` tab

BibLaTeX manual says:
> If there is no date field in an entry, biblatex will also consider the fields year and month
for backwards compatibility with traditional BibTeX.

PR #2236, opened 2016-11-05
BibLaTeX mode: Add month and year field to deprecated tab

grimes2:
Issue #2233: In BibLaTeX the `date` field is recommended. For backward compatibility the BibTeX fields `month` and `year` are also accepted. I've added both fields to the deprecated fields tab, although the year field is present in the required fields tab too.

### DIFF
--- a/src/main/java/net/sf/jabref/gui/entryeditor/EntryEditor.java
+++ b/src/main/java/net/sf/jabref/gui/entryeditor/EntryEditor.java
@@ -263,8 +263,7 @@ public class EntryEditor extends JPanel implements EntryContainer {
                 // Get all optional fields which are deprecated
                 usedOptionalFieldsDeprecated.retainAll(optionalFieldsAndAliases);
 
-                // Get deprecated parts of the date field
-                usedOptionalFieldsDeprecated.add(FieldName.YEAR);
+                // Get other deprecated fields
                 usedOptionalFieldsDeprecated.add(FieldName.MONTH);
 
                 // Add tabs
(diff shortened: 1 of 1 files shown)


stefan-kolb:
@koppor has an opinion here?!

koppor:
In my experience, in a typical setting, one does not know the correct date, but only the year. Together with an experienced librarian (@georgd), I maintain a high-quality example .bib file at https://github.com/latextemplates/biblatex-lni/blob/master/LNI-examples.bib. There, we use `year` and `urldate`, but never `date`.

I currently don't see, where the month file is currently shown.

So I would vote for keeping as is and not merge.

grimes2:
The problem is, that an existing month field is not shown at all. The month field should be shown in deprecated or in other fields tab. I prefer the deprecated tab, because the month field is mentioned in the BibLaTeX manual. I think, it is better to remove the year field from deprecated fields, because it is present in required field.

koppor:
The year is IMHO not in the depricated fields, is it? Could you provide a screenshot?

We can add the month to the deprecated fields.

tobiasdiez:
The biblatex manual states

> biblatex will also consider the fields year and month
for backwards compatibility with traditional BibTeX but this is not encouraged as
explicit year and month are not parsed for date meta-information markers or
times and are used as-is.

which is a polite way to say "they are deprecated". Thus I support @grimes2 here and think these two fields should be added to the deprecated tab. @koppor you can also put only the year in the date field if you don't know the month/day of publication.

koppor:
Typically, bibtex things from the web have year and not date. This IMHO
won't change in the future... I'm also in favour for putting journal in the
required fields and not other fields. Reason: Main bib file biblatex, but
paper bib file in bibtex. No converter for biblatex to bibtex and not
forcing everyone to use biblatex/jabref.

PR #2248, opened 2016-11-08
Moved return out of for-loop

dierkes:
I found some time to look into the code. It seems, only the last conditional statement was accidentally placed inside the for loop.

I fixed that and now the behavior is as before, as far I can see.


dierkes:
This pull request belongs to #1899

Siedlerchr:
Thanks for your contribution! Yeah, that could have been the reason... for the failure.
I will check it later.

Siedlerchr:
Looks good!
The failing tests are just Some failing fetchers
Can be merged into the branch 
Thanks for your contribution

PR #2250, opened 2016-11-08
[WIP] Use v8 engine for CSL

shitikanth:
Changes made:
- Added a compile dependency on com.eclipsesource.j2v8 in build.gradle

We are using citeproc-java library in CitationStyleGenerator to generate previews, which itself is basically a wrapper around the javascript library citeproc-js. Switching to V8 engine dramatically improves the time taken to generate these previews. For example, on my laptop, generating preview used to take about 20s for the first preview, and 10s for each subsequent preview; using V8 reduces both by a factor of 10.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -65,6 +65,26 @@ configurations {
     integrationTestRuntime.extendsFrom testRuntime
 }
 
+def getV8Dependency() {
+    def arch = System.getProperty('os.arch')
+    def dep = null
+    if (OperatingSystem.current().isWindows()) {
+        dep = 'win32_x86'
+        if (arch.equals('amd64')) {
+            dep += '_64'
+        }
+    } else if (OperatingSystem.current().isMacOsX() && (arch.equals('amd64') || arch.equals('x86_64'))) {
+        dep = 'macosx_x86_64'
+    } else if (OperatingSystem.current().isLinux() && arch.equals('amd64')) {
+        dep = 'linux_x86_64'
+    }
+    if (dep == null) {
+        logger.error("Could not find V8 runtime compatible to this system")
+        return null
+    }
+    return 'com.eclipsesource.j2v8:j2v8_' + dep + ':4.5.0'
+}
+
 dependencies {
     compile 'com.jgoodies:jgoodies-common:1.8.1'
     compile 'com.jgoodies:jgoodies-forms:1.9.0'
@@ -123,6 +143,8 @@ dependencies {
     compile 'org.citationstyles:styles:1.0.1-SNAPSHOT'
     compile 'org.citationstyles:locales:1.0.1-SNAPSHOT'
     compile 'de.undercouch:citeproc-java:1.0.0-SNAPSHOT'
+
+    compile getV8Dependency()
 
     testCompile 'junit:junit:4.12'
     testCompile 'org.mockito:mockito-core:2.2.9'
--- a/build.gradle
+++ b/build.gradle
@@ -144,7 +144,10 @@ dependencies {
     compile 'org.citationstyles:locales:1.0.1-SNAPSHOT'
     compile 'de.undercouch:citeproc-java:1.0.0-SNAPSHOT'
 
-    compile getV8Dependency()
+    // if available, use platform specific build of V8
+    if (getV8Dependency() != null) {
+        compile getV8Dependency()
+    }
 
     testCompile 'junit:junit:4.12'
     testCompile 'org.mockito:mockito-core:2.2.9'
--- a/external-libraries.txt
+++ b/external-libraries.txt
@@ -177,6 +177,6 @@ Licence: Apache License, Version 2.0
 Id:      com.eclipse.j2v8
 Project: J2V8
 URL:     https://github.com/eclipsesource/J2V8
-Licence: EPL- 1.0
+Licence: EPL-1.0
 
 The last entry has to end with an empty line. Otherwise the entry is not present in About.html.
--- a/external-libraries.txt
+++ b/external-libraries.txt
@@ -174,4 +174,9 @@ Project: Citeproc-Java
 URL:     http://michel-kraemer.github.io/citeproc-java/
 Licence: Apache License, Version 2.0
 
+Id:      com.eclipse.j2v8
+Project: J2V8
+URL:     https://github.com/eclipsesource/J2V8
+Licence: EPL- 1.0
+
 The last entry has to end with an empty line. Otherwise the entry is not present in About.html.


matthiasgeiger:
Hi and thanks for your contribution.

However, I suppose it will be difficult to actually use `com.eclipsesource.j2v8` in this way: If the concrete packages are in fact platform dependent it won't be possible to provide a platform independet jar file anylonger.

shitikanth:
Ooh good point! If the slow performance of CSL citation generation isn't an issue for others, I wouldn't bother about it too much. Maybe we can include V8 in the platform specific binaries though?

build.gradle
```diff
@@ -65,6 +65,26 @@ configurations {
...
+            dep += '_64'
+        }
+    } else if (OperatingSystem.current().isMacOsX() && (arch.equals('amd64') || arch.equals('x86_64'))) {
+        dep = 'macosx_x86_64'
+    } else if (OperatingSystem.current().isLinux() && arch.equals('amd64')) {
+        dep = 'linux_x86_64'
+    }
+    if (dep == null) {
+        logger.error("Could not find V8 runtime compatible to this system")
+        return null
```
koppor: Can't we just return an empty string here so that it compiles at other platforms, too?

koppor:
Note that the code is copied from https://github.com/michel-kraemer/citeproc-java/blob/master/citeproc-java/build.gradle#L26

https://github.com/michel-kraemer/citeproc-java/blob/master/citeproc-java/build.gradle#L56:

```
    if (getV8Dependency() != null) {
        provided getV8Dependency()
}
```

Could that be a solution? Maybe replace `provided` by `compile`?

build.gradle
shitikanth: Hmm, that is a good idea, but the jar we produce will still not be platform independent (as @matthiasgeiger mentioned). Maybe we can define different gradle tasks for a platform independent jar and platform specific builds.

What do you guys feel about just directly implementing citeproc natively in Java? I can take it up if you think that it will be worthwhile.

koppor: Seeing that (i) the current rendering is very slow and (ii) others will surely use that library too, citeproc natively in Java is a huge win for the community. 

I fear, however, that it will be a huge effort and we would also need that effort in JabRef itself.

koppor:
I would vote for generating specific JARs and package them for MacOSX and Windows. For linux, we could generate two JARs, too.

koppor:
Could you also update [external-libraries.txt](https://github.com/JabRef/jabref/blob/master/external-libraries.txt) as described at https://github.com/JabRef/jabref/blob/master/CONTRIBUTING.md#when-adding-a-library. We need that to ease debian packaging (refs https://github.com/koppor/jabref/issues/135).

koppor:
I used your patch here with the new engine. It is very fast. Seeing that, I think, we should invest time for a correct build of JabRef and not reimplementing the whole functionality in Java. That time should better be spent in JabRef itself 😇

Siedlerchr:
Another option would be to check if it works wirh jdk8 nashorn engine as discussed here https://github.com/michel-kraemer/citeproc-java/issues/28

shitikanth:
@Siedlerchr, I did try building with the nashorn engine, the performance wasn't improved by very much at least on my machine.

external-libraries.txt
```diff
@@ -174,4 +174,9 @@ Project: Citeproc-Java
 URL:     http://michel-kraemer.github.io/citeproc-java/
 Licence: Apache License, Version 2.0
 
+Id:      com.eclipse.j2v8
+Project: J2V8
+URL:     https://github.com/eclipsesource/J2V8
+Licence: EPL- 1.0
```
koppor: Could you remove the space before `1.0` so it is a valid [SPDX license identifier]( CC-BY-SA-3.0)?

shitikanth: Done!

koppor:
@shitikanth Could you try to adapt the gradle build file to create a plattform-specific jar and a plattform-neutral jar? The speedup of the V8 engine is huge and I would really like to get this into the next JabRef release.

shitikanth:
@koppor I am not familiar with JabRef's build/release process, so it will help me if can you specify what behaviour you want. For example, does the following expected behaviour look okay to you?

* `./gradlew releaseJar`: creates platform specific build like JabRef-<version>-linux_x86-64.jar 
* `./gradlew shadowJar`: creates a platform independent build.

matthiasgeiger:
I think during release the shadowJar variant is used which creates a fat-jar. The needed behavior would be to create 5 different fat-Jars: Windows x86, Win x64, Linux, MacOSX,  and independent - which is then used to create the platform specific installers using install4j. 

@stefan-kolb Can you assist here? Is it possible to use different jars in install4j?

stefan-kolb:
Yes, shadowJar is used right now as far as I know. As for the executables, we probably need to add more `launchers` inside install4j for every configuration.

koppor:
To illustrate the performance gain, [I made](https://screentogif.codeplex.com/) two videos:

## without V8

without-v8

## with V8

with-v8

stefan-kolb:
Hm, I ask myself why this stuff is taking so long anyway? It is just a String creation wtf?

koppor:
http://eclipsesource.com/blogs/2014/11/17/highly-efficient-java-javascript-integration/:

> Unlike other JS Runtimes (including JV8, Jav8), J2V8 takes a primitive based approach, resulting in much less garbage.

I am not in the internals of JavaScript engines and cannot really compare [Rhino](https://github.com/mozilla/rhino) and [Nashorn](http://www.oracle.com/technetwork/articles/java/jf14-nashorn-2126515.html) to [J2V8](https://github.com/eclipsesource/J2V8).

koppor:
IMHO the alternative is to implement citeproc-java directly in Java as suggested by @shitikanth at https://github.com/JabRef/jabref/pull/2250#discussion_r87088785. Maybe, this could be a project for the time between the years? 😇

koppor:
I merged to `use-V8-for-citeproc` to enable other developers continuing working on it.

build.gradle
rmzelle: > What do you guys feel about just directly implementing citeproc natively in Java? I can take it up if you think that it will be worthwhile.

Hi all, CSL maintainer here. I have never worked on a citeproc implementation myself, but just a friendly warning that supporting the full CSL 1.0.1 specification is a significant undertaking. I'm cc'ing some developers who might give you some additional pointers/warnings.

(cc @michel-kraemer (citeproc-java), @fbennett (citeproc-js), @inukshuk (citeproc-ruby))

michel-kraemer: I can confirm this. In fact, I initially thought about creating a pure Java implementation myself, but citeproc-js is so mature and well-maintained, you'll probably never reach the same quality -- at least I would never be able to do so.

Anyway, there are a couple of options to improve the performance of citeproc-java. You already discovered you can use V8, although it isn't documented anywhere. You might be able to improve the performance further by caching the processor instance. But maybe you already do so, I don't know your code well enough. If you're interested we can discuss this topic further. I'd be more than happy to help.

michel-kraemer:
> Hm, I ask myself why this stuff is taking so long anyway? It is just a String creation wtf?

The `citeproc.js` file is 664 KB large and the code is very complex. It seems Nashorn is not very fast when compiling and running such a file for the first time. However, since it uses JIT compilation, it gets very fast the longer the program runs and the more often you execute the file.

Issue #2239, opened 2016-11-07
Localization of upcoming release 3.7


matthiasgeiger:
Dear @JabRef/translators ,

we are about to release the next version of JabRef.

It would be great if you find the time to translate all, or some :wink: of the missing translations.

The current status is:

### Localization files status (2016-11-07 08:52 - Branch `master` `dd7a757`)

| Property file | Keys | Keys translated | Keys not translated | % translated |
| ------------- | ---- | --------------- | ------------------- | ------------ |
| JabRef_en.properties | 1576 | 1576 | 0 | 100 |
| JabRef_da.properties | 1576 | 956 | 620 | 60 |
| JabRef_de.properties | 1576 | 1564 | 12 | 99 |
| JabRef_es.properties | 1576 | 1368 | 208 | 86 |
| JabRef_fa.properties | 1576 | 5 | 1571 | 0 |
| JabRef_fr.properties | 1576 | 1564 | 12 | 99 |
| JabRef_in.properties | 1576 | 1210 | 366 | 76 |
| JabRef_it.properties | 1576 | 1565 | 11 | 99 |
| JabRef_ja.properties | 1576 | 1554 | 22 | 98 |
| JabRef_nl.properties | 1576 | 642 | 934 | 40 |
| JabRef_no.properties | 1576 | 953 | 623 | 60 |
| JabRef_pt_BR.properties | 1576 | 1189 | 387 | 75 |
| JabRef_ru.properties | 1576 | 1143 | 433 | 72 |
| JabRef_sv.properties | 1576 | 1394 | 182 | 88 |
| JabRef_tr.properties | 1576 | 1560 | 16 | 98 |
| JabRef_vi.properties | 1576 | 961 | 615 | 60 |
| JabRef_zh.properties | 1576 | 1207 | 369 | 76 |

| Property file | Keys | Keys translated | Keys not translated | % translated |
| ------------- | ---- | --------------- | ------------------- | ------------ |
| Menu_en.properties | 108 | 108 | 0 | 100 |
| Menu_da.properties | 108 | 93 | 15 | 86 |
| Menu_de.properties | 108 | 108 | 0 | 100 |
| Menu_es.properties | 108 | 93 | 15 | 86 |
| Menu_fa.properties | 108 | 93 | 15 | 86 |
| Menu_fr.properties | 108 | 108 | 0 | 100 |
| Menu_in.properties | 108 | 93 | 15 | 86 |
[...]

PR #2256, opened 2016-11-09
Localization in Russian

mlep:
Files from wuw, reformated a bit.
(related to issue https://github.com/JabRef/jabref/issues/2239)

### DIFF
--- a/src/main/resources/l10n/JabRef_ru.properties
+++ b/src/main/resources/l10n/JabRef_ru.properties
@@ -78,7 +78,7 @@ All_fields=Все_поля
 
 All_subgroups_(recursively)=Все_подгруппы_(рекурсивно)
 
-Always_reformat_BIB_file_on_save_and_export=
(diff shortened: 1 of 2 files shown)


Siedlerchr:
There seems to be some keys more than once inside, according to the failed test:

`net.sf.jabref.logic.l10n.LocalizationConsistencyTest > allFilesMustHaveSameKeys FAILED

    java.lang.AssertionError: Missing keys of ru expected:<[]> but was:<[Connection_to_%0_server_established., New_technical_report, The_BibEntry_you_currently_work_on_has_been_deleted_on_the_shared_side., The_connection_to_the_server_has_been_terminated.]>`

koppor:
I think, it's the other way round - these keys are missing. One has to run `c:\tools\python2\python scripts\syncLang.py update -e` and handle the result. I fixed it in 2441675677812ae3d852b80, which means, this is merged.

PR #2260, opened 2016-11-10
Update JabRef_es.properties

jtornero:
Update localization strings for JabRef - spanish


matthiasgeiger:
Great! Thank you for the huge update!

PR #2265, opened 2016-11-13
Fix translation warning

koppor:
In the JavaFX branch, JabRef outputs

```
09:52:15.355 [AWT-EventQueue-0] WARN  net.sf.jabref.logic.l10n.Localization - Warning: could not get message translation for "Check integrity" for locale en
09:52:15.356 [AWT-EventQueue-0] WARN  net.sf.jabref.logic.l10n.Localization - Warning: no message translation for "Check integrity" for locale en
```

This is fixed with this PR.

When digging deeper, it seems also other calls to `Localization.lang` should be replaced by `Localization.menuTitle`.

For other cases, like "Copy_BibTeX_key_and_title" (or "Open Database"), some calls might get replaced. The RightClickMenu calls `lang` with that string, whereas JabReFrame calls `menuTitle`. I would opt for changing the call of RightClickMenu. Reason: When adding a new language, first the menu strings are translated, then the other strings. To offer a huge coverage of the translation, these translated strings should be used at so many places as possible.

"Open file" also appears twice, this should be consolidated.

"Quit JabRef" seems to be at the wrong file: It is only in the full language, but this is a global menu action.


lenhard:
I agree with your suggestions regarding the usage of `menuTitle`.

As for this PR, however, the localization tests fail. So please fix them and we should be ready to merge.

Issue #684, opened 2016-01-20
ISBN to BibTeX Fetch updates


ajbelle:
Issues  that seem left over from http://sourceforge.net/p/jabref/bugs/1241/

I tried 0-444-41881-4 and  0444418814 which return  an “Invalid ISBN” message. It is a valid ISBN and all other ISBN sites find the book. The message is misleading and should be “ISBN metadata unavailable”. Perhaps unlike http://manas.tungare.name/software/isbn-to-bibtex/  http://www.ebook.de/ does not accept  the old 10 digit ISBN. I tried 9780444418814 (the 13 digit converted ISBN) and ebook.de returned the same message. SOMETHING SHOULD BE ADDED TO THE HELP FILE TO EXPLAIN THIS MISLEADING “Invalid ISBN” MESSAGE. 

Trying to figure out why I read http://jabref.sourceforge.net/faq.php "Q: I have an ISBN. Is it possible to create an entry directly out of the ISBN?" and http://jabref.sourceforge.net/help/ISBNtoBibTeXHelp.php which are still to reflect what has changed, unlike the JabRef Help src/main/resources/help/en/ISBNtoBibTeXHelp.html (It is good that the embedded help is current). PLEASE UPDATE THE ONLINE CONTENT.

koppor:
Thank you for your report. This is OUTSTANDING.

If you read the [CHANGELOG of 2.11 beta 4](https://github.com/JabRef/jabref/blob/dev_2.11/CHANGELOG), you would have read ""ISBN to BibTeX" fetcher now uses eBook.de's API (fixes bug [#1241](https://sourceforge.net/p/jabref/bugs/1241/))". Then,you would have read some background information. Sure, you cannot have read the private conversation between me and @manastungare. We came up with the solution that http://www.ebook.de (via @freiheit-com) implemented a completely new API for us. It has the drawback, as you noted, that only books available via ebook.de can be converted to bibtex.

Maybe you have noticed that we are under heavy load with other things. Maybe it is unnoticed, but we are NOT paid for JabRef at all.

It would be very helpful if you would try to [fork JabRef](https://github.com/JabRef/jabref/blob/master/CONTRIBUTING.md) and change the error message to "no information about this ISBN is available from ebook.de". The source of the homepage is available at https://github.com/JabRef/htdocs. We really need help migrating that to www.jabref.org. See #133 and #111.

If your valuable time allows it, please support us! This would really bring JabRef a huge step forward. :running:

[...]

PR #2266, opened 2016-11-13
When an ISBN is not available on ebook.de, the error message is more clear.

koppor:
When fetching a valid ISBN from ebook.de, but the site has no data of the ISBN, the error message just showed "FileNotFoundException". This PR fixes that.

The issue itself is a kind of regression as it has been fixed for verison 3.3 - see #684.

### DIFF
--- a/src/main/java/net/sf/jabref/logic/importer/IdBasedParserFetcher.java
+++ b/src/main/java/net/sf/jabref/logic/importer/IdBasedParserFetcher.java
@@ -1,6 +1,7 @@
 package net.sf.jabref.logic.importer;
 
 import java.io.BufferedInputStream;
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.MalformedURLException;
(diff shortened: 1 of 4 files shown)


Siedlerchr:
Lgtm failing Test seems to be the astrophysics fetcher

koppor:
The test failed because of the changed implementation. I adapted the test accordingly in cdf6261.

MathSciNet timeouts, which seems to be unrelated to the improvement above.

src/main/java/net/sf/jabref/logic/importer/IdBasedParserFetcher.java
```diff
@@ -80,6 +81,9 @@ default void doPostCleanup(BibEntry entry) {
             return Optional.of(entry);
         } catch (URISyntaxException e) {
             throw new FetcherException("Search URI is malformed", e);
+        } catch (FileNotFoundException e) {
```
matthiasgeiger: Why do you catch this here and not at the IsbnFetcher?

src/main/java/net/sf/jabref/logic/importer/IdBasedParserFetcher.java
```diff
@@ -80,6 +81,9 @@ default void doPostCleanup(BibEntry entry) {
...
+            LOGGER.debug("Id not found");
```
matthiasgeiger: Exception should be logged.

src/main/java/net/sf/jabref/logic/importer/IdBasedParserFetcher.java
koppor: The IsbnFetcher returns the URL only. The fetching is done in this class.

PR #2278, opened 2016-11-16
URLs can be passed as command line arguments to the import

tobiasdiez:
If an URL is passed as an argument to the `-import` or `-importToOpen` command in the console, then the file is first downloaded and afterwards imported as usual. This functionality will enable JabFox to pass PDFs directly to JabRef for importing.

Further changes:
- [Changed](https://github.com/JabRef/jabref/compare/master...tobiasdiez:commandLine?expand=1#diff-e415888225bbbeb20a948c9ff07de910R278) the log level for parser errors. `ImportFormatReader.importUnknownFormat` first tries to open the file as a bib-file and thus this logging resulted in many log-entries if the file is not in the BibTeX format.
- Fixed an NPE which occurred when the ParserResult of the imported file had no path attached.

~~Will add a changelog entry and documentation change later. DONE~~


src/main/java/net/sf/jabref/logic/importer/ImportException.java
```diff
@@ -0,0 +1,17 @@
+package net.sf.jabref.logic.importer;
+
+public class ImportException extends Exception {
```
lenhard: Why do we need a new Exception here? I mean, what is the benefit of knowing that this is specific to the importer? My main point is that we do not really have a strategy as to when we add a new exception and when to use a specific exception.

How about just reusing the existing JabRefException? Does the job as well, as far as I see.

tobiasdiez: Well, the code would work with JabRefException instead of ImportException, but you would loose the semantics that the exception is really related to the import and not because JabRef runs out of memory (or another general problem). But I have no strong opinion about it (mainly because you are right that we don't have a general strategy regarding exceptions).

ImportException now derives from JabRefException. Is this a compromise you can live with?

src/main/java/net/sf/jabref/cli/ArgumentProcessor.java
```diff
@@ -536,54 +538,65 @@ public boolean isBlank() {
...
-                    Path file;
-                    if (OS.WINDOWS) {
-                        file = Paths.get(data[0]);
-                    } else {
-                        file = Paths.get(data[0].replace("~", System.getProperty("user.home")));
-                    }
-                    ParserResult result = Globals.IMPORT_FORMAT_READER.importFromFile(data[1], file);
+        String address = data[0];
+        Path file;
+        if (address.startsWith("http://") || address.startsWith("https://")) {
```
Siedlerchr: What about FTP?

src/main/java/net/sf/jabref/cli/ArgumentProcessor.java
```diff
@@ -536,54 +538,65 @@ public boolean isBlank() {
...
+                System.err.println(
+                        Localization.lang("Problem downloading from %1", address) +
+                        e.getLocalizedMessage());
+                return Optional.empty();
+            }
+        } else {
+            if (OS.WINDOWS) {
+                file = Paths.get(address);
+            } else {
+                file = Paths.get(address.replace("~", System.getProperty("user.home")));
```
Siedlerchr: I see not reason for doing this.
a Path object is os independent and it can handle relative and absolute file path

src/main/java/net/sf/jabref/logic/net/URLDownload.java
```diff
@@ -186,17 +188,55 @@ private void copy(InputStream in, Writer out, Charset encoding) throws IOExcepti
...
+     */
+    public Path downloadToTemporaryFile() throws IOException {
+        // Determine file name and extension from source url
+        String sourcePath = source.getPath();
+
+        // Take everything after the last '/' as name + extension
+        String fileNameWithExtension = sourcePath.substring(sourcePath.lastIndexOf('/') + 1);
+
+        // View the last '.' as separator between the name and the extension
+        int dotPosition = fileNameWithExtension.lastIndexOf('.');
```
Siedlerchr: We already have a Util method in FileUtil.getExtension

src/main/java/net/sf/jabref/cli/ArgumentProcessor.java
tobiasdiez: Good point, thanks for the suggestion. Added

src/main/java/net/sf/jabref/cli/ArgumentProcessor.java
tobiasdiez: I didn't add this, it was already there. Since I have no linux system right now to check if it works without the special "~" handling, I prefer not to change it right now. Feel free to open a new PR.

src/main/java/net/sf/jabref/logic/net/URLDownload.java
```diff
@@ -186,17 +188,55 @@ private void copy(InputStream in, Writer out, Charset encoding) throws IOExcepti
...
+    /**
+     * Downloads the web resource to a temporary file.
+     *
+     * @return the path to the downloaded file.
+     */
+    public Path downloadToTemporaryFile() throws IOException {
+        // Determine file name and extension from source url
+        String sourcePath = source.getPath();
+
+        // Take everything after the last '/' as name + extension
```
tobiasdiez: `source` is a URL and thus `source.getPath()` is just a string and not a nio.Path

src/main/java/net/sf/jabref/logic/net/URLDownload.java
tobiasdiez: Thanks for the hint. Changed it now.

src/main/java/net/sf/jabref/logic/net/URLDownload.java
```diff
@@ -186,17 +188,55 @@ private void copy(InputStream in, Writer out, Charset encoding) throws IOExcepti
...
+    @Deprecated
     public void downloadToFile(File destination) throws IOException {
+        downloadToFile(destination.toPath());
+    }
+
+    public void downloadToFile(Path destination) throws IOException {
 
         try (InputStream input = new BufferedInputStream(openConnection().getInputStream());
-             OutputStream output = new BufferedOutputStream(new FileOutputStream(destination))) {
+                OutputStream output = new BufferedOutputStream(new FileOutputStream(destination.toFile()))) {
```
tobiasdiez: Nice!

PR #2281, opened 2016-11-17
Fix #2279: File permissions are kept upon save

### LINKED ISSUE #2279: Jabref changes the file permissions of the database
JabRef version 3.7 on Ubuntu 16.04

If you save the database it changes the permissions of the file to 600 (read/write for user and no access to anybody else). This can be a problem in shared databases where all users belong to the same group, but are different users.

tobiasdiez:
Fixes #2279.
JabRef does not directly writes to the bib-file upon save but creates a temporary file which is then copied over the original one. In this process the original file permissions get lost. Should be fixed with this PR (I have not tried it...).
I pushed the code to the branch https://github.com/JabRef/jabref/tree/filePermissions but for some reason the setup files do not show up on [builds.jabref.org](http://builds.jabref.org/)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ This project **does not** adhere to [Semantic Versioning](http://semver.org/).
 This file tries to follow the conventions proposed by [keepachangelog.com](http://keepachangelog.com/).
 Here, the categories "Changed" for added and changed functionality,
 "Fixed" for fixed functionality, and
-"Removed" for removed functionality is used.
+"Removed" for removed functionality are used.
 
 We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#NUM`.
 
@@ -14,6 +14,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 ### Changed
 
 ### Fixed
+- We fixed an issue where the file permissions of the .bib-file were changed upon saving [#2279](https://github.com/JabRef/jabref/issues/2279).
 
 ### Removed
 
--- a/src/main/java/net/sf/jabref/logic/exporter/FileSaveSession.java
+++ b/src/main/java/net/sf/jabref/logic/exporter/FileSaveSession.java
@@ -4,6 +4,9 @@ import java.io.IOException;
 import java.nio.charset.Charset;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.attribute.PosixFilePermission;
+import java.util.EnumSet;
+import java.util.Set;
 
 import net.sf.jabref.logic.util.io.FileBasedLock;
 import net.sf.jabref.logic.util.io.FileUtil;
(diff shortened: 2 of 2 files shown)


lenhard:
Good job! Give it some time. If there is no build tomorrow, try restarting. Then @g-mos should give it a try and see if it keeps permissions.

g-mos:
I would be happy to test whatever is needed. Since I am new to this, inform me and tell me, when and what I need to download.

koppor:
@tobiasdiez Could you create a branch at the JabRef main repo?

tobiasdiez:
@koppor I pushed to a branch in the main rep (https://github.com/JabRef/jabref/tree/filePermissions) but no build appears under http://builds.jabref.org/.

matthiasgeiger:
A test failed at circleCI... I've restarted the build: https://circleci.com/gh/JabRef/jabref/8328

tobiasdiez:
Thanks @matthiasgeiger, it worked!

@g-mos could you please install the development version from http://builds.jabref.org/filePermissions/ and test if the permission issue is indeed resolved. Thank you!

stefan-kolb:
LGTM 👍 
Just for clarification: why don't we write directly to the bib file? Would be cleaner?!

g-mos:
I can confirm that the permissions of the files are kept. Therefore the issue at hand is fixed.
BUT! Due to the logic as @tobiasdiez described every time a different user saves the file user/group owners of the file change. For common usage this is not important, even for my scenarios this is not important. I presume that you need the "opening new file philosophy" for triggering when actual changes happen in a group/collaboration environment. But to add to @stefan-kolb idea, instead of creating a new file can you clone at the beginning and open the clone? Then save to the original? Just as an idea?
Either way thanks @tobiasdiez for the quick fix!

CHANGELOG.md
```diff
@@ -4,7 +4,7 @@ This project **does not** adhere to [Semantic Versioning](http://semver.org/).
 This file tries to follow the conventions proposed by [keepachangelog.com](http://keepachangelog.com/).
 Here, the categories "Changed" for added and changed functionality,
 "Fixed" for fixed functionality, and
-"Removed" for removed functionality is used.
+"Removed" for removed functionality are used.
```
Siedlerchr: Why "are"?
Functionality is singular 
Either you let the is or you use functionalities

tobiasdiez: the object is "categories" which is plural.

tobiasdiez:
@g-mos Thanks for your feedback (and reporting the issue in the first place).
I will now merge this PR, so the fix will be included in the next version.
I don't think we will completely change how files are saved in the near future. I think the reason for writing everything to a new file is that it provides an additional layer of security. If something gets messed up during the writing process, the original file is untouched and without damage.

g-mos:
@tobiasdiez I agree for the security layer. My idea is 
a. copy the file, so that permissions, owners stay the same
b. open the copy
c. if the user saves --> save to the original file.
I do not know if this is feasible....

Siedlerchr:
@g-mos  Due to a problem on windows I slightly had to change the code. Follow up is in  #2286 
I would like you to test the build from here:
http://builds.jabref.org/copyPermissions/

g-mos:
Unfortunately @Siedlerchr, this new version returns to the original problem. The permissions are completely destroyed. rw for user and no access to everybody else. It is also changing of user/group ownership to the user that shaved the jabref file. Sorry for the bad news.

Siedlerchr:
@g-mos Okay, thanks for the feedback. I will investigate this further. In the meantime you can revert back to the version before.

Issue #1359, opened 2016-05-06
Collect usage statistics


tobiasdiez:
Automatically (and of course anonymously) collect usage statistics, for example about:
- changed preferences (to determine better default values)
- number of open databases, size of db, ...
- used functionality (to extract main use cases)

Maybe also occasionally ask questions about the way users use JabRef and/or what they want (a lite version of the recent survey).

Of course, the should have to option to opt out of this data collecting.

simonharrer:
We could leverage Google Analytics for this which allows to send arbitrary events to the cloud for analysis.

tobiasdiez:
In this case, we may use https://github.com/siddii/jgoogleanalytics or some related project.

Siedlerchr:
As an alternative we could use Piwik:
https://github.com/piwik/piwik-java-tracker 
It can be self-hosted and would be a bit more user-privacy-friendly

simonharrer:
I do not like this self-hosted stuff, as it draws developer resources away from the more important things - improving JabRef :)

tobiasdiez:
Decision from devcall: use google.

Siedlerchr:
Google analytics no longer works. There is no actual version which can be used.

lenhard:
This can be closed as it has been implemented in #2283 with Microsoft Azure.

PR #2283, opened 2016-11-18
Implement #1359: collect telemetry

tobiasdiez:
In this PR, [Microsoft Azure Application Insights](https://docs.microsoft.com/en-us/azure/application-insights/app-insights-overview) is used to record the following information about how JabRef is used:
- Number of users and sessions
- Dialogs opened (so far only the About dialog is tracked, as proof of concept)
- New database is opened, along with how many entries it contains
- All exceptions are automatically recorded
- A few user information (Country, OS, screen size, JabRef version, Java version)

If you give your ok, I will continue and implement the following:
- Track all dialogs
- Add "opt-out" preference

In https://github.com/JabRef/jabref/issues/1359, we decided to use google analytics for this, but as Microsoft already provided a nice interface for java, I went for this route.


lenhard:
Well, since it was decided in the devcall to implement this, it is ok for me if you go ahead. Regarding the choice of the tracking technology, I trust your skills. The code that is needed to track a dialog seems to be very simple in any case.

I am just not sure if an opt-out is the way to go, or if it should rather be an opt-in. Maybe we could display a dialog asking for consent on first startup. That would be more privacy-friendly in my point of view and we do not want JabRef to appear that it tries to grab user data. I am sure there would be enough people who are willing to contribute their data even in case of opt-in.

matthiasgeiger:
:+1: for opt-in

src/main/resources/ApplicationInsights.xml
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0" encoding="utf-8"?>
+<ApplicationInsights xmlns="http://schemas.microsoft.com/ApplicationInsights/2013/Settings" schemaVersion="2014-05-30">
+ 
+  <!-- The key from the Azure portal: -->
+  <InstrumentationKey>c2435fd3-cb59-423c-846b-04898b101cc1</InstrumentationKey>
```
stefan-kolb: We should probably inject this later via an ENV variable?

stefan-kolb:
Nice that you give it a try 😄 👍 GA had no appripriate library?

grimes2:
Why do you want to spy almost everything I do?

lenhard:
@grimes2 My guess (I haven't participated in the discussion on this) is that this would enable us to see which features are actually used and which are not. If we had had data on content selector usage, we would not have removed them. At the same time, we would be able to see if there is a point at all in optimizing JabRef usage for really large databases or if this is a waste of time.

src/main/resources/ApplicationInsights.xml
koppor: Security by obscurity. In that way, the key is not in the source, but in TravisCI environment variables, where only JabRef developers have access to by using "echo $key" in the travis.yml in a pull request. If we can somehow encrypt the JAR file, we can further prevent others from stealing the key.

tobiasdiez:
I added tracking of other dialogs and changed the default to "do not share telemetry data". Also, after the first minute of using JabRef, the user is asked whether he wants to share is data. I actually preferred to have a more non-invasive notification similar to 
!(ie)[http://media.askvg.com/articles/images3/Speedup_Browsing_Disable_Addons_Popup_IE9.png].
JavaFX provides a NotificationPanel exactly for this scenario, however, it doesn't work if the rest of the application is in swing.

src/main/resources/ApplicationInsights.xml
tobiasdiez: Done!

lenhard:
I tried it out locally, the dialog appears as desired, and the option can be disabled via the preferences. Code-wise, I have no objections.

However, since this is a very sensitive topic, I think we should make the dialog a little more explanatory. Currently it says 

> To improve the user experience, we would like to collect data on the features you use. No personal data will be collected.


While everything is in that statement, I think we should be more explicit and repeat ourselves. Suggestion: 

> To improve the user experience, we would like to collect anonymous statistics on the features you use. We will only record what features you access and how often you do it. We will neither collect any personal data nor the content of bibliographic items. The collected data will be stored on XXX and we will only use it to prioritize our development efforts.

I know this is quite long, but since this is a delicate topic, I think it is necessary. Btw.: where is the data actually stored?

@JabRef/developers I think this is so important that everyone should consider it briefly.

And a last comment: We will get flack from the Linux freaks when they find out that we dare to include a microsoft library. So be prepared for it ;-)

tobiasdiez:
I like your suggestion for the dialog text and updated it accordingly. The data is store on Microsoft Azure server...no idea where exactly 😄 .

The reason for the error message is that a correct instrumentation key is only inserted by travis and thus you get these messages when you run JabRef from the IDE.

lenhard:
By the way, this is what I just got after installing soapUI:
soapui

Of course, they have a much more explicit privacy policy page for that.

PR #2287, opened 2016-11-19
Add fallback exception handler

tobiasdiez:
Previously, uncaught exceptions were simply ignored by our logging framework. With these changes, they are properly logged in log4j.
One would actually want to show a nice error dialog to the user, but well, this is something for another time https://github.com/JabRef/jabref/issues/2288.

### DIFF
--- a/src/main/java/net/sf/jabref/gui/util/FallbackExceptionHandler.java
+++ b/src/main/java/net/sf/jabref/gui/util/FallbackExceptionHandler.java
@@ -1,0 +1,24 @@
+package net.sf.jabref.gui.util;
+
+import org.apache.logging.log4j.LogManager;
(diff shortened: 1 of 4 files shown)


src/main/java/net/sf/jabref/gui/util/FallbackExceptionHandler.java
```diff
@@ -0,0 +1,24 @@
+package net.sf.jabref.gui.util;
```
lenhard: This does not belong in the gui package, but in JabRef's root package, where the classes that use it reside. Please move it.

tobiasdiez: Ok, done!

Issue #2221, opened 2016-11-01
Customizable drop down menus for general fields are gone


wilthan:
Until 3.6 I had a drop down menu set up in a custom general field.
The code in the bib file was:
```
"@Comment{jabref-meta: selector_status:approved;captured;captured, references   only;marked;needGDCmod;not to capture;order;ordered;cancelled by library;out of    scope;printed;processing;recieved;status;}"
```

This option is now gone and the line is deleted when I save the file with:
JabRef 3.7dev--snapshot--2016-10-18--master--8f7ae51
windows 7 6.1 amd64 
Java 1.8.0_66

It's a bug for me but I'm not sure if it was decided to remove this feature.

```
Put the excerpt of the log file here
```

koppor:
They have been removed in #2068.

I personally never used them. I guess, your use case is to quickly change the content of a field with pre-defined words?

Could you provide a screenshot of the selectors in action? Could you elaborate a bit more why this was a useful feature and why the auto completion does not help?

An alternative could be the [special fields](http://help.jabref.org/en/SpecialFields), but they are hard-coded, whereas you seem to search for arbitrary flags.

@lenhard @tobiasdiez

lenhard:
Wow, so there are indeed people who seemed to have used that feature. It should not be too hard to revert the PR, I guess, but the question is if we really want to. As far as I see it, adding a special term to some field (keywords, other, etc.) would achieve the same and we would not have to maintain and drag along the UI parts. @wilthan Some motivation on why this feature is important, should be re-added and cannot simply be replaced by some other existing mechanism would be nice! Then we could discuss what we do at the next devcall.

koppor:
There was also the [feature request 669 on sourceforge](https://sourceforge.net/p/jabref/feature-requests/669/), which required the feature for "the systematic review of medical outcome literature, an important process in the practice of evidence based medicine (EBM)". If I understood that correctly, with customizable combo boxes, JabRef could possible ease the review process of researchers.

lenhard:
@koppor: So did they ever develop the plugin they promised in that feature request? I am unable to find it.

wilthan:
The difference to auto complete for me is that you get a list of options presented. If you don't know the available options you can't use autocomplete.
I implemented a custom field labeled: status
The options there were: 
- approved
- captured
- references only
- marked
- order (from library)
- ordered
- out of scope

With the drop down my students were able to just pick from the list to update the status of a specific reference. 
[...]

PR #2301, opened 2016-11-20
[WIP] Readd Drop Down content selectors

Siedlerchr:
First attempt at readding the code for 
#2221 
Refs #2068 

Code does not compile yet, some adjustments need to be made regarding Metadata stuff etc


Siedlerchr:
@tobiasdiez  There are some things with the Metadata to make it compile again. Could you please have a look? thanks!

tobiasdiez:
How to handle the metadata is indeed a problem. I remember that I removed the content selectors mainly because they were the last not-so-easy-to-convert-to-the-new-metadata item. The next few days I will not find the time to assist you here, sorry.

Siedlerchr:
Still some language related errors.

lenhard:
The localization keys existed before the feature was removed. I wonder why the revert did not restore them in the language files?

Siedlerchr:
I did a manual revert, because there were some more changes which would have led to conflicts otherwise.
Was somehow decided.

matthiasgeiger:
As this was a rather old feature I suppose there were translations for most languages... Can you readd the translated values as well? Or should I go  digging in the history? 😉

lenhard:
So I re-implemented parsing and serialization of content selectors and added tests for it. It is probably not the most beautiful fashion, but I need to (want to) link to the old code and not have to re-write the GUI. 

Speaking of the GUI: The content selectors seem to be working fine if the MetaData is in the bib file. The GUI in the EntryEditor is not actually beautiful, but I guess it has never been (does anybody feel like polishing the UI?). However, I do not see Options –> Manage content selectors. Maybe that has not been ported from  #2068 yet? Maybe someone who is more familiar with the menu than me can have a look? :)

Finally, the test currently say:
```
net.sf.jabref.logic.help.HelpFileTest > referToValidPage FAILED
```
Although there seems to be a help page here: http://help.jabref.org/en/ContentSelector Someone certainly has a pointer for that?

Siedlerchr:
Regarding the help file, there is an enum called HelpFiles or similar  for
the name of the help page

I can take a look at the dialog thing in the menu

Am 30.11.2016 5:10 nachm. schrieb "Jörg Lenhard" <notifications@github.com>:
>
> So I re-implemented parsing and serialization of content selectors and
added tests for it. It is probably not the most beautiful fashion, but I
need to (want to) link to the old code and not have to re-write the GUI.
>
> Speaking of the GUI: The content selectors seem to be working fine if the
MetaData is in the bib file. The GUI in the EntryEditor is not actually
beautiful, but I guess it has never been (does anybody feel like polishing
the UI?). However, I do not see Options –> Manage content selectors. Maybe
that has not been ported from #2068 yet? Maybe someone who is more familiar
with the menu than me can have a look? :)
>
> Finally, the test currently say:
>
> net.sf.jabref.logic.help.HelpFileTest > referToValidPage FAILED
>
> Although there seems to be a help page here:
http://help.jabref.org/en/ContentSelector Someone certainly has a pointer
for that?

tobiasdiez:
I pushed a commit which should take care of the menu item (hopefully, I directly edited in on github).

lenhard:
@tobiasdiez Indeed, it does! Nice :) That means the feature is back!

Regarding the help files, there is `CONTENT_SELECTOR("ContentSelector"),` in `net.sf.jabref.logic.help.HelpFile`, so the test really should not fail, should it?

lenhard:
Ok, seems we are back on track, only the google scholar test is failing now, which, I hope, is just a random connection problem on the side of travis. The same might have applied for the help file test.

So now we can do some more code review and at a certain point, have the users from #2221 try out the feature to make sure that it works.

src/main/java/net/sf/jabref/gui/BasePanel.java
```diff
@@ -193,6 +193,8 @@
     // when a Back or Forward operation is being processed:
     private boolean backOrForwardInProgress;
 
+    private final Map<String, EntryEditor> entryEditors = new HashMap<>();
```
tobiasdiez: If I see it correctly, then this map is always empty. Even if the content selectors need to be updated, I would strongly prefer that the event framework is used instead of push-forward through the BasePanel.

src/main/java/net/sf/jabref/gui/ContentSelectorDialog2.java
```diff
@@ -0,0 +1,513 @@
...
+
+    /**
+     *
+     * @param owner the parent Window (Dialog or Frame)
+     * @param frame the JabRef Frame
+     * @param panel the currently selected BasePanel
+     * @param modal should this dialog be modal?
+     * @param fieldName the field this selector is initialized for. May be null.
+     */
+    public ContentSelectorDialog2(Window owner, JabRefFrame frame, BasePanel panel, boolean modal, String fieldName) {
```
tobiasdiez: Remove the "2" in the name?

src/main/java/net/sf/jabref/gui/FieldContentSelector.java
```diff
@@ -0,0 +1,226 @@
...
+    public void rebuildComboBox() {
+        comboBox.removeAllItems();
+
+        // TODO: CO - What for?
+        comboBox.addItem("");
+        for (String item : metaData.getContentSelectors(editor.getFieldName())) {
+            comboBox.addItem(item);
+        }
+    }
+    // Not used since the comboBox is not editable
```
tobiasdiez: I think we can delete these comments, right?

src/main/java/net/sf/jabref/gui/entryeditor/EntryEditor.java
```diff
@@ -484,7 +502,14 @@ private void setupToolBar() {
         } else if (fieldExtras.contains(FieldProperty.JOURNAL_NAME)) {
             // Add controls for switching between abbreviated and full journal names.
             // If this field also has a FieldContentSelector, we need to combine these.
-            return FieldExtraComponents.getJournalExtraComponent(panel, editor, entry, getStoreFieldAction());
+
+            // If this field also has a FieldContentSelector, we need to combine these.
```
tobiasdiez: two times the same comment?

src/main/java/net/sf/jabref/gui/importer/actions/AppendDatabaseAction.java
```diff
@@ -152,7 +155,16 @@ private static void mergeFromBibtex(JabRefFrame frame, BasePanel panel, ParserRe
                 frame.getGroupSelector().addGroups(newGroups, ce);
             });
         }
+        /*
+        FIXME: import selectors
```
tobiasdiez: Indeed :-)

src/main/java/net/sf/jabref/model/metadata/ContentSelector.java
```diff
@@ -0,0 +1,24 @@
...
+import java.util.Collections;
+import java.util.List;
+
+public class ContentSelector {
+
+    private final String fieldName;
+
+    private final List<String> values;
+
+    public ContentSelector(String fieldName, List<String> keys) {
```
tobiasdiez: rename keys -> values

src/main/java/net/sf/jabref/model/metadata/MetaData.java
```diff
@@ -147,6 +151,23 @@ public boolean isProtected() {
         return isProtected;
     }
 
+    public Optional<ContentSelectors> getContentSelectors() {
+        return Optional.ofNullable(contentSelectors);
```
tobiasdiez: I think `contentSelectors` is never null, but of course may be empty. Thus the `Optional` part should be removed.

src/main/java/net/sf/jabref/model/metadata/ContentSelectors.java
```diff
@@ -0,0 +1,62 @@
...
+    public static ContentSelector parse(String key, String values) {
+        Objects.requireNonNull(key);
+        Objects.requireNonNull(values);
+
+        List<String> valueList = Arrays.asList(values.split(";"));
+
+        return new ContentSelector(key, valueList);
+    }
+
+    public Map<String, List<String>> getSelectorData() {
```
tobiasdiez: I don't like that `ContentSelectors` exposes essentially all its data directly to the outside. People will reimplement the same things again (getSelectorValuesForField == getSelectorData.get). I think the only usage of getSelectorData so far is in the foreach statements, for which it would be preferable to implement `Iterable`.

src/main/java/net/sf/jabref/model/metadata/ContentSelectors.java
```diff
@@ -0,0 +1,62 @@
...
+        }
+
+        return Collections.emptyList();
+    }
+
+    public void removeSelector(String fieldName) {
+        contentSelectors.remove(fieldName);
+    }
+
+    public List<ContentSelector> getContentSelectors(){
```
tobiasdiez: See remark below for getSelectorData

src/main/java/net/sf/jabref/model/metadata/MetaData.java
```diff
@@ -147,6 +151,23 @@ public boolean isProtected() {
...
+    public void addContentSelector(ContentSelector contentSelector) {
+        this.contentSelectors.addContentSelector(contentSelector);
+        postChange();
+    }
+
+    public void clearContentSelectors(String fieldName) {
+        contentSelectors.removeSelector(fieldName);
+    }
+
+    public List<String> getContentSelectors(String fieldName) {
```
tobiasdiez: rename to `getContentSelectorValuesForField` or really return the content selector. By the way, the name `ContentSelector` is really strange. It is more a provider than a selector...

src/main/java/net/sf/jabref/model/metadata/MetaData.java
```diff
@@ -248,16 +269,18 @@ public boolean equals(Object o) {
...
-                .equals(citeKeyPatterns, metaData.citeKeyPatterns) && Objects.equals(userFileDirectory,
-                metaData.userFileDirectory) && Objects.equals(defaultCiteKeyPattern, metaData.defaultCiteKeyPattern)
-                && Objects.equals(saveActions, metaData.saveActions) && mode == metaData.mode && Objects.equals(
-                defaultFileDirectory, metaData.defaultFileDirectory);
+        return (isProtected == metaData.isProtected) && Objects.equals(groupsRoot, metaData.groupsRoot)
+                && Objects.equals(encoding, metaData.encoding)
+                && Objects.equals(saveOrderConfig, metaData.saveOrderConfig)
+                && Objects.equals(citeKeyPatterns, metaData.citeKeyPatterns)
+                && Objects.equals(userFileDirectory, metaData.userFileDirectory)
+                && Objects.equals(defaultCiteKeyPattern, metaData.defaultCiteKeyPattern)
```
tobiasdiez: Add contentSelectors

src/main/resources/l10n/JabRef_da.properties
```diff
@@ -519,8 +519,6 @@ Field_to_filter=Felt_som_skal_filtreres
 
 Field_to_group_by=Grupperingsfelt
 
-File=Fil
```
tobiasdiez: This entry is deleted and readded in every localization file.

src/main/resources/l10n/JabRef_da.properties
```diff
@@ -2305,3 +2303,11 @@ Recommended_for_%0=
 This_might_be_caused_by_reaching_the_traffic_limitation_of_Google_Scholar_(see_'Help'_for_details).=
 
 Problem_downloading_from_%1=
+
+Manage_content_selectors=
+<field_name>=
```
tobiasdiez: I think these strings should be localized without the brackets surrounding them.

src/test/java/net/sf/jabref/logic/exporter/MetaDataSerializerTest.java
```diff
@@ -48,7 +49,18 @@ public void serializeSingleSaveAction() {
...
+        List<String> values = new ArrayList(4);
+        values.add("approved");
+        values.add("captured");
+        values.add("received");
+        values.add("status");
+
+        metaData.addContentSelector(new ContentSelector("status", values));
+
+        Map<String, String> expectedSerialization = new TreeMap<>();
+        expectedSerialization.put(MetaData.SELECTOR_META_PREFIX + "status",
```
tobiasdiez: Inline `MetaData.SELECTOR_META_PREFIX` so that it is clear if e.g. a underscore is added or not.

lenhard:
@tobiasdiez I addressed most of your comments. I agree that "content selectors" is probably not the most appropriate name for the stuff this feature does, but it is the name under which it has existed for a long time and is known, at least by some people. I would say that consistency in naming tops appropriateness of the name.

Furthermore, I removed the entry editors update code, as you asked for. The updating of the content selectors did not even work with this code. I tried to replace it with something else and failed. So, for now you have to restart JabRef to get a change in the content selectors to work. Maybe someone from @JabRef/developers can provide a hint on how the refresh of the entry editor should work?

Finally, I will leave the problems in the localization for @Siedlerchr to fix :)

matthiasgeiger:
Closing and reopening of the entry editor does not work? (Potentially with selecting another entry?)

lenhard:
@matthiasgeiger Unfortunately not. The changed data makes it into the MetaData, but the GUI does not notice. It is only up-to-date when the MetaData is re-read from disk and the EntryEditor is re-built. Closing and reopening the bib file does the trick, or restarting JabRef

lenhard:
So we would like to get this into v3.8, which hopefully can be released in december. What is missing:
 - [ ] correct the language files
 - [ ] have selector keywords updated without reloading of the bib file
 - [ ] Have the users play around with it to see if it really works

I was hoping that @Siedlerchr could do the first two points, and hopefully soon :-) @Siedlerchr what is your opinion on that?

Siedlerchr:
Yeah, I will try to work on it. At least I tried to fix the language files, yesterday. 
I will continue working on it the weekend

src/main/java/net/sf/jabref/gui/contentselector/FieldContentSelector.java
```diff
@@ -161,19 +160,18 @@ private void selectionMade() {
...
-        // TODO: could improve checking as not do add the same item twice
-        if (!"".equals(editor.getText())) {
+        // not the first word and no duplicate -> we need a comma
+        if (!"".equals(currentText) && !currentText.contains(chosen)) {
             editor.append(FieldContentSelector.this.delimiter);
         }
 
-        editor.append(chosen);
+        // no duplicate -> add it
+        if (!currentText.contains(chosen)) {
```
tobiasdiez: I think it is better to reuse the KeywordList class here since keyword duplication is not as easy as it appears on first sight. For example, if the field contains "some test, something else" and I would like to add "test", then the simple "contains" solution marks this as duplicate although "same test" is not the same as "test" 😼.

PR #2303, opened 2016-11-21
Update citeproc dependency to 1.0.1

matthiasgeiger:
see https://github.com/michel-kraemer/citeproc-java/releases

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -122,7 +122,7 @@ dependencies {
     // need to use snapshots as the stable version is from 2013 and doesn't support v1.0.1 CitationStyles
     compile 'org.citationstyles:styles:1.0.1-SNAPSHOT'
     compile 'org.citationstyles:locales:1.0.1-SNAPSHOT'
-    compile 'de.undercouch:citeproc-java:1.0.0-SNAPSHOT'
+    compile 'de.undercouch:citeproc-java:1.0.1'
 
     testCompile 'junit:junit:4.12'
     testCompile 'org.mockito:mockito-core:2.2.15'
--- a/src/test/java/net/sf/jabref/logic/citationstyle/CitationStyleGeneratorTest.java
+++ b/src/test/java/net/sf/jabref/logic/citationstyle/CitationStyleGeneratorTest.java
@@ -11,16 +11,15 @@ import static org.junit.Assert.assertEquals;
 public class CitationStyleGeneratorTest {
 
     @Test
-    public void testCarriageReturn() {
+    public void testIgnoreCarriageReturn() {
         BibEntry entry = new BibEntry();
-        entry.setField(FieldName.AUTHOR, "Doe, John and\rDoe, Jane");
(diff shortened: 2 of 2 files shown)


matthiasgeiger:
Test is failing: 
```
net.sf.jabref.logic.citationstyle.CitationStyleGeneratorTest > testCarriageReturn FAILED

    org.junit.ComparisonFailure: expected:<...ight-inline">J. Doe [and J. ]Doe, .</div>

      </div...> but was:<...ight-inline">J. Doe [John and]Doe, .</div>

      </div...>

        at org.junit.Assert.assertEquals(Assert.java:115)

        at org.junit.Assert.assertEquals(Assert.java:144)

        at net.sf.jabref.logic.citationstyle.CitationStyleGeneratorTest.testCarriageReturn(CitationStyleGeneratorTest.java:24)
```

As I'm not an expert for the CSL styles... is this an issue or just an irrelevant change? @bartsch-dev @koppor

Siedlerchr:
We should update asap, because version 1.0.0 is no longer avaiable at jcenter/maven repo:
https://bintray.com/bintray/jcenter/de.undercouch%3Aciteproc-java

koppor:
I have no clue. Just @Ignore

bartsch-dev:
The new version just removes the carriage return (`\r`) which is bad as this possibly connects two lines.

matthiasgeiger:
Thanks for the explanation. I already adapted the test to be no longer failing. 
Let's keep an eye on the discussion -  but I'll merge this anyway as currently all builds are failing due to a missing citeproc dependency.

koppor:
Side note: Maven Central still offers version 1.0.0 - https://mvnrepository.com/artifact/de.undercouch/citeproc-java/1.0.0

I don't get, why it is removed on bintray (https://bintray.com/bintray/jcenter/de.undercouch%3Aciteproc-java)

@michel-kraemer Could you give us a hint why version 1.0.0 was removed on bintray?

matthiasgeiger:
It was the '-SNAPSHOT' Version we used and which has been removed after the official 1.0.0 version

michel-kraemer:
Hey guys. I see two problems here.

a) Version 1.0.0 is not available on Bintray. That's very strange. I only upload to Maven Central and Bintray should sync automatically. It must be a problem with Bintray actually, but if you like I can try to publish 1.0.0 to Bintray too.

b)`\r` is replaced by nothing and not a space. I'm sorry if this causes problems. It seems there are files out there containing only a `\r` without a `\n`? If you can confirm this in the other thread I can release a fix.

michel-kraemer:
Regarding b): I fixed it here https://github.com/michel-kraemer/citeproc-java/commit/dc800d17d8bdab0817e5ef624843a40fb6d1a116

If you wish I can release 1.0.2 shortly.

koppor:
Thank you for your feedback. I think, an update is not necessary: See https://github.com/JabRef/jabref/pull/2306

PR #2304, opened 2016-11-21
Replace usage of Threads and priorities with thread pool

lenhard:
The current priority-based thread handling in JabRef is bad for two reasons:
 - Using and managing threads directly is bad practice. Since Java 1.6, there is `ExecutorServices` for this.
 - Relying on thread priorities is highly JVM-dependent and not portable. It is quite pointless for us to do this.

This PR replaces the manual thread-handling with a second thread pool for all threads that were low-priority before. The main (only) difference to other JabRef threads is that low-priority threads sometimes run forever and need to be interrupted when JabRef should shut down. The new structure keeps this.


src/main/java/net/sf/jabref/Globals.java
```diff
@@ -61,8 +61,7 @@ public static void startBackgroundTasks() {
         Globals.streamEavesdropper = StreamEavesdropper.eavesdropOnSystem();
 
         Globals.fileUpdateMonitor = new FileUpdateMonitor();
-        JabRefExecutorService.INSTANCE.executeWithLowPriorityInOwnThread(Globals.fileUpdateMonitor,
-                "FileUpdateMonitor");
+        JabRefExecutorService.INSTANCE.executeInterruptableTask(Globals.fileUpdateMonitor);
```
simonharrer: I really liked the fact that a single thread was running in the background and I could find this thread when looking at the thread dump during a debug session. Maybe use a single threaded executor service just for this file update monitor?

lenhard: This shouldn't be a problem. Maybe it makes sense for the specific threads, which are very prominent. Using an executor makes lifecycle management a little nicer at least.

I have to look into the remote thread as well, I guess it makes sense for that one, too.

lenhard:
So, what started out as a simple improvement ran into some issues. This is all due to the, in my point of view, quite unfortunate implementation of the remote functionality. This is heavily tied to being managed in a thread specifically and the lifecycle heavily depends on that. I was not successful in refactoring it towards a runnable within a reasonable amount of time and reverted that part of the PR.

The rest is no problem: Also the former low-priority threads are now managed in a thread pool as opposed to a custom thread list and in case someone ever wants to do a manual prioritization it would be a one-liner to add it back. Moreover, threads can be named and are named as before. The remote thread is kept as is and also managed by `JabRefExecutorService`, but with a little stricter requirements: There will never be more than one thread of this type in a JabRef instance, which was not ensured before.

I have installed the compiled version from builds.jabref, played around a little, and everything seems fine. I would appreciate it if someone else would also do some testing (no visible error should happen and JabRef should terminate in an orderly fashion) and then this is good to go from my side.

src/main/java/net/sf/jabref/JabRefExecutorService.java
```diff
@@ -64,80 +71,83 @@ public void executeAndWait(Runnable command) {
...
+    }
+
+    class NamedRunnable implements Runnable {
 
-        private final Runnable runnable;
-        private final ConcurrentLinkedQueue<Thread> startedThreads;
+        private String name;
 
-        public Thread thread;
+        private Runnable task;
```
simonharrer: final für beide Felder.

simonharrer:
I do remember the pain with RemoteThread. Hm, would something like this work? https://techblog.bozho.net/interrupting-executor-tasks/

lenhard:
I would just leave the RemoteThread as it is in this PR now. The problem is that the whole remote functionality is dependent on this being a RemoteThread and it also depends on various lifecycle methods of the Thread class itself (calls for Thread.State, etc.). This just won't work as soon as you use an ExecutorService and changing it would mean rewriting the remote functionality. Purely for a refactoring, this is not worth it imho.

src/main/java/net/sf/jabref/JabRefExecutorService.java
lenhard: Will do!

lenhard:
We checked this in the devcall and I will merge it now. The build breaks at the moment because of the BibLatexEntryTypes, but that is also the case for master, so it does not prevent merging.

PR #2307, opened 2016-11-21
Add complete entries with fetched infos

tobiasdiez:
Bibliographic information from web resources can now be used to complete existing entries. This functionality can be accessed via a new button in the entry editor.

I think also the IdBasedFetcher like DOI can implement the `EntryBased` interface and thus can be made accessible to this update function. It that case, `FetchAndMergeWorker` may be replaced by the new `EntryFetchAndMergeWorker`. Something for a new PR through.

image


src/main/java/net/sf/jabref/gui/entryeditor/EntryEditor.java
```diff
@@ -439,6 +443,27 @@ private void setupToolBar() {
...
+            fetcherPopup.add(new JMenuItem(new AbstractAction(fetcher.getName()) {
+
+                @Override
+                public void actionPerformed(ActionEvent e) {
+                    new EntryFetchAndMergeWorker(panel, getEntry(), fetcher).execute();
+                }
+            }));
+        }
+        JButton fetcherButton = new JButton(IconTheme.JabRefIcon.REFRESH.getIcon());
+        fetcherButton.setToolTipText(Localization.lang("Update with bibliographic information from the web."));
```
lenhard: Imho, there is no reason for adding the period in the end. JabRef is not consistent in this, some tool tips end with periods and some do not. In my point of view, punctuation is only needed if there is a block of text for the tool tip. If it is a single sentence, no punctuation looks better.

So please remove that one :)

src/main/java/net/sf/jabref/gui/mergeentries/MergeWithFetchedEntryAction.java
```diff
@@ -0,0 +1,36 @@
...
+
+    private final BasePanel basePanel;
+
+    public MergeWithFetchedEntryAction(BasePanel basePanel) {
+        this.basePanel = basePanel;
+    }
+
+    @Override
+    public void action() throws Throwable {
+        {
```
lenhard: What is the reason for this additional level of indentation?

onewiki:
Sorry I don't know if this is the right place to ask my question, that is, whether it is possible to fetch the abstract when clicking "Get Bibtex data from DOI" (along with other data Author/Journal, etc.). Thanks!

matthiasgeiger:
@onewiki This is not possible, as the information for DOIs are resolved using [crossref.org](https://www.crossref.org). And crossref.org does not provide abstracts, see this example: http://search.crossref.org/?q=10.2331%2Fsuisan.32.804

onewiki:
@matthiasgeiger Thanks for explaining.

onewiki:
@matthiasgeiger Sorry to bother you with another question (since you are knowledgeable on DOI etc.): do all articles (in social science) have DOIs, or only the ones published after year 2000?

Siedlerchr:
@matthiasgeiger  This is not really right: CrossRef does provide a way to get the abstract, however, this is only possible using the XML version, see the comment:
https://github.com/CrossRef/rest-api-doc/issues/98#issuecomment-227132063

I recently checked that, because of a feature request in the forum:
http://discourse.jabref.org/t/export-abstract-from-text/324/2

@onewiki  It depends which article a DOI has. I recently had one article from 1987 in the area of computer science which had a DOI assigned. As Publishers have to pay a fee for using a DOI, it might be that not all are willing to pay it, especially OpenAcces models.

matthiasgeiger:
Generally not all academic works have DOIs, e.g., Technical Reports published at my university only have a ISSN for the series and a "URN" as an identifier. Also workshop and conference publications exist without DOIs. *Most* journal articles however have DOIs - but there are also journals not using DOIs. 

In my main database there are some articles published prior 2000 which have a DOI, e.g.: http://doi.org/10.1145/69605.2085 - but most older articles don't have DOIs.

onewiki:
@Siedlerchr 
Thanks, but I am a novice; could you please explain a little more what this means? "If you have a DOI use the API but append ".xml" to the end." How to do that?

Siedlerchr:
@onewiki  The hint with the xml was for @matthiasgeiger. It's an implementation detail for calling the  API service from JabRef, you can't  do anything with.

onewiki:
@Siedlerchr Are you saying I have to wait for the Jabref team to incorporate this feature into the program (so that it will be available in a future version)?

src/main/java/net/sf/jabref/gui/entryeditor/EntryEditor.java
koppor: As this is a tooltip, the final dot should really be removed.

lenhard:
@tobiasdiez Please fix the few very minor things mentioned above, merge the language files and then this is ready to go into master!

PR #2310, opened 2016-11-23
Replace usages of Throwable with Exception

lenhard:
Reviewing #2307 I noticed that `BaseAction` and `AbstractWorker` use `Throwable` in their method signature. I cannot find a reason for linking to the highest possible error type here. After all, this has nothing to do with JVM errors or the like. 

I moved the type one step down in the hierarchy to `Exception` which is still basic enough to cover every possible.


Siedlerchr:
LGTM!

Issue #2273, opened 2016-11-15
export via command line fails


WastlM:
JabRef versions 3.3 - 3.7 on Ubuntu 14.04 and 16.04.

Steps to reproduce:

1. Download JabRef's jar file for any of the releases 3.3 to 3.7
2. Create or find a bib file, `references.bib`, say (an example is given below)
3. Try `java -jar JabRef-3.7.jar -n -o references.html,tablerefsabsbib references.bib` (other export formats do not work either)

In current JabRef 3.7, no html file is produced and the output is:
```
09:57:04.556 [AWT-EventQueue-0] INFO  net.sf.jabref.logic.importer.OpenDatabase - Opening: references.bib
Exporting: references.html
Could not export file 'references.html': null
```
**Note:** exporting bibfiles via the GUI works flawlessly.

In JabRef 3.2, the html file is produced correctly and the output is:
```
Nov 15, 2016 9:51:39 AM net.sf.jabref.JabRef openBibFile
INFO: Opening: references.bib
Exporting: references.html
Nov 15, 2016 9:51:39 AM net.sf.jabref.exporter.ExportFormat performExport
WARNING: []
```

I used the following simple `references.bib`:
```
@Book{key,
  Title = {Title},
  Author = {An Author},
  Publisher = {A Publisher},
  Year = {2016},
  Address = {Address}
}
```

har-wradim:
I can confirm the bug for v. 3.8 (Ubuntu 14.04), it is IMO a very serious problem.

Each one of the exporters listed in `jabref -h`:

    for format in MSBib bibordf bibtexml din1505 docbook endnote harvard html iso690rtf iso690txt listrefs misq mods ods oocalc oocsv ris simplehtml tablerefs tablerefsabsbib; do
    	jabref -n -o references.$format,$format references.bib
    done

fails with:

    22:19:31.879 [AWT-EventQueue-0] INFO  net.sf.jabref.logic.importer.OpenDatabase - Opening: references.bib
    Exporting: references.<format>
    Could not export file 'references.<format>': null

Siedlerchr:
Confirmed on Windows 7 x64, too.
 I will look into that.

Siedlerchr:
I could boil it down to a NPE in ExportFormat 

```
java.lang.NullPointerException: Entries must not be null!
	at java.util.Objects.requireNonNull(Unknown Source)
	at net.sf.jabref.logic.exporter.ExportFormat.performExport(ExportFormat.java:198)
[...]

PR #2316, opened 2016-11-28
Fix exporting via commandline in no gui mode

Siedlerchr:
Fix for #2273
Adjusted javadoc


CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - We fixed an issue when JabRef restores its session and a shared database was used: The error message "No suitable driver found" will not appear.
 - Update check now correctly notifies about new release if development version is used. [#2298](https://github.com/JabRef/jabref/issues/2298)
 - Fixed [#2311](https://github.com/JabRef/jabref/issues/2311): The DBLP fetcher has been rewritten and is working again.
+- Fixed [#2273](https://github.com/JabRef/jabref/issues/2273): Export via commandline in no-gui modus is now working again.
```
tobiasdiez: modus -> mode

src/main/java/net/sf/jabref/cli/ArgumentProcessor.java
```diff
@@ -208,7 +211,7 @@ private boolean exportMatches(List<ParserResult> loaded) {
                             matches);
                 } catch (Exception ex) {
                     System.err.println(Localization.lang("Could not export file") + " '" + data[1] + "': "
-                            + ex.getMessage());
+                            + ExceptionUtils.getStackTrace(ex));
```
tobiasdiez: Did you changed this on purpose? All the other error reporting in this class seems to display the message instead of the stack trace (not sure what is better, but it should be consistent).

src/main/java/net/sf/jabref/logic/exporter/ExportFormat.java
```diff
@@ -179,23 +179,12 @@ private Reader getReader(String filename) throws IOException {
...
-     * @throws Exception   if any other error occurred during export.
-     * @see net.sf.jabref.logic.exporter.IExportFormat#performExport(BibDatabaseContext, String, Charset, List)
-     */
+
     @Override
     public void performExport(final BibDatabaseContext databaseContext, final String file,
             final Charset encoding, List<BibEntry> entries) throws Exception {
-        Objects.requireNonNull(databaseContext);
-        Objects.requireNonNull(entries);
+        Objects.requireNonNull(databaseContext, "Database Context must not be null!");
```
tobiasdiez: To be honest, in my opinion these string messages don't add more relevant information.

src/main/java/net/sf/jabref/cli/ArgumentProcessor.java
Siedlerchr: Yes, because if you look into the issue, ex.getMessage simply returned "null" and there was no stacktrace visible, which made it hard to locate the bug.
`Could not export file 'references.html': null`

Siedlerchr: I looked at all other cases and there is either a logger attached or a specific Exception, e.g. a SaveException/JabRef Exception catched and the message printed, which is fine enough.
Only this two line were problematic, as the getMessage returned null

Siedlerchr:
Included feedback

Issue #1879, opened 2016-08-29
Installer Code Signing


stefan-kolb:

stefan-kolb:
@koppor Any news?

stefan-kolb:
We are just targeting Windows code signing with this certificate? Install4j says we need a Developer ID Application something from Apple to sign the code...

stefan-kolb:
This certificate will only work for Windows? As we don't have an Apple code signing certificate?!

stefan-kolb:
https://docs.travis-ci.com/user/encrypting-files/

koppor:
Note that only one encryption key can be stored per time: https://docs.travis-ci.com/user/encrypting-files/#Encrypting-multiple-files

stefan-kolb:
```
 travis login --org --github-token $ && travis encrypt-file jabref-2016.p12 --add
```

stefan-kolb:
http://stackoverflow.com/questions/6654543/how-to-create-pfx-with-my-chain-of-certificates

stefan-kolb:
http://charles-blog.appspot.com/2012/11/Export-Private-Key-from-pkcs12-file

koppor:
Thanks to @stefan-kolb, a test build is available at https://builds.jabref.org/cert/

matthiasgeiger:
Is working only partially...:

![grafik](https://cloud.githubusercontent.com/assets/676652/20681115/62d9c8d8-b5a2-11e6-97ec-3dc3c838a016.png)

The UAC dialog shown after clicking "Trotzdem ausführen" looks good (except that only `Oliver Kopp, Open Source Developer` is mentioned but not directly JabRef 😕

koppor:
Mentioning me instead of JabRef is the condition of the [Certum Open Source Code Signing](https://www.certum.eu/certum/cert,offer_en_open_source_cs.xml).

Quoting from their conditions, I received via Email:

> OPEN SOURCE CODE SIGNING CERTIFICATES MAY BE ISSUED ONLY TO PRIVATE INDIVIDUALS !

> When issuing a certificate the "Open Source Developer"  term is automatically added to the "CN" (Common Name) field. This is a necessary condition to obtain a certificate.

stefan-kolb:
Yeah, I don't see why this means it is only partially working. The name is intentional.

PR #2320, opened 2016-11-29
Installer Code Signing #1879

stefan-kolb:
- [ ] Win
- [ ] OSX


koppor:
LGTM

[LGTM](https://lgtm.in/i/Mur1n32nq)

stefan-kolb:
Can someone please try this on OSX @chochreiner. I also signed the dmg but I'm not sure that this works with the current certificate (even quite sure that it doesn't).

behrinml:
Standard response (double click):
screen shot 2016-11-29 at 16 53 18

By right click -> open:
screen shot 2016-11-29 at 16 53 31

With the signed version:
screen shot 2016-11-29 at 16 59 42

So it seems to work but gives of course the standard warning for Applications from outside the App Store.

chochreiner:
hmm
for me, there is no difference between the app from the master and the cert branch
Which certificate are you currently using?

behrinml:
Which settings do you have for installing applications? For me it is set to "Allow apps downloaded from Mac App Store and identified developers".

I think if you use one of the other two settings there will be no differences between the builds...

chochreiner:
For me its: "App Store and identified developers" (on MacOS Sierra)

behrinml:
Ok.. Now I am confused. Just checked on MacOS Sierra from a colleague and there is no difference between the builds. Maybe it is something about the OS (I am using OS X El Capitan) or I have changed something else which interferes, f.e. disabled SIP to get Spaces back, but this shouldn't make a difference. I can check next week on OS X Mavericks or later today on Mac OS X Snow Leopard.. but I think the latter has no signing...

stefan-kolb:
Maybe we need an extra certificate issued or registered with the App Store for MacOSX? Needs to be investigated later. For now it works for Windows at least.

Issue #1445, opened 2016-05-24
Integrity check for possibly wrong DOIs


stefan-kolb:
- [x] syntactically wrong DOIs
- [ ] semantically wrong DOIs false assignment, not the correct DOI (has influence on PDF retrieval!)

simonharrer:
:+1: 

Could be a performance issue. 

Maybe we should make the integrity check asynchronous based on events of database changes, so that all integrity violations are always kept in sync eventually. But this would require a lot of code.

stefan-kolb:
A maybe I meant Cleanup not integrity check?!

simonharrer:
@stefan-kolb is the solution of the integrity check sufficient for you? Can this be closed then?

stefan-kolb:
Uhm, well this was not meant to be a check for syntactically wrong DOIs but more for DOIs that do not fit the current entry, i.e., not the correct DOI for an entry.

matthiasgeiger:
> No description provided.

There's a reason for this description field. 😜

simonharrer:
@stefan-kolb ah ok. But this would require to change the architecture of the integrity checks. At the moment, the integrity checks are fast as they use only local knowledge. If you want to use global knowledge from the internet, this would introduce considerable delays. Not sure if this is feasible without investing a lot of time.

stefan-kolb:
Yes, we already discussed this before. Probably we should implement such a check but only call it when a full text PDF is queried or the BibTeX auto completion is called. If this check fails we should alert the user that the DOI is possibly wrong.

stefan-kolb:
I thought about several ideas how to validate if the DOI is the correct DOI for an entry, but I think we just have to accept the DOI right now if there is one given. Algorithms to check if it is the right DOI for the entry may also fail quite often so we have to trust the user and he has to find out by himself that he put the wrong DOI. Another option would be to put a manual validate DOI functionality with warnings but I think this is not useful enough right now to be justified.

PR #2327, opened 2016-12-01
Add DOI integrity check

simonharrer:
Refs #1445

Adds an integrity check to determine whether we can handle the content in the DOI field of a bib entry.


Siedlerchr:
I have some valid DOIs, which look a bit weird, maybe  you could add a test for them:

`10.17487/rfc1436`
`10.1002/(SICI)1097-4571(199205)43:4<284::AID-ASI3>3.0.CO;2-0`

matthiasgeiger:
I don't think the additional tests proposed by @Siedlerchr are necessary here. They are more tests of the DOI implementation and not tests of the integrity checker. (the second DOI is even already a testcase in https://github.com/JabRef/jabref/blob/master/src/test/java/net/sf/jabref/logic/util/DOITest.java).

Apart from that: LGTM

Issue #2206, opened 2016-10-27
Integrity check to report BibTex keys that do not fit the expected scheme


stefan-kolb:
Example: 

Custom types (or even other types?) might miss the author field which leads to a not well-formed BibTeX key if no exception schema is defined.
We should probably warn in case something like this happens or at least provide an integrity check for this?

```
-@Standard{iso2011,
+@Standard{2011,
   title       = {{Systems and software engineering -- System and software Quality Requirements and Evaluation (SQuaRE) -- System and software quality models}},
   institution = {ISO/IEC},
   number      = {25010},
   year        = {2011},
   owner       = {Administrator},
@@ -563,11 +543,11 @@
   publisher = {Springer},
   year      = {2013},
   doi       = {10.1007/978-3-642-36981-0_75},
 }
```

simonharrer:
I think a check to determine that the current bibtex key deviates from the one that would be generated is also helpful to reveal the entries which have been imported but deviate from the set generation scheme. 

Of course, in a save action, the key could always be generated, but I am not sure if this is advisable.

matthiasgeiger:
Implemented in #2328.

PR #2328, opened 2016-12-01
Add bibtex key deviation check

simonharrer:
Refs #2206 

Adds integrity check to determine all entries which bibtex keys deviate from the set bibtex key generation pattern.

### DIFF
--- a/src/main/java/net/sf/jabref/logic/integrity/BibtexkeyDeviationChecker.java
+++ b/src/main/java/net/sf/jabref/logic/integrity/BibtexkeyDeviationChecker.java
@@ -32,7 +32,7 @@ public class BibtexkeyDeviationChecker implements Checker {
         String key = valuekey.get();
 
         // generate new key
-        String generatedKey = BibtexKeyPatternUtil.generateNewCityKey(bibDatabaseContext, entry, bibtexKeyPatternPreferences);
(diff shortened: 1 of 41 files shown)


src/main/java/net/sf/jabref/logic/integrity/BibtexkeyDeviationChecker.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.Optional;
+
+import net.sf.jabref.logic.bibtexkeypattern.BibtexKeyPatternPreferences;
+import net.sf.jabref.logic.bibtexkeypattern.BibtexKeyPatternUtil;
+import net.sf.jabref.logic.integrity.IntegrityCheck.Checker;
+import net.sf.jabref.logic.l10n.Localization;
+import net.sf.jabref.model.database.BibDatabaseContext;
+import net.sf.jabref.model.entry.BibEntry;
+
+public class BibtexkeyDeviationChecker implements Checker {
```
stefan-kolb: maybe rename to BibtexKeyDeviationChecker?

simonharrer: There is another class which is called BibtexkeyChecker, so I used the naming convention from this other Checker class.

PR #2332, opened 2016-12-02
Add initial .editorconfig

koppor:
This adds an initial [editorconfig](http://editorconfig.org/) to ensure that all files are encoded in UTF-8 and end with a new line.


Siedlerchr:
Looks good, but at Least eclipse needs a plugin for recognizing that. 
You should document that somewhere

PR #2337, opened 2016-12-03
Add intellij code style

tobiasdiez:
Contains:
- Import style according to wiki.
- Automatic line breaks

### DIFF
--- a/config/IntelliJ Code Style.xml
+++ b/config/IntelliJ Code Style.xml
@@ -1,0 +1,23 @@
+<code_scheme name="Project">
+  <option name="CLASS_COUNT_TO_USE_IMPORT_ON_DEMAND" value="999" />
+  <option name="NAMES_COUNT_TO_USE_IMPORT_ON_DEMAND" value="999" />
+  <option name="PACKAGES_TO_USE_IMPORT_ON_DEMAND">
+    <value />
+  </option>
+  <option name="IMPORT_LAYOUT_TABLE">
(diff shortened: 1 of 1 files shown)


Siedlerchr:
Is t his compatible with the eclipse style?
And you are missing a newline at EOF

stefan-kolb:
Actually I don't like this automatic line breaks at all. Why not let the developer choose when to break the line? We are not in the 90s where desktop resolutions were so low that they had to restrict the line length for devs.

lenhard:
1 .The import style, as far as I can see, is set as we have done with checkstyle (so this is very much appreciated).
2. As stated by @stefan-kolb automatic line breaking should not be turned on.
3. I had a look at more intellij options. There is more that can be configured like in the Eclipse style, i.e.
  - Use spaces for tabs
  - a tab is 4 spaces of indent
  - Set spaces for all kind of constructs, i.e., after if, while, etc.
  - There are also many options to force braces for the aforementioned statements

I am not sure if many of these things are set by Intellij by default and you do not have to specify this explicitly, but please have a check and modify the configuration if needed.

koppor:
It the "spaces for all kinds of contructs" really in place? I saw

> if(preset.getFormatterCleanups().isEnabled()) 

in a commit after this merge.

tobiasdiez:
Yes it is on by default, but I think IntelliJ does not reformat the code on changes (only if I tell it to do so).

PR #2339, opened 2016-12-04
Fix biblatex cleanup for date field

### LINKED ISSUE #2335: Biblatex Cleanup for moving year+ month to Date field not working
Version 3.7, I think it dit not work in 3.6, too.
According to the code, the Biblatex Cleanup should cleanup/move the fields year + months etc to date

However, it currently gives always "No Entry needed a cleanup".
'Tested with different entrytypes and both year and month were set.

```
@Article{Orlowski;2011,
  author        = {Adam Czarnecki; Cezary Orłowski;},
  title         = {Application of Ontology In the ITIL Domain},
  year          = {2011},
  month         = jan,
}
```

https://github.com/JabRef/jabref/blob/master/src/main/java/net/sf/jabref/logic/cleanup/BiblatexCleanup.java

Siedlerchr:
This fixes #2335 
Moved biblatex Cleanup to new test


Siedlerchr:
Failing test is something from Medline, not related

src/main/java/net/sf/jabref/logic/cleanup/BiblatexCleanup.java
```diff
@@ -35,22 +35,26 @@
...
-                    changes.add(new FieldChange(entry, FieldName.MONTH, oldMonth.orElse(null), null));
-                });
-            }
-        });
+        // If there already exists a value for the field date, it is not overwritten
+        String date = "";
+        if (entry.getField(FieldName.DATE).isPresent()) {
+            date = entry.getField(FieldName.DATE).get();
+        }
+        if (date.isEmpty()) {
```
tobiasdiez: I think `if (StringUtils.isBlank (entry.getField(FieldName.DATE)))` is cleaner (not sure about the naming)

src/main/java/net/sf/jabref/logic/cleanup/BiblatexCleanup.java
```diff
@@ -35,22 +35,26 @@
...
+            entry.getFieldOrAlias(FieldName.DATE).ifPresent(newDate -> {
+                Optional<String> oldYear = entry.getField(FieldName.YEAR);
+                Optional<String> oldMonth = entry.getField(FieldName.MONTH);
+
+                entry.setField(FieldName.DATE, newDate);
```
tobiasdiez: If I remember correctly, then set/clearField already return the right FieldChange

src/test/java/net/sf/jabref/logic/cleanup/BiblatexCleanupTest.java
```diff
@@ -0,0 +1,78 @@
...
+public class BiblatexCleanupTest {
+
+    @Rule
+    public TemporaryFolder bibFolder = new TemporaryFolder();
+
+    private CleanupWorker worker;
+
+    @Before
+    public void setUp() throws IOException {
+
```
tobiasdiez: Remove empty line

src/test/java/net/sf/jabref/logic/cleanup/BiblatexCleanupTest.java
```diff
@@ -0,0 +1,78 @@
...
+        MetaData metaData = new MetaData();
+        BibDatabaseContext context = new BibDatabaseContext(new BibDatabase(), metaData, bibFolder.newFile("test.bib"));
```
tobiasdiez: Do we really need a create a file?

src/test/java/net/sf/jabref/logic/cleanup/BiblatexCleanupTest.java
```diff
@@ -0,0 +1,78 @@
...
+                        JabRefPreferences.getInstance().getFileDirectoryPreferences()));
+    }
+
+    @Test
+    public void convertToBiblatexMovesJournalToJournalTitle() {
+        CleanupPreset preset = new CleanupPreset(CleanupPreset.CleanupStep.CONVERT_TO_BIBLATEX);
+        BibEntry entry = new BibEntry();
+        entry.setField(FieldName.JOURNAL, "test");
+
+        worker.cleanup(preset, entry);
```
tobiasdiez: directly use the BiblatexCleanup.cleanup method instead of using the worker (makes tests and initialization more transparent)

src/test/java/net/sf/jabref/logic/cleanup/CleanupWorkerTest.java
```diff
@@ -308,17 +308,6 @@ public void cleanupLatexMergesTwoLatexMathEnvironments() {
     }
 
     @Test
-    public void convertToBiblatexMovesJournalToJournalTitle() {
```
tobiasdiez: I would leave this test here to make sure that the CleanupWorker works correctly wrt to the "Convert to BibLatex" option; the other test class for the BibLatexCleanupJob then only tests this class directly.

src/main/java/net/sf/jabref/logic/cleanup/BiblatexCleanup.java
Siedlerchr: That was part of the code I didn't touch

src/main/java/net/sf/jabref/logic/cleanup/BiblatexCleanup.java
Siedlerchr: I used this approach, because it captures 2 purposes and checks the optonal
a) If the field is not present the date is empty
b) if the field is present, we assign the value, however, it still might be empty

src/main/java/net/sf/jabref/logic/cleanup/BiblatexCleanup.java
```diff
@@ -25,33 +26,22 @@
             entry.getField(oldFieldName).ifPresent(oldValue -> {
                 if (!oldValue.isEmpty() && (!entry.getField(newFieldName).isPresent())) {
                     // There is content in the old field and no value in the new, so just copy
-                    entry.setField(newFieldName, oldValue);
-                    changes.add(new FieldChange(entry, newFieldName, null, oldValue));
-
-                    entry.clearField(oldFieldName);
-                    changes.add(new FieldChange(entry, oldFieldName, oldValue, null));
+                    changes.add(entry.setField(newFieldName, oldValue).orElse(null));
```
tobiasdiez: entry.setField(...).ifPresent(changes::add), otherwise you add null values to the list which is not so nice.

src/main/java/net/sf/jabref/logic/cleanup/BiblatexCleanup.java
```diff
@@ -25,33 +26,22 @@
...
         }
-
         // Dates: create date out of year and month, save it and delete old fields
         // If there already exists a value for the field date, it is not overwritten
         String date = "";
         if (entry.getField(FieldName.DATE).isPresent()) {
             date = entry.getField(FieldName.DATE).get();
         }
-        if (date.isEmpty()) {
+        if (StringUtils.isBlank(date)) {
```
tobiasdiez: isBlank also accepts an optional I think (if not, please add a corresponding overload), hence `Optional<String> date = entry.getField(FieldName.DATE)` should work.

Siedlerchr:
However  in  this  case  that  would  result  in  the  original  bug
 again, when checked for ifPresent !

If the field is not yet in the biblatex entry (which is obviously the case
when you only have year), it returns an EmtyOptional and thus you would
have the same situation as in the original code.
This is why I splitted it.




Am 05.12.2016 4:21 nachm. schrieb "Tobias Diez" <notifications@github.com>:

*@tobiasdiez* commented on this pull request.
------------------------------

In src/main/java/net/sf/jabref/logic/cleanup/BiblatexCleanup.java
<https://github.com/JabRef/jabref/pull/2339#pullrequestreview-11398588>:

>          // Dates: create date out of year and month, save it and delete old fields
         // If there already exists a value for the field date, it is
not overwritten
         String date = "";
         if (entry.getField(FieldName.DATE).isPresent()) {
             date = entry.getField(FieldName.DATE).get();
         }
-        if (date.isEmpty()) {
+        if (StringUtils.isBlank(date)) {

isBlank also accepts an optional I think (if not, please add a
corresponding overload), hence Optional<String> date =
entry.getField(FieldName.DATE) should work.

—
You are receiving this because you authored the thread.
<https://github.com/JabRef/jabref/pull/2339#pullrequestreview-11398588>,
or mute
the thread

Siedlerchr:
Failing Snap CI is due to HTTP error 502 in AstrophysicsDataSystemTest

koppor:
Could you please merge master again - I merged https://github.com/JabRef/jabref/pull/2333 and thus SnapCI only does the basic tests as described at https://github.com/JabRef/jabref/wiki/CI#snapci 😇

tobiasdiez:
@Siedlerchr: sorry I don't understand you comment.
Let go through this case by case with the following code:
````
Optional<String> date = entry.getField(FieldName.DATE);
Boolean isBlank = StringUtils.isNotBlank(date);
````

case | date | isBlank |
---- | --- | --- |
no date at all | Optional.empty | true
empty string | Optional.of("") | true
something else | Optional.of("something") | false

You want to move year and month to the date field exactly in the first two cases, i.e., if and only if `isBlank == true`. What do I miss?

**Comment by Siedlerchr** [comment]
Ah now I understand your comment, the first case was not obvious for me, that an empty should return true.

**Comment by Siedlerchr** [comment]
@tobiasdiez The confusion came from the naming stuff. I previously had the apache.commons StringUtil class and not the jabref specific one. Changed that.

File: src/test/java/net/sf/jabref/logic/cleanup/BiblatexCleanupTest.java [hunk 9]
```diff
@@ -0,0 +1,50 @@
...
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+
+public class BiblatexCleanupTest {
+
+    @Rule
+    public TemporaryFolder bibFolder = new TemporaryFolder();
```
koppor: Is this variable ever used?

koppor:
Besides the micro comment: LGTM :+1:

src/test/java/net/sf/jabref/logic/cleanup/CleanupWorkerTest.java
```diff
@@ -307,11 +308,11 @@ public void cleanupLatexMergesTwoLatexMathEnvironments() {
     public void convertToBiblatexMovesJournalToJournalTitle() {
         CleanupPreset preset = new CleanupPreset(CleanupPreset.CleanupStep.CONVERT_TO_BIBLATEX);
         BibEntry entry = new BibEntry();
-        entry.setField("journal", "test");
+        entry.setField(FieldName.JOURNAL, "test");
```
tobiasdiez: Please revert these changes, as hard-coded strings in tests are actually desirable. The logic for the conversation moves FieldName.JOURNAL to FieldName,JOURNALTITLE. Hence if somebody changes `FieldName.JOURNAL = "stupid"` the test still passes but the functionality is broken.

Issue #1323, opened 2016-04-29
"Look up full text document" shortcut; download all


vogler:
"Quality > Look up full text document" is great, but what if I want to do this for all entries?
It has no shortcut and it is not in the list of key bindings.
Also, if I have multiple entries selected, it only processes the first one.

Version 3.3

Siedlerchr:
I fully agree with you, that would be a great enhancement :+1:  .  Stumbled across this a couple of days ago, too.

stefan-kolb:
Good enhancement indeed! Although we need a completely new workflow for it.
My suggestion:

Easy:
- progress bar downloading all stuff in the background, auto renaming and linking files, just a message dialog how many files could be linked.

Downsides: A lot of wrong downloads could be there.

Advanced:
- Big window like `merge dialog`
- On the left list of all entries (key, author, title)
- On the right possibility for a larger PDF preview to judge if the download is correct
- Download is triggered in the background
- Some options to correct/delete wrong files

envhyf:
I have the same issue. Is there any alternative methods to achieve that?(I mean downloading full text for all the entries I selected)

stefan-kolb:
Not right now, but we are working on it :+1:

stefan-kolb:
@oscargus This was implemented by you a few weeks ago right? Inside the main menu `quality - lookup full text document`? If so, we should rename the description to `Retrieve full text documents for the whole library` or anything alike.

stefan-kolb:
@vogler @envhyf You can find the functionality under `quality - lookup full text documents`. Just select multiple documents (but not to many 😄 )

koppor:
Follow up at #2344 

There is no shortcut. Maybe <kbd>Ctrl</kbd> + <kbd>F7</kbd> as <kbd>F7</kbd> is "Automatically set file links" and <kbd>Shift</kbd> + <kbd>F7</kbd> is "Find unlinked files"?

PR #2027, opened 2016-09-21
Disable context menu action on selection of multiple entries

motokito:
If multiple entries are selected, some  context menu, toolbar and menu actions are now disabled.
Disabled actions:
- Print entry preview
- Send as email
- Attach file
- Edit entry
- Open folder (enable only, if file exist)
- Open file (enable only, if file exist)
- Open URL or DOI (enable only, if url or doi exist)
- Merge entries (enable only on select of two entries)

These actions make no sense to be used when there are multiple entries selected.
For further information have a look at #463 

Context menu before change:

2016-09-22_15h50_28

Context menu after change:

2016-09-22_15h50_45

Edit menu before change:

2016-09-22_15h51_28

Edit menu after change:

2016-09-22_15h51_55

Bibtex menu before change:

2016-09-22_15h52_21

Bibtex menu after change:

2016-09-22_15h52_44

Quality menu before change:

2016-09-22_16h34_33

Quality menu after change:

2016-09-22_16h34_52

Tools menu before change:

2016-09-22_16h35_17

Tools menu after change:

2016-09-22_18h08_02

Toolbare button "Edit Entry" before change:

2016-09-22_15h53_17

Toolbare button "Edit Entry" after change:

2016-09-22_15h53_44

Toolbare button "Merge entries" before change:

2016-09-22_16h38_10

Toolbare button "Merge entries" after change:

2016-09-22_16h37_00

Please take a look for merge. THX 😄 

- [x] Change in CHANGELOG.md described
- [x] Screenshots added (for bigger UI changes)
- [x] Manually tested changed features in running JabRef
- [ ] Internal QS

PR #2344, opened 2016-12-07
Follow up for #1323 Batch Fulltext downloader

stefan-kolb:
- 	Rename menu item to plural for finding full texts
- 	Disable menu item when there are no documents selected for fulltext retrieval
- 	Fix layout of file dialog and increase size of file path significantly
old:
image

new:
image


src/main/java/net/sf/jabref/gui/externalfiles/FindFullTextAction.java
```diff
@@ -41,12 +41,26 @@ public FindFullTextAction(BasePanel basePanel) {
 
     @Override
     public void init() throws Exception {
+        if (basePanel.getSelectedEntries().size() == 0) {
```
matthiasgeiger: Wouldn't be a better approach to disable the menu item (as it is already done for various other actions)?

If selection size() is == 0 at this point you can just throw an exception.

koppor: Refs #2027

stefan-kolb: Not 100 % sure about this, but done.

src/main/java/net/sf/jabref/gui/JabRefFrame.java
```diff
@@ -1591,6 +1594,9 @@ public void updateEnabledState() {
 
             boolean twoEntriesSelected = current.getSelectedEntries().size() == 2;
             setEnabled(twoEntriesOnlyActions, twoEntriesSelected);
+
+            boolean oneOrMoreEntriesSelected = current.getSelectedEntries().size() >= 1;
```
koppor: Why not checking for `!isEmpy()`?

src/main/java/net/sf/jabref/gui/externalfiles/FindFullTextAction.java
```diff
@@ -41,26 +41,14 @@ public FindFullTextAction(BasePanel basePanel) {
...
     @Override
     public void init() throws Exception {
-        if (basePanel.getSelectedEntries().size() == 0) {
-            JOptionPane.showMessageDialog(
-                    basePanel,
-                    Localization.lang("Select at least one entry to perform this operation"),
-                    Localization.lang("No entry selected"),
-                    JOptionPane.ERROR_MESSAGE);
-            return;
+        if (basePanel.getSelectedEntries().size() > 0) {
```
koppor: `!isEmpty()`, maybe add a logger.DEBUG statement?

PR #2348, opened 2016-12-08
Fix #1446 Only add non empty paths

### LINKED ISSUE #1446: Better default for saving files
If the preferences are reset, and I want to download a file, JabRef suggests that I store it under `C:\`. I would prefer to use a more sophisticated location, e.g., relative to the current bib file perhaps. As normally, one does not have access rights to save directly on `C:\`, wwhich can leave to strange behavior.

stefan-kolb:
Suggest current BIB folder when no defaults are set.

### DIFF
--- a/src/main/java/net/sf/jabref/model/database/BibDatabaseContext.java
+++ b/src/main/java/net/sf/jabref/model/database/BibDatabaseContext.java
@@ -176,6 +176,7 @@ public class BibDatabaseContext {
     * @return The default directory for this field type.
     */
     public List<String> getFileDirectories(String fieldName, FileDirectoryPreferences preferences) {
+        // TODO: all of these folder will be also searched for local PDFs. If this includes large folders like user.home this might take forever.
         List<String> fileDirs = new ArrayList<>();
 
         // 1. metadata user-specific directory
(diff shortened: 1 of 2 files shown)


src/main/java/net/sf/jabref/model/database/BibDatabaseContext.java
```diff
@@ -176,6 +176,7 @@ public boolean isBiblatexMode() {
     * @return The default directory for this field type.
     */
     public List<String> getFileDirectories(String fieldName, FileDirectoryPreferences preferences) {
+        // TODO: all of these folder will be also searched for local PDFs. If this includes large folders like user.home this might take forever.
```
stefan-kolb: Yes but it uses this functionality. The reason is, I added the user.dir as a 5th option to avoid the error message when a DB is not saved yet and the preferences are resetted (see #1446) but this lead to a "endless" looping of the find unlinked files because of the amount of files that typically reside inside the user.dir. This is more or less a hint what changes here can currently cause. I understand that this is not totally intuitive to anyone right now.

stefan-kolb: Anyways, feel free to just remove the comment via Github.

Siedlerchr: The findAssociated files gets a list of the dirs but uses a lazy stream (Files.find) to populate a HashSet.
Should not be that performance problem.
However, I see the findAssociatedFiles method could be improved to directly search on the stream. But this is time for a new PR

PR #2349, opened 2016-12-08
Integrity checker improvements

stefan-kolb:
Follow up for #1779

Linking to Bibentries inside the integrity checker should be done via a reference to the entry not the key.
- [x] Focus by Id rather than Bibtexkey
- [x] Also the field Bibtexkey can now be focused. I changed the internal list from unmodifiable and returned another unmodifiable list. This needs some close examination in my opinion as I'm not 100% sure if I introduced any side effects. Please also test the feature if it is still working as expected.


tobiasdiez:
I did a quick check and it appears to work.

matthiasgeiger:
Uhh... Browsing this from my mobile I'm not Sure... But @tobiasdiez you have closed the PR and not merged it 😉

src/main/java/net/sf/jabref/model/database/BibDatabase.java
```diff
@@ -152,6 +151,21 @@ public boolean containsEntryWithId(String id) {
...
     /**
+     * Finds the entry with a specified ID.
+     *
+     * @param id
+     * @return The entry that has the given id
+     */
+    public synchronized Optional<BibEntry> getEntryById(String id) {
+        for (BibEntry entry : entries) {
+            if (entry.getId().equals(id)) {
+                return Optional.of(entry);
```
Siedlerchr: Stream with findFirst or findAny

stefan-kolb: You can directly edit this if you find it more intuitive.

stefan-kolb:
Anything that needs to be changed here?

Siedlerchr:
From my point of view, it's good. I merge it in

Issue #2243, opened 2016-11-07
CrossRef entry improvements


stefan-kolb:
- [x] Table cell not high enough, select button cannot be seen
- [x] Rename Select to `Go to entry`: Problem here is that we use a generic `EntryLinkListEditor` for such fields so goto entry may not be appropriate for all of those fields.
- [x] ClassCastException because auto completion is used with the new NOT JTextComponent field inside the crossref field: Maybe reenable again with access to different Java UI component: `AutoCompleteListener.java`line 274

PR #2354, opened 2016-12-08
Crossref improvements

stefan-kolb:
See #2243

- Increase heigth of crossref field
- Rename Select to Jump to entry

image

### DIFF
--- a/src/main/resources/l10n/JabRef_de.properties
+++ b/src/main/resources/l10n/JabRef_de.properties
@@ -2311,4 +2311,4 @@ Could_not_find_any_bibliographic_information.=
 BibTeX_key_%0_deviates_from_generated_key_%1=
 DOI_%0_is_invalid=
 
-Jump_to_entry=
+Jump_to_entry=Springe_zu_Eintrag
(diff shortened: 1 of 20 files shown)


src/main/resources/l10n/JabRef_en.properties
```diff
@@ -2311,3 +2311,4 @@ Could_not_find_any_bibliographic_information.=Could_not_find_any_bibliographic_i
 BibTeX_key_%0_deviates_from_generated_key_%1=BibTeX_key_%0_deviates_from_generated_key_%1
 DOI_%0_is_invalid=DOI_%0_is_invalid
 
+Jump_to_entry=
```
koppor: Shouldn't that be filled?

src/main/resources/l10n/JabRef_de.properties
```diff
@@ -2310,3 +2310,5 @@ Update_with_bibliographic_information_from_the_web=
 Could_not_find_any_bibliographic_information.=
 BibTeX_key_%0_deviates_from_generated_key_%1=
 DOI_%0_is_invalid=
+
+Jump_to_entry=
```
koppor: "Springe zu Eintrag"

koppor:
Besides the minor comments, it looks good.

Could you adapt the functionality of "Jump to entry" so that it works too, if only one entry is selected.

not-functioning


I would place "+" and "-" right next to the field so that the UI is not that ugly, but that's a matter of taste.

Issue #2247, opened 2016-11-08
Large memory usage on Windows 7, freezes on Fedora 24


CsatiZoltan:
JabRef version 3.6 consumes a lot of memory under Windows 7 (more than 500 MB, see the attached screenshot) and sometimes it freezes (than it's allocating more and more, unbounded. I killed the process after 1500 MB). On Fedora 24 freezes much more often. The former versions were totally stable on Windows 7.
![jabref](https://cloud.githubusercontent.com/assets/12742890/20092923/ceef83fe-a59a-11e6-865b-d87b24e1b10c.PNG)

koppor:
Please try the latest dev build

stefan-kolb:
Thank you for your report :+1:
This should be fixed in current `master`. Please try the latest build from http://builds.jabref.org/master.

CsatiZoltan:
After installing the latest build from your link, it didn't become better. When I wanted to close it, it didn't exit, but the allocated memory was not freed (see the screenshot). I had to kill the process.
![capture](https://cloud.githubusercontent.com/assets/12742890/20103337/d87fa00a-a5c9-11e6-915f-dc9016a6c600.PNG)

stefan-kolb:
Can you give us some steps to reproduce the memory leak?

CsatiZoltan:
I just opened it and added an article. Here is the .bib file.
[SaddlePoint.zip](https://github.com/JabRef/jabref/files/578455/SaddlePoint.zip)

stefan-kolb:
Ah thanks great! We will investigate this!

grimes2:
500 MB Memory is normal for Jabref. You system seems to have only 2 GB RAM.

CsatiZoltan:
No, my system has 16 GB of RAM.

Siedlerchr:
Do you have the 64bit java runtime environment installed?


2016-11-08 18:19 GMT+01:00 Zoltán Csáti <notifications@github.com>:

koppor:
Refs #2166 #2175

tobiasdiez:
I can confirm at least the large memory footprint. After opening a normal db (mine has 500 entries and a few groups) JabRef eats only 200 MB of RAM. Now open the entry editor and run through a few entries (using the arrow keys so that a new entry editor is generated for the entries). Result: over 1 GB of RAM usage. Btw: the same db with 3.6 only needed < 100 MB RAM.

stefan-kolb:
[...]

PR #2356, opened 2016-12-09
Update swingx from 1.6.1 to 1.6.4

koppor:
The changelog of swingx is available there: https://java.net/jira/browse/SWINGX?selectedTab=com.atlassian.jira.jira-projects-plugin:changelog-panel

The only changelog entry with "memory leak" is https://java.net/jira/browse/SWINGX-299. I don't know, wether this is related.

I am aware of https://github.com/JabRef/jabref/issues/271. Thus, I manually tested which version to use and 1.6.4 seems to be the latest version usable.

We should check whether this helps solving #2247.

- [n/a] Change in CHANGELOG.md described

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -71,7 +71,9 @@ dependencies {
     compile 'com.jgoodies:jgoodies-forms:1.9.0'
     compile 'com.jgoodies:jgoodies-looks:2.7.0'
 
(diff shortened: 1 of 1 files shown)


lenhard:
I tested this and, at least for me, there seems to be no flickering effect. So, I think we should go ahead.

The only long-term solution will be to get rid of swingx. Hopefully, this is possible in the course of JavaFX.

PR #2357, opened 2016-12-09
Resolves #802 Warn when unknown entry types are used inside Bibtex/Biblatex

### LINKED ISSUE #802: Warn when unknown entry types are used inside Bibtex/Biblatex database
For example if one uses entry types of Biblatex inside a bibtex-defined database.

stefan-kolb:
Quick shot at the checker. WDYT?


stefan-kolb:
I'm not sure if we 100% need this check just thinking...

src/main/java/net/sf/jabref/logic/integrity/BibTeXEntryTypeChecker.java
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * BibTeX mode only checker
+ */
+public class BibTeXEntryTypeChecker implements Checker {
+    private static final List<EntryType> BIBTEX = BibtexEntryTypes.ALL;
+    private static final List<EntryType> BIBLATEX = BibLatexEntryTypes.ALL;
+    private static final List<String> EXCLUSIVE_BIBLATEX = filterEntryTypesNames(BIBLATEX, isNotIncludedIn(BIBTEX));
+
+    public BibTeXEntryTypeChecker() {
```
Siedlerchr: Why empty constructor?

src/main/java/net/sf/jabref/logic/integrity/BibTeXEntryTypeChecker.java
```diff
@@ -0,0 +1,44 @@
...
+import net.sf.jabref.model.entry.BibtexEntryTypes;
+import net.sf.jabref.model.entry.EntryType;
+
+/**
+ * BibTeX mode only checker
+ */
+public class BibTeXEntryTypeChecker implements Checker {
+    private static final List<EntryType> BIBTEX = BibtexEntryTypes.ALL;
+    private static final List<EntryType> BIBLATEX = BibLatexEntryTypes.ALL;
+    private static final List<String> EXCLUSIVE_BIBLATEX = filterEntryTypesNames(BIBLATEX, isNotIncludedIn(BIBTEX));
```
tobiasdiez: I don't see the need for these static lists and would favor if the "exclusive biblatex" list is constructed in the check method.

src/main/java/net/sf/jabref/logic/integrity/BibTeXEntryTypeChecker.java
```diff
@@ -0,0 +1,44 @@
...
+
+    /**
+     * Will check if the current library uses any entry types from another mode.
+     * For example it will warn the user if he uses entry types defined for Biblatex inside a BibTeX library.
+     */
+    @Override
+    public List<IntegrityMessage> check(BibEntry entry) {
+        if (EXCLUSIVE_BIBLATEX.contains(entry.getType())) {
```
tobiasdiez: I think `if (EntryTypes.getType(entry.getType(), Current BibDatabaseMode) == false)` also detects undefined entry types in BibLatexMode

src/main/java/net/sf/jabref/logic/integrity/BibTeXEntryTypeChecker.java
```diff
@@ -0,0 +1,44 @@
...
+            return Collections.singletonList(
+                    new IntegrityMessage(Localization.lang("Entry type %0 is only defined for Biblatex but not for BibTeX", entry.getType()), entry, "bibtexkey")
+            );
+        }
+        return Collections.emptyList();
+    }
+
+    private static List<String> filterEntryTypesNames(List<EntryType> types, Predicate<EntryType> predicate) {
```
tobiasdiez: In my opinion, it makes sense to move this logic to the EntryTypes class and provide a `isExclusiveBibLatex(String type)` method there.

lenhard:
Ok, looks good to me code-wise. Since this had been reviewed by two other devs and you fixed their issues, I will merge this right away.

PR #2358, opened 2016-12-09
Fix Issue 2012: Use bibtexformat for month field in medlinefetcher

### LINKED ISSUE #2012: month fields in braces results in a warning

zellerdev:
Fix #2012 
Checks if the month field is present. In the case it's present, the month field is set to bibtex format using MonthUtil.


koppor:
Could you add a test case?

zellerdev:
I just updated the test cases. As far as I understood it, this should be the normal behavior. So no additional tests should be necessary.

Issue #1873, opened 2016-08-27
Group names are matched wrong


cogsys-michael:
JabRef version 3.6 (also 3.5) on Ubuntu 14.04

Group membership of entries is stored in the groups-field. Groups are identified by their name. Multiple groups are separated by commas. It seems that during matching (for marking group memberships in the groups interface), group names are split at commas and spaces. These facts lead to the following undesired behaviours:
1. If an entry's group field contains 'a b' it is considered a member the groups 'a', 'b', and 'a b'.
2. If an entry's group field contains 'a' it is considered a member of the groups 'something'/'a' and 'anything'/'a'.
3. If an entry's group field contains 'a b', you can remove it from group 'a'. Resulting in the entry being considered a member of 'b' (in combination with 2., possibly in some other branch of the group tree).

This behaviour might completely break the group-entry relations.

Steps to reproduce:
1. download and open [jabref_bug.bib.txt](https://github.com/JabRef/jabref/files/440689/jabref_bug.bib.txt)
2. Add the entry 'demo' to the group 'some rating'/'very good'
3. Observe, that the entry now belongs to all four (leaf) groups
4. Remove the entry from group 'some rating'/'good'
5. Observe that the entry now only belongs to group 'other rating'/'very'

cogsys-michael:
Possible duplicate of #1495.

matthiasgeiger:
Hi Michael!
Thanks for your report. I can confirm the behavior - and yes, you are right this a quite serious problem (and also related to #1495 but not exactly a duplicate).

Regards
Matthias

koppor:
Thank you for reporting this issue. We think, that is already fixed in our development version and consequently the change will be included in the next release.

We would like to ask you to use a development build from https://builds.jabref.org/master and report back if it works for you.

PR #2361, opened 2016-12-10
Fix a few group related issues

### LINKED ISSUE #2334: "Remove entry selection from this group" removes entries from similarly named groups
JabRef 3.7
windows 10 10.0 amd64 
Java 1.8.0_111

Steps to reproduce:

1. Create a Bibtex file with the groups "mygroup" and "mygroup_alternative". Assign a few entries to both of the groups.
2. Select the entries in "mygroup" and remove them from the group using "Remove entry selection from this group".
3. The entries have been removed from both "mygroup" and "mygroup_alternative". Indeed, if you check the group assignments of the entries they now contain "_alternative" as group assignment, i. e., the first part of the group name has been deleted. As no group "_alternative" exists, the items appear to not be assigned to any group.

Note that this bug is related to a workaround/solution to the issue of group hierarchy in JabRef:
https://github.com/JabRef/jabref/issues/1495

If someone was to use

Asthma
    Asthma treatment
    Asthma diagnosis

as proposed in https://github.com/JabRef/jabref/issues/1495, I suspect that the same problem would arise (?).

tobiasdiez:
This PR fixes #2334, #1873 and #1681.
Since I didn't understood the groups code, I also refactored it. Main changes:
- Split `KeywordGroup` class into `WordKeywordGroup`, which splits field content into words and then matches them, and `RegexKeywordGroup` that uses a regular expression for this job.
- Extract serialization of groups to `GroupSerializer` (previously every group decided about its own serialization via the `toString` method)
- Move a few things from model and logic to gui since they mostly contained ui-related stuff (e.g. GroupDesrciptions)

---


lenhard:
The groups code is a mess :-( Thanks for being willing to put up with this!

Unfortunately, the amount of changes is too large for meaningful review and my knowledge on groups and the related code is too limited. I would trust your ability, and the amount of tests that you added here together with the working build. Cosmetic changes can be made on a later occasion. I am sure this is not the last visit to the groups code.

Since this is on your private branch, we have no build on builds.jabref.org. If we had, we could ask the people that reported the issues to try out this version. Any chance of getting a build for this?

src/main/java/net/sf/jabref/Groups.uml
```diff
@@ -0,0 +1,91 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Diagram>
+  <ID>JAVA</ID>
+  <OriginalElement>net.sf.jabref.model.groups.AbstractGroup</OriginalElement>
```
koppor: Why not using [PlantUML](https://plugins.jetbrains.com/plugin/7017) for documenting?

Which tool did you use?

tobiasdiez: IntelliJ

koppor:
- [ ] dragging an entry to a group does not refresh the main table. (keyword-based group)
- [ ] context menu shows "Remove group, keep subgroups" even though there are no subgroups
- [ ] deletion of a static group does not remove the assignment of entries to it
- [ ] clicking on a group and pressing <kbd>Del</kbd> triggers the dialog "Do you want to remove the selcted entries". If I press <kbd>Cursor Down</kbd>, the next group is selected. Thus <kbd>Del</kbd> should really apply to the current group.

Nothing important or me. Think, this was also happening using the old groups code.

Issue #2209, opened 2016-10-29
DOI to Bibtex failing to catch journaltitle from sciencedirect


jdpipe:
When attempting to import the following DOI to Jabref 3.6 (release), the journal title is not picked up. This has occurred for a number of DOIs; I suspect there is some problem in general for papers coming from the ScienceDirect site. FWIW there is already a Bibtex export feature in ScienceDirect, so perhaps that can be reused in some way.

Here is the DOI that gave problems:
http://dx.doi.org/10.1016/j.renene.2016.04.033

koppor:
Refs #1603

## Information exported by ScienceDirect

```
@article{Giostri2016449,
title = "Microalgae cofiring in coal power plants: Innovative system layout and energy analysis ",
journal = "Renewable Energy ",
volume = "95",
number = "",
pages = "449 - 464",
year = "2016",
note = "",
issn = "0960-1481",
doi = "http://dx.doi.org/10.1016/j.renene.2016.04.033",
url = "http://www.sciencedirect.com/science/article/pii/S0960148116303147",
author = "A. Giostri and M. Binotti and E. Macchi",
keywords = "Renewable energy",
keywords = "Biomass",
keywords = "Bioenergy",
keywords = "Microalgae",
keywords = "Raceway pond",
keywords = "Cofiring "
}
```

## Information retrieved by JabRef
```
@article{Giostri_2016,
	doi = {10.1016/j.renene.2016.04.033},
	url = {http://dx.doi.org/10.1016/j.renene.2016.04.033},
	year = 2016,
	month = {sep},
	publisher = {Elsevier {BV}},
	volume = {95},
	pages = {449--464},
	author = {A. Giostri and M. Binotti and E. Macchi},
	title = {Microalgae cofiring in coal power plants: Innovative system layout and energy analysis},
	journal = {Renewable Energy}
}
```

## Comments

In both bibtex entries, the field `journal` exists. Could you elaborate more what you think is missing?

koppor:
I think, it's the typical issue with bibtex vs. biblatex:

![grabbed_20161101-113108](https://cloud.githubusercontent.com/assets/1366654/19887073/b7ad217a-a026-11e6-81f1-ca3830285667.png)

You could execute the cleanup operation "Convert to BibLatex format", which fixes the journal field:

![grabbed_20161101-113146](https://cloud.githubusercontent.com/assets/1366654/19887082/c7667e40-a026-11e6-8631-95fa7f5de383.png)

[...]

Issue #521, opened 2015-12-15
Add support for biblatex field aliases


koppor:
In the biblatex documentation at Section 2.2.5, "Field Aliases" are described. 
- address for location
- annote for annotation (jurabib)
- archiveprefix for eprinttype (arXiv)
- journal for journaltitle
- key for sortkey
- pdf for file (JabRef compatibility :sunny:)
- primaryclass for eprintclass (arXiv)
- school for institution

Issue: When switching from bibtex to biblatex mode, the field "journal" is written as optional field, even though it is a required field.
### quick hack

In the serialization of JabRef, aliased fields should be treated like the original fields.

Issue with the quick hack: The aliased field is still treated as optional field or not treated at all. This seems to be inconsistent.
### related bug reports
- [support for journal alias](https://sourceforge.net/p/jabref/bugs/1014/). Reported to be resolved somehow in "JabRef v2.11 beta"

cbersch:
The related bug report on sf isn't relevant, since this was only about the mere possibility to use the `journaltitle` field at all. It didn't try at all to handle field aliases properly.

koppor:
We have some support in the entry table: #24 and some code has been written for the journaltitle field: #1720 and #15

The entry editor, however, still does not support alias and this causes issues when fetching entries: #2209

tobiasdiez:
This is now implemented in the latest development version (at least if I understand it correctly). Could you please check the build from http://builds.jabref.org/master/. Thanks!

PR #2364, opened 2016-12-11
Be more friendly when using journal alias field

koppor:
I'd like to copy bibtex entries to biblatex databases and vice versa. When using required fields only, this works perfectly. When in biblatex mode, JabRef does not display the value of `journal` in the required fields, because JabRef demands the field `journaltitle`. This is not only confusing for me, but also for other users: #2209.

I know that a clean solution is #521. This, however, won't happen this year.

I also know that a bibtex-to-biblatex converter and biblatex-to-bibtex converter is another solution. This, however, forces users to run these when copying and pasting between databases. This could be solved by running these converters silently when copying and pasting. When JabRef is used in parallel to other software (such as Notepad++), this does not help. I cannot force everyone to use JabRef. Thus, simply supporting `journal` in JabRef is the solution for me. 

The biblatex manual states:

> journal field (literal)
> An alias for journaltitle, provided for BibTeX compatibility. See § 2.2.2.

Thus, it is not deprecated, but an alias and should IMHO be considered as full alternative for `journaltitle`.

<s>I am aware that this patch makes it hard for users using full biblatex with `journaltitle` instead of `journal`. I am assuming that not much users are aware of the different modes of JabRef and that they switch back and forth for the same file or for the bibentries.</s>

We now show both `journaltitle` and `journal` if these fields exists in as required fields.

This whole discussion somehow refs https://github.com/JabRef/jabref/issues/160

 * Minor improvement: Enhanced integrity checks testing for biblatex-only fields to be aware of more fields (e.g., `location`).
 * All fetchers return now bibtex. (Before: ArXiv and msbib created biblatex)

## Screenshot: Current JabRef
grabbed_20161211-183240

## Screenshot: Updated JabRef
grabbed_20161212-030057

## Checklist

- [n/A] Check documentation status (Issue created for outdated help page at [help.jabref.org](https://github.com/JabRef/help.jabref.org/issues)?)

### DIFF
--- a/src/main/java/net/sf/jabref/logic/importer/fetcher/ArXiv.java
+++ b/src/main/java/net/sf/jabref/logic/importer/fetcher/ArXiv.java
@@ -343,7 +343,7 @@ public class ArXiv implements FulltextFetcher, SearchBasedFetcher, IdBasedFetche
             abstractText.ifPresent(abstractContent -> bibEntry.setField(FieldName.ABSTRACT, abstractContent));
             getDate().ifPresent(date -> bibEntry.setField(FieldName.DATE, date));
             primaryCategory.ifPresent(category -> bibEntry.setField(FieldName.EPRINTCLASS, category));
-            journalReferenceText.ifPresent(journal -> bibEntry.setField(FieldName.JOURNAL, journal));
+            journalReferenceText.ifPresent(journal -> bibEntry.setField(FieldName.JOURNALTITLE, journal));
             getPdfUrl().ifPresent(url -> bibEntry
                     .setFiles(Collections.singletonList(new ParsedFileField("online", url, "PDF"))));
             return bibEntry;
--- a/src/main/java/net/sf/jabref/logic/importer/fetcher/ArXiv.java
+++ b/src/main/java/net/sf/jabref/logic/importer/fetcher/ArXiv.java
@@ -16,7 +16,6 @@ import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.parsers.ParserConfigurationException;
 
-import net.sf.jabref.logic.TypedBibEntry;
 import net.sf.jabref.logic.help.HelpFile;
 import net.sf.jabref.logic.importer.FetcherException;
 import net.sf.jabref.logic.importer.FulltextFetcher;
@@ -26,7 +25,6 @@ import net.sf.jabref.logic.importer.SearchBasedFetcher;
 import net.sf.jabref.logic.importer.util.OAI2Handler;
 import net.sf.jabref.logic.util.DOI;
 import net.sf.jabref.logic.util.io.XMLUtil;
-import net.sf.jabref.model.database.BibDatabaseMode;
 import net.sf.jabref.model.entry.BibEntry;
 import net.sf.jabref.model.entry.BibtexEntryTypes;
 import net.sf.jabref.model.entry.FieldName;
@@ -345,10 +343,9 @@ public class ArXiv implements FulltextFetcher, SearchBasedFetcher, IdBasedFetche
             abstractText.ifPresent(abstractContent -> bibEntry.setField(FieldName.ABSTRACT, abstractContent));
             getDate().ifPresent(date -> bibEntry.setField(FieldName.DATE, date));
             primaryCategory.ifPresent(category -> bibEntry.setField(FieldName.EPRINTCLASS, category));
-            journalReferenceText.ifPresent(journal -> bibEntry.setField(FieldName.JOURNALTITLE, journal));
-            getPdfUrl().ifPresent(url -> (new TypedBibEntry(bibEntry, BibDatabaseMode.BIBLATEX))
+            journalReferenceText.ifPresent(journal -> bibEntry.setField(FieldName.JOURNAL, journal));
+            getPdfUrl().ifPresent(url -> bibEntry
                     .setFiles(Collections.singletonList(new ParsedFileField("online", url, "PDF"))));
-
             return bibEntry;
         }
     }
--- a/src/main/java/net/sf/jabref/logic/integrity/NoBibtexFieldChecker.java
+++ b/src/main/java/net/sf/jabref/logic/integrity/NoBibtexFieldChecker.java
@@ -1,31 +1,32 @@
 package net.sf.jabref.logic.integrity;
 
-import java.util.Collections;
 import java.util.List;
-import java.util.Optional;
+import java.util.SortedSet;
+import java.util.TreeSet;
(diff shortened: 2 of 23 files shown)


lenhard:
Hehe, let us see how political this will get :-)

Code-wise the PR looks fine and I give my +1 for merging. The only blocker is that, for some reason, the GUI tests are failing.

Siedlerchr:
You could also create a discussion point at https://github.com/plk/biblatex

koppor:
In case, I want to start a discussion, I'm going to do that on the [DANTE e.V. mailing list](https://lists.dante.de/mailman/listinfo/dante-ev) or same [DANTE e.V. event](https://www.dante.de/events.html). 😇

koppor:
The GUI tests are also failing on master: https://travis-ci.org/JabRef/jabref/builds/183054056

Refs https://github.com/JabRef/jabref/pull/1700

matthiasgeiger:
I don't like this.

Despite I'm not using the Biblatex mode I think it is the wrong approach to use the "deprecated" field instead of the intended field only because we are too lazy to implement our fetchers in the right way or because the user is too lazy to convert copied BibTeX entries to biblatex :wink:

tobiasdiez:
`journaltitle` is the main biblatex field for journal names and hence JabRef should support it as a first class citizen. This means, it should be displayed as a required field and "to biblatex" should convert `journal` to `journaltitle`. Moreover, the journal is not the only controversial field: date is the preferred way to store the publication date but bibtex only supports `year` and `month`. In my opinion, we shouldn't reinterpret the "biblatex standard" just because some users have problems. 

But of course I can see that the new conventions are sometimes inconvenient if users switch between BibTeX and BibLaTeX. As far as I see it, the main problem is that "journal" lands in "deprecated" although it is an alias to a required field. Would it be a solution to show the journal field right below "journaltitle" in "required" if it has a value? Then it would work similar to year+month vs date.

koppor:
@matthiasgeiger `journal` is not deprecated, it is just an alias field.

koppor:
@tobiasdiez Yes, that would be a solution.

matthiasgeiger:
Therefore I put it in quotes ;-)

src/main/java/net/sf/jabref/logic/importer/fetcher/ArXiv.java
```diff
@@ -345,10 +343,9 @@ public BibEntry toBibEntry(Character keywordDelimiter) {
             abstractText.ifPresent(abstractContent -> bibEntry.setField(FieldName.ABSTRACT, abstractContent));
             getDate().ifPresent(date -> bibEntry.setField(FieldName.DATE, date));
             primaryCategory.ifPresent(category -> bibEntry.setField(FieldName.EPRINTCLASS, category));
-            journalReferenceText.ifPresent(journal -> bibEntry.setField(FieldName.JOURNALTITLE, journal));
-            getPdfUrl().ifPresent(url -> (new TypedBibEntry(bibEntry, BibDatabaseMode.BIBLATEX))
+            journalReferenceText.ifPresent(journal -> bibEntry.setField(FieldName.JOURNAL, journal));
```
tobiasdiez: Please leave `journaltitle` here since the returned entry is in biblatex format (see date above)

src/main/java/net/sf/jabref/model/entry/BibEntry.java
```diff
@@ -824,4 +827,17 @@ private void invalidateFieldCache(String fieldName) {
...
+    public Optional<FieldChange> setFiles(List<ParsedFileField> files) {
+        Optional<String> oldValue = this.getField(FieldName.FILE);
+        String newValue = FileField.getStringRepresentation(files);
+
+        if (oldValue.isPresent() && oldValue.get().equals(newValue)) {
+            return Optional.empty();
+        }
+
+        this.setField(FieldName.FILE, newValue);
+        return Optional.of(new FieldChange(this, FieldName.FILE, oldValue.orElse(""), newValue));
```
tobiasdiez: return this.setField should be enough

src/main/java/net/sf/jabref/logic/integrity/NoBibtexFieldChecker.java
```diff
@@ -1,31 +1,32 @@
...
+import java.util.TreeSet;
+import java.util.stream.Collectors;
 
 import net.sf.jabref.logic.integrity.IntegrityCheck.Checker;
 import net.sf.jabref.logic.l10n.Localization;
 import net.sf.jabref.model.entry.BibEntry;
+import net.sf.jabref.model.entry.EntryConverter;
 import net.sf.jabref.model.entry.FieldName;
 
 public class NoBibtexFieldChecker implements Checker {
```
tobiasdiez: Rename to NoBibLatex ?

src/test/java/net/sf/jabref/logic/integrity/NoBibTexFieldCheckerTest.java
```diff
@@ -0,0 +1,45 @@
...
+        BibEntry entry = new BibEntry();
+        entry.setField("journal", "test");
+        assertEquals(Collections.emptyList(), checker.check(entry));
+    }
+
+    @Test
+    public void journaltitleIsRecognizedAsBibLaTeXOnlyField() {
+        BibEntry entry = new BibEntry();
+        entry.setField("journaltitle", "test");
+        assertNotEquals(Collections.emptyList(), checker.check(entry));
```
tobiasdiez: This and location should report an error, shouldn't they?

src/main/java/net/sf/jabref/model/entry/BibEntry.java
koppor: Done.

src/main/java/net/sf/jabref/logic/integrity/NoBibtexFieldChecker.java
koppor: Added class comment

> This checker checks whether the entry does not contain any field appearing only in BibLaTeX (and not in BibTex)

Thus, the name is IMHO correct.

src/test/java/net/sf/jabref/logic/integrity/NoBibTexFieldCheckerTest.java
koppor: Why? `journaltitle` and `location` are BibLaTeX only. Thus, everything is allright.

koppor: Fixed in c7841a6dcf839ccd8a2bfe7dcbb1481ec11dfba7

koppor:
All comments addressed. Should be good to go now.

PR #2368, opened 2016-12-12
Never manually set ID of a BibEntry

### LINKED ISSUE #1721: Wrong usage of BibEntry field `id`
The field `id` in `BibEntry` is the internal ID used by JabRef and **NOT** the bibtexKey (which is set using (`setCiteKey(...)`). Moreover, the ID should be unique.

However, a lot of the importers and also at other spots either the bibtexkey, or a fixed value is used to set the ID.

I'm not sure whether this is already causing real trouble, but it is at least a source for potential bugs if we rely on the uniqueness of the IDs somewhere in the code.

Thus: `setId()` and the constructors of `BibEntry` should always use `IdGenerator.next()` to create a valid, unique ID value.

stefan-kolb:
The internal ID of a BibEntry should never be set manually. Closes #1721

DONE:
- Removed constructors that allow setting the ID of an entry
- __Every__ BibEntry now gets an ID on creation
- A clone of a BibEntry __always gets a unique ID__ now
- Only option to change ID of an entry is now via `setID()`

PENDING:
- [ ] Is IDGenerator really global and consistent between calls now?
- [ ] Can we import Bibentries without an open Database? See c884745. Importers used a static ID value before. Not 100% sure if this change has any side effects.
- [ ] setID may still be replaced for some occurences?! Maybe we can even remove this execpt for a few test cases or so.

### DIFF
--- a/src/main/java/net/sf/jabref/logic/importer/fileformat/EndnoteImporter.java
+++ b/src/main/java/net/sf/jabref/logic/importer/fileformat/EndnoteImporter.java
@@ -253,8 +253,9 @@ public class EndnoteImporter extends Importer {
                 hm.put(FieldName.PAGES, artnum);
             }
 
-            BibEntry b = new BibEntry(DEFAULT_BIBTEXENTRY_ID, type); // id assumes an existing database so don't
+            // id assumes an existing database so don't
             // create one here
+            BibEntry b = new BibEntry(type);
             b.setField(hm);
             if (!b.getFieldNames().isEmpty()) {
                 bibitems.add(b);
--- a/src/main/java/net/sf/jabref/logic/importer/fileformat/EndnoteImporter.java
+++ b/src/main/java/net/sf/jabref/logic/importer/fileformat/EndnoteImporter.java
@@ -253,8 +253,6 @@ public class EndnoteImporter extends Importer {
                 hm.put(FieldName.PAGES, artnum);
             }
 
-            // id assumes an existing database so don't
-            // create one here
             BibEntry b = new BibEntry(type);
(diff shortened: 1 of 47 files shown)


stefan-kolb:
This is ready for review now. Please carefully see the changes in the initial post and think about possible side effects I might have missed! I also need some advice for the setID() replacement.

src/main/java/net/sf/jabref/logic/importer/fileformat/EndnoteImporter.java
```diff
@@ -253,8 +253,9 @@ else if ("P".equals(prefix)) {
                 hm.put(FieldName.PAGES, artnum);
             }
 
-            BibEntry b = new BibEntry(DEFAULT_BIBTEXENTRY_ID, type); // id assumes an existing database so don't
+            // id assumes an existing database so don't
```
koppor: I think, this comment can be deleted, can't it?

koppor:
LGTM. Shouldn't we (i) use this Id as HashCode? or (ii) get rid of the Id completely?

stefan-kolb:
We could use it as hash code. We also use it to distinguish duplicated/cloned entries + in our integrity checks to jump to an entry (that might have no key for example).

koppor:
I don't quickly get why we cant use `equals()` etc. OK, probably the code needs to be rewritten accordingly.

stefan-kolb:
Maybe we could do this eventually...

koppor:
Code looks good. Your isses seem hard to tackle. We merge after v3.8 and then test in production. 😇

PR #2369, opened 2016-12-12
Remove four Blob listeners

arnobl:
Hi guys,

Sorry in advance for the long text.

We are researchers in software engineering that currently work on user interface (UI) bad coding practices. We are studying UI listeners, in particular listeners that manage several widgets, like the following listener extracted from JabRef:

```
public void valueChanged(ListSelectionEvent evt) {
    Object source = evt.getSource();
    if (familyList.equals(source)) {
        String family = familyList.getSelectedValue();
        if (family != null) {
            familyField.setText(family);
        }
    } else if (sizeList.equals(source)) {
        String size = sizeList.getSelectedValue();
        if (size != null) {
            sizeField.setText(size);
        }
    } else if (styleList.equals(source)) {
        String style = styleList.getSelectedValue();
        if (style != null) {
            styleField.setText(style);
        }
    }
    updatePreview();
}
```

In such listeners, the source widget is identified using if/switch statements. Thanks to empirical studies we have indications that such a practice has a negative impact on the code quality. We call UI listeners that control three or more widgets, Blob Listeners (see this PDF document for more details: https://hal.inria.fr/hal-01308625v2/document).

We identified in JabRef six instances of the Blob listener and for four of them we have a refactoring solution. This pull request (PR) removes these four Blob listeners by splitting them into atomic UI listeners (i.e. they manage a single widget). Each commit of this PR refers to one or two Blob listeners. We manually tested JabRef for any regression. The new code is less complex in terms of cyclomatic complexity (CC) and lines of code (LoC): the new code has -38 LoCs and -21 of CC on the refactored files (according to Sonar analyses).

Complementary to this PR, we have several questions for the JabRef's developers:

1/ Do you think that coding UI listeners that manage several widgets is a bad coding practice?

2/ Do you think that the threshold of three or more widgets per listener relevant for characterising a Blob listener?

3/ Do you think that the refactored code relevant to remove Blob listeners?

4/ Do you have any idea of the reason of the presence of Blob listeners in your code?
For example, lack of programming experience? Lack of anonymous functions (aka lambdas) in Java 7 (and previous JDKs)?

5/ We identified two Blob listener for which we do not have a refactoring solution:
https://github.com/JabRef/jabref/blob/master/src/main/java/net/sf/jabref/gui/autocompleter/AutoCompleteListener.java#L70
https://github.com/JabRef/jabref/blob/master/src/main/java/net/sf/jabref/gui/BasePanel.java#L1409
Do you have any idea if these listener have to and can be refactored?

6/ Do you have suggestions about other bad user interface coding practices? This can concern controllers, commands, GUIs, widgets, data bindings, GUI testing, etc.

### DIFF
--- a/src/main/java/net/sf/jabref/gui/EntryTypeList.java
+++ b/src/main/java/net/sf/jabref/gui/EntryTypeList.java
@@ -1,7 +1,6 @@
 package net.sf.jabref.gui;
 
 import java.awt.GridBagConstraints;
-import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
 import java.util.List;
 import java.util.Optional;
@@ -42,7 +41,7 @@ public class EntryTypeList extends FieldSetComponent implements ListSelectionLis
         gbl.setConstraints(def, con);
         add(def);
         list.addListSelectionListener(this);
-        def.addActionListener(this);
+        def.addActionListener(e -> def.setEnabled(false));
         def.setEnabled(false);
         remove.setEnabled(false);
     }
@@ -124,16 +123,6 @@ public class EntryTypeList extends FieldSetComponent implements ListSelectionLis
 
     public void addDefaultActionListener(ActionListener l) {
         def.addActionListener(l);
-    }
-
-    @Override
-    public void actionPerformed(ActionEvent e) {
-        // Default button pressed.
-        if (e.getSource() == def) {
-            def.setEnabled(false);
-        } else {
-            super.actionPerformed(e);
-        }
     }
 
     @Override
--- a/src/main/java/net/sf/jabref/gui/FieldSetComponent.java
+++ b/src/main/java/net/sf/jabref/gui/FieldSetComponent.java
@@ -162,17 +162,6 @@ class FieldSetComponent extends JPanel {
         } else {
             sel = new JComboBox<>(preset.toArray(new String[preset.size()]));
             sel.setEditable(true);
-//            sel.addActionListener(e -> {
-//                if ("comboBoxChanged".equals(e.getActionCommand()) && (e.getModifiers() == 0)) {
-//                    // These conditions signify arrow key navigation in the dropdown list, so we should
-//                    // not react to it. I'm not sure if this is well defined enough to be guaranteed to work
-//                    // everywhere.
-//                    return;
-//                }
-//                String s = sel.getSelectedItem().toString();
-//                addField(s);
-//                sel.getEditor().selectAll();
-//            });
             gbl.setConstraints(sel, con);
             add(sel);
         }
--- a/src/main/java/net/sf/jabref/gui/FieldSetComponent.java
+++ b/src/main/java/net/sf/jabref/gui/FieldSetComponent.java
@@ -13,7 +13,6 @@ import java.util.ArrayList;
 import java.util.Enumeration;
 import java.util.HashSet;
 import java.util.List;
-import java.util.Objects;
 import java.util.Set;
 
 import javax.swing.Box;
@@ -40,7 +39,7 @@ import net.sf.jabref.preferences.JabRefPreferences;
 /**
  * @author alver
  */
-class FieldSetComponent extends JPanel implements ActionListener {
+class FieldSetComponent extends JPanel {
 
     private final Set<ActionListener> additionListeners = new HashSet<>();
     protected final JList<String> list;
@@ -94,8 +93,16 @@ class FieldSetComponent extends JPanel implements ActionListener {
         list = new JList<>(listModel);
         list.setSelectionMode(ListSelectionModel.MULTIPLE_INTERVAL_SELECTION);
         // Set up GUI:
-        add.addActionListener(this);
-        remove.addActionListener(this);
+        add.addActionListener(e -> {
+            // Selection has been made, or add button pressed:
+            if ((sel != null) && (sel.getSelectedItem() != null)) {
+                String s = sel.getSelectedItem().toString();
+                addField(s);
+            } else if ((input != null) && !"".equals(input.getText())) {
+                addField(input.getText());
+            }
+        });
+        remove.addActionListener(e -> removeSelected()); // Remove button pressed
 
         setLayout(gbl);
         con.insets = new Insets(1, 1, 1, 1);
@@ -118,8 +125,8 @@ class FieldSetComponent extends JPanel implements ActionListener {
             con.weightx = 0;
             up = new JButton(IconTheme.JabRefIcon.UP.getSmallIcon());
             down = new JButton(IconTheme.JabRefIcon.DOWN.getSmallIcon());
-            up.addActionListener(this);
-            down.addActionListener(this);
+            up.addActionListener(e -> move(-1));
+            down.addActionListener(e -> move(1));
             up.setToolTipText(Localization.lang("Move up"));
             down.setToolTipText(Localization.lang("Move down"));
             gbl.setConstraints(up, con);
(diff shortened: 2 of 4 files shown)


Siedlerchr:
First of all, thank you very much for taking the time to analyze JabRef and other OpenSource. We are "proud" that we are relevant enough to be considered in your analysis ;)
I quick read your referenced paper and find the insights very interesting. 

To answer some of your additional questions, I first have to state, that I am not a long standing JabRef developer and that many of the original developers are no longer active and this of course changed over the time. However, as I have a background as Software developer and worked with several software systems, I think I can answer your questions.

> 1/ Do you think that coding UI listeners that manage several widgets is a bad coding practice?

In general I would agree to you, following the Separation of Concern principle. However, there might be situations where this can not be achieved fully, e.g. due to limiting implementations provided by the framework.

>2/ Do you think that the threshold of three or more widgets per listener relevant for characterising a Blob listener?

Strictly speaking I would say, more than one or two are definetly an indicator. However, setting the threshold to low could lead to many false positives.

>3/ Do you think that the refactored code relevant to remove Blob listeners?

I took a look at the refactored code and would say it is relevant. 

>4/ Do you have any idea of the reason of the presence of Blob listeners in your code?

As the orginal JabRef version is very old (first release in 2003 or earlier ) and during the time there have been working many devs on it with a different skill set. I would say we both have technical (lack of features) and skill related aspects. Despite that, as far as I would say from a personal view,  the idea of following coding best practices and standards has been becoming more the focus in the last years. Especially in university lectures etc where still Java 2 was the state of the art for many years, despite newer versions.

5/ We identified two Blob listener for which we do not have a refactoring solution:

We will check them 

6/ Do you have suggestions about other bad user interface coding practices? This can concern controllers, commands, GUIs, widgets, data bindings, GUI testing, etc.

Apart from the design and interaction of the GUI, I think a typical bad practices is a more general practice, having logic code inside the GUI classes. Another thing would be a calling GUI code from other threads without invoking or  executing long operations in the gui code.

Mabye some others @JabRef/developers  have some more ideas on that ;)

Thank you very much for your useful analysis.
Feel free to contribute more ;)

src/main/java/net/sf/jabref/gui/EntryTypeList.java
```diff
@@ -127,16 +126,6 @@ public void addDefaultActionListener(ActionListener l) {
     }
 
     @Override
-    public void actionPerformed(ActionEvent e) {
-        // Default button pressed.
-        if (e.getSource() == def) {
-            def.setEnabled(false);
-        } else {
-            super.actionPerformed(e);
```
lenhard: If I read the code correctly, the propagation of the event to super gets lost in your refactoring. Please make sure that you preserve the existing functionality, unless it is a bug.

src/main/java/net/sf/jabref/gui/FieldSetComponent.java
```diff
@@ -149,13 +156,23 @@ private FieldSetComponent(String title, List<String> fields, List<String> preset
...
         if (preset == null) {
             input = new JTextField(20);
-            input.addActionListener(this);
+            input.addActionListener(e -> addField(input.getText()));
             gbl.setConstraints(input, con);
             add(input);
         } else {
             sel = new JComboBox<>(preset.toArray(new String[preset.size()]));
             sel.setEditable(true);
-            //sel.addActionListener(this);
```
lenhard: Usually, commented-out code should be deleted. So instead of extending them, please remove these lines.

src/main/java/net/sf/jabref/gui/EntryTypeList.java
arnobl: The refactoring is ok since the super call is also refactored in the same commit (the super class does not implement ActionListener anymore and thanks to the inheritance the super call is now useless). 

I looked at the type hierarchy of these two classes to check that this change has no impact on other classes.

src/main/java/net/sf/jabref/gui/FieldSetComponent.java
arnobl: Done

lenhard:
Regarding your follow-up questions, I have little to add on top of the comments of @Sie
dlerchr 

Overall, I like the refactoring since it enforces separation of concerns more clearly. The only downside I see is that this bloats the initialization code somewhat, as we usually add the listeners at the place where we create the UI components.

1. Hard to say. I does not strictly violate the MVC pattern, I would say. Therefore, it highly depends on the concrete situation. Overall, I like your solution.
2. I agree that more than one could be used as threshold, but generalizing this is probably not possible.
3. I assume that by "relevant" you mean "suitable". In my point of view, the solution is ok.
4. As @Siedlerchr said, JabRef is old. It has grown and evolved without a structure or plan over many years and many people with a different background contributed to it. In such an environment, a certain degree of chaos is inevitable. Availability of Java language features are probably not the reason.
5. The UI classes do not have to be refactored as long as they are working. We are trying to maintain a higher degree of quality in the other parts of jabref (btw, you might want to have a look at https://github.com/JabRef/jabref/wiki/High-Level-Documentation ). In this setting, the UI really has the lowest priority in terms of code quality. Refactoring is welcome of course, but we are not actively changing UI classes just for the reason of quality improvement. Our illusive hope is that everything will get better when we switch to JavaFX.
6. Our most pressing need is a framework that allows us to write easy and concise UI tests that remain stable.

src/main/java/net/sf/jabref/gui/EntryTypeList.java
arnobl: Also, I manually tested JabRef by using the impacted widgets.

lenhard: Ok, thanks! It might have been unnecessary during all of its existence, or maybe it had a purpose at some point in time that got lost. In any case, its removal is positive.

arnobl: Great!
The super call was not useless. However, the super class is also refactored so that its widgets now register lambda listeners that EntryTypeList now inherits.

arnobl:
Indeed, the detection of code smells always depends on the context. That is why the threshold should be customised by developers depending on their needs.

Regarding JavaFX, a very nice testing framework is TestFX: https://github.com/TestFX/TestFX
I wrote tutorials on how to write simple and complex tests and how to factorise test code with this library:
https://torgen-engineering.blogspot.fr/2015/11/testing-javafx-applications-with-testfx.html
https://torgen-engineering.blogspot.fr/2015/12/gui-testing-how-to-compare-javafx-gui.html

I am also porting an app from Swing to JFX: https://github.com/arnobl/latexdraw/tree/master/latexdraw-core/net.sf.latexdraw
This port is supported by a large UI testing suite powered by TestFX:
https://github.com/arnobl/latexdraw/tree/master/latexdraw-core/net.sf.latexdraw/src/test/test/gui

Hope it may help. Porting to JFX is not an easy task.

Thanks a lot for your feedback!
If other devs want to give their opinion, your welcome.

lenhard:
Thanks for your suggestions! Do have experience with FX support in OpenJdk? We would welcome any hints regarding pitfalls with this one.

One more thing regarding your solution here that came to my mind. A major advantage of your solution, in my point of view, is that the top level class no longer needs to implement `ActionListener`. In terms of information hiding, this is a major advantage.

Finally, I tested the listeners in the in a running JabRef and everything seems to work fine. Let us give some other devs the opportunity to have a look at the code. Nevertheless, this might make it into v3.8 on friday.

arnobl:
Thanks again for the feedback.
Even if I am an open-sourcer, I do not really like OpenJDK and always use and recommend the OracleJDK (I know, shame on me). The main reason is that many users reported me issues that do not appear with the OracleJDK.
Ubuntu/Debian releases the OpenJFX library but I never use it. Fedora do not package OpenJFX yet (https://bugzilla.redhat.com/show_bug.cgi?id=1145303).

Issue #938, opened 2016-03-11
BibTeXML vs. bibteXMP


koppor:
JabRef 3.2

It seems that JabRef offers a second kind of XML serialization in BibTeX:

```
xmlns:bibtex='http://jabref.sourceforge.net/bibteXMP/'
```

IMHO, it is not worth to keep two different XML Schemas for an XML serialization of BibTeX. AFAIK, there isn't even one for JabRef's XML. Therefore, I propose that we should use [BibTeXML ](http://bibtexml.sourceforge.net/) only and migrate old XMP meta data to the BibTeXML format.

XMP examples can be found at https://github.com/JabRef/jabref/blob/fc82796f1f99c00a01ccdb9dd4fff792a40b7e75/src/test/java/net/sf/jabref/logic/xmp/XMPUtilTest.java#L139.

koppor:
Refs #898

oscargus:
I'm not really following the argumentation. One may argue of different export formats, but how is it relevant that they are both XML? Isn't it more of an issue if it is a relevant format in itself?

koppor:
Context: The format is used for storing BibTeX data in XML files using the XMP functionality (follow `net.sf.jabref.logic.xmp.XMPSchemaBibtex`). This PDF meta data is used by other people to exchange PDFs with the correct bibliographic data without being forced to send the bib entry along with the PDF in two files.

I am arguing that JabRef uses a proprietary format which is not used elsewhere. Thus, our XMP data cannot be processed by other software. I see the point, that the last commit at the [current BibTeXML repository](https://github.com/Zearin/BibTeXML) is from 2011. Nevertheless, I vote for joining forces. These formats are too similar to go into different directions.

I see following alternatives:

 1. Replace JabRef's bibteXMP by the canonical bibtex representation
 2. Completely use RDF. There seem multiple BibTeX2RDF converters available: https://www.w3.org/wiki/ConverterToRdf#BibTex
 3. Maybe, OWL is also an option: http://zeitkunst.org/bibtex/0.1/
 4. Move to BiBTeXML (as outlined in the original issue)
 5. Keep everything as is

Somehow, the current code seems to use "Dublin Core", which reads good. Maybe, that code can just be used and the other serialization using `{http://jabref.sourceforge.net/bibteXMP/}bibtex` can be removed completely. Needs to be investigated further.

oscargus:
Ah, OK, so bibteXMP is JabRef's own format? Then it clearly makes more
[...]

PR #357, opened 2015-11-17
Add a few tests to the field formatter checking multiline normalization

tobiasdiez:
Inspired by the old (and fixed) bug [SF #1191](http://sourceforge.net/p/jabref/bugs/1191/).

PR #2371, opened 2016-12-12
Reenable more tests

stefan-kolb:
- [x] Corrections of wrong Bibtex serializations: I inverted the tests for the file field so they test if the fields are not changed for \n and \t.
- [x] Edge case behavior: Not implemented right now and won't be as @lenhard says. Removed for now until implemented.
- [x] Author parsing
The result is: `Expected :Other, A. N.; Actual   :Other, Jr., A. N.` We decided that this is expected behavior for now.
- [x] Layouter test: logic is just not handling `\n\n` to `<br>` conversion. Cannot be handled anymore as the BibTeX parser strips duplicated whitespaces by default!
- [x] XMP test: Fixed
- [x] OO test: Fixed
- [x] Ignored GUI tests: Will stay ignored for now.


stefan-kolb:
@lenhard You can probably comment a little bit on the test that were previously inside the parser.

stefan-kolb:
I tend to either fix the ignored tests or remove them. For some of them I don't even know what they should do...

stefan-kolb:
@JabRef/developers I need help from all of you that have expertise in at least one of these topics!

tobiasdiez:
The tests for the file field are a bit strange. I'm not sure why tabs and newlines are bad there. But, anyways, the parser shouldn't remove whitespace; this should be done in a separate cleanup.

The result of the author parsing also looks good to me (although one could argue that the Jr part is unimportant and should  be deleted in an abbreviated representation...).

koppor:
XMP: This is JabRef's implemenation to store BibTeX data inside PDFs to ease sharing of BibTeX data along with the PDF. Refs #938 https://github.com/JabRef/jabref/pull/1096#issuecomment-220529040. Think, we have to switch to Dublin Core or to biblatexml. This takes more than a few hours. Since I am fan of this way of sharing, I'd like to keep the tests to be able to think of backward compatibility.

lenhard:
My 2 cents:
 - newline / tab elimination: That happens in `FieldContentParser`, which is called during parsing, and does work. However, the file field is explicitly exempted from this behavior, so the test fails (trying to fix the content of the `file` field is calling for problems). I think this should be changed to some other field, say `title` and the tests would be fine.
 - edge case: This is documented in #1212 and does not work. With the current structure of the parser, this is also far from trivial, since it would need a lot of look-ahead in the parsing. My strategy has been to ignore it until some users actually complain about it. I would keep on doing that.
 - author parsing: No idea really, I would keep ignoring it.
 - The html stuff I am not sure as well. I haven't really touched those formatters so far.

stefan-kolb:
@lenhard Thanks for your input. So I just need to change the first tests to another field right now, e.g. title?!
I will remove the dge case then for now.
Looks like the author parser is correct.
So I'm just missing a resolution for the HTML and XMP stuff 😄 👍

matthiasgeiger:
> However, the file field is explicitly exempted from this behavior, so the test fails (trying to fix the content of the file field is calling for problems). I think this should be changed to some other field, say title and the tests would be fine.

Or: Rewrite the tests to check that the content of the file field is NOT changed. @stefan-kolb You should check whether tests for these two aspects are already in place - if so, you can just delete the `ignored` ones

lenhard:
@stefan-kolb @matthiasgeiger If you invert the tests, then use either `review` or `abstract`. These are the only hard-coded fields where we avoid newline elimination. The other ones can be configured in the preferences and the file field is a default. But if a user really wants to, he can remove it from that list.

stefan-kolb:
Ok, I fixed or removed all tests now. Only one ignored test is remianing in logic, which will be fixed by @matthiasgeiger in his custom entries PR?! The GUI tests still remian ignored.

src/test/java/net/sf/jabref/logic/xmp/XMPUtilTest.java
```diff
@@ -1205,16 +1204,15 @@ public void testCommandLineSinglePdf() throws IOException, TransformerException,
...
      */
     @Test
-    @Ignore
     public void testCommandLineByKey() throws IOException, TransformerException {
 
         File tempBib = File.createTempFile("JabRef", ".bib");
         try (BufferedWriter fileWriter = Files.newBufferedWriter(tempBib.toPath(), StandardCharsets.UTF_8)) {
             fileWriter.write(t1BibtexString());
             fileWriter.write(t2BibtexString());
+            fileWriter.close();
```
lenhard: Why do you need the close statement here? `fileWrite` is opened in a try-with-resources, so it will be closed at the end.

src/test/java/net/sf/jabref/logic/xmp/XMPUtilTest.java
```diff
@@ -1205,16 +1204,15 @@ public void testCommandLineSinglePdf() throws IOException, TransformerException,
...
 
             { // First try canh05
```
lenhard: Not your change, but why is this bracket opening in a new line? Can you fix the formatting in this method?

stefan-kolb: The bracket below? This is some old stuff. I already forgot what this does in Java...

stefan-kolb:
@lenhard Thanks for the review, some good input here!

src/main/java/net/sf/jabref/logic/openoffice/OOBibStyle.java
```diff
@@ -149,12 +149,11 @@ public OOBibStyle(File styleFile, LayoutFormatterPreferences prefs,
...
     }
 
-    public OOBibStyle(String resourcePath, LayoutFormatterPreferences prefs)
-            throws IOException {
+    public OOBibStyle(String resourcePath, LayoutFormatterPreferences prefs) throws IOException {
         this.prefs = Objects.requireNonNull(prefs);
         this.encoding = StandardCharsets.UTF_8;
         setDefaultProperties();
-        initialize(OOBibStyle.class.getResource(resourcePath).openStream());
+        initialize(OOBibStyle.class.getResourceAsStream(resourcePath));
```
simonharrer: be aware that the inputstream is null if the resource cannot be found, resulting in a NPE.

src/test/java/net/sf/jabref/logic/xmp/XMPUtilTest.java
```diff
@@ -1205,51 +1204,49 @@ public void testCommandLineSinglePdf() throws IOException, TransformerException,
...
+
+        // Now try OezbekC06
+        try (ByteArrayOutputStream s = new ByteArrayOutputStream()) {
+            System.setOut(new PrintStream(s));
+            try {
+                XMPUtilMain.main(new String[]{"OezbekC06", tempBib.getAbsolutePath(), pdfFile.getAbsolutePath()});
+            } finally {
+                System.setOut(sysOut);
             }
         }
```
simonharrer: For something like this, I would prefer to create a utility method that can be called. Should be useful on other occasions as well. You can pass in a lambda function which will be called from the utility method, and the method could simply return the value of the sysout as a String.

src/main/java/net/sf/jabref/logic/openoffice/OOBibStyle.java
stefan-kolb: Is this any different than before?

src/test/java/net/sf/jabref/logic/importer/fileformat/BibtexParserTest.java
```diff
@@ -1220,38 +1206,23 @@ public void parsePreservesMultipleSpacesInFileField() throws IOException {
...
-     * @author Uwe Kuehn
-     * @author Andrei Haralevich
-     */
     @Test
-    @Ignore("Ignoring, since the parser is not responsible for fixing the content. This should be done later")
-    public void parseRemovesNewlineInFileField() throws IOException {
-        ParserResult result = BibtexParser.parse(new StringReader("@article{canh05,file = {ups \n\tsala}}"),
+    public void parsePreservesNewlineInAbstractField() throws IOException {
+        ParserResult result = BibtexParser.parse(new StringReader("@article{canh05,abstract = {ups \nsala}}"),
                 importFormatPreferences);
```
simonharrer: Would it make sense to create a BibtexParser.parse(string, prefs) method?

src/test/java/net/sf/jabref/logic/xmp/XMPUtilTest.java
```diff
@@ -1205,51 +1204,49 @@ public void testCommandLineSinglePdf() throws IOException, TransformerException,
...
         } finally {
-            if (!tempBib.delete()) {
-                System.err.println("Cannot delete temporary file");
+            System.setOut(sysOut);
+        }
+
+        // PDF should be annotated:
+        List<BibEntry> l = XMPUtil.readXMP(pdfFile, xmpPreferences);
+        Assert.assertEquals(1, l.size());
+        assertEqualsBibtexEntry(t1BibtexEntry(), l.get(0));
```
simonharrer: I do not like to assert the size of a list, because the result would simply say that the count mismatches, not which elements are in the list. Why not use the assertEqualsBibtexEntryList method if there is some which compares a list containing t1BibtexEntry with l?

src/test/java/net/sf/jabref/logic/xmp/XMPUtilTest.java
```diff
@@ -1205,51 +1204,49 @@ public void testCommandLineSinglePdf() throws IOException, TransformerException,
...
+                XMPUtilMain.main(new String[]{"OezbekC06", tempBib.getAbsolutePath(), pdfFile.getAbsolutePath()});
+            } finally {
+                System.setOut(sysOut);
             }
         }
+
+        // PDF should be annotated:
+        l = XMPUtil.readXMP(pdfFile, xmpPreferences);
+        Assert.assertEquals(1, l.size());
+        assertEqualsBibtexEntry(t2BibtexEntry(), l.get(0));
```
simonharrer: See above.

src/test/java/net/sf/jabref/logic/xmp/XMPUtilTest.java
stefan-kolb: To be honest, I didn't want to refactor the tests but just get them running again.

src/test/java/net/sf/jabref/logic/xmp/XMPUtilTest.java
```diff
@@ -1205,51 +1204,49 @@ public void testCommandLineSinglePdf() throws IOException, TransformerException,
...
+
+        if (!tempBib.delete()) {
+            // TODO: file is still locked
+            System.err.println("Cannot delete temporary file");
+        }
```
stefan-kolb: For some reason the file cannot be deleted and is still locked but I dunno where yet (and don't really care).

src/test/java/net/sf/jabref/logic/importer/fileformat/BibtexParserTest.java
stefan-kolb: I dunno, amybe @lenhard knows.

lenhard: Well, it is what the `BibtexParser.parse` method does internally. If you parse multiple times subsequently, there might be a little performance impact, if you parse only once then it does not matter.

src/test/java/net/sf/jabref/logic/xmp/XMPUtilTest.java
tobiasdiez: Assert.assertEquals(Collections.singleton(tlBibtexEntry()), l) should work (and please rename `l` to something better :-)

src/main/java/net/sf/jabref/cli/XMPUtilMain.java
```diff
@@ -176,7 +176,7 @@ private static void usage() {
         System.out.println("  xmpUtil <bib> <pdf>");
         System.out.println("");
         System.out
-        .println("To report bugs visit http://jabref.sourceforge.net");
+        .println("To report bugs visit https://github.com/JabRef/jabref/issues");
```
koppor: `https://issues.jabref.org`, pleae

src/main/java/net/sf/jabref/logic/openoffice/OOBibStyle.java
koppor: Maybe we can wrap it inside `Objects.requireNonNull()` and also add some `Objects.requireNonNull` for `resourcePath` and `prefs` before lin 153

src/test/java/net/sf/jabref/logic/importer/fileformat/BibtexParserTest.java
```diff
@@ -786,19 +786,6 @@ public void parseWarnsAboutUnmatchedContentInEntry() throws IOException {
     }
 
     @Test
-    @Ignore("Ignoring because this is an edge case")
```
koppor: I would like to know, what's happening here. Can the test be adapted to the current behavior and possibly commented, what should happen?

I would need that if someone comes over and wants to replace our parser 😇

src/test/java/net/sf/jabref/logic/layout/LayoutTest.java
```diff
@@ -86,15 +82,6 @@ public void testHTMLChar() throws IOException {
     }
 
     @Test
-    @Ignore
```
koppor: Why not adapting it to the current behavior? I think, we drop `<br>` now, which is fine. Why not leaving the additional test?

src/test/java/net/sf/jabref/logic/xmp/XMPUtilTest.java
koppor: Yeah, we wanted to use a canonical string representation for the bib entries.

The current implementation of BibEntryAssert also relies on `equals` so it is very much OK to use the code @tobiasdiez proposed.

src/test/java/net/sf/jabref/logic/xmp/XMPUtilTest.java
koppor: Why not just calling `deleteOnExit()`? This would prevent these kind of issues.

src/test/java/net/sf/jabref/logic/xmp/XMPUtilTest.java
koppor: OK for leave as is - I copied @simonharrer's comment to https://github.com/koppor/jabref/issues/6.

src/test/java/net/sf/jabref/logic/importer/fileformat/BibtexParserTest.java
stefan-kolb: It just doesn't work right now as @lenhard says. So there is no sense to test it at the moment.

src/test/java/net/sf/jabref/logic/layout/LayoutTest.java
stefan-kolb: If that's really what we want to achieve here \n\n -> ' ' 
I'm not sure about this that's why I removed the test.

stefan-kolb: In other words: That is the current behavior, yes. But should it really be the correct behavior? In my opinion we should only test correct behavior not as is behavior. As the correct behavior for example <br> is not working right now I cannot really test it.

src/test/java/net/sf/jabref/logic/importer/fileformat/BibtexParserTest.java
koppor: JabRef returns one entry, which pretty good. I would like to have this checked in tests if we adress it.

The test has to be adapted accordingly.

Result is 

`@article{test,
  author = {author bracket #too##much#}
}`

Would it be OK if you include this as expected case, even if the `}` gets lost...

This documents the drawback of JabRef.

OK, additional documentation at https://help.jabref.org/en/Bibtex would be better, but none of us has time for that.

src/test/java/net/sf/jabref/logic/layout/LayoutTest.java
koppor: I think, we decided to remove newlines in non-multiline fields. Is it https://github.com/JabRef/jabref/pull/357?

Even if that is not the desired behavior, I would add `// TODO: The correct behavior is` to indicate that someone should work on it.

Alternative: open an issue on GitHub. - I am currently optimizing (i) to keep the number of opened issues low and (ii) to have documentation close to the code and not spread around.

src/test/java/net/sf/jabref/logic/xmp/XMPUtilTest.java
Siedlerchr: A better way to use a TemporaryFolder rule and create the file there.  Then junit keeps track of deleting it

src/test/java/net/sf/jabref/logic/xmp/XMPUtilTest.java
stefan-kolb: Does not work right now author is not normailzed or so:
Expected  
```
[@article{canh05,
  author = {Crowston, K. and Annabi, H. and Howison, J. and Masango, C.},
  booktitle = {Hawaii International Conference On System Sciences (HICSS)},
  owner = {oezbek},
  timestamp = {2006.05.29},
  title = {Effective work practices for floss development: A model and propositions},
  url = {http://james.howison.name/publications.html},
  year = {2005}
}]
```
Actual  
```
[@article{canh05,
  author = {K. Crowston and H. Annabi and J. Howison and C. Masango},
  booktitle = {Hawaii International Conference On System Sciences (HICSS)},
  owner = {oezbek},
  timestamp = {2006.05.29},
  title = {Effective work practices for floss development: A model and propositions},
  url = {http://james.howison.name/publications.html},
  year = {2005}
}]
```

I would skip this for now.

src/main/java/net/sf/jabref/logic/openoffice/OOBibStyle.java
stefan-kolb: Added some requireNonNulls.

Issue #2343, opened 2016-12-07
ISBN Generator output does not follow convention for author names


duelle:
JabRef 3.7
Archlinux Kernel 4.8.12-2
Java 1.8.0_112 (OpenJDK)

The results of the ISBN Generator are not following the conventions for author names which leads to references that cannot be resolved in LaTeX.

Steps to reproduce:
1. In menu: BibTeX -> New Entry
2. Change ID type to "ISBN"
3. Enter for ID: 0321331885
4. Press "Generate"
5. The author field content is not correctly formatted:
Result: 
```David Steinberg, Frank Budinsky, Marcelo Paternostro, Ed Merks```
Should be: 
```Steinberg, David and Budinsky, Frank and Paternostro, Marcelo and Merks, Ed```

The result of the request http://www.ebook.de/de/tools/isbn2bibtex?isbn=0321331885 is:
```
@book{9780321331885,
  Author = {David Steinberg, Frank Budinsky, Marcelo Paternostro, Ed Merks},
  Title = {EMF. Eclipse Modeling Framework},
  Publisher = {Addison Wesley},
  Year = {2009},
  Date = {2009-01-12},
  PageTotal = {XXIX},
  EAN = {9780321331885},
  ISBN = {0321331885},
  URL = {http://www.ebook.de/de/product/6903954/david_steinberg_frank_budinsky_marcelo_paternostro_ed_merks_emf_eclipse_modeling_framework.html}
}
```
The entry above is just an example. I tested this with several other books having mutliple authors with a similar outcome.
Obviously the JabRef implementation is not the issue but the service that is hardcoded into the application.

stefan-kolb:
So this is a problem of the returned BibTeX of ebook.de?! The only thing we could do is to run a normalization on the returned data. This is one thing you should be able to do manually as well to fix this issue. 

![image](https://cloud.githubusercontent.com/assets/2141507/20969632/2c5ca426-bc8a-11e6-9bae-4468e0620ef6.png)

As you can see the BibTeX key is also not formated as expected. 

**For me, the question here is:**
@JabRef/developers Should we add a middleware after the fetchers to fix such issues (BibTeX key generation and other normalization) or should this be the responsibility of the user?

koppor:
Responsibility of the fetcher. We have `net.sf.jabref.logic.importer.SearchBasedParserFetcher#doPostCleanup`

duelle:
@stefan-kolb Ah, good to know, thank you for the hint!

[...]

PR #2374, opened 2016-12-13
Include https://bibtex.chimbori.com/ as fallback for the ebook.de fetcher

koppor:
The fetcher offered by ebook.de does not offer bibtex for all valid ISBNs. This PR adds https://bibtex.chimbori.com/ as fallback for the ebook.de fetcher

Long discussion at: https://github.com/JabRef/jabref/issues/684#issuecomment-266541507

I created a meta IsbnFetcher and two separate fetchers for ebook.de and chimbori. I readded the URL field as this is the deal with both ebook.de and chimbori.

Fixes https://github.com/JabRef/jabref/issues/684


src/main/java/net/sf/jabref/logic/importer/fetcher/IsbnFetcher.java
```diff
@@ -3,68 +3,53 @@
...
-        URIBuilder uriBuilder = new URIBuilder("http://www.ebook.de/de/tools/isbn2bibtex");
-        uriBuilder.addParameter("isbn", identifier);
-        return uriBuilder.build().toURL();
-    }
+        this.ensureThatIsbnIsValid(identifier);
 
-    @Override
-    public Parser getParser() {
-        return new BibtexParser(importFormatPreferences);
+        IsbnViaEbookDeFetcher isbnViaEbookDeFetcher = new IsbnViaEbookDeFetcher(importFormatPreferences);
```
tobiasdiez: I don't like the architecture that a fetcher controls other fetcher. All of them should be completely independent and their working-together should be controlled at a higher level. Moreover, I would allow the user to choose between "ISBN (ebook.de)" and "ISBN (Chimbori/Amazon)", or is the quality of ebook's metadata really superior?

src/main/java/net/sf/jabref/logic/importer/fetcher/IsbnViaChimboriFetcher.java
```diff
@@ -0,0 +1,87 @@
...
+    public Optional<BibEntry> performSearchById(String identifier) throws FetcherException {
+        if (StringUtil.isBlank(identifier)) {
+            return Optional.empty();
+        }
+
+        this.ensureThatIsbnIsValid(identifier);
+
+        HttpResponse<String> postResponse;
+        try {
+            postResponse = Unirest.post("https://bibtex.chimbori.com/isbn-bibtex")
```
tobiasdiez: We should rewrite IdBasedParserFetcher so that it also supports POST actions (and maybe uses Unirest for it).

src/main/java/net/sf/jabref/logic/importer/fetcher/IsbnViaEbookDeFetcher.java
```diff
@@ -0,0 +1,45 @@
...
+    public URL getURLForID(String identifier) throws URISyntaxException, MalformedURLException, FetcherException {
+        this.ensureThatIsbnIsValid(identifier);
+        URIBuilder uriBuilder = new URIBuilder("http://www.ebook.de/de/tools/isbn2bibtex");
+        uriBuilder.addParameter("isbn", identifier);
+        return uriBuilder.build().toURL();
+    }
+
+    @Override
+    public void doPostCleanup(BibEntry entry) {
+        // We MUST NOT clean the URL. this is the deal with ebook.de
```
tobiasdiez: I don't see that we have such an agreement with ebook.de

src/main/java/net/sf/jabref/logic/importer/fetcher/IsbnFetcher.java
koppor: I think, it should be transparent to the user, so that he does not need think, what fetcher he wants to use. We could also offer all three fetchers to the user so that if he really wants to choose.

It is the deal with Chimbori that we use ebook.de first to reduce load on his server.

src/main/java/net/sf/jabref/logic/importer/fetcher/IsbnViaChimboriFetcher.java
koppor: Not sure whether it is worth the effort - I think, we currently have only one exception for that.

koppor:
Update: Fixes #2343

src/main/java/net/sf/jabref/logic/importer/fetcher/AbstractIsbnFetcher.java
```diff
@@ -0,0 +1,37 @@
...
+import net.sf.jabref.logic.help.HelpFile;
+import net.sf.jabref.logic.importer.FetcherException;
+import net.sf.jabref.logic.importer.IdBasedParserFetcher;
+import net.sf.jabref.logic.importer.ImportFormatPreferences;
+import net.sf.jabref.logic.importer.Parser;
+import net.sf.jabref.logic.importer.fileformat.BibtexParser;
+import net.sf.jabref.logic.l10n.Localization;
+import net.sf.jabref.logic.util.ISBN;
+
+public abstract class AbstractIsbnFetcher implements IdBasedParserFetcher {
```
tobiasdiez: Here IdBasedFetcher should be sufficient and then IsbnViaEbookDeFetcher implements IdBasedParserFetcher (in addition to AbstractIsbnFetcher)

koppor: In IsbnViaChimboriFetcher, I need both `doPostCleanup()` and `getParser()` which are not offered by `IdBasedFetcher`.

Don't want to introduce a new intermediate class...

manastungare:
> It is the deal with Chimbori that we use ebook.de first to reduce load on his server. The deal was made in personal emails I exchanged with the author of Chimbori

This was our conversation regarding the version that I had self-hosted, before I set up the dedicated service at bibtex.chimbori.com. Feel free to send *all* your traffic to this new endpoint!

PR #2384, opened 2016-12-14
[FX] Use native file dialog

tobiasdiez:

### DIFF
--- a/src/main/java/net/sf/jabref/gui/FileDialog.java
+++ b/src/main/java/net/sf/jabref/gui/FileDialog.java
@@ -4,60 +4,41 @@ import java.awt.Component;
 import java.io.File;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.EnumSet;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.FutureTask;
 import java.util.stream.Collectors;
 
 import javax.swing.JFileChooser;
-import javax.swing.JOptionPane;
-import javax.swing.filechooser.FileFilter;
-import javax.swing.filechooser.FileNameExtensionFilter;
+
+import javafx.application.Platform;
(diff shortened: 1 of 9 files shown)


src/main/java/net/sf/jabref/gui/FileDialog.java
```diff
@@ -200,21 +176,23 @@ public FileDialog updateWorkingDirPref() {
...
-
-            if (!extensions.isEmpty() && !fileChooser.accept(file)) {
-                return Optional.of(Paths.get(file.getPath() + extensions.iterator().next().getFirstExtensionWithDot()));
-            }
-
-            return Optional.of(file.toPath());
-        }
-        return Optional.empty();
+        configureFileChooser();
+        return runInJavaFXThread(() -> Optional.ofNullable(fileChooser.showSaveDialog(null)).map(File::toPath));
```
koppor: Please document, why `null` is passed here. We wanted to avoid passing `null`, but it seems the API required it.

src/main/java/net/sf/jabref/gui/FileDialog.java
```diff
@@ -154,43 +129,44 @@ public FileDialog updateWorkingDirPref() {
...
     public Optional<Path> showDialogAndGetSelectedFile() {
-        fileChooser.setDialogType(JFileChooser.OPEN_DIALOG);
-
-        if (showDialogAndIsAccepted()) {
-            return Optional.of(fileChooser.getSelectedFile().toPath());
-        }
-
-        return Optional.empty();
+        configureFileChooser();
+        return runInJavaFXThread(() -> Optional.ofNullable(fileChooser.showOpenDialog(null)).map(File::toPath));
```
koppor: Please add a comment, why `null` is passed

koppor:
Besides minor comments, LGTM :+1:. Also tested locally with file/open and Quality/find unlinked files

src/main/java/net/sf/jabref/gui/FileDialog.java
tobiasdiez: Yes, thats the only way to call the showSaveDialog method. Sorry, but I'm not in the mood to add comments for such things as "null" passed 😸 .

Issue #2391, opened 2016-12-17
Erratic behavior when clicking on "Get fulltext"


avaucher:
JabRef development version on Linux (and probably other platforms)

In some cases, the behavior when clicking on "Get fulltext" is erratic or, at least, unexpected and/or misleading.
Two cases are explained below.

## Steps to reproduce:

First, in "File -> Database properties", set "General file directory" to the directory with your files to link. 

### Case A: When multiple entries are selected.

1. Have two bibtex entries with Bibtexkeys "a" and "b"
2. Create corresponding files "a.pdf" and "b.pdf" in the directory chosen before.
3. Double-click on "a" to select it and show the editor view
4. In the editor view, go to the "General" tab.
5. With Ctrl, click on the entry "b" so that both "a" and "b" are selected, and the editor view still shows entry "a"
6. Click on "Get fulltext"

Result: Both "a.pdf" and "b.pdf" are added to entry "a" !!

### Case B: When adding an entry after performing a search.

0. Create corresponding files "a.pdf" and "b.pdf" in the directory chosen before.
1. Have a bibtex entries with Bibtexkeys "a", containing the text "AAA"
2. Select entry "a"
3. Perform a search for "AAA". Only the entries containing "AAA" are listed.
4. Create a new entry. In the editor view, give it the bibtexkey "b"
5. Under the "General" tab, click on "Get fulltext"

Result: Instead of linking the file "b.pdf", it will link the file "a.pdf" !!

koppor:
When #2270 is solved, this issue also gets solved.

koppor:
The fix provided by @avaucher at https://github.com/JabRef/jabref/issues/2392 also provides a solution, which works, because the action of the button is local to the entry editor.

PR #2392, opened 2016-12-17
When clicking on "Get fulltext" in the editor view, link the files for the entry being edited instead of looking for the selected entries

avaucher:
Bug fix for issue number 2391.

When clicking on "Get fulltext" from the editor view, the autoSetLinks function should be performed only to the entry being edited.
Before this pr, the autoSetLinks function uses the selected entries, which can causes problems in different cases (See issue number #2391 for details).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 ### Changed
 
 ### Fixed
+- Fixed [#2391](https://github.com/JabRef/jabref/issues/2391): Clicking on "Get Fulltext" button sets links correctly for the entry being edited.
 
 ### Removed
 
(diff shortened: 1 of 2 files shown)


tobiasdiez:
Thank you very much for your contribution 👍. Your changes look good to me. Could you please add a changelog entry in https://github.com/JabRef/jabref/blob/master/CHANGELOG.md and then we can merge your PR!

matthiasgeiger:
I am browsing from my mobile so I can't check my self... Is this action also available via the menu when no entry editor is open? If yes this change might break this functionality...

koppor:
These are different actions. We have "Quality" - "Lookup fulltext documents". This only does an online search. At "Get fulltext", first, a local lookup is made.

Thus, this quick fix is OK.

In the long run, we have to work on https://github.com/koppor/jabref/issues/192

koppor:
@avaucher  Thank you for the quick solution 👍

PR #2396, opened 2016-12-17
[WIP] Introduce explicit list of ALL_BIBLATEX_ONLY_FIELDS

### LINKED ISSUE #2390: Integrity check shows wrongfully that institution is a biblatex only element
JabRef version JabRef 3.8
windows 10 10.0 amd64 
Java 1.8.0_112

Steps to reproduce:

1. Have DB with bibtex mode and a techreport entry
2. Integrity check reports that institution is only for biblatex (proof below)
3. But this is wrong, see https://de.wikipedia.org/wiki/BibTeX

image

koppor:
This fixes #2390. The last implementation used `EntryConverter.FIELD_ALIASES_LTX_TO_TEX`. This implementation is based on `BibtexEntryTypes` and `BiblatexEntryTypes`. 

## Issue
`JabRefPreferences.CUSTOM_TAB_NAME_num_` should be used to really determine the fields to check. This, however, is really a **huge** architectural rework.

It has to be checked, in following (seldom) situation:

- Field X is defined in JabRef's BiblatexEntryTypes
- Field X is added by the user to a general tab

JabRef now complains that X is Biblatex only.

## Further changes
 * ISBN and ISBN are now BibTeX fields, too.
   - Reason: Even though, both are not listed in [WikiPedia](https://en.wikipedia.org/wiki/BibTeX) and the [lncs style](ftp://ftp.springer.de/pub/tex/latex/llncs/latex2e/splncs03.bst), I see them so hard tied to a book and article, that I would add it. In the case of a DOI, JabRef (more or less) enforces it by offering that field in the "General" tab. Moreover, [it is part of plainnat.bst](http://tex.stackexchange.com/a/52046/9075) and [natbib](https://www.ctan.org/pkg/natbib) is a popular package.
 * `FieldName.COMMENT` instead of `FieldName.COMMENTS` (affects OaiImporter only)
 * Preferences: Use FieldName constants at `setLanguageDependentDefaultValues()`

### DIFF
--- a/src/main/java/net/sf/jabref/logic/integrity/NoBibtexFieldChecker.java
+++ b/src/main/java/net/sf/jabref/logic/integrity/NoBibtexFieldChecker.java
@@ -1,6 +1,5 @@
 package net.sf.jabref.logic.integrity;
 
-import java.util.Arrays;
 import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
@@ -10,24 +9,19 @@ import net.sf.jabref.logic.l10n.Localization;
 import net.sf.jabref.model.entry.BibEntry;
 import net.sf.jabref.model.entry.BibLatexEntryTypes;
 import net.sf.jabref.model.entry.BibtexEntryTypes;
-import net.sf.jabref.model.entry.FieldName;
+import net.sf.jabref.model.entry.InternalBibtexFields;
 
 /**
  * This checker checks whether the entry does not contain any field appearing only in BibLaTeX (and not in BibTeX)
  */
 public class NoBibtexFieldChecker implements Checker {
 
-
     private List<String> getAllBiblatexOnlyFields() {
         Set<String> allBibtexFields = BibtexEntryTypes.ALL.stream().flatMap(type -> type.getAllFields().stream()).collect(Collectors.toSet());
-
-        // fields as set at net.sf.jabref.preferences.JabRefPreferences.setLanguageDependentDefaultValues()
-        final List<String> JABREF_GENERAL_FIELDS =  Arrays.asList(FieldName.CROSSREF, FieldName.KEYWORDS, FieldName.FILE, FieldName.DOI, FieldName.URL, FieldName.COMMENT, FieldName.OWNER, FieldName.TIMESTAMP, FieldName.ABSTRACT, FieldName.REVIEW);
-
         return BibLatexEntryTypes.ALL.stream()
                 .flatMap(type -> type.getAllFields().stream())
                 .filter(fieldName -> !allBibtexFields.contains(fieldName))
-                .filter(fieldName -> !JABREF_GENERAL_FIELDS.contains(fieldName))
+                .filter(fieldName -> !InternalBibtexFields.DEFAULT_GENERAL_FIELDS.contains(fieldName))
                 .sorted()
                 .collect(Collectors.toList());
     }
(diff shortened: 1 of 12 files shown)


src/test/java/net/sf/jabref/logic/integrity/NoBibTexFieldCheckerTest.java
```diff
@@ -22,6 +22,36 @@ public void addressIsNotRecognizedAsBibLaTeXOnlyField() {
     }
 
     @Test
+    public void afterwordIsRecognizedAsBibLaTeXOnlyField() {
+        BibEntry entry = new BibEntry();
+        entry.setField("afterword", "test");
+        IntegrityMessage message = new IntegrityMessage("BibLaTeX field only", entry, "afterword");
+        List<IntegrityMessage> messages = checker.check(entry);
+        assertEquals(messages, Collections.singletonList(message));
```
tobiasdiez: change order of arguments (and maybe inline `messages`)

src/main/java/net/sf/jabref/logic/integrity/NoBibtexFieldChecker.java
```diff
@@ -1,34 +1,43 @@
...
-        SortedSet<String> allBibLaTeXOnlyFields = new TreeSet<>();
-        allBibLaTeXOnlyFields.addAll(EntryConverter.FIELD_ALIASES_LTX_TO_TEX.keySet());
 
-        // file is both in BibTeX and BibLaTeX
-        allBibLaTeXOnlyFields.remove(FieldName.FILE);
+    private List<String> getAllBiblatexOnlyFields() {
+        Set<String> allBibtexFields = BibtexEntryTypes.ALL.stream().flatMap(type -> type.getAllFields().stream()).collect(Collectors.toSet());
+
+        // fields as set at net.sf.jabref.preferences.JabRefPreferences.setLanguageDependentDefaultValues()
+        final List<String> JABREF_GENERAL_FIELDS =  Arrays.asList(FieldName.CROSSREF, FieldName.KEYWORDS, FieldName.FILE, FieldName.DOI, FieldName.URL, FieldName.COMMENT, FieldName.OWNER, FieldName.TIMESTAMP, FieldName.ABSTRACT, FieldName.REVIEW);
```
tobiasdiez: The naming would indicate that it is a static variable, but it is not. Same applies to `ALL_BIBLATEX...` below.
Moreover, if this list is reused somewhere else then please extract it and reuse it (e.g. to FieldName ?)

src/test/java/net/sf/jabref/logic/integrity/NoBibTexFieldCheckerTest.java
koppor: Done

src/main/java/net/sf/jabref/logic/integrity/NoBibtexFieldChecker.java
koppor: Ups. - I don't know where to put it. It is specific to JabRef and can be re-configured by the user. Thus, I should read it from the preferences. Introducing a good preferences object is currently much effort for me. I would say: as soon as a user complains that the integrity check does not work with his customized tabs, I will spend the effort.

Nevertheless, I moved the constant to `InternalFields` and use it in both preferences and the checker.

PR #2399, opened 2016-12-18
[WIP] Use x11 selection

### LINKED ISSUE #2389: JabRef and the X Window clipboard/selections
[JabRef version 3.8~pre2+ds-1 from Debian/experimental. But the issue is not specific to this version.]

The X Window System has usually three "selections": primary/XA_PRIMARY - copying by highlighting with the first mouse button, pasting by clicking the third (middle) button; secondary/XA_SECONDARY (never seen this in action); and clipboard/XA_CLIPBOARD which acts like the clipboard on e.g. Windows (Ctrl-C, Ctrl-V and copy/paste menu items or buttons).

It seems that JabRef only supports the last version (the "clipboard"). This can be seen e.g.
- in the "Copy DOI URL" function which copies the URL into the XA_CLIPBOARD selection (and not (also) into XA_PRIMARY);
- in the "New entry from plain text" dialog, where (first of all neither middle-clicking/XA_PRIMARY nor Ctrl-V/XA_CLIPBOARD in the text box does anything but) using the "Paste" button pastes the contents of the XA_CLIPBOARD selection.

I don't know how to make Java under X11 use the XA_PRIMARY selection as well but I'm sure this is not the first time in the history of Java that this question comes up :)

Cheers,
gregor

koppor:
Fixes #2389.

- [ ] Distinguish between middle click and <kbd>Ctrl</kbd> + <kbd>V</kbd>.

## Current solution

When SELECTION, the "selection buffer" of X11 is available, use it also at copy and paste. This is a quick solution. I cannot test it locally. I only checked that the copy and paste on windows still works.

- [n/a] Manually tested changed features in running JabRef


koppor:
Hint by @mkarg (Thank you!)

One has to use https://docs.oracle.com/javase/8/javafx/api/javafx/scene/input/Clipboard.html. Otherwise, it would not work.

snisnisniksonah:
I looked into it and found this https://bugs.openjdk.java.net/browse/JDK-8088117 and while i couldn't find anything similar for the oracle jdk i couldn't get it to work with that either.
So maybe put this on hold for now?

koppor:
We migrate to JavaFX, so this hack is (i) not working and (ii) there is no work-around. We have to re-investigate as soon as JavaFX can cope with the secondary buffer.

Issue #2403, opened 2016-12-19
Timestamp does not work


mlep:
JabRef 3.8, but also JabRef-3.8.1-dev--snapshot--2016-12-19--master--8cfbc41.jar

java version "1.8.0_66"
Java(TM) SE Runtime Environment (build 1.8.0_66-b17)
Java HotSpot(TM) 64-Bit Server VM (build 25.66-b17, mixed mode)

The timestamp features does not work well:
- a click on `Today` raises an exception (see below)
- a click on `Clear` does not clear the field content
- a double-click in the field sets the timestamp value to the current date (i.e. today). Is it the intended behavior?

```
15:10:15.023 [AWT-EventQueue-0] INFO  net.sf.jabref.logic.importer.OpenDatabase - Opening: /home/fdarboux/Documents/Biblio/biblio.bib
15:10:15.964 [AWT-EventQueue-0] INFO  net.sf.jabref.logic.importer.OpenDatabase - Opening: /home/fdarboux/Documents/Biblio/darboux.bib
15:11:06.100 [AWT-EventQueue-0] ERROR net.sf.jabref.FallbackExceptionHandler - Uncaught exception Occurred in Thread[AWT-EventQueue-0,6,main]
java.time.DateTimeException: Unable to obtain ZonedDateTime from TemporalAccessor: 2016-12-19 of type java.time.LocalDate
	at java.time.ZonedDateTime.from(ZonedDateTime.java:565) ~[?:1.8.0_66]
	at net.sf.jabref.gui.date.DatePickerButton.dateChanged(DatePickerButton.java:57) ~[JabRef-3.8.1-dev--snapshot--2016-12-19--master--8cfbc41.jar:?]
	at com.github.lgooddatepicker.components.DatePicker.zInternalSetLastValidDateAndNotifyListeners(DatePicker.java:846) ~[JabRef-3.8.1-dev--snapshot--2016-12-19--master--8cfbc41.jar:?]
	at com.github.lgooddatepicker.components.DatePicker.zEventTextFieldChanged(DatePicker.java:896) ~[JabRef-3.8.1-dev--snapshot--2016-12-19--master--8cfbc41.jar:?]
	at com.github.lgooddatepicker.components.DatePicker.zInternalSetDateTextField(DatePicker.java:831) ~[JabRef-3.8.1-dev--snapshot--2016-12-19--master--8cfbc41.jar:?]
[...]

PR #2405, opened 2016-12-19
Datepicker fix

matthiasgeiger:
Fix for #2403 - and as I was just touching this stuff again: I replaced our custom `EasyDateFormat` class with a standard `DateTimeFormatter`


koppor:
Documentation at http://help.jabref.org/en/TimeStamp

Does not work with master and v3.8 on my side. v3.7 does not compile any more ("Could not find de.undercouch:citeproc-java:1.0.0-SNAPSHOT"). Does also not work with v3.6. 

The solution is to reset the preferences.

Solution works for me. @matthiasgeiger Thank you for the fix.

mlep:
Does not work for me even after resetting the preferences... (JabRef-3.8.1-dev--snapshot--2016-12-20--master--b655bfd.jar)

lenhard:
@mlep: After resetting the preferences, the timestamp field appears for me and the datepicker works as desired. 

The dev version you mention in your comment is not the most recent one, maybe it did not yet include the fix. Can you try again with JabRef-3.8.1-dev--snapshot--2016-12-20--master--62d55e1.jar

mlep:
Indeed, the calendar is displayed now.
HOWEVER, the `Clear` function does not work as expected: the timestamp field is not cleared. If I click on `Today`, then `Clear` works.

lenhard:
@mlep Indeed! Clear only works if a proper date is already in the field. If you insert something else, such as `abc`, then clear does not work.

@matthiasgeiger Can you please address this? Since this PR is already merged, I guess we need a follow-up PR.

mlep:
Note: for me, it does not clear even if the field content looks good, like:
 ` timestamp     = {2016-10-26},`

Issue #2117, opened 2016-10-04
Entry-based fetchers should respect BibTeX key pattern of JabRef


AEgit:
JabRef 3.7dev--snapshot--2016-10-03--master--565676c
windows 10 10.0 amd64 
Java 1.8.0_101
1. Set default pattern for Bibtex key generation to [auth][year]
2. Generate a new item/article using the DOI (Ctrl+N --> DOI: mydoi --> Generate)
3. The generated bibtex key, has the following format: [auth]_[year], i. e. there's an underscore between the author's name and the year; but there shouldn't be an underscore

EXAMPLE:
Use the following doi: 10.1671/0272-4634(2007)27[316:psgesn]2.0.co;2

It generates "Huttenlocker_2007" as Bibtex key, but you would expect "Huttenlocker2007".

This behaviour only appeared after the changes made to the DOI import, which is now no longer part of the Web search. Is there another option that I haven't set or is this a bug?

zellerdev:
I will have a look at this

koppor:
Could you also set the focus on the input field after CTRL n - or is this
already done? 😇

AEgit:
@zellerdev: No, it appears to be restricted to the DOI fetcher. At least the Google Scholar one doesn't show this unexpected behaviour.

stefan-kolb:
Actually this is no bug. It is the standard behavior. The reason is that the fetcher does not use the BibTeX key pattern to generate a key but just uses the key that is fetched from the web!

stefan-kolb:
Problem here is, if we use the BibTeX key pattern from the preferences we introduce a new dependency or tighter coupling of the fetcher with the preferences...

tobiasdiez:
I think we need to implement a second layer around the fetcher which after-processes the fetched entries (correct cite key, apply cleanup operations, etc. pp).

AEgit:
@stefan-kolb: But how comes this was not the case in earlier versions of Jabref? Furthermore, I'm not sure, whether it is just the key, that is fetched from the web - if I use the DOI fetcher, I ALWAYS get the abovementioned BibTex key format. If I download the BibTex entry for an article manually and then import it into Jabref, the BibTex can be the same as the one fetched by JabRef but it can also differ from it. Is that to be expected?

stefan-kolb:
1. I don't know if this was any different before.
2. The current fetcher grabs the whole BibTeX document from the web.
[...]

PR #2413, opened 2016-12-21
Enable entry generatation based on title

koppor:
This fixes https://github.com/koppor/jabref/issues/179

In the [new entry dialog](http://help.jabref.org/en/BaseFrame#adding-a-new-entry), one can use the id fetcher "Title" to create a bib entry based on the title. It uses the functionality from the DOIFetcher.

Discussion: There is no [merge entries](https://help.jabref.org/en/MergeEntries) dialog shown. The title by the publisher is always used. I would say that this is OK as the title casing of the publisher should take precedence. If the user wants to overwrite it, he has to use the entry editor and paste the title again in the title field.

- [x] Check documentation status (Issue created for outdated help page at [help.jabref.org](https://github.com/JabRef/help.jabref.org/issues)?) -> https://github.com/JabRef/help.jabref.org/issues/147


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 ## [Unreleased]
 
 ### Changed
+- When [adding a new entry](http://help.jabref.org/en/BaseFrame#adding-a-new-entry), one can select "title" to create a full BibTeX entry based ona title.
```
matthiasgeiger: space: "on a"

matthiasgeiger:
LGTM

Regarding the merge entry: This is a more general problem of all IDBasedFetchers - as mentioned by @stefan-kolb in https://github.com/JabRef/jabref/issues/2117#issuecomment-255052260: We currently do not perform the same checks and modifications on fetched entries as we do/did for the normal web fetchers. Thus, there is no check for duplicates.

koppor:
I wanted to say that "BPELscript ..." (search string) gets "{BPELscript}" (returned by the publisher; which is good in this case), but sometimes something strange is returned by the fetcher and the user title (used for searching) might have to be preferred.

PR #2414, opened 2016-12-21
Replace calls of BibtexParser.parse

lenhard:
This is a follow-up to #2139, so that we can finally get the PR into master.

I created a new branch since I have no access rights on the previous one, took over all commits from there, and tried to merge this with the current master. The point is just to get the code into master now. I will not implement additional suggestions from #2139


lenhard:
I double-checked the code and there should be nothing problematic left. Essentially this is just a few replacements of the static `BibtexParser.parse` with `new BibtexParser(...).parse`.

When you merge, please do not squash so that the commits by @oscargus are preserved.

tobiasdiez:
I also extracted the initialization of the parser to the setup method in the tests (as proposed in #2139). Oscar was right; it is not nice to simply mark a method as deprecated and let other people cleanup the dirt - thanks for your work @oscargus and @lenhard.

tobiasdiez:
Waaaaahhhh I clicked on squash & merge. How do I revert and merge as "normal" ??

Issue #2385, opened 2016-12-15
Move file to file directory can overwrite files


davydden:
JabRef version  on 

Steps to reproduce:

1. Have two different files attached to the paper of the same type (i.e. `.pdf`)
2. Do "Move file to file directory" twice and observe that one file is overriden by another

Siedlerchr:
I am having problems to reproduce the error and need some more details on your setup:

I have a database-specific General file directory C:\XXXX\asdfasdf
I have an entry with 2 different pdf files.  Both pdf files are in C:\YYYY\
Bib filie lies in C:\YYY\biblocation 

I execute **only** Cleanup: Move files to file directory + make links relative 
The cleanup moves the files to the target directory (ignoring fiileDirPattern atm)  correctly, but reports that no entry needed a cleanup. 

Did you execute Rename PDF Cleanup, too?

davydden:
> I execute only Cleanup: Move files to file directory + make links relative

cleanup is fine, i was doing it manually in `General` tab of the the entry by right clicking on each .pdf and the `Move file to file directory`.

Siedlerchr:
I think I found one of the problems: When you do a right click in the EntryEditor int he General Tab on the the FileList, it does not correctly move the file nor does it respect  the fileDirPattern. I just figured out, that the EntryEditor ist in fact using a **copy** (!) of the code in cleanup.  When implementing #1899  I was not aware of this fact.... 👊 😕

Siedlerchr:
@davydden  It's been some time, but I finally implemented a solution. The GUI code in the Entry Editor now uses the same functionality as the normal Cleanup does. This means, files are no longer overwritten.
You can test the development version here:
http://builds.jabref.org/moveFileDir/

I tested following cases:
A.pdf
B.pdf
=> Both get moved to fileDir (also respecting fileDir Pattern)
So with `fileDirPattern=\EntryType `you have `\Article\A.pdf and \Article\B.pdf`

Renaming works the same:
Article Author = Some Author 
Year 2011
=> First file selected will be renamed to bibtexkey
=> Second file selected will be just moved to directory 

I  now only have to implement some Preview  GUI dialog, but the core functionality works.

davydden:
@Siedlerchr thank you very much, i will try to test it soon...

lenhard:
We think this is fixed in current master, hence I am closing it.

Feel free to reopen if the problem reappears.

PR #2415, opened 2016-12-21
Fix several File Cleanup + Rename issues 

Siedlerchr:
Separated Move + Rename Cleanup Logic
Move + Rename files in General Tab -> File now use internally the cleanup operations logic
Separated GUI context menu operations for move + cleanup
Rename files no longer moves files, it only renames them.
Files are not overwritten if they exist or have the same name.

- [x] TODO: Apply new targetFileDir in Test 
- [x] TODO: Adapt test finally
- [x] TODO: Rework GUI action code
 
Fix for #2385 and followup from #1899 

Includes Fix for #2454 and #2526 

jabrefgeneralfilefield

grafik

jabrefmovefile


Siedlerchr:
GUI: CleanupPreset Panel + Metadata.getDefaultFileDirectory seems not to check bibtex location somehow

Siedlerchr:
TODO: Find a wy to cleanup only one file

Siedlerchr:
I will investigate this, but also I checked the [biblatex](http://ftp.fau.de/ctan/macros/latex/contrib/biblatex/doc/biblatex.pdf) manual which states:

> An alias for institution, provided for BibTeX compatibility. The institution
> field is used by traditional BibTeX for technical reports whereas the school field
> holds the institution associated with theses. Te biblatex package employs the
> generic field name institution in both cases. See §§ 2.2.2 and 2.3.4.

Siedlerchr:
- [ ] TODO: Add some preview dialog stuff

Siedlerchr:
- [ ] Prevent FileDirPattern subfolder creation on each run if already exists

src/main/java/org/jabref/logic/cleanup/MoveFilesCleanup.java
```diff
@@ -16,64 +23,113 @@
...
+            ParsedFileField field) {
+
+        this(databaseContext, fileDirPattern, fileDirectoryPreferences, prefs);
+        this.singleFileFieldCleanup = field;
     }
 
     @Override
     public List<FieldChange> cleanup(BibEntry entry) {
-        if(!databaseContext.getMetaData().getDefaultFileDirectory().isPresent()) {
+        Optional<Path> firstExistingFileDir = databaseContext.getFirstExistingFileDir(fileDirectoryPreferences);
```
tobiasdiez: I don't like that so much logic is hidden in a cleanup process. In my opinion it would be better to extract as much logic as possible to separate classes and reuse them here and also in the other places (instead of reusing the cleanup).
In more detail I propose to:
- Create a class that wraps around `List<ParsedFileField>` and that takes care of moving all linked files (and providing nice helper methods like `replace(one file, by another one)` 
- Almost everything in the [for loop below](https://github.com/JabRef/jabref/pull/2415/files#diff-697281a5461cd043769c3bbf1331a9deR87) can be extracted to the ParsedFileField class (and be split in smaller helper methods like ParsedFileField.exists())
- The code is a bit complicated since you constantly have to pass the fileDirectoryPreferences around (not your fault!). What you think about passing these information once into `TypedBibEntry.getFiles` as arguments, which are then stored in the ParsedFileField class? Thus the ParsedFileField class actually knowns if the file exists in one of the directories, what its preferred name is, can rename and move itself...

Siedlerchr: Sounds like a good idea. However, as my time is currently limited due to exams I would rather vote to merge it in as it is, as it fixes several important bugs and I would create a followup PR for making the code strucutred

tobiasdiez: Its probably fine... maybe somebody else should also have a look at the code before merge.

src/main/java/org/jabref/gui/externalfiles/MoveFileAction.java
```diff
@@ -13,178 +13,84 @@
...
+ * Action for moving a file that is linked  from an entry in JabRef.
  */
 public class MoveFileAction extends AbstractAction {
 
-    private static final Log LOGGER = LogFactory.getLog(MoveFileAction.class);
-
     private final JabRefFrame frame;
     private final EntryEditor eEditor;
     private final FileListEditor editor;
+    private final CleanupPreferences prefs = Globals.prefs.getCleanupPreferences(new JournalAbbreviationLoader());
```
Siedlerchr: I now pass the prefs directly

lenhard:
I had a look at the code, which in general is good, and fixed a few minor issues directly (missing getters/setters that were missing well before this PR, but this is a good time for a fix). I also played around a little with the UI and tried to find errors, but was unsuccessful.

In the spirit of progress and as suggested by @Siedlerchr  https://github.com/JabRef/jabref/pull/2415#discussion_r103486029 I am merging now directly. More far fetching suggestions by @tobiasdiez should be addressed in a follow-up PR, as discussed above.

PR #2429, opened 2016-12-23
Check Integrity: No BibTeX or BibLaTeX field

grimes2:
This integrity check detects non BibTeX/BibLaTeX fields. This gives the opportunity to check for possibly wrong spelling.


Siedlerchr:
What about custom entry fields?
They would then be reported every time


Am 23.12.2016 9:44 vorm. schrieb "Jürgen Lange" <notifications@github.com>:

koppor:
Any news on this? 💤 

It would be good if the recommendation of @Siedlerchr would be implemented.

grimes2:
Custom entry fields? Do you mean fields that does not belong to an specific entry type (for example: field journal in type book). This is a good idea, but maybe for another checker.

Siedlerchr:
@grimes2  You can either have a) custom entry types (for example  `@MyCustomEntrType`) 
and when you check BibTex -> Custom Entry Types, you can add custom fields. For example you can add a field MyCustomInternalID or whatever as a (required) field for article.

You should take care that you do not report them.  And another thing. When you sometimes import data from other publishers/portals set, you can have custom fields, too. For example, if you import a bib file  from ACM Portal, you typically have a field like  acmid. I am not sure if this should be reported then or not...

[Example from ACM](http://dl.acm.org/citation.cfm?id=2392777.2392781)

koppor:
Help: https://help.jabref.org/en/CustomEntryTypes

Am 20.01.2017 20:09 schrieb "Christoph" <notifications@github.com>:

grimes2:
Ahh! Good point. But I think, most style files (bst) don't print custom entry fields. So these fields should maybe be reported. I'm thinking (breeding) in the moment about things, that have nothing to do with JabRef. So I would like to pass the implementation to someone else.

stefan-kolb:
Closing this issue for now. I think we need to rediscuss this for an addition. Currently, I don't see the immediate use case for this.

koppor:
I stumbled over this while searching for something else. I like the motivation of "typos". The use case is that a JabRef-pro opens a bib file authored **by someone else** not using JabRef. That person might have introduced typos: `authors`, `autor`, `jahr`, ... -- and not knowing it... This checker would report it.

We have a list of all known bibtex and biblatex fields. The checker should "just" use these hard-coded fields (and not introduce another set of fields in the code). -- Then, it is IMHO good to go.

PR #2435, opened 2016-12-27
Detect unused journals by checking whether there are abbreviations fo…

simonharrer:
…r the journals available.

I have 500 bibtex entries, out of which 125 are journal articles. With the article bibtex type, there are a few issues:
- It happens that an article should actually have the inproceedings type, but Google Scholar generated it wrong, and it was not corrected and checked.
- That the known journal is misspelled. This is hard to detect manually.

As a solution, I propose to use an integrity check which displays all the articles which have journals that are not in the abbreviation repository. This has helped me to determine misspellings, wrongly classified articles, and improve the abbreviations list overall as well.

This integrity check would also advertise our journal abbreviation feature more, and encourage people to contribute to journal abbreviations lists. Maybe, we should really create a cloud based service for that?

What do you think?


matthiasgeiger:
> What do you think?
I think you should not spend your time in improving your bibliography and not in improving the tool which is uaed to organize it but in producing some more actual Content :wink:

lenhard:
The code looks fine. Since I use neither the integrity checks nor journal abbreviation lists, it is a bit hard for me to tell how aggressive we should tune the warnings. I guess the more we increase the amount of warnings, the more we should think about making the integrity check more configurable or introducing levels of severity of warnings, etc. etc.

The travis builds seemed to have failed because of the GUI tests, so I restartet them. If they succeed, this is good to go from my side.

tobiasdiez:
Some of the checker unit tests fail. Could you please fix them and then directly merge? Thanks.

stefan-kolb:
Ping 😄

lenhard:
I fixed the remaining issues in the PR and am merging this now.

PR #2437, opened 2016-12-28
Group all checker which only check the value of one field

tobiasdiez:
Most of the integrity checker only work on one field and thus I introduced a new abstract class which groups the common functionality.
(I know we don't want pure refactor PRs anymore, sorry for that) 


simonharrer:
To me, the code is worse than before. Less clear, less self-contained. And just for saving a few lines of code which are very simple. Just my 2 cents.

tobiasdiez:
Thanks @simonharrer for your feedback. Besides "saving a few lines of code" I had the following motivation in mind:
- Invoke these FieldChecker as validators in the entry editor (i.e. show errors directly while editing an entry)
- Move from strings to valued types. Most of the checkers work with strings, instead of using data objects (for example, they use regular expressions to match pages instead of parsing the string to a Page object and let this class check if it's a normalized page string; the ISSN checker is an exception and in a style I prefer). My hope was that all this parsing can also be moved to the parent FieldChecker class, so that each checker only operates on a parsed object (like `ISSN`). 

But probably all these things can also be accomplished without the FieldChecker as abstraction layer. @JabRef/developers if you think that it over-complicates things, just close this PR.

simonharrer:
Ok, I see. Thanks for clarification. The PR makes more sense to me now. :-)

In that case, I would vote for a FieldChecker interface that is implemented by each of these classes you refactored. Then, another class can bridge between a Checker and a FieldChecker like an adapter. The FieldCheckers can then be easily used within the entry editor.

Working with generics could be the way to go for your second point, but I am unsure how this will look. The idea behind this change could be to use the decorator pattern. On the outside, there is a String FieldChecker, and on the inside an ISSN FieldChecker, for instance.

tobiasdiez:
Thanks @simonharrer for this suggestion! I now implemented the FieldCheckers via the adapter pattern and think it is more clear than it was before. 
Hopefully you finish your thesis quickly so that you have more time for JabRef again 😄

PR #2445, opened 2017-01-02
Escaping of escape symbols in the MetaData

### LINKED ISSUE #2426: Backslashes are repeatedly escaped on save in Content-Selectors
Hello,

JabRef version: JabRef 3.8; linux 4.4.0-57-generic amd64; Java 1.8.0_111 on Ubuntu 16.04
Steps to reproduce:

1. Create a new, empty database (bibtex mode)
2. Add a keyword under manage content selectors for journal: Test {\"U}mlaut
3. Hit Save
4. Check comment in editor: ... Test {\\"U}mlaut ...
5. Restart Jabref
6. Hit Save
7. Check comment in editor: ...Test {\\\\"U}mlaut

So with every save each Backslash is escaped again and again. 

### 1. Save
```
% Encoding: ISO-8859-1

@Comment{jabref-meta: databaseType:bibtex;}

@Comment{jabref-meta: selector_journal:Test {\\"U}mlaut;}
```

### 2. Save
```
% Encoding: ISO-8859-1

@Comment{jabref-meta: databaseType:bibtex;}

@Comment{jabref-meta: selector_journal:Test {\\\\"U}mlaut;}
```

I hope this provides all necassary and relevant information. I hope I haven't missed something. :-)

lenhard:
Fixes #2426 and is ready for review.

# Old discussion
Attempts to fix  #2426.

Is not yet functional, since the issue goes somewhat deeper and affects LaTeX-encoding things in all MetaData, also groups. The problem is line 64 in MetaDataSerializer:
```
 stringBuilder.append(StringUtil.quote(dataItem, MetaData.SEPARATOR_STRING, MetaData.ESCAPE_CHARACTER)).append(MetaData.SEPARATOR_STRING);
```
For every MetaData item, all special characters are quoted, including the quoting symbol (`\`). Thus, if there is a `\` in the MetaData, it is quoted during serialization, resulting in `\\`, which is read as such and requoted on the next serialization, resulting in `\\\\` and so on. 

Unfortunately, unquoting the text during parsing and quoting it again during writing does not help, since a single `\` will simply be lost. E.g.: unquoting `\"U` would result in `"U`, which upon quoting anew would stay `"U`.

The same problem should occur for groups, if I am not mistaken.

I now have the following options to resolve the problem:
 1. Add special handling for content selectors to avoid quoting for them
 2. Change `StringUtil.quote` to avoid quoting the quoting char (`\`)
 3. Add another method in StringUtil that avoids quoting the quoting char and use it for MetaData in general.

I am afraid that changing `StringUtil.quote` has wider implications that I am not aware of., especially relating to groups Hence, there is a need for discussion. 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 
 ### Fixed
 - The formatter for normalizing pages now also can treat ACM pages such as `2:1--2:33`.
+- Backslashes in content selectors are now corretly escaped. Fixes [#2426](https://github.com/JabRef/jabref/issues/2426).
 
 ### Removed
 
--- a/src/main/java/net/sf/jabref/logic/exporter/MetaDataSerializer.java
+++ b/src/main/java/net/sf/jabref/logic/exporter/MetaDataSerializer.java
@@ -61,7 +61,7 @@ public class MetaDataSerializer {
             StringBuilder stringBuilder = new StringBuilder();
             for (String dataItem : metaItem.getValue()) {
                 stringBuilder.append(StringUtil.quote(dataItem, MetaData.SEPARATOR_STRING, MetaData.ESCAPE_CHARACTER)).append(MetaData.SEPARATOR_STRING);
-
+                
                 //in case of save actions, add an additional newline after the enabled flag
                 if (metaItem.getKey().equals(MetaData.SAVE_ACTIONS)
                         && (FieldFormatterCleanups.ENABLED.equals(dataItem)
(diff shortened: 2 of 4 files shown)


tobiasdiez:
I don't understand why unquoting on parse and quoting on writing does not work.
The input `\"U` will be written as `\\\"U` (as in the java string enquoting) and upon parsing be restored as `\"U`. Of course, a single backslash in the bib file gets lost but JabRef never should write just one backslash but always two. 

As for the StringUtil.quote method, I think, you can edit it directly. If I remember correctly, then its purpose is to ensure that strings are properly encoded for writing them in the bib-file (this applies to all kind of metadata - groups are nothing special). Probably, it would be good to move this method actually to the metadata serializer / parser.

tobiasdiez:
Build is fine expect a fetcher problem -> Merge

PR #2459, opened 2017-01-12
Fix conversion of tilde n

### LINKED ISSUE #2458: "\~" Escape Sequence Does Not Display Correctly in Entry Table
JabRef 3.8.1 on Linux Mint 17.3

The Latex backslash tilde escaped diacritic "\\\~" does not display correctly in the entry table. So, for example, the bibtex entry of "Monta\\\~{n}a" displays as "Monta na". Note the space in the middle of the word. The "\\" escape character seems to be ignored and the "~" character treated as a non-breaking space.

Peter

lenhard:
And although I said that I would not work on JabRef for the next two weeks, here is a PR. This stuff is addictive and I need help :-/

Fixes #2458. Author name formatting in the main table is handled different from the formatting of the remaining columns and we simply forgot to perform the unicode conversion also for the author column.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,10 +15,10 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 
 ### Fixed
 - The formatter for normalizing pages now also can treat ACM pages such as `2:1--2:33`.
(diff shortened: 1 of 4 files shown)


Siedlerchr:
Failing tests are again the damn fetchers...
However, I merge itin

PR #2464, opened 2017-01-13
Fix conversion of "'n" and "\'{n}" from LaTeX to Unicode

lenhard:
This is a follow up for fixing  #2458

The combinations `\'n` and `\\'{n}` are converted from LaTeX to different symbols in Unicode.

This was really hard to track down. It was not a problem in our conversion maps, but in `LatexToUnicode`. The ultimate reason is that this code is just an utter mess of hacks and I could only fix this by adding another ugly map on top of this heap.

We have already identified this class to cause performance issues and, as you can see here, it is very difficult to maintain. We should consider removing the functionality and rewriting it from scratch or (preferably) replace it with an external library. After all, converting a LaTeX string to Unicode seems sufficiently well-defined that there is some library for this out there.

### DIFF
--- a/src/test/java/net/sf/jabref/logic/layout/format/LatexToUnicodeFormatterTest.java
+++ b/src/test/java/net/sf/jabref/logic/layout/format/LatexToUnicodeFormatterTest.java
@@ -92,4 +92,10 @@ public class LatexToUnicodeFormatterTest {
     public void testTildeN () {
         assertEquals("Montaña", formatter.format("Monta\\~{n}a"));
     }
+
+    @Test
+    public void testApostrophN () {
+        assertEquals("Maliński", formatter.format("Mali\\'{n}ski"));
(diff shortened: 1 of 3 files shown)


src/test/java/net/sf/jabref/logic/layout/format/LatexToUnicodeFormatterTest.java
```diff
@@ -92,4 +92,10 @@ public void unknownCommandWithEmptyArgumentIsKept() {
     public void testTildeN () {
         assertEquals("Montaña", formatter.format("Monta\\~{n}a"));
     }
+
+    @Test
+    public void testApostrophN () {
+        assertEquals("Maliński", formatter.format("Mali\\'{n}ski"));
+        assertEquals("Maliŉski", formatter.format("Mali'nski"));
```
tobiasdiez: Only one assert per tests pls...... 😢

PR #2466, opened 2017-01-14
Fixed bug when assigning refs to groups.

milleratotago:
Program hung if user clicked ExpandAll or CollapseAll dialog button when
assigning refs to groups. Also changed CollapseAll to show 1st-level
children of AllEntries node.

Any thoughts? Big fan of JabRef but new to open source, git, etc.  Let me know if anything else is required.

CC @chm052


milleratotago:
Hi Tobias,

Thanks for the feedback--I'm glad the code looks ok to you.

I will do as you say to modify the changelog asap--probably tomorrow.

Jeff Miller

________________________________

koppor:
We merged it and will add the CHANGELOG entry later. We want it to be part of the next release.

Thank you the patch 👍

PR #2474, opened 2017-01-20
[WIP] Use CERMINE as PDF parser

tobiasdiez:
This PR replaces our own PDF parser with [CERMINE](https://github.com/CeON/CERMINE).
In my tests, this library was able to extract (relatively) correct information from a wide variety of articles. It had some problems with books and thesises (what is the plural of a thesis?) through.
As far as I understand it, it uses neural networks that try to analyze the PDF on a structural level (e.g. the title is often placed rather prominently). More information can be found in a [paper](http://link.springer.com/article/10.1007/s10032-015-0249-8).
In summary:

Pros:
- Good detection level
- Works for a lot of different journals
- OpenSource and actively developed
- Provides facilities to extract bibliographic information from a citation string (i.e. plain text to BibEntry converter would be possible). Can be tested [online](http://adalab.icm.edu.pl/api/citation_parser.html?q=L.+Bolikowski%2C+N.+Houssos%2C+P.+Manghi%2C+J.+Schirrwagen%2C+%22Data+as+First-class+Citizens%2C%22+D-Lib+Magazine%2C+vol.+21%2C+no.+1%2F2%2C+2015.)
- Also extracts document structure (i.e. sections) and a list of referenced articles! See for [example](https://pub.phyks.me/paper_references_extractor/cermine_out/4945dbf177156d481f686f116b9fce5b.pdf.cermine.xml)

Cons:
- Has problems with books/thesis
- Increases the size of JabRef from around 40 mb to 70 mb.

A comparison of different metadata extract tools can be found in the [following blog post](https://known.phyks.me/2016/comparison-of-tools-to-fetch-references-for-scientific-papers), which ends with the following summary
> the machine learning based models (Cermine and Grobid) seems to give far better results than the simple approach taken by pdfextract, at the expense of being more difficult to build and run. Cermine gives a bunch of infos, too much in my opinion, and I think Grobid is given the most reusable and complete results.

I tried to include [Grobid](https://github.com/kermitt2/grobid) but it appears to be [hard to use](https://github.com/kermitt2/grobid/issues/59) and has some [problems on Windows (until recently?)](https://github.com/kermitt2/grobid/issues/92).

@koppor I added you as reviewer since you wrote the PdfImporter as far as I understood it.


src/main/java/net/sf/jabref/model/entry/Date.java
```diff
@@ -0,0 +1,62 @@
...
+import java.time.DateTimeException;
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+import java.time.temporal.TemporalAccessor;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Optional;
+
+public class Date {
```
Siedlerchr: Sure you want to call that Date? Will create a lot of confusion with the import hell..

Siedlerchr:
Really like the idea!

koppor:
Grobid is Apache-licensed: https://github.com/kermitt2/grobid/blob/master/LICENSE

koppor:
This is still `WIP`, because [CERMINE](https://github.com/CeON/CERMINE) is [A-GPL-licensed](https://github.com/CeON/CERMINE/blob/master/LICENSE). This is on of the highest risks in open-source-software.

Maybe, someone should re-try [grobid](https://github.com/kermitt2/grobid).

koppor:
We are currently trying to focus on other things. :fire:
We will come back later to this topic, therefore closing adding [on-hold label](https://github.com/JabRef/jabref/issues?utf8=%E2%9C%93&q=label%3Aon-hold+) and closing the issue for now.

Issue #2477, opened 2017-01-22
Error importing groupstree version 3 generated by Better BibTeX for Zotero


retorquere:
I'm the author of Zotero Better BibTeX; my extension generates BibTeX including a JabRef groupstree that is intended to import cleanly into JabRef. Recenty I've had reports that this import fails; this means my implementation of the groupstree is faulty, but as I cannot find any documentation on the format, I don't know in what way it is faulty.

I have a sample at https://drive.google.com/open?id=0BxFpK0V-elKWSVE1ejdLVkdiNXc, if someone could have a look at what I did wrong it'd be hugely appreciated. The BBT-generated groupstree is very simple and always only has ExplicitGroup entries.

lenhard:
Hi @retorquere 

Thanks for your report and thanks for supporting an integration with JabRef. Recently, there have been considerable changes in the way in which information about groups is serialized. This is probably the reason for why an import of your generated BibTeX no longer works. Unfortunately, the current serialization is not documented as well.

I am far from being an expert on groups and hopefully our group expert @tobiasdiez can find the time to look at this. The most notable change is that the group information is serialized together with an entry in a field and the meta data only contain the structure of the groups tree. Here is an example of one entry with a group:
```
@Article{somekey,
  author    = {Author Author},
  title     = {Title Title},
  groups    = {group},
}

@Comment{jabref-meta: groupstree:
0 AllEntriesGroup:;
1 ExplicitGroup:group\;0\;;
}
```

Maybe we should take this issue as an opportunity to document the group syntax of JabRef.

retorquere:
Oh wow that's a very significant change. This would mean group get some kind of unique ID then. It'd be good to know what kind of restrictions are in place on group IDs as I would have to generate them, and the Zotero collection names from which I'd have to do so will almost certainly contain illegal characters. And the names are not by necessity unique even on a single parent.

Is this format already out in the wild? Or is there still opportunity to weigh in on the format?

retorquere:
Unless https://github.com/JabRef/jabref/issues/1495 is still at play? Would that mean that an entry belongs to any group which happens to have that exact name?

lenhard:
As far as I am aware, special characters in groups should not be a problem. However, somebody else should confirm this. 

[...]

PR #2488, opened 2017-01-24
Catch NumberFormatException if context can't be parsed in groups

tobiasdiez:
Fixes https://github.com/JabRef/jabref/issues/2477 in the sense that a more detailed error message is displayed. 
The functional change is https://github.com/JabRef/jabref/pull/2488/files#diff-48e24aa13e9cd41a20bf4e79f4ad7d6bR129, while the rest is a bit of code cleanup (introduction of helper methods, removal of unused code, change public -> private if applicable).

### DIFF
--- a/src/main/java/net/sf/jabref/logic/importer/ParserResult.java
+++ b/src/main/java/net/sf/jabref/logic/importer/ParserResult.java
@@ -9,6 +9,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.stream.Collectors;
 
 import net.sf.jabref.model.database.BibDatabase;
 import net.sf.jabref.model.database.BibDatabaseContext;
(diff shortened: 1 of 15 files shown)


src/main/java/net/sf/jabref/logic/importer/ParserResult.java
```diff
@@ -46,36 +42,48 @@ public ParserResult(BibDatabase database) {
...
 
+    private static String getErrorMessage(Exception exception) {
+        String errorMessage = exception.getLocalizedMessage();
+        if (exception.getCause() != null) {
+            errorMessage += " Caused by: " + exception.getCause().getLocalizedMessage();
+        }
+        return errorMessage;
+    }
+
+    public static ParserResult getNullResult() {
```
lenhard: This smells a lot like singleton. There are potential problems in the current implementation, which do probably not manifest, but let us improve this anyway: If you always hand out the same `NULL_RESULT`, a caller of the method can modify this result (e.g. `NULL_RESULT.getDatabase().add(...)`), which could theoretically lead to interesting observations at other places.

Long story short: Please always create a new empty ParserResult here. Thus, you can also remove `NullResult`.

tobiasdiez: Good suggestion. I completely inlined the `getNullResult` method as I don't see any need to encapsulate a constructor call in a new method.

PR #2499, opened 2017-01-30
Display LaTeX-free version of field content in entry preview panel

### LINKED ISSUE #2498: "\'{}" Escape Sequence Does Not Display Correctly in Entry Preview
JabRef 3.8.2
windows 10 10.0 amd64 
Java 1.8.0_121

Steps to reproduce:

1. Create a new entry with "Pie{\\'{n}}kowski, G." as author.
2. The author name is correctly displayed as Pieńkowski in the author column of the entry table. This was a fix for https://github.com/JabRef/jabref/issues/2458 introduced in JabRef 3.8.2
3. The author name, however, is NOT correctly displayed in the entry preview of the article. Instead it is shown as Pieŉkowski, G.

Somewhat similar issues have been reported here:
https://github.com/JabRef/jabref/issues/2458
https://github.com/JabRef/jabref/issues/2460

lenhard:
Fixes #2498.

The PR implements a new method in `BibEntry`: `getResolvedFieldOrAliasLatexFree` that builds upon our existing `LatexFree` methods and extends this functionality to resolved fields. This method can be reused in the UI parts of JabRef that should rather display a Unicode version of a field than a LaTeX version, such as the entry preview.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
     - All file dialogs now use the native file selector of the OS. [#1711](https://github.com/JabRef/jabref/issues/1711)
 
 ### Fixed
+- LaTeX symbols are now escaped to Unicode in the entry preview panel. Fixes [#2498](https://github.com/JabRef/jabref/issues/2498).
 
 ### Removed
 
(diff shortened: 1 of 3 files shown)


Siedlerchr:
There seems to be some problems now with the layout files if you look at travis output

lenhard:
It seems that using a formatter for HTML conversion in the layout conflicts with the LatexToUnicode conversion. 

I'm not sure what our take on this is. Maybe this PR is not the solution at all? Maybe the correct solution would be to use LatexToUnicode in the entry preview configuation?

tobiasdiez:
> Maybe the correct solution would be to use LatexToUnicode in the entry preview configuration?

Sounds like the better solution in my opinion. I think the formatter can be specified in the layout file, something like `\format[LatexToUnicode]{....}`.

lenhard:
> \begin{author} \format[Authors(LastFirst,Initials,Semicolon,Amp),HTMLChars,LatexToUnicode]{\author}<BR>\end{author}

This is the way to go. Documented in #2498. This PR can be closed.

Issue #2467, opened 2017-01-14
Add option to create group in Add To Group Dialog


Siedlerchr:
One thing that recently annoyed me all the time.

1. Select an Entry
2. Add to Group
3. You can only choose existing groups here

=> Add possibility to create a new Group in that window, too.

Siedlerchr:
This feature was replaced by drag and drop

PR #2508, opened 2017-02-01
Groups button and context menu for adding groups

tobiasdiez:
Add a button (and context menu) for adding groups in the new interface. Some todo's are still open, because they are easier to implement as soon as the majority of the group stuff is converted to javafx.


lenhard:
I played around a little with the UI interface. The tree and the context menu are fine I think. However (not using groups that much), it took me quite some time to locate the button you talked about. 

The button was located at the bottom of the interface also in the old groups interface, right? I think the rewrite of the groups panel gives us an opportunity to try to improve a few things. Personally, I think it would be better if the button were moved to the top, right under the heading of the groups sidepane and above the "All Entries" group. What do you think?

Also, support for assigning entries to groups by drag and drop would be cool :-)

tobiasdiez:
You are of course right; we should use the rewrite to improve the interface. Thats why I moved the "new group" button from the top to the bottom 😄. My reasoning was that in the majority of the cases you are interested in selecting a group in order to find a assigned entry. Once you have an established group tree, the "new group" functionality is used very rarely. Thus the groups tree is way more important than the add button and should thus be put more prominently, i.e., at the very top. On the other hand, I can see that the add button at the bottom looks a bit disconnected from the groups tree (especially if you have only a few groups). Not sure how to improve this through.

But in the end, I have not a very strong opinion about the location of the button and it is very easy to change. My proposal would be to leave the button at the bottom for the moment and revisit this question as soon as more buttons/functionality is added to the groups pane.

lenhard:
I don't have a strong opinion either :-) @JabRef/developers What does the rest of you think?

Siedlerchr:
It would be nice if you could provide a screenshot. Otherwise I don't really care where it it as long as it is there. This is what I wanted in  #2467

Issue #2506, opened 2017-02-01
Import RIS from Elsevier does not get journal name


tagMacher:
JabRef version 3.8.2 on Ubuntu 16.04

This problem is also observed for versions 3.7, 3.8, and 3.8.1.

Steps to reproduce:

1. Export RIS for any journal article from ScienceDirect.com. Save file on local hard-disk.
2. Drag-drop RIS file onto JabRef window. This opens up the import dialog, Click ok after ensuring 'keep' checkbox for the article entry is selected.
3. Check the created entry for journal name - it is seen to be blank.

The RIS file created by ScienceDirect has a JO entry for journal name - this used to be correctly downloaded earlier. The last correctly working version is 3.6. Sample RIS file from ScienceDirect is enclosed (as ZIP archive) - RIS file can be seen in any text editor. Log pasted below.

[sciencea49f51c3.ris.zip](https://github.com/JabRef/jabref/files/743866/sciencea49f51c3.ris.zip)

```
10:18:35.724 [JabRef CachedThreadPool] ERROR net.sf.jabref.logic.xmp.XMPUtil - XMP reading failed
java.io.IOException: Error: End-of-File, expected line
	at org.apache.pdfbox.pdfparser.BaseParser.readLine(BaseParser.java:1519) ~[JabRef-3.8.2.jar:?]
	at org.apache.pdfbox.pdfparser.PDFParser.parseHeader(PDFParser.java:372) ~[JabRef-3.8.2.jar:?]
	at org.apache.pdfbox.pdfparser.PDFParser.parse(PDFParser.java:186) ~[JabRef-3.8.2.jar:?]
	at org.apache.pdfbox.pdmodel.PDDocument.load(PDDocument.java:1235) ~[JabRef-3.8.2.jar:?]
	at org.apache.pdfbox.pdmodel.PDDocument.load(PDDocument.java:1202) ~[JabRef-3.8.2.jar:?]
	at net.sf.jabref.logic.xmp.XMPUtil.loadWithAutomaticDecryption(XMPUtil.java:127) ~[JabRef-3.8.2.jar:?]
	at net.sf.jabref.logic.xmp.XMPUtil.readXMP(XMPUtil.java:160) ~[JabRef-3.8.2.jar:?]
	at net.sf.jabref.logic.xmp.XMPUtil.hasMetadata(XMPUtil.java:1110) ~[JabRef-3.8.2.jar:?]
[...]

PR #2510, opened 2017-02-02
Fix journal title in ris importer

tschechlovdev:
Fix for #2506.
The problem was that the T2 field was used for the journal title, if it comes first. Now T2 is only used for the journal title, if the field JO does not appear. I've also added the file from science direct as test case.

### DIFF
--- a/src/main/java/net/sf/jabref/logic/importer/fileformat/RisImporter.java
+++ b/src/main/java/net/sf/jabref/logic/importer/fileformat/RisImporter.java
@@ -129,7 +129,11 @@ public class RisImporter extends Importer {
                         fields.put(FieldName.TITLE, fields.get(FieldName.TITLE).replaceAll("\\s+", " ")); // Normalize whitespaces
                     } else if ("BT".equals(tag)) {
                         fields.put(FieldName.BOOKTITLE, value);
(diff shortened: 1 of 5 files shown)


src/main/java/net/sf/jabref/logic/importer/fileformat/RisImporter.java
```diff
@@ -129,7 +129,11 @@ public ParserResult importDatabase(BufferedReader reader) throws IOException {
                         fields.put(FieldName.TITLE, fields.get(FieldName.TITLE).replaceAll("\\s+", " ")); // Normalize whitespaces
                     } else if ("BT".equals(tag)) {
                         fields.put(FieldName.BOOKTITLE, value);
-                    } else if ("T2".equals(tag) || "JO".equals(tag)) {
+                    } else if ("T2".equals(tag)&&(fields.get(FieldName.JOURNAL) == null || "".equals(fields.get(FieldName.JOURNAL)))){
```
lenhard: Could you fix the formatting in the condition here? I am referring to whitespaces between operators `&&` and the operands.

tschechlovdev: Done. Also added a changelog entry.

tobiasdiez:
Thanks!

PR #2515, opened 2017-02-03
Highlight groups that match any/all of the entries selected in the main table.

tobiasdiez:
If a group contains all selected entries, then the "hit counter" turns green (and if only a few of the selected entries are contained in the group, then yellowish).

untitled

This replaces the "highlight groups" feature under the "Groups" menu.
(Best part: removes code from BasePanel and JabRefFrame)

I'm sorry that this PR contains a lot of modifications that are essentially only code rearrangements. 


src/main/java/net/sf/jabref/gui/BasePanel.java
```diff
@@ -2179,6 +1886,10 @@ public boolean isUpdatedExternally() {
         return updatedExternally;
     }
 
+    public void setUpdatedExternally(boolean b) {
+        updatedExternally = b;
```
Siedlerchr: Please use a more speaking variable name instead of b, maybe the same name as for the variable

src/main/java/net/sf/jabref/gui/BasePanel.java
```diff
@@ -2374,60 +2116,261 @@ public SearchAndOpenFile(final BibEntry entry, final BasePanel basePanel) {
...
+                frame.getGroupSelector().revalidateGroups();
+            }
+        };
+
+        /**
+         * Only access when you have the lock of the task instance
+         *
+         * Guarded by "task"
+         */
+        private TimerTask timerTask = new TimerTask() {
```
Siedlerchr: Better use a SheduledExecuterService: 
http://stackoverflow.com/questions/409932/java-timer-vs-executorservice

src/main/java/net/sf/jabref/gui/BasePanel.java
```diff
@@ -2374,60 +2116,261 @@ public SearchAndOpenFile(final BibEntry entry, final BasePanel basePanel) {
         }
     }
 
+    private class GroupTreeListener {
+
+        private final Runnable task = new Runnable() {
```
Siedlerchr: You could use a lamdba here, too

lenhard:
I know that it is hard to build a group-related PR with fewer LOC changes, but I still don't like reviewing a PR of this size ;-) On a very superficial scroll-over level, the code looks ok.

Instead, I played around a little with the UI. It took me a while to understand how the highlighting features was meant to work, but once I did, I could not find bugs. And I tried hard. Just one thought:
 - All entries seems to be always green, even if no entries are selected. Should it not be gray in this case? This is a sematics question, so just implement it in the same way as it has been before.

From my user's perspective, the PR is ready to be merged.

tobiasdiez:
I will merge this right now as it is. 
@Siedlerchr your comments concerned code that was just moved and not really modified (and I don't really want to touch the code in BasePanel and MainFrame since these components glue everything together but there are absolutely no tests in place...)

@lenhard The all entries hits number is no longer highlighted. Thanks for the suggestion. Do you have any ideas how to make the feature more self-explanatory?

lenhard:
@tobiasdiez Not really. Once you start thinking about it, you end up with a more or less similar solution. One minor thing maybe: When clicking a group I was expecting that all entries in the group become selected (that is: blue). It took me a while to understand that they were sorted to the top and the unselected entries were greyed out. But if it has been that way and nobody really complains about it, it is propably best to leave it that way (in the abscense of a better idea). Also, I did the testing with our gigantic testing database from Aegit. Propably the highlighting is more intuitive if you use it with a database where you built the grouping yourself.

tobiasdiez:
There are two modes for groups:
- Gray-out entries that are not matched 
- Filter, i.e. hide non-matched entries
I didn't changed anything at this behaviour.
One of the next PRs will move the option to switch between these modes to the preferences.

matthiasgeiger:
> One of the next PRs will move the option to switch between these modes to the preferences.

To the `Options` -> `Preferences` preferences? Or do you want to recreate the "old" groups settings menu?

tobiasdiez:
I prefer to collect all the options in the preferences dialog. I think it is also not a setting that a user changes very often. What do you think?

matthiasgeiger:
In general I don't agree: Aspects as "Union vs. Intersect" and "Inverted" are more things that should be directly changeable in the groups UI. 
Switching between "hide" and "float" is more an edge case - as I cannot come up with an use case which requires switching the style - this is just a personal preference.

tobiasdiez:
Of course you are right that the other settings (union, intersect, inverted, etc) would make sense as a popup-menu (as they are currently implemented) or as a toggle button. But these other options are not very high on my priority list as I don't really understand why one needs them (maybe they are better implemented as search commands like `not in mygroup` and `in groupa and groupb`)

matthiasgeiger:
Yeah you need them to accomplish "searches" like the ones you proposed (Find all papers, a) which are in both selected groups, b) which are in one or both of the selected groups, c) which are not in the selected group. 
Of course these are mere "shortcuts" for normal searches, but I'm not sure whether an ordinary JabRef user without some basic knowledge of formal logic can formulate this that easy. 

Maybe we could implement this a bit different as it is now using toggle buttons as for the normal search: Toggle button for "invert" + toggle button for "union", which can only be selected if multiple groups are selected (Side note: This is currently not possible in the FX version :wink:). And, as this is basically just a search that is performed, we could consider that the "search term" will be automatically appended at the search field of the search bar. This could either educate the user how to formulate such searches manually - or confuse him :wink: This also might have been some side effects we should think about (e.g., if a "global search" is already active.

Issue #2489, opened 2017-01-26
Hangs after editing abstract of existing entry


ziggystar:
## Versions

- JabRef version 3.8.1 
- on Ubuntu 16.04
 - java version "1.8.0_121" (oracle)

## Issue
The program hangs with no CPU utilization after editing an abstract (removing a HTML code and replacing it with the word 'and'). I don't know whether this is reproducible, simply editing an abstract didn't do it. 

But I have a thread dump attached. I think it's some kind of deadlock/race condition, and I hope you can diagnose it from the stack traces.
[jabref-thread-dump.zip](https://github.com/JabRef/jabref/files/732157/jabref-thread-dump.zip)

koppor:
Could you a) try to reproduce? b) provide us the abstract?

Here, I can edit the abstract without issues.

Am 26.01.2017 12:31 schrieb "ziggystar" <notifications@github.com>:

ziggystar:
I tried for a bit and I can't reproduce it. I don't know exactly what I clicked after editing the abstract. 

I looked a bit into the sources and couldn't find an obvious issue with the locking.

- The deadlock is not visible inside the stack trace, for me. Not within the Swing thread itself, and also no other thread appears to be holding a lock. This hinges on me recognizing the places where a lock is held. See next.
- I roughly checked the places where `writeLock().lock()` (by find usages on `writeLock()`) is called and they all have an `unlock()` within a `finally`, except one which could be an issue if unchecked exceptions are thrown/**caught** within the JabRef source code:
    ````
        sortedEntries.getReadWriteLock().writeLock().lock();
        comparatorChooser.appendComparator(PAD, 0, false);
        sortedEntries.getReadWriteLock().writeLock().unlock();
    ````

Here is the relevant part of the stack; apparently I hit a key that triggered the save database action. That's quite possible.

```
"AWT-EventQueue-0" #14 prio=6 os_prio=0 tid=0x00007f365c782000 nid=0x4919 waiting on condition [0x00007f36121c0000]
<snip>
	at ca.odell.glazedlists.impl.java15.LockAdapter.lock(J2SE50LockFactory.java:85)
	at net.sf.jabref.gui.maintable.MainTableDataModel$StartStopListFilterAction.update(MainTableDataModel.java:237)
	at net.sf.jabref.gui.maintable.MainTableDataModel$StartStopListFilterAction.start(MainTableDataModel.java:229)
[...]

Issue #2482, opened 2017-01-23
Seemingly Random freeze on PDF import


paulsputer:
v3.8.1, windows 10

Great work on this version the _get BibTex from doi_ functionality is working better than ever. Unfortunately this version often seems to suffer a freeze when importing PDFs.  Initially I thought it was caused by some interaction with the _quality assured_ field as it seemed to freeze after setting a few _quality assured_ fields and then importing a PDF.  However, I've also had it freeze without setting those.  On a restart it imports the file no problem so seems to be some state issue.

Would love to give more details but the freeze prevents me from accessing the in-app log.  Is there a quick way to get it to print the log to a file without running in dev mode?

Perhaps related to #410 #2309

paulsputer:
Can confirm this also occurs on v3.8.2

It appears to be a state issue as importing [this PDF file](http://dx.doi.org/10.1007/s11263-013-0644-x) caused a crash.  However, when importing into a fresh database it did not cause a crash.

Any hints on how to turn on file based logging on the release version would be much appreciated.

matthiasgeiger:
Hi! 
Can you please describe the concrete workflow of your "importing the pdf"? There are different ways to do this - and there might be even different understandings of what "import" actually means ;-)

Regarding the logging: We have not implemented a file based logging mechanism - but you can run the `jar`-Version of JabRef from the windows console which will show each log statement directly without the need to open the JabRef error console dialog.

paulsputer:
Hi @matthiasgeiger sure.

1. File explorer window open of directory containing PDF files
2. Select a single PDF file and drag and drop into JabRef list panel
3. Confirm dialog to relocate/rename PDF
4. File relocated and metadata displayed in JabRef

In the case of the error occurring the JabRef UI is unresponsive however there are no drawing errors.  The window can be relocated and it redraws correctly.  CPU usage for JabRef indicated at 0%, only way to close JabRef is via task manager.

Thanks for the hint on running the jar version I'll try that and see if it provides some further clues

Siedlerchr:
As I cant reproduce that here, do you make use of the XMP feature, e.g. creating metatadat from the pdf or do you solely attach the pdf?

paulsputer:
[...]

PR #2517, opened 2017-02-03
fix for #2482 deadlock on PDF import

paulsputer:
Prevented deadlock for #2482 when importing PDF by removing UI update via multiple threads.  Likely to be similar resolution for #2489 but insufficient information on non AWT thread to identify where fix should be made.


Siedlerchr:
Thank you very much for your contribution and your debugging this issue! I will merge this in now.

PR #2527, opened 2017-02-07
Fix error when path is no valid directory

### LINKED ISSUE #2522: Not possible to save a database
JabRef_windows-x64_4_0_0-dev--snapshot--2017-02-05--multicleanup--ddecdfd3f
Windows 7
java 1.8.0_121

Steps to reproduce:

1. open a new database 
2. input from medline PMID 27151464
3. run cleanup
4. try **save** or **save as**

this does not give any reaction; the program when run exit notices there is something to save but the file dialog does not open.

```
Put the excerpt of the log file here
```
btw: where to find the log file?

Siedlerchr:
Fixes the exception Folder parameter must be a valid folder
Internal fix, 4.0.0 only
When no working directory is set and last used path was an actual file, File Dialog could not be opened.
Fixes #2522

### DIFF
--- a/src/main/java/net/sf/jabref/gui/FileDialog.java
+++ b/src/main/java/net/sf/jabref/gui/FileDialog.java
@@ -2,6 +2,7 @@ package net.sf.jabref.gui;
 
 import java.awt.Component;
 import java.io.File;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Collection;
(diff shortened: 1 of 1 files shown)


stefan-kolb:
LGTM, but what is the normal behavior for system dialogs? If I select a file ijnstead of a directory, is it possible and selects the parent folder or is the select button inactive?

Siedlerchr:
@stefan-kolb  My fix has only impact when **opening** the dialog, that it opens in a directory without an exception . It has no impact on the selection of files etc.

Parent does not mean parent directory, but the directory itself (the parent of the file):
Example: Your last selected file was X:\abc\defg.pdf  
When you try to set the initalDirectory to this complete path (including filename), it gives an exception, because it is obviously no a folder (directory). Therefore, the fix now sets the dir  to X:\abc (which is the parent of X:\abc\defg.pdf

stefan-kolb:
Ah **** sorry for that! Yeah I know, i was aware of that.

PR #2529, opened 2017-02-08
Rework File links logic

stefan-kolb:
Major changes:
- Download from URL will now insert at __first__ position (formerly last)
- Get Fulltext from URL will now insert at __first__ position(formerly last)
- Add Button will now insert at __last__ position (formerly at selected row)

image

### DIFF
--- a/src/main/java/net/sf/jabref/gui/externalfiles/FindFullTextAction.java
+++ b/src/main/java/net/sf/jabref/gui/externalfiles/FindFullTextAction.java
@@ -91,10 +91,11 @@ public class FindFullTextAction extends AbstractWorker {
                         basePanel.getBibDatabaseContext(), entry);
                 try {
                     def.download(result.get(), file -> {
-                        FileListTableModel tm = new FileListTableModel();
-                        entry.getField(FieldName.FILE).ifPresent(tm::setContent);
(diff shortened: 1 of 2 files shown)


src/main/java/net/sf/jabref/gui/externalfiles/FindFullTextAction.java
```diff
@@ -91,10 +91,11 @@ public void update() {
...
                 try {
                     def.download(result.get(), file -> {
-                        FileListTableModel tm = new FileListTableModel();
-                        entry.getField(FieldName.FILE).ifPresent(tm::setContent);
-                        tm.addEntry(tm.getRowCount(), file);
-                        String newValue = tm.getStringRepresentation();
+                        FileListTableModel fileLinkModel = new FileListTableModel();
+                        entry.getField(FieldName.FILE).ifPresent(fileLinkModel::setContent);
+                        // add full text file link at first position
+                        fileLinkModel.addEntry(0, file);
```
Siedlerchr: What happens if there is already a file linked?
Will the linked file then be moved one down?

stefan-kolb: yeah, the file link model is just a linked list

Issue #2455, opened 2017-01-12
Add Crossref.org as source for web search


matthiasgeiger:
As [complained here at the forum](http://discourse.jabref.org/t/the-google-scholar-fetcher-should-also-fetch-the-doi/360/6?u=matthiasgeiger) the Google Scholar BibTeX quality is not the best. As there are no "good" databases for all research disciplines present in JabRef we should add the crossref.org service to the web search fetchers (we are already using it for DOItoBibTeX).

Implementation should be straightforward but I've currently no time to do it...

REST API Documentation: https://github.com/CrossRef/rest-api-doc/blob/master/rest_api.md
Web search: http://search.crossref.org/

stefan-kolb:
Wishes for fulltext fetcher will be collected at #2581.

wujastyk:
Tried out the new crossref fetcher in 4.0.0 beta today, but it gave an error message:

![screenshot from 2017-05-02 11-04-31](https://cloud.githubusercontent.com/assets/762246/25629568/3264e97c-2f27-11e7-94cc-66f8cc58ddd3.png)

Other fetchers, GVK etc., are working, and the network is fine.

matthiasgeiger:
Unfortunately the crossref service is quite unstable 😥 there are quite often some problems at their Server side which prevent correct fetching of entries...

wujastyk:
That's annoying, of course.  But I don't think the instability of Crossref accounts for the problem we're seeing, though, because some searches (e.g., "zysk") _always_ give a result, and others ("wujastyk") _always_ give an error.  Plus, searching for a problem-term at crossref.org itself works fine.

tobiasdiez:
@wujastyk thanks for the report! There was a problem in our parsing logic which should be resolved now. 
Could you please check the build from http://builds.jabref.org/master/. (be advised that some parts are still a bit unstable, especially the groups panel and the entry editor)

wujastyk:
Yes, that fixed the problem!  Many thanks!

When I ran JabRef-4.0.0-dev--snapshot--2017-05-03--master--a16e6123d.jar, I
had to install the Oracle 8 version of java.  It wouldn't run with
java-1.8.0-openjdk-amd64.  Error message:

$ java -jar JabRef-4.0.0-dev--snapshot--2017-05-03--master--a16e6123d.jar
Error: Could not find or load main class org.jabref.JabRefMain

With Oracle 8, the program ran, but I got these messages, which didn't stop
the program working just fine:

[...]

PR #2531, opened 2017-02-08
Check more results returned by CrossRef API for matching

### LINKED ISSUE #2431: Can download article by DOI but not title
If I click the New Bibtex Entry button, paste this DOI:

    http://dx.doi.org/10.1016/j.engappai.2016.01.001

into the ID field, and then click the Generate button, then Jabref successfully downloads the article's BibTeX entry.

But if I instead paste the article's title into the ID field, change the ID Type to Title, and then click Generate, I get the error shown in this screenshot:

image

I'm using the title string:

    Traffic flow optimization: A reinforcement learning approach

This is with: 

    JabRef 3.8.1 windows 7 6.1 amd64  Java 1.8.0_111

stefan-kolb:
Resolves #2431

Changes:
- We will now check 5 results instead of just the first one for a matching title.

### DIFF
--- a/src/main/java/net/sf/jabref/logic/importer/fetcher/CrossRef.java
+++ b/src/main/java/net/sf/jabref/logic/importer/fetcher/CrossRef.java
@@ -27,11 +27,16 @@ import org.json.JSONObject;
  */
 public class CrossRef {
     private static final Log LOGGER = LogFactory.getLog(CrossRef.class);
-    private static final RemoveBracesFormatter REMOVE_BRACES_FORMATTER = new RemoveBracesFormatter();
 
     private static final String API_URL = "http://api.crossref.org";
+    // number of results to lookup from crossref API
+    private static final int API_RESULTS = 5;
+
     private static final Levenshtein METRIC_DISTANCE = new Levenshtein();
+    // edit distance threshold for entry title comnparison
     private static final int METRIC_THRESHOLD = 4;
+
+    private static final RemoveBracesFormatter REMOVE_BRACES_FORMATTER = new RemoveBracesFormatter();
 
     public static Optional<DOI> findDOI(BibEntry entry) {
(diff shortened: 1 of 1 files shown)


src/main/java/net/sf/jabref/logic/importer/fetcher/CrossRef.java
```diff
@@ -84,33 +90,37 @@ private static String enhanceQuery(String query, BibEntry entry) {
...
+                // additional check, as sometimes subtitle is needed but sometimes only duplicates the title
+                if (data.getJSONArray("subtitle").length() > 0) {
+                    String dataWithSubTitle = dataTitle + " " + data.getJSONArray("subtitle").getString(0);
+
+                    if (editDistanceIgnoreCase(entryTitle, dataWithSubTitle) <= METRIC_THRESHOLD) {
+                        return Optional.of(data.getString("DOI"));
+                    }
+                }
+            } catch(JSONException ex) {
+                return Optional.empty();
```
Siedlerchr: Mabye useful to add a logging here in case the json format changes?

matthiasgeiger: This should be detected by our Tests. Shouldn't it?

However, logging here does no harm.

stefan-kolb: The test will fail yes.

matthiasgeiger:
LGTM

Do you want to implement https://github.com/JabRef/jabref/issues/2455 as well?  :heart_eyes_cat:

matthiasgeiger:
Ok I interpret this as a "no" :wink:

stefan-kolb:
nah will do that later, maybe at jabcon. is a separate task.

Issue #2465, opened 2017-01-13
Switch to latex2unicode lib instead of own handling


Siedlerchr:
This would solve some serious issues:

Idea: Use https://github.com/tomtung/latex2unicode
Provides a simple extendable grammar.

We then would only need a conversion to html

Refs #2063 #207 #1252

tobiasdiez:
Or: https://github.com/JabRef/jabref/issues/1252

lenhard:
I set up a working implementation using latex2unicode in the latex2unicode branch. Here is the diff: https://github.com/JabRef/jabref/compare/latex2unicode

I had some intial setup problems, but the author helped me to resolve that in  tomtung/latex2unicode#1 

The library is really easy to use and would be great in terms of code complexity. It also fixes a few conversion problems we are having. However, here are the results from the benchmark:

On current master:
```
Benchmark                             Mode  Cnt      Score       Error  Units
Benchmarks.latexToUnicodeConversion  thrpt   20  100036.407 ▒ 1406.243  ops/s
Benchmarks.parallelSearch            thrpt   20     733.635 ▒   74.242  ops/s
Benchmarks.parse                     thrpt   20      26.130 ▒    0.653  ops/s
Benchmarks.search                    thrpt   20     374.335 ▒   17.097  ops/s
```

Using latex2unicode
```
Benchmark                             Mode  Cnt      Score       Error  Units
Benchmarks.latexToUnicodeConversion  thrpt   20    1936.683 ▒   75.321  ops/s
Benchmarks.parallelSearch            thrpt   20     744.081 ▒   78.945  ops/s
Benchmarks.parse                     thrpt   20      23.939 ▒    1.574  ops/s
Benchmarks.search                    thrpt   20     402.994 ▒   16.057  ops/s
```

This is a drop in op/s by two orders of magnitude. It is also noticeable when starting up JabRef. I hate to say this since I really like the library, but it looks as if we cannot use it due to performance.

tobiasdiez:
Did you tried the benchmark without the normalizer and the replacement of tildes? Maybe these post-and preconversations are a problem.
But probably the preformance problems comes from the fact that latex2unicode really uses a grammar to parse the latex code. In this case, I would propose to merge both projects:
- reuse our parser
- but convert similar to latex2unicode, in particular use something similar to the helper methods https://github.com/tomtung/latex2unicode/tree/master/src/main/scala/com/github/tomtung/latex2unicode/helper to make the code more readable.

[...]

Issue #2063, opened 2016-09-25
Latex to Unicode in author field fails


grimes2:
JabRef 3.7dev--snapshot--2016-09-25--master--40df60b
windows 10 10.0 amd64 
Java 1.8.0_101
1. author field in entry editor: Sto{\"{\i}}anova, Ivanka
2. author field in main table:  Stoanova, Ivanka
3. should be: Stoïanova, Ivanka

Siedlerchr:
Confirmed. Will take a look at it.

Siedlerchr:
Ah, I found the problem:

For some reason I don't know there exist 2 Variants of these character, `latin small letter i with diaeresis (U+00EF)`
ftp://ftp.dante.de/tex-archive/macros/xetex/latex/xecjk/xunicode-symbols.pdf

However, JabRef only supports the first one: `\^{i}`

stefan-kolb:
@oscargus is the expert here 😄

Siedlerchr:
The problem is theoretically easy to fix, but practically we have the problem that the mappings are stored in a HashMap and we have no chance to add a double entry.  Fuu.. I hate HashMaps sometimes.

koppor:
Refs https://github.com/koppor/jabref/issues/145

Siedlerchr:
@koppor  Sure. I found that piece of code. However, 

`00EF  ï  ` has two possible mappings: `\"{i}` or `\"{\i}` (notice the slash before the i) 
The latter one is used here, latest version (inlcuding unicode8)
https://www.w3.org/2003/entities/2007xml/unicode.xml (warning, file is around 5MB)

Both versions are valid. The question is: Do we find a solution to support multiple variants or do we stick with just one variant and then do nothing?

The current behaviour is more or less removing the character if no conversion possible

tobiasdiez:
I think you can just add a new entry to the `CONVERSION_LIST`. In the Unicode -> Latex direction only one of these entries is kept, but both should be added to the Latex -> Unicode map.

Siedlerchr:
That was may inital idea, too: However the point is, we have a  **HashMap**

https://github.com/JabRef/jabref/blob/ee688c44765f11344226da531b63f9df345a6585/src/main/java/net/sf/jabref/model/strings/HTMLUnicodeConversionMaps.java#L852
` public static final Map<String, String> LATEX_UNICODE_CONVERSION_MAP = new HashMap<>();`

[...]

Issue #207, opened 2015-10-04
Combining accents not working in Unicode converter (and more)


oscargus:
Just to remember for myself (or anyone else):

The Unicode to LaTeX converter does not handle combining accents, i.e., `"a\u0301"` = á does not convert into `{\'{a}}` as it should. Interestingly enough it seems like the Java compiler would also convert `"\u0301a"` to á, but `"a\u0301e"` = áe, so clearly the combining accent should be applied to the preceding character.

Double combining accents are not supported in either HTML nor Unicode to LaTeX conversion. While some of the codes are there, the behavior is not correct. For example, `b&#x35E;c` or `"b\u035Cc"`, meaning b͞c, should be converted to something clever, probably not `{\textdoublemacron{b}}c` which is the current result from HTML.

oscargus:
How complete should this functionality really be? When I look around, I notice that there are huge conversion lists which could be implemented. On the other hand, how likely is it that e.g. ⬲ (HTML: `&#x2B32;`, Unicode: `"\u2B32"`) shows up in an imported field?

Should we keep it relevant or complete?

oscargus:
The old list was pretty much constructed by actively searching in IEEE Xplore for entries that were likely to end up with math symbols etc. That covered most of the HTML entries at least. I guess one can do something similar with Medline to cover even more symbols that might occur.

koppor:
This seems somehow related to #161, which can now be closed, can't it? If I get you right, `latex2utf8` and `LaTeX::Encode` now support less characters.

I would use that tools as benchmark and not do more encoding.

Since [biber](http://biblatex-biber.sourceforge.net/) is a perfect replacement for bibtex, I'd even suggest to do as least effort as possible here and encourage the users to switch to biblatex. The idea of #161 was only to gain a lot with little effort. Now, it sounds like much effort for little gain :see_no_evil:

ThomasA:
Careful with Biber... Biblatex *is* great, but some journals do not 
accept LaTeX submissions using Biblatex - only BibTeX. As far as I know, 
using `\usepackage[utf8]{inputenc}` and `bibtex8` instead of `bibtex` 
can get you a fair bit of the way there.

koppor:
[...]

PR #2532, opened 2017-02-09
Switch to Latex2unicode

lenhard:
Motivation is discussed in #2465. Fixes many conversion problems, such as those discussed in #2063 and #207.


stefan-kolb:
Please remove the old latex2unicode formatter code, otherwise good to go.

src/main/java/net/sf/jabref/model/strings/LatexToUnicodeAdapter.java
```diff
@@ -0,0 +1,18 @@
...
+import java.util.Objects;
+
+import com.github.tomtung.latex2unicode.LaTeX2Unicode;
+
+/**
+ * Adapter class for the latex2unicode lib. This is an alternative to our LatexToUnicode class
+ */
+public class LatexToUnicodeAdapter {
+
+    public static String format(String inField) {
```
tobiasdiez: I would merge this class into the `LatexToUnicodeFormatter`. I don't see any advantage of having two classes that do essentially the same.

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
     - Redesigned journal abbreviations dialog.
     - Redesigned error console.
     - All file dialogs now use the native file selector of the OS. [#1711](https://github.com/JabRef/jabref/issues/1711)
+- Switch to the latex2unicode library https://github.com/tomtung/latex2unicode for converting LaTeX to unicode
```
tobiasdiez: use markdown links

src/test/java/net/sf/jabref/logic/layout/format/LatexToUnicodeFormatterTest.java
```diff
@@ -28,9 +29,10 @@ public void testFormatStripLatexCommands() {
     @Test
     public void testFormatTextit() {
         // See #1464
-        assertEquals("text", formatter.format("\\textit{text}"));
+        assertEquals("\uD835\uDC61\uD835\uDC52\uD835\uDC65\uD835\uDC61", formatter.format("\\textit{text}"));
```
tobiasdiez: [d835](http://www.charbase.com/d835-unicode-invalid-character) is the symbol for an invalid character. Is this conversation a bug or indeed the expected outcome?

lenhard: It is not only the symbol for an invalid character, it is also the symbol for a Java escape (shown on the page you linked). I arrived at this sequence by copying the italicized text in Intellij into the code which transformed it into these hexes.

If you remove the first hex, then the first letter of the expected output turns into a `?` for Intellij, so this really seems to be the escape sequence.

src/main/java/net/sf/jabref/model/strings/LatexToUnicodeAdapter.java
lenhard: This is not possible, since we need the unicode conversion in the model to internally convert fields of `BibEntries`. If we just had the formatter, we would have a dependency from model to logic, which is forbidden in the architecture.

CHANGELOG.md
lenhard: done

src/test/java/net/sf/jabref/logic/layout/format/LatexToUnicodeFormatterTest.java
matthiasgeiger: italics

Okay... I really kinda works - Although is looks strange ;-)

lenhard: I don't expect anyone to ever use italicized text in their entries - after all, why would you? But isn't it cool to know that with the converter we could theoretically support it? :-)

tomtung: Some more details here https://en.wikipedia.org/wiki/Universal_Character_Set_characters#Surrogates :)

lenhard:
Since the build is passing and there are two approved reviews, I will merge this into master now and close the related issues.

PR #2538, opened 2017-02-12
Fix import of MS Office xml in case of invalid month

Siedlerchr:
Fix for:  http://discourse.jabref.org/t/msbib-import-not-working/463
Fix month field.shortname returning null in case of invalid Month

~~- [ ] Check documentation status (Issue created for outdated help page at [help.jabref.org]~~(https://github.com/JabRef/help.jabref.org/issues)?)


src/main/java/net/sf/jabref/logic/msbib/BibTeXConverter.java
```diff
@@ -84,7 +83,12 @@ public static BibEntry convert(MSBibEntry entry) {
         }
         if (entry.month != null) {
             Month month = MonthUtil.getMonth(entry.month);
-            fieldValues.put(FieldName.MONTH, month.shortName);
+            //if we encouter an UnknownMonth/NULL_OBJECT month, the shortname returns null
+            if (month.shortName == null) {
```
tobiasdiez: I think it is cleaner if you modify the `getMonth` method to return `Optional<Month>̀  instead of `NULL` objects.

Siedlerchr: Yeah. Thought about remodeling the stuff. The NULL_OBJECT stuff and UnknownMonth is really shitty,

stefan-kolb:
Can you please add a test for this if possible?

src/main/java/net/sf/jabref/logic/msbib/BibTeXConverter.java
simonharrer: You can use the `isValid` method of a `Month` to determine whether it is a valid month.

Siedlerchr: Done and added also test

Siedlerchr:
Check for valid month and added test

PR #2541, opened 2017-02-14
Enhance integrity check tests

stefan-kolb:
See https://github.com/koppor/jabref/issues/162

### DIFF
--- a/src/test/java/net/sf/jabref/logic/integrity/IntegrityCheckTest.java
+++ b/src/test/java/net/sf/jabref/logic/integrity/IntegrityCheckTest.java
@@ -6,6 +6,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
+import java.util.UUID;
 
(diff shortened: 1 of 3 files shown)


src/test/java/net/sf/jabref/logic/integrity/IntegrityCheckTest.java
```diff
@@ -308,6 +309,32 @@ public void testDOIChecks() {
...
+        BibDatabase bibDatabase = new BibDatabase();
+        bibDatabase.insertEntry(entry);
+        BibDatabaseContext context = new BibDatabaseContext(bibDatabase, new Defaults());
+
+        new IntegrityCheck(context,
+                JabRefPreferences.getInstance().getFileDirectoryPreferences(),
+                createBibtexKeyPatternPreferences())
+                .checkBibtexDatabase();
+        // equals
+        clonedEntry.equals(entry);
```
tobiasdiez: Use assertEquals otherwise the test never fails.

stefan-kolb: god, good catch :D

PR #2545, opened 2017-02-15
Show pdf comments

lenhard:
This is a follow-up to #1883 which I cannot modify because I have no commit access to that branch.

I tested the features under Windows and it works fine overall. Opening the PDF at the position of the annotation does not work for me, hence I removed the button that provided this functionality. At this time (four month of inactivity in the other PR), the only point is to get the functionality in that actually works.


src/main/java/net/sf/jabref/logic/pdf/FileAnnotationCache.java
```diff
@@ -0,0 +1,43 @@
...
+
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+import net.sf.jabref.model.database.BibDatabaseContext;
+import net.sf.jabref.model.entry.BibEntry;
+import net.sf.jabref.model.pdf.FileAnnotation;
+
+import org.apache.commons.collections4.map.LRUMap;
```
simonharrer: Could we use a class from google  guava instead? Because of this LRUMap class, the whole commons collections library has to be included.

lenhard: To be done in the future:  #2546

kafran:
@lenhard is this really hard to implement? [#662](https://github.com/JabRef/jabref/issues/662#issuecomment-280012908). It would be really helpful to open the annotation on the PDF location.

JabRef could store the page number/label of the annotation and the user could configure its preferred PDF reader as: `exec evince -i $PAGE $FILE_PATH` or `exec firefox $FILE_PATH#page=$PAGE`

lenhard:
@kafran It is all relative of course ;-)

Storing the page number/ label of the annotation is already done right now, I guess. What makes this a lot of effort is if we want offer built-in configurations for a number of pdf readers and operating system (since we would have to test that). I guess if we just offer the user to execute a command he has to configure himself, it would get easier. We would still get support requests from people asking how things should work for their particular configuration, though.

Anyway, I guess this could be done. I don't see anyone from the current developers volunteering right now (@jabref/developers feel free to disagree), but you are very welcome to go ahead with an implementation and to provide a pull request!

PR #2551, opened 2017-02-15
Initial idea to show "Parsing..." during run

koppor:
When using the "New Entry from Plain Text dialog", there seems to be no reaction when pushing "Parse with FreeCite" (@lenhard).

grafik

This PR tries to solve this by changing the button label. However, it does not work. I have no clue. Does someone take to take over?


src/main/java/net/sf/jabref/gui/plaintextimport/TextInputDialog.java
```diff
@@ -456,6 +457,11 @@ public boolean okPressed() {
      * @return true if successful, false otherwise
      */
     private boolean parseWithFreeCiteAndAddEntries() {
+        SwingUtilities.invokeLater(() -> {
+            parseWithFreeCiteButton.setText(Localization.lang("Parsing..."));
+            parseWithFreeCiteButton.setEnabled(false);
```
Siedlerchr: You disable it here. Is this what you want?

koppor: Yes. Same thing as with the entry from id thing at New BibEntry...

Siedlerchr:
I'll take a look at this

Siedlerchr:
I tried it and both accept and Parse With FreeCite (whatever that means) works:
On acecpt a new empty entry is created, on Parse with FreeCite the bibtex code I pasted in the field is entered in author title...

koppor:
I just want to have a progress indicator: Disabled button with other text...

The functionality itseld works.

Am 07.03.2017 18:13 schrieb "Christoph" <notifications@github.com>:

koppor:
Expected - only with "Parsing..."

fuuu gif

Siedlerchr:
Okay, I looked into it. It does work (uncomment the dispose of the dialog for a moment and you see that the title changes). However this is done very fast, The freecite parsing does not take long enough that it is visible

tobiasdiez:
What if we just write `Creating new entry...` instead of `Searching...`? This applies to fetcher as well as to the "text parser". Anyway, I wouldn't invest to much energy in such a small thing (my opinion).

koppor:
OK, too much time spent here. I'll just close it.

PR #2557, opened 2017-02-16
[WIP] Refactored around the FileAnnotationCache.

lynyus:
The Cache now works as intended, i.e., fully automatically instead of manual maintenance.


src/main/java/net/sf/jabref/gui/entryeditor/EntryEditor.java
```diff
@@ -1102,7 +1107,7 @@ public void stateChanged(ChangeEvent event) {
 
 
     class DeleteAction extends AbstractAction {
-        public DeleteAction() {
+        DeleteAction() {
```
koppor: I think, checkstyle or codacy or some other toal moans because of package private. Nevertheless, I think, it is OK to reduce visibility.

src/main/java/net/sf/jabref/gui/entryeditor/FileAnnotationTab.java
```diff
@@ -66,78 +59,73 @@
...
+    FileAnnotationTab initializeTab(FileAnnotationTab tab, Map<String, List<FileAnnotation>> cachedFileAnnotations) {
+        this.annotationsOfFiles = cachedFileAnnotations;
+
+        if (tab.isInitialized) {
+            return tab;
+        }
 
-        if(!tab.isInitialized) {
-            
             try {
```
koppor: indent? - Think, indent is wrong here.

src/main/java/net/sf/jabref/gui/entryeditor/FileAnnotationTab.java
```diff
@@ -66,78 +59,73 @@
...
-        return tab;
+
     }
 
     /**
-     * Adds pdf comments from all attached pdf files belonging to the entry selected in the main table and
-     * shows those from the first file in the comments tab
+     * Adds pdf annotations from all attached pdf files belonging to the entry selected in the main table and
+     * shows those from the first file in the file annotations tab
      * @throws IOException
```
koppor: I would remove that, `@throws` javadoc, because it does not add any more information.

src/main/java/net/sf/jabref/logic/pdf/EntryAnnotationImporter.java
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * Filter files with a web address containing "www."
+     *
+     * @return a list of file parsed files
+     */
+    public List<ParsedFileField> getFilteredFileList() {
+        return FileField.parse(this.entry.getField(FieldName.FILE).get()).stream()
+                .filter(parsedFileField -> parsedFileField.getLink().toLowerCase().endsWith(".pdf"))
+                .filter(parsedFileField -> !parsedFileField.getLink().contains("www.")).collect(Collectors.toList());
```
koppor: I'm not sure whether this line is really valid. Did you find a test case for that? I assume, a user put an URL in the file field. This should not be checked here, but at other places (integrity check, ...)

src/main/java/net/sf/jabref/gui/entryeditor/EntryEditor.java
lynyus: It seems like codacy does not like default visibility at all

src/main/java/net/sf/jabref/logic/pdf/EntryAnnotationImporter.java
lynyus: actually this is a one-to-one copy of what has been there before. I don't have the domain-specific insight if it makes sense.

lynyus:
Hey, I have tried to fix the issues from your review. Check my comments on the particular parts above.

PR #2563, opened 2017-02-17
Make automatic groups a first-class citizien

tobiasdiez:
So after you guys had your fun in Lugano (and harvested a big number of changed lines), I want to have my share too. Thus here I'm with a new groups-related PR.

Previously there was the option to automatically generate a bunch of groups. For example based on the keywords contained in a certain field. However, this process was a one-time thing, which means that once the groups were created they were not keep in sync with changes to the entries.
In this PR, I introduce the automatic groups as a new type of groups that based on a certain criteria automatically create subgroups and keep them synchronized with the changes.

The current implementation has a few caveats which I would like to address in a new PR:
- Automatic groups are not yet written to the bib file (and parsed).
- The update process only happens if you switch between databases and the groups tree is reinitialized. 
- The UI in the groups edit dialog is probably not the greatest...

Remark: Our architecture tests threw a NPE instead of a meaningful message in case of an issue. This is also fixed.

----


src/main/java/org/jabref/gui/groups/GroupDialog.java
```diff
@@ -304,39 +358,55 @@ public void actionPerformed(ActionEvent e) {
...
-            setContext(editedGroup.getHierarchicalContext());
-        } else { // creating new group -> defaults!
+        if (editedGroup == null) {
+            // creating new group -> defaults!
             explicitRadioButton.setSelected(true);
             setContext(GroupHierarchyType.INDEPENDENT);
+        } else {
+            if (editedGroup.getClass() == WordKeywordGroup.class) {
+                WordKeywordGroup group = (WordKeywordGroup) editedGroup;
+                nameField.setText(group.getName());
```
Siedlerchr: I would prefer a switch case here, but that is a minor thing

src/main/java/org/jabref/gui/groups/GroupDialog.java
```diff
@@ -304,39 +358,55 @@ public void actionPerformed(ActionEvent e) {
...
+                keywordGroupSearchField.setText(group.getSearchField());
+                keywordGroupSearchTerm.setText(group.getSearchExpression());
+                keywordGroupCaseSensitive.setSelected(group.isCaseSensitive());
+                keywordGroupRegExp.setSelected(false);
+                keywordsRadioButton.setSelected(true);
+                setContext(editedGroup.getHierarchicalContext());
```
Siedlerchr: Move set context out of the ifs as it is called always with the same parameter 
Same for set Name

lenhard:
You call a mere thousand changes "line harvesting"? Pathetic! ;-) Fun aside, please do not take this as a justification to blow up the number of LOC in your PR.

I have a more general comment: The handling of special fields is pretty much the same as you do here. Have a look at the (deprecated) class `org.jabref.logic.specialfields.SpecialFieldUtils`. This class synchronizes the keyword field with different other fields. You want the same here: To synchronize one field (groups) with some others (author, keywords). It would be really cool if we could extract this synchronization logic in a more abstract fashion, so that it is reused for the special fields and the automatic groups. Otherwise, we would have several competing implementations for the same thing in JabRef.

I haven't given much thought at how to unify these two features with the common logic, but maybe you could give it a try here? The special fields are much less ugly now than they used to be, promise!

tobiasdiez:
I fear I cannot completely follow you @lenhard.
- If I understand the special fields correctly, then they sync the content of some field (like `rank = 1`) with the keywords field (`keywords = rank1`).
- The automatic groups take the contents of some field (e.g. `keywords = foo, bar`) and creates dynamic groups for each keyword (i.e. a group matching `foo` and one for `bar`). However, these groups are only created for the display and not written to groups tree in the bib file. Moreover, this features is completely independent of the `groups` field that is only used for the static groups.

Thus the only logic that seems to be shared by these features is the extraction of a list of keywords from a string and this is already encapsulated in the `KeywordList` class. What do I miss?

koppor:
Side topic: @stfnklb and me discussed whether we should use jrkeywords as
field. Reason: keywords are the keywords from the paper and users want to
keep them. - By this, we separate paper data from meta data on
bibliographical data...

Am 22.02.2017 20:57 schrieb "Tobias Diez" <notifications@github.com>:

lenhard:
@tobiasdiez From your answer, I think that you have understood me correctly.

It was just an idea and might not be worthwhile. Since the special fields synchronization already builds upon `KeywordList`, this might be as far as we can get. So, from my point of view you can proceed with this PR.

@koppor: Although what you say is true, I do not know of any user complaint about this behavior (which is telling). In contrast, changing the semantics of the keyword field in JabRef will confuse a lot of users, I would expect.

src/main/java/org/jabref/gui/groups/GroupDialog.java
tobiasdiez: According to the Java language specifications
> The type of the Expression must be char, byte, short, int, Character, Byte, Short, Integer, String, or an enum type, or a compile-time error occurs.

thus switch over a class type does not work.

tobiasdiez:
Ok, I merge now and will have a second look about the SpecialFields later.

Issue #245, opened 2015-10-19
Attached file delete and rename functionality 


ajbelle:
@koppor asked me to log an issue (feature request?) covering the "Delete" functionality currently available through the "Rename files" plugin https://sourceforge.net/p/jabref/discussion/318824/thread/8f56b49a/  Apparently plugins will not operate after JabRef 2.80 :-( While deleting attached files using the current plugin has the risks listed in the discussion, the manual alternative is a PIA making it "Attached File Delete" important if using JabRef as a literature repository manager, especially if merging files form existing repositories (e.g. EndNote). 
It would be intuitively clear where to find what is to be deleted to have the delete button option listed in data entry tab via the exisiting hover over file 'right mouse button' menu containing  “Open, Open Folder, Rename File, Move to file directory". . 

It should add the following to what the "Rename files" plugin already does.
- [ ] Checks whether the file is required by another entry and ask if deletion is still needed!
- [ ] Present a large Warning that is has no way of knowing if other .bib files link to the files selected  for deletion. Or have the ability to add a list BIB files (say in preferences) for checking and to check them for file links in the background.
- [x] Warn that deletion is immediate and permanent with no move to "Trash" and there is no <Undo>. Better still move the file to a "deleted file folder", making cross-platform implementation of safety net easier.
- [ ] Selecting files by highlighting entries for deletion when multiple files attached exist.
- [x] If it does not delete it should not automatically remove the link, or must warn that deletion was unsuccessful.
- [x] Have an option to "open containing folder" like PDF-XChange when deleting to manually check and clean up if desired (Endnote creates redundent folders that require deleting also). 
- [ ] Make it work for files linked through <Preferences><External programs>"Legacy file Fields". I notice that ""Move to file directory" is currently unable to negotiate the legacy settings for my configuration E:\CRC Share\registry\endnote\the.Data\PDF\
- [ ] It could be made "Enable at own risk" in setting so users don't use it without realising and tarnishing JabRefs excellent reliability reputation.
I know that is a lot of programming, but would make Jabref much more capable for repository management.

simonharrer:
In the current master branch, there is a "delete local file" action attached to the right-click menu in the generals tab for each file.

jhshinn:
It would be great if a hot key is assigned for the "delete local file" function.

ajbelle:
[...]

PR #2569, opened 2017-02-19
More safely delete files

stefan-kolb:
See #245

- confirmation before file deletion
- change language strings so the user knows this action is permanent


src/main/java/net/sf/jabref/gui/fieldeditors/FileListEditor.java
```diff
@@ -218,13 +218,18 @@ public void actionPerformed(ActionEvent actionEvent) {
...
             int row = getSelectedRow();
+
             // no selection
-            if (row != -1) {
+            if (row == -1) {
+                return;
+            }
 
+            int userConfirm = JOptionPane.showConfirmDialog(null, Localization.lang("Are you sure you want to delete this file?"), Localization.lang("Warning"), JOptionPane.YES_NO_OPTION);
+            if (userConfirm == JOptionPane.YES_OPTION) {
```
Siedlerchr: Please don't use simple YES/NO Buttons: 
 http://www.uxdesignedge.com/2010/06/are-you-sure-how-to-write-effective-confirmations/

Better add explicit named Buttons Delete file permanently and Cancel:
As I just did this yesterday for another case, here is a sample code:

```
String[] options = {Localization.lang("Move file"), Localization.lang("Cancel")}; 
int dialogResult = JOptionPane.showOptionDialog(frame, "Move file to file directory" + fileDir.get(),
              "Move",
                   JOptionPane.INFORMATION_MESSAGE, JOptionPane.YES_NO_CANCEL_OPTION, null, options, options[0]);
            if (dialogResult == JOptionPane.YES_OPTION) {
```

stefan-kolb: Hm, one can argue if this really adds so much value. I think the functionality is clear with the question `Do you really want to delete this file?`. It just adds more Strings that need to be translated. However, I'm OK with implementing this. 

In general, what do @JabRef/developers think about this?

Siedlerchr: There are just a few more Strings, but it really helps to improve usability a  lot! Here is a more detailed explanation . http://ux.stackexchange.com/questions/756/what-are-some-alternatives-to-the-phrase-are-you-sure-you-want-to-xyz-in-confi

See also https://github.com/koppor/jabref/issues/149

stefan-kolb: I understand it, tho not sure if I really agree 😄

PR #2575, opened 2017-02-20
Check similarity of entry when using DOI retrieval with ArXiV

### LINKED ISSUE #2570: ArXiV fetcher returns wrong entries when DOI is not in catalog
Every time I am using the quality -> lookup full text documents I get a pdf which is completely irrelevant
Any suggestions ?

I am using jabref 3.8.2 , java-runtime 8, antergos os , gnome 3.22.2

stefan-kolb:
Resolves #2570


lenhard:
Looks good to me! The failing push build is just because of the medline fetcher.

However, you keep forgetting to add a Changelog entry :-) Please add one.

PR #2579, opened 2017-02-21
Update IEEEJournalList

Ede123:
Update for `IEEEJournalListCode.txt` and `IEEEJournalListText.txt` compiled from [IEEEfull.bib](http://tug.ctan.org/tex-archive/macros/latex/contrib/IEEEtran/bibtex/IEEEfull.bib) and [IEEEabrv.bib](http://tug.ctan.org/tex-archive/macros/latex/contrib/IEEEtran/bibtex/IEEEabrv.bib) respectively.


Siedlerchr:
Thank you very much for your contribution. I am not an expert in the IEEE Abbrev codes but to me it looks good.

lynyus:
Would it be advisable to write an automated script that fetches those lists automatically before each release?

Ede123:
The files were last changed 2015/08/28. Probably not be worth the effort...

koppor:
Thank you for the PR! I think, syncing it with existing lists, is a very good idea!

Is there a script you wrote or is it manual effort? 🌞

Ede123:
No, half-auto using regex and Notepad++

Issue #1985, opened 2016-09-15
Enforce String literal inside Localization.lang() calls


stefan-kolb:
Introduced by the global search bar of Stupro.
We need to enforce that we use Strings or concatenated Strings inside our Localization.lang calls and no variables. otherwise our detection LocalizationConsistencyTest cannot work correctly.
Fixed by https://github.com/JabRef/jabref/commit/05e7f0def59b7c34707f9e0a1f3dc17db678d1f0.

We should accept a String in "" or a concatenated String " " +var +  "" ... as it is required sometimes for insertions %0.

stefan-kolb:
@JabRef/stupro @koppor Would be nice if Stupro can take care of this.

PR #2582, opened 2017-02-22
Localization tests

stefan-kolb:
Test for #1985

### DIFF
--- a/src/test/java/org/jabref/logic/l10n/LocalizationParser.java
+++ b/src/test/java/org/jabref/logic/l10n/LocalizationParser.java
@@ -290,7 +290,10 @@ public class LocalizationParser {
                     } else if (c == ')') {
                         brackets--;
                     }
-                    buffer.append(c);
+                    // skip closing brackets
+                    if (brackets != 0) {
+                        buffer.append(c);
+                    }
                     index++;
                 }
                 // trim newlines and whitespace
--- a/src/test/java/org/jabref/logic/l10n/LocalizationParser.java
+++ b/src/test/java/org/jabref/logic/l10n/LocalizationParser.java
@@ -75,6 +75,14 @@ public class LocalizationParser {
         }
     }
 
+    public static Set<LocalizationEntry> findLocalizationParametersStringsInJavaFiles(LocalizationBundleForTest type)
+            throws IOException {
+        return Files.walk(Paths.get("src/main"))
+                .filter(LocalizationParser::isJavaFile)
+                .flatMap(path -> getLocalizationParametersInJavaFile(path, type).stream())
+                .collect(Collectors.toSet());
+    }
+
     private static Set<LocalizationEntry> findLocalizationEntriesInJavaFiles(LocalizationBundleForTest type)
             throws IOException {
         return Files.walk(Paths.get("src/main"))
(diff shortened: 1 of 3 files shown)


stefan-kolb:
@JabRef/developers  Hm, our org.jabref move obviously lost all information on who did author the files :(
But I think this one was @simonharrer . I didn't understand your code 100% but tried to add a quick test for the localization problems. Tests are passing but I'm not sure whether I damaged your logic, especially the strange quotation replacements and rereplacements.

src/test/java/org/jabref/logic/l10n/LocalizationParser.java
```diff
@@ -75,6 +75,14 @@
         }
     }
 
+    public static Set<LocalizationEntry> findLocalizationParametersStringsInJavaFiles(LocalizationBundleForTest type)
+            throws IOException {
+        return Files.walk(Paths.get("src/main"))
```
Siedlerchr: You could directly use the Files.find method and pass the filter predicate as argument.

src/test/java/org/jabref/logic/l10n/LocalizationParser.java
```diff
@@ -259,6 +269,39 @@ public boolean containsKey(String key) {
...
+                matcher = LOCALIZATION_START_PATTERN.matcher(content);
+            } else {
+                matcher = LOCALIZATION_MENU_START_PATTERN.matcher(content);
+            }
+            while (matcher.find()) {
+                // find contents between the brackets, covering multi-line strings as well
+                int index = matcher.end();
+                int brackets = 1;
+                StringBuilder buffer = new StringBuilder();
+                while (brackets != 0) {
```
Siedlerchr: Look in our StringUtil class, there is already a method for finding braces

src/test/java/org/jabref/logic/l10n/LocalizationParser.java
stefan-kolb: Thanks for the hints, but this is old code which I didnt write and don't wanna change if not really necessary.

stefan-kolb: Just looked at it. Doesn't seem like a real change to me except that its a parameter then?!

src/test/java/org/jabref/logic/l10n/LocalizationParser.java
stefan-kolb: Thanks for the hints, but this is old code which I didnt write and don't wanna change if not really necessary.

Siedlerchr: But you created a test for it right? So you could check it....

src/test/java/org/jabref/logic/l10n/LocalizationParser.java
Siedlerchr: A bit different. Instead of collecting all files and then filtering on the list(the stream), the Files.find directly invokes the predicate on each file and the file is only included in the stream if it matches:

>  This method walks the file tree in exactly the manner specified by the walk method. For each file encountered, the given BiPredicate is invoked with its Path and BasicFileAttributes. The Path object is obtained as if by resolving the relative path against start and is only included in the returned Stream if the BiPredicate returns true. Compare to calling filter on the Stream returned by walk method, this method may be more efficient by avoiding redundant retrieval of the BasicFileAttributes. 

https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#find-java.nio.file.Path-int-java.util.function.BiPredicate-java.nio.file.FileVisitOption...-

>  look in FileUtil.findFiles

PR #2583, opened 2017-02-22
Keyword - Special field synchronization

### LINKED ISSUE #2580: Editing keywords clears priority, ranking, quality and relevance
JabRef version 3.8.2 on Windows 10

Editing keywords clears the fields priority, ranking, quality and relevance

Steps to reproduce:

__Write values of special fields as separate fields to BiBTeX__ enabled!
1. Add a value to any of the fields: priority, ranking, quality and relevance
2. Edit the keyword field
3. Save or change focus from the keywords field in the entry editor

The fields priority, ranking, quality and relevance are now removed.

No events reported in the log

stefan-kolb:
Fixes #2580 
Synch should now work appropriately.

Few questions here @koppor and all others:

- [x] If keyword sync is enabled, there is the possibility that we have multiple `rank1, rank4` keywords.
Only the first one is synched with the special field. Was this always like this? Seems like a bug to me. 

- [x] If we change the preferences from write special fields to file to keyword synchronization the data gets lost. Did I introduce this or was this a bug?


stefan-kolb:
@lenhard talked to @koppor This was always the case. writable fields and keyword fields were never synchronized. we have to defer this until a reimplementation.

PR #2585, opened 2017-02-23
Replace URLDownloading logic

stefan-kolb:
Imho the Downloading logic should be based on some stable HTTP library.

Tho I'm not sure which one to choose. Some handle JSON very well, too.
@JabRef/developers WDYT?
- https://github.com/Mashape/unirest-java Is convenient but seems to be not so well maintained
- http://hc.apache.org/index.html Probably the best alternative but more native code
- https://github.com/google/google-http-java-client
- http://square.github.io/okhttp/

Also, it seems like we already have a few of these libraries inside our code and should standardize on one?

See also: http://stackoverflow.com/questions/1322335/what-is-the-best-java-library-to-use-for-http-post-get-etc


src/test/java/org/jabref/logic/net/MimeTypeDetectorTest.java
```diff
@@ -31,35 +32,29 @@ public void handlePermanentRedirections() {
                 )
         );
 
-        assertTrue(MimeTypeDetector.isPdfContentType(redirectedUrl));
+        assertTrue(new URLDownload(redirectedUrl).isMimeType("application/pdf"));
```
tobiasdiez: Introduce the shortcut method `isPdf` or create an enum for the mime types? I don't like the magic strings...

src/main/java/org/jabref/logic/importer/fetcher/IEEE.java
```diff
@@ -54,7 +54,7 @@
             if (doi.isPresent() && doi.get().getDOI().startsWith(IEEE_DOI) && doi.get().getURI().isPresent()) {
                 // Download the HTML page from IEEE
                 String resolvedDOIPage = new URLDownload(doi.get().getURI().get().toURL())
-                        .downloadToString(StandardCharsets.UTF_8);
+                        .asString(StandardCharsets.UTF_8);
```
tobiasdiez: introduce overload of `asString` which uses UTF_8 by default.

Issue #1407, opened 2016-05-17
Problem removing group with long names


juniorcaemj:
JabRef version 3.2 on Windows 10

User Interface Problem

If a group with a long name been removed, the JOptionPane box is created too large, vanishing "Yes" and "No" buttons.

Steps to reproduce:
1. Create a group with a long name
2. Try remove group

![1](https://cloud.githubusercontent.com/assets/17187418/15323642/6ec92f58-1c19-11e6-863d-e982b50b2b90.jpg)
![2](https://cloud.githubusercontent.com/assets/17187418/15323643/6ed08b54-1c19-11e6-9f18-da8d28d4bef0.png)

matthiasgeiger:
Thanks for your report.

If this is a real problem for you, simply delete the group just hit "enter" as "Yes" is preselected.

However, this seems to be only a theoretical problem, so we might fix this in a later version but first concentrate on other more important aspects.

juniorcaemj:
Hello @matthiasgeiger, I report this problem because it's a work (especifically UI Tests) for a Software Testing course, and we utilize JabRef as a case study. So I found interesting report it to the community.

simonharrer:
Thanks for the info. We welcome that JabRef is used in testing case studies. :+1: 

What would be the expect solution? Truncate the group name after a specific number of letters?

We also have UI tests in place at https://github.com/JabRef/jabref/tree/master/src/integrationTest/java/net/sf/jabref/gui - you could also have a look there. They are only the tip of the iceberg, and need further improvements. :)

matthiasgeiger:
Nice to hear, that JabRef is used in a practical Software Testing course. :+1: 

Hopefully not only as a bad example :wink:  We already know that the test coverage of our code should be improved, and we are already working on it. But especially on the field of UI testing there are only a few tests present - as Simon just mentioned.

If you are implementing practical, executable tests in your course we'd love to integrate them into our tests - if this is possible!

If you have other findings, or tips for us to improve JabRef just contact us!

juniorcaemj:
@simonharrer I thinking in this solution too, limit the size of group name, doesn't make any sense a group with a very very long name..

@matthiasgeiger I will read the documentation and submit my tests (and maybe a solution), as soon as I can.

stefan-kolb:
[...]

Issue #2561, opened 2017-02-17
Parsing entry with unbalanced braces doesn't show error message


AEgit:
JabRef 3.8.2
windows 10 10.0 amd64 
Java 1.8.0_121

I think this is not actually a bug (although I'm not sure), but I just wanted to report it as a reference in case someone comes across this problem.

Steps to reproduce:
1. Add the following entry to your JabRef database:

```
@Article{Redding2006,
  author   = {Redding, David W. and Mooers, Arne {\O}.},
  title    = {{Incorporating {evolutionary} {measures} into {conservation} {prioritization}}},
  journal  = {Conservation Biology},
  year     = {2006},
  volume   = {20},
  pages    = {1670--1678},
}
```

2. Check the entry in the entry preview. It is correctly displayed as:

> Redding, D. W. & Mooers, A. Ø. Incorporating evolutionary measures into conservation prioritization Conservation Biology, 2006, 20, 1670-1678

3. Now change the special character of the second author's name. Change `{\O}.` to `{\{O}}.`. You end up with the following entry:

```
@Article{Redding2006,
  author   = {Redding, David W. and Mooers, Arne {\{O}}.},
  title    = {{Incorporating {evolutionary} {measures} into {conservation} {prioritization}}},
  journal  = {Conservation Biology},
  year     = {2006},
  volume   = {20},
  pages    = {1670--1678},
}
```

4. Save the database and close JabRef.
5. Restart JabRef and reopen the database. The entry preview of the article is no longer correctly displayed. Instead, the following is shown:

> Redding, D. W. & Mooers, A. Ø#.#.

Indeed, the BibTex source panel also shows just:

```
@Article{Redding2006,
author   = {Redding, David W. and Mooers, Arne {\{O}}.}
```

6. Close JabRef and open the bibtex file with a text editor of your choice. Search for Redding2006. As you can see, the article entry is still complete. Change the author's name back to `{\O}`. The entry should look again like this:

```
@Article{Redding2006,
  author   = {Redding, David W. and Mooers, Arne {\O}.},
  title    = {{Incorporating {evolutionary} {measures} into {conservation} {prioritization}}},
  journal  = {Conservation Biology},
  year     = {2006},
  volume   = {20},
  pages    = {1670--1678},
}
```

7. Restart JabRef and open the database. The entry is again correctly displayed.

[...]

PR #2587, opened 2017-02-24
Add "Remove group and subgroups" option

tobiasdiez:
Issue #1407 is fixed in the process (long names are handled well by default in javafx, see screenshot below).
snipimage


src/main/java/org/jabref/gui/groups/GroupSelector.java
```diff
@@ -407,13 +403,13 @@ private void showPopup(MouseEvent e) {
             if (node.canBeEdited()) {
                 editGroupPopupAction.setEnabled(false);
                 addGroupPopupAction.setEnabled(false);
-                removeGroupAndSubgroupsPopupAction.setEnabled(false);
+                //removeGroupAndSubgroupsPopupAction.setEnabled(false);
```
stefan-kolb: Why only commented out?

src/main/java/org/jabref/gui/groups/GroupSelector.java
```diff
@@ -407,13 +403,13 @@ private void showPopup(MouseEvent e) {
...
                 removeGroupKeepSubgroupsPopupAction.setEnabled(false);
             } else {
                 editGroupPopupAction.setEnabled(true);
                 addGroupPopupAction.setEnabled(true);
                 addGroupPopupAction.setNode(node);
-                removeGroupAndSubgroupsPopupAction.setEnabled(true);
+                //removeGroupAndSubgroupsPopupAction.setEnabled(true);
```
stefan-kolb: Why only commented out?

src/main/java/org/jabref/gui/groups/GroupSelector.java
tobiasdiez: The new button is always enabled which is not what you want. Thus this old code functions as a reminder for myself to change it.

lenhard:
If I newly create groups in some bib file and delete them using the options, everything works.

However, I just tested this with our large group file and unfortunately it does not work. I right-click a group (for instance "Life" as the first group in the large bib file), select "Remove group and subgroups", click ok in the message box. Then, nothing happens and the groups stay as they are. There are no messages in the log.

Am I missing something? Can you maybe have a try with our huge bib file?

lenhard:
Looks like other people see a group-related performance degradation: https://github.com/JabRef/jabref/issues/2561#issuecomment-283417338 So, this is most likely not specific to this PR.

tobiasdiez:
@lenhard I can't reproduce the problem with the huge bib file. Thus I will merge this PR for now and would ask you to open a new issue with more details of how to reproduce it.

Issue #1904, opened 2016-09-01
Proposed enhancement: Search for groups


AEgit:
I propose to add a search functionality just for groups (unfortunately my Java knowledge is nearly non-existent, so I can only propose this, but not contribute). For large databases with many groups (I have >900 groups) it is a bit of a hassle having to scroll to the group of interest. It would be nice to have a search feature (similar to what is already available articles) just for groups. When searching for a group you could then immediately hop to the group of interest, instead of having to scroll to it.

Siedlerchr:
@oscargus  Relates to #1876, Is there a "group" pseudo field already?

oscargus:
One can search for groups=xxx in the normal search and that will highlight/filter the entries in that group.

@AEgit Is that enough or do you mean having a search box in the groups side panel to easily filter out groups there? (Which I think is a quite reasonable way to implement it.)

AEgit:
@oscargus 

I was thinking about a search box in the groups sides panel - the issue with filtering is, that you still have to scroll in the group side panel, if you want to see the actual group (not just the articles entries associated with the group). Reason for this could be, that you want to assign the group to another group (as a subgroup) - currently that can only be done by scrolling to the actual group.

Furthermore, by filtering you are not able to see entries, that are part of a subgroup of the group that you are searching for.

So, if it is possible, I would really appreciate to see an additional search box in groups side panel to filter out groups.

tobiasdiez:
Good idea!
I think, I'll try to implement it during the redesign of the groups interface (work-in-progress, https://github.com/JabRef/jabref/pull/1438).

AEgit:
Now that version 3.7 has been released, could we reopen this please? I would still be very interested in this functionality, and, as far as I can tell, the redesign of the groups interface (https://github.com/JabRef/jabref/pull/1438) is currently on hold, so there might be other options to implement this?

matthiasgeiger:
This is quite some effort. And as the new interface will eventually replace the current one, we won't implement it for now as we would have to implement it twice, which is not really reasonable with our limited resources.

AEgit:
Hmmm, ok - could we then reopen "Groups JavaFX Rework" thread (https://github.com/JabRef/jabref/pull/1438)? I just want to make sure that this problem is not forgotten. Or do people remember this issue, even when it's closed?

tobiasdiez:
We remember everything! 😄 
[...]

PR #2588, opened 2017-02-24
Implement #1904: filter groups

tobiasdiez:
Add textbox add the bottom of the groups panel which allows for filtering the groups by name. Implements #1904.

snipimage 002

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
     - Redesigned group panel.
       - Number of matched entries is always shown.
       - The background color of the hit counter signals whether the group contains all/any of the entries selected in the main table. 
+      - Added a possibility to filter the groups panel [#1904](https://github.com/JabRef/jabref/issues/1904)
       - Removed edit mode.
     - Redesigned about dialog.
     - Redesigned key bindings dialog.
--- a/src/main/java/org/jabref/gui/groups/GroupNodeViewModel.java
+++ b/src/main/java/org/jabref/gui/groups/GroupNodeViewModel.java
@@ -27,6 +27,7 @@ import org.jabref.model.groups.AbstractGroup;
 import org.jabref.model.groups.AllEntriesGroup;
 import org.jabref.model.groups.AutomaticGroup;
 import org.jabref.model.groups.GroupTreeNode;
+import org.jabref.model.strings.StringUtil;
 
 import com.google.common.eventbus.Subscribe;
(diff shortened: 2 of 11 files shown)


Siedlerchr:
Great idea. However, I would place the search bar on top of the screen and not at the bottom, or alternatively is there a way to reuse the existing search bar and add an option to search for groups?

tobiasdiez:
Maybe it's indeed a good idea to move the search bar and button to the top...as this is an easy change I would postpone it for the moment and see what kind of further user controls are necessary in the group pane.

stefan-kolb:
On the bottom looks good to me.

lenhard:
I think having the search box at the bottom is fine for now, since we can also change it easily.

The feature works nicely with normal-size bib files, but I just checked it with the large bib file and there the feature scales very badly. If I enter something into the search field, it freezes and nothing happens for one or two minutes until the rearrangement of the groups appears (as desired). 

Do you see a chance to speed this up? Maybe some parallelism, using `parallel()` in streams, could help?

lenhard:
Looks like other people see a group-related performance degradation: https://github.com/JabRef/jabref/issues/2561#issuecomment-283417338 So, this is most likely not specific to this PR and you could go for a merge.

However, we really need to follow up with a performance improvement PR.

lenhard:
Since, as indicated in https://github.com/JabRef/jabref/issues/2561#issuecomment-283494292 the degraded performance for large bibs is expected at the moment and will be tackled separately, I think this is ready for merge (which I am doing right away).

matthiasgeiger:
One suggestion as I just stumbled over it: I think the searching should be case insensitive (although this might further slow down the searching ;-))

AEgit:
Should the performance issue with the groups filtering (which is still massive) also be added to the 4.0 milestone?

lenhard:
@AEgit I am not exactly sure which issue you are referring to, but yes: We should try to improve performance before a final release. So if there is an open issue that should be in 4.0, please point me to it!

AEgit:
@lenhard The performance issue I'm referring to happens with large databases (>10k entries, ~1k static groups) in conjunction with the groups filtering. When you try to find a group using the newly implemented groups search/filter, you have to wait for several minutes for the search to complete. During that time JabRef behaves as if it had crashed. You mentioned this problem yourself in this comment:
https://github.com/JabRef/jabref/pull/2588#issuecomment-283042778

Should I open a new ticket for this?

lenhard:
@AEgit Yes, please open a new ticket for this, so that we do not forget it.

AEgit:
Done ;)
https://github.com/JabRef/jabref/issues/2852

PR #2592, opened 2017-02-27
Fix pdf file import when only one file selected and no entry created

### LINKED ISSUE #2577: PDF importation freezes when filename contains a curly bracket
When I try to import PDFs via the Quality -> Find unlinked files, the process freezes when a PDF containing a curly bracket ( "{" or "}" ) in its filename is encountered.

JabRef version 4.0.0-dev (snapshot 2017-02-18) on Win 10 (1607)

Steps to reproduce:

1. Open Quality -> Find unlinked files
2. Select a folder containing at least one PDF with a curly bracket in its name
3. Scan directory
4. Make sure the offending file is selected
5. Apply

After the last step, the process goes through all valid files (showing "Importing into database... M of N"), but freezes the whole JabRef once it gets to the offending file.

Siedlerchr:
Fixes #2577 
Problem is not the filename 
Problem: You still get a warning that one file could not be imported, although it successfully worked.

I have unfortunately no idea if this code is at all correct, the DroppedFileHandler stuff looks suspicious to me, as it is normally used wit Drag and drop.

**Question**: Is there somewhere a method for linking pdfs?


lenhard:
I agree with Tobias. I added a changelog entry and am merging directly. 

I also tried it out locally according to the description listed in #2577 and can confirm that the error no longer appears. I would not care about the warning that still appears, it is sort of valid (nothing was imported).

PR #2601, opened 2017-03-02
Remove OrdinalsToSuperscriptFormatter from recommended biblatex save actions

### LINKED ISSUE #2596: The JabRef goes mad after clicking `Enable saving actions` in database properties.
Version information:
```
JabRef 3.8.2
windows 10 10.0 amd64 
Java 1.8.0_121
```

Steps to reproduce:

1. Open `database properties` dialog, and Click `Enable saving actions`.
2. Using `Tasks Manager` in Windows 10, you will see JabRef costing a high quota of CPU, esp. higher than 30% for my laptop.
3. The directory contains the `BibTex` file will continuously generate new *.sav.log file and delete it again and again.
4. When open the error console, there is nothing in the `Exception` tab.

Only one log of opening my database found in the error console
```
20:15:08.224 [AWT-EventQueue-0] INFO  net.sf.jabref.logic.importer.OpenDatabase - Opening: D:\git\JabRefBibs\MyFullResearch.bib
```

Additionally, when I disabled the saving actions, everything becomes still at once.

lenhard:
Fixes #2596

The LatexToUnicode is conflicting with the OrdinalsToSuperscriptFormatter but both are recommended for the same fields for biblatex. This PR removes the latter. Since biblatex is usually used together with UTF-8 and Unicode conversion should be sufficient.


Siedlerchr:
That's really weird. I never had problems with that setting

koppor:
Yeah, I worked with that for ages. 🌞

Siedlerchr:
Is this maybe due to the new lib?

lenhard:
Yes, I would expect that this is due to the new lib. I assume that it leads to the following:
1. Ordinals formatter converts `2nd` to `2\textsuperscript{nd}`
2. Latex formatter converts `2\textsuperscript{nd}` to `2ⁿᵈ`
3. Ordinals formatter interprets `2ⁿᵈ` as `2nd` and converts it to `2\textsuperscript{nd}`
4. And so on

I might be wrong though, I did not debug this. 

@Siedlerchr @koppor: Do you have problems with it now? What would you recommend instead?

koppor:
When using the defaults in the "Cleanup entries" dialog, everything works fine.

grafik

Maybe, it is possible to have different defaults for save actions and for cleanup? I really need that functionality.

lenhard:
@koppor The problem are the defaults for biblatex, not for BibTeX, which are depicted in your screenshot above.

An alternative solution, and maybe even a better one, might be to have the ordinals formatter correctly recognize 2ⁿᵈ as ordinals, so that it does not try to replace that. Should be doable, but right now I cannot make any promises time-wise. So, if anybody wants to take this over, be my guest.

koppor:
I am not sure whether all unicode fonts support 2ⁿᵈ. Therefore, I really like the idea to have the ordinals formatter recognize that.

matthiasgeiger:
I second Jörgs and Tobias' opinion...

koppor:
I really like the ordinals-to-suprscript formatter as it provides great formatted bibliographies.

People write "2nd conference on something", which should appear as "2ⁿᵈ conference on something" in the proceedings. If I understand the fix correctly, biblatex papers won't have the superscript (more or less) automatically from now on.

Since most publishers use bibtex, the solution for me is to have an unclean personal biblatex file and always to the bibtex conversion when working on a paper. The superscript will be injected when copying the entry to the paper by  a [cleanup](http://help.jabref.org/en/CleanupEntries)..

Refs https://github.com/JabRef/jabref/issues/160.

Issue #2604, opened 2017-03-03
BibTeX Key Pattern results don't match with expectations


RolfStarre:
Hi!

I encounterd some unexpected behaviour in the creation of keys, I'm using JabRef 3.8.2 on windows 10 10.0 amd64 Java 1.8.0_121.  

Steps to reproduce:

1. In Preferences -> BibTeX key generator set the Default pattern field to [title][shorttitle]
2. Generate the key, for example for the document in the screenshot

Based on the documentation on http://help.jabref.org/en/BibtexKeyPatterns I was expecting the following output:
For [title]: HeuristicsandUsabilityGuidelinesfortheCreationandEvaluationofFuninVideogames
For [shorttitle]: Heuristicsandusability 

Intead I got Heuristicsandusabilityguidelinesforthecreationandevaluationoffuninvideogames
and Heuristicsusabilityguidelines.  

![image](https://cloud.githubusercontent.com/assets/5969100/23546773/6e02e358-0001-11e7-97b5-d3da4b69be9b.png)

```
Put the excerpt of the log file here
```

lenhard:
Related to #2589

Since this is the second report regarding key patterns in a short time, maybe there are some general issues with key patterns at the moment.

RolfStarre:
Hey, thanks for the response :)

I was working on it but I noticed afterwards that @Siedlerchr was already working on it as well in this [pull request](https://github.com/JabRef/jabref/pull/2597), so I wasn't sure if I should continue.

Fixing the [shorttitle] option shouldn't be too hard, but I was unable to find where the [title] option is handled in the function makeLabel in the class BibtexKeyPatternUtil.

Siedlerchr:
@RolfStarre  Thanks for your interest, as I had not yet the time to investigate why the Google Guava solution does not really work as expected, you could take a look at it. Or we could use your solution, I saw that you already added a test for it... Feel free to propose your PR here.
We are always looking contributors...;)

Regarding the [title] option, it is handled implicity in the else-part where just the field content is returned:
` return entry.getFieldOrAlias(val).orElse("");`

RolfStarre:
Cool thanks, I'll continue working on it then :)

I'm going to try to fix the shortTitle and title cases as well then before I create the pull request. 
I'll try to see what the Google Guava does and if I can use it.

Siedlerchr:
This can be closed right?

RolfStarre:
Yup 👍

Issue #2589, opened 2017-02-24
Camel key doesn't seem to be working


traversc:
Hi, I can't get the camel key in the key pattern to work for some reason.  I'm using the latest release: 

Here is my key pattern: 
[auth][year][camel]

Here is my bibtex file: 

> % Encoding: UTF-8
> 
> @Article{Trapnell2009,
>   author  = {Trapnell, Cole and Pachter, Lior and Salzberg, Steven L.},
>   title   = {TopHat: discovering splice junctions with RNA-Seq},
>   journal = {Bioinformatics},
>   year    = {2009},
>   volume  = {25},
>   number  = {9},
>   pages   = {1105-1111},
>   issn    = {1367-4803, 1460-2059},
>   doi     = {10.1093/bioinformatics/btp120},
>   type    = {Journal Article},
>   url     = {http://bioinformatics.oxfordjournals.org/content/25/9/1105
> http://www.ncbi.nlm.nih.gov/pubmed/19289445
> http://bioinformatics.oxfordjournals.org/content/25/9/1105.abstract},
> }
> 
> @Comment{jabref-meta: databaseType:bibtex;}
>

lenhard:
Hi @traversc 

What exactly is the error? Looking at the description here http://help.jabref.org/en/BibtexKeyPatterns I would say that the key `Trapnell2009` is exactly what I would expect from the key pattern `[auth][year][camel]`. 

Do you expect more author names? Then you should use `[authors][year][camel]`.

traversc:
Apologies if I misunderstood the documentation.  

It says: 

>[camel]: Capitalize and concatenate all the words of the title. For example, An awesome paper on JabRef will become AnAwesomePaperOnJabref

So following that pattern when I use just [camel], shouldn't it output: TophatDiscoveringSpliceJunctionsWithRnaseq?  However, I just get a blank.  

Thanks

Siedlerchr:
@traversc You have to use` [title][camel]` to generate the key
`TopHat:discoveringsplicejunctionswithRNA-Seq`

Siedlerchr:
I took a look at the code and interestingly there is no check for the [camel] option.  I will provide a fix

matthiasgeiger:
Hi! 
Thanks to @RolfStarre the problem should be fixed in the recent development builds. 
You can check out fixed versions at: https://builds.jabref.org/master

PR #2610, opened 2017-03-03
Fix title-related key patterns in BibtexKeyPatternUtil

RolfStarre:
Related to [#2604](https://github.com/JabRef/jabref/issues/2604) and [#2589](https://github.com/JabRef/jabref/issues/2589)

In the class BibtexKeyPatternUtil I've added cases for [title] and [camel] to try and make them conform to the [documentation](http://help.jabref.org/en/BibtexKeyPatterns).

@Siedlerchr was also working on this issue [here](https://github.com/JabRef/jabref/pull/2597). I found out that  
`CaseFormat.LOWER_CAMEL.to(CaseFormat.UPPER_CAMEL, entry.getField(FieldName.TITLE).orElse("").replaceAll("\\s+", ""));`
didn't work because to be able to transform the title into upper_camel this way the title has to be in  lower_camel case initially, which is often not the case.

I've added some test cases and changed a few others to conform to the documentation. 
I still need to take a look at the failing tests.

### DIFF
--- a/src/main/java/org/jabref/logic/bibtexkeypattern/BibtexKeyPatternUtil.java
+++ b/src/main/java/org/jabref/logic/bibtexkeypattern/BibtexKeyPatternUtil.java
@@ -8,6 +8,7 @@ import java.util.Locale;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Scanner;
+import java.util.StringJoiner;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -791,7 +792,7 @@ public class BibtexKeyPatternUtil {
      * Capitalises the significant words of the "title" field in the given BibTeX entry
      */
     public static String camelizeSignificantWordsInTitle(String title) {
-        StringBuilder stringBuilder = new StringBuilder();
+        StringJoiner stringJoiner = new StringJoiner(" ");
         String formattedTitle = formatTitle(title);
         Boolean camelize;
 
@@ -808,25 +809,22 @@ public class BibtexKeyPatternUtil {
                     }
                 }
                 // We want to capitalize significant words and the first word of the title
-                if (camelize || (stringBuilder.length() == 0)) {
+                if (camelize || (stringJoiner.length() == 0)) {
                     word = word.substring(0, 1).toUpperCase() + word.substring(1);
                 } else {
                     word = word.substring(0, 1).toLowerCase() + word.substring(1);
                 }
 
-                if (stringBuilder.length() > 0) {
-                    stringBuilder.append(' ');
-                }
-                stringBuilder.append(word);
+                stringJoiner.add(word);
             }
         }
 
-        return stringBuilder.toString();
+        return stringJoiner.toString();
     }
 
 
     public static String removeSmallWords(String title) {
-        StringBuilder stringBuilder = new StringBuilder();
+        StringJoiner stringJoiner = new StringJoiner(" ");
         String formattedTitle = formatTitle(title);
 
         try (Scanner titleScanner = new Scanner(formattedTitle)) {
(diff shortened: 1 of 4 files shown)


Siedlerchr:
Codewise looks good! 👍  Thank you very much for your contribution. Would you please check the failing tets? (Just click Details on the Red-Marked Travis Build and you are able to see the log). From what I saw there is jut another unit test which tests key generation which has to be adapted!

RolfStarre:
@Siedlerchr 
I've taken a look at the failing tests and after making a small change to the code and to some tests it's now working.

Some tests failed because most of the formatter classes only work if the input is a non concatenated string.

Siedlerchr:
From my point of view it looks good.  However, another dev should take a look, too.
 If you could just add a changelog entry it would be nice. Otherwise we can create it on merge.

src/test/java/org/jabref/logic/bibtexkeypattern/MakeLabelWithDatabaseTest.java
```diff
@@ -191,21 +191,21 @@ public void generateKeyEmptyFieldColonInDefaultText() {
...
-        assertEquals(Optional.of("AapoJ"), entry.getCiteKeyOptional());
+        assertEquals(Optional.of("AAPoJ"), entry.getCiteKeyOptional());
     }
 
     @Test
     public void generateKeyShorttitle() {
         bibtexKeyPattern.setDefaultValue("[shorttitle]");
         BibtexKeyPatternUtil.makeAndSetLabel(bibtexKeyPattern, database, entry, preferences);
-        assertEquals(Optional.of("awesomepaperJabRef"), entry.getCiteKeyOptional());
+        assertEquals(Optional.of("Anawesomepaper"), entry.getCiteKeyOptional());
```
koppor: I don't get, why JabRef is missing here.

src/main/java/org/jabref/logic/bibtexkeypattern/BibtexKeyPatternUtil.java
```diff
@@ -742,18 +748,122 @@ private static String getTitleWordsWithSpaces(int number, String title) {
...
+                String word = titleScanner.next();
+
+                for (String smallWord : Word.SMALLER_WORDS) {
+                    if (word.equalsIgnoreCase(smallWord)) {
+                        continue mainl;
+                    }
+                }
+
+                if (stringBuilder.length() > 0) {
+                    stringBuilder.append(' ');
```
koppor: Normally, we would use StringJoiner to avoid this kind of code.

RolfStarre: Ah I copied this from some of the existing code, I'll try to change it to use a StringJoiner!

I'll also add a changelog entry.

koppor: Thank you! :+1: We're just trying to improve the code quality wherever possible 😇

RolfStarre: No problem :)

I've made the suggested code change and added a change to the changelog.

koppor:
I wonder whether it would be possible to add test cases to MakeLabelWithDatabaseTest.java with all patterns listed at http://help.jabref.org/en/BibtexKeyPatterns.

I saw that you also added `[camel]` and added a test for that in BibtexKeyPatternUtilTest.java. To test the "user experience", I would like to ask to add some tests for that in MakeLabelWithDatabaseTest.

Strangely, `camel".equals(val)` is not test covered even though it should?!

grabbed_20170305-212351

https://codecov.io/gh/JabRef/jabref/src/ee7445652023f95e7c623c75c93058b9f2f4adf9/src/main/java/org/jabref/logic/bibtexkeypattern/BibtexKeyPatternUtil.java

If it is too much effort, we can just go ahead and merge!

RolfStarre:
I have added a test for the [camel] case to MakeLabelWithDatabaseTest.java
Hope the `camel".equals(val)` is covered now :)

Are there any additional test cases I should add?

koppor:
Yeah, it's covered: https://codecov.io/gh/JabRef/jabref/src/6aa925b7ec6fc2e1deb497a898e076bc747630da/src/main/java/org/jabref/logic/bibtexkeypattern/BibtexKeyPatternUtil.java

I'll go ahead with the merge and look forward to other tests / general improvements. We have a huge list of smaller tasks at https://github.com/koppor/jabref/issues.

Thank you for the good work!

koppor:
We just discovered that our documentation was wrong and that `camel` should be a modifier: `title:camel`.

It should work with `title:capitalize`.

See https://github.com/koppor/jabref/issues/237

RolfStarre:
Ah OK, so the camel case should also be moved in the code 😃

PR #2567, opened 2017-02-18
Fix NPE on tab switching in EntryEditor between Mr Dlib

Siedlerchr:
and File Annotations
Adjust formatting

~~- [ ] Change in CHANGELOG.md described~~
~~- [ ] Tests created for changes~~
~~- [ ] Screenshots added (for bigger UI changes)~~
- [X] Manually tested changed features in running JabRef
~~- [ ] Check documentation status (Issue created for outdated help page at [help.jabref.org]~~(https://github.com/JabRef/help.jabref.org/issues)?)
~~- [ ] If you changed the localization: Did you run `gradle localizationUpdate`?~~

PR #2616, opened 2017-03-05
Fixes displaying of Mr DLib recommendations

koppor:
Fixes https://github.com/koppor/jabref/issues/231

Mr DLib did never load the recommendations, because of an NPE fix (#2567)

### DIFF
--- a/src/main/java/org/jabref/gui/entryeditor/EntryEditor.java
+++ b/src/main/java/org/jabref/gui/entryeditor/EntryEditor.java
@@ -1148,7 +1148,7 @@ public class EntryEditor extends JPanel implements EntryContainer {
                 }
 
                 // When the tab "Related articles" gets selected, the request to get the recommendations is started.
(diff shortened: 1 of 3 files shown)


src/main/java/org/jabref/gui/entryeditor/EntryEditor.java
```diff
@@ -1148,7 +1148,7 @@ public void stateChanged(ChangeEvent event) {
                 }
 
                 // When the tab "Related articles" gets selected, the request to get the recommendations is started.
-                if (activeTab instanceof EntryEditorTabRelatedArticles) {
+                if (activeTab == relatedArticlePanel) {
```
Siedlerchr: Hm. I really wonder why this previously didn't work. But well, if it works now, then it's fine...

koppor: EntryEditorTabRelatedArticles was AFAIK used as child. `activeTab` was a `JPanel`.

PR #2623, opened 2017-03-06
Fix2613

### LINKED ISSUE #2613: Cannot call TexStudio from JabRef
JabRef version 3.8.2on 

I've set the right path /Applications/texstudio.app in the preference, but when I click on the icon in menu bar I receive the message: Could not call executable '/Applications/texstudio.app'

 Thanks for the support.

chochreiner:
Provided a MacOS specific routine for opening Tex editors to import citations to fix #2613.
The solution has been tested with Texstudio and Texmaker

- [x ] Change in CHANGELOG.md described
- [x ] Manually tested changed features in running JabRef


Siedlerchr:
Looks good to me as a small fix. In general I wondered if it isn't useful to integrate the PushToApplication somewhere in the desktop.os classes. I mean there are already methods for opening a file/with an application for each os.

stefan-kolb:
LGTM 👍 Would also love to see more OS-specific stuff integrated in the Desktop classes.

PR #2624, opened 2017-03-06
Refactored around the PDF Annotation Importer

lynyus:
- I mainly refactored the PdfAnnotationImporter. It now uses java.nio and has sane error handling.
- Furthermore, I made the FileAnnotations immutable.
- Deleted an unused (was renamed but not deleted) Interface
- The only actual change for the user is that the dates from the FileAnnotations shown in the FileAnnotationsTab are nicely formatted now.

- I am not so satisfied with the Optional field in FileAnnotation. However, I have not come up with a better solution to model that relationship from highlighted text to the sticky note attached to it.


src/main/java/org/jabref/model/pdf/FileAnnotation.java
```diff
@@ -1,127 +1,116 @@
...
+        this.annotationType = annotation.getSubtype();
+        this.linkedFileAnnotation = Optional.of(linkedFileAnnotation);
     }
 
-    public String getAuthor() {
-        return author;
-    }
+    private String prettyPrint(String date) {
+        // Sometimes this is null, not sure why.
+        if (date == null) {
```
lynyus: This seems to be the case if the guava cache is loaded.

koppor:
Current state: LGTM 👍

src/main/java/org/jabref/model/pdf/FileAnnotation.java
```diff
@@ -1,127 +1,116 @@
...
-    private String commentId;
-    private String author;
-    private String date;
-    private int page;
-    private String content;
-    private String annotationType;
-    private FileAnnotation linkedFileAnnotation;
     private final static int ABBREVIATED_ANNOTATION_NAME_LENGTH = 45;
+    private final static String ANNOTATION_DATE_REGEX = "D:\\d+";
+    public final String author;
```
lenhard: I know that this is not a problem of mutability here, but is there a good reason for breaking with the Java conventions for this and the following attributes? If there is no added benefit, then I would much prefer to have public getters instead of public attributes for the simple reason that this is the convention. Otherwise it always looks fishy when such a variable is accessed in another class and I have to look up `FileAnnotation` to check that this is not a grave access bug.

lynyus:
Hm good point. I personally think that this mentioned convention ist quite obsolete nowadays unless you work with a framework that requires it.People blindly generate trivial setters and getters (see diff) for all attributes driving the principle of information hiding ad absurdum and bloating up the code with unnecessary methods.However, if this is the consensus that we stick with the getters convention I ofc can change it.On Mar 7, 2017 8:52 AM, Jörg Lenhard <notifications@github.com> wrote:@lenhard commented on this pull request.



In src/main/java/org/jabref/model/pdf/FileAnnotation.java:
>      private final static int ABBREVIATED_ANNOTATION_NAME_LENGTH = 45;
+    private final static String ANNOTATION_DATE_REGEX = "D:\\d+";
+    public final String author;

I know that this is not a problem of mutability here, but is there a good reason for breaking with the Java conventions for this and the following attributes? If there is no added benefit, then I would much prefer to have public getters instead of public attributes for the simple reason that this is the convention. Otherwise it always looks fishy when such a variable is accessed in another class and I have to look up FileAnnotation to check that this is not a grave access bug.

lenhard:
In this particular case, the point of the conventions is mainly to ease the understanding of the code and, thereby, consume less developer time. From my point of view, this is worthwhile and I would opt for adding getters.

However, maybe someone from @JabRef/developers disagrees with this? In this case we should discuss our preferred solution and reach a consensus in the devcall that is to be followed for all of JabRef.

Siedlerchr:
I generally vote for getters and setters
Because of information hiding principle
Encapsulation, makes it easier to read and allows for future modifications

Am 07.03.2017 10:47 vorm. schrieb "Jörg Lenhard" <notifications@github.com>:

src/main/java/org/jabref/model/pdf/FileAnnotation.java
```diff
@@ -1,127 +1,116 @@
...
 
-    public FileAnnotation(final PDAnnotation annotation, final int page ){
+    /**
+     * Creating a normal FileAnnotation from a PDAnnotation.
+     *
+     * @param annotation The actual annotation that holds the information
+     * @param pageNumber The page of the pdf where the annotation occurs
+     */
+    public FileAnnotation(final PDAnnotation annotation, final int pageNumber) {
         COSDictionary dict = annotation.getDictionary();
```
tobiasdiez: Can you please reuse the "dummy" construction, i.e. `this(annotation.getModifiedDate(), pageNumber, ...)`. This comment also applies to the second constructor below.

src/main/java/org/jabref/model/pdf/FileAnnotation.java
```diff
@@ -1,127 +1,116 @@
...
+        this.linkedFileAnnotation = Optional.of(linkedFileAnnotation);
     }
 
-    public String getAuthor() {
-        return author;
-    }
+    private String prettyPrint(String date) {
+        // Sometimes this is null, not sure why.
+        if (date == null) {
+            return date;
```
tobiasdiez: Then better set date here to the empty string (otherwise the null-check has to propagate to every usage of  the field `FileAnnotation.date`

src/main/java/org/jabref/model/pdf/FileAnnotation.java
```diff
@@ -1,127 +1,116 @@
...
+        }
 
-    public void setAuthor(String author) {
-        this.author = author;
-    }
+        // normal case for an imported annotation
+        if (date.matches(ANNOTATION_DATE_REGEX)) {
```
tobiasdiez: Maybe even store the date as an instance of  `Date` ? Then the gui can decide how to display it.

src/main/java/org/jabref/logic/pdf/PdfAnnotationImporter.java
```diff
@@ -39,115 +44,123 @@
...
+        if (!path.toString().toLowerCase(Locale.ROOT).endsWith(".pdf")) {
+            LOGGER.warn(String.format("File %s does not end with .pdf!", path.toString()));
+            return Optional.empty();
+        }
 
-            if(path.toLowerCase(Locale.ROOT).endsWith(".pdf")){
-               return PDDocument.load("/"+ path);
+        if (!Files.exists(path)) {
+            Optional<File> importedFile = FileUtil.expandFilename(context, path.toString(),
+                    JabRefPreferences.getInstance().getFileDirectoryPreferences());
```
tobiasdiez: can you please pass the `fileDirectoryPreferences` as an argument to the constructor. Everything in logic and model should be independent of the `JabRefPreferences` class (we should add a architecture test for that too...)

stefan-kolb:
I don't see a need for Getters if the Objects are immutable, just clutters the code.

matthiasgeiger:
Well... I think adding the `devcall` label was a good idea of Jörg ;-)

koppor:
+1 for devcall. 🌞 

Until then: https://projectlombok.org/features/GetterSetter.html

src/main/java/org/jabref/model/pdf/FileAnnotation.java
lynyus: fixed in bd542d6

src/main/java/org/jabref/model/pdf/FileAnnotation.java
lynyus: fixed in f977614

src/main/java/org/jabref/model/pdf/FileAnnotation.java
lynyus: done in bd542d6

src/main/java/org/jabref/logic/pdf/PdfAnnotationImporter.java
lynyus: hmm I'm not sure how to do that. On the calling side of PdfAnnotationImporter (EntryAnnotationImporter) I only have the BibDatabaseContext available. Not sure how to obtain the correct path from there?

lynyus:
Thanks for the feedback, esp. @tobiasdiez. I fixed these issues and made the FileAnnotationTab better overall by removing invalid (i.e. unparsable annotations).

src/main/java/org/jabref/logic/pdf/PdfAnnotationImporter.java
```diff
@@ -39,115 +44,123 @@
...
+        highlightAnnotations.add(new FileAnnotation(annotation, pageIndex + 1, annotationBelongingToHighlighting));
+        return highlightAnnotations;
+    }
+
+    private String extractHighlightedText(PDPage page, PDAnnotation annotation) throws IOException {
+        //highlighted text has to be extracted by the rectangle calculated from the highlighting
+        PDFTextStripperByArea stripperByArea = new PDFTextStripperByArea();
+        COSArray quadsArray = (COSArray) annotation.getDictionary().getDictionaryObject(COSName.getPDFName("QuadPoints"));
+        String highlightedText = "";
+        for (int j = 1,
```
Siedlerchr: I don't know if you changed it, but it would be nice to have a bit better speaking variable names, as I here don't get from a quick view what j and k stand for.

lynyus:
You are absolutely right. But to be honest in this particular case I just regard this method as a black box and hope nobody has to touch it again.

matthiasgeiger:
We decided that we want to stick with the good old Getters.

src/main/java/org/jabref/logic/pdf/PdfAnnotationImporter.java
tobiasdiez: Yes this change would propagate to EntryAnnotationImporter and FileAnnotationCache (and probably further places) until one ends up in gui. It's not that important, you can also leave it as it is right now.

Issue #2572, opened 2017-02-20
Cannot change file folder of download fulltext file


stefan-kolb:
![image](https://cloud.githubusercontent.com/assets/2141507/23121265/0c23f7e8-f760-11e6-855c-2e51361e7bf5.png)

Not possible to select different folder.

Siedlerchr:
Do you get an error when clicking opening like "no valid folder"?
Maybe similar to #2527

stefan-kolb:
No, I just cannot select a folder but only a file.

stefan-kolb:
Actually, also the add file link and edit file link behave nearly the same and have wrong titles but should be dependent on the use case:
- add: select a file (not title=save file)
- fulltext download (change directory only and append file name)
- edit entry: select a file (not save file)

PR #2625, opened 2017-03-07
Change dialog of download fulltext to save dialog

Siedlerchr:
Fix for #2572 
When running Get fulltext, the dialog is now a save dialog which enables to select a different folder or even a different filename. The initial filename which JabRef assigns is now also visible in the file dialog

~- [ ] Tests created for changes~
~- [X] Screenshots added (for bigger UI changes)
~- [ ] Check documentation status (Issue created for outdated help page at [help.jabref.org]~(https://github.com/JabRef/help.jabref.org/issues)?)
~- [ ] If you changed the localization: Did you run `gradle localizationUpdate`?~

savefile

grafik

### DIFF
--- a/src/main/java/org/jabref/gui/FileDialog.java
+++ b/src/main/java/org/jabref/gui/FileDialog.java
@@ -35,6 +35,7 @@ import org.apache.commons.logging.LogFactory;
  */
 @Deprecated
 public class FileDialog {
+
     private static final Log LOGGER = LogFactory.getLog(FileDialog.class);
 
     private final FileChooser fileChooser;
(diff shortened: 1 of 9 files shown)


src/main/java/org/jabref/gui/FileDialog.java
```diff
@@ -114,6 +115,14 @@ public void setDefaultExtension(FileExtensions extension) {
     }
 
     /**
+     * Sets the initial file name, useful for saving dialogs
+     * @param fileName
+     */
+    public void setInitialFileName(String fileName) {
```
tobiasdiez: Instead of extending a deprecated class, can the FileDialog below be replaced by the new DialogService?

Siedlerchr: done. I extended the new DialogService class now

stefan-kolb:
LGTM 👍

PR #2626, opened 2017-03-08
F4 opens selected file in current JTable context

### LINKED ISSUE #2355: F4 should open currently selected file
<kbd>F4</kbd> opens the first attached file. When in the General tab another file is selected, this should be opened.

not-functioning

This is https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=445301

stefan-kolb:
Resolves #2355

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - The `Move linked files to default file directory`-Cleanup operation respects the `File directory pattern` setting 
 - We separated the `Move file` and `Rename Pdfs` logic and context menu entries in the `General`-Tab for the Field `file` to improve the semantics
 - A scrollbar was added to the cleanup panel, as a result of issue [#2501](https://github.com/JabRef/jabref/issues/2501)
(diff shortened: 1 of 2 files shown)


lenhard:
I just tested this locally and it works if I have selected another file in the entry editor.

However, the following seems a little counter-intuitive to me:
 1. I have the second file of an entry selected in the entry editor, everything works as expected.
 2. Next, I click the same entry in the main table. Both the entry in the main table and the second file in the entry editor are still highlighted.
 3. When I push F4 now, the first attached file is opened and not the one that is still highlighted in the entry editor. This seems counter-intuitive.

This is just a minor thing, and if in doubt the PR can be merged as is (code-wise it is fine). However, it would be nice if the feature were more consistent.

stefan-kolb:
I understand this, but this is a focus problem then. I'm not sure how to fix this. Initially, I also didn't want to implement this feature here, but Olly wanted JabRef to behave like this. For me F4 opening the first link regardless of the current focus was ok.

lenhard:
@stefan-kolb Sure, I understand. This is no blocker and from my point of view, the PR can also be merged as is. I just guess that at some point a user will stumble upon this and report it as an issue.

PR #2628, opened 2017-03-09
Filenames

stefan-kolb:


src/main/java/org/jabref/logic/util/io/FileUtil.java
```diff
@@ -84,6 +82,28 @@ public static String getFileName(String fileNameWithExtension) {
...
 
     /**
+     * Returns a valid filename for most operating systems.
+     *
+     * Currently, only the length is restricted to 255 chars.
+     */
+    public static String getValidFileName(String fileName) {
+        String nameWithoutExtension = getFileName(fileName);
+
+        if (nameWithoutExtension.length() > 255) {
```
lynyus: I would suggest to extract this magic number into `public static final int MAX_FILENAME_LENGTH`.  This can then be reused by the test as well.

stefan-kolb:
THX for the Review,but it is Not ready Yet ;)

Am 17.03.2017 6:07 nachm. schrieb "Linus Dietz" <notifications@github.com>:

lynyus:
I guess this is ready now to be merged. I don't recall any other problems (we talked about that in the devcall) with the linked files, since @Andrew-Lev already takes care of https://github.com/JabRef/jabref/issues/3172 in  https://github.com/JabRef/jabref/pull/3188

src/main/java/org/jabref/logic/cleanup/RenamePdfCleanup.java
```diff
@@ -125,22 +122,12 @@ public RenamePdfCleanup(boolean onlyRelativePaths, BibDatabaseContext databaseCo
...
-                    Path parent = settingsDir.get();
-                    String newFileEntryFileName;
-                    if (parent == null) {
-                        newFileEntryFileName = targetFileName;
-                    } else {
-                        newFileEntryFileName = parent.relativize(newPath).toString();
-                    }
-                    newFileList.add(new LinkedFile(description, newFileEntryFileName, type));
-                }
+                settingsDir.ifPresent(path -> newFileList.add(new LinkedFile(description, path.relativize(newPath).toString(), type)));
```
koppor: shouldn't be here an `orElse` to ensure that  in `newFileList` the entry is always added, even if there is no `getFirstExistingFileDir`?

Maybe, we can be sure that this always returns something in this state. If we are, we have to add

    if (!settingsdir.isPresent) LOGGER.debug("no directory determined. This is strange.");

Siedlerchr: an orElse would be the wrong approach here because it would then resolve to the JabRef install directory.

src/main/java/org/jabref/logic/cleanup/RenamePdfCleanup.java
```diff
@@ -150,28 +137,24 @@ public RenamePdfCleanup(boolean onlyRelativePaths, BibDatabaseContext databaseCo
...
     public String getTargetFileName(LinkedFile flEntry, BibEntry entry) {
         String realOldFilename = flEntry.getLink();
 
-        StringBuilder targetFileName = new StringBuilder(FileUtil
-                .createFileNameFromPattern(databaseContext.getDatabase(), entry, fileNamePattern, layoutPrefs)
-                .trim());
-        //Add extension to newFilename
-        targetFileName.append('.').append(FileHelper.getFileExtension(realOldFilename).orElse("pdf"));
-        return targetFileName.toString();
+        String targetFileName = FileUtil.createFileNameFromPattern(
```
Siedlerchr: Why do you drop the StringBuilder here?

src/main/java/org/jabref/logic/cleanup/RenamePdfCleanup.java
Siedlerchr: In fact I think the optional could be dropped because the bib file dir should always be a valid directory

src/main/java/org/jabref/logic/cleanup/RenamePdfCleanup.java
koppor: Just normal String concatenation. I thin, the StringBuilder did not introduce more readability. In the current code, I find it more readable. But, maybe, I am a 2005 programmer 🎉

Siedlerchr: As I just read the String concat gets nonetheless optimized as a StringBuilder on the compiler side, so I guess it's safe here. 

At least when doing concats in loops you should stick to StringBuilder because its more efficient

lynyus: I don't see the point of using a StringBuilder there either, it feels like it clutters the code. Used String.format() as an alternative before, but I think this is the most readable variant.

src/main/java/org/jabref/logic/cleanup/RenamePdfCleanup.java
```diff
@@ -121,13 +121,14 @@ public RenamePdfCleanup(boolean onlyRelativePaths, BibDatabaseContext databaseCo
...
+                // Change the path for this entry
                 String description = oldLinkedFile.getDescription();
                 String type = oldLinkedFile.getFileType();
 
-                //We use the file directory (if none is set - then bib file) to create relative file links
-                Optional<Path> settingsDir = databaseContext.getFirstExistingFileDir(fileDirectoryPreferences);
-                settingsDir.ifPresent(path -> newFileList.add(new LinkedFile(description, path.relativize(newPath).toString(), type)));
+                // We use the file directory (if none is set - then bib file) to create relative file links.
+                // The .get() is legal without check because the method will always return a value.
+                Path settingsDir = databaseContext.getFirstExistingFileDir(fileDirectoryPreferences).get();
```
Siedlerchr: Could you add a test case for that?
Just to make sure we don't miss anything

lynyus: What do you mean exactly? The tests already cover this to some extend. https://codecov.io/gh/JabRef/jabref/compare/c416818deba486605779a375a8f09f605c8dec9c...e4604444bdf8f1ed93678736183248d25924ae93/src/src/main/java/org/jabref/logic/cleanup/RenamePdfCleanup.java

Siedlerchr: Sorry, did not see that! I merge it in now

Issue #2599, opened 2017-03-02
A few issues with the new groups panel in version 4


AEgit:
JabRef 4.0.0-dev--snapshot--2017-03-01--braces-checking-followup--488825af5

This ticket is supposed to mention a few problems with the implementation of the groups panel in the current development version 4 of JabRef.

I realized that saving a huge database (>10,000 entries) takes now much longer than with version 3.8.2 (database was saved nearly instantaneously), especially when the entry preview is open. Indeed, the usage of the CPU suddenly spikes to nearly 100% and it takes several seconds for JabRef to save the database. In the meantime it behaves as if JabRef had crashed. This might be related to some issues of the groups feature which I report below. Performance issues with version 4 have also been reported elsewhere: https://github.com/JabRef/jabref/pull/2587 and https://github.com/JabRef/jabref/pull/2588

Generally the implementation of the groups feature in JabRef 4 is:

- [ ] slower than the old one (especially when you try to save changes in your database while the entry preview is open - see above). Furthermore, the first time you try to expand "All entries" it takes a couple of seconds before the expanded groups are shown.
- [x] does not allow to hide the number of entries (which might cause some performance issues on large databases): hit number is calculated async, thus there shouldn't be a performance problem
- [x] cannot be opened with all subgroups expanded as default: fixed (expansion status will be kept across sessions)
- [x] displays group titles with underscore, e.g. "Alpha_taxonomy" as "Alpha_(t)axonomy", i. e. brackets are added to the first character following the underscore: consequence of a known bug https://github.com/JabRef/jabref/issues/2664
- [x] does not offer the same options that were available with the old group panel, when rightclicking onto one of the groups. E.g., the menu points 
  - [x] "Edit group": https://github.com/JabRef/jabref/pull/2674
  - [x] "Add group", 
  - [x] "Remove groups and subgroups"
  - [x] "Remove group keep subgroups": https://github.com/JabRef/jabref/pull/2674
  - [x]  "Remove subgroups": https://github.com/JabRef/jabref/pull/2674
  - [x]  "Expand subtree": removed in https://github.com/JabRef/jabref/pull/2697
  - [x]  "Collapse subtree": removed in https://github.com/JabRef/jabref/pull/2697
  - [x]  "Move: Up, Down, Left, Right": removed in https://github.com/JabRef/jabref/pull/2697
  - [x]  "Sort alphabetically"
  - [x]  "Assign entry selection to this group": https://github.com/JabRef/jabref/pull/2674
[...]

PR #2632, opened 2017-03-11
Add drag and drop for groups -> groups

tobiasdiez:
Enable drag and drop in the group panel.
I also had to remove the "worker" aspect of the group worker, ie. now the entry list is updated immediately (this might lead to performance problems in bigger db's; but without this change JabRef would just hang after drop and down).


Siedlerchr:
I will add my changes from the other PR to this one

matthiasgeiger:
@Siedlerchr Really? Isn't the scope of this PR different? 
You are working on "assign entries to groups by drag-and-drop" whereas this PR deals with "restructure the groups tree with drag-and-drop".

However, it yours and @tobiasdiez's decision whether this makes sense from the technical aspect.

Siedlerchr:
In fact it uses the same methods for listening/reaction of an DND Event. My changes would be simply an else case.

matthiasgeiger:
Allright, then forget my objections :wink:

Siedlerchr:
@tobiasdiez  Drag and drop of entries works now as expected 🥇

src/main/java/org/jabref/gui/groups/GroupTreeViewModel.java
```diff
@@ -67,7 +67,8 @@ public StringProperty filterTextProperty() {
      * We need to notify the {@link StateManager} about this change so that the main table gets updated.
      */
     private void onSelectedGroupChanged(GroupNodeViewModel newValue) {
-        if (currentDatabase != stateManager.activeDatabaseProperty().get()) {
+        //TODO: Can one of this be empty?
+        if (currentDatabase.get() != stateManager.activeDatabaseProperty().get().get()) {
```
Siedlerchr: @tobiasdiez  I am not sure if we should compare the values or if it is enough to check the object refs

tobiasdiez: I think `equals` is the cleanest solution

tobiasdiez:
Thanks @Siedlerchr for making the groups great again!

src/main/java/org/jabref/gui/groups/GroupTreeController.java
```diff
@@ -136,9 +136,54 @@ public void initialize() {
...
+                    if (source.isPresent()) {
+                        source.get().moveTo(row.getItem());
+                        success = true;
+                    }
+                }
+                if (dragboard.hasContent(DragAndDropDataFormats.ENTRIES)) {
+                    TransferableEntrySelection entrySelection = (TransferableEntrySelection) dragboard
+                            .getContent(DragAndDropDataFormats.ENTRIES);
+
+                    row.getItem().addEntriesToGroup(entrySelection.getSelection());
```
Siedlerchr: Forgot to set succes true in the entries else part

Siedlerchr: Pushed to master

matthiasgeiger:
Short question here: Drag-and-drop generally works in current `master` - but the main table is not updated at once. I.e., after dropping an entry to the currently selected group it is still grayed out as an "non-member". After reselecting the group it is displayed correctly in the "floating" list of group members.

Is this a known problem? Or should I open an issue?

tobiasdiez:
I added it to the list of open group-related problems: https://github.com/JabRef/jabref/issues/2599

Issue #210, opened 2015-10-05
Import window is displayed in background


Ede123:
I'm using JabRef 2.11b4 on Windows 7 x64 with Java 8 Update 60.

I have set Firefox to open .bib files with JabRef (JabRef.exe in JabRef directory to be specific).

If JabRef is already opened in background an import window is shown for the new entry.
The problem is that this import window is not brought to the foreground (i.e. the window is not focused). Therefore the import window is always hidden behind the browser window. Since the import window is not shown in task bar either, it seems as if the import had failed (unless one "digs" for the import window with Alt+Tab).

stefan-kolb:
@Ede123 Can you reproduce this issue with the current master branch?

Ede123:
I have  to check. Are there somewhere pre-built versions available or do I have to build from source?

koppor:
The build artifacts are only available via deep link: https://circle-artifacts.com/gh/JabRef/jabref/674/artifacts/0/home/ubuntu/jabref/build/releases/JabRef-2.80dev--snapshot--2015-10-09--8fe4c28.jar and https://circle-artifacts.com/gh/JabRef/jabref/674/artifacts/0/home/ubuntu/jabref/build/releases/JabRefSetup-2.80dev--snapshot--2015-10-09--8fe4c28.exe

lenhard:
Closing, since @Ede123 seems to have lost interest.

Feel free to reopen if the problem reappears.

PR #2639, opened 2017-03-12
Add option to copy title of bibentries

ohuang12:
Implements issue [#210](https://github.com/koppor/jabref/issues/210) from [@koppor](https://github.com/koppor).

Adds an option to copy the title of a bibtex entry when right clicking one (or more) to the clipboard.
Also adds the option to copy the title using `Edit -> Copy title`.

edit_menu
right_click_menu


src/main/java/org/jabref/gui/BasePanel.java
```diff
@@ -808,6 +811,33 @@ private void paste() {
...
 
+    private void copyTitle() {
+        List<BibEntry> bes = mainTable.getSelectedEntries();
+        if (!bes.isEmpty()) {
+            storeCurrentEdit();
+            List<String> titles = new ArrayList<>(bes.size());
+
+            // Collect all non-null keys.
+            for (BibEntry be : bes) {
+                be.getTitle().ifPresent(titles::add);
```
Siedlerchr: I think you could use a Stream together with filter(title::isPresent).collect(Collectors.toList)

src/main/java/org/jabref/model/entry/BibEntry.java
```diff
@@ -622,6 +622,15 @@ public String getAuthorTitleYear(int maxCharacters) {
     }
 
     /**
+     * Returns the title of the given BibTeX entry as an optional.
+     *
+     * @return the title of a BibTeX entry or null if no title was found.
+     */
+    public Optional<String> getTitle() {
+        return getField(FieldName.TITLE);
```
lynyus: Not sure what @JabRef/developers think about returning an optional there, but I think instead the error handling (absence of title) should be done already in this getter.
Thus I would vote for returning an empty string of the optional is absent.

ohuang12: I've based the getter for the title on the other present getters in the `BibEntry` class. 
All these getters also return an empty `Optional` in case one isn't found and leave the error handling to be performed by the methods using it.

I can change it to return an empty string, if that seems to be the better option.

src/main/java/org/jabref/gui/BasePanel.java
ohuang12: I've tried using a stream using your suggest now for retrieving the list of titles.
However, it required an additional mapping to be done in order to form the list of titles.

tobiasdiez: It also works with flatMap and a helper in OptionalUtil (not sure how the method is called that transforms an optional into a stream)

src/main/java/org/jabref/model/entry/BibEntry.java
tobiasdiez: I think `Optional<String>` is indeed the more consistent way to go. Also has the advantage that the methods of the optional can be used (map, ifPresent etc).

Siedlerchr: Optional is fine here. It follows the same pattern for each field.

matthiasgeiger: 👍 for `Optional<>` - but the javadoc should be updated accordingly: `@return the title of a BibTeX entry or null if no title was found.`

I think this is introduced by copy&paste but we should not introduce new issues with new code. 
Another, rather nitpicky aspect: In another snippet you created variables like `bes` - don't cheat by looking - but what exactly is stored in this variable? :wink: 

Using the name "selectedBibEntries" would be way more self-explanatory...

matthiasgeiger: Okay after some discussion with @lynyus and after checking the code again, I would recommend to remove this getter completely as it does not provide any benefit. Just directly use `getField(FieldName.TITLE)` in the code.

src/main/java/org/jabref/gui/BasePanel.java
ohuang12: I couldn't find the helper to transform an optional into a stream in the [Optional](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html) documentation. 
I believe this is something that was added later in JDK 9.
I looked into [other ways](http://stackoverflow.com/questions/22725537/using-java-8s-optional-with-streamflatmap) to use flatMap, but these did not look very nice to me.

Do you think I should change it any way or just leave it like it is now (with the filter + map)?

src/main/java/org/jabref/model/entry/BibEntry.java
ohuang12: Yes, that was careless of me 😄.
I have updated the java doc for `getTitle` to reflect what it actually does.

I was aware that `bes` contained all the selected `BibEntry` objects, but left the naming like that to be consistent with the other naming. 
I have changed the variable names to be more descriptive.

src/main/java/org/jabref/gui/BasePanel.java
tobiasdiez: Leave it like this, my comment was just nitpicking. I was referring to our [toStream method](https://github.com/JabRef/jabref/blob/master/src/main/java/org/jabref/model/util/OptionalUtil.java#L24).

Siedlerchr:
So as we all agreed on this and the code is looking good, I will merge this in now.
Thank you very much for your contribution!

PR #2640, opened 2017-03-12
[WIP] Put the PDFAnnotationImporter under Test, enhance FileAnnotationTab

lynyus:
I have created tests for the PDF importer. However, the coverage has yet to be improved and I'm planning to extend the FileAnnotationTab in this pull request. So this is the follow-up for #2624, but this time in TDD.

- [x] Support Underline Annotations
- [x] Support Shapes with Comments
- [x] Fix correct highlighting/shading of the TextAreas in the FileAnnotationTab
- [x] Fix reload annotations Button


src/main/java/org/jabref/logic/pdf/FileAnnotationCache.java
```diff
@@ -35,7 +38,13 @@ public FileAnnotationCache(BibDatabaseContext context) {
      * @param entry entry for which to get the annotations
      * @return Map containing a list of annotations in a list for each file
      */
-    public Map<String, List<FileAnnotation>> getFromCache(BibEntry entry) throws ExecutionException {
-        return annotationCache.get(entry);
+    public Map<String, List<FileAnnotation>> getFromCache(BibEntry entry) {
+        LOGGER.info(String.format("Loading Bibentry '%s' from cache.", entry.getField(BibEntry.KEY_FIELD).get()));
```
tobiasdiez: maybe change to Logger.debug?

src/test/java/org/jabref/logic/pdf/PdfAnnotationImporterTest.java
```diff
@@ -0,0 +1,130 @@
...
+
+public class PdfAnnotationImporterTest {
+
+    private AnnotationImporter importer = new PdfAnnotationImporter();
+
+    @Test
+    public void noAnnotationsWriteProtected() {
+
+        List<FileAnnotation> annotations = importer.importAnnotations(Paths.get("src/test/resources/pdfs/write-protected.pdf"));
+        assertEquals(0, annotations.size());
```
tobiasdiez: it is preferable to check versus an empty list: `assertEquals(Collections.emptyList(), annotations)`. In the case, when the test fails you get a more helpful error message.

src/test/java/org/jabref/logic/pdf/PdfAnnotationImporterTest.java
```diff
@@ -0,0 +1,130 @@
...
+    @Test
+    public void noAnnotationsMinimal() {
+        List<FileAnnotation> annotations = importer.importAnnotations(Paths.get("src/test/resources/pdfs/minimal.pdf"));
+        assertEquals(0, annotations.size());
+    }
+
+    @Test
+    public void inlineNoteMinimal() {
+        List<FileAnnotation> annotations = importer.importAnnotations(Paths.get("src/test/resources/pdfs/minimal-inlinenote.pdf"));
+        assertEquals(1, annotations.size());
```
tobiasdiez: As above, check versus an expected list:
`FileAnnotation expected = new .....` and `assertEquals(Collections.singletonList(expected), annotations))`

src/main/java/org/jabref/gui/entryeditor/FileAnnotationTab.java
```diff
@@ -267,52 +287,91 @@ private JPanel setUpButtons() {
...
-            highlightTxtArea.setText("N/A");
-            highlightTxtArea.setEnabled(false);
+            contentTxtArea.setText(delocalize(annotation.getContent()));
+            contentTxtArea.setEnabled(true);
+            markedTxtArea.setText("N/A");
+            markedTxtArea.setEnabled(false);
+        }
+    }
+
+    private String delocalize(String content) {
```
tobiasdiez: Why does the content is first localized and then delocalized again?! I just don't understand this part of the code.

src/test/java/org/jabref/logic/pdf/PdfAnnotationImporterTest.java
lynyus: This is not so useful, as a FileAnnotation is a quite complex class (especially if it has a linkedAnnotation) and some fields are just not of interest for the tests. I would suggest to leave it as it is.

src/test/java/org/jabref/logic/pdf/PdfAnnotationImporterTest.java
lynyus: fixed in c4aed22921514b9998976d5057d74b5fc4e7a9ab

src/main/java/org/jabref/gui/entryeditor/FileAnnotationTab.java
lynyus: Well, the localization must be done where the class is used instead of the model.
As I introduced more annotation types like underline annotations or polygons, I ran out of symbols for the icon in the annotationList, thus, I have prepended the type of the marking to the title of the marking in the list, so that the user is aware what type of annotation it is.
However, in the actual content field this should not be there any more, as it is shown in the label in front of the TextArea.

src/main/java/org/jabref/logic/pdf/FileAnnotationCache.java
lynyus: fixed in 2894c9ec561cb3476a646fc13b0b55c5625a0c5a

src/main/java/org/jabref/logic/pdf/EntryAnnotationImporter.java
```diff
@@ -43,15 +47,19 @@ public EntryAnnotationImporter(BibEntry entry) {
...
         AnnotationImporter importer = new PdfAnnotationImporter();
+
         //import annotationsOfFiles if the selected files are valid which is checked in getFilteredFileList()
-        this.getFilteredFileList().forEach(parsedFileField -> annotations.put(parsedFileField.getLink(),
-                importer.importAnnotations(Paths.get(parsedFileField.getLink()), context)));
+        for (ParsedFileField parsedFileField : this.getFilteredFileList()) {
+            Optional<File> expandedFileName = FileUtil.expandFilename(databaseContext, parsedFileField.getLink(),
+                    JabRefPreferences.getInstance().getFileDirectoryPreferences());
+            expandedFileName.ifPresent(file -> annotations.put(file.toString(), importer.importAnnotations(Paths.get(file.toString()))));
+        }
```
Siedlerchr: Use file.toPath() directly

src/main/java/org/jabref/logic/pdf/PdfAnnotationImporter.java
```diff
@@ -43,24 +38,22 @@
...
         List<FileAnnotation> annotationsList = new LinkedList<>();
-        try (PDDocument document = PDDocument.load(validPath.toString())) {
+        try (PDDocument document = PDDocument.load(path.toString())) {
             List pdfPages = document.getDocumentCatalog().getAllPages();
             for (int pageIndex = 0; pageIndex < pdfPages.size(); pageIndex++) {
                 PDPage page = (PDPage) pdfPages.get(pageIndex);
                 for (PDAnnotation annotation : page.getAnnotations()) {
-                    if (annotation.getSubtype().equals(FDFAnnotationHighlight.SUBTYPE)) {
-                        annotationsList.add(createHighlightAnnotations(pageIndex, page, annotation));
+                    if (annotation.getSubtype().equals(FDFAnnotationHighlight.SUBTYPE) ||
```
Siedlerchr: You should change the order of equals, the constant first and then the annotation.getsubType as it theoretically could ne null and would result in an NPE.

src/main/java/org/jabref/gui/entryeditor/FileAnnotationTab.java
Siedlerchr: Sorry, that sounds really complex.
Why don't you just pass the name of the type to the GUI and then add the translation there?

src/main/java/org/jabref/gui/entryeditor/FileAnnotationTab.java
```diff
@@ -358,6 +417,9 @@ public Component getListCellRendererComponent(JList<?> list, Object value, int i
                 case FDFAnnotationHighlight.SUBTYPE:
                     label.setIcon(IconTheme.JabRefIcon.MARKER.getSmallIcon());
                     break;
+                case FDFAnnotationUnderline.SUBTYPE:
```
Siedlerchr: Is/Can FileAnnotation be made an enum?

src/test/java/org/jabref/logic/pdf/PdfAnnotationImporterTest.java
Siedlerchr: The junit message would show the objects values for exepcted vs actual as it is done in the fetchers or importers for example.
So you can directly compare them and see any differences, here you would get only Expected 1 actual 0 in case a test would fail -> Not helpful

src/main/java/org/jabref/logic/pdf/PdfAnnotationImporter.java
```diff
@@ -70,31 +63,34 @@
...
-        FileAnnotation annotationBelongingToHighlighting = new FileAnnotation(
+    private FileAnnotation createMarkedAnnotations(int pageIndex, PDPage page, PDAnnotation annotation) {
+        FileAnnotation annotationBelongingToMarking = new FileAnnotation(
                 annotation.getDictionary().getString(COSName.T), FileAnnotation.extractModifiedTime(annotation.getModifiedDate()),
-                pageIndex + 1, annotation.getContents(), FDFAnnotationText.SUBTYPE, Optional.empty());
+                pageIndex + 1, annotation.getContents(), annotation.getSubtype(), Optional.empty());
 
         try {
-            annotation.setContents(extractHighlightedText(page, annotation));
+            if (annotation.getSubtype().equals(FDFAnnotationHighlight.SUBTYPE) || annotation.getSubtype().equals(FDFAnnotationUnderline.SUBTYPE)) {
```
Siedlerchr: same as above

src/test/java/org/jabref/logic/pdf/PdfAnnotationImporterTest.java
lynyus: Hm. After @tobiasdiez comment I have tried it already but the message is: (see 8ed6abf1a9a3a486f7d62b84e153b1a14130c18d)
```
java.lang.AssertionError: Comparison fails. WHY? expected: java.util.Collections$SingletonList<[inline note annotation]> but was: java.util.LinkedList<[inline note annotation]>
Expected :java.util.Collections$SingletonList<[inline note annotation]> 
Actual   :java.util.LinkedList<[inline note annotation]>
```
First of all I don't get why the comparison fails.
Second the message is really not helpful.

src/main/java/org/jabref/logic/pdf/EntryAnnotationImporter.java
lynyus: fixed in 1dfa504dd206b3b65463eee4cfccf776a7a566f4

src/main/java/org/jabref/gui/entryeditor/FileAnnotationTab.java
lynyus: Because the `DefaultListModel<FileAnnotation> listModel;` uses the `FileAnnotation.toString()` method directly to render the `JList<FileAnnotation> annotationList;`. There we need the localized name of the `AnnotationType`, whereas in the `private final JTextArea markedTxtArea;` this would be a duplication, so I stripped it.

I'm not proficient in GUI programming, and this was the easiest solution I found.

src/main/java/org/jabref/logic/pdf/PdfAnnotationImporter.java
lynyus: yes. bcf75168294e25f0769cf8f90ff847321032ee95

src/main/java/org/jabref/logic/pdf/PdfAnnotationImporter.java
lynyus: yes. bcf75168294e25f0769cf8f90ff847321032ee95

src/main/java/org/jabref/gui/entryeditor/FileAnnotationTab.java
Siedlerchr: OKay, then It's okay

src/main/java/org/jabref/gui/entryeditor/FileAnnotationTab.java
Siedlerchr: Ah I just checked it's strings..

src/test/java/org/jabref/logic/pdf/PdfAnnotationImporterTest.java
```diff
@@ -45,14 +47,21 @@ public void noAnnotationsMinimal() {
...
+        assertEquals(1, note.getPage());
         assertEquals("inline note annotation", note.getContent());
-        assertFalse(note.hasLinkedAnnotation());
         assertEquals("FreeText", note.getAnnotationType());
-        assertEquals(1, note.getPage());
+        assertFalse(note.hasLinkedAnnotation());
+        // End old test
+
+        FileAnnotation expected = new FileAnnotation("Linus Dietz", LocalDateTime.of(2017, 3, 12, 20, 25), 1, "inline note annotation", "FreeText", Optional.empty());
+        assertEquals("Comparison fails. WHY?", Collections.singletonList(expected), annotations);
```
Siedlerchr: Not directly to see, but I did some debuggin and it  fails because the FileAnnotation does not provide an equals method

Siedlerchr: I have added an equals method now

lynyus: Ah. That makes sense. Thanks a lot!

src/test/java/org/jabref/logic/pdf/PdfAnnotationImporterTest.java
```diff
@@ -49,38 +49,34 @@ public void noAnnotationsMinimal() {
...
+        assertEquals(LocalDateTime.of(2017, 3, 12, 20, 17, 24), note.getTimeModified());
+        assertEquals(1, note.getPage());
+
         assertEquals("A simple pop-up note", note.getContent());
-        assertFalse(note.hasLinkedAnnotation());
         assertEquals("Text", note.getAnnotationType());
-        assertEquals(1, note.getPage());
+        // End old test style
+
+        assertEquals(expected,
```
lynyus: Thanks to @Siedlerchr for pointing out the necessity to override the `.equals()` method in `FileAnnotation`, however the reason for this was to actually improve the message which is displayed by jUnit if the test fails.
@tobiasdiez, can you give a hint why the message is better with the Collections.singletonList() variant? (see https://github.com/JabRef/jabref/pull/2640#discussion_r106727820)

src/main/java/org/jabref/gui/entryeditor/FileAnnotationTab.java
tobiasdiez: I think it the classical problem, that `model` objects are directly used in the gui to display information (and thus the model class gets modified in ways that are undesirable from a model view point, but needed in the gui). Here I would propose that:
- `FileAnnoation` stores and returns the content as extracted from the PDF. It also provides way to determine its type (preferable as an enum)
- You create a wrapper `FileAnnotationViewModel` that prepares the data to be displayed in the GUI (e.g. there is a method that converts the annotation type to a localized string, overwrites toString such that the content is displayed etc).

src/test/java/org/jabref/logic/pdf/PdfAnnotationImporterTest.java
tobiasdiez: To see the value of comparison with lists, we need to assume that the method under test returns something unexpected. For example:
- It returns two items instead of one. With the "old style" you just get `expected 1 but got 2` while comparison with the singleton list displays something like `Expected "item [property 1, ...]" but got "item [property 1, ....], item2 [value 1, ... ]"`. So you get more details about the item that is wrongly returned, which is often helpful in fixing bugs. 
- The method still returns only one item but with the wrong values. In this case, the old style gives you `expected "value1" but got "wrong value"`. If you compare against an expected item you directly get also information about the other properties (thus you directly see if just one value is off or a completely different item is returned). 

Another important advantage is that, if in the future the class is extended by an additional property, you just need to adapt the equals method and all your tests automatically take this new property in account.

lynyus: Well, if I change any of the tests in the PdfAnnotationImporterTest such that one attribute of the single list element mismatches I don't get the rich comparison you are describing. Nevertheless I have implemented it as you have proposed.

PR #2641, opened 2017-03-13
Improve issue reporting a bit

### LINKED ISSUE #2573: Integrate github plugins for automatic bug report
Issues made with JabRef directly result in very short issues (see #2571). Is it possible to show the issue template (https://github.com/JabRef/jabref/blob/master/.github/ISSUE_TEMPLATE.md) in the JabRef UI to force the user to fill that out?

stefan-kolb:
See #2573.
I also wanted to add the log details to the report without the necessity to paste them later on. Dunno why this was implemented this way initially (See https://github.com/JabRef/jabref/pull/1383). However, the explorer will cut off the arguments at some point so this did not work for me.
I  think it is not convenient if the user needs to add this info manually.
Maybe someone else has an idea how to change this.

I still think the current addition fixes #2573 for now.

image

image

### DIFF
--- a/src/main/java/org/jabref/gui/errorconsole/ErrorConsoleViewModel.java
+++ b/src/main/java/org/jabref/gui/errorconsole/ErrorConsoleViewModel.java
@@ -29,8 +29,8 @@ import org.apache.http.client.utils.URIBuilder;
 import org.apache.logging.log4j.core.LogEvent;
 
 public class ErrorConsoleViewModel extends AbstractViewModel {
-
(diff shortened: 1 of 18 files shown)


stefan-kolb:
@tobiasdiez Actually it is not the restriction of the URL which is quite high (over 2000 i think), but the call from the command line, e.g., explorer.exe "https://www....". Seems like the explorer does not take such a long argument.

Siedlerchr:
From my point of view looks good. Will merge it in now

PR #2644, opened 2017-03-14
Refactor OO connection

### LINKED ISSUE #2484: Jabref 3.8.1 can't connect to open office 4.1.3
JabRef version 3.8.1 on Mac OS X 10.9.5

connecting to OpenOffice 4.1.3 fails with a message that the file /Applications/OpenOffice.app/Contents/Resources/java/unoil.jar cannot be found.

In the OpenOffice application folder the Resources/java folder is missing. The jar file is located in /Applications/OpenOffice.app/Contents/MacOS/classes/

A workaround seems to be in 'repairing' the OpenOffice application on the disk by using Terminal as follows:
_$ cd /Applications/OpenOffice.app/Contents/Resources
$ ln -sf ../MacOS/classes/ java_

After this workaround JabRef connected to the running instance of OO, correctly reporting the file that was open.

As a caveat, I have not tried using this beyond establishing connection.

A better solution would be for JabRef to look for jar files in both the Macos/classes and Resources/java folders.

stefan-kolb:
Also fixes #2484

Trying to simplify OO detection logic.
- Removed cancel option in auto detection logic
- Removed some of the preferences variables for OO directories and try to keep only the root directory and search for the necessary files inside this directory
- Added two file finder helper methods inside FileUtil
- Lots of variable, method, and class renamings
- Used java.nio were possible
- Some graphical UI improvements

Tested on
- [x] Windows with LibreOffice 5 and multiple installations
- [ ] Windows Open Office 4 (will only work with 32-bit JRE!)
```
java.lang.UnsatisfiedLinkError: C:\Program Files (x86)\OpenOffice 4\program\jpipe.dll: Can't load IA 32-bit .dll on a AMD 64-bit platform
```
- [x] Linux LibreOffice 5
`unoil.jar` is missing. However, it would even work without that jar. Not sure whats the problem here and why it is gone. Two other jars `java_uno` and `unoloader`are there. Looks like we can maybe include the uno libraries from Gradle (https://mvnrepository.com/artifact/org.openoffice) rather than loading them from the classpath of the installation?! Maybe a license problem @koppor ?
For Ubuntu you need to install the additional package:
http://packages.ubuntu.com/de/xenial/libreoffice-java-common.
- [x] Linux Open Office
- [x] OSX OpenOffice
- [x] OSX LibreOffice


src/main/java/org/jabref/gui/openoffice/OpenOfficePanel.java
```diff
@@ -460,11 +446,25 @@ private void connect(boolean auto) {
         }
     }
 
+    private void loadOpenOfficeJars(String installationPath) throws IOException {
+        List<String> jarFiles = Arrays.asList("unoil.jar", "jurt.jar", "juh.jar", "ridl.jar");
```
Siedlerchr: You can directly use Stream.of (...) here..

stefan-kolb:
@JabRef/developers @chochreiner I think this has improved quite a bit now. Need your help on a few question (especially LibreOffice 5 Linux Jar poblem) and your testing here!

Siedlerchr:
I can test Windows and ubuntu with libre office 
Regarding the inclusion of jars, aren't the jars version depended on the installed version of Open Office/libre Office?

src/main/java/org/jabref/gui/openoffice/DetectOpenOfficeInstallation.java
```diff
@@ -0,0 +1,183 @@
...
+        progressDialog.dispose();
+    }
+
+    private Optional<Path> selectInstallationPath() {
+        JOptionPane.showMessageDialog(parent,
+                Localization.lang("Unable to autodetect OpenOffice/LibreOffice installation. Please choose the installation directory manually."),
+                Localization.lang("Could not find OpenOffice/LibreOffice installation"),
+                JOptionPane.INFORMATION_MESSAGE);
+        // TODO Dialog service
+        JFileChooser fileChooser = new JFileChooser(new File(System.getenv("ProgramFiles")));
```
Siedlerchr: You Can easily switch to the new dialog service, you just need to call the show function using the executor util function runinjavafxthread

src/main/java/org/jabref/gui/openoffice/OpenOfficePanel.java
```diff
@@ -524,13 +495,13 @@ private void showConnectDialog() {
         ooJars.setText(preferences.getJarsPath());
         JButton browseOOJars = new JButton(Localization.lang("Browse"));
         browseOOJars.addActionListener(e ->
-                new FileDialog(frame).showDialogAndGetSelectedFile()
+                new FileDialog(frame).showDialogAndGetSelectedDirectory()
                         .ifPresent(f -> ooJars.setText(f.toAbsolutePath().toString()))
```
Siedlerchr: Same as above,

src/main/java/org/jabref/logic/openoffice/OpenOfficeFileSearch.java
```diff
@@ -1,121 +1,80 @@
...
 
-        for (String rootPath : sourceList) {
-            File root = new File(rootPath);
-            File[] dirs = root.listFiles(File::isDirectory);
-            if (dirs != null) {
-                for (File dir : dirs) {
-                    if (dir.getPath().contains("OpenOffice") || dir.getPath().contains("LibreOffice")) {
-                        dirList.add(dir);
+        for (Path programDir : programDirectories) {
+            File[] subDirs = programDir.toFile().listFiles(File::isDirectory);
```
Siedlerchr: I think Files.find or Files.walk can be used here as well, because a path can also be a directory

Siedlerchr: Or even better use a DirectoryStream:
http://howtodoinjava.com/best-practices/java-8-list-all-files-example/
https://docs.oracle.com/javase/8/docs/api/java/nio/file/DirectoryStream.html

chochreiner:
open office works fine on macos
libreoffice cannot be detected automatically

stefan-kolb:
@chochreiner Great thanks! Is LibreOffice missing the `unoil.jar` inside the `program/classes` folder probably?
The logic basically is:
1. Look inside `/Applications` for folder names that contain (case insensitive) `libreoffice | openoffice` folder names
2. Check for `soffice.bin` executable mostly inside `program/`
3. Check for the UNO Jars `"unoil.jar", "jurt.jar", "juh.jar", "ridl.jar"` mostly inside `/prgram/classes/`.

chochreiner:
libreoffice seems to have a new internal package structure compared to OO
they have migrated to dylibs

stefan-kolb:
Hm, I'm not sure what this means for our current logic. As far as it is implemented right now every connection will fail if any of those libs are missing. Still it was perfectly fine for a few tests even without the `unoil.jar`.

src/main/java/org/jabref/logic/openoffice/OpenOfficeFileSearch.java
stefan-kolb: Yeah i tried all of these. Seems more code and edge cases handling than the current one, e.g. filter only directories catch IOExceptions...

src/main/java/org/jabref/gui/openoffice/DetectOpenOfficeInstallation.java
stefan-kolb: Can I already use these dialogs for selecting directories only? Looks like they can only be configured as file dialogs.

Siedlerchr:
LibreOffice provides this libs: 
https://mvnrepository.com/artifact/org.libreoffice

According to his project https://github.com/LibreOffice/loeclipse the libs can be found:

> Search for the LibreOffice jars unoil.jar juh.jar jurt.jar ridl.jar unoloader.jar. You find them here:
> 
>     macOS: /Applications/LibreOffice.app/Contents/Resources/java
>     Ubuntu: /usr/lib/libreoffice/program/classes
>     Windows: C:\Program Files\LibreOffice 5\program\classes\
> 

Now there should be no more project errors.

stefan-kolb:
@Siedlerchr I installed LO 5 on Windows and Ubuntu and the `unoil.jar` is missing, the rest is there.

Siedlerchr:
@stefan-kolb  Under Windows 10, LO 5.3 64Bit  the unoil.jar is there in `C:\Program Files\LibreOffice 5\program\classes,` for Ubuntu you need to install the additional package:
http://packages.ubuntu.com/de/xenial/libreoffice-java-common

stefan-kolb:
@Siedlerchr Hm, that's bad for us as the user needs to install that manually...probably there is an equal problem with OSX.

src/main/java/org/jabref/gui/openoffice/DetectOpenOfficeInstallation.java
stefan-kolb: We need a DirectoryChooser there as it seems.

chochreiner:
i can confirm that all jars are availalable in: /Applications/LibreOffice.app/Contents/Resources/java

Siedlerchr:
We then should add this at least for linux to the installation instructions that the -common-java package is required

stefan-kolb:
@chochreiner Hm, ok then the automatic detection should work...

stefan-kolb:
@Siedlerchr @tobiasdiez I have added a new directory chooser in the FX dialog service. Can you please take a look? One thing that I noticed is that all of these dialogs have no parent (i.e., not model), therefore hide in the background if the user clicks somewhere else. Dunno if we want them to behave like this...

chochreiner:
the OpenOffice provides both soffice.bin and soffice; LibreOffice only provides soffice --> with this fix, the automatic detection works for me

stefan-kolb:
@chochreiner Ah nice, thanks so much Christoph!

stefan-kolb:
@Siedlerchr Can you add your find for Ubuntu to the doucumentation please?

stefan-kolb:
@JabRef/developers Otherwise I'm done with this. Thanks everybody for your help and input 😄 👍

stefan-kolb:
I also added our finds to the documentation under __Known issues__.

PR #2649, opened 2017-03-15
Paying off technical debt: almost all utility classes have a private constructor now.

TRvanRossum:
I also removed an unused private method. According to SonarQube scans, this removes at least five days of technical debt.


stefan-kolb:
Hm, not sure if I like the addition of those constructors just to prevent the instantiation of those classes?! 
As it seems this is not done anywhere anyway.
I don't see much use right now, maybe you can clarify this a bit.
Lets see what @JabRef/developers  think.

src/main/java/org/jabref/gui/exporter/ExportAction.java
```diff
@@ -33,6 +33,9 @@
 
     private static final Log LOGGER = LogFactory.getLog(ExportAction.class);
 
+    private ExportAction() {
```
Siedlerchr: I doubt that this is useful here. I am sure Export Action is somewhere called

Siedlerchr:
I would have not cared about this, as it is normally pretty obvious that you should not create an object of a class which only has static methods, but it is actually recommended http://www.informit.com/articles/article.aspx?p=1216151&seqNum=4

But I think we agreed a while ago that we focus more on fixing/implementing issues (where you can resolve some code quality issues locally) than solely on internal code quality.

PS: You could also take a look at the modernizer test output log and maybe fix some of the mentioned issues as well (I would suggest ignoring group related stuff for the moment)

src/main/java/org/jabref/gui/exporter/ExportAction.java
TRvanRossum: I just had a look at that. `ExportAction` is indeed instantiated in `org.jabref.gui.preftabs.PreferencesDialog`, but that class has its own `ExportAction` class in it. `PreferencesDialog` also has no import for `org.jabref.gui.exporter.ExportAction`, so the instantiation is most likely an instantiation of the inner class.

Siedlerchr: Ah okay, good to know, seems like they are different exportActions here. 
Some time ago I was working with these action classes and therefore was just crurious.

TRvanRossum:
The merge conflicts should be solved now.

PR #2650, opened 2017-03-15
Performance improvements related to import

tobiasdiez:
Adding a larger number of entries (e.g. after "Web search" or import) currently takes very long. This PR tries to remove some overhead: add `insertEntries` method that admits many entries as arguments so that events are only fired once (and not after every entry). 
The performance is better with these changes but still not good enough (importing something into a very huge db leads to JabRef not responding).

Moreover, instead of creating many threads for the computation of the "hit number" in the groups panel, a thread pool is used.


src/main/java/org/jabref/JabRefExecutorService.java
```diff
@@ -147,6 +119,31 @@ public void shutdownEverything() {
         // kill the remote thread
         stopRemoteThread();
         // timer doesn't need to be canceled as it is run in daemon mode, which ensures that it is stopped if the application is shut down
+
+        Globals.taskExecutor.shutdown();
```
lenhard: With this statement you are creating a dependency from this relatively independent class to the gui package. I really do not like this. I think the shutting down of the gui-related `taskExecutor` should not happen in this class, especially since this one line is really the only position where it interacts with `JabRefExecutorService`. This is not cohesive.

src/main/java/org/jabref/Globals.java
```diff
@@ -19,32 +21,26 @@
     public static final RemoteListenerServerLifecycle REMOTE_LISTENER = new RemoteListenerServerLifecycle();
 
     public static final ImportFormatReader IMPORT_FORMAT_READER = new ImportFormatReader();
-
-
+    public static final TaskExecutor taskExecutor = new DefaultTaskExecutor();
```
lenhard: `Globals` is a class I would like to see only go down in size, but I see that the thread pool has to be managed somewhere. I would suggest to add a method here that shuts down everything, the GUI-independent thread pools and this GUI-dependent thread pool and to call this method form `JabRefMain` instead of the current code.

src/main/java/org/jabref/gui/groups/GroupNodeViewModel.java
```diff
@@ -184,10 +188,11 @@ private void calculateNumberOfMatches() {
         // We calculate the new hit value
         // We could be more intelligent and try to figure out the new number of hits based on the entry change
         // for example, a previously matched entry gets removed -> hits = hits - 1
-        new Thread(()-> {
-            int newHits = groupNode.calculateNumberOfMatches(databaseContext.getDatabase());
-            Platform.runLater(() -> hits.setValue(newHits));
-        }).start();
+        BackgroundTask<Integer> calculateHits = BackgroundTask
+                .run(() -> groupNode.calculateNumberOfMatches(databaseContext.getDatabase()))
```
lenhard: This seems little, but is very important: Please rename the method `run()` in `BackgroundTask` to something else, suggestion: `wrap()`. `run()` has a very specific meaning in parallel Java code and by just looking at this code here, I thought: "This code runs a task not in parallel, but on the calling thread". And this is not at all what this code does. It just wraps a `Callable` as a `BackgroundTask`, execution comes later. The name should reflect that.

src/main/java/org/jabref/gui/util/DefaultTaskExecutor.java
```diff
@@ -19,9 +21,27 @@
...
+
     @Override
     public <V> void execute(BackgroundTask<V> task) {
-        new Thread(getJavaFXTask(task)).start();
+        executor.submit(getJavaFXTask(task));
+    }
+
+    @Override
+    public void shutdown() {
+        executor.shutdown();
```
lenhard: Since this is about gui-related task, I assume it would be ok to try to kill them when shutting down instead of waiting until they are finished. Therefore, you should use `shutdownNow()` here, which will interrupt tasks that are executing at the moment of the invocation of the method.

src/main/java/org/jabref/model/database/BibDatabase.java
```diff
@@ -193,21 +194,32 @@ public synchronized boolean insertEntry(BibEntry entry) throws KeyCollisionExcep
...
-        entries.add(entry);
-        entry.registerListener(this);
+    private synchronized void insertEntries(List<BibEntry> newEntries, EntryEventSource eventSource) throws KeyCollisionException {
+        Objects.requireNonNull(newEntries);
 
-        eventBus.post(new EntryAddedEvent(entry, eventSource));
-        return duplicationChecker.isDuplicateCiteKeyExisting(entry);
+        for (BibEntry entry : newEntries) {
+            String id = entry.getId();
+            if (containsEntryWithId(id)) {
```
lenhard: This code is probably the reason why we are experienced performance issues when importing entries to larger DBs. Maybe we can find a better way for this check, but this does not have to be part of the current PR.

tobiasdiez: I'm not sure that this is really code is really the problem - `containsEntryWithId` is just a lookup in a relatively small (< 10.000) hashset. I have the feeling that our logic and model code is quite optimized or at least performs reasonable well but there are some problems with GUI code (e.g. adding 10 entries at the same time leads to 10 separate redraws of the maintable).

tobiasdiez:
Thanks for the valuable feedback! You raised good points, which I used to improve the code. Will merge now (as soon as build passes).

Issue #2173, opened 2016-10-17
Google Scholar fetching not working


Djeed:
JabRef 3.6
linux 3.13.0-86-generic amd64 
Java 1.8.0_101 on 131-Ubuntu SMP Thu May 12 23:33:13 UTC 2016

Google Scholar fetching returns an error message:
"Error while fetching from Google Scholar".
As I have read in previous threads, there is a recurrent problem with this. Could it be that Google Scholar has changed something from their API again?

matthiasgeiger:
Please try the current Dev builds. The issue in JabRef 3.6 should be already fixed, but is mit yet released in a 3.7 Version.

You can find the builds at https://builds.jabref.org/master

Djeed:
The latest build (.jar version) is working properly regarding this issue. 
Thank you!

koppor:
Duplicate of #1886.

I would vote to leave this issue open (and label with `fixed-in-devBuilds` as more and more users are asking for a fix.

Djeed:
Hello, after a few minutes of work, the error pops up again :(
"Error while fetching from Google Scholar. Please try again later and/or check your network connection". Not sure whether it is another change from Scholar or something about the dev version in my computer...

matthiasgeiger:
The problem with google scholar is that fetching via JabRef is creating a lot of requests at the Google Servers: One for the search page + 10 for the found entries (`bib` download for each entry). These (quick) requests are detected by Google as machine-requests - and if you perform some more searches - your IP address will be "blocked". I.e., a "captcha" is shown which cannot be passed by JabRef. 

Thus, the search does not work for "mass downloads" of entries. If you are using Firefox you might consider using the [Plugin "JabFox"](https://addons.mozilla.org/de/firefox/addon/jabfox/) written by one of our developers (@tobiasdiez) to add entries directly from the browser which will not be blocked.

AEgit:
As a workaround, what you could try is to change your IP address (using a VPN service) and potentially also your MAC address. See:
http://help.jabref.org/en/GoogleScholar

and

https://github.com/JabRef/jabref/issues/2046

Djeed:
I see... So it is rather undetermined when the fetching will block right? Hmm... I guess this feature will keep being problematic. Anyway thanks for the info!

tobiasdiez:
@matthiasgeiger Can we somehow detect that Google is blocking the request (using the response code or analyzing the returned html)? Then we could at least show a better error message.

[...]

PR #2654, opened 2017-03-16
Disalbe Google Scholar tests on all CI environments

koppor:
Google Scholar tests currently fail. This PR just follows the comment given within the test cases and disables the Google Scholar tests on **all** CI Servers.

We might revert this if we have time to fix Google Scholar.

Refs #2173

### DIFF
--- a/src/test/java/org/jabref/logic/importer/fetcher/GoogleScholarTest.java
+++ b/src/test/java/org/jabref/logic/importer/fetcher/GoogleScholarTest.java
@@ -53,7 +53,7 @@ public class GoogleScholarTest {
     @Test
     public void linkFound() throws IOException, FetcherException {
         // CI server is blocked by Google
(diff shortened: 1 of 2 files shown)


stefan-kolb:
For me they do not fail all the time. I checked them locally. They just randomly fail?!

Siedlerchr:
I think they fail when they are executed more than once in a couple of
minutes, e.g when you push several commits in a few seconds and each time
the build is triggered they will fail

At least that's what I noticed

Am 16.03.2017 3:15 nachm. schrieb "Stefan Kolb" <notifications@github.com>:

For me they do not fail all the time. I checked them locally. They just
randomly fail?!

—
You are receiving this because you are subscribed to this thread.
<https://github.com/JabRef/jabref/pull/2654#issuecomment-287070299>, or mute
the thread

PR #2661, opened 2017-03-17
Add possibility to clear the log at runtime

### LINKED ISSUE #2660: Add an option to clear the log inside the error console
In order to reproduce issues without restarting JabRef.

stefan-kolb:
Resolves #2660

@tobiasdiez The dialog is not modal, but does not update itself when new events arrive, i.e., needs to be closed and opened again to see new events. Is this expected? Can we fix this?


tobiasdiez:
Yeah @stefan-kolb is now an active JavaFX developer 😼 . Code looks good, so I'll merge this. Actually, new logs should be displayed automatically in the console (this is why it is modal) - thus this is probably a bug (can you please open a new issue; I don't have time right now to further investigate this; thanks).

Issue #457, opened 2015-12-03
Add Key Pattern-specific Regular Expressions


jkulesza:
My request is to incorporate regular expressions into the key pattern dialog to modify the key patterns.  It looks like this may have been [done previously](https://github.com/korv/Jabref-plugins) when Plugins were available.  This would effectively extend the `:(x)` functionality there which only matches on null fields.

For example, I hope that the following would be a valid "Book" key pattern:

`Book_[keywords]_[year]_[authors1]_[title:regex(" ","-")]`

which would replace all spaces in the title by dashes.

lenhard:
@jkulesza Thanks for this suggestion. You are welcome to provide a pull request.

tobiasdiez:
This was implemented with https://github.com/JabRef/jabref/pull/3096. 
Could you please check the build from http://builds.jabref.org/master/. Thanks!

PR #2663, opened 2017-03-17
[WIP] Testing bibtex key patterns and fixing title_case/capitalize modifiers

RolfStarre:
Related to https://github.com/koppor/jabref/issues/237
I'm working on adding tests for key patterns + key patterns in combination with modifiers.

I found out why the title_case and capitalize modifiers didn't work with shorttitle. It went wrong because before the modifier was applied the shorttitle was already a concatenated string, so the formatter would just see it as a single word.
For example if we have the title "An Awesome Paper on JabRef", it would first turn into "AnAwesomePaper", and then the modifier would turn it into "Anawesomepaper".

### DIFF
--- a/src/main/java/org/jabref/logic/bibtexkeypattern/BibtexKeyPatternUtil.java
+++ b/src/main/java/org/jabref/logic/bibtexkeypattern/BibtexKeyPatternUtil.java
@@ -726,7 +726,7 @@ public class BibtexKeyPatternUtil {
      * Determines "number" words out of the "title" field in the given BibTeX entry
      */
     public static String getTitleWords(int number, String title) {
-        return keepLettersAndDigitsOnly(getTitleWordsWithSpaces(number, title));
+        return getTitleWordsWithSpaces(number, title);
     }
 
     /**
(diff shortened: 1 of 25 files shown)


Siedlerchr:
Just as an idea, maybe you can check if you could include that feature request here?
https://github.com/JabRef/jabref/issues/457

RolfStarre:
@Siedlerchr I've made a first attempt at adding the feature ^_^

RolfStarre:
I have a problem with testing the regexFormatter, it doesn't take protected parts of a sentence into account. 

I don't really know how I can write the code so that it does take the protected characters into account, since it currently only seems to be done on a letter by letter basis.

Siedlerchr:
Maybe anyone of the other @JabRef/developers  has an idea on how to check that

koppor:
@tobiasdiez Maybe you have an idea?

tobiasdiez:
Sorry, I've no deep knowledge about regular expressions in particular and about the formatters in general.

koppor:
I understand the question https://github.com/JabRef/jabref/pull/2663#issuecomment-287796215 as follows:

If I replace regex `a` by `X` and I apply it at `abc{abc}`, the result is `Xbc{Xbc}`, but it should be `Xbc{abc}`.

As one cannot change the regex engine, I would `{...}` replace by something not matched by the regular expression, then apply the regex and then change it back.

1. Replace protected terms
1a. Replace first `{...}`by [unicode character 2580](https://unicode-table.com/de/#2580)
1b. Replace second `{...}`by [unicode character 2581](https://unicode-table.com/de/#2581)
1.c etc.
2. Apply regex
3. replace back

koppor:
Closing this issue due to inactivity :zzz: 

@shitikanth will take over in another PR, because it is unlikely, that we get access to https://github.com/delftswa2017/jabref

PR #2666, opened 2017-03-18
Add sorting of all groups and subgroups, recursively

Siedlerchr:
Implements some features from #2599
Recursively sort all groups or selected groups 

Replaced Screenshot

grafik

Sometimes the sorting is not always directly correct visible after sorting, This is somehow an update problem of the gui I think. When you switch db back and forth it's correctly there.

~~- [ ] Change in CHANGELOG.md described~~
~~- [ ] Tests created for changes~~
~~- [ ] Check documentation status (Issue created for outdated help page at [help.jabref.org](https://github.com/JabRef/help.jabref.org/issues)?)~~


src/main/java/org/jabref/gui/groups/GroupTreeController.java
```diff
@@ -43,14 +43,21 @@
 
     private static final Log LOGGER = LogFactory.getLog(GroupTreeController.class);
 
-    @FXML private TreeTableView<GroupNodeViewModel> groupTree;
-    @FXML private TreeTableColumn<GroupNodeViewModel, GroupNodeViewModel> mainColumn;
-    @FXML private TreeTableColumn<GroupNodeViewModel, GroupNodeViewModel> numberColumn;
-    @FXML private TreeTableColumn<GroupNodeViewModel, GroupNodeViewModel> disclosureNodeColumn;
-    @FXML private CustomTextField searchField;
+    @FXML
+    private TreeTableView<GroupNodeViewModel> groupTree;
```
tobiasdiez: Please leave the formatting as it is right now. It is kind of a JavaFX-thing that normally the @FXML tag is on the same line (and I like it that way because otherwise the private fields declaration takes more space)

Siedlerchr:
Yeah I will change that, however the fxml tag are not in the formatting yet
I will try to change that as well

Am 19.03.2017 12:07 vorm. schrieb "Tobias Diez" <notifications@github.com>:

*@tobiasdiez* approved this pull request.

The code looks good. I would, however, argue that "sort subgroups" should
suffice - if you want to sort the whole tree then just right click the
root. And maybe add "recursively" to the menu text to make it a bit clearer.

By the way, we should discuss at some point which other menu entries make
sense and which we should remove instead of reimplementing them (e.g. I
don't see a point in the move actions).
------------------------------

In src/main/java/org/jabref/gui/groups/GroupTreeController.java
<https://github.com/JabRef/jabref/pull/2666#discussion_r106793485>:

> @@ -43,14 +43,21 @@

     private static final Log LOGGER =
LogFactory.getLog(GroupTreeController.class);

-    @FXML private TreeTableView<GroupNodeViewModel> groupTree;
-    @FXML private TreeTableColumn<GroupNodeViewModel,
GroupNodeViewModel> mainColumn;
-    @FXML private TreeTableColumn<GroupNodeViewModel,
GroupNodeViewModel> numberColumn;
-    @FXML private TreeTableColumn<GroupNodeViewModel,
GroupNodeViewModel> disclosureNodeColumn;
-    @FXML private CustomTextField searchField;
+    @FXML
+    private TreeTableView<GroupNodeViewModel> groupTree;

Please leave the formatting as it is right now. It is kind of a
JavaFX-thing that normally the @FXML <https://github.com/FXML> tag is on
the same line (and I like it that way because otherwise the private fields
declaration takes more space)

—
You are receiving this because you authored the thread.
<https://github.com/JabRef/jabref/pull/2666#pullrequestreview-27736134>,
or mute
the thread

Siedlerchr:
@tobiasdiez  Changed localization, but I dunno how to display the menu entry only when the root nod eis selected?
I read about adding a change listener and then setting the contextmenu. However, I am not sure how that plays together with the binding stuff

tobiasdiez:
In `createContextMenuForGroup` you can use `isRoot` of the `GroupNodeViewModel group` to determine if the right-clicked group is the root or not...and then create the menu depending on this value.

However, I would completely leave of the `sort subgroups` option.

Siedlerchr:
I thin we should stuck with simply having the sort groups recursively. Then we could omit the explicit chek for root node

Issue #2652, opened 2017-03-16
Incorrect display in the author entry column


bvdalen:
Version 3.8.2, confirmed in JabRef 4.0.0-dev--snapshot--2017-03-16--master--4323923a8, Windows 7 6.1 amd64 and Java 1.8.0_92.

The below BibLaTeX file illustrates the following problems in JabRef's display of the author entry column:

1. The single name "Rodriguez Fernandez, José María" in Item:1 is turned into two authors F. Rodriguez and J. María. This does not happen in the alternative format in Item:2, nor in combination with another author in either format (Item:3 and Item:4).

2. In the format "Waard, Gerard de" the prefix "de" is abbreviated as "d." and comes at the end of the name as displayed in the author column, but in the alternative equivalent format "Gerard de Waard" it is not abbreviated and appears at the beginning (but is correctly ignored in sorting). I think the latter display form is preferable, but in any case it reads easier if the prefix is not abbreviated.

3. Besides the prefix "de" also the first of a double surname seems to be ignored in sorting the author column for certain (but not all) input formats: Item:2 and Item:3 are apparently sorted in under F, but item:4 correctly under R.

Sample BibLaTeX file:
```
% Encoding: UTF-8

@Article{Item:1,
  author    = {Rodriguez Fernandez, José María},
  title     = {L'arc du ciel},
  date      = {2016},
}

@Article{Item:2,
  author    = {José María {Rodriguez Fernandez}},
  title     = {L'arc du ciel},
  date      = {2016},
}

@Article{Item:3,
  author    = {José María {Rodriguez Fernandez} and Millás Vallicrosa, José María},
  title     = {L'arc du ciel},
  date      = {2016},
}

@Article{Item:4,
  author    = {Rodriguez Fernandez, José María and Millás Vallicrosa, José María},
  title     = {L'arc du ciel},
  date      = {2016},
}

@Article{Item:5,
  author    = {Waard, Gerard de},
  title     = {Jabref's Issues},
  date      = {2017},
}

@Article{Item:6,
  author    = {Gerard de Waard},
  title     = {Jabref's Issues},
  date      = {2017},
}

@Article{Item:7,
  author    = {José Martinez},
  title     = {Jabref's Issues},
  date      = {2017},
}

@Comment{jabref-meta: databaseType:biblatex;}
```

Maformatiker:
I think the second point is not a bug. The two possible formats are de Waard, Gerard and Gerard de Waard.
[...]

PR #2669, opened 2017-03-19
Fixed issue #2652

Maformatiker:
Commas should only be interpreted as separators of several authors if there is no possible interpretation that conforms to the BibTeX rules, i.e. if there are more than two commas.


Maformatiker:
Hello Tobias,
I looked at the falied tests and realized that one test fails exactly because he expects what I considered unintended behavior. So my proposed change is not just an undisputed bugfix but due to different expectations on which we have to agree. The question is how to interpret inputs of the form

1. Canon der Barbar, Alexander der Große
2. Rinnooy Kan, Alexander H. G.
3. Rinnooy Kan, Alexander Hendrik George
4. Rodriguez Fernandez, José María

Currently such inputs are interpreted as two authors and normalized to
1. der Barbar, Canon and der Große, Alexander
2. Kan, Rinnooy and Alexander H., G.
3. Kan, Rinnooy and George, Alexander Hendrik
4. Fernandez, Rodriguez and María, José

I think the entries should be interpreted as one author each, i.e.
1, lastname="Canon der Barbar", firstname="Alexander der Große"
2. lastname="Rinnooy Kan", firstname="Alexander H. G."
3. lastname="Rinnooy Kan", firstname="Alexander Hendrik George"
4. lastname="Rodriguez Fernandez", firstname="José María"

However this leads to the error in the JUnit tests.

Siedlerchr:
Thank your for your effort. As you already noticed this is a serious issue.  This really depends on the data you have. Some sites export all authors with commas if I remember correclty

tobiasdiez:
Thanks for the additional investigative research 😄. I'm strongly in favor that `Rinnooy Kan, Alexander Hendrik George` is parsed as one author! The biblatex code contains a [test](https://github.com/plk/biber/blob/3a5b7e6aaef9e87a021b97a4b8b14a9c300e2328/t/names.t#L868) that the similar input `von Berlichingen zu Hornberg, Johann Gottfried` is parsed as
````
      given          = {'Johann Gottfried', initial => ['J', 'G']},
      family         = {'Berlichingen zu Hornberg', initial => ['B', 'z', 'H']},
      prefix         = {'von', initial => ['v']},
      suffix         = {undef, initial => undef},
````
In my opinion, we should follow biblatex here. However, I remember that we already had a long discussion about similar issues some time ago https://github.com/JabRef/jabref/pull/1470. Indeed, we should probably discuss this in the next devcall.

**Comment by matthiasgeiger** [comment]
Okay and how to distinguish `Berlichingen zu Hornberg, Johann Gottfried` from `Tobias Diez, Matthias Geiger` ?

Or even more abstract:
`A a B, C D, E F` 

How much authors are this? 3? 2? What are the given and family names? 

This is not possible to detect the authors free from errors as there various different formats out there in the wild. Especially, as long as only `,` are used as seperators I don't see a way to "fix" this...

File: src/main/java/org/jabref/model/entry/AuthorList.java [hunk 1]
```diff
@@ -167,7 +167,8 @@ public static AuthorList parse(String authors) {
 
         // Handle case names in order lastname, firstname and separated by ","
         // E.g., Ali Babar, M., Dingsøyr, T., Lago, P., van der Vliet, H.
-        if (!authors.toUpperCase(Locale.ENGLISH).contains(" AND ") && !authors.contains("{") && !authors.contains(";")) {
+        if (!authors.toUpperCase(Locale.ENGLISH).contains(" AND ") && !authors.contains("{") && !authors.contains(";")
+                && ((authors.length() - authors.replace(",", "").length()) > 2)) {
```
koppor: Can the last condition be extracted into a variable?

koppor:
Decision: AB, CD -> one author, AB, CD, EF: Three authors

koppor:
As far as I understand, this will partially revert https://github.com/JabRef/jabref/pull/1470, which was implemented because of https://github.com/JabRef/jabref/issues/1181, which was risen because of https://github.com/koppor/jabref/issues/125.

The discussion revealed that users have to change `John Maindonald, W. John Braun` manually to `John Maindonald and W. John Braun`, but `A B, C D, E F` is automatically converted to `B, A and D, C and F, E`.

lenhard:
So how should this PR proceed? @koppor: I do no longer remember the discussion from the devcall. Can we get the PR mergeable? If so, what needs to be done?

koppor:
I included the decisions (one comma should be treated as special case) and added a test case. Will merge after travis succeeded.

koppor:
@Maformatiker Thank you for working on this and improving our JavaDoc. I hope, that you are satisfied with our updates. Looking forward to more pull requests 😇

PR #2681, opened 2017-03-23
Fix #2680 and fix #2667: Swing errors are catched properly and without freezing

### LINKED ISSUE #2667: Error console does not update on new events
The dialog is not modal, but does not update itself when new events arrive, i.e., needs to be closed and opened again to see new events.

### LINKED ISSUE #2680: Lookup DOI (sometimes) freezes JabRef
New in 4.0 Development version!

tobiasdiez:
Hopefully, fixes that the error console is not updated properly #2667 as well as that Swing errors lead to frezzing #2680. At least for me, it works now without problems. @stefan-kolb @lynyus can you please also check it.

For the error console, I took the suggestion https://github.com/JabRef/jabref/issues/2667#issuecomment-288742546 by @matthiasgeiger. The freezing after a Swing message occurred since 
̀`LOGGER.warn("No localized message exception message defined. Falling back to getMessage().");`
lead to recursive calls (actually, Log4j has some logic to prevent deadlocks but these don't take effect since the original message is thrown in the Swing thread, while the above "no localized messsage" is posted in the JavaFX thread.). I simply changed it to log level "debug", which effectively hides it in most cases. As a positive side effect, the scroll issue mentioned in https://github.com/JabRef/jabref/issues/2667#issuecomment-288714639 is also fixed.

### DIFF
--- a/src/main/java/org/jabref/JabRefException.java
+++ b/src/main/java/org/jabref/JabRefException.java
@@ -5,9 +5,8 @@ import org.apache.commons.logging.LogFactory;
 
 public class JabRefException extends Exception {
 
-    private String localizedMessage;
-
     private static final Log LOGGER = LogFactory.getLog(JabRefException.class);
+    private String localizedMessage;
 
     public JabRefException(String message) {
         super(message);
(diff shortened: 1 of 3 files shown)


stefan-kolb:
Will look at it. Please do not close #2680 tho, as it includes further things that need to be fixed.

tobiasdiez:
@matthiasgeiger You are right, my changes are more a workaround than a proper solution. However, since the deadlock only appears if an exception is logged in the Swing thread and afterwards the GUIAppender also logs something, I decided not to invest more time. Probably a application-wide lock for the appender is needed.

Issue #2615, opened 2017-03-05
Ctrl+a selects first entry if entry editor is opened


koppor:
JabRef version as of today on Win 10.

Steps to reproduce:

1. Double click third entry
2. Click on third entry to focus it on the main table
3. Press <kbd>Ctrl</kbd> + <kbd>A</kbd>
4. Expected behavior: All entries are selected.

![ctrl-a](https://cloud.githubusercontent.com/assets/1366654/23591197/15690b1c-01ed-11e7-8453-067ccf7f38a1.gif)

lenhard:
Related: #2322 

Wasn't this the expected behavior? But anyway, it would be perfectly fine by me to ditch emacs-specific behavior and just do a select all.

Btw.: An interesting color scheme you have there :-)

Siedlerchr:
Isn't that depending on the key settings you have enabled?
I personally prefer that option to select all entries and not the current
text

2017-03-06 8:45 GMT+01:00 Jörg Lenhard <notifications@github.com>:

koppor:
Ctrl+A should always select all text and not the first entry in the table. This was the behavior in JabRef 3.2 and changed somewhere from there to today.

![jr32 gif](https://cloud.githubusercontent.com/assets/1366654/23661577/82a85b2c-034d-11e7-8485-af93f03cf1f0.gif)

stefan-kolb:
Not sure what's wrong there. We just call `actions.put(Actions.SELECT_ALL, (BaseAction) mainTable::selectAll);` in BasePanel. selectAll is a JDK method on the MainTable, i.e., JTable.

lenhard:
I guess some other class, probably something related to the entry editor is stealing the event. The first press of CTRL+A then puts focus on the main table so that the second action is not stolen. I guess it will require some searching to find the guilty one.

koppor:
Introduced in https://github.com/JabRef/jabref/pull/2199

@bartsch-dev Can you have a look?

PR #2683, opened 2017-03-24
Fix selecting all entries with shortcut

bartsch-dev:
Fixes https://github.com/JabRef/jabref/issues/2615

It seems that I was a bit too fast when refactoring this code block.


koppor:
When the entry editor is opened, the first click on the main table causes no reaction or reshuffles the entry table. Is this caused by this fix or is this an old issue?

Ctrl+A works if I manage to focus the entry table.

bartsch-dev:
woops, missed another one

PR #2684, opened 2017-03-24
FileAnnotationTab now only shows the FileName, not the full path.

### LINKED ISSUE #2676: Filename field in file annotations tab is too small when filename exceeds length of field
Happens when filename is too long for the default field length.

image

lynyus:
Fixes #2676


Siedlerchr:
Okay, LGTM

PR #2692, opened 2017-03-29
Add PDF Viewer

tobiasdiez:
Since @lynyus wants to implement a fulltext pdf search 😛, we need a proper way to display the results. In this PR a very plain PDF viewer is implemented. Most  #PDFs are displayed correctly, maybe with a few small glitches here and there. Basic features like scrolling, changing pages and zooming work. In addition, while being in "live mode" the linked document of the currently selected BibEntry is shown - this is very handy in a two-display setup since one can have JabRef's maintable on one screen and the always-current PDF on the second.

untitled

Feedback welcome! (I will update the language files before merge.)

- Similar project: https://github.com/IDRSolutions/maven-OpenViewerFX-src
- Extended example: https://github.com/james-d/PdfViewer/blob/master/src/pdfviewer/PdfViewer.fxml and https://github.com/james-d/PdfViewer/blob/master/src/pdfviewer/Controller.java

----


stefan-kolb:
Code looks ok to me 👍 Need to try it out, still.

src/main/java/org/jabref/gui/documentviewer/DocumentViewerController.java
```diff
@@ -0,0 +1,126 @@
+/*
+ * Copyright (C) 2003-2016 JabRef contributors.
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
```
Siedlerchr: The header still days GPL

stefan-kolb:
Ok, tested it a bit. Some thoughts:
- live view is nice 👍 
- functionality currently quite hidden via `view - show document viewer`. Probably need a magnifier glass somewhere. Not sure where tho.
- PDF rendering is a bit slow.
- Also, it does not render backgrounds correctly and also the casing of some texts seem to be odd
image
image
image
image
- Highlighting does not display correctly
image
- Comments cannot be seen
- I can no longer navigate through the main panel with UP/DOWN buttons

__We already had such a viewer before and removed it__. Isn't it possible to do this natively with Acrobat or so? I doubt we will ever get a really good rendering by ourselves.

tobiasdiez:
Thanks for the feedback. I also noticed some of these rendering issues (and, in addition, some math characters are displaced or not shown at all). I did some investigation and it seems that most of these problems are fixed/improved upon with PDFBox 2.0. But the update to 2.0 is still not possible for us https://github.com/JabRef/jabref/pull/1096. The only other decent PDF rendering engine I found for java is itext, but it is licensed under the incompatible AGPL (this is also the reason why #2474 is blocked at the moment). 

We already had a PDF viewer? I was only aware of the PDF thumbnail, which was totally buged.
The idea behind this PR is not to implement a full-fledged PDF viewer and editor, but only provide a reasonable good minimal viewer. Its main purpose should be to let the user quickly decide whether he has picked the right document for further reading. In addition, I plan to show the preview also in the import dialog to give more context there. Some further plans are integration with and enrichment of the core bibliographic features of JabRef: pdf fulltext search, visualization of cites / cited by relationships, automatic import of referenced literature.
Thus the idea is kind of similar to the PDF previewer in firefox: provide a means for a quick inline PDF preview for convenience but don't try to write yet another PDF editor. 

I will have a look why the navigation in the maintable with arrow keys does not work anymore.

src/main/java/org/jabref/gui/documentviewer/DocumentViewerController.java
tobiasdiez: Good catch! Some of the code was written in an era, when such copyright notices were still included 😄

lenhard:
And one more thing that just came to my mind: What are the licenses of the libraries you are including here?

src/main/java/org/jabref/model/entry/ParsedFileField.java
```diff
@@ -1,7 +1,14 @@
 package org.jabref.model.entry;
 
+import java.io.File;
 import java.net.URL;
+import java.nio.file.Path;
 import java.util.Objects;
+import java.util.Optional;
+
+import org.jabref.logic.util.io.FileUtil;
```
tobiasdiez: May I also put `FileUtil` in `model` since it is our abstraction of the `File` object? Reason: I really want the `ParsedFileField` to have a method that returns the path to the linked file. There are around 15 places where `expandFilename` is used upon `ParsedFileField::getLink`.

tobiasdiez:
@lenhard Thanks for the additional feedback. 
I changed the log output of pdfbox to errors, so all the warnings are hidden now. The "-fx-background-color" messages seem to be a bug in JavaFX at least it referenced a code point in JavaFX's own stylesheet `modena.css`. 

Moreover, I added the licenses: Apache 2.0 and BSD 2, so this shouldn't be a problem.

src/main/java/org/jabref/model/entry/ParsedFileField.java
lenhard: The thing is, `FileUtil` has more dependencies to logic classes, which would then also need to move to model. At some point, the distinction between the two packages would be completely lost.

Can't you just add a method `public static Path getPath(ParsedFileField fileField)` to `FileUtil` instead? Not much better than the current version, I know, but at least a little. Alternatively, extract the methods you need, and just those, into a new `FileUtil` class in model (maybe with a better name). It wouldn't hurt to break up these util classes into more meaningful smaller units. They all devolve into god classes...

tobiasdiez: I moved the method to FileUtil as you proposed and will open a new PR with the refactoring since it turned out to be a bit more extensive than I have expected.

PR #2703, opened 2017-04-02
Fix #628: implement hierarchical keywords

### LINKED ISSUE #628: Feature: Hierarchical Keywords

tobiasdiez:
The automatic keywords group now supports hierarchical keywords. 
The changes mainly concentrate on the keywords class...so this may have some unwanted side effects in other parts (but I didn't noticed anything unusual in my tests).

Note: Bib files containing an automatic keyword from prior 4.0 tests can no longer be opened and the lines containing `AutomaticKeywordGroup` have to be deleted.

### DIFF
--- a/src/main/java/org/jabref/model/groups/AutomaticGroup.java
+++ b/src/main/java/org/jabref/model/groups/AutomaticGroup.java
@@ -4,8 +4,8 @@ import java.util.Set;
 
 import javafx.collections.ObservableList;
 
-import org.jabref.logic.util.TreeCollector;
 import org.jabref.model.entry.BibEntry;
+import org.jabref.model.util.TreeCollector;
 
 public abstract class AutomaticGroup extends AbstractGroup {
     public AutomaticGroup(String name, GroupHierarchyType context) {
--- a/src/main/java/org/jabref/model/groups/AutomaticGroup.java
+++ b/src/main/java/org/jabref/model/groups/AutomaticGroup.java
@@ -2,6 +2,9 @@ package org.jabref.model.groups;
 
 import java.util.Set;
 
+import javafx.collections.ObservableList;
+
+import org.jabref.logic.util.TreeCollector;
 import org.jabref.model.entry.BibEntry;
 
 public abstract class AutomaticGroup extends AbstractGroup {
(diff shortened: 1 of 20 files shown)


src/main/java/org/jabref/model/groups/AutomaticGroup.java
```diff
@@ -2,7 +2,10 @@
 
 import java.util.Set;
 
+import javafx.collections.ObservableList;
```
lenhard: I know this is convenient, but I am strongly against it. We do not want the model to be dependent on a GUI technology. If you want this conversion to be done based on `ObservableList`, it will have to be done elsewhere.

src/main/java/org/jabref/model/util/TreeCollector.java
```diff
@@ -0,0 +1,104 @@
...
+import java.util.Set;
+import java.util.function.BiConsumer;
+import java.util.function.BiPredicate;
+import java.util.function.BinaryOperator;
+import java.util.function.Function;
+import java.util.function.Supplier;
+import java.util.stream.Collector;
+
+import javafx.collections.FXCollections;
+import javafx.collections.ObservableList;
```
lenhard: Again, no javafx in model

src/main/java/org/jabref/model/groups/AutomaticGroup.java
tobiasdiez: I agree that we don't want to have GUI technology in our model and logic classes. However, these observable lists and properties have nothing to do with the GUI-related part of JavaFX - you can use them in a command line application if you want. An observable list is just a list with some built-in event listeners, i.e. a different form of List combined with the google event stuff. Similar the properties in in `javafx.beans` are just fields with event handlers.

Thus I strongly vote for allowing `javafx.collections` and `beans` in model and logic.

matthiasgeiger:
> Note: Bib files containing an automatic keyword from prior 4.0 tests can no longer be opened and the lines containing AutomaticKeywordGroup have to be deleted.

Isn't there a possibility to migrate this? From a user point of view this is not really acceptable...

tobiasdiez:
@matthiasgeiger I formulated it a bit awkwardly. Automatic keywords groups are a new feature in 4.0, so this only hits users of the current development version - not users of 3.8.2.

src/main/java/org/jabref/model/groups/AutomaticGroup.java
lenhard: Decision in devcall: make exception for `javafx.collections.*`

PR #2704, opened 2017-04-02
Write groups under tag `group` instead of `groupstree` to enhance compatibility with previous versions

tobiasdiez:
Right now, files written with 4.0 cannot be opened in older versions. The new format of the groups leads to a NPE. 
As a way out, I changed the tag for the groups tree from `groupstree` to `group` (`groups` was already used by previous version way back). Thus older versions completely ignore the group tree and no longer try to parse it, so no NPE is thrown. On the downside, however, the groups tree is deleted upon save in the older version. Still a bit better than a crashing Jabref in my opinion.

### DIFF
--- a/src/main/java/org/jabref/logic/importer/util/MetaDataParser.java
+++ b/src/main/java/org/jabref/logic/importer/util/MetaDataParser.java
@@ -61,37 +61,36 @@ public class MetaDataParser {
             }
 
             switch (entry.getKey()) {
-            case MetaData.GROUPSTREE:
+                case MetaData.GROUPSTREE:
                 case MetaData.GROUPSTREE_LEGACY:
-                metaData.setGroups(GroupsParser.importGroups(value, keywordSeparator));
-                break;
-            case MetaData.SAVE_ACTIONS:
-                metaData.setSaveActions(Cleanups.parse(value));
-                break;
-            case MetaData.DATABASE_TYPE:
-                metaData.setMode(BibDatabaseMode.parse(getSingleItem(value)));
-                break;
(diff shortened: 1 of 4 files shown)


Siedlerchr:
`On the downside, however, the groups tree is deleted upon save in the older version. Still a bit better than a crashing Jabref in my opinion.`

I wonder why JabRef does not simply ignore the groups Tree then and deletes it on save.

lenhard:
@tobiasdiez Even the old versions from 3.1 or 3.2 onward (not sure which) should not delete the new `groups` tree, since at some point we implemented the feature that JabRef does no longer delete random text from the bib file but keeps it. Anything prior to v3.0 will certainly delete it.

Anyway, I think this is a good compromise. Then people with older versions can still open the file and when they wonder that the groups tree is not correct (or gone), we can tell them to upgrade. 

I just think you should use `groups` instead of `group` as the key in the meta data, since this is normally not about a tree consisting of a single group, but of multiple groups.

tobiasdiez:
Well, even 3.8.2 deletes the `group` tag, probably since it is an unknown metadata item. 

I could use `groups` but it was already used in a very (?) old JabRef version...not sure which.

lenhard:
I see, with metadata the serialization might be different.

Hm, if it has been used before, we probably should not use `groups`. Since this name goes into the bib files of many people and will annoy us if we change it later, let us discuss a good version in the next devcall.

koppor:
Decision: `grouping`

Alternatives: `grouping`, `groupies`, `group`, `groups`

PR #2705, opened 2017-04-02
Fix a few smaller group related things 

tobiasdiez:
Will merge immediately since they are only small changes...


Siedlerchr:
You increased the StringUtil method, which resulted in an architecture test error
@lenhard  won't be amused :-P

lenhard:
I am not angry for increasing StringUtil, I am angry for just ignoring the broken test that resulted from the increase and merging this PR into master. That effectively breaks all builds on master. At the very least, you should change the test so that it passes.

I just did this as a quick-fix in d1c8569c5b1c169896517425082753fee34ee838

tobiasdiez:
Yes sorry didn't saw the failing test (first only the language tests were failing and I fixed them and merged directly afterwards). Thanks for the fix.

Issue #2539, opened 2017-02-14
Call for function: export the pdfs of all the selected entrys to a new folder


Vaver:
some times, i want share some of the literature pdfs to friends, it is very easy to select the wanted entrys from the Jabref, but hard to pick out the seleted pdfs.

I hope there is a function to export the pdfs of the selected entrys to a new folder.

thanks

AEgit:
I agree, that would a very nice feature to have. I guess, it is also already somewhat implemented in JabRef, as JabRef is already able to move files from one folder to another. Just copying (instead of moving) the files from one folder to another, user-specified folder would probably already do the trick, I guess.

Siedlerchr:
@Vaver  @AEgit  I have implemented the export functionality that works for all linked files:

http://builds.jabref.org/exportPDF/
You can use it like any other export function. Select the entries then choose  Export -> Files Exporter 
However there is currently one thing I can not change: You have to enter an arbitrary string in the save dialog as "filename".  The entered name is ignored, it is just needed to get the directory.

AEgit:
@Siedlerchr : Thank you very much! I can confirm that this feature works in the posted build!

And immediately a warning or potential improvement against (my) typical noob behaviour. I went for "Export" instead of "Export selected". This lead to JabRef copying over all linked files (in my case ~100 GB) - even when you stop JabRef, it will continue copying over those files. Only by killing the relevant process I was able to stop the copy process.
To prevent this, it might be helpful to add a warning or something along the lines (or let the noobs be noobs ;)).

Thanks again!

Siedlerchr:
Thanks for reporting. I will try to see if I can add a break to stop the process.

Siedlerchr:
@Vaver  @AEgit  I recently finished an improved functionality of this feature. Could you please test and give feedback on it?
http://builds.jabref.org/exportPdf/

AEgit:
JabRef 4.1-dev--snapshot--2017-10-26--exportPdf--88faa7ce3
Windows 10 10.0 amd64 
Java 1.8.0_151

@Siedlerchr This looks really good! Thank you very much!

[...]

PR #2706, opened 2017-04-02
[WIP] Create PDF file Exporter

Siedlerchr:
Create an exporter that enables the export of Files.
Only caveat: You have to enter an arbitrary filename in the save dialog. The exporter then uses the directory.

Implements #2539


src/main/java/org/jabref/logic/exporter/IExportFormat.java
```diff
@@ -38,18 +37,4 @@
...
-     * Gets the path as a java.nio.path instead of a string.
-     *
-     * @param databaseContext the database to export from.
-     * @param file  the Path to the file to write to.The path should be an java.nio.Path
-     * @param encoding  The encoding to use.
-     * @param entries  A list containing all entries that
-     * should be exported. The list of entries must be non null
-     * @throws Exception
-     */
-    void performExport(BibDatabaseContext databaseContext, Path file, Charset encoding, List<BibEntry> entries)
```
tobiasdiez: I actually prefer these export methods that take the file as a `Path` instead of string.

src/main/java/org/jabref/logic/exporter/LinkedFileExporter.java
```diff
@@ -0,0 +1,50 @@
...
+import org.jabref.logic.TypedBibEntry;
+import org.jabref.logic.util.io.FileUtil;
+import org.jabref.model.database.BibDatabaseContext;
+import org.jabref.model.entry.BibEntry;
+import org.jabref.model.entry.ParsedFileField;
+
+public class LinkedFileExporter extends ExportFormat {
+
+    public LinkedFileExporter() {
+        super("Files Exporter", "Files", null, null, ".*");
```
tobiasdiez: Maybe "Export linked files" as name? not sure.

src/main/java/org/jabref/logic/exporter/IExportFormat.java
Siedlerchr: Yeah, I wanted to use it, too. However, the method with that signature never gets called which led to an NPE.
In ExportAction only the one with String is called.

koppor:
Should we also move the "Send es email" functionality to the export menu?

simonharrer:
Hm, I would distinguish between Export and Share. Send as Email sounds to be me like sharing it whereas Export is a lowerlevel primitive which is not always for sharing the data.

src/main/java/org/jabref/logic/exporter/IExportFormat.java
tobiasdiez: Maybe change `ExpoortAction` then? For now the `Path` method can be implemented as `default` and invoking the stringy one.

tobiasdiez:
I like the suggestion by @simonharrer. "Share" fits better than "Export".

Siedlerchr:
DevCall Decision: Move to Tools -> Copy Files

Siedlerchr:
@JabRef/developers  I need some help here. I want to execute the thing in a background thread and make it cancable. Looked already at some other, but it's not clear to me, why some of them are implemented in Frame and some in BasePanel.  And I never worked with this AbstractWorker stuff. Any help welcome

lynyus:
@Siedlerchr how would you describe the status of this PR? Is it feasible to include it into 4.0 or should we postpone it to 4.1?

Siedlerchr:
I would postpone it to 4.1, as we want the 4.0 to be stable

Am 08.08.2017 10:26 vorm. schrieb "Linus Dietz" <notifications@github.com>:

koppor:
This branch is followed up at https://github.com/JabRef/jabref/pull/3147, because it was renamed to `export Pdf`.

koppor:
This branch is followed up at https://github.com/JabRef/jabref/pull/3147, because it was renamed to `exportPdf`.

PR #2711, opened 2017-04-05
Avoid conversion of single underscores

### LINKED ISSUE #2664: Latex2Unicode Formatter: Captial letters after an underscore are wrapped in braces
While checking the problem with the Underscores in the groups from  #2599, I noticed it's a problem of the Latex2Unicode formatter:  

Create a new entry with title `Underscore_Underscore`
Save
In the main table the entry is shown as `Underscore_(U)nderscore`

lenhard:
Fixes #2664

Single underscores that are not followed by `{` are not converted to unicode during the LaTeX to unicode conversion. This is not actually correct according to the rules of LaTeX, but seems to be an important exception required by our users.

This is achieved by the following:
 1. Replace all single underscores with `JABREFUNDERSCORE`
 2. Do the unicode conversion
 3. Replace `JABREFUNDERSCORE` with an underscore

TODO: We should find a better magic word instead of `JABREFUNDERSCORE`, because every character will be processed by latex2unicode and that costs performance. However, it should still be long and unique enough so that we avoid accidental conversions. Suggestions anyone?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - Entries with a single corporate author are now correclty exported to the corresponding `corporate` author field in MS-Office XML. [#1497](https://github.com/JabRef/jabref/issues/1497)
 - Improved author handling in MS-Office Import/Export
 - The `day` part of the biblatex `date` field is now exported to the corresponding `day` field in MS-Office XML. [#2691](https://github.com/JabRef/jabref/issues/2691)
+- Single underscores are not converted to during the LaTeX to unicode conversion, which does not follow the rules of LaTeX, but is what users requre. [#2664](https://github.com/JabRef/jabref/issues/2664)
 
 ### Fixed
  - We fixed an issue of duplicate keys after using a fetcher, e.g., DOI or ISBN [#2867](https://github.com/JabRef/jabref/issues/2687)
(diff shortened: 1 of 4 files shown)


Siedlerchr:
What about the Unicode Replacement Char?
http://www.fileformat.info/info/unicode/char/fffd/index.htm

CHANGELOG.md
```diff
@@ -64,6 +64,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - Entries with a single corporate author are now correclty exported to the corresponding `corporate` author field in MS-Office XML. [#1497](https://github.com/JabRef/jabref/issues/1497)
 - Improved author handling in MS-Office Import/Export
 - The `day` part of the biblatex `date` field is now exported to the corresponding `day` field in MS-Office XML. [#2691](https://github.com/JabRef/jabref/issues/2691)
+- Single underscores are not converted to during the LaTeX to unicode conversion, which does not follow the rules of LaTeX, but is what users requre. [#2664](https://github.com/JabRef/jabref/issues/2664)
```
Siedlerchr: typo: to Latex during the Latex to Unicode cconversion ;)

src/test/java/org/jabref/logic/layout/format/LatexToUnicodeFormatterTest.java
```diff
@@ -156,4 +156,13 @@ public void testApostrophO() {
     public void testApostrophC() {
         assertEquals("O'Connor", formatter.format("O'Connor"));
     }
+
+    @Test
+    public void testCustomUnderscoreConversion() {
+        // our custom version which should preserve the _
+        assertEquals("Lorem ipsum_lorem ipsum", formatter.format("Lorem ipsum_lorem ipsum"));
```
tobiasdiez: Can you please split the test into two tests and convert your comments into nice test names 😸

CHANGELOG.md
lenhard: Done

src/test/java/org/jabref/logic/layout/format/LatexToUnicodeFormatterTest.java
lenhard: Of course I can ;-)

lenhard:
Excellent idea @Siedlerchr I have implemented this using the replacement char (and it still works).

lenhard:
As discussed in the devcall: This can be merged aber fixing checkstyle

Issue #2617, opened 2017-03-06
Improve Architecture Tests


lenhard:
I just noticed that our architecture tests currently do not cover static imports and, therefore, are missing at least one dependency that we want to forbid.

The class `org.jabref.logic.pdf.PdfAnnotationImporter` currently includes this statement `import static org.jabref.gui.importer.actions.OpenDatabaseAction.LOGGER`. Apart from being totally unnecessary (there would be no problem in using an own logger for the class), this is also forbidden. A fix to this issue should:
 1. remove the dependency in `PdfAnnotationImporter` and using an own logger.
 2. enhance the `ArchitectureTests` to detect static imports (which could lead to discovering further errors that might not be as easy to fix)

simonharrer:
Idea: there may be checkstyle rules that can be configured to cover our architecture tests.

lynyus:
Oh, that was me. Fixed this instance in #2618. Unfortunately, I don't have time to investigate 2.

However, I am happy to have raised attention to this problem by committing this mistake.

tobiasdiez:
Further architecture related desirable tests are:
- logic and viewModel: no javafx except javafx.collections
- view, viewModel and controller: no Globals and JabRefGUI, no direct FXDialogService
- all fxml files can be loaded successfully
- no fetcher is used directly outside of the import package

lynyus:
There is a new tool called ArchUnit, maybe this is helpful. https://github.com/TNG/ArchUnit

PR #2719, opened 2017-04-07
Extend architecture tests to javafx and javafx.collections

lenhard:
Extends the architecture tests to also cover `javafx` and `javafx.collections`, as discussed in #2617.

It seems that it is high time for this, since javafx classes are already being misused, as this PR uncovers. Thus, this currently fails the build. The misuses are:
 - org.jabref.logic.exporter.GroupSerializer/GroupSerializerTest/GroupsParserTest uses javafx.scene.paint.Color
 - org.jabref.logic.journals.Abbreviation uses javafx.beans.property.SimpleStringProperty
 - test/org.jabref.logic.l10n.LocalizationParser uses FXMLLoader/PlatformImpl

Although we might think about making exceptions for the beans and tests, I don't see why the Color class should be directly accessed in the gui.


lenhard:
Thanks for your reply:
 - Looking at the Color class, it is just a representation of the RGB code, so not too problematic. I am still not 100% sure. Assuming we would switch to a new GUI technology, would we be using javafx.Color objects, or would we have to refactor the model classes to only use the RGB code string? I would really like to have a third opinion from someone from @JabRef/developers on this, so that we can sort of say that we have a majority consensus.
 - I don't think very highly of our custom event bus (and have always said so), so bean properties would be acceptable to me, especially considering the fact that we allow observable collections, which are essentially just "bean data structures"
 - It might make sense to exclude the whole test sources from the architecture tests. Imho, they are not really relevant for static architecture checks. I will have a look at that.

lenhard:
I have been thinking a little more about `Color`. Isn't this the very same thing as we have in https://github.com/JabRef/jabref/blob/master/src/main/java/org/jabref/gui/IconTheme.java ?

There, we also have all the color codes, but in the gui package and not in model. So what is our reasoning to allow the javafx version of Color in model, but not the awt version? Also, had we used the awt version in model, wouldn't that have meant that we would have had to rewrite all such usages for the javafx migration? If yes, isn't that a good argument for not allowing the javafx version of `Color` in model?

tobiasdiez:
I think the issue with the `Color` class is that previously there was no need to have anything color-related in the model as no class in model relied upon it - thus there was never situation where the usage of `awt.color` in model was on the table. From an architectural standpoint, the cleanest solution is probably to implement our own version of `Color` in model and use this. The gui contains then a converter from this `Color` to whatever color is used in the corresponding UI framework. However, this is additional work and apart from making a migration to a new GUI easier I don't see any advantages. 
@simonharrer as I have our discussion about various architectural points in good memory, I would value your opinion on this matter (if you find the time to look into it).

I can't remember that I was creative enough to use the full path to hide the dependency from our architecture tests. But good to know that this works, I will now definitely (ab)use it and make my life way easier 😄 . Of course I will fix it (tomorrow).

koppor:
@tobiasdiez The fix will go into this PR?

koppor:
* Color: +1 to add it as exception, because it is "just" a data model. I would not implement a separate color class, even though this would be the cleanest solution. (Effort saving)
* Beans: +1 to allow them
* Tests: Yes, please exclude them.

koppor:
Side comment: [Macker](https://innig.net/macker/index.html) seems to be a tool, which also implements architecture checks. Example: https://innig.net/macker/example/layering/src/macker.xml. OK, it is outdated, but I thought, it is nice to know. 😇

lenhard:
I still think that making an exception for `Color` is a bad decision that will come back to bite us. In the course of moving to JavaFX, we are currently losing a lot of independence in our model packages - something that we invested a lot of effort when trying to get rid of the swing dependencies.

But if this is the consensus in the team, I'll bow to that. I created exceptions for the remaining FX classes and this PR should be ready to go when the build succeeds on travis.

PR #2723, opened 2017-04-08
Refactor parsed file field

tobiasdiez:
Inspired by a discussion with @lenhard in #2692. In detail, the following changes were performed:
- Split `FileField` into `FileFieldWriter` and `FileFieldParser`
- Move `getFiles` from `TypedBibEntry` to `BibEntry` (the later already contained a `setFiles` method)
- Add `isOnlineLink` and `findIn` to `ParsedFileField`
- Extract a few methods from `logic.util.io.FileUtil` to `model.util.FileHelper` (idea for a better name?)
- Rename `ParsedFileField` to `LinkedFile`

----


src/main/java/org/jabref/model/util/FileHelper.java
```diff
@@ -0,0 +1,123 @@
...
+        return getFileExtension(file.getName());
+    }
+
+    /**
+     * Returns the extension of a file name or Optional.empty() if the file does not have one (no "." in name).
+     *
+     * @param fileName
+     * @return The extension (without leading dot), trimmed and in lowercase.
+     */
+    public static Optional<String> getFileExtension(String fileName) {
```
koppor: Can you add a JavaDoc comment why you implemented that method on your own instead of using [Files.getFileExtension(fileName);](http://stackoverflow.com/a/15985486/873282)? We have the Guava libraries and should use them 😇

src/main/java/org/jabref/model/entry/FileFieldWriter.java
```diff
@@ -0,0 +1,66 @@
...
+        return getStringRepresentation(Collections.singletonList(field));
+    }
+
+    /**
+     * Encodes a two-dimensional String array into a single string, using ':' and
+     * ';' as separators. The characters ':' and ';' are escaped with '\'.
+     * @param values The String array.
+     * @return The encoded String.
+     */
+    public static String encodeStringArray(String[][] values) {
```
koppor: Can you add a JavaDoc why this method is public? The other below is private, I miss the symmetry.

src/main/java/org/jabref/model/entry/BibEntry.java
```diff
@@ -847,6 +848,21 @@ private void invalidateFieldCache(String fieldName) {
...
+    /**
+     * Gets a list of linked files.
+     *
+     * @return the list of linked files, is never null but can be empty
+     */
+    public List<LinkedFile> getFiles() {
+        //Extract the path
+        Optional<String> oldValue = getField(FieldName.FILE);
+        if (!oldValue.isPresent()) {
+            return new ArrayList<>();
```
koppor: Why not `Collections.emptyList()`?

src/main/java/org/jabref/model/entry/LinkedFile.java
```diff
@@ -1,23 +1,31 @@
...
 import java.util.Objects;
+import java.util.Optional;
 
-public class ParsedFileField {
+import org.jabref.model.database.BibDatabaseContext;
+import org.jabref.model.metadata.FileDirectoryPreferences;
+import org.jabref.model.util.FileHelper;
 
-    private static final ParsedFileField NULL_OBJECT = new ParsedFileField("", "", "");
+public class LinkedFile {
```
koppor: Could you add a JavaDoc stating the intension and usage of this class? Also, why no TypedBibEntry is required anymore?

src/main/java/org/jabref/model/util/FileHelper.java
Siedlerchr: I thik he just moved the method, we previously had it in FileUtils

src/main/java/org/jabref/gui/externalfiles/SynchronizeFileField.java
```diff
@@ -160,8 +153,8 @@ public void run() {
                         boolean deleted = false;
 
                         // Get an absolute path representation:
-                        Optional<File> file = FileUtil.expandFilename(flEntry.getLink(), dirsS);
-                        if ((!file.isPresent()) || !file.get().exists()) {
+                        Optional<Path> file = field.findIn(panel.getBibDatabaseContext(), Globals.prefs.getFileDirectoryPreferences());
+                        if ((file.isPresent()) && Files.exists(file.get())) {
```
Siedlerchr: Why did you change the logic here?` Previously it was ! present

src/main/java/org/jabref/model/util/FileHelper.java
```diff
@@ -0,0 +1,123 @@
...
+     */
+    public static Optional<Path> expandFilename(final BibDatabaseContext databaseContext, String name,
+                                                        FileDirectoryPreferences fileDirectoryPreferences) {
+        Optional<String> extension = getFileExtension(name);
+        // Find the default directory for this field type, if any:
+        List<String> directories = databaseContext.getFileDirectories(extension.orElse(null), fileDirectoryPreferences);
+        // Include the standard "file" directory:
+        List<String> fileDir = databaseContext.getFileDirectories(fileDirectoryPreferences);
+        // Include the directory of the BIB file:
+        List<String> al = new ArrayList<>();
```
Siedlerchr: A better variable name would be nice

src/main/java/org/jabref/model/util/FileHelper.java
koppor: We nevertheless could use Guava 😇

tobiasdiez: Our code is slightly different (extensions are required to have at length > 1) and uses optional instead of empty strings. Thus I think it is not worth using Guava (maybe we can even remove the dependency on guava at some point).

src/main/java/org/jabref/model/entry/FileFieldWriter.java
tobiasdiez: It is public because it is used somewhere else. You are probably right and this method should be private, but I have no idea what "FileListTableModel" is doing...

src/main/java/org/jabref/model/entry/BibEntry.java
tobiasdiez: done

src/main/java/org/jabref/model/entry/LinkedFile.java
tobiasdiez: Of course I can.

src/main/java/org/jabref/gui/externalfiles/SynchronizeFileField.java
tobiasdiez: Nice catch! Was probably a copy&paste error on my side. Fixed

src/main/java/org/jabref/model/util/FileHelper.java
tobiasdiez: Yep

Issue #2682, opened 2017-03-23
Look up DOI improvements


stefan-kolb:
From #2680 

- [x] Lookup DOI action for the button is not in a separate thread (ugly but present long time ago)
- [x] GUI freezes when exception from above is catched.  Maybe related to the start inside the FX thread  9b28d36. Also @lynyus reported that the whole GUI stopped working for him, so maybe this also triggers the freeze problem here and we need to fix the sart in FX thread method somehow.
- [x] Error in CrossRef (500) https://github.com/CrossRef/cayenne/issues/5: This happens randomly!
`http://api.crossref.org/works?query.title=A+break+in+the+clouds%3A+towards+a+cloud+definition&rows=20&offset=0` ->
```
{"status":"error","message-type":"exception","message-version":"1.0.0","message":{"name":"class java.lang.RuntimeException","description":"java.lang.RuntimeException: Solr returned a partial result set","message":"Solr returned a partial result set","cause":null}}
```
- [x] Error in Crossref 503 & 504
maybe log to the console and in the status bar for the user that the service is unavailable.
Follow up #2729

PR #2728, opened 2017-04-12
Move DOI fetching to separate thread #2682

stefan-kolb:
@lynyus Can you check if my code is fine conceptually?


tobiasdiez:
We also have https://github.com/JabRef/jabref/blob/master/src/main/java/org/jabref/JabRefExecutorService.java. 

If you want, I can take care of this issue since I'm currently rewriting the DOI editor component in JavaFX.

PR #2730, opened 2017-04-12
Allow failures for fetcher test

stefan-kolb:
image

As many of @JabRef/developers are annoyed by randomly failing fetcher tests, here's an option to test them but ignore them for the overall build output.

Waiting for your comments if we should do it this way for now.


koppor:
I like it. Better than a cronjob-approach bullying us each Monday. We can add a TOP for "[Topics discussed in each telephone conference](https://github.com/JabRef/jabref/wiki/Minutes#topics-discussed-in-each-telephone-conference)" to discuss the status regularly.

stefan-kolb:
🤣

PR #2731, opened 2017-04-12
Improve date parsing

tobiasdiez:
This PR improves a few small things about the date parsing:
- There were a few places where dates were parsed, for example in `NormalizeDateFormatter`, `BibEntry` and in some importer. All these code snippets are now merged in a new `Date` class. This resulted in more robust code and slightly improved parsing performance.
- The `MonthUtil` class is now an enum with a few static helper methods.  

----


src/main/java/org/jabref/model/entry/Date.java
```diff
@@ -0,0 +1,126 @@
...
+     * Try to parse the following formats
+     *  - "M/y" (covers 9/15, 9/2015, and 09/2015)
+     *  - "MMMM (dd), yyyy" (covers September 1, 2015 and September, 2015)
+     *  - "yyyy-MM-dd" (covers 2009-1-15)
+     *  - "dd-MM-yyyy" (covers 15-1-2009)
+     *  - "d.M.uuuu" (covers 15.1.2015)
+     *  - "uuuu.M.d" (covers 2015.1.15)
+     * The code is essentially taken from http://stackoverflow.com/questions/4024544/how-to-parse-dates-in-multiple-formats-using-simpledateformat.
+     */
+    public static Optional<Date> parse(String dateString) {
```
Siedlerchr: Theoretically we should support all allowed date formats from biblatex: EDTF format
2.3.8 Date and Time Specifications
A while ago I looked into that and found that there is no easy way to support all in java.

src/main/java/org/jabref/model/entry/Month.java
```diff
@@ -0,0 +1,134 @@
...
+import java.util.Optional;
+
+import org.jabref.model.strings.StringUtil;
+
+/**
+ * Represents a month object.
+ */
+public enum Month {
+
+    JANUARY("January", "jan", "01", "#jan#", 0),
```
Siedlerchr: With the JAVA 8 Date Time changes, month now are 1-12 and no longer 0-11

Siedlerchr:
In general LGTM

src/main/java/org/jabref/model/entry/Date.java
lynyus: I have moved this into issue #2753

src/test/resources/org/jabref/logic/importer/fileformat/MsBibImporterTest7.bib
```diff
@@ -5,7 +5,7 @@ @Article{Orlowski;2011
   title        = {Application of Ontology In the ITIL Domain},
   journal      = {Information Systems Architecture and Technology: Service Oriented Networked Systems},
   year         = {2011},
-  month 	   = {mar},
+  month 	   = {03},
```
tobiasdiez: Isn't `month = mar` the correct format for BibTex?!

koppor: It's only, because many styles define the string `mar`. At least this is what I read from the discussion at https://tex.stackexchange.com/q/70455/9075.

Issue #2701, opened 2017-04-01
"Get fulltext" finds too many files when using regexp **/[year]_[auth].*\\.[extension]


sauliusg:
JabRef version JabRef 3.8.2 linux 4.8.0-41-generic amd64 Java 1.8.0_121on Ubuntu 16.04 finds all files that match only the year, not the year_author when the

    **/[year]_[auth].*\\.[extension]

 search pattern is used.

Steps to reproduce:

1. mkdir -p ~/tmp/jabref/pdf; cd ~/tmp/jabref
2. touch pdf/1983_Feringa_509.pdf pdf/1983_Author_123.pdf pdf/1983_unknown.pdf
3. jabref &
4. unpack the attached "bibliography.bib" (from the attached "bibliography.zip", contains just one entry) to the working directory;
5. jabref bibliography.bib
6. Options|Preferences|File: tick "use regular expression search"; set regexp to:

    **/[year]_[auth].*\\.[extension] 

or, alternatively, to:

    **/[year]_[auth]_[firstpage].*\\.[extension]

7. Edit the entry; go to "General"; remove the file; click "Get fulltext"
=> JabRef finds all files that start with pdb/1983_*, although one would expect that if finds only pdf/1983_Feringa_509.pdf (the versions up to JabRef 2.10 behaved as expected).

Screenshots are attached.

The bug could be reproduced also in the e10e62fc8a11f5cea0135593df0900904d34ad35 'master' commit using './gradlew run' and Eclipse (Neon), running the JabRefMain class.

[bibliography.zip](https://github.com/JabRef/jabref/files/887710/bibliography.zip)
![screenshot from 2017-04-01 17-34-47](https://cloud.githubusercontent.com/assets/10668420/24579631/aa596e3c-1701-11e7-8a03-743cfcbfdf34.png)
![screenshot from 2017-04-01 17-35-09](https://cloud.githubusercontent.com/assets/10668420/24579634/afe0c1b6-1701-11e7-94b7-fe3600708204.png)

My file tree configuration:

    saulius@koala jabref/ $ pwd
    /home/saulius/tmp/jabref
    saulius@koala jabref/ $ tree
    .
    ├── bibliography.bib
    ├── bibliography.bib.sav
    ├── bibliography.zip
    └── pdf
            ├── 1983_Author_123.pdf
            ├── 1983_Feringa_509.pdf
            ├── 1983_unknown.pdf
            └── Feringa1983.pdf

PR #2732, opened 2017-04-12
Fix 2701 too may files found

sauliusg:
I would like to offer my fix for the issue  2701: JabRef would find too many files when they are search with the [year]_[auth].* pattern. The reason seems to be a 'null' database passed to the BibtexKeyPatternUtil.makeLabel(BibEntry, String, Character, BibDatabase) method, and a null pointer exception is triggered. As a result, the '[auth]' part of the pattern was erased and ignored. With the suggested fix, JabRef uses the '[auth]' part of the pattern and finds just the files with the matching author, as expected. A relevant unit test was added; no other tests were broken.

- [ ] Change in CHANGELOG.md described -- no functionality changes, restored the previous state
- [x ] Tests created for changes -- 1 unit test added;
- [ ] Screenshots added (for bigger UI changes) -- no GUI changes;
- [ ] Manually tested changed features in running JabRef -- none
- [x ] Check documentation status (Issue created for outdated help page at [help.jabref.org](https://github.com/JabRef/help.jabref.org/issues)?) -- the issue #2701 still open.
- [ ] If you changed the localization: Did you run `gradle localizationUpdate`? -- no localisation changes.

### DIFF
--- a/src/test/java/org/jabref/logic/bibtexkeypattern/MakeLabelWithoutDatabaseTest.java
+++ b/src/test/java/org/jabref/logic/bibtexkeypattern/MakeLabelWithoutDatabaseTest.java
@@ -11,7 +11,6 @@ public class MakeLabelWithoutDatabaseTest {
 
     // private GlobalBibtexKeyPattern pattern;
     private BibEntry entry;
-    private String patternString;
 
     @Before
     public void setUp() {
--- a/src/test/java/org/jabref/logic/bibtexkeypattern/MakeLabelWithoutDatabaseTest.java
+++ b/src/test/java/org/jabref/logic/bibtexkeypattern/MakeLabelWithoutDatabaseTest.java
@@ -1,0 +1,33 @@
+package org.jabref.logic.bibtexkeypattern;
+
+import org.jabref.model.entry.BibEntry;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+public class MakeLabelWithoutDatabaseTest {
+
+    // private GlobalBibtexKeyPattern pattern;
+    private BibEntry entry;
+    private String patternString;
+
+    @Before
+    public void setUp() {
+        entry = new BibEntry();
+        entry.setField("author", "John Doe");
+        entry.setField("year", "2016");
+        entry.setField("title", "An awesome paper on JabRef");
+    }
+
+    @Test
+    public void makeLabelForFileSearch() {
+        String label = BibtexKeyPatternUtil.makeLabel(entry,
+                /*value=*/ "auth",
+                /*keywordDelimiter=*/ ',',
+                /*database=*/ null);
+        assertEquals("Doe", label);
+    }
+
+}
--- a/src/test/java/org/jabref/logic/bibtexkeypattern/MakeLabelWithoutDatabaseTest.java
+++ b/src/test/java/org/jabref/logic/bibtexkeypattern/MakeLabelWithoutDatabaseTest.java
@@ -9,7 +9,6 @@ import static org.junit.Assert.assertEquals;
 
 public class MakeLabelWithoutDatabaseTest {
 
-    // private GlobalBibtexKeyPattern pattern;
     private BibEntry entry;
 
     @Before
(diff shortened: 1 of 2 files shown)


Siedlerchr:
Thank you very much for your contribution 👍 . Codewise it looks very good to me. If the test passes and another dev gives his okay, we can merge it.

lynyus:
Thanks for the contribution @sauliusg . Great, when users find an issue and even fix it themselves (;

> The reason seems to be a 'null' database passed to the BibtexKeyPatternUtil.makeLabel(BibEntry, String, Character, BibDatabase) method, and a null pointer exception is triggered

I will investigate this tomorrow, because I'm not entirely sure if your fix solves the symptoms or we should dig a bit deeper and eliminate the main problem, which could be that bloody 'null' database.

sauliusg:
Dear  Siedlerchr, dear Lynyus,

thank you for the instant reply!

I have got feedback from Codacy and fixed: removed an unused variable from the new unit test. Also, I added another second unit test to cover the second 'if' statement added by me.

> I will investigate this tomorrow, because I'm not entirely sure if your fix solves the symptoms or we 
> should dig a bit deeper and eliminate the main problem, which could be that bloody 'null' database.

Please have a look. It seems to me that the 'null' can be traced to the explicit 'null' argument passed to it.

Regards,
Saulius

tobiasdiez:
@lynyus your suggestions sounds reasonable. We should try to not pass null as an argument. Moreover, the label generation code probably needs a refactoring (especially, get rid of the huge amount of static methods). But this will be probably a bigger project. While I would welcome PRs from you or @sauliusg in this direction, I think the current PR is an acceptable workaround within the boundaries of the current code infrastructure.

src/test/java/org/jabref/logic/bibtexkeypattern/MakeLabelWithoutDatabaseTest.java
```diff
@@ -0,0 +1,46 @@
...
+        entry = new BibEntry();
+        entry.setField("author", "John Doe");
+        entry.setField("year", "2016");
+        entry.setField("title", "An awesome paper on JabRef");
+    }
+
+    @Test
+    public void makeLabelForFileSearch() {
+        String label = BibtexKeyPatternUtil.makeLabel(entry,
+                /*value=*/ "auth",
```
tobiasdiez: I think we never used comments to indicate the name of the argument and thus would prefer that you remove them.

src/test/java/org/jabref/logic/bibtexkeypattern/MakeLabelWithoutDatabaseTest.java
```diff
@@ -0,0 +1,46 @@
...
+import org.jabref.model.entry.BibEntry;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+public class MakeLabelWithoutDatabaseTest {
+
+    // private GlobalBibtexKeyPattern pattern;
```
tobiasdiez: Use remove the unused variable completely.

sauliusg: Agreed, this var will probably never be needed.

src/test/java/org/jabref/logic/bibtexkeypattern/MakeLabelWithoutDatabaseTest.java
sauliusg: Hmmm, indicating “Named Function Parameter” Comments is a recommended practice from the Dustin Boswell, Trevor Foucher "The Art of Readable Code", for languages that do not support named parameters, and I personally find such code easier to read, with not so much overhead. But if this is not used on JabRef and you do not find them useful, then, sure, please just remove them. Should I make the requested changes and recommit, or will you andle these issues?

lynyus: Yes please, IntelliJ, for example automatically displays the parameter names. 
parameter_names

sauliusg: Both issues fixed, please find a new push.

Siedlerchr:
Okay, lgtm! I merge it in! Thank you very much for your contribution! 👍  Feel free to take a look at other issues

sauliusg:
Thanks!

PR #2752, opened 2017-04-16
Add test for ordinals

koppor:
This is a follow up for https://github.com/JabRef/jabref/issues/2596 and https://github.com/JabRef/jabref/pull/2601. I'd like to test how I can reproduce the behavior of ordinals.

Can someone explain me, why I don't get Unicode as result here?

### DIFF
--- a/src/main/java/org/jabref/model/strings/LatexToUnicodeAdapter.java
+++ b/src/main/java/org/jabref/model/strings/LatexToUnicodeAdapter.java
@@ -21,7 +21,7 @@ public class LatexToUnicodeAdapter {
         Objects.requireNonNull(inField);
 
         String toFormat = underscoreMatcher.matcher(inField).replaceAll(replacementChar);
-        toFormat = Normalizer.normalize(LaTeX2Unicode.convert(toFormat), Normalizer.Form.NFKC);
+        toFormat = Normalizer.normalize(LaTeX2Unicode.convert(toFormat), Normalizer.Form.NFC);
         return underscorePlaceholderMatcher.matcher(toFormat).replaceAll("_");
     }
 }
--- a/src/test/java/org/jabref/logic/layout/format/LatexToUnicodeFormatterTest.java
+++ b/src/test/java/org/jabref/logic/layout/format/LatexToUnicodeFormatterTest.java
@@ -29,7 +29,7 @@ public class LatexToUnicodeFormatterTest {
     @Test
     public void testFormatTextit() {
         // See #1464
-        assertEquals("text", formatter.format("\\textit{text}"));
+        assertEquals("\uD835\uDC61\uD835\uDC52\uD835\uDC65\uD835\uDC61", formatter.format("\\textit{text}"));
     }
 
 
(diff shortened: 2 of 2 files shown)


lynyus:
I have fixed the ordinals, with the side-effect that e.g., italics are now also converted into unicode. What do you think about that?

koppor:
Refs https://github.com/JabRef/jabref/issues/160

I think, I begin to like it, because the display of JabRef in the UI will improve. As stated in https://github.com/JabRef/jabref/pull/2601#issuecomment-294490145, I will step away from biblatex in papers and so, we can move on here.

lenhard:
As indicated in #2784, this breaks support for italics in the main table, which was working before. We should revert it and find another way to support ordinals.

lynyus:
Hi @lenhard, the problem is you always loose something.
The key for the conversions lies in the Normalizer settings of the [LaTexToUnicodeConverter#24](https://github.com/JabRef/jabref/blob/master/src/main/java/org/jabref/model/strings/LatexToUnicodeAdapter.java#L24)

I have made this handy chart of the four options and what works when (on my linux system).

|      | Ordinals | textit | Special Chars, eg. polish names, acute, umlauts |
|------|----------|--------|-------------------------------------------|
| NFD  |  Yes      |   Yes     |             No                              |   
| NKC  |     Yes     |   Yes     |            Yes                               |   
| NFKD |     No     |    No    |              No                             |   
| NFKC |     No     |     No   |               Yes                            |   

In conclusion I would suggest to investigate what factors influence the bug that so far was only seen on @AEgit's system,

AEgit:
@lynyus : Let me know which additional configuration details are needed. I still encounter the bug in:
JabRef 4.0.0-dev--snapshot--2017-07-10--master--e80f450ee
Windows 10 10.0 amd64 
Java 1.8.0_131

lynyus:
@AEgit the problem is: I have no idea :/

lynyus:
One solution might be to convert italics back into regular ascii after the latex2unicode conversion?

AEgit:
Hmmm, ok - am I really the only one with a Windows system, who encounters this problem? Because if that's the case, it might be my/my system's fault. But as far as I can tell Siedlerchr confirmed the issue on his Windows 10 installation (https://github.com/JabRef/jabref/issues/2784). Therefore it sounds like a Windows issue. What about the Mac users?

lynyus:
Okay, I'm on my Windows now and I have these problems. Somehow I must have overread @Siedlerchr's comment or misinterpreted it :/ All the time I was assuming that I had checked it successfully on my windows machine.

Mea culpa, I'll be working on a fix now.

AEgit:
No worries, and thank you for your help!

lenhard:
@lynyus Nice table, thanks for the overview :) Sure, NKC is the right option here. The problem is that we're using a Swing table as maintable and as far as I've found it just refuses to work with unicode italics (unless you have appropriate fonts installed in your system, but thats not the default). The best solution would be to get rid of the maintable in our move to JavaFX. Let's do that in the future, but let's not do it right now, because we should rather build a stable entry editor.

I'll suggest something different: @AEgit How important is it really that the text in the main table is in italics? Would it be acceptable if the text would just be displayed as plain text? Because in that case, we could just eliminate the `\textit` command before we do the unicode conversion.

AEgit:
@lenhard To be honest, I'm not really bothered by the fact that the text in the main table is not in italics. As long as the entry preview shows it in italics (which it currently does), I'm pretty happy with it. Having the proper format in the main table as well, would be nice, but it is not a show-stopper for me at the moment.

lynyus:
I have a PR in preparation, hang on

PR #2754, opened 2017-04-17
Entry editor goes JavaFX

tobiasdiez:
So now that we have released the beta version, we can go back to the alpha state with this PR 😄.
I actually just wanted to have a look at the autocompletion and use some built-in JavaFX functionality for it. However, just changing the JTextarea's to their JavaFX equivalent wasn't possible since the editor components (additional buttons next to the text area) required JTextareas.
Thus, eating the foul egg (= Easter version of "in den sauren Apfel beißen"), I started to migrate the whole entry editor to JavaFX. So far only the DOI component is re-implemented.

Points to notice:
- Using JavaFX bindings, there is no longer the need to manually update the entry after a changed happened. This should fix these nasty bugs that sometimes changes in the editor got lost when performing a different action. Moreover, the main table is instantly updated with the changes in the editor.
- I opted for a slightly different architecture. Previously, there were the text field and additional components. Now these are combined into one FieldEditor.
- As for the groups interface, I will continue slowly to reimplement the other field editors.
- The style is not that different from the Swing version. For the buttons, I only used images:
snipimage

------


src/test/java/org/jabref/SearchQueryHighlightListenerTest.java
```diff
@@ -34,6 +26,7 @@ public void setUp() {
         when(loader.getProtectedTermsLists()).thenReturn(Collections.emptyList());
     }
 
+    /*
```
koppor: Why not ignoring the whole test class?

src/main/java/org/jabref/gui/util/BackgroundTask.java
```diff
@@ -27,30 +28,51 @@ private BackgroundTask(Callable<V> callable) {
...
+    /**
+     * Sets the {@link Consumer} that is invoked after the task is successfully finished.
+     */
     public BackgroundTask<V> onSuccess(Consumer<V> onSuccess) {
         this.onSuccess = onSuccess;
         return this;
     }
 
-    public V call() throws Exception {
+    V call() throws Exception {
```
koppor: I wonder why some methods became package-private

Siedlerchr: Maybe some cleanup action on saving?

tobiasdiez: These methods should not be called directly, but only by the TaskExecutor.

src/test/java/org/jabref/SearchQueryHighlightListenerTest.java
tobiasdiez: The problem is these tests don't compile anymore (TextArea no longer gives access to the highlighter). I marked it now as a todo for later.

stefan-kolb:
@JabRef/developers Again, I really like what you are doing!, however this really holds us back from releasing a major 4.0 version for a long?! time again.

PR #2758, opened 2017-04-18
Updated Spanish Translation

jtornero:
Updated menu and properties file.


Siedlerchr:
Thank you very much for your contribution, but unfortunately you seem to have some duplicated and obsolete keys which you must remove from your file:
`Export_with_selected_format, Remove_entry_selection_from_this_group`

And these ones are duplicated: `Remove_entry_from_import, Donate_to_JabRef, From_import, Discard_changes, Filled, Field_is_missing` which have to be removed.

it would be nice, if you could fix this. If you need help, then we can help you.

koppor:
Thank you for the large update!

Could you eventually also check all strings containing "library". Pre-4.0 used to use "database", which we decided to change to "library". Can you adapt these things, too? (Refs https://github.com/JabRef/jabref/issues/2095)

jtornero:
@koppor @saivert Well, I've revisited the files and I've updated all the strings. As always, I have some problems while fighting with git :) , so maybe It's better to attach the files to this message. 
[JabRef_es.translation.zip](https://github.com/JabRef/jabref/files/942185/JabRef_es.translation.zip)

By the way.. I though that new strings to be translated appeared appended to the file.. it looks like it isn't this way, so I have translated many _hidden_ strings. Next time I'll look for untranslated strings in all the file.

Cheers

jorge

Siedlerchr:
@jtornero Thanks again for the contribution. I took care of the git stuff in #2773 
Please have a look there
 To find untranslated strings you can use a regular expression search like  `=\r\n (or =\n)` in a text editor

jtornero:
@Siedlerchr Yes, that's the way I use to do it. Thanks.

PR #2759, opened 2017-04-18
More JavaFX editors

tobiasdiez:
A few more field editors are reimplemented in JavaFX.
I will fix the failing language test later.


Siedlerchr:
Looks okay to me. Would be nice if you could add some screenshots

stefan-kolb:
@JabRef/developers Whereas I like the work that is done here, it is a step back from the current UI. Some functionality is missing other functionality is not as accessible as before. We should have discussed this at least shouldn't we?

Siedlerchr:
@stefan-kolb  I think the buttons are now setup as icons here.

lenhard:
I agree with Stefan. It seems that the crossref, file, and timestamp field editors are missing certain functions.

Also if the text on the buttons is gone, then we at least need the tool tips.

tobiasdiez:
@stefan-kolb as outlined in the previous PR, I will convert each editor peu à peu and in the end the full functionality should be restored. For the text buttons vs icons, we should probably discuss this in the devcall.

PR #2763, opened 2017-04-19
Fix 'library' vs 'database' Chinese translation.

solrex:
- [ ] Fix 'library' vs 'database' Chinese translation in `JabRef_zh.properties` and `Menu_zh.properties`


Siedlerchr:
Thank you very much for your contribution! 👍

Issue #2760, opened 2017-04-19
Drag'n'drop of a file to the main table does not work any more


koppor:
I used to drop a file into the main table to open it. Now, I get

    02:52:00.742 [AWT-EventQueue-0] ERROR org.jabref.gui.groups.EntryTableTransferHandler - Drop type error
    java.lang.ClassCastException: java.io.File cannot be cast to java.nio.file.Path

<details>
02:52:00.742 [AWT-EventQueue-0] ERROR org.jabref.gui.groups.EntryTableTransferHandler - Drop type error
java.lang.ClassCastException: java.io.File cannot be cast to java.nio.file.Path
        at org.jabref.gui.groups.EntryTableTransferHandler.handleDraggedFiles(EntryTableTransferHandler.java:296) ~[main/:?]
        at org.jabref.gui.groups.EntryTableTransferHandler.importData(EntryTableTransferHandler.java:122) [main/:?]
        at javax.swing.TransferHandler.importData(TransferHandler.java:827) [?:1.8.0_121]
        at javax.swing.TransferHandler$DropHandler.drop(TransferHandler.java:1544) [?:1.8.0_121]
        at java.awt.dnd.DropTarget.drop(DropTarget.java:455) [?:1.8.0_121]
        at javax.swing.TransferHandler$SwingDropTarget.drop(TransferHandler.java:1282) [?:1.8.0_121]
        at sun.awt.dnd.SunDropTargetContextPeer.processDropMessage(SunDropTargetContextPeer.java:538) [?:1.8.0_121]
        at sun.awt.dnd.SunDropTargetContextPeer$EventDispatcher.dispatchDropEvent(SunDropTargetContextPeer.java:852) [?:1.8.0_121]
        at sun.awt.dnd.SunDropTargetContextPeer$EventDispatcher.dispatchEvent(SunDropTargetContextPeer.java:776) [?:1.8.0_121]
        at sun.awt.dnd.SunDropTargetEvent.dispatch(SunDropTargetEvent.java:48) [?:1.8.0_121]
        at java.awt.Component.dispatchEventImpl(Component.java:4744) [?:1.8.0_121]
        at java.awt.Container.dispatchEventImpl(Container.java:2294) [?:1.8.0_121]
        at java.awt.Component.dispatchEvent(Component.java:4711) [?:1.8.0_121]
        at java.awt.LightweightDispatcher.retargetMouseEvent(Container.java:4888) [?:1.8.0_121]
[...]

PR #2764, opened 2017-04-19
Fix Drag and drop of file by first converting to File and then to path

Siedlerchr:
Fix for #2760

Internal change only. Was introduced by fa6e0740fdc66afe05f0b54a507f58198f6e2000


Siedlerchr:
I don't think is possible as this is some internal java definition in the
background which all work on File instead path
Otherwise the cast would have been Successful at first

Am 19.04.2017 3:54 nachm. schrieb "Tobias Diez" <notifications@github.com>:

PR #2771, opened 2017-04-20
Fix-for-issue-2760

LouisCochinho:
Fix drag and drop of file by first converting to File and then to Path


lynyus:
Thank you for your contribution! It is now the case that we have two solutions for this problem, yours and https://github.com/JabRef/jabref/pull/2764.

Siedlerchr:
Thank you for your interest in contribution. Unfortunately for you I fixed that particular bug yesterday.  ;)
But there are plenty of more things you can do if you look at the issues. If you need help at any point, feel free to ask

Issue #2777, opened 2017-04-21
Illegal file annotation type


psychelzh:
JabRef 4.0.0-beta
Windows 10 10.0 amd64 
Java 1.8.0_131

There happened to strange behavior when click on some group.

![image](https://cloud.githubusercontent.com/assets/13306020/25273728/b121e15e-26bf-11e7-87b2-231650d0bb98.png)

Paste your log details here.

<details>
<summary>Detail information:</summary>

```
Opening: D:\git\JabRefBibs\MyFullResearch.bib
Unable to parse stored date for field 'date' with current settings. Clear button in calender popup will not work.
Uncaught exception Occurred in Thread[AWT-EventQueue-0,6,main]
com.google.common.util.concurrent.UncheckedExecutionException: java.lang.IllegalArgumentException: No enum constant org.jabref.model.pdf.FileAnnotationType.INK
	at com.google.common.cache.LocalCache$Segment.get(LocalCache.java:2217)
	at com.google.common.cache.LocalCache.get(LocalCache.java:4154)
	at com.google.common.cache.LocalCache.getOrLoad(LocalCache.java:4158)
	at com.google.common.cache.LocalCache$LocalLoadingCache.get(LocalCache.java:5147)
	at com.google.common.cache.LocalCache$LocalLoadingCache.getUnchecked(LocalCache.java:5153)
	at org.jabref.logic.pdf.FileAnnotationCache.getFromCache(FileAnnotationCache.java:43)
	at org.jabref.gui.entryeditor.FileAnnotationTab.addAnnotations(FileAnnotationTab.java:110)
	at org.jabref.gui.entryeditor.FileAnnotationTab.initializeTab(FileAnnotationTab.java:89)
	at org.jabref.gui.entryeditor.EntryEditor$TabListener.lambda$stateChanged$0(EntryEditor.java:1163)
	at java.awt.event.InvocationEvent.dispatch(Unknown Source)
	at java.awt.EventQueue.dispatchEventImpl(Unknown Source)
	at java.awt.EventQueue.access$500(Unknown Source)
	at java.awt.EventQueue$3.run(Unknown Source)
	at java.awt.EventQueue$3.run(Unknown Source)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(Unknown Source)
	at java.awt.EventQueue.dispatchEvent(Unknown Source)
[...]

PR #2778, opened 2017-04-21
Fixes handling of unknown PDAnnotation types.

lynyus:
refs. #2777


src/main/java/org/jabref/model/pdf/FileAnnotationType.java
```diff
@@ -12,14 +22,33 @@
...
     }
 
+    /**
+     * Determines the FileAnnotationType of a raw PDAnnotation. Returns 'UNKNOWN' if the type is currently not in our
+     * list of FileAnnotationTypes.
+     *
+     * @param annotation the raw PDAnnotation
+     * @return The determined FileAnnotationType
+     */
+    public static FileAnnotationType parseFileAnnotationType(PDAnnotation annotation) {
```
tobiasdiez: I think we normally called such methods simply `parse` since the rest should be clear from the class name. But you can leave it also like that.

koppor:
Please change method name, resolve the conflict and then merge.

src/main/java/org/jabref/model/pdf/FileAnnotationType.java
```diff
@@ -12,14 +22,33 @@
     FREETEXT("FreeText"),
     STRIKEOUT("Strikeout"),
     LINK("Link"),
+    UNKNOWN("Unknown"),
     NONE("None");
 
     private final String name;
+    private static final Log LOGGER = LogFactory.getLog(FileAnnotationType.class);
```
stefan-kolb: Please put the logger attribute directly under the enum declaration.

src/main/java/org/jabref/model/pdf/FileAnnotationType.java
```diff
@@ -12,14 +22,33 @@
...
+
```
stefan-kolb: Obsolete line?!

PR #2794, opened 2017-04-25
Fixes #2789 Add Referer to API call

### LINKED ISSUE #2789: IEEEXplore fetcher seems to be broken
JabRef version 4.0.0-beta on Windows 10

Steps to reproduce:

1. Perform a web search using the "IEEEXplore" fetcher
2. result is always "Invalid search term"

no more logs shown...

stefan-kolb:
See #2789.
> JabRef version 4.0.0-beta on Windows 10
>Steps to reproduce:
>Perform a web search using the "IEEEXplore" fetcher
>result is always "Invalid search term"

Please check if this works for you.

### DIFF
--- a/src/main/java/org/jabref/gui/importer/fetcher/IEEEXploreFetcher.java
+++ b/src/main/java/org/jabref/gui/importer/fetcher/IEEEXploreFetcher.java
@@ -105,6 +105,7 @@ public class IEEEXploreFetcher implements EntryFetcher {
             //add request header
             dl.addHeader("Accept", "application/json");
             dl.addHeader("Content-Type", "application/json");
+            dl.addHeader("Referer", "http://ieeexplore.ieee.org/search/searchresult.jsp");
 
             // set post data
             dl.setPostData(postData);


Siedlerchr:
Could you please add a test? The IEEE-Tests we already have do not cover this fulltext finder

stefan-kolb:
@Siedlerchr Unfortunately it has a lot of GUI and Preferences interdependencies. As it seems we do not have tests for any of those query fetchers yet?!

matthiasgeiger:
Of course we have some test for those fetchers - but only for those which have been rewritten to the new structure, e.g., https://github.com/JabRef/jabref/blob/master/src/test/java/org/jabref/logic/importer/fetcher/DBLPFetcherTest.java which tests https://github.com/JabRef/jabref/blob/master/src/main/java/org/jabref/logic/importer/fetcher/DBLPFetcher.java

stefan-kolb:
Yes, I was talking about the old ones!

koppor:
Is there an Issue for rewriting those; tagged with "beginner"? 😇

Am 26.04.2017 12:14 schrieb "Stefan Kolb" <notifications@github.com>:

Yes, I was talking about the old ones!

—
You are receiving this because you are subscribed to this thread.
<https://github.com/JabRef/jabref/pull/2794#issuecomment-297335649>, or mute
the thread

stefan-kolb:
Tried to add a test, but got stuck with so many dependencies and calls on different Preferences classes :(

PR #2798, opened 2017-04-26
Rework file editor in JavaFX

tobiasdiez:
The file editor now looks like this:
look
Moreover, I changed the behavior slightly to reduce the amount of popuped dialogs:
- "Add new file" directly adds the linked file (and guesses the file type) instead of showing the "edit linked file" dialog
- "Download from URL" adds the download to the file list and showing a progress bar:
filedownloadinprogress
- Files that are found by the "Auto link" feature are automatically added to the files list (but not serialized in the files field) and can be accepted (i.e. serialized) by clicking the accept button:
fileautofound

The code was ugly and is still not very nice (global dependencies everywhere). Maybe I will find the time later to refactor the code even more.
Some features are still missing (e.g. right-click menu).

----


src/main/java/org/jabref/gui/fieldeditors/LinkedFilesEditorViewModel.java
```diff
@@ -0,0 +1,286 @@
...
+        return files.get();
+    }
+
+    public ListProperty<LinkedFileViewModel> filesProperty() {
+        return files;
+    }
+
+    public void addNewFile() {
+        List<Path> fileDirectories = databaseContext.getFileDirectoriesAsPaths(Globals.prefs.getFileDirectoryPreferences());
+        Path workingDirectory = fileDirectories.stream()
```
Siedlerchr: There is a method I once crated: getFirstFileDirectory (or similar) which exactly does this. I think it is in databaseContext, too

src/main/java/org/jabref/logic/util/io/FileUtil.java
```diff
@@ -168,99 +163,26 @@ public static boolean renameFile(Path fromFile, Path toFile, boolean replaceExis
...
         }
 
-        for (String dir : dirs) {
-            if (dir != null) {
-                File result = shortenFileName(fileName, dir);
-                if ((result != null) && !result.equals(fileName)) {
-                    return result;
-                }
+        for (Path dir : dirs) {
+            if (file.startsWith(dir)) {
```
Siedlerchr: Be careful about this, you probably want to use toString.startswith dir.toString
http://stackoverflow.com/questions/30157072/why-is-path-startswith-behaviour-different-from-a-string-startswith-operatio

src/main/java/org/jabref/model/database/BibDatabaseContext.java
```diff
@@ -154,8 +163,7 @@ public boolean isBiblatexMode() {
      * @return Optional of Path
      */
     public Optional<Path> getFirstExistingFileDir(FileDirectoryPreferences preferences) {
-        return getFileDirectories(preferences).stream().filter(s -> !s.isEmpty()).map(p -> Paths.get(p))
-                .filter(Files::exists).findFirst();
+        return getFileDirectoriesAsPaths(preferences).stream().filter(Files::exists).findFirst();
         //Filter for empty string, as this would be expanded to the jar-directory with Paths.get()
```
Siedlerchr: Add this comment please to the method above where you check for the isEmpty

src/main/java/org/jabref/model/database/BibDatabaseContext.java
```diff
@@ -154,8 +163,7 @@ public boolean isBiblatexMode() {
      * @return Optional of Path
      */
     public Optional<Path> getFirstExistingFileDir(FileDirectoryPreferences preferences) {
```
Siedlerchr: This is the method I was refereing to earlier

src/main/java/org/jabref/gui/fieldeditors/LinkedFilesEditorViewModel.java
Siedlerchr: getFirstExistingFileDirectory is the name of the method

tobiasdiez: Thanks, I changed the code to use the other method now.

src/main/java/org/jabref/logic/util/io/FileUtil.java
tobiasdiez: I really want to check that the `file` is a subpath of `dir`, so I think this is fine.

tobiasdiez:
I merged this PR now so that I can continue working on the editor stuff. Further feedback will of course be incorporated.

PR #2813, opened 2017-05-01
Editor tab goes JavaFX

tobiasdiez:
The whole entry tab is now converted to use JavaFX controls instead of Swing and I did a bit of restyling.
Before:
image
After:
image


lenhard:
Just a thought: Can we wait with merging that until after the final v4.0 release?

I really like your work and am happy with the effort you put into updating the GUI, but I think that it is time for a regular release and there I would rather like to see something stable and tested. There might be some hidden problems in the new version of the entry editor that we only find out after a longer period of actually using it.

lenhard:
Ok, in this case, I'd like to see some small improvements before merging. Here is how it looks like for me atm:
jabref

Could you make sure that:
 - the field names are not cut in the entry editor?
 - there is at least one point of space between the sidebar right next to the entry editor and the field names? I mean a whitespace between the sidebar and, for example, the text "Volume". Otherwise it looks as if the text is glued to the sidebar
 - the icons you use for the buttons in the entry editor follow the color scheme of JabRef? Right now, they are all just black.

koppor:
Maybe we should do a "release candidate"? -> JabRef 4.0-rc1 😇

tobiasdiez:
@lenhard thanks for the feedback. I now fixed all the points you mentioned except the color. The problem is that currently the color is controlled via preferences while JavaFX prefers the color scheme as a css file (the JavaFX colors are defined in [Main.css](https://github.com/JabRef/jabref/blob/master/src/main/java/org/jabref/gui/Main.css)). Thus the easiest solution is to wait until most of the swing stuff is replaced and then allow the user to configure the look by editing the css file.

lenhard:
Then we need a pretty damn good way of modifying the css file via the JabRef gui ;-) What does our "color expert" say to all of this? @matthiasgeiger

Siedlerchr:
At least during development or however, there is a cool app that allows you to change the css at runtime: https://github.com/McFoggy/cssfx

Issue #2805, opened 2017-04-28
ui wording: related articles


ilippert:
Following http://discourse.jabref.org/t/integrating-related-articles-from-mrdlib-into-jabref/240/37
I propose to adapt the wording in the preferences: "show recommendations" should be "show related articles"

lenhard:
This has been done in #2819 and the issue can be closed.

PR #2819, opened 2017-05-04
Revised the localization for 'show_recommendations'

lynyus:
This is one solution to https://github.com/JabRef/jabref/issues/2805.
* Concerning the wording I suggest using the exact technical term 'tab' for referring to this part of the UI. I think tabs are quite ubiquitous, especially in today's browsers. However, I acknowledge concerns that it might be harder to translate.

I'm happy to hear what you think!


stefan-kolb:
I'm ok with that. It's funny tho that the term 'related' does not really fit the actual results of Mr.Dlib right now ;)

src/main/resources/l10n/JabRef_de.properties
```diff
@@ -2302,7 +2302,7 @@ Do_you_want_to_recover_the_library_from_the_backup_file?=Möchten_Sie_die_Biblio
 Firstname_Lastname=Vorname_Nachname
 
 Recommended_for_%0=Empfohlen_für_%0
-Show_recommendations=Empfehlungen_anzeigen
+Show_'Related_Articles'_tab=Reiter_für_Empfehlungen_anzeigen
```
koppor: "Reiter" nur i.d. Deutschen Übersetzung? In den anderen? 😱

Issue #2825, opened 2017-05-06
Using right-click->attch file does not open the file dialog in the selected directory


bernhard-kleine:
JabRef 4.0.0-dev--snapshot--2017-05-06--master--98e2e4379
Windows 7 6.1 amd64 
Java 1.8.0_131

Steps to reproduce:

1. right-click any entry
2. choose attach-file
3. a file dialog which displays a window with files in the selected directory, does not open

The error in previous snapshot version that clicking the + in edit window -> General does not open the window is gone. One can use the dialog there to link files, once again.

```
Put the excerpt of the log file here
```

Siedlerchr:
Can reproduce this.
```
java.lang.NullPointerException: null
	at org.jabref.gui.filelist.FileListEntryEditor.lambda$0(FileListEntryEditor.java:105) ~[bin/:?]
	at org.jabref.gui.filelist.FileListEntryEditor$3.lambda$0(FileListEntryEditor.java:247) ~[bin/:?]
	at java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:311) ~[?:1.8.0_131]
	at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:756) ~[?:1.8.0_131]
	at java.awt.EventQueue.access$500(EventQueue.java:97) ~[?:1.8.0_131]
	at java.awt.EventQueue$3.run(EventQueue.java:709) ~[?:1.8.0_131]
	at java.awt.EventQueue$3.run(EventQueue.java:703) ~[?:1.8.0_131]
	at java.security.AccessController.doPrivileged(Native Method) ~[?:1.8.0_131]
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:80) ~[?:1.8.0_131]
	at java.awt.EventQueue.dispatchEvent(EventQueue.java:726) ~[?:1.8.0_131]
	at java.awt.EventDispatchThread.pumpOneEventForFilters(EventDispatchThread.java:201) [?:1.8.0_131]
	at java.awt.EventDispatchThread.pumpEventsForFilter(EventDispatchThread.java:116) [?:1.8.0_131]
	at java.awt.EventDispatchThread.pumpEventsForFilter(EventDispatchThread.java:109) [?:1.8.0_131]
	at java.awt.WaitDispatchSupport$2.run(WaitDispatchSupport.java:184) [?:1.8.0_131]
[...]

PR #2826, opened 2017-05-06
Fix invalid folder when passed file/path is empty string

Siedlerchr:
Return the correct working dir
Fix for #2825 

Internal/4.0.0 beta change only 

~~- [ ] Change in CHANGELOG.md described~~
~~- [ ] Tests created for changes~~
~~- [ ] Check documentation status (Issue created for outdated help page at [help.jabref.org](https://github.com/JabRef/help.jabref.org/issues)?)~~


src/main/java/org/jabref/model/util/FileHelper.java
```diff
@@ -115,7 +116,9 @@
         Objects.requireNonNull(directory);
 
         Path file = Paths.get(filename);
-        if (Files.exists(file)) {
+        //Explicitly check for an empty String, as File.exists returns true on that empty path, because it maps to the default jar location
```
tobiasdiez: I think it would be clearer if you just add a 
````
if (filename.isEmpty())
    return directory;
````
statement.

Siedlerchr: Will do that.
The failing tests are still somehow weird as locally the run through partly.. I still think there's something with the prefs which has some dubious side effects

Siedlerchr:
I found the problem. It was not due to preferences, but due to an empty filename which would create a folder outside the junit tmp dir and would stay persistent.

PR #2264, opened 2016-11-13
[WIP] Enable PDF fulltext search

koppor:
This PR enables a fulltext search of all linked PDFs.

It is an initial implementation skeleton, where work can be started from.

Issue #2263, opened 2016-11-13
Advanced PDF features


JingyunLiang:
JabRef version 3.6 on window8

more features:

1. ...add the global search from PDFs
2. ...add an inner PDF reader, which can be used to read papers easily and make notes
3. ...make adding remarks easier

```
Put the excerpt of the log file here
```

koppor:
This feature request provides highlights given at https://github.com/JabRef/jabref/wiki/FeatureRequests-Sorted#better-management-of-my-linked-files-please and https://github.com/JabRef/jabref/wiki/FeatureRequests-Sorted#make-a-better-use-of-my-linked-files-please

The search within PDFs is started at https://github.com/JabRef/jabref/pull/2264. We need help to finish that features.

A PDF reader was included until version 3.1 of JabRef. [It was removed in version 3.2](https://github.com/JabRef/jabref/blob/master/CHANGELOG.md#removed-3), because it was very ugly and we will never achieve that the inner PDF pdf reader is as good as a professional PDF reader such as SumatraPDF or Acrobat Reader. Therefore, we focus our energy to a good improvement of these readers. See #1883

Regarding the remarks: Do you talk about PDF remarks or about "Review" comments, which should go into the "Review" tab?

In case you want to help us improving JabRef, we would like to point you to our [CONTRIBUTING.md](https://github.com/JabRef/jabref/blob/master/CONTRIBUTING.md). You can also contact us via [slack](http://jabref.slack.com/), email or via the forum if you need support.

Siedlerchr:
We could include the pdfjs lib which is used on Firefox

koppor:
It has difficulties in handling comments. For instance, highlight text, put a comment on it and open the PDF in pdfjs. The comment on the text is not displayed.

@Joeran (Docear) also wrote some thoughts on integrating a PDFViewer: http://www.docear.org/2012/05/22/dreaming-of-an-integrated-java-pdf-viewer-for-docear/

A feature comparison of current PDF readers is provided at http://www.docear.org/support/user-manual/#PDF_Editors

Joeran:
we tried for a while to integrate some pdf viewer into docear, but we failed, i.e. every library we looked at had some serious problems or did run properly with docear. of course, that was 4 years ago and lots of things might have changed since then. anyway, i generally would advise to let users use their own preferred pdf reader and try to make jabref work with those other PDF readers. 

[...]

PR #2838, opened 2017-05-12
Search in PDF Files

LinusDietz:
This implements a fulltext-search feature for linked PDF files.

search

All linked pdf files are indexed using [Apache Lucene](https://lucene.apache.org/). Search-results are displayed in a google-esque fashion in an additional tab in the entry editor (visible only if a fulltext search is active).
Users can enable/disable fulltext-search with a toggle-button in the search-bar.


build.gradle
```diff
@@ -308,26 +314,26 @@ shadowJar {
 
         // this is an adapter required for generating a fat jar with correct log4j2 output
 
-        com.github.edwgiz.mavenShadePlugin.log4j2CacheTransformer.PluginsCacheFileTransformer target = new com.github.edwgiz.mavenShadePlugin.log4j2CacheTransformer.PluginsCacheFileTransformer();
+        com.github.edwgiz.mavenShadePlugin.log4j2CacheTransformer.PluginsCacheFileTransformer target = new com.github.edwgiz.mavenShadePlugin.log4j2CacheTransformer.PluginsCacheFileTransformer()
 
         @Override
         boolean canTransformResource(FileTreeElement element) {
-            return target.canTransformResource(element.getPath());
```
tobiasdiez: @koppor what is your opinion about these code style related changes?

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Thu Apr 13 18:56:36 CEST 2017
+#Fri May 12 13:26:57 CEST 2017
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-3.5-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-3.5-all.zip
```
tobiasdiez: I think we changed it to `-bin` on purpose...don't remember the details through. What is the reason for this change?

src/main/java/org/jabref/logic/pdf/search/indexing/DocumentReader.java
```diff
@@ -0,0 +1,94 @@
...
+import static org.jabref.model.pdf.search.SearchFieldConstants.KEYWORDS;
+import static org.jabref.model.pdf.search.SearchFieldConstants.SUBJECT;
+
+public final class DocumentReader {
+
+    private final BibEntry entry;
+    private final PDFTextStripper pdfTextStripper = new PDFTextStripper();
+
+    public DocumentReader(BibEntry bibEntry) throws IOException {
+        if (!bibEntry.getField(FieldName.FILE).isPresent()) {
```
tobiasdiez: In my opinion the `readPdfContents` should just return an empty document (maybe `Optional<Document>`?) if the file field is not present. Makes the class a bit easier to use and there might be more reasons why a straightforward reading of the file is not possible (encrypted, not found, permission error etc).

src/main/java/org/jabref/logic/pdf/search/indexing/DocumentReader.java
```diff
@@ -0,0 +1,94 @@
...
+
+        this.entry = bibEntry;
+        pdfTextStripper.setLineSeparator("\n");
+    }
+
+    /**
+     * Reads the content and metadata from a pdf file
+     */
+    public Document readPdfContents() throws IOException {
+        Path pdfPath = Paths.get(this.entry.getField(FieldName.FILE).get());
```
tobiasdiez: The format of the files field is more complicated than a simple file path. In the end it contains a list of [linked files](https://github.com/JabRef/jabref/blob/master/src/main/java/org/jabref/model/entry/LinkedFile.java). You can get the list of linked files using `BibEntry.getFiles()`.

src/main/java/org/jabref/logic/pdf/search/indexing/DocumentReader.java
```diff
@@ -0,0 +1,94 @@
...
+            addMetaData(pdfDocument, newDocument);
+            return newDocument;
+        } catch (IOException e) {
+            throw new IOException("Could not read pdf file: " + pdfPath + "!", e);
+        }
+    }
+
+    private void addMetaData(PDDocument pdfDocument, Document newDocument) {
+        PDDocumentInformation info = pdfDocument.getDocumentInformation();
+        addStringField(newDocument, AUTHOR, info.getAuthor());
```
tobiasdiez: Maybe we should use the metadata from the BibEntry? Not sure what the purpose of these information is.

src/main/java/org/jabref/logic/pdf/search/indexing/DocumentReader.java
```diff
@@ -0,0 +1,94 @@
...
+        newDocument.add(new StringField(field, value, Field.Store.YES));
+    }
+
+    private boolean isValidField(String value) {
+        return !(value == null || value.trim().isEmpty());
+    }
+
+    private void addContentIfNotEmpty(PDDocument pdfDocument, Document newDocument) throws IOException {
+        String pdfContent = pdfTextStripper.getText(pdfDocument);
+        if (!pdfContent.trim().isEmpty()) {
```
tobiasdiez: There is a method in StringUtils for this (isNotBlank I think).

build.gradle
```diff
@@ -141,6 +141,12 @@ dependencies {
     compile group: 'com.microsoft.azure', name: 'applicationinsights-core', version: '1.0.+'
     compile group: 'com.microsoft.azure', name: 'applicationinsights-logging-log4j2', version: '1.0.+'
 
+    compile 'org.apache.lucene:lucene-core:6.5.1'
```
Siedlerchr: If not already done, please also add it to the libraries.txt

src/main/java/org/jabref/logic/pdf/search/retrieval/PdfSearcher.java
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * Search for results matching a query in the Lucene search index
+     *
+     * @param searchString a pattern to search for matching entries in the index, must not be null
+     * @param maxHits      number of maximum search results, must be positive
+     * @return a result set of all documents that have matches in any fields
+     */
+    public PdfSearchResults search(String searchString, int maxHits) throws IOException {
+        if (Objects.requireNonNull(searchString, "The search string was null!").isEmpty()) {
```
Siedlerchr: This is a bit weird, with the isEmpty 
The Objects.requireNonNull will throw an NPE if the field is null. And you can omit the second parameter.
This is a breaking condition. So afterwards you can safely assume that searchString is not null

src/test/java/org/jabref/logic/pdf/search/indexing/DocumentReaderTest.java
```diff
@@ -0,0 +1,102 @@
...
+        assertEquals("Example2017", doc.get(SearchFieldConstants.KEY));
+        assertFalse(doc.get(SearchFieldConstants.CONTENT).isEmpty());
+        assertEquals("LaTeX with hyperref package", doc.get(SearchFieldConstants.CREATOR));
+    }
+
+    @Test
+    public void thesisExampleTest() throws IOException {
+        Path example = Paths.get("src/test/resources/pdfs/thesis-example.pdf");
+
+        BibEntry entry = mock(BibEntry.class);
```
Siedlerchr: I think you don't need to mock the BibEntry class, you can normally use it and then just call addField(...) or similar

src/test/java/org/jabref/logic/pdf/search/indexing/DocumentReaderTest.java
```diff
@@ -0,0 +1,102 @@
...
+    public void metaDataTest() throws IOException {
+        Path example = Paths.get("src/test/resources/pdfs/metaData.pdf");
+
+        BibEntry entry = mock(BibEntry.class);
+        when(entry.getField(FieldName.FILE)).thenReturn(Optional.of(example.toString()));
+        when(entry.getCiteKeyOptional()).thenReturn(Optional.of("MetaData2017"));
+
+        Document doc = new DocumentReader(entry).readPdfContents();
+
+        assertEquals("MetaData2017", doc.get(SearchFieldConstants.KEY));
```
Siedlerchr: Wouldn't it make sense to compare two Document-objects? (Needs an equals method then)
e.g. Document expected = new Document(....)

src/test/java/org/jabref/logic/pdf/search/indexing/PdfIndexerTest.java
```diff
@@ -0,0 +1,151 @@
...
+
+public class PdfIndexerTest {
+
+    private PdfIndexer indexer;
+    private BibDatabase database;
+
+    @Before
+    public void setUp() throws IOException {
+        this.indexer = new PdfIndexer();
+        this.database = mock(BibDatabase.class);
```
Siedlerchr: Same as above, you don't need to mock them
You only need to mock classes which operate on preferences

gradle/wrapper/gradle-wrapper.properties
lynyus: My IntellJ keeps reverting this to `all` all the time. Seems like in one commit I forgot to disable it. :/

src/main/java/org/jabref/logic/pdf/search/indexing/DocumentReader.java
lynyus: I would argue against that. After all indexing is a background task and if the reading the PDF fails it is worth throwing an exception instead of returning any useless value which will not be used anywhere.

src/main/java/org/jabref/logic/pdf/search/indexing/DocumentReader.java
lynyus: Well, I think the keywords field can be very useful. The others are probably redundant with the bibEntry. Generally, I will think about if it makes sense to incorporate bibtex fields into the lucene search.

src/test/java/org/jabref/logic/pdf/search/indexing/DocumentReaderTest.java
lynyus: Well, the question is if this is supposed to be a unit or an integration test.
I have decided to write a unit test here to test the DocumentReader and ONLY the DocumentReader.

Of course in this test the difference to a small integration test is not much. Would propose to keep it like that and add further integration tests soon.

src/test/java/org/jabref/logic/pdf/search/indexing/DocumentReaderTest.java
lynyus: The Document is a Lucene class.

src/main/java/org/jabref/logic/pdf/search/indexing/DocumentReader.java
lynyus: Uhm yes. Will change that and write some integration tests.

build.gradle
lynyus: done

lynyus:
I think I have incorporated most of the feedback by now.

What remains an open for discussion is actually the use case behind this feature which brings us to the question which information should go into the lucene search index and which shouldn't.

Currently the following is indexed:
  * The entry UID (Not searchable, but potentially useful to map back to the BibEntry)
  * The CiteKey from the BibEntry
  * The Author as stated in the PDF metadata
  * The content of the PDF
  * The Title as stated in the PDF metadata
  * The Subject as stated in the PDF metadata (redundant with Title?!)
  * The Keywords as stated in the PDF metadata
 
We could add more information from the BibEntry (like the name of the linked PDF file or the Name of the Journal/Conference) if you think this makes sense. Also if we cannot extract all metadata from the PDF does it make sense to use the ones from the BibEntry (i.e., keywords).

Also we really must decide how this PDF search relates to the existing JabRef search. I would be interested in what your original thoughts were, @koppor!

lenhard:
4.0 has been release and 4.1 is not far off.

What is missing to integrate this in master? A parallelization of the indexing process and the integration of the search in the UI, right?

koppor:
- [ ] indexing in the background
- [ ] separate search to enable the normal search box for table search. This somehow contradicts OneNote, but we think, separation is good.

lenhard:
Ping :)

Any chance of getting this into master before JabCon? I'd hate to have to move this to 4.3.

lenhard:
It seems that @LinusDietz is not going to continue with this. Unless someone else steps in, we'll have to close it.

koppor:
Some usefile links were given at https://github.com/JabRef/jabref/pull/2264#issuecomment-290912524

koppor:
Related Work: Mendeley

grafik

koppor:
We added `lucene.jar` because of https://stackoverflow.com/a/60589932/873282

Future work: Add automatic build of `lucene.jar` in the main build workflow. Gradle sub project + copy task.

Siedlerchr:
I may need to do a similar workaround for the Apollo stuff,so this may be a good starting point

koppor:
Lucene 9 should work: https://twitter.com/jpountz/status/1406508603830411265?s=19

src/main/java/org/jabref/gui/StateManager.java
```diff
@@ -28,6 +32,8 @@
 
 import com.tobiasdiez.easybind.EasyBind;
 import com.tobiasdiez.easybind.EasyBinding;
+import org.apache.commons.logging.Log;
```
Siedlerchr, line 35: Must be 
```
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;
```

btut: Thanks! Thats what you get for trusting Itellij to autoguess imports...

src/main/java/org/jabref/model/search/rules/ContainBasedSearchRule.java
```diff
@@ -1,27 +1,43 @@
...
 public class ContainBasedSearchRule implements SearchRule {
 
     private final boolean caseSensitive;
+    private final boolean fulltext;
 
-    public ContainBasedSearchRule(boolean caseSensitive) {
+    private String lastQuery;
+    private List<SearchResult> lastSearchResults;
+
+    public ContainBasedSearchRule(boolean caseSensitive, boolean fulltext) {
```
Siedlerchr, line 26: I think you could use an enum here for all modes and combine them? Like an EnumSet, then you can get rid of the booleans https://eddmann.com/posts/using-bit-flags-and-enumsets-in-java/

btut: The values for caseSensitive and fulltext mainly comes from the state of the respective toggle button (a boolean). If we change to an EnumSet, you would have to construct the EnumSet from the booleans. I don't think that would make the code any better.

src/main/java/org/jabref/logic/pdf/search/indexing/DocumentReader.java
btut: I will remove most metadata from the index as the search in the metadata will be kept as-is. I will create an ADR on why.

btut:
7804060 is quite large. Let me explain a little.

I agreed with @koppor that it would be best to have one lucene search index per database.
This means that we can store the index and only update it when file-links change, instead of having one big index for all open databases that needs to be updated each time a database is opened/closed/created...

Since the index is needed in the search-rules, I need to pass the BibDatabaseContext all the way down to the search-rules (so they can find the correct index) and this required more changes than I would have hoped.

All suggestions for better solutions are very welcome!
For the time being, I did not adapt the tests in the hopes someone comes up with a better plan.

btut:
I think edc6fe4 is better, still curious what others think though.

koppor:
This refs https://github.com/JabRef/jabref/issues/2263, but does not fix it, because following two issues are not fixed

- ...add an inner PDF reader, which can be used to read papers easily and make notes
- ...make adding remarks easier

src/main/java/org/jabref/model/search/rules/RegexBasedSearchRule.java
```diff
@@ -92,7 +90,11 @@ public PdfSearchResults getFulltextResults(String query, BibEntry bibEntry) {
             this.lastQuery = query;
             lastSearchResults = List.of();
             try {
-                PdfSearcher searcher = PdfSearcher.of(databaseContext);
+                if (Globals.stateManager.getActiveDatabase().isEmpty()) {
```
Siedlerchr, line 93: Although we agreed to use the Globals here, I would  nonetheless add it to the constructor 
e.g. this.stateManager = Globals.stateManager so it's easier to spot for future refactorings and not  hidden somehwere in a method down below

btut: Done in 85beb5c

src/main/java/org/jabref/model/pdf/search/SearchFieldConstants.java
```diff
@@ -3,13 +3,10 @@
...
     public static final String CONTENT = "content";
-    public static final String SUBJECT = "subject";
-    public static final String KEYWORDS = "keywords";
-    public static final String KEY = "key";
     public static final String MODIFIED = "modified";
 
-    public static final String[] PDF_FIELDS = new String[]{PATH, TITLE, AUTHOR, CONTENT, SUBJECT, KEYWORDS, KEY, MODIFIED};
+    public static final String[] PDF_FIELDS = new String[]{PATH, CONTENT, MODIFIED};
+
+    public static final String VERSION = "0.2a";
```
btut, line 11: This shows how changes in indexing should happen. If Fields or Tokenizers change, the index needs to be rebuilt. Changing the version changes the path the index is stored at. If JabRef is opened after such an update, the index is rebuilt at the new location.

What we need is a mechanism to handle out-of-date indices to make sure they don't fill up users file systems.

I would suggest: At startup, silently delete all indices that were built with a different version than what is currently running.
This would probably be the best solution for most users. For people running multiple versions of JabRef at the same time (and if there was an index-update between them) JabRef would rebuild the index each time they switch versions. I think this is acceptable and probably a nieche use-case.

btut:
Currently, we have the following behaviour:
- Someone performs a search
- Only entries that match the search are shown

This may lead to a problem.
The more generic search terms are, the more likely it becomes that all linked files match them (e.g. if I search for "something like this", the word "this" will be present in most documents and therefore all bib-entries match the search).
The search feature is then useless.

Lucene does provide a score for the results, so we could:
1. only consider the best 5 to be matches (currently we limit at 100)
2. use it to sort the bib-entries when a search is performed

Option 1 is a quick fix, option 2 requires some tinkering on how to make this user-friendly (people probably don't want to lose the filtering they set up / is it easy to understand that the entries on top match the search better? / how do we weight matches in the bib-entry information against fulltext search results?)

Siedlerchr:
> only consider the best 5 to be matches (currently we limit at 100)

This seems like a reasonable approach, just limit to 10 maybe

src/main/java/org/jabref/gui/LibraryTab.java
```diff
@@ -846,4 +856,64 @@ public void listen(EntriesRemovedEvent removedEntriesEvent) {
...
+                List<LinkedFile> addedFiles = new ArrayList<>(newFileList);
+                addedFiles.remove(oldFileList);
+                List<LinkedFile> removedFiles = new ArrayList<>(oldFileList);
+                removedFiles.remove(newFileList);
+
+                try {
+                    indexingTaskManager.addToIndex(PdfIndexer.of(bibDatabaseContext, preferencesService.getFilePreferences()), fieldChangedEvent.getBibEntry(), addedFiles, bibDatabaseContext);
+                    indexingTaskManager.removeFromIndex(PdfIndexer.of(bibDatabaseContext, preferencesService.getFilePreferences()), fieldChangedEvent.getBibEntry(), removedFiles);
+                } catch (IOException e) {
+                    e.printStackTrace();
```
Siedlerchr, line 909:
```suggestion

```

src/main/java/org/jabref/gui/entryeditor/fileannotationtab/FulltextSearchResultsTab.java
```diff
@@ -0,0 +1,74 @@
...
+    @Override
+    protected void bindToEntry(BibEntry entry) {
+        if (!shouldShow(entry)) {
+            return;
+        }
+        PdfSearchResults searchResults = stateManager.activeSearchQueryProperty().get().get().getRule().getFulltextResults(stateManager.activeSearchQueryProperty().get().get().getQuery(), entry);
+        StringBuilder content = new StringBuilder("<p>" + Localization.lang("Search results") + " </p>");
+
+        for (SearchResult searchResult : searchResults.getSearchResults()) {
+            content.append("<p>Found match in <a href=");
```
Siedlerchr, line 53: Localize

src/main/java/org/jabref/gui/search/RebuildFulltextSearchIndexAction.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    private void rebuildIndex() {
+        if (!shouldContinue || stateManager.getActiveDatabase().isEmpty()) {
+            return;
+        }
+        try {
+            currentLibraryTab.get().getIndexingTaskManager().createIndex(PdfIndexer.of(databaseContext, filePreferences), databaseContext.getDatabase(), databaseContext);
+        } catch (IOException e) {
+            dialogService.notify(Localization.lang("Failed to access fulltext search index"));
```
Siedlerchr, line 68: I would also log the original exception, as an IOException can have many reasons it's otherwise hard to debug.
And then  dialogService.notify(Localization.lang("Failed to access fulltext search index." See error log for details.));

src/main/java/org/jabref/logic/pdf/search/indexing/DocumentReader.java
```diff
@@ -0,0 +1,138 @@
...
+     *
+     * @return An Optional of a Lucene Document with the (meta)data. Can be empty if there is a problem reading the LinkedFile.
+     */
+    public Optional<Document> readLinkedPdf(BibDatabaseContext databaseContext, LinkedFile pdf) {
+        Optional<Path> pdfPath = pdf.findIn(databaseContext, filePreferences);
+        if (pdfPath.isPresent()) {
+            try {
+                return Optional.of(readPdfContents(pdf, pdfPath.get()));
+            } catch (IOException e) {
+                LOGGER.info("Could not read pdf file: " + pdf.getLink() + "!", e);
```
Siedlerchr, line 65: Shouldn't this be LOGGER.error?

Siedlerchr: See http://www.slf4j.org/faq.html#paramException,the last parameter must be exception then
                LOGGER.info("Could not read pdf file {}!,pdf.getLInk() e);

src/main/java/org/jabref/logic/pdf/search/indexing/DocumentReader.java
```diff
@@ -0,0 +1,138 @@
...
+        try {
+            PDFTextStripper pdfTextStripper = new PDFTextStripper();
+            pdfTextStripper.setLineSeparator("\n");
+
+            String pdfContent = pdfTextStripper.getText(pdfDocument);
+            if (StringUtil.isNotBlank(pdfContent)) {
+                newDocument.add(new TextField(CONTENT, pdfContent, Field.Store.YES));
+            }
+        } catch (IOException e) {
+            LOGGER.info("Could not read contents of PDF document " + pdfDocument.toString(), e);
```
Siedlerchr, line 131: can also be paramterized

src/main/java/org/jabref/model/pdf/search/SearchFieldConstants.java
Siedlerchr: Yep, I think this is fine.
What about portable versions?

src/main/java/org/jabref/model/search/rules/ContainBasedSearchRule.java
```diff
@@ -56,6 +81,28 @@ public boolean applyRule(String query, BibEntry bibEntry) {
...
+
+        if (!query.equals(this.lastQuery)) {
+            this.lastQuery = query;
+            lastSearchResults = List.of();
+            try {
+                PdfSearcher searcher = PdfSearcher.of(databaseContext);
+                PdfSearchResults results = searcher.search(query, 100);
+                lastSearchResults = results.getSortedByScore();
+            } catch (IOException e) {
+                e.printStackTrace();
```
Siedlerchr, line 102: LOGGER.error(...)

src/test/java/org/jabref/logic/pdf/search/retrieval/PdfSearcherTest.java
```diff
@@ -0,0 +1,99 @@
...
+    private PdfSearcher search;
+
+    @BeforeEach
+    public void setUp() throws IOException {
+        FilePreferences filePreferences = mock(FilePreferences.class);
+        // given
+        BibDatabase database = new BibDatabase();
+        BibDatabaseContext context = mock(BibDatabaseContext.class);
+        when(context.getFileDirectories(Mockito.any())).thenReturn(Collections.singletonList(Path.of("src/test/resources/pdfs")));
+        when(context.getFulltextIndexPath()).thenReturn(Path.of("src/test/resources/luceneTestIndex"));
```
Siedlerchr, line 38: Better is to use the @TempDir , this ensure that anything inside is deleted after the test
https://www.baeldung.com/junit-5-temporary-directory

src/test/resources/org/jabref/logic/importer/fileformat/MsBibImporterTestTranslator.xml
```diff
@@ -1,30 +1,31 @@
...
-         </b:Translator>
-      </b:Author>
-      <b:RefOrder>1</b:RefOrder>
-      <b:Month>07</b:Month>
-      <b:Day>1</b:Day>
-      <b:YearAccessed>2018</b:YearAccessed>
-      <b:MonthAccessed>07</b:MonthAccessed>
-      <b:DayAccessed>1</b:DayAccessed>
-   </b:Source>
+<b:Sources xmlns:b="http://schemas.openxmlformats.org/officeDocument/2006/bibliography"
```
Siedlerchr, line 2: Does the test still work? I remember that iat some point I had problems with the formatting

Siedlerchr (changes requested):
overall looks good to me, some small changes

src/main/java/org/jabref/model/pdf/search/SearchFieldConstants.java
btut: Good point, I need to try what the AppDirs library returns in such a case.

src/main/java/org/jabref/logic/pdf/search/indexing/PdfIndexer.java
```diff
@@ -0,0 +1,248 @@
...
+            // Check if a document with this path is already in the index
+            try {
+                IndexReader reader = DirectoryReader.open(directoryToIndex);
+                IndexSearcher searcher = new IndexSearcher(reader);
+                TermQuery query = new TermQuery(new Term(SearchFieldConstants.PATH, linkedFile.getLink()));
+                TopDocs topDocs = searcher.search(query, 1);
+                // If a document was found, check if is less current than the one in the FS
+                if (topDocs.scoreDocs.length > 0) {
+                    Document doc = reader.document(topDocs.scoreDocs[0].doc);
+                    long indexModificationTime = Long.valueOf(doc.getField(SearchFieldConstants.MODIFIED).stringValue());
```
calixtus, line 222:
```suggestion
                    long indexModificationTime = Long.parseLong(doc.getField(SearchFieldConstants.MODIFIED).stringValue());
```

src/main/java/org/jabref/logic/pdf/search/retrieval/PdfSearcher.java
```diff
@@ -0,0 +1,74 @@
...
+import org.apache.lucene.store.SimpleFSDirectory;
+
+import static org.jabref.model.pdf.search.SearchFieldConstants.PDF_FIELDS;
+
+public final class PdfSearcher {
+    private static final Log LOGGER = LogFactory.getLog(PdfSearcher.class);
+
+    private final Directory indexDirectory;
+
+    private PdfSearcher(Directory indexDirectory) throws IOException {
```
calixtus, line 34: IOException is never thrown

src/main/java/org/jabref/logic/pdf/search/retrieval/PdfSearcher.java
```diff
@@ -0,0 +1,74 @@
...
+        this.indexDirectory = indexDirectory;
+    }
+
+    public static PdfSearcher of(BibDatabaseContext databaseContext) throws IOException {
+        return new PdfSearcher(new SimpleFSDirectory(databaseContext.getFulltextIndexPath()));
```
calixtus, line 39: SimpleFSDirectory is deprecated. See above.

src/main/java/org/jabref/logic/search/SearchQuery.java
```diff
@@ -58,13 +58,15 @@ String format(String regex) {
     private final String query;
     private final boolean caseSensitive;
     private final boolean regularExpression;
+    private final boolean fulltext;
     private final SearchRule rule;
 
-    public SearchQuery(String query, boolean caseSensitive, boolean regularExpression) {
+    public SearchQuery(String query, boolean caseSensitive, boolean regularExpression, boolean fulltext) {
```
calixtus, line 64: I know, they weren't enums before, but as the SearchQuery is used a few times, would it be possible to quickly make enums of the bools?

src/main/java/org/jabref/model/database/BibDatabaseContext.java
```diff
@@ -211,4 +215,12 @@ public void convertToLocalDatabase() {
     public List<BibEntry> getEntries() {
         return database.getEntries();
     }
+
+    public Path getFulltextIndexPath() {
+        Path appData = Path.of(AppDirsFactory.getInstance().getUserDataDir(JabRefFrame.FRAME_TITLE, SearchFieldConstants.VERSION, "org.jabref"));
```
calixtus, line 220: JabRefFrame is totally a GUI thing, something we need to totally rework sometime in the future too. This is very bad in the model module. We need somehow a more generic solution here. @koppor

src/main/java/org/jabref/model/pdf/search/EnglishStemAnalyzer.java
```diff
@@ -0,0 +1,25 @@
...
+
+public class EnglishStemAnalyzer extends Analyzer {
+
+    @Override
+    protected TokenStreamComponents createComponents(String fieldName) {
+        Tokenizer source = new StandardTokenizer();
+        TokenStream filter = new LowerCaseFilter(source);
+        filter = new StopFilter(filter, EnglishAnalyzer.ENGLISH_STOP_WORDS_SET);
+        filter = new DecimalDigitFilter(filter);
+        filter = new PorterStemFilter(filter);
```
calixtus, line 21: Looks strange. Maybe rename to stopFiler, decimalDigitFilter, porterStomFilter. Not really sure, what is easier to read.

src/main/java/org/jabref/model/pdf/search/PdfSearchResults.java
```diff
@@ -0,0 +1,45 @@
...
+public final class PdfSearchResults {
+
+    private final List<SearchResult> searchResults;
+
+    public PdfSearchResults(List<SearchResult> search) {
+        this.searchResults = Collections.unmodifiableList(search);
+    }
+
+    public PdfSearchResults() {
+        this.searchResults = Collections.unmodifiableList(Collections.emptyList());
```
calixtus, line 16: Collections.emptyList() already is unmodifiable. Wrapper is without use.

src/main/java/org/jabref/model/pdf/search/PdfSearchResults.java
```diff
@@ -0,0 +1,45 @@
...
+            }
+            return 0;
+        });
+        return Collections.unmodifiableList(sortedList);
+    }
+
+    private List<SearchResult> getSortedByAlphabet() {
+        // TODO implement sorting
+        throw new RuntimeException("Not implemented");
+    }
```
calixtus, line 36: Whats the plan here?

src/main/java/org/jabref/model/pdf/search/PdfSearchResults.java
```diff
@@ -0,0 +1,45 @@
...
+        List<SearchResult> sortedList = new ArrayList<>(searchResults);
+        sortedList.sort((searchResult, t1) -> {
+            if (searchResult.getLuceneScore() < t1.getLuceneScore()) {
+                return -1;
+            }
+            if (searchResult.getLuceneScore() > t1.getLuceneScore()) {
+                return 1;
+            }
+            return 0;
+        });
```
calixtus, line 29:
```suggestion
        sortedList.sort((searchResult, t1) -> Float.compare(searchResult.getLuceneScore(), t1.getLuceneScore()));
```

Suggested by IntelliJ 👀

src/main/java/org/jabref/model/pdf/search/SearchFieldConstants.java
calixtus: But not multiple versions of JabRef but of the Indexer? Since users can turn of the listening to the port and run two instances of JabRef of the same version on the same time...

src/main/java/org/jabref/model/pdf/search/SearchResult.java
```diff
@@ -0,0 +1,89 @@
...
+    private final String content;
+    private final long modified;
+
+    private final float luceneScore;
+    private String html;
+
+    public SearchResult(IndexSearcher searcher, Query query, ScoreDoc scoreDoc) throws IOException {
+        this.path = getFieldContents(searcher, scoreDoc, PATH);
+        this.content = getFieldContents(searcher, scoreDoc, CONTENT);
+        this.modified = Long.valueOf(getFieldContents(searcher, scoreDoc, MODIFIED));
```
calixtus, line 36:
```suggestion
        this.modified = Long.parseLong(getFieldContents(searcher, scoreDoc, MODIFIED));
```

src/main/java/org/jabref/model/pdf/search/SearchResult.java
```diff
@@ -0,0 +1,89 @@
...
+        this.luceneScore = scoreDoc.score;
+
+        TokenStream stream = TokenSources.getTokenStream(CONTENT, content, new EnglishStemAnalyzer());
```
calixtus, line 39: The whole tokenization of Lucene seems to be reworked and TokenSources.getTokenStream is marked deprecated. In the lates api documentation of lucene this doesn't even appear anymore. So this needs somehow to be redone.
Maybe you'll find some resources here (https://lucene.apache.org/core/8_9_0/core/org/apache/lucene/analysis/package-summary.html) or some hints here (https://stackoverflow.com/a/50785820).

src/main/java/org/jabref/model/search/rules/SearchRules.java
```diff
@@ -12,30 +12,30 @@ private SearchRules() {
     /**
      * Returns the appropriate search rule that fits best to the given parameter.
      */
-    public static SearchRule getSearchRuleByQuery(String query, boolean caseSensitive, boolean regex) {
+    public static SearchRule getSearchRuleByQuery(String query, boolean caseSensitive, boolean regex, boolean fulltext) {
```
calixtus, line 15: enums instead of booleans could be shared with SearchQuery. See above.

src/main/java/org/jabref/model/search/rules/ContainBasedSearchRule.java
calixtus: Depends. Readability is the leading concern...
See SearchQuery etc. above and below

src/main/java/org/jabref/preferences/SearchPreferences.java
```diff
@@ -7,11 +7,13 @@
     private final SearchDisplayMode searchDisplayMode;
     private final boolean isCaseSensitive;
     private final boolean isRegularExpression;
+    private final boolean isFulltext;
 
-    public SearchPreferences(SearchDisplayMode searchDisplayMode, boolean isCaseSensitive, boolean isRegularExpression) {
+    public SearchPreferences(SearchDisplayMode searchDisplayMode, boolean isCaseSensitive, boolean isRegularExpression, boolean isFulltext) {
```
calixtus, line 12: enums instead of bools could be used here too

calixtus (changes requested):
Phew, 53 files reviewed.
Some suggestions, some came from IntelliJ, some nitpicks, some questions, some important remarks. Enjoy 😅

lucene-jar/lib/src/main/java/lucene/jar/Library.java
```diff
@@ -0,0 +1,10 @@
+/*
+ * This Java source file was generated by the Gradle 'init' task.
+ */
+package lucene.jar;
+
+public class Library {
+    public boolean someLibraryMethod() {
+        return true;
+    }
+}
```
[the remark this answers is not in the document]
btut: I thought it was necessary to build the shadow-jar but apparently, the whole file is unnecessary.
The build process cannot be automated, one needs to regenerate the jar file when updating lucene.
However, starting from release 9, this workaround should not be necessary anymore and the shadow-jar can be removed.

src/main/java/org/jabref/logic/pdf/search/indexing/IndexingTaskManager.java
```diff
@@ -0,0 +1,188 @@
...
+    @Override
+    protected Void call() throws Exception {
+        // update index to make sure it is up to date
+        this.updateProgress(1, 1);
+        return null;
+    }
+
+    public Directory getIndexDirectory() {
+        return this.getIndexDirectory();
+    }
```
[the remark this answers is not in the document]
btut: Must be leftover from when there was only one index. Don't know how the endless recursion came to happen^^. Good catch!

src/main/java/org/jabref/logic/pdf/search/indexing/IndexingTaskManager.java
```diff
@@ -0,0 +1,188 @@
...
+    public void flushIndex(PdfIndexer indexer) {
+        enqueueTask(new BackgroundTask<Void>() {
+            @Override
+            protected Void call() throws Exception {
+                this.updateProgress(-1, 1);
+                indexer.flushIndex();
+                return null;
+            }
+        });
+    }
```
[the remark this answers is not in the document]
Siedlerchr: I think I saw an event bus subscriber earlier, this could be the reason why intellj says it's not called

src/main/java/org/jabref/model/pdf/search/SearchFieldConstants.java
btut: Yes, I introduced a new version number here because I assume the indexing to be updated much less frequently than JabRef. So if the fields stay the same between JabRef 5.4 and 5.5, users can keep the same index and could even run 5.4 and 5.5 at the same time.
Having two instances of JabRef running simultaneously with the same library open could lead to issues where one of them holds the lock on the index. Is this a scenario we have to consider?

src/main/java/org/jabref/logic/pdf/search/indexing/IndexingTaskManager.java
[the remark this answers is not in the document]
btut: The subscribers don't call these either I think. They only call the versions of the methods that add/remove single entries or files.
The signatures were already defined in @LinusDietz' code, I just updated the code where necessary.
I think they are safe to remove, but could be helpful sometime in the future. Is the policy to not have any dead (uncalled) code at all?

src/main/java/org/jabref/logic/pdf/search/indexing/IndexingTaskManager.java
```diff
@@ -0,0 +1,188 @@
...
+                task.executeWith(taskExecutor);
+                this.progressProperty().bind(task.progressProperty());
+            });
+        }
+        runningTask = task;
+        runningTask.onFinished(() -> {
+            this.progressProperty().unbind();
+            this.updateProgress(1, 1);
+            runningTask = null;
+        });
```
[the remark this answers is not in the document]
btut: Yes this is kind of hacky.
At the very least we need to change the name of the runningTask variable. It's not actually the running task, but the last task that was added.
Therefore, if more than one is in the queue, we make the new task run after the one that was last finished is done (onFinished).
I'll have a look at the Queue implementation, I'll just need to figure out how that can be connected to the BackgroundTask.

src/main/java/org/jabref/model/database/BibDatabaseContext.java
btut: Well, we could just use a literal 'JabRef'...

src/main/java/org/jabref/model/pdf/search/EnglishStemAnalyzer.java
btut: This is how people seem to do it. You basically chain them.
If you find it hard to read, instead of introducing a variable I would rather nest the calls.
`filter = new PorterStemFilter(new DecimalDigitFilter(new StopFilter(...)))`

src/main/java/org/jabref/model/pdf/search/PdfSearchResults.java
btut: Leftover from @LinusDietz. I guess it was just an idea, I don't see a usecase for it right now and would vote to remove it.

src/main/java/org/jabref/model/pdf/search/EnglishStemAnalyzer.java
calixtus: Meh... don't really have an opinion on this. maybe just leave it as it is.

src/main/java/org/jabref/model/pdf/search/PdfSearchResults.java
btut: Wow. good catch, IntelliJ.

src/main/java/org/jabref/model/database/BibDatabaseContext.java
calixtus: No, no magic strings. Maybe olly got an idea.

src/main/java/org/jabref/logic/pdf/search/indexing/IndexingTaskManager.java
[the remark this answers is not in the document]
calixtus: Ah ok, it's already a linked list. that way.
A java Queue is just a FIFO list

src/main/java/org/jabref/model/search/rules/ContainBasedSearchRule.java
btut: TBH, I don't think enum-sets are easier to read than booleans. If I look at a signature and see it expects an EnumSet, I need to check that set to understand what the options are. There is no need to do that with (named) booleans.
For my argument that I would need to convert them from the toggle-buttons, we could do that like described [here](https://stackoverflow.com/a/35851125/5401297).

src/main/java/org/jabref/model/pdf/search/SearchFieldConstants.java
btut: For the portable versions: I can only check linux and windows. I now log the location of the index when opening a database, @Siedlerchr could you check on mac please?

src/main/java/org/jabref/logic/pdf/search/indexing/DocumentReader.java
```diff
@@ -0,0 +1,138 @@
...
+            return newDocument;
+        }
+    }
+
+    private void addMetaData(PDDocument pdfDocument, Document newDocument, Path resolvedPdfPath) {
+        try {
+            BasicFileAttributes attributes = Files.readAttributes(resolvedPdfPath, BasicFileAttributes.class);
+            addStringField(newDocument, MODIFIED, String.valueOf(attributes.lastModifiedTime().to(TimeUnit.SECONDS)));
+        } catch (IOException e) {
+            LOGGER.error("Could not read timestamp for " + resolvedPdfPath, e);
```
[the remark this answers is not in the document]
btut: I noticed that some classes use apache commons logging (outside of this PR only CleanupUrlFormatter) while most use slf4j. Was there a migration to slf4j from when this PR was originally started?
The thing is, what you are suggesting only works in slf4j, so if this is the case I would migrate the classes in this PR to slf4j.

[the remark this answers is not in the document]
btut: I noticed that in code-howtos.md, it says to use slf4j, but the given code is for apache (LogFactory.getLog instead of LoggerFactory.getLogger). Shall I fix that in a separate PR?

src/main/java/org/jabref/logic/pdf/search/indexing/IndexingTaskManager.java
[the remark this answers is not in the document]
btut: I misunderstood your original comment. I thought the Queue you were suggesting was already tailored for tasks, but its just the regular java.util.Queue. In that case it would not change much in terms of functionality, but it would be easier to read I think.

src/test/resources/org/jabref/logic/importer/fileformat/MsBibImporterTestTranslator.xml
btut: I can't tell which test this is associated to. I would guess the MsBibImporterTest (which passes for me - just as all other tests), but can't find any reference there to this file.

src/main/java/org/jabref/logic/pdf/search/indexing/IndexingTaskManager.java
```diff
@@ -0,0 +1,188 @@
...
+import org.jabref.model.database.BibDatabaseContext;
+import org.jabref.model.entry.BibEntry;
+import org.jabref.model.entry.LinkedFile;
+
+import org.apache.lucene.store.Directory;
+
+/**
+ * Wrapper around {@link PdfIndexer} to execute all operations in the background.
+ */
+public class IndexingTaskManager extends BackgroundTask<Void> {
```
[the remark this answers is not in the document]
btut: void cannot be used as a type parameter since it's not a class, therefore you need to use Void. This is so it is clear that the return value is not used.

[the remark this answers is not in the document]
calixtus: Ah ok.  Special behavior of void. I'm probably thinking to formalistic.

src/main/java/org/jabref/logic/pdf/search/indexing/IndexingTaskManager.java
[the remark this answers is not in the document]
calixtus: Chose what's easier to read / clean code.

src/main/java/org/jabref/model/pdf/search/PdfSearchResults.java
btut: (talking about getSortedByAlphabet only, getSortedByScore is already used to put the most prominent search-results on top in the search results tab)

src/main/java/org/jabref/model/search/rules/ContainBasedSearchRule.java
btut: I converted to EnumSets in 44c975f6a.
I stopped at SearchPreferences not to mess with peoples preferences.

src/main/java/org/jabref/model/pdf/search/SearchFieldConstants.java
btut: Deletion of the out-of-date indices is in cb1c3935f

src/main/java/org/jabref/preferences/SearchPreferences.java
btut: I kept it here because I don't want to mess with peoples Search-Preferences (stored). How do you usually handle such updates?

src/main/java/org/jabref/logic/pdf/search/indexing/IndexingTaskManager.java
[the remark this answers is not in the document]
btut: I think an actual queue is more understandable. I love recursion, but in this case it's not necessary and an actual queue is better.
Implemented in 82b9d69

src/main/java/org/jabref/logic/pdf/search/indexing/DocumentReader.java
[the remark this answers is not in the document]
Siedlerchr: I don't know when exactly the migration was, but you should use the slf4j. Slf4j is a facade for the underlying logging implementation (log4j in this case)

src/test/resources/org/jabref/logic/importer/fileformat/MsBibImporterTestTranslator.xml
Siedlerchr: MsBibImporterTestFiles, it is a parameterized test for grab all bib and xml files in the folder

src/main/java/org/jabref/preferences/SearchPreferences.java
Siedlerchr: Have a look at the PreferencesMigration

src/main/java/org/jabref/gui/JabRefMain.java
```diff
@@ -139,4 +149,25 @@ private static void configureProxy(ProxyPreferences proxyPreferences) {
...
+                    LOGGER.info("Deleting out-of-date fulltext search index at {}.", path);
+                    Files.walk(path)
+                         .sorted(Comparator.reverseOrder())
+                         .map(Path::toFile)
+                         .forEach(File::delete);
+
+                }
+            }
+        } catch (IOException e) {
+            LOGGER.error("Could not access app-directory at {}", appData);
```
koppor, line 169:
```suggestion
            LOGGER.error("Could not access app-directory at {}", appData, e);
```

src/main/java/org/jabref/gui/JabRefMain.java
```diff
@@ -139,4 +149,25 @@ private static void configureProxy(ProxyPreferences proxyPreferences) {
...
+        }
+
```
koppor, line 171:
```suggestion
```

src/main/java/org/jabref/logic/pdf/search/indexing/DocumentReader.java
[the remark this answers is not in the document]
koppor: This refs https://github.com/JabRef/jabref/issues/5475#issuecomment-858083303.

src/main/java/org/jabref/logic/pdf/search/indexing/DocumentReader.java
```diff
@@ -0,0 +1,140 @@
...
```
koppor, line 101:
```suggestion
            LOGGER.error("Could not read timestamp for {}", resolvedPdfPath, e);
```

src/main/java/org/jabref/logic/pdf/search/indexing/PdfIndexer.java
```diff
@@ -0,0 +1,243 @@
...
+    /**
+     * Writes the file to the index if the file is not yet in the index or the file on the fs is newer than the one in
+     * the index.
+     * @param entry the entry associated with the file
+     * @param linkedFile the file to write to the index
+     */
+    private void writeToIndex(BibEntry entry, LinkedFile linkedFile) {
+        Optional<Path> resolvedPath = linkedFile.findIn(databaseContext, filePreferences);
+        if (resolvedPath.isEmpty()) {
+            LOGGER.warn("Could not find " + linkedFile.getLink());
```
koppor, line 204:
```suggestion
            LOGGER.warn("Could not find {}", linkedFile.getLink());
```

src/main/java/org/jabref/gui/JabRefMain.java
```diff
@@ -139,4 +149,25 @@ private static void configureProxy(ProxyPreferences proxyPreferences) {
             Authenticator.setDefault(new ProxyAuthenticator());
         }
     }
+
+    private static void clearOldSearchIndices() {
+        Path appData = Path.of(AppDirsFactory.getInstance().getUserDataDir(JabRefFrame.FRAME_TITLE, null, "org.jabref"));
+        Path currentIndexPath = Path.of(AppDirsFactory.getInstance().getUserDataDir(JabRefFrame.FRAME_TITLE, SearchFieldConstants.VERSION, "org.jabref"));
```
koppor, line 155: This appears twice here in the PR. Can it be factored out to another method? Just use `jabref` as string, no `FRAME_TITLE`.

src/main/java/org/jabref/model/database/BibDatabaseContext.java
koppor: I checked the documentation at https://github.com/harawata/appdirs. Magic string `jabref` is OK for me here.

koppor (changes requested):
small nitpicks, it is good to go otherwise

koppor:
This fixes https://github.com/JabRef/jabref/issues/6188

calixtus:
Im still not convinced with using the literal JabRef for index path and would like to talk about that in the call tonight. I wonder, if there isn't a better, more generic solution.

btut:
> Im still not convinced with using the literal JabRef for index path and would like to talk about that in the call tonight. I wonder, if there isn't a better, more generic solution.

I agree, especially because it is used in two distinct files and needs to be identical. However, I see how it is also problematic to use the one stored in JabRefFrame. Maybe we should define a new static variable in PdfIndexer?

src/test/resources/org/jabref/logic/importer/fileformat/MsBibImporterTestTranslator.xml
btut: All tests in MsBibImporterTestFiles pass for me.

src/main/java/org/jabref/logic/pdf/search/indexing/DocumentReader.java
btut: [ADR](https://adr.github.io/adr-manager/#/manager/JabRef/jabref/main/0022-search-metadata-directly,-without-indexing.md)

src/main/java/org/jabref/model/pdf/search/SearchFieldConstants.java
btut: Works well with the portable versions, index locations are as expected.

src/main/java/org/jabref/logic/pdf/search/indexing/DocumentReader.java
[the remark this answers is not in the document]
btut: I removed references to apache commons logging in #7907

src/main/java/org/jabref/logic/pdf/search/indexing/DocumentReader.java
```diff
@@ -122,6 +125,20 @@ private void addContentIfNotEmpty(PDDocument pdfDocument, Document newDocument)
...
+                    for (PDAnnotation annotation : page.getAnnotations( annotation -> {
+                        if (annotation.getContents() == null) {
+                            return false;
+                        }
+                        return annotation.getSubtype().equals("Text") || annotation.getSubtype().equals("Highlight");
+                    })) {
+                        newDocument.add(new TextField(ANNOTATIONS, annotation.getContents(), Field.Store.YES));
+                    }
+                } catch (IOException e) {
+                    e.printStackTrace();
```
Siedlerchr, line 138: LOGGER

btut: When will I learn... sorry.
Instead of logging the exception here I let it fall through to the other catch-block. Since it acesses the same file, the IO should happen on the earlier call anyways (at pdfTextStripper.getText).

koppor:
Result of call - we looked up https://lucene.apache.org/core/8_0_0/core/org/apache/lucene/store/NIOFSDirectory.html

> NOTE: NIOFSDirectory is not recommended on Windows because of a bug in how FileChannel.read is implemented in Sun's JRE. Inside of the implementation the position is apparently synchronized. See here for details. 

For now, we accept that and think, no issues occur.

koppor (approved):

calixtus (approved):

Siedlerchr:
FYI: mac index file is located under  /Users/christophs/Library/Application Support/JabRef/0.3a

koppor:
Directory is determined using the `appdirs` library: https://github.com/harawata/appdirs#appdirs

PR #2840, opened 2017-05-14
Convert crossref editor to tag-like interface using JavaFX

tobiasdiez:
The crossref editor is the last one still implemented in Swing. With this PR it is converted to JavaFX and looks like this:

image

-----


tobiasdiez:
This PR is now ready-for-review. There are still some points that can be improved but I think for now this is ok:
- Auto-completion to make it easier to add a new crossref.
- Popup over linked entry with further information and a button that lets the user jump to this entry in the main table.

src/main/java/org/jabref/gui/fieldeditors/LinkedEntriesEditor.java
```diff
@@ -0,0 +1,46 @@
...
+        this.fieldName = fieldName;
+        this.viewModel = new LinkedEntriesEditorViewModel(databaseContext);
+
+        ControlHelper.loadFXMLForControl(this);
+
+        linkedEntriesBar.setStringConverter(viewModel.getStringConverter());
+        linkedEntriesBar.setOnTagClicked((parsedEntryLink, mouseEvent) -> viewModel.jumpToEntry(parsedEntryLink));
+        Bindings.bindContentBidirectional(linkedEntriesBar.tagsProperty(), viewModel.linkedEntriesProperty());
+
+        //TODO: Add toolitp for tag: Localization.lang("Jump to entry")
```
lenhard: I assume this TODO is legacy and not coming from the current PR, right?

src/main/java/org/jabref/gui/fieldeditors/LinkedEntriesEditorViewModel.java
```diff
@@ -0,0 +1,55 @@
...
+
+            @Override
+            public ParsedEntryLink fromString(String key) {
+                return databaseContext.getDatabase().getEntryByKey(key).map(ParsedEntryLink::new).orElse(null);
+            }
+        };
+    }
+
+    public void jumpToEntry(ParsedEntryLink parsedEntryLink) {
+        // TODO: Implement jump to entry
```
lenhard: Related to the comment above: I would prefer it very much to remove the commented-out code here directly. Way back in time when JabRef was still full of such stuff, we decided to remove commented-out code.

tobiasdiez: Ok that policy is very understandable. In the current case, I'm not sure if we really should remove it. The "jump to entry" feature was present in previous versions and this PR removes it (temporarily). I don't see a nice way to re-implement it right now as we have no good interface to control the focus of the main table (I really don't want to directly use the `JabRefFrame` class). But removing the comment here and above results also in a change of the language files and the "Jump to entry" string has to be readded by the translators at some point in the future. 

Should I still remove it? I don't have a strong opinion about it.

lenhard: I see. It would be better to keep it in the language file. Thus, I would say it is ok to keep the code, but please extend the explanation, so that we know that this is not just the legacy stuff. Be verbose (no-problem with a multi-line comment) and explain the rationale and why it's problematic to re-add the code, like in you comment here.

tobiasdiez: Done

PR #2862, opened 2017-05-23
Fix export and import of MS office day/year/month acessed fields

### LINKED ISSUE #2859: MSBibEntry office xml import and (re-)export of yearAccessed monthAccessed dayAccessed
Bug report: After import of fields yearAccessed monthAccessed dayAccessed from an office xml file, they are stored in format mm dd, yy in the field msbib-accessed (or mm dd, yyyy).

Re-export fails as it expects format \d{2}\s*[.,-/]\s*\d{2}\s*[.,-/]\s*\d{2,4} but \d{2} \d{2}, \d{2,4} doesn't match.

Possible fixes: change DATE_PATTERN s.t. month and day may be separated by \s*[ .,-/]
OR line 168 of MSBibEntry.java to `sbDateAccesed.append(', ');` (Typo "Accesed" present in source code).

Additionally, shouldn't the biblatex field "urldate" be exported to the fields yearAccessed monthAccessed dayAccessed in an office xml file? And vice versa for import? In that case importing into msbib-accessed should be replaced by importing into urldate (except for possible cases where urldate wouldn't be appropriate).

In case that urldate doesn't correspond to yearAccessed monthAccessed dayAccessed in some or all cases, importing into msbib-accessed could be changed to yyyy-mm-dd (yy-mm-dd) to make it coherent with biblatex date format. Parse msbib-accessed as year month date if pattern \d{2,4}\s*-\s*\d{2}\s*-\s*\d{2} is matched, if not try DATE_PATTERN \d{2}\s*[ .,-/]\s*\d{2}\s*[.,-/]\s*\d{2,4} for month year day.

Siedlerchr:
Fix possible NPE in Date parsing
Fixes #2859 

### DIFF
--- a/src/main/java/org/jabref/logic/msbib/MSBibEntry.java
+++ b/src/main/java/org/jabref/logic/msbib/MSBibEntry.java
@@ -246,7 +246,6 @@ class MSBibEntry {
 
         Optional.ofNullable(dateAccessed).ifPresent(field -> addDateAcessedFields(document, rootNode));
 
-
         Element allAuthors = document.createElementNS(MSBibDatabase.NAMESPACE, MSBibDatabase.PREFIX + "Author");
 
         addAuthor(document, allAuthors, "Author", authors);
--- a/src/main/java/org/jabref/logic/msbib/MSBibEntry.java
+++ b/src/main/java/org/jabref/logic/msbib/MSBibEntry.java
@@ -162,7 +162,7 @@ class MSBibEntry {
                 Optional.ofNullable(monthAccessed),
                 Optional.ofNullable(dayAccessed));
 
-        dateAccessed = parsedDateAcessed.map(Date::getNormalized).orElse(null);
+        parsedDateAcessed.map(Date::getNormalized).ifPresent(date -> dateAccessed = date);
 
         NodeList nodeLst = entry.getElementsByTagNameNS("*", "Author");
         if (nodeLst.getLength() > 0) {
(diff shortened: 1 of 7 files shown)


src/main/java/org/jabref/logic/msbib/MSBibEntry.java
```diff
@@ -262,15 +244,16 @@ public Element getEntryDom(Document document) {
...
-            Matcher matcher = DATE_PATTERN.matcher(dateAccessed);
-            if (matcher.matches() && (matcher.groupCount() >= 3)) {
-                addField(document, rootNode, "Month" + "Accessed", matcher.group(1));
-                addField(document, rootNode, "Day" + "Accessed", matcher.group(2));
-                addField(document, rootNode, "Year" + "Accessed", matcher.group(3));
-            }
-        }
+        Optional<Date> parsedDateAcesseField = Date.parse(dateAccessed);
+        String yearAccessed = parsedDateAcesseField.flatMap(Date::getYear).map(accYear -> accYear.toString())
+                .orElse(null);
```
tobiasdiez: Instead of using null strings, I would prefer if you use the `ifPresent` method (since we probably don't want to add fields with null as value anyway).

src/main/java/org/jabref/model/entry/Date.java
```diff
@@ -33,15 +33,16 @@ public Date(TemporalAccessor date) {
     public static Optional<Date> parse(String dateString) {
         List<String> formatStrings = Arrays.asList("uuuu-M-d", "uuuu-M", "d-M-uuuu", "M/uu", "M/uuuu", "MMMM d, uuuu", "MMMM, uuuu",
                 "d.M.uuuu", "uuuu.M.d", "uuuu");
-        for (String formatString : formatStrings) {
-            try {
-                TemporalAccessor parsedDate = DateTimeFormatter.ofPattern(formatString).parse(dateString);
-                return Optional.of(new Date(parsedDate));
-            } catch (DateTimeParseException ignored) {
-                // Ignored
+        if (dateString != null) {
```
tobiasdiez: I don't like that null is a valid argument. We try to get rid of null values and should be consistent here. Either way, `if(StringUtil.isBlank(dateString)) return Optional.empty();` is probably preferable.

src/main/java/org/jabref/logic/msbib/MSBibEntry.java
Siedlerchr: The null values are an indicator that no value is present, and in this case it is consistent to the rest of the code. In the addField method there is already a  check ` if( value == null ) return `

src/main/java/org/jabref/model/entry/Date.java
Siedlerchr: Either we do a Objects.requireNonNull and leave the checking to the caller, or we use this isBlank Method. Currently it will throw an NPE from the TemporalAccessor.parse method if null is passed.

src/main/java/org/jabref/logic/msbib/MSBibEntry.java
tobiasdiez: Well, I also don't like the rest of the code exactly for this reason. But instead of rewriting everything I would vote for an incremental improvement and since you already converted parts to use optionals... (By the way, BibEntry has the distinction between setField and clearField precisely on similar grounds.)

src/main/java/org/jabref/model/entry/Date.java
tobiasdiez: +1 for throwing NPE to be consistent with the TemporalAccesor.parse method.

Siedlerchr:
Change the code to optionals. Merge it in now

Issue #2785, opened 2017-04-23
Allow resorting of groups using drag & drop


tobiasdiez:
Resort groups by drag and drop as raised in https://github.com/JabRef/jabref/issues/2656#issuecomment-287402637 and further discussed in https://github.com/JabRef/jabref/issues/2599.

AEgit:
To add upon this: When a group is assigned as subgroup to another group, the focus on this group is lost. Instead the group that was found below the initial group is selected.

1. Selected a group
2. Drag'n'drop onto another group, to reassign it as respective subgroup
3. The focus on this subgroup is lost in the groups panel - Instead the group that was initially found below the reassigned group is selected

AEgit:
I can confirm that the resorting of groups using drag'n'drop works in:
JabRef 4.0.0-dev--snapshot--2017-05-25--master--4c87dde9c
Windows 10 10.0 amd64 
Java 1.8.0_131

The issue I reported above, however, persists. Every time you assign a group to a new subgroup, the focus on your selected group is lost after the assignment. I will open a new ticket for this.

Issue #2787, opened 2017-04-23
Changes to groups and edits in entry editor no longer mark database as dirty


tobiasdiez:
For groups raised in https://github.com/JabRef/jabref/issues/2599#issuecomment-291474696 and also occurs in the current development build for changes in bibentries. 
Possible solution: main table should listen to changes to database + maintable (instead of the old push-notification).

koppor:
Refs https://github.com/JabRef/jabref/issues/2946

PR #2864, opened 2017-05-24
Implement #2785: resort groups using drag & drop

tobiasdiez:
With this PR it is now possible to resort groups using drag and drop.
When the user drags a group over the center of another one, then the target group is highlighted as follows
image
and the source group is added as a child. 

In contrast, when the user hovers over the bottom or top part of a group, then a line is added that indicates that the source group is added at this point (i.e. between Test 1 and Test 2 in this picture)
image

There are still some problems with the selection status after a drag and drop.


tobiasdiez:
Thanks for the positive feedback.

I'll fix the "not mark as dirty" issue as part of #2787 as in my opinion the cleanest solution for this is a global listener (instead of manually calling `markAsChanged` after every action).

PR #2866, opened 2017-05-25
[WIP] Sharelatex Integration

Siedlerchr:
Sharelatex Integration

Status:
TODO:
- [ ] GUI shows double entries on second connect
- [ ] Fix problems on error requires restart
- [ ] Add status messages 
- [ ] Fix problems with Accept changes dialog


src/main/java/org/jabref/gui/sharelatex/ShareLatexLoginDialogController.java
```diff
@@ -0,0 +1,80 @@
...
+                    String id = elem.getAsJsonObject().get("id").getAsString();
+                    String name = elem.getAsJsonObject().get("name").getAsString();
+                    String lastUpdated = elem.getAsJsonObject().get("lastUpdated").getAsString();
+                    String owner =  elem.getAsJsonObject().get("owner_ref").getAsString();
+                    System.out.println("ID " + id);
+                    System.out.println("Name " + name);
+                    System.out.println("LastUpdated " + lastUpdated);
+                    System.out.println("Owner" + owner);
+
+                    ShareLatexProjectViewModel model = new ShareLatexProjectViewModel(id, name, owner, lastUpdated);
```
Siedlerchr: @tobiasdiez  Could you please have a look if I did this javafx stuff right?
First a login dialog is shown (that part works) and on successful connect you get a list of projects which I want to show then in another dialog in a table 
My current question is: I am not sure how and where to pass a list of projects (the view model class)  to the other dialog/view.

tobiasdiez: A few points:
- This code here looks like it mostly belongs to the logic package.
- I would create a logic class that provides a convenient interface for the ShareLatex integration (e.g. methods like `login(username, password)` and `getProjects`). Lets call this class ShareLatexManager for the moment. 
- This manager class can be initialized in https://github.com/JabRef/jabref/blob/master/src/main/java/org/jabref/gui/DefaultInjector.java#L25 and then injected into each controller that needs access to the sharelatex functionality, for example similar to the journal abbreviation loader https://github.com/JabRef/jabref/blob/ae6d01be46c96ef57e9871dfeb208ab0b7aab295/src/main/java/org/jabref/gui/journals/ManageJournalAbbreviationsController.java#L43
- So you see that in JavaFX you almost never want to pass something directly to a new view. 
- The purpose of view models is to enrich or format the backend data in a nice way to display them directly in the GUI. For example, if one of your sharelatex projects have a property `dateCreated` but you only want to display the year of creation that the `ProjectViewModel` only contains `yearCreated`. But most of the work should really be done in the logic.

Hope this is helpful.

Siedlerchr: Thanks, got it to work now. By the way. Is is okay to fill the ProjectViewModel in the logic or does this violate the gui -> logic dependency? 
So should I create a model-container and use the ViewModel separately?

tobiasdiez: Yes, you have a model class (Project) in the model that is used in the logic, and then a simple wrapper (ProjectViewModel) around it for the gui.

src/main/java/org/jabref/gui/entryeditor/SourceTab.java
```diff
@@ -77,6 +78,15 @@ private Node createSourceEditor(BibEntry entry, BibDatabaseMode mode) {
             LOGGER.debug("Incorrect entry", ex);
         }
 
+        codeArea.richChanges().subscribe(x -> {
```
Siedlerchr: @tobiasdiez  I now found a way to get position information when some data is changed in the code editor, e.g. exactly the information I would need for the sharejs thing (I need the range information (e.g. at position 3 the character x was deleted)

Two problems:
1. This only applies to one entry - whereas I would need a similar function for the whole database content
The problem is that a simple value changed event is not sufficient, because there I only get the old and new value.
However, there is something like https://docs.oracle.com/javase/8/javafx/api/javafx/scene/control/TextFormatter.Change.html 
which applies for any controls that have TextInput


2.  This even only fires when I change sth in the code editor and I am not sure if it would be fired if I edit just the author.

Maybe you have some idea?

tobiasdiez: I don't think it is a good idea if the sharelatex stuff depends on the entry editor. We have a text comparison engine that is used in the merge dialog and directly gives you the information you need (what characters changed at which position).

Siedlerchr: This is outdated,I decoupled this from the entry editor a while ago

src/main/java/org/jabref/gui/sharelatex/ShareLatexProjectDialogViewModel.java
```diff
@@ -0,0 +1,61 @@
...
+        this.projects.clear();
+        this.projects.addAll(projectsToAdd.stream().map(ShareLatexProjectViewModel::new).collect(Collectors.toList()));
+    }
+
+    public SimpleListProperty<ShareLatexProjectViewModel> projectsProperty() {
+        return this.projects;
+    }
+
+    @Subscribe
+    public void listenToSharelatexEntryMessage(ShareLatexEntryMessageEvent event) {
```
Siedlerchr: @koppor Here I write the new content to the file

src/main/java/org/jabref/logic/sharelatex/WebSocketClientWrapper.java
```diff
@@ -0,0 +1,337 @@
...
+            } catch (IOException | InterruptedException e) {
+                Thread.currentThread().interrupt();
+                LOGGER.debug("Exception in taking from queue", e);
+            }
+        });
+
+    }
+
+    //Actual response handling
+    private void parseContents(String message) {
```
Siedlerchr: @koppor  Here I parse the responses I get from the server

src/main/java/org/jabref/logic/sharelatex/ShareLatexManager.java
```diff
@@ -0,0 +1,76 @@
...
+                connector.startWebsocketListener(projectID, database, preferences);
+            } catch (URISyntaxException e) {
+                LOGGER.error(e);
+            }
+            registerListener(ShareLatexManager.this);
+
+        });
+    }
+
+    public void sendNewDatabaseContent(BibDatabaseContext database) {
```
Siedlerchr: @koppor  Here the changes are sended, the bibtexdb is written into a string and then the changes are calculated

src/main/java/org/jabref/logic/sharelatex/ShareLatexParser.java
```diff
@@ -0,0 +1,175 @@
...
+                if (name.endsWith(".bib")) {
+                    return id;
+                }
+
+            }
+        }
+        return "";
+    }
+
+    public List<SharelatexDoc> generateDiffs(String before, String after) {
```
Siedlerchr: @koppor  This is what I do for generating the diffs for sending

koppor:
I had a phone call with @Siedlerchr some weeks ago. I pointed him to the user experience of the [shared database feature](http://help.jabref.org/en/SQLDatabase). Especially the conflict handling. It seems not be as easy as expected to switch to the behavior, but we agreed that it is worth a try to think of it. Personally, I'd like to have a good user experience here, because this feature might be a killer feature of JabRef.

build.gradle
```diff
@@ -129,6 +130,12 @@ dependencies {
     compile group: 'com.microsoft.azure', name: 'applicationinsights-core', version: '1.0.9'
     compile group: 'com.microsoft.azure', name: 'applicationinsights-logging-log4j2', version: '1.0.9'
 
+    compile "org.glassfish.tyrus.bundles:tyrus-standalone-client:1.13.1"
+    compile "org.glassfish.tyrus.ext:tyrus-extension-deflate:1.13.1"
+
+    compile "com.google.code.gson:gson:2.8.0"
```
koppor: Don't we have another JSON library in JabRef? Maybe via unirest-java?

build.gradle
```diff
@@ -129,6 +130,12 @@ dependencies {
...
+    compile "org.bitbucket.cowwoc:diff-match-patch:1.1"
```
koppor: Do we really need another diff library? Can't we use the diff library already used in the merge entries dialog? It is OK, if we do. Then, please explain it somehow. Preferrably using [MADR](https://github.com/adr/madr).

koppor: Do we really need another diff library? Can't we use the diff library already used in the merge entries dialog? It is OK, if we do. Then, please explain it somehow. Preferrably using [MADR](https://github.com/adr/madr).

build.gradle
koppor: Don't we have another JSON library in JabRef? Maybe via unirest-java?

src/test/java/org/jabref/logic/sharelatex/ShareLatexManagerTest.java
```diff
@@ -0,0 +1,33 @@
...
+
+public class ShareLatexManagerTest {
+
+    @Test
+    public void test() throws URISyntaxException {
+
+        List<ShareLatexProject> projects;
+        try {
+            ShareLatexManager manager = new ShareLatexManager();
+            manager.login("http://192.168.1.248", "joe@example.com", "test");
```
koppor: Could you comment where this magic IP comes from?

src/test/java/org/jabref/logic/sharelatex/ShareLatexManagerTest.java
```diff
@@ -0,0 +1,33 @@
...
+
+            //  manager.login("https://www.sharelatex.com", "email ", "password" );
+
+            projects = manager.getProjects();
+            assertFalse(projects.isEmpty());
+        } catch (IOException e) {
+            // TODO Auto-generated catch block
```
koppor: Please remove this comment. Please log exceptions. -- In tests: Why not letting the test method throw an exception as we do in other tests, too?

koppor:
I have some UI issues:

## tool bar

grafik

- [ ] Fix color

## connection dialog

grafik

- [ ] What do I need to enter for the server address? Why is there no value suggested? - I think, the value is `https://www.sharelatex.com`.
- [ ] Position the "Cancel" button more nicely. Maybe let it start at the same column as the text box?
- [ ] After pressing "Login", the button text should change to "Logging in..." and be disabled. Otherwise, I do not know that JabRef is doing something.

## Choose Project dialog

grafik

- [ ] The buttons "Synchronize Library" and "Cancel" should be swapped, because of consistency with the former dialog
- [ ] I do not know, what happens. Will it upload my whole library? There should be help text displayed. I would have expected, that JabRef downloads the bib file from the ShareLaTeX folder and stores it locally in a temporary folder not visible to the user. Similar to the shared database feature: I do not have a local copy (unless I want to).
- [ ] After pressing the synchronize button, my already opened library file is opened. Without an indicator that there is a connection. I think, in the SharedDatabase feature, there is an indicator such as `[synced]`.

## Tool bar

If I press the lion button again, I am asked to enter the connection details again. How do I know that the current library is synched?

## Menu

grafik

Do I really need to press the "Send changes to server" button? Why is there not an automatic sync like in the case of a shared database and PostgreSQL?

koppor:
The synchronization has issues with `&` in the the text content. When I add ISBN `978-0-471-44846-4` and sync the entry, the `&` in the URL field are escaped:

grafik

tobiasdiez:
From an architectural view point it would be very nice to have a common interface that defines the interaction with external databases (i.e. setup the connection/login, bidirectional sync of changes, etc). Then the current sql-based databases are one implementation, while your sharelatex feature is implemented as another implementation. In the best case the gui only works against the interface. In this way you also ensure that the user experience is the same for all the connections to external databases (e.g. don't run into some of the problems @koppor mentions above). I know this is a lot of refactoring work, but in my opinion definitely worth!

build.gradle
Siedlerchr: Yes, this is diff library is the same which is used on the server side of sharelatex

build.gradle
Siedlerchr: We already have Guava. And gson is a much better json library than the default implementation

src/test/java/org/jabref/logic/sharelatex/ShareLatexManagerTest.java
Siedlerchr: Still an artifact from local tests.

Siedlerchr:
@tobiasdiez  The main difference between the shared SQL stuff and Sharelatex is the way the synchro works: The sql stuff works on an per entry base while the sharelatex.com works on a diff level with respect to the whole bibtex code as once.  It simply sees the bibtex "file" as a single string. The diff I get from the server reports a line/col diff.  There is no concept of entries.

tobiasdiez:
ok, I see, that complicates things. Nonetheless, you need to have a way to convert the changes coming from ShareLatex in a way JabRef understands them (i.e. in terms of BibEntries), right? Or how do you handle conflicts right now? Ideally you would want to reuse the ChangeScanner class that is used from handle external changes to the file.

koppor:
- [ ] login data should be persisted similar to the shared-datbase feature

Siedlerchr:
@tobiasdiez atm I do a trick to receive the whole bibtex code, write it to the file and let the change scanner do it's work

tobiasdiez:
Ok, but then I don't see the challenge here. In abstract terms, the change notification thus gives you a list of pairs of bib entries `<before, after>`. This is the same for the sql stuff and for the sharelatex implementation (and even for external changes to the bib file). Thus all these (messy) implementation details can be hidden under exactly such an interface, where you listen for changes in the above form. Do I miss something?

koppor:
@Siedlerchr uses the file change detection thing while SharedDB uses the in
memory represention.

I would create the bib in memory representation using the server info,
check which entries are different and publish that info using the event bus

Am 20.10.2017 09:34 schrieb "Tobias Diez" <notifications@github.com>:

koppor:
- [ ] 404 at the URL should be caught. If I enter `https://www.sharelatex.com/` as URL, there should be no 404, but the `/` be removed. When setting `https://www.sharelatex.com` as default, this issue would be solved.

Siedlerchr:
I worked on the gui and made it more consistent with other dialogs and I added sharelatex.com as default (although you could use your own [instance in a docker container](https://github.com/sharelatex/sharelatex/wiki/Quick-Start-Guide#using-a-compose-file))

koppor:
In the File menu, only two of the three elements should be highlighted:

grafik

"Disconnect ..." should only be enabled when sync is active.

src/main/java/org/jabref/gui/sharelatex/ShareLatexProjectDialogViewModel.java
Siedlerchr: @tobiasdiez  Here I receive the whole remote bibtex db as string and overwrite the current active database.

My future vision would be: Apply delta changes (lines/cols-diff) from remote to local bibtex database. 
But that is still a bit more complicated

koppor:
"More nicelcy" from https://github.com/JabRef/jabref/pull/2866#issuecomment-337807498 did not mean to switch the positions of the buttons. "Login" should be kept right.

grafik

koppor:
- [ ] When closing and reopening a sahred database, the connection status is kept. Is this intented?

grafik

koppor:
Local save does not work any more:

grafik

But, review dialog does not popup. 2 clicks on "save" needed.

Siedlerchr:
@koppor  Regarding the save changes: Yes, I noticed that, too. Did not change any at the code regarding the saving. Seems to be coming from @tobiasdiez  changes to the DB Monitor.

Disconnect: Connection is kept on line, as long as you don't hit disconnect. 

Regarding the gui dialog:
I made the layout consistent with other dialogs. Cancel button is on the right.

Siedlerchr:
@koppor  I just pushed a new version which implements some entry merging based on the dialog you love so much.

1. Sharelatex server send me an update message -> I store the line number
2. I do my trick to get all entries from sharelatex -> List with entries
3. I  now identify an entry from the line number (if possible) 
4. If line number is part of entry -> I get that entry: If not I dumb the whole content as before -> External Change
5. Check if I have an entry with that **citekey** in the local db
6. If the entries do not equal I show the Merge Entries dialog.
7. Merged entries are stored locally

It works if you don't type too fast on the sharelatex server side.
Otherwise you will get that Merge Dialog everytime someone types a char and waits some time at the server side (as the update from the server is sent immediately)

koppor:
Summary of JabCon discussion:

- Multiple users should be supported "seamlessly"
- If user A and B work on the bibfile, a modification by user A on a bibtex entry B1 should not trigger any "changes" dialog if user B did not do any changes.
- A "Merge Entries" dialog should popup if users A and B modified the same entry.

How this works can be tried out using the [Shared SQL Database Feature](http://help.jabref.org/en/SQLDatabase) of JabRef, where @obraliar and me worked hard to  have a very nice user experience.

Technically, this is not that easy to achieve. However, users will be happy, if it "just works".

At each instance of JabRef:

- Local state L is known
- Remote state R is known
- In case an update is made remotely, the diff can be applied to R.
  So R' is known. Based on R and R', the entries E where a change was made, are known.
  Based on E, changes can be applied to L.
  If in L the entry was not modified: just apply the update.
  If in L the entry was modified: Start the merge entries dialog.

tobiasdiez:
What's the status here?

Siedlerchr:
Works theoretically, but needs some handling of the merge stuff.

koppor:
We discussed the usability issues at JabCon 2018. 

Meanwhile, it seems my requirements on usability and preventing huge "WTFs" of users are too high. I was hoping that @obraliar would step in as he also implemented the shared-DB-support for JabRef. Maybe [#hacktoberfest](https://hacktoberfest.digitalocean.com/) could motivate him?

My personal last resort is a hacking session at JabCon 2019.

tobiasdiez:
What's the status of this PR? Is somebody actively working on it? If not I would close it for now and reopen as soon as somebody has the time to continue with it.

PR #2873, opened 2017-05-28
Fix Divide by zero exception if rows is zero in Entry Editor Tab

Siedlerchr:
Internal thing, it could happen that when the fields array is empty the rows variable is zero which led to an Arithmetic Exception (divide by zero)


tobiasdiez:
Thanks! You still have the hard wrap at a certain line length on, don't you 😄 ?

PR #2876, opened 2017-05-30
Improve person names checker

tobiasdiez:
The integrity check now also tests that names are in BibTeX format.


src/main/resources/l10n/Menu_el.properties
```diff
@@ -1,31 +1,30 @@
+#!
```
lenhard: It seems that something went wrong when merging the language file here? Why is there a file header? Also, why are some of the localizations removed?

src/main/java/org/jabref/logic/integrity/PersonNamesChecker.java
```diff
@@ -15,6 +16,16 @@
...
             return Optional.of(Localization.lang("should end with a name"));
         }
+
+        // Check that the value is in one of the two standard BibTeX formats:
+        //  Last, First and ...
+        //  First Last and ...
+        AuthorList authorList = AuthorList.parse(value);
+        if (!authorList.getAsLastFirstNamesWithAnd(false).equals(value)
+                && !authorList.getAsFirstLastNamesWithAnd().equals(value)) {
+            return Optional.of(Localization.lang("Names are not in the standard BibTeX format."));
```
lenhard: Should we drop a hint here what the expected format actually is? Otherwise, the user might be left wondering what he should do to fix the warning.

tobiasdiez: I agree that the error message is not very detailed and leaves room for interpretation. However, since the check issues are reported in table form a very concise error message is required. Thus for now I don't see an easy way to improve it.

PR #2877, opened 2017-05-31
Fixes jacoco by excluding the fetcher tests from analysis

lynyus:
Two months ago Codecov stopped working. I have identified the problem being the failing fetcher tests.
By removing these from the jacoco test suite, the reports are generated again solving the issue.


koppor:
Yes, why not. When a new fetcher will be created, we will have to ensure, the code is tested. The functionality used outside the fetchers is surely covered using other tests.

Issue #2839, opened 2017-05-13
4.0.0-Beta not showing PDF File Annotations


kafran:
Text Highligh and Comments created with FoxitReader for Linux and Evince (with copy and past of the text inside the pop-up note, this is how it works on Docear) do not appear on File annotations tab.

lynyus:
Thanks for reporting this issue @kafran. We need some more information to resolve the issue.

Have you checked the event log (Help -> View event log) for messages related to this problem?
If there are not messages the easiest would be to send us a PDF file where the issue occurs.

kafran:
@lynyus Sure. https://gist.github.com/kafran/c6dc9ab84bb3a25b24b908b42d406d67

lynyus:
Thanks, that helps a lot. Will fix this soon!

kafran:
Thank you. @lynyus, also... is this possible [#2545](https://github.com/JabRef/jabref/pull/2545#issuecomment-301242416) Should I open a feature request?

lynyus:
Hi @kafran, the reported Issue should be fixed in the latest development branch. Could you please check the build from http://builds.jabref.org/master/. Thanks!

kafran:
Hey @lynyus, this is really getting good. I can see some annotations now. But, others are getting this error: https://gist.github.com/kafran/79903f6961a6e0757c13824ad5812662

Do you think this is JabRef or the PDF Reader (Foxit for Linux) I'm using? Some PDFs' who annotations works was annotated using the same software.

lynyus:
The problem lies in the the library we are using to extract these annotations.
I will investigate this "soon". Stay tuned. (:

kafran:
This new feature is getting awesome. Congratulations. =)
I'm not a developer but I use JabRef for helping me with my research at university. This new feature came in a good time when I was about to change my research workflow from Zotero/Docear/etc. to a more simple approach with BibLaTex + JabRef and TiddlyWiki for note taking. By the way, if someone else is looking for another way to organize a Master/Doctor Tesis research here is the tip: http://tesis.tiddlyspot.com/

This new feature gonna be handy to extract annotation and place on TiddlyWiki. And with the page information I can even save a link like this "file:///file.pdf#page=x" and open the file directly on the note on Firefox =D.

Thank you for this great work guys.

lynyus:
Hi @kafran, I had a look and created a fix for the bug you reported before https://github.com/JabRef/jabref/issues/2839#issuecomment-302058227.
[...]

PR #2878, opened 2017-06-01
Support Annotations Created by Foxit

lynyus:
Fixes some problems when Annotations were created using the Foxit Reader. (cf. https://github.com/JabRef/jabref/issues/2839#issuecomment-302058227)


lynyus:
Based on the comments of @kafran, I have made some more changes. I think this PR is ready now.

src/main/java/org/jabref/logic/pdf/PdfAnnotationImporter.java
```diff
@@ -71,6 +67,10 @@
     }
 
     private boolean isSupportedAnnotationType(PDAnnotation annotation) {
+        if (annotation.getSubtype().equals("Link") || annotation.getSubtype().equals("Widget")) {
```
Siedlerchr: I don't know how likely it is, but just to be sure you should change it to "Link".equals...  otherwise there might be a potenial NPE

lynyus: Good point, should not be the case, but that pdfbox library is really not to be trusted....

src/main/java/org/jabref/model/pdf/FileAnnotationType.java
```diff
@@ -50,6 +56,26 @@ public static FileAnnotationType parse(PDAnnotation annotation) {
...
 
+    /**
+     * Determines if a String is a supported marked FileAnnotation type.
+     *
+     * @param annotationType a type descriptor
+     * @return true if annotationType is a supported marked FileAnnotation type
+     */
+    public static boolean isMarkedFileAnnotationType(String annotationType) {
+        for (FileAnnotationType type : Collections.unmodifiableList(Arrays.stream(FileAnnotationType.values())
+                .filter(FileAnnotationType::isLinkedAnnotationType).collect(toList()))) {
```
Siedlerchr: An alternative and simpler solution would be to use Enum.valueOf... to parse the String. If it throws an Illegal Argument exception you simply return false...
Otherwise I would omit the loop and directly add a second filter Argument to the strea, and use anyMatch to return a boolean

Siedlerchr: And I don't understand why you wrap the Array in a list again

lynyus: good idea. fixed in b9ddbc4e7666598efc0e92ac08303972d1d78ce8

PR #2889, opened 2017-06-05
Fix DOI resolving by using https

Siedlerchr:
I think this is an important issue with small code change.  @JabRef/developers  Should we do a fix for 3.8,2?

For 3.8.2 https://github.com/JabRef/jabref/blob/v3.8.2/src/main/java/net/sf/jabref/logic/util/DOI.java

- [X] Tests created for change
~~- [ ] Check documentation status (Issue created for outdated help page at [help.jabref.org]~~(https://github.com/JabRef/help.jabref.org/issues)?)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - We added integration of the Library of Congress catalog as a fetcher based on the [LCCN identifier](https://en.wikipedia.org/wiki/Library_of_Congress_Control_Number). [Feature request 636 in the forum](http://discourse.jabref.org/t/loc-marc-mods-connection/636)
 
 ### Fixed
-- We fixed the adding of a new entry from DOI which led to a connection error [#2879](https://github.com/JabRef/jabref/issues/2897)
+- We fixed the adding of a new entry from DOI which led to a connection error. The DOI resolution now uses HTTPS to protect the user's privacy [#2879](https://github.com/JabRef/jabref/issues/2897)
 - We fixed the IEEE Xplore web search functionality [#2789](https://github.com/JabRef/jabref/issues/2789)
 - We fixed an error in the CrossRef fetcher that occurred if one of the fetched entries had no title
 - We fixed an issue that prevented new entries to be automatically assigned to the currently selected group [#2783](https://github.com/JabRef/jabref/issues/2783).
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - We added integration of the Library of Congress catalog as a fetcher based on the [LCCN identifier](https://en.wikipedia.org/wiki/Library_of_Congress_Control_Number). [Feature request 636 in the forum](http://discourse.jabref.org/t/loc-marc-mods-connection/636)
 
 ### Fixed
+- We fixed the adding of a new entry from DOI which led to a connection error [#2879](https://github.com/JabRef/jabref/issues/2897)
```
lynyus: Suggestion: Improve wording to make clear that now HTTPS is used to protect the privacy of the user.

gmonet:
Any binairie install plane? I can't create mine from jabref.install4j because i need a 'Windows key store'. I don't get what it means but I really need this update.
Thanks :)

matthiasgeiger:
This fix is already part of 4.0-beta2 or the recent Dev builds which can be downloaded here: https://builds.jabref.org/master 

So you don't need to build JabRef on your own.

Best,
Matthias

gmonet:
Thanks for your (amazing) fast answer.
Actually, i was speaking about Jabref 3.8. I tried 4.0-beta 2 but there are still some bugs (especially group display) and it crashes frequently.
The comit 2889 fixes DOI's problem for Jabred 3.8 (3.8.3 in the changelog) but it seems it's not distributed yet.

Thanks :)

jonasstein:
We are very interested on a 3.8.3 version for Gentoo too.

stefan-kolb:
I don't think we will release a 3.8.3. The only active branch and release version will be 4.x.

jonasstein:
Is there a special reason for that? It does not seem that the 4.x tree will have the stability of 3.8.x in the near future. We have many users who were interested in working on 3.8.2, but need doi.

lynyus:
Resources, ancient, hard to maintain technology.

lenhard:
@jonasstein Regarding the freeze when dragging to a group, one thing just came to my mind: What is the look and feel you are using? We had a few Linux users who experienced freezes with the GTK look and feel. Switching to a different look and feel (e.g. plastic) effectively solved their problem. Does that maybe also help in your case? If so, the trivial solution would be to set a different default look and feel on Linux. Maybe the solution is also too hidden in the documentation: https://help.jabref.org/en/Installation#freezes-when-running-jabref

jonasstein:
It was this one:
selection_383
I found that bug too and tried also all others. But this did not change the freeze.
selection_384

lenhard:
Ok, too bad. Thanks for testing!

Issue #2855, opened 2017-05-19
Command + X does not work properly


iTao8:
JabRef 4.0.0-dev--snapshot--2017-05-18--master--018173ebd
Mac OS X 10.11.6 x86_64 
Java 1.8.0_131

Steps to reproduce:
1. select an entry
2. double click the selected entry
3. select the title
4. Command +X
5. the selected entry is deleted

iTao8:
Similar problem #2842

Siedlerchr:
I found a solution and created a PR.

iTao8:
I think it has a similar problem with #2923, all are about the keybindings. For me, the Command + X has been fixed in 
JabRef 4.0.0-dev--snapshot--2017-06-18--master--26d1e2e07
Mac OS X 10.11.6 x86_64 
Java 1.8.0_131

Siedlerchr:
Thanks for your report. With the linked PR #2937  I also fixed the problem with doing CTRL+C inside the entry editor,e.g. inside the title field would always copy the whole entry bibtex code to the clipboard.
Could you please test that, too?

iTao8:
I test Command + C, Command + X, Command + Z, these keybindings work well in 
JabRef 4.0.0-dev--snapshot--2017-06-22--master--bee062c6a
Mac OS X 10.11.6 x86_64 
Java 1.8.0_131

tobiasdiez:
Thanks for the quick feedback!

iTao8:
This problem appears again in the current snapshot:

JabRef 4.0.0-dev--snapshot--2017-07-06--master--edd3f7cc1
Mac OS X 10.11.6 x86_64 
Java 1.8.0_131

Siedlerchr:
Hi, you have to reset the keybindings first (or if you use custom ones, you have to modify the preferences and add a plus sign between each modifier.  We ware working on a preference migraiton

iTao8:
This works well on Windows 7, but not on mac

JabRef 4.0.0-dev--snapshot--2017-07-10--master--e80f450ee
Mac OS X 10.11.6 x86_64 
Java 1.8.0_131

Siedlerchr:
Hi,
this should now be fixed again in the latest snapshot f23db0ee1a0130d7b835bb321acf5bf483cc7a0c. 
The snapshot should be ready in a couple of minutes.

PR #2892, opened 2017-06-06
Improve performance of entry editor

tobiasdiez:
This PR improves the performance of the entry editor to a satisfactory level (switching between entries leads to a short flicker of the entry editor but no real delay). Things changed:
- Tab panel is converted to JavaFX
- Each tab is initialized only upon focus

Moreover, a bit of the code was refactored (e.g. the creation of the different tabs is extracted from the EntryEditor class to many smaller classes, one for each tab).


src/main/java/org/jabref/gui/contentselector/ContentSelectorDialog.java
```diff
@@ -302,8 +310,7 @@ private void applyChanges() {
 
         // Update all selectors in the current BasePanel.
         if (changedFieldSet) {
-            // We have added or removed content selectors, update the entry editor
-            panel.rebuildAllEntryEditors();
+            // TODO: We have added or removed content selectors, update the entry editor
```
lenhard: What about this TODO? This shouldn't be merged as long as the PR still breaks the feature.

src/main/java/org/jabref/gui/entryeditor/FieldsEditorTab.java
```diff
@@ -0,0 +1,326 @@
...
+    private Region setupPanel(JabRefFrame frame, BasePanel bPanel, boolean compressed) {
+
+        //setupKeyBindings(panel.getInputMap(JComponent.WHEN_FOCUSED), panel.getActionMap());
+
+        editors.clear();
+        List<Label> labels = new ArrayList<>();
+
+        for (String fieldName : fields) {
+
+            // TODO: Reenable/migrate this
```
lenhard: Same here and at a few positions below. This looks like you are removing a few features including content selectors (again)?

src/main/java/org/jabref/gui/entryeditor/FieldsEditorTab.java
```diff
@@ -0,0 +1,326 @@
...
+                fieldEditor = new TextArea(field, null, getPrompt(field));
+                //parent.addSearchListener((TextArea) fieldEditor);
+                defaultHeight = fieldEditor.getPane().getPreferredSize().height;
+            }
+            
+            Optional<JComponent> extra = parent.getExtra(fieldEditor);
+            
+            // Add autocompleter listener, if required for this field:
+            /*
+            AutoCompleter<String> autoCompleter = bPanel.getAutoCompleters().get(field);
```
lenhard: Just a note: The auto completer is broken and therefore it is one of the few features that is fine to go.

src/main/java/org/jabref/gui/entryeditor/OtherFieldsTab.java
```diff
@@ -0,0 +1,36 @@
...
+import org.jabref.Globals;
+import org.jabref.gui.BasePanel;
+import org.jabref.gui.IconTheme;
+import org.jabref.gui.JabRefFrame;
+import org.jabref.logic.l10n.Localization;
+import org.jabref.model.entry.BibEntry;
+import org.jabref.model.entry.EntryType;
+
+public class OtherFieldsTab extends FieldsEditorTab {
+    public OtherFieldsTab(JabRefFrame frame, BasePanel basePanel, EntryType entryType, EntryEditor parent, BibEntry entry) {
```
lenhard: Very minor, but please add a blank between the class declaration and the method.

src/main/java/org/jabref/gui/entryeditor/SourceTab.java
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    /**
+     * NOTE: This method is only used for the source panel, not for the
+     * other tabs. Look at EntryEditorTab for the setup of text components
+     * in the other tabs.
+     */
+    private void setupJTextComponent(JTextComponent textComponent) {
+        /*
+        // TODO: Set up key bindings and focus listener for the code editor.
```
lenhard: This todo is quite important

src/main/java/org/jabref/gui/contentselector/ContentSelectorDialog.java
tobiasdiez: The content selector feature is not yet reimplemented in the new FX entry editor and this PR doesn't change anything in this respect. The purpose of this particular code line was to update the content selectors in the editor if the user changes them. But since content editors are currently not enabled, I can't say which is the best way to reimplement the update functionality.

src/main/java/org/jabref/gui/entryeditor/FieldsEditorTab.java
tobiasdiez: I actually didn't change anything here in this PR (github just marks the whole file as changed since it was renamed).

src/main/java/org/jabref/gui/entryeditor/OtherFieldsTab.java
tobiasdiez: done

src/main/java/org/jabref/gui/contentselector/ContentSelectorDialog.java
lenhard: Ok, I get it. If this is just a false marking by github and the TODOs have been there before, then this can just move forward.

However, we need to be sure that we do not release 4.0 without content selectors. Should we open an issue for this? Or will you keep this in mind and open a PR at some point?

src/main/java/org/jabref/gui/entryeditor/SourceTab.java
```diff
@@ -0,0 +1,249 @@
...
+            boolean emptyWarning = (newKey == null) || newKey.isEmpty();
+
+            if (newKey != null) {
+                entry.setCiteKey(newKey);
+            } else {
+                entry.clearCiteKey();
+            }
+
+            // First, remove fields that the user has removed.
+            for (Map.Entry<String, String> field : entry.getFieldMap().entrySet()) {
```
Siedlerchr: You could rewrite this to streams, using entrySet().stream()

src/main/java/org/jabref/gui/entryeditor/SourceTab.java
```diff
@@ -0,0 +1,249 @@
...
+                if (InternalBibtexFields.isDisplayableField(fieldName) && !newEntry.hasField(fieldName)) {
+                    compound.addEdit(
+                            new UndoableFieldChange(entry, fieldName, fieldValue, null));
+                    entry.clearField(fieldName);
+                    entryChanged = true;
+                }
+            }
+
+            // Then set all fields that have been set by the user.
+            for (Map.Entry<String, String> field : newEntry.getFieldMap().entrySet()) {
```
Siedlerchr: Same as aboe

src/main/java/org/jabref/gui/entryeditor/SourceTab.java
tobiasdiez: Agreed. The entry editor specific key bindings are now activated again and they work as expected. There are still issues with the key bindings, especially when it comes to more global ones - but this is already tracked as https://github.com/JabRef/jabref/issues/2855.

src/main/java/org/jabref/gui/contentselector/ContentSelectorDialog.java
tobiasdiez: I have this in mind. I plan to implement content selectors along with auto-completion in one of the next PRs (as soon as the big problems are resolved).

src/main/java/org/jabref/gui/entryeditor/SourceTab.java
tobiasdiez: You mean as entrySet().stream().foreach(...) ? I really like the old-style for loop more when it comes to a complicated body and find foreach only preferable when calling one method.

Siedlerchr: No, I thought about directly applying the 

> internalBibtexFields.isDisplayableField(fieldName) && !newEntry.hasField(fieldName)
as filter option in the stream and then applying anyMatch()
So: 

entry.getFieldMap().entrySet().stream().filter(field->InternalBibtexFields.isDisplayableField(field.gett )
https://www.mkyong.com/java8/java-8-filter-a-map-examples/

But now I see it may not make sense

tobiasdiez: Ah ok, that would be possible yes. Since this is only a slight improvement, I will leave the for-loop for now.

Issue #2893, opened 2017-06-07
Localization el (Greek) menu is not updated


tobiasdiez:
The `Menu_el.properties` file is not in sync with the other language files (e.g. it still contains `database` instead of `library`). Running the localizationUpdate gradle task results in some updates.
Moreover, the localization tests should be fixed and report this issue.

Siedlerchr:
Refs  #2835

PR #2895, opened 2017-06-08
Update Localization of Greek Translations

lynyus:
fixes https://github.com/JabRef/jabref/issues/2893.

### DIFF
--- a/src/main/java/org/jabref/logic/l10n/Languages.java
+++ b/src/main/java/org/jabref/logic/l10n/Languages.java
@@ -14,23 +14,24 @@ public class Languages {
         LANGUAGES = new TreeMap<>();
 
         // LANGUAGES contains mappings for supported languages.
+        LANGUAGES.put("Bahasa Indonesia", "in");
+        LANGUAGES.put("Brazilian Portuguese", "pt_BR");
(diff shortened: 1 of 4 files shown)


src/main/java/org/jabref/logic/l10n/Languages.java
```diff
@@ -14,23 +14,24 @@
...
         LANGUAGES.put("Norsk", "no");
-        LANGUAGES.put("Brazilian Portuguese", "pt_BR");
+        LANGUAGES.put("Persian (فارسی)", "fa");
         LANGUAGES.put("Russian", "ru");
+        LANGUAGES.put("Simplified Chinese", "zh");
         LANGUAGES.put("Svenska", "sv");
         LANGUAGES.put("Turkish", "tr");
         LANGUAGES.put("Vietnamese", "vi");
-        LANGUAGES.put("Simplified Chinese", "zh");
+        LANGUAGES.put("ελληνικά", "el");
```
Siedlerchr: Would it make sense to add it as Greek(...) or you remove the non-english name to stay consistent with Chinese, Japanese...

lynyus:
give me a sec before merging, I am just writing another test to avoid this issue in future

Issue #2274, opened 2016-11-15
2 links in file when autolink with RegEx and literature in subfolder


Schnet:
JabRef 3.7-dev--snapshot--2016-11-10--master--2441675
windows 7 6.1 amd64 
Java 1.8.0_111

Steps to reproduce:

1. BibTeX-File in a folder (e.g. /root) and literature in a subfolder (e.g. /root/literature)
2. names of files in /root/literature in the form "bibtexkey_*.pdf"
2. in JabRef perferences: 
    a. "Main file directory = /root/literature"
    b. "Use regular expression search = **/([bibtexkey]_).*\\.[extension]
3. Autolink yields "file = {bibtexkey_filename.pdf:literature\\bibtexkey_filename.pdf:PDF;bibtexkey_filename.pdf:root\\literature\\bibtexkey_filename.pdf:PDF},

lenhard:
@Schnet: Thanks for your report. In my point of view, this is a little bit of a edge case. The main file directory is a global setting that applies to all bib files. For this problem to matter, either you have just one bib file ever, or you store all your bib files in the same folder. Both ways seem strange to me, but I guess: different people, different workflows. Typically, I would expect one global file directory to which bib files from all over the system point to. I guess this thinking is why we did not run into the problem before.

Anyway, there should of course just be one entry if this is the exact same pdf. Hence, we should have a look at the duplicate detection in the autolink action. Since the current behavior not shoot down JabRef (both links should work), but just duplicates some information, I will categorize it as an enhancement issue and not as a bug.

Siedlerchr:
@tilsche submitted a PR which maybe fixed this bug.

We would like to ask you to use a development build from https://builds.jabref.org/master and report back if it works for you.

PR #2914, opened 2017-06-15
[WIP] Avoid duplicate results for recursive search in RegExpBasedFileFinder

tilsche:
The `RegExpBasedFileFinder` found the exact same files multiple times because 1) "**" recursively walks through subdirectories and 2) `Files.find` was called with a high `maxDepth`. So the file was found once from the root directory, and again from it's directory.

The suggested change addresses this by not recursively searching with Files.find. I presume that it was not intended that search was recursive without `**`. This may be related to #2274.


Siedlerchr:
If you look into the test classes, there is already one method for findFiles in the RegExp, so you can take this as a starting point. If you have any further questions, feel free to ask.
And many thanks for your contribution 👍

tilsche:
Thanks for the feedback. I added two tests, both failed without the change, but succeed now.

Siedlerchr:
Thank you very much for your contribution! I merge it in now!

PR #2926, opened 2017-06-19
Better icon support for groups

Siedlerchr:
Parse enum icon value from icon field by name (and apply to uppercase)
Only minor caveat: When you edit a group, you see only a unicode-box and not the name)

No valid icon: (the illegal is just a simple string)

groupsnovalidicon

Valid icon name: `facebook`
jabrefgroupwithicon

Icon Enum based on Material Design Icons v1.7.22

### DIFF
--- a/src/main/java/org/jabref/gui/IconTheme.java
+++ b/src/main/java/org/jabref/gui/IconTheme.java
@@ -292,6 +292,14 @@ public class IconTheme {
             return new FontBasedIcon(this.unicode, this.color);
         }
 
+        public List<MaterialDesignIcon> getUnderlyingIcons() {
+            return icons;
+        }
+
+        public MaterialDesignIcon getUnderlyingIcon() {
+            return icons.get(0);
+        }
+
         public FontBasedIcon getSmallIcon() {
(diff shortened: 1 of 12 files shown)


lenhard:
Unsurprisingly, I am very much opposed to this. We are reverting many structural improvements after all the effort we invested into getting the GUI dependencies out of the model.

I already grudgingly admitted the `iconCode` into the model, but I won't do the same with a `MaterialDesignIcon`. It's not too hard to extract a class into gui that maps `iconCode` to `MaterialDesignIcon` for the groups. I did something similar when I refactored the special fields. Please go for a similar route here.

tobiasdiez:
In this case we should store (and save in the bibfile) the icon not as the unicode but using the icon name.

Siedlerchr:
I moved the code to the ViewModel in GUI and it is called on getIconCode().
The icon name e.g. `facebook `or whatever you enter is stored in the IconCode field and now also displayed when you edit the group description

tobiasdiez:
Thanks. I did some small further changes so that now the icon name is used instead of the unicode (I think this is results in a more readable/compatible serialization in the bib file and I doubt somebody already managed to specify an icon :-) )

PR #2934, opened 2017-06-22
Page prefix

jhshinn:
This PR adds a new special field marker for BibTeX key generator [pageprefix]. It returns the non-digit prefix of `pages` (like "L" of L7) or "" if no non-digit prefix is found in `pages`. Some journals use non-digit letters for their pages, e.g. http://iopscience.iop.org/journal/2041-8205. (I have run `gradlew check` and there was no problem.)

### DIFF
--- a/src/main/java/org/jabref/logic/bibtexkeypattern/BibtexKeyPatternUtil.java
+++ b/src/main/java/org/jabref/logic/bibtexkeypattern/BibtexKeyPatternUtil.java
@@ -644,6 +644,8 @@ public class BibtexKeyPatternUtil {
                 }
             } else if ("firstpage".equals(val)) {
                 return firstPage(entry.getField(FieldName.PAGES).orElse(""));
+            } else if ("pageprefix".equals(val)) {
+                return Pageprefix(entry.getField(FieldName.PAGES).orElse(""));
             } else if ("lastpage".equals(val)) {
                 return lastPage(entry.getField(FieldName.PAGES).orElse(""));
             } else if ("title".equals(val)) {
@@ -1301,6 +1303,26 @@ public class BibtexKeyPatternUtil {
             return "";
         } else {
             return String.valueOf(result);
+        }
+    }
+
+    /**
+     * Return the non-digit prefix of pages
+     *
+     * @param pages
+     *            a pages string such as L42--111 or L7,41,73--97 or L43+
+     *
+     * @return the non-digit prefix of pages (like "L" of L7)
+     *         or "" if no non-digit prefix is found in the string
+     *
+     * @throws NullPointerException
+     *             if pages is null.
+     */
+    public static String Pageprefix(String pages) {
+        if (pages.matches("^\\D+.*$")) {
+            return (pages.split("\\d+"))[0];
+        } else {
+            return "";
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/jabref/logic/bibtexkeypattern/BibtexKeyPatternUtil.java
```diff
@@ -1305,6 +1307,26 @@ public static String firstPage(String pages) {
...
+     * @param pages
+     *            a pages string such as L42--111 or L7,41,73--97 or L43+
+     *
+     * @return the non-digit prefix of pages (like "L" of L7)
+     *         or "" if no non-digit prefix is found in the string
+     *
+     * @throws NullPointerException
+     *             if pages is null.
+     */
+    public static String Pageprefix(String pages) {
```
tobiasdiez: Please rename this method to `pagePrefix` and make it private.

koppor:
Dear @jhshinn. Also thank you from my side for your contribution. We are aware that writing tests is not as much fun as writing new code. During the development of JabRef we learned that tests are really vital: If we modernize the code, tests ensure that all functionality remains correctly in place. Further, this is some sort of documentation to enable other developers learning how the code works.

So, please, please, follow @tobiasdiez's request to add test cases. Would that be possible?

jhshinn:
Dear @koppor. Thanks a lot for your detailed exlanation. I've been busy recently, and had no time to deal with the requested revisions. I'm a newbie in java coding and used to be a far-far-end-user of JabRef. I've never done the *test* coding before, but I'll try to do it. When I need some help, I'll post again.

src/main/java/org/jabref/logic/bibtexkeypattern/BibtexKeyPatternUtil.java
jhshinn: @tobiasdiez Thanks a lot for your comments. I've revised the file as commented above.

jhshinn:
@tobiasdiez @koppor I've added a test for pagePrefix. Could you check if it is ok?

src/main/java/org/jabref/logic/bibtexkeypattern/BibtexKeyPatternUtil.java
```diff
@@ -1305,6 +1307,26 @@ public static String firstPage(String pages) {
...
+     *            a pages string such as L42--111 or L7,41,73--97 or L43+
+     *
+     * @return the non-digit prefix of pages (like "L" of L7)
+     *         or "" if no non-digit prefix is found in the string
+     *
+     * @throws NullPointerException
+     *             if pages is null.
+     */
+    public static String pagePrefix(String pages) {
+        if (pages.matches("^\\D+.*$")) {
```
Siedlerchr: It is better if you extract the regex to a constant pattern, by using Pattern.compile  , have a look at the titleChecker (logic.integrity package ) class, to see how it's done.
This construct compiles the regex and makes it faster when it is called multiple times

tobiasdiez:
Perfect, thanks! I'll merge it now.

src/main/java/org/jabref/logic/bibtexkeypattern/BibtexKeyPatternUtil.java
jhshinn: @Siedlerchr Thanks a lot for the comment. It would be very useful to me. I've found the same comment [here](https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html).

> though for repeated matches it is less efficient since it does not allow the compiled pattern to be reused.

PR #2939, opened 2017-06-24
Ask before deleting associated files from disk in entry

### LINKED ISSUE #2938: Deleting a file in the entry editor deletes the file on disk
This should not happen.

By default in the old entry editor,the file never got deleted in the filesystem
So we should add a second confirmation dialog

Siedlerchr:
Add new dialog

Fixes #2938 
grafik


Siedlerchr:
We can change the text always laster, but I suggest we leave it for the moment to have the functionality

tobiasdiez:
I just noticed that in the right click menu it is still `Permanently delete the local file` which is kind of confusing as you later get the option to decide wheter to delete the file or just the link. So to options:
a) provide two context menus for "Unlink" and "Delete local file" but not the dialog you added in this PR or
b) rename also the context menu entry

Can you please have a look and provide a follow-up PR? Thanks!

Siedlerchr:
Yes, but the dialog is still necessary when you hit the del key


Am 29.06.2017 1:33 nachm. schrieb "Tobias Diez" <notifications@github.com>:

PR #2943, opened 2017-06-28
Add a shortcut to "Quality - Look up full text documents"

jhshinn:
This PR adds a shortcut key <kbd>Alt</kbd> + <kbd>F7</kbd>  to the function "Quality"-"Look up full text documents". I've found that a issue was already raised in #1323 after finishing the code revision (hence, I happened to use the modifier <kbd>Alt</kbd> rather than <kbd>Ctrl</kbd> :-) )

### DIFF
--- a/src/main/java/org/jabref/gui/JabRefFrame.java
+++ b/src/main/java/org/jabref/gui/JabRefFrame.java
@@ -405,7 +405,7 @@ public class JabRefFrame extends JFrame implements OutputPrinter {
             IconTheme.JabRefIcon.MERGE_ENTRIES.getIcon());
     private final AbstractAction downloadFullText = new GeneralAction(Actions.DOWNLOAD_FULL_TEXT,
             Localization.menuTitle("Look up full text documents"),
-            Localization.lang("Look up full text documents"));
+            Globals.getKeyPrefs().getKey(KeyBinding.DOWNLOAD_FULL_TEXT));
     private final AbstractAction increaseFontSize = new IncreaseTableFontSizeAction();
     private final AbstractAction defaultFontSize = new DefaultTableFontSizeAction();
     private final AbstractAction decreseFontSize = new DecreaseTableFontSizeAction();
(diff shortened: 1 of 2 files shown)


src/main/java/org/jabref/gui/keyboard/KeyBinding.java
```diff
@@ -25,6 +25,7 @@
     CUT("Cut", Localization.lang("Cut"), "ctrl+X", KeyBindingCategory.EDIT),
     DECREASE_TABLE_FONT_SIZE("Decrease table font size", Localization.lang("Decrease table font size"), "ctrl+MINUS", KeyBindingCategory.VIEW),
     DELETE_ENTRY("Delete entry", Localization.lang("Delete entry"), "DELETE", KeyBindingCategory.BIBTEX),
+    DOWNLOAD_FULL_TEXT("Look up full text documents", Localization.lang("Look up full text documents"), "alt F7", KeyBindingCategory.QUALITY),
```
Siedlerchr: As a result of fixing the keybindind stuff, there now needs to be a plus sign between the modifiers and the key, otherwise it may not be recognized correctly.
Otherwise the code looks good, thanks for your contribution

jhshinn: @Siedlerchr Thanks for the correction~ I've added the sign.

src/main/java/org/jabref/gui/JabRefFrame.java
```diff
@@ -1434,7 +1434,9 @@ private void initActions() {
         twoEntriesOnlyActions.addAll(Arrays.asList(mergeEntries));
 
         atLeastOneEntryActions.clear();
-        atLeastOneEntryActions.addAll(Arrays.asList(downloadFullText, lookupIdentifiers));
+        atLeastOneEntryActions.addAll(Arrays.asList(downloadFullText));
+        atLeastOneEntryActions.clear();
```
tobiasdiez: I don't understand these changes: you first add the download action but then clear the list right after.

jhshinn: @tobiasdiez sorry for the confusion and thanks for the note. I misunderstood something while doing try and error. I've reverted the changes.

tobiasdiez:
I've no idea why the build fails, but since these changes should not be the reason for the failing test I'll merge it now.

PR #2952, opened 2017-06-30
Fix the function "Edit - Copy BibTeX key and link" 

jhshinn:
This PR fixes the function "Edit - Copy BibTeX key and link"  to pass a hyperlink rather than an HTML statement. A copied hyperlink is more useful than an HTML statement in practice, like when writing emails. This is related with [#549](https://github.com/JabRef/jabref/issues/549).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - Links in the Recommended Articles tab (Mr.DLib), when clicked, are now opened in the system's default browser. [2931](https://github.com/JabRef/jabref/issues/2931)
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


tobiasdiez:
LGTM. I just removed the comment that contained the non-working code (we had a discussion about such code and decided to almost always remove them as JabRef was cluttered with these kind of comments).

We should think about reimplementing the `ClipboardManager` using the JavaFX facilities. Follow-up PR in this direction is highly welcome :-).

Thanks again for you contribution!

jhshinn:
@tobiasdiez Thanks a lot for telling me the configuration file. I'm not sure if I did it correctly, because the only change was an addition of a blank line (-_-;). If something has to be done again, please let me know.

tobiasdiez:
Everything is fine now. Thanks!

Issue #2949, opened 2017-06-28
Pressing Escape does not close entry editor


tobiasdiez:

Siedlerchr:
Could be related to my last changes, we probably just need to add a Check
for the escape key

Am 28.06.2017 3:54 nachm. schrieb "Tobias Diez" <notifications@github.com>:

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/JabRef/jabref/issues/2949>, or mute the thread

koppor:
Somehow related:

- [ ] <kbd>Ctrl</kbd>+<kbd>S</kbd> does not work. Neither when focused in the entry nor when 
- [ ] Star for modified database not shown if I change something in the entry editor

Siedlerchr:
This does not work when you are in an entry editor field.

Siedlerchr:
Escape now works. For save, we first need to create a save action in the entry editor. I will create a follow up PR

tobiasdiez:
The save command shouldn't be handled by the entry editor but by a higher instance (probably JabRefMain).

Siedlerchr:
Yes, but the problem is that it is not that easy to filter out the command from the JFX panel, because we would need to map between awt keyEvents and javafx keyEvents/KeyCombination. Unfortuately there is no easy way to do this.
So I would suggest creating a save action analogous to the close action in EntryEditor/ setupKeyBindings

tobiasdiez:
This should be fixed in the latest development version. Could you please check the build from http://builds.jabref.org/master/. Thanks!

koppor:
Since two years, this does not work. At https://github.com/JabRef/jabref/pull/11190#issuecomment-2054006956, we decided to stick with <kbd>Ctrl</kbd>+<kbd>E</kbd> to open/close the entry editor.

PR #2957, opened 2017-07-02
Fix closing entry editor with ESC

Siedlerchr:
Rename Escape to ESC for javafx
Renamed Enum variables as key for the constant was amibiuous

First part of #2949 
Internal change only, from the new entry editor.

Edit// You may need to reset the keybinding for CLOSE_ENTRY_EDITOR in the Manage keybinding dialog, as the old stored value from the prefs no longer works for javafx.


src/main/java/org/jabref/gui/keyboard/KeyBinding.java
```diff
@@ -93,27 +93,35 @@
...
         this.category = category;
     }
 
-    public String getKey() {
-        return key;
+    /**
+     * This method returns the enum constant value
+     * @return
+     */
+    public String getConstant() {
```
tobiasdiez: Why do we need this "constant/key" as a unique identifier if we already have an enum?

Siedlerchr: This constant is just the name of the enum value. It is used in several cases to compare bindings and to map between the settings and the enum values dynamically. It is used in KeyBindingsRepositoriy 
I just renamed it, because "key" is ambigous in this content and always confused me.

PR #2959, opened 2017-07-03
In JUnit tests, always state the expected value before the actual

lynyus:
This is only an internal change, no functionality was actually changed. I thought it would be good to open a PR anyways to raise awareness.


Siedlerchr:
Ah good one!

Siedlerchr:
LGTM, no idea why freecite fails

PR #2965, opened 2017-07-04
[WIP] Complete rework of the auto completion

tobiasdiez:
This PR reenables the auto completion in the entry editor. It is more or less a complete rewrite of this feature using JavaFX techniques. Now a dropdown box shows the suggestions instead of the previously used inline overwrite.

Current status: the auto completion in the search bar does not work and the code needs further cleanup. However, in the entry editor it should be feature complete. It would be nice if you could test it and give early feedback. Thanks.


Siedlerchr:
codewise it looks good so far. But I have not yet tested it

lenhard:
If played around a little with a running JabRef. When I enabled the autocompletion in the preferences, first my entry editor seemed to have crashed. Clicking around in it had no effect and eventually it closed. When I reopened it, the autocompletion was working fine. So that doesn't seem to be a blocking issue.

I have to say that I quite like the dropdowns :) This looks way better than the old autocompletion functionality! But before this goes life, the search should be fixed as well. When I start entering terms there (for example the letter `N` in some bibliography of mine) it is giving me expressions of the form `Author{firstPart='Norman E.', firstAbbr='N. E.', vonPart='null', lastPart='Fenton', jrPart='null'}`.

tobiasdiez:
Thanks for the feedback.
The search bar now should work again. Since I don't expect that someone has the desire to review the 3k line differences, I'll merge this PR now. However, as usually, if you find some problematic code I commit to fix it.

PR #2978, opened 2017-07-07
Fix group storage

### LINKED ISSUE #2932: Group information in new database not stored
Current master.

I debugged a bit through and found that in the ` MetaDataSerializer.getSerializedStringMap ` the `metaData.getGroups() `returns null/empty optional. In the metadata class the `groupsRoot ` is null.

Furthermore I found out that `setGroups()` in `Metadata `is never called (except from the importer and some tests)

lenhard:
Fixes #2932

It seems that up to now all group related changes in the UI were not actually stored in the MetaData and lost on save. This PR resets the group-related information in the MetaData whenever something group-related changes.

Hopefully I found all positions where that may happen. If tested it manually, but someone else needs to do this, too!

It also seems that IntelliJ decided to remove trailing whitespaces in the Changelog when I added my entry...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,10 +11,10 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 ## [Unreleased]
 
 ### Changed
-- We moved the `adsurl` field to `url` field when fetching with the ADS fetcher. 
+- We moved the `adsurl` field to `url` field when fetching with the ADS fetcher.
 - We continued to improve the new groups interface:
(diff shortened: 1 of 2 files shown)


tobiasdiez:
I don't have time right now to have a closer look, so just a very quick remark/question: 
I think that up to now changes were actually stored correctly. The reference to the root group was passed from the metadata to the group view model and thus changes from the later class should propagate back and update the metadata automatically. But if you create a new database, then the root was actually null (instead of empty) and thus the whole procedure broke down.

So I'm not sure if this explicit update is really necessary. But if its works we can also just merge it :-). Thanks anyway for having a look at this issue!

Siedlerchr:
Tested it locally and now the group information is stored when you create a new db, Importing works also.

PR #2991, opened 2017-07-10
Check for different editions in the duplicate check

### LINKED ISSUE #2960: Duplicate finder ignores book edition
My bibtex file contains two editions of a book, but the "find duplicates" function considers them to be duplicates and tries to merge them.

Also, shouldn't the jabref gitup bug reporter accept bibtex files for uploading?  As it is, it's necessary to rename them to *.txt.

JabRef version 

JabRef 4.0.0-dev--snapshot--2017-06-29--master--7e50e84a1
Windows 7 6.1 amd64 
Java 1.8.0_131

Steps to reproduce:

1. Load the attached bib file
2. Quality--> Find Duplicates
3. A window should pop up with a merge suggestion

[example2editionBooks.txt](https://github.com/JabRef/jabref/files/1119705/example2editionBooks.txt)

lynyus:
Fixes #2960 by checking for different editions before the actual duplicate check.
If both BibEntries have a Edition field that mismatches, it cannot be a duplicate.

### DIFF
--- a/src/main/java/org/jabref/logic/bibtex/DuplicateCheck.java
+++ b/src/main/java/org/jabref/logic/bibtex/DuplicateCheck.java
@@ -26,7 +26,7 @@ import org.apache.commons.logging.LogFactory;
  * This class contains utility method for duplicate checking of entries.
  */
 public class DuplicateCheck {
-    public static double duplicateThreshold = 0.75; // The overall threshold to signal a duplicate pair
+    private static final double duplicateThreshold = 0.75; // The overall threshold to signal a duplicate pair
 
     private static final Log LOGGER = LogFactory.getLog(DuplicateCheck.class);
     /*
@@ -54,7 +54,8 @@ public class DuplicateCheck {
         DuplicateCheck.FIELD_WEIGHTS.put(FieldName.JOURNAL, 2.);
     }
 
-    private DuplicateCheck() { }
+    private DuplicateCheck() {
+    }
 
     /**
      * Checks if the two entries represent the same publication.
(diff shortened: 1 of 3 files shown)


src/main/java/org/jabref/logic/bibtex/DuplicateCheck.java
```diff
@@ -98,7 +102,16 @@ public static boolean isDuplicate(BibEntry one, BibEntry two, BibDatabaseMode bi
         return req[0] >= DuplicateCheck.duplicateThreshold;
     }
 
-    private static boolean hasSameIdentifier(BibEntry one, BibEntry two) {
+    private static boolean haveDifferentEditions(BibEntry one, BibEntry two) {
+        if (one.getField(FieldName.EDITION).isPresent() && two.getField(FieldName.EDITION).isPresent()) {
+            if (!one.getField(FieldName.EDITION).get().equals(two.getField(FieldName.EDITION).get())) {
```
Siedlerchr: You can direclty compare Optionals with equals, but be aware that two empty optionals also equal.
https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#equals-java.lang.Object-

src/main/java/org/jabref/logic/bibtex/DuplicateCheck.java
```diff
@@ -26,7 +26,7 @@
  * This class contains utility method for duplicate checking of entries.
  */
 public class DuplicateCheck {
-    public static double duplicateThreshold = 0.75; // The overall threshold to signal a duplicate pair
+    private static final double duplicateThreshold = 0.75; // The overall threshold to signal a duplicate pair
```
Siedlerchr: Just a very minor thing, the constant variable should be uppercase

src/main/java/org/jabref/logic/bibtex/DuplicateCheck.java
lynyus: Yeah, I'm aware of that. That's why I first checked both to be present, and then called `.get()` on both to explicitly state that this line will only work if there is a value in the optional.

Siedlerchr: Okay, good! Then it's fine

PR #2994, opened 2017-07-11
 Shutdown previus AutosaveManager and BackupManager during SaveAs 

### LINKED ISSUE #2947: Save as doesnt remove backup file
JabRef version 4.0 

Steps to reproduce:

1. Save as new lib
2. Open original library - the user is asked if he want to recover library from backup file.

As I could see in the folder still exist file bib.sav from original library.

mpele:
Hi, 
Could anyone check is everything done correctly in this simple bug fix #2947.


lenhard:
Hi @mpele 

Thanks for your take on this! Unfortunately, the current version will not do. If you shutdown the AutoSaveManager and the BackupManager, then there won't be any autosaves or backups anymore and this is not what we want here.

As far as I can see, the way to go would be the following:
 1. Check if the save as operation actually uses a different file path than the old one
 2. If this is true and the save operation is successful, remove the old .sav and .bak files explicitly

mpele:
@lenhard, as I tested it works. After SaveAs new .sav file is created (AutosaveManager and BackupManager are restarted later in lines 351 and 356).

lenhard:
I see, my bad! It sure makes sense to disable the two managers for the old file location. 

But the problem that remains is what if the save operation fails? Then, autosave and backup will no longer work for the old file. I still think it would be better to have a cleanup phase after the safe operation has been successful, i.e., shutting done the obsolete managers and removing any unnecessary files.

mpele:
Hi @lenhard,
Is the proposed code acceptable? Right now the shutdown of AutosaveManager and BackupManager is done if the save operation is successful.

lenhard:
I've gone through #2947 and the code seems to solve the problem: The sav file is gone and the user is no longer asked if he wants to restore a backup when opening the old file. 

On looking at the code of BackupManager and AutosaveManager, I am a little unsure why we restart them to begin with, but that is not your problem here. So from my side the code is good to go. 

We have the policy that two of the developers need to review every pull request before it is merged. So could somebody else from @JabRef/developers have a look at this?

src/main/java/org/jabref/gui/exporter/SaveDatabaseAction.java
```diff
@@ -316,7 +316,8 @@ public void saveAs() throws Exception {
      */
     public void saveAs(File file) throws Exception {
         BibDatabaseContext context = panel.getBibDatabaseContext();
-
+        Path oldFile = context.getDatabasePath().get();
```
Siedlerchr: Not your fault: I always wondered why this is an Optional, I think in this case I see no reason why it could be empty, so you are probably on the save side calling directly get.

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - We improved the duplicate checker such that different editions of the same publication are not marked as duplicates. [2960](https://github.com/JabRef/jabref/issues/2960)
 
 ### Fixed
+- We fixed the a bug that leaves .sav file after SaveAs [#2947]
```
Siedlerchr: A little typo: fixed a bug  and please inlcude the URL behind like we do with other entries

src/main/java/org/jabref/gui/exporter/SaveDatabaseAction.java
tobiasdiez: I think you don't have a database path if the library comes from a SQL server.

PR #2997, opened 2017-07-11
Changes in the entry editor mark the database as dirty

lenhard:
Partial fix to  #2787 

Changes in the entry editor now mark the database as dirty. This are somewhat proactive: even only opening the entry editor will mark the database as changed. I'd say this isn't a problem. I'd rather have it do the marking than forget one.

Essentially, this adds another listener to the properties in the entry editor that propagates the change whenever the property changes. I haven't figured out how to do this for groups.

### DIFF
--- a/src/main/java/org/jabref/gui/entryeditor/FieldsEditorTab.java
+++ b/src/main/java/org/jabref/gui/entryeditor/FieldsEditorTab.java
@@ -9,6 +9,7 @@ import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Stream;
 
+import javafx.beans.property.BooleanProperty;
 import javafx.geometry.VPos;
 import javafx.scene.Node;
 import javafx.scene.Parent;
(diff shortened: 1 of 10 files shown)


stefan-kolb:
I'd say this is a hack. Whereas it is better than missing a changed state for now, we should only mark the DB as dirty if we really changed something.

tobiasdiez:
I'm also not very happy with this solution from an architectural view: why should the gui handle the dirty/clean state and not the logic? I would propose to add listeners to the database/metadata in `DatabaseContext` which sets a corresponding flag. The UI can then listen to changes of this new property and display the star in the tab accordingly.

Siedlerchr:
The databaseChangedEvent is already realized with the EventBus system, so it just needs a couple of listeners if I understand it correctly. And one event bus can have multiple listeners for the same event

lenhard:
Ok, I see. So far it had been implemented only in the BasePanel, but I aggree that the cleanest solution is probably to track it in the database itself. This won't be as quick as the current solution, though :)

The consensus seems to be that we are not going in this direction, so I'll close this PR.

PR #3000, opened 2017-07-12
Fix mac keybinding by replacing ctrl it with meta key

Siedlerchr:
This should make the entry editor commands work again with Command+X /C etc

@chochreiner  Could you please test if that works now? Copy/Cut/Paste in the entry editor with Command key instead of control key.

Unfortunately I could not add a unit test, because the OS is a static class with final vars.


chochreiner:
Copy/Cut/Paste works fine with the command key 👍

PR #3001, opened 2017-07-12
Listen to change events for setting dirty status of database

### LINKED ISSUE #2787: Changes to groups and edits in entry editor no longer mark database as dirty

lenhard:
New attempt to fix #2787 and follow up to #2997 (much simpler now)

Uses the event bus feature. Now, the `BasePanel` simply registers as a listener to the event buses of its `BibDatabase` and `MetaData`. The listener sets the status of the panel to dirty whenever a change event occurs.

This is the first time that I can see any benefit in our event bus system :)

I've tested this manually in a running instance, but somebody else should please test too!


src/main/java/org/jabref/gui/BasePanel.java
```diff
@@ -204,6 +204,9 @@ public BasePanel(JabRefFrame frame, BibDatabaseContext bibDatabaseContext) {
         Objects.requireNonNull(bibDatabaseContext);
 
         this.bibDatabaseContext = bibDatabaseContext;
+        DatabaseChangeListener databaseChangeListener = new DatabaseChangeListener(this);
```
Siedlerchr: Why don't you just do registerListener(this)  and then do this? No need to create a new listener?

```java
@Subscribe 
public void listen(BibDatabaseContextChangedEvent event)
{
basePanel.markBaseChanged();
}
```

lenhard: Thanks, you are right! I started out with the intention of not making the BasePanel any bigger than necessary. That's why I added a listener class. But since this class turned out to be so dead simple, it can just as well go into BasePanel. 

I fixed this in the way you suggested.

PR #3003, opened 2017-07-12
Reimplement content selectors

tobiasdiez:
This PR adds the content selector back. They are not implemented as a dropdown option (as it was previously) but added to the list of auto completion suggestions. Viewing "content selectors" as predefined autocompletion options seemed more natural to me.

@koppor 
In which order should I put the following lines to adhere to the style conventions but at the same time don't introduce illegal forward references?
````java
   private final JScrollPane wPane = new JScrollPane(wordList);
   private final JList<String> wordList = new JList<>(wordListModel);
   private DefaultListModel<String> wordListModel = new DefaultListModel<>();
````


lenhard:
I've tested it locally and like it! I was skeptical first as to how well this will actually work with autocompletion, but the solution is good. The configured selectors are directly offered the instant I click into the field in the entry editor. In my point of view, this is a full replacement for a drop down menu and way more handy.

Nevertheless, we will have to explain that to the users when we publish 4.0 :) Otherwise, they will open JabRef, see no drop down box, and get angry. So this should definitely go into a blog post.

The code looks fine overall. This is a situation where FX really beats the old swing implementation. Good job overall!

tobiasdiez:
Thanks for the praising words 😄

PR #3005, opened 2017-07-13
javafx replacement for file dialog

### LINKED ISSUE #2970: Resizing of the window "Select files" does not scale properly
Windows should be scalable and move the buttons intelligent within a given range.
Scaling should be limited to the minimum/maximum size.

selection_082
selection_083
selection_084

```
JabRef 4.0.0-dev--snapshot--2017-07-03--master--6897b10fb
Linux 4.4.0-78-generic amd64 
Java 1.8.0_131
```

Siedlerchr:
Fixes #2970 

The open functionality has been removed -> Directly possible in the entry editor
Contextmenu on maintable ->attach file now opens a file chooser dialog and afterwards this dialog for editing, e.g. adding a description

grafik

Only resizable to max screen size: (on my 24" with 1920x1200)
grafik


src/main/java/org/jabref/gui/filelist/FileListDialogController.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    private void checkExtension() {
+        if (cmbFileType.getSelectionModel().isEmpty() && (!tfLink.getText().trim().isEmpty())) {
+
+            // Check if this looks like a remote link:
+            if (REMOTE_LINK_PATTERN.matcher(tfLink.getText()).matches()) {
+                Optional<ExternalFileType> type = ExternalFileTypes.getInstance().getExternalFileTypeByExt("html");
+                if (type.isPresent()) {
+                    cmbFileType.getSelectionModel().select(type.get());
```
Siedlerchr: @tobiasdiez 
I would like to move this nethod and some others to the view Model, but I don't know how/or if I should expose the **selection model** of the combobox to the view model.

src/main/java/org/jabref/gui/filelist/FileListDialogController.java
```diff
@@ -0,0 +1,157 @@
...
+    @FXML
+    private void initialize() {
+        viewModel = new FileListDialogViewModel(stateManager.getActiveDatabase().get());
+        setBindings();
+
+    }
+
+    private void setBindings() {
+
+        cmbFileType.itemsProperty().bindBidirectional(viewModel.externalFileTypeProperty());
```
Siedlerchr: @tobiasdiez  Here the binding is created, a simple list property for the combobox

tobiasdiez: this looks good! (a one-directional binding should suffice I think since it looks like the `itemsProperty` is never changed directly.)

src/main/java/org/jabref/gui/filelist/FileListDialogController.java
tobiasdiez: You have a few options here:
1) Create a property in the view model and bind the selected item property to it, similar to the following code:
https://github.com/JabRef/jabref/blob/cbd6844183c15479ea2429d3962197e78b1494c5/src/main/java/org/jabref/gui/keyboard/KeyBindingsDialogController.java#L37-L41
But I'm not sure if using a bidirectional binding works here out of the box, probably you need to use https://github.com/JabRef/jabref/blob/bee715d03ab6fddcd48af5a5cf9e134ae458d521/src/main/java/org/jabref/gui/util/BindingsHelper.java#L75 or a similar helper method.

2) Just return the to-be-selected item from the method in the view model and then really select it here in the controller.

Option 1 is definitely the cleaner solution but maybe a bit harder to setup (but definitely worth the initial effort since a similar problem might appear in the future).

koppor:
Discussion @Siedlerchr with @tobiasdiez required.

src/main/java/org/jabref/gui/filelist/AttachFileAction.java
```diff
@@ -27,9 +24,14 @@ public void action() {
...
-        FileListEntryEditor editor = new FileListEntryEditor(flEntry, false, true,
+
+        FileListDialogView dlg = new FileListDialogView();
+
+        Platform.runLater(() -> dlg.show());
+        /*   FileListEntryEditor editor = new FileListEntryEditor(flEntry, false, true,
                 panel.getBibDatabaseContext());
         editor.setVisible(true, true);
+
         if (editor.okPressed()) {
```
Siedlerchr: @tobiasdiez  How do I handle such cases in javafx?  The check for a return value? Should I do a showAndWait?
 My initial idea would be to pass the UndoManager to the javafx view model (state manager? or other injection)
There are some other clases in which a similar stuff is executed which seems to a bit more complicated, e.g. the
DownloadExternalFile class

tobiasdiez: Good question. I think the return value should be accessed per getters of the `FileListDialogView` (this seems to be the most flexible and easiest solution - you don't need to create "Dialog return value classes" for more complex scenarios). These getters in the view should ask the view model of the dialog about the values. You can get the controller with 
https://github.com/JabRef/jabref/blob/1172ed493c6dce05fcb427b4e665f5de3a7b3fc3/src/main/java/org/jabref/gui/AbstractView.java#L45
(just change the visibility) and then the view model with
https://github.com/JabRef/jabref/blob/1172ed493c6dce05fcb427b4e665f5de3a7b3fc3/src/main/java/org/jabref/gui/AbstractController.java#L20

For parameter injection, have a look at `CopyFilesDialogView`.

tobiasdiez:
Dialog only for edit and not for adding linked files (for this: just create a new linked file instance with path to file and empty directory)

src/main/java/org/jabref/gui/fieldeditors/LinkedFileViewModel.java
```diff
@@ -326,8 +326,13 @@ public boolean delete(FileDirectoryPreferences prefs) {
     }
 
     public void edit() {
-        FileListEntryEditor editor = new FileListEntryEditor(linkedFile, false, true, databaseContext);
-        SwingUtilities.invokeLater(() -> editor.setVisible(true, false));
+        LinkedFilesWrapper wrapper = new LinkedFilesWrapper();
+        wrapper.setLinkedFile(this.linkedFile);
+        LinkedFileEditDialogView dialog = new LinkedFileEditDialogView(wrapper);
+        if (dialog.showAndWait()) {
+            // TODO: What do when saved?
```
Siedlerchr: @tobiasdiez  Do you have an idea how I can store the changes to the linkedFile I made in the dialog?
I can get back a new linkedFile object but no idea how to apply the changes to the orginal obj

tobiasdiez: It should suffice to use the setters of the old linked file (the changes should then automatically be stored in the entry; at least if the linked file is already shown in the entry editor).

Siedlerchr:
Failing tests are only the l10n ones from the maintable branch

src/main/java/org/jabref/gui/BasePanel.java
```diff
@@ -387,8 +386,8 @@ public void init() {
 
                 if (entries.isEmpty()) { // None selected. Inform the user to select entries first.
                     dialogService.showWarningDialogAndWait(
-                            Localization.lang("Autogenerate BibTeX keys"),
-                            Localization.lang("First select the entries you want keys to be generated for."));
+                    Localization.lang("Autogenerate BibTeX keys"),
```
tobiasdiez: I think, there is again something wrong with the eclipse code style since almost all formatting changes in this file feel wrong.

Siedlerchr: Yeah. I know. I am still figuring out how to get it right:
I am currently tweaking the behaviour a bit to not be so agressive
- Default indentation factor: I can only control how many spaces the new line will be indented, based on the Default space indentation (4 spaces)

tobiasdiez: I think, I've found a solution: https://www.jetbrains.com/idea/download/ 😈 .

Siedlerchr:
Yeah, I thought I could get rid of the old classes completely, but that has to wait for a future PR. As far as I remember it's used in the import inspection window (lot of ugly code)

Siedlerchr:
@tobiasdiez  I tried to convert the EditfileDialog to the new dialog structure from #3880  but I do still get an NPE on loading the dialog:
```
caused by: java.lang.NullPointerException
	at org.jabref.gui.filelist.LinkedFileEditDialogView.initialize(LinkedFileEditDialogView.java:55) ~[main/:?]
```

tobiasdiez:
And what is null there? If line 54 runs without any problems, then the setup seems to be fine and the FXML nodes are injected. Maybe the `description` field is not defined in the fxml?

Siedlerchr:
@tobiasdiez  No, it fails directly in the initalize, The viewModel which gets initialized in the ctor is null.
When debugging it crashes directly after/when calling `ViewLoader.view(this)
                  .load()`
Btw, I noticed the same problem also occurs with the CopyLinkedFilesDialogView (Tools->Copy LInked Files)

tobiasdiez:
ok, this makes sense since the `initialization` method is called during `ViewLoader.load` and at this point the view model is not yet created. Just move the `viewModel = ...` line to the beginning of the `initialization` method and it should work.

Siedlerchr:
Works now.  I just additionally  had to add a close/cancel button to the dialogPane,  because otherwise the dialog is not closeable: https://docs.oracle.com/javase/8/javafx/api/javafx/scene/control/Dialog.html

src/main/java/org/jabref/gui/filelist/LinkedFilesWrapper.java
```diff
@@ -0,0 +1,19 @@
+package org.jabref.gui.filelist;
+
+import org.jabref.model.entry.LinkedFile;
+
+/**
+ * Workaround to inject {@link LinkedFile} into javafx controller
```
tobiasdiez: This should no longer be necessary.

src/main/java/org/jabref/gui/filelist/AttachFileAction.java
```diff
@@ -2,43 +2,61 @@
...
+        FileDialogConfiguration fileDialogConfiguration = new FileDialogConfiguration.Builder()
+                                                                                               .withInitialDirectory(Globals.prefs.get(JabRefPreferences.WORKING_DIRECTORY))
+                                                                                               .build();
+
+        dialogService.showFileOpenDialog(fileDialogConfiguration).ifPresent(newFile -> {
+            LinkedFile newLinkedFile = new LinkedFile("", newFile.toString(), "");
+
+            LinkedFilesWrapper wrapper = new LinkedFilesWrapper();
+            wrapper.setLinkedFile(newLinkedFile);
+            LinkedFileEditDialogView dialog = new LinkedFileEditDialogView(wrapper);
```
tobiasdiez: Do we really want to show the linked file editor dialog here? In my experience, I just click "OK" anyway and would expect that this is the case for most users (recall how long it took until somebody complained that the description was not shown in the file list). At least, I would add a preferences that controls the appearance of the dialog.

src/main/java/org/jabref/gui/filelist/AttachFileAction.java
```diff
@@ -2,43 +2,61 @@
...
-                panel.markBaseChanged();
-            }
-        }
+
+        FileDialogConfiguration fileDialogConfiguration = new FileDialogConfiguration.Builder()
+                                                                                               .withInitialDirectory(Globals.prefs.get(JabRefPreferences.WORKING_DIRECTORY))
+                                                                                               .build();
+
+        dialogService.showFileOpenDialog(fileDialogConfiguration).ifPresent(newFile -> {
+            LinkedFile newLinkedFile = new LinkedFile("", newFile.toString(), "");
```
tobiasdiez: Shouldn't the file type be guessed here based on the file extension?

src/main/java/org/jabref/gui/filelist/AttachFileAction.java
```diff
@@ -2,43 +2,61 @@
...
+            wrapper.setLinkedFile(newLinkedFile);
+            LinkedFileEditDialogView dialog = new LinkedFileEditDialogView(wrapper);
+
+            Optional<LinkedFile> editedLinkeDfile = dialog.showAndWait();
+            editedLinkeDfile.ifPresent(file -> {
+
+                Optional<FieldChange> fieldChange = entry.addFile(file);
+
+                if (fieldChange.isPresent()) {
+                    UndoableFieldChange ce = new UndoableFieldChange(entry, FieldName.FILE,
```
tobiasdiez: There should be constructor of `UndoableFieldChange` that directly accepts a `FieldChange`. If not, please add one and use it here.

src/main/java/org/jabref/gui/filelist/LinkedFilesEditDialogViewModel.java
```diff
@@ -0,0 +1,126 @@
...
+import org.jabref.model.database.BibDatabaseContext;
+import org.jabref.model.entry.LinkedFile;
+import org.jabref.model.util.FileHelper;
+import org.jabref.preferences.JabRefPreferences;
+import org.jabref.preferences.PreferencesService;
+
+public class LinkedFilesEditDialogViewModel extends AbstractViewModel {
+
+    private static final Pattern REMOTE_LINK_PATTERN = Pattern.compile("[a-z]+://.*");
+    private final StringProperty linkProperty = new SimpleStringProperty("");
```
tobiasdiez: the convention is to not add the suffix `Property` here.

src/main/java/org/jabref/gui/filelist/LinkedFilesEditDialogViewModel.java
```diff
@@ -0,0 +1,126 @@
...
+
+    public StringProperty descriptionProperty() {
+        return descriptionProperty;
+    }
+
+    public ListProperty<ExternalFileType> externalFileTypeProperty() {
+        return externalfilesTypes;
+    }
+
+    public ObjectProperty<ExternalFileType> getSelectedExternalFileType() {
```
tobiasdiez: this method should be called `selectedExternalFileTypeProperty` to adhere to JavaFX conventions

src/main/java/org/jabref/gui/filelist/LinkedFilesEditDialogViewModel.java
```diff
@@ -0,0 +1,126 @@
...
+    private final ObjectProperty<ExternalFileType> selectedExternalFileType = new SimpleObjectProperty<>();
+    private final BibDatabaseContext database;
+    private final DialogService dialogService;
+    private final PreferencesService preferences;
+
+    public LinkedFilesEditDialogViewModel(LinkedFile linkedFile, BibDatabaseContext database, DialogService dialogService, PreferencesService preferences2) {
+        this.database = database;
+        this.dialogService = dialogService;
+        this.preferences = preferences2;
+        externalfilesTypes.set(FXCollections.observableArrayList(ExternalFileTypes.getInstance().getExternalFileTypeSelection()));
```
tobiasdiez: Please extract `ExternalFileTypes.getInstance()` as a dependency (this singleton is as bad as the other Globals).

src/main/java/org/jabref/gui/filelist/LinkedFilesEditDialogViewModel.java
```diff
@@ -0,0 +1,126 @@
...
+    public void openBrowseDialog() {
+        String fileText = linkProperty().get();
+
+        Optional<Path> file = FileHelper.expandFilename(database, fileText, preferences.getFileDirectoryPreferences());
+
+        Path workingDir = file.orElse(Paths.get(preferences.get(JabRefPreferences.WORKING_DIRECTORY)));
+        String fileName = Paths.get(fileText).getFileName().toString();
+
+        FileDialogConfiguration fileDialogConfiguration = new FileDialogConfiguration.Builder()
+        .withInitialDirectory(workingDir)
```
tobiasdiez: indentation is wrong

src/main/java/org/jabref/gui/filelist/LinkedFilesEditDialogViewModel.java
```diff
@@ -0,0 +1,126 @@
...
+
+    public LinkedFilesEditDialogViewModel(LinkedFile linkedFile, BibDatabaseContext database, DialogService dialogService, PreferencesService preferences2) {
+        this.database = database;
+        this.dialogService = dialogService;
+        this.preferences = preferences2;
+        externalfilesTypes.set(FXCollections.observableArrayList(ExternalFileTypes.getInstance().getExternalFileTypeSelection()));
+        setValues(linkedFile);
+    }
+
+    private void checkExtension() {
```
tobiasdiez: rename to `setExtension`? I would also add the link as a parameter to make it clear that the method guesses the extension based on the path.

src/main/java/org/jabref/gui/filelist/LinkedFilesEditDialogViewModel.java
```diff
@@ -0,0 +1,126 @@
...
+            // If the file is below the file directory, make the path relative:
+            List<Path> fileDirectories = database.getFileDirectoriesAsPaths(preferences.getFileDirectoryPreferences());
+            path = FileUtil.shortenFileName(path, fileDirectories);
+
+            linkProperty().set(path.toString());
+            checkExtension();
+        });
+    }
+
+    //
```
tobiasdiez: trailing comment

src/main/java/org/jabref/gui/filelist/LinkedFilesEditDialogViewModel.java
```diff
@@ -0,0 +1,126 @@
...
+    public void setValues(LinkedFile entry) {
```
tobiasdiez: `entry` is a bit unfortunate name and usually reserved for bibentries.

src/main/java/org/jabref/preferences/PreferencesService.java
```diff
@@ -11,4 +12,10 @@
     KeyBindingRepository getKeyBindingRepository();
 
     void storeJournalAbbreviationPreferences(JournalAbbreviationPreferences abbreviationsPreferences);
+
+    FileDirectoryPreferences getFileDirectoryPreferences();
+
+    String get(String key);
```
tobiasdiez: Please don't add generic methods like this! The idea was that the `PreferencesService` provides a typed interface for the preferences. Thus here it would be `set/getWorkingDirectory` with `nio.path` as argument/return value.

src/main/resources/org/jabref/gui/filelist/LinkedFileEditDialog.fxml
```diff
@@ -0,0 +1,53 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
tobiasdiez: Please add this file in the `java` folder and not `resource`

src/main/resources/org/jabref/gui/filelist/LinkedFileEditDialog.fxml
```diff
@@ -0,0 +1,53 @@
...
+               </GridPane.margin>
+            </Button>
+            <Label text="Filetype" GridPane.rowIndex="2">
+               <padding>
+                  <Insets bottom="10.0" left="10.0" right="10.0" top="10.0" />
+               </padding>
+            </Label>
+         </children>
+      </GridPane>
+   </content>
```
tobiasdiez: You can also add `buttontypes` here in the FXML file.

src/main/resources/org/jabref/gui/filelist/LinkedFileEditDialog.fxml
```diff
@@ -0,0 +1,53 @@
...
+            <ColumnConstraints hgrow="SOMETIMES" minWidth="10.0" prefWidth="108.0" />
+        </columnConstraints>
+        <rowConstraints>
+          <RowConstraints minHeight="10.0" prefHeight="30.0" vgrow="SOMETIMES" />
+          <RowConstraints minHeight="10.0" prefHeight="30.0" vgrow="SOMETIMES" />
+          <RowConstraints minHeight="10.0" prefHeight="30.0" vgrow="SOMETIMES" />
+        </rowConstraints>
+         <children>
+               <Label minWidth="95.0" text="%Link">
+                   <padding>
```
tobiasdiez: padding and layout stuff should be controlled using css and not in the fxml file.

src/main/resources/org/jabref/gui/filelist/LinkedFileEditDialog.fxml
```diff
@@ -0,0 +1,53 @@
...
+<?import javafx.scene.layout.ColumnConstraints?>
+<?import javafx.scene.layout.GridPane?>
+<?import javafx.scene.layout.RowConstraints?>
+
+<DialogPane maxHeight="1.7976931348623157E308" maxWidth="1.7976931348623157E308" minHeight="-Infinity" prefHeight="150.0" prefWidth="536.0" xmlns="http://javafx.com/javafx/8.0.141" xmlns:fx="http://javafx.com/fxml/1" fx:controller="org.jabref.gui.filelist.LinkedFileEditDialogView">
+   <content>
+      <GridPane>
+        <columnConstraints>
+          <ColumnConstraints hgrow="NEVER" minWidth="10.0" prefWidth="100.0" />
+          <ColumnConstraints hgrow="ALWAYS" maxWidth="1.7976931348623157E308" minWidth="10.0" prefWidth="200.0" />
```
tobiasdiez: maxWidth = infinity ?

src/main/java/org/jabref/gui/copyfiles/CopyFilesDialogView.java
```diff
@@ -28,11 +29,14 @@ public CopyFilesDialogView(BibDatabaseContext bibDatabaseContext, CopyFilesResul
         this.setTitle(Localization.lang("Result"));
         this.setResizable(true);
 
+        this.getDialogPane().getButtonTypes().addAll(ButtonType.CLOSE);
```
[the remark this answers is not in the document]
Siedlerchr: The problem was that with an only OK button I was not able to close the dialog neither via the X in the corner nor via the OK button. It needs to have at least a button with Type CANCEL_CLOSE 
Read the last paragraph: https://docs.oracle.com/javase/8/javafx/api/javafx/scene/control/Dialog.html

src/main/java/org/jabref/gui/filelist/AttachFileAction.java
Siedlerchr: The filetype guesing happens when the EditDialog is shown

src/main/java/org/jabref/gui/filelist/LinkedFilesEditDialogViewModel.java
```diff
@@ -61,17 +61,17 @@ private void checkExtension() {
...
         Optional<Path> file = FileHelper.expandFilename(database, fileText, preferences.getFileDirectoryPreferences());
 
         Path workingDir = file.orElse(Paths.get(preferences.get(JabRefPreferences.WORKING_DIRECTORY)));
         String fileName = Paths.get(fileText).getFileName().toString();
 
         FileDialogConfiguration fileDialogConfiguration = new FileDialogConfiguration.Builder()
-        .withInitialDirectory(workingDir)
-        .withInitialFileName(fileName)
-        .build();
+                                                                                               .withInitialDirectory(workingDir)
```
Siedlerchr: @tobiasdiez  this is how the indentation at column level works: As I said before it's not possible to align it at the dot position. So when I use indencation on column  all things are then wrapped after Builder (Github doesn't show it that wide).

src/main/java/org/jabref/gui/filelist/AttachFileAction.java
Siedlerchr: Previously there wasn't event he FileChooser shown .Instead it only showed the LinkedFilesEditor in which I then had to click open

src/main/java/org/jabref/gui/copyfiles/CopyFilesDialogView.java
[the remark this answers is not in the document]
Siedlerchr: The Problem was that the OK button was not at the Dialog Pane, but inside the content and therefore did not work

src/main/java/org/jabref/gui/filelist/LinkedFilesEditDialogViewModel.java
```diff
@@ -102,19 +102,19 @@ public void setValues(LinkedFile entry) {
         }
     }
 
-    public StringProperty linkProperty() {
+    public StringProperty link() {
         return linkProperty;
```
tobiasdiez: Sorry, I wasn't clear enough. The private field should be called `link` with public methods `linkProperty()` and `get/setLink`, with the latter optional.

src/main/java/org/jabref/gui/filelist/AttachFileAction.java
tobiasdiez: Ok, then was the previous behavior even worse. What advantage do you see of displaying the dialog every time?

Siedlerchr: You can add a description and or choose a different file.

tobiasdiez: If you really think that the majority of users wants to add a description, then keep the current behavior....

Siedlerchr:
I fixed all things open and did the renamings.  So  I merge this now

Issue #2996, opened 2017-07-11
Enable DeclarationOrder in Eclipse and IntelliJ


koppor:
[DeclarationOrder](http://checkstyle.sourceforge.net/apidocs/com/puppycrawl/tools/checkstyle/checks/coding/DeclarationOrderCheck.html) should be set in the IntelliJ config - or disabled.

1. Class (static) variables. First the public class variables, then the protected, then package level (no access modifier), and then the private.
1.  Instance variables. First the public class variables, then the protected, then package level (no access modifier), and then the private.
1. Constructors
1. Methods

koppor:
Refs https://github.com/JabRef/jabref/pull/2765#issuecomment-297405985
@snisnisniksonah

tobiasdiez:
Same applies for the rule `'METHOD_DEF' has more than 1 empty lines before. [EmptyLineSeparator]`

tobiasdiez:
I found a checkstyle IntellJ plugin, which imports a checkstyle configuration and changes the style accordingly. I also added a description of how to setup everything in the wiki. So this issue can be closed in my opinion.

PR #3006, opened 2017-07-13
Don't abort build when there are checkstyle violations

lynyus:
Upon wish of @lenhard, nothing more to say here

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -392,6 +392,7 @@ checkstyle {
     // do not use other packages for checkstyle, excluding gen(erated) sources
     checkstyleMain.source = "src/main/java"
     toolVersion = '7.6.1'
+    ignoreFailures = true
 }
 
 task release(dependsOn: ["media", "releaseJar"]) {


koppor:
Why?

This improves our code quality. If not the author, who else will clean up the code?

Which checkstyle rules are annoying?

We all know  (do we?) that JabRef's code was (and still is) a mess (at some places). I strongly believe that checkstyle rules are one solution to the problem.

koppor:
I give it a go. If the active developers don't see any value in checkstyle checks, I give it a go.

Refs https://github.com/JabRef/jabref/issues/2996

Siedlerchr:
Well, also I think checkstyle warnings are annoying, they necessary. And it's not much work when you just have to press ctrl+f in eclipse to do it.
The only thing I am not happy about is that sort order of declarations. That could create problems when initializing fields.

lynyus:
Hey @koppor, the intention is not to abolish checkstyle, but to improve the coding experience.

When working on a PR, the most important thing are the unit tests to pass. When executing gradle check it takes 59 secs to only compile the java code and run the checkstyle tests. If there is ANY problem because I didn't hit the code formatter the build will abort at this point. That's why suggested to make them optional. Of course they should be adhered.

In an ideal case they SHOULD fail on travis, but not locally. Does anybody know if that's possible in an elegant way?

lenhard:
If I recall the discussion correctly, then we agreed on the following: This build should execute even if their are checkstyle warnings (so that you get feedback on the tests), but it should still fail. Is this what this PR achieves?

I hate checkstyle and complain all the time, but I concede that the rules improve code quality.

PR #3008, opened 2017-07-13
Don't enforce default Windows font

lynyus:
Check below https://github.com/JabRef/jabref/pull/3008#issuecomment-315136683

> 
> Default Arial has only 3988 glyphs (cf. https://en.wikipedia.org/wiki/Unicode_font#List_of_Unicode_fonts), where italics don't seem to be supported.
> 
> To fix the issue at hand I would propose to change the font to GNU Unifont](https://en.wikipedia.org/wiki/GNU_Unifont) which supports 63,449 glyphs.
> 
> License-wise there shouldn't be any issues:
> 
> > The software on this site, unless otherwise noted, is released under the terms of the GNU General Public License (GNU GPL) version 2.0, or (at your option) a later version. The precompiled fonts are released under the terms of the GNU GPL version 2, or (at your option) a later version, with the exception that embedding the font in a document does not in itself bind that document to the terms of the GPL.
>  [source: unifoundry.com](http://www.unifoundry.com/)
> 

unifont

### DIFF
--- a/src/main/java/org/jabref/preferences/JabRefPreferences.java
+++ b/src/main/java/org/jabref/preferences/JabRefPreferences.java
@@ -452,7 +452,7 @@ public class JabRefPreferences implements PreferencesService {
             defaults.put(WIN_LOOK_AND_FEEL, UIManager.getSystemLookAndFeelClassName());
             defaults.put(EMACS_PATH, "emacsclient");
         } else if (OS.WINDOWS) {
-            defaults.put(FONT_FAMILY, "Arial");
+            defaults.put(FONT_FAMILY, "Unifont");
             defaults.put(WIN_LOOK_AND_FEEL, "com.jgoodies.looks.windows.WindowsLookAndFeel");
             defaults.put(EMACS_PATH, "emacsclient.exe");
         } else {
--- a/src/main/java/org/jabref/preferences/JabRefPreferences.java
+++ b/src/main/java/org/jabref/preferences/JabRefPreferences.java
@@ -452,7 +452,6 @@ public class JabRefPreferences implements PreferencesService {
             defaults.put(WIN_LOOK_AND_FEEL, UIManager.getSystemLookAndFeelClassName());
             defaults.put(EMACS_PATH, "emacsclient");
         } else if (OS.WINDOWS) {
-            defaults.put(FONT_FAMILY, "Unifont");
             defaults.put(WIN_LOOK_AND_FEEL, "com.jgoodies.looks.windows.WindowsLookAndFeel");
             defaults.put(EMACS_PATH, "emacsclient.exe");
         } else {


lenhard:
On my Win7 system, this font seems to be unavailable. I cannot select it in the preferences dialog and resetting the preferences does not help as well.

Maybe the real solution here is to tell the users they should install and configure a font that supports italics? Then, we would solve this issue by documentation?

lynyus:
I have removed the font default now. On my win7 system it defaults to "Dialog" and the italic texts are shown.
@lenhard, can you check again what happens on Windows 7?

lenhard:
I get a font named "Dialog", but unfortunately no italics.

lynyus:
@lenhard because my windows system time is broken....

lynyus:
So what we need is any font that is available and works on all Windows versions :/

AEgit:
I might add something on this.
I agree that not seeing the italics in the title is itself not a big issue (although it would be nice to have it at a certain point in time).
I disagree, however, with the assumption that not many users have `\textit` in their title. Genus and species names in biology have to (!) be written in italics. I guess therefore, that every user that is somewhat affiliated with biology will have articles that have italics in their titles.

koppor:
Why removing the default font? Has "Dialog" more characters than Arial? OK, could be OK for me...

However, I checked stackoverflow: [embedding a font in Java](https://stackoverflow.com/q/4487871/873282) and also [that](https://stackoverflow.com/a/8365030/873282).

koppor:
It seems that there is no font that we can embed perfectly.

lenhard:
@AEgit As you've probably read already, we won't do more changes here. Ultimately, this is a problem of the capabilities of the fonts installed in the underlying operation system. So it is a little out of the reach for JabRef to do much about it. In newer OSs, Win10 for instance, there seem to be plenty of fonts that can handle Unicode. If you're on Win7, there's Cambria Math that does the trick.

AEgit:
Hmmm, couldn't you just use a font that is already part of the respective operating system (i.e., without installing a new one), that is capable of handling Unicode? So, e.g. in Windows 7, JabRef would default to "Cambria Math" as default font, and in Windows 10 to any other font you mentioned (I'm on Windows 10 so I guess I should have less issues).

AEgit:
Nevermind, I see, that's exactly what you have done in the current development version:
JabRef 4.0-dev--snapshot--2017-07-19--master--2f12fecfb
Windows 10 10.0 amd64 
Java 1.8.0_141

Cheers and thank you very much!

Siedlerchr:
For summary: LIst of fonts supporting Math Unicode blocks (For MacOS apparently FreeSerif works good)
http://www.fileformat.info/info/unicode/block/mathematical_alphanumeric_symbols/fontsupport.htm

Issue #2966, opened 2017-07-05
Improve startup time


tobiasdiez:
For some reason JabRef 4.0 needs really long to start.

Siedlerchr:
I think this is related to the telemetry stuff

lenhard:
Yes, the telemetry is probably an influencing factor.

Other things are:
 - Class loading: We've added a lot of libraries recently. Class loading takes time, especially in the beginning when a lot of classes need to be loaded to build the GUI. The only way to improve this is by reducing the amount of libraries we use. We've done this around 2 years ago, but looks like it has crept in again
 - Loading of FXML files: They're nice, but they need to be loaded at some point and that likely includes startup. I have no evidence for this, but my guess is that compiled code loads faster.

Siedlerchr:
One thing I noticed: In Gloabs.java we should do:
        JabRefExecutorService.INSTANCE.execute(() -> startTelemetryClient());
instead of starting the telemetry client directly

tobiasdiez:
I did some investigations.

Setting up the main table takes nearly 21 seconds. The main contribution is coming from
a) sorting the entries (11 secs)
b) loading the journal abbreviations for the auto completion (8 secs)
![image](https://user-images.githubusercontent.com/5037600/28171504-e16ad036-67e8-11e7-98bd-4fb5ba334afe.png)
 
Rest rest of the interface also takes a bit of time (5 secs) but that is negligible in comparison to the above numbers. Main reason here is setting up the JavaFX stage and loading the FXML files for the group interface (2 secs) and the search bar (0.5 secs). There is not really anything we can do about it (except migrating everything to JavaFX so that not every JavaFX control needs its own stage). 
What I found interesting is that all these `LOGGER = getLog(....)` statements take 5 seconds (can we fix this? we do not really need the option to have different loggers for different classes, right?).
![image](https://user-images.githubusercontent.com/5037600/28172077-c6fc7c20-67ea-11e7-8b59-713c497b270b.png)

Telemetry (1 sec) and our update notification (2.5 secs) take some time due to the HTTP requests but both are performed on different threads so this should do no harm.

lenhard:
Thanks a lot for this, this makes for a much more well-founded discussion. Apart from the log issues you describe, my 5 cents are the following:

The sorting is done by the glazed lists library that we are desperately wanting to get rid of. Maybe the solution would be to migrate the main table to jabref in the nearer future.

[...]

PR #3009, opened 2017-07-13
Improve performance of journal abbreviation loader

tobiasdiez:
This PR improves the performance of the journal abbreviation loader/repository/parser and thus is a partial fix for https://github.com/JabRef/jabref/issues/2966.
Previously, loading and parsing the journal repository files took nearly 10 seconds and now almost nothing:
image
Main boost came from using `equalsIgnoreCase` instead of using `equals(text.toLowerCase())` and changing `LinkedList` to `HashSet` as we don't care about the ordering.
I had to adopt two tests which relied on the sorting.

Moreover, only one copy of the list is now stored in memory opposed to the 1 + 4 maps previously (equates to 1/10 RAM consummation). In principle, this makes the journal auto completion slower but I couldn't notice any negative impact.


lenhard:
Awesome :) I've tested it locally and I have the feeling that startup is faster. I've played around a little with autocompletion, but I don't notice any difference.

PR #3012, opened 2017-07-14
Mark LibraryOfCongressTest as a fetcher test

lynyus:
This test is currently failing with a 500 Error. I would suggest to add it to the allowed failures and open an issue.


lenhard:
You are right to do so. I'll merge this directly (although I am the only reviewer), so that we have a clean build again.

PR #3015, opened 2017-07-15
[WIP] Use jcabi-log to reduce startup time

koppor:
This implements https://github.com/JabRef/jabref/issues/2966#issuecomment-315127129:

    Logger.debug(this, ...) instead of LOGGER.debug(...)

> 13:56:54.678 [AWT-EventQueue-0] INFO  org.jabref.logic.importer.OpenDatabase - Opening: C:\git-repositories\jabref\jabref\src\test\resources\testbib\complex.bib

#3017 is also included here, because of `net.corda:jfx` depends on `org.slf4j:jcl-over-slf4j`, but we want to pipe slf4j to jcl.

@tobiasdiez Could you check the startup time?


koppor:
Rumors said that simonharrer uses jcabi-log in current projects. 😇 

## Without this PR

10% time for logging startup

grafik


## With this PR

7% time for logging startup

grafik

koppor:
@lenhard You are right, the logging of exceptions changed (not that obviously documented - see https://github.com/jcabi/jcabi-log/issues/68).

    Logger.debug("Error during X", e);

becomes

    Logger.debug("Error during X: %[exception]s", e);

I began to rewrite our code, but I am unsure whether we really want that. Maybe we should write our own static log methods which check whether the last parameter is instance of Throwable, add `: %[exception]s` to the formatting string, and finally delegate to jcabi-log?

lenhard:
3% improvement is an improvement, so we should go for it :)

Regarding the exception: I think the way you suggest is the right one. You should write a wrapper class, say `org.jabref.JabRefLogger` with static methods for the different log levels and one for an exception. Then this logger should be used in JabRef. Otherwise we would be duplicating `%[exception]s` all over the code

Siedlerchr:
Yes, but instead of checking explicitly for a throwable, just create one wrapper method overlaod which excepts an Throwable as second argument.

koppor:
Done. See [org.jabref.Logger](https://github.com/JabRef/jabref/pull/3015/files#diff-42a62e13854bd284a29625feccd2b3b8). We still (?) have many classes with static methods. https://github.com/JabRef/jabref/pull/3015/commits/0c0d26b004fb3ac947defe77e9ba4ae3eb4c1baf#diff-f64d223f20f6b0e53b00c8dbd2be2ce1R62 I cannot fix that, but just ignore them in the ArchitectureTests.

stefan-kolb:
Not sure if changing the interface of such a well-known functionality like logging is really the way to go. Don't you think this will lead to fragmented code as not everyone is used to it?

src/main/java/org/jabref/Logger.java
```diff
@@ -0,0 +1,153 @@
...
+                             final String msg, final Throwable e, final Object... args) {
+        // error is always logged. Therefore, no isErrorEnabled(Object) exists
+        com.jcabi.log.Logger.error(source, msg + ": %[exception]s", merge(args, e));
+    }
+
+    /**
+     * Merges args and e together into an array starting with args
+     */
+    private static Object[] merge(Object[] args, Throwable e) {
+        final int length = args.length;
```
Siedlerchr: Guava has a method for Array concat:
https://google.github.io/guava/releases/19.0/api/docs/com/google/common/collect/ObjectArrays.html#concat(T%5B%5D,%20T%5B%5D,%20java.lang.Class)

docs/adr/index.md
```diff
@@ -0,0 +1,7 @@
+# Architectural Decision Records
+
+This lists the architectural decisions for Eclipse Winery™
```
stefan-kolb: Stuff from Winery....

src/main/java/org/jabref/Logger.java
```diff
@@ -0,0 +1,153 @@
...
+ * We use jcabi-log (see ADR-0001)
+ *
+ * We need to add a wrapper around it as developers won't write Logger.debug(this, "msg: %[exception]s", e) for every logged exception
+ */
+public class Logger {
+
+    /** begin: direct pass-through **/
+
+    public static void trace(final Object source, final String msg) {
+        com.jcabi.log.Logger.trace(source, msg);
```
koppor: Yes, there would be conflicts.

koppor:
I agree that this is uncommon. However, I think (i) it is more convenient to work without Logger factory, if one wants "gschwind mal was loggen". (ii) We have the architecture test in place complaining if someone uses apache commons logging 😇 .

lenhard:
Why is there a problem with using the Logger in static methods? Because you need to provide an instance as the source of the message? If that's the case, why don't you just do something like this: `Logger.debug(MyClass.class, "My log message");`

The style of logging should be the same everywhere, in static and non-static methods. If we can't find a uniform way, logging will just degrade overtime and we end up with a mess.

koppor:
jcabi-log needs `this`, because it also provides information on threads. See here for more features: http://log.jcabi.com/

In case we don't want to have this feature, I can rewrite the layer so that `this` does not need passed. Think, then we are at TinyLog. I'm unsure, why we dropped that in another project...

lenhard:
Hm, how about at least adding a few methods to the current solution that does not require `this`? So that we can use the new logger also from static methods. That would be a compromise (we get the threading information if we want, but we don't have to). What do zou think? Would that work?

stefan-kolb:
I'd still favor keeping the current logging. I don't see too much improvements and a few problems with the new solution.

koppor:
Logger.debug(CurrentClass.class, ...) works. I also like the fact, I can
just do Logger.debug(this, ...) in a class where there currently is no
logger... Maybe an issue for the devcall... 🕺

Am 21.07.2017 12:47 schrieb "Stefan Kolb" <notifications@github.com>:

koppor:
We do want to change the usual way of doing logging for just saving some seconds at the startup time.

koppor:
If I redid this, I would give [tinylog](https://tinylog.org/v2/) a try.

Issue #3014, opened 2017-07-15
Reimplement MappedList.kt


koppor:
During the discussion of #2966, we also discussed issues with too many classes.

We have a import to a JavaFX helper class [net.corda.client.jfx.utils.MappedList](https://github.com/corda/corda/blob/master/client/jfx/src/main/kotlin/net/corda/client/jfx/utils/MappedList.kt). The dependency causes many additional libraries to be loaded. I tried to [convert Kotlin to Java](https://stackoverflow.com/a/40762755/873282), but it is not as easy as promised.

Can we reimplement the behavior of this class instead of fetching a client for a [ledger platform](https://github.com/corda/corda)?

```
+--- net.corda:jfx:0.12.1
|    +--- net.corda:core:0.12.1
|    |    +--- org.jetbrains.kotlin:kotlin-stdlib-jre8:1.1.2
|    |    |    \--- org.jetbrains.kotlin:kotlin-stdlib-jre7:1.1.2
|    |    |         \--- org.jetbrains.kotlin:kotlin-stdlib:1.1.2
|    |    |              \--- org.jetbrains:annotations:13.0
|    |    +--- org.jetbrains.kotlin:kotlin-reflect:1.1.2
|    |    |    \--- org.jetbrains.kotlin:kotlin-stdlib:1.1.2 (*)
|    |    +--- com.google.code.findbugs:jsr305:3.0.1
|    |    +--- org.apache.logging.log4j:log4j-core:2.7 -> 2.8.2
|    |    |    \--- org.apache.logging.log4j:log4j-api:2.8.2
|    |    +--- org.slf4j:jcl-over-slf4j:1.7.25
|    |    |    \--- org.slf4j:slf4j-api:1.7.25
|    |    +--- org.slf4j:slf4j-api:1.7.25
|    |    +--- com.google.guava:guava:21.0 -> 22.0 (*)
|    |    +--- io.reactivex:rxjava:1.2.4
|    |    +--- com.esotericsoftware:kryo:4.0.0
|    |    |    +--- com.esotericsoftware:reflectasm:1.11.3
|    |    |    |    \--- org.ow2.asm:asm:5.0.4
|    |    |    +--- com.esotericsoftware:minlog:1.3.0
|    |    |    \--- org.objenesis:objenesis:2.2
|    |    +--- de.javakaffee:kryo-serializers:0.41
|    |    |    \--- com.esotericsoftware:kryo:4.0.0 (*)
[...]

PR #3016, opened 2017-07-15
Google Guava over Apache Commons Lang3

koppor:
We don't have an [Architectural Decision Record](https://adr.github.io/) why we included Apache Commons Lang. We introduced the dependency within https://github.com/JabRef/jabref/pull/1572.

I don't know, why Apache Commons Lang3 was chosen. The [current opinion](https://stackoverflow.com/a/4543114/873282) is that Guava is more modern and that new projects should use Google Guava instead of Apache Commons Lang3.

Should we replace Apache Commons Lang3 or should we just keep it, because it is not that huge and we have other class issues (see https://github.com/JabRef/jabref/issues/3014).

Refs https://github.com/JabRef/jabref/issues/2966#issuecomment-314369222

## Update

- I replaced most places where we use Apache Commons Lang3 and there is a Google Guava equivalent
- We have some places, where there is no Google Guava equivalent, so I kept that
- Moved all architecture tests to `org.jabref.architecture`.
- I began to use [ArchUnit](https://github.com/TNG/ArchUnit/) for checking that. However, I have issues in using the `notAnnotatedWithRule` (see https://github.com/TNG/ArchUnit/pull/19).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -125,8 +125,6 @@ dependencies {
 
     compile 'commons-logging:commons-logging:1.2'
 
-    compile 'org.apache.commons:commons-lang3:3.6'
-
     compile 'org.jsoup:jsoup:1.10.3'
     compile 'com.mashape.unirest:unirest-java:1.4.9'
     compile 'info.debatty:java-string-similarity:0.24'
@@ -152,6 +150,9 @@ dependencies {
     testCompile 'com.github.tomakehurst:wiremock:2.6.0'
     testCompile 'org.assertj:assertj-swing-junit:3.6.0'
     testCompile 'org.reflections:reflections:0.9.11'
+
+    testCompile 'com.tngtech.archunit:archunit-junit:0.4.0'
+    testCompile 'org.slf4j:slf4j-jcl:1.7.25' // required by ArchUnit to enable logging over jcl
 }
 
 sourceSets {
--- a/external-libraries.txt
+++ b/external-libraries.txt
@@ -1,6 +1,9 @@
 This document lists the fonts, icons, and libraries used by JabRef.
 This file is manually kept in sync with build.gradle and the binary jars contained in the lib/ directory.
 
+One can list all depdencies by using Gradle task `depdencyReport`.
+It generated the file [build/reports/project/dependencies.txt](build/reports/project/dependencies.txt).
+
 # Legend
 
 ## License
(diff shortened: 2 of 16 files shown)


Siedlerchr:
I like the guava stuff. I don't see any big advantage in the apache commons stuff which the Guava has not.
And it seems like the Guava has also an Escape method:
https://github.com/google/guava/blob/9b94fb3965c6869b0ac47420958a4bbae0b2d54c/android/guava-tests/test/com/google/common/html/HtmlEscapersTest.java

lenhard:
I am also in favor of replacing commons lang with guava. So, if anybody wants to go ahead with this, be my guest.

koppor:
Fun fact:

```
+--- com.microsoft.azure:applicationinsights-core:1.0.+ -> 1.0.8
|    +--- eu.infomas:annotation-detector:3.0.4
|    +--- commons-io:commons-io:2.4
|    +--- org.apache.commons:commons-lang3:3.1 -> 3.4
|    +--- org.apache.httpcomponents:httpclient:4.3.5 -> 4.5.2 (*)
|    \--- com.google.guava:guava:12.0.1 -> 22.0 (*)
```

koppor:
HtmlToUnicode is unfortunately not the same as Escaping HTML:

`&Epsilon;` gets `&amp;Epsilon;` instead of `E` 😄

koppor:
What I stumbled upon while searching for various commons lang replacements: [Ecilpse collections](http://www.eclipse.org/collections/).

koppor:
- I replaced most places where we use Apache Commons Lang3 and there is a Google Guava equivalent
- We have some places, where there is no Google Guava equivalent, so I kept that
- I began to use [ArchUnit](https://github.com/TNG/ArchUnit/) for checking that. However, I have issues in using the `notAnnotatedWithRule` (see https://github.com/TNG/ArchUnit/pull/19).

The ArchRule doNotUseApacheCommonsLang3 reads easy:

    public static final ArchRule doNotUseApacheCommonsLang3 =
            noClasses().that().areNotAnnotatedWith(ApacheCommonsLang3Allowed.class)
            .should().accessClassesThat().resideInAPackage("org.apache.commons.lang3");

However, this rule completely ignores the annotation `ApacheCommonsLang3Allowed` and I have no clue how to solve that.

PR #3017, opened 2017-07-15
Minimize depdencies fetched by using net.corda:jfx

koppor:
This is to reduce the mass of transitive dependencies fetched by using corda.net. See https://github.com/JabRef/jabref/issues/3014.


lenhard:
The conversion would be the more desirable solution.

But if we don't get the conversion done before the release of 4.0, we might consider to merge it anyway. I'd say: Let's keep this PR open until either someone does the conversion or we want to release 4.0.

koppor:
I assigned https://github.com/JabRef/jabref/issues/3014 to a milestone, so we are reminded that there is some work left.

I will merge this in as it will surely reduce the size of our JAR.

PR #3026, opened 2017-07-18
Add checkstyle as separate TravisCI test

koppor:
Add checkstyle as separate test to speedup feedback.


tobiasdiez:
Is it on purpose that the first job (`TEST_SUITE=check OPTIONS=modernizer`) still runs checkstyle? As of now checkstyle is run twice during the build and thus the build times are increased by about 5 mins.

Issue #2853, opened 2017-05-19
Key binding DELETE tries to delete whole entry when file is selected in general tab


stefan-kolb:
![image](https://cloud.githubusercontent.com/assets/2141507/26243464/3e901a06-3c8c-11e7-98a9-17fa8b204aeb.png)

1. select file in general tab
2. press DELETE key

RESULT: JabRef wants to delete whole entry
EXPECTED: File assignment is removed

tobiasdiez:
This is the same underlying problem as https://github.com/JabRef/jabref/issues/2855.

PR #3030, opened 2017-07-20
Fix keybinding in entry editor in localized installation

Siedlerchr:
Check keycode to avoid localization problems

Follow up from #2853 

### DIFF
--- a/src/main/java/org/jabref/gui/keyboard/KeyBindingRepository.java
+++ b/src/main/java/org/jabref/gui/keyboard/KeyBindingRepository.java
@@ -127,7 +127,7 @@ public class KeyBindingRepository {
     }
 
     public Optional<KeyBinding> mapToKeyBinding(java.awt.event.KeyEvent keyEvent) {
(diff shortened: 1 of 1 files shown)


matthiasgeiger:
Alternative solution without using the deprecated function:

```java
    private Optional<KeyCode> determineKeyCodeFromKeyEvent(java.awt.event.KeyEvent keyEvent) {
        // special treatment of localized KeyEvents
        if (keyEvent.getKeyCode() == java.awt.event.KeyEvent.VK_DELETE) {
            return Optional.of(KeyCode.DELETE);
        }
        
        return Arrays.stream(KeyCode.values()).filter(k -> k.getName().equals(java.awt.event.KeyEvent.getKeyText(keyEvent.getKeyCode()))).findFirst();
    }
```

But this only solves the problem for the "delete" key - so we still might have other yet unidentified problems.

So we should decide which route to take ;-)

koppor:
I would give a +1 for @Siedlerchr's solution, since we surely do not catch all possible keys and the maintable is our next target after the entry editor 😇.

PR #3032, opened 2017-07-20
Replace Reimplement MappedList.kt

lynyus:
Fixes https://github.com/JabRef/jabref/issues/3014.

I have mainly used this gist as an idea https://gist.github.com/TomasMikula/8883719, the alternative would have been https://gist.github.com/mikehearn/a2e4a048a996fd900656

### DIFF
--- a/src/main/java/org/jabref/gui/util/MappedList.java
+++ b/src/main/java/org/jabref/gui/util/MappedList.java
@@ -8,12 +8,17 @@ import javafx.collections.ListChangeListener;
 import javafx.collections.ObservableList;
 import javafx.collections.transformation.TransformationList;
 
+/**
+ * MappedList implementation based on https://gist.github.com/TomasMikula/8883719
+ *
+ * @author https://github.com/TomasMikula
+ */
 public final class MappedList<A, B> extends TransformationList<A, B> {
 
     private final Function<B, A> mapper;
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/jabref/gui/util/MappedList.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    @Override
+    public int getSourceIndex(int index) {
+        return index;
+    }
+
+    @Override
+    public A get(int index) {
+        return mapper.apply(super.getSource().get(index));
```
tobiasdiez: This implementation of the MappedList looks identical to the one in the EasyBind package but is different from `MappedList.kt`. 
The mapped list from the corda package runs the `mapper` when the item is [inserted](https://github.com/corda/corda/blob/master/client/jfx/src/main/kotlin/net/corda/client/jfx/utils/MappedList.kt#L19)/updated while this implementation here invokes the `mapper` on every `get`-call. Since the `get` method is called relatively often (e.g. if you scroll in the listview), the EasyBind implementation should only be used for quick mappers. Can you please revert the change and/or write a direct reimplementation of the corda mapped list since we otherwise run again in performance problems with huge groups trees.

lynyus: Good point, will do!

PR #3033, opened 2017-07-20
Initial idea for architectural decision records

koppor:
I'd like to improve the code documentation of JabRef to make is easier for newcomers to understand the code. One part of that is to document the ideas behind of some code. This PR adds an initial directory and template for that.

I tried to collect background information at http://adr.github.io/.

### DIFF
--- a/docs/adr/0000-use-architectural-decision-records.md
+++ b/docs/adr/0000-use-architectural-decision-records.md
@@ -1,0 +1,16 @@
+# Use Architectural Decision Records
+
+We need to record the architectural decisions made on this project.
+
+## Considered Alternatives
+
+* No record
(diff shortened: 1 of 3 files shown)


Siedlerchr:
Related: https://github.com/JabRef/jabref/wiki/Code-Howtos

koppor:
This is also related: https://github.com/JabRef/jabref/blob/master/CONTRIBUTING.md#when-making-an-architectural-decision -- no one every did it until now.

I would move everything from the wiki to docs/ and enable gh-pages build from master-branch/docs and render it at https://jabref.github.io/jabref. I think, 90% of the information can be moved, but pages like https://github.com/JabRef/jabref/wiki/Templates and https://github.com/JabRef/jabref/wiki/Acquiring-Contributors should be moved somewhere else.

PR #3037, opened 2017-07-21
Fix navigation in entry editor increases modifies font size

Siedlerchr:


Siedlerchr:
The only thing which does not yet work is ctrl+MINUS, it always decreases the font size although I added the exception here. Did not yet figure out the problem. Maybe @lenhard  can have a look at that, too.

Siedlerchr:
Found the problem. The decreasing font size was found first and returned before the entry editor panel constant, because both have the same keymapping

PR #3061, opened 2017-08-01
Closes #3027 and updates install4j to v7

### LINKED ISSUE #3027: Update java version to Update 141 (JDK 8u141)
Released yesterday
http://www.oracle.com/technetwork/java/javase/8u141-relnotes-3720385.html

stefan-kolb:
We need to check if the install4j migration did change something inside the installer.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -10,17 +10,17 @@ dependencies:
     # openssl aes-256-cbc -e -in ./buildres/jabref-cert-2016.p12 -out jabref-cert-2016.p12.enc -k {PASSWORD}
     - openssl aes-256-cbc -d -in ./buildres/jabref-cert-2016.p12.enc -out ./buildres/jabref-cert-2016.p12 -k $CERTIFICATE
     - scripts/prepare-install4j.sh
-    - install4j6/bin/install4jc --verbose --license=$INSTALL4J_KEY --win-keystore-password $CERTIFICATE_PW --mac-keystore-password $CERTIFICATE_PW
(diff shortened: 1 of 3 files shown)


Siedlerchr:
Could you provide a Screenshot if you changed the layout?

stefan-kolb:
Waiting for the build, this was just an auto migration by intsall4j7.

stefan-kolb:
@JabRef/developers Windows x64 installation looked ok on my end.

stefan-kolb:
@chochreiner Thanks for testing. Not sure if I we can disable this, I didn't see an option yet, although I dunno where to look really 😄

chochreiner:
The warning looks like this after the second installation/update of Jabref after the installation and before the "File association" step 
I've had a look at the install4j configuration, but I could not find the setting to turn this of.

screen shot 2017-08-01 at 11 33 04

Siedlerchr:
Just another idea: I think we already discussed that a while ago:
Add a page/button to open donations, e..g like in Tortoise SVN:
grafik

stefan-kolb:
@lenhard I don't have an icon for it other than that it would be a 5 minute task 😄

stefan-kolb:
@chochreiner Probably caused by adding this to the Info.plist? <key>NSSupportsAutomaticGraphicsSwitching</key>
<string>YES</string>
Do we still need this?

stefan-kolb:
Or maybe did the deinstallation not work properly?

chochreiner:
The deinstallation should have worked fine and I would keep the NSSupportsAutomaticGraphicsSwitching attribute, since there may be quite some macbooks around with two graphic cards

PR #3069, opened 2017-08-02
Reimplement MappedList using a backigList

lynyus:
This is the followup for https://github.com/JabRef/jabref/pull/3032#discussion_r129271280.

This is different from https://github.com/JabRef/jabref/pull/3032, as now the mapping only happens on insert.


Siedlerchr:
Could you add a test to ensure correct behaviour?

lynyus:
@Siedlerchr Actually the test coverage is already not too bad.
coverage

lynyus:
@Siedlerchr actually the test coverage covers the important parts of the method.
coverage

lynyus:
@tobiasdiez I manually tested it, however I don't have many groups. Please check that.

Issue #2989, opened 2017-07-10
Cut/Paste article


mpele:
JabRef version - development release

Steps to reproduce:

1. Cut article from some group
2. Paste to another group

The article is copied to another group but it is not removed from original group.

AEgit:
I can confirm this issue for:
JabRef 4.0.0-dev--snapshot--2017-07-10--master--e80f450ee
Windows 10 10.0 amd64 
Java 1.8.0_131

tobiasdiez:
How do you cut and paste "from the group"?
The usual Cut/Paste operations are semantically performed against the database, that is, cut removes the entry from the database and paste inserts it (in a possible different database). All information like the group membership are preserved in the process. So for example if you have two databases with the same groups then copying/cutting an entry and pasting it in the other database still assigns the entry to the correct group. Thus the behavior you describe is expected.

If you want to change the group membership you can right click the entry and use "Add/Remove from group" or use the corresponding options in the right click menu of the group.

AEgit:
@tobiasdiez: Ah fair enough, if that's the expected behaviour there's no issue to report. Following @mpele I had the impression that "cut" was supposed to remove an article from a group and "paste" add it to the selected group.

PR #3071, opened 2017-08-03
[WIP] Delete or remove from group

mpele:
The default dialog for confirmation of delete entries is changed so the user have option to choose if he wants to delete entry, or to just remove entry from active group. I have few times deleted articles from database not being aware that article belongs to other groups. 
Those changes should make clear to user what is happening when he press "Delete".


src/main/resources/l10n/JabRef_en.properties
```diff
@@ -1172,7 +1172,7 @@ The_chosen_date_format_for_new_entries_is_not_valid=The_chosen_date_format_for_n
 
 The_chosen_encoding_'%0'_could_not_encode_the_following_characters\:=The_chosen_encoding_'%0'_could_not_encode_the_following_characters\:
 
-The_entry_is_the_member_of_groups=The entry is the member of groups
+The_entry_is_the_member_of_groups=The entry is the member of groups:
```
Siedlerchr: The values on the left and on the right must be equal in the English properties file

src/main/java/org/jabref/gui/BasePanel.java
```diff
@@ -767,12 +771,71 @@ private void cut() {
      *            "deleted". If true the action will be localized as "cut"
      */
     private void delete(boolean cut) {
-        List<BibEntry> entries = mainTable.getSelectedEntries();
+    	boolean bRemoveFromGroup = false;
```
Siedlerchr: Please no hungarian notation, so just without the prefix

src/main/java/org/jabref/gui/BasePanel.java
```diff
@@ -767,12 +771,71 @@ private void cut() {
...
+
         if (entries.isEmpty()) {
             return;
         }
-        if (!cut && !showDeleteConfirmationDialog(entries.size())) {
-            return;
+    	
+    	String selectedGroup = "";
+    	if (Globals.stateManager.getSelectedGroup(bibDatabaseContext).size() > 0) {
+    		selectedGroup = Globals.stateManager.getSelectedGroup(bibDatabaseContext).get(0).getName();
```
Siedlerchr: You should add a check for All Entries, because you can't delete from the All Entries group

src/main/java/org/jabref/gui/BasePanel.java
```diff
@@ -767,12 +771,71 @@ private void cut() {
...
+    	}
+
+        //get list of all groups
+        Optional<GroupTreeNode> groups = bibDatabaseContext.getMetaData().getGroups();
```
Siedlerchr: You have to check the optional either with ifPresent or isPresent to see if there is a value otherwise your next line will produce an NPE if there are no groups

src/main/java/org/jabref/gui/BasePanel.java
```diff
@@ -767,12 +771,71 @@ private void cut() {
...
+
+        if (groups.get().getName().contentEquals(groups.get().getName())) { //checks if is selected any group
+        	//TODO define dialog not to offer "remove from group"
+        }
+
+		Object[] options = {Localization.lang("Delete_from_database"),
+			Localization.lang("Remove_from_group") + " \"" + selectedGroup + "\"",
+			Localization.lang("Cancel")};
+		int n = JOptionPane.showOptionDialog(frame,
+				Localization.lang("The_entry_is_the_member_of_groups") + entries.get(0).getField("groups"),
```
Siedlerchr: I think it would be best to put this dialog in a loop, because it it confusing if you delete many entries it displays only the group information from the first entry.
It get's displayed as Optional[Group1, Group2].  
You should add a check before the dialog: (assuming you have a loop)
```java
for(BibEntry entry:  entries)
{
  if (entry.getField("groups").isPresent())
  {
    show....
  }
}
```

Siedlerchr:
Thank you very much for your contribution! We appreciate this! 👍  Please have a look a the review comments.

tobiasdiez:
I have to admit that I don't really get the idea behind your pull request. Almost all the actions (like mouse up/down, cut + paste, etc) are performed against the main table. So if I press up, I select the previous entry and similarly for the other actions. With your proposed changes, the delete key would get a complete new semantics as it also performs an action against the group panel. I find this a bit confusing and incoherent; but maybe I only missunderstand your intention.

(An example coming from a different domain would be a music player that allows tagging of songs. If you filter your music by a given tag and then hit delete, all the music players I know just delete the song from the library instead of asking you if you want to remove the tag. As in JabRef, the tag/group feature is just for filtering and not like a file directory.)

mpele:
@tobiasdiez I am not sure that I understand what is wrong with synergy between modules. 
If I do not like some song - I do not like it. But if i think that some article is not adequate for some research area (group) it doesn't disqualify it for other research areas (groups). And we are not talking about simply filters, we are talking about entity field that defines static groups that entity belongs to. 

I personally would like the proposed functionality so i made this PR. I understand that something is  expected behavior  https://github.com/JabRef/jabref/issues/2989#issuecomment-314389139 but, for me, it is not intuitive. So, from my point of view, the user should be informed about side effects.

src/main/java/org/jabref/gui/BasePanel.java
mpele: @Siedlerchr Thank you for review. If it is generally acceptable, I will make some more checks.

tobiasdiez:
Ok, I see your point. We just have different views about the role of groups (filter vs basket). So the added functionality is fine for me. But may I ask you to add a checkbox with "Remember choice and do not show dialog again" so that it is easy to hide the additional step/dialog if one is happy with the interpretation "delete key = always remove entry from database". Thanks.

lynyus:
@tobiasdiez what would you say, is this something we should include in 4.0 or even the beta3?

tobiasdiez:
I would say we wait until 4.1 with merging of this PR as we try to converge to a stable version and should try to minimize changes that are not bug fixes.

lenhard:
4.0 is released! What is missing to get this PR into master? As far as I can see it is only to fix the merge conflicts?

mpele:
This PR is not ready for master. The logic and some dialogs should be rewritten.

koppor:
@mpele Do you have time to continue working on this?

mpele:
I will try to finish this functionality in 2-3 weeks. I hope that is not a problem.
And I will do it in new branch because it has been a lot of changes.

lenhard:
@mpele Sounds fine, thanks for following up!

Issue #2998, opened 2017-07-11
The autocompletion does not work perfectly


iTao8:
JabRef 4.0.0-dev--snapshot--2017-07-11--master--e35b6ac2c
Mac OS X 10.11.6 x86_64 
Java 1.8.0_131

The bugs: 
1. The autocompletion does not work in the author and journal fields. 

2. The autocompletion dropdown box gives many same items. When you search something once more, there are two dropdown boxes overlap there.

![screen shot 2017-07-12 at 6 47 45 am](https://user-images.githubusercontent.com/28705547/28094335-9cc7078a-66ce-11e7-8a02-64ab54fa8818.png)
![screen shot 2017-07-12 at 6 48 05 am](https://user-images.githubusercontent.com/28705547/28094343-a2713444-66ce-11e7-9815-9e6af682f719.png)

tobiasdiez:
Thanks for the quick feedback! 

This should be fixed in the latest development version. Could you please check the build from http://builds.jabref.org/master/. Thanks!
Sometimes you still get the same suggestion in the search bar. This happens when the same author is entered in different ways (for example, once as "Tao, I." and also as "Tao, IdontKnow" so that the suggestion is "Tao, I." in both cases). There is nothing I can do at the moment but reporting the issue at []().

Moreover, I couldn't replicate the two overlapping dropdown boxes (on Windows). What are the exact steps to reproduce this problem?

iTao8:
JabRef 4.0-dev--snapshot--2017-07-19--master--894799b2f
Mac OS X 10.11.6 x86_64 
Java 1.8.0_131

The autocompletion dropdown box for the journal field covers the input
![screen shot 2017-07-21 at 6 27 11 pm](https://user-images.githubusercontent.com/28705547/28460120-3a96268c-6e43-11e7-8089-d62c815a35d8.png)

iTao8:
@tobiasdiez To reproduce the overlapping dropdown box, you just need to search an author two or more times. 

It seems that the autocompletion in the search bar can only work for the author, not for journal and some other thing, is it?

My current version is 
JabRef 4.0-dev--snapshot--2017-07-19--master--894799b2f
Mac OS X 10.11.6 x86_64
Java 1.8.0_131

tobiasdiez:
[...]

Issue #3068, opened 2017-08-02
Autocompletion in search field


lc9275:
JabRef 4.0-dev--snapshot--2017-08-02--master--63414b030
Windows 10 10.0 amd64 
Java 1.8.0_66

Steps to reproduce:

1. Type last name of the author say "Lastname"
2. A suggestion pops up "Lastname, F." and you can already see the item you want in the list
3. You hit enter, no item is found

This happens in particular if the author field is encoded as e.g. "F. Lastname and G. Rdsfdsf and Z. Ddslkfjddf" instead of "Lastname, F. and Rdsfdsf, G. and Ddslkfjddf, Z."

lc9275:
As a side note, is there a way to turn auto-completion of? It is very slow and instead of being helpful it just slows me down.

lenhard:
I can confirm this, author autocompletion has a chance of breaking main table selection. However, let me point out that if you don't hit enter, table sorting works as expected and the correct entries are displayed. So as a quick fix for your problem: don't hit enter.

You can turn of the autocompletion for the entry editor under preferences -> Entry editor -> Enable word/name completion. However, that disables the autocompletion only for the entry editor. We should extend it so that it covers the search box as well.

Doc73:
Similar issue:

1. I want to add author in the form "surname, name"
2. I write "surname" --> autocompletion suggests "surname, n."
3. I continue typing "name" and... then? If I press the Enter key, editor prints "surname, n." and the only way to write "surname, name" appears to be to go with the mouse on another field, because other keys (ex. tab) don't work

It is very annoying!

lenhard:
@Doc73 You might want to have a look at the autocompletion preferences under Options->Preferences->Entry Editor.

There's different ways you can set the autocompletion for names. I have selected "Autocomplete names in Firstname Lastname format only" and "Use full firstname whenever possible". With this setting, I cannot reproduce the problem you describe, I always get the full author name.

Moreover, when you do not like the suggested autocompletion, you can just hit the escape key. Then the autocompletion suggestion disappears and there's no need to select a different field.

Doc73:
@lenhard 
Many thanks! I could not find this section in preferences, maybe due to italian translation: "Entry editor" becomes "Modifica voci". I can not imagine a better translation: in this case, English is clearer to me.

I would suggest: "Editor delle voci"

lenhard:
I am glad to hear it helped :)

[...]

PR #3078, opened 2017-08-06
Only update auto completion suggestions upon major changes

tobiasdiez:
Fixes https://github.com/JabRef/jabref/issues/2998. Reused logic from recent change to the backup manager to only propagate mayor changes.


iTao8:
@tobiasdiez I didn't see any improvement. The problems are still there.

lenhard:
@iTao8 Please clarify. This is meant to fix https://github.com/JabRef/jabref/issues/2998#issuecomment-316968028 and from what I see it does. 

As Tobias wrote in the issue, there is nothing to be done about author duplicates and so far no one managed to replicate overlapping dropdown box here (I tried as well and couldn't).

iTao8:
@lenhard 1. Do you mean no one can reproduce the overlapping dropdown box? 
2. I reported another problem in #2998: the autocompletion dropdown box for the journal field covers the input. When you are typing a journal name in the journal filed, the input was covered by the autocompletion dropdown box, so you can not see any inputing. 
3. I find new problems. I have a .bib for all my publications (20 papers) and my name are all listed as "Wang, Tao". When I input "Wang, Tao" in the search bar and click "enter", the input automatically changed to "Wang, T." and only ONE entry was found. 
4. Another problem. If I input "WangTao" in the search bar, and then I found this is wrong and want to move the cursor and insert a space between "Wang" and "Tao", but the left key and right key do not work.

lenhard:
@iTao8 Thanks for clarifying! Here's my feedback:

1. Tobias and me have tried and couldn't reproduce it. I am not aware of anyone else trying, but as long as we can't reproduce this, we can't fix it.

2. Same here, no one could reproduce the overlap in the journal field (I tried just now). Maybe this is all OSX specific? @chochreiner Can you see how these things look on your machine?

3. This is already tracked in #3068 There are two simple solutions to this right now: 1. Don't hit enter. The main table will be sorted as you type anyway. 2. Instead of enter, hit escape. Then the autocompletion suggestions go away.

4. This one I can reproduce :) And I agree with you that the left and right keys should work differently. Could you open a new issue for this? @Siedlerchr This is about key bindings (again). Can you have a look?

tobiasdiez:
@iTao8 Thank you very much for your feedback. 
In fact this PR was not meant to fix any of the issues you mention. I only fixed that you don't get "Te" and "Tes" as autocompletion suggestions if you write "Test" in the field (I thought this is what you meant with "autocompletion covers input").

May I ask you to open a new issue of each of the problems you describe separately. I know this is more work on your side, but would help us to keep track of all the issues and not skip one. Thanks!
(Concerning the overlapping dropdown box: this is maybe a MacOS specific problem).

iTao8:
@lenhard @tobiasdiez Thanks very much! 

The autocompletion dropdown box covers the input, see the screenshot. The dropdown box started at the bottom of the title filed, so all the input letters in the journal field are covered by this box. 
28460120-3a96268c-6e43-11e7-8089-d62c815a35d8

chochreiner:
I can confirm the overlap on MacOS (latest development build) 
screen shot 2017-08-07 at 12 44 55

matthiasgeiger:
Also reproducible on Windows if there is not enough space to display the whole list of suggestions (i.e., the suggestions would be placed over the task bar)

lenhard:
@matthiasgeiger Indeed! Maybe we should just make the size of the drop down list shorter by default.

tobiasdiez:
Its a known bug of the library we use: https://bitbucket.org/controlsfx/controlsfx/issues/701/auto-complete-box-overlaps-textfield-when

PR #3085, opened 2017-08-07
Only load the telementry service as a background task if used

chochreiner:
Up to now, the telemetry service was always running (even if the reporting was turned of).
This initialization already requires a lot of time during startup, which is already rather long on MacOS

I hope I have caught all usages of the telemetry client to avoid nullpointer exceptions, but i would be great if @lynyus could have a look at it.

### DIFF
--- a/src/main/java/org/jabref/Globals.java
+++ b/src/main/java/org/jabref/Globals.java
@@ -1,6 +1,7 @@
 package org.jabref;
 
 import java.awt.Toolkit;
+import java.util.Optional;
 import java.util.UUID;
 
 import org.jabref.collab.FileUpdateMonitor;
@@ -80,8 +81,8 @@ public class Globals {
 
     private static void stopTelemetryClient() {
         if (Globals.prefs.shouldCollectTelemetry()) {
-            telemetryClient.trackSessionState(SessionState.End);
-            telemetryClient.flush();
+            getTelemetryClient().ifPresent(client -> client.trackSessionState(SessionState.End));
+            getTelemetryClient().ifPresent(client -> client.flush());
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/jabref/gui/JabRefDialog.java
```diff
@@ -54,6 +54,8 @@
     }
 
     private <T extends JabRefDialog> void trackDialogOpening(Class<T> clazz) {
-        Globals.getTelemetryClient().trackPageView(clazz.getName());
+        if (Globals.getTelemetryClient()!=null) {
```
koppor: Checkstyle will hit you! 😈 

I think, this should be converted to an Optional, shouldn't it? Then the usual `Globals.getTelemtryClient().ifPresent(client -> client....)`

lynyus:
@chochreiner my macbook has not yet arrived , so I cant test it on MacOs.

chochreiner:
The changes work fine on MacOS. I've just added you, bc you have added the telemetry part and I might have overlooked sth, which should also use the "null" checks.

Siedlerchr:
The failing tests seem to be about the Help pages, otherwise LGTM

koppor:
Go ahead merging, I checked the new commit. The Optional.ofNullable in the
getter should not be a performance issue

Am 07.08.2017 22:04 schrieb "Christoph" <notifications@github.com>:

src/main/java/org/jabref/Globals.java
```diff
@@ -73,12 +74,16 @@ public static void startBackgroundTasks() {
...
+            startTelemetryClient();
+        }
     }
 
     private static void stopTelemetryClient() {
-        telemetryClient.trackSessionState(SessionState.End);
-        telemetryClient.flush();
+        if (Globals.prefs.shouldCollectTelemetry()) {
+            getTelemetryClient().ifPresent(client -> client.trackSessionState(SessionState.End));
+            getTelemetryClient().ifPresent(client -> client.flush());
```
Siedlerchr: If you have only a single method call, you can do some syntactic sugar by doing `ifPresent(TelementryClient::flush)`
https://docs.oracle.com/javase/tutorial/java/javaOO/methodreferences.html

PR #3089, opened 2017-08-09
Add Guava Cache to CitationStyleCache

snisnisniksonah:
This uses Guava's Cache instead of a HashMap for caching.

This enables us to limit the number of cached entries. The current number 1024 is chosen arbitrarily. Maybe, 100 could be enough.

### DIFF
--- a/src/main/java/org/jabref/logic/citationstyle/CitationStyleCache.java
+++ b/src/main/java/org/jabref/logic/citationstyle/CitationStyleCache.java
@@ -1,7 +1,5 @@
 package org.jabref.logic.citationstyle;
 
-import java.util.HashMap;
-import java.util.Map;
 import java.util.Objects;
 
 import org.jabref.model.database.BibDatabaseContext;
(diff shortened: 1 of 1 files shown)


src/main/java/org/jabref/logic/citationstyle/CitationStyleCache.java
```diff
@@ -1,58 +1,61 @@
...
     public CitationStyleCache(BibDatabaseContext bibDatabaseContext) {
         this(bibDatabaseContext, CitationStyle.getDefault());
     }
 
     public CitationStyleCache(BibDatabaseContext bibDatabaseContext, CitationStyle citationStyle) {
-        this.setCitationStyle(citationStyle);
+        this.citationStyle = citationStyle;
+        citationStyleCache = CacheBuilder.newBuilder().maximumSize(CACHE_SIZE).build(new CacheLoader<BibEntry, String>() {
+            @Override
+            public String load(BibEntry entry) throws Exception {
```
Siedlerchr: Can this be converted to a Lambda expression?

koppor: According to https://stackoverflow.com/a/34424577/873282, this is not possible. The basic class has more than one abstract method (see https://github.com/google/guava/blob/master/guava/src/com/google/common/cache/CacheLoader.java#L127).

PR #3090, opened 2017-08-09
Add validation to entry editor

tobiasdiez:
I just remembered that my initial motivation to convert the entry editor to JavaFX came from my desire to display the integrity check messages directly in the entry editor. This PR accomplishes exactly this. 
image
The location and style of the tooltip is not perfect but as of now these things cannot be changed since they are hardcoded in `controlsfx` (will create a feature request).

What do you think, should this still merged before the beta3 release? In the end it is only a minor change and shouldn't break anything.


lenhard:
I have tested it locally and it looks good. I don't think it's a problem to merge it even into beta3, since it is a minor change and no new feature.

I have one request for a usability optimization: the warning sign is very small and it seems that I have to locate the mouse on exactly one pixel so that the message appears. Do you see any chance of increasing the size of the warning sign or making it easier to display the warning message?

tobiasdiez:
No sadly I didn't found a way to specify the size of the graphic. These things are all hard-coded in the controlsfx class (which contains the comment `// TODO we shouldn't hardcode this - defer to CSS eventually`). I created [an issue](https://bitbucket.org/controlsfx/controlsfx/issues/788/make-style-of-graphicvalidationdecoration) and will revist the style once this is possible.

PR #3092, opened 2017-08-10
Fetcher timestamp

### LINKED ISSUE #3074: Adding from DOI does not generate timestamp
JabRef version 
JabRef 4.0-dev--snapshot--2017-08-03--master--e9fbf1963
Linux 4.4.0-21-generic amd64 
Java 1.8.0_144
on
Ubuntu 14.04

Steps to reproduce:

1. 
Check on Options --> Preferences --> General --> Mark new entries with addition date
Check on Options --> Preferences --> General --> Overwrite
Field name: timestamp

2. Add an entry by DOI. I have tried several and the same behavior happens. Here is a sample: 10.1002/jae.2557

3. Entry is added, but timestamp is not populated.

I have confirmed that adding an entry manually still creates a timestamp.

stefan-kolb:
Fixes #3074.


koppor:
+1 for separating a preference class. I am unsure, whether we introduced static getters. Our current documentation in https://github.com/JabRef/jabref/wiki/Code-Howtos#how-to-work-with-preferences is somewhat shortcoming in that regard. I think, `LayoutFormatterPreferences` are a good example about our new way to handle preferences.

I think, the only change required is to let JabRefPreferences have a method `getTimestampPreferences`.

stefan-kolb:
Ok, I tried to create a own preference class. Not sure if I've done that correctly (as intended at the moment). Should I also replace calls to the preferences inside the General tab?

Issue #3052, opened 2017-07-26
Focus entry editor tab when opening entry editor 


Siedlerchr:
Open entry editor.
Tab is not focused, Press CTRL+ PLUS or  CTRL+ Minus to navigate: Notice that focus is still on table and therefore changes the size.

You first have to click on the tab or in a field to focus it and therefore make the navigation work correctly.

halx4:
Hello, I currently taking a shot on this one and would like some clarification.
After opening the entry editor, the focus should go to the tab of entry editor or to a field?

I think the field solution is more sensible as the user can, at this point, start typing as well as changing tab but if the focus goes to tab, the user has to press TAB button to start typing. 
Thank you!

Siedlerchr:
Thanks for your interest in this, I agree the field solution would be more useful. So the focus then should go to the first field on that page.

If you need help or have any further questions on this, just ask.

halx4:
hi Siedlerchr,
Initially I tried to propagate the `requestFocus()` invocation down the hierarchy of nodes, checking in every step that everything is ok. After initialization , I press the "new BibTeX Library" and immediately the 
"new BibTeX entry" button. The focus goes away from the last button I pressed and goes to another item in the UI.
It is working as expected until (inclusively) I request focus from `EntryEditor` (which extends a `JPanel`), meaning when I invoke `super.requestFocus()`, inside `EntryEditor`, the `EntryEditor` gets the focus properly. 
The hierarchy of elements continues as follows:
`EntryEditor` has a `JFXPanel` and inside the `JFXPanel` is a scene which has a `TabPane` wich has `EntryEditorTab`s etc...
The problem is that when I propagate the `requestFocus()` call on a child of `EntryEditor`, the focus does not change (it stays on the last button I pressed).
I have checked that all the parents of the nodes I try to give focus to are visible which is a requirement according to javadocs. I tried enclosing `requestFocus()` callings in `Platform.runLater();` with no luck.

I notices the UI hierarchy consists of mixed swing and javaFX elements and maybe this has to do something with my problem. 

I created a branch to demonstrate the problem named *problemDemonstration*. I would expect the focus to go to the EntryEditor's first tab but it does not.

Any help would be greatly  appreciated! Thanks!

Siedlerchr:
Hi,
thanks for investigating this. You are probably right with the focus problem being part of the swing/javafx interoperability.
Maybe it's related to this bug?
[...]

PR #3095, opened 2017-08-10
Fix for issue 3052

halx4:
This is an atempt to fix the problem with entry editor focus as described in #3052. In detail, when `EntryEditor` opens up, it is not focused automatically. The desired behavior is to focus the first `Textfield` on any of the `EntryEditor`'s tabs.

The way the fix is implemented: when `EntryEditor` is opened, the `JFXPanel` in which is the `TabPane` of the `EntryEditor` is getting the focus, The focus is given implicitly to a focusTraversable node contained in the `JFXPanel`. The selection of the node seems to be "random" (even though consistent for a given tree of nodes). As far as I've tested it consistently focuses the first `Textfield` on any tab, which is the desired behavior. I make this pull requested as suggested in #3052 discussion by @Siedlerchr .

I should mention that I ran `gradle check` which results in "6 tests failed" in my setup but this was also the result when I ran the test before making any change.


tobiasdiez:
Thanks for the PR!

PR #3096, opened 2017-08-10
Followup to 2663 Regexformatter

snisnisniksonah:
The Regexformatter now handles protected groups correctly. This is a followup to #2663. There, the regex modifier was introduced and a minor issue with shorttitle was fixes


src/main/java/org/jabref/logic/formatter/Formatters.java
```diff
@@ -68,7 +70,15 @@ private Formatters() {
 
     public static Optional<Formatter> getFormatterForModifier(String modifier) {
         Objects.requireNonNull(modifier);
-        Optional<Formatter> formatter = ALL.stream().filter(f -> f.getKey().equals(modifier)).findAny();
+        Optional<Formatter> formatter;
+
+        if (modifier.matches("regex.*")) {
+            String regex = modifier.substring(5);
```
Siedlerchr: please define this magic number somewhere as constant

src/main/java/org/jabref/logic/formatter/bibtexfields/RegexFormatter.java
```diff
@@ -0,0 +1,95 @@
...
+
+    @Override
+    public String getExampleInput() {
+        return "Please replace the spaces";
+    }
+
+    public static void setRegex(String rex) {
+        // formatting is like ("exp1","exp2"), we want to remove (" and ")
+        String rexToSet = rex;
+        rexToSet = rexToSet.substring(2, rexToSet.length() - 2);
```
Siedlerchr: please define this magic number as constant with a useful name, so it is clear what it does

src/main/resources/l10n/JabRef_da.properties
```diff
@@ -950,6 +952,8 @@ Reference_library=Referencelibrary
 
 Refine_supergroup\:_When_selected,_view_entries_contained_in_both_this_group_and_its_supergroup=Undergruppe\:_Vis_poster_indeholdt_både_i_denne_gruppe_og_gruppe_over
 
+Regex=
```
Siedlerchr: Regex is the same as regular expression, I would reuse that existing translation

src/main/java/org/jabref/logic/formatter/Formatters.java
```diff
@@ -65,6 +65,8 @@
 
     public static final List<Formatter> ALL = new ArrayList<>();
 
+    private static final int LENGTH_OF_REGEX_PREFIX = 5;
```
koppor: Can `5` be replaced by `"regex".length()``?

src/main/java/org/jabref/logic/formatter/bibtexfields/RegexFormatter.java
```diff
@@ -87,7 +90,7 @@ public String getExampleInput() {
     public static void setRegex(String rex) {
         // formatting is like ("exp1","exp2"), we want to remove (" and ")
         String rexToSet = rex;
-        rexToSet = rexToSet.substring(2, rexToSet.length() - 2);
+        rexToSet = rexToSet.substring(REMOVE_OUTER_QUOTES, rexToSet.length() - REMOVE_OUTER_QUOTES);
```
koppor: I would replace the first `REMOVE_OUTER_QUOTES` by `"\"(".length()` and the second one by `")\"".length()`. Maybe as two variables `LENGTH_OF_QUOTE_AND_OPENING_BRACE` and `LENGTH_OF_CLOSING_BRACE_AND_QUOTE`.

Issue #3063, opened 2017-08-01
Tab "Other" does not get updated after editing the biblatex source


koppor:
Steps to reproduce:

1. Open an entry in the entry editor. The entry must have an other field such as `bibsource`
2. Switch to biblatex source
3. change `booktitle` to `mybooktitle`
4. Go to "Other fields"
5. "mybooktitle" does not appear as field

Refs #2002, #2064

Example entry:

```
@InProceedings{DBLP:conf/zeus/KoppLUW11,
  author      = {Oliver Kopp and Frank Leymann and Tobias Unger and Sebastian Wagner},
  title       = {Towards The Essential Flow Model},
  bibsource   = {dblp computer science bibliography, http://dblp.org},
  booktitle = {3rd Central-European Workshop on Services and their Composition, Services und ihre Komposition, {ZEUS} 2011, Karlsruhe, Germany, February 21-22, 2011. Proceedings},
}
```

lenhard:
@koppor Can you please check if #3097 solves this problem for you?

matthiasgeiger:
Fixed by #3097

PR #3097, opened 2017-08-11
Store changes in source tab when new entry is selected

### LINKED ISSUE #3086: Changes in Source Tab are not saved correctly without switching to another tab
JabRef version 4.0-beta2 on Debian 9 Stretch AMD64

Steps to reproduce:

1. Add a book entry
2. Add `title`, `author` etc. and save
3. Add `subtitle` and save
4. Switch to another entry
5. Return to the previous entry --> `subtitle` disappeared!!!!

It is very annoying and happens also with other fields.

lenhard:
Fixes #3086 

The problem was that the source tab only stored changes on focus change and this did not happen when you just select a new entry in the main table.

The PR modifies an attribute in the `EntryEditor` (`movingToDifferentEntry`) and turns it into a `BooleanProperty` so that you can bind to it and react to changes.

EDIT: Also fixes a second (unreported) issue: Now the base panel is marked as dirty when you type something in the source tab.

EDIT 2: Fixes also  #3063

Now the Other fields tab is rebuilt whenever an other field is newly added or removed at the BibEntry. This is achieved with the event system and a new subtype of FieldChangedEvent.


src/main/java/org/jabref/gui/entryeditor/EntryEditor.java
```diff
@@ -202,6 +210,58 @@ public void keyPressed(java.awt.event.KeyEvent e) {
...
+        if (isOtherFieldsTabSelected) {
+            tabbed.getSelectionModel().select(tab);
+        }
+    }
+
+    private void addNewOtherFieldsTabIfNeeded() {
+        OtherFieldsTab tab = new OtherFieldsTab(frame, panel, entryType, this, entry);
+        if (tab.getFields().size() > 0) {
+            // add it at default index, but that is just a guess
+            int defaultIndex = 4;
```
Siedlerchr: Please add a constant for the magic number

lenhard: Done

Issue #2857, opened 2017-05-21
Group membership not indicated by colour if JabRef is opened and entry is selected immediately


AEgit:
JabRef 4.0.0-dev--snapshot--2017-05-20--master--01e854829
Windows 10 10.0 amd64 
Java 1.8.0_131

1. Open a large database (e.g, >6000 entries and several 100 static groups)
2. If you don't wait for a couple of seconds, before you start selecting an entry but instead choose to immediately select one, once the databse is visible (i.e., the graphical user interface of JabRef has appeared), the respective group of the selected entry WILL NOT be indicated by the green number
3. Once that has happened, no matter what entry you select, its group membership will not be indicated by colours in the groups panel
4. To overcome this issue, you have to close JabRef, then reopen it and wait for a couple of seconds, after having opened the database. I guess that some modules (?) are still loaded in the background - if you are too quick in using JabRef, the loading of those modules will be stopped, leading to this behaviour (which was, indeed, quite difficult to replicate, but I think I've now found the pattern)

AEgit:
JabRef 5.0-dev--snapshot--2019-09-18--master--1601770f4
Windows 10 10.0 amd64 
Java 11.0.4-bellsoft

This appears to be fixed in the latest dev version. Cheers!

PR #3105, opened 2017-08-15
Remove entries from prior groups if move operation is selected

### LINKED ISSUE #3101: Move to group works like add to group
JabRef version 4.0-beta2 on Ubuntu 17.04 with oracle JVM

Steps to reproduce:

1. Add an entry to a group
2. Right click and press move to group
3. The entry is added to the new group but not removed from the old one

lenhard:
Fixes #3101

There is one problem left to conclude this. It does the proper operation on the selected entries and updates the group tree. However, it does not update the coloring in the group tree. @tobiasdiez Could you give me a hint on how I can achieve this?


lenhard:
Unfortunately, using the `StateManager` does not solve the coloring for me.

But to be honest, it seems rather broken anyway, irrespective of this PR. I have a database with 3 entries and 2 groups. The only way to get a coloring is to click an entry with a group in a newly started JabRef without opening the entry editor. That coloring never changes even if I click different entries. Does coloring actually work for you?

I think this PR is ready to go. Someone else should check the fix in a running JabRef. The coloring problem should be investigated in another issue/PR.

tobiasdiez:
In general it works for me but I also noticed that sometimes the color is not updated. Probably related to https://github.com/JabRef/jabref/issues/2857.

PR #3107, opened 2017-08-15
New home for snap -> buildres/snapcraft

koppor:
This addresses https://github.com/JabRef/jabref/commit/c2fb270eefda2b9d16ec656c924a5e2c6d759c56#commitcomment-23656145

The normal home for snapcraft files (especially, `/snap/gui/jabref.deskop`) is `/snap`. See https://snapcraft.io/docs/build-snaps/metadata#fixed-assets. I could not find documentation, why `/snap/.snapcraft` was generated.

With this PR, we do not follow the snap convention, but enforce our own folder layout by placing stuff in `buildres`. To make the use of `snapcraft` explicit, I also named the folder `snapcraft` instead of `snap`.


lynyus:
Is this ready for review? I guess this is not relevant for the beta-3 anyways?
Also, at some point maybe this should be documented in the wiki.

koppor:
Build goes through (locally using https://hub.docker.com/r/koppor/snapcraft-java-de/), so this change really works.

Issue #919, opened 2016-03-08
Integrate journal abbreviations into JabRef


koppor:
Currently, the repository https://github.com/JabRef/reference-abbreviations contains several abbreviation lists. They should be maintained in JabRef itself and should be loaded automatically on startup.
- [ ] check performance
- [ ] check whether the files are disjunct
- [ ] check that the abbreviations are correct
- [ ] merge https://github.com/JabRef/reference-abbreviations into JabRef's code

oscargus:
As a big fan of using strings for journal abbreviations, I also believe that it would be nice to define strings for the journals, similar to how it already works for the IEEE journals. The potential drawback is that we also must provide a .bib-file with all those strings, in both abbreviated and non-abbreviated form, but it would be quite nice. Also, this might help in showing the correct name in the table, while showing the string in the entry editor.

koppor:
@oscargus, if I understand correctly, you propose to completely drop the abbrevation system and to maintain `bib` files with `@string`s?

oscargus:
From my perspective, yes, but my suggestion was really only to enable it, not dropping it. Using the IEEE-version one toggles between Full Journal Name -> Abbr. J. Name -> #string# and back again. To me that seems like the best option, those who want can use full or abbreviated and I can use strings. :-)

On the other hand with the dual bib-files discussed in other issues and a "complete" list of abbreviations it is less of an issue. The main reason for using strings is to get consistent spelling/naming/abbreviation of journals (and conference proceedings titles).

This also relates to #323

lenhard:
@koppor The journal abrreviation feature is implemented, right? I recall that @tobiasdiez recently improved its performance. So this issue can be closed, right?

And this means that #323 can also be closed, right?

koppor:
[The feature](http://help.jabref.org/en/JournalAbbreviations) is available [since version 2.0b](https://github.com/JabRef/jabref/blob/v2.11.1/CHANGELOG#L1050). My issue is we have **multiple** lists:

1. In the code:

![grafik](https://user-images.githubusercontent.com/1366654/29402879-db9b6c74-8336-11e7-9f8e-1bdcea26c5b2.png)

2. Some other lists at external repository:

https://github.com/JabRef/abbrv.jabref.org

To make the lists more accessible, I proposed to merge these lists into JabRef (fourth bullet point). For that, quality of these lists has to be checked (second and third bullet point).

Regarding the performance, I haven't checked and I am not sure whether we have a JMH test for that.

[...]

PR #3123, opened 2017-08-18
[WIP] Integrate journal abbreviations into JabRef

lenhard:
Meetings are good for easy programming tasks...

This takes care of parts of #919. It takes the existing built-in journal lists and transforms them from text files into source code. We've found the abbreviation loading to be a performance problem. This should reduce this problem a little, because we replace file parsing with class loading and the reading of the lists can be optimized by the compiler. Changes are completely internal to the abbreviation code.

This doesn't check the lists for consistency, the other parts of #919. It just takes the existing lists and turn them into source code.

This doesn't need to be merged into 4.0, but can be included later.

EDIT: Turns out our built-in list has 15195. This is too large for a Java class and I am getting the error `Error:(8, 45) java: code too large` on compilation. So we need to think about this. We can leave the built-in list as a text file (no performance improvement) or distribute it into multiple source code files (seems rather brittle).


koppor:
In case we want to keep the Java-class, I would add a transformation transforming the list into Java-Classes. End users are maintaining the abbreviation lists (https://github.com/JabRef/abbrv.jabref.org/pulls?q=is%3Apr+is%3Aclosed) and I assume, they do not want to maintain Java-Code.

Does loading of the lists still take that much time? Maybe we should ship [SQLite](https://www.sqlite.org/)?

lenhard:
But those are different lists, right? This PR was just about the integration of the built-in lists. For the gigantic list, there's no clean way to convert this into code, so we might have to load it anyway. The IEEE lists could go into code, but they probably don't matter performance-wise anyway.

I am not even sure if this PR is worth following up.

koppor:
Yes, these are different lists - and this also confuses users of JabRef. Why is there https://github.com/JabRef/jabref/blob/master/src/main/resources/journals/journalList.txt and http://abbrv.jabref.org/.

Current state: On http://abbrv.jabref.org/ at the end, the internal list is linked somehow...

I would not follow up here, but possible follow up in the same context:

A) support three kinds of modes: full, abbreviated, shortest unique abbreviation
B) add support for abbreviating conference names.

Siedlerchr:
What about storing them as Properties file?   That would be actual equivalent of storing it into code
This was suggested here: https://stackoverflow.com/a/2407930
And here is an analysis of Performance: https://stackoverflow.com/a/10756739

lenhard:
@Siedlerchr This sounds like a good suggestion.

Issue #3117, opened 2017-08-16
The General tab in the entry editor is not scrollable.


bernhard-kleine:
JabRef version JabRef 4.0-beta3
Windows 7 6.1 amd64 
Java 1.8.0_131

The display is cluttered!

![grafik](https://user-images.githubusercontent.com/9216487/29386744-6f24ff8c-82dd-11e7-9a61-6a93a339b968.png)

lenhard:
@bernhard-kleine Thanks for reporting.

The fields in a tab of the entry editor are shrunk to fit in the display space that's available to that tab. When there's not enough space for all fields, a scroll pane for the complete tab should appear (it does for me).

What would you suggest to change to make the display less cluttered in the general tab?

bernhard-kleine:
@lenhard 
When there is scrolling possible why not here, the keywords field is not properly visible, so there should be scrolling.

lenhard:
Ok, thanks. The layouting is done by a library we are using, so I have no immediate solution. Maybe it's possible to increase the field size to two rows by default.

lenhard:
@bernhard-kleine Could you perhaps try out this version http://builds.jabref.org/increase-textarea-size/ It assigns a minimum height to all text fields, so hopefully keywords should be more readable and the scroll bar should appear more easily. Do you think that's better? Would you want an even bigger minimum size or a smaller one?

bernhard-kleine:
IMHO this looks much better.

lenhard:
Perfect! The PR is merged now, so we can closed this issue.

Issue #2941, opened 2017-06-24
Linked files are badly aligned


jonasstein:
The text in form fields is badly aligned and has sometimes a grey background.

JabRef 4.0.0-dev--snapshot--2017-06-18--master--c0ca85e07
Linux 4.9.16-gentoo+ amd64 
Java 1.8.0_131

![selection_302](https://user-images.githubusercontent.com/298682/27512441-f960f09c-593f-11e7-89ac-a0e419b2501c.png)

tobiasdiez:
The display of linked files is also not ideal on Windows. I already tried a few days ago to fix these alignment problems but wasn't successful (all text fields and other controls in the editor appear to have a negative top margin - no idea why). If somebody else wants to play around, he is happily invited to do so. For the moment, I wouldn't assign it to the 4.0 milestone since it is not critical (the information is readable, just not as nice as is could be).

sauliusg:
I can confirm the bug on:
JabRef 4.0-dev--snapshot--2017-07-21--master--46f58d6
LinuxMint-18.1
Java 1.8.0_131-8u131-b11-0ubuntu1.16.04.2-b11
Sure, not critical, but important to attract new users ;)
Cheers,
Saulius

lenhard:
Since this is not a functional error, but UI polishing, I'll relabel it from "bug" to "enhancement".

lenhard:
@jonasstein @sauliusg  Could you perhaps try out this version http://builds.jabref.org/increase-textarea-size/ It assigns a minimum height to all text fields, so hopefully file attachments should be more readable. It just means that you might have to scroll in the general tab. Do you think that's better? Would you want an even bigger minimum size or a smaller one?

lenhard:
We consider this fixed thanks to #3141, so I am closing this issue.

PR #3141, opened 2017-08-20
Increase size of file and keywords editors

lenhard:
In the new EntryEditor, we have the problem that sometimes text in the TextAreas is hard to read, because the areas are very narrow. The reason for this is that Java tries to fit them in the available display space. Examples: #3130 #3117 #2941

This PR sets a minimum size for all TextAreas in the EntryEditor. The value of 70px is rather arbitrary and I'll be happy to get some thoughts if this is good or if more or less is desirable. With this minimum size all TextAreas are nicer to read, even in the cramped up General tab. The only "problem" is that you now need to scroll in the General tab most of the time. 

All in all I think this should definitly go into 4.0!


tobiasdiez:
To be honest, I'm not a big fan of the increased size. I prefer to have as much information on the screen as possible and not to have to scroll. On the other hand, I acknowledge that the current layout is also suboptimal as the text fields are not expanding on their own when they contain much information. Maybe it is already enough to increase the `prefered` height (instead of minimum)?

lenhard:
@tobiasdiez Ok, you have the same opinion as @koppor I guess.

Increasing the preferred height didn't help unfortunately. We can leave most field editors as they are, but we have to increase the size of the file and the the keywords editor. Those are the two editors in the cramped general tab where users could enter information with more than a single line. That doesn't look good at all and really does annoy users.

I've tried to change just these two editors in code, but failed. Whenever I call `setMinWidth` in https://github.com/JabRef/jabref/blob/master/src/main/java/org/jabref/gui/fieldeditors/LinkedFilesEditor.java or change the `minHeight` in the related fxml file, the min height of all field editors in the tab increase. Any hint?

tobiasdiez:
The problem is probably https://github.com/JabRef/jabref/blob/master/src/main/java/org/jabref/gui/entryeditor/FieldsEditorTab.java#L166, where the rows are set to have the same height.

matthiasgeiger:
Quick idea: If the row constraints are the problem, it might be an option to use two rows for the File/Keyword editors

lenhard:
@matthiasgeiger Not sure if that's easily possible. The best solution is probably to write new subclass of EntryEditorTab.

tobiasdiez:
Or extend the `FXFieldEditor` by a `getWeight` method, which you can use in https://github.com/JabRef/jabref/blob/master/src/main/java/org/jabref/gui/entryeditor/FieldsEditorTab.java#L114.

lenhard:
Ok, I've implemented a hack that assigns twice the space of normal rows to the keyword and file field editors.

As I said, this is a hack. But before I turn this into something more acceptable (along your suggestion @tobiasdiez ), I would like to get confirmation from the team that this is a solution that we can live with visually in the JabRef UI. If not, than I am just wasting my time with GUI hacking.

lenhard:
And now I refactored it anyway... JabRef continues to ruin my lunch time.

koppor:
I tried out the new solution locally and it looks good enough. On two-column tabs, the things shrink too much, but we discussed in the devcall that it is difficult to determine the right "min-height". (screen resolution etc).

So, good to merge.

lenhard:
As discussed in the dev call, I am merging this now.

PR #3147, opened 2017-08-22
Export pdf/linked files 

### LINKED ISSUE #2539: Call for function: export the pdfs of all the selected entrys to a new folder

Siedlerchr:
Fixes #2539 

Followup from #2706  because somehow my new changes did not update the PR correctly.

grafik


src/main/java/org/jabref/gui/actions/ExportLinkedFilesAction.java
```diff
@@ -0,0 +1,120 @@
...
+import org.jabref.model.entry.LinkedFile;
+import org.jabref.model.util.FileHelper;
+import org.jabref.model.util.OptionalUtil;
+import org.jabref.preferences.JabRefPreferences;
+
+import org.controlsfx.dialog.ProgressDialog;
+
+public class ExportLinkedFilesAction extends AbstractAction {
+
+    private final BiFunction<Path, Path, Path> resolvePathFilename = (p, f) -> {
```
lenhard: Please use meaningful names (not p, f) even in lambdas

src/main/java/org/jabref/gui/actions/ExportLinkedFilesAction.java
```diff
@@ -0,0 +1,120 @@
...
+        return p.resolve(f.getFileName());
+    };
+    private final DialogService ds = new FXDialogService();
```
lenhard: Same here, better name please.

src/main/java/org/jabref/gui/actions/ExportLinkedFilesAction.java
```diff
@@ -0,0 +1,120 @@
...
+        List<BibEntry> entries = JabRefGUI.getMainFrame().getCurrentBasePanel().getSelectedEntries();
+
+        Optional<Path> path = DefaultTaskExecutor
+                .runInJavaFXThread(() -> ds.showDirectorySelectionDialog(dirDialogConfiguration));
+
+        BibDatabaseContext databaseContext = JabRefGUI.getMainFrame().getCurrentBasePanel().getDatabaseContext();
+
+        long totalFilesCount = entries.stream().flatMap(entry -> entry.getFiles().stream()).count();
+
+        Service<Void> service = new Service<Void>() {
```
lenhard: The enclosing `actionPerformed()` is a really big method. Please do some refactoring and make it smaller. The first thing would be to extract this class, at least to an inner class instead of an anonymous one.

src/main/java/org/jabref/gui/actions/ExportLinkedFilesAction.java
```diff
@@ -0,0 +1,120 @@
...
+                        return null;
+                    }
+                };
+            }
+        };
+
+        service.start();
+
+        DefaultTaskExecutor.runInJavaFXThread(() -> {
+            ProgressDialog dlg = new ProgressDialog(service);
```
lenhard: And again, more meaningful names please :-)

koppor:
Is it possible to add this to the context menu? I once added "Send as email" to the menu as I also needed a similar functionality for sending entries as email:

grafik

koppor:
@Siedlerchr Your PR did not update as you named the branches differently: `exportPdf` vs.  `exportPDF`. The letters `D` and `F` are cased differently. Thus, these are different branches. 🎉

koppor:
Current accessibility:

grafik

Reading https://github.com/JabRef/jabref/pull/2706#issuecomment-292110772 and the following comments, I do not like the word "Export" in the Tools menu. It took me 5 minutes to find the functionality.

1. Context menu: No entry for that
2. File menu: No entry for that
3. File -> Export: Cannot select "PDFs" somehow.

Suggestion:

Rename to "Copy attached files to folder..."

Further minor issues

- [ ] progress always shows "exporting file 0 of x"
  grafik
- [ ] Last chosen folder is not remembered. The browse-dialog always starts in `%HOME`
- [ ] Pressing "Cancel" when choosing the folder also triggers the export.

lynyus:
Added devcall label to discuss milestone

koppor:
Stays at v4.1 milestone as v4.1 will be our PDF feature release.

Siedlerchr:
I think I fixed all issues raised by @koppor  and I also create a log file in the export dir.

src/main/java/org/jabref/JabRefMain.java
```diff
@@ -51,6 +51,7 @@ public static void main(String[] args) {
     public void start(Stage mainStage) throws Exception {
         Platform.setImplicitExit(false);
         SwingUtilities.invokeLater(() -> start(arguments));
+
```
lenhard: Why is there a blank line here?

src/main/java/org/jabref/gui/actions/ExportLinkedFilesAction.java
```diff
@@ -0,0 +1,164 @@
...
+            Button cancelButton = (Button) dialogPane.lookupButton(ButtonType.CANCEL);
+            cancelButton.setOnAction(evt -> {
+                service.cancel();
+                progressDialog.close();
+            });
+            progressDialog.showAndWait();
+        });
+    }
+
+    private class ExportService extends Service<Void> {
```
lenhard: This is rather nitpicky, but what about extracting this into a separate class file in the same package with default visibility? You could also extract parts of `call()` into separate private methods. All this wouldn't change anything about the functionality, but it would make the code a little less chunky and more readable.

src/main/java/org/jabref/JabRefMain.java
Siedlerchr: Oops.

src/main/java/org/jabref/gui/actions/ExportLinkedFilesAction.java
Siedlerchr: I see no reason why I should extract that in a new class. It's just a single Service for one purpose.
But I think I can surely extract some private methods

src/main/java/org/jabref/gui/actions/ExportLinkedFilesAction.java
```diff
@@ -0,0 +1,169 @@
...
+            Service<Void> exportService = new ExportService();
+            startServiceAndshowProgessDialog(exportService);
+        });
+    }
+
+    private <V> void startServiceAndshowProgessDialog(Service<V> service) {
+        DefaultTaskExecutor.runInJavaFXThread(() -> {
+            service.start();
+
+            ProgressDialog progressDialog = new ProgressDialog(service);
```
tobiasdiez: Can you please extract the "show progress dialog" part of this method to the `DialogService` class.

src/main/java/org/jabref/gui/actions/ExportLinkedFilesAction.java
```diff
@@ -0,0 +1,169 @@
...
+                    updateProgress(0, totalFilesCount);
+
+                    try (BufferedWriter bw = Files.newBufferedWriter(exportPath.get().resolve(LOGFILE), StandardCharsets.UTF_8)) {
+
+                        for (int i = 0; i < entries.size(); i++) {
+
+                            List<LinkedFile> files = entries.get(i).getFiles();
+
+                            for (int j = 0; j < files.size(); j++) {
+                                updateMessage(Localization.lang("Copying file %0 of BibEntry %1", Integer.toString(j + 1), Integer.toString(i + 1)));
```
tobiasdiez: I think we never used `BibEntry` before in the user interface (it is the class name...). What about only "entry" or "item"?

src/main/java/org/jabref/gui/actions/ExportLinkedFilesAction.java
```diff
@@ -0,0 +1,169 @@
...
+                                Thread.sleep(500); //TODO: Adjust/leave/any other idea?
```
tobiasdiez: This sleep makes the whole progress of copying files really slow. I would remove it even through then the user has no chance to actually read the message for small files (but that is ok in my opinion).

src/main/java/org/jabref/gui/actions/ExportLinkedFilesAction.java
```diff
@@ -0,0 +1,169 @@
...
+
+                                String fileName = files.get(j).getLink();
+                                Optional<Path> fileToExport = FileHelper.expandFilenameAsPath(fileName,
```
tobiasdiez: Use 
https://github.com/JabRef/jabref/blob/fae290caa38aedbea3a2d093419e5ade4e62d620/src/main/java/org/jabref/model/entry/LinkedFile.java#L131

src/main/java/org/jabref/gui/actions/ExportLinkedFilesAction.java
```diff
@@ -0,0 +1,169 @@
...
+                int numberError;
+                Optional<Path> newPath;
+
+                @Override
+                protected Void call()
+                        throws InterruptedException, IOException {
+                    updateMessage(Localization.lang("Copying files..."));
+                    updateProgress(0, totalFilesCount);
+
+                    try (BufferedWriter bw = Files.newBufferedWriter(exportPath.get().resolve(LOGFILE), StandardCharsets.UTF_8)) {
```
tobiasdiez: In my opinion, a log file is an inconvenient solution for the user. Why don't log all the errors to an internal list and display this list at the end in a new dialog?

src/main/java/org/jabref/logic/exporter/ExportFormat.java
```diff
@@ -313,12 +313,6 @@ public void performExport(final BibDatabaseContext databaseContext, final String
 
     }
 
-    @Override
-    public void performExport(final BibDatabaseContext databaseContext, Path file, final Charset encoding,
```
tobiasdiez: Please don't delete this `path` based methods. I know they are not yet used but they are actually preferable to the string or file-based methods (these should actually be deprecated).

src/main/java/org/jabref/gui/menus/RightClickMenu.java
```diff
@@ -102,6 +103,7 @@ public RightClickMenu(JabRefFrame frame, BasePanel panel) {
 
         add(new GeneralAction(Actions.SEND_AS_EMAIL, Localization.lang("Send as email"), IconTheme.JabRefIcon.EMAIL.getSmallIcon()));
         addSeparator();
+        add(new ExportLinkedFilesAction());
```
tobiasdiez: Do you really think that this export functionality is used that often that we should add it to the right click menu?

src/main/java/org/jabref/gui/actions/ExportLinkedFilesAction.java
tobiasdiez: One reason is that you could (and should) write tests for the `ExportService`.

src/main/java/org/jabref/logic/exporter/ExportFormat.java
Siedlerchr: I deleted it on purpose because it is not working at all as I explained in my other PR or earlier.
It always lead to an exception.

src/main/java/org/jabref/gui/menus/RightClickMenu.java
Siedlerchr: It was requested by @koppor  And I think it makes sense to have it in the context menu

Siedlerchr:
@tobiasdiez  I followed your idea and added a TableView showing the log messages and the file.
I plan to replace the true/false values with Icons: (green checkmark and red x)
grafik

tobiasdiez:
Looks good! Do you want to display this dialog everytime, even through everything went fine and all files were exported successfully? Or just when an error occurs? I'm unsure but would tend to the latter.

src/main/java/org/jabref/gui/actions/ExportLinkedFilesAction.java
koppor: Not fixed.

Siedlerchr:
From my point of view, all remaining issues should be resolved and this can be merged.

src/main/java/org/jabref/gui/actions/CopyFilesResultViewModel.java
```diff
@@ -0,0 +1,50 @@
...
+import de.jensd.fx.glyphs.materialdesignicons.MaterialDesignIcon;
+
+public class CopyFilesResultViewModel {
+
+    private final StringProperty file = new SimpleStringProperty("");
+    private final ObjectProperty<MaterialDesignIcon> icon = new SimpleObjectProperty<>(MaterialDesignIcon.ALERT);
+
+    private final StringProperty message = new SimpleStringProperty("");
+
+    public CopyFilesResultViewModel(String file, boolean success, String message) {
```
tobiasdiez: maybe change the name `Result` -> `ResultItem` to make it clearer that is one item in the result window.

tobiasdiez: Please change the file from string to path

src/main/java/org/jabref/gui/actions/CopyFilesResultViewModel.java
```diff
@@ -0,0 +1,50 @@
...
+            this.icon.setValue(MaterialDesignIcon.CHECK);
+        }
+
+    }
+
+    public StringProperty getFile() {
+        return file;
+    }
+
+    public void setFile(String file) {
```
tobiasdiez: are these setters needed? I find an immutable solution better.

src/main/java/org/jabref/gui/actions/ExportLinkedFilesAction.java
```diff
@@ -0,0 +1,143 @@
...
+        });
+    }
+
+    private void startServiceAndshowProgessDialog(Service<List<CopyFilesResultViewModel>> exportService) {
+        DefaultTaskExecutor.runInJavaFXThread(() -> {
+            exportService.setOnSucceeded(value -> {
+                DefaultTaskExecutor.runInJavaFXThread(() -> showDialog(exportService.getValue()));
+            });
+            exportService.start();
+            DialogService dlgService = new FXDialogService();
```
tobiasdiez: you already have a dialog service at class level -> reuse?

src/main/java/org/jabref/gui/actions/ExportLinkedFilesAction.java
```diff
@@ -0,0 +1,143 @@
...
+            dlgService.showCanceableProgressDialogAndWait(exportService);
+
+        });
+    }
+
+    private void showDialog(List<CopyFilesResultViewModel> data) {
+        Dialog<ButtonType> dlg = new Dialog<>();
```
tobiasdiez: Can you please extract the result dialog to a seperate fxml file, as all the other dialogs.

src/main/java/org/jabref/gui/actions/ExportLinkedFilesAction.java
```diff
@@ -0,0 +1,143 @@
...
+        TableColumn<CopyFilesResultViewModel, String> colFile = new TableColumn<>(Localization.lang("File"));
+        TableColumn<CopyFilesResultViewModel, MaterialDesignIcon> colIcon = new TableColumn<>(Localization.lang("Status"));
+        TableColumn<CopyFilesResultViewModel, String> colMessage = new TableColumn<>(Localization.lang("Message"));
+
+        colFile.setCellValueFactory(cellData -> cellData.getValue().getFile());
+        colMessage.setCellValueFactory(cellData -> cellData.getValue().getMessage());
+        colIcon.setCellValueFactory(cellData -> cellData.getValue().getIcon());
+
+        colIcon.setCellFactory(column -> {
+            return new TableCell<CopyFilesResultViewModel, MaterialDesignIcon>() {
```
tobiasdiez: I think there is a table cell factory helper (similar to list view). If not, then I would be nice if you could create one.

src/main/java/org/jabref/gui/actions/ExportLinkedFilesService.java
```diff
@@ -0,0 +1,125 @@
...
+import org.jabref.model.entry.LinkedFile;
+import org.jabref.model.util.OptionalUtil;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+public class ExportLinkedFilesService extends Service<List<CopyFilesResultViewModel>> {
+
+    private static final Log LOGGER = LogFactory.getLog(ExportLinkedFilesAction.class);
+    private static final String LOGFILE = "exportLog.log";
```
tobiasdiez: I think the log file is not superfluous with the new result dialog, isn't it?

src/main/java/org/jabref/gui/actions/ExportLinkedFilesService.java
```diff
@@ -0,0 +1,125 @@
...
+import org.jabref.logic.util.io.FileUtil;
+import org.jabref.model.database.BibDatabaseContext;
+import org.jabref.model.entry.BibEntry;
+import org.jabref.model.entry.LinkedFile;
+import org.jabref.model.util.OptionalUtil;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+public class ExportLinkedFilesService extends Service<List<CopyFilesResultViewModel>> {
```
tobiasdiez: I'm not sure but wasn't a service intended to be reused while a task is for single work items? I think `Task` would fit better here.

src/main/java/org/jabref/logic/exporter/ExportFormat.java
tobiasdiez: ok...

src/main/java/org/jabref/gui/actions/ExportLinkedFilesService.java
Siedlerchr: I still like the idea of having a log file there.

src/main/java/org/jabref/gui/actions/ExportLinkedFilesAction.java
Siedlerchr: Will do later. All other issues should be resolved now

src/main/java/org/jabref/gui/actions/ExportLinkedFilesService.java
Siedlerchr: Yeah, I did not really know the difference

src/main/java/org/jabref/gui/copyfiles/CopyFilesDialogViewModel.java
```diff
@@ -0,0 +1,28 @@
...
+
+import org.jabref.gui.AbstractViewModel;
+
+public class CopyFilesDialogViewModel extends AbstractViewModel {
+
+    private final SimpleListProperty<CopyFilesResultItemViewModel> copyFilesResultItems = new SimpleListProperty<>(
+            FXCollections.observableArrayList());
+
+    //TODO: How do I add the data here?
+    public CopyFilesDialogViewModel(List<CopyFilesResultItemViewModel> items) {
```
Siedlerchr: How do I get the data here? Should I add a field/property to the state Manager?
This is still one of the things which freaks me out of our javafx concept and for which I can't get my head around

tobiasdiez: Yes, this is a bit ugly.

In the view you create an "injection context"
https://github.com/JabRef/jabref/blob/455965e931ca347e5caf91f52ded8e42cf11881c/src/main/java/org/jabref/gui/entryeditor/fileannotationtab/FileAnnotationTabView.java#L13-L22
which then injects your data directly into the controller by
https://github.com/JabRef/jabref/blob/455965e931ca347e5caf91f52ded8e42cf11881c/src/main/java/org/jabref/gui/entryeditor/fileannotationtab/FileAnnotationTabController.java#L43-L46

src/main/resources/org/jabref/gui/copyfiles/CopyResultsDialog.fxml
```diff
@@ -0,0 +1,30 @@
...
+            <TableView fx:id="tvResult">
+              <columns>
+                <TableColumn fx:id="colStatus" prefWidth="75.0" text="Status" />
+                <TableColumn fx:id="colMessage" prefWidth="75.0" text="Message" />
+                  <TableColumn fx:id="colFile" prefWidth="75.0" text="File" />
+              </columns>
+            </TableView>
+            <ButtonBar prefHeight="40.0" prefWidth="200.0">
+               <buttons>
+                  <Button defaultButton="true" mnemonicParsing="false" onAction="#close" text="OK" />
```
tobiasdiez: you should always annotate buttons in a button bar with a proper ["button data"](https://docs.oracle.com/javase/8/javafx/api/javafx/scene/control/ButtonBar.ButtonData.html), which lets JavaFX position these buttons appropriately according to the style guidelines of the current os.

src/main/java/org/jabref/gui/copyfiles/CopyFilesDialogViewModel.java
Siedlerchr: But that does only work if I have a single View and not a dialog, right?
I still find this complicated and I don't see how I can achieve this with my custom dialog. I think I will add a property to the stage manager

Siedlerchr: I really would appreciate, if you could hook into this  PR and add the things

tobiasdiez: I think, as AbstractDialog derives from AbstractView, you just need to add a new constructor to the former one that relays everything to the view.

Sorry, I have enough "Baustellen" at the moment.

src/main/java/org/jabref/gui/copyfiles/CopyFilesResultListDependency.java
```diff
@@ -0,0 +1,22 @@
+package org.jabref.gui.copyfiles;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class CopyFilesResultListDependency {
```
Siedlerchr: @tobiasdiez  I have created this wrapper class to inject it

src/main/java/org/jabref/gui/copyfiles/ExportLinkedFilesAction.java
```diff
@@ -76,7 +71,11 @@ private void startServiceAndshowProgessDialog(Task<List<CopyFilesResultItemViewM
     }
 
     private void showDialog(List<CopyFilesResultItemViewModel> data) {
-        Dialog<ButtonType> dlg = new Dialog<>();
+
+        CopyFilesDialogView dlg = new CopyFilesDialogView(databaseContext, new CopyFilesResultListDependency(data));
```
Siedlerchr: @tobiasdiez  Here I am passing the instance of the object, but it doesn't display any data.

src/main/java/org/jabref/gui/copyfiles/CopyFilesResultListDependency.java
Siedlerchr: If I dont' intilaize the ArrayList with Empty, it will give an NPE in my View Model which leads to the conclusion that this are somehow different objects. I remember having similair problems wit the Sharelatex Integration and ended up making it static. But maybe there is another solution?

tobiasdiez: Oh damn...thats ugly. It is not possible to directly inject the list of results?
You also hit this bug here: https://github.com/AdamBien/afterburner.fx/issues/71 this is why you need the no-arg constructor.

tobiasdiez: I'll have a look at it later.

Siedlerchr: The list can not be directly injected, because List is an interface which of course can not be instantiated...that's where our default injector throws an error 
I tried to add handling for list but then generics are a problem...

Siedlerchr: @tobiasdiez  I found the issue. The "key" of the Hashmap and the name of the with inject annotated variable differed! That's why the data was not passed correct

tobiasdiez: With https://github.com/JabRef/jabref/pull/3334 you should be able to use `@Inject List<whatever> toBeInjected` or other complex types (they do not need to have a no-args constructor anymore). Of course the names of the "key" and the "field" have to coincide to make this work.

lynyus:
The result dialog is displayed way too small on my display.

screen shot 2017-10-25 at 18 24 34

When I try to resize it, a magical fourth column appears.
screen shot 2017-10-25 at 18 29 19

src/main/java/org/jabref/gui/copyfiles/CopyFilesTask.java
```diff
@@ -0,0 +1,115 @@
...
+import org.jabref.model.entry.LinkedFile;
+import org.jabref.model.util.OptionalUtil;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+public class CopyFilesTask extends Task<List<CopyFilesResultItemViewModel>> {
+
+    private static final Log LOGGER = LogFactory.getLog(ExportLinkedFilesAction.class);
+    private static final String LOGFILE = "exportLog.log";
```
lynyus: I am unsure if this logfile is needed.

If it is, I would encode a timestamp into the filename to be able to distinguish several exports into the same folder.

src/main/java/org/jabref/gui/copyfiles/ExportLinkedFilesAction.java
```diff
@@ -0,0 +1,68 @@
...
+import org.jabref.preferences.JabRefPreferences;
+
+public class ExportLinkedFilesAction extends AbstractAction {
+
+    private final DialogService dialogService = new FXDialogService();
+    private BibDatabaseContext databaseContext;
+    private List<BibEntry> entries;
+
+    public ExportLinkedFilesAction() {
+        super(Localization.lang("Copy attached files to folder..."));
```
lynyus: The terminology should be consistent. I suggest "Copy linked files to folder.."

src/main/java/org/jabref/gui/copyfiles/CopyFilesDialogController.java
```diff
@@ -0,0 +1,58 @@
...
+        viewModel = new CopyFilesDialogViewModel(copyfilesresultlistDependency);
+        setupTable();
+    }
+
+    private void setupTable() {
+        colFile.setCellValueFactory(cellData -> cellData.getValue().getFile());
+        colMessage.setCellValueFactory(cellData -> cellData.getValue().getMessage());
+        colStatus.setCellValueFactory(cellData -> cellData.getValue().getIcon());
+
+        colFile.setCellFactory(new ValueTableCellFactory<CopyFilesResultItemViewModel, String>().withText(item -> item).withTooltip(item -> item));
```
lynyus: is there a way to prevent that fourth column appearing https://github.com/JabRef/jabref/pull/3147#issuecomment-339389635 ?

Siedlerchr:
@lynyus  I changed the column size and there is also a tooltip for the file column

Issue #3145, opened 2017-08-22
Validation messages are displayed at wrong position


tobiasdiez:
Latest master. The validation messages in the entry editor are shifted to the bottom (by what appears to be a fixed constant):
![image](https://user-images.githubusercontent.com/5037600/29570175-f90a1d8c-8755-11e7-9b57-2ff7e3e7bbfa.png)

lenhard:
Confirmed, but it's not only that.

For me, DOI and URL warnings in the general fields tab are shifted upwards by a threshold.

I just reset my master branch to the beta3 and it was already broken there. Now I wonder, did the feature ever work? Were there library updates that broke this? We surely didn't change the code.

tobiasdiez:
No it never worked as desired (I just tested it with my original branch). I filed [a bug report at controlsfx](https://bitbucket.org/controlsfx/controlsfx/issues/791/update-position-of-decoration-when-target).

lenhard:
Damn :-( How did we not see that? I was very convinced that I saw it working.

What does that mean for 4.0? I would suggest that we add a preference options for enabling/disabling validation that is initially set to disabled. When the feature works properly, we can add the UI part to that allows you to enable the preference. That way, we keep the code, but do not annoy users.

matthiasgeiger:
Strangely if I click into an "invalid" editor field the sign is then positioned at the right position...  and yes: I first noticed this in the branch of @lenhard's resizing - and I was pretty sure that I have never seen it displayed in that wrong way before... 😕

koppor:
@matthiasgeiger I would assume that with the CSS patch by @lenhard leads to the initial size not being 0 and thus the issue https://bitbucket.org/controlsfx/controlsfx/issues/791/update-position-of-decoration-when-target is risen:

> The problem is that size listener is only registered if the initial size was 0

Please sign it at bitbucket and cast a vote for the issue. Maybe that helps to get the CustomFX library fixed.

lenhard:
@koppor My CSS patch is not the cause. We are talking about the master branch here and my patch is still an unmerged PR.

lenhard:
The validation feature is now masked by a preference option. It will be re-enabled once the bug is fixed in controlsfx.

Issue #3100, opened 2017-08-14
Add preference option to disable field validation


lenhard:
In #3090 @tobiasdiez introduced field validation in the entry editor. This is clearly beneficial, but I can image that some users do not want it.

We should add a preference option that disables field validation for the entry editor.

lenhard:
@tobiasdiez Could you please implement this one little addition to your validation feature. It is no big deal and we really need it for 4.0 (user's are complaining about the validation).

tobiasdiez:
I'll have a look next week but can't promise anything.

tobiasdiez:
Sorry, I will not manage to do this in the foreseeable future.

lenhard:
The option (default false) is now implemented and merged. The UI to change it will be added when the feature is fixed in controlsfx. 

Hence this can be removed from 4.0

PR #3154, opened 2017-08-25
Add preference to disable validation in the entry editor by default

lenhard:
Adds a preference option for enabling/disabling validation in the entry editor: #3100 (but not the UI)

The reason is because the validation is simply broken at the moment #3145

The preference option is set to false by default and there is no UI to change it. So this turns off validation in the entry editor. We can add the UI to turn it on later, when the display bug in controlsfx has been fixed.

When this is merged, the related issues can be removed from the 4.0 milestone.

### DIFF
--- a/src/main/java/org/jabref/gui/fieldeditors/EditorValidator.java
+++ b/src/main/java/org/jabref/gui/fieldeditors/EditorValidator.java
@@ -1,0 +1,16 @@
+package org.jabref.gui.fieldeditors;
+
+import de.saxsys.mvvmfx.utils.validation.ValidationStatus;
+import de.saxsys.mvvmfx.utils.validation.visualization.ControlsFxVisualizer;
+import org.jabref.Globals;
+import org.jabref.preferences.JabRefPreferences;
+
+public class EditorValidator {
+
+    public static void configureValidation(ValidationStatus status, EditorTextArea area) {
+        if (Globals.prefs.getBoolean(JabRefPreferences.VALIDATE_IN_ENTRY_EDITOR)) {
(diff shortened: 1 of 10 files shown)


src/main/java/org/jabref/gui/fieldeditors/EditorValidator.java
```diff
@@ -0,0 +1,17 @@
...
+import org.jabref.Globals;
+import org.jabref.preferences.JabRefPreferences;
+
+import de.saxsys.mvvmfx.utils.validation.ValidationStatus;
+import de.saxsys.mvvmfx.utils.validation.visualization.ControlsFxVisualizer;
+
+public class EditorValidator {
+
+    public static void configureValidation(ValidationStatus status, EditorTextArea area) {
+        if (Globals.prefs.getBoolean(JabRefPreferences.VALIDATE_IN_ENTRY_EDITOR)) {
```
tobiasdiez: Can you please not use `Globals.prefs`. I really tried hard to make the JavaFX stuff free of `Globals`. You can simply pass a preference instance to the editor (similar how it is done in the OwnerEditor) and then to this method here (mabye as `new EntryValidator(preferences).configureValidation(...))`).

lenhard: Done! Good point also, we should thrive to keep this independent from `Globals`. I even spotted one more usage in `FieldEditors` that you seem to have overlooked ;-)

lenhard:
Since the review comments have been addressed and for the sake of getting things done, I am merging this now.

PR #3155, opened 2017-08-25
Update .gitignore

### LINKED ISSUE #3106: New gitignore config ignores lib folder
In #3091 we updated our gitignore to a suggestion from gitignore.io

Unfortunately, this includes *.jar and this makes git ignore our /lib folder. Tracking this down just cost me a few hours....

@koppor What are your thoughts on this? It means that we cannot blindly take the suggestions from gitignore.io Is there a way to exclude jar files from their generated configuration?

koppor:
This implements https://github.com/JabRef/jabref/issues/3106#issuecomment-322461597 and thus fixes #3106.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,6 @@
 # no generated files in version control
 src/main/gen/
 
-# do not distribute Oracle's JDBC driver
-lib/ojdbc.jar
-
 # private data
 /buildres/jabref-cert-2016.p12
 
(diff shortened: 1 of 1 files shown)


lenhard:
You could add a Changelog entry that references the issue and then this is good to go.

koppor:
I thought, we are only recording changes for the end user and no internal
changes? 😇

Am 25.08.2017 14:43 schrieb "Jörg Lenhard" <notifications@github.com>:

lenhard:
Ok, fine :-D I am merging.

Issue #2946, opened 2017-06-28
External changes of open entry are not synced-back to the entry editor


tobiasdiez:
Changes like generating a new bibtexkey or renaming the file are not displayed in the entry editor.

lenhard:
@tobiasdiez Is it possible that the `ObjectBinding` between the editor and the `BibEntry` breaks when you change something in the map of its fields, for instance by calling `fields.put()`? Can you please have a look at that?

lenhard:
@tobiasdiez Sorry, but this one is not fixed yet. It works now if you change something in the source tab. Then the changes are synced back to the field editor.

But it does not work in case of external changes. If you apply save actions, the changes only become visible if you reopen the file. If you change the file externally in a text editor, the changes become visible in the main table but not in the entry editor. They only become visible when the entry editor is recreated.

tobiasdiez:
I tried it with the "convert to biblatex" cleanup and it worked (the year was moved to date and this change was correctly shown in the entry editor). Same with the "generate bibtex key" action.
What do I have to do to reproduce the problem you described?

lenhard:
@tobiasdiez I'll have to correct myself regarding safe actions. They do work. But the external changes don't. Do the following:
 1. open a bib file in jabref and a text editor (e.g. notepad++)
 2. change the content of the author field in notepad to something else, say `test`
 3. Go back to JabRef, the files changed dialog pops up. Accept the changes
 4. Now, the new author is displayed in the main table, but not in the entry editor.

tobiasdiez:
Ok, probably the merge dialog deletes the old bibentry and creates a new one instead of updating the one already present in the database. Thus the binding to the entry editor gets destroyed.
Should I create a new issue for this or do we silently ignore this problem until someone complains? 😄

lenhard:
That's the clue. The review panel says that there's an entry added and an entry removed, so that's why the relation breaks.

So that's not really critical, since I don't see the use case. Why would someone change the exact same entry in JabRef and a text editor in parallel, except if it's me who's testing. If a different entry is changed, the entry editor is recreated anyway.

Let's keep this open and I'll rename it and remove it from the 4.0 milestone.

lenhard:
The remaining part is fixed by #3156.

PR #3156, opened 2017-08-25
Close entry editor when the shown entry is removed externally

lenhard:
Fixes the remaining part of #2946

When you externally delete an entry that is opened in the entry editor and accept the change in the dialog in JabRef, then the entry editor closes. That way, it prevents you from editing an entry that does no longer exist.

As a side note: The `BasePanel` is ripe for a refactoring. There are a good dozen classes that I could extract. In the light of only doing bug fixes before the 4.0 release, I'll avoid that now. But when we've released, this class gets it...

### DIFF
--- a/src/main/java/org/jabref/gui/BasePanel.java
+++ b/src/main/java/org/jabref/gui/BasePanel.java
@@ -2105,6 +2105,9 @@ public class BasePanel extends JPanel implements ClipboardOwner, FileUpdateListe
             // if the entry that is displayed in the current entry editor is removed, close the entry editor
             if (isShowingEditor() && currentEditor.getEntry().equals(entryRemovedEvent.getBibEntry())) {
                 currentEditor.close();
+            }
+
+            if (selectionListener.getPreview().getEntry().equals(entryRemovedEvent.getBibEntry())) {
                 selectionListener.setPreviewActive(false);
             }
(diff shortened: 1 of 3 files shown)


src/main/java/org/jabref/gui/BasePanel.java
```diff
@@ -2097,6 +2098,18 @@ public void listen(EntryAddedEvent addedEntryEvent) {
         }
     }
 
+    private class EntryRemovedListener {
```
Siedlerchr: Why do you introduce this private class? Why not directly add the listener to the BasePanel?

lenhard: Because there are several other listeners in `BasePanel` that are all implemented as separate classes. That way it is easier to one day extract the listeners into different files and reduce the size of `BasePanel`

Btw.: `BasePanel` currently has no closing method, so these listeners are never unregistered and are essentially a memory leak. Not too big a problem, though: A new `BasePanel` is created very rarely and the listeners are very small (another reason for the separate class at the moment).

lenhard:
For the sake of getting things done. I'll merge now directly, even though there's only one review. I guess because of our current shortage in personal we'll have to deviate from the two-reviewer rule now and then.

PR #3179, opened 2017-08-31
syncLang.py: link in markdown output

mlep:
To allow for a more direct edit of the localization files.

Please, triple check, I am in no way a python programmer.

### DIFF
--- a/scripts/syncLang.py
+++ b/scripts/syncLang.py
@@ -11,6 +11,7 @@ import logger
 
 RES_DIR = "src/main/resources/l10n"
 STATUS_FILE = "status.md"
+URL_BASE = "https://github.com/JabRef/jabref/tree/master/"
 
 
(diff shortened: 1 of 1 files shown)


koppor:
Tried id out locally - and it works. Thank you for the PR!

lynyus:
as a follow up: I have tested it as well, works.

What about automatically generating this data with each push to master and publish a html version on jabref.org?

mlep:
It would be nice!

mlep:
Although: it would be nice if the order of the languages would be always the same.

lynyus:
That should be feasible. Is that all the requirements you can think of? I would volunteer to program that. Would you be so kind to file an issue to separate the discussion and point out all potential requirements?

mlep:
Done (Issue #3181)

PR #3180, opened 2017-09-01
[WIP] Use v8 for citeproc

koppor:
This is a try to get the V8 engine into JabRef. This is a follow up to #2250, where a discussion of this is made.

V8 requires different JARs to be built as the V8 JARs are platform dependent. This is not easily possible and the [stackoverflow question](https://stackoverflow.com/q/43459663/873282) has no answer directly applicable.

Open issues:

- [ ] Produce multiple fat jars
- [ ] Integration in install4j
- [ ] Update of the homepage

Context: When a user uses a preview based on citation styles, the preview gets slow:

grafik

To illustrate the performance gain, I made two videos using [screentogif](https://screentogif.codeplex.com/):

## without V8

without-v8

## with V8

with-v8


koppor:
Alternative could be https://github.com/graalvm/graaljs

Issue #3181, opened 2017-09-01
syncLang.py: improvements


mlep:
Following comments in PR #3179, here are ideas for improving syncLang.py:
- automatically generate the translation status with each push to master and publish a html version on jabref.org
- keep a constant order for the table of languages: English first, then alphabetical (or all alphabetical). (I guess it currently depends on the OS)
- put the table of menus first (so that is it not discouraging newcomers with a huge number of strings waiting for translation).
- align to the right the columns with numbers (easier to read, but just a tiny detail).

lynyus:
- [ ] Update to python3
- [ ] Apply object-oriented design principles

LinusDietz:
Successort Issue: https://github.com/JabRef/help.jabref.org/issues/171

PR #3184, opened 2017-09-01
Improve SyncLang.py

lynyus:
See Issue #3181.

- [X]  Update to Python 3
- [ ]  automatically generate the translation status with each push to master and publish a html version on jabref.org
- [x]  keep a constant order for the table of languages: English first, then alphabetical
- [x]  put the table of menus first (so that is it not discouraging newcomers with a huge number of strings waiting for translation).
- [x]  align to the right the columns with numbers (easier to read, but just a tiny detail).
- [x]  Apply object-oriented design principles

### DIFF
--- a/scripts/syncLang.py
+++ b/scripts/syncLang.py
@@ -36,7 +36,7 @@ def call_command(command: str) -> str:
     return subprocess.check_output(command.split(" ")).decode("utf-8").rstrip()
 
 
-def get_filename(filepath):
+def get_filename(filepath) -> str:
     """
     removes the res_dir path
 
@@ -46,7 +46,7 @@ def get_filename(filepath):
     return filepath.replace("{}\\".format(RES_DIR), "")
 
 
-def read_file(filename, encoding="UTF-8"):
+def read_file(filename, encoding="UTF-8") -> list:
     """
     :param filename: string
     :param encoding: string: the encoding of the file to read (standard: `UTF-8`)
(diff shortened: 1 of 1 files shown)


mlep:
Additional task: in a new column of the tables, flag files with anomalies (such as space characters found in localized strings).

Is there any other special characters for localized strings?

lynyus:
@mlep, I think we should rather write a test case for that in Java, so that it is not possible to commit such malformed strings

mlep:
Fine. Great!

lynyus:
I have refactored that script, however I have not yet tried to incorporate it in the build pipeline. I would merge this for now and think about the build automation in another PR.

Siedlerchr:
Can you check the case if the syncing does correctly add/remove keys when I do a rename of a key?
I recently had the problem with the old version of the script in a branch.
E..g. I previously created a new key e.g.  Exporting_file_X_of_Bibtex_Entry 
Synced that to all lang files
Afterwards I just renamed the BIbEntry to "entry", ran it again and the script did not replace the changes in the other files. I guess it was because it was at the same line. After I moved the renamed key to another position, it worked

scripts/syncLang.py
```diff
@@ -1,277 +1,211 @@
...
+    def key_from_line(line) -> str:
+        """
+        Tries to extract the key from the line
+
+        :param line: unicode string
+        :return: unicode string: the key or None
+        """
+        if line.find("#") != 0 or line.find("!") != 0:
+            index_key_end = line.find("=")
+            while (index_key_end > 0) and (line[index_key_end - 1] == "\\"):
```
Siedlerchr: Just curious: Is this platform independent line endings?

lenhard:
Bump.

Can we get this PR ready to merge?

scripts/syncLang.py
lynyus: Yep, I have enforced the line ending handling.

lynyus:
@Siedlerchr Yep, updating and deleting entries works for me.

@lenhard I think we can merge.

lenhard:
Ok, perfect! Then I'll do this right away. 

@lynyus I am not 100% sure if #3181 is ready to be closed based on this PR or what it means for #3397. Could you please have a look at these issues and close or comment?

Issue #3175, opened 2017-08-31
Performance issues with open entry editor


AEgit:
JabRef 4.0-dev--snapshot--2017-08-31--master--54e0994ba
Windows 10 10.0 amd64 
Java 1.8.0_141

This isn't really a bug, it's more an observation I made that I wanted to share, as it might affect some users.

Steps to reproduce:

1. Open an item in the entry editor
2. Keep the entry editor open and select another item
3. The new item is opened in the entry editor
4. Do this multiple times
5. The Windows Task Manager shows a massive increase in the used CPU power (up to 95% on a Corei7 4+4HT system)
6. Indeed, the responsiveness of the entry editor does not feel as quick as with 3.8.2 - it is still ok on my machine, I guess, but people with a slower CPU might experience bigger performance issues

Note, that this might be also related to the size of the databse (>6k entries).
Note furthermore, that this issue does not appear, if the entry editor is closed.

lenhard:
The entry editor also seems to be the main source of our memory problems, see #3113.

lenhard:
@AEgit Could you do me a favour and try out the version available here http://builds.jabref.org/entry-editor-performance/

This is a tiny improvement and I would be surprised if it has a big effect performance-wise. Its purpose is more to pave the way for an actual performance improvement that is still to come. But do you maybe notice something, e.g. a little improvement in the responsiveness of the entry editor?

Moreover, it creates a visual effect: When selecting a new entry, then the tabs in the entry editor sort of "fly in". I am interested in your opinion about this effect. Do you maybe like it, is it no problem, or do you find it totally annoying? If the latter is the case, then this shouldn't go into 4.0 anymore.

AEgit:
JabRef 4.0-dev--snapshot--2017-09-02--entry-editor-performance--548537f12
Windows 10 10.0 amd64 
Java 1.8.0_141

@lenhard This version does seem to be slightly faster (although, as you say, the improvement is probably rather small). It still does use, however, a lot of CPU performance.

Regarding the visual effect: It is problem better without it - if you quickly change between different items it can happen that the tabs don't load properly or that it takes a while for them to appear (see appended image).
[Dok1.pdf](https://github.com/JabRef/jabref/files/1272306/Dok1.pdf)

koppor:
Maybe, we have to use YourKit (see https://github.com/JabRef/jabref/issues/3113#issuecomment-322826845) to profile where the CPU is used?

Siedlerchr:
[...]

Issue #3113, opened 2017-08-16
JabRef should release unused heap more frequently


lenhard:
It seems that we have a memory leak with the entry editor on current master, latest commit 0326d7778230e640105381c4c8dc7f548f580c55

JabRef 4.0-dev
Windows 7 6.1 amd64 
Java 1.8.0_121

To reproduce:
 - Open entry editor and select entry in the main table
 - Press down key to cycle through the database
 - Watch RAM consumption grow and not go down again

@JabRef/developers Please see if you can reproduce this on your system. This seems quite serious to me and a blocker for 4.0.

Any hints where this might come from are very much appreciated. @koppor Could you maybe try tracking this down with git bisect?

lenhard:
@tobiasdiez I suspect that bindings are the cause of this all. 

I know your time is limited, but can you please have a look at this? It's really a game changer...

lynyus:
Confirmed. Had about 2.5GB allocated RAM after 5 minutes of clicking around in a Bibtex File with about 50 entries. Closing the EntryEditor did not help:/

JabRef 4.0-beta3
Mac OS X 10.12.6 x86_64 
Java 1.8.0_144

tobiasdiez:
The main reason for the memory leak is that the event listener in the source tab is never unregistered: https://github.com/JabRef/jabref/blob/master/src/main/java/org/jabref/gui/entryeditor/SourceTab.java#L55

After uncommenting this line, the memory still increases a bit for each new entry but not as much as before. I tried to further investigate the issue but couldn't really pinpoint the cause.

First of all, mostly int arrays are left over after opening a bunch of entries (the screenshot shows the difference in the memory consummation/allocated objects between a clean start and after the entry editor was opened for a few entries):

![image](https://user-images.githubusercontent.com/5037600/29373346-37e721f0-82ae-11e7-96aa-70ce2685ae09.png)

However, most of these arrays are actually no longer referenced and can be removed by garbage collection (no idea, why these are kept):
![image](https://user-images.githubusercontent.com/5037600/29373317-1f018838-82ae-11e7-9b00-ae7b8811c394.png)

It appears that these numbers (and the other objects like HashMaps) have their origin in JavaFX. The bigger int arrays are a kind of raster image of older entry editors. Moreover, a few other objects are referenced by a cache in the JavaFX style manager (but again, these are weak hash maps). 

[...]

PR #3187, opened 2017-09-02
Avoid recreation of the EntryEditor

lenhard:
This is one baby step towards fixing the performance and memory issues we are having with the entry editor, e.g. #3175 #3113

Previously, we've created a new EntryEditor instance every time a new entry was selected. This PR avoids that. Instead of recreating a new entry editor, the existing one is reused and just all its contents, i.e. the tabs, are recreated. This doesn't help significantly and we still have the peaks when cycling through the main table with key pressed, but it's a step in the right direction. Here's an overview
jabref
The peaks are from cycling through the main table with down key pressed, but as we can see garbage collection still reduces the used heap massively.

The next step (in a new PR) would be to avoid a recreation of the tabs and to implement a proper lazy initialization of them, so that they only consume memory and processing power when they are actually visible.

I've tested it manually and hope that this doesn't add hidden bugs. After all, the tabs are still created anew so there shouldn't be update problems. Nevertheless, this should receive some manual testing from someone else before it is merged. Lastly, since the tabs are created anew there's now a visible effect when selecting a new entry: The new tabs sort of "fly in". I have no problems with that, but if others find that annoying then this PR should not yet be merged before for 4.0 or alternatively only when the tabs are not recreated.


koppor:
JabRef used to have an entry editor cache. We disabled that cache, because of issues with the "Other Fields" tab and customized entries.  Thus, when reviewing, we have to check https://github.com/JabRef/jabref/issues/2064 and https://github.com/JabRef/jabref/pull/2075.

src/main/java/org/jabref/gui/entryeditor/EntryEditor.java
```diff
@@ -153,71 +153,56 @@
...
-        setupToolBar();
+        layout = new BorderLayout();
+        setLayout(layout);
 
         container = OS.LINUX ? new CustomJFXPanel() : new JFXPanel();
+        // Create type-label
+        typeLabel = new TypeLabel("");
+        setEntry(entry);
+        setupToolBar();
+        container.setScene(new Scene(tabbed));
```
tobiasdiez: I'm not sure but I think `setScene` should be called in the JavaFX thread (`DefaultTaskExecutor.runInJavaFXThread`).

src/main/java/org/jabref/gui/entryeditor/EntryEditor.java
```diff
@@ -153,71 +153,56 @@
...
     private final BooleanProperty movingToDifferentEntry = new SimpleBooleanProperty();
-    private final EntryType entryType;
+    private EntryType entryType;
     private SourceTab sourceTab;
+    private final BorderLayout layout;
+    private TypeLabel typeLabel;
 
-    public EntryEditor(JabRefFrame frame, BasePanel panel, BibEntry entry, String lastTabName) {
-        this.frame = frame;
+    public EntryEditor(BasePanel panel, BibEntry entry) {
```
tobiasdiez: I would remove the `entry` from the constructor to make it clearer that the entry editor can be reused and is not specific to the entry.

src/main/java/org/jabref/gui/entryeditor/EntryEditor.java
```diff
@@ -153,71 +153,56 @@
...
-                            break;
-                        default:
-                            //do nothing
-                    }
-                }
-            }
-        });
         DefaultTaskExecutor.runInJavaFXThread(() -> {
-            addTabs(lastTabName);
+            addTabs(this.getVisibleTabName());
```
tobiasdiez: is `getVisibleTabName` used somewhere outside of this class? if not then it can be converted to just a field and don't need to be exposed.

koppor:
If I have "Other fields" and switch to an entry without "Other fields", the FIRST tab "Required field" is focused, not "Deprecated Fields", which is the tab BEFORE the tab "Other fields".

halirutan:
I used [YourKit](https://www.yourkit.com/) tonight to profile the current master and looked over this PR. I tested it by using the infinite scroll with the entry editor open so that JabRef was pushed to its limits. Be aware that the things below might be wrong but I hope they give some helpful insight:

## Localization memory

First, I was stunned that the objects that are purely connected to getting the translation strings are so large. In the current master branch, they accounted for 6% of the memory of the complete thread. I marked the line:

mem1

I looked into the implementation and saw that the list of keys is always recreated. I was wondering why that is. When I change the language in the preferences, I get notified, that I have to restart JabRef. So why don't we hold one `HashSet` that is initialized once? I [hacked this down](https://gist.github.com/halirutan/3d772207945527de14f5881d1f676dce) just to test it (the implementation is not correct as it stands!). After that, the memory usage basically vanished. Compare the size and the recorded objects.

mem2

## `EntryEditorTab` and *lazy initialization*

This base-class is laid out to support late initialization. Unfortunately, this is not done. Most notably `org.jabref.gui.entryeditor.FieldsEditorTab#FieldsEditorTab` calls `setupPanel` in its constructor. So everytime `org.jabref.gui.entryeditor.EntryEditor#addTabs` is called, which is every time we now set a new entry (although the `EntryEditor` is reused), a full-fledged panel is constructed. I don't see why this is necessary when we only need to construct this when the user wants to see this tab. [Another hack](https://gist.github.com/halirutan/135ec7f263c61a23305a1ecf59ee5a8f) that moved this to the `initialize` method.

## Going one step further

This PR is a first good step. My question is: Shouldn't this also be done one step deeper? What I propose is the following:

- All editor tabs are instantiated only once for the `EntryEditor`. Tabs (I mean the header) must be instantaneously available to the entry editor so that they are not rebuilt every time as you can see in the last link below.
- When the editor updates, it decides which tabs should be shown and which not
- The initialization of each tab is done when the tab gets focus. There is no need to fill it with data and set it up before that

## Comparison of speed

I tested this speed of the `EntryEditor` with the largest example I could find in the examples and used again infinite scrolling. The difference is astounding. Look at the following for the current master branch implementation and see how long it takes until the editor is loaded at the end of the scrolling (btw, I have an Intel(R) Core(TM) i7-5960X CPU @ 3.00GHz with 64GB of RAM here!)

[Slow Version](https://youtu.be/i4Jb2YuTvc4)

Now this PR with the fix for the initialization and the Location.

[Fast Version](https://youtu.be/bPAI9atxDQA)

lenhard:
@halirutan First of all: Wow, just wow. It is a very rare sight to see a new contributor put in as much effort and diligence as you do. You have my kudos. Your work is excellent.

# Regarding the localization
Your suggestion is a valid improvement which I fully support. It's not directly linked to the topic of this PR, though. So, could you turn this into a separate pull request? I hope that it does not take too much effort to turn this into a fully correct implementation, but from glancing at the code I don't think it would. 

# Regarding the entry editor
The lazy initialization has my upvote as well. I wonder why this hasn't been changed to support lazy initialization before, and my guess is that it is just an oversight by several people including myself.

What you list under "going one step further" is also what I had in mind. The "problem" that this faces is that the content of a tab varies depending on what kind of entry is displayed. For example, an `article` has different required fields than an `inproceedings` entry. I think this difference has been the original reason for rebuilding the complete entry editor every time selection changes. It could be a little tricky to get that right every time, but it is something that we should definitely do to improve performance.

# How to proceed from here
The performance improvements you show are just awesome. We should get all this into master and I suggest the following: 
 - I fix the issues mentioned above and get this PR ready for merge. I do not implement that the lazy initialization or localization fix, since I do not want to take credit for something that you have found and repaired.
 - In the meantime, you can create two new PRs, one for the localization and one for the lazy initialization. As far as I can see, this should not result in merge conflicts to this one.

When we have integrated all this, we can have a go at a proper one-time creation of the tabs. If you are interested in trying this, be my guest.

Once again, thank you a lot!

src/main/java/org/jabref/gui/entryeditor/EntryEditor.java
lenhard: Done

src/main/java/org/jabref/gui/entryeditor/EntryEditor.java
lenhard: Done

src/main/java/org/jabref/gui/entryeditor/EntryEditor.java
lenhard: It is used in the `BasePanel` and the main table, so it needs to be visible.

lenhard:
I have double checked  #2064 and  #2075 manually and the optional fields are recreated as desired.

@koppor I don't really see why the focus should jump to adjacent tab when switching to an entry that does not have optional fields (also why to the one before and not the one after). So unless I have a good reason to jump to a particular tab, I am against implementing special handling for that.

From my point of view, this PR is ready to merge and than we can go for the further improvements discussed above.

lynyus:
Okay, then I'll be brave and merge this now.

Issue #3172, opened 2017-08-31
Rename File Fails Silently


lynyus:
Renaming a linked file fails if there already exists a file with the name.

In the General tab of the Entry Editor: Right click on the linked file -> Rename File. A popup that states the new file name appears. Confirming it results in a failure without any message. The other file is not overwritten and the operation apparently has no effect.

I would expect that a popup message appears and asks to user to confirm the overwrite.

JabRef 4.0-dev
Mac OS X 10.12.6 x86_64 
Java 1.8.0_144

Siedlerchr:
Have a look at 
```
public class LinkedFileViewModel extends AbstractViewModel {

 public void rename() { 
```

Andrew-Lev:
It sounds like adding a check for existing files in that directory with the targetFileName and then showing the error dialog would take care of this.  I can fix this fairly quickly, want me to work on it?

Andrew-Lev:
Didnt take very long at all actually.  I'll make a pull request once someone confirms.

![samefilenameerror](https://user-images.githubusercontent.com/17808285/29958237-8db997f8-8ea6-11e7-85f5-1ba9a0d0a718.png)

lynyus:
Yeah, cool stuff! Just open the PR, and we'll have a look at it!

lynyus:
fixed in #3211. Thanks @Andrew-Lev!

PR #3188, opened 2017-09-03
Added check for existing files when renaming linked file

Andrew-Lev:
#3172 Added a check in LinkedFileViewModel.rename() to see if a file which is
being renamed already exists in the target location.  If such a file is
found then an error popup displays informing the user of the conflict.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -128,7 +128,7 @@ dependencies {
     compile group: 'com.microsoft.azure', name: 'applicationinsights-logging-log4j2', version: '1.0.+'
 
     testCompile 'junit:junit:4.12'
-    testCompile 'org.mockito:mockito-core:2.9.0'
+    testCompile 'org.mockito:mockito-core:2.10.0'
     testCompile 'com.github.tomakehurst:wiremock:2.8.0'
     testCompile 'org.assertj:assertj-swing-junit:3.8.0'
     testCompile 'org.reflections:reflections:0.9.11'


src/main/java/org/jabref/gui/fieldeditors/LinkedFileViewModel.java
```diff
@@ -175,7 +175,14 @@ public void rename() {
                     Localization.lang("Cancel"));
 
             if (confirm) {
-                pdfCleanup.cleanup(entry);
+                Path targetFilePath = file.get().getParent().resolve(targetFileName);
```
koppor: For me, this is too much logic in the UI. We really aim to have the logic separated from the GUI. See https://github.com/JabRef/jabref/wiki/High-Level-Documentation

It costed us much headaches to refactor JabRef and also costs us headaches now. However, this really increases testability and enables extraction of libraries.

Therefore, may I ask whether it is possible to move this logic into RenamePdfCleanup somehow? Meaning: Is it possible to get a getter for `targetFileAlreadyExsists()`?

In RenamePdfCleanup, we also have `pathsDifferOnlyByCase`, which is not respected by this code: https://github.com/JabRef/jabref/blob/master/src/main/java/org/jabref/logic/cleanup/RenamePdfCleanup.java#L101

src/main/resources/l10n/JabRef_da.properties
```diff
@@ -503,6 +503,9 @@ File_directory_is_not_set_or_does_not_exist\!=Filbiblioteket_er_ikke_sat_eller_e
 
 File_exists=Filen_eksisterer
 
+File_already_exists=
```
koppor: Please do not add this new key. Please reuse `File_exists` two lines above.

In case you think, including `already` is a better fit, please rename the translation key and adapt `JabRef_en.properties` accordingly (the keys have to match there).

src/main/resources/l10n/JabRef_da.properties
```diff
@@ -503,6 +503,9 @@ File_directory_is_not_set_or_does_not_exist\!=Filbiblioteket_er_ikke_sat_eller_e
...
+File_'%0'_already_exists.=
```
koppor: We have `'%0'_exists._Overwrite_file?` in JabRef. Maybe, the code can be adapted to allow file overwrite instead of just cancelling the request?

Andrew-Lev:
After looking at the code in RenamePdfCleanup.cleanup I'm a little confused.
There's a note that seems to contradict the log output and the logic.  Can you clear that up for me?

```
if (Files.exists(newPath) && !pathsDifferOnlyByCase) {
                // we do not overwrite files
                // Since File.exists is sometimes not case-sensitive, the check pathsDifferOnlyByCase ensures that we
                // nonetheless rename files to a new name which just differs by case.
                // TODO: we could check here if the newPath file is linked with the current entry. And if not, we could add a link
                LOGGER.debug("There already exists a file with that name " + newPath.getFileName()
                        + " so I won't rename it");
                newFileList.add(flEntry);
```

koppor:
Windows setting:

- Rename a.pdf to A.pdf.
- Files.exist("A.PDF") returns true. 
- We really want to do this name change, because we want to change `a` to `A`

This check allows exactly that. If `!pathsDifferOnlyByCase` was missing, JabRef would complain that `A.pdf` already existed. On Windows, however, this is not true, as `a.pdf` exists, but not `A.pdf`. But due to the FAT32 case-insentive thingy, both `a.pdf` and `A.pdf` exist.

Is that the thing, which confused or is it something different?

Andrew-Lev:
Ok that makes sense.

I went over your suggestions, and things turned out a bit more complicated than expected.  The only way I could see to make the check case-insensitive was to iterate through a list of files in the target directory.  I had to do this because we only have the original file path and the target file path when this runs.  I also extracted pathsDifferByCaseOnly to a method so I could use it here.

Let me know what you think. 

```
    public boolean fileAlreadyExists(LinkedFile flEntry, BibEntry entry) {
        String targetFileName = getTargetFileName(flEntry, entry);
        Path targetFilePath = flEntry.findIn(databaseContext,
                fileDirectoryPreferences).get().getParent().resolve(targetFileName);
        Path oldFilePath = flEntry.findIn(databaseContext, fileDirectoryPreferences).get();
        boolean matchedByCase = false;
        try (Stream<Path> filesList = Files.list(oldFilePath.getParent())) {
            //Need to iterate through filenames in directory and compare to target filename
            for (Object path : filesList.toArray()) {
                if (pathsDifferOnlyByCase(targetFilePath, path.toString())) {
                    matchedByCase = true;
                    break;
                }
            }
        } catch (IOException e) {
            LOGGER.error("Could not get the list of files in target directory", e);
        }

        //check if file already exists at target and make sure the difference isnt case only
        if (!Files.exists(targetFilePath) && !matchedByCase) {
            return false;
        } else {
            return true;
        }
    }
```

I suppose the code that iterates could be extracted to its own method and only run when the first condition from the below isnt met, if theres any concern that this might slow things down.
`if (!Files.exists(targetFilePath) && !matchedByCase) {`

Siedlerchr:
Your approach looks good, however I don't see the reason why you need to iterate over all files:
You just have to extract/copy the following steps in a new method:

We want to rename X:\ABC\A.pdf into X:\ABC\B.pdf (we ignore the creation of the specific filenames for the moments) 

`       Optional<Path> expandedOldFile = flEntry.findIn(databaseContext, fileDirectoryPreferences);`
Here we get the old existing file which the user wants to rename: X:\ABC\A.pdf 

Then in the following code we build the new FileName (the target of the renaming) 
```         
            String targetFileName = getTargetFileName(flEntry, entry);
            Path newPath = expandedOldFile.get().getParent().resolve(targetFileName); 
```
We go to the directory of the old file and resolve it against the TargetFileName, so in newPath is now stored the full path to the target file - which hopefully does not yet exist. 
So new Path is now the directory (parent) of the file A.pdf -> ABC and we do now resolve ABC-> B.pdf
=> X:\ABC\B.pdf 

Here we compare the full path of the old file name with the new file and check if A.pdf and B.pdf are the same paths (as koppor explained above, we treat A.pdf and a.pdf as the same file)
            String expandedOldFilePath = expandedOldFile.get().toString();
            boolean pathsDifferOnlyByCase = newPath.toString().equalsIgnoreCase(expandedOldFilePath)
                    && !newPath.toString().equals(expandedOldFilePath);

   if (Files.exists(newPath) && !pathsDifferOnlyByCase) {
```
That is the actual check: If there already exists an actual File X:\ABC\B.pdf

**Comment by Andrew-Lev** [comment]
The reason I chose to iterate is because there might be another file from a different entry with the same name in that folder, but with different case.

Here is a scenario that I ran into while testing.
Entry A - File aaa 
Entry B - File bbb
Test: Rename aaa to bbB = failed and not detected unless we iterate

The method pathsDifferOnlyByCase checks the old name (aaa) against the new name (bbB), but does not check the new name against other files in that directory.
Files.exists method will check for bbB and return that it doesnt exist, but then still fail because bbb exists in the directory.  The directory could contain any variety of files such as BBB bbb BbB BBb etc.

**Comment by Siedlerchr** [comment]
Ah
Good catch! Now I see your problem.
However, if you directly apply the filter operation on the stream you are more efficient and a stream is lazily populated.
Check out that link for an example how to use files find wth a condition 


http://winterbe.com/posts/2015/03/25/java8-examples-string-number-math-files/

**Comment by Andrew-Lev** [comment]
That seems very useful, thanks for letting me know.  How's this?

```
        try (Stream<Path> stream = Files.list(oldFilePath.getParent())) {
            matchedByCase = stream
                    .map(String::valueOf)
                    .anyMatch(path -> path.equalsIgnoreCase(targetFilePath.toString()));
        } catch (IOException e) {
            LOGGER.error("Could not get the list of files in target directory", e);
        }
```

Siedlerchr:
Hi, just add the method and commit then I can check out your branch. For overwriting or renaming look at FileUtil, the method take a Boolean for overwriting

Andrew-Lev:
Local branch got a bit messy and I ended up resetting it without knowing that this pull request would get closed.  I opened up a new one with the latest code changes.

koppor:
No need to open a new one. You can push to the old branch. This PR gets
updated then. One can reopen a PR easily by clicking on Reopen.

Am 12.09.2017 05:21 schrieb "Andrew" <notifications@github.com>:

Andrew-Lev:
Sorry I didnt know about that.  Will do so in the future.

PR #3191, opened 2017-09-03
Testcoverage for AutoEvent and EntryLinkList

sammann-fnordbedarf-de:
I want to  increase testcoverage. This is my first pull request with minor changed to get used to the contribution process.

none, too little changed.
Added tests for AutoEvent and EntryLinkList
2859 done, 139 failed, 6 ignored. Failed tests had nothing to do with my changed.

### DIFF
--- a/src/test/java/org/jabref/model/database/event/AutosaveEventTest.java
+++ b/src/test/java/org/jabref/model/database/event/AutosaveEventTest.java
@@ -1,14 +1,7 @@
 package org.jabref.model.database.event;
 
-import org.jabref.model.database.EntrySorter;
-import org.jabref.model.entry.BibEntry;
 import org.junit.Test;
 
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.Comparator;
-
-import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 
 public class AutosaveEventTest {
--- a/src/test/java/org/jabref/model/database/event/AutosaveEventTest.java
+++ b/src/test/java/org/jabref/model/database/event/AutosaveEventTest.java
@@ -1,0 +1,21 @@
+package org.jabref.model.database.event;
+
+import org.jabref.model.database.EntrySorter;
+import org.jabref.model.entry.BibEntry;
+import org.junit.Test;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Comparator;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+
+public class AutosaveEventTest {
+
+    @Test
+    public void givenNothingWhenCreatingThenNotNull() {
+        AutosaveEvent e = new AutosaveEvent();
+        assertNotNull(e);
+    }
+}
--- a/src/test/java/org/jabref/model/entry/EntryLinkListTest.java
+++ b/src/test/java/org/jabref/model/entry/EntryLinkListTest.java
@@ -17,7 +17,7 @@ import static org.junit.Assert.assertEquals;
 public class EntryLinkListTest {
 
     @Test
-    public void givenFieldValueAndDatabaseWhenParsingThenResultExpected() {
+    public void givenFieldValueAndDatabaseWhenParsingThenExpectLink() {
         String fieldValue = "test";
         BibDatabase bibDatabase = new BibDatabase();
         List<ParsedEntryLink> links = EntryLinkList.parse(fieldValue, bibDatabase);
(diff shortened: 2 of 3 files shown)


src/test/java/org/jabref/model/entry/EntryLinkListTest.java
```diff
@@ -0,0 +1,38 @@
...
+        assertEquals(fieldValue, link.getKey());
+        assertEquals(bibDatabase, link.getDataBase());
+        assertEquals(Optional.empty(), link.getLinkedEntry());
+    }
+
+    @Test
+    public void givenNullFieldValueAndDatabaseWhenParsingThenExpectLinksSizeZero() {
+        List<ParsedEntryLink> links = EntryLinkList.parse(null, bibDatabase);
+
+        assertEquals(0, links.size());
```
Siedlerchr: It it better when you compare the list itself instead of the size, makes it easier to spot eventual errors.
So:  assertEquals(Collections.emptyList(), links)

src/test/java/org/jabref/model/entry/EntryLinkListTest.java
```diff
@@ -0,0 +1,38 @@
...
+
+    String fieldValue = "test";
+    BibDatabase bibDatabase = new BibDatabase();
+
+    @Test
+    public void givenFieldValueAndDatabaseWhenParsingThenExpectLink() {
+        List<ParsedEntryLink> links = EntryLinkList.parse(fieldValue, bibDatabase);
+        ParsedEntryLink link = links.get(0);
+
+        assertEquals(fieldValue, link.getKey());
```
Siedlerchr: Please use only one assert per test, make it easier to spot erros on failing tests and I would recommend directly comparing the ParsedEntryLink: Just create a new ParsedEntrylink with the expected values and compare it. 
You may need to implement/override (or let it generate from your IDE) a toString() method, otherwise your junit test will probably fail (Encountered that recently in another project)

sammann-fnordbedarf-de:
@Siedlerchr Thanks for your review!

I implemented the requested changes.

src/test/java/org/jabref/model/database/event/AutosaveEventTest.java
```diff
@@ -0,0 +1,14 @@
+package org.jabref.model.database.event;
+
+import org.junit.Test;
+
+import static org.junit.Assert.assertNotNull;
+
+public class AutosaveEventTest {
+
+    @Test
+    public void givenNothingWhenCreatingThenNotNull() {
```
koppor: What is this test useful for? When creating a java object, the object is always non-null.

Maybe, it could be check for no exceptions being thrown. Maybe add `AndNoExceptionThrown`.

src/test/java/org/jabref/model/entry/EntryLinkListTest.java
```diff
@@ -0,0 +1,45 @@
...
+        assertEquals(key, link.getKey());
+    }
+
+    @Test
+    public void givenFieldValueAndDatabaseWhenParsingThenExpectDataBase() {
+        assertEquals(database, link.getDataBase());
+    }
+
+    @Test
+    public void givenFieldValueAndDatabaseWhenParsingThenExpectEmptyLinkedEntry() {
```
koppor: Is it possible to add a "real" test linking a "real" other entry?

src/test/java/org/jabref/model/database/event/AutosaveEventTest.java
nanooq: Good point, I will add another test for exceptions.

Yes, the tests concern exception handling. While trivial, they may turn out for regression tests later. In general, I would like to increase overall testcoverage of classes, then methods, then lines. 

I start with class covering, which sometimes is trival, but allows me to get used to this projects conventions.

src/test/java/org/jabref/model/entry/EntryLinkListTest.java
nanooq: I do not understand. Is a "real" other entry just a `new BibEntry()` for you?

koppor:
Entry linking is described from the user's side at http://help.jabref.org/en/EntryLinks

Example bib file: https://github.com/JabRef/jabref/blob/master/src/test/resources/testbib/crossref.bib

Entry `DBLP:conf/wicsa/ZimmermannWKG15` links to `DBLP:conf/wicsa/2015`.

So, I am leaning to ask whether it is possible to create two BibEntry objects `source` and `target`. The `target` has the BibTeX key `target` and the source `crossrefs` `target`.

koppor:
Since this PR brings the test coverage forward, we merge this PR.

Thank you for your contribution @nanooq. We hope that we will see further increase of the test coverage. This really helps to stabilize JabRef.

PR #3193, opened 2017-09-03
Improvement in Japanese translation

koji33:
Removed spacing error and improved some translations


Siedlerchr:
Thanks for the update! Now the files should be consistent

Issue #2284, opened 2016-11-18
Metadata properties on shared databases.


g-mos:
JabRef version 3.7 on Ubuntu 16.04 with PostgreSQL database

Steps to reproduce:
The metadata of the user are saved in the database as common metadata. Therefore the last users general file directory is saved in the database and when an other user opens the database he loads these metadata, and this general file directory
This also affects #2255 
Furthermore is the groups structure somehow handled in SQL databases?

koppor:
Firstly, #2219 has to be fixed. Otherwise, metadata is lost anyway.

When you use the "User-specific file directory", everything should work as expected: Each user has its own file directory. This assumes that each user has a separate Windows/Linux/Mac login name. This name is used at the user-specific-file-directory-setting. The general file directory, as you also noticed, applies for all users.

In any case, each user can configure his JabRef instance at Options->Preferences->File->"Main file directory". This directory also applies for shared databases.

All in all, I think, this is more a documentation issue (http://help.jabref.org/en/FileLinks is outdated) than a real issue of JabRef. Could you try with the "user-specific dirctory"?

The storage of the group structure has changed since version 3.4 (https://github.com/JabRef/jabref/pull/1276). It still has some issues (https://github.com/JabRef/jabref/issues/1495), but works well in other regards.

g-mos:
@koppor, thank you for your reply.
Concerning "User-specific file directory" I will reply to that tomorrow, when I have multiple users in the database environment. We have a common user logging in the database, so the general directory was false. Definitely the documentation needs more clarification!
Concerning the groups I will put my two cents in #1495, since in a shared database things are missing!

g-mos:
I am now using the JabRef 3.8-dev--snapshot--2016-11-21--master--7362d02
 version which fixes the #2219 problem. This had a significant impact as @koppor suggested. In the database when you use user-specific file directory these user information are also stored in the db and can be handled properly. Furthermore in this version the groups tree is also saved in the metadata information.
Yet the program seems a little unstable. If you pull  changes from the shared db (which by the way I believe is the wrong name for these button, sync with db might be more accurate) the program hangs with no error messages. Sometimes the groups info are not saved when you close jabref (but I cannot steadily reproduce this error).

koppor:
@g-mos If understand correctly, your issue is solved, but JabRef is unstable in this regard.

Can we have a chat on skype or slack? Feel free to write me on my email address published on my GitHub profile.

g-mos:
[...]

Issue #2647, opened 2017-03-15
Group information not saved in a Postgres db


g-mos:
JabRef version JabRef-4.0.0-dev on Ubuntu 16.04

Steps to reproduce:

1. Opened a Postegres literature db
2. Created some groups
3. Closed and reopend the db. The groups have disappered

koppor:
When #2284 is fixed, this should be fixed, too.

tgkolda:
I have the same problem with 

JabRef 4.0.0-dev--snapshot--2017-06-09--master--b5f1e7d7e
Windows 7 6.1 amd64 
Java 1.8.0_66

Thanks for any help! I use the groups frequently and really need them to work. I've switched to the latest snapshot due to Bug #2898. I'm really stuck because I need both groups and DOIs to work! Thanks for any help.

lenhard:
@koppor: You've written above that this is fixed by #2284. That has been merged, so I guess we can close the issue?

PR #3195, opened 2017-09-03
Fix local db and shared db sync #2284

betheunique:
Fixed database sync issue. While syncing with shared db there was no listener for GroupUpdateEvent but there is MetaData listener.
Two Scenarios : -
1. Either we can add GroupUpdateEvent listener.
2. or we can reused MetaData listener (as GroupUpdate is also part of Metadata). 

I have used 2 scenario as per code reuse. I just added a post event to MetaData.


Siedlerchr:
Cool stuff! Thanks for your contribution! Will check and test it tomorrow

koppor:
There are free cloud offerings to test - see http://help.jabref.org/en/SQLDatabase#try-it-out.

I am currently not using the shared database feature, but I think, this fixes https://github.com/JabRef/jabref/issues/2284 and presumably also https://github.com/JabRef/jabref/issues/2647.

@g-mos, @tgkolda,  @Fab7c4 could you comment whether the JabRef build at http://builds.jabref.org/bg-shared-db-sync/ works for you?

PR #3196, opened 2017-09-04
Fix tooltip for hotkey

koppor:
Before

grafik

After:

grafik


src/main/java/org/jabref/gui/util/KeyStrokeToString.java
```diff
@@ -0,0 +1,350 @@
...
+
+import javax.swing.KeyStroke;
+
+/**
+ * Class code adapted from http://www.java2s.com/Tutorial/Java/0260__Swing-Event/ConvertingaKeyStroketoaString.htm
+ */
+public class KeyStrokeToString {
+
+    public static String keyStrokeToString(KeyStroke key) {
+        StringJoiner sj = new StringJoiner(" ");
```
lenhard: I hate to nitpick, but could you assign more meaningful names to this and the following variable?

koppor: I googled for that and just added it. I think, it was always broken.

Siedlerchr: This KeyStrokeToSTring class looks really odd. I am not sure if we really need it for javafx

src/main/java/org/jabref/gui/entryeditor/EntryEditor.java
```diff
@@ -917,7 +918,8 @@ public void actionPerformed(ActionEvent e) {
         private AutoLinkAction() {
             putValue(Action.SMALL_ICON, IconTheme.JabRefIcon.AUTO_FILE_LINK.getIcon());
             putValue(Action.SHORT_DESCRIPTION,
-                    Localization.lang("Automatically set file links for this entry") + " (Alt-F)");
+                    Localization.lang("Automatically set file links for this entry") +
+                            " (" + KeyStrokeToString.keyStrokeToString(Globals.getKeyPrefs().getKey(KeyBinding.AUTOMATICALLY_LINK_FILES)) + ")");
```
Siedlerchr: Why don't you call Globals.getKeyPrefs().get(KeyBinding.AUTOMATICALLY_LINK_FILES).get()=
The get Method returns already the binding name (or the default binding, e.g. ctrl+z as an Optional<String>  so no need to rely on the KeyStroke stuff

koppor: Did not find the method.

koppor:
@Siedlerchr Thank you for the hint. I hope, I got the Optional thingy right

    Globals.getKeyPrefs().get(KeyBinding.AUTOMATICALLY_LINK_FILES).map(b -> " (" + b + ")").orElse(""))

Issue #2942, opened 2017-06-24
Add copy and paste enabled textbox with version information to about window


jonasstein:
I suggest to add a copy and paste enabled textbox with version information to the about window. At the moment there is only a copy to clipboard button. It would be nice to see the information already in the window. There should be also more detailed information on the Java environment
![jabref_about](https://user-images.githubusercontent.com/298682/27512491-13343ee6-5942-11e7-8a53-be38d1baa1e9.png)
.

Siedlerchr:
At least the snapshot information should be displayed and copied.

koppor:
+1 for implementing it 😇

steph37tours:
So I agree to look at this.
I suppose I have firstable to find the form and add the textbox.

Siedlerchr:
@steph37tours In our wiki you find some information about the package structure and some explainations to the architecture to get an understanding how things work together: https://github.com/JabRef/jabref/wiki/High-Level-Documentation

Even I often struggle to locate a specific gui dialog in the package structure.. So I give you a little hint:  Look at gui.help

steph37tours:
Thank you @Siedlerchr, I found the gui.help.
As long as for instance I've got any experience into javafx. Does I have to install scenebuilder ?

Siedlerchr:
@steph37tours Yes, it might come in handy and eases the gui creation. The resource files (fxml) are stored in the respective folder under /src/main/resources/ 
If you use eclipse, I can recommend installing the e(fx)clipse plugin: http://www.eclipse.org/efxclipse/index.html

steph37tours:
Actually version show is "unknow" and I've got any developers, etc...
How to have some info to test ?

steph37tours:
I use a TextArea.

steph37tours:
I create a String into the controller. Into a Label, I see it, but not into the textarea.

jonasstein:
Please do not get me wrong, but I suggest you use a chat like IRC for discussion in spite of a bug tracker. I (and all others) get a notification for each activity here and it is rather about general Java programming/ Jabref structure than the bug. Anyway, thank you for taking care of this ticket and your effort.

Siedlerchr:
@steph37tours  Agree with @jonasstein  You can add me at Skype for more discussions:
My Skype name is  `siedlerkiller2.0`

lynyus:
@jonasstein This issue was fixed in https://github.com/JabRef/jabref/pull/3205

[...]

PR #3198, opened 2017-09-05
localisationTestForInvalidStrings

steph37tours:
…te for instance


lynyus:
Hey @steph37tours, this goes into the right direction. I think the test fails in the right places, but the message is quite big.
What would be better is something like this:
`Found an invalid character in the <language> localization of <JabRef | Menu>: The <key | value> <actual string> contains a space!`

This would require to iterate over all property strings, similar as in `LocalizationConsistencyTest.keyValueShouldBeEqualForEnglishPropertiesMenu()`.
Please give it another shot.

steph37tours:
Thank you @Lynyus. I see what you mean. I'll correct this tomorrow if it's ok :-).

koppor:
I am unsure whether JabRef really cannot handle underscores. I once looked at the code and I think, we could get rid of using the underscores and always use spaces. Could you check that? I think `_` is so 2005 😇

lynyus:
Wow @koppor (;

lynyus:
@koppor to make this crystal clear: Do we need to fix the underscores or not? 
In https://github.com/JabRef/jabref/blob/ebbeb1d249edf7134abf92d24c2031ae54d19881/src/main/java/org/jabref/logic/l10n/LocalizationKey.java#L14 there is a comment that fixes the spaces, but not the `#`, `!`, `=`, and `:`.
So is there need to pursue this PR, or not?

Siedlerchr:
@koppor  @lynyus  That is specified in the java docs:
http://docs.oracle.com/javase/8/docs/api/java/util/Properties.html#load(java.io.Reader)

> .A comment line has an ASCII '#' or '!' as its first non-white space character; comment lines are also ignored and do not encode key-element information
In addition to line terminators, this format considers the characters space (' ', '\u0020'), tab ('\t', '\u0009'), and form feed ('\f', '\u000C') to be white space. 

> The key contains all of the characters in the line starting with the first non-white space character and up to, but not including, the first unescaped '=', ':', or white space character other than a line terminator

koppor:
Then, we can have the values without underscores? That would improve the language files a lot!

lynyus:
The question is, of course, why they were introduced in the first place. Is there any pitfall if we remove them?!?

Siedlerchr:
From what I understand the key can't contain whitespaces or it would be recognized as key:
`A key is all the characters up to the first whitespace or a key/value separator `
http://kajabity.com/kajabity-tools/java-properties-classes/the-java-properties-file-format/

steph37tours:
I just push a change. I hope it is what you want. I've got difficult to test it.
And I'm not sure for the "throws IOException" Eclipse ask  me for remove it.
Thank you for help and permit to participate. It was very interesting for me.

src/test/java/org/jabref/logic/l10n/LocalizationConsistencyTest.java
```diff
@@ -183,6 +183,28 @@ public void localizationParameterMustIncludeAString() throws IOException {
         }
     }
 
+    @Test
+    public void localisationTestForInvalidStrings() throws IOException {
```
lynyus: Indeed, you can remove the throws statement.

src/test/java/org/jabref/logic/l10n/LocalizationConsistencyTest.java
```diff
@@ -183,6 +183,28 @@ public void localizationParameterMustIncludeAString() throws IOException {
...
+                String propertyFilePath = String.format("/l10n/%s_%s.properties", bundle, lang);
+
+                // read in
+                Properties textKeys = LocalizationParser
+                        .getProperties(propertyFilePath);
+                //parse object "textKeys" to find any spaces
+                for (Map.Entry<Object, Object> entry : textKeys.entrySet()) {
+                    String expectedKeyEqualsKey = String.format("%s=%s", entry.getKey(), entry.getKey());
+                    String actualKeyEqualsValue = String.format("%s=%s", entry.getKey(),
+                            entry.getValue().toString().replace(" ", ""));
```
lynyus: I would suggest that you check for spaces using String.contains(). Then there is no need to compare strings. later.

src/test/java/org/jabref/logic/l10n/LocalizationConsistencyTest.java
```diff
@@ -183,6 +183,28 @@ public void localizationParameterMustIncludeAString() throws IOException {
...
+            for (String lang : Languages.LANGUAGES.values()) {
+                String propertyFilePath = String.format("/l10n/%s_%s.properties", bundle, lang);
+
+                // read in
+                Properties textKeys = LocalizationParser
+                        .getProperties(propertyFilePath);
+                //parse object "textKeys" to find any spaces
+                for (Map.Entry<Object, Object> entry : textKeys.entrySet()) {
+                    String expectedKeyEqualsKey = String.format("%s=%s", entry.getKey(), entry.getKey());
+                    String actualKeyEqualsValue = String.format("%s=%s", entry.getKey(),
```
lynyus: This is not correct. While in the english localization the key must equal the value, in all other languages the the key is different from the value.

src/test/java/org/jabref/logic/l10n/LocalizationConsistencyTest.java
```diff
@@ -183,6 +183,28 @@ public void localizationParameterMustIncludeAString() throws IOException {
...
+                            entry.getValue().toString().replace(" ", ""));
+
+
+                    assertEquals("Found an invalid character in the "+ lang + "localization of" + bundle + ": The " + entry.getKey().toString() + ":" + expectedKeyEqualsKey + " contains a space!",expectedKeyEqualsKey, actualKeyEqualsValue);
```
lynyus: This looks fine, however make sure to include some spaces and quotes to make the sentence more readable
`Found an invalid character in the inlocalization ofJabRef: The pairs_processed:pairs_processed=pairs_processed contains a space! `
=>
`Found an invalid character in the 'in' localization of 'JabRef': The 'pairs_processed:pairs_processed=pairs_processed' contains a space! `

Also the error should be apparent from the message. The String above does not have a space.

steph37tours: I give you push access. I'm correcting my code.

steph37tours: I think it's ok now. But I didn't success to test ! It seems I don't know how to do.

src/test/java/org/jabref/logic/l10n/LocalizationConsistencyTest.java
```diff
@@ -183,6 +183,23 @@ public void localizationParameterMustIncludeAString() throws IOException {
...
+        for (String bundle : Arrays.asList("JabRef", "Menu")) {
+            for (String lang : Languages.LANGUAGES.values()) {
+                String propertyFilePath = String.format("/l10n/%s_%s.properties", bundle, lang);
+
+                // read in
+                Properties textKeys = LocalizationParser
+                        .getProperties(propertyFilePath);
+                //parse object "textKeys" to find any spaces
+                for (Map.Entry<Object, Object> entry : textKeys.entrySet()) {
+                    assertTrue("Found an invalid character in the " + lang + " localization of " + bundle + " : The " + entry.getKey().toString() + " : " + entry.getValue().toString() + " contains a space!", entry.getValue().toString().contains(" "));
```
lynyus: The statement must be inverted. We want the test to fail if there is a '` `' in the String.
Also, you only verify the values, but not the keys of the entry.
Just add another line where you assert that there are also no spaces in the key.

src/test/java/org/jabref/logic/l10n/LocalizationConsistencyTest.java
```diff
@@ -183,6 +183,23 @@ public void localizationParameterMustIncludeAString() throws IOException {
         }
     }
 
+    @Test
+    public void localisationTestForInvalidStrings() {
```
lynyus: I would suggest to make this name a bit more precise: maybe locali**z**ationTestForInvalidCharacters?

src/test/java/org/jabref/logic/l10n/LocalizationConsistencyTest.java
steph37tours: @lynyus, I commit and push again. 
I correct the message with "At key", which seems to me more clear to find the line.
For the test : I use Junit. It find some failures

java.lang.AssertionError: Found an invalid character in the in localization of JabRef : proces_berpasangan At key : pairs_processed contains a space!

If I correct the P into Pairs_processed, it runs.

But I've got the same trouble in New_article , but I don't find why...

lynyus: Have you tried using assertFalse instead if assertTrue?

lynyus:
So if you change the Assertions to assertFalse or negate the condition the tests should be operational and we can see if they pass the test on our build server
`assertFalse("Found an invalid character in the " + lang + " localization of " + bundle + " : " + entry.getValue().toString() + " At key : " + entry.getKey().toString() + " contains a space!", entry.getValue().toString().contains(" "));`

steph37tours:
Yes, I change and commit. I found spaces in pt_BR localization :

java.lang.AssertionError: Found an invalid character in the pt_BR localization of JabRef : Normalizar para o formato_de_nome_do_BibTeX At key : Normalize_to_BibTeX_name_format contains a space!

steph37tours:
So it is done now for me ?

lynyus:
Can you push your commit please? (; You'll have to pull first, as I have already fixed the localization files on your branch.

src/test/java/org/jabref/logic/l10n/LocalizationConsistencyTest.java
steph37tours: Yes. It runs !

steph37tours:
sorry, it doesn't want to push ! It seems I have to pull your change before. I 'm looking at...

steph37tours:
I'm sorry I can't push !
I've got the error that the branch countain work I doesn't have in local.
I try git pull
git pull --rebase without success for instance !

lynyus:
on the current branch try the following:
`git fetch`
`git merge`
`git push`

steph37tours:
git merge said not remote branch for the current

steph37tours:
git merge Localisation_test_for_invalid_Strings 
said up-to-date

lynyus:
Hm. Maybe try a GUI client like Github Desktop to figure everything out. Or clone the repository again and apply your changes again.

steph37tours:
I'm sorry, I just clone and try to push again with the same error message. I continue to look for a solution.

steph37tours:
I don't know what to do.
Into Eclipse, the menu team/pull give to me an error "Missing commit"

steph37tours:
perhaps I have to do another fork ?

steph37tours:
It seems it's ok now. I don't know what happen exactly.
Really thank you for this first experience, thank you for the time you share with me to help.

src/test/java/org/jabref/logic/l10n/LocalizationConsistencyTest.java
```diff
@@ -183,6 +182,25 @@ public void localizationParameterMustIncludeAString() throws IOException {
...
+            for (String lang : Languages.LANGUAGES.values()) {
+                String propertyFilePath = String.format("/l10n/%s_%s.properties", bundle, lang);
+
+                // read in
+                Properties textKeys = LocalizationParser
+                        .getProperties(propertyFilePath);
+
+                //parse object "textKeys" to find any spaces
+                for (Map.Entry<Object, Object> entry : textKeys.entrySet()) {
+                    assertFalse("Found an invalid character in the " + lang + " localization of " + bundle + " : " + entry.getValue().toString() + " At key : " + entry.getKey().toString() + " contains a space!", entry.getValue().toString().contains(" "));
```
Siedlerchr: Just a minor thing: if you concatenate Strings and objects (no matter which type) with` "some text " + object    `
You don't need to do an explicit call to "toString()" The method is automatically invoked. But beware, in the case where you want to call a methond on a string like the contains, you have to use the toString

steph37tours: Thank you @Siedlerchr for this explanation.

lynyus:
I'll merge this now, thanks for the efforts, @steph37tours! I hope you have learned something and are still eager to contribute. (;
How about this? https://github.com/JabRef/jabref/issues/2942 It's a small feature in the UI, and everybody would be quite happy if somebody finds time to implement it!

steph37tours:
Thank you very much @lynyus !
I learn many things with this experience.
So I agree and wish to continue ! :-)

koppor:
Good to see that you won the git war :). -- For learning git, I would recommend the tutorials listed at https://github.com/dictcp/awesome-git.

steph37tours:
Thank you @koppor for the link ! I'll study it !

PR #3203, opened 2017-09-07
When groups are removed, also remove them from entries

### LINKED ISSUE #3190: Deleting a group via groups interface does not remove that group reference from BibTex source
JabRef 4.0-dev
Linux 4.4.0-92-generic amd64
Java 1.8.0_131

Steps to reproduce:

1. Open group interface. (Groups -> toggle groups interface)
2. Create a group
3. Add an item to the group
4. Right click on the group in the group interface -> remove group and subgroups

The group gets removed but the reference to it in the BibTeX source of each item that was in the group does not. (i.e. groups = {abc}).
This causes an unstable state where jabref might freeze if you try adding that item to another group because of a null pointer.

 [JavaFX Application Thread] ERROR org.jabref.FallbackExceptionHandler - Uncaught exception Occurred in Thread[JavaFX Application Thread,5,main]
java.lang.NullPointerException: null

lenhard:
Fixes #3190


src/main/java/org/jabref/gui/groups/GroupTreeController.java
```diff
@@ -81,10 +81,10 @@ public void initialize() {
                 (newSelectedGroups) -> newSelectedGroups.forEach(this::selectNode);
         Consumer<List<TreeItem<GroupNodeViewModel>>> updateViewModel =
                 (newSelectedGroups) -> {
-                    if (newSelectedGroups == null) {
-                        viewModel.selectedGroupsProperty().clear();
-                    } else {
+                    try {
                         viewModel.selectedGroupsProperty().setAll(newSelectedGroups.stream().map(TreeItem::getValue).collect(Collectors.toList()));
+                    } catch (NullPointerException e) {
```
Siedlerchr: Why do you catch a NPE here? Looks strange

lenhard: I know. For some reason this Binding triggers a NullPointerException when a group is removed, even though there was a null check directly before. I tried hard, but I could not pin down the actual cause of this (or which object actually is null). The debugger is really of no use with lambdas and bindings. But look at the complete code change in this class, it's just a different way of doing the same thing with a null check.

Apart from that, the NPE doesn't have any effect. The removal works anyway and the selection is cleared. All that this line does is to avoid a stack trace in the log.

I can offer to add a code comment explaining this, but I don't have more time to investigate why this is triggered (spent enough on that already). So it's either this solution or a stack trace in the log.

Siedlerchr: Okay, add a comment and I am fine. Not the best idea yeah but I know how hard it is do debug binding lamdba stuff

lenhard: Okay, I took the time to rewrite that part in an imperative fashion, track down the bug, and fix it properly. It is totally insane... `List<TreeItem<GroupNodeViewModel>> newSelectedGroups` is a list that can actually contain `null` values. When you iterate over that list in an absolutely standard fashion, e.g. `for (TreeItem<GroupNodeViewModel> model : newSelectedGroups) {` than `model` can actually be null.

If that's not a violation of the basic contract of how collections work in Java, then I don't know what would be.

Siedlerchr: This is really odd

Siedlerchr: The NPE is valid in the for each loop, it does not check against NULL, it only skips the loop if the list is empty. So this is more of a design issue that the TreeItem could be null at one poin
https://stackoverflow.com/questions/85190/how-does-the-java-for-each-loop-work?rq=1

lenhard: Interesting, thanks :) You never finish learning about even the most basic aspects of the language.

Anyway, the code avoids the NPE now through proper checking, so this PR should be ready to go.

src/main/java/org/jabref/gui/groups/GroupTreeController.java
```diff
@@ -81,11 +89,7 @@ public void initialize() {
                 (newSelectedGroups) -> newSelectedGroups.forEach(this::selectNode);
         Consumer<List<TreeItem<GroupNodeViewModel>>> updateViewModel =
```
tobiasdiez: With these changes, I think the `updateViewModel` consumer can be replaced by `this::updateSelection` (or whatever the exact syntax is for these generics)

src/main/java/org/jabref/gui/groups/GroupTreeController.java
```diff
@@ -52,15 +52,23 @@
 
     private static final Log LOGGER = LogFactory.getLog(GroupTreeController.class);
 
-    @FXML private TreeTableView<GroupNodeViewModel> groupTree;
-    @FXML private TreeTableColumn<GroupNodeViewModel, GroupNodeViewModel> mainColumn;
-    @FXML private TreeTableColumn<GroupNodeViewModel, GroupNodeViewModel> numberColumn;
-    @FXML private TreeTableColumn<GroupNodeViewModel, GroupNodeViewModel> disclosureNodeColumn;
-    @FXML private CustomTextField searchField;
+    @FXML
```
tobiasdiez: I prefer the condensed 1-line form for javafx classes since these @FXML and @Inject tags make the code rather clumsy.... but both forms are fine with me (we should just stick to one convention).

src/main/java/org/jabref/model/database/BibDatabase.java
```diff
@@ -613,4 +613,15 @@ public String generateSharedDatabaseID() {
     public DuplicationChecker getDuplicationChecker() {
         return duplicationChecker;
     }
+
+    public List<BibEntry> getEntriesInGroup(String groupName) {
```
tobiasdiez: The GroupsTreeNode class already contains the `getMatchingGroups` method https://github.com/JabRef/jabref/blob/master/src/main/java/org/jabref/model/groups/GroupTreeNode.java#L164, which slightly differs from this code here but is very similar in its purpose (its also matches dynamics and search groups). Thus I would propose to move the new code to a new method `GroupTreeNode::getMatchingExplicitGroups`. Moreover, please reuse `ExplicitGroup:contains` https://github.com/JabRef/jabref/blob/master/src/main/java/org/jabref/model/groups/WordKeywordGroup.java#L105 for the group membership check.

src/main/java/org/jabref/gui/groups/GroupTreeController.java
lenhard: This was done by the IntelliJ autoformat function. I'll gladly change it back.

src/main/java/org/jabref/gui/groups/GroupTreeController.java
lenhard: Ok, I'll have a check and see what I can do.

src/main/java/org/jabref/model/database/BibDatabase.java
lenhard: Ok, thanks. I'll look into these things. I added this method basically because I'm not at all an expert in the group code and didn't know about these methods.

src/main/java/org/jabref/gui/groups/GroupTreeController.java
lenhard: Done

src/main/java/org/jabref/model/database/BibDatabase.java
lenhard: Moved and refactored. I kept the method name, though.

lenhard:
Ok, I managed to get this done and I think I've addressed everything as the reviewers desired. It can be merged from my point of view. So please have a look and merge if you think it is ok.

src/main/java/org/jabref/gui/groups/GroupTreeViewModel.java
```diff
@@ -248,13 +249,23 @@ public void removeGroupAndSubgroups(GroupNodeViewModel group) {
...
             writeGroupChangesToMetaData();
         }
     }
 
+    private void removeGroupsAndSubGroupsFromEntries(GroupNodeViewModel group) {
+        for (GroupNodeViewModel child: group.getChildren()) {
+            removeGroupsAndSubGroupsFromEntries(child);
+        }
+
+        List<BibEntry> entriesInGroup = group.getGroupNode().getEntriesInGroup(this.currentDatabase.get().getEntries());
```
tobiasdiez: You need to check if the `group` is an explicit group. Otherwise, removing a dynamic keyword-based group also removes the keyword from the entries that are matched by this group - this behavior is probably not what we want.

lenhard: Ok, thanks for clarifying. But then what do I do if it is a dynamic keyword group? No removal at all?

tobiasdiez: Yes, I would just ignore every other type of group but the explicit group (or ask the user for confirmation about the removal).

lenhard: Ok, done!

lenhard:
One more try :) Now, only ExplicitGroups are removed from entries, keyword-based groups are kept. I hope this is the last missing puzzle piece.

tobiasdiez:
Sorry for being pestic, but with this implementation keyword groups that are children of the deleted explicit group are still handled wrongly. I directly committed a fix. If you are fine with it, you have my ok for merge.

By the way, in order to convince you about the testability of Javafx code, you are invited to write a test covering the new behavior... 😈

lenhard:
Thanks, challenge accepted :) I did not only add one test, but three.

I will merge when the build goes through.

Issue #3199, opened 2017-09-05
LatexEditor: defunct ==> drop JabRef support


mlep:
The latest version of LatexEditor is from 2009 (see http://www.latexeditor.org/news.html).
Maybe the "Push to LatexEditor" could be removed from JabRef?

h93xV2:
I would like to solve this issue.

h93xV2:
If this feature does need to be removed, I would like to do so.

BKDevelop:
Hey,
I am new to the codebase and also still fairly new to java development. My sleepy brain thinks that I solved that issue, but I will check again tomorrow morning before making a pull request.

Cheers,
Ben

PR #3204, opened 2017-09-08
Drop latex editor support

BKDevelop:
Fixes #3204 
This removes  LatexEditor support ("Push to LatexEditor") in JabRef. 

As stated in #3199 the last update to LatexEditor was made in 2009 (see http://www.latexeditor.org/news.html ) and the software can be considered outdated.

I removed LatexEditor from the applications list in PushToApplications class. I deleted the PushToLatexEditor class since it is now unused.
gradlew check was successful.

I am new here (first pull request) and also still fairly new to Java, so any c&c would be greatly appreciated. Especially let me know if either the practice of commits or this request is not up-to-standard :)

Cheers,
Ben

edit: 
removed LatexEditor from Preferences and corrected the URL link style in Changelog :)


CHANGELOG.md
```diff
@@ -39,7 +39,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - Renaming files now truncates the filename to not exceed the limit of 255 chars [#2622](https://github.com/JabRef/jabref/issues/2622)
 
 ### Removed
-
+- We removed support for LatexEditor, as it is not under active development. (https://github.com/JabRef/jabref/issues/3199a)
```
Siedlerchr: Please follow the styles from the other entries for creating the URL links. And I think you have an 'a' too much at the issue number

Siedlerchr:
For your first PR you did all right 🍾.  If you want to further contribute to JabRef, I would suggest looking at the Wiki here https://github.com/JabRef/jabref/wiki where we have collected some JabRef specific things regarding the architecture and also some Java tips

CHANGELOG.md
lynyus: yeap. `[#3199](https://github.com/JabRef/jabref/issues/3199)` would be correct.

BKDevelop: Oops, totally overlooked that :D Should be fixed now, thanks guys

lynyus:
I'll merge this now, thanks for the contribution!  🎉 
Have a look at the other [issues](https://github.com/JabRef/jabref/issues), those labeled with [beginner](https://github.com/JabRef/jabref/issues?q=is%3Aissue+is%3Aopen+label%3Abeginner) are usually easier to fix if you don't have much prior experience with JabRef!

PR #3210, opened 2017-09-11
Add file description to gui and fix sync bug

### LINKED ISSUE #3208: Description of LinkedFiles is not synced to the bibtex code/database
Existing files and descriptions are shown, but not changed on editing.

Follow up from  https://github.com/JabRef/jabref/issues/2930#issuecomment-328354137

Siedlerchr:
Fixes #3208


tobiasdiez:
The problem is that the changes in the underlying `LinkedFile` are not propagated to the list `LinkedFilesEditorViewModel.files` so that the serialization https://github.com/JabRef/jabref/blob/2c5e7f2940481d66d50573e6fc40858f55b1cc47/src/main/java/org/jabref/gui/fieldeditors/LinkedFilesEditorViewModel.java#L74-L82 is never called. By the way, also changes to the file name or icon do not trigger an update and thus are not serialized.

As a solution, I propose to convert the data fields in `LinkedFile` to real properties, add `Observable[] getObservables()` to `LinkedFile` and return these observables here:
https://github.com/JabRef/jabref/blob/2c5e7f2940481d66d50573e6fc40858f55b1cc47/src/main/java/org/jabref/gui/fieldeditors/LinkedFileViewModel.java#L109-L111

Ps: Can you pls add a screenshot how the file editor now looks like with the description added (especially when the file link is relatively long). Thanks.

Siedlerchr:
@tobiasdiez  I did as you suggested, but still does not work.  I think it has something to do with the LIstProperty, but I am not that deep inside javafx binding stuff to know why it is not

src/main/java/org/jabref/model/entry/LinkedFile.java
```diff
@@ -19,42 +27,49 @@
...
+        return link.get();
     }
 
     public void setLink(String link) {
-        this.link = link;
+        this.link.setValue(link);
+    }
+
+    public Observable[] getObservables() {
+        return new Observable[] {this.downloadProgress, this.isAutomaticallyFound};
```
tobiasdiez: you need to return `description, link, fileType` here (and move `downloadProgress` and `isAutomaticallyFound` back to the viewmodel).

The observable tell javafx which properties mark the object as changed, i.e. if `description` is changed this should mark the `LinkedFile` as changed and thus also `LinkedFileViewModel` and thus finally the list https://github.com/JabRef/jabref/blob/23c6ac6308dc4df72057eec0806570f635c1ed0b/src/main/java/org/jabref/gui/fieldeditors/LinkedFilesEditorViewModel.java#L55.
For more details see https://docs.oracle.com/javase/8/javafx/api/javafx/collections/FXCollections.html#observableList-java.util.List-javafx.util.Callback- or https://stackoverflow.com/questions/31687642/callback-and-extractors-for-javafx-observablelist

Siedlerchr:
Yeah! Got it now!
grafik

src/main/java/org/jabref/gui/fieldeditors/LinkedFileViewModel.java
```diff
@@ -107,7 +109,11 @@ public void acceptAsLinked() {
     }
 
     public Observable[] getObservables() {
-        return new Observable[] {this.downloadProgress, this.isAutomaticallyFound};
+        List<Observable> observables = new ArrayList<>(Arrays.asList(linkedFile.getObservables()));
+        observables.add(downloadOngoing);
+        observables.add(downloadProgress);
```
tobiasdiez: also add `isAutomaticallyFound`

src/main/java/org/jabref/model/entry/FileFieldWriter.java
```diff
@@ -7,7 +7,7 @@
 
 public class FileFieldWriter {
 
-    private FileFieldWriter() {
+    FileFieldWriter() {
```
tobiasdiez: Why is this constructor not private anymore?

src/main/java/org/jabref/gui/fieldeditors/LinkedFileViewModel.java
notuntoward: Just budding in here... 

The file list would be cleaner if the filenames weren't shown when there is a description, as was done in JabRef 3.x.  If there are very long descriptions, it will still be cluttered, but in my opinion., tooltips aren't an improvement.  

The problem with descriptions only in tooltips is that you have to mouse over each one.  This makes JabRef harder to use:

1.  It forces you to lift you hands off the keyboard, but more importantly,
2. You can't scan the descriptions at a glance.  Instead, you have to mouse over the first one, then mouse over the next one (which causes the previous description to vanish), and so on.

You can simulate the user experience by having somebody else make you a JabRef entry with, say five files with slightly different descriptions -- this is what your own entries will look like to you, a couple years after you have made them.  You won't remember anything.

Now, try to decide which file to open.  With tooltips, you see only one description at a time, like you're viewing them through a straw.  Now compare that with the JabRef 3.x experience, where all descriptions are visible; with just one glance, you can comprehend all of them.

Siedlerchr: Sounds reasonable!

Siedlerchr: I now changed the order: Description is shown first and then the file link

src/main/java/org/jabref/model/entry/FileFieldWriter.java
Siedlerchr: Was a relict from testing. Didn't see that it had static method on the first try

src/main/java/org/jabref/gui/fieldeditors/LinkedFileViewModel.java
notuntoward: Come to think of it, you could even make the filename a tooltip.  Once you've got the description, the filename is probably not so useful, but some situation, you still might want to know it.

But then again, there is the "see everything at a glance" argument I just made...

src/main/java/org/jabref/gui/fieldeditors/LinkedFilesEditorViewModel.java
```diff
@@ -69,6 +69,7 @@ public LinkedFilesEditorViewModel(String fieldName, AutoCompleteSuggestionProvid
                 text,
                 LinkedFilesEditorViewModel::getStringRepresentation,
                 this::parseToFileViewModel);
+
```
lenhard: Very minor, but as far as I can see this blank line is unnecessary.

src/main/java/org/jabref/model/entry/LinkedFile.java
```diff
@@ -63,31 +76,50 @@ public boolean equals(Object o) {
...
+     * Writes serialized object to ObjectOutputStream, automatically called
+     * @param out {@link ObjectOutputStream}
+     * @throws IOException
+     */
+    private void writeObject(ObjectOutputStream out) throws IOException {
+        out.writeUTF(getFileType());
+        out.writeUTF(getLink());
+        out.writeUTF(getDescription());
+        out.flush();
+
```
lenhard: Very minor, but as far as I can see this blank line is unnecessary.

src/main/java/org/jabref/model/entry/LinkedFile.java
```diff
@@ -63,31 +76,50 @@ public boolean equals(Object o) {
...
+    /**
+     * Reads serialized object from ObjectInputStreamm, automatically called
+     * @param in {@link ObjectInputStream}
+     * @throws IOException
+     */
+    private void readObject(ObjectInputStream in) throws IOException {
+        fileType = new SimpleStringProperty(in.readUTF());
+        link = new SimpleStringProperty(in.readUTF());
+        description = new SimpleStringProperty(in.readUTF());
+
```
lenhard: Very minor, but as far as I can see this blank line is unnecessary.

tobiasdiez:
I just discovered that there is additional space between the icon and the file name if the file doesn't have a description. Can you please fix this. Thanks!

Siedlerchr:
Fixed in master.

PR #3211, opened 2017-09-12
Added logic to check whether linked file already exists

Andrew-Lev:
#3172 
Added a check in LinkedFileViewModel.rename() to see if a file which is
being renamed already exists in the target location. If such a file is
found then an error popup displays informing the user of the conflict.
User will be prompted to choose whether to overwrite or cancel.

Also see #3188 for additional details.


Andrew-Lev:
This work fine and includes all the requested changes from the previous pull request with one caveat.  If the file being overridden is related to a different entry and has different case, such as Entry 1 = [abc] renamed to [def], Entry 2 = [DEF], then the 2nd entry file association will target a now non-existing file.  Its possible to get around this by searching the database for any entries that have the old filename [DEF] and changing the association to the new filename, but I dont know if thats something desireable so I decided to push this as is for now.

src/main/java/org/jabref/logic/cleanup/RenamePdfCleanup.java
```diff
@@ -158,4 +159,35 @@ public String getTargetFileName(LinkedFile flEntry, BibEntry entry) {
...
+            matchedByDiffCase = stream
+                    .filter(name -> name.toString().equalsIgnoreCase(targetFilePath.toString()))
+                    .findFirst()
+                    .get();
+        } catch (IOException e) {
+            LOGGER.error("Could not get the list of files in target directory", e);
+        }
+
+        if (!Files.exists(targetFilePath) && (matchedByDiffCase == null)) {
+            return null;
```
tobiasdiez: Please use [Optional<Path>](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html) as a return type and return an empty optional here instead of  `null`.

src/main/java/org/jabref/logic/cleanup/RenamePdfCleanup.java
```diff
@@ -158,4 +159,35 @@ public String getTargetFileName(LinkedFile flEntry, BibEntry entry) {
...
+        Path oldFilePath = flEntry.findIn(databaseContext, fileDirectoryPreferences).get();
+
+        //Check if file already exists in directory with different case.
+        //This is necessary because other entries may have such a file.
+        Path matchedByDiffCase = null;
+        try (Stream<Path> stream = Files.list(oldFilePath.getParent())) {
+            matchedByDiffCase = stream
+                    .filter(name -> name.toString().equalsIgnoreCase(targetFilePath.toString()))
+                    .findFirst()
+                    .get();
```
Siedlerchr: findFirst also returns an optional, so you could just forward that one from here

src/main/java/org/jabref/logic/cleanup/RenamePdfCleanup.java
Siedlerchr: and here you can then check the above optional with isPresent and avoid the null

Andrew-Lev:
Ok I changed to an Optional and removed !Files.exists(targetFilePath) as its redundant.  I'm still getting the hang of git and egit so hopefully the merge didnt mess anything up.

Siedlerchr:
@Andrew-Lev  I stick to git and git gui from command line. A lot easier than struggling with the eclipse git client

src/main/java/org/jabref/logic/cleanup/RenamePdfCleanup.java
```diff
@@ -158,4 +159,29 @@ public String getTargetFileName(LinkedFile flEntry, BibEntry entry) {
...
         return unsuccessfulRenames;
     }
+
+    /**Check to see if a file already exists in the target directory.  Search is not case sensitive.
+    *
+    * @param flEntry
+    * @param entry
+    * @return First identified path that matches an existing file.  This name can be used in subsequent calls to override the existing file.
+    */
+    public Optional<Path> fileAlreadyExists(LinkedFile flEntry, BibEntry entry) {
```
lynyus: This is maybe a bit nitpicky: Think of a more intuitive name for the method that reflects the return type. `fileAlreadyExists` sounds to me like it would return a boolean. How about `findExistingFile` or something similar?

src/main/java/org/jabref/gui/fieldeditors/LinkedFileViewModel.java
```diff
@@ -175,7 +176,20 @@ public void rename() {
                     Localization.lang("Cancel"));
 
             if (confirm) {
-                pdfCleanup.cleanup(entry);
+                Optional<Path> fileConflictCheck = pdfCleanup.fileAlreadyExists(linkedFile, entry);
+                if (!fileConflictCheck.isPresent()) {
+                    pdfCleanup.cleanup(entry);
+                } else {
+                    confirm = dialogService.showConfirmationDialogAndWait(
```
lynyus: You could reduce code-duplication by extracting this method call into a private method.

lynyus:
Thanks, this looks really good now!

PR #3214, opened 2017-09-13
Real test linking real other entry

sammann-fnordbedarf-de:

_Why_
Based on https://github.com/JabRef/jabref/pull/3191#issuecomment-326902439 I added two tests concerning cross referencing. 

_What_
1. It is based on two dummy entries, where source is cross referencing target.
2. It reads the database from mentioned example bib file: https://github.com/JabRef/jabref/blob/master/src/test/resources/testbib/crossref.bib and asserts cross reference of two given entries.

_Watchout_
- I was not able to clean code method `assertSourceCrossrefsTarget` further due to the `Optional<>` concept.
- I wrote a small TestDataBuilder `BibEntryBuild.` It is an inner class, because I am uncertain where to put it.
- I wrote an crossref.bib-encapsulating class `DatabaseFromFile` to keep test code readable. It is an inner class, because I did not find a similar wrapper nor did I know where to put it. Its construction uses a path to a file to be parsed to a database.


src/test/java/org/jabref/model/entry/EntryLinkListTest.java
```diff
@@ -42,4 +67,102 @@ public void givenNullFieldValueAndDatabaseWhenParsingThenExpectLinksIsEmpty() {
...
+        boolean crossrefs = false;
+        Optional<String> sourceCrossref = source.getField(FieldName.CROSSREF);
+        Optional<String> targetCiteKey = target.getCiteKeyOptional();
+        if (sourceCrossref.isPresent()
+                && targetCiteKey.isPresent()) {
+            String actualcrossrefValue = sourceCrossref.get();
+            String expectedCrossrefValue = targetCiteKey.get();
+            crossrefs = actualcrossrefValue.contains(expectedCrossrefValue);
+        }
+        assertTrue(crossrefs);
```
Siedlerchr: Are you aware that you can directly call assertEquals on two optionals?
https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#equals-java.lang.Object-

src/test/java/org/jabref/model/entry/EntryLinkListTest.java
```diff
@@ -42,4 +67,102 @@ public void givenNullFieldValueAndDatabaseWhenParsingThenExpectLinksIsEmpty() {
         links = EntryLinkList.parse(null, database);
         assertTrue(links.isEmpty());
     }
+
+    @Test
+    public void givenTargetAndSourceWhenSourceCrossrefTargetThenSourceCrossrefsTarget() {
+        target = BibEntryBuild.ing().withId("target").withCiteKey("target").now();
```
Siedlerchr: I don't see why it's necessary to have a BibEntry build class?
Why not just simply create two new BibEntrys? And you won't need the actual database file.

e.g.
```
database = new BibDatabase()
BibEntry source = new BibEntry()
source.setCiteKey("source")
source.setField(Crossref, "target")
database.insert(source)
BibEntry target = new BibEntry()
target.setCiteKey("target")
database.insert(target)

```

Siedlerchr: Put that in the before method and it will be "freshly created" for every test

tobiasdiez: `BibEntry` actually contains a `withField` method if you want to use a more fluent api,
https://github.com/JabRef/jabref/blob/23a0610ea7cba94172e5d9f0b4b6e0591003d667/src/main/java/org/jabref/model/entry/BibEntry.java#L734-L737
If you want, you can also add a new method `withCiteKey` along these lines.

sammann-fnordbedarf-de:
Thank you for your valuable feedback, I learned a lot and implemented all change requests

PR #3224, opened 2017-09-17
Testcoverage preferences

sammann-fnordbedarf-de:
Fixed a typo in https://github.com/JabRef/jabref/compare/master...sammann-fnordbedarf-de:testcoverage-preferences?expand=1#diff-b58b89dd0bdb3e1dad1c0ae7801bae2b: from previe to preview

Added tests for creating and building classes in preferences-package in order to increase text coverage of classes. 

Constructing classes expects nothing thrown, building classes also. Where needed the most trivial variables were used. 

Increasing tests for methods will be more rigid and are planned after (nearly) full class coverage. 

### DIFF
--- a/src/main/java/org/jabref/preferences/PreviewPreferences.java
+++ b/src/main/java/org/jabref/preferences/PreviewPreferences.java
@@ -12,9 +12,9 @@ public class PreviewPreferences {
     private final String previewStyleDefault;
 
 
-    public PreviewPreferences(List<String> previewCycle, int previeCyclePosition, int previewPanelHeight, boolean previewPanelEnabled, String previewStyle, String previewStyleDefault) {
+    public PreviewPreferences(List<String> previewCycle, int previewCyclePosition, int previewPanelHeight, boolean previewPanelEnabled, String previewStyle, String previewStyleDefault) {
         this.previewCycle = previewCycle;
-        this.previewCyclePosition = previeCyclePosition;
+        this.previewCyclePosition = previewCyclePosition;
         this.previewPanelHeight = previewPanelHeight;
         this.previewPanelEnabled = previewPanelEnabled;
         this.previewStyle = previewStyle;
@@ -51,7 +51,7 @@ public class PreviewPreferences {
 
     public static class Builder {
         private List<String> previewCycle;
-        private int previeCyclePosition;
+        private int previewCyclePosition;
         private int previewPanelHeight;
         private boolean previewPanelEnabled;
         private String previewStyle;
@@ -60,7 +60,7 @@ public class PreviewPreferences {
 
         public Builder(PreviewPreferences previewPreferences) {
             this.previewCycle = previewPreferences.getPreviewCycle();
-            this.previeCyclePosition = previewPreferences.getPreviewCyclePosition();
+            this.previewCyclePosition = previewPreferences.getPreviewCyclePosition();
             this.previewPanelHeight = previewPreferences.getPreviewPanelHeight();
             this.previewPanelEnabled = previewPreferences.isPreviewPanelEnabled();
             this.previewStyle = previewPreferences.getPreviewStyle();
@@ -69,15 +69,15 @@ public class PreviewPreferences {
 
         public Builder withPreviewCycle(List<String> previewCycle) {
             this.previewCycle = previewCycle;
-            return withPreviewCyclePosition(previeCyclePosition);
+            return withPreviewCyclePosition(previewCyclePosition);
         }
 
         public Builder withPreviewCyclePosition(int position) {
-            previeCyclePosition = position;
-            while (previeCyclePosition < 0) {
-                previeCyclePosition += previewCycle.size();
+            previewCyclePosition = position;
+            while (previewCyclePosition < 0) {
+                previewCyclePosition += previewCycle.size();
             }
-            previeCyclePosition %= previewCycle.size();
+            previewCyclePosition %= previewCycle.size();
             return this;
         }
 
@@ -97,7 +97,7 @@ public class PreviewPreferences {
         }
 
         public PreviewPreferences build() {
-            return new PreviewPreferences(previewCycle, previeCyclePosition, previewPanelHeight, previewPanelEnabled, previewStyle, previewStyleDefault);
+            return new PreviewPreferences(previewCycle, previewCyclePosition, previewPanelHeight, previewPanelEnabled, previewStyle, previewStyleDefault);
         }
     }
 
--- a/src/test/java/org/jabref/preferences/PreviewPreferenceTest.java
+++ b/src/test/java/org/jabref/preferences/PreviewPreferenceTest.java
@@ -1,0 +1,41 @@
+package org.jabref.preferences;
+
+import org.junit.Test;
+
+import java.util.List;
+
+public class PreviewPreferenceTest {
+
+    private List<String> previewCycle = null;
+    private int previewCyclePosition = 0;
+    private int previewPanelHeight = 0;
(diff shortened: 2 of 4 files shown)


Siedlerchr:
you have to be careful when testing preferences, it can has side effects with the actual stored preferences,  that's why we try to mock them away in tests

src/test/java/org/jabref/preferences/PreviewPreferenceTest.java
```diff
@@ -0,0 +1,41 @@
...
+        PreviewPreferences previewPreferences =
+                new PreviewPreferences(
+                        previewCycle,
+                        previewCyclePosition,
+                        previewPanelHeight,
+                        previewPanelEnabled,
+                        previewStyle,
+                        previewStyleDefault
+        );
+        new PreviewPreferences.Builder(previewPreferences);
```
koppor: Please always use Assert in each `@Test`

src/test/java/org/jabref/preferences/PreviewPreferenceTest.java
```diff
@@ -0,0 +1,41 @@
...
+
+    private List<String> previewCycle = null;
+    private int previewCyclePosition = 0;
+    private int previewPanelHeight = 0;
+    private boolean previewPanelEnabled = false;
+    private String previewStyle = "";
+    private String previewStyleDefault = "";
+
+    @Test
+    public void givenNothingWhenCreatingPreviewPreferenceThenNothingThrown() {
```
lynyus: What is the benefit of having this test?
Is it good design that the `PreviewPreferences` class accepts null values in the constructor?
If it isn't, we should not have a test that fails as soon as somebody adds parameter validation...

src/test/java/org/jabref/preferences/VersionPreferencesTest.java
```diff
@@ -0,0 +1,21 @@
...
+import org.jabref.logic.util.Version;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+public class VersionPreferencesTest {
+
+    @Test
+    public void givenNothingWhenCreatingThenNothingThrown() {
+        new VersionPreferences(Version.parse(""));
```
lynyus: Here the same again. Nice that the test succeeds, but what do we learn from that?

src/test/java/org/jabref/preferences/VersionPreferencesTest.java
```diff
@@ -0,0 +1,21 @@
...
+    }
+
+    @Test
+    public void givenIgnoredVersionStringWhenGetIgnoredVersionThenUnknown() {
```
lynyus: In this test the system under test is the `VersionPreferences` class.
However, your assertion tests a very specific behaviour of the `Version` class. That's a mismatch for me.

src/main/java/org/jabref/preferences/PreviewPreferences.java
```diff
@@ -12,9 +12,9 @@
     private final String previewStyleDefault;
 
 
-    public PreviewPreferences(List<String> previewCycle, int previeCyclePosition, int previewPanelHeight, boolean previewPanelEnabled, String previewStyle, String previewStyleDefault) {
+    public PreviewPreferences(List<String> previewCycle, int previewCyclePosition, int previewPanelHeight, boolean previewPanelEnabled, String previewStyle, String previewStyleDefault) {
```
lynyus: Thanks for spotting this error!

src/test/java/org/jabref/preferences/PreviewPreferenceTest.java
sammann-fnordbedarf-de: I get your point. 
Implementing this test helped me understand jabref / Preview Preference better. And as there was not a basic test yet, I thought it might be useful for others to see a "skeleton" constructor. 

On another level, to which I have not jet gained enought understanding: May I ask, what is the benefit of not having this test?  If it hinders good design, then I would argue to implement good design, instead of not testing not so good design. Unfortunately, I cannot contribute to this level of discussion or programming yet (I am working on it, though).

src/test/java/org/jabref/preferences/VersionPreferencesTest.java
sammann-fnordbedarf-de: I understand that you are missing learnings from this simple test. At least for regression testing or in case of refactoring, this test may already be helpful. In future, more complex test methods may be added to this class.

src/test/java/org/jabref/preferences/VersionPreferencesTest.java
sammann-fnordbedarf-de: I would disagree with you about the mismatch. My current understanding of ```VersionPreferences``` leads me to think, that it is a specific subset of ```Version```. 
I cannot remember now, why I stumbled over ```getIgnoredVersion()```, but I used this base method to fiddle around with loading preferences. 
If the test does not belong in this class, where would you put it?

sammann-fnordbedarf-de:
@lynyus You wrote "Testing JabRef is more than line coverage." from which I conclude, that it is at least line coverage. To my knowledge, there is no document to "HowTo Test" or QA yet, so I started with the easiest at the lowest point, while looking at the code. If this is to your dislike, I apologize and am also fine with removal of these tests. 

@Siedlerchr I am interested in mocking away ```VersionPreferences``` and argue for adding examples in its test class for others to see how to do it. In my opinion, the respective testclass would be an intuitive to place to have such mocking-show-cases. Could you please point me to a "nice" mocking case, I would then like to implement "how-to-mock-this-class"-test.   

@lenhard I beg your pardon, somehow testing JabRef UI caught my attention and I forgot about this ticket. I added assertions and argued for keeping simple tests. Please proceed with this ticket as you choose.

src/test/java/org/jabref/preferences/VersionPreferencesTest.java
tobiasdiez: What @lynyus (probably) meant is the following: you usually write a test to verify a certain desired behavior and not to proof the absence of an undesired one.

To make this more concrete, consider the difference between the test method you added and
````java
public void isEmptyForEmptyVersion() {
     versionPreferences = new VersionPreferences(new Version());
     assertEqual(new Version(), versionPreferences.getIgnoredVersion());
}
````
Both tests would fail if the constructor of `VersionPreferences` throws an exception for some reason. However, this new version of the test also fails in case the constructor does not work as expected, namely if the version is not stored in the `ignoredVersion` field.

That being said, normally you don't need to write tests for simple constructors and properties. If you see this as a learning opportunity, you are of course welcome to write such tests.

lenhard:
@sammann-fnordbedarf-de If I read the comments correctly, then this is not yet in a mergable state, but you are the only one with the access rights to modify the branch since it is your fork. So unless you address the comments, we can only close it or copy your code to a new branch on which we can work.

@lynyus @Siedlerchr What is your opinion here? Just close or copy over to new branch?

Siedlerchr:
@sammann-fnordbedarf-de  Regarding mocking: Have a look at the Cleanup Test classes, e.g. RenamePDF or MoveFiles leanup. Another example where I had to use mocking intensive was recently in the AutoSetFileLinksUtilTest class

tobiasdiez:
But here no mocking is needed since the preference classes are actually the subject of the test. But since they do not involve `JabRefPreferences` there shouldn't be any side effects.

LinusDietz:
I am closing this now. We have spent much time with these "improvements" and discussions that I have the feeling that this PR is not going into a truly useful direction any time soon.

Issue #3229, opened 2017-09-20
Improve error handling when renaming a linked file


tobiasdiez:
JabRef 4.0-dev--snapshot--2017-09-20--master--b0489d5b3
Windows 10 10.0 amd64 
Java 1.8.0_131

Steps to reproduce:

1. Link a pdf to an entry and open the pdf.
2. Right click the file in the entry editor and choose "Rename file"
3. Rename fails since access is denied (which is ok), but no error message is shown to the user. The following exception is logged however.

<details>
<summary>Detail information:</summary>

```
Renaming Files failed
java.nio.file.FileSystemException: file.pdf -> new file name.pdf: The process cannot access the file because it is being used by another process.

	at sun.nio.fs.WindowsException.translateToIOException(Unknown Source)
	at sun.nio.fs.WindowsException.rethrowAsIOException(Unknown Source)
	at sun.nio.fs.WindowsFileCopy.move(Unknown Source)
	at sun.nio.fs.WindowsFileSystemProvider.move(Unknown Source)
	at java.nio.file.Files.move(Unknown Source)
	at org.jabref.logic.util.io.FileUtil.renameFile(FileUtil.java:195)
	at org.jabref.logic.cleanup.RenamePdfCleanup.cleanup(RenamePdfCleanup.java:121)
	at org.jabref.gui.fieldeditors.LinkedFileViewModel.performRenameWithConflictCheck(LinkedFileViewModel.java:198)
	at org.jabref.gui.fieldeditors.LinkedFileViewModel.rename(LinkedFileViewModel.java:186)
	at org.jabref.gui.fieldeditors.LinkedFilesEditor.lambda$createContextMenuForFile$7(LinkedFilesEditor.java:238)
	at com.sun.javafx.event.CompositeEventHandler.dispatchBubblingEvent(CompositeEventHandler.java:86)
	at com.sun.javafx.event.EventHandlerManager.dispatchBubblingEvent(EventHandlerManager.java:238)
	at com.sun.javafx.event.EventHandlerManager.dispatchBubblingEvent(EventHandlerManager.java:191)
	at com.sun.javafx.event.BasicEventDispatcher.dispatchEvent(BasicEventDispatcher.java:58)
	at com.sun.javafx.event.EventDispatchChainImpl.dispatchEvent(EventDispatchChainImpl.java:114)
	at com.sun.javafx.event.EventUtil.fireEventImpl(EventUtil.java:74)
[...]

PR #3233, opened 2017-09-24
Implemented issue #3229

derebaba:
Added an error dialog if the file is open in another process and cannot be renamed. [#3229]. This is my first contribution. Please give me feedback if you see any problems.


lenhard:
Hi @derebaba 

Thanks for your pull request! Your contribution is certainly very welcome and looks good overall.

I still have two points that should be addressed:
 - You add two new localization entries here. The way in which you use the localization strings in the code is fine, but you need to add the keys to the language files, otherwise the build will fail. To do so, you need to add the keys manually to the English file `jabref/src/main/resources/l10n/JabRef_en.properties` and then execute `./gradlew localizationUpdate` on the command line, which will synchronize the keys in all files. There's also a pretty good description of that when you run the tests and the localization test fails.
 - Apart from that, you are using the older file api `java.io` for doing the rename check. We try to get rid of that and try to use the new file api `java.nio` instead. I would expect something like `Files.isWritable()` to do the trick. Can you please adjust your solution to use the NIO API?

derebaba:
Hi @lenhard 

Thanks for your feedback. Are the keys ordered in any way in properties file? Or should I add new keys at the end?

Siedlerchr:
@derebaba  You should first check if there is already a key with the same/or similar text which you can reuse. Otherwise add it to the end and.

derebaba:
I synchronized the keys and removed the check which used the old api. However, `Files.isWritable()` returns true even if the file is open in another process. I searched a lot but I don't think the check is possible with the new api. Therefore, I moved the error dialog code to the catch block in FileUtil.java.

src/main/java/org/jabref/logic/util/io/FileUtil.java
```diff
@@ -198,6 +202,9 @@ public static boolean renameFile(Path fromFile, Path toFile, boolean replaceExis
                 return Files.move(fromFile, fromFile.resolveSibling(toFile)) != null;
             }
         } catch (IOException e) {
+            DIALOGSERVICE.showErrorDialogAndWait(
```
Siedlerchr: That is not a good idea, as you are now violating the architecture: https://github.com/JabRef/jabref/wiki/High-Level-Documentation
You are having gui components in a logic method. This FileUtil method is used in many places.
So I suggest you create a method which throws the exception

tobiasdiez: I think we need to help @derebaba here a bit (since architectural questions are probably not a good start into a project). 

It is also not possible to simply move the dialog service stuff to `RenamePdfCleanup` since this class also resides in `logic`. 
Thus I would propose the following:
1. Extract the body of the `FileUtil.renameFile` method to a new method `FileUtil.renameFileWithException`
https://github.com/JabRef/jabref/blob/c47d2a468a798c6ca83b8080e1301513717b1fff/src/main/java/org/jabref/logic/util/io/FileUtil.java#L194-L199
2. Similarly, extract the body of the cleanup method to a new method `cleanupWithException` (which then uses the new `renameFileWithException` method) https://github.com/JabRef/jabref/blob/8a420b4fcc4a97a2d8d42f4641d0061d22982b05/src/main/java/org/jabref/logic/cleanup/RenamePdfCleanup.java#L61
3. Catch the exception and show the error dialog in the gui:
https://github.com/JabRef/jabref/blob/c47d2a468a798c6ca83b8080e1301513717b1fff/src/main/java/org/jabref/gui/fieldeditors/LinkedFileViewModel.java#L206

It's a bit complicated but the best I can come up with (especially since the `FileUtil` and `CleanupJob` classes are designed to hide exceptions instead of to expose them). @JabRef/developers what do you think?

derebaba: Thanks for the comprehensive explanation. I got a little confused at first. Check out my latest commit. It should do the trick if I understand correctly.

tobiasdiez: Thanks for the quick implementation! The code is exactly how I meant it, good job!
The only problem left is the code duplication from the old methods and the new ones that throw an exception. But this should be easy to fix by calling the `withException` method in the original one; along the following lines:
````java
public static boolean renameFile(Path fromFile, Path toFile, boolean replaceExisting) {
        try {
            return renameFileWithException(fromFile, toFile, replaceExisting);
        } catch (IOException e) {
            LOGGER.error("Renaming Files failed", e);
            return false;
        }
}
````
Similarly for the `cleanup` method. If this is changed as well, I give my 👍 for merge.

derebaba: I removed the duplicate code as you explained. It is my first contribution ever to an open source project. Thank you for being so helpful.

PR #3238, opened 2017-09-27
Directory pattern bracketed expressions

sauliusg:
- [ x] Change in CHANGELOG.md described
- [ x] Tests created for changes
- [ x] Manually tested changed features in running JabRef

The offered PR implements functionality discussed in my "File linking functionality update?" proposal (http://discourse.jabref.org/t/file-linking-functionality-update/723). It is now possible to use bracketed expressions (like "[bibtexkey]" or "[year]\_[auth]\_[firstpage]") consistently in:

- BibTex key generator patterns;
- File search regular expressions (for automatic file linking);
- Import file name patterns;
- Import directory patterns.

To maintain consistent formatting, I have created a new BracketedExpressionExpander class which provides its main methods, 'expandBrackets()'. They essentially contain all bracket expansion functionality moved from BibtexKeyPatternUtil, and the BibtexKeyPatternUtil class now extends BracketedExpressionExpander, and uses its static methods. The BracketedExpressionExpander is also used in FileUtil and RegExpBasedFileFinder, avoiding the previous duplication of functionality in BibtexKeyPatternUtil.makeLabel(...) and RegExpBasedFileFinder.getFieldAndFormat(...) methods (both now use BracketedExpressionExpander.expandBrackets() for dealing with *all* aspects of bracket expansion in one place).

The bracketed expressions now can be used for Import File name patterns and Import File Directory patterns, providing consistent behavior of all "pattern" tabs in the JabRef "Preferences" dialogue; it is also possible to use underscores in the Import File and Directory patterns (e.g. "[year]\_[auth]\_[pages]") and to place Import destination directories deeper in the tree (e.g. "by-author/[authIni1]/[auth]" Directory Pattern now works as expected, creating "by-author/P/Pezzotti/2013\_Pezzotti\_211301.pdf" for the reference: Pezzotti, G. Raman spectroscopy of piezoelectrics Journal of Applied Physics, 2013, 113, 211301).

To retain functionality that was previously implemented by the '\entrytype' Layout pattern, a new bracketed expression "[entrytype]" was created and tested. It is suggested that Layout manager is used exclusively for describing layouts in citation lists, and bracketed expressions are used for describing file system and database entry patterns.

The functionality was tested under LinuxMint 18.1 and Ubuntu 16.04. Unfortunately, I could not test it on Windows hosts since I do not have any of them available for Java testing. I have put in provisions to deal with both '/' and '\' directory separators in Directory Patterns, but I would be grateful if someone checks them on a Windows system.

Sincerely,
Saulius


sauliusg:
I have fixed code according those Codacy remarks that were feasible. The rest is as follows:

- "Avoid really long classes: public class BracketedExpressionExpander {...":
-- actually, the class BracketedExpressionExpander is **shorter** than the BibtexKeyPatternUtil.java which it originated from, so the split in two classes was an overall improvement of this parameter ;) The rest would require separate refactoring, IMHO.

- "Avoid really long methods.", "The method generateInstitutionKey() has an NPath complexity of 10566908" and the friends:
-- These methods were ported from BibtexKeyPatternUtil.java; their refactoring would be a separate issue. Also, the code there is pretty straightforward (I had no problem to read an comprehend it), so teh Codacy complain might be too blunt.

- "Avoid unused private fields such as 'javafxPanel'.":
-- This is hack which I could not avoid; documented in comments. Any suggestions ow to improve that are welcome.

- "Avoid unused constructor parameters such as 'layoutPrefs'.":
-- These are left in place so that the old interface of MoveFilesCleanup and RenamePdfCleanup does not change; the old methods are marked as @Deprecated. If you approve that the LayoutFormatterPreferences are indeed no longer desired for the file maintenance functionality, I will remove them.

lynyus:
Wow, what a great PR! We are shortly before releasing JabRef 4.0 (hopefully this weekend), thus it will only be part of the first minor release after 4.0. Nevertheless, thanks for the efforts, and I'll start reviewing this soon!

.gitignore
```diff
@@ -1,3 +1,6 @@
+# Ignore local Make configuration files:
+Makeconf
```
lenhard: Out of curiosity: Did you really use make in some way while implementing this PR?

src/main/java/org/jabref/gui/externalfiles/DownloadExternalFile.java
```diff
@@ -310,8 +310,7 @@ private void downloadFinished() {
 
     private String getSuggestedFileName(String suffix) {
         String plannedName = FileUtil.createFileNameFromPattern(databaseContext.getDatabase(), entry,
-                Globals.prefs.get(JabRefPreferences.IMPORT_FILENAMEPATTERN),
-                Globals.prefs.getLayoutFormatterPreferences(Globals.journalAbbreviationLoader));
```
lenhard: Can you clarify some more why the `LayoutFormatterPreferences` are gone here? Doesn't this break existing formatting?

src/main/java/org/jabref/gui/externalfiles/DroppedFileHandler.java
```diff
@@ -33,7 +33,6 @@
 import org.jabref.gui.undo.UndoableFieldChange;
 import org.jabref.gui.undo.UndoableInsertEntry;
 import org.jabref.logic.l10n.Localization;
-import org.jabref.logic.layout.LayoutFormatterPreferences;
```
lenhard: Same as above: Why is it save to remove the `LayoutFormatterPreferences`?

src/main/java/org/jabref/gui/externalfiles/DroppedFileHandler.java
```diff
@@ -351,18 +350,19 @@ private boolean showLinkMoveCopyRenameDialog(String linkFileName, ExternalFileTy
         moveRadioButton.setText(Localization.lang("Move file to file directory"));
         renameCheckBox.setText(Localization.lang("Rename file to").concat(": "));
 
-        LayoutFormatterPreferences layoutPrefs = Globals.prefs
-                .getLayoutFormatterPreferences(Globals.journalAbbreviationLoader);
+        // FIXME: Remove unused code later S.G.
```
lenhard: Please remove the commented-out code here.

src/main/java/org/jabref/logic/bibtexkeypattern/BibtexKeyPatternUtil.java
```diff
@@ -33,335 +18,15 @@
 /**
  * This is the utility class of the LabelPattern package.
  */
-public class BibtexKeyPatternUtil {
+public class BibtexKeyPatternUtil extends BracketedExpressionExpander {
```
lenhard: I think it's a really good idea to extract some code here into a separate class. Good job!

src/main/java/org/jabref/logic/cleanup/MoveFilesCleanup.java
```diff
@@ -31,17 +31,21 @@
     private final BibDatabaseContext databaseContext;
     private final FileDirectoryPreferences fileDirectoryPreferences;
 
-    private final LayoutFormatterPreferences layoutPrefs;
+    // FIXME: remove unused variables later S.G.
```
lenhard: Please remove the commented-out code here.

src/main/java/org/jabref/logic/cleanup/MoveFilesCleanup.java
```diff
@@ -31,17 +31,21 @@
...
+    // private final LayoutFormatterPreferences layoutPrefs;
     private final String fileDirPattern;
 
     private LinkedFile singleFileFieldCleanup;
 
+    // FIXME: remove the unused 'layoutPrefs' parameter later (S.G.); so far, however, it is left to
```
lenhard: Please remove the commented-out code here.

src/main/java/org/jabref/logic/cleanup/MoveFilesCleanup.java
```diff
@@ -31,17 +31,21 @@
...
+    // prevent constructor signature change.
     public MoveFilesCleanup(BibDatabaseContext databaseContext, String fileDirPattern,
             FileDirectoryPreferences fileDirectoryPreferences, LayoutFormatterPreferences layoutPrefs) {
         this.databaseContext = Objects.requireNonNull(databaseContext);
         this.fileDirPattern = Objects.requireNonNull(fileDirPattern);
         this.fileDirectoryPreferences = Objects.requireNonNull(fileDirectoryPreferences);
-        this.layoutPrefs = Objects.requireNonNull(layoutPrefs);
+        // FIXME: Remove unused code later S.G.
```
lenhard: Please remove the commented-out code here.

src/main/java/org/jabref/logic/cleanup/RenamePdfCleanup.java
```diff
@@ -33,18 +33,22 @@
     private final BibDatabaseContext databaseContext;
     private final boolean onlyRelativePaths;
     private final String fileNamePattern;
-    private final LayoutFormatterPreferences layoutPreferences;
+    // FIXME: remove unused class member later (S.G.):
```
lenhard: Please remove the commented-out code here.

src/main/java/org/jabref/logic/cleanup/RenamePdfCleanup.java
```diff
@@ -33,18 +33,22 @@
...
+    // private final LayoutFormatterPreferences layoutPreferences;
     private final FileDirectoryPreferences fileDirectoryPreferences;
     private int unsuccessfulRenames;
     private LinkedFile singleFieldCleanup;
 
+    // FIXME: remove unused constructor argument 'layoutPreferences' later; for now,
```
lenhard: Please remove the commented-out code here.

src/main/java/org/jabref/logic/cleanup/RenamePdfCleanup.java
```diff
@@ -33,18 +33,22 @@
...
+    // however, the argument is retained in order not to change the class interface (S.G.):
     public RenamePdfCleanup(boolean onlyRelativePaths, BibDatabaseContext databaseContext, String fileNamePattern,
                             LayoutFormatterPreferences layoutPreferences,
                             FileDirectoryPreferences fileDirectoryPreferences) {
         this.databaseContext = Objects.requireNonNull(databaseContext);
         this.onlyRelativePaths = onlyRelativePaths;
         this.fileNamePattern = Objects.requireNonNull(fileNamePattern);
-        this.layoutPreferences = Objects.requireNonNull(layoutPreferences);
+        // FIXME: remove unused code later (S.G.):
```
lenhard: Please remove the commented-out code here.

src/main/java/org/jabref/logic/util/io/FileNameCleaner.java
```diff
@@ -18,18 +18,16 @@
             10, 11, 12, 13, 14, 15, 16, 17, 18, 19,
             20, 21, 22, 23, 24, 25, 26, 27, 28, 29,
             30, 31, 34,
-            42, 47,
```
lenhard: Why did you remove 47 here? Why is it no longer an illegal char? This needs a good justification or should be reverted.

src/main/java/org/jabref/logic/util/io/FileNameCleaner.java
```diff
@@ -18,18 +18,16 @@
...
+            42,
             58,
             60, 62, 63,
-            92,
```
lenhard: Same as for 47 above.

src/main/java/org/jabref/logic/util/io/RegExpBasedFileFinder.java
```diff
@@ -57,58 +56,14 @@ public static String expandBrackets(String bracketString, BibEntry entry, BibDat
         Matcher m = SQUARE_BRACKETS_PATTERN.matcher(bracketString);
         StringBuffer s = new StringBuffer();
         while (m.find()) {
-            String replacement = getFieldAndFormat(m.group(), entry, database, keywordDelimiter);
+            String replacement = BracketedExpressionExpander.expandBrackets(m.group(), keywordDelimiter, entry, database);
```
lenhard: Since you are modifying this piece of code anyway, could you improve it a little and give `m` a more meaningful name, e.g. `matcher`.

src/main/java/org/jabref/logic/util/BracketedExpressionExpander.java
```diff
@@ -0,0 +1,1356 @@
...
+import org.jabref.model.entry.FieldName;
+import org.jabref.model.entry.Keyword;
+import org.jabref.model.entry.KeywordList;
+import org.jabref.model.strings.StringUtil;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+/**
+ * @author saulius
```
lenhard: We're trying to get rid of all the `@author` annotations, so please remove this. Thanks to git, this code will be attributed to you anyway via your commit. No need to write it into the source.

src/test/java/org/jabref/logic/util/BracketedExpressionExpanderTest.java
```diff
@@ -0,0 +1,153 @@
...
+import org.jabref.model.entry.BibtexEntryTypes;
+import org.jabref.model.entry.BibtexString;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+public class BracketedExpressionExpanderTest {
+    // The javafxPanel instace is only created to suppress the ugly
```
lenhard: That is indeed quite ugly. As far as I can see this comes from the using the logger. It's a little weird, because I am almost sure that the logger didn't trigger these exceptions earlier, but this is probably not caused by this PR.

I think I would prefer it if you removed the JFXPanel here. It is better to have exceptions written into the console (the tests work anyway) than to have dependencies such as this in the code. But maybe another developer might disagree with me here.

src/main/java/org/jabref/logic/util/BracketedExpressionExpander.java
```diff
@@ -0,0 +1,1356 @@
...
+    private final BibEntry bibentry;
+
+    public BracketedExpressionExpander() {
+        this.bibentry = null;
+    }
+
+    /**
+     * @param bibentry
+     */
+    public BracketedExpressionExpander(BibEntry bibentry) {
```
tobiasdiez: I'm not a big fan of the semantics here. Instead I would propose that `BracketedExpressionExpander` actually accepts the pattern as a constructor argument and then provides a method `expand` that accepts a BibEntry as an argument and uses the contained information to expand the pattern. My proposal is a bit similar to the regular expression pattern class.
Moreover, I would then rename the `BracketedExpressionExpander` class to something like `GenericPattern` or `BacketedPattern` (actually I would prefer just `Pattern` but this name is already taken by the java regex pattern class).

What do you think?

src/main/java/org/jabref/logic/util/io/FileUtil.java
```diff
@@ -279,6 +283,50 @@ public static String createFileNameFromPattern(BibDatabase database, BibEntry en
...
 
     /**
+     * Determines filename provided by an entry in a database
+     *
+     * @param database        the database, where the entry is located
+     * @param entry           the entry to which the file should be linked to
+     * @param fileNamePattern the filename pattern
+     * @return a suggested fileName
+     */
+    public static String createFileNameFromPattern(BibDatabase database, BibEntry entry, String fileNamePattern) {
```
tobiasdiez: We are trying to get rid of these static helper methods. Similar to my remark above, I would propose to create a `FileNamePattern` class, that accepts the file name pattern and directory pattern as constructor arguments and provides these two methods as instance methods.

sauliusg:
On 2017-09-29 12:02, Jörg Lenhard wrote:
> In .gitignore
> <https://github.com/JabRef/jabref/pull/3238#discussion_r141811777>:
> 
>> @@ -1,3 +1,6 @@
> +# Ignore local Make configuration files:
> +Makeconf
> 
> Out of curiosity: Did you really use make in some way while implementing
> this PR?

Yes :). GNU Make is my favourite tool since decades, and sometimes its
just faster for me to get things running with it. So I made myself a
small Makefile with 'make run/make test' targets for CLI-testing the
classes I wrote. Also, Eclipse/Gradle are *sooo* RAM-hungry... my 4GB
laptop swaps like hell :) when running them.

As a side note, we use GNU Make for testing our Unix/Linux CLI tools
extensively, and for organising computations. But for JabRef, I guess,
there is no sense to have a also Make-based tester, your 3000+ JUnit
tests are the impressive lot and should do the job. Therefore I removed
the Makefile from my final pull request; I can remove the ignore
patterns too if you want :)

More one the PR code and the comments in a short while...

Regards,
Saulius

-- 
Dr. Saulius Gražulis
Vilnius University Institute of Biotechnology, Saulėtekio al. 7
LT-10257 Vilnius, Lietuva (Lithuania)
fax: (+370-5)-2234367 / phone (office): (+370-5)-2234353
mobile: (+370-684)-49802, (+370-614)-36366

sauliusg:
On 2017-09-29 12:02, Jörg Lenhard wrote:
>   * Please clarify a little more about what happened to the
>     |LayoutPreferences| and why it is safe to remove them at some
>     points. Does this not break current behavior? Maybe you've already
>     explained that, but please bear with me and do it again.

Sure; I did not explain this point fully and will elaborate below.

The short answer would be: in the code before the suggested PR, file
name patterns were handled by a LayoutHelper, and that class has
LayoutFormatterPreferences argument in its constructor. The suggested PR
code uses BracketedExpressionExpander (or BacketedPattern as Tobias
suggest to rename it), which is made by splitting the
BibTexKeyPatternUtil into two classes. The BibTexKeyPatternUtil did not
require LayoutFormatterPreferences, and neither does
BracketedExpressionExpander.

The longer discussion would be as follows: in the current master (since
I guess JabRef 4.0) RegExpBasedFileFinder uses BibtexKeyPatternUtil to
generate regexps for autolinking files, but the Import file pattern uses
LayoutHelper to generate imported file names. Autodownloading was one of
the main reasons for us to switch to JabRef 4.0, but we used
[year]_[auth]_[firstpage].*\\.pdf pattern for file autolinker (with
>4000 files in a tree), and the downloader offered only the "\bibtexkey"
and "\bibtexkey - \begin{title}\title\end{title}" patterns; so the
downloaded files were not found by autolinker :(.

I have tried the "\year_\auth_\firstpage" as a download pattern, but it
turned out that underscore (and hyphen) were regarded as letters, so my
pattern did not work, and I could not find a way to escape the
underscore ("\year\_\auth\_\firstpage" did not work either...). I
confirmed that removing '_' and '-' from the LayoutHelper did allow me
to use the "\year_\bibtexkey_\pages" pattern, after the change:

src/main/java/org/jabref/logic/layout/LayoutHelper.java:
-      if (!Character.isLetter((char) c) && (c != '_') && (c != '-')) {
+      if (!Character.isLetter((char) c)) {

(see also my comments at
https://github.com/sauliusg/jabref/commit/477db399a3d0b888d557127850c216533e0e62c0#diff-2d48ba893857485e0758e284be6ba290)

However, such change could potentially break the functionality of
layouts for generating reference lists, and that would be disastrous. So
I decided not to pursue this line. Instead, I wrote a discussion page
offering to unify pattern syntax for bibtexkeys, file linking regexps,
import file directories and import file directories. Indeed, anyone who
sees "Import|Filename format pattern" in "Preferences" will immediately
assume that the same pattern syntax as in "BibTeX key generator|Default
pattern" is used, and will be surprised if that does not work.

Moreover, at the time of branching "Filename format pattern" worked
reasonably only with the two provided default patterns, and "File
directory pattern" seemed to ignore patterns altogether; thus, I think
few people will be using the layout-style ("\bibtexkey - \title")
patterns, and changing syntax to the equivalent "[bibtexkey] - [title]"
would not hurt much, especially if we implement, as you propose, a
PreferenceMigration tool. The benefits of using the same language for
all patterns are obvious and are acknowledged by Tobias as well :).

Programmatically, everything fits together well -- all interfaces match
and all tests pass. The end user, of course, will see the difference in
behaviour. On one hand, there is an improvement: one can use the same
"[auth][year]" pattern or similar in all Preference fields; one can also
use similar patterns for directories, like "PDF/[year]/[auth]" and get
the files nicely organised on the spot.

On the other hand, the "bracketed expressions" are at the moment weaker
than layouts. For example, there is no conditional expansion, so there
is no equivalent of the "\begin{title} - \title\end{title}" construct
(where "-" is only inserted if the title is present). For me, this does
not seem to be a show stopper, much less so than discrepancies in
bibtexkey and import file expressions. I see however that some people
might want this functionality.

I would therefore suggest the following strategy: one could integrate
the current bracketed expressions as they are now, and I promise write
extensions of these expressions which support conditional expansion and
nested expressions, e.g.:
[title? - ] would expand to " - " only if title is present, and the
"\bibtexkey\begin{title} - \title\end{title}" would become
"[bibtexkey][title? - ][title]" (quite readable to my taste...). Another
obvious expansions would be:

[author?auth][!author&&editor?ed. [editor]] and so on.

Nested booleans and nested bracketed expressions would be supported; I
would suggest a simple recursive descent parser for this (a similar
thing is already started in the LayoutHelper :).

I could offer a PR with this functionality, and you have a look and if
you decide to go for it, JabRef would have full expression support in
both BibTeX keys, file patterns and directory patterns.

I hope this clarifies the PR, and sorry for the long text...

>   *There's some commented out code that you should remove.

Will do that.

>   * You are changing preference-related stuff here, e.g. moving from
>     |\entrytype| to |[entrytype]| in the CleanupPreferences and more.
>     From what I can see that means that you should write a preference
>     migration that changes older preferences a user might have to your
>     new format automatically. Please have a look at
>     |org.jabref.migrations.PreferencesMigration| and implement a
>     migration for your changes.

This is an important point which I have overlooked! Thanks for the hint,
I'll look at the class and implement a migration. I'll come back to you
if I have questions regarding migration strategies.

Regards,
Saulius

-- 
Dr. Saulius Gražulis
Vilnius University Institute of Biotechnology, Saulėtekio al. 7
LT-10257 Vilnius, Lietuva (Lithuania)
fax: (+370-5)-2234367 / phone (office): (+370-5)-2234353
mobile: (+370-684)-49802, (+370-614)-36366

sauliusg:
Hi,

On 2017-09-29 12:02, Jörg Lenhard wrote:
> In src/main/java/org/jabref/logic/util/io/FileNameCleaner.java
> <https://github.com/JabRef/jabref/pull/3238#discussion_r141814168>:
> 
>> @@ -18,18 +18,16 @@
>              10, 11, 12, 13, 14, 15, 16, 17, 18, 19,
>              20, 21, 22, 23, 24, 25, 26, 27, 28, 29,
>              30, 31, 34,
> -            42, 47,
> 
> Why did you remove 47 here? Why is it no longer an illegal char? This
> needs a good justification or should be reverted.
> 
> ------------------------------------------------------------------------
> 
> In src/main/java/org/jabref/logic/util/io/FileNameCleaner.java
> <https://github.com/JabRef/jabref/pull/3238#discussion_r141814198>:
> 
>>              58,
>              60, 62, 63,
> -            92,
> 
> Same as for 47 above.
> 
> ------------------------------------------------------------------------

The removed codes are for the ASCII "/" and "\" chars, the directory
name separators for Unix-like systems and for Windows.

Before, they were considered illegal chars everywhere, and directory
patterns like "PDF/[auth]" would yield directories "PDF_Author", clearly
not a desired result.

I have thus introduced another function, cleanDirectoryName(), that uses
the shortened table ("/" and "\" chars are now legal in directory
names). In file names, the "/" and "\" chars are illegal, and
cleanFileName() checks them explicitly in addition to the table chars.

The cleanDirectoryName() is used to check directory names in file mover
cleanup; cleanFileName() is used, as before, to check PDF file names
after renaming. The behaviour of cleanFileName() should not have changed
(regression tests pass), and the cleanDirectoryName() is used only in
one place in the code, to check directory names of the moved files (an
there are two tests to check its functionality).

I opted to have one common table and two functions to preserve SPOT
(single point of truth) for the character list, which is extensive and
in addition needs special maintenance (needs to be sorted). Actually, we
should have test for that, shouldn't we?

Hope this is OK. I'll comment this in the source.

Regards,
Saulius

-- 
Dr. Saulius Gražulis
Vilnius University Institute of Biotechnology, Saulėtekio al. 7
LT-10257 Vilnius, Lietuva (Lithuania)
fax: (+370-5)-2234367 / phone (office): (+370-5)-2234353
mobile: (+370-684)-49802, (+370-614)-36366

src/main/java/org/jabref/logic/util/BracketedExpressionExpander.java
sauliusg: I bet this is the "user friendly" Eclipse that did it :). Removed. Will push update shortly.

src/main/java/org/jabref/logic/util/io/RegExpBasedFileFinder.java
sauliusg: Fixed: m -> matcher, s -> expandedStringBuffer

.gitignore
sauliusg: Yes, GNU Make is my favourite project management tool :)
But, agreed, not needed for JabRef. Removed the .gitignore lines.

src/main/java/org/jabref/logic/bibtexkeypattern/BibtexKeyPatternUtil.java
sauliusg: I agree regarding the separation. I'm glad you like the implementation; some people say that "extends is evil"....

src/main/java/org/jabref/gui/externalfiles/DroppedFileHandler.java
sauliusg: Removed; will push in a short while.

src/main/java/org/jabref/gui/externalfiles/DownloadExternalFile.java
sauliusg: See my e-mail reply below.

src/main/java/org/jabref/gui/externalfiles/DroppedFileHandler.java
sauliusg: See my e-mail reply below.

src/main/java/org/jabref/logic/cleanup/MoveFilesCleanup.java
sauliusg: Fixed.

src/main/java/org/jabref/logic/cleanup/MoveFilesCleanup.java
sauliusg: Fixed.

src/main/java/org/jabref/logic/cleanup/RenamePdfCleanup.java
sauliusg: Fixed.

src/main/java/org/jabref/logic/cleanup/RenamePdfCleanup.java
sauliusg: Fixed.

src/main/java/org/jabref/logic/cleanup/RenamePdfCleanup.java
sauliusg: Fixed; although, I have still left a FIXME regarding the unused 'layoutPreferences' constructor argument; I'll remove it when we all agree that it is safe to drop dependency on the LayoutFormatterPreferences class. The latter change will require more refactoring.

src/main/java/org/jabref/logic/util/io/FileNameCleaner.java
sauliusg: See comment in my e-mail below.

src/main/java/org/jabref/logic/util/io/FileNameCleaner.java
sauliusg: See comment in my e-mail below.

src/test/java/org/jabref/logic/util/BracketedExpressionExpanderTest.java
sauliusg: Removed JFXPanel, but added FIXMEs to the tests that trigger the exception logging -- we should find out what goes on there. The exceptions can not be caught in the test case code (!), apparently meaning that they are thrown and logged in a separate thread (JUnut?).

Only two test cases, unbalancedBracketExpansionTest and unbalancedLastBracketExpansionTest, cause this behaviour, since only they use the logger to report syntax errors. Another solution would be to drop logger dependency from BracketedExpressionExpander and throw exceptions on sytax errors, expecting that the caller catches and logs them appropriately. I would like this approach more, but currently there is a risk that uncaught exceptions will crash the application or produce obscure notifications.

src/main/java/org/jabref/logic/util/BracketedExpressionExpander.java
sauliusg: Thaks for the renaming suggestion! I agree, 'BracketedExpressionExpander' is a long and clumsy name. I would opt for 'BacketedPattern': 'GenericPattern', I think, is too generic ;), and Pattern is taken, as you say. I've refactored the class and tests to use 'BacketedPattern' and 'BacketedPattern.expand' instead of 'BacketedPattern.exoandBrackets'.

sauliusg: As for semantics, I thing one can do both ways. I assumed that the the common situation will be to have a pattern that we want to expand, and then have an (abstract) expander that contains all necessary information (bibentry, preferenceses, databases, etc.). I also assumed that one may want to use the same expander for several patterns in a row; this last assumption does not hold for JabRef.

So, I could change the interface to public BracketedPattern(String pattern, BibEntry bibentry), or public BracketedPattern(String pattern).

I must admit however that to my taste it seems a terrible abuse of OO to make an object as a fancy way of passing another parameter to a function; i.e. having 'X x = new X(y); x.f(z)' instead of simply 'f(y,z)'. Functions should stay functions, and object should be "clever" values used in those functions, IMHO :).

In the above case, 'expandBrackets' seems to be more like a function that takes two arguments, a pattern (which is always a string) and the context information (which may be BibEntry, Database, DatabaseContext, depending on your needs, and may change in the future). Having BracketedPattern.expand( String pattern ) would allow flexibility in the future: one could use descendent classes of BracketedPattern (with different context information) everywhere where the '.expand( bracketedPattern )' method is called, without changing and interface. Or one might have an AbstractBracketedPattern interface/abstract class, and implement '.expand( bracketedPattern )' in different ways. 

Given the above considerations, the current interface, 'public BracketedPattern(BibEntry bibentry)' and 'String BracketedPattern.expand(String pattern)' seem the most flexible and future-proof.

What are your thoughts?

src/main/java/org/jabref/logic/util/io/FileUtil.java
sauliusg: The 'static String createFileNameFromPattern(...)' method was given, and I was trying to change as little as possible, so opted to make a static ''

src/main/java/org/jabref/logic/util/BracketedExpressionExpander.java
tobiasdiez: You are right,
>'expandBrackets' seems to be more like a function that takes two arguments, a pattern (which is always a string) and the context information (which may be BibEntry, Database, DatabaseContext, depending on your needs, and may change in the future)

So in principle there are two logical ways to split these arguments to the `expandBrackets` function:
a) The pattern as an argument to the constructor and the context information as arguments to the `expandBrackets` method.
b) The other way around, that is the context information as arguments to the constructor and the pattern as an argument to the `expandBrackets` method.

You favor method b) while I would use a). There are good arguments for both ways. Lets make a deal: I'll present my reasoning for a) and if I can convince you, you change the code accordingly. If not then the code stays like it is right now and I give my +1 for merge 😸 

So:
- In principle, we can do a pre-processing of the pattern in the constructor and then reuse the `expand` method for different entries to save time.
- You can create the `BracketedPattern` from the string in some place (say in the preference class) and then pass it to other methods (for example, the rename cleanup operation). Then only the preference class needs to know how to deserialize the pattern from a string, while the other methods in the chain only care that the are given a `BracketedPattern`. (Background: there was a time, when strings where used for everything in JabRef: File names, entry field names and values, preferences, ... literally everywhere. Since about 2 year we are now trying to encapsulate more and more of these strings in special classes, e. g. `AuthorList` or dedicated preference classes like `LayoutPreferences`. I would like to see the same for the pattern.)
- If we should decide to change or replace the pattern format, one would only need to change the places where the conversion from string to the pattern happens (which is considerably less overhead then all the places where the pattern is expanded). 
- If a class is named `BracketedPattern` I would assume that it encapsulates a pattern. 😈

sauliusg:
Hi,

On 2017-09-30 11:44, Tobias Diez wrote:
> You favor method b) while I would use a). There are good arguments for
> both ways. Lets make a deal: I'll present my reasoning for a) and if I
> can convince you, you change the code accordingly. If not then the code
> stays like it is right now and I give my +1 for merge 😸

Agreed :)

>   * In principle, we can do a pre-processing of the pattern in the
>     constructor and then reuse the |expand| method for different entries
>     to save time.

Pre-processing is optimisation, and our interface design decisions
should not crucially depend on (hypotetical) optimisations, I would say.
Code readability, IMHO, is more important.

Also, you can do the proposed optimisation also with the current
interface, say by using memoisation.

And finally, I can turn your argument on its head an day that I can do
reprocessing with the BibEntry to save processing time with different
patterns (and then apply my own counter-argument to it (does it get
recursive at this stage? :))

So I think No1 is a weak argument, but:

>   * You can create the |BracketedPattern| from the string in some place
>     (say in the preference class) and then pass it to other methods (for
>     example, the rename cleanup operation). Then only the preference
>     class needs to know how to deserialize the pattern from a string,
>     while the other methods in the chain only care that the are given a
>     |BracketedPattern|. (Background: there was a time, when strings
>     where used for everything in JabRef: File names, entry field names
>     and values, preferences, ... literally everywhere. Since about 2
>     year we are now trying to encapsulate more and more of these strings
>     in special classes, e. g. |AuthorList| or dedicated preference
>     classes like |LayoutPreferences|. I would like to see the same for
>     the pattern.)

This I buy, I think its a serious consideration.

>   * If we should decide to change or replace the pattern format, one
>     would only need to change the places where the conversion from
>     string to the pattern happens (which is considerably less overhead
>     then all the places where the pattern is expanded).

Agreed, this is an argument in favour of abstraction of patterns from
plain strings.

>   * If a class is named |BracketedPattern| I would assume that it
>     encapsulates a pattern. 😈

:) Sure, but this is the name you suggested in the first place :) Mine
was ...Expander, you remember? Anyway, the names should go after design,
not the other way round, shouldn't they?

So, to sum up, I think arguments 2) and 3) (1-based) are really in
favour of your suggestion. So I implement 'public
BracketedPattern(String Pattern)' and '...expand(bibentry)'.

Regards,
Saulius

-- 
Dr. Saulius Gražulis
Vilnius University Institute of Biotechnology, Saulėtekio al. 7
LT-10257 Vilnius, Lietuva (Lithuania)
fax: (+370-5)-2234367 / phone (office): (+370-5)-2234353
mobile: (+370-684)-49802, (+370-614)-36366

sauliusg:
Hi,

On 2017-09-30 11:44, Tobias Diez wrote:
> a) The pattern as an argument to the constructor and the context
> information as arguments to the |expandBrackets| method.
> b) The other way around, that is the context information as arguments to
> the constructor and the pattern as an argument to the |expandBrackets|
> method.

I have implemented the (a) variant of the BracketedPattern interface.

Please note, however:

- I had to implement a default constructor, since BibtexKeyPatternUtil
extend BracketedPattern and BibtexKeyPatternUtil so far only has a
default constructor; I do not want to refactor even further the use of
BibtexKeyPatternUtil -- let's leave it for the next PR if needed :)

This means that at the moment there is a possibility to construct a
BibtexKeyPatternUtil and BracketedPattern instance without a pattern,
and this will later trigger a NullPointerException if used indiscriminately.

- For the same reason I have left the expandBrackets(...) static method
– it is used by static methods in BibtexKeyPatternUtil and possibly
elsewhere. Again, if you want to get rid of those, more refactoring and
testing would necessary.

At the moment, however, all functionality that was intended for this PR
works, is tested, adheres to JabRef coding standards (I hope :).

Regards,
Saulius

-- 
Dr. Saulius Gražulis
Vilnius University Institute of Biotechnology, Saulėtekio al. 7
LT-10257 Vilnius, Lietuva (Lithuania)
fax: (+370-5)-2234367 / phone (office): (+370-5)-2234353
mobile: (+370-684)-49802, (+370-614)-36366

lynyus:
okay, this looks good to be tested with real users. since we release 4.0 today, I would kindly ask you to fix the conflict in the changelog, then I would merge this.

Issue #3143, opened 2017-08-22
Import entry from clipboard


buhtz:
Some websites offer bibliographic data in machinereadable format. e.g. a Bibtex oder RIS text. But there is no download link. The text is just displayed so you can copy and paste it.

JabRef should be able to read from clipboard e.g. the content of a RIS file.
```
TY  - BOOK
ED  - Kempf, Ute
ED  - Wrede, Birgitta
ID  - 2912032
SN  - ISBN 978-3-932869-19-8
TI  - Gender-Effekte. Wie Frauen die Technik von morgen gestalten
VL  - Band 19
ER  - 
```

This would prevent me to open a text editor, paste the text in there, store it as a RIS file, reopen it with jabref, ... ;)

lenhard:
We sort of have that already. 

Try in the menu: BibTeX -> New entry from plain text -> select the type -> paste the text.

Afaik, it only works for BibTeX and the workflow could certainly be optimized, though.

buhtz:
Mhm... I see. But it could be "better" - IMO.

1. Internaly all import formats supported by JabRef should be used by the "entry from plain text" feature, too.
2. And the access to the clipboard feature should be in the "import into new database"/"import into the current database" area. I am not sure how this could be solved in the UI because the "import" currently opens a file dialog.

stefan-kolb:
Just as a side note: pasting entries works as long as they are in BibTeX format.

koppor:
@buhtz Can you imagine opening a pull request for that? We have [code howtos](https://github.com/JabRef/jabref/wiki/Code-Howtos) and [guidelines for setting up a local workspace](https://github.com/JabRef/jabref/wiki/Guidelines-for-setting-up-a-local-workspace). I think, your feature request is "just" iterating over all available EntryFetchers and checking if the fetcher accepts the format.

buhtz:
@kopper Sorry I am far away from Java and out of time ressource. Just keep this Issue open to document the enhancement. Of course I understand your ressource situation also.

lenhard:
This is fixed by #3243

PR #3243, opened 2017-10-01
Fix for issue 3143: Import entry from clipboard in different formats

125m125:
see #3143
Use all importers to parse pasted entries. This allows the user to paste entries in different formats and not ony BibTex.


lynyus:
Hi @125m125, as I see you have some troubles with our checkstyle rules. I recommend you to use our IDE config files from https://github.com/JabRef/jabref/tree/master/config. Which IDE are you using? We recommend IntelliJ IDEA or Eclipse.
You can see if the tests a pass by calling `gradle check`. If you only want to check for the code style rules, you can call `gradle checkstyleMain`.

If you have any questions, feel free to ask them at slack: https://jabref.slack.com

125m125:
I am using Eclipse and just had to format a file. Because the saveAction is set to only reformat edited lines a missing newline in the imports wasn't corrected. Also thanks for pointing out gradle check. I somehow missed that command.

Siedlerchr:
@125m125  If you use eclipse, just run` gradlew eclipse `to import the correct code formatting rules

src/main/java/org/jabref/logic/importer/ImportFormatReader.java
```diff
@@ -215,4 +218,39 @@ public UnknownFormatImport importUnknownFormat(Path filePath) throws ImportExcep
 
         throw new ImportException(Localization.lang("Could not find a suitable import format."));
     }
+
+    public UnknownFormatImport importUnknownFormatFromString(String data) throws ImportException {
```
tobiasdiez: As far as I can see, this is mostly a copy of `importUnknownFormat(Path filePath)` and I would like to have a solution without code duplication. Maybe something along the following lines works?
````java
private UnknownFormatImport importUnknownFormat(Function<Importer, ParserResult> importDatabase, Function<Importer, Boolean> isRecognizedFormat) {
    // original method with obvious changes imFo.importDatabase -> importDatabase.apply(imFo)
}

public UnknownFormatImport importUnknownFormat(Path filePath) {
    return importUnknownFormat(importer -> importer.importDatabase(filePath, importFormatPreferences.getEncoding(), importer -> importer .isRecognizedFormat(filePath, importFormatPreferences.getEncoding());
}

public UnknownFormatImport importUnknownFormat(String data) {
    return importUnknownFormat(importer -> importer.importDatabase(data), importer -> importer .isRecognizedFormat(data));
}
````

File: src/main/java/org/jabref/logic/importer/ImportFormatReader.java [hunk 2]
```diff
@@ -215,4 +218,39 @@ public UnknownFormatImport importUnknownFormat(Path filePath) throws ImportExcep
...
+        // stores ref to best result, gets updated at the next loop
+        List<BibEntry> bestResult = null;
+        long bestResultCount = 0;
+        String bestFormatName = null;
+
+        // Cycle through all importers:
+        for (Importer imFo : getImportFormats()) {
+            try (StringReader in = new StringReader(data); BufferedReader input = new BufferedReader(in)) {
+                ParserResult parserResult = imFo.importDatabase(input);
```
tobiasdiez: Can you please extract the method `Importer.importDatabase(String data)`. This new method can be reused at least in the `MrDLibFetcher`. Similarly, for the `isRecognizedFormat`.

src/test/java/org/jabref/logic/importer/ImportFormatReaderIntegrationTest.java
```diff
@@ -55,6 +55,20 @@ public void testImportFormatFromFile() throws Exception {
...
     }
 
+    @Test
+    public void testImportUnknownFormatFromString() throws Exception {
+        String data = new String(Files.readAllBytes(file), StandardCharsets.UTF_8);
+        try {
+            assertEquals(count, reader.importUnknownFormatFromString(data).parserResult.getDatabase().getEntries().size());
+        } catch (ImportException e) {
+            // msbib test file does not contain an author
+            if ("msbib".equals(format)) {
```
tobiasdiez: Why is this special treatment necessary for the import from a string, but not when the same content is imported from a file?

125m125:
Thanks for the feedback. I integrated the requests with the last commit.
I also changed the PdfXmpImporter to return false when checking the format instead of throwing an exception to prevent the need for special treatment and make it consistent with the PdfContentImporter.

lenhard:
Ok, then I'll add this to the 4.1 milestone.

tobiasdiez:
I think the error message comes from
https://github.com/JabRef/jabref/blob/ae6d01be46c96ef57e9871dfeb208ab0b7aab295/src/main/java/org/jabref/logic/importer/fileformat/MrDLibImporter.java#L53-L60
In my opinion, you can just delete these `Logger.log` statements (since its quite expected that a file in a wrong format is passed to `isRecognizedFormat` and this importer is not used outside of internal code anyways).

PR #3273, opened 2017-10-06
Set WM_CLASS to org-jabref-JabRefMain

michaellass:
WM_CLASS is used by certain Un*x window managers, such as GNOME, to map windows
to the corresponding application. Currently WM_CLASS is set to java-lang-Thread
which is very generic and not unique to JabRef. Instead, use reflection to set
it to org-jabref-JabRefMain which resembles the main class name.

The implementation in this PR is based on this blog post:
http://elliotth.blogspot.de/2007/02/fixing-wmclass-for-your-java.html
By now you find similar solutions in several Java applications. I'm not aware of a more convenient way to manually set WM_CLASS.

I tested the change on a Linux system running GNOME 3 and also on macOS to rule out any cross-platform issues. Also I chose the WM_CLASS name that matches the StartupWMClass in [buildres/snapcraft/jabref.desktop](../blob/master/buildres/snapcraft/jabref.desktop), so no changes are necessary in there.

### DIFF
--- a/src/main/java/org/jabref/gui/GUIGlobals.java
+++ b/src/main/java/org/jabref/gui/GUIGlobals.java
@@ -2,6 +2,7 @@ package org.jabref.gui;
 
 import java.awt.Color;
 import java.awt.Font;
+import java.awt.Toolkit;
 import java.util.HashMap;
 import java.util.Map;
 
@@ -13,6 +14,7 @@ import org.jabref.gui.externalfiletype.ExternalFileTypes;
 import org.jabref.gui.keyboard.EmacsKeyBindings;
 import org.jabref.gui.specialfields.SpecialFieldViewModel;
 import org.jabref.logic.l10n.Localization;
+import org.jabref.logic.util.OS;
 import org.jabref.model.entry.FieldName;
 import org.jabref.model.entry.specialfields.SpecialField;
 import org.jabref.preferences.JabRefPreferences;
@@ -155,6 +157,18 @@ public class GUIGlobals {
 
         GUIGlobals.currentFont = new Font(Globals.prefs.get(JabRefPreferences.FONT_FAMILY),
                 Globals.prefs.getInt(JabRefPreferences.FONT_STYLE), Globals.prefs.getInt(JabRefPreferences.FONT_SIZE));
+
+        // Set WM_CLASS using reflection for certain Un*x window managers
+        if (!OS.WINDOWS && !OS.OS_X) {
+            try {
+                Toolkit xToolkit = Toolkit.getDefaultToolkit();
+                java.lang.reflect.Field awtAppClassNameField = xToolkit.getClass().getDeclaredField("awtAppClassName");
+                awtAppClassNameField.setAccessible(true);
+                awtAppClassNameField.set(xToolkit, "org-jabref-JabRefMain");
+            } catch (Exception e) {
+                // ignore any error since this code only works for certain toolkits
+            }
+        }
 
     }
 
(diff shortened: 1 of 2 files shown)


Siedlerchr:
Hm, I appreciate this 👍 , but I will think it will backfire to us with java 9 or later , because we are effectively accessing internal Apis (for sure jdeps will complain), but as we still have do to this for the accent problem on linux I think it's worth. And I see no alternative.

Another thing I found is the following: I do not fully understand the discussion and the resulting code change. But what I understand is that they added a property`/method to the c++ code which sets this gtk stuff in javafx.
Maybe someone with more insight can help to clarify this

https://bugs.openjdk.java.net/browse/JDK-8097949

src/main/java/org/jabref/gui/GUIGlobals.java
```diff
@@ -156,6 +157,16 @@ public static void init() {
         GUIGlobals.currentFont = new Font(Globals.prefs.get(JabRefPreferences.FONT_FAMILY),
                 Globals.prefs.getInt(JabRefPreferences.FONT_STYLE), Globals.prefs.getInt(JabRefPreferences.FONT_SIZE));
 
+        // Set WM_CLASS using reflection for certain Un*x window managers
+        try {
```
lenhard: I guess we should only execute this code, if we are actually running on a unix system, right? That means it should all be wrapped in an `if (OS.LINUX)`

tobiasdiez:
According to the above link, for JavaFX it should suffice to call `setApplicationName` http://hg.openjdk.java.net/openjfx/8u60/rt/file/89345a49e71f/modules/graphics/src/main/java/com/sun/javafx/application/PlatformImpl.java#l109. But I have no idea how to do this...

src/main/java/org/jabref/gui/GUIGlobals.java
michaellass: This would also exclude Unix systems like FreeBSD, wouldn't it?

michaellass:
> According to the above link, for JavaFX it should suffice to call setApplicationName

It seems to me like this should happen automatically when the application is launched. But for some reason this is either not happening or the name is set to java-lang-Thread instead. I will try to find out what is going on here...

michaellass:
Here are some new results:
- LauncherImpl calls PlatformImpl.setApplicationName() during application startup
- this in turn calls setName() on the Application object
- both methods get the correct name ("org-jabref-JabRefMain") and this is successfully stored in the name attribute of the Application object
- the value of the name attribute does not change afterwards
- getName() of the Application object is never called

This leaves the big question: Why does my window manager still show "java-lang-Thread" as WM_CLASS? In case this is only an issue on specific Linux systems, here are information about mine:
* Desktop / window manager: GNOME 3.26.1
* GTK: 2.24.31
* JRE: openjdk 8u144
* JavaFX: openjfx 8u121
* GNOME runs on wayland but jabref is started as an X11 application using xwayland

Siedlerchr:
That is indeed mysterious: I can confirm this with Xfce, however in the Task Manager under Xfce it's correctly displayed

```
libgtk-3-0:amd 3.18.9-1ubun amd64       
libgtk2.0-0:am 2.24.30-1ubu amd64 
xprop|grep WM_CLASS
WM_CLASS(STRING) = "sun-awt-X11-XFramePeer", "java-lang-Thread"
```
http://i.imgur.com/jfwCfuR.png

src/main/java/org/jabref/gui/GUIGlobals.java
lenhard: In all my time in the JabRef team, I've never heard of someone running it on FreeBSD. Anyway, the alternative solution is `!OS.WINDOWS && !OS.OSX` :)

tobiasdiez:
Thanks for investigating this. I suppose the problem is again that we are mixing JavaFX and Swing. I'm fine with the `awt`-solution if it works as expected. Please add the `if` check @lenhard proposed and this is ready for merge.

src/main/java/org/jabref/gui/GUIGlobals.java
michaellass: I added that alternative variant now. Thanks for your review!

Issue #3247, opened 2017-10-02
Command line use doesn’t exit with -n on completion


davidaknowles:
JabRef version 4.0-beta3. Mac OS X 10.12.6

Why I care: 
http://discourse.jabref.org/t/command-line-use-doesnt-exit-with-n-on-completion/798

Steps to reproduce:
`java -jar JabRef.jar -n` hangs indefinitely.

MrTosz:
I'll have a go at it, if you don't mind.

stefan-kolb:
Fixed in #3274

PR #3274, opened 2017-10-06
Terminate JavaFX thread when running with "nogui" option

MrTosz:
Fix for Issue #3247

When JabRef was run with the '-n' parameter, i.e. the "No GUI" option, like:
`java -jar JabRef.jar -n`
or any other reason for `ArgumentProcessor#shouldShutDown()` returning `true` took place, the application should shut down after performing custom export filters etc. However, `Platform.exit()` wasn't called, so the JavaFX thread kept running indefinitely. Simply added this call to resolve the problem.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 
 ### Fixed
  - We fixed an issue where JabRef would not terminated after asking to collect anonymous statistics [#2955 comment](https://github.com/JabRef/jabref/issues/2955#issuecomment-334591123)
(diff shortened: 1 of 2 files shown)


Siedlerchr:
Thanks for your contribution, would you be so kind to add a changelog entry?
Then we can merge

MrTosz:
Sorry for omitting the changelog entry -- I wasn't sure if such fix qualified for a mention there. Will add it in a moment.

Siedlerchr:
@MrTosz  Thanks! The rule is simple: If it didn't work/has not been present in the version before, than we add a changelog entry. We are looking forward to more contributions from you!
I now merge your changes it.

PR #3291, opened 2017-10-12
Remove newlines in PDF annotations

### LINKED ISSUE #3280: When PDF file is highlighted without any comment, File annotations display "N/A", it is better to leave it empty
JabRef version 4.0 on Windows 10 

Just as the following screenshot, N/A seems a little intrusive for users.

na

stefan-kolb:
Resolves #3280


src/main/java/org/jabref/model/pdf/FileAnnotation.java
```diff
@@ -106,7 +106,10 @@ private String parseContent(final String content) {
             return "";
         }
 
-        return content.trim();
+        // remove line breaks
+        String processedContent = content.replace("\r\n", " ").replace("\n", " ").replace("\r", " ");
```
tobiasdiez: Can you please reuse the remove new lines formatter here, otherwise this is good to go.

stefan-kolb: @tobiasdiez This does not work with our current conventions, i had it this way before...: Adhere to architecture tests do not use logic from within model

tobiasdiez: Ok, thats again one of these cases where I don't like our strict architecture tests. One could argue that the Formatter can actually life in model as well, but I don't want to start a war here :-)

stefan-kolb: me neither ;-)

src/main/java/org/jabref/logic/formatter/bibtexfields/RemoveNewlinesFormatter.java
```diff
@@ -24,7 +24,7 @@ public String getKey() {
     public String format(String value) {
         Objects.requireNonNull(value);
 
-        return value.replace("\r\n", " ").replace("\n", " ").trim();
+        return value.replace("\r\n", " ").replace("\n", " ").replace("\r", " ").trim();
```
lynyus: I would also add a special treatment for hyphenated words.
Algorithm sketch: If the last character before the line termination symbol is a minus, replace the minus and the line-terminator with an empty string.

```
"...Jab-\n"
"ref is cool!\n" 
```
=>
```
"...JabRef is cool!"
```

stefan-kolb: What about Key-Account-Manager?

lynyus: Well... bad luck? words with a dash are quite rare, at least less frequent than hyphens at line breaks.

This is a problem, where we cant make a good decision.

stefan-kolb: Words with a dash are quite rare in English, maybe...

stefan-kolb: We can just add your solution and see if someone complains. I guess leaving the chars as they are is also not a bad decision.

lenhard: We've discussed this in the DevCall: Please go for the solution suggested by @LinusDietz

src/main/java/org/jabref/model/pdf/FileAnnotation.java
tobiasdiez: You can use the field formatter it in the view model of the file annotation instead of in the actual model (i.e. the actual model contains the text as it was in the pdf, but it is shown without line breaks).
https://github.com/JabRef/jabref/blob/efdf91a7d277a950204672c2c823d7e6dfe65305/src/main/java/org/jabref/gui/entryeditor/fileannotationtab/FileAnnotationViewModel.java#L19-L25

stefan-kolb:
@JabRef/developers Can someone do the localization update for me? The script doesn't work on my end 😢

Siedlerchr:
@stefan-kolb  I neither can run the script

lenhard:
Well if #3184 was broken, we can just revert the merge.

lenhard:
@stefan-kolb The localization script is working again. Can you merge master into this and then it should be ready to go.

LinusDietz:
Merging now, as this was discussed already in the devcall. The failing test concerning freecite is unrelated and has been fixed in https://github.com/JabRef/jabref/pull/3445.

PR #3293, opened 2017-10-12
File open dialog should have "supported formates" file type

### LINKED ISSUE #3255: import: file open dialog should have "supported formates" filetype
Using 4.0 stable with Windows 10.

The file open dialog while import should have a "supported formates" filetype. The list of supported file types is quite long and it costs time to find the correct type.

stefan-kolb:
Resolves #3255 

image

### DIFF
--- a/src/main/java/org/jabref/gui/importer/ImportFormats.java
+++ b/src/main/java/org/jabref/gui/importer/ImportFormats.java
@@ -81,6 +81,7 @@ public class ImportFormats {
                 DialogService ds = new FXDialogService();
 
                 FileChooser fs = ds.getConfiguredFileChooser(fileDialogConfiguration);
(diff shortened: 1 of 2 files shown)


src/main/java/org/jabref/gui/importer/ImportFormats.java
```diff
@@ -81,6 +81,7 @@ public void actionPerformed(ActionEvent e) {
                 DialogService ds = new FXDialogService();
 
                 FileChooser fs = ds.getConfiguredFileChooser(fileDialogConfiguration);
+                fs.getExtensionFilters().add(0, allImports);
```
Siedlerchr: Is there no setter for setting all filters?
Could be merged otherwise

PR #3300, opened 2017-10-13
Update libs

lynyus:
Merge if travis succeeds.


Siedlerchr:
Be careful with controlsfx9 , it is built on/with java 9 support.

Siedlerchr:
You could also check the fontawesomefx v 9 library (also with java 9 support)
http://www.jensd.de/wordpress/?p=2686

lynyus:
yeah, im on the train in the bavarian nirvana, that's why I cant download all these libs.

PR #3303, opened 2017-10-14
Open Unix filenames with spaces

### LINKED ISSUE #3264: Unable to "open folder" in case of filename with spaces
JabRef version 4.00 on CentOS 7.1

Expected:
When selecting the "Open folder" from the entry context menu, the folder containing the article should open.

What Happens:
Jabref opens a lot of locations based on the number of words in the name. Every word in the name is appended to the root location and get opened in Nautilus. 
For instance - if the file name is "someone2010 - A Sample file.pdf" saved in ~/, then the following folder will be invoked - 

~/someone2010
~/A
~/Sample
...

Steps to reproduce:
1. Create a BibTeX entry.
2. Attach a file with space in the name
3. Right-click on the entry, Select "Open Folder" or Ctrl + Shift + O

stefan-kolb:
Resolves #3264
We need to check if we need to escape the other unix paths too.


Siedlerchr:
I have to try if this works, but whille you are onto the code, you could take this as a chance to resolve the "alert/warning" from LGTM regarding String conceatenation in cmd parameters:
https://lgtm.com/projects/g/JabRef/jabref/alerts/?mode=tree&severity=error&rule=7870097
https://lgtm.com/rules/7870097/
Could you try to resolve this as an array? I think that would also resolve the problem I guess

stefan-kolb:
Should I add something similar for the other executables in Linux too?

Siedlerchr:
@stefan-kolb  Yes, make it consistent and then we can merge

stefan-kolb:
@LinusDietz Does my solution work then? Can you test it. I changed it to your \ suggestion.

stefan-kolb:
@Siedlerchr Does my solution work then? Can you test it. I changed it to your \ suggestion.

lenhard:
@Siedlerchr @LinusDietz Ping :-) Is this working? Can we merge?

LinusDietz:
I think it should, but I lack a linux system in question. (KDE or Gnome). @Siedlerchr, do you have Gnome or KDE?

LinusDietz:
I have tested the nautilus command, so I'm merging now. Don't have a kde environment at the moment.

PR #3305, opened 2017-10-15
Improvement for Java FX font rendering on Linux

halirutan:
The font-rendering of Java FX on Linux seems to be generally not as good as on other systems. I found the repo [Java FX hates Linux](https://github.com/woky/javafx-hates-linux) helpful which made a comparison for different settings for the *prism rendering engine*.

That being said, I found simply setting

``` java
System.setProperty("prism.text", "t2k");
System.setProperty("prism.lcdtext", "true");
```
provides the best font rendering without having to make changes in css that might affect other systems.

This pull-request adds a checkbox under Settings -> Appearance -> Look and Feel to turn this on after a restart. I incorporated it into the preference framework. On Linux, it is turned on in the default settings. 

settings

When turning it is turned on, the fonts are less blurry and the color bleaching is highly reduced (although we don't use gray scale font smoothing!). Compare these two screenshots:

1. [Tweaks turned off](http://i.imgur.com/BsFhpmk.png)
2. [Tweaks turned on](http://i.imgur.com/Ld8UJm2.png)


halirutan:
I would have provided localization, but it seems this is done on Windows and when I run `localizationUpdate`, all property files will be changed because the line separator is different. I can provide descriptions for English and German:

```
Tweak_font_rendering_for_entry_editor_on_Linux=Tweak_font_rendering_for_entry_editor_on_Linux
Tweak_font_rendering_for_entry_editor_on_Linux=Verbessertes_Schrift-Rendering_für_Linux
```

lynyus:
Are you sure that the screenshots are correctly named? I find http://i.imgur.com/BsFhpmk.png way better, however you state that font tweaking is OFF in this version.

src/main/java/org/jabref/gui/preftabs/AppearancePrefsTab.java
```diff
@@ -247,8 +253,12 @@ public void setValues() {
     public void storeSettings() {
         // L&F
         prefs.putBoolean(JabRefPreferences.USE_DEFAULT_LOOK_AND_FEEL, !customLAF.isSelected());
+        final boolean oldFxTweakValue = prefs.getBoolean(JabRefPreferences.FX_FONT_RENDERING_TWEAK);
+        prefs.putBoolean(JabRefPreferences.FX_FONT_RENDERING_TWEAK, fxFontTweaksLAF.isSelected());
         prefs.put(JabRefPreferences.WIN_LOOK_AND_FEEL, classNamesLAF.getSelectedItem().toString());
-        if ((customLAF.isSelected() == useDefaultLAF) || !currentLAF.equals(classNamesLAF.getSelectedItem().toString())) {
+        if ((customLAF.isSelected() == useDefaultLAF)
```
lynyus: Can you extract this into a private method named something like `isRestartRequired`?

Siedlerchr:
Regarding the line endings, check out this post which explains the configs:
https://stackoverflow.com/questions/10418975/how-to-change-line-ending-settings

halirutan:
First, @Siedlerchr: thanks for the link. I will look into it. For this PR, I will fix the line endings manually, but maybe it is possible to put [something like `-crlf`](https://stackoverflow.com/a/2517442/1078614) into `.gitattributes` for the property files under l10n.

@lynyus I will do the refactoring and include the localization. 

> Are you sure that the screenshots are correctly named?

Yes, I'm very sure, but I admit that I might be very pedantic about something like this as I'm working in image processing. However, I'm happy to give a demonstration: First, a screenshot from the swing text. which render OK (not as great as in OSX or Windows, but hey, it's Linux):

swing

You see that the antialiasing creates colors around the characters. Now, close your eyes *almost* completely to create a *blurred impression*. You'll notice that despite the color-antialiasing, the font appears consistent meaning no character sticks out.

Let's take a small portion from the FX fonts in the bottom

fx1

No matter how hard you will try, the vertical lines of the I, H, and the first vertical of the M feel too heavy compared to the extremely thin and sharp last vertical line of the M. When working, these things give the Entry Editor font a very inconsistent feeling.

Let's look at the same portion with the FX-tweak turned on.

fx2

Not completely awesome, but definitely closer to the swing fonts. I hope that proves my case and I dearly hope that it will work as good on other Linux systems.

lynyus:
Thanks for the explanation. I see the same effects that you explain when I zoom in, it is just that the screenshots without the tweaks look somehow clearer and more harmonic to read. What do you think, @JabRef/developers?

halirutan:
@lynyus Let me be a visitor for while :) I'm happy if I can help and I'm using JabRef regularly, but I'm not sure how much time I have beside my other stuff. I'm fine with making a PR for the moment.

koppor:
I like the screenshots with the tweaks.

Pro tip: One can drop images directly into the text boxes and GitHub uploads it transparently. No need for external hosters as stackoverflow does.

src/main/java/org/jabref/JabRefGUI.java
```diff
@@ -272,6 +272,13 @@ private void setLookAndFeel() {
                     LOGGER.warn("Unable to find requested look and feel", e);
                 }
             }
+
+            // On Linux, Java FX fonts look blurry per default. This can be improved by using a non-default rendering
+            // setting. See https://github.com/woky/javafx-hates-linux
+            if (Globals.prefs.getBoolean(JabRefPreferences.FX_FONT_RENDERING_TWEAK)) {
```
stefan-kolb: Why don't we just enable this for Linux by default then?

lynyus: Well, I think it is okay like this. Giving the Users the possibility to opt out is a good thing here, as all screens are different and maybe some users prefer the heavier font over the lighter one (to me the untweaked font looks somehow more appealing)

stefan-kolb: hm ok.

lynyus:
@halirutan thanks for the contribution! Such expert knowledge is always highly appreciated. If you need a break from research, you are always welcome to have a look at our issues. (;

halirutan:
@koppor Pro-pro-tip: The screenshot util on Linux can be configured to automatically upload images to imgur and gives you the link back. Only a Ctrl+V is required :smiley:

Issue #3313, opened 2017-10-17
Page views and exceptions are no longer tracked succesfully in Azure


tobiasdiez:
For some reason the released version of 4.0 does not send page views and exceptions to Azure. The latest message that was logged was on 
> 4.0-dev--snapshot--2017-09-29--master--6cca2ed32

It is also strange that customs events (new database / new entry) are still logged properly. Even stranger, starting in the same time period "track session: start / end" events appear, which were not logged before.

tobiasdiez:
Ok, there was a [release](https://github.com/Microsoft/ApplicationInsights-Java/releases) of azure 1.0.10 on the 30th september, which resulted in a lot of similar issues, e.g. https://github.com/Microsoft/ApplicationInsights-Java/issues/441.

tobiasdiez:
Fixed by https://github.com/JabRef/jabref/commit/8840da5767bd05e6bd538ff344807c65b3eeace6

PR #3317, opened 2017-10-17
Update applicationinsights from  1.0.9 -> 1.0.10

lynyus:
Applicationinsights is outdated.
Is there any reason, for not updating it? 

If so it should be stated in the gradle build file.

refs https://github.com/JabRef/jabref/commit/8840da5767bd05e6bd538ff344807c65b3eeace6

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -126,8 +126,8 @@ dependencies {
 
     compile 'com.github.tomtung:latex2unicode_2.12:0.2.2'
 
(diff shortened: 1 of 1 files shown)


tobiasdiez:
Yes, the 1.0.10 version is really buggy and does not send any telemetry (page views / exceptions) at all. See https://github.com/JabRef/jabref/issues/3313. Thus I downgraded it to 1.0.9. A new version 1.0.11 should be released soon and we can hopefully upgrade without problems.

PR #3320, opened 2017-10-18
Increase relative size of abstract field in editor

tobiasdiez:
Someone had the abstract field in the "required" tab. Since all fields were treated the same way, the abstract field was as big as the other fields resulting in a inconvenient display. See http://discourse.jabref.org/t/entry-preview-in-version-4/827/7. With this PR the solution for the keywords field is adopted and the abstract gets a bigger weight.

~~Not yet ready for review, a bit of code cleanup is still coming.~~

### DIFF
--- a/src/main/java/org/jabref/model/entry/FieldProperty.java
+++ b/src/main/java/org/jabref/model/entry/FieldProperty.java
@@ -17,7 +17,6 @@ public enum FieldProperty {
     EDITOR_TYPE,
     PAGINATION,
     TYPE,
-    CROSSREF,
     ISO_DATE,
     ISBN,
     EPRINT,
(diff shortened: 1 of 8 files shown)


src/main/java/org/jabref/model/entry/FieldProperty.java
```diff
@@ -20,16 +17,13 @@
     EDITOR_TYPE,
     PAGINATION,
     TYPE,
-    CROSSREF,
```
lenhard: Why is crossref gone here?

tobiasdiez: It was not used (only assigned) and there is `single/multiple entry link` which looks the proper generalization.

tobiasdiez:
Yes, the weights were previously used to determine the size of the field editors. I think our new approach is superior since now the model does not need to know anything about an entry editor at all and the size of the fields is determined in the gui, as it should be.

I'll merge this now.

Issue #3326, opened 2017-10-20
GUI button label is not correctly translated


halirutan:
If you set a different language than English, the search filter button is not correctly translated. In the image below, it is always "Filter" and "Float" no matter the language settings I have. The tooltips for these buttons are not translated either

![img](http://i.imgur.com/pt9REqi.png)

The reason for this is that the string values are initialized as static fields in `org.jabref.gui.search.SearchDisplayMode` when the user language was not loaded. Therefore, they will never be translated.

lenhard:
Should be fixed by #3327

PR #3327, opened 2017-10-20
Fix location bundle with fast access

halirutan:
This PR addresses the issue I mentioned under [Localization memory](https://github.com/JabRef/jabref/pull/3187#issuecomment-337464813). To provide a reasonable implementation and not only a bugfix I had to change the layout. These things bothered me the most:

- `LocalizationBundle` and `Localization` were calling each other. We needed `Localization` to create the keys in the bundle and on the other hand, the `LocalizationBundle` was only used in the `Localization` class
- `Localization` seemed like a quick fix to provide messages as bundle to Java FX
- On each call to any function in these classes, the unescaping of messages was done again. This, in particular, was the reason why showed up as hot spot in the profiler

To unify the implementation, I based the complete `Localization` framework on two instances (menus and messages) of `LocalizationBundle` and made this an internal class as it should not be used without it. The bundles are (re)created when `setLanguage` is called and contain the unescaped keys and messages for the specified language as a fast hash table.

This enables two things: (1) We can pass the `LocalizationBundle` instance directly to Java FX with `getMessages` and since we only pass the reference to the hash table, it will be instantaneous. (2) Every call to `lang` or `menuTitle` is a simple lookup in the hash table and a replacement of possible parameters in the string.

The public API for `Localization`, meaning `lang`, `menuTitle`, `getMessages` and `setLanguage` did not change and works as expected. During profiling/debugging I noticed, however, that there is a flaw in the JabRef startup. The language is set after the initialization of the preferences. At least one preference setting uses a localized string in a static field. Since when uninitialized, the `Localization` reverts back to "en", these strings will never be in the user language (see [this issue](https://github.com/JabRef/jabref/issues/3326)). Furthermore, due to this access, the `Localization` framework will always be initialized several times. Once in the init of the preferences and once further down in `main`.

This can only be fixed by pushing `setLanguage` to the initialization of the preferences which I did. One additional minor thing was that one message key was used escaped and had to be unescaped. Other than that, this implementation worked out of the box and passed all existing localization tests.

I commented my code heavily and would be glad if someone would review it. Don't forget that I'm looking only about 5 days on the code base and it is very likely I have overlooked something.


src/main/java/org/jabref/logic/l10n/Localization.java
```diff
@@ -1,41 +1,88 @@
...
+    /**
+     * Public access to all messages that are not menu-entries
+     * @param key The key of the message in unescaped form like "All fields"
+     * @param params Replacement strings for parameters %0, %1, etc.
+     * @return The message with replaced parameters
+     */
+    public static String lang(String key, String... params) {
+        if (localizedMessages == null) {
+            // I'm logging this because it should never happen
+            LOGGER.error("Messages are not initialized in " + Localization.class);
```
Siedlerchr: Yozu can omit the class part when writing a log message, as the Logger is already instantiated for the current class and therefore knows its name.

src/main/java/org/jabref/logic/l10n/Localization.java
```diff
@@ -48,53 +95,108 @@ public static void setLanguage(String language) {
...
 
     /**
-     * In the translation, %0, ..., %9 is replaced by the respective params given
+     * Helper function to create a hash-table from the key/value pairs of a bundle.
+     * @param baseBundle JabRef language bundle with keys and values for translations.
+     * @return Lookup table for the baseBundle.
+     */
+    private static Hashtable<String, String> createLookupTable(ResourceBundle baseBundle){
+        final ArrayList<String> baseKeys = Collections.list(baseBundle.getKeys());
+        Hashtable<String, String> lookup = new Hashtable<>(baseKeys.size());
```
Siedlerchr: HashTable is deprecated, better use HashMap.
And with java 8 streams and the Collection methods you an even make this code a lot nicer:
https://www.mkyong.com/java8/java-8-convert-list-to-map/

src/main/java/org/jabref/logic/l10n/Localization.java
```diff
@@ -48,53 +95,108 @@ public static void setLanguage(String language) {
...
+     * @param bundle The {@link LocalizationBundle} which means either {@link Localization#localizedMenuTitles} or {@link Localization#localizedMessages}.
+     * @param idForErrorMessage Identifier-string when the translation is not found.
+     * @param key The lookup key.
+     * @param params The parameters that should be inserted into the message
+     * @return The final message with replaced parameters.
      */
-    protected static String translate(ResourceBundle resBundle, String idForErrorMessage, String key, String... params) {
-        Objects.requireNonNull(resBundle);
+    private static String lookup(LocalizationBundle bundle, String idForErrorMessage, String key, String... params) {
+        if (key == null) {
```
Siedlerchr: Better use the Objects.requireNonNull here. It will automatically throw an NPE

src/main/java/org/jabref/logic/l10n/Localization.java
```diff
@@ -48,53 +95,108 @@ public static void setLanguage(String language) {
...
-    public static String lang(String key, String... params) {
-        if (messages == null) {
-            setLanguage("en");
+    /**
+     * A bundle for caching localized strings. Needed to support JavaFX inline binding.
+     */
+    private static class LocalizationBundle extends ResourceBundle {
+
+        // I'm using a Hashtable since it allows to get the keys directly as an Enumeration
+        private Hashtable<String, String> lookup;
```
Siedlerchr: As mentioned above, use as HashMap

src/main/java/org/jabref/logic/l10n/Localization.java
```diff
@@ -48,53 +95,108 @@ public static void setLanguage(String language) {
...
+     * @param locale Localization to use.
+     */
     private static void createResourceBundles(Locale locale) {
-        messages = ResourceBundle.getBundle(RESOURCE_PREFIX, locale, new EncodingControl(StandardCharsets.UTF_8));
-        menuTitles = ResourceBundle.getBundle(MENU_RESOURCE_PREFIX, locale, new EncodingControl(StandardCharsets.UTF_8));
+        ResourceBundle messages = ResourceBundle.getBundle(RESOURCE_PREFIX, locale, new EncodingControl(StandardCharsets.UTF_8));
+        ResourceBundle menuTitles = ResourceBundle.getBundle(MENU_RESOURCE_PREFIX, locale, new EncodingControl(StandardCharsets.UTF_8));
+        // Just for the case something really stupid happened.
+        if (messages == null || menuTitles == null) {
+            LOGGER.error("Could not load language translation files in " + Localization.class);
```
lenhard: Same remark regarding the class files as by @Siedlerchr above

src/main/java/org/jabref/logic/l10n/Localization.java
```diff
@@ -48,53 +95,108 @@ public static void setLanguage(String language) {
...
+            throw new NullPointerException();
```
lenhard: I think throwing an `IllegalStateException` would be better here. Just for the more descriptive exception type.

src/main/java/org/jabref/logic/l10n/Localization.java
```diff
@@ -48,53 +95,108 @@ public static void setLanguage(String language) {
...
+            this.lookup = lookupTable;
         }
-        return translate(messages, "message", key, params);
-    }
 
-    public static String menuTitle(String key, String... params) {
-        if (menuTitles == null) {
-            setLanguage("en");
+        public final Object handleGetObject(String key) {
+            if (key == null) {
```
lenhard: As above, rather use `Objects.requireNonNull` here.

halirutan:
I fixed the issues. Is there a nicer way to create a `HashMap` from the `Collectors.toMap` than simply wrapping it around?

Siedlerchr:
Unfortunately there is no simpler solution for the Map thing. 
Looks good to me, but you have still some checkstyle issues, here is a template? for Intellij:
https://github.com/JabRef/jabref/tree/master/config

src/main/java/org/jabref/logic/l10n/Localization.java
halirutan: I used

```
        Objects.requireNonNull(messages, "Could not load " + RESOURCE_PREFIX + " resource.");
        Objects.requireNonNull(menuTitles, "Could not load " + MENU_RESOURCE_PREFIX + " resource.");
```

which throws NPE, but gives a description.

src/main/java/org/jabref/logic/l10n/Localization.java
halirutan: Done.

src/main/java/org/jabref/logic/l10n/Localization.java
halirutan: Done

lenhard:
Since the checkstyle problems are fixed, I'll merge this now.

Thanks a lot for your contribution!

PR #3334, opened 2017-10-22
Fix initialization using injection context for complex objects

tobiasdiez:
In order to help @Siedlerchr, I fixed a bug in the framework which we use to inject dependencies in the javafx controller. I proposed a PR https://github.com/AdamBien/afterburner.fx/pull/80 but as he is a bit slow with integrating fixes, I think we should directly include the fix in JabRef for the moment (and remove it as soon as a new version of afterburner.fx with the fix is released).

### DIFF
--- a/src/main/java/org/jabref/gui/BaseInjector.java
+++ b/src/main/java/org/jabref/gui/BaseInjector.java
@@ -1,0 +1,236 @@
+package org.jabref.gui;
+
+/*
+ * #%L
+ * afterburner.fx
+ * %%
+ * Copyright (C) 2013 Adam Bien
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


Siedlerchr:
Why don't we create a fork of the original project and include other fixes as well, e.g. the system errr output (there is already a Pr regarding that)

lenhard:
The problem here is the license. We cannot just copy the code and use it here, because it is licensed differently. We will need to go for a fork / separate library.

@koppor will contact the author of afterburner per mail. Maybe this wakes him up.

tobiasdiez:
I really like license problems :-). I would actually help Adam Bien to maintain this library. I think, of now JabRef is one of the main users of afterburner and thus we should be interested in having a stable and working platform. (I don't like a separate fork).

Issue #3335, opened 2017-10-23
Can the context-menus for the field editor be reused?


halirutan:
A great deal of work was done lately on the entry editor. One last thing that waves a red flag there are (surprise, surprise) the context menus. Believe it or not, those tooltips are waste more memory than you would think considering those are simple text things.

Let me make it clear: One run with opened editor through the 200 entry example bibfile. This does now only use 30MB of memory on its way. The "required tab" is of course always visible and therefore, the contents of this tab needs to be initialized and filled with data. Now hold your breath: 22MB of those 30MB are the context menus for the text fields.

<img width="1550" alt="screen shot 2017-10-23 at 11 50 28" src="https://user-images.githubusercontent.com/511683/31882869-709c03d4-b7e8-11e7-849f-015e6a536862.png">

I have seen that for each of those text fields, every tooltip and every context menu is a new instance. The question is, is it possible to make this more static. Tooltips can surely be initialized once and reused. But maybe it is possible to do this to a certain extent for the whole context menu like the `CaseChangeMenu`? Tooltips alone would account for 12MB and basically, only the 4 classes under `package org.jabref.gui.fieldeditors.contextmenu` need to be changed.

Is this possible?

tobiasdiez:
Thanks again for your investigations!

You can try to modify
https://github.com/JabRef/jabref/blob/d922e5597f7496515ca5e8d274190539405cc5ae/src/main/java/org/jabref/gui/fieldeditors/EditorTextArea.java#L54-L63
to take a `Supplier<List<MenuItem>>` as the argument (instead of the already created list of items) and generate the list in the overriden `populate` method. This might work (depending on whether `populate` is always called or only when the context menu is needed).

But I wouldn't spend to much energy on the context menus. If this is now the biggest contributor, we are in quite a luxury position and think we shouldn't overpush.

halirutan:
@tobiasdiez Thanks for the tip!

lenhard:
Fixed with #3340

PR #3340, opened 2017-10-24
Used late initialization for context menus

halirutan:
I believe I could bring down the performance impact of the context menus to a bare minimum. I reported this in https://github.com/JabRef/jabref/issues/3335 and @tobiasdiez provided me with the hint how this could work. Against his suggestion to not overshoot, I decided to prove my point that even those small context menus have a significant impact because things will sum up in the end and those menus are newly instantiated for every visible field in the editor.

The starting point was the current master-branch and profiling it by going through only 50 bib-entries. This means that the context menus are instantiated for the visible text fields. This has an impact on performance and memory and, unfortunately, it is not only a tiny fraction. 

## Memory

Notable are the columns "Objects", which are all objects that are created by this method and underlying sub calls and the overall "Size".

mem1

With almost 40% of the overall memory size, it's definitely something to look into.

## Performance

The performance is with 8% not so large but still noticeable. 

perf1

## Fixing it

By using the `Supplier` interface, we can move the instantiation to the point where the user actually right-clicks into the field. I debugged this, and it seems to work nice. The outcome was as expected. Since no one actually triggers the context menu by just browsing through the list, it vanished altogether. Here is the used memory with this PR created by exactly scrolling through the 50 entries. First the memory footprint

mem2

and here the performance impact

perf2

In conclusion, this means as long as the user doesn't need a context menu, we have no memory or performance impact at all.

## Note

I updated the `checkstyle` version in the Gradle built. I like to write comments, and now it doesn't prevent me any longer from using `{@link ...}` in doc-comments to classes that are otherwise not used.

- [x ] Manually tested changed features in running JabRef

### DIFF
--- a/src/main/java/org/jabref/gui/fieldeditors/EditorTextArea.java
+++ b/src/main/java/org/jabref/gui/fieldeditors/EditorTextArea.java
@@ -3,6 +3,7 @@ package org.jabref.gui.fieldeditors;
 import java.net.URL;
 import java.util.List;
 import java.util.ResourceBundle;
+import java.util.function.Supplier;
 
 import javafx.fxml.Initializable;
 import javafx.scene.control.ContextMenu;
@@ -49,14 +50,16 @@ public class EditorTextArea extends javafx.scene.control.TextArea implements Ini
     }
 
     /**
-     * Adds the given list of menu items to the context menu.
+     * Adds the given list of menu items to the context menu. The usage of {@link Supplier} prevents that the menus need
+     * to be instantiated at this point. They are populated when the user needs them which prevents many unnecessary
+     * allocations when the main table is just scrolled with the entry editor open.
      */
-    public void addToContextMenu(List<MenuItem> items) {
+    public void addToContextMenu(Supplier<List<MenuItem>> items) {
         TextAreaSkin customContextSkin = new TextAreaSkin(this) {
             @Override
             public void populateContextMenu(ContextMenu contextMenu) {
                 super.populateContextMenu(contextMenu);
-                contextMenu.getItems().addAll(0, items);
+                contextMenu.getItems().addAll(0, items.get());
             }
         };
         setSkin(customContextSkin);
--- a/src/main/java/org/jabref/gui/fieldeditors/IdentifierEditor.java
+++ b/src/main/java/org/jabref/gui/fieldeditors/IdentifierEditor.java
@@ -1,14 +1,11 @@
 package org.jabref.gui.fieldeditors;
 
-import java.util.ArrayList;
-import java.util.List;
 import java.util.Optional;
 
 import javafx.event.ActionEvent;
 import javafx.fxml.FXML;
 import javafx.scene.Parent;
 import javafx.scene.control.Button;
-import javafx.scene.control.MenuItem;
 import javafx.scene.control.Tooltip;
 import javafx.scene.layout.HBox;
 
@@ -25,10 +22,14 @@ import org.jabref.preferences.JabRefPreferences;
 
 public class IdentifierEditor extends HBox implements FieldEditorFX {
 
-    @FXML private IdentifierEditorViewModel viewModel;
-    @FXML private EditorTextArea textArea;
-    @FXML private Button fetchInformationByIdentifierButton;
-    @FXML private Button lookupIdentifierButton;
+    @FXML
+    private IdentifierEditorViewModel viewModel;
+    @FXML
+    private EditorTextArea textArea;
+    @FXML
+    private Button fetchInformationByIdentifierButton;
+    @FXML
+    private Button lookupIdentifierButton;
     private Optional<BibEntry> entry;
 
     public IdentifierEditor(String fieldName, TaskExecutor taskExecutor, DialogService dialogService, AutoCompleteSuggestionProvider<?> suggestionProvider, FieldCheckers fieldCheckers, JabRefPreferences preferences) {
@@ -43,12 +44,11 @@ public class IdentifierEditor extends HBox implements FieldEditorFX {
         lookupIdentifierButton.setTooltip(
                 new Tooltip(Localization.lang("Look up %0", FieldName.getDisplayName(fieldName))));
 
-        List<MenuItem> menuItems = new ArrayList<>();
         if (fieldName.equalsIgnoreCase(FieldName.DOI)) {
-            menuItems.addAll(EditorMenus.getDOIMenu(textArea));
+            textArea.addToContextMenu(EditorMenus.getDOIMenu(textArea));
+        } else {
+            textArea.addToContextMenu(EditorMenus.getDefaultMenu(textArea));
         }
-        menuItems.addAll(EditorMenus.getDefaultMenu(textArea));
-        textArea.addToContextMenu(menuItems);
 
         new EditorValidator(preferences).configureValidation(viewModel.getFieldValidator().getValidationStatus(), textArea);
     }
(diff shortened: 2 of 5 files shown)


stefan-kolb:
Nice work @halirutan Anytime your PRs are ready for review please add a 'ready-for-review' label.

halirutan:
@stefan-kolb At the moment, I'm not sure I'm allowed to do this.

stefan-kolb:
@halirutan Oops, ok I thought you already have acces 😄

src/main/java/org/jabref/gui/fieldeditors/IdentifierEditor.java
```diff
@@ -25,10 +22,14 @@
 
 public class IdentifierEditor extends HBox implements FieldEditorFX {
 
-    @FXML private IdentifierEditorViewModel viewModel;
-    @FXML private EditorTextArea textArea;
-    @FXML private Button fetchInformationByIdentifierButton;
-    @FXML private Button lookupIdentifierButton;
+    @FXML
```
tobiasdiez: Our convention is that these `@FXML` tags should be kept on the same line (I think this is also configured in the intellj style)

src/main/java/org/jabref/gui/fieldeditors/contextmenu/EditorMenus.java
```diff
@@ -14,42 +15,73 @@
...
-        menuItems.add(new SeparatorMenuItem());
-        return menuItems;
+    /**
+     * The default menu that contains functions for changing the case of text and doing several conversions.
+     *
+     * @param textArea text-area that this menu will be connected to
+     * @return default context menu available for most text fields
+     */
+    public static Supplier<List<MenuItem>> getDefaultMenu(TextArea textArea) {
+        return () -> {
```
tobiasdiez: I'm not sure what is more readable: Hiding the supplier interface here like you did or instead writing something like `addToContextMenu(() -> getDefaultMenu(textArea))`. @JabRef/developers any opinions on this?

src/main/java/org/jabref/gui/fieldeditors/IdentifierEditor.java
halirutan: It's what IntelliJ did, therefore I thought it might be overlooked in the first place. I can change this

lenhard: This is a fetish of @tobiasdiez :P For the sake of being consistent to what we have done before, please keep the `@FXML` in the same line a the instance variable.

src/main/java/org/jabref/gui/fieldeditors/contextmenu/EditorMenus.java
lenhard: I think I am in favor of the second solution, i.e. putting the supplier into `EditorTextArea.addToContextMenu()`. Reasoning: With the current solution, the lazy initialization is dispersed over two classes. Changes are required in `EditorTextArea` and `EditorMenus`. If I understand everything correctly, we could build the supplier inside `addToContextMenu()` only. That would be a very localized change that hides the lazy initialization as far as possible.

My perception might be wrong, though. Please correct me if this is the case.

src/main/java/org/jabref/gui/fieldeditors/EditorTextArea.java
```diff
@@ -49,14 +50,16 @@ public EditorTextArea(String text) {
...
+     * allocations when the main table is just scrolled with the entry editor open.
      */
-    public void addToContextMenu(List<MenuItem> items) {
+    public void addToContextMenu(Supplier<List<MenuItem>> items) {
         TextAreaSkin customContextSkin = new TextAreaSkin(this) {
             @Override
             public void populateContextMenu(ContextMenu contextMenu) {
                 super.populateContextMenu(contextMenu);
-                contextMenu.getItems().addAll(0, items);
+                contextMenu.getItems().addAll(0, items.get());
```
lenhard: If I understand everything correctly, we could build the supplier right here and still accept a `List<MenuItem>` as method parameter. That would keep the changes very localized.

More discussion see below.

halirutan: Hmm, then I misunderstood the whole thing. Every text-field editor needs to attach its own context menu. Als long as I have wrapped it in a supplier, the code is not instantiated. So how is for instance `IdentifierEditor` supposed to attach a non-instantiated context menu?

```
textArea.addToContextMenu(EditorMenus.getDefaultMenu(textArea));
```

Assume we don't do the wrapping inside `EditorMenus`, then the menu would be instantiated exactly there. Except when I put the `Supplier` wrapper at this place. But then we have the suppliers scattered over even more classes. What do I miss?

src/main/java/org/jabref/gui/fieldeditors/IdentifierEditor.java
halirutan: Found it. We need to set this in our code style for IDEA

img

src/main/java/org/jabref/gui/fieldeditors/EditorTextArea.java
tobiasdiez: No, Jörg, the supplier has to be built outside of this method since otherwise the context menu items are already created. With the help of a supplier, we are able to delay the creation until the user opens the context menu (since only then the `get` method is called).

lenhard: Ok, then I was wrong. Thanks for clarifying! In that case, please disregard my comment and keep the solution.

halirutan:
@lenhard I wanted to comment on this

> Here's my memory profile of the keydown behavior:

The keydown behavior is a good general indicator, but there are some drawbacks. The main issue with it is that it can be heavily misleading. Assume for a moment that we ignore everything else and look only at the impact of showing the entry editor. When we have fixed an issue like the one in this PR that does not only affect memory but performance as well, then the speed of creating the entry editor will improve. That directly impacts how many entries you can visit on holding the key down.

Now assuming that we could speed up things by 50% but couldn't reduce the memory impact, then your memory monitor will make the impression if things had gone worse because it will fill the memory faster. That was the main reason why I reported both the memory and the performance separately under controlled circumstances (visiting exactly 50 entries).

What we would need (maybe we already have this) are what we call [model level functional tests](https://www.jetbrains.org/intellij/sdk/docs/basics/testing_plugins.html) in IDEA plugin development. So, for instance, a test that visits n entries and simulates a click or starts JabRef and performs a search.

Those tests shouldn't be run by travis but they should be available for profiling to clearly pinpoint if we have a regression or a speedup.

lenhard:
@halirutan You are right of course.

We have some GUI tests in `jabref/src/test/java/org/jabref/gui/`  but they are run on travis only, because they are rather annoying to run locally. Also, they are not focused on performance at all.

We also have a few performance tests in `jabref/src/jmh/java/org/jabref/benchmarks/`
 but they are not related to the GUI and this is where our major performance problems are at the moment.

It would be good to have the tests you describe and, as usual, it depends on someone being willing to put the effort into. Most people (including myself) are a little shy when it comes to UI testing in JabRef, because the user interface has been heavily changed in the last year, meaning you have to re-write the tests all the time. If you are interested in doing something here, I would suggest you have a look at the current GUI tests, but if you want to use a better framework for building such tests, be my guest!

A new minor release of JabRef is imminent. Since I definitely want to have this in, I'll merge it right now. It is complete as far as I can see (the other refactoring I proposed does not work, as we have discussed).

Thank you for your contribution.

koppor:
@halirutan - Did you sign up for https://hacktoberfest.digitalocean.com/?

halirutan:
@koppor I've seen it but I thought it was a JabRef internal thing. After all, many of the contributors are located in Germany and I thought it was a fun way of dedicating a month to pushing things forward.

PR #3353, opened 2017-10-26
Refactor collab package

tobiasdiez:
This PR refactors the collab package and splits it into gui and logic parts. More precisely, the code that was used to determine the differences between two databases now resides in logic, while the rest went to the gui package (one might try to extract even more code to logic but this is a first good start in my opinion). 

I tried to not change the functionality. This of course means that I tried not to introduce bugs (although I'm sure there will be small glitches) but it also implies that I didn't improved the code too much. If somebody has time, he or she is really invited to have a look at the feature. Many things are not yet implemented (e.g. a real metadata difference) and the ui is just ugly. 

However, there is one thing which did changed: previously it was a kind of 3-way merge between the last saved version, the current state and the external changes to the bib file. However, the current state was taken into account only sometimes (and sometimes in the wrong way). Thus I removed this part completely and moved to a 2-way merge between the last saved version and the external changes. In rare cases, this might lead to an unnecessary notification of external changes (namely exactly when the external changes coincide with what the user changed since the last save). 

If you merge, it probably preferable to not smash the commits.


lenhard:
Even if this is not as big a refactoring as it could have been, I appreciate it very much. At least it is moving the package structure a little more in the direction in which we intend to be going. After the refactoring, the `pdfimport` and `shared` packages are the last problematic parts. Especially the `shared` package is a real problem, because it introduces a transitive dependency from `model` into `gui`, but that is a different topic.

I have played around with a running JabRef instance. Regardless of what external change I do, it notifies me of an additional change in the meta data (but the change is empty). That is probably related to the unnecessary notifications you mention above. I don't care too much if JabRef notifies me when external changes do not actually change file content, but it is annoying that it notifies me of a meta data change, when I just change an entry. Could you please look into this?

Siedlerchr:
Previously: an external change occurs, the Change Monitor occures and notifies. However, even after accepting all changes and saving, the change monitor would kick in again. So I would suggest to handle that case somewhere that when I accepted my external changes and press save that I don't get that dialog again

tobiasdiez:
@lenhard good observation! The content selectors didn't had a proper equals method and thus were always reported as a difference (this problems should have occurred also before the refactoring since I didn't change anything in the metadata comparison). I now fixed it and even added a test covering it.

@Siedlerchr this sounds like a reasonable thing to do. However, this PR is only about the refactoring and as I said above I'll try to not touch the functionality. Thus I will leave this to a follow-up.

Issue #3352, opened 2017-10-25
Duplication when using  entry editor


sambo57u:
This is 4.1dev Oct.25 on Linux x86_64 with java 1.8.0_152.

1. Open a bib file
2. Double clck on an entry so that the source editor window opens up
3. Click into the editor source window
4. Then click onto one of the other bib entries shown in the table above the editor
5. The other entry is duplicated

I also had a problem with changes done via the source editor not being saved. It may be related to
above. Will investigate further.

halirutan:
This is neat and it was an overlook by the reviewers (including myself) who approved the PR https://github.com/JabRef/jabref/pull/3333 from @tobiasdiez. What happens is that on moving to a new entry we call `bindToEntry` which in turn creates a new source editor. This will call `storeSource` with the new entry [here](https://github.com/JabRef/jabref/blob/master/src/main/java/org/jabref/gui/entryeditor/SourceTab.java#L89-L92). The `entry` is already the new one but the code in the code area is the old one and it overwrites it.

It should be an easy fix and maybe the right time to review if the communication flow does really what we intend. What I would suggest is:

- if someone types into this field, the `storeEntry` method should be called after a tiny pause to make changes instantly visible if the code is correct
- switching tabs or moving to a different entry will ensure as well that modifications are reflected
- while typing, changes are only reflected when the parser says it's valid BibTeX code
- when switching tabs or moving to a different and the code is incorrect, we probably should send a warning instead of silently not saving the code

Currently, this does not work correctly. Making changes that invalidate the entry will be silently ignored when you switch tabs or entries.

lenhard:
The source tab with its dependencies on the other tabs is always a problem... It would be good if this got fixed before we release 4.1.

Most of what you write seems reasonable, except for reflecting changes while a user is typing in the source tab. If we immediately store every character change (as was the case for some time during development between 3.8.2 and 4.0), then we really annoy all users who have autosave or a sync to a database enabled. I think it would be best if we store the contents every time the text area looses focus or the entry changes.

Siedlerchr:
@lenhard didn't you implement the CoarseChangeFilter which only fires, if one leaves the field? Can't we re-use that here?

halirutan:
@lenhard 

> If we immediately store every character change (as was the case for some time during development between 3.8.2 and 4.0), then we really annoy all users who have autosave or a sync to a database enabled. 

[...]

PR #3360, opened 2017-10-27
Source tab entry duplication

lenhard:
Adapts the source tab a little better to the new and no longer re-created entry editor. It is a fix for the problem described in this comment https://github.com/JabRef/jabref/issues/3352#issue-268580818 The remaining problems in the issue are not addressed here, so the issue should stay open when merging this and the update flow in the source tab should be reworked in further PRs.

### DIFF
--- a/src/main/java/org/jabref/gui/entryeditor/SourceTab.java
+++ b/src/main/java/org/jabref/gui/entryeditor/SourceTab.java
@@ -78,7 +78,7 @@ public class SourceTab extends EntryEditorTab {
         }
     }
 
-    private Node createSourceEditor(BibEntry entry, BibDatabaseMode mode) {
+    private Node createSourceEditor(BibDatabaseMode mode) {
         codeArea = new CodeArea();
         codeArea.setWrapText(true);
         codeArea.lookup(".styled-text-area").setStyle(
@@ -86,12 +86,12 @@ public class SourceTab extends EntryEditorTab {
         // store source if new tab is selected (if this one is not focused anymore)
         EasyBind.subscribe(codeArea.focusedProperty(), focused -> {
             if (!focused) {
-                storeSource(entry);
+                storeSource();
             }
         });
 
         try {
-            String srcString = getSourceString(entry, mode);
+            String srcString = getSourceString(this.currentEntry, mode);
             codeArea.appendText(srcString);
         } catch (IOException ex) {
             codeArea.appendText(ex.getMessage() + "\n\n" +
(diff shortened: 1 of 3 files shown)


halirutan:
To give some feedback and to spare you unnecessary work. I had already rewritten this yesterday and I'm sorry for not saying that I look over this. When I saw that you started this too, I fetched your branch and my initial idea was, just to provide the functionality to make `Ctrl+S` work.

However, I believe the information flow for the `EntryEditor` is extremely confusing so I ended up rewriting this completely. The main idea is to make the `EntryEditor`implement [`EntryEditorInfo`](https://gist.github.com/halirutan/d6386973d7939fd6e2d721673b6139ce) that provides access for each tab. Therefore, we don't pass entries, entry-types, contexts, etc. between the editor and its tabs, instead, each tab accesses the information when it needs it. This let me through out all the methods that pass things and the tabs are only *notified* when they get focus or when a new entry is set by the `EntryEditor`. This makes things much more localized and easier to follow.

I would like to push this on a new branch and test it carefully. It will include the changes you made here. The issue with `Ctrl+S` can be resolved by putting this in the `setupKeyBindings()` for the EntryEditor where we act on (but don't consume) that save/saveAll events:

``` java
                    switch (keyBinding.get()) {
                        case CUT:
                        case COPY:
                        case PASTE:
                        case CLOSE_ENTRY_EDITOR:
                        case DELETE_ENTRY:
                        case SELECT_ALL:
                        case ENTRY_EDITOR_NEXT_PANEL:
                        case ENTRY_EDITOR_NEXT_PANEL_2:
                        case ENTRY_EDITOR_PREVIOUS_PANEL:
                        case ENTRY_EDITOR_PREVIOUS_PANEL_2:
                            e.consume();
                            break;
                        case SAVE_ALL:
                        case SAVE_DATABASE:
                        case SAVE_DATABASE_AS:
                            sourceTab.saveCurrentSource();
                            break;
                        default:
                            //do nothing
                    }
```

The things I tested worked but there is one complicated situation: User edited the bibtex source and made a syntax error in it. Then he clicks on a different entry in the main table. Besides that the current "let me edit my wrong code" does not work as it stands (exception is not caught), there is a deeper problem: The main table is already on a different entry and the `EntryEditor` has already fetched the new entry. To make a successful *revert&edit* we need to tell the main table to jump back to the erroneous entry and set it in `EntryEditor`. 

@lenhard Shall I put my changes on a different branch? In this way we have a quick fix with your PR and can test things carefully.

lenhard:
@halirutan Thanks for your input! What you write sounds reasonable and I agree that the information flow in the entry editor is crying for a larger refactoring.

I have just pushed a fix to repair the build here and when it has passed we could merge this already. You can do your changes on a different branch, then. And we can give it a thorough look and testing.

Just be aware that even more problems apart from syntax error correction might come up. The tabs are not only dependent on the entry editor, but also partly dependent on each other. E.g. if a field value changes in the source tab, it should be changed in another tab as well and vice versa (not so hard) and if I add a new optional field in the source tab, it should make the optional tab make appear (somewhat harder).

src/main/java/org/jabref/gui/entryeditor/SourceTab.java
```diff
@@ -116,11 +116,6 @@ public boolean shouldShow(BibEntry entry) {
 
     @Override
     protected void bindToEntry(BibEntry entry) {
-        // store source if new entry is selected in the maintable and the source tab is focused
-        if (codeArea != null && codeArea.focusedProperty().get()) {
-            DefaultTaskExecutor.runInJavaFXThread(() -> storeSource());
-        }
-
```
halirutan: The problem here is btw that you run it in a separate thread. This means, the moment `storeSource` starts to work, a new entry is already set and what you do is to overwrite the correct new entry with itself.

lenhard:
Perfect! Then I'll merge this directly so that we have a quick fix to the issue and you can go ahead with your larger refactoring.

PR #3362, opened 2017-10-27
Integrity check "Abbreviation Detection" detects abbreviated names for journals and booktitles based on the internal list instead of only looking for "." signs. 

### LINKED ISSUE #3144: Journal abreviation and integrity check
JabRef version 3.8.2 and 4.0 on macOS 10.12

Using integrity check on my database, I noticed that as soon as there is a point (".") in the journal's name, it creates a warning when I check the integrity of my database. I do not know if it is really a bug or a feature but it is not really convenient. For example, according to http://www.ams.org/msnhtml/serials.pdf (a good source for journal abreviations I guess)
ESAIM. Mathematical Modelling and Numerical Analysis = ESAIM Math. Model. Numer. Anal.  

but since there is a point in the full name (ESAIM.   ...) it is considered by JabRef as an abbreviation and gives a warning during integrity check. I tried to add this journal to my personal journal list but it does not change anything.

If it is not a bug, I suggest to check if it is a known full name (done in 4.0 I think) or if it is a known abbreviation, otherwise Jabref could give this kind of warning: "journal's name not recognized".

simonharrer:
Fixes #3144 

#hacktoberfest


simonharrer:
Updated. Please have a look.

Issue #3217, opened 2017-09-13
Again with the editor panels


wujastyk:
JabRef 4.0-dev--snapshot--2017-09-13--master--343545ee8
Linux 4.10.0-33-generic amd64 
Java 1.8.0_144

Steps to reproduce:

1. open a blank Article for editing
2. Go to Optional Fields tab and enter the DOI.  Download the DOI info and "replace" the current (empty) fields.
3. The Required Fields and Optional Fields tabs are now populated correctly
3. The biblatex source tab is empty except for the DOI

See video log at [https://www.dropbox.com/s/3vwxppehvkyulp6/jabref%20report.mp4?dl=0](https://www.dropbox.com/s/3vwxppehvkyulp6/jabref%20report.mp4?dl=0)

Siedlerchr:
This looks like a similar problem we had with #3208  that the underlying model is updated but not the view model

wujastyk:
JabRef 4.1-dev--snapshot--2017-10-09--master--1b030b459
Linux 4.10.0-37-generic amd64 
Java 1.8.0_144

1. Create entry in the "biblatex source" tab
2. switch to "required entry" tab - there's nothing there. 

See [video]([url](https://www.dropbox.com/s/tpr52kq3ugn1ox5/jabref%20report.mp4?dl=0)).

lenhard:
I just tested this on current master with the DOI `10.1103/PhysRevLett.119.177201`. Now, the source tab and the other tabs are populated correctly, so this is almost fixed.

Almost, since I also get an exception. A normal user would not notice this since the entry editor keeps working, but it still cannot be good:
```
[JavaFX Application Thread] ERROR org.jabref.FallbackExceptionHandler - Uncaught exception occurred in Thread[JavaFX Application Thread,5,main]
java.lang.IndexOutOfBoundsException: [0, 11) is not a valid range within [0, 1)
	at org.reactfx.util.Lists.checkRange(Lists.java:128) ~[reactfx-2.0-M5.jar:?]
	at org.reactfx.util.Lists.checkRange(Lists.java:123) ~[reactfx-2.0-M5.jar:?]
	at org.reactfx.util.SparseList.getPresentCountBetween(SparseList.java:278) ~[reactfx-2.0-M5.jar:?]
	at org.reactfx.util.SparseList.collect(SparseList.java:310) ~[reactfx-2.0-M5.jar:?]
[...]

PR #3366, opened 2017-10-28
Update entry for every change in the source panel

tobiasdiez:
In https://github.com/JabRef/jabref/issues/3352#issuecomment-339910785, @lenhard suggested to update the entry as soon as the source code is changed (instead of only on focus lost). This is accomplished in this PR. 
Jörg mentioned that the most direct solution does not work since
> the store operation will move the cursor to the end of the source tab.

This problem is circumvented by implementing a proper bidirectional binding between the entry and the source code, which prevents update cycles.

Moreover, in case of a parse error, the message is no longer shown in a dialog but inline as a notification:
image

### DIFF
--- a/src/main/java/org/jabref/gui/entryeditor/EntryEditor.java
+++ b/src/main/java/org/jabref/gui/entryeditor/EntryEditor.java
@@ -83,7 +83,6 @@ import org.jabref.model.EntryTypes;
 import org.jabref.model.database.BibDatabase;
 import org.jabref.model.entry.BibEntry;
 import org.jabref.model.entry.EntryType;
-import org.jabref.model.entry.event.EntryChangedEvent;
 import org.jabref.model.entry.event.FieldAddedOrRemovedEvent;
 import org.jabref.preferences.JabRefPreferences;
 
@@ -196,7 +195,6 @@ public class EntryEditor extends JPanel implements EntryContainer {
 
     public void setEntry(BibEntry entry) {
         this.entry = Objects.requireNonNull(entry);
-        entry.registerListener(this);
         entryType = EntryTypes.getTypeOrDefault(entry.getType(),
                 this.frame.getCurrentBasePanel().getBibDatabaseContext().getMode());
 
@@ -221,11 +219,6 @@ public class EntryEditor extends JPanel implements EntryContainer {
     public synchronized void listen(FieldAddedOrRemovedEvent event) {
         // Rebuild entry editor based on new information (e.g. hide/add tabs)
         recalculateVisibleTabs();
-    }
-
-    @Subscribe
-    public synchronized void listen(EntryChangedEvent event) {
-        DefaultTaskExecutor.runInJavaFXThread(() -> sourceTab.updateSourcePane(entry));
     }
 
     /**
@@ -520,7 +513,6 @@ public class EntryEditor extends JPanel implements EntryContainer {
     }
 
     private void unregisterListeners() {
-        this.entry.unregisterListener(this);
         removeSearchListeners();
 
     }
--- a/src/main/java/org/jabref/gui/entryeditor/SourceTab.java
+++ b/src/main/java/org/jabref/gui/entryeditor/SourceTab.java
@@ -21,6 +21,7 @@ import org.jabref.gui.undo.NamedCompound;
 import org.jabref.gui.undo.UndoableChangeType;
 import org.jabref.gui.undo.UndoableFieldChange;
 import org.jabref.gui.util.BindingsHelper;
+import org.jabref.gui.util.DefaultTaskExecutor;
 import org.jabref.logic.bibtex.BibEntryWriter;
 import org.jabref.logic.bibtex.InvalidFieldValueException;
 import org.jabref.logic.bibtex.LatexFieldFormatter;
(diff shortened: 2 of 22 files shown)


halirutan:
I have several issues with this on my machine. 

1. Open Entry go to source tab and invoke Generate Key button or Quality -> Autogenerate BibTeX keys throws an exception and leaves everything in an undefined state where the source editor doesn't work anymore (likewise each external change like Clean Entry, Mark Entry, will 
2.  Using Change Entry Type button doesn't update the source view
3. Changing source to something invalid and moving to a different tab or entry silently reverts the entry to its original content (which would be OK for me and which makes life so much easier)
4. Undo doesn't work correctly. It kind of mixes reverting things I did earlier and throws an exception. In the end, it clears the source editor

Some of the things might be older issues but leaving the source editor unusable should be introduced by this PR.

tobiasdiez:
@halirutan Thank you for the intensive testing. 

- 1+4: Oh yes, again one of these "not on JavaFX thread"-related issues. Should be fixed now
- 2: Changing the entry type indeed does not update the source code. However, this also didn't worked before if I'm not mistaken.
- 3: I don't want to store invalid input. This would really complicate things.

tobiasdiez:
@lenhard Also thanks to you for the intensive testing! I've not realized how many bugs this PR actually fixes. The exception you get is also present in the current master https://github.com/JabRef/jabref/issues/3375. I actually cannot reproduce it, so its not possible to solve or analyze it (however: it seems to be a problem with the code editor itself, `flowless` is the library that provides the editor).
Thus I'm taking the liberty to merge the PR after the build is successful.

Sidenote: you are right, debugging these bindings is a nightmare...on the other hand, they are really powerful and make things easier (as the 2-way binding in this PR shows; setting up something similar with our event system would be way harder).

@halirutan I'm really sorry that my changes overlapped with your's. I read it to late that you already had a solution for some of these issues. Feel free to work on the editor, I'll stay put and work at other things now. (A more detailed communication on these matters is hard at the moment since my skype is not working due to the great firewall and vpn issues...)

sambo57u:
I just tried testing the latest build which I believe includes this change. Changes now a properly saved
in the entry editor except when one tries to put a curly bracket, e.g. for capital letters, I am getting an
what seems to be an inconsequential error message saying that "unable to parse the entry ....".
I think as soon as you type in a { if this is going to be stored then there is a mismatch in brackets.

jonasstein:
I tried this release too, here (#3383) it even "destroys" the source field until I restart Jabref. In the Example I documented there is also a `{` perhaps this is the root of evil.

tobiasdiez:
@sambo57u Thank you very much for trying the new build. If you add an additional { jabref can no longer parse the entry successful and thus does not stores the changes. But as soon as you close the bracket } JabRef should start to update the entry again. I just pushed a fix to the newest development version that makes the error message at the bottom disappear as soon as the code can be parsed again.

sambo57u:
Trying 4.1dev build 2017-11-01 04:42
There seems to be other problems creeping into the entry editor:
1. Changes are not saved at all
2. When the source editor is invoked by double clicking on an entry the clicking the mouse cursor somewhere in the source has no effect of moving the cursor (remains at the top left position). But actually it seems to have moved because if you type a letter it types it where you clicked the cursor and
at that point cursor also moves.

lenhard:
@tobiasdiez It seems that several other people are experiencing problems with the source tab that are affecting the complete entry editor, e.g. also https://github.com/JabRef/jabref/issues/3217#issuecomment-341149456 I guess the binding as it stands here does not work properly.

I think the merge of this might have been a little too early. Could you please investigate this issue (sort of soon, sorry), because the master branch is kind of broken at the moment. I am not sure, but maybe an alternative would be to revert the merge, reopen the closed issues linked above, and go for a new PR. After all, it might be better to have no sync between source tab and other tabs than to have no entry editor at all. But maybe the whole problem is easily fixed soon in a follow-up PR and then reverting would not be necessary.

tobiasdiez:
@lenhard Yes, sorry for merging a bit to early. I thought that since the index out of bounds exception was already a known issue, I did not introduce a new bug and thus everything is fine. But, as we learned now, the fixes in this PR actually make it so that users hit the IndexOutOfBounds bug way more often (since now the external updates actually work...). 

Hopefully https://github.com/JabRef/jabref/pull/3390 provides a temporary fix until https://github.com/FXMisc/RichTextFX/issues/637 is solved.

lenhard:
@tobiasdiez Thanks for the quick action :) I'll try to review the new PR as soon as possible and then we will hopefully be back in a working state soon.

sambo57u:
I am only seeing the 11-01 12;11 build on master. Testing that two problems remain, the mouse movement and nothing is saved. Even after making a change one does ctrl+s and sees a message that the bib file is saved, it is really not saved.

sambo57u:
As of 4.1dev 2017-11-03 07:17 version no change in the source editor is saved, plus the cursor issue.

lenhard:
@sambo57u As of now, the fix is not yet merged into master, so this is expected. If you want to try out the current version of the fix that seems to repair the saving, you can get it here: http://builds.jabref.org/storeSourceFix/

sambo57u:
Hi, unfortunately the version from http://builds.jabref.org/storeSourceFix/ have the same problem. This was fixed at some point except the issue of getting an error message when one tried to enter a curly bracket. At this point no changes are saved in my test.

lenhard:
Ok, I also have some update problems in that left. Looks like they won't get fixed as part of that PR. Well have to see if someone addresses them.

PR #3370, opened 2017-10-28
Refactorings

Siedlerchr:
Merge when build succesufl


Siedlerchr:
@tobiasdiez  I now moved the common code to an own method in a class in the gui.
It can't be moved to logic because it depends on ExternalFileType
I tried adding a test for it, but even with extensive mocking, ExternalFileTypes are again the problem - Because it's a singleton class...Mockito is not able to mock it.

tobiasdiez:
I think you included your autoset link fixes by accident.

Issue #2721, opened 2017-04-07
Set/clear/rename fields: append to a field


buhtz:
The dialog in `Quality -> Set/clear/rename fields` has no way to append a value to a field.

**e.g.** I want to add a new `keyword` to multiple entries without losing the existing keywords.

There is a [workaround](http://discourse.jabref.org/t/add-hyphenation-german-to-each-some-entries/493/7?u=buhtz) for that:

> Create a keyword-based group containing the expression you want to add and drag the entries onto this new group.

This works but is not the usual way of a user I think.

Anirudh257:
Hey, I am a newbie and I would love to help you out.

koppor:
@Anirudh257 Thank you for offering your help. Have you seen https://github.com/JabRef/jabref/blob/master/CONTRIBUTING.md? We hope that you have IntelliJ in place. This will help you much.

tobiasdiez:
@Anirudh257 As a starting point have look at the class https://github.com/JabRef/jabref/blob/7a1b15abb58016b40e21a72124edf9da17cd08f2/src/main/java/org/jabref/gui/actions/MassSetFieldAction.java that handles the "Set/clear/rename fields" functionality. If you need further help, just ask here (or create a PR with your initial ideas).

Anirudh257:
Sure, i will take a look into it.

PR #3395, opened 2017-11-03
Fix for issue #2721 append to a field

ethanwharris:
#2721 
Add the append option to the dialog.


src/main/java/org/jabref/gui/actions/MassSetFieldAction.java
```diff
@@ -235,6 +251,10 @@ public void actionPerformed(ActionEvent e) {
                 ce.addEdit(MassSetFieldAction.massRenameField(entryList, fields[0], renameTo.getText(),
                         overwrite.isSelected()));
             }
+        } else if (append.isSelected()) {
+            for (String field1 : fields) {
```
lenhard: Nitpick: `field1` is not really a good variable name here. The loop iterates over all fields and not just the first. Please rename this to `field`.

src/main/java/org/jabref/gui/actions/MassSetFieldAction.java
```diff
@@ -281,6 +301,31 @@ private static UndoableEdit massSetField(Collection<BibEntry> entries, String fi
...
+     * @param text            The value to set. A null in this case will simply preserve the current field state.
+     * @return A CompoundEdit for the entire operation.
+     */
+    private static UndoableEdit massAppendField(Collection<BibEntry> entries, String field, String text) {
+
+        if (text == null) {
+            text = "";
+        }
+
+        NamedCompound ce = new NamedCompound(Localization.lang("Append field"));
```
lenhard: Since this is new code, please do proper naming (and don't get inspired by all the bad naming in JabRef). So please call this `compoundEdit` (or something similar) instead of `ce`.

src/main/java/org/jabref/gui/actions/MassSetFieldAction.java
```diff
@@ -281,6 +301,31 @@ private static UndoableEdit massSetField(Collection<BibEntry> entries, String fi
...
+        for (BibEntry entry : entries) {
+            Optional<String> oldVal = entry.getField(field);
```
lenhard: Same here: Why not call it `oldValue`.

ethanwharris:
@lenhard Thanks for your review! Totally agree, have made changes as requested and updated some of the old stuff to match.

src/main/java/org/jabref/gui/actions/MassSetFieldAction.java
```diff
@@ -53,26 +53,30 @@
     private JRadioButton selected;
     private JRadioButton clear;
     private JRadioButton set;
+    private JRadioButton append;
     private JRadioButton rename;
     private JComboBox<String> field;
-    private JTextField text;
+    private JTextField setText;
```
LinusDietz: I know this has been there befor your PR, but I would  like to encourage you to change the naming of these JTextFields.

Background: In Java you usually use verbs or verbs+nouns for method names. When having some buttons in a GUI I can still understand that it is acceptable to name the variables after the button's functionality. However, when it comes to these Textfields, I would say a verb+noun combination is really misleading.

How about naming them something like `textFieldSet`, `textFieldAppend`, `textFieldRename`?

ethanwharris: @LinusDietz Thank you for your review. Those names are much better! Have made the changes as requested.

LinusDietz:
@ethanwharris nice! We'll merge this as soon as another maintainer approves.

ethanwharris:
@lenhard Thank you for your feedback. Have now updated the menu localization to retain original translation, have also reflected this in the JabRef localization.

lenhard:
@ethanwharris Ok, perfect! I've looked at the code again and this seems to be ready for merge now (which is what I will do right away).

Thank you very much for your contribution! Feel very much invited to contribute more :-)

PR #3398, opened 2017-11-04
Fix problems in source editor

tobiasdiez:
So here is now a real fix of these problems in the source editor (index out of bounds, paste, select all, update,...). The problem was that updates to the editor were not performed on the javafx thread although they were wraped in a `runInJavaFXThread` call. The reason was that this method would actually submit tasks to the executor, which would then start a new thread. So the fix only happens in `DefaultTaskExecutor`, the rest of the changes are minor cosmetic improvements to get rid of globals or base panels in the source tab.

### DIFF
--- a/src/main/java/org/jabref/gui/entryeditor/EntryEditor.java
+++ b/src/main/java/org/jabref/gui/entryeditor/EntryEditor.java
@@ -339,7 +339,7 @@ public class EntryEditor extends JPanel implements EntryContainer {
         tabs.add(new RelatedArticlesTab(Globals.prefs));
 
         // Source tab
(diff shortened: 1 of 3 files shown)


Siedlerchr:
Looks good to be, however I am wondering why the javafx thread check did not work as specified?

tobiasdiez:
It worked...if the call were made on the javafx thread, then the runnable was send to the executor. However, the executor starts a new thread for each submitted task.

Siedlerchr:
I tested it locally and apart from getting any error hint on invalid bibtex (which is simply reverted) it works as expected. So I merge this in now

PR #3399, opened 2017-11-04
Add simple gui test

tobiasdiez:
This PR adds a simple GUI test for the source editor, which I used to track down the index of out bounds exception. It uses [TestFX](https://github.com/TestFX/TestFX), which seems to be a good framework for javafx gui tests. For example, it is possible to run the unit tests in a headless method - thus allowing it to run on travis. For me, the biggest advantage is that you can isolate different parts of the system (i.e. only add the source tab) and ignore the rest of the application. Thus you can write "unit tests for ui".

Disclaimer: I will not invest any more work in these gui tests. But since some of you were quite in favor of having a gui test suite, I thought I share with you what I have so far.


Siedlerchr:
And please have a look at the errors, you need a BasePanel

koppor:
I think, it does not need any XServer running etc? Thus, I lean towards not including it in our existing GUI tests, but eitehr keep it in the main testing block or create a new category javafxguitests.

src/test/java/org/jabref/gui/entryeditor/EntryEditorTest.java
```diff
@@ -0,0 +1,66 @@
...
+    public Scene scene;
+    public CodeArea area;
+    TabPane pane;
+    SourceTab sourceTab;
+
+    @Override
+    public void start(Stage stage) throws Exception {
+        area = new CodeArea();
+        area.appendText("some example\n text to go here\n across a couple of \n lines....");
+        sourceTab = new SourceTab(new BibDatabaseContext(), new LatexFieldFormatterPreferences());
```
LinusDietz: The Argument CountingUndoManager is missing in the signature of this test. I have tried to add it by simply adding `new CountingUndoManager()`, but then the test fails. Can you have a look at it?

tobiasdiez:
It now compiles. However, I need to use the preferences in the test and thus had to add an exception to the architecture tests. Since this is often a critical area of discussion, I would like to have another reviewers' ok before merging. Thanks.

src/main/java/org/jabref/gui/entryeditor/SourceTab.java
```diff
@@ -71,7 +73,7 @@ private static String getSourceString(BibEntry entry, BibDatabaseMode type, Late
     private CodeArea createSourceEditor() {
         CodeArea codeArea = new CodeArea();
         codeArea.setWrapText(true);
-        codeArea.lookup(".styled-text-area").setStyle("-fx-font-size: " + Globals.prefs.getFontSizeFX() + "pt;");
+        codeArea.lookup(".styled-text-area").setStyle("-fx-font-size: " + preferences.getFontSizeFX() + "pt;");
```
Siedlerchr: Why not pass the font size directly as int parameter? Would also work for test

tobiasdiez: That would work but there are 2 other accesses to the preferences. Thus 3 new constructor arguments just to get rid of the preference class seems a bit too much for me.

tobiasdiez:
As before, I need this PR to continue working on the javafx migration of the maintable. Thus I merge this now but, of course, implement further feedback on how to improve the code. (For some reason the build on travis does not started for the merge commit...but it was green before..so hopefully the build still passes).

PR #3400, opened 2017-11-04
Fix static localized text

tobiasdiez:
I had some problems with not initialized resource bundles (an issue on my side), which lead me to discover a few instances where localized text were accessed statically. Such text would never be translated correctly. Should be fixed with this PR. (I didn't add a changelog entry since it is a really minor fix.)

### DIFF
--- a/src/main/java/org/jabref/gui/search/SearchDisplayMode.java
+++ b/src/main/java/org/jabref/gui/search/SearchDisplayMode.java
@@ -1,5 +1,7 @@
 package org.jabref.gui.search;
 
+import java.util.function.Supplier;
+
 import org.jabref.logic.l10n.Localization;
 
 /**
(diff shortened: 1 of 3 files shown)


halirutan:
The PR looks OK to me, although it's only cosmetics and not necessary. I just wanted to point out that these static texts should correctly be translated after my localization PR that cached the messages. The idea behind statically caching all messages was that it's anyway not possible to make a complete language change in the GUI without restarting JabRef. We always had the warning for the user to restart JabRef after setting a new language.

Back then, I saw these static initializations as well and the main issue with them was, that they were initialized before the localization framework was initialized with the correct language. Therefore, in former times there was a hidden, but non-critical bug. In main:

1. The preferences were initialized. In them, at least one of the files you fixed was statically instantiated.
2. Since the localization was not ready at this point, it used English as fallback
3. Therefore, these things were statically initialized with English everytime no matter of the user language

I, therefore, reordered the init phase so that the language and localization were ready when the first access to it happened. This ensures correct language for the static things I found.

Nevertheless, this is a good PR and although I haven't tested it, it looks correct. If you like, you can refactor the initialization phase again which means moving

```
Localization.setLanguage(prefs.get(LANGUAGE, "en"));
```

from `org.jabref.preferences.JabRefPreferences#JabRefPreferences` back to line 71 in `JabRefMain`.

Issue #3164, opened 2017-08-28
new entry displayed twice in sql database with jabref 4


napseis:
hi,

When connected to a remote database (mariadb), when I add an entry, the entry is always added twice. If i want to remove one, both are removed.

This happens with Jabref 4 final, but works ok in 3.8.2

Furthermore, entries added using version 3.8.2 appear correctly when opened with jabref 4, i don't see any duplicate.

tell me if i can provide something more.

Thanks

MarioLipinski:
I think this just a presentation bug after adding the entry to the database. It seems the entry is only added once.

napseis:
hi,
Maybe, but there is something more, as I said, when I create with 3.8, i see it only once when i open the database with 4 beta3

napseis:
It is indeed displayed bug. If database is opened with 3.x verion, entry shows up once. But it's quite annoying, in version 4 I have to close database and reopen it in order to fix the display. Pulling change from sql doesn't  refresh correctly.

mwessley:
Having the same Problem and sadly did not find a solution yet

mwessley:
Any Ideas on why this bug occurs so I could go on myself and fix this?

tobiasdiez:
I have no idea but can point you to the relevant code: since it appears to be a problem with the update of the maintable, I would start debugging https://github.com/JabRef/jabref/blob/ebbeb1d249edf7134abf92d24c2031ae54d19881/src/main/java/org/jabref/gui/maintable/ListSynchronizer.java (best set breakpoints in the listener methods). This class ensures that the backend (List<BibEntry>) is properly synchronized with what is shown (EventList<BibEntry>). The list synchronizer is bound to the backend here: https://github.com/JabRef/jabref/blob/c6f39cbab33e56bdc704d9575efc661b78c7b9c2/src/main/java/org/jabref/gui/BasePanel.java#L1231

koppor:
@obraliar Can you have a look, please?

obraliar:
@koppor yes :)
@napseis @mwessley I would appreciate if you could test this fix on your local machine.

napseis:
hi,
I tried JabRef 4.1-dev--snapshot--2017-11-07--master--efdf91a7d , and it persists, but i think the fix is not yet incorporated in this snapshot ?

koppor:
@obraliar What do you mean by "this fix"? 🕺

lenhard:
[...]

PR #3412, opened 2017-11-08
Fix event callback for shared databases

obraliar:
This PR fixes [issue 3164](https://github.com/JabRef/jabref/issues/3164).

The problem was about execution order and an event callback caused by `MetaData`.
Appropriate parameters were introduced by overloading affected methods.


src/main/java/org/jabref/model/metadata/MetaData.java
```diff
@@ -64,9 +64,15 @@ public MetaData() {
         return Optional.ofNullable(saveOrderConfig);
     }
 
-    public void setSaveOrderConfig(SaveOrderConfig saveOrderConfig) {
+    public void setSaveOrderConfig(SaveOrderConfig saveOrderConfig, ChangePropagation postChanges) {
```
lenhard: You are adding extending the method signatures in this core class, and I do not like that. The `ChangePropagation` parameter is added to every method and that pollutes the signatures. 

Please remove `ChangePropagation` from all method signatures. Instead, add an additional constructor for `MetaData` that takes the parameter and store it in an instance variable. This should have the same effect, but cause much less code changes.

obraliar: That was also my first idea but then I saw [MetaData.java#L249](https://github.com/JabRef/jabref/blob/master/src/main/java/org/jabref/model/metadata/MetaData.java#L249) which was already there and also uses `ChangePropagation` in this context. However now we are using a `set` method to avoid the signature pollution :).

lenhard:
Agreed, this is much more concise :) Good work!

When the test is there we can merge.

koppor:
DevCall decision: There won't be tests coming, but we need the fix. We merge to have it in the 4.1 release.

lenhard:
Not optimal, but if that was the decision during the call, then fair enough.

This is mergeable, I'll merge it now. Since @Siedlerchr only merged master and added a changelog entry, I guess it's ok to do a squash.

PR #3413, opened 2017-11-08
Fix some error prone warnings

stefan-kolb:


src/test/java/org/jabref/logic/xmp/XMPUtilTest.java
```diff
@@ -856,7 +856,7 @@ public void testReadMultiple() throws Exception {
      */
     @Test
     public void testWriteMultiple() throws IOException, TransformerException {
-        List<BibEntry> l = new LinkedList<>();
+        List<BibEntry> l = new ArrayList<>();
```
Siedlerchr: I know, is not your code, but when you are here, you could provide at least a better name for that list.

PR #3415, opened 2017-11-08
Adapt localization.gradle to python 3

### LINKED ISSUE #3414: Error in SyncLang.py
@LinusDietz  Could you please take a look?
```
$ ./gradlew localizationUpdate
Starting a Gradle Daemon (subsequent builds will be faster)
Task :jmh Last added: null
:localizationUpdate  File "E:\workspace\JabRef\jabref\scripts\syncLang.py", line 17
    def get_current_branch(self) -> str:
                                ^
SyntaxError: mismatched input '-' expecting COLON
 FAILED

```

LinusDietz:
Fixes #3414

### DIFF
--- a/localization.gradle
+++ b/localization.gradle
@@ -8,8 +8,9 @@ task localizationStatusVersion {
             commandLine "python3", "--version"
             standardOutput = os
         }
-        assert os.toString().startsWith("Python 3.")
-        assert Integer.parseInt(String.valueOf(os.toString().charAt(9))) >= 6
+        if(!os.toString().startsWith("Python 3.") || Integer.parseInt(String.valueOf(os.toString().charAt(9))) < 6){
+            throw new Exception ("You need at least Python 3.6 to run syncLang!")
+        }
     }
 }
 
--- a/localization.gradle
+++ b/localization.gradle
@@ -2,57 +2,47 @@ repositories {
     jcenter()
 }
 
-configurations {
-    jython
+task localizationStatusVersion {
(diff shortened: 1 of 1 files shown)


lenhard:
I just checkout this branch, ran `./gradlew localizationUpdate` and got the following error:
```
                                                               
* Where:                                                       
Build file 'C:\workspaces\jabref\jabref\build.gradle' line: 30 
                                                               
* What went wrong:                                             
A problem occurred evaluating script.                          
> A problem occurred starting process 'command 'python3''      
```
So this seems to not to be working yet.

LinusDietz:
Well, do you have python3 installed?I know, the solution before with jython was more elegant.The problem is: it doesn't support python3, meaning that it must be pre installed on the user's machine.If this is an infeasible option, I would have to Port it back to a jython-supported version.On Nov 9, 2017 09:16, Jörg Lenhard <notifications@github.com> wrote:I just checkout this branch, ran ./gradlew localizationUpdate and got the following error:
                                                               
* Where:                                                       
Build file 'C:\workspaces\jabref\jabref\build.gradle' line: 30 
                                                               
* What went wrong:                                             
A problem occurred evaluating script.                          
> A problem occurred starting process 'command 'python3''      

So this seems to not to be working yet.

lenhard:
Ok, I wasn't aware of that. In that case, I am strongly for reverting the changes and going back to the old localization script. Forcing people to install additional dependencies defeats the whole point of using a build system. Everything should be doable with only Java installed so that you can run gradle.

@JabRef/developers We need a decision on this.

Issue #3354, opened 2017-10-26
Position of Special Field Columns is reset


devBeginner:
JabRef version 4.1-dev on Windows

Steps to reproduce:

1. Activate the special fields (rank/priority/...)
2. Change the position of the columns in the table
3a. Go to preferences and click OK, then the columns are shifted to the right of the table
3b. Also after restarting jabref...

lenhard:
It is news to me that you should be able to configure the order of fields in the main table. @JabRef/developers Did a feature like this ever exist?

Siedlerchr:
Interesting. I never cared about this, but I could reproduce this. If I rearrange the columns, the order is not saved upon exiting. But as far as I remember, nobody touched the Maintable code for ages. 
Maybe there is/was a reason to not save this?

lenhard:
@Siedlerchr So this has actually been working at some point? You could actually rearrange columns just by dragging and JabRef would store that somehow?

Siedlerchr:
@lenhard  I actually have no idea if this ever worked. Mabye @koppor  remembers this

matthiasgeiger:
This should work for all "Standard" columns - However the configuration of the Special field order was never possible as they are treated completely different and I had not the nerves to refactor this shitload of messy code 😜

devBeginner:
Maybe it would be helpfull to shift them to the left (should not be as hard as treating them like the other columns).

devBeginner:
@lenhard You can rearrange the "non special" columns and then update the column order in preferences to the one you set up in the table. This is working well, but as mentioned, not for the special columns.

devBeginner:
Hey there,

I managed out, that the order of columns is given by the updateTableFormat() in org\jabref\gui\maintable\MainTableFormat.java.

By moving the code for the special columns after the URL_Column part, the special columns are placed right after url column, in case, that they are activated. I was able to test it locally and seems to work.

Siedlerchr:
Thanks for your analysis of the problem. So if the cols are after URL their position will be stored?

I would say, go for a PR so we can see the changes in the code

devBeginner:
As matthiasgeiger said, storing the position is still not possible. What happens is, that at redrawing the maintable, the special field columns are always added directly after the URL column. The question is, if anyone beside me finds it more convenient to have the columns added to the left.

I will open a PR today or tomorrow so you can see what I mean.

tobiasdiez:
[...]

PR #3416, opened 2017-11-08
Changed fixed position of special columns. #3354

devBeginner:
When activating the special columns in the preferences, all of them are added to the right of the maintable. Each restart of Jabref or rebuild of the maintable caused the columns to be added on the right side. #3354 

I changed the fixed position of the special columns from the right of the maintable to the right of the URL column. I think this position makes it easier to work with these columns, because you have interesting information like read status in the beginning of the table. 

I did this by rearranging the order in which the columns are added to the maintable. Because the special columns are treated different to the basic columns, you cannot change order within  preferences.

### DIFF
--- a/src/main/java/org/jabref/gui/maintable/MainTableFormat.java
+++ b/src/main/java/org/jabref/gui/maintable/MainTableFormat.java
@@ -113,19 +113,6 @@ public class MainTableFormat implements TableFormat<BibEntry> {
                 tableColumns.add(builder.createFileIconColumn(desiredColumn));
             }
         }
-
-        // Add 'normal' bibtex fields as configured in the preferences
-        // Read table columns from prefs:
-        List<String> colSettings = Globals.prefs.getStringList(JabRefPreferences.COLUMN_NAMES);
-
-        for (String columnName : colSettings) {
-            // stored column name will be used as columnName
-            // There might be more than one field to display, e.g., "author/editor" or "date/year" - so split
-            // at MainTableFormat.COL_DEFINITION_FIELD_SEPARATOR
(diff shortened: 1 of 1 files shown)


matthiasgeiger:
Just a quick remark from the Off as I do Not have the time to check.

I think due to a proposal by @koppor we intentionally moved the columns from the left to the far right side... We should check that

lenhard:
Let us discuss the positioning of the special fields among the developers during the next devcall.

koppor:
I think, we discussed that two years ago, but I don't remember. For me, moving it to the front is OK. But that might disturb new users of JabRef, because they first see the stars...

Do we need an additional preference? ^^

devBeginner:
I could be nice to have the choice/option to either insert them in the beginning or end of the table .  Another solution would be splitting the special columns and move only the stars/rating to the end of table.

lenhard:
@devBeginner We've now talked this through among the core developers.

Unfortunately, we've decided against this solution. Moving the fields to the front for all users is just too strong a thing to do and might upset users. After all, there might be people who are very happy with having them on the right side. A fixed hard-coded moving of the special fields is not the way to go.

This should be solved in an alternative way, either:
- Fixing the bug that prevents the manual resorting of special fields to be stored and saved among different settings
- Introducing a new preference option that manages the order of special fields as suggested by @koppor 

We know that these things are not trivial, but ultimately they are a better way to address the underlying problem.

You're very welcome to take one of these directions and you can do so in this PR if you want.

koppor:
@devBeginner Sorry not to hear any feedback from you. We will close this, but hope, you will continue working on it --> we will be happy to reopen then.

Issue #3310, opened 2017-10-16
Check for Java version


tobiasdiez:
Lately, we get a lot of issue reports that can be fixed with a simple update to a newer Java version. I propose that we add a check at the start that informs the user that his java version is outdated. (Maybe helpful https://stackoverflow.com/questions/2591083/getting-java-version-at-runtime or https://stackoverflow.com/questions/4638994/how-to-know-jdk-version-from-within-java-code). 

Moreover, if I remember correctly then the installer (on Windows) also installs the JRE. Is it possible to configure the installer to also update it, when an older version is already installed?

Finally, we should also display a warning message that JabRef does not support Java 9 at the moment.

lenhard:
I fully support this! It should be a must for the 4.1 release.

lynyus:
refs https://github.com/JabRef/jabref/issues/3073

halirutan:
@tobiasdiez Can you comment on

- Where should the information about min and max supported version number go? Preferences, Globals, build-settings.properties?
- What would be a good thing to track? Min-version, Max-version?
- I've seen issues where the Java FX version might be older than Java itself. I guess I read it on SO, but don't remember exactly. In the light of this, should we test this as well?
- How do we decide which versions are acceptable?

halirutan:
Some more info since java is not constant through different major versions and different vendors:

01:32 < halirutan> Hey. `System.getProperty("java.version")` gives me "1.8.0_144". Is this naming convention 
                   used on all systems and all java version? Is there a comparator available in the standard 
                   library for these version strings?
01:32 < Maldivia> halirutan: nope :D
01:33 < Maldivia> halirutan: it was changed in java 9 :D
01:33 < halirutan> Maldivia: Ahh.. so 10min of googling didn't failed me.
01:33 < Maldivia> halirutan: also, it's slightly different for IBM JDK
01:35 < halirutan> We kind of need to ensure not only that users have say 1.8 but even what patch they use.
01:35 < Maldivia> halirutan: why?
01:36 < halirutan> Maldivia: We ran into problems with JabRef if users has e.g. 1.8.0_74 and not say 1.8.0_144.
01:36 < halirutan> I didn't want to hack something myself to compare versions. That's why I asked.
01:37 < Maldivia> well, across all openjdk jvms, you should be able to use java.version
01:37 < halirutan> Slightly simpler question: is there a list of all java version numbers? :)
01:37 < Maldivia> no
[...]

PR #3418, opened 2017-11-09
Checking Java version on startup

halirutan:
This is not quite ready for review, but I appreciate some feedback. Since `localizationUpdate` seems broken, the messages are currently not added and I'm waiting for the fix.The logic behind the java check is as follows:

1. In `gradle.build` we can define the `minRequiredJavaVersion` which is added as a value to `build.properties` as this was the most fitting place and it makes the version check quite flexible. I did not use an upper bound for the version but a boolean `allowJava9` indicator. The reasoning behind it is that a developer who is actually working on the Java 9 support just has to change this value to make JabRef start.
2. I put the version check as a method into `BuildInfo`. Other possible places would be `Globals` or when we give the build-info as a parameter then `JavaVersionCheck` itself. Suggestions?
3. The version check is called pretty early in `main` (as soon as localization is set up). I didn't bother to put the single error dialog into a separate class. Therefore, the logic of checking, showing the message and probably exiting is done directly there. Is this acceptable?

The version check itself has the only requirement to successfully access `System.getProperty("java.version")`. I implemented a very *optimistic* check that doesn't fail if we cannot determine or compare the versions. Only if we are sure that the used version is not sufficient, it will fail.

This will solve
https://github.com/JabRef/jabref/issues/3310


src/main/java/org/jabref/JabRefMain.java
```diff
@@ -65,6 +69,25 @@ private static void start(String[] args) {
         }
 
         Globals.prefs = preferences;
+
+        // Check if we are running an acceptable version of Java
+        final BuildInfo buildInfo = Globals.BUILD_INFO;
```
tobiasdiez: I would extract this new code at least to a new method.

src/main/java/org/jabref/logic/util/JavaVersionCheck.java
```diff
@@ -0,0 +1,67 @@
...
+ * very early stage in the JabRef startup.
+ */
+public class JavaVersionCheck {
+
+    // We expect version in the form 1.8.0_101 and need . and _ as delimiter to separate numbers
+    private static final Pattern DELIMITER = Pattern.compile("[._]");
+    // Be adventurous and assume that we can access this property always!
+    private static final String JAVA_VERSION = System.getProperty("java.version");
+
+    private JavaVersionCheck() {
```
tobiasdiez: I would propose to make this class non-static. You then have a (public) constructor that accepts the current version string and a default constructor which uses `JAVA_VERSION`. In this way you can also easily write a test to verify the methods in this class.

halirutan: Yes, good catch.

src/main/java/org/jabref/JabRefMain.java
halirutan: A new method of `JabRefMain` or a different place?

tobiasdiez: In my opinion, the code can stay in `JabRefMain`.

src/main/java/org/jabref/JabRefMain.java
```diff
@@ -65,6 +69,25 @@ private static void start(String[] args) {
...
+                    Localization.lang("Your current Java version (%0) is not supported. Please install version %1 or higher.",
+                    JavaVersionCheck.getJavaVersion(),
+                    buildInfo.getMinRequiredJavaVersion()
+                    )
+            );
+            if (!buildInfo.isAllowJava9()) {
+                versionError.append("\n");
+                versionError.append(Localization.lang("Note that currently, JabRef does not run with Java 9."));
+            }
+            final JFrame frame = new JFrame();
```
lenhard: Since we are trying to move to JavaFX, would it be possible to display the error message with a FX window instead of a `JFrame`?

Siedlerchr: Have a look at the FXDialogService.class. there are already some predefined dialogs for fx

halirutan: @Siedlerchr @lenhard That was on purpose. I was using `JFrame` because what good will it be if I check the Java version, try to display a message and crash with an exception because FX is not available? :smile:

Can we safely assume the FXDialog will always work?

halirutan: @lenhard I made JabRef quit only when Java 9 is used. Otherwise it displays the message but continues

halirutan:
When I remember correctly, that *lgtm analysis thing* failed on every single PR I submitted so far..

src/main/java/org/jabref/JabRefMain.java
Siedlerchr: @halirutan  if no openjfx is avaiable it will crash nonetheless because the JabRef main inherits from Application (javafx.application.Application). That is the thing which you can't check for directly.
The only thing would be we use a "Pre-Main-Class" that does nit depend on javafx. But I am unsure if this would work

halirutan: @Siedlerchr Ha.. overlooked this. My genius idea was to put the test at the very beginning before any of the fx stuff is instantiated. OK, I'm looking into fx dialog tonight.

halirutan:
OK, after understanding the whole situation, I think this PR would have only one advantage: It could check if the user uses an old built of 1.8. Since JabRef doesn't even start for 1.6 or 9 or with missing FX, the Java check would fail to act on most of its target audience. The Java check needs to be run separately with a small program that runs on all JVMs.

On Skype @Siedlerchr suggested to put it in the installer. However, what happens if someone installs JabRef with an OK JVM and then upgrades Java to 9? Maybe this is acceptable but I wanted to point it out. In any case, as it stands the PR doesn't make sense. I will close this but leave the branch alive.

If someone is going to work on putting a check into the installer or in a small check program that is started prior JabRef, we have a starting point and `JavaVersion` doesn't need to be re-implemented completely.

tobiasdiez:
I think this PR is already a huge step in the right direction. Even if the current version don't catch a missing OpenFX or Java9, it does report an error in case somebody still uses an older Java8 version like 1.8.31 that is known to cause bugs in JabRef. Thus I would like to get this PR merged.

I agree it would be also nice to handle the OpenFX & Java9 cases, but since this requires additional overhead (as you wrote, a small luncher that then starts JabRef) I don't think this will happen soon.

halirutan:
@tobiasdiez Then, I could rework the Dialog to be replaced by an FX version. The question for the correct Java version we want to enforce still stands. I currently used 1.8.0_144 which is not the latest one but quite new. I additionally checked many issues to see which version reporters use. Many indeed use _152 while one guy still had _76 (or something close to that).

The big question that cannot be answered is, what version is recent enough to not present issues.. hmm..

tobiasdiez:
Yes, 1.8.0_144 should be fine (at least I'm not aware of any bug that affects us but was fixed in a later java version). So please convert the dialog to JavaFX and this can be merged in my opinion.

halirutan:
@Siedlerchr @tobiasdiez  Quick question. If I use the FX Dialog at this point, I have to run it in the FX thread with 

```
FXDialogService dialogService = new FXDialogService();                                           
DefaultTaskExecutor.runInJavaFXThread(                                                           
        () ->                                                                                    
        dialogService.showErrorDialogAndWait(Localization.lang("Error"), versionError.toString())
);
```

This, however, shows the dialog but doesn't wait for it to be closed and resumes to start JabRef. Additionally, it throws an exception

```
java.lang.NullPointerException: null
	at org.jabref.gui.FXDialog.setLocationRelativeToMainWindow(FXDialog.java:137) ~[JabRef-4.1-dev.jar:?]
	at org.jabref.gui.FXDialog.lambda$new$1(FXDialog.java:98) ~[JabRef-4.1-dev.jar:?]
...
```
since the rest of FX is not set up at this point. Any tips?

tobiasdiez:
Yes, just keep the current swing dialog 😄  If it would work instantaneously, then using JavaFX is of course preferable.

Siedlerchr:
Oh okay, then stick to the Swing stuff. No problem.

Issue #2594, opened 2017-02-28
Java 9/JDK9 compatibility - blocks: Java 11/JDK11 compatibility


Siedlerchr:
JabRef does not run on Java9

This issue collects all java 9 related problems. Starting point: http://discourse.jabref.org/t/cannot-start-jabref-3-7-3-6-using-java-9-on-ubuntu-16-04/361/8

## Not supported JREs

This presents some JRE outputs to support issue finding during search.

Keywords: Java9, Java 9, JRE9, JRE 9

```
openjdk version "9.0.1"
OpenJDK Runtime Environment (build 9.0.1+11-Debian-1)
OpenJDK 64-Bit Server VM (build 9.0.1+11-Debian-1, mixed mode)
```

## Development hints
- Check for dependenices: https://wiki.openjdk.java.net/display/JDK8/Java+Dependency+Analysis+Tool

Siedlerchr:
Apple/OSX compatibility:

> These new methods replace the functionality of the internal APIs contained in the OS X package com.apple.eawt which are not accessible by default in JDK 9. Note that the package com.apple.eio is no longer accessible and no public replacement exists in JDK 9.

http://openjdk.java.net/jeps/272

kaimast:
I get the following error when trying to run JabRef on Linux with OpenJDK9. Should I file a new ticket for this? 

>kai@chipad:~/Downloads$ java -jar ./JabRef-3.8.2.jar
Apr 17, 2017 10:42:01 AM java.util.prefs.FileSystemPreferences$1 run
INFO: Created user preferences directory.
10:42:02.130 [AWT-EventQueue-0] INFO  net.sf.jabref.migrations.PreferencesMigrations - Migrating old custom entry types.
10:42:02.389 [AWT-EventQueue-0] ERROR net.sf.jabref.FallbackExceptionHandler - Uncaught exception Occurred in Thread[AWT-EventQueue-0,6,main]
java.lang.NoClassDefFoundError: javax/xml/bind/JAXBException
	at net.sf.jabref.logic.importer.ImportFormatReader.resetImportFormats(ImportFormatReader.java:56) ~[JabRef-3.8.2.jar:?]
	at net.sf.jabref.JabRefMain.start(JabRefMain.java:78) ~[JabRef-3.8.2.jar:?]
	at net.sf.jabref.JabRefMain.lambda$main$0(JabRefMain.java:40) ~[JabRef-3.8.2.jar:?]
[...]

PR #3421, opened 2017-11-09
[WIP] First attempt for JDK11 compilation - #Java11

Siedlerchr:
This addresses https://github.com/JabRef/jabref/issues/2594.

I played around a bit with the jigsaw module stuff.
Does not yet compile complete, but it's a step in the right direction.
If anyone is interested in this stuff, please move ahead. 


koppor:
It seems, we are going to ship different jars for different JREs. Think, we have to continue working on https://github.com/JabRef/jabref/pull/3180 to get this done?!

Siedlerchr:
Yes, however, there is also the experimental gradle stuff to create multi release jar. https://github.com/FrVaBe/Java9/tree/master/multi-release-jar

build.gradle
```diff
@@ -34,9 +35,11 @@ group = "org.jabref"
 version = "4.1-dev"
 project.ext.threeDotVersion = "4.0.3.1"
 project.ext.install4jDir = hasProperty("install4jDir") ? getProperty("install4jDir") : (OperatingSystem.current().isWindows() ? 'C:/Program Files/install4j6' : 'install4j6')
-sourceCompatibility = 1.8
-targetCompatibility = 1.8
+sourceCompatibility = 1.9
+targetCompatibility = 1.9
```
koppor: I would still let it work on Java8. So no update.

lenhard:
One of our core problems here at the moment seems to be a split package problem, i.e. several of the modules we use (open office stuff, jgoodies, ...) export the same package from different modules. When that happens, Java will report an error of the form
```
Error:java: the unnamed module reads package com.sun.star.util from both ridl and unoil
```
I am still looking for a solution to this that is not too much effort. One seems to be to package the two modules together manually into a single module that JabRef could depend upon. That is absolutely no solution for us, because it means that we would have to repackage tons of outdated and no longer maintained libraries.

Siedlerchr:
Well, at least jgoodies will be removed by your other PR.
Stay some other moduels. According to the jdk9 documentaiton, when they are unnamed modules, they can still be split packages.
http://www.logicbig.com/tutorials/core-java-tutorial/modules/split-packages/

A better toolfor identifying split classes: 
https://github.com/AdoptOpenJDK/jsplitpkgscan

In general:
https://dzone.com/articles/the-legacy-developers-guide-to-java-9

lenhard:
A hands-on guide that has helped me to resolve some problems with internal accesses:

https://blog.codefx.org/java/java-9-migration-guide

florian-beetz:
Hey, I'm a student of @stefan-kolb and I'm planning to write my Bachelor's Thesis on Java 9 Migration. I'd like to have a look on this.

lenhard:
I am glad to hear that! Just a remark: Although we try to add module files here, the module system is not our first priority when it comes to Java 9. Getting rid of all accesses to internal libraries is the initial obstacle. So, as a start, you might want to look at that problem.

Siedlerchr:
With the removal of our customjfx component in the future #3779 we will be one step closer to no internal dependencies. I think we should consider basing this branch on the maintable beta?

build.gradle
```diff
@@ -24,7 +24,8 @@ plugins {
     id "de.sebastianboegl.shadow.transformer.log4j" version "2.1.1"
     id "com.simonharrer.modernizer" version '1.5.0-1'
     id 'me.champeau.gradle.jmh' version '0.4.3'
-    id 'net.ltgt.errorprone' version '0.0.13'
+    id 'com.zyxist.chainsaw' version '0.1.3'
+    //id 'net.ltgt.errorprone' version '0.0.13'
```
koppor: Can't we support error prone any more? This should be commented inside build.gradle

Siedlerchr: at that time it did not work. Therefore I uncommented that. But have no clue if it's just the version or module stuff.

Siedlerchr:
Another interesting aspect:  Dependencies on JAVA EE modules ( java.mail for example) etc  are soon no longer part of JAVA SE.  http://openjdk.java.net/jeps/320  The good news: Nearly all EE modules are available as reference implementation on github and can be integrated as an example.
https://javaee.github.io/javaee-spec/Specifications

florian-beetz:
I currently have a version that compiles with Java 9 in my fork, where I temporarily removed the dependencies with split packages.

However I've got some new problems:
* When I try to run JabRev with the gradle run goal, it seems like no resources can be read. Log4j prints a warning that it is unconfigured and the ResourceBundles are null.

* When I run the assemble goal and run the JAR manually JabRef starts, but opening FX dialogs does not work, because the `PostConstruct` annotation can not be found. It works with the `--add-modules=java.xml.ws.annotation` flag, but I can't add it to the module-info, because gradle depends on a library also providing these annotations. 

I'd really appreciate if someone could have a look into this.

Siedlerchr:
Hi, I am not that deep into that, but I think in such cases you need to specify
But this might help:
https://discuss.gradle.org/t/java-9-build-9-ea-170-issues/22767/8

florian-beetz:
Hey, thanks for your help.

I just tried that, but this seems to just add the modules to the module path during compilation, but not actually at runtime.

tobiasdiez:
@florian-beetz That are already good news. Nice work! 
Which class or library uses `PostConstruct`? Is it only [afterburner.fx](https://github.com/AdamBien/afterburner.fx/issues/75)? If yes, then I'll have a look at it in the coming days since I'm currently in the progress of partly rewriting that library.

florian-beetz:
@tobiasdiez Thank you.
As far as I could tell it is afterburner.fx and Guava, that have a dependency on `com.google.code.findbugs:jsr305`.

tobiasdiez:
At least for guava, this should  be fixed if I understand https://github.com/google/guava/issues/2782#issuecomment-293444426 correctly.

florian-beetz:
I was able to fix the JSR 305 problem now.
Guava's dependency on JSR 305 is actually optional, but was made a runtime dependency due to some special use case (https://github.com/google/guava/issues/2721), so explicitly excluding the dependency helped.

Also there is currently an open Guava issue to replace JSR 305 completely https://github.com/google/guava/issues/2960

stefan-kolb:
@florian-beetz Nice 👍 Would you like to share your current progress via a PR with us?

florian-beetz:
@stefan-kolb Sure, should I just open a new one? I can't seem to add my changes to this one.

stefan-kolb:
I guess so, we can close this one then in favor of your new one!

Siedlerchr:
@stefan-kolb  Maybe we can give him push rights for the branches here?

stefan-kolb:
Is that possible? Would also be fine but I guess it complicates things unecessesarily.

florian-beetz:
I created a new pull requests including the changes of this one: https://github.com/JabRef/jabref/pull/3888.

Siedlerchr:
@florian-beetz  @stefan-kolb  I added him as a contributor with write permission. So he now can push to that branch here. Makes it easier for us to check it out

florian-beetz:
@Siedlerchr Thanks. I just closed the other PR and pushed to this one.

build.gradle
```diff
@@ -168,7 +185,8 @@ dependencies {
     testCompile 'org.reflections:reflections:0.9.11'
     testCompile 'org.xmlunit:xmlunit-core:2.5.1'
     testCompile 'org.xmlunit:xmlunit-matchers:2.5.1'
-    testCompile 'com.tngtech.archunit:archunit-junit:0.5.0'
+    // TODO: has a split package between archunit and archunit-junit
+//    testCompile 'com.tngtech.archunit:archunit-junit:0.5.0'
```
Siedlerchr: I just discovered that they  converted that to  java9 compatibility in a later snapshot:  https://github.com/TNG/ArchUnit/issues/47

build.gradle
```diff
@@ -133,12 +133,13 @@ dependencies {
...
     //compile 'de.codecentric.centerdevice:javafxsvg:1.3.0'
     compile 'de.jensd:fontawesomefx-materialdesignfont:1.7.22-4'
+    compile 'de.jensd:fontawesomefx-commons:9.1.2'
     compile 'de.saxsys:mvvmfx-validation:1.7.0'
     compile 'de.saxsys:mvvmfx:1.7.0'
     compile 'org.fxmisc.easybind:easybind:1.0.3'
     compile 'org.fxmisc.flowless:flowless:0.6.1'
     compile 'org.fxmisc.richtext:richtextfx:0.9.0'
-    compile 'com.sibvisions.external.jvxfx:dndtabpane:0.1'
+//    compile 'com.sibvisions.external.jvxfx:dndtabpane:0.1'
```
Siedlerchr: Dnd TabPane has been introduced offically with java 10:
https://bugs.openjdk.java.net/browse/JDK-8190717

florian-beetz: Good to know! Java 9 removed some (internal) classes this library uses. 
I don't think we'll be able to have the D'n'd feature in a Java 9 release unless someone patches the library.  But as Java 9 is already superseded, I guess this is not really high priority.

Siedlerchr: Agree with you. As soon we have it running with java 9, we can focus on the migration to java 10 ;)

src/main/java/module-info.java
```diff
@@ -0,0 +1,70 @@
...
+	requires java.xml.bind; // Deprecated!
+    requires jdk.xml.dom;
+
+    // Annotations (@PostConstruct)
+    requires java.annotation;
+
+    // Microsoft application insights
+    requires applicationinsights.core;
+
+    requires glazedlists.java15;
```
Siedlerchr: I think this one of the dependencies we hopefully can remove soon. Old swing stuff.

florian-beetz:
## Remove illegal reflective access
Java 9 does not allow reflective access into different modules at all unless they explicitly "open" packages for reflective access (see bcfde5666abd9b8d88df627585e4d2ad206868f2).
* [ ] `org.jabref.gui.maintable.SmartConstrainedResizePolicy`
This class determines how the Maintable columns are resized and accesses to non-visible members. I'm pretty clueless how to solve this issue. 
* [ ] `org.jabref.gui.groups.GroupTreeView`
Makes a ControlsFX `CustomTextField` clearable. There is already an [issue](https://bitbucket.org/controlsfx/controlsfx/issues/330/making-textfieldssetupclearbuttonfield) for this problem at ControlsFX.
* [ ] `org.jabref.gui.GUIGlobals`
Sets X11's WM_CLASS using internal AWT API. Related OpenJDK [Issue from 2007 🙄](https://bugs.openjdk.java.net/browse/JDK-6528430)
* [ ] `org.jabref.gui.search.GlobalSearchBar`
Accesses the autocompletion popup of the search bar.
## Upgrade to newer library versions
* [ ] **LibreOffice UNO API**
The 4 LibreOffice JARs we use all export the package `com.sun.star`. I currently have an [open issue](https://bugs.documentfoundation.org//show_bug.cgi?id=117331). 
A simple workaround would be to repackage the 4 JARs into a single JAR.
* [ ] **latex2unicode**
By convention, Scala includes the Scala version in artifact names, separated with `_`, but `xxx_2.12` is converted to the module name `xxx.2.12` which is not a valid module name. This could be easily solved by adding an [`Automatic-Module-Name`](http://branchandbound.net/blog/java/2017/12/automatic-module-name/). latex2unicode has dependencies on other Scala libraries with the same problem.
Relevant issues: https://github.com/tomtung/latex2unicode/issues/10, https://github.com/lihaoyi/fastparse/issues/184
* [ ] **ApplicationInsights**
ApplicationInsights exports the Guava package `com.google.thirdparty`. This is [planned to be fixed in the next version](https://github.com/Microsoft/ApplicationInsights-Java/issues/661).
* [ ] **Guava**
Guava's dependency on the (unofficial) JSR 305 annotations. This is kind of a non-issue as the dependency can be excluded in the build script, but Guava will probably exclude JSR 305 in a future version.
* [ ] **ArchUnit**
ArchUnit has a split package. There is already a SNAPSHOT release compatible with Java 9.
* [ ] **JavaFX SVG**
JavaFX SVG has a dependency on Apache Batik, which has a split package (`org.w3c.dom`) with the `java.xml` module.
As far as I can tell, the only SVG used is the JabRef logo in the about dialog, so the simplest solution would be to replace the image with a rasterized version or maybe we could use [SVGPath](https://docs.oracle.com/javase/8/javafx/api/javafx/scene/shape/SVGPath.html).
* [ ] **ControlsFX**
ControlsFX has only partial support for Java 9, but they are [working on it](https://bitbucket.org/controlsfx/controlsfx/issues/780/controlsfx-should-support-java-9).
## Misc/Minor
* [ ] **DndTabPane**
The library is not compatible with Java 9, but the functionality is implemented in Java 10 natively anyway.
* [ ] **Update Start Script**
Java 9 has some new command line arguments, so the start scripts that gradle generates, needs to be updated.
I currently run JabRef with
```bash
java \
    --illegal-access=debug \
    --add-opens javafx.swing/javafx.embed.swing=org.jabref \
    --add-opens java.desktop/java.awt=spin \
    --add-opens javafx.controls/javafx.scene.control=org.jabref \
    --add-exports javafx.base/com.sun.javafx.runtime=controlsfx \
    -p . \
    -m org.jabref/org.jabref.JabRefMain
```
* [ ] **Spin**
Spin accesses internal API in AWT. Since spin is a swing library, I assume it will be removed in the future anyway.
* [ ] **Citation Styles**
Currently no citation styles can be loaded and I am not really sure why.

florian-beetz:
> Latex2Unicode: I think the easiest option is, if you directly provide a PR at that repo. ;)

I hoped I could avoid digging too deep into SBT, but I guess that is the best option :) Also the development of fastparse seems kind of inactive.

I opened PR #4035 for the SVG and #4034 for WM_CLASS.

I think the problem with the citation styles is related to how Java 9 loads resources from different modules, might also be a problem that all citation styles are in the default package, but I'm not sure yet.

Siedlerchr:
I just updated the maintable-beta branch gradle file with the latest version. Archunit is now 0.8.0 and the release should be compatible with java 9

florian-beetz:
@tobiasdiez Yes, we can ignore that for now. In fact, that's what I'm currently doing with the `--add-opens` in my launch script.

> I don't know of a good way to remove the reflection code.

Good to know that I'm not the only one :)

Siedlerchr:
@florian-beetz  we recently changed the postgressql driver to a new one, so you may need to update the module info

florian-beetz:
@Siedlerchr Already noticed that :) Apparently they also have no Automatic-Module-Name, [yet](https://github.com/pgjdbc/pgjdbc-parent-poms/pull/13), so that is another library that needs updating.

Also the 2.1.1 of ApplicationInsights apparently still has the conflict with Guava.

tobiasdiez:
So the current status is that we have to wait until all libraries listed above are updated?

florian-beetz:
If we do not want to rely on "hacking" the module system with the `--add-*` switches: Yes.

Siedlerchr:
For JDK10 we definitely can't use some of the hacking switches. But for a jdk9 release I think it would be okay. I would propose we release the maintable-beta with jdk9.  In the meantime I would suggest trying to get rid of the old swing libs (spin etc)

Siedlerchr:
The Spin library sounds really obsolete to me, but we use it in our AbstractWorker. 
But I am not into that Swing Threading stuff in detail. Maybe @lenhard  can add some information on this. 
https://sourceforge.net/projects/spin/

florian-beetz:
I had a discussion with @stefan-kolb today about releasing Java 9 compatible versions of JabRef.

Currently, we are releasing exploded fat JARs, where all dependencies are repackaged into it. I think this will become a problem as soon as more libraries start adopting Java 9 with a `module-info.java` because this is placed in the root directory of the JAR, so merging this will probably become difficult. Apart from that, an exploded fat JAR is also loaded as one module, so this is also kind of working around the module system (but will probably solve some of the above problems above, as I discovered those by testing with the non-fat JAR).

Also, I don't know if everyone is aware of this, but the version in this PR is incompatible with Java 8. This is because ControlsFX 9.0.0 is already compiled with Java 9, so it can not be run using Java 8, same goes for JabRef itself when we include a `module-info.java`.

We also talked about using [jlink](https://steveperkins.com/using-java-9-modularization-to-ship-zero-dependency-native-apps/) to bundle the necessary JRE modules for releases. I think if we don't use jlink, we should either ship the dependencies in separate JARs as in the regular (non-fat) distribution package or build an "EAR-style" fat JAR, where this JAR contains the dependencies again as JARs.

What do you think?

stefan-kolb:
Shipping JR with the runtime could avoid/make it easier to track bugs like #3943

Siedlerchr:
I see no problem, at least for widows we can just put the libs next to the jar.
For Mac the dmg is probably similar. 

But what about Linux users? I think the gradle plugin we use for shadow build can generate a tar file which includes the libs and a bash script. Maybe this is an option?

florian-beetz:
Yes, the `assemble` target already produces a zip and a tar file, containing all libraries and run scripts for both Linux and Windows. That's what I'm currently using to test the module stuff.

florian-beetz:
Just a quick status update:

So, since I have to wait right now for updates of libraries to be released, I had a look into making #3704 work with JPMS.
Seems like the chainsaw plugin for Gradle cannot correctly build multi-module projects right now: https://github.com/zyxist/chainsaw/issues/34
There is still the possibility to just do [everything manually](https://guides.gradle.org/building-java-9-modules/).

I also had a look at removing Spin, but this seems like a really big task on its own and I think this needs some rework of the JabRef's threading model in general.

And apparently, I was a bit quick with suggesting using jlink, as it requires *every* module to have a real module descriptor and not just an automatic module name. I guess that will take some time until every library fully supports Java 9.

stefan-kolb:
> I also had a look at removing Spin, but this seems like a really big task on its own and I think this needs some rework of the JabRef's threading model in general.

This would be an important change for JabRef 😄

stefan-kolb:
@florian-beetz can you please base all PRs on master now. maintable-beta is now master.

stefan-kolb:
Application Insights should be fixed?! https://github.com/Microsoft/ApplicationInsights-Java/pull/683

florian-beetz:
@stefan-kolb Yes, but version 2.1.2 is not yet released.

tobiasdiez:
@florian-beetz could you please give an update what still needs to be done a) to have a workable release with Java 9 and b) additional nice-to-have things (which are not essential for a successful build). Thanks!

florian-beetz:
TL;DR: 
* see [comment above](https://github.com/JabRef/jabref/pull/3421#issuecomment-389229101)
* **working release**: 
  * latex2unicode dependencies are apparently abandoned, distribute manually patched versions with JR or maintain a fork
  * citationstyles should probably be a git submodule, because the encapsulation is wierd with resource JARs
* **nice to have**: 
  * repackage LibreOffice and distribute it with JR as a single JAR, or disable the integration for this release
  * fix internal access, for now it can be allowed with command line flags
  * release with Java 10 so D'n'D works in the main table
* distribution of fat JARs vs non-fat JARs needs to be looked into

---

@tobiasdiez I took a bit of time off after I finished my Bachelor Thesis on the migration and let watching the issues a bit slack, so I hope I can recall everything correctly.
I think my [comment above](https://github.com/JabRef/jabref/pull/3421#issuecomment-389229101) is still up to date.

I think the issues with the highest priority are the latex2unicode issues and the citation styles:

The issue in latex2unicode itself is finished, but I guess the maintainer of its dependencies fastparse and sourcecode is no longer active. I had a discussion with @stefan-kolb about that a while ago and we discussed the possibilities to either rewrite latex2unicode in Java, as it is actually not much the library does and we already have dependencies on parser generators anyway, use something like Jitpack for building my forks, or patching the libraries manually and distributing them in the `libs/` directory, because all that needs to be done is setting the `Automatic-Module-Name` in the JAR manifest.

I think the issue with the citation styles being unable to be loaded is caused by the new rules of encapsulation and this can not be easily bypassed, all styles are in the default package and that can not be exported and additionally no empty packages (packages without classes, resources apparently do not count) can also not be exported. I think the easiest solution would be to simply remove the dependencies and add the repositories they are based on as submodules to be included as resources in the JabRef JAR itself.

The issues with internal access are not essential for a successful build, as it can be allowed with the command line flags above.
That the D'n'D does not work with Java 9 is also a minor issue, I guess.

If you think the LibreOffice integration is an essential feature that needs to be in the release, I think the easiest solution is to manually repackage the 4 JARs into one and distributing it also in the `libs/` directory. Otherwise, I've disabled it in this PR completely.

Lastly, the building of the fat JAR needs to be looked into (see [comment above](https://github.com/JabRef/jabref/pull/3421#issuecomment-392746603)), because I don't know how duplicate files in multiple dependencies (`module-info.class`) are handled. Maybe there is some way to merge the module descriptors, but I haven't looked into that. Just shipping as non-fat JARs should be possible, though.

I hope to continue contributing when I'm back at university mid October. Also let me know if anyone is interested in my Bachelor Thesis on my work here and I will send you a copy.

tobiasdiez:
Thanks for detailed status report! So in general we are not far away from a workable solution. Nice! I pinged the guys from farstparse and sourcecode - hopefully they include your PRs soon (there was some activity recently in these projects). That leaves:
- [ ] Include citationstyles as submodule
- [ ] Merge the 4 LibreOffice jars (the feature is not really essential, but I don't see them fixing the issue soon)
- [ ] Have a workable fat jar (this is essential!)

It would be nice if you could work on this issues in the future. But no hurry, for now enjoy your free time!

florian-beetz:
I will look into it, but I'm pretty busy atm. so I don't know when I'll find time for it.

Sadly, I also didn't really have time to work on any other open points.

Siedlerchr:
@florian-beetz  I just merged the current master into the branch and I hope I did not destroy all things with the merge conflicts. It would be really nice if you could look over it.
As far as I see the LibreOffice jars problem is not yet resolved. What can we do about this? put them all in a single jar?

florian-beetz:
@Siedlerchr Thanks for merging. I just checked it out. There were some minor issues I fixed (error prone compiler plugin was missing from build script and the com.jfoenix module was not required)

I left the LibreOffice issues unfixed in my commit, but I think the easiest solution would be to repackage all four JARs into a single one and assign it a module name like `org.jabref.thirdparty.libreoffice` or something.

florian-beetz:
I've written a Gradle task to bundle LibreOffice as one JAR. I hope I found all the places where I commented out LibreOffice. This version now is runnable again, but I could not test the LibreOffice integration in depth.

Sadly, Gradle apparently does not notice dependencies that are generated during the build, because dependency resolution is already completed, so the task has to be executed manually when the LibreOffice dependencies change. This is also why I checked in the generated libreoffice.jar.

Siedlerchr:
As the Libre office jars don't change that often it's  not a problem to execute the task manually if we document that somewhere.

So one of the biggest things missing currently is the latex2unicode stuff

tobiasdiez:
For the latex2unicode, @florian-beetz could you please update your PR https://github.com/lihaoyi/fastparse/pull/185. The other one was already merged, so that it's hopefully only a matter of time until they publish a working version. Thanks!

Siedlerchr:
@florian-beetz  There also seems to be some errors with the testCompile, travis reports this errors for nearly every module;
`error: the unnamed module reads package com.github.jknack.handlebars.helper from both handlebars.helpers and handlebars`

tobiasdiez:
I've invested a bit of time today, and got JabRef to compile with Java 11 with only a few minor glitches. The `gradle run` task now works. (In the process, I had to make the cli a bit more robust regarding unknown options.)

Left to do:
- [ ] Use latex2unicode again
- [ ] Fix logger appender (see error message when starting JabRef)
- [ ] Remove additional command line interface (not our fault: https://github.com/java9-modularity/gradle-modules-plugin/issues/65)

Siedlerchr:
I would have removed the module info stuff completely and only go with the classpath

build.gradle
```diff
@@ -123,25 +148,23 @@ dependencies {
 
     compile 'com.google.guava:guava:27.1-jre'
 
+    compile group: 'javax.annotation', name: 'javax.annotation-api', version: '1.3.2'
+
     // JavaFX stuff
-    compile 'de.jensd:fontawesomefx-commons:8.15'
-    compile 'de.jensd:fontawesomefx-materialdesignfont:1.7.22-4'
-    compile 'de.saxsys:mvvmfx-validation:1.8.0'
+    compile 'de.jensd:fontawesomefx-commons:9.00'
```
Siedlerchr: Probably you can use 11.0

src/test/java/org/jabref/logic/l10n/LocalizationParser.java
```diff
@@ -223,6 +226,17 @@ public boolean containsKey(String key) {
                      .collect(Collectors.toList());
     }
 
+    private static void setStaticLoad(FXMLLoader loader) {
+        // Somebody decided to make "setStaticLoad" package-private, so let's use reflection
+        try {
+            Method method = FXMLLoader.class.getDeclaredMethod("setStaticLoad", boolean.class);
```
Siedlerchr: Maybe we can add our arguments to the PR discussion? https://github.com/javafxports/openjdk-jfx/pull/44/files

src/test/resources/org/jabref/logic/exporter/MsBibExportFormatTest2.xml
```diff
@@ -1,21 +1,21 @@
...
-<b:URL>http://dx.doi.org/10.1007/BF03250842</b:URL>
-<b:DOI>10.1007/BF03250842</b:DOI>
-<b:Author/>
-<b:Pages>237-248</b:Pages>
-<b:JournalName>Wirtschaftsinformatik</b:JournalName>
-<b:Number>3</b:Number>
-<b:City>a</b:City>
-<b:ThesisType>type</b:ThesisType>
-</b:Source>
+    <b:Source>
```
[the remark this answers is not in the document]
tobiasdiez: These changes actually fix most of the failing xml export test. The few test that are still failing are due to the missing latex2unicode.

tobiasdiez:
The build now passes and thanks to @siordache a "custom runtime image" is successfully created using `gradlew jlinkZip`.  It is also possible to generate a launchable exe file using `gradlew jpackage` after downloading the [preview version of jdk 13](https://jdk.java.net/jpackage/). The same command should also create an installer, but that requires to install [Inno Setup](http://www.jrsoftware.org/isinfo.php) or something comparable; [details about jpackage](https://blogs.oracle.com/jtc/a-brief-example-using-the-early-access-jpackage-utility).

It would be nice if somebody (@Siedlerchr @koppor ?) could take over from here and create the installer. Either by changing the previous install4j setup (which [apparently supports Java 11](https://www.ej-technologies.com/products/install4j/whatsnew7.html)) or by playing around with `jpackage` (which is probably the best solution in the long-term). 

Not sure how to proceed from here. Codewise this PR is almost finished (needs a bit of cleanup). I see the following options:
- create installer and deployment pipeline to distribute the custom runtime image (which includes all necessary java and javafx libraries), and then completely switch to java 11 
- maintain master (java 8) and this branch (java 11)
- investigate if we can distribute multi-release builds supporting java 8 and java 11 at the same time

Personally, I prefer the first option because we then don't have to worry about missing javafx libraries or strange copies of jre on mac (or similar version problems that we had before).

Siedlerchr:
@tobiasdiez  I will look at the installer. We just need to have an instance of the jdk11+openjfx11 somewhere avaiable and we need to make sure that libfxwebkit etc are included.

tobiasdiez:
Thanks. The image created by jlink contains everything that is needed: jre, JavaFX and all libraries.

tobiasdiez:
@Siedlerchr https://github.com/Santulator/Santulator/blob/master/package/PACKAGING.md might be helpful

Siedlerchr:
I am confused with the installer stuff atm, but it should also be possible to use the jlink stuff in install4j
However @tobiasdiez  jlink fails at my sys: (running at my sys jdk11)

```
> Task :jlink FAILED
Error: Modules org.jabref.merged.module and javafx.graphics export package javafx.scene.text to module org.jabref
```

homocomputeris:
@florian-beetz [ContorsFX 11](https://github.com/controlsfx/controlsfx) claims to have Java 9 support, maybe that checkbox can be ticked, eventually?

Siedlerchr:
@homocomputeris  Thanks for the info. We are aware of that and in generak JabRef runs under jdk11.
What is missing is the creation of the installer.

Siedlerchr:
Seems like the new install4j8 has better suport for modules jdlk:
https://www.ej-technologies.com/products/install4j/whatsnew8.html

Siedlerchr:
@tobiasdiez  I tried to merge the current version but seems like I broke the build and we have some split package or so again. Any idea where it might come from?

gradlew.bat
```diff
@@ -14,7 +30,7 @@ set APP_BASE_NAME=%~n0
 set APP_HOME=%DIRNAME%
 
 @rem Add default JVM options here. You can also use JAVA_OPTS and GRADLE_OPTS to pass JVM options to this script.
-set DEFAULT_JVM_OPTS=
+set DEFAULT_JVM_OPTS="-Xmx64m" "-Xms64m"
```
Siedlerchr: We definitely should add more memory.
This is some ugly gradle behavior that they changed memory to 64m

koppor: This is `gradlew.bat` distributed by Gradle Inc. This are the options for JabRef compilation. I did not notice any difficulties when compiling JabRef with the *default* distributed by Gradle Inc. Did you?

In other words: This is not JabRef's setting.

In case, there are difficulties, we should set `GRADLE_OPTS` (as indicated at the comment) since I don't want to patch a generated file.

koppor:
What about the checks at https://github.com/JabRef/jabref/pull/3421#issuecomment-389229101? Are they all resolved? - I think, we either added workarounds (https://github.com/JabRef/jabref/pull/4431) or just removed the feature (https://github.com/JabRef/jabref/pull/4694). @florian-beetz Could you provide us with an update.

koppor:
I updated to Install4J 8. As soon as I have a new key, I can fill the variable `INSTALL4J8_KEY` and this PR should be good to go.

New: The JRE is contained within the `exe`/`dmg` file. This is good, because we do not need to keep up https://files.jabref.org/jres/.

Siedlerchr:
We still have the issue with the split package. There is some dependency which needs to added as exception again. jdeps coud help to identify the root cause

koppor:
@Siedlerchr What is the symptom of a "split package"? JabRef available at http://builds.jabref.org/jdk9/ starts fine. Would it be possible to merge this PR and fix the split modules later?

src/main/java/org/jabref/gui/remote/JabRefMessageHandler.java
```diff
@@ -9,20 +9,29 @@
...
+            List<ParserResult> loaded = argumentProcessor.getParserResults();
+            for (int i = 0; i < loaded.size(); i++) {
+                ParserResult pr = loaded.get(i);
+                boolean focusPanel = i == 0;
+                Platform.runLater(() ->
+                        // Need to run this on the JavaFX thread
+                        JabRefGUI.getMainFrame().addParserResult(pr, focusPanel)
+                );
+            }
+        } catch (ParseException e) {
```
Siedlerchr: This must be from our package in org.jabref.logic.importer

Siedlerchr:
@koppor  https://www.logicbig.com/tutorials/core-java-tutorial/modules/split-packages.html
The problem is that we have someohow two packages that read from handlebars
https://blog.codefx.org/java/java-9-migration-guide/#Split-Packages

src/main/java/org/jabref/gui/remote/JabRefMessageHandler.java
```diff
@@ -18,20 +17,17 @@
...
-                boolean focusPanel = i == 0;
-                Platform.runLater(() ->
-                        // Need to run this on the JavaFX thread
-                        JabRefGUI.getMainFrame().addParserResult(pr, focusPanel)
-                );
-            }
-        } catch (ParseException e) {
-            LOGGER.error("Error when parsing CLI args", e);
+
+        ArgumentProcessor argumentProcessor = new ArgumentProcessor(message, ArgumentProcessor.Mode.REMOTE_START);
```
tobiasdiez: Now we have the problem: /home/travis/build/JabRef/jabref/src/main/java/org/jabref/gui/remote/JabRefMessageHandler.java:21: error: unreported exception ParseException; must be caught or declared to be thrown
        ArgumentProcessor argumentProcessor = new ArgumentProcessor(message, ArgumentProcessor.Mode.REMOTE_START);

Siedlerchr: This is totally odd. I first replaced it with our ParseException. Then it complained that the exception is never thrown. So I removed it. Makes no sense

tobiasdiez: It does not throw our Parse Exception but the one from apache cli.

build.gradle
```diff
@@ -191,9 +191,9 @@ dependencies {
     compile group: 'com.microsoft.azure', name: 'applicationinsights-logging-log4j2', version: '2.4.0'
 
     testCompile 'junit:junit:4.12'
-    testImplementation 'org.junit.jupiter:junit-jupiter:5.5.1'
-    testRuntimeOnly 'org.junit.vintage:junit-vintage-engine:5.5.1'
-    testCompile 'org.junit.platform:junit-platform-launcher:1.5.1'
+    testImplementation 'org.junit.jupiter:junit-jupiter:5.4.2'
```
Siedlerchr: From the 5.5 release notes:
https://junit.org/junit5/docs/current/release-notes/index.html#release-notes-5.5.0
>Explicit Java module descriptors

Probably we need to import the modules

Siedlerchr: https://github.com/junit-team/junit5/blob/master/README.md#junit-platform

tobiasdiez: Yap, should work now.

build.gradle
```diff
@@ -191,9 +191,9 @@ dependencies {
     compile group: 'com.microsoft.azure', name: 'applicationinsights-logging-log4j2', version: '2.4.0'
 
     testCompile 'junit:junit:4.12'
```
Siedlerchr: I guess this can removed now?

tobiasdiez: You can try it, but we are still using some of the old junit4 parameterized tests

jabref.install4j
```diff
@@ -37,155 +36,157 @@
...
+      <variable name="versionFourDots" value="3.1.0.0" />
+      <variable name="buildFileName" value="JabRef-3.1dev.jar" />
+      <variable name="version" value="3.1dev" />
     </variables>
-    <mergedProjects />
-    <codeSigning macEnabled="false" macPkcs12File="" windowsEnabled="false" windowsKeySource="pkcs12" windowsPvkFile="" windowsSpcFile="" windowsPkcs12File="" windowsPkcs11Library="" windowsPkcs11Slot="">
-      <windowsKeystoreIdentifier issuer="" serial="" subject="" />
-      <windowsPkcs11Identifier issuer="" serial="" subject="" />
-    </codeSigning>
+    <jreBundles jdkProviderId="Liberica" release="OpenJDK 11.0.4">
```
tobiasdiez: Why is it necessary to specify that here, instead of reusing the `jlink` gradle task?

scripts/download-install4j-and-jres.sh
```diff
@@ -5,13 +5,4 @@ if [ ! -d ~/downloads ]; then
...
-
-# fetch JREs
-if [ ! -d ~/.install4j7/jres ]; then
-  mkdir -p ~/.install4j7/jres
-fi
-cd ~/.install4j7/jres
-wget --quiet -nc https://files.jabref.org/jres/windows-x86-1.8.0_172.tar.gz
-wget --quiet -nc https://files.jabref.org/jres/windows-amd64-1.8.0_172.tar.gz
-wget --quiet -nc https://files.jabref.org/jres/macosx-amd64-1.8.0_172_unpacked.tar.gz
+wget --quiet -nc --show-progress https://download-gcdn.ej-technologies.com/install4j/install4j_unix_8_0.tar.gz
```
tobiasdiez: rename file now to `download-install4j` ? maybe combine it with `extract-install4j`?

tobiasdiez:
I went through the PR again and the following things should be fixed before merging:
- [ ] Reenable wiremock tests (MimeTypeDetectorTest)
- [ ] Reenable archunit tests
- [ ] Decide what heap size to specify
- [ ] Reimplement stuff concerning `TextAreaSkin` and `TextFieldSkin`
- [ ] Fix and reenable CitationStyleTests
- [ ] Create working installer

Morerover, the following needs to be fixed at some point but we cannot do much at the moment:
- [ ] Reenable modernizer as soon as https://github.com/simonharrer/gradle-modernizer-plugin/pull/7 is merged
- [ ] Remove workaround for latex2unicode dependencies as soon as they are updated with proper module names: https://github.com/lihaoyi/fastparse/pull/185
- [ ] Remove openoffice workaround as soon as they provide proper module information: https://bugs.documentfoundation.org/show_bug.cgi?id=117331
- [ ] Update error-prone as soon as https://github.com/google/error-prone/issues/1240 is fixed
- [ ] Remove access to internal libraries which mostly come from controls-fx: https://github.com/controlsfx/controlsfx/blob/452e23e4d9e6d2e2fcddb1640e4fc5fc258221a1/build.gradle#L12-L41
- [ ] Specify `add-opens` and `add-exports` using `moduleOptions` as soon as https://github.com/java9-modularity/gradle-modules-plugin/issues/89 is fixed
- [ ] Simplify configuration of jlink as soon as more packages are modularized
- [ ] Use jpackage as soon as java 13 is released
- [ ] Reenable WebFetcher tests as soon as https://github.com/ronmamo/reflections/issues/202 is fixed

koppor:
Does "https://github.com/JabRef/jabref/pull/3421#issuecomment-515430078" mean, that we have to replace controlsfx completely.

koppor:
@tobiasdiez

- I put "not blocking for a release" to some points, is that right?
- At jlink, I get `error: module java.xml.bind reads package thredds.catalog from both cdm and grib`. Is that normal? Can this be ignored or is jlink also not finished here?

koppor:
Current state regarding jlink - I have no idea how to fix that.

```
$ gradlew jlink
Starting a Gradle Daemon, 1 busy Daemon could not be reused, use --status for details

> Configure project :
Found module name 'org.jabref'

> Task :compileJava
C:\git-repositories\jabref\circleci20\src\main\java\module-info.java:68: error: module not found: httpclient
  requires httpclient;
           ^
C:\git-repositories\jabref\circleci20\src\main\java\module-info.java:73: error: module not found: commons.lang3
  requires commons.lang3;
                  ^
error: module java.xml.bind reads package javax.activation from both java.activation and jakarta.activation
error: module java.xml.bind reads package com.sun.activation.registries from both java.activation and jakarta.activation
error: module java.xml.bind reads package com.sun.activation.viewers from both java.activation and jakarta.activation
error: module java.xml.bind reads package thredds.catalog from both cdm and grib
```

tobiasdiez:
These errors probably occur because some libraries were updated in the master branch (first guess: slfjlog 2.0 should be put back to 1.8>

koppor:
Fix refs https://github.com/JabRef/jabref/pull/5217

Siedlerchr:
https://github.com/TestFX/TestFX#java-11

koppor:
@Siedlerchr We already have that in: https://github.com/JabRef/jabref/blob/jdk9/build.gradle#L111 

The issue currently is:

    Module org.testfx contains package org.assertj.core.error.future, module org.assertj.core exports package org.assertj.core.error.future to org.testfx

In other words: 

- package A is contained in package B
- package A is exported by package C

with A being `org.assertj.core.error.future`, B being `org.testfx`, C being `org.assertj.core`

This is a contradition

koppor:
New dependencies can introuce spooky issues: https://github.com/JabRef/jabref/pull/5217

koppor:
Hint by @tobiasdiez: https://github.com/java9-modularity/gradle-modules-plugin#patching-modules-to-prevent-split-packages

koppor:
All tests green. GuiTests are properly ignored now.

Before that, we had ` || true` do allow failures, but due to some gradle magic for forking, it was hanging:

```
To honour the JVM settings for this build a new JVM will be forked. Please consider using the daemon: https://docs.gradle.org/5.6/userguide/gradle_daemon.html.

Starting process 'Gradle build daemon'. Working directory: /home/travis/.gradle/daemon/5.6 Command: /home/travis/oraclejdk11/bin/java --add-opens java.base/java.util=ALL-UNNAMED --add-opens java.base/java.lang=ALL-UNNAMED --add-opens java.base/java.lang.invoke=ALL-UNNAMED --add-opens java.prefs/java.util.prefs=ALL-UNNAMED -XX:MaxMetaspaceSize=256m -XX:+HeapDumpOnOutOfMemoryError -Xms256m -Xmx512m -Dfile.encoding=UTF-8 -Duser.country=US -Duser.language=en -Duser.variant -cp /home/travis/.gradle/wrapper/dists/gradle-5.6-bin/ez4di095eg6cfomw17rqphhsg/gradle-5.6/lib/gradle-launcher-5.6.jar org.gradle.launcher.daemon.bootstrap.GradleDaemon 5.6

Successfully started process 'Gradle build daemon'

An attempt to start the daemon took 1.717 secs.

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

The build has been terminated
```

PR #3422, opened 2017-11-10
Fix #3411: ordering of fields in customized entry types works again

### LINKED ISSUE #3411: BibTeX - customize entry types ordering issue
Using jabref-4.1-dev 2017-11-07 with java 1.8.0_152 on Linux x86_64.

1. Open a bib file
2. Select Bibtex - customize enttry types from menu
3. Choose Article type (default)
4. Select author and move it to top clcking the up button, order some other things
5. Click Apply

The ordering immediately goes back to its original order and saving the bib file adheres to that
ordering in bibtex entries. I presume this is not the "required" ordering for Article type, in which
case the  possibility to reorder should not be there. I have tested other functions but removing
items does seem to work, namely after the apply button they stay removed.

tobiasdiez:
merge if test passes since it is only a very minor code change.

### DIFF
--- a/src/main/java/org/jabref/gui/customentrytypes/EntryCustomizationDialog.java
+++ b/src/main/java/org/jabref/gui/customentrytypes/EntryCustomizationDialog.java
@@ -8,7 +8,7 @@ import java.awt.GridLayout;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
 import java.util.ArrayList;
-import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


src/main/java/org/jabref/gui/customentrytypes/EntryCustomizationDialog.java
```diff
@@ -342,7 +336,7 @@ private void deleteType(String name) {
                 }
             }
             EntryTypes.removeType(name, bibDatabaseMode);
-            updateEntriesForChangedTypes(Arrays.asList(name.toLowerCase(Locale.ENGLISH)));
+            updateEntriesForChangedTypes(Collections.singletonList(name.toLowerCase(Locale.ENGLISH)));
```
Siedlerchr: You are aware that Collections,singletonList is immutable?

tobiasdiez: Thanks for the hint, I was actually not aware of it. In this case, however, it does not matter since `updateEntriesForChangedTypes` only invokes `contains` on the list.

Issue #3397, opened 2017-11-03
Untranslated text 


tobiasdiez:
In todays master, for example:

Warning: could not get message translation for "Create_blank_entry_linking_the_PDF" for locale en
Warning: could not get message translation for "Create_entry_based_on_XMP-metadata" for locale en
Warning: could not get message translation for "Create_entry_based_on_content" for locale en
Warning: could not get message translation for "Only_attach_PDF" for locale en

The problem is the old format using the underscore. I think the were missed, when @halirutan improved the language handling. Our tests should be adapted to report warnings in this cases (just report them as not found).

Siedlerchr:
Refs #3184

halirutan:
@tobiasdiez I was aware of that and carefully checked that all usages did not use underscores. There was at this time only one instance that needed to be replaced. I'm sorry, I should have pointed this out more clearly back then: To make a fast cache, the un-quoting and translating of the messages is done on the first instantiation. Messages in code `lang(..)` must not contain the underscored variant but need to be the real message (which is more natural anyway).

tobiasdiez:
@halirutan Its perfectly fine! I actually like that the underscore version does no longer work (it was a relic of old times).

halirutan:
I added a test that will take care of this.

```
java.lang.AssertionError: Language keys must not be used quoted in code! Use "This is a message" instead of "This_is_a_message".
Please correct the following entries:
[
"Create_entry_based_on_content" (src/main/java/org/jabref/gui/preftabs/ImportSettingsTab.java)
, 
"ID-based_entry_generator" (src/main/java/org/jabref/gui/EntryTypeDialog.java)
, 
"Only_attach_PDF" (src/main/java/org/jabref/gui/preftabs/ImportSettingsTab.java)
, 
"Create_entry_based_on_XMP-metadata" (src/main/java/org/jabref/gui/preftabs/ImportSettingsTab.java)
, 
"Create_blank_entry_linking_the_PDF" (src/main/java/org/jabref/gui/preftabs/ImportSettingsTab.java)
] 
```

PR #3424, opened 2017-11-10
Fix test for quoted lang messages

halirutan:
Fix for https://github.com/JabRef/jabref/issues/3397

(I made some newlines in the `findMissingLocalizationKeys` so that they can be better seen in the console)

### DIFF
--- a/src/test/java/org/jabref/logic/l10n/LocalizationConsistencyTest.java
+++ b/src/test/java/org/jabref/logic/l10n/LocalizationConsistencyTest.java
@@ -117,6 +117,24 @@ public class LocalizationConsistencyTest {
     }
 
     @Test
+    public void languageKeysShouldNotBeQuotedInFiles() throws IOException {
+        final List<LocalizationEntry> quotedEntries = LocalizationParser
+                .findLocalizationParametersStringsInJavaFiles(LocalizationBundleForTest.LANG)
+                .stream()
+                .filter(key -> key.getKey().contains("_") && key.getKey().equals(new LocalizationKey(key.getKey()).getPropertiesKey()))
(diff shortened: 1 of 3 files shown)


src/test/java/org/jabref/logic/l10n/LocalizationConsistencyTest.java
```diff
@@ -117,6 +117,24 @@ public void keyValueShouldBeEqualForEnglishPropertiesMessages() {
     }
 
     @Test
+    public void languageKeysShouldNotBeQuotedInFiles() throws IOException {
+        final List<LocalizationEntry> quotedEntries = LocalizationParser
+                .findLocalizationParametersStringsInJavaFiles(LocalizationBundleForTest.LANG)
+                .stream()
+                .filter(key -> key.getKey().contains("_") && key.getKey().equals(new LocalizationKey(key.getKey()).getPropertiesKey()))
```
LinusDietz: I wonder if `.contains('_')` will produce false positives one day. Probably not very likely, as there are no problem so far.

halirutan: @LinusDietz We need *some* test or we will get a lot of false positives for messages that consist only of a single word like "Cancel". I didn't put too much thought into this. It was the first that came to mind.

LinusDietz: @halirutan I agree with the approach and don't have any other reasonable idea to solve this. My remark is entirely serves karma purposes, to have a 'told you so moment' if somebody actually needs to put an underscore into a localized String (;

PR #3428, opened 2017-11-11
Feature java version check again

halirutan:
OK, I added some comment about the Java 9 check and why we should leave it in for the moment. The rest is unchanged but updated to the latest master.


halirutan:
@tobiasdiez I'm not exactly sure how this would have been possible here. For some sub-test sure, but would it have been possible to make nested loops where I test all java8 versions against all java9 version when I provide them as parameter test?

tobiasdiez:
Yes, I think that is possible. The junit 5 documentation contains the code
````java
@ParameterizedTest
@MethodSource("stringAndIntProvider")
void testWithMultiArgMethodSource(String first, int second) {
    assertNotNull(first);
    assertNotEquals(0, second);
}

static Stream<Arguments> stringAndIntProvider() {
    return Stream.of(Arguments.of("foo", 1), Arguments.of("bar", 2));
}
````
By changing `stringAndIntProvider` you can generate every possible combination of java 8 and java 9 version strings. Anyway, no need to change it for this PR. Was just a remark for the next time.

halirutan:
Can this be merged? If yes, would some of the veterans do this?

Siedlerchr:
@halirutan  If you are ready, add the label "ready for review". From my point of view it's good now.

PR #3430, opened 2017-11-13
Add support for junit 5

tobiasdiez:
This PR adds junit 5 support and migrates the database tests to it (more as a proof of concept then a serious migration attempt).


LinusDietz:
Well, migrating to Junit 5 is more than a bit of search and replace, as there have been some more complex api changes.... We should agree on a roadmap for migrating junit 5.

For the time being... how well can junit 4 and junit 5 tests coexist in our CI-pipeline?

tobiasdiez:
@LinusDietz Yes, of course, there were some API changes - but mostly the tests run without any big changes (main exception: assertEquals with custom messages and parametrized tests).

@JabRef/developers Apparently the filtering by categories (or tags as they are now called) does not work for old-style tests. Thus, all fetcher, database and gui tests have to be migrated to junit 5 immediately. Overall our tests seem to pass when run with the vintage engine. Thus there is no need to migrate everything at once.

tobiasdiez:
So I managed the build to pass successfully, hence this PR is now really ready for review.

tobiasdiez:
I removed the plural s from the categories since this is apparently the naming convention for "tags" in junit 5 (probably because you now can mark a single test with a certain tag).

LinusDietz:
@tobiasdiez that removal of the plural 's' was not applied to  the merge commit. I made a quickfix: https://github.com/JabRef/jabref/commit/ee10e82823c794a57ace144109f6dcf3b69a3890

tobiasdiez:
Ah thanks! Should have waited until the build from the merge commit was completed. Mea culpa.

build.gradle
```diff
@@ -2,6 +2,15 @@ import org.gradle.internal.os.OperatingSystem
 
 // to update the gradle wrapper, execute ./gradlew wrapper --gradle-version 4.0
 
+buildscript {
+    repositories {
+        mavenCentral()
```
lenhard: I am wondering: Isn't jcenter (the site we have been using so far) a superset of maven central? And we remember that we used maven central before and then replaced it with jcenter, because of the availability of some library.

Why do you add maven central here?

Siedlerchr: Yes, agree, jcenter is superior. However, we could add it as second repo, in case some libs are not found on jcenter

tobiasdiez: This whole buildscript tag is just a workaround since as of now junit 5 does not support the id-syntax, ie something like
̀ ```
plugins {
id 'junit5' version '5'
}
````

Issue #3432, opened 2017-11-14
Flags (relevant, qualityAssured, ...) cannot be removed persistently from entries


derTimme:
I'm using JabRef version 4.0 on ArchLinux (Linux 4.13.11-1-arch amd64) with Java 1.8.0_144.

Steps to reproduce:

0. Set "Options" - "Preferences" - "Entry table columns" - "Synchronize with keywords"
1. Create a new empty Biblatex library (or use an existing one, doesn't make a difference for me)
2. Add a new entry (of any type, I tested with article)
3. Add a flag like "relevant" via the entry table (either using the right click menu or the corresponding column)
4. Save the library - the flag gets stored in the `keyword` field of the entry
5. Remove the flag in the entry table and save the library again
6. The flag is still present in the `.bib` file and shows up in JabRef when opening that library again

The log just says "Opening `<filename>`".
The same happens when setting multiple flags on a single entry.

\[Edit 1\]: Problem only occurs when syncing special fields with keywords.

Siedlerchr:
Could you please test with the latest version from builds.jabref.org

derTimme:
Yes - I'm out of time for today, will try it tomorrow.

Am 14. November 2017 21:46:25 MEZ schrieb Christoph <notifications@github.com>:

derTimme:
I tried it with JabRef 4.1-dev--snapshot--2017-11-13--master--c1d425a31 and got the same results.

derTimme:
Currently investigating, maybe I can fix it myself.
It seems like these special fields (as they are called in the code) are represented in two different ways:

- In the running instance, they are represented as additional biblatex fields (e.g. `relevance={relevant}`), these fields are *not* shown in the `biblatex source` tab of the entry editor however.
- In the stored `.bib` file, they are represented as a list of keywords (in the `keywords` field)

When opening a database, the `keywords` field is parsed in `org.jabref.logic.importer.OpenDatabase#loadDatabase(File, ImportFormatPreferences)` and the "runtime" fields are created.

derTimme:
Investigated a bit further: The problem only occurs when special fields are synced with the `keywords` field, if special fields are synced as separate bibtex fields, everything works as expected.
I updated my initial comment accordingly.

derTimme:
It would be nice to have a little discussion with the more experienced devs here (I started looking at the code two days ago...) about what needs to be done.
[...]

PR #3438, opened 2017-11-17
Fix removal of special fields when syncing them via keywords

derTimme:
Fixes a bug where JabRef special fields (such as 'printed', 'relevant', ...) were not cleared (e.g. removing the 'printed' flag) when using synchronization of special fields via the bibtex keywords field.

See #3432 for more details.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,7 +47,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
  - We fixed an issue where fetching entries from crossref that had no titles caused an error [#3376](https://github.com/JabRef/jabref/issues/3376)
  - We fixed an issue where the same Java Look and Feel would be listed more than once in the Preferences. [#3391](https://github.com/JabRef/jabref/issues/3391)
  - We fixed an issue where errors in citation styles triggered an exception when opening the preferences dialog [#3389](https://github.com/JabRef/jabref/issues/3389)
- - We fixed an issue where special fields (such as **printed**) could not be cleared when syncing special fields via the keywords [#3432](https://github.com/JabRef/jabref/issues/3432)
+ - We fixed an issue where special fields (such as `printed`) could not be cleared when syncing special fields via the keywords [#3432](https://github.com/JabRef/jabref/issues/3432)
 
  ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
  - We fixed an issue where fetching entries from crossref that had no titles caused an error [#3376](https://github.com/JabRef/jabref/issues/3376)
  - We fixed an issue where the same Java Look and Feel would be listed more than once in the Preferences. [#3391](https://github.com/JabRef/jabref/issues/3391)
  - We fixed an issue where errors in citation styles triggered an exception when opening the preferences dialog [#3389](https://github.com/JabRef/jabref/issues/3389)
+ - We fixed an issue where special fields (such as **printed**) could not be cleared when syncing special fields via the keywords [#3432](https://github.com/JabRef/jabref/issues/3432)
 
  ### Removed
 
--- a/src/main/java/org/jabref/logic/specialfields/SpecialFieldsUtils.java
+++ b/src/main/java/org/jabref/logic/specialfields/SpecialFieldsUtils.java
@@ -46,7 +46,7 @@ public class SpecialFieldsUtils {
 
         KeywordList keyWords = specialField.getKeyWords();
         Optional<Keyword> newValue = entry.getField(specialField.getFieldName()).map(Keyword::new);
-        newValue.map(value -> entry.replaceKeywords(keyWords, newValue, keywordDelimiter))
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -47,6 +47,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
  - We fixed an issue where fetching entries from crossref that had no titles caused an error [#3376](https://github.com/JabRef/jabref/issues/3376)
  - We fixed an issue where the same Java Look and Feel would be listed more than once in the Preferences. [#3391](https://github.com/JabRef/jabref/issues/3391)
  - We fixed an issue where errors in citation styles triggered an exception when opening the preferences dialog [#3389](https://github.com/JabRef/jabref/issues/3389)
+ - We fixed an issue where special fields (such as **printed**) could not be cleared when syncing special fields via the keywords [#3432](https://github.com/JabRef/jabref/issues/3432)
```
Siedlerchr: Just a very minor issue, for field names in changelog we usually use backticks

src/main/java/org/jabref/logic/specialfields/SpecialFieldsUtils.java
```diff
@@ -46,9 +46,12 @@
 
         Optional<Keyword> newValue = entry.getField(specialField.getFieldName()).map(Keyword::new);
         KeywordList keyWords = specialField.getKeyWords();
-
-        Optional<FieldChange> change = entry.replaceKeywords(keyWords, newValue, keywordDelimiter);
-        change.ifPresent(changeValue -> fieldChanges.add(changeValue));
+        if (newValue.isPresent()) {
```
Siedlerchr: You can chain nested optionals with map, it will then only be executed if a value is present

src/test/java/org/jabref/logic/specialfields/SpecialFieldsUtilsTest.java
```diff
@@ -69,4 +72,18 @@ public void syncSpecialFieldsFromKeywordCausesNoChangeWhenKeywordsAreEmpty() {
...
+    @Test
+    public void updateFieldRemovesSpecialFieldKeywordWhenKeywordSyncIsUsed() {
+        BibEntry entry = new BibEntry();
+        SpecialField specialField = SpecialField.PRINTED;
+        Keyword specialFieldKeyword = specialField.getKeyWords().get(0);
+        // Add the special field
+        SpecialFieldsUtils.updateField(specialField, specialFieldKeyword.get(), entry, true, true, ',');
+        // Remove it
+        List<FieldChange> changes = SpecialFieldsUtils.updateField(specialField, specialFieldKeyword.get(), entry, true, true, ',');
+        assertTrue(changes.size() > 0);
```
Siedlerchr: Please avoid checking such things with size, instead directly create a list expected and use assertEquals

src/test/java/org/jabref/model/entry/BibEntryTests.java
```diff
@@ -369,6 +372,36 @@ public void getKeywordsReturnsParsedKeywordListFromKeywordsField() {
...
+        keywordEntry.putKeywords(Arrays.asList("kw1", "kw2"), ',');
+        Optional<FieldChange> change = keywordEntry.removeKeywords(new KeywordList(), ',');
+        assertEquals(Optional.empty(), change);
+    }
+
+    @Test
+    public void removeKeywordsWithNonExistingKeywordsDoesNothing() {
+        keywordEntry.putKeywords(Arrays.asList("kw1", "kw2"), ',');
+        Optional<FieldChange> change = keywordEntry.removeKeywords(KeywordList.parse("kw3, kw4", ','), ',');
+        assertEquals(Optional.empty(), change);
```
Siedlerchr: As mentioned above, better to directly compare the result to a pre defined list with the keywords which you already use to add the keywords
So just do expected= arrrays.aslist(kw1,kw2)

CHANGELOG.md
derTimme: Even if it does NOT reference a bibtex field? With "printed", I refer to the "special field" here.

derTimme:
Made the requested changes.

src/main/java/org/jabref/model/entry/BibEntry.java
```diff
@@ -688,6 +678,22 @@ public KeywordList getResolvedKeywords(Character delimiter, BibDatabase database
         return KeywordList.parse(keywordsContent, delimiter);
     }
 
+    public Optional<FieldChange> removeKeywords(KeywordList keywordsToRemove, Character keywordDelimiter) {
+        KeywordList keywordList = getKeywords(keywordDelimiter);
+        keywordList.removeAll(keywordsToRemove);
+        return putKeywords(keywordList, keywordDelimiter);
+    }
+
+    public Optional<FieldChange> replaceKeywords(KeywordList keywordsToReplace, Optional<Keyword> newValue,
```
tobiasdiez: Can you please change the type of `newValue` to be just `Keyword` (instead of an optional around it). Accepting the optional, but then not properly acting on empty ones, actually lead to this problem. So I think we should fix its root.

derTimme:
Thanks @tobiasdiez for pointing that out - fixed it.

PR #3439, opened 2017-11-17
Add Hint for checking master version

Siedlerchr:
I added a tasklist checkbox with the hint to verify the problem with the latest master

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -2,8 +2,12 @@
 Feature requests,  questions and general feedback is now handled at http://discourse.jabref.org 
 Thanks! --> 
 
+[] I have tested the latest master version from http://builds.jabref.org/master/ and the problem persists
+
+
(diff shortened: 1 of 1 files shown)


halirutan:
Good idea. We have too many issue reports that still use 4.0 and report things that are already fixed. I think at the current pace of changes that are made, working towards a big 4.1 release is not the best course of action. We have reworked so many things. Shouldn't we push an update soon?

LinusDietz:
Actually, I'm against this.
The underlying problem is, that we have introduced too many bugs in our major releases lately which were, of course, reported quite often.

As @halirutan said, we should release more often and the problem should go away again.

Siedlerchr:
Even when the release is stable, the dev Version could already contain a bug fix for an issue present in release. Therefore it makes sense to check the latest dev before

halirutan:
I'm definitely in favor of this as well. It is a good thing to ask users "hey, have you checked our latest dev release?" Especially, since some of them might not even be aware of that there is an active develop-version that they can try.

Issue #3440, opened 2017-11-17
Mac OS X dmg is corrupted for the nightly builds


halirutan:
In this issue https://github.com/JabRef/jabref/issues/3437#issuecomment-345182095, the reporter pointed out he could not install the `dmg` file from the development builds. I tried to reproduce the problem on my MacBook (10.12.6) and I got as well the following error:

<img width="532" alt="screen shot 2017-11-17 at 18 26 57" src="https://user-images.githubusercontent.com/511683/32960567-a33b70cc-cbc5-11e7-89a8-e2e5fe5eba73.png">

That is something we need to fix ASAP. Could the dev who is responsible for the release building on CI look at this? I can help and test, but I cannot build the release myself as there seem to be some local settings that are required that I don't have.

Siedlerchr:
Could you try to test with an older version of here?
http://builds.jabref.org/fixfetcherisbn/
I am unsure whether it is caused by a certficiation issue (the code signing expired) or if its caused by a commit

halirutan:
@Siedlerchr This one works.

ncihnegn:
My Firefox download is corrupted but curl works fine.

halirutan:
OK, I googled a bit around and I believe I found the problem:

```
codesign -v JabRef_macos_4_1-dev--snapshot--2017-11-13--master--c1d425a31.dmg

JabRef_macos_4_1-dev--snapshot--2017-11-13--master--c1d425a31.dmg: CSSMERR_TP_CERT_EXPIRED
```

The certificate of the older `dmg` is not expired.

Siedlerchr:
@halirutan  Yes, what I already thought. We are aware that our codesigning is expired and we are in the process of getting a new one.
In the meantime I found this solution: https://github.com/processing/processing/issues/4705#issuecomment-269241488

LinusDietz:
We disabled code signing until we obtain the certificate.

PR #3441, opened 2017-11-18
Improve searching for JRE in install4j and set max version

### LINKED ISSUE #3434: JabRef Java version check does not work correclty if java9 and java8 are installed parallel
Have both java 9.01 and java 8 installed: (latter one is default)

grafik

C:\Windows\System32>java -version
java version "1.8.0_152"
Java(TM) SE Runtime Environment (build 1.8.0_152-b16)
Java HotSpot(TM) 64-Bit Server VM (build 25.152-b16, mixed mode)

Siedlerchr:
Should fix #3434 
add  max version to 1.8

I tested this under Windows 10 with jdk9 and jdk9 installed and it now works correclty
@halirutan  @tobiasdiez  could you please test, if this works for you, too?

### DIFF
--- a/jabref.install4j
+++ b/jabref.install4j
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<install4j version="7.0.1" transformSequenceNumber="7">
+<install4j version="7.0.3" transformSequenceNumber="7">
   <directoryPresets config="build/releases/${compiler:buildFileName}" />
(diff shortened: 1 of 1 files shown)


halirutan:
@Siedlerchr I can only test it on Linux and Mac and there it works. Since this is the `install4j` branch, can we try to include a fix for https://github.com/JabRef/jabref/issues/3440? :smile: 

I compared the master branch (where the dmg doesn't work) with the one you mentioned. I could not find any relevant difference. The building of the dmg is done by install4j as well, correct? How is it possible that this breaks, when we haven't changed anything the `install4j` settings? :confused:

Siedlerchr:
Regarding the broken dmg, I absolutely have no idea. Maybe it's Firefox who corrupts the file? Another reason could be that our code signing certificate no longer is valid. 
The dmg is built with the installer as well on circle CI, I checked the logs, but could not see anything regarding this

halirutan:
@Siedlerchr What's got Firefox (the browser?) to do with it?

Issue #3381, opened 2017-10-31
Html code in the tooltip of the search field


jonasstein:
```
JabRef 4.1-dev--snapshot--2017-10-31--master--42ea6bc9c
Linux 4.12.12-gentoo+ amd64 
Java 1.8.0_152
```
There is html code in the tooltip of the search field (click to enlarge the screenshot)

![jabref_376](https://user-images.githubusercontent.com/298682/32246612-d24045da-be7f-11e7-8fb4-6411cb6ab6f4.png)

jupf:
Hey, I looked into this issue and would be glad to take it.

It's possible to display a WebView inside the Tooltip which renders the HTML as a graphic. Or was there any other solution intended?

Does anyone know why the background of the search field tooltip is black-ish? Shouldn't it be grey or light blue? Seems strange to me and I couldnt find out why.

This is my first approach to contribute in an open source project, i already read your contributing guide. Is there something else to consider?

jonasstein:
As user I would expect plain text with proper line break

jupf:
It would be text, but with the possibility to make some words e.g. bold

Siedlerchr:
Hi, thanks for your interest in this issue. Both javafx and swing support html toolips out of the box. 
See https://docs.oracle.com/javafx/2/ui_controls/tooltip.htm
. Regarding the color, I think it depends on the look and feel you choose.

Some General Tipps for contributing, make sure to set up your IDE accordingly so that you have the code formatter rules we use.
Just create a new branch and add a PR. If you have questions or need help, just ask

Siedlerchr:
I think you should use the html option, to make sure the author names are shown correctly

tobiasdiez:
@jupf Thanks for your interest in contributing!

JavaFX actually does not support HTML in its controls and I think this is a good design. Instead of using a webview (which would work in principle), JavaFX introduced the [TextFlow](https://docs.oracle.com/javase/8/javafx/api/javafx/scene/text/TextFlow.html) to support rich text.
Thus I would propose to change https://github.com/JabRef/jabref/blob/ebbeb1d249edf7134abf92d24c2031ae54d19881/src/main/java/org/jabref/logic/search/rules/describer/SearchDescriber.java#L6 to return a `TextFlow` instead of `String`. Note that as a consequence of this the `search/rules/describer/` package should move to gui instead. If you feel not yet comfortable enough to do this refactoring, it is no problem - we can help you with it or take over this final step.
[...]

PR #3443, opened 2017-11-18
Fix for issue #3381

jupf:
Because the javafx Tooltip used in the [GlobalSearchBar](https://github.com/JabRef/jabref/blob/master/src/main/java/org/jabref/gui/search/GlobalSearchBar.java) cannot render HTML, @tobiasdiez recommended in #3381  to create a TextFlow object in the [SearchDescriber](https://github.com/JabRef/jabref/blob/master/src/main/java/org/jabref/logic/search/rules/describer/SearchDescriber.java) classes for structuring the Text. 

I made the changes to return a TextFlow and its working fine so far without changing the logic that is used to build the text.  I moved the search.rules.describer package like proposed from the logic package to the gui package. The existing tests were moved accordingly but still needs to be rewritten for the new TextFlow results.

There are some open questions i would like to ask:

1. With no Styling of the Tooltip, the readability was really bad (black text on grey background), so I changed the styling to look like the other Tooltips in the bar. Hope it was okay to do this in the context of this issue? I created a new styleclass for this in the Main.css to not change any other styled elements.
2. the [SearchQuery](https://github.com/JabRef/jabref/blob/master/src/main/java/org/jabref/logic/search/SearchQuery.java) class stores a description of the search. It is created with the SearchDescribers. But they are now inside the gui package and return a TextFlow. So there is a javafx dependency in the SearchQuery class which resides in the logic package at the moment. This doesnt sound right to me? Should the class be moved to the gui package too?
3. The [GroupDialog](https://github.com/JabRef/jabref/blob/master/src/main/java/org/jabref/gui/groups/GroupDialog.java) class uses a SearchQuery object to get a search description. This class is written with Swing and is not compatible to the now returned TextFlow. If I understand it right, the project is migrating to use javaFX. So I want to propose that I write a compatibility method in the GroupDialog class to create an HTML string from the given TextFlow to be compatible while the GroupDialog is still using Swing. 
4. Sadly the localized strings used in the SearchDescribers included the HTML tag elements.. So in many cases it is not possible to use existing translations.. just because the tags are gone. I'm not sure what to do about this. It could be possible to remove the tags manually from the translations. Not sure how long this would take..

That's how the tooltip looks with the changes:
screenshot-2017-11-18--13 15 28

There seems to be something wrong with the logic of the tooltip text creation, the first few words repeat itself. And it already happened prior to my changes (see the image in #3381) But I didnt had time to look very closely into this. Should I open a new issue for this problem?

I hope I didn't forget anything. I'm happy to hear your feedback!


Siedlerchr:
First of all thanks for your contribution! Moving all logic code to gui seems like the only option here, but in general should be avoided.  What about having a different approach: Keepiing all stuff string based in logic and in the GUI you have a helper method for converting the HTML tags (are only a few) from the localization into the RTF equivalent and by adding it to a textflow. I see no reason why this should not work.

If you look into the localization stuff, you see there are only a couple of `<BR> and <b>/<i>` tags, so nothing really advanced. 
https://github.com/JabRef/jabref/blob/master/src/main/resources/l10n/JabRef_en.properties

jupf:
After thinking a little longer about the package problem. I think the architecture is actually not right if the SearchDescribers generate HTML code and stay in the logic package. Because HTML code is only usable in a GUI use case and has not really anything to do with logic. So maybe they were not really in the right package to begin with if their only purpose is to create a structured text for the gui?

I think the SearchQuery objects should not hold a description stored. If it is removed, there is no further moving of classes necessary. At the places the description member of SearchQuery is used, it can be generated with the SearchDescribers class like in the SearchQuery class at the moment.

IMHO the package structure would be most honored that way

It would save the translations if the SearchDescriber creates an HTML string which can be converted later to a textflow like you said. But if the project really is moving towards javafx it seems wrong to me to keep html strings for structuring text because in javafx this is the task of a TextFlow. I would try to look into removing the HTML tags from the translations to not lose them and make a step forward using javafx

tobiasdiez:
Also a big thanks from me! I don't have much time at the moment, so just a short remark concerning your point 2. I think it should be possible to completely remove the `description` field from the `SearchQuery` class. Just expose the `rule` (and `query`) via getters and then call the ̀ SearchDescriptor.getDescriptionFor(...)` every time you need the description in the gui. In this way, the description completely disappears from the logic package.

(Concerning 4: You can also embed JavaFX controls in swing, which is probably easier to do then creating a conversion to html)

jupf:
Status update:
* I removed the description from the SearchQuery class so it can remain in the logic package
* I recreate an html string in the GroupDialog class for compatibility (was very easy this way)
* modified the Tests for the SearchDescriber Implementations to properly check the TextFlow description
* created Tests for the utility class I created

Now there is only one problem left: the broken localizations.
I looked into the localization files and it could be possible to rescue a lot of translations by just manually deleting the html tags from the translations (when they are at the beginning or end). But in some cases, after translating the text, the tags are positioned in the middle of words. There it would be necessary to create a new translation because I'm not able to just delete the tag from the beginning or end.

The only other alternative i see is using the old html strings for the localization and then break them up and convert them into javafx texts.. IMHO this would be really ugly and not a good readable solution

What do you think? Any other possible solutions?

jupf:
Sadly its not so simple with TextFlows. They consist of multiple Text objects and you can set the font on each Text. So if there is a String like this:
XXX XXX XXXX <b>XX</b> 
that translates into:
XX XXX <b>XXX</b> XX XXXX
Then there are three Text objects needed: Text("XX XXX"), Text("<b>XXX</b>"), Text("XX XXXX")

I will try to find an acceptable abstraction so the code will not be completely unreadable and the old translations can be used

jupf:
So, now the old strings are used and the localization is working again. I tried to make it as readable as possible. 

What do you think now? Any more things to do? Anything else I should have considered?

src/main/java/org/jabref/gui/Main.css
```diff
@@ -66,6 +66,12 @@
     -fx-effect:none;
 }
 
+.search-tooltip {
```
tobiasdiez: I agree that the default JavaFX tooltips are hard to read! But I'm not a fan of the old swing tooltips either. I would suggest to use a grayish color scheme with less opacity (>= 90%): [inspiration](https://material.io/guidelines/components/tooltips.html#tooltips-tooltips-desktop). Maybe increasing the text size (by default its 0.85em).

Please override the default `.tooltip` style so that tooltips look consistent.

src/main/java/org/jabref/gui/groups/GroupDialog.java
```diff
@@ -566,7 +571,9 @@ private void updateComponents() {
             s1 = searchGroupSearchExpression.getText().trim();
             okEnabled = okEnabled & !s1.isEmpty();
             if (okEnabled) {
-                setDescription(new SearchQuery(s1, isCaseSensitive(), isRegex()).getDescription());
+                SearchQuery searchQuery = new SearchQuery(s1, isCaseSensitive(), isRegex());
+                setDescription(fromTextFlowToHTMLString(SearchDescribers.getSearchDescriberFor(searchQuery.getRule(),
```
tobiasdiez: I think it makes sense to introduce an overload of `getSearchDescriberFor` that accepts a `SearchQuery` (or even change the signature directly).

src/main/java/org/jabref/gui/util/TextUtil.java
```diff
@@ -0,0 +1,116 @@
...
+                break;
+            case BOLD:
+                font = Font.font("System Regular", FontWeight.BOLD, size);
+                break;
+            case ITALIC:
+                font = Font.font("System Regular", FontPosture.ITALIC, size);
+                break;
+            case MONOSPACED:
+                font = Font.font("Monospaced", size);
+                break;
```
Siedlerchr: Just add a default case.

src/main/java/org/jabref/gui/util/TextUtil.java
```diff
@@ -0,0 +1,116 @@
...
+                        + "' exists multiple times in the original string");
+            }
+        } else {
+            throw new IllegalStateException("It is not allowed that the toReplace string: '" + replacement.toReplace
+                    + "' does not exist in the original string");
+        }
+    }
+
+    public static class TextReplacement {
+        String toReplace;
```
Siedlerchr: Please make the vars private final

src/test/java/org/jabref/gui/util/TextUtilTest.java
```diff
@@ -0,0 +1,150 @@
...
+
+    @Test
+    public void testTextToHTMLStringWithLinebreaks() {
+        String testText = "this\nis a\ntest text";
+        Text text = TextUtil.createText(testText, 12, TextUtil.TextType.NORMAL);
+        String htmlString = TextUtil.textToHTMLString(text);
+        Assert.assertEquals("this<br>is a<br>test text", htmlString);
+    }
+
+    private boolean checkIfTextsEqualsExpectedTexts(List<Text> texts, List<Text> expectedTexts) {
```
tobiasdiez: Please extract this method to a new util class, since it is used in many tests.
Also remove the `System.out` statements.

src/test/java/org/jabref/gui/search/ContainsAndRegexBasedSearchRuleDescriberTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+import static org.junit.Assert.assertTrue;
+
+public class ContainsAndRegexBasedSearchRuleDescriberTest {
+
+    private boolean checkIfDescriptionEqualsExpectedTexts(TextFlow description, List<Text> expectedTexts) {
+        if (expectedTexts.size() != description.getChildren().size())
+            return false;
+        Text expectedText;
+        for (int i = 0; i < expectedTexts.size(); i++) {
```
Siedlerchr: Better convert the two lists to sets and use streams 
https://stackoverflow.com/a/30346012 (second approach)
Furthermore, as this method is also used in the other test it makes sense to extract it as one TestUtil method, to a new or an existing "testutil" class. I don' kntpw

jupf: but the order is important for the test so I dont think a set is helpful here. Every item has to be in the right position with the correct properties.

Siedlerchr: Yeah, okay, then stick with this. But please extract it to a new class so that it can be reused across the tests

src/main/java/org/jabref/gui/Main.css
jupf: The problem is that the TextFlow is not styled together with the tooltip its inside. So only the background can be changed with the tooltip class when using a TextFlow. It would be needed to add a style class to each used texts inside the textflow i think? then i was able to change the text font. But this would mean it would be necessary to make a text style class for each different style: bold, italic, monospaced etc. In my tests the css style had overwritten the programatically set font.

So im a little confused what to do here. It seems difficult to make a standardized look here. Depending on what is used inside the tooltip: normal text <-> TextFlow

Even if i create all the different text style classes, they had to be used explicitly in all other places. So no automatic default style possible here i think? Or is there any other possible solution i dont see?

tobiasdiez: You are talking about the font size? Using `-fx-font-size` does not work?
Moreover, you can influence the look of sub-items by using the appropriate css selector. For example, with `.tooltip > .text` you should be able to modify the look of the text elements in the tooltip. However, I have to admit that I don't really understand where you problem lies.

jupf: The selector ```.tooltip > .text { ...``` does not work here. The text element has no default style class.  
I just found out I could use this: ```.tooltip Text { ...``` but then all texts have the same style. But we need to have some Texts with Bold, etc. properties and then the Text.setFont method for setting the font progammatically is not working anymore..

tobiasdiez: That is strange. Usually styles specified in the code take priority over css files. You could add [PseudoClasses](https://docs.oracle.com/javase/8/javafx/api/javafx/css/PseudoClass.html) (e.g. `bold`) to the text element and style them via the selector `.tooltip Tex :bold`. 

But maybe the best solution is to actually leave all text related styles at their default value and only play a bit with the background color?

jupf: That would be the easiest solution :) 

But a good styling would be nice too.

I played around a bit and the following would also be possible:  
```css
.tooltip Text {
    /* this styles all texts inside a tooltip */
    -fx-fill: rgba(255, 255, 255, 0.9);
    -fx-font-size: 1em;
}

.tooltip > TextFlow > .tooltip-text-bold {
    -fx-font-weight: bold;
}

.tooltip > TextFlow > .tooltip-text-italic {
    -fx-font-style: italic;
}

.tooltip > TextFlow > .tooltip-text-monospaced {
    -fx-font-family: monospace;
}
```
So every text gets a basic styling and when you want to make one Text e.g. bold you can add the ```"tooltip-text-bold"``` style class, then additionally the text will be bold. It is even possible to combine styles like italic and bold

I think I could integrate this rather easily. I didn't look into PseudoClasses but if you say it is the best way to go I can look into them. So what should it be?

tobiasdiez: Your solution looks very clean! Go for it, if it works! (Pseudo classes are very similar, not sure what is preferable here).

jupf:
So, I think I made all your requested changes, it now looks like this with the new styling:
screenshot-2017-11-27--17 49 24

tobiasdiez:
Thank you very much for the quick follow-ups and, of course, for your contribution in general. We are looking forward to see more PRs from you 😉

jupf:
Your Welcome! :grin:

PR #3445, opened 2017-11-20
Declare FreeCiteImportestTest as Fetcher Test

LinusDietz:
FreeCite seems to be down atm. While the issues looks to be temporary (http://freecite.library.brown.edu/) is also down, I have declared the test as a fetcher test.


LinusDietz:
merging to fix build

PR #3450, opened 2017-11-22
Fix JUnit tests on MacOS

LinusDietz:
- [x] Tests are passing
- [x] Re-enable disabled tests so they run also on mac.
- [x] Convert tests in question to JUnit5 (I propose to gradually transition to JUnit5 by updating all tests that we touch.)

### DIFF
--- a/src/test/java/org/jabref/gui/keyboard/KeyBindingsDialogViewModelTest.java
+++ b/src/test/java/org/jabref/gui/keyboard/KeyBindingsDialogViewModelTest.java
@@ -10,16 +10,17 @@ import javafx.scene.input.KeyCombination;
 import javafx.scene.input.KeyEvent;
 
 import org.jabref.gui.DialogService;
+import org.jabref.logic.util.OS;
 import org.jabref.preferences.PreferencesService;
 
 import org.junit.jupiter.api.BeforeEach;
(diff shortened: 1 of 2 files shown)


tobiasdiez:
If it is not possible to fix the tests on MacOS, I propose to use the following method to disable those tests on MacOS: https://blog.codefx.org/libraries/junit-5-conditions/

LinusDietz:
I have finally finished this PR. I had a look at @tobiasdiez suggestion, however this functionality is not available from maven and the license is unclear. Furthermore, I think it's a bit overkill.

So what I did in the end is to disable the 5 tests on MacOs by adding an assumtion in the first line of the test:
```
assumeFalse(OS.OS_X);
```
This aborts the test without letting it fail.

lenhard:
@LinusDietz Why did you request a review from me? I do not have a Mac and cannot meaningfully test this.

LinusDietz:
Well you can believe me that the tests succeed in MacOs, I rather wanted your opinion if the way I selectively disabled the tests using `assumeFalse(OS.OS_X);` is okay.

Siedlerchr:
The quesition would be rather, why the specific tests do not suceed under Mac?

PR #3454, opened 2017-11-24
Replace swingx with standard swing

lenhard:
In order to be able to properly move to Java 9 (see #2594), we need to get rid of all accesses of internal APIs. Unfortunately, we also use libraries that make such internal accesses and we need to find a solution for that.

One of these libraries is swingx. Originally by Sun, it is no longer maintained since quite a while we could not even upgrade to the newest version, because of bugs in the library. It is time to get rid of this dependency. This PR replaces the usage of swingx with regular Swing components. Only the new entries dialog and the side pane are affected. For the new entries dialog, the change is trivial (just using a regular layout manager). For the side pane, a few changes are needed, but I can make it roughly resemble the old layout. 

There's one bigger change, though, that I happen to like very much: The side panes are no longer exclusive, you can display them together. That means that I can have the web search and the groups panel open at the same time. Personally, I think that's pretty cool and I would like to keep it that way. You can of course still disable the side panes independent of each other. Here's a screenshot:
sidepane

----

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -80,10 +80,6 @@ dependencies {
     compile 'com.jgoodies:jgoodies-forms:1.9.0'
     compile 'com.jgoodies:jgoodies-looks:2.7.0'
 
-    // Later versions cannot be used. Tested until 1.6.5-1.
-    // See https://github.com/JabRef/jabref/issues/271 for details.
-    compile 'org.swinglabs.swingx:swingx-core:1.6.4'
-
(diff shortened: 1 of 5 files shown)


Siedlerchr:
Yeah! Cool, pretty good work!

matthiasgeiger:
> There's one bigger change, though, that I happen to like very much: The side panes are no longer exclusive, you can display them together. That means that I can have the web search and the groups panel open at the same time.

This was already possible before 😉

lenhard:
Huh? Good to know, then at least no one can blame it on me when something related to that breaks ;-)

PR #3457, opened 2017-11-25
Use JUnit5 conditions to disable some tests on CI server

tobiasdiez:
Instead of assumptions, we can now disable tests on the CI server using a new `@DisabledOnCIServer` annotation. I converted the fetcher tests affected to junit 5.
Moreover, I extracted some common tests (null / empty) to a new parametrized fulltextfetcher class.

----


tobiasdiez:
Good question! IntelliJ warned about it and this is why I changed it. If you look that the examples from the JUnit team, they also always use the default visibility (e.g. [here](http://junit.org/junit5/docs/current/user-guide/#writing-tests-standard)). 
With junit 4 all tests had to be public and this restriction was removed in the new version.

I have no strong opinion about this.

lenhard:
In that case, I'd say we go for default visibility in all JUnit5 tests, regardless of whether they're new or migrated from a JUnit4 tests.

tobiasdiez:
Ok, I now used the default visibility everywhere.

Siedlerchr:
Default Visibility is normally considered an anti pattern and is marked by coday:
DefaultPackage

Since: PMD 3.4

Use explicit scoping instead of the default package private level.

PR #3458, opened 2017-11-25
Remove dependency to jgoodies-looks

lenhard:
Jgoodies-looks uses internal APIs (see https://github.com/JabRef/jabref/issues/2594#issuecomment-346829982 ), which means that it causes problems with Java 9. This PR removes this dependency. The biggest change is that two Look and Feels go and that particularily affects Linux users.

I have used the Nimbus Look and Feel as the new default for Linux, but we can change that or even search for better Java 9-compatible Look and Feels. I have also added a preferences migration in case a user has set one of the incompatible Look and Feels.

I would suggest that we wait with merging this after we have release 4.1. Java 9 does not yet matter for 4.1 and we should give the Linux users a stable JabRef that still has the old Look and Feels before moving on.

----


src/main/java/org/jabref/JabRefGUI.java
```diff
@@ -251,10 +249,7 @@ private void setLookAndFeel() {
             if (UIManager.getCrossPlatformLookAndFeelClassName().equals(lookFeel)
                     && !System.getProperty("java.runtime.name").contains("OpenJDK")) {
                 // try to avoid ending up with the ugly Metal L&F
-                Plastic3DLookAndFeel lnf = new Plastic3DLookAndFeel();
-                MetalLookAndFeel.setCurrentTheme(new SkyBluer());
-                com.jgoodies.looks.Options.setPopupDropShadowEnabled(true);
-                UIManager.setLookAndFeel(lnf);
+                UIManager.setLookAndFeel(new NimbusLookAndFeel());
```
Siedlerchr: you should set the LookAndFeel based on className, as recommend by the jdk api:
   UIManager.setLookAndFeel("javax.swing.plaf.metal.MetalLookAndFeel");

lenhard: Thanks for pointing out! Changed it.

src/main/java/org/jabref/gui/preftabs/AppearancePrefsTab.java
```diff
@@ -60,8 +60,6 @@
         public static Set<String> getAvailableLookAndFeels() {
 
             Set<String> lookAndFeels = new HashSet<>();
-            lookAndFeels.add("com.jgoodies.looks.plastic.Plastic3DLookAndFeel");
-            lookAndFeels.add("com.jgoodies.looks.windows.WindowsLookAndFeel");
 
             for (LookAndFeelInfo info : UIManager.getInstalledLookAndFeels()) {
```
tobiasdiez: I think the rest of the method can now be replaced by the one-liner `return getInstalledLookAndFeels().stream().map(LookAndFeelInfo:getClassName).collect(toSet)`  ?

lenhard: Not exactly (it's and array), but almost: `Arrays.stream(UIManager.getInstalledLookAndFeels()).map(LookAndFeelInfo::getClassName).collect(Collectors.toSet())`

koppor:
Can't incorporating this new L&F in 4.1 give us a huge user-base to test? 4.1 will be a release with christmas colors and then we will release 4.2 somewhere around 6th january with the default colors again. - Hoping that this works with the new color configuration ^^.

koppor:
This is how JabRef looks after upgrading:

grafik

@LinusDietz could you provide a Linux screenshot?

tobiasdiez:
To be honest, I wouldn't spend to much time on swing look and feels. They disappear as soon as we finish migrating (most) of the ui to JavaFX.

src/main/java/org/jabref/preferences/JabRefPreferences.java
```diff
@@ -466,12 +466,12 @@ private JabRefPreferences() {
             defaults.put(WIN_LOOK_AND_FEEL, UIManager.getSystemLookAndFeelClassName());
             defaults.put(EMACS_PATH, "emacsclient");
         } else if (OS.WINDOWS) {
-            defaults.put(WIN_LOOK_AND_FEEL, "com.jgoodies.looks.windows.WindowsLookAndFeel");
+            defaults.put(WIN_LOOK_AND_FEEL, "com.sun.java.swing.plaf.windows.WindowsLookAndFeel");
```
koppor: According to https://bugs.openjdk.java.net/browse/JDK-8136366, this is a non-public L&F. But reading along, it seems, we are "correctly" using it...

lenhard:
@koppor: For me JabRef looked like this all the time ;-) I have been using the windows landf all along. The big difference in look might rather be on Linux.

Siedlerchr:
Okay, here are some linux screenshots:
Xubuntu 16.10,  Theme: Greybird 
Plastic Look and Feel - jgoodies (seems like the same as metal?)
linuxplasticxubuntu

Nimbus L&F
linuxnimbusxubuntu

Metal L&F
linuxmetalxubuntu

GTK L&F (broken)
linuxgtkxubuntu

lenhard:
Since 4.1 is released, we can merge this as agreed.

PR #3460, opened 2017-11-27
Only build master branch on push

tobiasdiez:
Our current workflow is to create a branch on the main JabRef fork and then open a PR for it. This, however, leads to two travis builds: one for the new push to JabRef and one for the PR.
There seems to be no elegant way to force travis to

- always build PRs
- but don't build PRs from main branches twice

See https://github.com/travis-ci/travis-ci/issues/6612. I followed [the suggested solution on stackoverflow](https://stackoverflow.com/a/31882307/873661) and only white-listed the master branch.
Thus pushes to master still trigger a build, as well as all PRs. However, pushes to other branches on main does not result in build. Thus only one build is executed if you open a PR from a main branch. Disadvantage: you have to open a PR to get a build from a different branch. I don't think this is a huge disadvantage since it is exactly our current workflow. 

----

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,6 +51,10 @@ after_failure:
   # show test results if build fails
   - $TRAVIS_BUILD_DIR/scripts/after-failure.sh
 
+branches:
+  only: 
+    - master
+
 # cache gradle dependencies
 # https://docs.travis-ci.com/user/languages/java#Caching
 before_cache:


koppor:
I have to confess, when I start to prepare pull requests, I push my current developments and check whether travis fails every now and then. But this is not a show-stopper.

grafik

koppor:
DevCall: PR build merges master before and then builds. Thus, it really makes sense to build both.

lenhard:
Some more explanation: In rare case there is a justification for both builds, namely when the combination of PR and master breaks. This doesn't happen often, but I recall at least one case where this happened to me. There's no real cost for us in executing two builds. That's why we decided to close this PR.

koppor:
Note: This PR still builds PRs, but not the respective branch.

koppor:
The time saving is relevant short before a release, when we run into travis limits.

koppor:
We include it until someone gets upset in his workflow

We accept: "local builds of incompatible things are not built any more."

lenhard:
@JabRef/developers It seems that we are now no longer getting builds for PRs on builds.jabref.org That's sad, because we can no longer point users to test a particular release. 

I think we should revert this PR.

LinusDietz:
Nope, this is because of #3499.

lenhard:
Oh well, people keep finding reasons why my perception of this PR is wrong ;-) 

Let's get the darn code signing fixed...

PR #3465, opened 2017-11-29
Added a "Clear text" button in right click menu within the text boxes

weidafan:
Adresses koppor#198

It was requested to have a new button within the right click menu of text boxes when creating or editing library entries. See this [link](https://github.com/koppor/jabref/issues/198?_pjax=%23js-repo-pjax-container)  for description

We created a new class that extended MenuItem to create the instance of the ClearField.
We added the instance into the menuItem array so that it would be included in the right click menu.

There is a possibility to make each text box a clearable text box, with a red "x" in the corner to clear the text box even quicker. We could not target where to change the type of text box, so that feature still remains. 

----


src/main/java/org/jabref/gui/fieldeditors/contextmenu/ClearField.java
```diff
@@ -0,0 +1,18 @@
...
+
+import javafx.scene.control.MenuItem;
+import javafx.scene.control.TextArea;
+
+import org.jabref.logic.formatter.bibtexfields.NormalizeNamesFormatter;
+import org.jabref.logic.l10n.Localization;
+
+class ClearField extends MenuItem {
+
+    private final TextArea opener;
```
tobiasdiez: I think there is no need for this private field since the variable is only used in the constructor anyway.

src/main/java/org/jabref/gui/fieldeditors/contextmenu/ClearField.java
```diff
@@ -0,0 +1,18 @@
...
+
+    public ClearField(TextArea opener) {
+        super(Localization.lang("Clear text"));
+        this.opener = opener;
+        setOnAction(event -> opener.setText(new NormalizeNamesFormatter().format("")));
```
tobiasdiez: Why do you use the names formatter here? I think `TextArea` has a `clear` method, or otherwise `setText("")` should work.

brandonwalker08: Hi, I was on the team with Weida that was working on this issue. We now realize that the things you mention were not needed. We were trying to follow how the other menu item classes were built, and that is why we added the private field. We found the class to add the menu item to the menuItem array by finding a similar issue that dealt with the Normalize name option. But we see now that we made it much more complicated than it needed to be. We should be able to fix these this week if we need to. 
Thanks

brandonwalker08:
We also see that the continuous build failed? I'm not sure how adding a small class could break it, so we'll look into this. If anyone sees an issue as well please let us know. Thanks

Siedlerchr:
Hi, you simply forgot to add a languagefile entry (from the end of the travis log)
```
    => java.lang.AssertionError: DETECTED LANGUAGE KEYS WHICH ARE NOT IN THE ENGLISH LANGUAGE FILE

1. PASTE THESE INTO THE ENGLISH LANGUAGE FILE

2. EXECUTE: gradlew localizationUpdate

[
Clear_text=Clear_text
] expected:<[]> but was:<[Clear_text (src/main/java/org/jabref/gui/fieldeditors/contextmenu/ClearField.java LANG)]>
```

PR #3473, opened 2017-11-30
Fetcher for IACR eprints

derTimme:
I'm working on a new fetcher for [IACR eprints](https://eprint.iacr.org/), as I need a lot of these for my current work.

IACR does not provide a real API, so the information needs to be parsed out of their (really simple) HTML.

@devs: Are there any special requirements for fetchers? (I couldn't find any, but the whole HTML parsing thing might be a problem concerning maintainability - on the other hand, I use IACR for quite a while now and the web interface hasn't changed (visibly at least)).

The idea is to have the following features:

- [x] Fetch based on ID (like `2017/1556`)
- [x] Fetched entry is of type `misc` and contains:
  - [x] Title
  - [x] Authors
  - [x] Url (contains the version)
  - [x] Note with `\url{https://eprint.iacr.org/<identifier>}` (recommended by IACR)
  - [x] Howpublished like `Cryptology ePrint Archive, Report <identifier>` (recommended by IACR)
  - [x] Version (most recent)
  - [x] Date (of the most recent version)
  - [x] Abstract
- [x] The encoding fits the library encoding (see b76a3f742f593e775a93997d67c94b2d954fc485, e. g. [`2017/1095`](https://eprint.iacr.org/2017/1095))

The following could also make sense:

- Get fulltext (if that means to download the pdf - I haven't used this feature with other fetchers yet, so I don't really know what exactly it does)
- Get the "also published in" information (see e.g. [this paper](https://eprint.iacr.org/2017/1118)) and DOI, if present (e.g. [this paper](https://eprint.iacr.org/2016/119))
- Get keywords
- Search in IACR preprints (e.g. for title) - but I suspect this would require us to download the complete archive or somehow "fill out" their search form (by crafting the corresponding POST request) programatically

----


src/main/java/org/jabref/logic/importer/fetcher/IacrEprintFetcher.java
```diff
@@ -0,0 +1,116 @@
...
+            } catch (Exception e) {
+                // Just skip
+            }
+        }
+
+        Collections.sort(formattedDates, Collections.reverseOrder());
+        return formattedDates.get(0);
+    }
+
+    private static String getValueBetween(String from, String to, String haystack) {
```
Siedlerchr: You could see if there is already a method in StringUtils

src/main/java/org/jabref/logic/importer/fetcher/IacrEprintFetcher.java
```diff
@@ -0,0 +1,116 @@
...
+
+
+        for (String rawDate : rawDates) {
+            try {
+                rawDate = rawDate.trim();
+                Matcher dateMatcher = Pattern.compile("[a-z ]+(\\d{1,2} [A-Za-z][a-z]{2} \\d{4})").matcher(rawDate);
+                if (dateMatcher.find()) {
+                    Date date = DATE_FORMAT_WEBSITE.parse(dateMatcher.group(1));
+                    formattedDates.add(DATE_FORMAT_BIBTEX.format(date));
+                }
```
Siedlerchr: Please always catch the most specific exceptions.

Siedlerchr:
Hi,
I think the html parsing is not really a problem. We already use that in other cases as well. You could also take a look at jsoup for parsing html contents/finding dat between elements (should be already included in our deps)

derTimme:
The problem with using libraries to retrieve the information from html is that their html seems to be quite buggy (using `<p />` for example).

src/main/java/org/jabref/logic/importer/fetcher/IacrEprintFetcher.java
derTimme: Thanks, there was indeed one. Keeping the functionality in a separate method though to have the error handling (nothing found) in one place.

derTimme:
From my point of view, the fetcher with its core features is finished.
The additional features mentioned in the initial comment are something I might tackle later on.

src/test/java/org/jabref/logic/importer/fetcher/IacrEprintFetcherTest.java
```diff
@@ -0,0 +1,121 @@
...
+import java.util.Optional;
+
+import org.jabref.logic.importer.FetcherException;
+import org.jabref.logic.importer.ImportFormatPreferences;
+import org.jabref.model.entry.BibEntry;
+import org.jabref.model.entry.BiblatexEntryTypes;
+import org.jabref.model.entry.FieldName;
+import org.jabref.testutils.category.FetcherTest;
+
+import org.junit.Before;
```
tobiasdiez: We just started to use JUnit 5 and it would be nice if you could use the new api (some of the other fetcher tests are already migrated).

src/test/java/org/jabref/logic/importer/fetcher/IacrEprintFetcherTest.java
```diff
@@ -0,0 +1,121 @@
...
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.mockito.Answers;
+
+import static org.junit.Assert.*;
+import static org.mockito.Mockito.mock;
+
+@Category(FetcherTest.class)
```
tobiasdiez: When converting to JUnit 5, please replace the category by the `FetcherTest` annotation.

src/main/java/org/jabref/logic/importer/fetcher/IacrEprintFetcher.java
```diff
@@ -37,6 +38,7 @@
     private static final Predicate<String> IDENTIFIER_PREDICATE = Pattern.compile("\\d{4}/\\d{3,5}").asPredicate();
     private static final String CITATION_URL_PREFIX = "https://eprint.iacr.org/eprint-bin/cite.pl?entry=";
     private static final String DESCRIPTION_URL_PREFIX = "https://eprint.iacr.org/";
+    private static final Charset WEBSITE_CHARSET = Charset.forName("iso-8859-1");
```
Siedlerchr: very very minor, you can directly use the predefined enum Constant: 
https://docs.oracle.com/javase/8/docs/api/java/nio/charset/StandardCharsets.html#ISO_8859_1

derTimme: Thanks for pointing that out - was looking for something like that, but apparently didn't look long enough...

Siedlerchr: For other cases where Java provides default enum variables, they all start with StandardXXX, for example for file opening there exists: StandardOpenOption, not really obvious if you search for it ;)

derTimme: Jep, I expected them in something like Charsets or directly as constants in the Charset class...

src/main/java/org/jabref/logic/importer/fetcher/IacrEprintFetcher.java
```diff
@@ -0,0 +1,158 @@
...
+import org.apache.commons.logging.LogFactory;
+
+public class IacrEprintFetcher implements IdBasedFetcher {
+
+    public static final String NAME = "IACR eprints";
+
+    private static final Log LOGGER = LogFactory.getLog(IacrEprintFetcher.class);
+    private static final Pattern DATE_FROM_WEBSITE_PATTERN = Pattern.compile("[a-z ]+(\\d{1,2} [A-Za-z][a-z]{2} \\d{4})");
+    private static final DateFormat DATE_FORMAT_WEBSITE = new SimpleDateFormat("dd MMM yyyy");
+    private static final DateFormat DATE_FORMAT_BIBTEX = new SimpleDateFormat("yyyy-MM-dd");
```
Siedlerchr: SimpleDateFormat is outdated, it has been replaced by several other constructs in java8: 
http://www.baeldung.com/java-8-date-time-intro
Or see example 18 here for an idea: http://javarevisited.blogspot.de/2015/03/20-examples-of-date-and-time-api-from-Java8.html

src/main/java/org/jabref/logic/importer/fetcher/IacrEprintFetcher.java
```diff
@@ -0,0 +1,158 @@
...
+        }
+
+        return entry;
+    }
+
+    private Optional<BibEntry> createEntryFromIacrCitation(String validIdentifier) throws FetcherException {
+        String bibtexCitationHtml = getHtml(CITATION_URL_PREFIX + validIdentifier);
+        String actualEntry = getValueBetween("<PRE>", "</PRE>", bibtexCitationHtml);
+
+        Optional<BibEntry> entry;
```
Siedlerchr: I am not sure, but you probably need to initialize it with Optional.empty() or you could get still an NPE if no entry is found

src/test/java/org/jabref/logic/importer/fetcher/IacrEprintFetcherTest.java
```diff
@@ -0,0 +1,119 @@
...
+    private static BibEntry delgado2017;
+
+    @BeforeAll
+    public static void setUp() {
+        fetcher = new IacrEprintFetcher(mock(ImportFormatPreferences.class, Answers.RETURNS_DEEP_STUBS));
+
+        abram2017 = new BibEntry();
+        abram2017.setType(BiblatexEntryTypes.MISC);
+        abram2017.setField("bibtexkey", "cryptoeprint:2017:1118");
+        abram2017.setField(FieldName.ABSTRACT, "The decentralized cryptocurrency Bitcoin has experienced great success but also encountered many challenges. One of the challenges has been the long confirmation time. Another challenge is the lack of incentives at certain steps of the protocol, raising concerns for transaction withholding, selfish mining, etc. To address these challenges, we propose Solida, a decentralized blockchain protocol based on reconfigurable Byzantine consensus augmented by proof-of-work. Solida improves on Bitcoin in confirmation time,  and provides safety  and liveness assuming the adversary control less than (roughly) one-third of the total mining power.\n");
```
Siedlerchr: Please no abstracts in tests, as they are usually subject to copyright of the publisher

src/main/java/org/jabref/logic/importer/fetcher/IacrEprintFetcher.java
derTimme: If the javadoc on BibtexParser.singleFromString is correct, it should always return an entry or an Optional.empty().
But I might as well initialize it...

src/test/java/org/jabref/logic/importer/fetcher/IacrEprintFetcherTest.java
derTimme: I'll look into the license details - not having the abstract there is a bit of a problem as fetching the abstract is part of the functionality; therefore it should be tested.
On fist glance, the whole article is published under CC-BY or CC-BY-NC; therefore this shouldn't be a problem as the authors are mentioned right next to the abstract text.
But if you still don't want the abstracts, I can modify the tests accordingly.

src/main/java/org/jabref/logic/importer/fetcher/IacrEprintFetcher.java
tobiasdiez: you can also just return the entry directly in the try construct. This is in my opinion the most readable solution.

src/test/java/org/jabref/logic/importer/fetcher/IacrEprintFetcherTest.java
```diff
@@ -0,0 +1,119 @@
...
+import org.mockito.Answers;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.mockito.Mockito.mock;
+
+@FetcherTest
+public class IacrEprintFetcherTest {
+
+    private static IacrEprintFetcher fetcher;
```
tobiasdiez: Please not static and use `BeforeEach` instead `BeforeAll` (better initialize the fetcher and entries fresh for each test)

src/main/java/org/jabref/logic/importer/fetcher/IacrEprintFetcher.java
derTimme: Didn't know that - I'm working on changing it.

src/test/java/org/jabref/logic/importer/fetcher/IacrEprintFetcherTest.java
tobiasdiez: Do you have a published article that can be received via IACR? In this case you hold the copyright and there is no problem.

src/main/java/org/jabref/logic/importer/fetcher/IacrEprintFetcher.java
derTimme: 👍

src/test/java/org/jabref/logic/importer/fetcher/IacrEprintFetcherTest.java
derTimme: Not yet ;) But then I'll just remove the abstracts for now and maybe just check that they are not empty.

src/test/java/org/jabref/logic/importer/fetcher/IacrEprintFetcherTest.java
derTimme: Thanks, it's my first time using JUnit 5 and I just looked at some examples online which used `BeforeAll`...
Should have been suspicious when adding the `static`s...

derTimme:
I just ran some "manual" tests for very old eprints and discovered that they indeed changed something in 2000 - up to 2000, the eprints don't have versions and the date format is a bit different.
I made this "WIP" again and am working on also supporting these old entries.

derTimme:
I now implemented support for pre-2000 entries; also added some tests for these as the pre-2000 entries don't seem to really have a "strong" standard format.
Luckily there aren't too much of these (about 65), so the tests just cover all of them.

I will now do some more "manual" testing, but I think everything will be finished in the next few hours.

src/main/resources/l10n/JabRef_da.properties
```diff
@@ -1819,6 +1819,11 @@ Copy_BibTeX_key_and_title=
 File_rename_failed_for_%0_entries.=
 Merged_BibTeX_source_code=
 Invalid_DOI\:_'%0'.=Ugyldig_DOI\:_'%0'.
+Invalid_IACR_identifier\:_'%0'.=
```
tobiasdiez: Please try to use a more generic version in these strings. As of now, they are not reusable in other fetchers or situations. E.g. just use `invalid identifier` or replace `ICAR` by a parameter slot.

derTimme:
I took a look at the `IdBasedParserFetcher` and if you insist, I could use it - but this fetcher works slightly different in that it requests *two* different URLs and builds the resulting entry with information from both those URLs.

tobiasdiez:
ok, then leave it like that. It was just a suggestion that came to my mind while browsing the code.

derTimme:
@lenhard I see your point - I haven't thought about that as I usually only use the "New entry" -> "ID based" way to fetch entries.
After looking into it, I don't think it belongs in the web search side bar as it only searches based on the ID of a paper; the DOI fetcher isn't in the side bar either. This makes sense from my point of view, as the web search side bar doesn't give any clues that it expects an ID. But if you think it makes sense to add it there, I can do that.

lenhard:
@derTimme Thanks for the explanation! What you write makes sense. There's no need to add the fetcher to the web search side bar as well. 

Unfortunately, there are some merge conflicts in the language files now. Could you please resolve those and then this is ready to go into master.

derTimme:
I'm really confused right now... I tried to merge in master and apparently it worked - in my local repo, the localization files look fine and I can do
```
git checkout master
git merge iacr_eprint_fetcher
```
without any conflicts...

Siedlerchr:
@derTimme  You have to first configure the JabRef/jabref repo as remote repository. Your master branch is still on the version when it was first forked. It does not get synced automatically.

[Gitbub Help on adding a cloned repo with a remote one](https://help.github.com/articles/fork-a-repo/#step-3-configure-git-to-sync-your-fork-with-the-original-spoon-knife-repository)

Add JabRef as remote repo:

```
git remote add upstream https://github.com/JabRef/jabref.git
git checkout <your branch>  
git merge upstream/master 
```

The last one merges the changes from the upstream repo (in this case the JabRef/jabref main repo)

Siedlerchr:
Okay, as all conflicts are resolved and the reviews are okay, I merge it now into master!
Thank you very much for your contribution! We hope you continue your contribution ;)

tobiasdiez:
I'll merge this now before you have to cope again with changes on the master branch. I'm sorry for the inconvenience caused by our recent change to the language files. Thanks for your contribution!

derTimme:
Thank you :)

PR #3470, opened 2017-11-29
[WIP] New Crowdin translations

koppor:
In comparison to transifex, we have the automatic GitHub integration.

## Pros and cons of crowdin

- `+`: Easy translation interface. Users can start joining at https://crwd.in/jabref.
- `+`: Nice status overview - see https://crowdin.com/project/jabref - internal link: https://crowdin.com/project/jabref/settings#translations - fixes fixes https://github.com/JabRef/help.jabref.org/issues/171
- `+` Translators do not require a login at GitHub and do not need to work with pull requests etc.
- `+` Translations can still be made using our GitHub workflow
- Exclamation marks are escaped. However, this is no issue, because we unescape them: https://github.com/JabRef/jabref/blob/d45d7ca381c31fab2e41e0693948b11e5068887b/src/main/java/org/jabref/logic/l10n/LocalizationKey.java#L25
- `-` At least one review of the translated string is required
- `-` We cannot create [our translation status page](https://github.com/JabRef/help.jabref.org/blob/gh-pages/en/TranslatingGUI.md#localization-files-status-2017-11-29-1957---branch-master-42ba568ea) anymore (using `syncLang.py`) as crowdin either (i)  does not put a key at all or (ii) puts the English translation in the properties file. Checking for equality does not help as `OK` is translated to `OK`.
- `-` It seems that generating the pull request has to be triggered manually.

I would vote to go for crowdin to make it easy for interested people to work on the translation.

## TODOs

- [ ] Check why our language file checker fails and adapt it accordingly
- [x] `id` vs `in`. Result: The ISO-Code for Indonesia changed from "in" to "id". Java uses only the old one. See https://docs.oracle.com/javase/7/docs/api/java/util/Locale.html#Locale(java.lang.String), but Crowdin uses the new one. With a new mapping configuration on Crowdin, this is fixed.

PR #3479, opened 2017-12-02
Prepare usage of crowdin: Remove underscores from l10n

koppor:
This PR prepares the usage of crowdin: No more strange `_` in our translation files. With this PR we are not bound to crowdin; we can still use our current GitHub worfklow. Furthermore, crowdin pushes new translations using pull requests.

A detailed discussion of crowdin can be found at https://github.com/JabRef/jabref/pull/3470.

grafik

### DIFF
--- a/src/main/resources/l10n/Menu_da.properties
+++ b/src/main/resources/l10n/Menu_da.properties
@@ -1,7 +1,3 @@
-#!
-#! created/edited by Popeye version 0.55 (github.com/JabRef/popeye)
-#! encoding:UTF-8
-
 Abbreviate\ journal\ names\ (ISO)=Forkort tidsskriftsnavn (ISO)
 Abbreviate\ journal\ names\ (MEDLINE)=Forkort tidsskriftsnavn (MEDLINE)
 About\ JabRef=Om &JabRef
--- a/src/main/resources/l10n/Menu_da.properties
+++ b/src/main/resources/l10n/Menu_da.properties
@@ -2,139 +2,139 @@
 #! created/edited by Popeye version 0.55 (github.com/JabRef/popeye)
 #! encoding:UTF-8
 
-Abbreviate_journal_names_(ISO)=Forkort_tidsskriftsnavn_(ISO)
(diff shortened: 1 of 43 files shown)


src/main/resources/l10n/Menu_da.properties
```diff
@@ -1,140 +1,136 @@
...
-#! created/edited by Popeye version 0.55 (github.com/JabRef/popeye)
-#! encoding:UTF-8
-
-Abbreviate_journal_names_(ISO)=Forkort_tidsskriftsnavn_(ISO)
-Abbreviate_journal_names_(MEDLINE)=Forkort_tidsskriftsnavn_(MEDLINE)
-About_JabRef=Om_&JabRef
-Append_library=Tilføj_ind&hold_fra_library
-Autogenerate_BibTeX_keys=&Autogenerer_BibTeX-nøgler
-Close_library=L&uk_library
+Abbreviate\ journal\ names\ (ISO)=Forkort tidsskriftsnavn (ISO)
```
Siedlerchr: But now you replaced the underscores with backslashes. Compared to that I find underscores better

koppor:
Doesn't work otherwise in the keys, because white spaces are treated as
delimiter between key and value.
https://de.m.wikipedia.org/wiki/Java-Properties-Datei

Using underscore was a JabRef proprietary way. Using escapes seems to be
the usual way.

Also note that all the escaping in thr value field is optional!

Am 02.12.2017 12:52 schrieb "Christoph" <notifications@github.com>:

koppor:
- [ ] adapt script to handle new entries also with `\ `

DevCall: merge after that is fixed.

koppor:
OK, `org.jabref.logic.l10n.LocalizationConsistencyTest#findMissingLocalizationKeys` now works also with the value vield:

```
java.lang.AssertionError: DETECTED LANGUAGE KEYS WHICH ARE NOT IN THE ENGLISH LANGUAGE FILE
1. PASTE THESE INTO THE ENGLISH LANGUAGE FILE
2. EXECUTE: gradlew localizationUpdate
[
No\ more\ escaped\ characters\ in\ the\ value\ field=No more escaped characters in the value field
] 
Expected :[]
```

PR #3484, opened 2017-12-04
[WIP] Delete from group

mpele:
This pull request is continuation of #3071

The default dialog for confirmation of delete entries is changed so the user have option to choose if he wants to delete entry, or to just remove entry from active group. I have few times deleted articles from database not being aware that article belongs to other groups. 
New dialog should make clear to user what is happening when he press "Delete" and offer to just remove from selected groups.

At this phase I need help with defining text in dialog, so any advice is welcome.

----

### DIFF
--- a/src/main/java/org/jabref/gui/BasePanel.java
+++ b/src/main/java/org/jabref/gui/BasePanel.java
@@ -20,6 +20,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
@@ -44,6 +45,7 @@ import javafx.application.Platform;
 
 import org.jabref.Globals;
 import org.jabref.JabRefExecutorService;
+import org.jabref.gui.DeleteDialog.DeleteResult;
 import org.jabref.gui.actions.Actions;
 import org.jabref.gui.actions.BaseAction;
 import org.jabref.gui.actions.CleanupAction;
@@ -143,7 +145,6 @@ import org.jabref.model.groups.GroupTreeNode;
 import org.jabref.preferences.JabRefPreferences;
 import org.jabref.preferences.PreviewPreferences;
 import org.jabref.shared.DBMSSynchronizer;
-
 import com.google.common.eventbus.Subscribe;
 import com.jgoodies.forms.builder.FormBuilder;
 import com.jgoodies.forms.layout.FormLayout;
@@ -769,7 +770,7 @@ public class BasePanel extends JPanel implements ClipboardOwner {
      */
     private void delete(boolean cut) {
     	boolean bRemoveFromGroup = false;
-    	boolean bDelete = false;
+    	boolean delete = false;
     	List<FieldChange> changesRemove = new ArrayList<>();
     	
     	List<BibEntry> entries = mainTable.getSelectedEntries();
@@ -778,35 +779,53 @@ public class BasePanel extends JPanel implements ClipboardOwner {
             return;
         }
     	
-    	String selectedGroup = "";
-    	if (Globals.stateManager.getSelectedGroup(bibDatabaseContext).size() > 0) {
-    		selectedGroup = Globals.stateManager.getSelectedGroup(bibDatabaseContext).get(0).getName();
-    	}
-
-        //get list of all groups
+        
+        //create list of all groups of selected entries
+        Set<String> allEnriesGroups = new HashSet<String>(); {
+		};
+		for(BibEntry entrie : entries){
+			if(entrie.getField("groups").isPresent()){
+				System.out.println(entrie.getField("groups"));
+				String[] groups = entrie.getField("groups").get().split(",");
+				for(String group : groups){
+					allEnriesGroups.add(group);
+				}
+			}
+		}
+        System.out.println(allEnriesGroups);
+        
+        
+    	//get list of all groups
         Optional<GroupTreeNode> groups = bibDatabaseContext.getMetaData().getGroups();
 
         if (groups.get().getName().contentEquals(groups.get().getName())) { //checks if is selected any group
         	//TODO define dialog not to offer "remove from group"
         }
 
-		Object[] options = {Localization.lang("Delete_from_database"),
-			Localization.lang("Remove_from_group") + " \"" + selectedGroup + "\"",
-			Localization.lang("Cancel")};
-		int n = JOptionPane.showOptionDialog(frame,
-				Localization.lang("The_entry_is_the_member_of_groups") + entries.get(0).getField("groups"),
-				Localization.lang("Delete_remove_entry"),
-				JOptionPane.YES_NO_CANCEL_OPTION,
-				JOptionPane.QUESTION_MESSAGE,
-				null,
-				options,
-				options[2]);
+        
+        DeleteDialog diag = new DeleteDialog(Globals.stateManager.getSelectedGroup(bibDatabaseContext), allEnriesGroups);
+        diag.setVisible(true);
 
-		if (n == 0) { // Delete
-			bDelete = true;
-		} else if (n == 1) { //Remove from group(s) 
+        DeleteResult resultDialog = diag.getResult();
+        System.out.println(resultDialog);
+        
+//		Object[] options = {Localization.lang("Delete_from_database"),
+//			Localization.lang("Remove_from_group") + " \"" + selectedGroup + "\"",
+//			Localization.lang("Cancel")};
+//		int n = JOptionPane.showOptionDialog(frame,
+//				Localization.lang("The_entry_is_the_member_of_groups") + entries.get(0).getField("groups"),
+//				Localization.lang("Delete_remove_entry"),
+//				JOptionPane.YES_NO_CANCEL_OPTION,
+//				JOptionPane.QUESTION_MESSAGE,
+//				null,
+//				options,
+//				options[2]);
+
+		if (resultDialog == DeleteResult.DELETE_FROM_DATABASE) { // Delete
+			delete = true;
+		} else if (resultDialog == DeleteResult.REMOVE_FROM_GROUPS) { //Remove from group(s) 
 			bRemoveFromGroup = true;
-		} else if (n == 2) { //Cancel
+		} else if (resultDialog == DeleteResult.CANCEL) { //Cancel
 			return;
 		}
 
@@ -831,7 +850,7 @@ public class BasePanel extends JPanel implements ClipboardOwner {
 		}
 
         //if (!cut && !bRemoveFromGroup && !showDeleteConfirmationDialog(entries.size())) {
-        if (!cut && !bDelete) {
+        if (!cut && !delete) {
              return;
         }
 
--- a/src/main/java/org/jabref/gui/BasePanel.java
+++ b/src/main/java/org/jabref/gui/BasePanel.java
@@ -769,12 +769,10 @@ public class BasePanel extends JPanel implements ClipboardOwner {
      *            "deleted". If true the action will be localized as "cut"
      */
     private void delete(boolean cut) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/jabref/gui/BasePanel.java
```diff
@@ -764,14 +769,57 @@ private void cut() {
...
         if (entries.isEmpty()) {
             return;
         }
-        if (!cut && !showDeleteConfirmationDialog(entries.size())) {
-            return;
+    	
+        //create list of all groups of selected entries
+        Set<String> allEnriesGroups = new HashSet<String>(); 
+
+        for (BibEntry entrie : entries) {
```
lenhard: The correct spelling here should be `entry`.

src/main/java/org/jabref/gui/BasePanel.java
```diff
@@ -764,14 +769,57 @@ private void cut() {
...
+        	if (entrie.getField("groups").isPresent()) {
+        		System.out.println(entrie.getField("groups"));
```
lenhard: I guess this is for you to help in debugging, but in the final PR, there shouldn't be any `System.out` or commented-out code left. I'll not add a comment at every place in this PR, but I am sure you get the point.

src/main/java/org/jabref/gui/BasePanel.java
```diff
@@ -764,14 +769,57 @@ private void cut() {
      *            "deleted". If true the action will be localized as "cut"
      */
     private void delete(boolean cut) {
```
lenhard: In the end, you could probably extract parts of this method into separate methods or maybe even extract a class here and move that into `org.jabref.gui.groups`. I am a bit wary about `BasePanel` growing even more. The class is already far too big.

src/main/java/org/jabref/gui/DeleteDialog.java
```diff
@@ -0,0 +1,129 @@
+package org.jabref.gui;
```
lenhard: This should be moved into `org.jabref.gui.groups`.

src/main/java/org/jabref/gui/DeleteDialog.java
```diff
@@ -0,0 +1,129 @@
...
+import javax.swing.JLabel;
+import javax.swing.JComponent;
+
+import java.awt.event.ActionListener;
+import java.awt.event.KeyEvent;
+import java.util.Iterator;
+import java.util.Set;
+import java.awt.event.ActionEvent;
+
+public class DeleteDialog extends JDialog {
```
lenhard: I think `GroupDeleteDialog` would be an even better name here.

src/main/java/org/jabref/gui/DeleteDialog.java
```diff
@@ -0,0 +1,129 @@
...
+		if (!allEnriesGroups.isEmpty()) {
+			message += Localization.lang("Selected_entries_belongs_to_group(s)") + ": <br/><ul>";
+			Iterator<String> iterator = allEnriesGroups.iterator();
+			while (iterator.hasNext()) {
+				message += "<li>" + iterator.next() + "</li>"; 
+			}
+			message += "</ul>" + Localization.lang("If_you_delete_articles_it_will_be_removed_from_all_listed_groups") + "<br>";
+		}
+
+		JLabel lblThisEntitiesAre = new JLabel("<html> "+ message + " </html>");
```
lenhard: Funny name :) How about just `userMessage` or something similar?

src/main/java/org/jabref/gui/DeleteDialog.java
```diff
@@ -0,0 +1,129 @@
...
+		panel.add(lblThisEntitiesAre);
+
+		JPanel buttonPane = new JPanel();
+		getContentPane().add(buttonPane);
+		buttonPane.setLayout(new BoxLayout(buttonPane, BoxLayout.X_AXIS));
+
+		JButton btnDeleteFromDatabase = new JButton(Localization.lang("Delete_from_database"));
```
lenhard: As above, please spell out terms in names, e.g. `deleteFromDatabase`

src/main/java/org/jabref/gui/DeleteDialog.java
```diff
@@ -0,0 +1,129 @@
...
+		setTitle("Delete");
+		setModal(true);
+		getContentPane().setLayout(new BoxLayout(getContentPane(), BoxLayout.Y_AXIS));
+
+		JPanel panel = new JPanel();
+		getContentPane().add(panel);
+
+		String message = "";
+		if (selectedGroups.size() > 0) {
+			message += Localization.lang("Currently_there_are_selected_group(s)")+": <br/><ul>";
```
lenhard: I would suggest `There_are_group(s)_selected`

src/main/java/org/jabref/gui/DeleteDialog.java
```diff
@@ -0,0 +1,129 @@
...
+			for(int i = 0; i < selectedGroups.size(); i++){
+				message += "<li>"+selectedGroups.get(i).getName() + "</li>"; 
+			}
+			message += "</ul>" + Localization.lang("If_you_would_like_to_remove_selected_entities_from_selected_groups_<br/>press_button")
```
lenhard: Suggestion `Do_you_want_to_remove_the_selected_entries_from_the_selected_groups`

src/main/java/org/jabref/gui/DeleteDialog.java
```diff
@@ -0,0 +1,129 @@
...
+				message += "<li>"+selectedGroups.get(i).getName() + "</li>"; 
+			}
+			message += "</ul>" + Localization.lang("If_you_would_like_to_remove_selected_entities_from_selected_groups_<br/>press_button") 
+			+ "\" " 
+			+ Localization.lang("Remove_from_groups") 
+			+ "\"<br/><br/>";
+		}
+		
+		if (!allEnriesGroups.isEmpty()) {
+			message += Localization.lang("Selected_entries_belongs_to_group(s)") + ": <br/><ul>";
```
lenhard: Suggestion: `Selected_entries_belong_to_group(s)`

src/main/java/org/jabref/gui/DeleteDialog.java
```diff
@@ -0,0 +1,129 @@
...
+			Iterator<String> iterator = allEnriesGroups.iterator();
+			while (iterator.hasNext()) {
+				message += "<li>" + iterator.next() + "</li>"; 
+			}
+			message += "</ul>" + Localization.lang("If_you_delete_articles_it_will_be_removed_from_all_listed_groups") + "<br>";
```
lenhard: Suggestion: `If_you_delete_the_entries_they_will_be_removed_from_all_listed_groups`

src/main/java/org/jabref/gui/DeleteDialog.java
tobiasdiez: And everything without the underscores since otherwise the tests fail.

mpele: If this format is not correct, could you give me one example of text in Localization.lang()?

src/main/java/org/jabref/gui/DeleteDialog.java
mpele: The suggested text doesn't give message to user about action that should or should not be done.

src/main/java/org/jabref/gui/groups/DeleteDialog.java
```diff
@@ -0,0 +1,125 @@
...
+import javax.swing.JLabel;
+import javax.swing.JComponent;
+
+import java.awt.event.ActionListener;
+import java.awt.event.KeyEvent;
+import java.util.Iterator;
+import java.util.Set;
+import java.awt.event.ActionEvent;
+
+public class DeleteDialog extends JDialog {
```
tobiasdiez: Can you please use JavaFX to implement this dialog. We already have a quite extensive `DialogService` class that you can use.

src/main/java/org/jabref/gui/groups/DeleteDialog.java
```diff
@@ -0,0 +1,125 @@
...
+	 */
+	public DeleteDialog(ObservableList<GroupTreeNode> selectedGroups, Set<String> allEnriesGroups) {
+		setTitle("Delete");
+		setModal(true);
+		getContentPane().setLayout(new BoxLayout(getContentPane(), BoxLayout.Y_AXIS));
+
+		JPanel panel = new JPanel();
+		getContentPane().add(panel);
+
+		String message = "";
```
tobiasdiez: In my opinion this dialog right now is a bit to complicated. My proposal:
Main text: "The entry / entries also belongs to the currently selected group(s). Do you want to remove it from the group or delete it completely?" (maybe with better english :-) )
Options: "Remove from group(s)", "Delete from database", "Cancel"

tobiasdiez: Moreover, I miss a "Remember this choice" checkbox.

src/main/java/org/jabref/gui/groups/DeleteDialog.java
```diff
@@ -0,0 +1,125 @@
...
+		if (selectedGroups.size() > 0) {
+			message += Localization.lang("There_are_group(s)_selected") + ": <br/><ul>";
+			for (int i = 0; i < selectedGroups.size(); i++) {
+				message += "<li>" + selectedGroups.get(i).getName() + "</li>"; 
+			}
+			message += "</ul>" + Localization.lang("If_you_would_like_to_remove_selected_entities_from_selected_groups_<br/>press_button") 
+			+ "\" " 
+			+ Localization.lang("Remove_from_groups")
```
tobiasdiez: Language text no longer work with underscore. Just use a normal space character.

src/main/java/org/jabref/gui/BasePanel.java
```diff
@@ -758,20 +764,57 @@ private void cut() {
...
+        for (BibEntry entry : entries) {
+        	if (entry.getField("groups").isPresent()) {
+        		System.out.println(entry.getField("groups"));
+        		String[] groups = entry.getField("groups").get().split(",");
+        		for (String group : groups){
+        			allEnriesGroups.add(group);
+        		}
+        	}
+        }
+        System.out.println(allEnriesGroups);
```
tobiasdiez: Please remove these debugging statements.

src/main/java/org/jabref/gui/groups/DeleteDialog.java
mpele: Hi @tobiasdiez, 
Thanks for comments.
The main purpose of this PR is to notify that some entry could belong to some groups that couldn't be identified by watching on screen (groups that are not selected). If we remove that information the new dialog is useless.

tobiasdiez: Ok, then I don't understand the use case. I thought you wanted to prevent that a user by accident deletes an entry although he just wanted to remove it from the currently selected groups. Why is the information about the other groups important? Moreover, if you select an entry, then groups that contain this entry are highlighted in the group pane so the information is already there (in some form).

stefan-kolb:
What is the current status of this PR? Will you be working on this in the near future?

mpele:
According to requests from Tobias this PR is deprecated.

koppor:
We generally like the idea and think, it will help users to work with JabRef more efficently. In the context of our JavaFX migration, it would be nice to have the dialog in JavaFX. Moreover, some nasty details (as outlined by @tobiasdiez) should be treated. @mpele would you be so nice to invest the energy to work on that?

mpele:
First of all, due the some deadlines that I have, I am not able to promise
that it will be finished in next few weeks.
And most important, I am not sure what is expected to be done. According to
yours opinion what should be the context of the dialog and what choices
should be offered to user?

stefan-kolb:
Ping 😄

Issue #3307, opened 2017-10-15
Cannot resolve '\textendash', etc. when displaying article title


psychelzh:
JabRef version `4.0` on Windows 10

See the following screenshot. `\textendash` cannot be resolved to the corresponding character.

![image](https://user-images.githubusercontent.com/13306020/31583271-bd65aa0e-b1ca-11e7-8d63-6a282aaa4f71.png)

![image](https://user-images.githubusercontent.com/13306020/31584134-f5601196-b1da-11e7-9504-fa87875c7413.png)

psychelzh:
Maybe other `LaTex` [Text commands (encoding)](http://tex.loria.fr/ctan-doc/macros/latex/doc/html/usrguide/node28.html) are not resolved, either.

lenhard:
For the record, this is not a problem in latex2unicode which translates the `\textendash` correctly. I guess this s a part of the html translation for the preview.

psychelzh:
Yeah, you are right.
![image](https://user-images.githubusercontent.com/13306020/31586301-9b2011dc-b201-11e7-9644-220b48b2eae1.png)

lenhard:
I added the command to the conversion table. Could you please check out the version here: http://builds.jabref.org/textendash/ and let me know if the problem is resolved at your end?

psychelzh:
Thank you, and I will test it tomorrow, because it is too late here in China.

psychelzh:
Greatm, it works! Here is a screenshot.
![image](https://user-images.githubusercontent.com/13306020/31592237-64bed580-b259-11e7-829e-2b0c28786d40.png)

lenhard:
It's a bit difficult to say what all means here, since we don't really know what is missing.

Could you perhaps provide a list of encoding symbols that you want, but which do not work at the moment? With such a list, it should be no big hassle to complete the implementation.

psychelzh:
No more wrongly encoding ones encountered, only some supported ones, say `\textquotedblleft` and `\textquotedblright`. I guess there could be a complete list in the `LaTex` documentation, e.g. [one](http://tex.loria.fr/ctan-doc/macros/latex/doc/html/usrguide/node28.html) that I had found, but would it be a huge work?

lenhard:
We're trying to minimize developer time consumption :) Hence, it would be cool if someone just checks which of the things listed on the page you you linked are missing.

But overall there are not too many symbols, so this should be doable.

psychelzh:
[...]

PR #3487, opened 2017-12-05
Add \textquotesingle to the html conversion table

psychelzh:
`LaTex` use various escape strings to denote certain symbols. Here I add {\textquotesingle} to the html conversion table, which represents apostrophe (').

----

### DIFF
--- a/src/main/java/org/jabref/logic/util/strings/HTMLUnicodeConversionMaps.java
+++ b/src/main/java/org/jabref/logic/util/strings/HTMLUnicodeConversionMaps.java
@@ -474,7 +474,7 @@ public class HTMLUnicodeConversionMaps {
             {"35", "", "\\#"}, // Hash
             {"36", "dollar", "\\$"}, // Dollar
             {"37", "#37", "\\%"}, // Percent (&percnt; is not displayed correctly in the entry preview)
-            {"39", "apos", "'"}, // Apostrophe                        
+            {"39", "apos", "'"}, // Apostrophe
             {"40", "lpar", "("}, // Left bracket
             {"41", "rpar", ")"}, // Right bracket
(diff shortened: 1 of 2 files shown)


psychelzh:
Add one more symbol. Related to #3307

src/main/java/org/jabref/logic/util/strings/HTMLUnicodeConversionMaps.java
```diff
@@ -475,6 +475,7 @@
             {"36", "dollar", "\\$"}, // Dollar
             {"37", "#37", "\\%"}, // Percent (&percnt; is not displayed correctly in the entry preview)
             {"39", "apos", "'"}, // Apostrophe
+            {"39", "apos", "{\\textquotesingle}"}, // Apostrophe
```
lenhard: Sorry, but I don't think this is going to work this way. This clashes with the already existing mapping for unicode value 39 in the line above and will lead to unpredictable results later.

I think it would be better to add the conversion as part of the "manual additions" to `LATEX_HTML_CONVERSION_MAP` and `LATEX_UNICODE_CONVERSION_MAP` , like for the `relax` term. This should have the same effect, but it avoids the clash for the other maps.

psychelzh: So how can it be done? I just do not know the details. I think `{\textquotesingle}` as an alias of `'`.

lenhard: Have a look at lines 897 and 898 in the same class. You can add the symbol in the very same way.

lenhard:
Thanks, perfect now :) I'll merge directly.

Issue #3471, opened 2017-11-29
space character causes cursor to jump to start of line


wujastyk:
JabRef 4.1-dev--snapshot--2017-11-27--master--42a959d1d
Linux 4.10.0-40-generic amd64 
Java 1.8.0_151

1. Open JR
2. Go to one of the editing tabs
3. Go to the end of a piece of text, like author or title
4. Press space bar
5. Cursor repositions itself at the start of the line. This happens with all subsequent space-bar presses. 

I don't understand why nobody else has reported this, since it makes JR unusable.  It's been an issue for over a week.  I guess it's a local problem on my system.  I normally run IBUS and the m17n keyboard library, but this problem still exists if I unload those utilities.

Maybe there's something else going on with my space-key?

Log copy:

Opening: /home/dom/Dropbox/localtexmf/bibtex/bib/biblio4-utf8.bib
Uncaught exception occurred in Thread[JavaFX Application Thread,5,main]
java.lang.NullPointerException
	at org.jabref.gui.search.GlobalSearchBar.lambda$new$5(GlobalSearchBar.java:196)
	at com.sun.javafx.application.PlatformImpl.lambda$null$172(PlatformImpl.java:295)
	at java.security.AccessController.doPrivileged(Native Method)
	at com.sun.javafx.application.PlatformImpl.lambda$runLater$173(PlatformImpl.java:294)
	at com.sun.glass.ui.InvokeLaterDispatcher$Future.run(InvokeLaterDispatcher.java:95)
	at com.sun.glass.ui.gtk.GtkApplication._runLoop(Native Method)
	at com.sun.glass.ui.gtk.GtkApplication.lambda$null$48(GtkApplication.java:139)
	at java.lang.Thread.run(Thread.java:748)

Siedlerchr:
I noticed that this happens sometimes on my windows system, too. But I just can not really reproduce it, But at least you got an exception in your log (maybe an other issue but still helpful)

tobiasdiez:
I also had it today. Not sure about the origin (no exception on my side).

sambo57u:
I can confirm this on Linux x86_64 with the above setup (but no exception).

wujastyk:
Temporary workaround: if you type any character into a field, say "x", you can then enter data normally to the left of the "x".  It's acting as a sort of terminator, and stops the cursor misbehaving when space is entered.

tobiasdiez:
I'll have a look at this issue this weekend, if nobody else is quicker.

[...]

PR #3497, opened 2017-12-08
Fix NPE in search bar

tobiasdiez:
In #3471 a NPE was mentioned. I'm not sure which object was null (it should not be the scene, since this is just set and the css file should also be always found). Moreover, I could not reproduce the exception, so this fix is a shot in the blue.

----


Siedlerchr:
Seems like this could be a LInux thing and related to our customjfx panel, see also the log in #3500

tobiasdiez:
I merge it now since it is still a slight improvement over the existing code. I'll have a look at the azure reports if a similar exception gets reported in the next days.

PR #3499, opened 2017-12-08
Update to new Code Signing Certificate 2017

LinusDietz:
- [x] check build pipeline
- [ ] Changelog

Works for me on MacOS, but there's a problem with Windows: The root certificate (TeleSec) is not configured to allow code signing per [default](https://www.pki.dfn.de/faqpki/faqpki-codesigning0). The user could, in theory, change that, but well... who does that.

This solution: Don't sign Windows, sign Macos.

Test binaries: http://builds.jabref.org/new-cert


LinusDietz:
This is a screenshot on MacOS if you try to install a binary **WITHOUT** code signing. It's not possible.
screen shot 2017-12-08 at 15 25 18
screen shot 2017-12-08 at 15 25 23

LinusDietz:
This is a screenshot (thanks @Siedlerchr ) when you try to install an entirely unsigned .exe on windows.
64d45a4b-96a7-4f8d-a065-e4f17b092be3

LinusDietz:
I'll close this now, as the certificate does not seem to be accepted by the major operating systems.

PR #3503, opened 2017-12-08
Update internal state of DatabaseChangeMonitor when external changes …

### LINKED ISSUE #3498: external changes to bib file make JabRef go mad
- [ ] I have tested the latest master version from http://builds.jabref.org/master/ and the problem persists

> “JabRef_macos_4_1-dev--snapshot--2017-12-08--master--6f39151a7.dmg” is damaged and can’t be opened. You should move it to the Trash.

JabRef version 4.1-dev

Steps to reproduce:

1. make changes to bib file with another editor
2. JabRef will ask you to review the changes, you acknowledge them
3. you try to save the bib file; JabRef will tell you that the file has had been modified externally
4. you go for "save" and JabRef will tell you that the library is protected and that it won't save until "changes have been reviewed"
5. you try again with "Review changes" in the save dialog but nothing happens
6. you now want to close and reopen JabRef, but it will tell you that your library has changed
7. again, "save changes" won't work so you for "discard"
8. it will hang with the message "waiting for save operation to finish"
9. you kill JabRef

lenhard:
…are resolved

Attempts to fix #3498

It seems that the internal state of DatabaseChangeMonitor was not updated correctly when external changes were marked as resolved, causing the monitor to misbehave when you tried to save afterwards. I am not sure how this bug came into being and what caused the update to get lost. Adding an additional update step seems to resolve the problem, but I would really appreciate some more testing of this branch by other people.

----

### DIFF
--- a/src/main/java/org/jabref/gui/BasePanel.java
+++ b/src/main/java/org/jabref/gui/BasePanel.java
@@ -1984,7 +1984,7 @@ public class BasePanel extends JPanel implements ClipboardOwner {
     }
 
     public void updateTimeStamp() {
-        changeMonitor.ifPresent(DatabaseChangeMonitor::updateTimestampAndFileSize);
+        changeMonitor.ifPresent(DatabaseChangeMonitor::markAsSaved);
     }
 
     public Path getTempFile() {
--- a/src/main/java/org/jabref/gui/BasePanel.java
+++ b/src/main/java/org/jabref/gui/BasePanel.java
@@ -1984,7 +1984,7 @@ public class BasePanel extends JPanel implements ClipboardOwner {
     }
 
     public void updateTimeStamp() {
-        changeMonitor.ifPresent(DatabaseChangeMonitor::markAsSaved);
+        changeMonitor.ifPresent(DatabaseChangeMonitor::updateTimestampAndFileSize);
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/jabref/gui/BasePanel.java
```diff
@@ -1984,7 +1984,7 @@ public void resetChangeMonitor() {
     }
 
     public void updateTimeStamp() {
-        changeMonitor.ifPresent(DatabaseChangeMonitor::markAsSaved);
+        changeMonitor.ifPresent(DatabaseChangeMonitor::updateTimestampAndFileSize);
```
tobiasdiez: I would actually prefer to rename `updateTimeStamp()` to `markAsSaved`. That the `DatabaseChangeMonitor` uses the timestamp and file size to keep track of the last saved state, should be an implementation detail that is hidden to the outside world.

lenhard: Alright, I changed the name back to what it was before.

lenhard:
It seems we're having a bit of a shortage of reviewers at the moment ;)

Since this is such a small one-line change that actually fixes a bug, I'll merge this now even if there's only one review.

Issue #2533, opened 2017-02-09
Citation preview incrementally increases memory usage with each preview


kingwarrick:
JabRef 3.8.2
windows 10 10.0 amd64 
Java 1.8.0_121

RAM is used during preview, but not fully released. With more previews, less is released until run-away usage occurs. After ~4-5 previews (or ~500MB RAM, not sure which is more significant), no appreciable amount of RAM is released, with subsequent previews adding 100-500MB per preview to total RAM usage. Behaviour changes after ~1400MB RAM use, with slower incremental use and more frequent release of small amounts of ram.

Tested with different preview styles (no change); interesting to note how CPU use in IEEE preview generation spikes to ~40%, whereas APA spikes to ~20%. 

NOT tested with any release versions.

Steps to reproduce [RAM use in square brackets]:

1. Open JabRef and bibtex database (~900 entries) 
2. Single click on entry to show preview box, wait for preview to show
3. Single click on another entry, wait for preview to show

This shows in the exceptions tab of the error console:
```
ANTLR Tool version 4.5.3 used for code generation does not match the current runtime version 4.6ANTLR Runtime version 4.5.3 used for parser compilation does not match the current runtime version 4.6ANTLR Tool version 4.5.3 used for code generation does not match the current runtime version 4.6ANTLR Runtime version 4.5.3 used for parser compilation does not match the current runtime version 4.6ANTLR Tool version 4.5.3 used for code generation does not match the current runtime version 4.6ANTLR Runtime version 4.5.3 used for parser compilation does not match the current runtime version 4.6ANTLR Tool version 4.5.3 used for code generation does not match the current runtime version 4.6ANTLR Runtime version 4.5.3 used for parser compilation does not match the current runtime version 4.6
```

kingwarrick:
Similar to this:
https://github.com/JabRef/jabref/issues/2247
but for preview rather than entry editing

>tobiasdiez commented on Nov 8 2016 
>I can confirm at least the large memory footprint. After opening a normal db (mine has 500 entries and a few groups) JabRef eats only 200 MB of RAM. **Now open the entry editor and run through a few entries (using the arrow keys so that a new entry editor is generated for the entries).** Result: over 1 GB of RAM usage. Btw: the same db with 3.6 only needed < 100 MB RAM.

koppor:
Yes, we cache the preview to gain speed when showing the second time.

@bartschdev What is our policy for cache clearance?

[...]

PR #3504, opened 2017-12-08
[WIP] Convert entry preview panel to JavaFX

tobiasdiez:
I migrated the entry preview to JavaFX. My hope was to resolve the memory issues #2533, but this was sadly not successful (thus the issues really come from the citation style cache and not from the display). 
Notable differences:
- The default JavaFX font seems to be bigger. In my opinion, it looks better that way.
- The description for two options in the right-click menu has a display bug. The reason is that I use `lang.menuTitle` to get the text but menu titles contain the acceleration key binding. Probably one needs to convert the main toolbar to JavaFX as well in order to fix this problem.

----


src/main/java/org/jabref/gui/BasePanel.java
```diff
@@ -239,6 +245,16 @@ public BasePanel(JabRefFrame frame, BibDatabaseContext bibDatabaseContext) {
...
 
         this.getDatabase().registerListener(new UpdateTimestampListener(Globals.prefs));
+
+        this.preview = new PreviewPanel(this, getBibDatabaseContext());
+        DefaultTaskExecutor.runInJavaFXThread(() -> frame().getGlobalSearchBar().getSearchQueryHighlightObservable().addSearchListener(preview));
+        this.previewContainer = setupPreviewContainer(preview);
+    }
+
+    private JFXPanel setupPreviewContainer(PreviewPanel preview) {
+        JFXPanel container = OS.LINUX ? new CustomJFXPanel() : new JFXPanel();
```
lenhard: I don't like the fact this chunk of three lines is duplicated all over the place. Maybe add a new factory class that provides a method to build a `JFXPanel` but hides the OS aspect.

src/main/java/org/jabref/gui/PreviewPanel.java
```diff
@@ -51,134 +39,103 @@
...
             }
         });
+    }
 
+    private ContextMenu createPopupMenu() {
+        MenuItem copyPreview = new MenuItem(Localization.lang("Copy preview"), IconTheme.JabRefIcon.COPY.getGraphicNode());
+        copyPreview.setOnAction(event -> copyPreviewToClipBoard());
+        MenuItem printEntryPreview = new MenuItem(Localization.lang("Print entry preview"), IconTheme.JabRefIcon.PRINTED.getGraphicNode());
+        printEntryPreview.setOnAction(event -> print());
+        MenuItem previousPreviewLayout = new MenuItem(Localization.menuTitle("Previous preview layout"));
```
lenhard: On my machine, the localization string appears prefixed with an `&`. Even if that's not nicely solvable here, could you at least implement a quick hack to remove the preceeding `&`?

Also, in the Swing version the context menu text also listed the shortcuts for the menu item. This should be the case here as well.

src/main/java/org/jabref/gui/PreviewPanel.java
```diff
@@ -51,134 +39,103 @@
...
+        previousPreviewLayout.setOnAction(event -> basePanel.ifPresent(BasePanel::previousPreviewStyle));
+        MenuItem nextPreviewLayout = new MenuItem(Localization.menuTitle("Next preview layout"));
```
lenhard: Same here with the display bug.

src/main/java/org/jabref/gui/PreviewPanel.java
```diff
@@ -264,140 +221,76 @@ public BibEntry getEntry() {
...
-                } catch (PrinterException e) {
-                    // Inform the user... we don't know what to do.
-                    JOptionPane.showMessageDialog(PreviewPanel.this,
-                            Localization.lang("Could not print preview") + ".\n" + e.getMessage(),
-                            Localization.lang("Print entry preview"), JOptionPane.ERROR_MESSAGE);
-                    LOGGER.info("Could not print preview", e);
-                }
-            });
-        }
+    public void print() {
```
lenhard: This method does not work on my machine. In the swing version a dialog would appear. Here, the printer icon pops up in the taskbar, but no documents are added.

src/main/java/org/jabref/gui/bibtexkeypattern/ResolveDuplicateLabelDialog.java
```diff
@@ -53,9 +58,12 @@ public ResolveDuplicateLabelDialog(BasePanel panel, String key, List<BibEntry> e
             JCheckBox cb = new JCheckBox(Localization.lang("Generate BibTeX key"), !first);
             b.appendRows("1dlu, p");
             b.add(cb).xy(1, row);
-            PreviewPanel pp = new PreviewPanel(null, entry, null);
-            pp.setPreferredSize(new Dimension(800, 90));
-            b.add(new JScrollPane(pp)).xy(3, row);
+            PreviewPanel previewPanel = new PreviewPanel(null, null);
+            previewPanel.setEntry(entry);
+            JFXPanel container = OS.LINUX ? new CustomJFXPanel() : new JFXPanel();
```
lenhard: Again the construction of the panel should be outsourced somewhere.

src/main/java/org/jabref/gui/collab/EntryAddChangeViewModel.java
```diff
@@ -1,29 +1,36 @@
...
 
     public EntryAddChangeViewModel(BibEntry diskEntry) {
         super(Localization.lang("Added entry"));
         this.diskEntry = diskEntry;
 
-        PreviewPanel pp = new PreviewPanel(null, diskEntry, null);
-        sp = new JScrollPane(pp);
+        PreviewPanel previewPanel = new PreviewPanel(null, null);
+        previewPanel.setEntry(diskEntry);
+        container = OS.LINUX ? new CustomJFXPanel() : new JFXPanel();
```
lenhard: Outsource as above.

src/main/java/org/jabref/gui/collab/EntryDeleteChangeViewModel.java
```diff
@@ -39,8 +44,10 @@ public EntryDeleteChangeViewModel(BibEntry memEntry, BibEntry tmpEntry) {
         LOGGER.debug("Modified entry: " + memEntry.getCiteKeyOptional().orElse("<no BibTeX key set>")
                 + "\n Modified locally: " + isModifiedLocally);
 
-        PreviewPanel pp = new PreviewPanel(null, memEntry, null);
-        sp = new JScrollPane(pp);
+        PreviewPanel previewPanel = new PreviewPanel(null, null);
+        previewPanel.setEntry(memEntry);
+        container = OS.LINUX ? new CustomJFXPanel() : new JFXPanel();
```
lenhard: Again outsource as above.

src/main/java/org/jabref/gui/importer/ImportInspectionDialog.java
```diff
@@ -230,7 +234,9 @@ public ImportInspectionDialog(JabRefFrame frame, BasePanel panel, String undoNam
         centerPan.setLayout(new BorderLayout());
 
         contentPane.setTopComponent(new JScrollPane(glTable));
-        contentPane.setBottomComponent(preview);
+        JFXPanel container = OS.LINUX ? new CustomJFXPanel() : new JFXPanel();
```
lenhard: Again outsource as above.

src/main/java/org/jabref/gui/mergeentries/MergeEntries.java
```diff
@@ -179,8 +185,11 @@ private void initialize() {
         // Setup a PreviewPanel and a Bibtex source box for the merged entry
         mainPanel.add(boldFontLabel(Localization.lang("Merged entry")), CELL_CONSTRAINTS.xyw(1, 6, 6));
 
-        entryPreview = new PreviewPanel(null, mergedEntry, null);
-        mainPanel.add(entryPreview, CELL_CONSTRAINTS.xyw(1, 8, 6));
+        entryPreview = new PreviewPanel(null, null);
+        entryPreview.setEntry(mergedEntry);
+        JFXPanel container = OS.LINUX ? new CustomJFXPanel() : new JFXPanel();
```
lenhard: Again outsource as above.

src/main/java/org/jabref/gui/openoffice/StyleSelectDialog.java
```diff
@@ -154,7 +158,9 @@ private void init() {
         builder.add(new JScrollPane(table)).xyw(1, 3, 5);
         builder.add(addButton).xy(3, 5);
         builder.add(removeButton).xy(5, 5);
-        builder.add(preview).xyw(1, 7, 5);
+        JFXPanel container = OS.LINUX ? new CustomJFXPanel() : new JFXPanel();
```
lenhard: Again outsource as above.

src/main/java/org/jabref/gui/preftabs/PreviewPrefsTab.java
```diff
@@ -116,10 +122,13 @@ private void setupLogic() {
...
         btnTest.addActionListener(event -> {
             try {
-                PreviewPanel testPane = new PreviewPanel(null, TestEntry.getTestEntry(), null)
-                        .setFixedLayout(layout.getText());
-                testPane.setPreferredSize(new Dimension(800, 350));
-                JOptionPane.showMessageDialog(PreviewPrefsTab.this, new JScrollPane(testPane), Localization.lang("Preview"), JOptionPane.PLAIN_MESSAGE);
+                PreviewPanel testPane = new PreviewPanel(null, null);
+                testPane.setFixedLayout(layout.getText());
+                testPane.setEntry(TestEntry.getTestEntry());
+                JFXPanel container = OS.LINUX ? new CustomJFXPanel() : new JFXPanel();
```
lenhard: Again outsource as above.

src/main/java/org/jabref/gui/search/SearchResultFrame.java
```diff
@@ -149,7 +154,10 @@ private void init(String title) {
         entryTable.addMouseListener(new TableClickListener());
 
         contentPane.setTopComponent(sp);
-        contentPane.setBottomComponent(preview);
+
+        JFXPanel container = OS.LINUX ? new CustomJFXPanel() : new JFXPanel();
```
lenhard: Again outsource as above.

halirutan:
@tobiasdiez @lenhard I hacked a version of the `CitationStyleGenerator` that manages a static instance of `CSL`. You find it in this [gist](https://gist.githubusercontent.com/halirutan/b9d74d414524edf9e534aea2307f4628/raw/393a1c3c05d58938401de818c6b45c1bcbaae488/CitationStyleGenerator.java). It vastly improves performance and memory footprint. The only sad thing is, that `CSL` itself needs so much memory, even for one single instance. I find over 500MB not acceptable even if many machines nowadays are equipped with at least 8GB, laptops might not.

Remarks: `CSL` is in general not threadsafe, but I'm not sure this is a problem with our current implementation. I tried to use `ThreadLocal` as well, but we have an armada of worker threads and `doInBackground` selects one. This means that several thread-local `CSL` instances are created which is not helping much.

The current behavior is that the first Preview takes as long as it took and until the style or the output format is changed, all follow-up calls are cheap, fast and don't use much memory. In numbers this means that while before one call to create the preview text needed 70.000 ms under profiling, it now takes on average about 800ms. The memory is better as well. Before we had about 85MB per preview and now 100 previews need 125MB (including the first expensive call).

lenhard:
@halirutan From my point of view, we could just scrap the CSLs again and remove them from JabRef completely. But I think @koppor has a strong opinion on it. 

Maybe the right way to go is to have them improve their implementation and I've seen that you've already opened an issue for that https://github.com/michel-kraemer/citeproc-java/issues/40 Let's hope they react.

koppor:
Yeah, CSL was wished by many users and is a strong selling point 😇

To improve speed, we added the CSL cache. Think, with JavaScript V8, it
also can get much better. Maybe, we should put energy in that...

Reimplementing csl-proc in Java is also possible, but much effort...

Am 13.12.2017 09:39 schrieb "Jörg Lenhard" <notifications@github.com>:

halirutan:
I checked out this branch and I would have included my fix in here, but we still have the big problem, that UTF8 fonts are not displayed correctly. This might be related to [this question I asked recently](https://stackoverflow.com/questions/46954858/display-unicode-characters-in-a-jlabel-on-os-x) although now it's for JavaFX and I'm on Linux.

Anyway, in the old preview, if an entry contains e.g. `textit{..}`, it looks like this in the preview

img

In the JavaFX preview from this branch, the same entry displays like this

img

If you like to try this out, here is the shortened entry

```
@PhdThesis{deBraga2001,
  author    = {Michael deBraga},
  title     = {{The postcranial anatomy of \textit{Procolophon} (Parareptilia: Procolophonidae) and its implications for the origin of turtles}},
  school    = {Department of Zoology, University of Toronto},
  year      = {2001},
  type      = {PhD thesis},
  address   = {Canada},
  note      = {303 pp.}
}
```

My PR for the CSL performance should be almost unrelated to this here and can later easily be merged if we find it works acceptably.

src/main/java/org/jabref/gui/BasePanel.java
tobiasdiez: Good remark! Fixed.

src/main/java/org/jabref/gui/PreviewPanel.java
tobiasdiez: I removed the & and added the display of the shortcut.

src/main/java/org/jabref/gui/PreviewPanel.java
tobiasdiez: ohh gosh. By default, I print to PDF and thus the PDF print dialog poped up and everything seemed to work just fine. But, of course, in general you want to have a confirmation dialog and a possibility to change the printer. Luckily, javafx provides a default dialog for printing, so this was easy to implement.

tobiasdiez:
Thanks for the feedback. I revised the code accordingly. 

@halirutan I couldn't replicate the problem with `textit` but I only tested it for the default preview layout (not cls). I would like to revisit this problem as soon as your PR with the simplified cls inclusion is merged so that testing gets easier.

koppor:
latex2unicode uses the math alphabet instead of the normal alphabet for `textit`. Note hat in mathematics, `mathit` is the command for italics text, not `textit`.

See https://github.com/tomtung/latex2unicode/blob/a3610c3b8161b5d2e604e1bebc9a1a1d780e4474/src/main/scala/com/github/tomtung/latex2unicode/helper/Unary.scala#L647

and compare 𝐻 with 𝘏:

grafik

Maybe we should patch latex2unicode?

Siedlerchr:
From what I read, the mathematic area is the only font which provides italics. And I am unsure whether we really should support that.

halirutan:
Question: If there was a mystical `latex2html`, could this be a replacement for all usages of the latex2unicode converter?

For the Mathematica plugin, I had to transform the doc of built-in function to html. Fortunately, I could even use MathML and now the function docs contain everything from italics, simple formulas, and integrals.

Siedlerchr:
@halirutan  There is a latex2html converter. But not all cases should be handled by this, for example you typically want names with accents or other weird characters to be displayed as Unicode.

Siedlerchr:
I propose the following which may make our life a lot easier:
- Main Table is a swing component which uses HTML
- The entry preview is soon a WebView Component, so it's also HTML
- CSL gets also written to HTML, if I am right

So why don't we use the latex2html converter on both of them?
This would make our life a lot easier.
Maybe we should discuss this in a devcall?

halirutan:
@Siedlerchr That was the reason behind my question. I ran several times into this issue on both Linux and OS X and it was the reason for the [question on SO](https://stackoverflow.com/questions/46954858/display-unicode-characters-in-a-jlabel-on-os-x). There, it was also suggested to use HTML and I'm also confident, that it would make life easier.

In addition to "it is rendered at all" we hopefully get a better quality. On Linux the italics look so ugly that it is almost not bearable. Therefore, if easily possible, I vote for using HTML for rendering LaTeX.

lenhard:
One reason for using latex2unicode was that our own custom conversion was relatively incomplete and partly broken. I am not sure on the state of the html conversion, but I wouldn't presume that it's perfect. So if we go for displaying html everywhere (which is perfectly fine in my point of view), it would also be cool if we had a robust Java library that does the LaTeX to html conversion. I am not sure if such a library exists, though, and then we will have to make do with the current conversion.

The main reason for latex2unicode was to enable a proper search of entries independent of LaTeX commands. As far as I understand, this shouldn't conflict with displaying html and can be kept as is.

tobiasdiez:
So how do we want to proceed with this PR? The discussion concerning latex2unicode/html is a bit decoupled from the actual purpose of this PR (which is just to move forward with the migration to JavaFX). Maybe it also makes sense to extract our latex2html code and merge it into the latex2unicode library we use.

koppor:
@Siedlerchr The displayed characters in my screenshot are BOTH (!) Unicode. The first UNICODE math the second UNICODE italics. It was a screenshot of Notepad++!

koppor:
I remember the discussions with @oscargus who argued that our latex2html converter is the most complete one. Maybe, we should offer it as library? Refs #110

halirutan:
@tobiasdiez I really would like to concentrate on the JavaFX part of this PR as well, but until the unicode issue is resolved, this breaks preview for entries that contain any LaTeX on my machine. As long as I'm not the only one where this happens, I see no way we could merge this PR without annoying a lot of users. Therefore, we need to discuss the latex2XXX issue somewhere first. Do you have a suggestion how we should proceed?

koppor:
+1 for switchting to latex2html in this case (entry preview). Thus, moved to v4.2 and not for v4.1 release.

tobiasdiez:
I now used latex2html for the preview and it seems to work fine. I also implemented the citation style provider as discussed with @halirutan in his recent PR.

Sadly, I made some mistakes with while checking out my code and now cannot push to this branch anymore (git claims "everything is up-to-date" - well it isn't). So continued at https://github.com/JabRef/jabref/pull/3574.

Siedlerchr:
@tobiasdiez  in that case you have to do a force push, e.g. when amending your last commit.

halirutan:
@tobiasdiez Only regarding the

> Sadly, I made some mistakes with while checking out my code

When you merge updates into this branch and you use rebase, then you cannot push it again. If you are sure no one else has used your branch, you have to make a force push and everything is fine. Rebasing your (local) branch looks nicer because you don't get a cluttered history wich merge commits, but it should only be used you are working on a local branch.

tobiasdiez:
Thanks for the help but the problem was not about the rebase/force pushed. I created the branch on another pc and used the github gui tool to check it out on a new one. For some reason git now thinks the remote branch associated to my local code is `JabRef/javafxPreview` instead of just `javafxPreview`. I could probably remove/readd the remote branch but it was easier to just create a new PR.

koppor:
You can edit .git/config directly for changing branch config. You can also
use gitk --all to check, what's going on.

Finally, git push -f origin HEAD:BranchNameAtOrigin always works;
regardless of your local branch name. No need to create a new PR then...

Am 23.12.2017 22:22 schrieb "Tobias Diez" <notifications@github.com>:

PR #3506, opened 2017-12-09
Fix #3359: Automatically remove colon and apostrophe from key pattern

### LINKED ISSUE #3359: key generator fails to parse colon
It seems that in 4.0 (Mac version), when I customize the key generator to [author][year][veryshorttitle], the generated keys will include the trailing colon if there is one in the title. It also includes other special characters like (`). 

I remember this wasn't the case in 3.6.

tobiasdiez:
Fixes #3359. Colons and apostrophes are now removed from the generated key pattern.
I also refactored the key generator, mainly converting the static methods to instance methods.

----


tobiasdiez:
@lenhard I manually tested most use cases, where the code changes (except OpenOffice since I don't use it). What are our current plans for the 4.1 release? When do we want to release? If there is still some time before the release I would prefer if this PR is merged before the release.

Siedlerchr:
You may want to add the percent sign and the ampersand (&), which is problematic as it creates a problem with biber/biblatex

src/main/java/org/jabref/logic/bibtexkeypattern/BibtexKeyGenerator.java
```diff
@@ -18,33 +22,95 @@
...
+                    newKey.append(c);
+                }
+            }
+            return newKey.toString();
+        }
+
+        StringBuilder newKey = new StringBuilder();
+        for (int i = 0; i < key.length(); i++) {
+            char c = key.charAt(i);
+            if (!Character.isWhitespace(c) && ("{}(),\\\"#~^':`".indexOf(c) == -1)) {
```
Siedlerchr: You should reeally extract this to a constant, same with th e one abvoe

tobiasdiez:
Should this still go into 4.1?

lenhard:
@tobiasdiez When you fix the checkstyle issues and conflicts then this PR can be merged.

PR #3507, opened 2017-12-09
Add option to remove the link when file is no longer found

### LINKED ISSUE #3493: It should be possible to delete an attached file

grafik

There also should be a `-` icon at the right side, shouldn't it?

Siedlerchr:
Fixes  #3493

When pressing the DEL key and the file can not be found, this dialog pops up and you can remove the  file from the entry

 
```
Opening: X:\Users\CS\Documents\_JABREFTEMP\bibliography.bib
Opening: X:\Users\CS\Documents\_JABREFTEMP\newBiblatexTEst.bib
Could not find file 'O’Keefe2009 - Feature selection and weighting methods in sentiment analysis.pdf'.
Could not find file 'Saggion2007 - Ontology-based Information Extraction for Business Intelligence.pdf'.

```
----


src/main/java/org/jabref/gui/fieldeditors/LinkedFileViewModel.java
```diff
@@ -263,10 +264,9 @@ public void moveToDefaultDirectory() {
 
     public boolean delete() {
         Optional<Path> file = linkedFile.findIn(databaseContext, Globals.prefs.getFileDirectoryPreferences());
-        if (file.isPresent()) {
-
-            ButtonType removeFromEntry = new ButtonType(Localization.lang("Remove from entry"));
+        ButtonType removeFromEntry = new ButtonType(Localization.lang("Remove from entry"));
```
tobiasdiez: Can you please specify the button type [YES](https://docs.oracle.com/javase/8/javafx/api/javafx/scene/control/ButtonBar.ButtonData.html#YES) so that the order of the buttons is correct.

tobiasdiez: Moreover, I would suggest a more descriptive label like "(Only) unlink file" or "Remove file link".

src/main/java/org/jabref/gui/fieldeditors/LinkedFileViewModel.java
```diff
@@ -289,15 +289,22 @@ public boolean delete() {
...
-            } else {
-                dialogService.showErrorDialogAndWait(
-                        Localization.lang("File not found"),
-                        Localization.lang("Could not find file '%0'.", linkedFile.getLink()));
-                return true;
+            }
+
+        } else {
+            Optional<ButtonType> buttonType = dialogService.showCustomButtonDialogAndWait(AlertType.ERROR, Localization.lang("File not found"),
+                    (Localization.lang("Could not find file '%0'.", linkedFile.getLink()) + OS.NEWLINE + Localization.lang("Pressing 'remove' will remove the file from the entry")),
```
tobiasdiez: The last sentence "Pressing..." is rather self-explanatory. I would remove it.

koppor:
+1 or not showing the dialog at all. We have the undo possibility and I would not think that users accidently press <kbd>Del</kbd>. Would be nice if I could quickly delete attachments if JabRef goes crazy ^^

PR #3508, opened 2017-12-10
Implement #3491: Clicking the PDF icon in the linked files editor opens the file

### LINKED ISSUE #3491: Double click on an entry on file list should open the PDF
As is:

grafik

opens:

grafik

Should be: 

- [ ] Double click directly opens the PDF
- [ ] Furthermore, there is no button to open the PDF. The Lupe searches for a PDF, but does not open one

tobiasdiez:
Resolves #3491. A click on the PDF icon on a linked file in the entry editor opens the file. This is consistent with the behavior in the main table and thus I didn't added a new button and / or changed the double-click behavior as it was suggested in the issue.

----


Siedlerchr:
That's a nice solution. Btw, do you have any idea why the second file in the listvie does not indicate the blue selection ? Although it is selected

tobiasdiez:
@lenhard The beauty of JavaFX! 
@Siedlerchr No idea. Similarly, there are these annoying padding problems of the first row. The "beauty" of JavaFX! 😄

PR #3368, opened 2017-10-28
Rework AutoSetFileLinks

Siedlerchr:
Fixes #3346 

- [x] Change in CHANGELOG.md described
- [x] Tests created for changes
- [ ] Screenshots added (for bigger UI changes)
- [x] Manually tested changed features in running JabRef
- [ ] Check documentation status (Issue created for outdated help page at [help.jabref.org](https://github.com/JabRef/help.jabref.org/issues)?)
- [ ] If you changed the localization: Did you run `gradle localizationUpdate`?

PR #3509, opened 2017-12-10
Fixes #3472: Files starting with BibTeX key of a similar entry are no longer found by mistake

### LINKED ISSUE #3472: Default settings for filename format and autolink files don't play nice together.
I ran into the need to autolink files. I had generated the bibtex keys using (the reasonable) default setting `[auth][year]`. And renamed the files using a filename format `\bibtexkey - ...`

As example consider I have following papers:

- Einstein 1902 "Kinetische Theorie des ...
- Einstein 1902 "Thermodynamische Theorie der...
- Einstein (year unknown) "Ein den Elementarprozess der...
- Unknown author - unknown year "hello...

The autogenerated keys would be "Einstein1902", "Einstein1902a", "Einstein", "".  The problem is that "Einstein1902" will try to autolink both 1902 papers.  The "Einstein" will try to autolink all three.  An the one with the empty key will try to autolink all of the papers.

I think it would be good to have save defaults.   

For myself I adjusted the autolink regex so that it checks for a whitespace or dot after the bibtex key - I think that might be a saver default.
However the case of empty keys still should be treated.

tobiasdiez:
Fixes #3472.
The auto-link algorithm just checked if the file starts with the BibTeX key. This leads to a lot of false positives since "Einstein1902a.pdf" is found for the entry "Einstein1902" although this pdf probably belongs to a second paper by Einstein that was published in the same year. 
This is fixed in this PR.
With the proposed implementation also the pdf file "Einstein1902aboutSomething.pdf" is no longer found (since there is really no possibility in telling the difference to "Einstein1902a - a second paper.pdf"). Thus the user is advised to add some non-letter character after the key.

----


src/main/java/org/jabref/gui/BasePanel.java
```diff
@@ -2016,9 +2016,9 @@ public void searchAndOpen() {
 
             // Run the search operation:
             FileFinder fileFinder = FileFinders.constructFromConfiguration(Globals.prefs.getAutoLinkPreferences());
-            List<Path> files = fileFinder.findAssociatedFiles(entry, dirs, extensions);
+            Collection<Path> files = fileFinder.findAssociatedFiles(entry, dirs, extensions);
             if (!files.isEmpty()) {
-                Path file = files.get(0);
+                Path file = files.iterator().next();
```
Siedlerchr: I see no advantage in using Collection here

src/main/java/org/jabref/gui/externalfiles/AutoSetFileLinksUtil.java
```diff
@@ -26,22 +27,33 @@
...
+    private List<Path> directories;
+    private AutoLinkPreferences autoLinkPreferences;
+    private ExternalFileTypes externalFileTypes;
 
-    public List<LinkedFile> findassociatedNotLinkedFiles(BibEntry entry, BibDatabaseContext databaseContext, FileDirectoryPreferences fileDirPrefs, AutoLinkPreferences autoLinkPrefs, ExternalFileTypes externalFileTypes) {
+    public AutoSetFileLinksUtil(BibDatabaseContext databaseContext, FileDirectoryPreferences fileDirectoryPreferences, AutoLinkPreferences autoLinkPreferences, ExternalFileTypes externalFileTypes) {
+        this(databaseContext.getFileDirectoriesAsPaths(fileDirectoryPreferences), autoLinkPreferences, externalFileTypes);
+    }
+
+    public AutoSetFileLinksUtil(List<Path> directories, AutoLinkPreferences autoLinkPreferences, ExternalFileTypes externalFileTypes) {
```
Siedlerchr: I am against creating a second constructor. It was my initial idea in #3368  and guess who supposed the current alternative? 😆 
See your comment https://github.com/JabRef/jabref/pull/3368#discussion_r147604206

src/main/java/org/jabref/logic/util/io/CiteKeyBasedFileFinder.java
```diff
@@ -32,40 +33,34 @@
     }
 
     @Override
-    public Map<BibEntry, List<Path>> findAssociatedFiles(List<BibEntry> entries, List<Path> directories, List<String> extensions) {
+    public Multimap<BibEntry, Path> findAssociatedFiles(List<BibEntry> entries, List<Path> directories, List<String> extensions) {
         Objects.requireNonNull(directories);
         Objects.requireNonNull(entries);
 
-        Map<BibEntry, List<Path>> result = new HashMap<>();
+        Multimap<BibEntry, Path> result = ArrayListMultimap.create();
```
Siedlerchr: If you structure the code similar to the one in AutoSetLinks, you won't need a multimap:
```
 for (BibEntry entry : entries) {

                List<LinkedFile> linkedFiles = util.findassociatedNotLinkedFiles(entry, databaseContext, Globals.prefs.getFileDirectoryPreferences(), Globals.prefs.getAutoLinkPreferences(), ExternalFileTypes.getInstance());

                if (ce != null) {
                    for (LinkedFile linkedFile : linkedFiles) {
```

src/main/java/org/jabref/gui/externalfiles/AutoSetFileLinksUtil.java
tobiasdiez: I don't understand what you mean. The linked comment is about the list of entries and not about how the `AutoSetFileLinksUtil` gets the things needed (like directories and preferences). I think, the current setup makes sense: on initialization, specify everything the auto set functionality needs to know. Why don't you like these constructors?

src/main/java/org/jabref/logic/util/io/CiteKeyBasedFileFinder.java
tobiasdiez: Thanks! This is really good feedback. I just realized that the finder never gets a list of entries as argument but always a single entry. Thus the design can be simplified drastically without mulitmaps or maps to collections.

src/main/java/org/jabref/logic/util/io/FileFinder.java
```diff
@@ -17,10 +19,9 @@
      * @param directories The root directories to search.
      * @param extensions  The extensions that are acceptable.
      */
-    Map<BibEntry, List<Path>> findAssociatedFiles(List<BibEntry> entries, List<Path> directories, List<String> extensions);
+    Multimap<BibEntry, Path> findAssociatedFiles(List<BibEntry> entries, List<Path> directories, List<String> extensions);
 
-    default List<Path> findAssociatedFiles(BibEntry entry, List<Path> directories, List<String> extensions) {
-        Map<BibEntry, List<Path>> associatedFiles = findAssociatedFiles(Collections.singletonList(entry), directories, extensions);
-        return associatedFiles.getOrDefault(entry, Collections.emptyList());
+    default Collection<Path> findAssociatedFiles(BibEntry entry, List<Path> directories, List<String> extensions) {
```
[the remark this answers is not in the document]
tobiasdiez: The multimap only returns a collection because it actually is not implemented as a list (the sorting is random). But now we don't need multimaps anymore so we can go back to lists!

src/main/java/org/jabref/gui/externalfiles/AutoSetFileLinksUtil.java
Siedlerchr: Ah sorry, I could not see all code in the line on my mobile, so I was a b9it confused. Forget about my remark,

src/test/java/org/jabref/logic/util/io/RegExpBasedFileFinderTests.java
```diff
@@ -56,7 +57,7 @@ public void testFindFiles() {
         RegExpBasedFileFinder fileFinder = new RegExpBasedFileFinder("**/[bibtexkey].*\\\\.[extension]", ',');
 
         //when
-        List<Path> result = fileFinder.findAssociatedFiles(localEntry, dirs, extensions);
+        Collection<Path> result = fileFinder.findAssociatedFiles(localEntry, dirs, extensions);
```
Siedlerchr: Here you can change them back to list again

src/main/java/org/jabref/gui/externalfiles/AutoSetFileLinksUtil.java
tobiasdiez: No problem!

src/test/java/org/jabref/logic/util/io/RegExpBasedFileFinderTests.java
tobiasdiez: Done!

koppor:
How does this relate with the preferences?

grafik

Documentation at http://help.jabref.org/en/FileLinks#auto-linking-files

Are these settings still respected?

tobiasdiez:
@koppor Yes, these settings are still valid. The fix applies only if you select the first option "Auto link files that start with the BibTeX key".

src/main/java/org/jabref/logic/util/io/CiteKeyBasedFileFinder.java
```diff
@@ -32,54 +33,57 @@
...
+        return result.stream().sorted().collect(Collectors.toList());
+    }
+
+    private boolean matches(String filename, String citeKey) {
+        boolean startsWithKey = filename.startsWith(citeKey);
+        if (startsWithKey) {
+            // The file name starts with the key, that's already a good start
+            // However, we do not want to match "JabRefa" for "JabRef" since this is probably a file belonging to another entry published in the same time / same name
+            char charAfterKey = filename.charAt(citeKey.length());
+            return !StringUtil.contains(BibtexKeyPatternUtil.CHARS, charAfterKey);
```
Siedlerchr: Better direclty use the java method: String.contains(Character.toString(c))

Siedlerchr: And also take a look at the codacy

tobiasdiez:
@koppor Why did you removed the "ready-for-review" flag and marked this PR as WIP? I was and I am still waiting for a second review, otherwise this PR is ready from my point of view. The minor issues pointed-out by @Siedlerchr are now also fixed.

koppor:
Think, because I watched this PR, liked it and just wanted others to focus on other PRs.

PR #3523, opened 2017-12-12
Refactor shared package into the architecture

lenhard:
The new Java 9 module system brings us the possibility to actually modularize inside Java, i.e. to distribute our source code into multiple modules and specify dependencies and constraints among them, much in the way that we do it now with tests. But to extract modules, there can be no cyclic dependencies among them.

The `model` package is our best candidate for extracting a new module. Here the last dependency that coupled it to transitively to everything else was the `shared` package. This PR refactors the shared package and puts most of it into `logic`. I broke the dependency from `model` by introducing a number of interface that the `logic` layer implements. These interfaces will have just one implementation now, but at least they break the dependency cycle.

This is just a refactoring, so I think it should be merged after the release of 4.1

----


src/main/java/org/jabref/gui/exporter/SaveDatabaseAction.java
```diff
@@ -318,10 +320,13 @@ public void saveAs(File file) throws Exception {
 
         if (context.getLocation() == DatabaseLocation.SHARED) {
             // Save all properties dependent on the ID. This makes it possible to restore them.
-            DBMSConnectionProperties properties = context.getDBMSSynchronizer().getDBProcessor()
-                    .getDBMSConnectionProperties();
-            new SharedDatabasePreferences(context.getDatabase().generateSharedDatabaseID())
-                    .putAllDBMSConnectionProperties(properties);
+            DatabaseSynchronizer synchronizer = context.getDBMSSynchronizer();
+            if (synchronizer instanceof DBMSSynchronizer) {
+                DatabaseConnectionProperties properties = ((DBMSSynchronizer) synchronizer).getDBProcessor()
```
tobiasdiez: If you add `getDBMSConnectionProperties` to the `Synchronizer` interface, then you probably don't need the if statement and the explicit cast here.

lenhard: Thanks, nicely spotted! So far, I was focused on the refactoring only, but the code certainly has a lot of potential for improvements. Implementing your suggestion was fairly easy and I fixed this at a second position as well.

koppor:
This refs https://github.com/JabRef/jabref/issues/110. I am still planning to work on that as I want to provide our logic package as standalone library.

lenhard:
@koppor Ah, I was looking for that issue :) Maybe we should re-open it, since we now have a tangible use case (the new Java module system). Once the model package is modularized, the logic package will not be far off.

lenhard:
@JabRef/developers With another review, this could be merged.

src/main/java/org/jabref/logic/shared/DBMSConnectionProperties.java
```diff
@@ -1,20 +1,22 @@
...
+import org.jabref.model.database.shared.DatabaseConnectionProperties;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
 /**
  * Keeps all essential data for establishing a new connection to a DBMS using {@link DBMSConnection}.
  */
-public class DBMSConnectionProperties {
+public class DBMSConnectionProperties implements DatabaseConnectionProperties {
```
Siedlerchr: I only don't get this, isn't this supposed to be same or are DataBaseConnectionProperties more general?

src/main/java/org/jabref/model/database/shared/DatabaseSynchronizer.java
```diff
@@ -0,0 +1,24 @@
...
+
+    String getDBName();
+
+    void pullChanges();
+
+    void closeSharedDatabase();
+
+    void registerListener(Object listener);
+
+    void openSharedDatabase(DatabaseConnection connection) throws DatabaseNotSupportedException, SQLException;
```
Siedlerchr: I would let this rather be more general, by not having the concrete SQLException here. I could implement this interface for my Sharelatex Connection.

src/main/java/org/jabref/logic/shared/DBMSConnectionProperties.java
lenhard: `DBMSConnectionProperties` are the only implementation of `DatabaseConnectionProperties`, they're essentially the same. The sole reason for the existence of the interface is that it breaks the dependency from model to elsewhere.

src/main/java/org/jabref/model/database/shared/DatabaseSynchronizer.java
lenhard: Ok, I've inlined the `SQLException` in favor of an `IllegalStateException`.

Issue #3526, opened 2017-12-13
External Group Changes result in NPE


Siedlerchr:
current master:
Open a   bib file
Modify the bib file in an external editor., add for example this group tree information:
```

@Comment{jabref-meta: groupstree:
0 AllEntriesGroup:;
1 ExplicitGroup:12\;0\;;
2 ExplicitGroup:1\;0\;;
2 ExplicitGroup:2\;0\;;
2 ExplicitGroup:3\;0\;;
2 ExplicitGroup:4\;0\;;
2 ExplicitGroup:5\;0\;;
2 ExplicitGroup:6\;0\;;
2 ExplicitGroup:7\;0\;;
2 ExplicitGroup:8\;0\;;
2 ExplicitGroup:9\;0\;;
2 ExplicitGroup:10\;0\;;
2 ExplicitGroup:11\;0\;;
2 ExplicitGroup:12\;0\;;
2 ExplicitGroup:13\;0\;;
2 ExplicitGroup:14\;0\;;
2 ExplicitGroup:15\;0\;;
0 ExplicitGroup:16\;0\;;
2 ExplicitGroup:17\;0\;;
}

```
Review changes lead to NPE

```
21:17:40.388 [AWT-EventQueue-0] ERROR org.jabref.FallbackExceptionHandler - Uncaught exception occurred in Thread[AWT-EventQueue-0,6,main]
java.lang.NullPointerException: null
	at org.jabref.gui.collab.GroupChangeViewModel.makeChange(GroupChangeViewModel.java:50) ~[main/:?]
	at org.jabref.gui.collab.ChangeDisplayDialog.lambda$2(ChangeDisplayDialog.java:94) ~[main/:?]
	at javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:2022) ~[?:1.8.0_152]
	at javax.swing.AbstractButton$Handler.actionPerformed(AbstractButton.java:2348) ~[?:1.8.0_152]
	at javax.swing.DefaultButtonModel.fireActionPerformed(DefaultButtonModel.java:402) ~[?:1.8.0_152]
	at javax.swing.DefaultButtonModel.setPressed(DefaultButtonModel.java:259) ~[?:1.8.0_152]
	at javax.swing.plaf.basic.BasicButtonListener.mouseReleased(BasicButtonListener.java:252) ~[?:1.8.0_152]
	at java.awt.Component.processMouseEvent(Component.java:6533) ~[?:1.8.0_152]
[...]

Issue #3189, opened 2017-09-03
Renaming a group via groups interface does not change the group name in the interface


Andrew-Lev:
JabRef 4.0-dev
Linux 4.4.0-92-generic amd64 
Java 1.8.0_131

Steps to reproduce:

0. Open group interface. (Groups -> toggle groups interface)
1. Create a group
2. Add an item to the group
3. Right click on the group in the group interface -> edit group
4. Change the group name

Group name in the interface stays the same.  If you check the group name by selecting an item and going to its Bibtex source you can see the change there.
![screenshot from 2017-09-02 18-09-52](https://user-images.githubusercontent.com/17808285/29999791-1e14ef46-900a-11e7-9ae3-2b2150bc8149.png)

Siedlerchr:
When you close and expand the group, the change should also be visible. No idea though why it is not visible immediately. The same issue occurs if you assign an icon or color to the group

AEgit:
Might be related to this:
https://github.com/JabRef/jabref/issues/3112

Ali96kz:
I will work on it

Ali96kz:
When I press edit group that show `org.jabref.model.groups.AbstractGroup#name` instead of `org.jabref.gui.groups.GroupNodeViewModel#displayName`, why? What's the difference between them?

tobiasdiez:
The `displayname` is a formatted version of `name`. For example, if `name` contains latex commands like `Hell\'{o}` then `displayname` is the unicode version `Helló`.

tobiasdiez:
This should be fixed in the latest development version. Could you please check the build from http://builds.jabref.org/master/. Thanks! Please remember to make a backup of your library before trying-out this version.

AEgit:
Can confirm this has been fixed in:

JabRef 5.0-dev--snapshot--2019-01-05--master--d281ece4b
Windows 10 10.0 amd64 
Java 1.8.0_191

PR #3529, opened 2017-12-14
Add null checks in GroupChangeViewModel

lenhard:
This avoids the NPEs in #3526. However, I can't get the groups side pane to repaint and depict the modified groups tree. I've tried different ways via calling the `GroupsSidePane` or setting selected groups in the `StateManager`, but no success.

@tobiasdiez How do I force the `GroupSidePane` to display a changed group tree?

----

### DIFF
--- a/src/main/java/org/jabref/gui/collab/GroupChangeViewModel.java
+++ b/src/main/java/org/jabref/gui/collab/GroupChangeViewModel.java
@@ -28,7 +28,11 @@ class GroupChangeViewModel extends ChangeViewModel {
 
     @Override
     public boolean makeChange(BasePanel panel, BibDatabase secondary, NamedCompound undoEdit) {
-        final GroupTreeNode root = panel.getBibDatabaseContext().getMetaData().getGroups().orElse(null);
+        GroupTreeNode root = panel.getBibDatabaseContext().getMetaData().getGroups().orElse(null);
+        if(root == null) {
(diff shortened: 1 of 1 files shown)


tobiasdiez:
Sadly updates to the group tree in general are ignored, see e.g. https://github.com/JabRef/jabref/issues/3189. Normally, one would like that all changes are automatically propagated using the event bus / javafx binding, but for some reason this does not work (sorry, I forgot the details).

lenhard:
Ok, thanks for clarifying. I am not sure what to do with this PR then. Under the hood it works (avoids the exceptions), but all of this is not visible to the user. Right now it requires a restart of the application.

tobiasdiez:
Is the view updated when you switch between databases? This always worked for me so far. I would say merge this PR here (since it fixes the NPE) and comment in the above issue that there are more update problems. I'll have a look at this problem, but not right now as this week is kind of a hell for me.

lenhard:
Yes, it is updated when switching between databases, e.g. all groups are removed and the all entries group is displayed when I remove the group statement. A rename of the group in the file doesn't carry over, so there are probably deeper problems.

I still have the problem that it tends to display the groups tree of the wrong database, but at least the change is reflected somehow.

Siedlerchr:
Can this be merge now?

lenhard:
Sure, go ahead.

PR #3530, opened 2017-12-14
Describe new workflow with Crowdin

koppor:
[Crowdin](http://translate.jabref.org/) automatically adds missing translations to the other languages.

With our `localizationUpdate` task, we also do that. This leads to conflicts at the Crowdin synchronizations.

This PR proposes to 

1. update the description of how to add new strings
2. update the tests to enable the translations being a subset of the existing translations.

After each sync, the translation will reappear. So we have a short time window where the translations are missing and that translators using GitHub do not see these new translations. I would accept that.

News: We have 17 new Indonesian translations, because a guy just translated some strings there at Crowdin. 🎉

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -82,10 +82,9 @@ because <additional rationale>.
 
 ### When adding a new Localization.lang entry
 Add new `Localization.lang("KEY")` to Java file.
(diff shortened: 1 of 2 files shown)


koppor:
Test fails because of

    java.io.IOException: Server returned HTTP response code: 503 for URL: http://ieeexplore.ieee.org/document/7421926/

PR #3533, opened 2017-12-15
Fix preview performance

### LINKED ISSUE #2533: Citation preview incrementally increases memory usage with each preview

halirutan:
This unifies the access to the CSL engine. The adapter class will hold one instance of it and only recreate it when the style changes. This improves the performance of from the second call to the Preview drastically and should provide an acceptable speed to fix #2533.

In addition, I extended the search for citation styles when JabRef is run during development.

I will add changelog and friends, but for the moment my biggest problem is that the unit tests are broken on master (at least for me).

----


src/main/java/org/jabref/logic/citationstyle/CSLAdapter.java
```diff
@@ -0,0 +1,116 @@
...
+    private static String style;
+    private static CitationStyleOutputFormat format;
+    private static JabRefItemDataProvider dataProvider = new JabRefItemDataProvider();
+    private static CSL cslInstance;
+
+    /**
+     * Creates the bibliography of the provided items. This method needs to run synchronized because the underlying
+     * CSL engine is not thread-safe.
+     */
+    static synchronized List<String> makeBibliography(List<BibEntry> bibEntries, String style, CitationStyleOutputFormat outputFormat) throws IOException {
```
lenhard: As far as I can see by looking at the code, this solution should be thread-safe. You could turn up parallelism by a tiny notch, but I cannot say if this has any tangible benefit. This could be done by not marking the whole method as synchronized, but just the part where the CSL engine is used. To do so, you would add a private lock instance variable, e.g. `private static Object lock = new Object();` and then refactor the method like this:
```
        List<CSLItemData> cslItemData = new ArrayList<>(bibEntries.size());
        for (BibEntry bibEntry : bibEntries) {
            cslItemData.add(bibEntryToCSLItemData(bibEntry));
        }

        synchronized(lock) {
            initialize(style, outputFormat);
            dataProvider.setData(cslItemData);
            cslInstance.registerCitationItems(dataProvider.getIds());
            Bibliography bibliography = cslInstance.makeBibliography();
            return Arrays.asList(bibliography.getEntries());
        }
```
You can refactor the code in this way or not, I leave this up to you. It will only have an effect if there are many concurrent calls to the method.

src/main/java/org/jabref/logic/citationstyle/CitationStyle.java
```diff
@@ -115,33 +118,51 @@ public static CitationStyle createCitationStyleFromFile(final String styleFile)
...
+        if (!STYLES.isEmpty()) {
+            return STYLES;
+        }
         try {
-            final List<CitationStyle> citationStyles = new ArrayList<>();
             String path = CitationStyle.class.getProtectionDomain().getCodeSource().getLocation().toURI().getPath();
 
+            // This is a quick fix to have the styles when running JabRef in a development environment.
+            // The styles.jar is not extracted into the JabRef.jar and therefore, we search the classpath for it.
+            if ((new File(path)).isDirectory()) {
```
lenhard: Whenever we add/modify file IO code in JabRef, we try to use the new Java API for that `java.nio`, not `java.io`. So please replace the condition with `Files.isDirectory(Paths.get(path))`.

src/main/java/org/jabref/logic/citationstyle/CSLAdapter.java
```diff
@@ -0,0 +1,116 @@
...
+ * In the current scheme, {@link #makeBibliography} can be called as usual
+ * SwingWorker task and to the best of my knowledge, concurrent calls will pile up and processed sequentially.
+ */
+public class CSLAdapter {
+
+    private static final BibTeXConverter BIBTEX_CONVERTER = new BibTeXConverter();
+    private static String style;
+    private static CitationStyleOutputFormat format;
+    private static JabRefItemDataProvider dataProvider = new JabRefItemDataProvider();
+    private static CSL cslInstance;
```
tobiasdiez: Is it needed, that everything is `static` here? I actually would prefer a usual class with non-standard methods/instance variables.

src/main/java/org/jabref/logic/citationstyle/CSLAdapter.java
```diff
@@ -0,0 +1,116 @@
...
+    private static class JabRefItemDataProvider implements ItemDataProvider {
+
+        private ItemDataProvider dataProvider = new ListItemDataProvider();
+
+        public void setData(List<CSLItemData> data) {
+            dataProvider = new ListItemDataProvider(data);
+        }
+
+        @Override
+        public CSLItemData retrieveItem(String id) {
```
tobiasdiez: I'm not sure if this is really an advantage but since you are already implementing a custom data provider, I would suggest the following:
- setData accepts a list of `BibEntry` 
- retriveItem gets the `BibEntry` with the given key and then returns the result of `bibEntryToCSLItemData`. This solution might be a bit more memory effective since you don't have an additional list (but it only makes sense if the `bibEntryToCSLItemData` method is not expensive).

halirutan: Ha funny. It's the exact same thing I have thought about. I wanted to wait until we are sure it works in general. I too would prefer to make a real `JabRefItemDataProvider` that works on `BibEntry`.

src/main/java/org/jabref/logic/citationstyle/CSLAdapter.java
halirutan: Yes. This would probably be better. We could then still instantiate several CSL adapters if we need one for whatever purpose.

src/main/java/org/jabref/logic/citationstyle/CitationStyle.java
halirutan: OK. No problem.

src/main/java/org/jabref/logic/citationstyle/CSLAdapter.java
halirutan: It is an easy fix indeed. I will profile where the bottleneck is when you have many items selected and click on `Copy Preview`. It might be useful to check if the loop itself should be parallelized as well. Copying the preview is too slow for my taste and I'm unsure it really comes from CSL this time. I will check.

lenhard:
@halirutan Thanks for the investigation and I hope you did not waste too much time on that. That is not correct bibtex, I just opened an old bib file and clicked around, not caring too much about the actual data. You can ignore that error in your PR here.

I think a viable strategy would actually be to absorb the exception completely. Exceptions in JabRef's log should indicate problems in JabRef. This is not a problem of JabRef, but of user data and the citation styles library. Therefore, I think it would be sufficient to create a log entry stating that the attempt to parse an entry results in an error from the citation styles library. But this can be part of a different PR.

src/main/java/org/jabref/logic/citationstyle/CSLAdapter.java
halirutan: I profiled this and it's having no significant impact on the performance. After refactoring the creation of the `cslItemData` was absorbed into our data provider and must be run synchronized as well or the data we provide to CSL will get out of synch.

src/main/java/org/jabref/logic/citationstyle/CSLAdapter.java
halirutan: I refactored this. `retrieveItem` cannot change its signature because it is called internally by CSL (that's why it is overridden). Still, the layout is much clearer now.

halirutan:
Things that need to work:

1. Fast scrolling through the entry liest with preview open
2. Edit -> Copy Preview for single and many items
3. On a selection right click and Copy citation

@lenhard I cannot reproduce that the styles are not found with the fat jar. Can you try to delete and rebuild it and show exactly what you try? I see all styles and I can switch them as usual. The fat jar has the same structure as the normal jar with all styles included as files.

lenhard:
I am doing the following:
 - Execute `gradlew build`
 - cd to `build/libs`
 - execute `java -jar JabRef-4.1-dev.jar`
 - For some reason there is nothing available to select in the preview panel (see picture), although the CSL files are contained in the fat jar (not as a jar, but as the single .csl files right in the root of the jar). The log says 
```
Opening: C:\Users\joerg\Desktop\jl-main\diss\tex\references.bib
something went wrong while adding the discovered CitationStyles to the list 
```
preview

Furthermore, I am getting several warnings on the console:

```
Dec 16, 2017 10:02:18 PM javafx.scene.CssStyleHelper calculateValue
WARNING: Could not resolve '-fx-info' while resolving lookups for '-fx-text-fill' from rule '*.log' in stylesheet jar:file:/C:/workspaces/jabref/build/libs/JabRef-4.1-dev-fat.jar!/org/jabref/gui/errorconsole/ErrorConsole.css
```

After these warnings I updated my current installation from Java 1.8.0_144 to Java 1.8.0_152 and the warnings were gone, but the rest stays the same (no citation styles in the list). My configuration is now

```
JabRef 4.1-dev
Windows 10 10.0 amd64 
Java 1.8.0_152
```

Siedlerchr:
I tested it and I found there is a problem with how you get the path. Will commit a fix.

Siedlerchr:
@halirutan  @lenhard  I fixed the path handling and improved the code for searching inside the jar file.
For me it works now when I directly run JabRef Main in Eclipse 🙌 😃

halirutan:
@Siedlerchr Can you point out what exactly went wrong with the path so that I don't make the mistake in the future? Otherwise nice clean up! I removed your `println`, I guess you simply forgot this and I renamed `jarfs` to `jarFs` to satisfy my spell check (I like to have absolutely no warning if possible :smiley: ).

@lenhard Can you test it again. Your approach is basically what I did too and here it worked.

src/main/java/org/jabref/logic/citationstyle/CitationStyle.java
```diff
@@ -115,36 +120,51 @@ public static CitationStyle createCitationStyleFromFile(final String styleFile)
...
-                Enumeration<JarEntry> entries = file.entries();
-                while (entries.hasMoreElements()) {
-                    String filename = entries.nextElement().getName();
-                    if (!filename.startsWith("dependent") && filename.endsWith("csl")) {
-                        CitationStyle citationStyle = CitationStyle.createCitationStyleFromFile(filename);
-                        if (citationStyle != null) {
-                            citationStyles.add(citationStyle);
-                        }
+
+            Path filePath = Paths.get(CitationStyle.class.getProtectionDomain().getCodeSource().getLocation().toURI());
```
Siedlerchr: @halirutan The Paths.get method has an overload for accepting an URI. 
The misleading method of the URI class (getPath) does not work  as itdoesn't inlcude the "scheme" part of the URI object which is used to identify the underlying FileSytem implementation.
This tutorial has a nice overview of "what is a path" https://docs.oracle.com/javase/tutorial/essential/io/path.html

If you happen to use a Classpath url or sth else, the recommended way is the same  Paths.get(Class.toUrl.toUri)

halirutan: That wasn't my code :) When I see this right, it was coming from Christian Bartsch but it doesn't matter. Good that you spotted it.

PR #3539, opened 2017-12-15
Fix options databinding by adding elements if they are not in the initial map

### LINKED ISSUE #3538: Custom data in Type field in techreport not stored
Originally reported in the forum:
http://discourse.jabref.org/t/content-of-field-type-disappears-when-saving/947

1. Create a techreport entry
2. Enter a type e.g. myCustomTypeOfReports
3. Move to other field -> Content is gone 

I suspect that it is a simliar databinding issue in relation to the date field #3483

Siedlerchr:
Fixes #3538 

I dunno if this is the best idea, but it works...

----

### DIFF
--- a/src/main/java/org/jabref/gui/fieldeditors/MapBasedEditorViewModel.java
+++ b/src/main/java/org/jabref/gui/fieldeditors/MapBasedEditorViewModel.java
@@ -59,7 +59,7 @@ public abstract class MapBasedEditorViewModel<T> extends OptionEditorViewModel<T
         try {
             return (T) string;
         } catch (ClassCastException ex) {
-            LOGGER.error(String.format("Could not cast string to type %1s. Try overriding the method in a subclass and provide a conversion from string to the concrete type %1s", string.getClass()), ex);
+            LOGGER.error(String.format("Could not cast string to type %1$s. Try overriding the method in a subclass and provide a conversion from string to the concrete type %1$s", string.getClass()), ex);
         }
         return null;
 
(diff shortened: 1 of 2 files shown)


tobiasdiez:
Can you please explain the reason for the bug? In general, the string converter should not modify the source or target map.

Siedlerchr:
The reason for the bug is that the fromString method does a lookup in the map with the current txt in the field. If you enter a non predefined String into it, the map lookup returns null because not found.

src/main/java/org/jabref/gui/fieldeditors/MapBasedEditorViewModel.java
```diff
@@ -38,7 +39,11 @@ public T fromString(String string) {
                 if (string == null) {
                     return null;
                 } else {
-                    return getItemMap().getOrDefault(string, null);
+
+                    if (getItemMap().get(string) == null) {
+                        getItemMap().put(string, (T) string);
```
tobiasdiez: Ok thanks!
According to the documentation
> a string converter property is provided to allow for developers to specify how to translate a users string into an object of type T, such that the value property may contain it. By default the converter simply returns the String input as the user typed it, which therefore assumes that the type of the editable ComboBox is String.

So you are right that the string converter should return something non-null even when the text is not recognized. However, there is no need to add it to the item map. Moreover, we need to handle the case when `T` is not a string type. Thus I would propose to create a new protected method that converts an unknown string to a `T` object. If not overridden, it should try to cast to `T` as in your code above (and throw an exception with details how to properly implement it if casting was not successful).

Siedlerchr: I now added a method for checking.
We also need such kind for dates to allow custom values which we currently don't support. All input in the field will be cleared if not parseable

PR #3543, opened 2017-12-17
Automate Gradle dependency Update checks

LinusDietz:
I have created a new build job on Travis, that automatically checks if there are zero out-of date dependencies.
Features:
- [x] set up specific build job
- [x] It is checked on PR, but as an allowed failure. Otherwise it might be annoying in WIP Pull requests.

As soon as anybody sees that a dependency is out-of-date, on travis, we can update it automatically.

Furthermore, I regrouped the gradle tasks that belong together.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,10 +20,7 @@ env:
     - TEST_SUITE=guiTest
     - TEST_SUITE=checkstyle
     - TEST_SUITE=codecov
-    - if type == cron
-      notifications:
-        email:
-        - linus@lynyus.de
+
 
 matrix:
     fast_finish: true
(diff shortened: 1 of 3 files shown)


LinusDietz:
I have added a new gradle task for automatically checking if there is an outdated dependency:
```
$ gradle checkOutdatedDependencies
[...]

* Where:
Build file '/Users/dietzl/git/jabref/build.gradle' line: 219

* What went wrong:
A problem occurred evaluating root project 'JabRef'.
> There are outdated dependencies in build.gradle!
   Run ./gradlew dependencyUpdates to see which. Expression: (dependencyReport.outdated.count == 0)
```

This is actually reliable, since I have added our 'special' out-of date dependencies as exceptions the the dependency script.

PR #3551, opened 2017-12-19
Make path relative when adding automatically found file

### LINKED ISSUE #3476: automatically added file link has absolute instead of relative path
- [x ] I have tested the latest master version from http://builds.jabref.org/master/ and the problem persists

JabRef 4.1-dev--snapshot--2017-11-30--master--d45d7ca38
Linux 3.16.0-38-generic amd64 
Java 1.8.0_151
on Mint 17 (Ubuntu 14.04)

Steps to reproduce:

1. import new entry
2. add file with bibtex key into file directory
3. open new entry in entry editor

I just switched from v4.0 to the latest development version, because Medline import did not work in v4.0 anymore. Unfortunately, now files appear to be added with absolute paths to the library. E.g., I added entry Lake2016 and Jabref now automatically adds the corresponding file as `/home/bitzer/papers/Lake2016.pdf` instead of as `Lake2016.pdf` as before. So I have to manually edit the file link to get the relative path. Is that intended behaviour and how do I get Jabref to automatically add relative links again?

Siedlerchr:
Fixes #3476

----

### DIFF
--- a/src/test/java/org/jabref/gui/externalfiles/AutoSetFileLinksUtilTest.java
+++ b/src/test/java/org/jabref/gui/externalfiles/AutoSetFileLinksUtilTest.java
@@ -1,13 +1,11 @@
 package org.jabref.gui.externalfiles;
 
-import java.nio.file.Path;
 import java.util.Collections;
 import java.util.List;
 import java.util.TreeSet;
 
 import org.jabref.gui.externalfiletype.ExternalFileTypes;
 import org.jabref.logic.util.io.AutoLinkPreferences;
-import org.jabref.logic.util.io.FileUtil;
 import org.jabref.model.database.BibDatabaseContext;
 import org.jabref.model.entry.BibEntry;
 import org.jabref.model.entry.LinkedFile;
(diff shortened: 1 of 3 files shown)


src/test/java/org/jabref/gui/externalfiles/AutoSetFileLinksUtilTest.java
```diff
@@ -30,22 +31,25 @@
...
         when(externalFileTypes.getExternalFileTypeSelection()).thenReturn(new TreeSet<>(externalFileTypes.getDefaultExternalFileTypes()));
 
     }
 
     @Test
     public void test() {
         //Due to mocking the externalFileType class, the file extension will not be found
-        List<LinkedFile> expected = Collections.singletonList(new LinkedFile("", file.toString(), ""));
+
+        String relativeFilePath = FileUtil.shortenFileName(file, dirs).toString();
```
tobiasdiez: Well, now the test uses the same logic as the method under test. Could you please hard code the expected value. It should be just `CiteKey.pdf`, right?

PR #3553, opened 2017-12-20
Fix group search performance

### LINKED ISSUE #2852: Performance problems with new groups search/filter and large databases
JabRef 4.0.0-dev--snapshot--2017-05-18--master--018173ebd
Windows 10 10.0 amd64 
Java 1.8.0_131

The newly implemented groups filter/search (https://github.com/JabRef/jabref/pull/2588) exhibits massive performance problems when used on a large database (>10,000 entries, ~1,000 static groups).

1. Open large database
2. Start to enter any term you want to search for in the "Filter Groups" part of the groups panel.
3. You will immediately see a massive performance decrease. It takes minutes to just complete entering the word you want to search for (because the filtering starts immediately). The groups filtering itself also takes minutes to complete. During that time JabRef behaves as if it had crashed.

This issue has been reported before (https://github.com/JabRef/jabref/pull/2588#issuecomment-283042778, https://github.com/JabRef/jabref/pull/2588#issuecomment-302629336) but I have been advised to open a new ticket so that the issue is not forgotten.

halirutan:
This includes currently 2 changes:
Fixes #2852

1. The input field for the group search is delayed now. It waits until the user has finished typing and starts the search afterward. This increases the usability and prevents lags during typing.
2. The RecursiveTreeElement caches the current predicate so that branches of the group tree are not tested over and over again. This is an ugly solution so far but we need to test if this or something similar (but beautiful) boosts the performance considerably.

----


src/main/java/org/jabref/gui/groups/GroupTreeController.java
```diff
@@ -87,7 +90,13 @@ public void initialize() {
                 this::updateSelection
         );
 
-        viewModel.filterTextProperty().bind(searchField.textProperty());
+        // We try to to prevent publishing changes in the searchfield directly to the search task that takes some time
+        // for larger group structures.
+        final Timer searchTask = FxTimer.create(Duration.ofMillis(400), () -> {
```
tobiasdiez: I think, it makes sense to extract this code to a helper method in `BindingsHelper`.

src/main/java/org/jabref/gui/groups/GroupTreeController.java
```diff
@@ -87,7 +90,13 @@ public void initialize() {
...
+            LOGGER.info("Run Search " + searchField.getText());
```
tobiasdiez: I think debug level is enough.

halirutan: That log is really only for testing so that I could inspect when a new search was triggered and if my "reset the timer on new key" does actually work. This will be removed.

halirutan:
@tobiasdiez 

> The creation of nodes is expensive since the hit counter triggers a run over all entries. @halirutan based on your experience with the performance investigation, is my observation correct?

I guess, this is what I saw this morning in the profiler. I didn't investigate further, but I was surprised that I saw constructor-hits. I thought the group-tree is static until it is changed by adding/deleting nodes but instead it seems the tree is rebuilt. I haven't worked through the whole implementation because there is a lot of stuff in it I haven't worked with before JabRef. This is another reason why I had the hope someone could join on this branch because I don't feel confident to implement a good fix for this fundamental flaw in the design.

Siedlerchr:
Refs #2852

src/main/java/org/jabref/gui/groups/GroupTreeController.java
koppor: Can the log level be changed to "debug" and the logging settings adjusted accordingly for you?

tobiasdiez:
Ok, I think I found a solution which reuses the children, does not need to store the predicate but nonetheless has a good performance.

halirutan:
@lenhard Have you seen the comment in the issue this PR should solve?

> @tobiasdiez Unfortunately the new version is not working as expected. While it still filters the groups, when a filtered group is selected, the associated entries are not updated accordingly in the main table view. This was not the case with the previous version..

lenhard:
@halirutan Sorry, I did not look at the issue again and just tested for performance, not correctness. Of course that bug should be fixed before merging.

tobiasdiez:
Bug fixed -> merge! 😃

halirutan:
Great!

PR #3567, opened 2017-12-22
Add new authors to mailMap

LinusDietz:
These are the new authors of Jabref. :tada: 

 I have contacted @MBunel  and @weidafan, whether they can provide us their full names, so let's wait until the release before we merge this.

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -90,9 +90,9 @@ Ulrik Stervbo <ulriks@users.sourceforge.net>
 Stefano Gariazzo <steog88@gmail.com>
 Lee Patton <lpatton@users.sourceforge.net>
 Jörg Lenhard <joerg.lenhard@kau.se>
-Jörg Lenhard <joerg@lenhard.info>
-Jörg Lenhard <lenhard@users.noreply.github.com>
-Jörg Lenhard <joerg.lenhard@uni-bamberg.de>
+<joerg.lenhard@kau.se> <joerg@lenhard.info>
+<joerg.lenhard@kau.se> <lenhard@users.noreply.github.com>
+Jörg Lenhard <joerg.lenhard@kau.se> <joerg.lenhard@uni-bamberg.de>
 Matthias Geiger <matthias.geiger@uni-bamberg.de>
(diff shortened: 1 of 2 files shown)


.mailmap
```diff
@@ -131,4 +131,7 @@ Berk Gureken <berkgureken@gmail.com>
 Abhishek Rai <abhisheksavitarai@gmail.com>
 Abhishek Rai <betheunique@users.noreply.github.com>
 Stéphane Curet <stephane.curet@orange.fr>
-Erdem Derebasoglu <erdemderebasoglu@hotmail.com>
\ No newline at end of file
+Erdem Derebasoglu <erdemderebasoglu@hotmail.com>
+Domenico Cufalo <cufalo@gmail.com>
+Tobias Bouschen <tobias.bouschen@googlemail.com>
+József Pallagi <pallagijoe@gmail.com>
```
Siedlerchr: Now new line at eof? is this correct?

.mailmap
```diff
@@ -90,9 +90,9 @@ Ulrik Stervbo <ulriks@users.sourceforge.net>
 Stefano Gariazzo <steog88@gmail.com>
 Lee Patton <lpatton@users.sourceforge.net>
 Jörg Lenhard <joerg.lenhard@kau.se>
-<joerg.lenhard@kau.se> <joerg@lenhard.info>
```
koppor: Please do not change this to the real name.

The intention is to map all different email addresses of one user to a single email address. The reason for that is that tools might have name + email or even the email only as unique key for code analysis. When having different emails, users might be treated as separate users.

LinusDietz: okay, the problem is just, that now if I run the `generate-authors` script a new name `Joerg Lenhard` appears. How can I do that properly?

.mailmap
LinusDietz: yes? Is there any reason against this?

koppor:
Then add the Name in front additionally:

name mail mail

Am 22.12.2017 19:19 schrieb "Linus Dietz" <notifications@github.com>:

PR #3580, opened 2017-12-26
Remove Dependency for JUnit4

LinusDietz:
The JUnit4 tests are handled by the `junit-vintage-engine`, so we don't need the old dependency anymore. 🎉


tobiasdiez:
LGTM. Since this is only a minor thing, I'll merge directly.

Issue #2581, opened 2017-02-22
Wishes for new fetchers


stefan-kolb:
Issue to collect wishes for new fetcher implementations. We should also add a prioritization here.

Fulltext:
- [x] https://oadoi.org/: Implemented with #3581 
- [x] American Physical Society (APS) #818

Entry-based:
- [x] Worldcat.org #1065 

ID-based:
- [x] Ottobib #2238 

Search-based:
- [x] Crossref.org #2455: Implemented with #2645

CfKu:
By the way, ottobib is worldcat.org based
Worldcat API manual: http://xisbn.worldcat.org/xisbnadmin/doc/api.htm#schema and example http://kitchingroup.cheme.cmu.edu/blog/2015/01/31/Turn-an-ISBN-to-a-bibtex-entry/

r0light:
Is there currently a need for new fetchers? Since all the issues are closed, it seems like this issue could also be closed

Siedlerchr:
@r0light  This issue is still valid, it was created to have a central list of fetcher wishes and not an issue for each fetcher to keep the better overview

allisonsampaio:
I developed the option to use ottobib based on the chimbori, then after it try the ebook.de and chimbori.com is made the requisition for ottobib.

I'm not sure if the method returns the expected ottobib bibtex. I'm putting here the IsbnViaOttoBibFetcher code and the part of the IsbnFetcher code that makes the request for ottobib.

`
        HttpResponse<String> postResponse;

        String BASE_URL = "https://www.ottobib.com/isbn/" + identifier + "/bibtex";

        try {
            postResponse = Unirest.post(BASE_URL)
                    .asString();
        } catch (UnirestException e) {
            throw new FetcherException("Could not retrieve data from ottobib.com", e);
        }
        if (postResponse.getStatus() != 200) {
            throw new FetcherException("Error while retrieving data from ottobib.com: " + postResponse.getBody());
        }`

`if (!bibEntry.isPresent()) {
            LOGGER.debug("No entry found at ebook.de and chimbori.com try ottobib");
            IsbnViaOttoBibFetcher isbnViaOttoBibFetcher = new IsbnViaOttoBibFetcher(importFormatPreferences);
            bibEntry = isbnViaOttoBibFetcher.performSearchById(identifier);
        }`

allisonsampaio:
[...]

PR #3581, opened 2017-12-27
Add oaDOI fulltext fetcher

tobiasdiez:
As wished in https://github.com/JabRef/jabref/issues/2581.

----

### DIFF
--- a/src/main/java/org/jabref/logic/importer/fetcher/oaDOI.java
+++ b/src/main/java/org/jabref/logic/importer/fetcher/oaDOI.java
@@ -1,0 +1,58 @@
+package org.jabref.logic.importer.fetcher;
+
+import java.io.IOException;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.util.Objects;
+import java.util.Optional;
+
+import org.jabref.logic.importer.FulltextFetcher;
+import org.jabref.model.entry.BibEntry;
+import org.jabref.model.entry.FieldName;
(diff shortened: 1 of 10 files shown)


src/main/java/org/jabref/logic/importer/fetcher/oaDOI.java
```diff
@@ -0,0 +1,58 @@
...
+import com.mashape.unirest.http.Unirest;
+import com.mashape.unirest.http.exceptions.UnirestException;
+import org.json.JSONObject;
+
+/**
+ * A fulltext fetcher that uses <a href="https://oadoi.org/">oaDOI</a>.
+ *
+ * @implSpec API is documented at https://oadoi.org/api/v2
+ */
+public class oaDOI implements FulltextFetcher {
```
lenhard: That's not an "ok" name according to the naming conventions for Java classes: they need to start with upper case. Please rename this to `OaDOI` or maybe `OaDoi`. Maybe `OpenAccessDoi` would be even better.

tobiasdiez: Good point, I wasn't sure about the naming conventions and this service seems to be really called `oaDOI`. But `OpenAccessDoi` is good.

tobiasdiez:
@lenhard It is only a fulltext fetcher. Thus it never appears in the GUI but is used when you click the "fetch fulltext" button in the entry editor. It looks for an URL for the paper based on the information in the entry (such as DOI).

PR #3591, opened 2017-12-30
Maintable leaves the Swing year and comes back in a new JavaFX dress

### LINKED ISSUE #3112: Colouring item count in groups panel does not work reliably
JabRef 4.0-dev--snapshot--2017-08-16--maintable-update--69f717336
Windows 10 10.0 amd64
Java 1.8.0_141

When selecting an item in the main table, the associated group (or rather the number giving the number of associated items) should be coloured. This is not the case with newly assigned items. It does appear, however, that this feature is generally buggy - also items that have not been newly assigned to a group are not always coloured in the group panel (sometimes it works, sometimes it does not).
It's difficult for me to replicate this behaviour. As far as I can tell, one way of experiencing reliably this bug is by doing the following.

1) Open a JabRef database
2) Select an item - the associated item count in the groups panel should be coloured (although sometimes it won't)
3) Assign this item to a new group - the item count of the new group will not be coloured

Note, that I've also noticed that colours can "persist" in the main table, although no or only one item is selected. Example:
1) Open a JabRef database
2) Select multiple items which are associated with different groups - the item counts of the different groups should be coloured
3) Now deselect all items and select only one that is associated with a single group - only the item count of that group should be coloured. But actually all item counts of the previously selected items/groups are still coloured.

### LINKED ISSUE #3356: LaTeX `\textit{..}` is not redered correctly in Main table
Entry to reproduce:

```
@Article{Abdala2007,
  author    = {Fernando Abdala},
  title     = {{{Redescription} {of} \textit{Platycraniellus elegans} ({Therapsida}, {Cynodontia}) {from} {the} {Lower} {Triassic} {of} {South} {Africa}, {and} {the} {cladistic} {relationships} {of} {eutheriodonts}}},
  journal   = {Palaeontology},
  year      = {2007},
  volume    = {50},
  number    = {3},
  pages     = {591--618},
  month     = {may},
  doi       = {10.1111/j.1475-4983.2007.00646.x},
  file      = {Abdala2007.pdf:Abdala2007.pdf:PDF},
  groups    = {Nanictosaurus, Ictidosuchoides},
  keywords  = {Platycraniellus elegans, Cynodontia, galesaurids, Early Triassic, South Africa.},
  publisher = {Wiley-Blackwell},
  url       = {http://dx.doi.org/10.1111/j.1475-4983.2007.00646.x},
}
```

Gives the following representation in the main table

screen shot 2017-10-26 at 12 38 55

The Preview is not affected.

```
MacOS Sierra 10.12.6
JabRef current master branch
Mac OS X 10.12.6 x86_64 
Java 1.8.0_144
```

### LINKED ISSUE #3532: Selecting new entry causes editing panel to become corrupted and not accept input
MacOS 10.12.6
java version "1.8.0_151"
Version: JabRef-4.1-dev--snapshot--2017-12-08--master--6f39151a7.jar
Invoked with "java -jar" on a command line

I moved to using the latest version of JabRef a couple of days ago (as above), and
have a new bug. JabRef works well for while, but at a certain point, trying
to add a new entry causes the field area to hang and become
non-responsive. (Figure hopefully included.) 

The following error is generated:

19:30:17.470 [JavaFX Application Thread] ERROR org.jabref.FallbackExceptionHandler - Uncaught exception occurred in Thread[JavaFX Application Thread,5,main]
java.lang.IndexOutOfBoundsException: [-2147483648, 16) is not a valid range within [0, 16)
	at org.reactfx.util.Lists.checkRange(Lists.java:128) ~[JabRef-4.1-dev--snapshot--2017-12-08--master--6f39151a7.jar:?]
	at org.reactfx.util.Lists.checkRange(Lists.java:123) ~[JabRef-4.1-dev--snapshot--2017-12-08--master--6f39151a7.jar:?]
	at org.reactfx.collection.MemoizationListImpl.forget(MemoizationList.java:184) ~[JabRef-4.1-dev--snapshot--2017-12-08--master--6f39151a7.jar:?]
	at org.fxmisc.flowless.CellListManager.cropTo(CellListManager.java:82) ~[JabRef-4.1-dev--snapshot--2017-12-08--master--6f39151a7.jar:?]
	at org.fxmisc.flowless.CellPositioner.cropTo(CellPositioner.java:29) ~[JabRef-4.1-dev--snapshot--2017-12-08--master--6f39151a7.jar:?]
	at org.fxmisc.flowless.Navigator.cropToNeighborhoodOf(Navigator.java:227) ~[JabRef-4.1-dev--snapshot--2017-12-08--master--6f39151a7.jar:?]
	at org.fxmisc.flowless.Navigator.placeStartAtMayCrop(Navigator.java:200) ~[JabRef-4.1-dev--snapshot--2017-12-08--master--6f39151a7.jar:?]
	at org.fxmisc.flowless.Navigator.visit(Navigator.java:129) ~[JabRef-4.1-dev--snapshot--2017-12-08--master--6f39151a7.jar:?]
	at org.fxmisc.flowless.StartOffStart.accept(TargetPosition.java:85) ~[JabRef-4.1-dev--snapshot--2017-12-08--master--6f39151a7.jar:?]
	at org.fxmisc.flowless.Navigator.layoutChildren(Navigator.java:78) ~[JabRef-4.1-dev--snapshot--2017-12-08--master--6f39151a7.jar:?]
	at javafx.scene.Parent.layout(Parent.java:1087) ~[jfxrt.jar:?]
	at org.fxmisc.flowless.VirtualFlow.layoutChildren(VirtualFlow.java:257) ~[JabRef-4.1-dev--snapshot--2017-12-08--master--6f39151a7.jar:?]
	at javafx.scene.Parent.layout(Parent.java:1087) ~[jfxrt.jar:?]
	at javafx.scene.Parent.layout(Parent.java:1093) ~[jfxrt.jar:?]
	at javafx.scene.Parent.layout(Parent.java:1093) ~[jfxrt.jar:?]
	at javafx.scene.Parent.layout(Parent.java:1093) ~[jfxrt.jar:?]
	at javafx.scene.Scene.doLayoutPass(Scene.java:552) ~[jfxrt.jar:?]
	at javafx.scene.Scene.preferredSize(Scene.java:1646) ~[jfxrt.jar:?]
	at javafx.scene.Scene.impl_preferredSize(Scene.java:1720) ~[jfxrt.jar:?]
	at javafx.stage.Window$9.invalidated(Window.java:846) ~[jfxrt.jar:?]
	at javafx.beans.property.BooleanPropertyBase.markInvalid(BooleanPropertyBase.java:109) ~[jfxrt.jar:?]
	at javafx.beans.property.BooleanPropertyBase.set(BooleanPropertyBase.java:144) ~[jfxrt.jar:?]
	at javafx.stage.Window.setShowing(Window.java:922) ~[jfxrt.jar:?]
	at javafx.stage.Window.show(Window.java:937) ~[jfxrt.jar:?]
	at com.sun.javafx.stage.EmbeddedWindow.show(EmbeddedWindow.java:58) ~[jfxrt.jar:?]
	at javafx.embed.swing.JFXPanel.lambda$addNotify$49(JFXPanel.java:799) ~[jfxrt.jar:?]
	at com.sun.javafx.application.PlatformImpl.lambda$null$172(PlatformImpl.java:295) ~[jfxrt.jar:?]
	at java.security.AccessController.doPrivileged(Native Method) ~[?:1.8.0_151]
	at com.sun.javafx.application.PlatformImpl.lambda$runLater$173(PlatformImpl.java:294) ~[jfxrt.jar:?]
	at com.sun.glass.ui.InvokeLaterDispatcher$Future.run(InvokeLaterDispatcher.java:95)
	~[jfxrt.jar:?]

The previous versions I was using,

JabRef-4.1-dev--snapshot--2017-11-13--master--c1d425a31.jar and

JabRef-4.1-dev--snapshot--2017-10-29--autofilelink--19481127d.jar

didn't have this problem.

screen shot 2017-12-14 at 7 40 10 pm

Clicking the check mark top-left of the panel closes and opens the panel correctly, but doesn't restore it to a working state. The following info is generated:

java.lang.RuntimeException: Requested texture dimensions (2516x2147483647) require dimensions (2516x0) that exceed maximum texture size (16384)
	at com.sun.prism.es2.ES2RTTexture.create(ES2RTTexture.java:220)
	at com.sun.prism.es2.ES2ResourceFactory.createRTTexture(ES2ResourceFactory.java:157)
	at com.sun.javafx.tk.quantum.UploadingPainter.run(UploadingPainter.java:117)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:308)
	at com.sun.javafx.tk.RenderJob.run(RenderJob.java:58)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at com.sun.javafx.tk.quantum.QuantumRenderer$PipelineRunnable.run(QuantumRenderer.java:125)
	at java.lang.Thread.run(Thread.java:748)

tobiasdiez:
2017:
old

2018:
image

There are still many many things missing for a fully working version but I thought I give you an early preview. Feedback welcome.
Happy new year everybody!

Fixes #3356, fixes #3532 and fixes #3112.

Known bugs: (ticked = resolved)
- [ ] Changes to entries don't mark database as changed.
- [x] Changes to entries are not reflected in main table.
- [x] Cleanup does not work.
- [x] Ctrl + A in entry editor selects all entries in table. [not reproducable]
- [x] Escape does not close entry editor.
- [x] Column width is not correct if the option "Resize to fit window width" is activated
- [ ] Select multiple entries by holding mouse button: [code to fix](https://community.oracle.com/thread/2621389)

Status of features:
- [x] Sorting (Shift + click for multi-column sort)
- [x] Filtering by search and groups 
- [ ] Context menus for the icons
- [ ] Right-click menu
- [ ] Setting special fields like reading status / priority / ... through the main table
- [ ] Marked entries are not highlighted
- [ ] Drag & drop to and from the table
- [ ] Float mode does not work
- [ ] Color cells in table depending on whether the field is optional/required

----


Siedlerchr:
Cool! Could you point me to new implementation? In your large commit I see only a lot of Icon changes

tobiasdiez:
Oh, sorry - I forgot to commit all files. The main changes are in the `gui.maintable` package.

Siedlerchr:
While you are working on this, you might also take a look at #3356  
We could use the Latex2Html converter and then convert that HTML formatting to the TextFlow which was done in the search component tooltips

tobiasdiez:
Yes, this issue is already fixed merely by switching to JavaFX - the unicode is correctly displayed (see above the second entry in the table, the first screenshot shows the boxes while the correct string is shown in the second screenshot).

src/main/java/org/jabref/gui/maintable/ColumnFactory.java
```diff
@@ -0,0 +1,185 @@
...
+import org.jabref.gui.specialfields.SpecialFieldViewModel;
+import org.jabref.gui.util.ValueTableCellFactory;
+import org.jabref.model.database.BibDatabase;
+import org.jabref.model.entry.FieldName;
+import org.jabref.model.entry.LinkedFile;
+import org.jabref.model.entry.specialfields.SpecialField;
+
+import org.fxmisc.easybind.EasyBind;
+
+class ColumnFactory {
```
Siedlerchr: I am not that happy with this generic name here... as this shadows the javafx classes 
Maybe MaintableColumnFactory? Same as with the other CellFactory

src/main/java/org/jabref/gui/maintable/ColumnFactory.java
```diff
@@ -0,0 +1,185 @@
...
+            }
+        }
+
+        // Add column for eprints
+        if (preferences.showEprintColumn()) {
+            columns.add(createIconColumn(IconTheme.JabRefIcons.WWW, FieldName.EPRINT));
+        }
+
+        // Add columns for other file types
+        for (String column : preferences.getExtraFileColumns()) {
```
Siedlerchr: Maybe also addAll possible?=

src/main/java/org/jabref/gui/maintable/StringTableColumn.java
```diff
@@ -0,0 +1,158 @@
...
+    @Override
+    public ObservableValue<String> getColumnValue(BibEntryTableViewModel entry) {
+        if (bibtexFields.isEmpty()) {
+            return null;
+        }
+        boolean isNameColumn = false;
+
+        Optional<String> content = Optional.empty();
+        for (String field : bibtexFields) {
+            content = entry.getResolvedFieldOrAlias(field, database.orElse(null));
```
Siedlerchr: Maybe stream with firstOrNull? or anyMatch?

src/main/java/org/jabref/migrations/FileLinksUpgradeWarning.java
```diff
@@ -248,17 +248,13 @@ private void makeChanges(BasePanel panel, ParserResult pr, boolean upgradePrefs,
     }
 
     private boolean showsFileInGenFields() {
-        boolean found = false;
-        EntryEditorTabList tabList = Globals.prefs.getEntryEditorTabList();
-        outer: for (int i = 0; i < tabList.getTabCount(); i++) {
-            List<String> fields = tabList.getTabFields(i);
-            for (String field : fields) {
+        for (Map.Entry<String, List<String>> tab : Globals.prefs.getEntryEditorTabList().entrySet()) {
+            for (String field : tab.getValue()) {
```
Siedlerchr: tab.stream.anymatch( entry -> entry.getKey.equals(FieldName.File)

lenhard:
One more step in the right direction. This PR has the potential to fix a lot of display bugs and we should be able to get rid of the annoying and unmaintained glazed lists library.

Nevertheless, I need to add a word of causing: We cannot allow the same situation as for the entry editor again. We've spent almost have a year in un-releaseable limbo because of that. The replacement for the main table needs to be absolutely bullet proof. However, the amount of  lines of code in this PR is too much for a meaningful review, so we need A LOT of manual testing by basically all developers.

I've played around a bit and I like the new layout of the maintable. What's a little strange is that every second row is highlighted differently. This should change for the final version.

tobiasdiez:
@Siedlerchr Thanks for the feedback. I'll improve the code accordingly.

@lenhard I completely agree! We should find a way to have a realeasable version while still work on the migration to JavaFX. The main problem is that the code is tightly coupled to the maintable (i.e. there is no abstraction layer before the BasePanel/MainTable that used in the other Swing dialogs). This has the potential upset that every single dialog may now trigger a bug (for example because it runs some code not on the JavaFX thread). So, even through I will try to thoroughly test the new code before marking it as "ready-for-review", there might still be countless bugs left which we need to find (and fix) in teamwork. Moreover, for the same reason, I think we also need to release a beta again.

That being said, the immediate question is: how much functionality should I try to migrate in this PR and what should follow as new PRs? As you said, the size of the changes makes the PR already unreviewable. But I would really like to have your feedback about some design questions. Especially since your aim should be not just to improve the gui but also decouple the code and make it more reusable.

Currently the new maintable is in a state where the very basic functionality works and the code is decoupled enough to actually write unit gui tests for it. There are still some bugs/feature that definitely need to be fixed before a users touches it (e.g. restore column width, context menus for the icons, a few display bugs). Then there are a few features that are not essential and rather isolated (e.g. "marking of entries"). These are probably good candidates for follow-up PRs. And finally there are features that I don't really see any benefit in reimplementing (e.g. coloring of table cells when the field is optional/required) or that are not really possible to implement with JavaFX (e.g. ["float search"](https://bugs.openjdk.java.net/browse/JDK-8091326)). Finally, the possible bug fixes @lenhard mentions should also follow as new PRs.

Any thoughts on this? (Sorry for the long text).

Siedlerchr:
I think we could have a separate main Table Beta Test-Release in an additional branch highly experimental and let some users test this functionality who are interested. Especially we would need some Mac/Linux guys

koppor:
We can create a `javafx` branch, where we can integrate not ready features, but completed work packages (such as a basic entry table in JavaFX)

The master branch must be release-ready at all times. Release-ready means: all features of the old versions should be working (or the dev team decided to remove/disable a feature, such as the auto completion).

tobiasdiez:
So, I used this version now in production during the past few days and fixed a lot of bugs. Although not all features are yet migrated (see list above), I mark this PR now as ready for review. All further changes are coming as extra PRs onto this branch. 

I would suggest that you try this version out and if nothing serious pops up, I would create a post in the forum advertising this PR and ask users about their opinion (something like a very early access program instead of several beta releases). What do you think?

lenhard:
I would be really cool if you could provide a table that lists all features related to the maintable that were present in the Swing version, e.g. sorting options, right click options, entry drop, column sorting, and could indicate if these features are working in the new version. Then we would have an overview of what features actually exist in the main table and we could have a meaningful discussion about which features we want to drop.

Personally, I don't really have an overview of all the things that the maintable can do. Could you perhaps provide this overview? I guess also @koppor know a thousand maintable features that I do not know.

koppor:
I changed the base branch to "maintable-beta". I would propose to proceed as follows:

1. @JabRef/developers review
2. Work-in review comments
3. Merge into `maintable-beta`.
4. Ask users to feedback to `maintable-beta`
5. Add new PRs on `maintable-beta`

Why? 

1. I think, PRs should be short and be closed fast. Not left open as the ones of @koppor. So, we can move on quickly. And we can also link to comments of this PR if necessary.
2. The discussions here are mixed on the implementation details and overall issues. I think, we should focus on implementation details 😇 

In case we need a overview-pr, we can create a PR from `maintable-beta` to `master` and describe the overall status there.

koppor:
I am *very* used to the floating search. See screenshots at https://github.com/JabRef/jabref/issues/3618.

tobiasdiez:
@lenhard I updated the first post to also include the features that are already migrated (it is only the very basic functionality). Right now we do not need to decide what features should be dropped since there is still a lot fundamental functionality to migrate (but we should speak about it in the next devcall). I just don't want this PR to explode even more and get feedback about some of the other refactorings / changes. Thus the early ready-for-review. 

@koppor Sounds like a good plan.

tobiasdiez:
@lenhard I updated the first post to also include the features that are already migrated (it is only the very basic functionality). Right now we do not need to decide what features should be dropped since there is still a lot fundamental functionality to migrate (but we should speak about it in the next devcall). I just don't want this PR to explode even more and get feedback about some of the other refactorings / changes. Thus the early ready-for-review. 

@koppor Sounds like a good plan.

Siedlerchr:
I fixed the Close Entry Editor with Escape bug

src/main/java/org/jabref/gui/maintable/StringTableColumn.java
tobiasdiez: not possible because `isNameColumn` has to be set.

src/main/java/org/jabref/migrations/FileLinksUpgradeWarning.java
tobiasdiez: Not sure that this is more readable, especially since the if statement still has to be there.

koppor:
@lenhard @Siedlerchr Is this architecturally OK? Then, I would propose to merge it into `maintable-beta` to reduce the size of PRs.

Note: The "real" PR of getting this into JabRef is https://github.com/JabRef/jabref/pull/3621 - there, code updates are only made as PRs to that branch.

src/main/java/org/jabref/model/entry/BibtexSingleField.java
```diff
@@ -8,7 +8,8 @@
  */
 public class BibtexSingleField {
 
-    public static final int DEFAULT_FIELD_LENGTH = 100;
+    // TODO: This constant should be moved to the gui package, probably to MainTableColumnFactory
+    public static final double DEFAULT_FIELD_LENGTH = 100;
```
lenhard: Would it be possible to move this as suggested in the comment?

tobiasdiez: This is not possible at the moment since the field is still used in other places in model. The new main table is free from this dependency but the import and search dialog still rely on it as the have before.

lenhard: Ok, no problem. Then just leave as it is.

PR #3603, opened 2018-01-04
Catch and propagate UncheckedIOException in FileFinder

tobiasdiez:
Catch UncheckedIOException and converted it to a usual IOException. Similarly for an `PatternSyntaxException`. These are ones of the rare exceptions that were never handled but are reported in azure.

----

- [ ] Change in CHANGELOG.md described (not worth it)


src/main/java/org/jabref/logic/util/io/CiteKeyBasedFileFinder.java
```diff
@@ -93,8 +94,8 @@ private boolean matches(String filename, String citeKey) {
             if (Files.exists(directory)) {
                 try (Stream<Path> files = Files.find(directory, Integer.MAX_VALUE, isFileWithCorrectExtension)) {
                     result.addAll(files.collect(Collectors.toSet()));
-                } catch (IOException e) {
-                    LOGGER.error("Problem in finding files", e);
+                } catch (UncheckedIOException e) {
```
Siedlerchr: That's interesting, why/when does this Exception occur?

tobiasdiez: To be honest, I have no idea. The documentation babbles something about "a late exception that is thrown when the stream is finished". I guess the file is found, added to the stream but when a bit later `toSet` is called it is no longer there. I also don't understand why it has to be an unchecked exception instead of a normal `IOException`.

Siedlerchr: I refactored the code slightly to get rid of these exception handling in streams.
I think this is should work better now because the stream is no longer directly closed and then accessed. But I could be mistaken.

src/main/java/org/jabref/logic/util/io/CiteKeyBasedFileFinder.java
```diff
@@ -85,18 +82,13 @@ private boolean matches(String filename, String citeKey) {
...
 
         Set<Path> result = new HashSet<>();
         for (Path directory : directories) {
             if (Files.exists(directory)) {
-                try (Stream<Path> files = Files.find(directory, Integer.MAX_VALUE, isFileWithCorrectExtension)) {
-                    result.addAll(files.collect(Collectors.toSet()));
-                } catch (UncheckedIOException e) {
-                    throw new IOException("Problem in finding files", e);
-                }
+                result.addAll(Files.find(directory, Integer.MAX_VALUE, isFileWithCorrectExtension).collect(Collectors.toSet()));
```
tobiasdiez: Can you please explain this change and why you removed the try-catch as the point of this PR was to actually catch this `UncheckedIOException`.

src/main/java/org/jabref/logic/util/io/CiteKeyBasedFileFinder.java
tobiasdiez: I don't think that works (see below). But thanks for the other fix, that's a nice improvement over my hacky code.

src/main/java/org/jabref/logic/util/io/CiteKeyBasedFileFinder.java
Siedlerchr: Your code did not compile, you can't throw an exception inside a lambda/stream. That's impossible. 
I did not realize that the unchecked exception can still occur. But we should just log it. I mean, what else can we do about it?`We can't rethrow it.

tobiasdiez: I just pushed a new solution. I think this kind of catch-rethrow should work.

tobiasdiez:
I merge this now so that it can serve as the base for https://github.com/JabRef/jabref/pull/3623.

Issue #3576, opened 2017-12-25
missing dot in files exported 


ziruizhuang:
issues with JabRef version 4.1
however, 4.0 is fine.

Steps to reproduce:

1. File - Export
2. Select MS2007 XML format
3. Save

Expected: a file saved with name "NAME.XML"
Got: a file saved with name "NAMEXML"

Siedlerchr:
Probably related to #3565

tgkolda:
I'm having the same problem.

tobiasdiez:
@ziruizhuang @tgkolda I think, I fixed the bug. Could you please try out the version from https://builds.jabref.org/fixExport3576/ and report back if these issues are indeed resolved. @tgkolda what do you mean with "custom export"?

tgkolda:
@tobiasdiez  Thanks for your response. The fixExport3576 version didn't run at all for me, so I couldn't check the fix. Sorry. As for custom export, I refer to this capability: https://help.jabref.org/en/CustomExports, of which I am an enthusiastic user.

tobiasdiez:
This should be fixed in the latest development version. Could you please check the build from http://builds.jabref.org/master/. Thanks!

ziruizhuang:
solved this issue, yet introduced another one.
The export format varies, so usually  we call a format "Microsoft Word Citation (*.XML)" "Open Document Citation (*.XML)", etc. 
Instead, in the latest snapshot(http://builds.jabref.org/master/JabRef_windows-x64_4_2-dev--snapshot--2018-01-05--master--d6435d56d.exe), there are multiple "XML file (*.XML)". Although they are all en-capsuled using XML, but they are different formats and now we cannot identify one from another.

PR #3606, opened 2018-01-05
Add exporter desc to enum analog to import

### LINKED ISSUE #3605: Export dialog with unclear format list
issues with JabRef version 4.2 snapshot (http://builds.jabref.org/master/JabRef_windows-x64_4_2-dev--snapshot--2018-01-05--master--d6435d56d.exe)

Steps to reproduce:

* File - Export

**Expected:** a list of formats such as "MS2007 XML format (*.XML)", "Open Document Citation (*.XML)"

**Got:** a list of formats with only the filename extension for indication, such as "XML file (*.XML)", "XML file (*.XML)"

Although they are all en-capsuled using XML, but they are different formats and now we cannot identify one from another. Other formats suffer from this issue as well.

Siedlerchr:
Fixes #3605

Follow up from #3576 

jabrefextension

----


src/main/java/org/jabref/logic/exporter/ExporterFactory.java
```diff
@@ -36,21 +35,21 @@ public static ExporterFactory create(Map<String, TemplateExporter> customFormats
...
-        exporters.add(new TemplateExporter("Harvard RTF", "harvard", "harvard", "harvard", FileType.RDF, layoutPreferences, savePreferences));
-        exporters.add(new TemplateExporter("ISO 690 RTF", "iso690rtf", "iso690RTF", "iso690rtf", FileType.RTF, layoutPreferences, savePreferences));
-        exporters.add(new TemplateExporter("ISO 690", "iso690txt", "iso690", "iso690txt", FileType.TXT, layoutPreferences, savePreferences));
-        exporters.add(new TemplateExporter("Endnote", "endnote", "EndNote", "endnote", FileType.TXT, layoutPreferences, savePreferences));
-        exporters.add(new TemplateExporter("OpenOffice/LibreOffice CSV", "oocsv", "openoffice-csv", "openoffice", FileType.CSV, layoutPreferences, savePreferences));
-        exporters.add(new TemplateExporter("RIS", "ris", "ris", "ris", FileType.RIS, layoutPreferences, savePreferences).withEncoding(StandardCharsets.UTF_8));
-        exporters.add(new TemplateExporter("MIS Quarterly", "misq", "misq", "misq", FileType.RTF, layoutPreferences, savePreferences));
+        exporters.add(new TemplateExporter(FileType.HTML.getDescription(), "html", "html", null, FileType.HTML, layoutPreferences, savePreferences));
```
tobiasdiez: It appears that now the description is always the description of the file type, right? Then it makes sense to create a new constructor of `TemplateExporter` that automatically sets the description.

Siedlerchr: I thought so, too but on the other hand we have the constructor needed for the custom exporter, so in the end it's the same

tobiasdiez: but at least we could add a new constructor, so that this list gets more readable, right?

PR #3623, opened 2018-01-09
Fix ErrorProne errors

koppor:
This fixes

```
/home/ubuntu/jabref/src/main/java/org/jabref/logic/citationstyle/CitationStyle.java:156: error: [StreamResourceLeak] Streams that encapsulate a closeable resource should be closed using try-with-resources
                List<Path> allStyles = Files.find(jarFs.getRootDirectories().iterator().next(), 1, (file, attr) -> file.toString().endsWith("csl")).collect(Collectors.toList());
                                                 ^
    (see http://errorprone.info/bugpattern/StreamResourceLeak)
  Did you mean 'List<Path> allStyles ;'?
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
1 error
```

### DIFF
--- a/src/main/java/org/jabref/logic/citationstyle/CitationStyle.java
+++ b/src/main/java/org/jabref/logic/citationstyle/CitationStyle.java
@@ -124,6 +124,7 @@ public class CitationStyle {
 
     /**
      * Provides the default citation style which is currently IEEE
+     *
      * @return default citation style
      */
     public static CitationStyle getDefault() {
@@ -132,6 +133,7 @@ public class CitationStyle {
 
     /**
      * Provides the citation styles that come with JabRef.
+     *
      * @return list of available citation styles
      */
     public static List<CitationStyle> discoverCitationStyles() {
@@ -155,13 +157,12 @@ public class CitationStyle {
 
             try (FileSystem jarFs = FileSystems.newFileSystem(Paths.get(path), null)) {
 
-                List<Path> allStyles;
                 try (Stream<Path> stylefileStream = Files.find(jarFs.getRootDirectories().iterator().next(), 1, (file, attr) -> file.toString().endsWith("csl"))) {
-                   allStyles = stylefileStream.collect(Collectors.toList());
-                }
-                for (Path style : allStyles) {
-                    CitationStyle.createCitationStyleFromFile(style.getFileName().toString())
-                            .ifPresent(STYLES::add);
+                    for (Path style : stylefileStream.collect(Collectors.toList())) {
+                        CitationStyle.createCitationStyleFromFile(style.getFileName().toString()).ifPresent(STYLES::add);
+                    }
+                } catch (UncheckedIOException e) {
+                    throw new IOException(e);
                 }
             }
             return STYLES;
@@ -212,5 +213,4 @@ public class CitationStyle {
     public int hashCode() {
         return Objects.hash(source);
     }
-
 }
--- a/src/main/java/org/jabref/logic/citationstyle/CitationStyle.java
+++ b/src/main/java/org/jabref/logic/citationstyle/CitationStyle.java
@@ -19,6 +19,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;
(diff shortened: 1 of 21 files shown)


src/main/java/org/jabref/logic/citationstyle/CitationStyle.java
```diff
@@ -153,8 +154,10 @@ public static CitationStyle getDefault() {
 
             try (FileSystem jarFs = FileSystems.newFileSystem(Paths.get(path), null)) {
 
-                List<Path> allStyles = Files.find(jarFs.getRootDirectories().iterator().next(), 1, (file, attr) -> file.toString().endsWith("csl")).collect(Collectors.toList());
-
+                List<Path> allStyles;
+                try (Stream<Path> stylefileStream = Files.find(jarFs.getRootDirectories().iterator().next(), 1, (file, attr) -> file.toString().endsWith("csl"))) {
```
Siedlerchr: This is again one of the ugly cases where an unchecked exception could be thrown. Because streams and lambdas can't throw checked exceptions. In this case it could be that the stream is already closed when calling collect. 
 Similar to #3606

tobiasdiez: Yes, please catch `IOUncheckedException` and rethrow it as `IOException`.

src/main/java/org/jabref/logic/util/io/FileUtil.java
```diff
@@ -331,8 +331,8 @@ public static String createDirNameFromPattern(BibDatabase database, BibEntry ent
      * @return the path to the first file that matches the defined conditions
      */
     public static Optional<Path> find(String filename, Path rootDirectory) {
-        try {
-            return Files.walk(rootDirectory)
+        try (Stream<Path> pathStream = Files.walk(rootDirectory)) {
```
tobiasdiez: Can you please check if `walk` may also throw an `IOUncheckedEx` and if yes, add it to the catch block.

src/main/java/org/jabref/logic/util/io/RegExpBasedFileFinder.java
```diff
@@ -178,14 +174,12 @@ public static String expandBrackets(String bracketString, BibEntry entry, BibDat
...
 
-                try {
-                    Path finalActualDirectory = actualDirectory;
-                    Files.walk(actualDirectory).forEach(subElement -> {
-                        // We only want to transverse directory (and not the current one; this is already done below)
-                        if (!finalActualDirectory.equals(subElement) && Files.isDirectory(subElement)) {
-                            res.addAll(findFile(entry, subElement, restOfFileString, extensionRegExp));
-                        }
-                    });
+                try (Stream<Path> pathStream = Files.walk(actualDirectory)) {
```
tobiasdiez: here also

src/main/java/org/jabref/logic/util/io/RegExpBasedFileFinder.java
```diff
@@ -197,14 +191,21 @@ public static String expandBrackets(String bracketString, BibEntry entry, BibDat
...
-                    (path, attributes) -> toMatch.matcher(path.getFileName().toString()).matches())
-                    .collect(Collectors.toList());
-            res.addAll(matches);
+        BiPredicate<Path, BasicFileAttributes> matcher = (path, attributes) -> toMatch.matcher(path.getFileName().toString()).matches();
+        res.addAll(collectMatchingFiles(actualDirectory, matcher));
+        return res;
+    }
+
+    private List<Path> collectMatchingFiles(Path actualDirectory, BiPredicate<Path, BasicFileAttributes> matcher) {
+        try (Stream<Path> pathStream = Files.find(actualDirectory, 1, matcher)) {
```
tobiasdiez: and here

src/main/java/org/jabref/logic/util/io/FileUtil.java
LinusDietz: Checked the JavaDoc and the sources... does not seem so.

LinusDietz:
@tobiasdiez @Siedlerchr Can you have a look if the exception handling is correct now? see c93ae63 for that

src/main/java/org/jabref/logic/util/io/RegExpBasedFileFinder.java
```diff
@@ -195,14 +191,24 @@ public static String expandBrackets(String bracketString, BibEntry entry, BibDat
...
-        return res;
+
+        return resultFiles;
+    }
+
+    private List<Path> collectFilesWithMatcher(Path actualDirectory, BiPredicate<Path, BasicFileAttributes> matcher) {
+        try (Stream<Path> pathStream = Files.find(actualDirectory, 1, matcher)) {
+            return pathStream.collect(Collectors.toList());
+        } catch (UncheckedIOException | IOException ioe) {
+            return Collections.emptyList();
```
tobiasdiez: I think this method should just rethrow as IOException.

tobiasdiez:
@LinusDietz Thanks, the code looks good to me!

Issue #3586, opened 2017-12-28
Provide and advertise a chat for developers


jonasstein:
The **bug tracker is often used as chat** (see #3382 and others), which is obviously not good.
I had a short search and found no IRC channel (or something similar), where new and old developers can simply discuss things which should not be archived together with the bug ticket.

Please create such a chat and advertise it on the related pages.

The head area of the project page would be a prominent place to advertise your chat. Even new contributors would see it:
![selection_430](https://user-images.githubusercontent.com/298682/34416788-789c85b8-ebf5-11e7-990a-2bf0d026578a.png)

halirutan:
I too prefer a chat what it comes to issues where a live discussion is helpful. We have an abandoned Gitter chat that you can find here:

https://gitter.im/JabRef/jabref

We had a similar discussion about a chat for IntelliJ Plugin Developers and there are many alternatives: Slack, Gitter, Discord, ... I much prefer Gitter because we are no large company where we need sub-channels and many features. Gitter can be used by everyone with his GitHub account. It supports markdown and has a panel for activities that happen on the repository. Screenshot sharing is as simple as dropping the image there. You can ping people or quote and reply to messages.

Everything else in Gitter is by far not as sophisticated as in Slack for instance, but it is absolutely sufficient IMO for the size fo the JabRef team.

Therefore, the answer to your question is, a chat exists and it only needs to be actively used.

LinusDietz:
https://jabref.slack.com/messages (;

jonasstein:
chat.freenode.net:6697 / non-SSL: 6667 
[irc://freenode/jabref(freenode:#jabref)
You could vote for it with thumbs up and see, which platform is most promising.

tobiasdiez:
> The bug tracker is often used as chat (see #3382 and others), which is obviously not good.

I really don't see a problem here. Most of the discussion was about the bug, how to reproduce it and where to find to find the code related to it. This is all useful information and good to have in clear connection to the issue (i.e. directly below it 😸 ). I do understand that such a discussion generates a bit of noise, but that also applies to a gitter/slack chat.

I added the slack chat to the [contributing FAQ](https://help.jabref.org/en/FAQcontributing) so that new programmers should be able to find it (yes, it is not as prominent as it could be, but should be sufficient until many new-comers break down our doors). I now play the dictator and simply close this issue 😈 .

halirutan:
[...]

PR #3630, opened 2018-01-12
Add gitter badge

koppor:
This addresses https://github.com/JabRef/jabref/issues/3586

In the DevCall, we decided, that we will try to use Gitter and see how it goes.

The main arguments were:
- easy to join for new users
- offer help for new developers since the burden of initial JabRef development is high for newcomers - even if we have https://github.com/JabRef/jabref/wiki/High-Level-Documentation and https://github.com/JabRef/jabref/wiki/Code-Howtos
- Nearly none of the developers uses Slack
- does not clutter the Skype chat of the developers
- developers can use [pinned tabs](https://support.mozilla.org/en-US/kb/pinned-tabs-keep-favorite-websites-open) to keep gitter always opened

We will shutdown gitter in case there won't be much support from @JabRef/developers there (because lack of time). Nevertheless, we give it a chance. 😇

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,6 +6,7 @@ Take a look at GitHub's excellent [help documentation] for a detailed explanatio
 We also have [code howtos](https://github.com/JabRef/jabref/wiki/Code-Howtos) and [guidelines for setting up a local workspace](https://github.com/JabRef/jabref/wiki/Guidelines-for-setting-up-a-local-workspace).
 
 In case you have any question, do not hesitate to write one of our [JabRef developers](https://github.com/orgs/JabRef/teams/developers) an email.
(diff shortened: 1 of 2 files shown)


lenhard:
Since testing gitter has been tentatively decided at the devcall, I'll merge this right away.

tobiasdiez:
The gitter chat is meant as a platform to discuss developer's question and not to give user support, right? In this case, I would suggest to follow the [rails example](https://github.com/rails/rails#contributing) and create a `contributing` section in the readme with the two badges "code helpers" and "gitter" to make the context clear. In this new section, we could also add links to the contributing FAQ and the related wiki pages to make contributing even easier.

lenhard:
Yes, you are right: gitter is for developer's questions. Maybe I was to quick with merging, apologies!

Issue #1750, opened 2016-08-16
Biblatex Export to Word 2007 xml: date field is not exported


sa3016:
JabRef 3.5 windows 10 10.0 amd64 Java 1.8.0_92

Jabref can import and recognize year of biblatex file, but cannot export xml 07 word format with year. 

Steps to reproduce:

1. prepare a biblatex file
2. import it into Jabref
3. export it as 07 word xml

I am asked to create an issue here. (http://discourse.jabref.org/t/export-bib-to-2007-office-xml/142)

Siedlerchr:
Thanks for your report! I tried with a recent 3.6 dev version and could not reproduce the behavior.
So i suggest, you should try again with the latest dev version from http://builds.jabref.org/master/

mlep:
I could reproduce this issue with the master build using the test BibLaTeX file http://mirrors.ctan.org/macros/latex/contrib/biblatex/bibtex/bib/biblatex/biblatex-examples.bib (file suggested in https://github.com/JabRef/jabref/wiki/BibTeX ).
In the xml file, most of entries no not have a year.

Siedlerchr:
@mlep Then I maybe misunderstood sth:  
If a year is set in the  `year `field , then it is in the xml file.  If it has a only a `date ` field set, the date is not exported. If it has both fields set, then the year is exported correctly.

Check the example `"he Importance of Being Earnest: A Trivial Comedy for Serious People`
A possible workaround is to have both fields set.

mlep:
My guess was: if there is information about the year in the BibLaTeX entry (in the field year or in the field date), this information should be found in the xml file.

Currently, because this filter exports the entries without this information, the xml file maybe, in practice, of limited use.

Siedlerchr:
Okay, so I file this as a bug. Because in biblatex we have either date or year field as required. And when only date field is there, then the export does not export any year/date things

oscargus:
Yes, we should add a mapping from date to year if there is no year field available.

Siedlerchr:
I will try to fix that

Siedlerchr:
As part of the reworking in #1765  , I already fixed the exportation of the year.
@sa3016  @mlep  Please try again with the newest version from here:
http://builds.jabref.org/office07/

If you notice any other things which look odd regarding the export, please report.

mlep:
Thanks for fixing the year. It works very well!

Other odd things (since you asked...):
[...]

PR #3632, opened 2018-01-12
 Export translator field to msoffice 2007 xml

Siedlerchr:
Upcoming issue from #1750
While testing this I noticed that the import was broken (due to the recent export filetype/extension changes), fixed that, too.

Directly use Bookauthors and Translator for import from xml
Add test 

----

### DIFF
--- a/src/test/resources/org/jabref/logic/importer/fileformat/MsBibImporterTestTranslator.bib
+++ b/src/test/resources/org/jabref/logic/importer/fileformat/MsBibImporterTestTranslator.bib
@@ -1,0 +1,15 @@
+% Encoding: UTF-8
+
+@Misc{2016,
+  title          = {MeinArtikel},
+  year           = {2016},
+  month          = jul,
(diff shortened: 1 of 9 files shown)


src/test/resources/org/jabref/logic/importer/fileformat/MsBibImporterTestTranslator.bib
```diff
@@ -0,0 +1,15 @@
+% Encoding: UTF-8
+
+@Misc{2016,
+  title          = {MeinArtikel},
```
tobiasdiez: This doesn't look like the bibtex version of the xml file below.

Siedlerchr: Thanks for noticing. But shouldn't travis normally have noticed this?

Siedlerchr:
As the feedback was okay I merge this in

Issue #3573, opened 2017-12-23
Gradle task localizationUpdate does not work


tobiasdiez:
`localizationUpdate` leads to resorting and adding of many language keys due to the recent changes. Thus, at the moment it is not possible to remove a pair from the english file and propagate this change to the other languages.

koppor:
I think, I can work on that in January. Meanwhile, you should be able to log in at crowdin and remove the key there. Need to make you maintainer there, though.

koppor:
Please DO NOT remove obsolete keys in other language files. This is done by crowdin automatically: https://github.com/JabRef/jabref/pull/3585

I will remove the task completely or have it output that there is no need for a check, because of crowdin!

tobiasdiez:
@koppor Right now we have tests in place that require to remove obsolete language keys from other languages. It would be nice if we could remove the `localizationUpdate` step completely (less things to do manually), but then the tests should be adopted accordingly.

koppor:
I thought, I adapted the tests. Seems, I missed something. Have to check...

Am 05.01.2018 12:01 schrieb "Tobias Diez" <notifications@github.com>:

tobiasdiez:
This test fails if you remove language keys from the english file but not from the others:
https://github.com/JabRef/jabref/blob/3b2d35f28f87038b1824bf1178d264cfb1984597/src/test/java/org/jabref/logic/l10n/LocalizationConsistencyTest.java#L50-L51
(If crowdin handles this case automatically, then there is no need for this test, right?)

A few other cases still refer to the `localizationUpdate` (which should not be invoked, right?), e.g.,
https://github.com/JabRef/jabref/blob/3b2d35f28f87038b1824bf1178d264cfb1984597/src/test/java/org/jabref/logic/l10n/LocalizationConsistencyTest.java#L140-L142

koppor:
It is not as straight-forward as I thought.

English escapes more characters than non-English:

    Executing\ command\ \"%0\"...=Executing command \"%0\"...

German:

    Executing\ command\ "%0"...=AusfÃ¼hrung des Kommandos "%0"...

Siedlerchr:
Any updates on this? I think we don't need the script anylonger or?

koppor:
A admin might still need it. Quoting: https://github.com/JabRef/jabref/pull/3635#issuecomment-357869322

[...]

PR #3635, opened 2018-01-15
Update documentation to new workflow using Crowdin

koppor:
Thanks to Crowdin, the workflow to add new keys is simpler.

This addresses https://github.com/JabRef/jabref/issues/3573, but not completely fixes it. More work is required at `syncLang.py`. That script, however, is not absolutely required any more. Crowdin de facto replaced it.

----

- [x] <s>If you changed the localization: Did you run `gradle localizationUpdate`?</s>


koppor:
The current state in `master` was confusing. So, I merged to keep things flowing. https://github.com/JabRef/jabref/issues/3573 is still open to track issues with the `localizationUpdate` task. That task won't be necessary for all contributors, only for maintainers wanting to cleanup the files (like sorting, ...). However, this is not absolutely necessary anymore since Crowdin basically takes care.

Issue #3634, opened 2018-01-13
Incorrect RIS fields


dsifford:
Greetings.

I'm the developer of the Academic Blogger's Toolkit plugin for WordPress and some of my users were having difficulty importing RIS from jabref. Upon further investigation, it looks like you are currently mis-keying some fields. 

https://github.com/JabRef/jabref/blob/723af6dddfb4339e5f5fec3bb38869344201c287/src/main/java/org/jabref/logic/importer/fileformat/RisImporter.java#L147-L152

According to the RIS specification, `A1-A4` are exclusively _author_ fields. Only the `ED` field should be used for editors.

Related: https://github.com/dsifford/academic-bloggers-toolkit/issues/409#issuecomment-357453191

dsifford:
Also, while I have your attention. It appears that your RIS exporter is leaving tons of blank lines (ostensibly where optional fields were skipped).

Here's an example that I just tried:

```txt
TY  - BOOK
AU  - Smith, Bob
AU  - Doe, Jan
AU  - Brown, Judy
A2  - Martin, Steve
A2  - Clark, Joe
T1  - Testing Book Title
PB  - Test Publisher

Y1  - 2015/october
VL  - 1












ER  -
```

lenhard:
Thanks for reporting! This shouldn't be too hard to fix.

Related issue: #2607

dsifford:
Sure thing. Happy to help.

Here's the list of field mappings that I use for RIS files: 

https://github.com/dsifford/astrocite/blob/668a9e4a0cb15a21a310d38e6e3f9ec5af7db9a0/packages/astrocite-ris/src/constants.ts#L62-L107

Also...

"stateful" fields: https://github.com/dsifford/astrocite/blob/668a9e4a0cb15a21a310d38e6e3f9ec5af7db9a0/packages/astrocite-ris/src/parser.ts#L58-L65

and finally, these:

https://github.com/dsifford/astrocite/blob/master/packages/astrocite-ris/src/parser.ts#L90-L109

The above fields include the standardized fields that you'd expect, but also non-standard (but highly used) fields that I've seen other applications export into (e.g., Zotero, Endnote, etc). 

Hope that's useful!

lenhard:
The import-related problems described in this issue should be fixed by #3642. The export-related formatting issues are still present, but since this is about formatting and not functional correctness, I'll reclassify this issue from bug to enhancement.

lenhard:
[...]

Issue #2607, opened 2017-03-03
RIS import ignores L2 ("Full-text link") field


buhtz:
Using JabRef 3.8.1 on Debian unstable.

I exported entries from Zotero as RIS format including the external files (e.g. snapshots). Zotero create a RIS-file and a `file`-directory for the external files.

e.g.
```

TY  - ELEC
TI  - git - Pull new updates from original GitHub repository into forked GitHub repository - Stack Overflow
UR  - http://stackoverflow.com/questions/3903817/pull-new-updates-from-original-github-repository-into-forked-github-repository
Y2  - 2016/08/30/11:53:28
L2  - files/5888/pull-new-updates-from-original-github-repository-into-forked-github-repository.html
```

When importing that into JabRef it looks like that the `L2` field is totaly ignored.
JabRef should be aware of that field and move that file to the setup "General file directory".

The same problem is when using "Biblatex" as export format in Zotero.

Currently this makes JabRef unusable for me because I am not able to import my data from Zotero. No snapshots of websites not PDF files etc

buhtz:
When looking into the code of `RisImproter.java` it looks like that JabRef even doesn't support the real RIS. See #2609.

matthiasgeiger:
Bibtex/biblatex export works fine for me...
All linked files/snapshots are accessible from within JabRef.

tokkot:
I found what I think is a more up-to-date specification for the RIS format in the manual for Reference Manager 12, [here](https://researchsoftware.com/sites/researchsoftware.com/files/product_attachments/v12_refman_userguide_0.pdf), page 581. But Reference Manager has been discontinued in favor of EndNote, so many EndNote defines the spec now.

lenhard:
This should be fixed by #3642

PR #3642, opened 2018-01-17
Extend RIS import with multiple fields

lenhard:
Meeting time is programming time :)

This addresses the import related things described in #3634 #2607

I basically went through the lists linked by @dsifford in the first issue, tried to adapt our import if possible and added custom fields when not. Our importer tests are still running.

----


src/main/java/org/jabref/logic/importer/fileformat/RisImporter.java
```diff
@@ -216,13 +210,39 @@ public ParserResult importDatabase(BufferedReader reader) throws IOException {
...
+                    } else if ("ID".equals(tag)) {
                         fields.put("refid", value);
                     } else if ("M3".equals(tag) || "DO".equals(tag)) {
                         addDoi(fields, value);
                     }
+                    // fields for which there is no direct mapping in the bibtext standard
+                    else if ("AV".equals(tag)) {
+                        fields.put("archive_location", value);
+                    } else if ("C3".equals(tag)) {
+                        fields.put("event", value);
```
tobiasdiez: eventtitle ?

tobiasdiez: eventtitle ?

src/main/java/org/jabref/logic/importer/fileformat/RisImporter.java
```diff
@@ -216,13 +210,39 @@ public ParserResult importDatabase(BufferedReader reader) throws IOException {
...
+                    } else if ("CN".equals(tag) || "VO".equals(tag)) {
+                        fields.put("call-number", value);
+                    } else if ("DB".equals(tag)) {
+                        fields.put("archive", value);
+                    } else if ("N1".equals(tag) || "RN".equals(tag)) {
+                        fields.put("note", value);
```
tobiasdiez: Note is a standard biblatex field and thus should be added to `FieldName`

src/main/java/org/jabref/logic/importer/fileformat/RisImporter.java
```diff
@@ -216,13 +210,39 @@ public ParserResult importDatabase(BufferedReader reader) throws IOException {
...
+                    } else if ("OP".equals(tag)) {
+                        fields.put("original-title", value);
+                    } else if ("RI".equals(tag)) {
+                        fields.put("reviewed-title", value);
+                    } else if ("RP".equals(tag)) {
+                        fields.put("status", value);
+                    } else if ("SE".equals(tag)) {
+                        fields.put("section", value);
+                    } else if ("ST".equals(tag)) {
+                        fields.put("shortTitle", value);
```
tobiasdiez: `shorttitle` is a standard biblatex field

src/main/java/org/jabref/logic/importer/fileformat/RisImporter.java
```diff
@@ -216,13 +210,39 @@ public ParserResult importDatabase(BufferedReader reader) throws IOException {
...
+                    } else if ("C2".equals(tag)) {
+                        fields.put("pubmed-identifier", value);
```
tobiasdiez: I'm not sure if we are already have a `PMID` field. If not, the (biblatex) standard is to specify it in the format
````
eprint = {<pmid>},
eprinttype = {pubmed},
````

File: src/main/java/org/jabref/logic/importer/fileformat/RisImporter.java [hunk 5]
```diff
@@ -216,13 +210,39 @@ public ParserResult importDatabase(BufferedReader reader) throws IOException {
...
+                    } else if ("TA".equals(tag)) {
+                        fields.put("translator", value);
```
tobiasdiez: `translator` is also a standard biblatex field

lenhard:
@tobiasdiez Thanks a lot! I didn't have biblatex in mind when setting the fields. Your comments are all integrated.

src/main/java/org/jabref/logic/importer/fileformat/RisImporter.java
lenhard: fixed

src/main/java/org/jabref/logic/importer/fileformat/RisImporter.java
lenhard: fixed

src/main/java/org/jabref/logic/importer/fileformat/RisImporter.java
lenhard: `shorttitle` wasn't even part of `FieldName` so far. I have added it now.

src/main/java/org/jabref/logic/importer/fileformat/RisImporter.java
lenhard: fixed

src/main/java/org/jabref/logic/importer/fileformat/RisImporter.java
lenhard: fixed

src/main/java/org/jabref/logic/importer/fileformat/RisImporter.java
```diff
@@ -129,21 +123,21 @@ public ParserResult importDatabase(BufferedReader reader) throws IOException {
                         fields.put(FieldName.TITLE, fields.get(FieldName.TITLE).replaceAll("\\s+", " ")); // Normalize whitespaces
                     } else if ("BT".equals(tag)) {
                         fields.put(FieldName.BOOKTITLE, value);
-                    } else if ("T2".equals(tag) && (fields.get(FieldName.JOURNAL) == null || "".equals(fields.get(FieldName.JOURNAL)))) {
+                    } else if (("T2".equals(tag) || "J2".equals(tag) || "JA".equals(tag)) && (fields.get(FieldName.JOURNAL) == null || "".equals(fields.get(FieldName.JOURNAL)))) {
```
Siedlerchr: This would have been optimal to use a switch case, which is very efficient for strings, but that's a bit of personal taste ;)

lenhard: You are right in terms of efficiency, but I really don't like switches. It is just to easy to forget the `break` statement. That doesn't happen if you stick to `if` and always use braces.

lenhard:
Is the `RISImporterTestFiles` class not executed during the travis build? Because those tests were actually partly broken. Anyway, I adapted them to the new import structure (which changes a few fields) and I added the RIS code from  #3634 as a new test case.

Siedlerchr:
Regarding the test, seems to be a general problem with the parameterized tests and the new junit ä.
I noticed the same with the office XML tests

lenhard:
Ok, then this is something that really needs fixing. I'll open an issue for it. Otherwise this PR is ready to merge from my point of view.

lenhard:
In the spirit of getting things done and because all reviewer comments have been addressed, I'll merge this now.

PR #3655, opened 2018-01-21
[WIP] Django style unchained.

tobiasdiez:
Code examples are better than words. So, let them speak. Instead of the current
````java
super.getFoo()
        .foo()
        .getBar()
        .bar();
````
we now get the nicely aligned
````java
super.getFoo()
     .foo()
     .getBar()
     .bar();
````

----

### DIFF
--- a/config/IntelliJ Code Style.xml
+++ b/config/IntelliJ Code Style.xml
@@ -31,6 +31,7 @@
     <option name="KEEP_BLANK_LINES_IN_DECLARATIONS" value="1" />
     <option name="KEEP_BLANK_LINES_IN_CODE" value="1" />
     <option name="KEEP_BLANK_LINES_BEFORE_RBRACE" value="0" />
+    <option name="ALIGN_MULTILINE_CHAINED_METHODS" value="true" />
     <arrangement>
       <rules>
         <section>


lenhard:
Code examples may be better than words, but only if they are really really obvious ;-) To be honest, I don't get why the new style is strictly better. I have nothing against it, but I also don't see any obvious benefit. To me, it just doesn't matter.

Could you convince me with words?

tobiasdiez:
Well, there are of course no real benefits. We will not perform 200% more efficient because of this change. It's just a matter of taste and for me fluent code (with builders or the stream api) does not look "nice". But of course this is a subjective opinion. For example, why is `map` indented (with respect to `getValues`) but `collect` is on the same level as `map`?
````java
return field.getValues().stream()
                .map(SpecialFieldValueViewModel::new)
````

lenhard:
Ok, I am buying into that :-) Could we do the same configuration for Eclipse here as well (as requested by @Siedlerchr ). Just to ensure that there is no constant reformatting of the code with different IDEs.

Siedlerchr:
https://stackoverflow.com/questions/6275785/wrapping-chained-method-calls-on-a-separate-line-in-eclipse-for-java
will try to check that the next days

PR #3658, opened 2018-01-22
Rename the Review Tab into Comments Tab

LinusDietz:
After a discussion with colleagues, it became apparent, that the Review Tab is maybe named a bit misleadingly. After all I know it is meant for comments about this paper, and I guess few really use it for actual peer-reviews.

~To keep consistency, I have just changed the wording.~

I wrote a migration and handled conflict cases by asking the user.

----

- [X] Test created


LinusDietz:
@tobiasdiez I did not change the field name to ensure backwards compatibility. (otherwise we would have to write a migration)
Is this a good  (enough) argument?

Second, the Tab is not an own class, but a Custom Tab, initialized here: https://github.com/JabRef/jabref/blob/cf8e6aac8263ce30eb2b042c8aacf2d88bbac4c8/src/main/java/org/jabref/preferences/JabRefPreferences.java#L978

tobiasdiez:
Yeah, backwards compatibility is a good point but on the other hand I also find it confusing that the content of the "comments" tab is stored in the field `review`. I don't know how to proceed. Maybe we should indeed write a migration...

LinusDietz:
Well, I just realized that the Comments field is also displayed in the General tab. :/ Thus, is might be that users have both the Review field and the Comments field filled.
Anyways, this makes it clear that the solution as it is right now cannot be used and we definitely have to write a migration if we want to rename this tab.

Proposed migration: Concatenate the Review to the Comment field, afterwards delete the Review field.
The Comment field should go from the General Tab to the Comments tab.

LinusDietz:
So as discussed this PR does the following:
- [x] Rename the Review Tab into Comments Tab
- [x] Remove the Comment Field from the General Tab
- [x] Migrate the Review Field into Comment according to https://github.com/JabRef/jabref/pull/3658#issuecomment-359502131

src/test/java/org/jabref/logic/importer/util/RenameReviewToCommentTest.java
```diff
@@ -0,0 +1,86 @@
...
+        actualEntry.setField(FieldName.COMMENT, "My Comment");
+
+        ParserResult actualParserResult = new ParserResult(Collections.singletonList(actualEntry));
+
+        BibEntry expectedEntry = createMinimalBibEntry();
+        expectedEntry.setField(FieldName.COMMENT, "My Comment\nReview:\nMy Review");
+
+        action.performAction(actualParserResult);
+
+        assertEquals(new ParserResult(Collections.singletonList(expectedEntry)).getDatabase().getEntryByKey("Entry1"),
```
tobiasdiez: why not just `assertEquals(expectedEntry, actualParserResult.getDatabase().getEntryByKey("Entry1"))`?

LinusDietz: good remark! (:

src/main/java/org/jabref/logic/importer/util/RenameReviewToComment.java
```diff
@@ -0,0 +1,39 @@
+package org.jabref.logic.importer.util;
```
lenhard: Could you move the complete class into `org.jabref.migrations`? Because that's what it actually does, it migrates an opened bib file to a newer version of JabRef. I think it makes sense to bundle all this stuff in a dedicated package, because then it gets easier for us to see what sort of conversions are performed on open.

While you're doing that, could you also move the `ConvertLegacyExplicitGroups`?

LinusDietz: Done. I also renamed these actions to what they are: migrations

src/test/java/org/jabref/migrations/RenameReviewToCommentTest.java
```diff
@@ -0,0 +1,82 @@
...
+        ParserResult actualParserResult = new ParserResult(Collections.singletonList(entry));
+
+        action.performMigration(actualParserResult);
+
+        assertEquals(entry, actualParserResult.getDatabase().getEntryByKey("Entry1").get());
+    }
+
+
+    @Test
+    public void reviewAndCommentField() {
```
koppor: I am not sure whether we should keep the two fields.

Maybe, I am storing the comments of reviewers in the `review` field? -- OK, then I should put it into the `reviews` (with `s`) field... 

I would let the user do it manually then. We have the "Other fields" tab, where "Review" will show up then, don't we?

koppor: Fine with all of it. I think, as usual, I am discussing an edge-case here. JabRef never merged field contents in its entire history. As far as I remember, I cannot disable this single migration, can I? If not, JabRef changes data in a more hard way as it did in the past.

Since no other software (hopefully!) does not use the "review" field, it is OK for me to merge it into the comment field. Let's hope for the best... In case there is software relying on this field, JabRef will remove it... OK, the user can always skip the migration...

LinusDietz: Actually, with the current implementation the user cannot skip the migration. It will be executed each times any database is opened!

To clarify:
Goal: Have a prominent position for users to make comments on their entries.

Opinionated observation: The Review Tab is less important than a Comments tab, a review is basically also a comment on the entry.

Proposed change: Rename the Review Tab to Comments Tab, consolidate fields. If both fields present merge.

Yes, this is a hard change, but what's the alternative? Keeping everything as it is/Close PR? Add another Tab? Swap the two fields? Make the migration optional? Let user choose the migration strategy? Discuss at devcall?

lenhard: How about just popping up a dialog IF a review field is found in the database? Just a dialog that informs the user that the review field will be appended to the comment field and removed from the entries. If the user does not want this, she should not save the file.

Good compromise imho and it informs the users what is actually going on.

koppor: I was only discussing the impact if **both** fields are present. The other silent migration (if only `review` is there) was OK for me. I like, however, the idea of @lenhard. 

I would like to have that dialog consistent with the old one at "Upgrade file"

grafik

tobiasdiez: I think a simple "the review field has been merged with the comments field" information is enough. If the user does not like our decision, then he is free to use other software :-)

koppor: @tobiasdiez And even the information should not be a dialog, but some kind of notification:

grafik

Maybe, this is is a task for #CloudRef. I sometimes have the feeling, we recode a UI framework in JabRef. Would it have been easier (in the long run) if we have used IntelliJ's UI as basis? Or some NetBeans UI designer (recommended by https://www.b-tu.de/fg-software-systemtechnik/team/mitarbeiter/markus-uhlig) or even the modern Eclipse thing? OK, JavaFX is way more modern than these :)

tobiasdiez: or just use controlsfx as we do: https://controlsfx.bitbucket.io/org/controlsfx/control/NotificationPane.html or https://controlsfx.bitbucket.io/org/controlsfx/control/Notifications.html

LinusDietz: @tobiasdiez The problem is the following: This is called before the JabRefFrame is created (where I would put the code for the Notification).
So where should I create such a Notification? Just creating an Object anywhere doesn't work, since just writing
```
Platform.runLater(() ->Notifications.create().title("Title Text").text("Hello World 0!"));
```
does not display anything.

tobiasdiez: Yeah, I think for the moment it best to just show a normal dialog. With the move to a javafx frame we should be able to use notifications in the future...

LinusDietz: Okay did that. You can even reject the merge now... however this will pop up again the next time one opens an unmigrated database.

LinusDietz:
I think this PR is quite ready now

LinusDietz:
@lenhard I have incorporated your suggestions. The solution I used for remembering the change in the database was to add a field in the ParserResult class and check that when instantiating the BasePanel.

src/main/java/org/jabref/logic/importer/ParserResult.java
```diff
@@ -19,6 +19,7 @@
 import org.jabref.model.metadata.MetaData;
 
 public class ParserResult {
+    public boolean wasChangedOnMigration = false;
```
lenhard: Sorry, but there is no way I am going to sign off a public instance variable such as this. Please follow the Java Conventions and use a getter and a setter.

src/main/java/org/jabref/migrations/MergeReviewIntoComment.java
```diff
@@ -0,0 +1,70 @@
...
+import org.jabref.gui.dialogs.MergeReviewIntoCommentUIManager;
+import org.jabref.logic.importer.ParserResult;
+import org.jabref.logic.l10n.Localization;
+import org.jabref.model.entry.BibEntry;
+import org.jabref.model.entry.FieldName;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class MergeReviewIntoComment implements PostOpenMigration {
```
tobiasdiez: I think most of this code should move to the logic package.

src/main/java/org/jabref/logic/importer/ParserResult.java
LinusDietz: Yeah I though somebody might complain I just find it so stupid... The `protection' is just the same with getters and setters...

lenhard: I know a new book that's about to come out that argues for keeping to conventions when programming Java ;-) Maintainability and all.

src/main/java/org/jabref/migrations/MergeReviewIntoComment.java
lenhard: I told him to move it there, because I want to have the migrations together. The migrations package should probably go into logic, but some of the migration classes are awefully coupled to UI classes.

tobiasdiez: ok, then the location is fine. But at least the new class should be completely decoupled in logic and gui since otherwise the migration of the migration package is getting harder with this PR.

LinusDietz: @tobiasdiez I guess you mean this line?
https://github.com/JabRef/jabref/blob/b246cfe89dcd97c229dc7b2c9cc79769d6227e32/src/main/java/org/jabref/migrations/MergeReviewIntoComment.java#L38

I can create another Class for this confirmation, but where would you put that?

src/main/java/org/jabref/logic/importer/ParserResult.java
Siedlerchr: The "protection" is not the same. Getters and setters encapsulate the variable and might perform (in future) addtional operations on them. That's what I liked about the automatic properties of C#.  Automatically generated getters and setters.

src/main/java/org/jabref/migrations/MergeReviewIntoComment.java
tobiasdiez: I have no deep knowledge about the migration code, so the following may be a bit naïve. For me, the expected flow is something like:
- You have a general migration manager class (in gui), which is getting called as soon as the user opens a database
- This manager has a list of possible migrations which are defined in logic and which provide two methods: one for determining if an action is needed and a second one to actually perform the action
- The manager first calls `actionNeeded` and maybe asks the user for confirmation, then runs `performAction` on the migration action.

Not sure how much of this architecture is already in place and how much you are willing to refactor for this PR. My initial comment was triggered by the fact that a new class was created that mixed UI and logic.

LinusDietz: Okay, I totally support your point, however this is all a bit tricky, as these PostOpenMigrations are all called before the main GUI is instatiated.

Suggestion: Lets merge this PR now, and I'll refactor the code in a sensible way in a followup PR. (I'm on the train tomorrow for at least 5 hours)

tobiasdiez: That's fine with me!

LinusDietz:
Thanks for all the feedback! I'll merge this now and start with the refactoring in the next PR.

src/test/java/org/jabref/migrations/RenameReviewToCommentTest.java
TarryDan: > I think a simple "the review field has been merged with the comments field" information is enough. If the user does not like our decision, then he is free to use other software :-)

Fundamentally, I don't see the problem with a spare field a user can choose to use or to ignore; but for users who have been using both fields for different types of data, this is a huge inconvenience. 

(either to manually edit every entry so the fields remain separate; or move to a new software; or keep an older version of jabref running)

As well as knocking my faith in the program going forwards (what fields can I rely on in the future?). 

& especially because a database manager isn't like a word processor (i.e.) where you're constantly creating new files, so you can have some .DOCs and some .DOCXs and it's not a big deal. A database is a longterm object, it needs to remain predictable over its lifespan and archive data reliably. Updating the software shouldn't lead to data loss (which is essentially what this is - I had data recorded in a field, that field no longer exists, and the data is no longer filed effectively).

The appeal of the separate Comment and Review fields *is* that it's purpose is uncertain - I can't appropriate the author field, say, for a new use because it's too likely that the author field will be useful for its original purpose. But a field with no clear function is extremely useful, as each user will likely have some unique need database designers cannot predict. The concept of a Review (specific data reviewing the entry in some way) and the concept of a Comment (any extra miscellaneous data to record with the entry) is distinct. 

To me this feels like removing a feature needlessly, without checking how people are using it.

(I've been using Jabref for over 6 years and only just updated my laptop, so I'm only now trying the new version - but that's a lot of data and wasted time, or difficulty and risk to keep an old laptop chugging along, or try and run an older version, or - I don't even know. The current users are the people who think the program is already good and have invested trust and time into it; and trustworthiness is an essential quality of a database, moreso than other kinds of program.)

Siedlerchr: You can use the custom entry types dialog to fully customize your entries (also reflected in the entry editor) and add a review field yourself.

ThiloteE: @TarryDan Sorry, this issue is causing you inconvenience.  What you could do to work around it for now is to open your (old) database with a [text editor](https://alternativeto.net/software/notepad-plus-plus/?license=opensource) and use the "find and replace" functionality to rename the review fields into a different field of your choice. In recent versions of JabRef, it is also possible to conveniently transfer field content to different fields via the automatic field editor (Edit > Automatic field editor > copy or move content), so if there ever were to happen another migration in future, it would be much easier.

koppor:
@TarryDan Interestingly, your comment and the follow-up are not shown in my web browsers. 

Thank you for your long comment and explanation. Especially the "faith" and trust of JabRef not altering data without asking. Back then, the group of developers of JabRef consistet of more progressive developers Now, the more thoughtful voices have more shares.

---

The workaround by @Siedlerchr DOES NOT WORK. We need to fix code @Siedlerchr 

The workaround by @Siedlerchr is as follows: Close the library, revert all changes. Start JabRef. File -> Preferences -> Entry types. Select "Article". Add "Review" as field:

image

Mark it as optional (AKA non-required) and multiline

image

Click on "Save"

Repeat for **all other** types.

**Restart  JabRef**

Then, the field appears again:

image

Save the library.

Close the library.

Open the library again.

JabRef displays **no warning**

image

"Review" moved to comment:

image

This is a BUG!

@TarryDan Thank you for raising this.  We will discuss internally how to proceed with that and then create a public issue. -- We have "preference migration", where it is IMHO "more OK" to more around data. For libraries it is "more difficult" (to unacceptable in my opinion).

src/main/java/org/jabref/migrations/MergeReviewIntoComment.java
koppor: The follow-up was done at https://github.com/JabRef/jabref/pull/3733

PR #3659, opened 2018-01-23
Jacoco toolversion

LinusDietz:
Try to get the failing `checkOutdatedDependencies` into green again.


koppor:
All tests are green. Thus, this PR is good to go? Is something left?

Siedlerchr:
The dependeny update still fails

LinusDietz:
@Siedlerchr  And I don't know why... These native gradle plugins are quite hard to configure. Locally, the gradle task even succeeds... :/

Siedlerchr:
Can't we just add an exception rule for this jacoo stuff?

LinusDietz:
yeap, but is that a solution? (;

tobiasdiez:
If I understand https://github.com/ben-manes/gradle-versions-plugin/issues/92 correctly, then the jacoco dependency should actually be hidden by default (but is shown sometimes, which is a bug). Thus I would vote to simply hide the update (it's not an important dependency anyway).

LinusDietz:
Current Status: No Dependencies found, since the `[gradle-versions-plugin](https://github.com/ben-manes/gradle-versions-plugin)` only provides support until gradle 4.3.

koppor:
We can go back to gradle 4.3. I don't see any advanced features. We should, however, document it, why we don't keep our build system updated ^^.

Siedlerchr:
Create an issue at their repo.

koppor:
I don't know how to test an self-made update of the script. When going to gradle 4.3, it was easy: https://github.com/ben-manes/gradle-versions-plugin/commit/e39a84307cea47892149f576461c80fdb3b95d2f

tobiasdiez:
We merge this now and will revisit the jacoco dependency at a later point.

PR #3661, opened 2018-01-23
Export no empty lines in RIS format

lenhard:
This fixes the remaining part of #3634

I had a closer look at the export / layout logic and wow, this is arcane! The code is more than 12 years old and almost as bad as the bibtex parser used to be.

This PR is based on two decisions:
 - I do not want to change the old layouting code. If I do, I am almost guaranteed to break it in many ways.
- I do not know if the removal of empty lines in exported text is really desirable in all cases. Maybe users wrote their custom formatters to explicitly include empty lines? I cannot tell.

We now have one case (RIS export) where empty lines are not desired. So, I implemented a solution that only changes the behavior of the RIS export, but leaves all others untouched. Nonetheless, it is very easy (single constructor parameter) to let other formats use empty line elimination as well. 

The empty line elimination is done using a hardly understandable, but fully functional regex.

----

### DIFF
--- a/src/main/java/org/jabref/logic/exporter/TemplateExporter.java
+++ b/src/main/java/org/jabref/logic/exporter/TemplateExporter.java
@@ -14,6 +14,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.regex.Pattern;
 
 import org.jabref.JabRefMain;
 import org.jabref.logic.layout.Layout;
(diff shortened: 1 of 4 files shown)


src/main/java/org/jabref/logic/exporter/TemplateExporter.java
```diff
@@ -257,7 +277,11 @@ public void export(final BibDatabaseContext databaseContext, final Path file,
 
                 // Write the entry
                 if (layout != null) {
-                    ps.write(layout.doLayout(entry, databaseContext.getDatabase()));
+                    if (deleteBlankLines) {
+                        ps.write(layout.doLayout(entry, databaseContext.getDatabase()).replaceAll("(?m)^\\s", ""));
```
Siedlerchr: Please extract the regex to Pattern. Compile. You  can then use it with
.matcher(structure).replaceAll(replace) 
And I would suggest a small comment explaining the regex

lenhard: No problem, done! Although I hardly suspect that this could become a performance problem due to frequent compilation. After all, this would be executed once per RIS entry and even a few thousand compilations probably won't make a noticeable difference.

But well, it surely doesn't hurt to compile only once.

Issue #3645, opened 2018-01-18
Parameterized tests are not executed on travis


lenhard:
Based on this discussion: https://github.com/JabRef/jabref/pull/3642#issuecomment-358591687

It seems that the move to JUnit5 has somehow turned off the parameterized tests on travis. That's bad, because it means that our importer tests are basically gone. We need a way to re-enable them, most likely by adapting them to JUnit5 syntax.

Siedlerchr:
If I execute them in Eclipse with junit5, they are listed as a single test

Siedlerchr:
For the migration to juit5. Refs https://github.com/junit-team/junit5/issues/1247

stefan-kolb:
Are these tests running now? PR is merged.

Siedlerchr:
Most of them. We still have a few left  especially those with Temporary folders. And not all of them are executed. 
However, an official support for them is currently planned for the next milestone (see linked issue above)

PR #3665, opened 2018-01-25
Migrate importer tests to JUnit5

lenhard:
In the course of https://github.com/JabRef/jabref/pull/3642#issuecomment-358581869 we realized that our current (JUnit4) parameterized tests are not executed on the (JUnit5) travis build. This is tracked here: #3645 I think this is a huge problem, because we are missing out on a lot of validation here. 

This PR is a migration of the paramterized RIS tests from JUnit 4 to JUnit 5. I'd like to put this forward to discussion, so that we can agree on how our parameterized JUnit 5 tests are to look like. When we have agreed, we can migrate all tests. So, please look at the code and write if you are happy with this.

[Note that parameters cannot be resolved for lifecycle methods](http://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests-lifecycle-interop) (`@BeforeEach`) in JUnit 5, hence the structure in the PR.

----


lenhard:
@tobiasdiez Thanks a lot for the feedback!

 - Regarding test arguments: I don't really see the point in splitting the single argument into two. We have used the convention that the `.bib` and `.importerFormat` files have the same name and why not keep leveraging that? Also, I find it better to do the path conversion in one place only, but if we want to pass a stream of paths we always have to do that beforehand
 - Regarding unification of importer tests: I have refactored around a little and arrived at a solution that should work for all importers. I quite like how that condenses the `RISImporterTestFiles` class.

Fun fact: I may have used JUnit 5 paramterization, but so far the assertions were still JUnit 4. I'll change that in the following.

src/test/java/org/jabref/logic/bibtex/BibEntryAssert.java
```diff
@@ -35,9 +35,9 @@
      */
     public static void assertEquals(Class<?> clazz, String resourceName, BibEntry entry)
             throws IOException {
-        Assert.assertNotNull(clazz);
-        Assert.assertNotNull(resourceName);
-        Assert.assertNotNull(entry);
+        Assertions.assertNotNull(clazz);
```
Siedlerchr: I would suggest, using a static import as before makes it more readable

src/test/java/org/jabref/logic/importer/fileformat/RISImporterTestFiles.java
```diff
@@ -1,58 +1,28 @@
...
-    public void setUp() throws URISyntaxException {
-        risImporter = new RisImporter();
-        risFile = Paths.get(RISImporterTest.class.getResource(fileName + ".ris").toURI());
-    }
-
-    @Parameters(name = "{0}")
-    public static Collection<String> fileNames() {
-        return Arrays.asList("RisImporterTest1", "RisImporterTest3", "RisImporterTest4a", "RisImporterTest4b",
+    private static Stream<String> fileNames() {
+        return Stream.of("RisImporterTest1", "RisImporterTest3", "RisImporterTest4a", "RisImporterTest4b",
```
Siedlerchr: It would be nice to have the RIS Importer collect the files dynamically like the other file tests do, e.g. starts with RIS and contains Test

lenhard: I will have a look at that and I will also refactor the remainder of the importer tests into this structure.

src/test/java/org/jabref/logic/bibtex/BibEntryAssert.java
lenhard: Done.

src/test/java/org/jabref/logic/importer/fileformat/RISImporterTestFiles.java
lenhard: I have extracted the relevant logic now for the RIS tests. However, it turns out that *the file reading logic in all importer tests is broken* (except for the RIS tests now). It just does not work, no files are found and in that case JUnit 4 didn't complain. JUnit 5 complains that there are no parameters. 

I am not sure if the importer tests never worked and people where just happy that they were green, or if they worked and got broken at some point in time. I'll gradually try to get the tests working now.

Siedlerchr:
The importer tests where running for each file, but that seems to have been ignored in the migration process by the junit 5 comp.
I looked at this, too but didn't find a way to get it to work again.

lenhard:
Yes, indeed. I'll gradually go through all parameterized importer tests and try to get them working in JUnit 5 syntax again as part of this PR.

lenhard:
Turns out, quite a lot of tests for different importers (BibTexML, Medline tested so far) actually fail and this is down to the implementations of the importers. Seems like these tests have never been executed before...

I am not sure what to do about this here. The point of this PR was a refactoring of the testing code, not a repair of broken importers.

Siedlerchr:
I would merge this if all tests are at least executed. Or better you merge this in a new branch where we will fix the importer then

src/test/java/org/jabref/logic/importer/fileformat/ImporterTestEngine.java
```diff
@@ -0,0 +1,56 @@
...
+
+import org.jabref.logic.bibtex.BibEntryAssert;
+import org.jabref.logic.importer.Importer;
+import org.jabref.model.entry.BibEntry;
+
+import org.junit.jupiter.api.Assertions;
+
+public class ImporterTestEngine {
+
+    private static final String TEST_RESOURCES = "src/test/resources/org/jabref/logic/importer/fileformat";
```
Siedlerchr: Maybe better to use Paths.get(ImporterTestEngine.class).getResource("").toUri())

Siedlerchr: I see you have the getPath already

Siedlerchr:
I converted and fixed some fetcher tests in another PR. I found that the GVK Test would also be a goo candidate for using the same schema as here

lenhard:
I have now migrated the complete importer tests to JUnit5 syntax.

Sadly, they are in a dire state. A lot of them are actually failing and seemed to have never been executed anyway. And even except from the refactorings that I have already done, there as a lot of things that can still be improved.

I think it's best if we take the current (failing) state and try to repair each importer and its tests here in the PR. This might be some work, but the importer tests cannot really continue as broken as they are now. Help on this is very much appreciated and should be possible without conflicts.

Siedlerchr:
You can ignore the failing fetcher test, this is already fixed. But I wonder why there is now an NPE with the Method Source

lenhard:
I should have more meetings, so that I get more programming done :)

I think I have repair all importers and their tests now. I'll see what travis says and I'll also do one more round of refactoring on the code (not now). When I am done, this can hopefully be merged.

Siedlerchr:
LGTM!, if you merge master in your branch you will get the fetcher test fixed as well

lenhard:
Ok, this can be merged from my point of view.

Although this is already a huge refactoring of the existing tests, there's certainly more that can be improved. I have a few things in mind, but I would rather keep them out of this already big PR. I'll tackle them later when this is merged.

lenhard:
That looks weird indeed. Looking at the first one: https://codecov.io/gh/JabRef/jabref/src/migrate-parameter-tests/src/main/java/org/jabref/logic/importer/fileformat/BibTeXMLImporter.java the `importDatabase` method is recorded as not covered, but when I execute the tests in Intellij there are twenty test files for that method alone. 

The key seems to be that the missing methods are the ones that use JUnit5 parameterized tests. It seems that Codecov does not record the coverage of these tests, but they are the core tests of the whole package.

But maybe the problem is still travis? How do I see from our travis output that the tests are really executed? I can push a failing test deliberately, but there surely is a better way?

tobiasdiez:
Since the travis log contains statement like 
>13:43:57.686 [main] WARN  org.jabref.logic.msbib.MSBibDatabase - Could not parse document

I suppose the tests are correctly invoked (in fact, the travis log exploded even more). Probably there is a problem with the gradle task `jacocoJunit5TestReport` (but there is no need to fix this in this PR...)

Siedlerchr:
I would merge this in and I think tthere might be a solution for the reports:
https://github.com/junit-team/junit5/issues/1024

PR #3666, opened 2018-01-26
Bring back the context menu

tobiasdiez:
before | after
------------ | -------------
before | image

Removed a few commands that are not needed often and thus don't need to be placed that prominently.
- Print entry preview: available through entry preview
- All commands related to marking: marking is not yet reimplemented
- Set/clear/append/rename fields: available through Edit menu
- Manage keywords: available through Edit menu
- Copy linked files to folder:  available through File menu

Changes in the architecture:
- Previously, an `Action` defined both the visual properties (text, icons, ...) as well as the behavior. These actions are usually defined in the base panel (or main frame).
- Now the visual elements are defined by an `ActionFX` (to be renamed to `Action` as soon as the old class can be removed) and the behavior by implementing the `Command` interface. This separation makes it possible to reuse the visual aspects without the ties to the behavior.
- In the long term, the commands should be defined outside of the main big classes. This is not done in this PR, because then the scope and size would explode. The extraction is best done peu à peu.

----


src/main/java/org/jabref/gui/actions/OldCommandWrapper.java
```diff
@@ -0,0 +1,50 @@
...
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+/**
+ * This wraps the old Swing commands so that they fit into the new infrastructure.
+ * In the long term, this class should be removed.
+ */
+public class OldCommandWrapper extends CommandBase {
+
+    private static final Log LOGGER = LogFactory.getLog(OldCommandWrapper.class);
```
Siedlerchr: Please use the new logging     private static final Logger LOGGER = LoggerFactory.getLogger(OldCommandWrapper.class);

Siedlerchr:
> Copy linked files to folder: moved to Tools menu
Already presend in the File menu
Open File PDF icon is missing

src/main/java/org/jabref/gui/actions/OldCommandWrapper.java
tobiasdiez: Your PR is not yet merged into `javafxTable` and I prefer to do this as soon as all the other maintable PRs are merged

tobiasdiez:
@Siedlerchr Thank you for the feedback. I implemented your suggestions accordingly.

src/main/java/org/jabref/gui/actions/OldCommandWrapper.java
Siedlerchr: Okay, just as a hint to make work then easier. Slf4j has a nice migrator tool (inlcuded in the zip download) https://www.slf4j.org/migrator.html to replace all occurences. Used that for the migration, too.

koppor:
Can't we have an "Edit" submenu here containing all the edit actions available?

For me, this conforms to "Lernförderlichkeit" and "Erwartungskonformität" from [EN ISO 9241-110 Grundsätze der Dialoggestaltung](https://de.wikipedia.org/wiki/EN_ISO_9241#EN_ISO_9241-110_Grunds%C3%A4tze_der_Dialoggestaltung). Especially the "Erwawrungskonformität", because I expect that all actions available for the current entry are available in the context menu.

PR #3674, opened 2018-01-29
Fix 2633 saving creates new database

### LINKED ISSUE #2633: Saving creates new database file instead of overwriting the old one
It seems that the __new version 3.8.2__ changes the saving manner. The old versions just overwrite the original file. I guess the new version rename the orginal file to the backup file and write a new database file.

This causes trouble for me (maybe also for others). I use hard links to point to the same database file in different folders so that the database can be updated automatically.

Currently the orginal file is either deleted or renamed. Hence the link is broken.

Will the developers consider to change it back to original saving manner?

ethanwharris:
Fix #2633 using the method suggested by @koppor. Switching from Files.copy to Files.write now simply overwrites the content of the file, leaving any hard links intact.


ethanwharris:
@lenhard Thanks for your review. Have performed a merge so hopefully that should be fixed now.

tobiasdiez:
The build still fails but so does the master (with the same failing tests). I thus merge this PR now. Thanks for your contribution!

PR #3681, opened 2018-01-31
Reimplement copy and paste of entries

tobiasdiez:
----

### DIFF
--- a/src/main/java/org/jabref/Globals.java
+++ b/src/main/java/org/jabref/Globals.java
@@ -4,6 +4,7 @@ import java.awt.Toolkit;
 import java.util.Optional;
 import java.util.UUID;
 
+import org.jabref.gui.ClipBoardManager;
 import org.jabref.gui.GlobalFocusListener;
 import org.jabref.gui.StateManager;
 import org.jabref.gui.keyboard.KeyBindingRepository;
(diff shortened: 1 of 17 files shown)


lenhard:
I have been playing around with a running JabRef. Pasting entries from external sources worked, cutting and pasting from the table as well. I was doing this for a while:
 - Cut entry (CTRL+X) from the table
 - Click to some position in the table
 - Hit paste (CTRL+V)

Then suddenly, all content in my table disappeared and I was unable to make it re-appear again. I was able to paste (which opened an entry editor with the pasted data), but nothing was shown in the main table. I could continue pasting and watch as the entry number increases in the group panel, but nothing got shown in the main table. Switching between databases did not help. Only closing and reopening the broken database helped. Then I could continue to just hit Cut+Paste, Cut+Paste, Cut+Paste, until eventually all content disappeared from the main table again and it said "No content in table" again. Unfortunately, there is nothing in the log to report.

tobiasdiez:
Thanks for the feedback.
@lenhard I followed your steps and eventually also had an empty table. The reason, however, is not a bug with the pasting/cutting code, but a focus problem: at some point the search bar steals the focus and thus you paste the bibtex code into the search bar. This filters the maintable and finds no entry.

I'll add this focus problem to the list of open problems and hope it resolves itself when the search bar is fully migrated to JavaFX.

PR #3685, opened 2018-02-02
Fixed typo, mayor to major

spogatetz8386:
I was reading through the file and I noticed a simple typo.


Siedlerchr:
Thanks for spotting the typo! Indeed we don't elect a mayor  when we are talking about major code changes  ;)

PR #3690, opened 2018-02-03
[WIP] File link deletion dialog improvements

### LINKED ISSUE #3679: Delete from disk should not be default option/selected on DELETE in file tab
image

Changing the buttons should be sufficient.

bdach:
This pull request fixes the following issues:

* When a file link in a bibliography entry is deleted by the user, a dialog is displayed. Previously, the default (first) dialog option deleted the linked file from disk. The default behaviour has been changed to just removing the file link from the entry.

  This is a fix to issue #3679 
Fixes #3679 

* When clicking the "Cancel" button in the file link deletion dialog, the file  was being deleted from the list. This behaviour has been fixed by changing the return value of the `delete()` method when the "Cancel" button is chosen.

## Example screenshot:

2018-02-04-001057_477x142_scrot

## Checklist:

- [x] Screenshots added

---

I am unsure whether the remaining checklist items are applicable to this pull request.
I would be willing to provide tests for the `delete()` method; however, the `DialogService` member cannot be stubbed-in/mocked as it is (as it is marked `final`), and therefore I can't control the returned chosen option from inside the unit test.

I would appreciate any pointers and suggestions.

### DIFF
--- a/src/main/java/org/jabref/gui/fieldeditors/LinkedFileViewModel.java
+++ b/src/main/java/org/jabref/gui/fieldeditors/LinkedFileViewModel.java
@@ -281,7 +281,7 @@ public class LinkedFileViewModel extends AbstractViewModel {
             Optional<ButtonType> buttonType = dialogService.showCustomButtonDialogAndWait(AlertType.INFORMATION,
                     Localization.lang("Delete '%0'", file.get().toString()),
                     Localization.lang("Delete the selected file permanently from disk, or just remove the file from the entry? Pressing Delete will delete the file permanently from disk."),
-                    deleteFromEntry, removeFromEntry, ButtonType.CANCEL);
+                    removeFromEntry, deleteFromEntry, ButtonType.CANCEL);
 
             if (buttonType.isPresent()) {
                 if (buttonType.get().equals(removeFromEntry)) {
--- a/src/main/java/org/jabref/gui/fieldeditors/LinkedFileViewModel.java
+++ b/src/main/java/org/jabref/gui/fieldeditors/LinkedFileViewModel.java
@@ -304,7 +304,7 @@ public class LinkedFileViewModel extends AbstractViewModel {
         } else {
             LOGGER.warn("Could not find file " + linkedFile.getLink());
         }
-        return true;
+        return false;
     }
 
     public void edit() {
--- a/src/main/java/org/jabref/gui/fieldeditors/LinkedFileViewModel.java
+++ b/src/main/java/org/jabref/gui/fieldeditors/LinkedFileViewModel.java
@@ -44,6 +44,8 @@ import de.jensd.fx.glyphs.materialdesignicons.MaterialDesignIcon;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static javafx.scene.control.ButtonBar.ButtonData;
+
 public class LinkedFileViewModel extends AbstractViewModel {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(LinkedFileViewModel.class);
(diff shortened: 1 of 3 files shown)


src/main/java/org/jabref/gui/fieldeditors/LinkedFileViewModel.java
```diff
@@ -281,7 +281,7 @@ public boolean delete() {
             Optional<ButtonType> buttonType = dialogService.showCustomButtonDialogAndWait(AlertType.INFORMATION,
                     Localization.lang("Delete '%0'", file.get().toString()),
                     Localization.lang("Delete the selected file permanently from disk, or just remove the file from the entry? Pressing Delete will delete the file permanently from disk."),
-                    deleteFromEntry, removeFromEntry, ButtonType.CANCEL);
+                    removeFromEntry, deleteFromEntry, ButtonType.CANCEL);
```
tobiasdiez: This change is fine, but could you please add [ButtonData.Yes](https://docs.oracle.com/javase/8/javafx/api/javafx/scene/control/ButtonBar.ButtonData.html#YES) as a second constructor argument for the `removeFromEntry` button. With this the buttons are automatically ordered correctly according to the current OS specifications.

bdach: Sure, not a problem. Would you like to have the change rebased into the first commit, or is a new one on top of the current two acceptable?

Siedlerchr:
Thanks for your contribution! I have tested the changes locally and it works as expected.
A little step for you, a great step for ~~humanity~~ JabRef users ;)
Just add the suggestions by  @tobiasdiez  and we can merge it in. 

PS: If you include the word  `Fixes #xxxx` then the corresponding issue will be closed automatically when the PR is merged.

src/main/java/org/jabref/gui/fieldeditors/LinkedFileViewModel.java
Siedlerchr: Just add them in a new commit. We usually squash all commits on merge

bdach:
Thank you for your pointers! I have locally added `ButtonData.YES`, and now I will attempt to prepare a test suite for the `remove()` method at least, and push everything out when it's ready. :smile:

bdach:
Okay, I pushed out the changes. Adding the button data restored the original button order (at least on Linux), but the remove button is now clearly the default, and as far as I can tell from [JavaFX docs](https://docs.oracle.com/javase/8/javafx/api/javafx/scene/control/ButtonBar.html) this is the desired behaviour.

The test code I'm not too proud of -- there's some weird hacking going on there to get the tests running, and I had to balance between not making them brittle and not changing the implementation just for tests' sake. If you have ideas for improvements, I'd be happy to go back and fix the ugliness.

src/test/java/org/jabref/gui/fieldeditors/LinkedFileViewModelTest.java
```diff
@@ -0,0 +1,154 @@
...
+        File tempFile = tempFolder.newFile();
+        linkedFile = new LinkedFile("", tempFile.getAbsolutePath(), "");
+        when(dialogService.showCustomButtonDialogAndWait(
+                any(AlertType.class),
+                anyString(),
+                anyString(),
+                any(ButtonType.class),
+                any(ButtonType.class),
+                any(ButtonType.class)
+        )).thenAnswer(invocation -> Optional.of(invocation.getArgument(3))); // first vararg - remove button
```
bdach: To explain myself with this -- the implementation uses strict reference equality for checking user choices (at least for the first two), so I had to either get out the parameter from the invocation like this, or modify the equality to use `ButtonData`, which made the implementation very weird looking.

bdach:
Well, if I'm this close, might as well try and make this PR as good as possible!

I've added the suggested changes. However, I did leave some of the parentheses as they were -- near those long mock setup calls to be precise, because I couldn't make them look decent otherwise.

tobiasdiez:
Nice, thanks again for your contributions. We are looking forward to see more PRs from you 😄 .

Issue #3673, opened 2018-01-29
Manually protecting text case by "Add {} around selected text" doesn't work


anassal:
- [x] I have tested the latest master version from http://builds.jabref.org/master/ and the problem persists

On:

JabRef 4.1 Windows 8.1 6.3 x86  Java 1.8.0_151
JabRef 4.2-dev Windows 8.1 6.3 x86 Java 1.8.0_151

Steps to reproduce:

1. Open entry editor 
2. Right click on a field (e.g., title) 
3. Select a text to protect its case, e.g., IEEE.
4. Right click --> protect items --> Add {} around selected text
5. Sometimes, {} appear for a moment around the selected text before quickly disappear.

Siedlerchr:
Win10 works for me in the latest master.  But could be related to that space character problem #3471

stefan-kolb:
Thank you for your report :+1:
However, I cannot reproduce this behavior on the current `master`. Can you please try the latest build from http://builds.jabref.org/master.

anassal:
Update: The problem occurred only when the library is saved (automatically in my case) and was because of the `all-text-fields[latex_to_unicode]` saveAction.

```
@Comment{jabref-meta: saveActions:enabled;
  all-text-fields[latex_to_unicode]
;} 
```

 Is this the intended behaviour of  `all-text-fields[latex_to_unicode]`?

lenhard:
@anassal Yes, this is the expected behavior of the safe action. Curly braces are special characters in LaTeX, so they will be removed when LaTeX is converted to Unicode. 

If you want to keep the braces in your fields, then you should not apply this save action. 

Since we have clarified this, I am closing the issue.

PR #3691, opened 2018-02-04
Fix some FX  Thread issue

Siedlerchr:
While investigating #3673  I noticed a thread error when changing the entry otherwise. 

I think this could also solve some other issues here 

----


tobiasdiez:
Is this fix also needed for the new maintable? I would prefer if you push such fixes directly onto this new branch.

Siedlerchr:
@tobiasdiez  There the fix is already present, the whole method is wrapped in the FX Thread in that branch

Issue #3344, opened 2017-10-25
Find Duplicates marks different book chapters as duplicates


notuntoward:
Quality --> "Find duplicates" marks different chapters of the same book as duplicates (see attached example).

JabRef 4.1-dev--snapshot--2017-10-23--master--451b6f3de
Windows 7 6.1 amd64 
Java 1.8.0_151

[notDupes.txt](https://github.com/JabRef/jabref/files/1413987/notDupes.txt)

lenhard:
The duplicate finder is not perfect (far from it) and seems to be happy if titles are the same. There is plenty of room for making the detection algorithm smarter.

Naitoreivun:
I would like to take a look at that issue :)

tobiasdiez:
@Naitoreivun Perfect! If you get stuck somewhere, feel free to post your question here or contact us directly in the [developer chat](https://gitter.im/JabRef/jabref).

PR #3694, opened 2018-02-05
differentiate inbooks with different chapters

Naitoreivun:
This is my simple solution for #3344 issue. I've added statement that checks authors, titles and chapters of two inbooks. What do you think?

I have several questions:
1. I'm not sure about using `compareSingleField` function for comparing chapters, because sometimes they are very similar, but different, for instance: 'chapter 1' and 'chapter 2'. Any idea? My is to remove 'chapter' word from strings and then make the comparison. 
2. Should I also add test for incollections?
3. What about pages? As I noticed they can be a replacement for chapters. Should I consider them in the code?
4. Can I add another commits with some cleanup and refactor for the files I've touched? I mean adding some finals to make code less error prone etc.


Siedlerchr:
Thanks for your contribution!
Regarding your questions, I can only answer some of them. 
1, I currently are on my mobile so I can't dig into that much but sounds like an idea. 
You could look what the biblatex manual says about the possibile contents of the field. That's our reference when in doubt about a field 

2, maybe also a good idea
3 could make sense if the same 
4 sure, code improvement is always good if it's within the scope of the implementation Regarding of finals, you can omit  them in parameters


Regarding test, you could migrate them to junit 5 if not already done. For examples you can look at the fetcher tests. Just a bit different Syntax

Naitoreivun:
Ok, so I added some more tests and PAGES field checks. Nevertheless I don't know how to upgrade test to JUnit5. Any tips?

@tobiasdiez - in that case compareSingleField would return for example EMPTY_IN_ONE cosntant, where my if statement require 'NOT_EQUAL' so there is no problem :) anyway I've added test to cover that case

Also I've made some cleanup and refactor. I know, a lot of finals, but imo it prevents silly mistakes like reassigning values to function parameters (especially error prone in constructors).


So... How about now?

stefan-kolb:
Thank you for your contribution 😄 👍

Issue #3535, opened 2017-12-15
Magnifier does not change its colors any more


koppor:
![grafik](https://user-images.githubusercontent.com/1366654/34044953-5e72e7f2-e1a7-11e7-9975-f800e03d6ebd.png)

JabRef has two search modes: https://help.jabref.org/en/Search#search-modes

In old versions, the active mode was indicted using a different color of the magnifier. Now, it just stays black...

koppor:
I would assume that changing colors can be implemented quicker than such a fancy search bar. 😇

koppor:
I labeled this as "bug", even though it is a "regression". We don't have the label "regression" and we aim for keeping the number of labels low, don't we? 🌞

koppor:
![search](https://user-images.githubusercontent.com/1366654/35925980-5eabdefe-0c27-11e8-9a2a-c2abc2b3509b.gif)

koppor:
Link to the search modes of v4: https://docs.jabref.org/v4/finding-sorting-and-cleaning-entries/search#search-modes

PR #3708, opened 2018-02-07
Try to change color of magnifier icon

lenhard:
I am trying to change the color of the magnifier icon as requested here: #3535

However, the icon refuses a color change. It even opts out of the color scheme of JabRef, since it is black instead of purple (despite the fact that the icon is configured as purple in the code).

@tobiasdiez How can I change the color of this icon? Any change will do, really. For reasons unknown to me the icon straight-out refuses the change. 

----

### DIFF
--- a/src/main/java/org/jabref/gui/search/SearchTextField.java
+++ b/src/main/java/org/jabref/gui/search/SearchTextField.java
@@ -1,8 +1,10 @@
 package org.jabref.gui.search;
 
+import javafx.scene.Node;
 import javafx.scene.control.TextField;
 
 import org.jabref.gui.IconTheme;
+import org.jabref.gui.util.DefaultTaskExecutor;
 import org.jabref.logic.l10n.Localization;
 
 import org.controlsfx.control.textfield.CustomTextField;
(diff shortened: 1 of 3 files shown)


src/main/java/org/jabref/gui/search/SearchTextField.java
```diff
@@ -13,7 +15,21 @@
     public static TextField create() {
         CustomTextField textField = (CustomTextField) TextFields.createClearableTextField();
         textField.setPromptText(Localization.lang("Search") + "...");
-        textField.setLeft(IconTheme.JabRefIcon.SEARCH.getGraphicNode());
+        Node node = IconTheme.JabRefIcon.SEARCH.getGraphicNode();
+        node.setStyle("-fx-text-fill: #00ffff");
```
tobiasdiez: -fx-fill should work, but do you really think that changing the icon color is a good idea? How intuitive (and important) is it to signal an advanced search mode by magenta?

lenhard: Unfortunately, `fx-fill` doesn't help. There seems to be a deeper problem with exactly this icon.

Actually, I am quite indifferent to this, but @koppor really wants it.

koppor: I really like the proposal at https://github.com/JabRef/jabref/issues/3535#issuecomment-352068826. However, JabRef had this functionality for at least one year. Animation at https://github.com/JabRef/jabref/issues/3535#issuecomment-363813333. - Simon and I had a long discussion on that. The decision was that we should present to the user that there are different search modes. They should "just work", but for the pros, there should be a hint which mode is active. Not that someone wonders why the search does not work. -- Furthermore, that icon makes it less hard to debug the search. It is clear, whether the search term is a grammar-based search or just plain text string search.

tobiasdiez: ok, I see your point. It is just again a feature that needs documentation before anybody understands what's going it. 

Butt if `fx-fill` and `fx-text-fill` don't work, I'm running out of ideas. I usually use a combination out of both and it worked without any problems so far. Maybe it works if you base this PR on the maintable-beta branch. I changed a few things regarding javafx icons there.

lenhard: I have tried some more (not the maintable branch, though) with no success. I think this is down to `CustomTextField` from controls-fx. It offers the functionality to add a node at the left or right side, but it seems to be impossible to change the style of this node.

Because of this, I've gone back to the previous version: A Swing icon next to the text field and everything just works. I know this is a step back, but if the `CustomTextField` cannot be made to use our icon theme, then we should probably migrate the whole `GlobalSearchBar` to JavaFX instead. Until this happens, I think it is wiser to use icons that preserve our color theme.

@tobiasdiez @koppor: I'll now leave it open to you to continue arguing, merge, or close this PR. I am fine with any solution.

koppor:
Screenshot: 
grafik

There is a space between the other icons. Until the cool solution from https://github.com/JabRef/jabref/issues/3535#issuecomment-352068826 is implemented, I am in favor of using this approach. Reason: This makes the JabRef feature explicit.

papers search.

tobiasdiez:
Well, if @koppor really wants this I'm fine with it although I bet that more users get confused by a cyan-highlighted icon than they are getting enlighted by it.

PR #3710, opened 2018-02-07
[WIP] Dublin Core

### LINKED ISSUE #938: BibTeXML vs. bibteXMP

johannes-manner:
Fixes #938 : Only supporting [Dublin Core](https://en.wikipedia.org/wiki/Dublin_Core)

Removed JabRef namespace for XMP 'xmlns:bibtex'. 
Removed some unused methods in gui/importer/EntryFromPDFCreator.java (discussed with @koppor and @stefan-kolb)
Deleted tests for XMPUtil. (New ones are in progress).

----


build.gradle
```diff
@@ -90,10 +90,10 @@ dependencies {
     compile 'com.jgoodies:jgoodies-common:1.8.1'
     compile 'com.jgoodies:jgoodies-forms:1.9.0'
 
-    // update to 2.0.x is not possible - see https://github.com/JabRef/jabref/pull/1096#issuecomment-208857517
-    compile 'org.apache.pdfbox:pdfbox:1.8.13'
-    compile 'org.apache.pdfbox:fontbox:1.8.13'
-    compile 'org.apache.pdfbox:jempbox:1.8.13'
+    // update possible because custom metadata format was dropped, https://github.com/JabRef/jabref/pull/3710 - https://github.com/JabRef/jabref/pull/1096
```
koppor: Just remove that line

src/main/java/org/jabref/logic/xmp/XMPUtil.java
```diff
@@ -318,51 +297,106 @@ public static PDDocument loadWithAutomaticDecryption(File file) throws IOExcepti
...
+     *
+     * @param dcSchema The document information from which to build a BibEntry.
+     * @return The bibtex entry found in the document information.
+     */
+    public static Optional<BibEntry> getBibtexEntry(DublinCoreSchema dcSchema,
+            XMPPreferences xmpPreferences) {
+
+        BibEntry entry = new BibEntry();
+
+        // Contributor -> Editor
```
koppor: I would not comment these functions as the implementation might change... The comments could co at `setEditor` etc. as method comment.

src/main/java/org/jabref/cli/XMPUtilMain.java
```diff
@@ -92,63 +90,62 @@ public static void main(String[] args) throws IOException, TransformerException
...
 
-                if (entries.isEmpty()) {
-                    System.err.println("Could not find BibEntry in " + args[0]);
-                } else {
-                    XMPUtil.writeXMP(new File(args[1]), entries, result.getDatabase(), false, xmpPreferences);
-                    System.out.println("XMP written.");
+                    if (entries.isEmpty()) {
+                        System.err.println("Could not find BibEntry in " + args[0]);
+                    } else {
+                        XMPUtil.writeXMP(new File(args[1]), entries, result.getDatabase(), false, xmpPreferences);
```
Siedlerchr: shouldn't this also be either a Path?

src/main/java/org/jabref/cli/XMPUtilMain.java
```diff
@@ -92,63 +90,62 @@ public static void main(String[] args) throws IOException, TransformerException
...
                     System.err.println("The given pdf does not contain any XMP-metadata.");
                 }
-                break;
+                return;
             }
 
             if (args[0].endsWith(".bib") && args[1].endsWith(".pdf")) {
-                ParserResult result = new BibtexParser(importFormatPreferences, Globals.getFileUpdateMonitor()).parse(new FileReader(args[0]));
+                try (FileReader reader = new FileReader(args[0])) {
+                    ParserResult result = new BibtexParser(importFormatPreferences, Globals.getFileUpdateMonitor()).parse(reader);
```
Siedlerchr: I am not sure, but does the parser accepts an inputStream? Then you could use Files.newInputStream...
would be actually preferable over a fileReader

johannes-manner: No overloaded parse method, which would accept an InputStream.

src/main/java/org/jabref/cli/XMPUtilMain.java
johannes-manner: No. The parameter is a file or sometimes also the filePath as a string is possible.

johannes-manner:
This pull request is ready for the final review and for merging into master :)

src/test/java/org/jabref/logic/xmp/XMPUtilTest.java
```diff
@@ -254,1144 +64,106 @@ public void setUp() throws IOException, COSVisitorException {
...
-
-            schemas = meta.getSchemasByNamespaceURI(XMPSchemaMediaManagement.NAMESPACE);
-            Assert.assertEquals(1, schemas.size());
-            XMPSchemaMediaManagement mm = (XMPSchemaMediaManagement) schemas.get(0);
-            Assert.assertEquals("17", mm.getSequenceList("xapMM:VersionID").get(0));
-
-        }
+    public void testReadEmtpyMetadata() throws IOException, URISyntaxException {
+        List<BibEntry> entries = XMPUtil.readXMP(XMPUtil.class.getResource("/org/jabref/logic/xmp/empty_metadata.pdf").toURI().getPath(), xmpPreferences);
+        Assert.assertEquals(0, entries.size());
```
tobiasdiez: `assertEquals(Collections.emptyList, entries)` to get a better error message in case the test fails.

src/test/java/org/jabref/logic/xmp/XMPUtilTest.java
```diff
@@ -254,1144 +64,106 @@ public void setUp() throws IOException, COSVisitorException {
...
-                StandardCharsets.UTF_8)) {
-            ParserResult result = new BibtexParser(importFormatPreferences, fileMonitor).parse(fr);
-
-            Assert.assertEquals("Arvind", result.getDatabase().resolveForStrings("#Arvind#"));
 
-            AuthorList originalAuthors = AuthorList.parse(
-                    "Patterson, David and Arvind and Asanov\\'\\i{}c, Krste and Chiou, Derek and Hoe, James and Kozyrakis, Christos and Lu, S{hih-Lien} and Oskin, Mark and Rabaey, Jan and Wawrzynek, John");
+        // read a bib entry from the tests before
+        List<BibEntry> entries = XMPUtil.readXMP(XMPUtil.class.getResource("/org/jabref/logic/xmp/PD_metadata.pdf").toURI().getPath(), xmpPreferences);
+        BibEntry entry = entries.get(0);
```
tobiasdiez: I think it is easier if you just create the entry by hand (and then write it, read and compare).

src/test/java/org/jabref/logic/xmp/XMPUtilTest.java
```diff
@@ -254,1144 +64,106 @@ public void setUp() throws IOException, COSVisitorException {
...
-        try (BufferedReader fr = Files.newBufferedReader(Paths.get("src/test/resources/org/jabref/util/twente.bib"),
-                StandardCharsets.UTF_8)) {
-            ParserResult result = new BibtexParser(importFormatPreferences, fileMonitor).parse(fr);
-
-            Assert.assertEquals("Arvind", result.getDatabase().resolveForStrings("#Arvind#"));
 
-            AuthorList originalAuthors = AuthorList.parse(
-                    "Patterson, David and Arvind and Asanov\\'\\i{}c, Krste and Chiou, Derek and Hoe, James and Kozyrakis, Christos and Lu, S{hih-Lien} and Oskin, Mark and Rabaey, Jan and Wawrzynek, John");
+        // read a bib entry from the tests before
+        List<BibEntry> entries = XMPUtil.readXMP(XMPUtil.class.getResource("/org/jabref/logic/xmp/PD_metadata.pdf").toURI().getPath(), xmpPreferences);
```
tobiasdiez: I think it is easier if you just create the entry by hand (and then write it, read and compare).

src/test/java/org/jabref/logic/xmp/XMPUtilTest.java
```diff
@@ -254,1144 +64,106 @@ public void setUp() throws IOException, COSVisitorException {
...
-        Assert.assertEquals(Optional.of("2003"), e.getField("year"));
-        Assert.assertEquals(Optional.of("�pt�mz�t��n"), e.getField("title"));
-        Assert.assertEquals("misc", e.getType());
-    }
-
-    /**
-     * Make sure that the privacy filter works.
+     * The month attribute in DublinCore is the complete name of the month, e.g. March.
+     * In JabRef, the format is #mar# instead. To get a working unit test, the JabRef's
+     * bib-entry is altered from #mar# to {March}.
```
tobiasdiez: Is there a reason why the month shouldn't be post-processed to get a proper bibtex value? We even have the `Month` class which handles parsing and converting to the correct output.

build.gradle
```diff
@@ -90,10 +90,9 @@ dependencies {
     compile 'com.jgoodies:jgoodies-common:1.8.1'
     compile 'com.jgoodies:jgoodies-forms:1.9.0'
 
-    // update to 2.0.x is not possible - see https://github.com/JabRef/jabref/pull/1096#issuecomment-208857517
-    compile 'org.apache.pdfbox:pdfbox:1.8.13'
-    compile 'org.apache.pdfbox:fontbox:1.8.13'
-    compile 'org.apache.pdfbox:jempbox:1.8.13'
+    compile 'org.apache.pdfbox:pdfbox:2.0.8'
+    compile 'org.apache.pdfbox:fontbox:2.0.8'
```
tobiasdiez: Further below (starting at line 218), we added exceptions for the update dependencies task. These are now invalid and should be removed.

src/test/java/org/jabref/logic/xmp/XMPUtilTest.java
johannes-manner: Done

src/main/java/org/jabref/logic/xmp/XMPUtil.java
```diff
@@ -95,87 +90,54 @@ private XMPUtil() {
...
-                            }
-                            result.add(entry.get());
+                // Only support Dublin Core since JabRef 4.2
+                DublinCoreSchema schema = meta.get().getDublinCoreSchema();
+                if (schema != null) {
+                    Optional<BibEntry> entry = XMPUtil.getBibtexEntry(schema, xmpPreferences);
+
+                    if (entry.isPresent()) {
+                        if (entry.get().getType() == null) {
+                            entry.get().setType(BibEntry.DEFAULT_TYPE);
```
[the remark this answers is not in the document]
johannes-manner: Done

src/main/java/org/jabref/logic/xmp/XMPUtil.java
```diff
@@ -95,87 +90,54 @@ private XMPUtil() {
...
-            // try the empty string as user password
-            StandardDecryptionMaterial sdm = new StandardDecryptionMaterial("");
-            try {
-                doc.openProtection(sdm);
-            } catch (BadSecurityHandlerException | CryptographyException e) {
-                LOGGER.error("Cannot handle encrypted PDF: " + e.getMessage());
-                throw new EncryptedPdfsNotSupportedException();
-            }
-        }
+    public static PDDocument loadWithAutomaticDecryption(File file) throws IOException {
```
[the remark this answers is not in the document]
johannes-manner: Done

build.gradle
johannes-manner: DONE

src/main/java/org/jabref/logic/xmp/XMPUtil.java
```diff
@@ -95,87 +90,54 @@ private XMPUtil() {
      * @throws IOException          If the file could not be written to or could not be found.
      */
     public static void writeXMP(String fileName, BibEntry entry,
-                                BibDatabase database, XMPPreferences xmpPreferences) throws IOException, TransformerException {
-        XMPUtil.writeXMP(new File(fileName), entry, database, xmpPreferences);
+            BibDatabase database, XMPPreferences xmpPreferences) throws IOException, TransformerException {
+        XMPUtil.writeXMP(Paths.get(fileName), entry, database, xmpPreferences);
```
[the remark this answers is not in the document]
johannes-manner: I tried to separete the logic in two utility classes and a shared utitlity class. 
Currently I don't see the benefit of getting "normal" reader and writer. I introduced two Extractors for the DocumentInformation and the DublinCore format. Maybe that's what you have intended.

I pushed a WIP state for another review. Maybe you have further suggestions, how to structure the package.

johannes-manner:
For my general understanding and my newbie state:

` public static void writeXMP(Path path,
            Collection<BibEntry> bibtexEntries, BibDatabase database,
            boolean writePDFInfo, XMPPreferences xmpPreferences) throws IOException, TransformerException {

        Collection<BibEntry> resolvedEntries;
        if (database == null) {
            resolvedEntries = bibtexEntries;
        } else {
            resolvedEntries = database.resolveForStrings(bibtexEntries, false);
        }

        try (PDDocument document = PDDocument.load(path.toFile())) {
            if (document.isEncrypted()) {
                throw new EncryptedPdfsNotSupportedException();
            }

            if (writePDFInfo && (resolvedEntries.size() == 1)) {
                XMPUtilWriter.writeDocumentInformation(document, resolvedEntries
                        .iterator()
                        .next(), null, xmpPreferences);
                XMPUtilWriter.writeDublinCore(document, resolvedEntries, null, xmpPreferences);
            }

            PDDocumentCatalog catalog = document.getDocumentCatalog();
            PDMetadata metaRaw = catalog.getMetadata();`

src/main/java/org/jabref/logic/xmp/DocumentInformationExtractor.java
```diff
@@ -0,0 +1,99 @@
...
+    private void extractKeywords() {
+        String s = documentInformation.getKeywords();
+        if (s != null) {
+            bibEntry.setField(FieldName.KEYWORDS, s);
+        }
+    }
+
+    private void extractSubject() {
+        String s = documentInformation.getSubject();
+        if (s != null) {
```
Siedlerchr: We once agreed on avoiding single char variables. Please use a meaningful name here and for the others

src/main/java/org/jabref/logic/xmp/DocumentInformationExtractor.java
```diff
@@ -0,0 +1,99 @@
...
+            bibEntry.setField(FieldName.ABSTRACT, s);
+        }
+    }
+
+    private void extractOtherFields() {
+        COSDictionary dict = documentInformation.getCOSObject();
+        for (Map.Entry<COSName, COSBase> o : dict.entrySet()) {
```
Siedlerchr: I think you could encapsulate part of this as a stream. at least the filtering and mapping to the key.
https://www.mkyong.com/java8/java-8-filter-a-map-examples/

src/main/java/org/jabref/logic/xmp/DublinCoreExtractor.java
```diff
@@ -0,0 +1,387 @@
...
+
+    /**
+     * Author in BibTex - Creator in DublinCore
+     *
+     * @param bibEntry The BibEntry object, which is filled during metadata extraction.
+     * @param dcSchema Metadata in DublinCore format.
+     */
+    private void extractAuthor() {
+        List<String> creators = dcSchema.getCreators();
+        if ((creators != null) && !creators.isEmpty()) {
```
Siedlerchr: I think we have a method in our own StringUtil class which combines both checks: IsNullOREmpty

johannes-manner:
For my general understanding and my newbie state:

` public static void writeXMP(Path path,
            Collection<BibEntry> bibtexEntries, BibDatabase database,
            boolean writePDFInfo, XMPPreferences xmpPreferences) throws IOException, TransformerException {

        Collection<BibEntry> resolvedEntries;
        if (database == null) {
            resolvedEntries = bibtexEntries;
        } else {
            resolvedEntries = database.resolveForStrings(bibtexEntries, false);
        }

        try (PDDocument document = PDDocument.load(path.toFile())) {
            if (document.isEncrypted()) {
                throw new EncryptedPdfsNotSupportedException();
            }

            if (writePDFInfo && (resolvedEntries.size() == 1)) {                // 1
                XMPUtilWriter.writeDocumentInformation(document, resolvedEntries
                        .iterator()
                        .next(), null, xmpPreferences);
                XMPUtilWriter.writeDublinCore(document, resolvedEntries, null, xmpPreferences);
            }

            PDDocumentCatalog catalog = document.getDocumentCatalog();
            PDMetadata metaRaw = catalog.getMetadata();`

For me, it makes no sense to write more than one BibEntry to the metadata of a PDF file.  Currently the implementation in XMPUtilWriter is also limited to a single element, but implemented as a list. (See line 260 and 110).

Can I drop the list?

src/main/java/org/jabref/logic/xmp/DublinCoreExtractor.java
```diff
@@ -0,0 +1,387 @@
...
+     * Year + Month in BibTex - Date in DublinCore is a combination of year and month information
+     *
+     * @param bibEntry The BibEntry object, which is filled during metadata extraction.
+     * @param dcSchema Metadata in DublinCore format.
+     */
+    private void extractYearAndMonth() {
+        List<String> dates = dcSchema.getUnqualifiedSequenceValueList("date");
+        if ((dates != null) && !dates.isEmpty()) {
+            String date = dates.get(0).trim();
+            Calendar c = null;
```
Siedlerchr: Please stick to the java 8 new date and time api: Use a DateTime or just a DateFormatter 
DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm");
LocalDateTime dateTime = LocalDateTime.parse(str, formatter);

src/main/java/org/jabref/logic/xmp/DublinCoreExtractor.java
```diff
@@ -0,0 +1,387 @@
...
+     * This method sets all fields, which are custom in bibtext and therefore supported by jabref, but which are not included in the DublinCore format.
+     * <p/>
+     * The relation attribute of DublinCore is abused to insert these custom fields.
+     *
+     * @param bibEntry The BibEntry object, which is filled during metadata extraction.
+     * @param dcSchema Metadata in DublinCore format.
+     */
+    private void extractBibTexFields() {
+        List<String> relationships = dcSchema.getRelations();
+        if (relationships != null) {
```
Siedlerchr: Why not use a stream with filter and map

src/main/java/org/jabref/logic/xmp/DublinCoreExtractor.java
```diff
@@ -0,0 +1,387 @@
...
+
+    /**
+     * Type is equivalent in both formats (BibTex and DublinCore)
+     *
+     * @param bibEntry The BibEntry object, which is filled during metadata extraction.
+     * @param dcSchema Metadata in DublinCore format.
+     */
+    private void extractType() {
+        List<String> l = dcSchema.getTypes();
+        if ((l != null) && !l.isEmpty()) {
```
Siedlerchr: Please avoid single char vairbales

src/main/java/org/jabref/logic/xmp/DublinCoreExtractor.java
```diff
@@ -0,0 +1,387 @@
...
+        // Fields for which not to write XMP data later on:
+        Set<String> filters = new TreeSet<>(xmpPreferences.getXmpPrivacyFilter());
+
+        for (Entry<String, String> field : bibEntry.getFieldMap().entrySet()) {
+
+            if (useXmpPrivacyFilter && filters.contains(field.getKey())) {
+                continue;
+            }
+
+            if (FieldName.EDITOR.equals(field.getKey())) {
```
Siedlerchr: I would prefer a switch/case here, but that's just my style. The others don't like them that much ;) So you can change it or leave it.

johannes-manner: I leave it :)

src/main/java/org/jabref/logic/xmp/DublinCoreExtractor.java
johannes-manner: Done with https://github.com/JabRef/jabref/pull/3710/commits/366aeed6a6870386c9eb4ba657332faff8f7d576

src/main/java/org/jabref/logic/xmp/XMPUtilReader.java
```diff
@@ -0,0 +1,124 @@
...
+     */
+    public static List<BibEntry> readXMP(Path path, XMPPreferences xmpPreferences)
+            throws IOException {
+
+        List<BibEntry> result = new LinkedList<>();
+
+        try (PDDocument document = loadWithAutomaticDecryption(path)) {
+            Optional<XMPMetadata> meta = XMPUtilReader.getXMPMetadata(document);
+
+            if (meta.isPresent()) {
```
Siedlerchr: You can use chained maps on the optionals for each step:
meta.map(m->m.getDublinCore().map(m->new DublinCore...).map(entry->extractBibEntry).ifPresent(result::add)

src/main/java/org/jabref/logic/xmp/DublinCoreExtractor.java
johannes-manner: Done with https://github.com/JabRef/jabref/pull/3710/commits/366aeed6a6870386c9eb4ba657332faff8f7d576. Please check the conditions again. It's not that natural to negate the isNullOrEmpty... Maybe another function would help in the StringUtils.

src/main/java/org/jabref/logic/xmp/XMPUtilWriter.java
```diff
@@ -0,0 +1,302 @@
...
+     * @param path     The path to write to.
+     * @param entry    The entry to write.
+     * @param database maybenull An optional database which the given bibtex entries belong to, which will be used to
+     *                 resolve strings. If the database is null the strings will not be resolved.
+     * @throws TransformerException If the entry was malformed or unsupported.
+     * @throws IOException          If the file could not be written to or could not be found.
+     */
+    public static void writeXMP(Path file, BibEntry entry,
+            BibDatabase database, XMPPreferences xmpPreferences) throws IOException, TransformerException {
+        List<BibEntry> l = new LinkedList<>();
```
Siedlerchr: Reason for a linked list? In most cases ArrayList is sufficient

src/main/java/org/jabref/logic/xmp/XMPUtilReader.java
johannes-manner: I find the current state more readable than the chained map operations... Do you have another suggestion?

src/main/java/org/jabref/logic/xmp/XMPUtilWriter.java
johannes-manner: That's a question, I posted above. This is code from a previous author and I want to drop the List at all. Please comment my question above.

src/main/java/org/jabref/logic/xmp/XMPUtilShared.java
```diff
@@ -0,0 +1,58 @@
...
+     */
+    public static boolean hasMetadata(Path path, XMPPreferences xmpPreferences) {
+        try {
+            List<BibEntry> bibEntries = XMPUtilReader.readXMP(path, xmpPreferences);
+            return !bibEntries.isEmpty();
+        } catch (EncryptedPdfsNotSupportedException ex) {
+            LOGGER.info("Encryption not supported by XMPUtil");
+            return false;
+        } catch (IOException e) {
+            // happens if no metadata is found, no reason to log the exception
```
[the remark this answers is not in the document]
johannes-manner: Done.

src/main/java/org/jabref/logic/xmp/XMPUtilReader.java
Siedlerchr: The chanining has the improvement that you don't have to care about nulls or empty optionals inside. The code at the end is only executed if none of the values is null or empty.

Did not see your question. Theoretically it could be that someone has CrossRef linked bibentries, (e.g. a bib entry with a book and another one with a chapter referring the book).
 but that's really and edge case. I think for most uses cases  one entry is enough. And if the code doesn't support it, then drop it.  let's see what @koppor  has to say

src/main/java/org/jabref/logic/xmp/DocumentInformationExtractor.java
johannes-manner: Done.

tobiasdiez:
I have no real idea concerning your question about the list. I would say you could safely remove it, but I'm not sure. What happens if multiple entries have the same file linked and write their metadata into it (e.g. the pdf is a book and I have a bunch of `bookchapters` as separate entries)?

johannes-manner:
`// Remove all current Dublin-Core schemas
        meta.removeSchema(meta.getDublinCoreSchema());`
XMPUtilWriter, line 147,148

Before new metadata is written to the PDF, all DublinCoreSchemas are removed. So I think, that there is currently no option to write more than one metaschema.
In your case, I would assume that the last written data is visible (others lost).

Siedlerchr:
Well, does the Dublin core schema specify. Multiple entries?

koppor:
Specifications are available at. http://dublincore.org/specifications/. The abstract model is there: http://dublincore.org/documents/abstract-model/

grafik

In case, I interpret that correctly, one record contains one description, which may contain multiple record sets.

This is the edge case, where one wants to write XMP to a proceedings.

Example:

- One chapter: https://link.springer.com/chapter/10.1007/978-3-540-79230-7_4
- Full proceedings: https://link.springer.com/book/10.1007/978-3-540-79230-7

src/test/java/org/jabref/logic/xmp/XMPUtilWriterTest.java
```diff
@@ -0,0 +1,162 @@
...
+            "  date       = {2006},\r\n" +
+            "  eprinttype = {asdf},\r\n" +
+            "  eventdate  = {2017-05-31},\r\n" +
+            "  owner      = {Christoph Schwentker},\r\n" +
+            "  timestamp  = {2016.11.07},\r\n" +
+            "  url        = {asdfasdfas},\r\n" +
+            "  urldate    = {2017-05-31},\r\n" +
+            "}";
+
+    private static final String vapnik2000 = "@Book{Vapnik2000,\r\n" +
```
tobiasdiez: Please create the BibEntries by hand (using `new BibEntry(), setField`) and not based on the string representation. The XMP test should be as autonomous as possible, especially they shouldn't fail if the BibParser is changed.

src/main/java/org/jabref/logic/xmp/XMPUtilReader.java
```diff
@@ -30,7 +35,7 @@ private XMPUtilReader() {
      * @param path The path to read the XMPMetadata from.
      * @return The XMPMetadata object found in the file
      */
-    public static Optional<XMPMetadata> readRawXMP(Path path) throws IOException {
+    public static Optional<List<XMPMetadata>> readRawXMP(Path path) throws IOException {
```
tobiasdiez: A `List` should always be non-null and thus it does not makes sense to wrap an `Optional` around it. The not-present case corresponds to an empty list, which you can check using `isEmpty()`.

src/test/java/org/jabref/logic/xmp/XMPUtilWriterTest.java
johannes-manner: DONE

src/main/java/org/jabref/logic/xmp/XMPUtilReader.java
johannes-manner: Thanks for this comment. Done 👍

src/test/java/org/jabref/logic/xmp/XMPUtilTest.java
johannes-manner: If there is no further reason (besides simplicity), I would stay with this.

src/main/java/org/jabref/logic/xmp/DocumentInformationExtractor.java
johannes-manner: I would stay with the implemented version. Currently it works and I'm not a stream fanatic 😋

johannes-manner:
I considered all comments to the source code and refactored my code. 

The last commit alters the behavior of the XMP import:
The previous implementation only imports the first entry and drops the others. The current implementation imports all metadata entries. My thoughts are as follows: It is easier to import additional entries and delete not needed ones compared to importing single entries by hand, if the needed entry is not the first one.

src/test/java/org/jabref/logic/xmp/XMPUtilReaderTest.java
```diff
@@ -0,0 +1,137 @@
...
+     * Tests the readRawXMP - method
+     *
+     * @throws IOException
+     * @throws URISyntaxException
+     * @throws ParseException
+     */
+    @Test
+    public void testReadArticleDublinCoreReadXMP() throws IOException, URISyntaxException, ParseException {
+
+        Path path = Paths.get(XMPUtilShared.class.getResource("/org/jabref/logic/xmp/article_dublinCore.pdf").toURI());
```
koppor: Does one really need the absolute path here? I thought, just the filename is enough, because it looks up in the current directory (where the resources are mirrored to). If this is the package org.jabref.logic.xmp, it should "just work".

src/test/java/org/jabref/logic/xmp/XMPUtilReaderTest.java
```diff
@@ -0,0 +1,137 @@
...
+     * <p/>
+     * Tests the readRawXMP - method
+     *
+     * @throws IOException
+     * @throws URISyntaxException
+     * @throws ParseException
+     */
+    @Test
+    public void testReadArticleDublinCoreReadXMP() throws IOException, URISyntaxException, ParseException {
+
```
koppor: Please remove empty line

src/test/java/org/jabref/logic/xmp/XMPUtilReaderTest.java
```diff
@@ -0,0 +1,137 @@
...
+        when(xmpPreferences.getKeywordSeparator()).thenReturn(',');
+
+        parser = new BibtexParser(importFormatPreferences, fileMonitor);
+    }
+
+    /**
+     * The month attribute in DublinCore is the complete name of the month, e.g. March.
+     * In JabRef, the format is #mar# instead. To get a working unit test, the JabRef's
+     * bib-entry is altered from #mar# to {March}.
+     * <p/>
```
koppor: Remove this and the next line. The next line should be included in the method name. Maybe, the method has to renamed to `testReadArticleDublinCoreReadRawXMP`?

src/test/java/org/jabref/logic/xmp/XMPUtilReaderTest.java
```diff
@@ -0,0 +1,137 @@
...
+     * Tests the readRawXMP - method
+     *
+     * @throws IOException
```
koppor: Remove empty `@throws` comments. It is already clear by the method signature which exceptions are thrown.

src/test/java/org/jabref/logic/xmp/XMPUtilReaderTest.java
```diff
@@ -0,0 +1,137 @@
+package org.jabref.logic.xmp;
+
```
koppor: Only one empty line (?)

src/main/java/org/jabref/pdfimport/PdfImporter.java
```diff
@@ -29,7 +29,7 @@
 import org.jabref.logic.l10n.Localization;
 import org.jabref.logic.util.UpdateField;
 import org.jabref.logic.util.io.FileUtil;
-import org.jabref.logic.xmp.XMPUtil;
+import org.jabref.logic.xmp.XMPUtilShared;
```
koppor: Maybe rename the class to `XmpUtilShared` to follow Google's Casing rules? (Similar for method names, ...)

src/test/java/org/jabref/logic/xmp/XMPUtilReaderTest.java
johannes-manner: Done

src/test/java/org/jabref/logic/xmp/XMPUtilReaderTest.java
johannes-manner: Of course 👍

src/test/java/org/jabref/logic/xmp/XMPUtilReaderTest.java
johannes-manner: done

src/test/java/org/jabref/logic/xmp/XMPUtilReaderTest.java
johannes-manner: done

src/test/java/org/jabref/logic/xmp/XMPUtilReaderTest.java
johannes-manner: Yes. I do not need the absolute path 👍

PR #3713, opened 2018-02-11
Add EndNote XML importer

tobiasdiez:
As wished [in the forum](http://discourse.jabref.org/t/import-from-bookends-or-endnote/1048).

----


tobiasdiez:
@Siedlerchr Can you please try to import the test file in Linux. The user reports that JabRef freezes after he confirms the import dialog but I can't reproduce the problem here on Windows. Thanks.

Siedlerchr:
Xubuntu 16.04 works fine. No issues.

Siedlerchr:
The problem is that in the xml file from the user there are records in which apparently many fields are null and thus result in an NPE.
I fixed some of them.

The root problem I see here is that this thing like Titles etc all are serialized to their own object and not to collections  and some of them might be null.
If it were mapped to collections we could just initalize it with empty collections and avoid the null shit. 
Maybe this approach helps:
http://blog.bdoughan.com/2012/11/creating-generic-list-wrapper-in-jaxb.html

Siedlerchr:
I think I found an idea:
First we generate an XSD file from a concrete xml file and then use that as input for xjc. This should hopefully be able to generate lists from what I read.. Or we convert the dtd to xsd first.

src/main/java/org/jabref/logic/importer/fileformat/EndnoteXmlImporter.java
```diff
@@ -107,9 +107,10 @@ public ParserResult importDatabase(BufferedReader reader) throws IOException {
                 // Check whether we have an article set, an article, a book article or a book article set
                 Xml root = (Xml) unmarshalledObject;
                 List<BibEntry> bibEntries = root.getRecords()
-                                                .getRecord().stream()
-                                                .map(this::parseRecord)
-                                                .collect(Collectors.toList());
+                        .getRecord()
```
tobiasdiez: @Siedlerchr Thanks for the fix! However, it appears there is still a problem with your eclipse style configuration (alignment at dot).

src/test/java/org/jabref/logic/importer/fileformat/ImporterTestEngine.java
```diff
@@ -47,9 +49,13 @@ public static void testIsNotRecognizedFormat(Importer importer, String fileName)
         Assertions.assertFalse(importer.isRecognizedFormat(getPath(fileName), StandardCharsets.UTF_8));
     }
 
-    public static void testImportEntries(Importer importer, String fileName, String fileType) throws IOException {
-        List<BibEntry> entries = importer.importDatabase(getPath(fileName), StandardCharsets.UTF_8).getDatabase()
-                .getEntries();
+    public static void testImportEntries(Importer importer, String fileName, String fileType) throws IOException, ImportException {
+        ParserResult parserResult = importer.importDatabase(getPath(fileName), StandardCharsets.UTF_8);
+        if (parserResult.isInvalid()) {
+            throw new ImportException(parserResult.getErrorMessage());
```
lenhard: Why not throw an IOException instead? Or maybe put the call to `importDatabase` into a try-catch that converts the IOException into an ImportException? 

Overall, it would be better to throw just one Exception type. Then you don't have to change all tests to `throws Exception`, which imho is ugly, because it is so generic.

src/main/java/org/jabref/logic/importer/fileformat/EndnoteXmlImporter.java
```diff
@@ -0,0 +1,212 @@
...
+        }
+        return false;
+    }
+
+    @Override
+    public ParserResult importDatabase(BufferedReader reader) throws IOException {
+        Objects.requireNonNull(reader);
+
+        try {
+            JAXBContext context = JAXBContext.newInstance("org.jabref.logic.importer.fileformat.endnote");
```
lenhard: Constructing a `JAXBContext` is very expensive. It is so expensive that you notice it even as a human. Please turn this into an instance variable, then the construction takes place only once per importer object. Or you could just store the `Unmarshaller`, which should be enough as well.

src/main/java/org/jabref/logic/importer/fileformat/EndnoteXmlImporter.java
```diff
@@ -0,0 +1,212 @@
...
+    @Override
+    public String getDescription() {
+        return "Importer for the EndNote XML format.";
+    }
+
+    @Override
+    public boolean isRecognizedFormat(BufferedReader reader) throws IOException {
+        String str;
+        int i = 0;
+        while (((str = reader.readLine()) != null) && (i < 50)) {
```
lenhard: That's a funny way of determining the format :-) Somewhere in the first 50 lines a `records` tag has to appear? Why up to 50? What if some random garbage comes before the `records` tag? Would JAXB still be able to parse that?

src/test/java/org/jabref/logic/importer/fileformat/ImporterTestEngine.java
tobiasdiez: Isn't the generic `Exception` the preferred choice for tests anyway? It has the advantage that you not need to change your tests if you decide to change the exception signature of the method under test.

I prefer to keep it that way. The error is not coming directly from the file system and thus it is not an IOException in my opinion.

src/main/java/org/jabref/logic/importer/fileformat/EndnoteXmlImporter.java
tobiasdiez: Good to know! I actually copied this part from the medline importer and will change it there too.

src/main/java/org/jabref/logic/importer/fileformat/EndnoteXmlImporter.java
tobiasdiez: That's another thing I just copied from the medline importer. Can jaxb test the file without actually parsing everything? If not I find the "record appears near the beginning" a relative good heuristic.

lenhard: Nope, JAXB will try to parse everything. So you could go with such a heuristic, but could you just improve the code a bit? E.g. no side effect (line reading) in the condition of the while, promoting the magic number 50 to a `private final static` and so on.

src/test/java/org/jabref/logic/importer/fileformat/ImporterTestEngine.java
Siedlerchr: In tests it doesn't matter which exceptions are thrown

Siedlerchr:
Coday is just complaining about the juntit 5 annotations, but that's not your fault. Otherwise from my side code is good as well.

PR #3714, opened 2018-02-13
Update entry editor when entry type changes

lenhard:
Fixes an issue partly described here: https://github.com/koppor/jabref/issues/289 
So far, changing the type of an entry did not update the type label in the entry editor or the structure of its tabs, which might offer different fields for a different type.

This turns the type attribute of `BibEntry` into a property so that it is possible to listen to this property and to react to changes. It also adds a `ChangeListener` to the `EntryEditor` that updates the label and the contents of the visible tab.

@tobiasdiez It would be cool if you could have a quick look at the listener related stuff to check that I remove it correctly and do not introduce any memory leaks.

----

### DIFF
--- a/src/main/java/org/jabref/gui/entryeditor/EntryEditor.java
+++ b/src/main/java/org/jabref/gui/entryeditor/EntryEditor.java
@@ -231,6 +231,10 @@ public class EntryEditor extends BorderPane {
         TypedBibEntry typedEntry = new TypedBibEntry(entry, bibDatabaseContext.getMode());
         typeLabel.setText(typedEntry.getTypeForDisplay());
 
+        entry.getTypeProperty().addListener(e -> {
+            typeLabel.setText(new TypedBibEntry(entry, bibDatabaseContext.getMode()).getTypeForDisplay());
+        });
+
         // Add type change menu
         ContextMenu typeMenu = new ChangeEntryTypeMenu().getChangeEntryTypePopupMenu(entry, bibDatabaseContext, undoManager);
         typeLabel.setOnMouseClicked(event -> typeMenu.show(typeLabel, Side.RIGHT, 0, 0));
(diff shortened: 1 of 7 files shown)


src/main/java/org/jabref/gui/entryeditor/EntryEditor.java
```diff
@@ -224,6 +234,10 @@ public void setEntry(BibEntry entry) {
 
             setupToolBar();
         });
+
+        // listen to type changes for rebuilding the currently visible tab
+        typeListener = new EntryTypeListener(this.entry, bibDatabaseContext.getMode(), typeLabel, tabbed);
+        this.entry.getTypeProperty().addListener(typeListener);
```
tobiasdiez: It is actually easier to use `Easybind.subcribe()` to listen to changes: you don't need a listener class and provide the code simply as a lambda expression. In this case, you get a subscription as the return value and call `unsubscribe` to, well, unsubscribe.

src/main/java/org/jabref/model/entry/BibEntry.java
```diff
@@ -200,6 +202,10 @@ public boolean hasCiteKey() {
      * Returns this entry's type.
      */
     public String getType() {
+        return type.getValue();
+    }
+
+    public StringProperty getTypeProperty() {
```
tobiasdiez: The convention is that such methods are not prefixed with `get`, i.e., use `typeProperty()` instead.

lenhard: Renamed.

src/main/java/org/jabref/gui/entryeditor/EntryEditor.java
lenhard: Thanks! I've refactored the listener into an EasyBind subscription as you suggested. Slowly but steadily I learn more about programming JavaFX :)

PR #3715, opened 2018-02-13
Group interface: Usability tweaks

Escoul:
Fixes [koppor#277](https://github.com/koppor/jabref/issues/277)

This fixes the name of the group editing window to "Add group" instead of "Edit Group" when adding a new group.
The group editing window can now also be called by double-clicking the group to be edited.

----

----

_Credit (for an exercise):
Rico Haas
David Zeck_

### DIFF
--- a/src/main/java/org/jabref/gui/groups/GroupDialog.java
+++ b/src/main/java/org/jabref/gui/groups/GroupDialog.java
@@ -148,7 +148,8 @@ class GroupDialog extends JabRefDialog implements Dialog<AbstractGroup> {
      *                    created.
      */
     public GroupDialog(JabRefFrame jabrefFrame, AbstractGroup editedGroup) {
-        super(jabrefFrame, Localization.lang("Edit group"), true, GroupDialog.class);
+        super(jabrefFrame, (editedGroup == null) ? Localization.lang("Add group") : Localization.lang("Edit group"),
+                true, GroupDialog.class);
 
         // set default values (overwritten if editedGroup != null)
         keywordGroupSearchField.setText(jabrefFrame.prefs().get(JabRefPreferences.GROUPS_DEFAULT_FIELD));
--- a/src/main/java/org/jabref/gui/groups/GroupTreeController.java
+++ b/src/main/java/org/jabref/gui/groups/GroupTreeController.java
@@ -178,7 +178,7 @@ public class GroupTreeController extends AbstractController<GroupTreeViewModel> 
             row.setOnMouseClicked((mouseClickedEvent) -> {
                 if (mouseClickedEvent.getButton().equals(MouseButton.PRIMARY)
                         && (mouseClickedEvent.getClickCount() == 2)) {
-                    GroupNodeViewModel groupToEdit = EasyBind.monadic(row.itemProperty()).getValue();
+                    GroupNodeViewModel groupToEdit = row.itemProperty().getValue();
                     if (groupToEdit != null) {
                         viewModel.editGroup(groupToEdit);
                     }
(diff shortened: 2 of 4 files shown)


src/main/java/org/jabref/gui/groups/GroupDialog.java
```diff
@@ -148,7 +148,8 @@ public Dimension getPreferredSize() {
      *                    created.
      */
     public GroupDialog(JabRefFrame jabrefFrame, AbstractGroup editedGroup) {
-        super(jabrefFrame, Localization.lang("Edit group"), true, GroupDialog.class);
+        super(jabrefFrame, (editedGroup == null) ? Localization.lang("Add group") : Localization.lang("Edit group"),
```
lenhard: I am really no friend of the inline syntax for conditional branching. In this special case (calling super), I'll go with it though. The reason for this is it's not possible to do a proper branching before the `super` call and the line is actually quite readable.

src/main/resources/l10n/JabRef_en.properties
```diff
@@ -386,6 +386,7 @@ Edit\ entry=Edit entry
 Save\ file=Save file
 Edit\ file\ type=Edit file type
 
+Add\ group=Add group
```
lenhard: There is already `add\ group` in line 57 of the same file. We really don't need localizations that differ only in the casing of the first word. Please remove the old (lower case) localization and make sure that everything still works.

src/main/java/org/jabref/gui/groups/GroupTreeController.java
```diff
@@ -173,6 +174,17 @@ public void initialize() {
             row.setDisclosureNode(null);
             row.disclosureNodeProperty().addListener((observable, oldValue, newValue) -> row.setDisclosureNode(null));
 
+            // Opens the group's edit window when the group gets double clicked
+            row.setOnMouseClicked((mouseClickedEvent) -> {
+                if (mouseClickedEvent.getButton().equals(MouseButton.PRIMARY)
+                        && (mouseClickedEvent.getClickCount() == 2)) {
+                    GroupNodeViewModel groupToEdit = EasyBind.monadic(row.itemProperty()).getValue();
```
lenhard: Why is there a call to `Easybind.monadic`? To get the selected group, you only need to call `row.itemProperty().getValue()` and that's it. Please change the line accordingly.

src/main/resources/l10n/JabRef_en.properties
Escoul: The old localization line is used by  org.jabref.gui.groups.UndoableAddOrRemoveGroup.getPresentationName() to name a undoable/redoable action. The problem is, that all actions of this kind are named in lowercase, it would be awkward to change just this name to uppercase.

src/main/java/org/jabref/gui/groups/GroupDialog.java
Escoul: Thanks, i think so too. The only alternative would have been to create a separate method just to determine the name of the frame.

src/main/java/org/jabref/gui/groups/GroupTreeController.java
Escoul: Good point, fixed in 3ae9885.

src/main/resources/l10n/JabRef_en.properties
lenhard: I'm not happy with having the essentially same translation key twice, but I get your point. We can leave this as it is.

lenhard:
Ok, with these changes the code is good to merge and, thus, we have positive reviews from two developers. So, I'll merge the PR. Once again, thanks for your contribution!

koppor:
@Escoul Thank you for working on that and giving good answers to the questions. What would you think to add an architectural decision on your conditional if? https://github.com/JabRef/jabref/blob/master/CONTRIBUTING.md#when-making-an-architectural-decision

All in all, we would be happy to see more contributions from your side. We use the label [good-first-issue](https://github.com/JabRef/jabref/labels/good%20first%20issue), where more issues will be tagged. You can also have a look at https://github.com/koppor/jabref/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22. Or hop by at my office in V38 in Stuttgart. 😇

PR #3718, opened 2018-02-14
Add cleanup operation for replacing ligatures

### LINKED ISSUE #3613: Add cleanup operation that replaces ligatures
We should add a cleanup operation that replaces unicode ligatures such as `ﬁ` (this is one character!) with their normal "expanded form", i.e. `fi` (two characters). Since they are hard to spot, result in compile errors and are not needed, this cleanup should be added to the list of recommended cleanups for both biblatex and bibtex.

lenhard:
Fixes #3613

I have used the mapping listed here: https://en.wikipedia.org/wiki/Typographic_ligature#Ligatures_in_Unicode_(Latin_alphabets)

Please note that german authors might be surprised by how this replaces the letter `ß`.

----

### DIFF
--- a/src/main/java/org/jabref/logic/layout/format/ReplaceUnicodeLigaturesFormatter.java
+++ b/src/main/java/org/jabref/logic/layout/format/ReplaceUnicodeLigaturesFormatter.java
@@ -1,0 +1,53 @@
+package org.jabref.logic.layout.format;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.regex.Pattern;
+
+import org.jabref.logic.l10n.Localization;
+import org.jabref.logic.layout.LayoutFormatter;
+import org.jabref.logic.util.strings.UnicodeLigaturesMap;
+import org.jabref.model.cleanup.Formatter;
+
+public class ReplaceUnicodeLigaturesFormatter implements LayoutFormatter, Formatter {
+
+    private Map<Pattern, String> ligaturesMap;
+
+    public ReplaceUnicodeLigaturesFormatter() {
+        ligaturesMap = new HashMap<>();
+        UnicodeLigaturesMap stringMap = new UnicodeLigaturesMap();
+        for (String key : stringMap.keySet()) {
+            ligaturesMap.put(Pattern.compile(key), stringMap.get(key));
+        }
+    }
+
+    @Override
+    public String getName() {
+        return Localization.lang("Replace Unicode ligatures");
+    }
+
+    @Override
+    public String getKey() {
+        return "remove_unicode_ligatures";
+    }
+
+    @Override
+    public String format(String fieldText) {
+        String result = fieldText;
+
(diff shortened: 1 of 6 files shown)


src/main/java/org/jabref/logic/layout/format/ReplaceUnicodeLigaturesFormatter.java
```diff
@@ -0,0 +1,53 @@
...
+    @Override
+    public String getKey() {
+        return "remove_unicode_ligatures";
+    }
+
+    @Override
+    public String format(String fieldText) {
+        String result = fieldText;
+
+        for (Pattern key : ligaturesMap.keySet()) {
```
tobiasdiez: I like the following code a bit more:
````
ligaturesMap.foreach((pattern, replacement) -> {
     result = pattern.matcher(result).replaceAll(replacement);
});
````
but this is probably a matter of taste....

File: src/main/java/org/jabref/logic/util/strings/UnicodeLigaturesMap.java [hunk 2]
```diff
@@ -0,0 +1,67 @@
+package org.jabref.logic.util.strings;
+
+import java.util.HashMap;
+
+public class UnicodeLigaturesMap extends HashMap<String, String> {
```
Siedlerchr: Why not using a BiMap from Guava? As far as I know we already do that for the html/latex mapping as well.
I would be in favor of this - as a way to convert back

src/main/java/org/jabref/logic/layout/format/ReplaceUnicodeLigaturesFormatter.java
lenhard: That's not possible because of side effects. `result` is overwritten in every iteration. You shouldn't do this in a lambda. Hence, I'd like to keep it as it is.

src/main/java/org/jabref/logic/util/strings/UnicodeLigaturesMap.java
lenhard: I had a quick look at this. `BibMap` is an interface and its implementations (e.g. `HashBiMap`) don't have constructors, so I cannot extend them.  So, I would have to implement all map methods anew.

Although it would be nice, let us wait with something like that until we have an actual use case. Right now, reversing a standard Java Map is much less code than hooking into the Guava collections.

lenhard:
Ok, I've taken out the conversion of ß and documented this in the code. If there are no other characters that we want to exclude then this PR should be good to go.

src/main/java/org/jabref/logic/layout/format/ReplaceUnicodeLigaturesFormatter.java
koppor: Annotate ADR? 😇 -> Y-Statement instead of huge MADR?

PR #3721, opened 2018-02-14
[WIP] Fix for issue 2831

### LINKED ISSUE #2831: Run "Abbreviate journal names" in parallel
JabRef 3.8.2
windows 10 10.0 amd64 
Java 1.8.0_131

As far as I can tell, the "Abbreviate journal names" function in JabRef is currently run serially. While this is fine for smaller databases, it can pose a serious performance problem when dealing with large (>10,000 entries) databases. I propose to run this function in parallel - this should speed up things quite a lot, I guess.

Brainsucker92:
Fixes #2831 

Added new method _executeAndReturn_ to the JabRefExecutorService that allows to run callables that return a value when the computation is completed. Using this method the abbreviations in _AbbreviateAction.java_ should run parallel now.

----

### DIFF
--- a/src/main/java/org/jabref/JabRefExecutorService.java
+++ b/src/main/java/org/jabref/JabRefExecutorService.java
@@ -89,19 +89,16 @@ public class JabRefExecutorService implements Executor {
      */
     public <T> Future<T> executeAndReturn(Callable<T> command) {
         Objects.requireNonNull(command);
-
         return executorService.submit(command);
     }
     
     public <T> List<Future<T>> executeAll(Collection<Callable<T>> tasks) throws InterruptedException {
         Objects.requireNonNull(tasks);
-        
         List<Future<T>> futures = null; 
-        
         try {
             futures =  executorService.invokeAll(tasks);
         } catch (InterruptedException e) {
-            LOGGER.debug("Invokation has been interrupted during execution.");
+            LOGGER.info("Invokation has been interrupted during execution.");
         }
         Objects.requireNonNull(futures);
         return futures;
--- a/src/main/java/org/jabref/JabRefExecutorService.java
+++ b/src/main/java/org/jabref/JabRefExecutorService.java
@@ -2,6 +2,7 @@ package org.jabref;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 import java.util.Timer;
@@ -80,7 +81,7 @@ public class JabRefExecutorService implements Executor {
     }
 
     /**
-     * This method allows to execute a callable task that provides a return value after the calculation is done.
+     * Executes a callable task that provides a return value after the calculation is done.
      * 
      * @param command The task to execute.
      * @return A Future object that provides the returning value.
@@ -90,13 +91,20 @@ public class JabRefExecutorService implements Executor {
         return executorService.submit(command);
     }
 
+    /**
+     * Executes a collection of callable tasks and returns a List of the resulting Future objects after the calculation is done. 
+     * 
+     * @param tasks The tasks to execute
+     * @return A List of Future objects that provide the returning values.
+     */
     public <T> List<Future<T>> executeAll(Collection<Callable<T>> tasks) {
         Objects.requireNonNull(tasks);
         List<Future<T>> futures = new ArrayList<>();
         try {
             futures = executorService.invokeAll(tasks);
         } catch (InterruptedException exception) {
-            LOGGER.error("Unble to execute tasks", exception);
+            LOGGER.error("Unable to execute tasks", exception);
+            return Collections.emptyList();
         }
         return futures;
     }
--- a/src/main/java/org/jabref/JabRefExecutorService.java
+++ b/src/main/java/org/jabref/JabRefExecutorService.java
@@ -110,10 +110,7 @@ public class JabRefExecutorService implements Executor {
     }
 
     public void executeInterruptableTaskAndWait(Runnable runnable) {
-        if (runnable == null) {
-            LOGGER.debug("Received null as command for execution");
-            return;
-        }
+        Objects.requireNonNull(runnable);
 
         Future<?> future = lowPriorityExecutorService.submit(runnable);
         while (true) {
--- a/src/main/java/org/jabref/JabRefExecutorService.java
+++ b/src/main/java/org/jabref/JabRefExecutorService.java
@@ -1,5 +1,7 @@
 package org.jabref;
 
+import java.util.Collection;
+import java.util.Objects;
 import java.util.Timer;
 import java.util.TimerTask;
 import java.util.concurrent.Callable;
@@ -40,19 +42,13 @@ public class JabRefExecutorService implements Executor {
 
     @Override
     public void execute(Runnable command) {
-        if (command == null) {
-            LOGGER.debug("Received null as command for execution");
-            return;
-        }
-
+        Objects.requireNonNull(command);
+        
         executorService.execute(command);
     }
 
     public void executeAndWait(Runnable command) {
-        if (command == null) {
-            LOGGER.debug("Received null as command for execution");
-            return;
-        }
+        Objects.requireNonNull(command);
 
         Future<?> future = executorService.submit(command);
         while (true) {
@@ -67,12 +63,9 @@ public class JabRefExecutorService implements Executor {
         }
     }
 
-    public boolean executeAndWait(Callable command) {
-        if (command == null) {
-            LOGGER.debug("Received null as command for execution");
-            return false;
-        }
-
+    public boolean executeAndWait(Callable<?> command) {
+        Objects.requireNonNull(command);
+        
         Future<?> future = executorService.submit(command);
         while (true) {
             try {
@@ -89,16 +82,25 @@ public class JabRefExecutorService implements Executor {
 
     /**
      * This method allows to execute a callable task that provides a return value after the calculation is done.
+     * 
      * @param command The task to execute.
      * @return A Future object that provides the returning value.
      */
     public <T> Future<T> executeAndReturn(Callable<T> command) {
-        if (command == null) {
-            LOGGER.debug("Received null as command for execution");
-            return null;
-        }
+        Objects.requireNonNull(command);
 
         return executorService.submit(command);
+    }
+    
+    public void executeAll(Collection<? extends Callable<?>> tasks) {
+        Objects.requireNonNull(tasks);
+        
+        
+        try {
+            executorService.invokeAll(tasks);
+        } catch (InterruptedException e) {
+            LOGGER.debug("Invokation has been interrupted during execution.");
+        }
     }
 
     public void executeInterruptableTask(final Runnable runnable) {
--- a/src/main/java/org/jabref/JabRefExecutorService.java
+++ b/src/main/java/org/jabref/JabRefExecutorService.java
@@ -89,14 +89,14 @@ public class JabRefExecutorService implements Executor {
         Objects.requireNonNull(command);
         return executorService.submit(command);
     }
-    
+
     public <T> List<Future<T>> executeAll(Collection<Callable<T>> tasks) {
         Objects.requireNonNull(tasks);
-        List<Future<T>> futures = new ArrayList<>(); 
+        List<Future<T>> futures = new ArrayList<>();
         try {
-            futures =  executorService.invokeAll(tasks);
-        } catch (InterruptedException e) {
-            LOGGER.info("Invokation has been interrupted during execution.");
+            futures = executorService.invokeAll(tasks);
+        } catch (InterruptedException exception) {
+            LOGGER.error("Unble to execute tasks", exception);
         }
         return futures;
     }
--- a/src/main/java/org/jabref/JabRefExecutorService.java
+++ b/src/main/java/org/jabref/JabRefExecutorService.java
@@ -1,5 +1,6 @@
 package org.jabref;
 
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Objects;
@@ -91,13 +92,12 @@ public class JabRefExecutorService implements Executor {
     
     public <T> List<Future<T>> executeAll(Collection<Callable<T>> tasks) {
         Objects.requireNonNull(tasks);
-        List<Future<T>> futures = null; 
+        List<Future<T>> futures = new ArrayList<>(); 
         try {
             futures =  executorService.invokeAll(tasks);
         } catch (InterruptedException e) {
             LOGGER.info("Invokation has been interrupted during execution.");
         }
-        Objects.requireNonNull(futures);
         return futures;
     }
 
--- a/src/main/java/org/jabref/JabRefExecutorService.java
+++ b/src/main/java/org/jabref/JabRefExecutorService.java
@@ -12,7 +12,6 @@ import java.util.concurrent.Future;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-
 /**
  * Responsible for managing of all threads (except Swing threads) in JabRef
  */
@@ -36,7 +35,8 @@ public class JabRefExecutorService implements Executor {
     private final Timer timer = new Timer("timer", true);
     private Thread remoteThread;
 
-    private JabRefExecutorService() { }
+    private JabRefExecutorService() {
+    }
 
     @Override
     public void execute(Runnable command) {
@@ -85,6 +85,20 @@ public class JabRefExecutorService implements Executor {
                 return false;
             }
         }
+    }
+
+    /**
+     * This method allows to execute a callable task that provides a return value after the calculation is done.
+     * @param command The task to execute.
+     * @return A Future object that provides the returning value.
+     */
+    public <T> Future<T> executeAndReturn(Callable<T> command) {
+        if (command == null) {
+            LOGGER.debug("Received null as command for execution");
+            return null;
+        }
+
+        return executorService.submit(command);
     }
 
     public void executeInterruptableTask(final Runnable runnable) {
--- a/src/main/java/org/jabref/JabRefExecutorService.java
+++ b/src/main/java/org/jabref/JabRefExecutorService.java
@@ -1,6 +1,7 @@
 package org.jabref;
 
 import java.util.Collection;
+import java.util.List;
 import java.util.Objects;
 import java.util.Timer;
 import java.util.TimerTask;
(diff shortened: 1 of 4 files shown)


src/main/java/org/jabref/gui/journals/AbbreviateAction.java
```diff
@@ -53,10 +57,18 @@ public void run() {
...
+            futures.add(result);
+        }
+        
+        for (Future<Boolean> future : futures) {
+            try {
+                if(future.get()) {
+                    count++;
+                }
+            } catch (InterruptedException | ExecutionException e) {
+                // Do nothing.
```
Siedlerchr: Just please add a Logger debug statement here, otherwise underlying eceptions may get swalloed

Brainsucker92: Alright, already done.

koppor:
@Brainsucker92 I added `Fixes #2831` to your PR description. Otherwise, GitHub does not link the issue properly. --> GitHub has no linking functionality in the header

src/main/java/org/jabref/JabRefExecutorService.java
```diff
@@ -87,6 +87,20 @@ public boolean executeAndWait(Callable command) {
         }
     }
 
+    /**
+     * This method allows to execute a callable task that provides a return value after the calculation is done.
+     * @param command The task to execute.
+     * @return A Future object that provides the returning value.
+     */
+    public <T> Future<T> executeAndReturn(Callable<T> command) {
+        if (command == null) {
```
koppor: Why this? Can't we just put `Objects.requireNonNull(command)`? I think, Item 49 of [Effective Java, 3rd edition](https://www.amazon.de/Effective-Java-Third-Joshua-Bloch/dp/0134685997?tag=jabref-21).

src/main/java/org/jabref/JabRefExecutorService.java
```diff
@@ -87,6 +87,20 @@ public boolean executeAndWait(Callable command) {
         }
     }
 
+    /**
+     * This method allows to execute a callable task that provides a return value after the calculation is done.
+     * @param command The task to execute.
```
koppor: Empty line before parameter description

src/main/java/org/jabref/JabRefExecutorService.java
```diff
@@ -87,6 +87,20 @@ public boolean executeAndWait(Callable command) {
...
+     * @return A Future object that provides the returning value.
+     */
+    public <T> Future<T> executeAndReturn(Callable<T> command) {
+        if (command == null) {
+            LOGGER.debug("Received null as command for execution");
+            return null;
```
koppor: `null` should never be used. Use Optionals. (See http://java.by-comparison.com/ and effective java).

In the concrete case, use `Objects.requireNonNull()`.

src/main/java/org/jabref/gui/journals/AbbreviateAction.java
```diff
@@ -54,9 +62,17 @@ public void run() {
...
             };
 
-            boolean result = JabRefExecutorService.INSTANCE.executeAndWait(callable);
-            if (result) {
-                count++;
+            Future<Boolean> result = JabRefExecutorService.INSTANCE.executeAndReturn(callable);
+            futures.add(result);
+        }
+
+        for (Future<Boolean> future : futures) {
```
koppor: Why not `ExecutorService.invokeAll()`? See https://stackoverflow.com/a/1250668/873282.

src/main/java/org/jabref/JabRefExecutorService.java
Brainsucker92: I did the null check this way because every other method in the class checked for null exactly the same way. In this case we should consider changing all other methods to `Objects.requireNonNull(...)`

koppor: Yeah, this is the preferred way. Hope, nothing breaks if this is done at other places. 😇

src/main/java/org/jabref/JabRefExecutorService.java
```diff
@@ -89,17 +82,26 @@ public boolean executeAndWait(Callable command) {
...
-            return null;
-        }
+        Objects.requireNonNull(command);
 
         return executorService.submit(command);
     }
+    
+    public void executeAll(Collection<? extends Callable<?>> tasks) {
+        Objects.requireNonNull(tasks);
+
```
koppor: Please remove empty line. - only one empty line at most 😇

src/main/java/org/jabref/JabRefExecutorService.java
```diff
@@ -89,17 +82,26 @@ public boolean executeAndWait(Callable command) {
...
     }
+    
+    public void executeAll(Collection<? extends Callable<?>> tasks) {
+        Objects.requireNonNull(tasks);
+        
+        
+        try {
+            executorService.invokeAll(tasks);
+        } catch (InterruptedException e) {
+            LOGGER.debug("Invokation has been interrupted during execution.");
```
koppor: I would do it at as `info` as it might be interesting for the user.

src/main/java/org/jabref/gui/journals/AbbreviateAction.java
koppor: Use `executeAll` from above?

Brainsucker92:
Alright, I made a couple of changes. Hopefully everything seems to be better and runs nice and smoothly.

src/main/java/org/jabref/JabRefExecutorService.java
```diff
@@ -87,6 +78,29 @@ public boolean executeAndWait(Callable command) {
...
+    
+    public <T> List<Future<T>> executeAll(Collection<Callable<T>> tasks) {
+        Objects.requireNonNull(tasks);
+        List<Future<T>> futures = null; 
+        try {
+            futures =  executorService.invokeAll(tasks);
+        } catch (InterruptedException e) {
+            LOGGER.info("Invokation has been interrupted during execution.");
+        }
+        Objects.requireNonNull(futures);
```
koppor: This is asserted by the API. I think, this line can be removed.

Typically `Objects.requireNonNull(...)` is only used at the beginning of a method to check parameters. Not in the middle.

src/main/java/org/jabref/JabRefExecutorService.java
```diff
@@ -87,6 +78,29 @@ public boolean executeAndWait(Callable command) {
...
+     * 
+     * @param command The task to execute.
+     * @return A Future object that provides the returning value.
+     */
+    public <T> Future<T> executeAndReturn(Callable<T> command) {
+        Objects.requireNonNull(command);
+        return executorService.submit(command);
+    }
+    
+    public <T> List<Future<T>> executeAll(Collection<Callable<T>> tasks) {
```
koppor: According to the discussion at https://stackoverflow.com/a/1250668/873282, all futures should be available. Thus, no need to return something.

src/main/java/org/jabref/gui/journals/AbbreviateAction.java
```diff
@@ -53,10 +63,20 @@ public void run() {
...
+        }
 
-            boolean result = JabRefExecutorService.INSTANCE.executeAndWait(callable);
-            if (result) {
-                count++;
+        // Execute the callables and wait for the results.
+        List<Future<Boolean>> futures = JabRefExecutorService.INSTANCE.executeAll(tasks);
+
+        // Evaluate the results of the callables.
+        for (Future<Boolean> future : futures) {
```
koppor: According to https://stackoverflow.com/a/1250668/873282, this is not required. See also above.

src/main/java/org/jabref/JabRefExecutorService.java
```diff
@@ -87,6 +78,29 @@ public boolean executeAndWait(Callable command) {
...
+        Objects.requireNonNull(tasks);
+        List<Future<T>> futures = null; 
+        try {
+            futures =  executorService.invokeAll(tasks);
+        } catch (InterruptedException e) {
+            LOGGER.info("Invokation has been interrupted during execution.");
```
koppor: An empty list should be returned.

src/main/java/org/jabref/gui/journals/AbbreviateAction.java
koppor: `futures.stream().map(f -> f.get()).countif(...` --> Please try to use Java8's streaming API - if possible. 😇

src/main/java/org/jabref/JabRefExecutorService.java
koppor: I hope, this goes well 😇

src/main/java/org/jabref/JabRefExecutorService.java
Brainsucker92: I just wanted to avoid returning null. I'll remove it.

koppor: Yeah, totally understandable. I think, that's why `@NonNull` annotations are there for... --> https://help.eclipse.org/neon/index.jsp?topic=%2Forg.eclipse.jdt.doc.user%2Ftasks%2Ftask-using_null_annotations.htm We are currently not using them in JabRef.

Otherwise, the whole could had to be cluttered with `Objects.requireNonNull(...)` statements, because `null` should be avoided in any case. 🔢

src/main/java/org/jabref/gui/journals/AbbreviateAction.java
koppor: I checked stackoverflow. It should read `count = futures.stream().filter(f -> f.get()).count();` Shouldn't it?

Brainsucker92: Isn't using a filter handier? Like this:
`futures.stream().filter(...).count()`?

Brainsucker92: Indeed. Using a filter seems to be the better option to me. Only downside of this is, we need to catch  the Exceptions within the filter... but ok, I'll do it that way.

src/main/java/org/jabref/JabRefExecutorService.java
```diff
@@ -87,6 +79,28 @@ public boolean executeAndWait(Callable command) {
...
+        Objects.requireNonNull(command);
+        return executorService.submit(command);
+    }
+    
+    public <T> List<Future<T>> executeAll(Collection<Callable<T>> tasks) {
+        Objects.requireNonNull(tasks);
+        List<Future<T>> futures = new ArrayList<>(); 
+        try {
+            futures =  executorService.invokeAll(tasks);
+        } catch (InterruptedException e) {
```
tobiasdiez: Please use `logger.error` and pass the exception as an argument. Btw, Invokation is German not English ;-)

src/main/java/org/jabref/gui/journals/AbbreviateAction.java
```diff
@@ -53,12 +62,24 @@ public void run() {
...
+        // Execute the callables and wait for the results.
+        List<Future<Boolean>> futures = JabRefExecutorService.INSTANCE.executeAll(tasks);
+
+        // Evaluate the results of the callables.
+        long count = futures.stream().filter(f -> {
+            boolean result;
+            try {
+                result = f.get();
+            } catch (InterruptedException | ExecutionException e) {
+                LOGGER.debug("Invokation has been interrupted during execution.");
```
tobiasdiez: again, pls `LOGGER.error`

src/main/java/org/jabref/gui/journals/AbbreviateAction.java
```diff
@@ -53,12 +62,24 @@ public void run() {
...
-            if (result) {
-                count++;
+        // Execute the callables and wait for the results.
+        List<Future<Boolean>> futures = JabRefExecutorService.INSTANCE.executeAll(tasks);
+
+        // Evaluate the results of the callables.
+        long count = futures.stream().filter(f -> {
+            boolean result;
+            try {
+                result = f.get();
```
tobiasdiez: Instead of the try, catch, finally construct it is actually easier to follow the rule "return as soon as possible" and write something like
````java
try {
   return f.get();
} catch (...) {
   Log;
   return false;
}
````

File: src/main/java/org/jabref/gui/journals/AbbreviateAction.java [hunk 15]
```diff
@@ -53,12 +62,24 @@ public void run() {
...
+        }
 
-            boolean result = JabRefExecutorService.INSTANCE.executeAndWait(callable);
-            if (result) {
-                count++;
+        // Execute the callables and wait for the results.
+        List<Future<Boolean>> futures = JabRefExecutorService.INSTANCE.executeAll(tasks);
+
+        // Evaluate the results of the callables.
+        long count = futures.stream().filter(f -> {
```
tobiasdiez: please don't use abbreviations, `future` is better than `f`.

src/main/java/org/jabref/JabRefExecutorService.java
Brainsucker92: Unfortunately, `Logger` does not provide any `PrintStream` to print the stacktrace as suggested here: https://stackoverflow.com/a/1149712 by using `e.printStackTrace(anyPrintStreamHere)`. So what's there preferred way to print the StackTrace to the LOGGER.error method?

src/main/java/org/jabref/gui/journals/AbbreviateAction.java
Brainsucker92: In this case I was looking at this anwser: https://stackoverflow.com/a/17405806 which prefers the way to return at the end of the method. I will change the code as you suggested.

src/main/java/org/jabref/JabRefExecutorService.java
```diff
@@ -87,6 +79,28 @@ public boolean executeAndWait(Callable command) {
...
+        return executorService.submit(command);
+    }
+
+    public <T> List<Future<T>> executeAll(Collection<Callable<T>> tasks) {
+        Objects.requireNonNull(tasks);
+        List<Future<T>> futures = new ArrayList<>();
+        try {
+            futures = executorService.invokeAll(tasks);
+        } catch (InterruptedException exception) {
+            LOGGER.error("Unble to execute tasks", exception);
```
koppor: Just to be sure: Do not miss the `return Collections.emptyList();` here.

src/main/java/org/jabref/gui/journals/AbbreviateAction.java
```diff
@@ -53,12 +62,21 @@ public void run() {
...
-            boolean result = JabRefExecutorService.INSTANCE.executeAndWait(callable);
-            if (result) {
-                count++;
+        // Evaluate the results of the callables.
+        long count = futures.stream().filter(future -> {
+            try {
+                return future.get();
+            } catch (InterruptedException | ExecutionException exception) {
+                LOGGER.error("Unable to retrieve value.", exception);
+                return false;
```
koppor: I would not break the loop here, just continue with the next future. If the `return false;` makes sense here, please add a comment why `false` is returned.

tobiasdiez: the return is for the lambda so everything is fine.

src/main/java/org/jabref/gui/journals/AbbreviateAction.java
```diff
@@ -22,6 +28,8 @@
     private String message = "";
     private final boolean iso;
 
+    private static final Logger LOGGER = LoggerFactory.getLogger(AbbreviateAction.class);
```
koppor: Travis fails becoaseu of checkstyle:

```
:checkstyleMain[ant:checkstyle] [ERROR] /home/travis/build/JabRef/jabref/src/main/java/org/jabref/gui/journals/AbbreviateAction.java:18: 'org.slf4j.Logger' should be separated from previous imports. [ImportOrder]
[ant:checkstyle] [ERROR] /home/travis/build/JabRef/jabref/src/main/java/org/jabref/gui/journals/AbbreviateAction.java:31:5: Static variable definition in wrong order. [DeclarationOrder]
```

Could you please fix that, too? -- Just move it before the other variable declarations.

Brainsucker92:
Actually I thinking if it makes sense to rename the `executeAndReturn` method in the ExecutorService simply to `execute` which overloads the execute method and allows to execute tasks with returning values and without.

koppor:
@Brainsucker92 If I understand correctly, you are searching for a feature not available in Java: https://stackoverflow.com/a/2439798/873282

Brainsucker92:
It will work perfectly, bacause both parameter lists of the functions are differently. This allows to create two methods with different returning values. Just tested it in my IDE and everything is just fine.

koppor:
Sorry, one more thing:

And, is that possible without huge effort? 😇 

Brainsucker92:
For testing the abbreviation I have to have a fairly large sized BibTeX database, which I currently don't have, right? So where to get any random database in order to test these changes?

koppor:
@Brainsucker92 I have to confess that we don't have a huge bibtex file. Could write a generator? 😇 This would help us so much for testing huge databases.

Siedlerchr:
@koppor  We have some script in /scripts/bib-file-generator.rb
Generates 100,000 times the same entry

tobiasdiez:
I think we tortured @Brainsucker92 long enough 😈

koppor:
@brainsucker92 Thank you for the PR, the quick reactions and the patience! We would love to see more PRs from you! Especially your Python skills are interesteing. :innocent: our syncLang.py script needs updates cause of our migration to crowd in...

Brainsucker92:
@koppor Thanks a lot for your appreciation! Of course I'd like to continue contributing to this project in future. Unfortunately, my skills in python are pretty rusty because last time I actively coded in python was back then with py2.5. Thus I don't have that much experience with modern Python3.X programming. But if there are some minor changes to make I'll give it a shot to do so. I had a quck look at syncLang.py and code quality seems to be pretty good in the script.

Siedlerchr:
The sync lang python script is still 2.x because it's executed with jython which does not work with python 3.x (we learned that the hard way... 😯)

koppor:
@Brainsucker92 One has to think into the aims what the script does. I think, it currently does not handle escapes the way as [Crowdin](http://crowdin.com/) does. Maybe, you can run the script and check what it modifies and which errors it reports. See https://github.com/JabRef/jabref/issues/3573#issuecomment-357653588 for a discussion. You can follow up there in case you want to dive into this thing.

koppor:
@Brainsucker92 Could you write a short personal email so that I have your contact data?

Brainsucker92:
@koppor sure, just sent a mail to kopp.dev@gmail.com

PR #3724, opened 2018-02-15
Fix Some Codacy Issues

LinusDietz:
I thought it might be good to reduce the number of Codacy issues, so I started hunting them.
To keep these changes reviewable, I will open a new PR for the next batch.

Also I plan to adapt the codacy configuration to reduce false positives.


LinusDietz:
@lenhard well that's why I have added microcommits with the corresponding messages

lenhard:
Point taken :) You just don't see these messages when you go to the code page of the PR

PR #3733, opened 2018-02-17
Refactor BibDatabase Migrations

LinusDietz:
This is the followup for https://github.com/JabRef/jabref/pull/3658#discussion_r168779345.

I have moved the functionality from org.jabref.migrations to `org.jabref.logic.importer.migrations`.

The code is called from the UI now in `org.jabref.gui.importer.actions.OpenDatabaseAction.performPostOpenActions()`, where also some other similar actions are called.

Since the build is broken at the moment due to external resources, I'll have to re-check later if everything is actually fine here.
The general structure should be ready to review anyways.


stefan-kolb:
Why logic.importer.migrations? Sure it would not be in a wrong place, but we put it in a top-level package on purpose as this should be immediately evident that this is caused by JabRef design decisions and not hidden inside logic.importer where standard formats should be put. Not sure what others think.

Siedlerchr:
Agree with @stefan-kolb  What has a migration to do with import?
Top level logic.migrations should be the way to got

lenhard:
+1 for logic.migrations

src/main/java/org/jabref/logic/importer/migrations/FileLinksUpgradeWarning.java
```diff
@@ -1,4 +1,4 @@
-package org.jabref.migrations;
+package org.jabref.logic.importer.migrations;
```
tobiasdiez: This class has to many UI-dependencies to be in `logic`.

Siedlerchr:
The failing test is probably due to caching on travis and a result from #3713.
I cleared the cache in travis for master and this build. Hope that's enough.,
If not, delete it again: https://travis-ci.org/JabRef/jabref/caches

src/main/java/org/jabref/gui/importer/actions/MergeReviewIntoComment.java
```diff
@@ -0,0 +1,23 @@
+package org.jabref.gui.importer.actions;
+
+import org.jabref.gui.BasePanel;
+import org.jabref.logic.importer.ParserResult;
+import org.jabref.logic.migrations.MergeReviewIntoCommentMigration;
+
+public class MergeReviewIntoComment implements GUIPostOpenAction {
```
Siedlerchr: Maybe rename that with the action suffix so that it is better distingusishable from the other migration

src/main/java/org/jabref/gui/importer/actions/MergeReviewIntoComment.java
```diff
@@ -0,0 +1,23 @@
...
+    public boolean isActionNecessary(ParserResult parserResult) {
+        return MergeReviewIntoCommentMigration.needsMigration(parserResult);
+    }
+
+    @Override
+    public void performAction(BasePanel basePanel, ParserResult parserResult) {
+        MergeReviewIntoCommentMigration migration = new MergeReviewIntoCommentMigration();
+
+        migration.performMigration(parserResult);
+        if (new MergeReviewIntoCommentConfirmation(basePanel).askUserForMerge(MergeReviewIntoCommentMigration.collectConflicts(parserResult))) {
```
Siedlerchr: I would split that into assignment and declaration. It's unforunately not so easy distinguishable what what ist, because all is named very similar

PR #3536, opened 2017-12-15
[WIP] Add support for German months at the normalizer functionality

koppor:
My library exports BibTeX in a strange format (http://www2.informatik.uni-stuttgart.de/zdi/buecherei/NCSTRL_listings/privat/Kopp_Oliver.bibtex.html). I cannot get this fixed, so JabRef has to.

----

- [x] Change in CHANGELOG.md described
- [x] Tests created for changes
- [ ] <s>Screenshots added (for bigger UI changes)</s>
- [x] Manually tested changed features in running JabRef
- [ ] Check documentation status (Issue created for outdated help page at [help.jabref.org](https://github.com/JabRef/help.jabref.org/issues)?)
- [ ] If you changed the localization: Did you run `gradle localizationUpdate`?

PR #3734, opened 2018-02-17
Add parser for German months (#3536)

Eiswindyeti:
Changed the month parse that it can also parse German months now. 
This should solve issue #3536.
Have I missed something?

(For PE1718: Name: Johannes Preßmar)

----


CHANGELOG.md
```diff
@@ -34,6 +34,7 @@ The new default removes the linked file from the entry instead of deleting the f
 - The group editing window can now also be called by double-clicking the group to be edited. [koppor#277](https://github.com/koppor/jabref/issues/277)
 - The magnifier icon at the search shows the [search mode](https://help.jabref.org/en/Search#search-modes) again. [#3535](https://github.com/JabRef/jabref/issues/3535)
 - We added a new cleanup operation that replaces ligatures with their expanded form. [#3613](https://github.com/JabRef/jabref/issues/3613)
+- We added the function to parse German month names.
```
Siedlerchr: Coressponding issue?

Eiswindyeti: Sorry I forgot. Fixed it.

src/main/java/org/jabref/model/entry/Month.java
```diff
@@ -93,9 +95,31 @@
...
+	 * "Oktober" or in German short form such as "Okt"
+	 * 
+	 * @param value,
+	 *            a String that represents a month in German form
+	 * @return the corresponding month instance, empty if input is not in German
+	 *         form
+         * @author Johannes Preßmar
+	 */
+	private static Optional<Month> parseGermanShortMonth(String value) {
+		String[] shortMonths = new DateFormatSymbols(Locale.GERMAN).getShortMonths();
```
Siedlerchr: Better use the new java 8 Date & Time Api 

Construt a  DateTimeFormatter.ofPattern("MMM", Locale.DE) 
Use YearMonth.parse  and then you can get a Numeric value from that
https://docs.oracle.com/javase/8/docs/api/java/time/YearMonth.html#parse-java.lang.CharSequence-

src/main/java/org/jabref/model/entry/Month.java
```diff
@@ -93,9 +95,31 @@
             int number = Integer.parseInt(value);
             return Month.getMonthByNumber(number);
         } catch (NumberFormatException e) {
-            return Optional.empty();
+            return parseGermanShortMonth(testString);
```
Siedlerchr: Better move the method above in the parse method, so the parseGermanMonth will be executed when the others returned empty. Just like it it is deone with the getMonthByShortName

src/main/java/org/jabref/model/entry/Month.java
Eiswindyeti: I fixed it. Is it better like this?

src/main/java/org/jabref/model/entry/Month.java
Eiswindyeti: Sorry, I don't know exactly what you mean.

src/main/java/org/jabref/model/entry/Month.java
Siedlerchr: Yep! Looks better now

src/main/java/org/jabref/model/entry/Month.java
Siedlerchr: Okay, I will try to explain:
Move the method parseGermanShortMonth before the catch block. and after the getMonthByShortName:
```
 month= Month.parseGerman....
if(month.isPresent)
return month;
```
The end result is the same, but it looks more logical in the whole parse method:
1. We check if the value is empty
2. We try to get the normal short month
3. we try german short month
4. last try:  try into parse it as number 

Otherwise from a quick look I would see and think: Okay, if the integer parsing fails it implies it is automatically a German month. And that might not be the case. In reality the integer parsing maybe failed because we entered 9999

src/main/java/org/jabref/model/entry/Month.java
```diff
@@ -93,9 +97,28 @@
...
+    
+    /**
+	 * Parses a month having the string in German standard form such as
+	 * "Oktober" or in German short form such as "Okt"
+	 * 
+	 * @param value,
+	 *            a String that represents a month in German form
+	 * @return the corresponding month instance, empty if input is not in German
+	 *         form
+	 * @author Johannes Pre�mar
```
koppor: Please do not put `@author` tags. We acknowledge our authors in `AUTHORS`. See https://github.com/JabRef/jabref/blob/master/CONTRIBUTING.md#author-credits. Please ensure that your git configuration `user.name` and `user.email` is correct.

src/main/java/org/jabref/model/entry/Month.java
```diff
@@ -93,9 +97,28 @@
...
+	 */
+	private static Optional<Month> parseGermanShortMonth(String value) {
+		try {
+		YearMonth yearMonth = YearMonth.parse("1969-" + value, DateTimeFormatter.ofPattern("yyyy-MMM", Locale.GERMAN));
```
koppor: The indent seems to be wrong here. Use auto-format of your IDE. Ctrl+Alt+L in IntelliJ.

src/test/java/org/jabref/model/entry/MonthTest.java
```diff
@@ -100,4 +100,36 @@ public void parseReturnsEmptyOptionalForInvalidInput() {
     public void parseReturnsEmptyOptionalForEmptyInput() {
         assertEquals(Optional.empty(), Month.parse(""));
     }
+    
+    @Test
```
koppor: The indent is wrong here.

src/main/java/org/jabref/model/entry/Month.java
koppor: <s>The idea of https://github.com/JabRef/jabref/pull/3536 was to parse all available languages. Also Spanish, Greek, etc. Therefore, all available locales were fetched at https://github.com/JabRef/jabref/pull/3536/files#diff-3c5fdac594bc2e649536f5ecdcb481ccR122 and iterated on. Is this possible here, too - or does it clutter the code?</s>

After more thinking, I would say, we do not iterate on all locales and keep German only, because only the literature list by http://www2.informatik.uni-stuttgart.de/zdi/buecherei/NCSTRL_listings/FAK/index.html causes these issues and I am not aware of any other systems.

src/test/java/org/jabref/model/entry/MonthTest.java
```diff
@@ -100,4 +100,36 @@ public void parseReturnsEmptyOptionalForInvalidInput() {
...
+		assertEquals(Optional.of(Month.OCTOBER), Month.parse("Okt"));
+		assertEquals(Optional.of(Month.NOVEMBER), Month.parse("Nov"));
+		assertEquals(Optional.of(Month.DECEMBER), Month.parse("Dez"));
+	}
+
+	@Test
+	public void parseCorrectlyByFullNameGerman() {
+		assertEquals(Optional.of(Month.JANUARY), Month.parse("Januar"));
+		assertEquals(Optional.of(Month.FEBRUARY), Month.parse("Februar"));
+		assertEquals(Optional.of(Month.MARCH), Month.parse("M�rz"));
```
koppor: Please also add a test for `Maerz`, which is returned by the Stuttgart System: http://www2.informatik.uni-stuttgart.de/zdi/buecherei/NCSTRL_listings/FAK/index.html

src/main/java/org/jabref/model/entry/Month.java
Eiswindyeti: Thanks for the detailed explanation. I adjusted it.

src/main/java/org/jabref/model/entry/Month.java
Eiswindyeti: Okey I understand, but I have done that for PE1718, but the worksheet said so for authorize. I deleted it and will put my name directly in the Pull Request.

src/main/java/org/jabref/model/entry/Month.java
Eiswindyeti: I use Eclipse, hopefully Ctrl + Shift + F will have the same result.

src/test/java/org/jabref/model/entry/MonthTest.java
Eiswindyeti: I adjusted it.

src/main/java/org/jabref/model/entry/Month.java
```diff
@@ -88,33 +88,41 @@
         if (testString.length() > 3) {
             testString = testString.substring(0, 3);
         }
-        Optional<Month> month = Month.getMonthByShortName(testString);
```
koppor: The indent is wrong here. I have no clue, what is going wrong on your side. Do you use tabs instead of spaces?

src/test/java/org/jabref/model/entry/MonthTest.java
Eiswindyeti: I also put in an if-statement for checking for `Maerz` in the method `parseGermanShortMonth`, because I was not sure if the Locale checks for this.

src/main/java/org/jabref/model/entry/Month.java
Eiswindyeti: I don't know what went wrong. I used the auto formatter of Eclipse, so it could be the tabs. Now it looks better after replacing the tabs with spaces by hand. I should definetly change the IDE if this is a common problem.

src/main/java/org/jabref/model/entry/Month.java
```diff
@@ -97,6 +105,29 @@
...
+     * Parses a month having the string in German standard form such as
+     * "Oktober" or in German short form such as "Okt"
+     * 
+     * @param value,
+     *            a String that represents a month in German form
+     * @return the corresponding month instance, empty if input is not in German
+     *         form
+     */
+	private static Optional<Month> parseGermanShortMonth(String value) {
+        if ("Mae".equals(value)) {
```
stefan-kolb: What if it is "mae" in lowercase?

src/test/java/org/jabref/model/entry/MonthTest.java
```diff
@@ -100,4 +100,38 @@ public void parseReturnsEmptyOptionalForInvalidInput() {
     public void parseReturnsEmptyOptionalForEmptyInput() {
         assertEquals(Optional.empty(), Month.parse(""));
     }
+    
+    @Test
+    public void parseCorrectlyByShortNameGerman() {
+        assertEquals(Optional.of(Month.JANUARY), Month.parse("Jan"));
```
stefan-kolb: Also add tests for lowercase?

src/main/java/org/jabref/model/entry/Month.java
```diff
@@ -74,28 +78,55 @@
...
+	 * "Oktober" or in German short form such as "Okt"
+	 * 
+	 * @param value,
+	 *            a String that represents a month in German form
+	 * @return the corresponding month instance, empty if input is not in German
+	 *         form
+	 */
+	private static Optional<Month> parseGermanShortMonth(String value) {
+		if ("Mae".equals(value)) {
+			return Month.getMonthByNumber(3);
```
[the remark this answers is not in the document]
Eiswindyeti: The test is then red. So I leave the if-statement in there?

koppor:
equalsIgnoreCase?

src/main/java/org/jabref/model/entry/Month.java
koppor: The implementation should use `equalsIgnoreCase`.

src/test/java/org/jabref/model/entry/MonthTest.java
Siedlerchr: Well, then just copy the test cases to a new method and then just call month. Parse with the lowercase name
And execute the tests then

src/main/java/org/jabref/model/entry/Month.java
```diff
@@ -85,9 +89,13 @@
             testString = testString.substring(0, 3);
         }
         Optional<Month> month = Month.getMonthByShortName(testString);
+        Optional<Month> monthGerman = Month.parseGermanShortMonth(testString);
```
koppor: Please move that after line 95. the German month only has to be determined if there is no English month.

src/main/java/org/jabref/model/entry/Month.java
koppor: Please fix indent.

src/test/java/org/jabref/model/entry/MonthTest.java
Eiswindyeti: Tests were added.

src/main/java/org/jabref/model/entry/Month.java
Eiswindyeti: Done.

src/main/java/org/jabref/model/entry/Month.java
```diff
@@ -115,6 +117,9 @@
      *         form
      */
 	private static Optional<Month> parseGermanShortMonth(String value) {
+        // support for lowercase German month
+        value = value.substring(0,1).toUpperCase() + value.substring(1);
```
koppor: Please remove this and the line above. Just use [equalsIgnoreCase](https://www.tutorialspoint.com/java/java_string_equalsignorecase.htm).

Eiswindyeti: Now no German months are accepted with lowercase, just `mae`.

koppor: I overlooked this case. I think, you have to change the if then to following

    if (("Mae".equalsIgnoreCase(value)) || ("Maerz".equalsIgnoreCase(value))) {

Please also ensure that a test case for `Maerz` is in there. If it was already, then I don't get why that code still worked.

koppor:
Please merge JabRefs' branch `master` (should be available at `upstream/master`) to resolve conflicts in CHANGELOG.md

src/main/java/org/jabref/model/entry/Month.java
tobiasdiez: The dateformatter is case-sensitive by default. (Adapting) the following code should work:
````java
DateTimeFormatter formatter = new DateTimeFormatterBuilder()
    .parseCaseInsensitive()
    .appendPattern("yyyy-MMM")
    .toFormatter(Locale.US);
````

lenhard:
The tests are still failing. From looking at the code, I think that "Mär" is not parsed correctly. Could you please investigate this?

Once all tests are passing on travis, I think we can merge this.

koppor:
I'll merge into `wip`, finish the PR and merge it into `master`.

koppor:
I fixed the issues at https://github.com/JabRef/jabref/pull/3742

1. The java files had the wrong encoding (UTF8 was not used)
2. The implementation for "mär" and "märz" was missing.

PR #3735, opened 2018-02-17
Convert Strings Dialog to javafx

Siedlerchr:
Convert Strings Dialog to javafx

I decided to take a shot at the String dialog.  But I have some questions about the **String** handling in the Database class. I know that I can define some user defined strings as kind of template and then reuse them in my bib entries.

The only thing which does not look that good is the position of the Validation error icon.
Needs perhaps some CSS tweaking, but for the moment not important.

- [x]  Store Strings back in DB
- [x] Validate Strings and label 

grafik

----


tobiasdiez:
Nice that you want the dialog to JavaFX. However, I'm not sure what questions you now have.

Siedlerchr:
@tobiasdiez I was wondering why the BibtexString is stored in a Map<String, BibTexString  in the databse.
The key of the Map is some generated id and is also stored in the BibTexString itself.
So I see no reason why the list of BibTexStrings is not a simple list or set.

https://github.com/JabRef/jabref/blob/cf8e6aac8263ce30eb2b042c8aacf2d88bbac4c8/src/main/java/org/jabref/model/database/BibDatabase.java#L278-L288

tobiasdiez:
This is probably for performance reasons (hashmap provides a quicker lookup than to transverse the list).

Siedlerchr:
I now tried around and got the Validation running somehow, but it doesn't really work
I just used an EmptyValidator, but although I have entered text, it is shown as "red" 
grafik
@tobiasdiez  Any idea how I have to configure that? I looked at the entry editor validation, but that is a bit different as always only one single property is validated

src/main/java/org/jabref/gui/strings/StringDialogController.java
```diff
@@ -46,17 +53,49 @@ private void initialize() {
...
+        });
+
         colLabel.setOnEditCommit(
                 (CellEditEvent<StringViewModel, String> t) -> {
-                    t.getTableView().getItems().get(
-                            t.getTablePosition().getRow()).setLabel(t.getNewValue());
+                    t.getTableView()
+                            .getItems()
+                            .get(
+                                    t.getTablePosition().getRow())
```
tobiasdiez: `t.getRowValue` is easier (and please rename `t` to something more descriptive.

tobiasdiez:
I strongly recommend to use the `mvvmfx.validation` framework instead of directly using controlsfx. The latter has the problem that UI components are mixed with validation logic. The `mvvmfx` has a good documentation at https://github.com/sialcasa/mvvmFX/wiki/Validation#validator.

Siedlerchr:
Okay, yes, that seems pretty streightfoward, except that I have now the problem:
I have my DialogViewModel and  the StringViewModel which has just the content properties. 
Should I merge this two ViewModels into one or is there another solution?
I think I did it that way in my Sharelatex PR, but with the consequence that I had all stuff double on each dialog opening

tobiasdiez:
Two view models are what you need here. One for the whole dialog (contains a list of all strings, the composite validation status, methods for load/save) and one for each string (determines how the string should be displayed and its validation status). The validation of the per-string view model should be bound to the text field similar to how you did it before.

Siedlerchr:
@tobiasdiez  Do you have an idea why the Validation error/warning icon is not showing? the validation is triggered, but it doesn't display anything

src/main/java/org/jabref/gui/strings/StringDialogController.java
```diff
@@ -0,0 +1,122 @@
...
+            return cell;
+
+        });
+
+        colContent.setCellFactory(column -> {
+
+            TextFieldTableCell<StringViewModel, String> cell = new TextFieldTableCell<>(new DefaultStringConverter());
+            column.setCellValueFactory(cellData -> {
+
+                visualizer.initVisualization(cellData.getValue().contentValidation(), cell);
```
tobiasdiez: Two ideas:
1. Is this code called when you enter data in the text field? If you hit enter and leave the edit mode?
2. Is the visualizer correctly installed (i.e. debug the `initVisualization` code). Maybe it is expecting the text field and not a cell...

Siedlerchr: Yes, I debugged it and it's called everytime. I also debugged through the whole code from controlsfx and have no idea what might be wrong.
I then thought it might be that the css class is missing, so I copied the stuff from the entry editor css, but still does not work.

The Textfield TableCell was working previously with the default settings after I registered the ValueExtractor for controlsfx. But this is no longer used and hasn't any effect.

Siedlerchr:
As the validation still does not show up I created an issue:
https://github.com/sialcasa/mvvmFX/issues/546

Siedlerchr:
Got it working now, not perfect, but at least it works. Currently it works only after you hit enter

grafik

tobiasdiez:
Any update on this? Already looks quite good.

Siedlerchr:
I will finish that soon. The most work is already done. currently a bit busy but still on my todo list

Siedlerchr:
@tobiasdiez  I reworked the dialog a bit, but now I have the problem that the OK an cancel button can't be clicked in that way that the dialog does not close.  I tried around and even uncommented the init Visualization. But to no effect.
I have a result converter setup, but does not work. Any idea what I did wrong?

Siedlerchr:
Found the cause, it was a wrong view loading code.

Siedlerchr:
I finally solved the validation @tobiasdiez , so this dialog is now ready for review

src/main/java/org/jabref/gui/strings/StringDialogViewModel.java
```diff
@@ -0,0 +1,85 @@
...
+        this.db = bibDatabase;
+        addAllStringsFromDB();
+
+    }
+
+    private void addAllStringsFromDB() {
+
+        Set<StringViewModel> strings = db.getStringKeySet()
+                                         .stream()
+                                         .map(string -> db.getString(string))
```
tobiasdiez: Why not directly get a collection of all strings from the database (is there a method like `getStringValueSet`? Pesonally, I would like to have only a method that returns the strings in the form of a list of `BibtexString`s)

src/main/java/org/jabref/gui/strings/StringDialogViewModel.java
```diff
@@ -0,0 +1,85 @@
...
+        allStrings.addAll(strings);
+
+    }
+
+    public ListProperty<StringViewModel> allStringsProperty() {
+        return this.allStrings;
+    }
+
+    public void addNewString() {
+        allStrings.add(new StringViewModel(newStringLabelProperty.getValue(), "New Content"));
```
tobiasdiez: Localized? replace by the non-empty message below?

src/main/java/org/jabref/gui/strings/StringDialogViewModel.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    public void removeString() {
+        allStrings.remove(selectedItemProperty.getValue());
+    }
+
+    private StringViewModel convertFromBibTexString(BibtexString bibtexString) {
```
tobiasdiez: move to `StringViewModel` or even better just create a constructor that admits a `BibtexString`.

src/main/java/org/jabref/gui/strings/StringDialogViewModel.java
```diff
@@ -0,0 +1,85 @@
...
+    public ObjectProperty<StringViewModel> seletedItemProperty() {
+        return this.selectedItemProperty;
+    }
+
+    public StringProperty newStringLabelProperty() {
+        return this.newStringLabelProperty;
+    }
+
+    public void save() {
+        for (StringViewModel model : allStrings) {
```
tobiasdiez: move this to a new method `BibDatabase.setStrings(Collection<BibtexString>))` ?

src/main/java/org/jabref/gui/strings/StringViewModel.java
```diff
@@ -0,0 +1,105 @@
...
+    private final StringProperty label = new SimpleStringProperty();
+    private final StringProperty content = new SimpleStringProperty();
+
+    private final Validator labelValidator;
+    private final Validator contentValidator;
+    private final CompositeValidator combinedValidator;
+
+    private final Function<String, ValidationMessage> labelValidatorFunction = input -> {
+        if (input == null) {
+            return ValidationMessage.error("May not be null");
```
tobiasdiez: localized?

src/main/java/org/jabref/gui/strings/StringViewModel.java
```diff
@@ -0,0 +1,105 @@
...
+    private final static Pattern IS_NUMBER = Pattern.compile("-?\\d+(\\.\\d+)?");
+
+    private final StringProperty label = new SimpleStringProperty();
+    private final StringProperty content = new SimpleStringProperty();
+
+    private final Validator labelValidator;
+    private final Validator contentValidator;
+    private final CompositeValidator combinedValidator;
+
+    private final Function<String, ValidationMessage> labelValidatorFunction = input -> {
```
tobiasdiez: Any reason why you define this methods as `Function` variables instead of directly as static methods?

src/main/java/org/jabref/gui/strings/StringViewModel.java
```diff
@@ -0,0 +1,105 @@
...
+            return ValidationMessage.error("May not be null");
+        } else if (input.trim().isEmpty()) {
+            return ValidationMessage.error(Localization.lang("Must not be empty!"));
+        } else {
+            return null; // everything is ok
+        }
+    };
+
+    public StringViewModel(String label, String content) {
+
```
tobiasdiez: empty lines... you really like them this time? ;-)

src/main/java/org/jabref/gui/undo/UndoableInsertString.java
```diff
@@ -15,14 +14,10 @@
     private static final Logger LOGGER = LoggerFactory.getLogger(UndoableInsertString.class);
 
     private final BibDatabase base;
-    private final BasePanel panel;
     private final BibtexString string;
 
-
-    public UndoableInsertString(BasePanel panel, BibDatabase base,
-            BibtexString string) {
+    public UndoableInsertString(BibDatabase base, BibtexString string) {
```
tobiasdiez: also move to the `string` package?

src/main/java/org/jabref/gui/undo/UndoableInsertString.java
```diff
@@ -37,7 +32,7 @@ public void undo() {
 
         // Revert the change.
         base.removeString(string.getId());
-        panel.updateStringDialog();
+
```
tobiasdiez: ;-)

src/main/java/org/jabref/gui/strings/StringDialogViewModel.java
Siedlerchr: Not that easy, it's a HashMap which explicitly checks the key before inserting and generates a new id.

  ```
  /**
     * Inserts a Bibtex String.
     */
    public synchronized void addString(BibtexString string) throws KeyCollisionException {
        if (hasStringLabel(string.getName())) {
            throw new KeyCollisionException("A string with that label already exists");
        }

        if (bibtexStrings.containsKey(string.getId())) {
            throw new KeyCollisionException("Duplicate BibTeX string id.");
        }

        bibtexStrings.put(string.getId(), string);
    }


```

src/main/java/org/jabref/gui/undo/UndoableInsertString.java
Siedlerchr: I was thinking about this, but decided against it, as in the undo package all other undo methods are also (UndoableRemoveEntry etc)

src/main/java/org/jabref/gui/strings/StringDialogViewModel.java
tobiasdiez: I meant you should simply refactor the code and move (most) of it to a new method in `BibDatabase` (not re-implement it in another way).

Siedlerchr:
Addressed all remaininig issues so far. Also extracted the column setup with validation to an extra Factory.
As explained earlier, decided not to move the undo stuff as all undo things are in one package now.
And moving the viewModel stuff adding the strings does not really make sense how the internal class is now architectured

Siedlerchr:
@tobiasdiez  Moved the code to the bibdatabase for adding new strings from a collection

PR #3739, opened 2018-02-18
[WIP] Toggle group intersection/union solved

### LINKED ISSUE #3269: 4.0 Groups: fast access to View options

stoever:
Added new menu entry with the needed action.
The action gets and sets the group intersection/union preferences.

Fixes #3269

PE1_Blatt_5
Aufgabe 7G

Matrikelnummer: 3325847
Name: Tobias Stöver

----

### DIFF
--- a/src/main/java/org/jabref/gui/BasePanel.java
+++ b/src/main/java/org/jabref/gui/BasePanel.java
@@ -750,6 +750,13 @@ public class BasePanel extends JPanel implements ClipboardOwner {
         actions.put(Actions.REMOVE_FROM_GROUP, new GroupAddRemoveDialog(this, false, false));
         actions.put(Actions.MOVE_TO_GROUP, new GroupAddRemoveDialog(this, true, true));
 
+        actions.put(Actions.INTERSECTION_PREF_GROUP, (BaseAction) () -> {
+            System.out.println("blubb");
+            boolean state = Globals.prefs.getBoolean(JabRefPreferences.GROUP_INTERSECT_SELECTIONS);
+            state = !state;
+            Globals.prefs.putBoolean(JabRefPreferences.GROUP_INTERSECT_SELECTIONS, state);
+        });
+
         actions.put(Actions.DOWNLOAD_FULL_TEXT, new FindFullTextAction(this));
     }
 
--- a/src/main/java/org/jabref/gui/BasePanel.java
+++ b/src/main/java/org/jabref/gui/BasePanel.java
@@ -750,11 +750,13 @@ public class BasePanel extends JPanel implements ClipboardOwner {
         actions.put(Actions.REMOVE_FROM_GROUP, new GroupAddRemoveDialog(this, false, false));
         actions.put(Actions.MOVE_TO_GROUP, new GroupAddRemoveDialog(this, true, true));
 
+        /**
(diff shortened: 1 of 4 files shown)


src/main/java/org/jabref/gui/BasePanel.java
```diff
@@ -750,6 +750,13 @@ public void update() {
         actions.put(Actions.REMOVE_FROM_GROUP, new GroupAddRemoveDialog(this, false, false));
         actions.put(Actions.MOVE_TO_GROUP, new GroupAddRemoveDialog(this, true, true));
 
+        actions.put(Actions.INTERSECTION_PREF_GROUP, (BaseAction) () -> {
+            System.out.println("blubb");
```
Siedlerchr: remove

src/main/java/org/jabref/gui/BasePanel.java
```diff
@@ -750,6 +750,13 @@ public void update() {
...
+            boolean state = Globals.prefs.getBoolean(JabRefPreferences.GROUP_INTERSECT_SELECTIONS);
+            state = !state;
```
Siedlerchr: This code is hard to understand and not clear on the first sight.
Better create two variables: e.g. union and intersection and then store the corresponding value

src/main/java/org/jabref/gui/preftabs/GroupsPrefsTab.java
```diff
@@ -115,6 +115,7 @@ public void setValues() {
         keywordSeparator.setText(prefs.get(JabRefPreferences.KEYWORD_SEPARATOR));
         autoAssignGroup.setSelected(prefs.getBoolean(JabRefPreferences.AUTO_ASSIGN_GROUP));
         multiSelectionModeIntersection.setSelected(prefs.getBoolean(JabRefPreferences.GROUP_INTERSECT_SELECTIONS));
+        multiSelectionModeUnion.setSelected(!prefs.getBoolean(JabRefPreferences.GROUP_INTERSECT_SELECTIONS));
```
Siedlerchr: this is really not a good idea. If you quickly go over the code you most certainly will oerlook the not sign.
The goal is to have clear code which is ideally understanding at first sight.
https://8thlight.com/blog/dariusz-pasciak/2015/05/28/alternatives-to-boolean-parameters.html

src/main/java/org/jabref/gui/BasePanel.java
Siedlerchr: Ideally you would have an enum with two states, e.g. GroupState... but in this case this might be an overkill.

src/main/java/org/jabref/gui/BasePanel.java
```diff
@@ -750,6 +750,15 @@ public void update() {
         actions.put(Actions.REMOVE_FROM_GROUP, new GroupAddRemoveDialog(this, false, false));
         actions.put(Actions.MOVE_TO_GROUP, new GroupAddRemoveDialog(this, true, true));
 
+        /**
```
Siedlerchr: Remove your author comments. 
We keep our authors in a separate file: https://github.com/JabRef/jabref/blob/master/AUTHORS

koppor:
@Siedlerchr please decide whether you take over or just close the PR.

PR #3740, opened 2018-02-19
Inline loc.org xsds

### LINKED ISSUE #3737: JabRef xjc gradle task fails due to outage of the library of congress
```
> Task :xjc
[ant:xjc] [WARNING] schema_reference.4: Failed to read schema document 'http://www.loc.gov/mods/xml.xsd', because 1) could not find the document; 2) the document could not be read; 3) the root element of the document is not <xsd:schema>.
```

see https://www.loc.gov/

> Websites Down Feb. 16-19; Reading Rooms and Library Shop Closed Feb. 17 and 19
Due to a planned power outage, most of the Library’s public websites (loc.gov and others) will be unavailable from 7 p.m. ET, Friday, Feb. 16 through Monday, Feb. 19. The Copyright online registration system will be unavailable beginning 5 p.m., Thursday, Feb. 15 throughout the weekend.

JabRef holidays! (;

side note: we should fix such a dependency on a website being available

stefan-kolb:
Fixes #3737 

Not sure if this works and is the way to go.
If we need to update the mods schema we need to alter the paths again.
Also, I don't know if we may inline this here.


LinusDietz:
We have combined our two efforts in inlining and updating the MODS XSD files.

This PR fixes https://github.com/JabRef/jabref/issues/3737 and https://github.com/koppor/jabref/issues/307

PR #3731, opened 2018-02-17
[WIP] Added Select all and Deselect all buttons to the "Review changes" dialog

N4tus:
Added "Select all" and "Deselect all" buttons to the "Review changes" dialog. The "Select all" button is the fix for https://github.com/koppor/jabref/issues/280 made by janek greif. The "Deselect all" button is an addition to that issue.

----

- [x] Change in CHANGELOG.md described
- [ ] Tests created for changes
- [ ] Screenshots added (for bigger UI changes)
- [x] Manually tested changed features in running JabRef
- [ ] Check documentation status (Issue created for outdated help page at [help.jabref.org](https://github.com/JabRef/help.jabref.org/issues)?)

PR #3742, opened 2018-02-19
[WIP] Add Accept All Changes to External Changes Dialog 

danielabajirov:
Fix for issue https://github.com/koppor/jabref/issues/280

Added a new ChoiceBox "Accept all changes" to the JabRefChangeDisplay that allows to select all the change and modify the status of the ChoiceBox from false to true so that all the cb now are selected when the computation is completed.

----

### DIFF
--- a/src/main/java/org/jabref/gui/collab/ChangeDisplayDialog.java
+++ b/src/main/java/org/jabref/gui/collab/ChangeDisplayDialog.java
@@ -30,6 +30,7 @@ class ChangeDisplayDialog extends JabRefDialog implements TreeSelectionListener 
     private final JTree tree;
     private final JPanel infoPanel = new JPanel();
     private final JCheckBox cb = new JCheckBox(Localization.lang("Accept change"));
+    private final JCheckBox cb2 = new JCheckBox(Localization.lang("Accept all change"));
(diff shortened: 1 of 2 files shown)


src/main/java/org/jabref/gui/collab/ChangeDisplayDialog.java
```diff
@@ -30,6 +30,7 @@
     private final JTree tree;
     private final JPanel infoPanel = new JPanel();
     private final JCheckBox cb = new JCheckBox(Localization.lang("Accept change"));
+    private final JCheckBox cb2 = new JCheckBox(Localization.lang("Accept all change"));
```
stefan-kolb: Better: "Accept all changes"?!

danielabajirov: Oh right ahah my bad.

Siedlerchr:
Changelog? And it looks like this is the same PR as  #3731

danielabajirov:
I didnt read it so the #3731 but yes the idea is the same now that i gave a look.

koppor:
Could you add a screenshot so that we can decide which one looks better? Codewise, this is beter than #3731.

danielabajirov:
Yep but after i fix a couple of things because i read the #3731 and need to implement the "deselect all changes" ChoiceBox and maybe i had an other idea. I'll post it as soon as i finish.

PR #3743, opened 2018-02-19
[WIP] Fix syncLangPy script 

### LINKED ISSUE #3573: Gradle task localizationUpdate does not work

Brainsucker92:
Fixes #3573 

As already mentioned in my comment here: https://github.com/JabRef/jabref/issues/3573#issuecomment-366490651
there was a logical issue hidden within the merging process when synchronizing the files. I fixed this issue and added/changed a couple of other things.

----

- [x] Manually tested changed features ~~in running JabRef~~

### DIFF
--- a/scripts/syncLang.py
+++ b/scripts/syncLang.py
@@ -26,6 +26,8 @@ try:
         requires Python 3.4 or higher
         """
         
+        BASE_DIRECTORY_NAME = 'jabref'
+        
         @staticmethod
         def getJabRefBaseDirectory():
             """
@@ -33,11 +35,11 @@ try:
             :return the path to JabRef's base directory as pathlib.Path object.
             """
             cwd = pathlib.Path.cwd()
-            if cwd.name == 'jabref':
+            if cwd.name == PathFinder.BASE_DIRECTORY_NAME:
                 return cwd
             
             for parent in cwd.parents:
-                if parent.name == 'jabref':
+                if parent.name == PathFinder.BASE_DIRECTORY_NAME:
                     return parent
             # TODO What to do if base directory could not be found?
     
--- a/scripts/syncLang.py
+++ b/scripts/syncLang.py
@@ -40,7 +40,7 @@ try:
             :return the path to JabRef's base directory as pathlib.Path object.
             """
             if pathlib.Path.cwd().name == 'jabref':
-                return pathlib.Path.cwd
+                return pathlib.Path.cwd()
             
             scriptLocation = PathFinder.getScriptLocation()
             for parent in scriptLocation.parents:
--- a/scripts/syncLang.py
+++ b/scripts/syncLang.py
@@ -456,9 +456,10 @@ class SyncLang:
             for line in main_lines:
                 key = main_keys.key_from_line(line)
                 if key is not None:
-                    # Do not write empty keys
-                    if keys[key] != "":
-                        other_lines_to_write.append(Keys.format_key_and_value(key=key, value=keys[key]) + "\n")
+                    if keys.has_key(key):
+                        # Do not write empty keys
+                        if keys[key] != "":
+                            other_lines_to_write.append(Keys.format_key_and_value(key=key, value=keys[key]) + "\n")
                 else:
                     other_lines_to_write.append(line)
 
--- a/scripts/syncLang.py
+++ b/scripts/syncLang.py
@@ -445,9 +445,9 @@ class SyncLang:
             num_keys_missing = len(keys_missing)
             num_keys_obsolete = len(keys_obsolete)
 
-            for missing_key in keys_missing:
+            # for missing_key in keys_missing:
                 # Missing keys are added with main translation by default.
-                keys[missing_key] = main_keys_dict[missing_key]
+                # keys[missing_key] = main_keys_dict[missing_key]
 
             for obsolete_key in keys_obsolete:
                 del keys[obsolete_key]
--- a/scripts/syncLang.py
+++ b/scripts/syncLang.py
@@ -315,6 +315,7 @@ class SyncLang:
     def __update_properties(self, main_property_file, other_property_files):
         main_lines = self.__read_file_as_lines(filename=main_property_file)
         main_keys = Keys(main_lines)
+        main_keys_dict = main_keys.translations_as_dict()
 
         main_duplicates = main_keys.duplicates()
         num_main_duplicates = len(main_duplicates)
@@ -350,7 +351,8 @@ class SyncLang:
 
             for missing_key in keys_missing:
                 logging.debug("Adding missing Key: " + missing_key)
-                keys[missing_key] = ""
+                # Missing keys are added with english translation by default.
+                keys[missing_key] = main_keys_dict[missing_key]
 
             for obsolete_key in keys_obsolete:
                 logging.debug("Deleting obsolete Key: " + obsolete_key)
@@ -361,8 +363,8 @@ class SyncLang:
                 key = main_keys.key_from_line(line)
                 if key is not None:
                     # Do not write empty keys
-                    # if keys[key] != "":
-                    other_lines_to_write.append(Keys.format_key_and_value(key=key, value=keys[key]) + "\n")
+                    if keys[key] != "":
+                        other_lines_to_write.append(Keys.format_key_and_value(key=key, value=keys[key]) + "\n")
                 else:
                     other_lines_to_write.append(line)
 
--- a/scripts/syncLang.py
+++ b/scripts/syncLang.py
@@ -27,25 +27,19 @@ try:
         """
         
         @staticmethod
-        def getScriptLocation():
-            """
-            :return the path this script is currently located as pathlib.Path object.
-            """
-            return pathlib.Path(__file__)
-        
-        @staticmethod
         def getJabRefBaseDirectory():
             """
             Searches the script's path backwards until it finds the matching base directory.
             :return the path to JabRef's base directory as pathlib.Path object.
             """
-            if pathlib.Path.cwd().name == 'jabref':
-                return pathlib.Path.cwd()
+            cwd = pathlib.Path.cwd()
+            if cwd.name == 'jabref':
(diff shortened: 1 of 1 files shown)


Siedlerchr:
Thanks a lot!

koppor:
## Escapings

Look at this:

grafik

English file:

    Executing\ command\ \"%0\"...=Executing command \"%0\"...
    The\ marked\ area\ does\ not\ contain\ any\ legible\ text!=The marked area does not contain any legible text!

I think, one has to rewrite the English file to NOT escape `!` and `"`.

## Empty Translations

Empty strings must not be added, because of the new Crowdin workflow.

Alternative: Add English as default  (and not the empty ones)

LinusDietz:
Does this run with jython?

Brainsucker92:
All of the code related to python 3 is within a try block. This keeps it's compatibility to python2, so it should be no problem to run it with jython.

By the way there is already a repository for a jython version which is compatible to python 3.5 here:
https://github.com/jython/jython3
But since it's not a stable version at all, I wouldn't recommend switching to jython3.

Brainsucker92:
@koppor In order to fix the issue with the escapings, I think I have to rework the logging output first. The output is currently not helpful at all, which is just printing a lot of unreadable spam into the console. Once this is fixed it will be much easier to see which keys are causing some trouble during the merging process.

Brainsucker92:
Quick update:
Today I've been working on the command-line interface using the [argparse ](https://docs.python.org/2/library/argparse.html) library. This took me way longer than I expected. Next step will be to rework the logging output, as mentioned before.

Siedlerchr:
That comes right in time for checking the localization on the maintable beta branch!

koppor:
- [ ] Would it be possible that `print missing` can also go through a complete directory and check all files at once? As default, it should use the directory `src/main/resources/l10n` so that `print missing` works without any additional options.

- [ ] It should output if nothing is found -- oh, it does not output if there was something. The `status` output missing keys

    > python scripts\syncLang.py print missing -f src\main\resources\l10n\JabRef_de.properties
    INFO:root:Unable to use PathFinder class.

Otherwise, it looks good to me. Have to change the Crowdin config then. @JabRef/developers JabRef will now contain values (the english text!) for each translation in the non-English files. This is supported by Crowdin.

Brainsucker92:
Ok, I changed the print command. The file argument is no longer required for all subcommands (missing, obsolete, duplicates).

You can now use `syncLang.py print missing`. But keep in mind, this might be a little spammy

tobiasdiez:
@koppor I have the vague memory that we already discussed the problematic of having English as the default value for the other language and decided against this option because it would make the established way of editing the language files directly more difficult. Not sure how relevant this still is because every translation contribution now seems to come through crowdin.

koppor:
@tobiasdiez I think, it was only because we could not recognize easily with the script that the keys have been translated. For instance, following line

    OK = OK

Is this translated or not? (In German yes, in Chinese not.)

IMHO this is the edge case. With the change that the English values come into the other language files we enable the guys NOT working with Crowdin to be able to translate. For instance, @josephace135 wants to use GitHub to get rewarded via https://utopian.io.

Brainsucker92:
@koppor Yes, those test cases are possible for sure, by using the [unittest](https://docs.python.org/2/library/unittest.html) library.
I might add another python file with sone test cases later.

By the way, I will be at Stuttgart University at 3:15 pm today

koppor:
@Brainsucker92 forget about that test cases.

We keep the non-translated strings removed. Reason: IntelliJ can perfectly deal with it:

grafik

koppor:
@Brainsucker92 I am sorry to say, but could you remove the functionality to add missing keys in the other language files? Reporting etc. is OK, but not the addition...

Reason: I find non-existing lines better than existing lines with strings from other languages.

Brainsucker92:
Sure, I can do that. But i think i'll do some other changes as well. The  `__update_properties` method seems to be way to overloaded by things that should be moved to external methods.

Brainsucker92:
Sorry for the delay. I was working on another improvement of the script but due to upcoming exams I think I won't be able to complete the improvement any time soon. So I simply uploaded the change you requested.

Siedlerchr:
Hi, I just tried to run your script from gradle under Windows and got (apart for the missing python 3 exception):
I tried to run that on a new empty property file for #3776 
```
$ ./gradlew localizationUpdateExtended
Task :jmh Last added: null
:localizationUpdateExtendedINFO:root:Unable to use PathFinder class.
Traceback (most recent call last):
  File "E:\workspace\JabRef\jabref\scripts\syncLang.py", line 660, in <module>
    main()
  File "E:\workspace\JabRef\jabref\scripts\syncLang.py", line 656, in main
    parsed_args.func(parsed_args)
  File "E:\workspace\JabRef\jabref\scripts\syncLang.py", line 594, in update_command
    syncer.update_properties()
  File "E:\workspace\JabRef\jabref\scripts\syncLang.py", line 296, in update_properties
    self.__update_jabref_properties()
  File "E:\workspace\JabRef\jabref\scripts\syncLang.py", line 303, in _SyncLang__update_jabref_properties
    self.__update_properties(main_property_file=self.main_jabref_preferences, other_property_files=self.__other_jabref_properties())
  File "E:\workspace\JabRef\jabref\scripts\syncLang.py", line 460, in _SyncLang__update_properties
    if keys[key] != "":
KeyError: u'What\\ is\\ Mr.\\ DLib?'
 FAILED
```

Brainsucker92:
Thanks for the hint. Should be fixed now, try again.

Siedlerchr:
I fixed it myway (I switched to my Pyhton 3 installation)

```python
         if key is not None:
                    # Do not write empty keys
                    if keys.get(key) != "" and keys.get(key) is not None:
                        other_lines_to_write.append(Keys.format_key_and_value(key=key, value=keys[key]) + "\n")
                else:
```

That worked for me. However, no keys are copied to my empty properties. Only newlines and comments.
```
INFO:root:Processing file 'Menu_tl.properties' with 0 Keys
INFO:root:	added 110 missing keys
```

Just create a new empty property file:
Menu_tl.properties
JabRef_tl.properties

=> No keys are inserted


https://stackoverflow.com/questions/33727149/dict-object-has-no-attribute-has-key

Brainsucker92:
Yes, because that's what @koppor requested to change. 

> could you remove the functionality to add missing keys in the other language files?

That's what I did.

Siedlerchr:
Okay: I have an idea. After I found out that missing properties make JabRef crash (Those keys that are used in the fxml files) it would be wise to either auto add these ones. Or you just could add an additional gradle task to add the missing keys.

Brainsucker92:
To be honest: In my opinion we should seriously consider rewriting the whole script to java, if there is no specific reason not to do this. The more changes are made on the script and the more functionality we add, the more complex the script becomes. And at some point python is no longer a viable language because it's primarily well suited for small and simple code which is slowly but surely overextended at some point. Rewriting into Java would also solve any compatibility problems with py2.x and py3, and also get rid of any jython requirement for the whole JabRef project (if there is no other python code in use).

Siedlerchr:
Yes, that sounds like a valid reasonable thing. I think for the moment we can keep it as is and we should discuss this in the next devcall.

koppor:
@tobiasdiez Is it really true that JabRef crashes due to missing keys? I thought, you wrote a comment at some other place that this is not really the case and this can be fixed. I would like to keep the files clean. Now that [utopian](https://utopian.io/) allows for having [crowdin](http://crowdin.com/) as platform to get [Steem Coins](https://bitcoinblog.de/2016/07/19/steem-die-kryptowaehrung-die-nicht-durch-mining-sondern-durch-content-entsteht/), I would say, we do not need to change our workflow and let the language files be lean and clean. Especially because of the issue `OK=OK`. Is this translated or not (see https://github.com/JabRef/jabref/pull/3743#issuecomment-367748764)

koppor:
I tried to collect the options using [MADR 2.0.1](https://adr.github.io/madr/)

# Only translated strings in language file

## Context and Problem Statement

JabRef has translation files `JabRef_it.properties`, ...
There are translated and unstranslated strings.
Which ones should be in the translation file?

## Decision Drivers

* Translators should find new strings to translate easily
* New strings to translate should be written into `JabRef_en.properties` to enable translation by the translators
* Crowdin should be kept as translation platform, because 1) it is much easier for the translators than the GitHub workflow and 2) it is free for OSS projects.

## Considered Options

* Only translated strings in language file
* Translated and untranslated strings in language file, have value the untranslated string to indicate untranslated
* Translated and untranslated strings in language file, have empty to indicate untranslated

## Decision Outcome

Chosen option: "Only translated strings in language file", because comes out best (see below.

## Pros and Cons of the Options

### Only translated strings in language file
* Good, because Crowdin supports it
* Bad, because translators need tooling to see untranslated strings
* Bad, because issues with FXML

### Translated and untranslated strings in language file, have value the untranslated string to indicate untranslated
* Good, because no issues with FXML
* Good, because Crowin supports it
* Bad, because untranslated strings cannot be identified easily in latin languages

### Translated and untranslated strings in language file, have empty to indicate untranslated
* Good, because untranslated strings can be identified easily
* Good, because works with FMXL (?)
* Bad, because Crowdin does not support it

koppor:
I just merged to move things forward, but the discussion on the options is not finalized.

koppor:
I think, we need the script if Crowdin changes its policy or vanishes from the market. In all other cases, our current workflow should be "good enough"? We can add new English language keys (by following the output of the failing tests). These are then synced to Crowdin. After a translator translated something, it goes into JabRef as pull request.

Nevertheless, it is nice to have and other projects may benefit from it. Thus, a pip package (and thus a separate repo) might be the way to go.

Siedlerchr:
Regarding the FXML issues, @tobiasdiez  fixed that in #3820

PR #3744, opened 2018-02-19
[WIP] Fix author parsing with ands 

mohamad-altaweel:
Fixes https://github.com/koppor/jabref/issues/294

I think that the issue is not with the junior suffix. 
first I tried to implement a new method but it didn't make any sense
I have tested inputs examples in the form fName lName,fName lName,fName lName,fName lName.....
so the problem is always with the "and" that comes before the Last author's name
before normalizing i added replaceAll method that replaces all and sepertaing between two names with a comma to get to my tested case and then normalizing it...

----

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -31,6 +31,7 @@ Brian Quistorff
 Brian Van Essen
 captain123
 Carlos Silla
+cerrisantos
 Christian Bartsch
 Christian Kopf
 Christoph Braun
@@ -89,6 +90,7 @@ Jeff Kerr
 Jeff Miller
 Jeffrey Kuhn
 Jens Döcke
+Joerg Lenhard
 Johannes Manner
 John David
 John Relph
@@ -138,6 +140,7 @@ Michael Spiegel
 Michael Wrighton
 Michel Baylac
 Mike Smoot
+Mohamad Altaweel
 Moritz Ringler
 Morten Alver
 Mélanie Tremblay
@@ -163,6 +166,7 @@ Predrag Milanovic
 Raik Nagel
 Renato Massao
 Richard Schneeman
+Robert
 Robert Jäschke
 Rolf Starre
 Rudolf Seemann
@@ -190,6 +194,7 @@ Stephan Lau
 Stephan Rave
 Stéphane Curet
 Sven Jäger
+Tarix
 The Gitter Badger
 Thiago Toledo
 Thomas Arildsen
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,6 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 ## [Unreleased]
 
 ### Changed
-- Changed Authorlist Format method to a new Method Last,First and Last,First. [#294]
- (https://github.com/koppor/jabref/issues/294)
 - Changed ID-based entry generator to store the last used fetcher. [#2796] (https://github.com/JabRef/jabref/issues/2796)
 - Reorganised annotation information on the right side of the "File annotations" tab. [#3109](https://github.com/JabRef/jabref/issues/3109)
 - We now show a small notification icon in the entry editor when we detect data inconsistency or other problems. [#3145](https://github.com/JabRef/jabref/issues/3145)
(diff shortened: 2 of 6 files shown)


src/main/java/org/jabref/logic/formatter/bibtexfields/NormalizeNamesFormatter.java
```diff
@@ -24,6 +24,7 @@ public String getKey() {
     @Override
     public String format(String nameList) {
         Objects.requireNonNull(nameList);
+        nameList = nameList.replaceAll(" and", ",");
```
tobiasdiez: Please don't change the `format` method but only `AuthorList.parse` since the former has almost no test coverage while for the latter exists an extensive test suite.

mohamad-altaweel: done :)

tobiasdiez: Thanks! Did you also run the tests in `AuthorListTests`. I'm pretty sure that some of them now fail. The problem is that now `A, B,C and D, E` and `A, B and C, D, E` are mapped to the same `A, B, C, D, E` and the parser can no longer decide between those two inequivalent inputs.

CHANGELOG.md
```diff
@@ -29,6 +29,8 @@ For more details refer to the [field mapping help page](http://help.jabref.org/e
 The new default removes the linked file from the entry instead of deleting the file from disk. [#3679](https://github.com/JabRef/jabref/issues/3679)
 
 ### Fixed
+- Fixed three commans and one and issue [#294]
```
Siedlerchr: Would you mind describing the fix in more detail?

koppor: Please use the format `[koppor#294](https://github.com/koppor/jabref/issues/294)`

AUTHORS
```diff
@@ -31,6 +31,7 @@ Brian Quistorff
 Brian Van Essen
 captain123
 Carlos Silla
+cerrisantos
```
koppor: It is strange that your commits include that change. Please try to remove it.

You can use `git reset upstream/master` and then re-create your commit using `git gui`.

src/main/java/org/jabref/logic/formatter/bibtexfields/NormalizeNamesFormatter.java
```diff
@@ -24,6 +24,7 @@ public String getKey() {
     @Override
     public String format(String nameList) {
         Objects.requireNonNull(nameList);
+
```
koppor: Please do not add empty lines in classes which you did not touch otherwise.

src/main/java/org/jabref/model/entry/AuthorList.java
```diff
@@ -129,6 +129,8 @@
     private final String[] authorLastFirstAnds = new String[2];
     private final String[] authorsLastFirst = new String[4];
     private final String[] authorsLastFirstFirstLast = new String[2];
+
+
```
koppor: Checkstyle will complain about too many empty lines, didn't it?

src/main/java/org/jabref/model/entry/AuthorList.java
```diff
@@ -164,7 +166,7 @@ public AuthorList() {
      */
     public static AuthorList parse(String authors) {
         Objects.requireNonNull(authors);
-
+        authors = authors.replaceAll(" and", ",");
```
koppor: Are you sure that his works in all cases?

What about

Lastname1, Jr1, Firstname1 and Lastname2, Jr2, Firstname2

(See https://tex.stackexchange.com/q/206760/9075)

This will produce three names, won't it?

Lastname2, Jr1 and Firstname1, Lastname2 and Jr2, Firstname2.

So, the quick fix is not a fix. Please add tests for normal handling of the "Jr" parts. Please google a bit for examples. The example given at https://tex.stackexchange.com/q/206760/9075 should be a good start.

koppor: Oh, I had the same comment as @tobiasdiez 

So, please add our examples as test cases.

koppor:
This PR did not go into deep into the issue. Therefore, I close it.

PR #3748, opened 2018-02-19
Add link to contribute.jabref.org

koppor:
I am not sure about a good way to link http://contribute.jabref.org. WDYT?


tobiasdiez:
What about putting it in the "Contributing" section? But not so prominently, the stuff written on this page is not that important (except the help which nobody updates except us and the translation stuff which works quite well).

koppor:
It is in the contributing section: 😇 

grafik

tobiasdiez:
ah sorry, its late. The location and style is fine with me. Maybe hide the url, e.g. "[We also need help with other things](contributing)".

koppor:
grafik

LinusDietz:
How about: "Not a Programmer? Learn how you can help."?

koppor:
grafik

PR #3704, opened 2018-02-07
[WIP] Multi module build

koppor:
This is a huge step towards https://github.com/JabRef/jabref/issues/110 and prepares https://github.com/JabRef/jabref/pull/3421 somehow.

This enables gradle multi module build.

The dependencies of "logic" were not as clean as I thought. I removed dependencies to GUI or moved packages back to gui (shared, remote).

We should get this in ASAP. 

Open issues:

- [ ] `src/main/gen` should be removed from version control - generated code like jabref.gui/src/main/gen/org/jabref/logic/importer/fileformat/mods/RoleTermDefinition.java is put under version control; is this desired?
- [ ] inherit more from root project avoid duplication of plugins dependencies and tasks (https://docs.gradle.org/current/userguide/multi_project_builds.html)
- [x] GUI does not start: `Cannot load file GroupTree.fxml`
- [x] Install4J does not find the generated JARs.
- [ ] The OO preferences should be fixed (reading/writing not fully supported yet). All other features should work. - will be done later https://github.com/koppor/jabref/issues/304
- [x] `jar` are not uploaded to CircleCI - are they stored somewhere else? http://builds.jabref.org/multi-module-build/
- [x]  Tests not executed on travis. `:org.jabref.gui:junitPlatformTest`  - this works locally though.
- [ ] Tests fail on travis. E.g., `[main] ERROR org.jabref.logic.l10n.Localization - Messages are not initialized before accessing All entries`

PR #3756, opened 2018-02-20
Refactor dublin core utility

johannes-manner:
The tool is now implemented as a console application with a read-evaluate-print cycle.
It can be used with and without a jar file and can be used independently of #3704.

----

- [x] Check documentation status (Updated help page at [help.jabref.org](https://github.com/JabRef/help.jabref.org/issues)?)


src/main/java/org/jabref/cli/XmpUtilMain.java
```diff
@@ -0,0 +1,128 @@
...
+        System.out.println("-------------------------------------------------");
+        System.out.print("Choose an option: ");
+    }
+
+    /**
+     * The tool is implemented as a console application with a read-evaluate-print cycle.
+     *
+     * @param args
+     */
+    public static void main(String[] args) {
```
Siedlerchr: Like the idea, but why not integrate it with the rest of our cli infrastruture? 
Maybe as part of an "Importer" or Exporter?
http://help.jabref.org/en/CommandLine#import-file--i-filenameimport-format

src/main/java/org/jabref/cli/XmpUtilMain.java
```diff
@@ -0,0 +1,128 @@
...
+        System.out.println("(2) Write entries in bib file to Pdf metadata");
+        System.out.println("    To report bugs visit https://issues.jabref.org");
+        System.out.println("-------------------------------------------------");
+        System.out.print("Choose an option: ");
+    }
+
+    /**
+     * The tool is implemented as a console application with a read-evaluate-print cycle.
+     *
+     * @param args
```
koppor: Remove empty parameter description 😇

koppor:
Regarding integration in JabRef: In https://github.com/JabRef/jabref/pull/3704, I separated the main method completely to be able to offer a non-fat jar...

However, I would agree to integrate everything in our normal CLI and then, in the context of the Java 9 refactorings (https://github.com/JabRef/jabref/pull/3421) create finer-grained compilation units.

johannes-manner:
I integrated the xmp functionality in the CLI infrastructure :)

After exiting my tool with '0' the graphical user interface starts... 
@Siedlerchr  How can I suppress the GUI?

Siedlerchr:
When I am right, you just need to call Platform.exit() (at least when it's the javafx application thread start )
Otherwise System.exit(0) (shuts down the Java vm and signals a success)

johannes-manner:
Platform.exit() does not work. Therefore I used the other option System.exit(0).

The implementation is now ready for review. 
Where can I write the help entry for the http://help.jabref.org/en/CommandLine webpage?

Siedlerchr:
@johannes-manner  Okay, good to hear! Help files are all in their own Repo. Either you click on that Octocat Github thing in the right corner, or you go to that repo:
https://github.com/JabRef/help.jabref.org/blob/gh-pages/en/CommandLine.md

src/main/java/org/jabref/cli/XmpUtilMain.java
```diff
@@ -0,0 +1,129 @@
...
+                    String pdfFile = consoleReader.readLine().trim();
+                    XmpUtilMain.writeBibFileToPdfMetadata(bibFile, pdfFile);
+                }
+            } catch (IOException | TransformerException e) {
+                System.err.println(e.getMessage());
+            }
+        }
+    }
+
+    public static void main(String[] args) {
```
tobiasdiez: Do we actually need this tool as a separate executable now that the functionality is included in the usual CLI?

johannes-manner: It's a convenient tool to add metadata to pdf files. If there is the possiblity to build a standalone jar with the multi-model build, why not use the chance to offer this tool to more users and make this tool more popular?

PR #3758, opened 2018-02-21
Reorder checklist in PR template and add "good commit message"

koppor:
I found https://github.com/joelparkerhenderson/git_commit_message in the net and I really like it. Maybe, we should promote that here, too? WDYT?

I also reordered the checklist at the PR template to go from UI to technical issues. Although the documentation is still the last bullet point.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -5,8 +5,9 @@
 
 ----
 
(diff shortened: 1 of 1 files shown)


lenhard:
I agree with Christoph, please put the Changelog entry to the top. After all, no Changelog entry is a blocker for merging most of the time, whereas a missing screenshot is not.

tobiasdiez:
And added tests are more important than manually tried it in JabRef ;-)

PR #3759, opened 2018-02-21
Update branch (again)

mohamean:
Update branch from origin with one from upstream


Siedlerchr:
You should now use the maintable beta branch. Do a git fetch upstream to fetch the latest version of the branch. It will tell you if your branch is even or not

PR #3761, opened 2018-02-21
Don't trim when migrating review field

LinusDietz:
fixes https://github.com/koppor/jabref/issues/311

### DIFF
--- a/src/main/java/org/jabref/logic/bibtex/FieldContentParser.java
+++ b/src/main/java/org/jabref/logic/bibtex/FieldContentParser.java
@@ -28,6 +28,7 @@ public class FieldContentParser {
         multiLineFields = new HashSet<>();
         // the following two are also coded in org.jabref.logic.bibtex.LatexFieldFormatter.format(String, String)
         multiLineFields.add(FieldName.ABSTRACT);
+        multiLineFields.add(FieldName.COMMENT);
         multiLineFields.add(FieldName.REVIEW);
         // the file field should not be formatted, therefore we treat it as a multi line field
         multiLineFields.addAll(prefs.getNonWrappableFields());
--- a/src/main/java/org/jabref/logic/migrations/MergeReviewIntoCommentMigration.java
+++ b/src/main/java/org/jabref/logic/migrations/MergeReviewIntoCommentMigration.java
@@ -48,7 +48,7 @@ public class MergeReviewIntoCommentMigration {
     private String mergeCommentFieldIfPresent(BibEntry entry, String review) {
         if (entry.getField(FieldName.COMMENT).isPresent()) {
             LOGGER.info(String.format("Both Comment and Review fields are present in %s! Merging them into the comment field.", entry.getAuthorTitleYear(150)));
-            return String.format("%s\n%s:\n%s", entry.getField(FieldName.COMMENT).get(), Localization.lang("Review"), review);
+            return String.format("%s\n%s:\n%s", entry.getField(FieldName.COMMENT).get().trim(), Localization.lang("Review"), review.trim());
         }
         return review;
     }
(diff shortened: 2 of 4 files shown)


koppor:
This does not help, because IN THE MIDDLE the newlines are deleted. 

```
@Article{AHP1996,
  author    = {R. Alur and G. J. Holzmann and D. Peled},
  title     = {An analyzer for Message Sequence Charts},
  journal   = {Software - Concepts and Tools},
  year      = {1996},
  pages     = {70–77},
  volume    = {17},
  number    = {2},
  keywords  = {TODO, race conditions},
  owner     = {koppor},
  review    = {TSE cites it as THE definition of "race conditions" (page 27)

In the literature about MSCs, the possibility that messages may be received in a different order than the one specified is usually called a race condition [2].},
  timestamp = {2011.04.17},
}
```

src/main/java/org/jabref/logic/migrations/MergeReviewIntoCommentMigration.java
```diff
@@ -48,7 +48,7 @@ public void performConflictingMigration(ParserResult parserResult) {
     private String mergeCommentFieldIfPresent(BibEntry entry, String review) {
         if (entry.getField(FieldName.COMMENT).isPresent()) {
             LOGGER.info(String.format("Both Comment and Review fields are present in %s! Merging them into the comment field.", entry.getAuthorTitleYear(150)));
-            return String.format("%s\n%s:\n%s", entry.getField(FieldName.COMMENT).get().trim(), Localization.lang("Review"), review.trim());
+            return String.format("%s\n%s:\n%s", entry.getField(FieldName.COMMENT).get(), Localization.lang("Review"), review);
```
koppor: Maybe, `%s` does something magic? Or does the `get()` do magic?

LinusDietz: wow. well I'll write a unit test for that and then we'll see.

Siedlerchr: I always forget the specifc string arguments formatter, too
`%$1s` etc- Maybe that helps?
An argument index is specified as a number ending with a “$” after the “%” and selects the specified argument in the argument list.
https://dzone.com/articles/java-string-format-examples

koppor: No, we just forgot that `COMMENT` is a mutli line field. We decided two or three years ago that JabRef rewraps all field values to have a good looking bibtex file. We make exceptions for **predefined** multi line fields. I am working on an update.

koppor:
I reverted the revert of trimming. I think, that is a good thing to do!

LinusDietz:
I'm approving this as well. Whoever merges, don't squash so @koppor gets credit as well!

src/main/java/org/jabref/logic/bibtex/FieldContentParser.java
```diff
@@ -28,6 +28,7 @@ public FieldContentParser(FieldContentParserPreferences prefs) {
         multiLineFields = new HashSet<>();
         // the following two are also coded in org.jabref.logic.bibtex.LatexFieldFormatter.format(String, String)
         multiLineFields.add(FieldName.ABSTRACT);
```
tobiasdiez: Please extract these fields that should be treated as multilines to `InternalBibtexFields` (or a better location). Moreover, a few places still work with https://github.com/JabRef/jabref/blob/733da312471c3abab691d61eccee061962357e1c/src/main/java/org/jabref/model/entry/FieldName.java#L107 instead of `comments`. This should be changed too.

koppor:
I treat this at hotfix. I copied the comment by @tobiasdiez to https://github.com/koppor/jabref/issues/314

koppor:
@LinusDietz I used the Co-authored-by: feature: https://github.com/blog/2496-commit-together-with-co-authors

PR #3762, opened 2018-02-21
Meaningful button review field migration

LinusDietz:
Fixes https://github.com/koppor/jabref/issues/312.

Sorry, the PR quality of https://github.com/JabRef/jabref/pull/3658 was really not so good.

### DIFF
--- a/src/main/java/org/jabref/gui/importer/actions/MergeReviewIntoCommentAction.java
+++ b/src/main/java/org/jabref/gui/importer/actions/MergeReviewIntoCommentAction.java
@@ -20,7 +20,7 @@ public class MergeReviewIntoCommentAction implements GUIPostOpenAction {
 
         migration.performMigration(parserResult);
         List<BibEntry> conflicts = MergeReviewIntoCommentMigration.collectConflicts(parserResult);
-        if (new MergeReviewIntoCommentConfirmationDialog(basePanel).askUserForMerge(conflicts)) {
+        if (!conflicts.isEmpty() && new MergeReviewIntoCommentConfirmationDialog(basePanel).askUserForMerge(conflicts)) {
             migration.performConflictingMigration(parserResult);
         }
     }
--- a/src/main/java/org/jabref/gui/importer/actions/MergeReviewIntoCommentConfirmationDialog.java
+++ b/src/main/java/org/jabref/gui/importer/actions/MergeReviewIntoCommentConfirmationDialog.java
@@ -25,13 +25,14 @@ public class MergeReviewIntoCommentConfirmationDialog {
                 .map(Optional::get)
                 .collect(Collectors.joining(",\n"));
 
-        int answer = JOptionPane.showConfirmDialog(
+        String[] options = {Localization.lang("Merge fields")};
+        int answer = JOptionPane.showOptionDialog(
                 panel,
                 bibKeys + " " +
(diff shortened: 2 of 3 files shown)


src/main/java/org/jabref/gui/importer/actions/MergeReviewIntoCommentConfirmationDialog.java
```diff
@@ -25,13 +25,14 @@ public boolean askUserForMerge(List<BibEntry> conflicts) {
...
-        int answer = JOptionPane.showConfirmDialog(
+        String[] options = {Localization.lang("Merge fields")};
+        int answer = JOptionPane.showOptionDialog(
                 panel,
                 bibKeys + " " +
                         Localization.lang("has/have both a 'Comment' and a 'Review' field.") + "\n" +
                         Localization.lang("Since the 'Review' field was deprecated in JabRef 4.2, these two fields are about to be merged into the 'Comment' field.") + "\n" +
                         Localization.lang("The conflicting fields of these entries will be merged into the 'Comment' field."),
-                Localization.lang("Review Field Migration"), JOptionPane.DEFAULT_OPTION, JOptionPane.WARNING_MESSAGE);
+                Localization.lang("Review Field Migration"), JOptionPane.DEFAULT_OPTION, JOptionPane.WARNING_MESSAGE, null, options, options[0]);
```
Siedlerchr: Maybe not a "Warning message", in fact it's an information message.  The user can't do anything about it.

LinusDietz: Well, infact it's a bit more important than just a simple notification, but I'd be open if others agree with you I'ts gonna be changed.

src/main/java/org/jabref/gui/importer/actions/MergeReviewIntoCommentAction.java
```diff
@@ -20,7 +20,7 @@ public void performAction(BasePanel basePanel, ParserResult parserResult) {
 
         migration.performMigration(parserResult);
         List<BibEntry> conflicts = MergeReviewIntoCommentMigration.collectConflicts(parserResult);
-        if (new MergeReviewIntoCommentConfirmationDialog(basePanel).askUserForMerge(conflicts)) {
+        if (!conflicts.isEmpty() && new MergeReviewIntoCommentConfirmationDialog(basePanel).askUserForMerge(conflicts)) {
```
koppor: I think, this makes it not better. Now, JabRef migrates WITHOUT ANY notification. The old way was better. It showed that something was migrated.

LinusDietz: Well, then lets discuss this again: The rationale behind #3658 was that the warning should be displayed iff there is BOTH a comment and and review field present and perform the migration silently nonetheless.

koppor:
Oh, JabRef should inform that something happend. Otherwise, it is not
clear, why there is a star. JabRef currently does not have the power to
show what it has changed... This is strange in the case of save actions,
but this can be explained.

IMHO consistency in the UI is an important feature. In the case of pdf/ps
fields, we show a popup. I can agree that we do not support "no migration"
and prevent users of creating bibtex entries with "comment". However, the
tool itself should communicate that it changed the database!

koppor:
grafik

PR #3764, opened 2018-02-22
Added <any file type> to the Import File Filter Dialog.

### LINKED ISSUE #3757: Add *.* to the Import File Filter Dialog
From the forum:

http://discourse.jabref.org/t/importing-bibtex-files-with-jabref-4-1-on-macos-sierra-10-12-6/1060

Currently JabRef only allows the import of files with the list of supported file extensions.
So, as there is now file extension .bibtex for bib files import we should add a *.* (All File types) filter.
The import should then use the automatism (as it's now the case for all avaiable import formats)

gerzmasn:
Added *.* (any file) to the File Filter Dialog. Now you can choose any file to import to try to import to your library. Fixes #3757.

----

### DIFF
--- a/src/main/java/org/jabref/gui/importer/ImportFormats.java
+++ b/src/main/java/org/jabref/gui/importer/ImportFormats.java
@@ -67,12 +67,12 @@ public class ImportFormats {
                 List<FileType> extensions = importers.stream().map(Importer::getFileType)
                         .collect(Collectors.toList());
                 FileChooser.ExtensionFilter allImports = FileFilterConverter.forAllImporters(importers);
-
                 FileDialogConfiguration fileDialogConfiguration = new FileDialogConfiguration.Builder()
                         .addExtensionFilter(allImports)
                         .addExtensionFilters(extensions)
                         .withInitialDirectory(Globals.prefs.get(JabRefPreferences.IMPORT_WORKING_DIRECTORY))
                         .build();
+                fileDialogConfiguration.getExtensionFilters().add(new FileChooser.ExtensionFilter("Any file", "*.*"));
                 DialogService dialogService = new FXDialogService();
                 DefaultTaskExecutor.runInJavaFXThread(() -> {
                     dialogService.showFileOpenDialog(fileDialogConfiguration)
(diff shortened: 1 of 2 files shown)


src/main/java/org/jabref/gui/importer/ImportFormats.java
```diff
@@ -67,12 +67,12 @@ public void actionPerformed(ActionEvent e) {
                 List<FileType> extensions = importers.stream().map(Importer::getFileType)
                         .collect(Collectors.toList());
                 FileChooser.ExtensionFilter allImports = FileFilterConverter.forAllImporters(importers);
-
                 FileDialogConfiguration fileDialogConfiguration = new FileDialogConfiguration.Builder()
                         .addExtensionFilter(allImports)
                         .addExtensionFilters(extensions)
                         .withInitialDirectory(Globals.prefs.get(JabRefPreferences.IMPORT_WORKING_DIRECTORY))
                         .build();
+                fileDialogConfiguration.getExtensionFilters().add(new FileChooser.ExtensionFilter("Any file", "*.*"));
```
Siedlerchr: If you could just wrap that "Any file" in Localization.lang, I would be happy.
Otherwise LGTM! :+1:

tobiasdiez: @Siedlerchr With this addition, is
https://github.com/JabRef/jabref/blob/58c2c7aa941744d9341454bfb5eeabeed77ab54e/src/main/java/org/jabref/gui/util/FileFilterConverter.java#L47-L49
now obsolete?

Siedlerchr: Still a little difference, but you are in general right.

tobiasdiez: So what exactly is the difference? Should we just remove the old one?

Siedlerchr: All available import formats  = All supported file extension
Any file = Any extension not registered as default

PR #3766, opened 2018-02-23
Update Architecture Tests to catch static imports

lenhard:
Fixes a part of #2617

Updates the architecture tests to JUnit5 and also catches violations in static imports now. I tested this by adding a static import and checking that the build breaks. It does, but of course I had to remove the cause again.

----

### DIFF
--- a/src/test/java/org/jabref/architecture/MainArchitectureTests.java
+++ b/src/test/java/org/jabref/architecture/MainArchitectureTests.java
@@ -14,11 +14,12 @@ import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
-import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeAll;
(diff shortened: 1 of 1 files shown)


src/test/java/org/jabref/architecture/MainArchitectureTests.java
```diff
@@ -104,8 +105,7 @@ public void firstPackageIsIndependentOfSecondPackage(String firstPackage, String
                     })
                     .collect(Collectors.toList());
 
-            Assert.assertEquals("The following classes are not allowed to depend on " + secondPackage,
-                    Collections.emptyList(), files);
+            Assertions.assertEquals(Collections.emptyList(), files, "The following classes are not allowed to depend on " + secondPackage);
```
Siedlerchr: I would prefer a static import of asset und

lenhard: Done!

PR #3778, opened 2018-02-26
Fix delete action and convert confirmation dialog to JavaFX

tobiasdiez:
Fixes the delete action in the maintable branch and the "do you really want to delete the entry"-dialog is converted to JavaFX. Moreover, a few lines of JavaFX-Swing-interaction code in `FXDialog` are deleted since it is no longer needed.

----

### DIFF
--- a/src/main/java/org/jabref/gui/BasePanel.java
+++ b/src/main/java/org/jabref/gui/BasePanel.java
@@ -1562,12 +1562,17 @@ public class BasePanel extends StackPane implements ClipboardOwner {
         if (Globals.prefs.getBoolean(JabRefPreferences.CONFIRM_DELETE)) {
             String title = Localization.lang("Delete entry");
             String message = Localization.lang("Really delete the selected entry?");
+            String okButton = Localization.lang("Delete entry");
+            String cancelButton = Localization.lang("Keep entry");
             if (numberOfEntries > 1) {
                 title = Localization.lang("Delete multiple entries");
                 message = Localization.lang("Really delete the %0 selected entries?", Integer.toString(numberOfEntries));
+                okButton = Localization.lang("Delete entries");
+                cancelButton = Localization.lang("Keep entries");
             }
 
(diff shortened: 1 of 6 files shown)


Siedlerchr:
You could take this as an opportunity to reword the question/dialog. Asking questions for yes/no answers is never a good idea https://github.com/JabRef/jabref/wiki/Code-Howtos#designing-gui-confirmation-dialogs

tobiasdiez:
@Siedlerchr Why not ask a question when you mean to ask one? In the stackoverflow thread that is linked in our wiki, the "never ask questions" rule is not mentioned and the examples discuss there actually show dialogs with questions. Microsoft seems to have the same opinion: https://msdn.microsoft.com/en-us/library/windows/desktop/dn742470.aspx

Siedlerchr:
It's better to state the actions in a user centric way. Just say sth like. You are about to delete the xxxx. Look at the file deletion dialog in the general tab. 

https://www.nngroup.com/articles/confirmation-dialog/

Here is a good explanation about asking questions in dialogs 

https://books.google.de/books?id=e75G0xIJju8C&pg=PA217&lpg=PA217&dq=ask+question+in+confirmation+dialog&source=bl&ots=6SLQCqcUBg&sig=dq2aJhgEcxrsWAnEJwVuXd6GFKo&hl=de&sa=X&ved=2ahUKEwjK8t2P98PZAhUICewKHfMiAvUQ6AEwCHoECAQQAQ#v=onepage&q=ask%20question%20in%20confirmation%20dialog&f=false

tobiasdiez:
I now changed the buttons from Yes/No to Delete entry / Keep entry. However, both links couldn't convince me that questions are bad. The first link even contains the standard Microsoft delete file dialog as a good example (and generally references the microsoft guidelines, which are full of "Do you really want..." questions). I see the reasoning why actions are usually better than questions, but in this case we really want to ask for confirmation. Anyway, the aim of this PR is not this dialog but actually fixing the delete function.

Issue #3028, opened 2017-07-20
Accentuation is not working ~^'`" (eg. ã ô é à ü)


kafran:
OS: Xubuntu 17.04
Computer keyboard map: Brazilian ABNT2
Java version: 
openjdk version "1.8.0_131"
OpenJDK Runtime Environment (build 1.8.0_131-8u131-b11-0ubuntu1.17.04.1-b11)
OpenJDK 64-Bit Server VM (build 25.131-b11, mixed mode)

JabRef version: JabRef-4.0-dev--snapshot--2017-07-19--master--2f12fecfb.jar

I **can't** insert (only) these accentuated character (comma separated) : ~, ^, `, ', "
Because of this, I can't make these letters: à, á, ã, â, ô, ó, ò, é, ê, ü, etc.
I can't insert these on Fields or Biblatex source tabs;
(Edit to reduce ambiguity) Inserting others special characters works ok: $, %, &, *, (, º

lenhard:
I've tried this with the current master branch on Win7. Everything works as expected, I can enter whatever characters I wish.

Is anyone able to reproduce this?

Siedlerchr:
Maybe related to this compose key stuff?

Am 20.07.2017 9:56 vorm. schrieb "Jörg Lenhard" <notifications@github.com>:

kafran:
On JabRef 3.8.2 at the same system I don't face this problem.

AEgit:
Cannot confirm for:
JabRef 4.0-dev--snapshot--2017-07-19--master--2f12fecfb
Windows 10 10.0 amd64 
Java 1.8.0_141

kafran:
Tested on another computer with a different keyboard. Its a Chromebook with GalliumOS (Ubuntu 16.04) and  openjdk version "1.8.0_131". The same problem.

Siedlerchr:
Can confirm on my Ubuntu 16.04 VM with oracle jdk 131, language setting en_gb m
Seems to be related to all kind of javafx fields inside swing, entering the text in pure Swing works.

@kafran  Could you check the event log, if you get the same error as I?

Opening: /home/osboxes/test
Opening: /home/osboxes/Documents/JABREFTEMP/references.bib
Uncaught exception Occurred in Thread[AWT-EventQueue-0,6,main]

```
java.lang.NullPointerException
	at javafx.embed.swing.JFXPanel.sendInputMethodEventToFX(JFXPanel.java:647)
[...]

PR #3779, opened 2018-02-26
Remove customjfx

Siedlerchr:
This removes the custom JfXPanel hack we introduced for  #3028  @koppor 
As now the main frame is javafx and no longer swing this hack is no longer needed and can be removed.

I did not completely remove the class, because some dialogs still require a jfxpanel with a scene

----


Siedlerchr:
Hm, good question. I will take a look at that.

Siedlerchr:
As far as I see the panel is only used to add the PreviewPanel which does not provide a way to input content. So we can safely remove that.

tobiasdiez:
Ok, but we may still need it in the future for example when the preferences are migrated pane by pane. If the library makes problems with Java 9, then I'm OK with merge. Otherwise I would keep it for the moment until the majority of the dialogs is also converted.

tobiasdiez:
So, is our custom fx hack a problem for java 9 or not? In the latter case I would vote for closing this PR; otherwise just merge it.

Siedlerchr:
I want to get rid of the class completely, it can be safely removed as JavaFX is our first class citizen.
For swing content we need to embed we can now use Swing Node 

I will update the PR to convert the remaining usage to fx

Siedlerchr:
I now converted the Preamble Editor to javafx. This was the only critical thing which remained and had a TextArea as input field.
All other cases where Customjfxpanel is used is uncrictical and only the previewpanel in the External Changes Dialog.

Siedlerchr:
@tobiasdiez Well, we can simply do it the other way round then. We wrap it in javafx by using a SwingNode. 
I did this with the cleanup operations already

PR #3783, opened 2018-02-27
Save order of columns across sessions

tobiasdiez:
After the user reorder the columns of the maintable per drag and drop, these changes are now saved in the preferences. Moreover, I refactored a bit how preferences are loaded from the global preferences class.

----


src/main/java/org/jabref/gui/preftabs/TableColumnsTab.java
```diff
@@ -438,14 +389,51 @@ public void setValues() {
...
                 tableRows.add(new TableRow(names.get(i)));
             }
         }
         rowCount = tableRows.size() + 5;
-        ncWidth = prefs.getInt(JabRefPreferences.NUMBER_COL_WIDTH);
+    }
+
+    /*** end: special fields ***/
+
+    static class TableRow {
```
Siedlerchr: Any reason why this has to be static?

Siedlerchr:
Can I still change the order and width in the settings?

src/main/java/org/jabref/gui/preftabs/TableColumnsTab.java
tobiasdiez: No idea... (the code quality in the preferences tabs is so bad anyway).

tobiasdiez:
@Siedlerchr Yes, this should still work.

src/main/java/org/jabref/gui/preftabs/TableColumnsTab.java
Siedlerchr: I personally see to reason to have this whole column stuff in the preferences anylonger. I think it was only added because we could not save the column reordering the user did directly on the maintable. I would vote to remove that stuff completely.

tobiasdiez: Well, currently, this is the only place where you can choose which columns are shown.

Siedlerchr: Why not adding a table column selector like it's possible in nearly every table 

grafik

Siedlerchr: Here is an example https://gist.github.com/Roland09/d92829cdf5e5fee6fee9

tobiasdiez: This is a good idea! I have to think a bit more about it and how to make it work for JabRef (we allow also columns of the form `author/editor` and have to many fields to display them all in a drop down - possible a combination of drop down and text field may work). But not as part of this PR.

PR #3621, opened 2018-01-09
[WIP] Main table JavaFX migration

tobiasdiez:
This PR tracks the progress of the migration of the maintable to JavaFX. See also the [Maintable in JavaFX project](https://github.com/JabRef/jabref/projects/4). The current version can be downloaded [here](https://builds.jabref.org/maintable-beta/).

image

Please report any bugs that appear or missing features, which are not listed below. Moreover, we appreciate feedback about the general design (what do guys like, what not)

Known bugs:
--
- [ ] Some input controls (like check boxes) look odd
- [ ] Smaller things outlined [here](https://github.com/JabRef/jabref/pull/3621#issuecomment-369646377) and [here](https://github.com/JabRef/jabref/pull/3621#issuecomment-378598558)
- [ ] Drag & drop to and from the table does not work #3765 
- [ ] Ranking is not shown (and thus not editable)
- [ ] Float mode does not work (partly because of [unresolved feature request for Java](https://bugs.openjdk.java.net/browse/JDK-8091326))  [code to fix](https://stackoverflow.com/questions/37768326/group-by-function-in-javafx-table)
- [ ] Linux / Mac OS: Resizing the columns per drag & drop does not work nicely (https://github.com/JabRef/jabref/pull/3621#issuecomment-368821666)
- [ ] Linux: Column separator are not shown on hover (https://github.com/JabRef/jabref/pull/3621#issuecomment-368821666)
- [ ] Performance of search is poor for large databases https://github.com/JabRef/jabref/issues/1513#issuecomment-361528430
- [x] <s>Magnifier icon is not colored correctly - we accept that workaround until we implement something really cool as proposed at https://github.com/JabRef/jabref/issues/3535</s> Follow up at https://github.com/koppor/jabref/issues/341
- [x] <s>Mouse wheel for tabs in entry editor works in the wrong way - up/down reversed in comparison to Firefox (by @koppor)</s> Follow up at https://github.com/JabRef/jabref/issues/4089
- [x] <s>When in the maintable, <kbd>Ctrl</kbd>+<kbd>F</kbd> does not focus the search field  (by @koppor)</s> Follow up at https://github.com/JabRef/jabref/issues/3062
- [x] Import does not work, see https://github.com/JabRef/jabref/pull/3621#issuecomment-387529415
- [x] Trailing thread even after JabRef is closed https://github.com/JabRef/jabref/pull/3621#issuecomment-372785852
[...]

PR #3785, opened 2018-02-27
Show empty group pane if no database is open

tobiasdiez:
Fixes an issue where the group tree of the last database is still shown even after the database was closed.
Reported in https://github.com/JabRef/jabref/pull/3621#issuecomment-368821666

----

### DIFF
--- a/src/main/java/org/jabref/gui/JabRefFrame.java
+++ b/src/main/java/org/jabref/gui/JabRefFrame.java
@@ -278,6 +278,7 @@ public class JabRefFrame extends BorderPane implements OutputPrinter {
          */
         EasyBind.subscribe(tabbedPane.getSelectionModel().selectedItemProperty(), e -> {
             if (e == null) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/jabref/gui/JabRefFrame.java
```diff
@@ -278,6 +278,7 @@ private void init() {
          */
         EasyBind.subscribe(tabbedPane.getSelectionModel().selectedItemProperty(), e -> {
             if (e == null) {
+                Globals.stateManager.activeDatabaseProperty().setValue(Optional.empty());
```
Siedlerchr: Hm, Optional sollte man eigentlich nicht als Parameter nutzen.

tobiasdiez: Yes, but this is the only way because `activeDatabaseProperty` uses the generic `Property<T>`, which provides `setValue(T value)`.

PR #3801, opened 2018-03-02
Dialogstojavafx

Siedlerchr:
I have converted a lot of dialogs to javafx.
I did not test each one.

Some of them are going to be replaced in the near future (e.g. the Strings dialog and the import of files/drag and drop stuff), so I did not convert them.

Dialogs which need special treatment are marked with TODO statements.
Some of the candidates are the ones that integrate html or the preview pane

----


src/main/java/org/jabref/gui/BasePanel.java
```diff
@@ -525,7 +525,7 @@ public void update() {
 
         actions.put(Actions.OPEN_URL, new OpenURLAction());
 
-        actions.put(Actions.MERGE_WITH_FETCHED_ENTRY, new MergeWithFetchedEntryAction(this));
+        actions.put(Actions.MERGE_WITH_FETCHED_ENTRY, new MergeWithFetchedEntryAction(this, frame));
```
tobiasdiez: could you please pass the `DialogService` directly, instead of the frame (makes the dependency clearer).

src/main/java/org/jabref/gui/collab/ChangeScanner.java
```diff
@@ -94,17 +94,21 @@ public void displayResult(final DisplayResultCallback fup) {
...
             fup.scanResultsResolved(true);
         }
     }
 
     private void storeTempDatabase() {
         JabRefExecutorService.INSTANCE.execute(() -> {
             try {
-                SavePreferences prefs = SavePreferences.loadForSaveFromPreferences(Globals.prefs).withMakeBackup(false)
-                        .withEncoding(panel.getBibDatabaseContext().getMetaData().getEncoding()
+                SavePreferences prefs = SavePreferences.loadForSaveFromPreferences(Globals.prefs)
```
tobiasdiez: The code style settings still seem to be off. The dots are not aligned (if this is not an artifact of the way github displays the code)

src/main/java/org/jabref/gui/BasePanel.java
```diff
@@ -657,7 +657,7 @@ public void update() {
         actions.put(Actions.REMOVE_FROM_GROUP, new GroupAddRemoveDialog(this, false, false));
         actions.put(Actions.MOVE_TO_GROUP, new GroupAddRemoveDialog(this, true, true));
 
-        actions.put(Actions.DOWNLOAD_FULL_TEXT, new FindFullTextAction(this));
+        actions.put(Actions.DOWNLOAD_FULL_TEXT, new FindFullTextAction(frame, this));
```
tobiasdiez: dito

src/main/java/org/jabref/gui/customentrytypes/EntryCustomizationDialog.java
```diff
@@ -101,7 +100,7 @@ private void initGui() {
         List<String> entryTypes = new ArrayList<>();
         entryTypes.addAll(EntryTypes.getAllTypes(bibDatabaseMode));
 
-        typeComp = new EntryTypeList(entryTypes, bibDatabaseMode);
+        typeComp = new EntryTypeList(frame, entryTypes, bibDatabaseMode);
```
tobiasdiez: here also

src/main/java/org/jabref/gui/externalfiles/DownloadExternalFile.java
```diff
@@ -160,9 +160,8 @@ public void download(URL url, final DownloadCallback callback) throws IOExceptio
             // TODO: stop editor dialog if this results in an error:
             mimeType = udl.getMimeType(); // Read MIME type
         } catch (IOException ex) {
-            JOptionPane.showMessageDialog(null, Localization.lang("Invalid URL") + ": " + ex.getMessage(),
-                    Localization.lang("Download file"), JOptionPane.ERROR_MESSAGE);
             LOGGER.info("Error while downloading " + "'" + res + "'", ex);
+            frame.getDialogService().showErrorDialogAndWait(Localization.lang("Download file"), Localization.lang("Invalid URL") + ": " + ex.getMessage(), ex);
```
tobiasdiez: This code should display the error message twice if I'm not mistaken.

src/main/java/org/jabref/gui/exporter/SaveDatabaseAction.java
```diff
@@ -140,7 +140,9 @@ public void run() {
 
                 // Save the database
                 success = saveDatabase(panel.getBibDatabaseContext().getDatabaseFile().get(), false,
-                        panel.getBibDatabaseContext().getMetaData().getEncoding()
+                        panel.getBibDatabaseContext()
```
tobiasdiez: Do these restylings happen automatically? If yes, could eclipse be configured differently? Sometimes they are good, sometimes they feel a bit over aggressive (like here, the two-lines solution was fine in my opinion)

src/main/java/org/jabref/gui/openoffice/OpenOfficePanel.java
```diff
@@ -672,13 +670,14 @@ private boolean checkThatEntriesHaveKeys(List<BibEntry> entries) {
...
+        frame.getDialogService().showErrorDialogAndWait(Localization.lang("Connection lost"),
                 Localization.lang("Connection to OpenOffice/LibreOffice has been lost. "
-                        + "Please make sure OpenOffice/LibreOffice is running, and try to reconnect."),
-                Localization.lang("Connection lost"), JOptionPane.ERROR_MESSAGE);
+                + "Please make sure OpenOffice/LibreOffice is running, and try to reconnect."));
+
     }
 
     private void reportUndefinedParagraphFormat(UndefinedParagraphFormatException ex) {
+        //TODO: Javafx does not support html as far as I know
```
tobiasdiez: Thats right, but the dialog does not seem to actually need html (the only tag is for the newline, right?)

src/main/java/org/jabref/gui/openoffice/OpenOfficePanel.java
```diff
@@ -694,6 +693,8 @@ private void reportUndefinedParagraphFormat(UndefinedParagraphFormatException ex
     }
 
     private void reportUndefinedCharacterFormat(UndefinedCharacterFormatException ex) {
+        //TODO: Javafx does not support html as far as I know
```
tobiasdiez: see above

src/main/java/org/jabref/gui/preftabs/AdvancedTab.java
```diff
@@ -156,10 +158,12 @@ public boolean validateSettings() {
                 throw new NumberFormatException();
             }
         } catch (NumberFormatException ex) {
-            JOptionPane.showMessageDialog(null,
+
+            frame.getDialogService().showErrorDialogAndWait(Localization.lang("Remote server port"),
                     Localization.lang("You must enter an integer value in the interval 1025-65535 in the text field for")
                             + " '" + Localization.lang("Remote server port") + '\'',
-                    Localization.lang("Remote server port"), JOptionPane.ERROR_MESSAGE);
+                    ex);
```
tobiasdiez: the exception as an argument is unneeded here, since the stack trace is not helpful at all in this situation: the user entered data in the wrong format and this is said already in the error message.

src/main/java/org/jabref/gui/preftabs/GeneralTab.java
```diff
@@ -64,8 +67,8 @@ public Component getListCellRendererComponent(JList<?> list, Object value, int i
         }
     }
 
-
-    public GeneralTab(JabRefPreferences prefs) {
+    public GeneralTab(JabRefFrame frame, JabRefPreferences prefs) {
```
tobiasdiez: directly pass the dialogservice pls

src/main/java/org/jabref/gui/preftabs/GeneralTab.java
```diff
@@ -255,10 +257,9 @@ public boolean validateSettings() {
             DateTimeFormatter.ofPattern(timeStampFormat.getText());
 
         } catch (IllegalArgumentException ex2) {
-            JOptionPane.showMessageDialog
-                    (null, Localization.lang("The chosen date format for new entries is not valid"),
-                            Localization.lang("Invalid date format"),
-                            JOptionPane.ERROR_MESSAGE);
+            JOptionPane.showMessageDialog(null, Localization.lang("The chosen date format for new entries is not valid"),
```
tobiasdiez: you missed this one ;-)

src/main/java/org/jabref/gui/preftabs/PreferencesDialog.java
```diff
@@ -94,9 +93,9 @@ public PreferencesDialog(JabRefFrame parent) {
         tabs.add(new ExportSortingPrefsTab(prefs));
         tabs.add(new NameFormatterTab(prefs));
         tabs.add(new XmpPrefsTab(prefs));
-        tabs.add(new NetworkTab(prefs));
-        tabs.add(new AdvancedTab(prefs));
-        tabs.add(new AppearancePrefsTab(prefs));
+        tabs.add(new NetworkTab(frame, prefs));
```
tobiasdiez: dialogservice everywhere

src/main/java/org/jabref/gui/preftabs/PreviewPrefsTab.java
```diff
@@ -127,13 +130,17 @@ private void setupLogic() {
                 testPane.setEntry(TestEntry.getTestEntry());
                 JFXPanel container = CustomJFXPanel.wrap(new Scene(testPane));
                 container.setPreferredSize(new Dimension(800, 350));
+
+                //TODO: Add dialog for showing pane
```
tobiasdiez: This should be easy using https://github.com/JabRef/jabref/blob/f7d96784dda0ee0d9adefd43b5a6803058284500/src/main/java/org/jabref/gui/DialogService.java#L114-L120

src/main/java/org/jabref/gui/importer/ImportInspectionDialog.java
```diff
@@ -733,16 +730,20 @@ public void actionPerformed(ActionEvent event) {
...
-                                        .lang("There are possible duplicates (marked with an icon) that haven't been resolved. Continue?"),
-                                Localization.lang("Disable this confirmation dialog"), false);
-                        int answer = JOptionPane.showConfirmDialog(ImportInspectionDialog.this, cbm,
-                                Localization.lang("Duplicates found"), JOptionPane.YES_NO_OPTION);
-                        if (cbm.isSelected()) {
-                            Globals.prefs.putBoolean(JabRefPreferences.WARN_ABOUT_DUPLICATES_IN_INSPECTION, false);
-                        }
-                        if (answer == JOptionPane.NO_OPTION) {
+
+                        //  return dialogService.showConfirmationDialogWithOptOutAndWait(title, message,
```
tobiasdiez: please remove these comments

src/main/java/org/jabref/gui/openoffice/StyleSelectDialog.java
```diff
@@ -272,9 +271,11 @@ private void setupPopupMenu() {
...
         // Create action listener for removing a style, also used for the remove button
         removeAction = actionEvent -> getSelectedStyle().ifPresent(style -> {
-            if (!style.isFromResource() && (JOptionPane.showConfirmDialog(diag,
-                    Localization.lang("Are you sure you want to remove the style?"), Localization.lang("Remove style"),
-                    JOptionPane.YES_NO_OPTION) == JOptionPane.YES_OPTION)) {
+            boolean removeStylePressed = frame.getDialogService().showConfirmationDialogAndWait(Localization.lang("Remove style"),
+                    Localization.lang("Are you sure you want to remove the style?"),
+                    Localization.lang("Remove style"),
+                    Localization.lang("Cancel"));
+            if (!style.isFromResource() && removeStylePressed) {
```
tobiasdiez: previously the dialog was only shown if `style.isFromResource` returned false. Is this change of behavior desirable?

src/main/java/org/jabref/gui/protectedterms/ProtectedTermsDialog.java
```diff
@@ -254,10 +253,12 @@ private void setupPopupMenu() {
         // Create action listener for removing a term file, also used for the remove button
         removeAction = actionEvent -> getSelectedTermsList().ifPresent(list -> {
 
-            if (!list.isInternalList() && (JOptionPane.showConfirmDialog(diag,
+            boolean removeProctedTermsFilePressed = frame.getDialogService().showConfirmationDialogAndWait(Localization.lang("Remove protected terms file"),
```
tobiasdiez: same here

src/main/java/org/jabref/gui/FXDialogService.java
```diff
@@ -68,7 +68,7 @@ private static FXDialog createDialog(AlertType type, String title, String conten
     }
 
     private static FXDialog createDialogWithOptOut(AlertType type, String title, String content,
-                                                   String optOutMessage, Consumer<Boolean> optOutAction) {
+            String optOutMessage, Consumer<Boolean> optOutAction) {
```
tobiasdiez: This seems to be another inconsistency between eclipse and intellj style

src/main/java/org/jabref/gui/shared/ConnectToSharedDatabaseDialog.java
```diff
@@ -115,9 +115,10 @@ public ConnectToSharedDatabaseDialog(JabRefFrame frame) {
     public void openSharedDatabase() {
 
         if (isSharedDatabaseAlreadyPresent()) {
-            JOptionPane.showMessageDialog(ConnectToSharedDatabaseDialog.this,
-                    Localization.lang("You are already connected to a database using entered connection details."),
-                    Localization.lang("Warning"), JOptionPane.WARNING_MESSAGE);
+
+            frame.getDialogService().showWarningDialogAndWait(Localization.lang("Warning"),
```
tobiasdiez: Use a better title?

src/main/java/org/jabref/gui/openoffice/StyleSelectDialog.java
Siedlerchr: Ah shit, I oversaw that condiation

src/main/java/org/jabref/gui/FXDialogService.java
Siedlerchr: I will investigate that

src/main/java/org/jabref/gui/push/PushToEmacs.java
```diff
@@ -61,11 +63,12 @@ protected void initSettingsPanel() {
     }
 
     @Override
-    public void pushEntries(BibDatabase database, List<BibEntry> entries, String keys, MetaData metaData) {
+    public void pushEntries(BibDatabase database, List<BibEntry> entries, String keys, MetaData metaData, DialogService dialogService) {
```
tobiasdiez: The `dialogService` makes more sense as a constructor argument, I think. Probably push it even to `AbstractPushToApplication`

Siedlerchr: I thought so, too, but the concrete subclass of PushToEntries (operation) is just  passed as constructur argument as a dependency, so it is not possible

src/main/java/org/jabref/gui/openoffice/OpenOfficePanel.java
```diff
@@ -672,41 +669,30 @@ private boolean checkThatEntriesHaveKeys(List<BibEntry> entries) {
...
-                                        "Your style file specifies the paragraph format '%0', "
-                                                + "which is undefined in your current OpenOffice/LibreOffice document.",
-                                        ex.getFormatName())
-                                + "<br>"
-                                + Localization
-                                        .lang("The paragraph format is controlled by the property 'ReferenceParagraphFormat' or 'ReferenceHeaderParagraphFormat' in the style file.")
-                                + "</html>",
-                        "", JOptionPane.ERROR_MESSAGE);
+
+        frame.getDialogService().showErrorDialogAndWait(Localization.lang("Undefined paragraph format"),
```
tobiasdiez: remove todo?

Siedlerchr: You were a second too early ;)

src/main/java/org/jabref/gui/push/PushToEmacs.java
tobiasdiez: Sorry, I don't understand what you mean. The push to application action only gets the `PushToApplicaiton` instance without knowing which concrete subclass it is. But this is not a problem as far as I can see since upon initialization each subclass would get a `DialogService` as a constructor argument.

Siedlerchr: The problem is: `PushToApplicationAction` already gets an instance of the concrete push operation and only calls `pushEntries.
And these push-Operations are defined and instantiated here. I can't just pass the DialogService as Constructor arg.
The only thing I could do is to create an additional setter for the Dialog Service

https://github.com/JabRef/jabref/blob/fd98056c76bc340ab7ed62944e13414b20d5a117/src/main/java/org/jabref/gui/push/PushToApplications.java#L16-L20

tobiasdiez: Why can't you pass the dialog service as a constructor argument to `PushToApplications`? It seems the only place where this class is initiated is https://github.com/JabRef/jabref/blob/fd98056c76bc340ab7ed62944e13414b20d5a117/src/main/java/org/jabref/gui/JabRefFrame.java#L580
where you have a dialog service at your disposal.

Siedlerchr: done

tobiasdiez:
Let's merge this before the size escalates ;-). Thanks a lot for your time and good work @Siedlerchr!

Issue #3561, opened 2017-12-20
dependency-updates-20171220111048


koppor:
* [ ] `com.microsoft.azure:applicationinsights-core:2.0.0-BETA`
* [ ] `com.microsoft.azure:applicationinsights-logging-log4j2:2.0.0-BETA`

tobiasdiez:
It appears that https://github.com/JabRef/jabref/pull/3559 reintroduced the problem that "page views" (i.e. dialogs) are not tracked. At least, the last version that reported this information was from the 19th. It could be, of course, that nobody yet opened a dialog in a newer version. 

Thus I reopen this issue. If you want to release a new version in the next few days, I would kindly ask you to revert https://github.com/JabRef/jabref/pull/3559. I'll keep an eye on the logs and will report in a few days again if dialogs are tracked or not.

PR #3559, opened 2017-12-20
Update applicationinsights from 1.0.9 -> 2.0.0-BETA

LinusDietz:
Can we update to this version @tobiasdiez?

Refs: https://github.com/JabRef/jabref/pull/3317

PR #3806, opened 2018-03-05
Update applicationinsights from 1.0.9 -> 2.0.1

LinusDietz:
@tobiasdiez this is another round of trying to update the applicationinsights.

refs https://github.com/JabRef/jabref/pull/3559, https://github.com/JabRef/jabref/pull/3317, https://github.com/JabRef/jabref/issues/3561, and https://github.com/JabRef/jabref/issues/3313

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -151,8 +151,8 @@ dependencies {
 
     compile 'com.github.tomtung:latex2unicode_2.12:0.2.2'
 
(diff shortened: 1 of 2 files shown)


Siedlerchr:
Seems like xmpUtil s has a wrong import for StringUtils.

tobiasdiez:
According to their changelog, the issues that prevented an upgrade are now fixed. Let's merge this and I'll have a look at the logs in a few days to check that the telemetry is indeed working as expected.

Issue #3818, opened 2018-03-08
Exception when trying to save linked file


erelsgl:
- [X ] I have tested the latest master version from http://builds.jabref.org/master/ and the problem persists

JabRef 4.2-dev--snapshot--2018-03-06--master--bb471a26d
Linux 4.4.0-104-generic amd64 
Java 1.8.0_161

Steps to reproduce:

1. Create an entry with a URL leading to a PDF file, like this: https://www.gla.ac.uk/media/media_452489_en.pdf
2. Right-click the entry in the entry table.
3. Choose "Copy linked files to folder...".
4. Choose your home folder.

<details>
copyfilesdialog.fxml not found, attempting with camel case
10:16:05.694 [JavaFX Application Thread] ERROR org.jabref.FallbackExceptionHandler - Uncaught exception occurred in Thread[JavaFX Application Thread,5,main]
java.lang.IllegalStateException: Cannot load org.jabref.gui.copyfiles.copyfilesdialog
	at com.airhacks.afterburner.views.FXMLView.loadSynchronously(FXMLView.java:101) ~[JabRef-4.2-dev--snapshot--2018-03-06--master--bb471a26d.jar:?]
	at com.airhacks.afterburner.views.FXMLView.initializeFXMLLoader(FXMLView.java:125) ~[JabRef-4.2-dev--snapshot--2018-03-06--master--bb471a26d.jar:?]
	at com.airhacks.afterburner.views.FXMLView.getView(FXMLView.java:137) ~[JabRef-4.2-dev--snapshot--2018-03-06--master--bb471a26d.jar:?]
	at org.jabref.gui.AbstractView.getView(AbstractView.java:28) ~[JabRef-4.2-dev--snapshot--2018-03-06--master--bb471a26d.jar:?]
	at org.jabref.gui.copyfiles.CopyFilesDialogView.show(CopyFilesDialogView.java:25) ~[JabRef-4.2-dev--snapshot--2018-03-06--master--bb471a26d.jar:?]
	at org.jabref.gui.copyfiles.CopyFilesAction.showDialog(CopyFilesAction.java:63) ~[JabRef-4.2-dev--snapshot--2018-03-06--master--bb471a26d.jar:?]
	at org.jabref.gui.copyfiles.CopyFilesAction.lambda$startServiceAndshowProgessDialog$3(CopyFilesAction.java:58) ~[JabRef-4.2-dev--snapshot--2018-03-06--master--bb471a26d.jar:?]
[...]

PR #3824, opened 2018-03-09
Provides download option in context menu and fixes #3614

### LINKED ISSUE #3614: Download no longer shows progress

recent master

Steps to reproduce:

1. Create entry and go to General tab
2. Click "Download from URL" button and insert a link to a relatively large PDF. Click ok.
3. When the download starts an empty progress bar is displayed. However, the progress is not updated and the progress indicator is hidden immediately.

tobiasdiez:
Inspired by the discussion in #3818, I added a "Download file" option in the context menu of a linked file (if it is an online link). Moreover, the bug #3614 is fixed.

This PR was brought to you with the support of coffee from the JabRef cup!

----

### DIFF
--- a/src/main/java/org/jabref/gui/fieldeditors/FieldEditors.java
+++ b/src/main/java/org/jabref/gui/fieldeditors/FieldEditors.java
@@ -51,7 +51,7 @@ public class FieldEditors {
         } else if (fieldExtras.contains(FieldProperty.OWNER)) {
             return new OwnerEditor(fieldName, preferences, suggestionProvider, fieldCheckers);
         } else if (fieldExtras.contains(FieldProperty.FILE_EDITOR)) {
-            return new LinkedFilesEditor(fieldName, dialogService, databaseContext, taskExecutor, suggestionProvider, fieldCheckers);
+            return new LinkedFilesEditor(fieldName, dialogService, databaseContext, taskExecutor, suggestionProvider, fieldCheckers, preferences);
         } else if (fieldExtras.contains(FieldProperty.YES_NO)) {
             return new OptionEditor<>(new YesNoEditorViewModel(fieldName, suggestionProvider, fieldCheckers));
         } else if (fieldExtras.contains(FieldProperty.MONTH)) {
--- a/src/main/java/org/jabref/gui/fieldeditors/LinkedFileViewModel.java
+++ b/src/main/java/org/jabref/gui/fieldeditors/LinkedFileViewModel.java
@@ -76,6 +76,9 @@ public class LinkedFileViewModel extends AbstractViewModel {
     private final XmpPreferences xmpPreferences;
     private final String fileNamePattern;
 
+    /**
+     * @deprecated use {@link #LinkedFileViewModel(LinkedFile, BibEntry, BibDatabaseContext, TaskExecutor, DialogService, JabRefPreferences, JournalAbbreviationLoader)} instead
+     */
     @Deprecated
     public LinkedFileViewModel(LinkedFile linkedFile, BibEntry entry, BibDatabaseContext databaseContext, TaskExecutor taskExecutor) {
         this(linkedFile, entry, databaseContext, taskExecutor, new FXDialogService(), Globals.prefs, Globals.journalAbbreviationLoader);
--- a/src/main/java/org/jabref/gui/fieldeditors/LinkedFileViewModel.java
+++ b/src/main/java/org/jabref/gui/fieldeditors/LinkedFileViewModel.java
@@ -1,6 +1,7 @@
 package org.jabref.gui.fieldeditors;
 
 import java.io.IOException;
+import java.net.MalformedURLException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -17,6 +18,7 @@ import javafx.beans.property.BooleanProperty;
 import javafx.beans.property.DoubleProperty;
 import javafx.beans.property.SimpleBooleanProperty;
 import javafx.beans.property.SimpleDoubleProperty;
+import javafx.beans.property.StringProperty;
 import javafx.scene.control.Alert.AlertType;
 import javafx.scene.control.ButtonType;
 
@@ -25,6 +27,8 @@ import org.jabref.gui.AbstractViewModel;
 import org.jabref.gui.DialogService;
 import org.jabref.gui.FXDialogService;
 import org.jabref.gui.desktop.JabRefDesktop;
+import org.jabref.gui.externalfiles.DownloadExternalFile;
+import org.jabref.gui.externalfiles.FileDownloadTask;
 import org.jabref.gui.externalfiletype.ExternalFileType;
 import org.jabref.gui.externalfiletype.ExternalFileTypes;
 import org.jabref.gui.filelist.FileListEntryEditor;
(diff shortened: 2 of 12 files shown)


src/main/java/org/jabref/gui/fieldeditors/LinkedFileViewModel.java
```diff
@@ -345,4 +352,113 @@ public void writeXMPMetadata() {
...
+    private Path constructSuggestedPath(Optional<ExternalFileType> suggestedType, List<Path> fileDirectories) {
+        String suffix = suggestedType.map(ExternalFileType::getExtension).orElse("");
+        String suggestedName = getSuggestedFileName(suffix);
+        Path directory;
+        if (fileDirectories.isEmpty()) {
+            directory = null;
+        } else {
+            directory = fileDirectories.get(0);
+        }
+        final Path suggestDir = directory == null ? Paths.get(System.getProperty("user.home")) : directory;
```
Siedlerchr: There is a method which always returns a file dir: getfirstExistingFileDir

src/main/java/org/jabref/gui/fieldeditors/LinkedFileViewModel.java
```diff
@@ -345,4 +352,113 @@ public void writeXMPMetadata() {
...
+        * [ 1548875 ] download pdf produces unsupported filename
+        *
+        * http://sourceforge.net/tracker/index.php?func=detail&aid=1548875&group_id=92314&atid=600306
+        * FIXME: rework this! just allow alphanumeric stuff or so?
+        * https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247(v=vs.85).aspx#naming_conventions
+        * http://superuser.com/questions/358855/what-characters-are-safe-in-cross-platform-file-names-for-linux-windows-and-os
+        * https://support.apple.com/en-us/HT202808
+        */
+        if (OS.WINDOWS) {
+            plannedName = plannedName.replaceAll("\\?|\\*|\\<|\\>|\\||\\\"|\\:|\\.$|\\[|\\]", "");
```
Siedlerchr: extract to a matcher?

src/main/java/org/jabref/gui/fieldeditors/LinkedFilesEditorViewModel.java
```diff
@@ -44,25 +39,23 @@
...
-
     private final ListProperty<LinkedFileViewModel> files = new SimpleListProperty<>(FXCollections.observableArrayList(LinkedFileViewModel::getObservables));
     private final BooleanProperty fulltextLookupInProgress = new SimpleBooleanProperty(false);
     private final DialogService dialogService;
     private final BibDatabaseContext databaseContext;
     private final TaskExecutor taskExecutor;
+    private final JabRefPreferences preferences;
 
-    public LinkedFilesEditorViewModel(String fieldName, AutoCompleteSuggestionProvider<?> suggestionProvider, DialogService dialogService, BibDatabaseContext databaseContext, TaskExecutor taskExecutor, FieldCheckers fieldCheckers) {
+    public LinkedFilesEditorViewModel(String fieldName, AutoCompleteSuggestionProvider<?> suggestionProvider, DialogService dialogService, BibDatabaseContext databaseContext, TaskExecutor taskExecutor, FieldCheckers fieldCheckers, JabRefPreferences preferences) {
```
Siedlerchr: I know that's its not taht good style to have methods with long params, but why not directly use the FileDirectoryPreferences and ImportFormatPrefernces as args?

src/main/java/org/jabref/gui/fieldeditors/LinkedFileViewModel.java
tobiasdiez: It turns out that this code is not needed since the `FileUtil` class already handles cleanup of file names

src/main/java/org/jabref/gui/fieldeditors/LinkedFilesEditorViewModel.java
tobiasdiez: Mhhh, I tried to collect all the preferences accesses and even for the LinkedFileViewModel it were 5 different settings that are needed. Thus I prefer to keep the general preferences class as dependency.

Siedlerchr:
If you fix checkstyle we can merge

src/main/java/org/jabref/gui/fieldeditors/FieldEditors.java
```diff
@@ -51,7 +51,7 @@ public static FieldEditorFX getForField(String fieldName, TaskExecutor taskExecu
         } else if (fieldExtras.contains(FieldProperty.OWNER)) {
             return new OwnerEditor(fieldName, preferences, suggestionProvider, fieldCheckers);
         } else if (fieldExtras.contains(FieldProperty.FILE_EDITOR)) {
-            return new LinkedFilesEditor(fieldName, dialogService, databaseContext, taskExecutor, suggestionProvider, fieldCheckers);
+            return new LinkedFilesEditor(fieldName, dialogService, databaseContext, taskExecutor, suggestionProvider, fieldCheckers, preferences);
```
koppor: Do really the **whole** preferences need to be passed? Didn't we want to implement information hiding here? 😇

src/main/java/org/jabref/gui/fieldeditors/LinkedFileViewModel.java
```diff
@@ -60,20 +70,32 @@
     private final DialogService dialogService;
     private final BibEntry entry;
     private final TaskExecutor taskExecutor;
+    private final FileDirectoryPreferences fileDirectoryPreferences;
+    private final CleanupPreferences cleanupPreferences;
+    private final LayoutFormatterPreferences layoutFormatterPreferences;
+    private final XmpPreferences xmpPreferences;
+    private final String fileNamePattern;
 
+    @Deprecated
```
koppor: Please add a JavaDoc pointing to the method one should use from now on.

src/main/java/org/jabref/gui/fieldeditors/LinkedFileViewModel.java
```diff
@@ -60,20 +70,32 @@
...
     public LinkedFileViewModel(LinkedFile linkedFile, BibEntry entry, BibDatabaseContext databaseContext, TaskExecutor taskExecutor) {
-        this(linkedFile, entry, databaseContext, taskExecutor, new FXDialogService());
+        this(linkedFile, entry, databaseContext, taskExecutor, new FXDialogService(), Globals.prefs, Globals.journalAbbreviationLoader);
     }
 
-    protected LinkedFileViewModel(LinkedFile linkedFile, BibEntry entry, BibDatabaseContext databaseContext,
-                                  TaskExecutor taskExecutor, DialogService dialogService) {
+    public LinkedFileViewModel(LinkedFile linkedFile, BibEntry entry, BibDatabaseContext databaseContext,
```
koppor: I think, only the `IMPORT_FILENAMEPATTERN` and `preferences.getFileDirectoryPreferences()` are needed. Why not pass them both (as quick fix regarding the information hiding). OK, a new preference object would even be better, but that costs more time.

src/main/java/org/jabref/gui/fieldeditors/LinkedFileViewModel.java
```diff
@@ -177,21 +199,21 @@ public void rename() {
...
-        Optional<Path> file = linkedFile.findIn(databaseContext, Globals.prefs.getFileDirectoryPreferences());
+        Optional<Path> file = linkedFile.findIn(databaseContext, fileDirectoryPreferences);
         if ((file.isPresent()) && Files.exists(file.get())) {
             RenamePdfCleanup pdfCleanup = new RenamePdfCleanup(false,
                     databaseContext,
-                    Globals.prefs.getCleanupPreferences(Globals.journalAbbreviationLoader).getFileNamePattern(),
-                    Globals.prefs.getLayoutFormatterPreferences(Globals.journalAbbreviationLoader),
-                    Globals.prefs.getFileDirectoryPreferences(), linkedFile);
+                    cleanupPreferences.getFileNamePattern(),
+                    layoutFormatterPreferences,
```
Siedlerchr: Actually layoutFormatterPrefernces can be removed from all cleanup ops (did this the maintable beta), but out of scope for this pr

src/main/java/org/jabref/gui/fieldeditors/LinkedFileViewModel.java
tobiasdiez: Also the auto-link preferences as well as the import preferences are needed. You are definitely right, a custom preference object unifying them all would be nice. To be honest, I don't have the motivation right now to create one (especially since the focus of this PR is a different one).

src/main/java/org/jabref/gui/fieldeditors/LinkedFileViewModel.java
tobiasdiez: Good observation. Fixed.

PR #3827, opened 2018-03-09
Update Eclipse style to intellij style

### LINKED ISSUE #3815: Adjust eclipse style to intellij (line wrapping)
Has intellij one or two default indendation lines? @tobiasdiez 
I now changed to align on column

[JabRefModdedStyleSettings.xml.txt](https://github.com/JabRef/jabref/files/1791055/JabRefModdedStyleSettings.xml.txt)

grafik

grafik

grafik

grafik

Siedlerchr:
Fixes #3815 
Screenshots of indentation in #3815 

(cherry pick to master afterwards)

----


config/README.md
```diff
@@ -15,3 +15,16 @@ Style-checks are done for each pull request and installing this cody style confi
 10. Press "OK"
 
 * Please let `.editorconfig` override the settings of IntelliJ
+
+
+# Eclipse:
+
+In case the Eclipse style needs to be adjusted for the eclipse.gradle file:
```
koppor: I don't understand that. Maybe, the gradle script needs to be rewritten to inject an exported XML file`?

Siedlerchr: I think I just found a simpler way. I will update the doc accordingly

Siedlerchr: I generally found a nice and simple way.  One can directly export the file in eclipse as EPF file in properties format
The main drawback is that it's not possible to import it with gradle (argh!).
So therefore the only solution would be to read the file, remove all unnecessary lines and write that into the .prefs file (ideally once). Then we can just copy over the prefs file to the workspace.

config/Eclipse Code Style.epf
```diff
@@ -0,0 +1,328 @@
+#Fri Mar 09 22:05:27 CET 2018
```
tobiasdiez: Is this file needed?

Siedlerchr: Yes, as a reference file. And this file can still be imported manually. 
Mabye someone more experienced in gradle/scripting can create a simple script to copy the things from that file to the workspace org.eclipse.jdt.ui.prefs file

koppor:
Gradle uses Groovy, which is similar to Java. You can use the normal Java
classes there. Appending with Java shouldn't be that hard, is it? 😇

Am 10.03.2018 15:25 schrieb "Christoph" <notifications@github.com>:

Siedlerchr:
I merge this for the moment. As we don't change our code style that often we can postpone the automatic creation of the prefs

PR #3839, opened 2018-03-12
Consistent FX color scheme for JabRef

halirutan:
The PR aims to make the look and feel of JabRef more consistent by providing a well-chosen set of base colors and deriving other colors from it. At the moment, I used a flat and bright scheme with the original JabRef blue as the theme color.

scrShot

In addition to the changes in the FX CSS files, I provide an additional true type font for the application icons that are not available in the Material Design Icon set. It can be loaded and used in exactly the same way as the Material Design Icons and only small changes to the code where necessary. As soon as we agreed on using this, I can make further adjustments and additions (currently, WinEd is missing) to the icons so that they fit perfectly into our toolbar. The one-colored icons should represent the original icon as much as possible and should be recognizable. The will look like this

i1 i2 i3 i4

For those who would like to help to work on this:

Our base style is `modena.css` and to make our restyling easy to use and to understand, we should try to redefine the color variables of `modena.css` as much as possible instead of explicitly introducing and using our own variables. This means, don't use explicit color settings in the definition of styles for a control. If it is a commonly used property e.g. the border color, try to find and overwrite the appropriate modena-variable. If you indeed need a specific color, then re-use the ones defined at the top of `Main.css`.

Remember that this is unfinished at the moment and I have rather commented out lines than simply deleting it.

----


lenhard:
I played around with the UI a little and I like it. I like the fact that the menu and tool bars are a little less color-intense than in the current maintable-beta.

You have my +1 for continuing with this line of work.

halirutan:
I made further improvements to the custom icons and they look at least acceptable now if put on the toolbar:

toolbar

For more information on this view issue https://github.com/JabRef/jabref/issues/3868. I pulled the latest changes from maintable-beta. I'm giving this free for review now. Note that there are still a lot of commented lines in the code to easily see what I have adjusted. Once we agree on something, I'll clean everything up.

src/main/java/org/jabref/gui/IconTheme.java
```diff
@@ -261,13 +264,13 @@ public static URL getIconUrl(String name) {
...
-        FILE_OPENOFFICE(IconTheme.getImage("openoffice")),
-        APPLICATION_EMACS(IconTheme.getImage("emacs")),
-        APPLICATION_LYX(IconTheme.getImage("lyx")),
-        APPLICATION_TEXSTUDIO(IconTheme.getImage("texstudio")),
-        APPLICATION_TEXMAKER(IconTheme.getImage("texmaker")),
-        APPLICATION_VIM(IconTheme.getImage("vim")),
-        APPLICATION_WINEDT(IconTheme.getImage("winedt")),
+        FILE_OPENOFFICE(JabRefMaterialDesignIcon.OPEN_OFFICE),
+        APPLICATION_EMACS(JabRefMaterialDesignIcon.EMACS),
+        APPLICATION_LYX(JabRefMaterialDesignIcon.LYX),
```
tobiasdiez: Are there other places that still use `IconTheme.getImage`? If not this method can be removed as well as `InternalFileIcon`.

src/main/java/org/jabref/gui/Main.css
```diff
@@ -1,69 +1,243 @@
...
+
+    /*-jr-theme: #2467ab;*/
+
+    /* This theme is the original JabRef dark blue color */
+    -jr-theme: #50618F;
+    /* The base gray. Most gray-tones in the application are derived from this color. */
+    -jr-base: #f0f0f0;
+
+    -jr-white: #ffffff;
+    -jr-gray-00: #f5f5f5;
```
tobiasdiez: what's the reason for the `00` suffix?

src/main/java/org/jabref/gui/Main.css
```diff
@@ -1,69 +1,243 @@
...
+    -jr-error: -jr-light-red;
+
+    /* Background-color for the small group view indicator for the number of hits */
+    -jr-group-hits-bg: -jr-gray-0;
+
+    /* Specific color for general tooltips */
+    -jr-tooltip-bg: -jr-info;
+    -jr-tooltip-fg: -jr-black;
+
+    /* Consistent size for headers of tab-pane and side-panels*/
```
tobiasdiez: Move this down to the end, since right now it appears in the middle of color definitions?

src/main/java/org/jabref/gui/Main.css
```diff
@@ -136,22 +323,22 @@
...
-    -fx-background-color: white;
+    -fx-background-color: -fx-body-color;
     -fx-background-insets: 0;
     -fx-background-radius: 0;
 }
 
-.menu-bar > .container > .menu-button > .label {
-    -fx-padding: 0.41777em 0.41777em 0.41777em 0.41777em; /* 5 5 5 5 */
-}
+/*.menu-bar > .container > .menu-button > .label {*/
```
tobiasdiez: What's the reason for uncommenting these menu bar customization? If I remember correctly, they make the menu items a bit bigger to align with the MaterialDesing specifications.

src/main/java/org/jabref/gui/Main.css
```diff
@@ -161,24 +348,54 @@
...
+    -fx-border-color: -jr-theme;
+    -fx-fill: -jr-theme;
+    -fx-text-fill: -jr-theme;
 }
 
 .tab-pane > .tab-header-area > .headers-region > .tab:selected {
-    -fx-background-color: -fx-light-background;
-    -fx-border-color: -fx-active-background;
+    /*-fx-background-color: -fx-control-inner-background;*/
+    -fx-border-color: -jr-theme;
```
tobiasdiez: I think it makes sense to introduce two new colors `-jr-accent` and `jr-accent-text` and use them for all selected/active states.

src/main/java/org/jabref/gui/Main.css
```diff
@@ -308,21 +526,21 @@
...
-    -fx-background-color: derive(#868786, 20%);
+    /*-fx-background-color: derive(-fx-outer-border, 10%);*/
     -fx-opacity: 0.2;
     -fx-background-radius: 0em;
 }
 
 .scroll-bar:horizontal .thumb,
 .scroll-bar:vertical .thumb {
-    -fx-background-color: derive(#757575, 60%);
+    -fx-background-color: -fx-outer-border;
```
tobiasdiez: Introduce a new color for the scrollbars? I could imagine a theme where the scrollbars pop-out a bit more.

src/main/java/org/jabref/gui/Main.css
```diff
@@ -365,38 +583,63 @@
...
 }
 
-.menu-bar > .container > .menu-button:hover,
-.menu-bar > .container > .menu-button:focused,
-.menu-bar > .container > .menu-button:showing {
-    -fx-background: derive(-fx-theme-base, 10%);
+.sidePaneComponentHeader .glyph-icon{
+    -fx-font-size: 16px;
 }
+/*.mainMenu {*/
```
tobiasdiez: I would prefer if you could reactivate these menu customizations.

src/main/java/org/jabref/gui/SidePaneComponent.java
```diff
@@ -101,8 +101,10 @@ public final Node getHeader() {
         BorderPane graphic = new BorderPane();
         graphic.setCenter(icon.getGraphicNode());
         BorderPane container = new BorderPane();
-        container.setLeft(graphic);
-        container.setCenter(new Label(title));
+//        container.setLeft(graphic);
+        final Label label = new Label(title);
+        label.getStyleClass().add("sidePaneComponentHeader");
```
tobiasdiez: I think it is confusing to use the same style class for two elements. The css selector `sidePaneComponentHeader > .label` should work.

src/main/java/org/jabref/gui/entryeditor/EntryEditor.css
```diff
@@ -10,12 +10,12 @@
 .editorPane {
     -fx-hgap: 10;
     -fx-vgap: 6;
-    -fx-background-color: -fx-light-background;
+    /*-fx-background-color: -jr-white;*/
```
tobiasdiez: Please make the editor background white again...

src/main/java/org/jabref/gui/entryeditor/EntryEditor.css
```diff
@@ -78,37 +86,54 @@
...
+    /*-fx-border-radius: 2 2 2 2;*/
+    /*-fx-background-radius: 2 2 2 2;*/
+    /*-fx-text-fill: -jr-icon;*/
+    /*-fx-fill: -jr-icon;*/
 }
 
+
 .tool-bar > .container > .separator {
-    -fx-background-color: #5e5f5e;
+    /*-fx-background-color: -jr-separator;*/
```
tobiasdiez: Reactivate?

src/main/java/org/jabref/gui/groups/GroupTree.css
```diff
@@ -1,10 +1,10 @@
 .tree-table-view, .tree-table-row-cell {
-    -fx-background-color: #dadad8;
+    -fx-background-color: -fx-control-inner-background;
```
tobiasdiez: I would introduce a new color `-jr-sidepane-background` here.

src/main/java/org/jabref/gui/groups/GroupTree.css
```diff
@@ -1,10 +1,10 @@
...
     -fx-table-cell-border-color: transparent; /* hide grid lines */
 }
 
 .tree-table-cell {
-    -fx-font-size: 105%;
+    /*-fx-font-size: 105%;*/
```
tobiasdiez: That's probably personal preference, but I prefer the bigger font size (and smaller for hits).

src/main/java/org/jabref/gui/groups/GroupTree.css
```diff
@@ -14,39 +14,39 @@
...
+    -fx-background-color: -jr-green;
 }
 
 .disclosureNodeColumn {
     -fx-alignment: top-right;
 }
 
 .tree-table-row-cell:dragOver-bottom {
-    -fx-border-color: #eea82f;
+    -fx-border-color: -jr-yellow;
```
tobiasdiez: this should reuse the same active/selected colors as for the main table, right?

src/main/java/org/jabref/gui/groups/GroupTree.css
```diff
@@ -76,9 +76,9 @@
     -fx-padding: 0.45em 0.2em 0.45em 0.2em;
 }
 
-.tree-table-row-cell:selected > .tree-table-cell > .tree-disclosure-node > .arrow {
-    -fx-background-color: white;
-}
+/*.tree-table-row-cell:selected > .tree-table-cell > .tree-disclosure-node > .arrow {*/
+    /*-fx-background-color: -jr-white;*/
```
tobiasdiez: In your screenshot, the arrow is never visible accept on hover because it is also white...

src/main/java/org/jabref/gui/maintable/MainTable.css
```diff
@@ -2,48 +2,51 @@
...
     -fx-alignment: center-left;
-    -fx-text-fill: -fx-unimportant;
+    -fx-text-fill: -jr-theme;
 }
 
 .table-view .column-header .glyph-icon {
     -fx-alignment: baseline-center;
-    -fx-text-fill: -fx-unimportant;
-    -fx-fill: -fx-unimportant;
+    -fx-text-fill: -jr-icon;
```
tobiasdiez: use fx-mid-text instead?

src/main/java/org/jabref/gui/maintable/MainTable.css
```diff
@@ -2,48 +2,51 @@
...
 
 .table-cell {
     -fx-padding: 0.5em 1em 0.5em 1em;
     -fx-cell-size: 4.0em;
 }
 
-.table-row-cell:odd {
-    -fx-background: -fx-control-inner-background;
-}
+/*.table-row-cell:odd {*/
```
tobiasdiez: Please reactivate this, because otherwise the rows are styled in a (ugly) strip-pattern.

Siedlerchr: I disagree. This is a usabiltily enhancement to difference between odd and even rows. Makes it much easier to see which entry is in which row.  So called "Zebra Striping"
http://uxmovement.com/content/9-design-techniques-for-user-friendly-tables/
https://alistapart.com/article/zebrastripingmoredataforthecase

Siedlerchr: See also here for some general ideas about desinging data tables:
https://uxdesign.cc/design-better-data-tables-4ecc99d23356

src/main/java/org/jabref/gui/IconTheme.java
halirutan: Refactored this out. I initially left it because maybe someone still needs a pixel icon in future. But since it is really easy now to add SVG icons if we indeed don't find anything appropriate in the existing ones, this can be removed.

src/main/java/org/jabref/gui/IconTheme.java
```diff
@@ -60,6 +61,8 @@
     }
 
     private static InputStream getMaterialDesignIconsStream() {
+        // TODO: The next line loads the additional JabRef-Icons. Most certainly wrong at this place
```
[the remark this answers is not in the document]
halirutan: I refactored this and used the same approach we (and Material Design Icons as well) use. It's now loaded the static initializer of the `IconTheme` class itself.

src/main/java/org/jabref/gui/Main.css
halirutan: It doesn't look good. Even the default padding is larger than it is in other applications, but it looks OK. Compare these two images that show additionally the header of the mail program. First, as I have it now

current

And here with the additional padding

old

The menu items look lost in the menu bar. The spacing is far too large. I really would like to keep the default setting.

Btw, in the colorful version, we have now on `maintable-beta`, this kind of made sense because it tried to mimic the look of today's web-pages with their very large top menus. If you decide to work on a different scheme, the large padding might be an option again. But in this scheme, it doesn't work and I'd like to keep it simple and unsurprising for the user.

src/main/java/org/jabref/gui/groups/GroupTree.css
halirutan: I'm against this. Although the font is only one pixel larger on my system and the anti-aliasing is also OK, it introduces a very notable font-difference to the main-table where there should be consistency.  I would accept if this is wanted by most devs, but from my point of view, it is not OK.

src/main/java/org/jabref/gui/groups/GroupTree.css
halirutan: OK, for a scheme that wants to set a different bg for the sidepane than in other controls this makes sense.

src/main/java/org/jabref/gui/entryeditor/EntryEditor.css
halirutan: I would leave it like that. The current separator is derived from the main colors and looks good. `-jr-separator` is not used (or even defined) anymore. It was from my first tests, where I tried to define these myself. When cleaning up, I'm going to remove it.

src/main/java/org/jabref/gui/Main.css
tobiasdiez: I still like the second screenshot more and find it more touch-friendly. But, well, we just have different tastes and I definitely understand your viewpoint.
Probably every dev should create an own theme and we then vote which should be the default 😸

src/main/java/org/jabref/gui/groups/GroupTree.css
tobiasdiez: Ok, convinced. The base font size should be the same. But I still think the hit-number can be smaller because this is kind of unimportant info.

By the way, did you change the padding of the tree node items? They have a bit tighter feeling than on the current master/maintable-beta, which probably results not from the difference of font-sizes.

src/main/java/org/jabref/gui/maintable/MainTable.css
tobiasdiez: Well, I don't want to start a rant about zebras (they are quite beautiful animals), but anyway here is the rant:
First of all, there are indeed a few studies (like [this one](http://alistapart.com/article/zebrastripingmoredataforthecase)) which show that zebra strips help users in some cases. One should note that the improvement in performance is relatively small (a few 1-3%) and only when the user is required to move vertically in a row to look for data. It is also still controversial whether the better performance comes from a superior design or just because testers where used to the striped style. On the other hand, zebra strips actually lead to worse performance when the task is to quickly identify a desired row containing given data because the eye takes these lines as fix points instead of scan quickly through the dataset. 

So zebra strips may have a place in bigger datasheets where it is required to jump in a specific row and the data spans multiple columns. We have, on the other hand, a very small dataset with like 20 entries on display and 5 or so columns. Moreover, and this is probably the biggest point, a row is highlighted when selected or hovered and thus you get all the benefits of the zebra style without the clean look of the plain table. 

Sorry for the rant but this is one of the things where people just follow the general advice in the form "zebra strips help to navigate in data table" without asking what are the boundary conditions for the performance increase and if they apply in the present case. This leads to a lot of documents /blog posts citing this as a always-good style to present data in a table format.

Siedlerchr: lets ask the other @JabRef/developers  on their opinion on this. I currently find it very hard to identify an entry in the list.  Especially, if you have maybe some entries in which not all required fields are visible. Compare this to the screenshot in the first post. The version from @halirutan  looks much better with the stripes. 
This is also the case when you have multiple files in the list view.

grafik

lenhard: I give my support for using zebra stripping.

tobiasdiez:
I've now merged this PR. We can fine-tune the design later and I'll provide a proposal for a dark theme in a new PR.

Issue #1975, opened 2016-09-13
Reuse query parser


tobiasdiez:
Instead of parsing a search query using a custom grammar we should reuse some existing query parser. For example, [Lucene's query parser](https://lucene.apache.org/core/2_9_4/queryparsersyntax.html) with [QueryParser.parse](https://lucene.apache.org/core/4_0_0/queryparser/org/apache/lucene/queryparser/classic/QueryParserBase.html#parse%28java.lang.String%29).

Triggered by https://github.com/JabRef/jabref/issues/1633#issuecomment-238762667. 
- Description of Lucene's syntax: http://www.solrtutorial.com/solr-query-syntax.html and https://lucene.apache.org/core/2_9_4/queryparsersyntax.html
- Lucene provides a few classes which let us modify the search syntax: https://lucene.apache.org/core/4_1_0/queryparser/index.html?org/apache/lucene/queryparser/flexible/core/parser/SyntaxParser.html
- Or we use the default Lucene syntax: https://lucene.apache.org/core/4_0_0/queryparser/org/apache/lucene/queryparser/classic/QueryParser.html

simonharrer:
Be aware that a lot of groups rely on the current search query grammar, and they will break if this is changed.

Le-wi:
So introducing the new parser would require to migrate all the old queries as well?

tobiasdiez:
The only groups effected are the "search based groups". I doubt that many users use them (and then probably only experienced users). Thus, I would say that displaying a warning dialog of the kind "some groups need to be updated by hand to work properly" is sufficient. @JabRef/developers opinions?

koppor:
Steps:

1. Dive into our powerful search syntax: https://docs.jabref.org/finding-sorting-and-cleaning-entries/search
2. Compare with lucene
3. Evaluate whether a switch is possible
4. Get rid off ANTLRv4
5. Add an ADR

koppor:
@DominikVoigt implemented a lucene parsing at https://github.com/JabRef/jabref/pull/6799. This could be a good basis to solve this issue.

koppor:
The follow-up ticket https://github.com/koppor/jabref/issues/341 should also be linked here.

koppor:
Decision note: During GSoC 2024 we found out that JabRef's search syntax covers more than existing search syntax. Moreover Apache Lucene syntax was strange. --> https://github.com/JabRef/jabref/pull/11803

PR #3871, opened 2018-03-21
Switch colors of search icon for the two search modes

### LINKED ISSUE #3870: Magnifier changes colors wrong
JabRef offers two search modes: normal search and the grammar based ("advanced") search. See https://help.jabref.org/en/Search#search-modes.

In JabRef 3.6, the initial search was the normal search. This changed in JabRef 3.7 somehow. This seems to be wrong.

Action points:

1. Dig into the change made between JabRef 3.6 and 3.7
2. Find out whether the change is right or wrong. Current assumption: It was wrong
3. Take appropriate action.

Refs https://github.com/JabRef/jabref/issues/3535 and the fix https://github.com/JabRef/jabref/pull/3708

lenhard:
Fixes #3870

----


koppor:
Grammar-based search is the advanced search. So, the magnifier should be cyan, when following search string is active:

    author=kolb and author=kopp

Taking examples from the documentation (https://help.jabref.org/en/Search#search-modes)

All of these are advanced, but they are not recognized as such:

    title|keywords = "image processing"

    (author = miller or title|keywords = "image processing") and not author = brown

Following is normal mode:

    progress “marine aquaculture”progress “marine aquaculture”

This is now with cyan. This is wrong.

Following is normal mode:

    progress

Correctly recognized as normal mode.

stefan-kolb:
If so, the whole detection is wrong?!

koppor:
Only the beginning of the detection is somehow wrong. For complex terms it is right, but for single words, it is wrong. See the examples above. Maybe, we changed something when going from JabRef 3.6 to 3.7.

lenhard:
Or maybe the detection never was entirely correct. If I have a lot of time on my hands, I might have a look at the detection algorithm. Otherwise we can close this PR and / or remove the coloring of the the search icon again. After all, what is the point of changing its color if the circumstances are incorrect?

tobiasdiez:
@lenhard If you indeed have time to look at the search algorithm, please replace it completely ;-) https://github.com/JabRef/jabref/issues/1975

koppor:
It worked in 3.6 and no one noticed that it broke at 3.7. So, the point is
to keep the idea and to fix it... The idea in other words: make users aware
of the feature and that the search might not work if the grammar based
search was intended, but not matched (typo in query)

Am 22.03.2018 19:53 schrieb "Jörg Lenhard" <notifications@github.com>:

Or maybe the detection never was entirely correct. If I have a lot of time
on my hands, I might have a look at the detection algorithm. Otherwise we
can close this PR and / or remove the coloring of the the search icon
again. After all, what is the point of changing its color if the
circumstances are incorrect?

—
You are receiving this because you commented.
<https://github.com/JabRef/jabref/pull/3871#issuecomment-375419286>, or mute
the thread

stefan-kolb:
Actually there should be tests for it then, as this is not a UI feature but the detection is logic based...

lenhard:
Did we update the antlr stuff after v3.6? Because the query "progress" actually is a valid grammar-based query that compiles with our antlr code. That's the reason for the "wrong" coloring. 

I have now added code that checks if a query consists only of word / whitespace characters and/or digits and uses a normal contains-based query for that. Thus, grammar-based searches are only executed when special symbols are used ALTHOUGH our grammar is more powerfull than that. The terms mentioned by @koppor above are now colored as expected. @koppor: I hope this limitation is what is wanted. Please do some testing to see if this matches your expectations.

src/main/java/org/jabref/logic/search/SearchQuery.java
```diff
@@ -85,7 +85,7 @@ private String getLocalizedRegularExpressionDescription() {
     }
 
     public boolean isGrammarBasedSearch() {
-        return rule instanceof GrammarBasedSearchRule;
+        return rule instanceof GrammarBasedSearchRule || rule instanceof RegexBasedSearchRule;
```
koppor: Should regex really be treated as grammar-based search? I would not treat it like that, because regex search is orthogonal to grammar-based search. One can activate it in addition.

src/main/java/org/jabref/model/search/rules/SearchRules.java
```diff
@@ -2,16 +2,20 @@
...
     }
 
+    private static final Pattern simpleExpression = Pattern.compile("(\\w|\\d|\\s)*");
+
     /**
      * Returns the appropriate search rule that fits best to the given parameter.
      */
     public static SearchRule getSearchRuleByQuery(String query, boolean caseSensitive, boolean regex) {
-        if (StringUtil.isBlank(query)) {
+        if (isSimpleQuery(query)) {
```
koppor: Reads good!

src/main/java/org/jabref/model/search/rules/SearchRules.java
```diff
@@ -25,6 +29,10 @@ public static SearchRule getSearchRuleByQuery(String query, boolean caseSensitiv
         }
     }
 
+    private static boolean isSimpleQuery(String query) {
+        return StringUtil.isBlank(query) || simpleExpression.matcher(query).matches();
```
koppor: Do you really need both? The current regex has a star (`*`) at the end, which means zero (`0`) ore more matches. So, the empty string should match, too.

src/main/java/org/jabref/logic/search/SearchQuery.java
lenhard: At least in the documentation: https://help.jabref.org/en/Search#search-modes they are listed under the heading of "advanced search". But I can easily change that of course.

koppor: I updated the help accordingly to also reflect that regular expressions are orthogonal. 😇 

Could you additionally add a comment there that this is the "Advanced Search" described in the help? I know, this contradicts [DDD](https://en.wikipedia.org/wiki/Domain-driven_design), but I think, it is OK in this place.

koppor:
Not what I expect:

`j` - OK:

grafik

`jö` - not OK:

grafik

Minor issue:

When I clear the field using Ctrl+A and then Delete (after an advanced search), the marker is still advanced:

grafik

Otherwise: LGTM :)

src/main/java/org/jabref/logic/search/SearchQuery.java
lenhard: Ok, I adapted the code accordingly.

src/main/java/org/jabref/model/search/rules/SearchRules.java
lenhard: Good suggestion, that sounds reasonable. I removed the StringUtil dependency and as far as I can see there is no difference in the search.

lenhard:
Thanks for the comments @koppor 

I fixed all the glitches you described and this PR should now be ready.

PR #3872, opened 2018-03-21
Fix freezing by replacing codeArea with TextArea

Siedlerchr:
I looked a bit more at the freezing problem and could nail it down to the CodeArea.
I replaced the CodeArea with a simple TextArea and now I don't get any freezes anylonger 

We don't use any extensive features of the TextArea, except for some styling.

We should adapt this for the 4.2 release as a workaround. As we don't have those freezes in the maintable-beta, I am in strong favor of using this.

I did some heavy clicking and scrolling and did not receive any freezes.

----

### DIFF
--- a/src/main/java/org/jabref/gui/BasePanel.java
+++ b/src/main/java/org/jabref/gui/BasePanel.java
@@ -1086,7 +1086,8 @@ public class BasePanel extends JPanel implements ClipboardOwner {
     }
 
     private boolean saveDatabase(File file, boolean selectedOnly, Charset enc,
-            SavePreferences.DatabaseSaveType saveType) throws SaveException {
+                                 SavePreferences.DatabaseSaveType saveType)
+            throws SaveException {
         SaveSession session;
         frame.block();
         final String SAVE_DATABASE = Localization.lang("Save library");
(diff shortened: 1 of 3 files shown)


Siedlerchr:
I tweaked it a bit and I do no longer get any errors/freezes/weird behaviour in the maintable.
However: We have somehow to adjust the scrollbar height, that currently makes the code editor jumps around

src/main/java/org/jabref/gui/BasePanel.java
```diff
@@ -585,7 +585,8 @@ public void update() {
                 .openConsole(frame.getCurrentBasePanel().getBibDatabaseContext().getDatabaseFile().orElse(null)));
 
         actions.put(Actions.PULL_CHANGES_FROM_SHARED_DATABASE, (BaseAction) () -> {
-            DatabaseSynchronizer dbmsSynchronizer = frame.getCurrentBasePanel().getBibDatabaseContext()
+            DatabaseSynchronizer dbmsSynchronizer = frame.getCurrentBasePanel()
+                    .getBibDatabaseContext()
```
tobiasdiez: I know, I'm getting on your nerves with formatting issues, but the dots should be aligned here, shouldn't they?

Siedlerchr: That's not that easily possible as I tried to explain in the other posts. There is no simple align on dot option in Eclipse. And aligning on column will sometimes bee to agressive.

Siedlerchr:
@tobiasdiez  Could you figure out why the entry editor now jumps around in size?

Siedlerchr:
The problem is that when I have a huge entry in my table and another smaller entry (e.g. two required felds set) , and I open the smaller entry,  the entry Editor uses nearly the whole screen:
although the Size of the bibtex source is hard set to 300.

image

image


Entry to test with:
```
@Article{Trapnell_2009,
  author          = {bbbbbbbbbbb},
  title           = {nnnnnnnnnnnn},
  journal         = {Deutsches Arzteblatt international},
  year            = {2009},
  date            = {2016},
  volume          = {25},
  number          = {9},
  issue           = {15},
  month           = {mar},
  pages           = {1105--1111},
  issn            = {1866-0452},
  doi             = {10.1093/bioinformatics/btp120},
  url             = {http://www.google.de},
  citation-subset = {IM},
  country         = {Germany},
  created         = {2016-05-06},
  file            = {:papers/TopHat_discoveringsplicejunctionswithRNA-Seq - TopHat_ Discovering Splice Junctions with RNA Seq.pdf:PDF;:demo/Witten2011  - Data Mining.pdf:PDF;:X\:/Users/CS/Documents/customSeetings.xml:xml;:MsBibExportFormatTest5.bib:bib;:/home/cik-dev/Documents/JabRefTmp/Trapnell_2009.pdf:PDF},
  issn-linking    = {1866-0452},
  keywords        = {rank4},
  nlm             = {PMC4860871},
  nlm-id          = {101475967},
  pii             = {arztebl.2016.0261},
  pmc             = {PMC4860871},
  pmid            = {27151464},
  publisher       = {Oxford University Press ({OUP})},
  pubmodel        = {Print},
  pubstatus       = {ppublish},
  revised         = {2016-05-23},
  timestamp       = {2017-07-15},
}
```

Siedlerchr:
Okay, I got it working now. The problem was that the setting of the entry editor preferences was executed in the FX Thread. I now extracted it and tada it works

Siedlerchr:
I used the version during my productive work and did not find any issues. Therefore I would propose to merge this in.  When merging into maintable-beta we have to revert that.

PR #2010, opened 2016-09-19
Rewrite localization script

bartsch-dev:
Relevant Issues: #1939

I rewrote the the localization script:

The Script has following commands available
- `$ python scripts/syncLang.py status [--extended]`
  - prints the current status to the terminal
  - `[--extended]` if the translations keys which create problems should be printed
  - usable with Gradle tasks
    - `$ gradle status`
    - `$ gradle statusExtended`

- `$ python scripts/syncLang.py markdown`
  - Creates a markdown file of the current status and opens it
  - usable with Gradle tasks
    - `$ gradle statusWithMarkdown`

- `$ python scripts/syncLang.py update [--extended]`
  - compares all the localization files against the English one and fixes unambiguous duplicates, removes obsolete keys, adds missing keys, and sorts them
  - `[--extended]` if the translations keys which create problems should be printed
  - usable with Gradle tasks
    - `$ gradle update`
    - `$ gradle updateExtended`
 
### Localization files status (2016-09-19 20:12 - Branch `rewriteLocalizationScript` `eef4325`)

| Property file | Keys | Keys translated | Keys not translated | % translated |
| ------------- | ---- | --------------- | ------------------- | ------------ |
| JabRef_en.properties | 1572 | 1572 | 0 | 100 |
| JabRef_da.properties | 1572 | 970 | 602 | 61 |
| JabRef_de.properties | 1572 | 1568 | 4 | 99 |
| JabRef_es.properties | 1572 | 1391 | 181 | 88 |
| JabRef_fa.properties | 1572 | 1 | 1571 | 0 |
| JabRef_fr.properties | 1572 | 1570 | 2 | 99 |
| JabRef_in.properties | 1572 | 1232 | 340 | 78 |
| JabRef_it.properties | 1572 | 1162 | 410 | 73 |
| JabRef_ja.properties | 1573 | 1571 | 2 | 99 |
| JabRef_nl.properties | 1572 | 646 | 926 | 41 |
| JabRef_no.properties | 1572 | 967 | 605 | 61 |
| JabRef_pt_BR.properties | 1572 | 1208 | 364 | 76 |
| JabRef_ru.properties | 1572 | 1161 | 411 | 73 |
| JabRef_sv.properties | 1572 | 1413 | 159 | 89 |
| JabRef_tr.properties | 1572 | 1572 | 0 | 100 |
| JabRef_vi.properties | 1572 | 837 | 735 | 53 |
| JabRef_zh.properties | 1572 | 1225 | 347 | 77 |

[...]

PR #390, opened 2015-11-25
Localization cleanup for 3.0

matthiasgeiger:
- [x] Adds missing keys to JabRef_en.properties
- [x] Check whether alternative, already existing keys can be reused
- [x] remove obsolete keys
- [x] Cleanup discrepancies listed by mlep
- [x] Add missing keys for all other languages
- [x] Integrate search localizations
- [ ] Resort? Non-Empty Alphebetically followed by Emphy Alphabetically?

PR #3873, opened 2018-03-22
Sorted language keys alphabetically

stefan-kolb:
As we have little context grouping i was thinking about sorting the keys alphabetically, so

- Words that should be translated consistently are close to each other
- translations are easier to find

Not a big deal but im putting this up for discussion.

### DIFF
--- a/src/main/resources/l10n/JabRef_en.properties
+++ b/src/main/resources/l10n/JabRef_en.properties
@@ -1,2354 +1,1649 @@
 %0\ contains\ the\ regular\ expression\ <b>%1</b>=%0 contains the regular expression <b>%1</b>
-
 %0\ contains\ the\ term\ <b>%1</b>=%0 contains the term <b>%1</b>
-
 %0\ doesn't\ contain\ the\ regular\ expression\ <b>%1</b>=%0 doesn't contain the regular expression <b>%1</b>
-
 %0\ doesn't\ contain\ the\ term\ <b>%1</b>=%0 doesn't contain the term <b>%1</b>
-
+%0\ driver\ not\ available.=%0 driver not available.
+%0\ entries\ found.\ To\ reduce\ server\ load,\ only\ %1\ will\ be\ downloaded.=%0 entries found. To reduce server load, only %1 will be downloaded.
+%0\ entries\ needed\ a\ clean\ up=%0 entries needed a clean up
 %0\ export\ successful=%0 export successful
-
+%0\ file=%0 file
+%0\ files\ found=%0 files found
+%0\ image=%0 image
+%0\ import\ canceled=%0 import canceled
 %0\ matches\ the\ regular\ expression\ <b>%1</b>=%0 matches the regular expression <b>%1</b>
-
 %0\ matches\ the\ term\ <b>%1</b>=%0 matches the term <b>%1</b>
-
+%0\ mode=%0 mode
+%0\ of\ %1=%0 of %1
+%0\ problem(s)\ found=%0 problem(s) found
+%0\ references\ found.\ Number\ of\ references\ to\ fetch?=%0 references found. Number of references to fetch?
+%0\ source=%0 source
+'%0'\ exists.\ Overwrite\ file?='%0' exists. Overwrite file?
+<b>All\ Entries</b>\ (this\ group\ cannot\ be\ edited\ or\ removed)=<b>All Entries</b> (this group cannot be edited or removed)
 <field\ name>=<field name>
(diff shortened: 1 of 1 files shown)


koppor:
Currently, there are logical groups inside the language file. Not everywhere, but sometimes. 

Because of "Words that should be translated consistently are close to each other" and because Words can also stay in the middle of translations, I vote for keeping the order in the language file.

The decision to have the ordering of the English file was taken in the context of https://github.com/JabRef/jabref/pull/1933.

Alphabetical ordering can be achieved by

- Using IntelliJ's alphabetical sorting functionality:
  grafik

- CrowdIn
  grafik

I agree that at first, alphabetically sorted files look "finish" whereas our current state seems to be confused. However, offering the possibility of the English file to be grouped, offers us to keep connected terms togeher.

See for instance "XMP metadata". It should be without hyphen in English (reviewers fault). Moreover, it is also translated differently in German. If these entries were kept together, the fix would be more easy!

    Create\ entry\ based\ on\ XMP-metadata=Eintrag aus XMP-Daten erstellen
    Do\ not\ write\ the\ following\ fields\ to\ XMP\ Metadata\:=Folgende Felder nicht in die XMP-Metadaten schreiben\:

Note in this example the term "XMP-metadata" appears NOT at the beginning, so alphabetical sorting destroys the connection between the translations.

I agree that these groups have to be created. Our tooling is ready to support us with the grouping, since our localization script sorts the non-English keys in the order of the English ones.

Other localization relevant PRs:

- https://github.com/JabRef/jabref/pull/2010
- https://github.com/JabRef/jabref/pull/2031

tobiasdiez:
+1 for alphabetic order. For me personally, the ordering is only relevant in one case: when one removes a localized string from the code and the corresponding language key has to be removed from the English file. This is easier with an alphabetic order.

Logical groups are a nice idea in principle but a) they are not really present right now since all new translations are almost always added at the end of the file b) it is more work to maintain the groups c) they only have diminishing returns in my opinion.

koppor:
@jabref/translators, what do you think? You are the ones wanting
high-quality translations. Do you care for grouping to be able to check for
consistency easily?

Am 22.03.2018 12:05 schrieb "Tobias Diez" <notifications@github.com>:

koppor:
@tobiasdiez My XMP example shows that grouping would have helped to a) spell "XMP metadata" always as such in English keys and b) ensure consistent translation. I know that currently no one volunteers to reorder the file, but enforcing alphabetical ordering removes that option completely.

koppor:
Having a grouped file also offers the possibility to have the Menu entries first (they are allowed to contain `&` menu shortcuts= and the other language keys else. How can we otherwise check for invalid characters in the language keys?

stefan-kolb:
If we would group it, we have to enforce that this grouping is kept consistent, which means:
- Every developer needs to find the right group for the translations (Not easy)
- Every Reviewer needs to be aware and check this for the language files

My reasoning is - all of your arguments are valid and nice - but simply won't happen. Therefore, I think sorting the stuff from time to time would be neat. 

But I have no strong opinion here as it doesn't matter that much.

Siedlerchr:
Well, for the maintable beta branch, we no longer need the menu keys with & signs. 
And often translations are reused, so the context is not always valid.
Java can't unzip files inside apache web server directory - java.util.zip.ZipError: zip END header not found

mlep:
For me grouping is unimportant: using Crowdin, and carrying out the translation often, untranslated keys are "naturally" grouped.
Grouping may be helpful for a new language. However, it will always remain difficult for some strings to get the suitable translation without the context. A full (approximate) translation will be improved later on by using JabRef.

koppor:
@Siedlerchr How do accelerator keys work then? How can I define the accelerator key for the translated keys? Think of `&File` vs. `&Datei`? Meaning: <kbd>Alt</kbd>+<kbd>F</kbd> opens the file menu in English JabRef and <kbd>Alt</kbd>+<kbd>D</kbd> the same menu in German. How can I (as developer) pre-configure that?

Siedlerchr:
Ask @tobiasdiez

tobiasdiez:
https://stackoverflow.com/questions/24499500/javafx-menu-first-letter-underline-decoration

koppor:
I see, I am the only one. I will keep my sorted translations and merge updates manually back. 

I will try to work together with @Brainsucker92 to be able to have `JabRef_en-grouped.properties` locally being synced with `JabRef_en.properties` (the corresponding `JabRef_*-groupes.properties` will be created accordingly). In that way, we can have both: the alphabetically ordered, where you all guys are working with and the grouped one, where I intend to work with. The `-grouped.properties` is for a **quality day once a year**. New keys are sorted into the groups, keys are reviewed for consistency, and languages we know will be reviewed for consistency, too.

koppor:
Alphabetical sort also discussed at https://github.com/JabRef/jabref/pull/390#issuecomment-160202579

PR #3878, opened 2018-03-22
Improve DOI resolution fetcher

stefan-kolb:
- Increase timeout as DOI resolution often fails from `5000`to `10000`
- Return DOI page if it directly redirects to a PDF (good guess that it is the fulltext then)

### DIFF
--- a/src/main/java/org/jabref/logic/importer/fetcher/DoiResolution.java
+++ b/src/main/java/org/jabref/logic/importer/fetcher/DoiResolution.java
@@ -48,7 +48,7 @@ public class DoiResolution implements FulltextFetcher {
                     connection.followRedirects(true);
                     connection.ignoreHttpErrors(true);
                     // some publishers are quite slow (default is 3s)
-                    connection.timeout(5000);
+                    connection.timeout(10000);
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/jabref/logic/importer/fetcher/DoiResolution.java
```diff
@@ -70,6 +72,11 @@
                         LOGGER.info("Fulltext PDF found @ " + sciLink);
                         pdfLink = links.get(0);
                     }
+                } catch (UnsupportedMimeTypeException type) {
+                    // this might be the PDF already as we follow redirects
+                    if (type.getMimeType().startsWith("application/pdf")) {
+                        return Optional.of(new URL(type.getUrl()));
+                    }
```
tobiasdiez: else log warning / rethrow?

stefan-kolb: Is this information relevant for us? We did not get a URL then, so we just return an empty optional (as before).

Siedlerchr:
Lgtm

PR #3880, opened 2018-03-22
Make it easier to create FXML dialogs

tobiasdiez:
Right now it is pretty difficult to create dialogs in JavaFX using FXML. For once, you have to create a bunch of files (fxml, view, controller, view model) and there are huge problems with passing data between the caller and the controller (e.g., no complex objects, cumbersome injection setup,...).

So I took the opportunity to rewrite parts of the afterburner framework to make the workflow more fluid. With these changes, the view class is merged with the controller and it is very easy to pass data to and from the controller. In particular, the code in https://github.com/JabRef/jabref/pull/3005 should become much simpler.

Moreover, I started to embed the content of a few Swing dialogs (cleanup and preferences) in a JavaFX window. This is kind of proof of concept, works pretty nicely in my opinion and is probably the only way to fix the z-hierarchy issues with the old swing dialogs. 

Git is not able to properly sort-out some of the changes (the merge of the view and controller classes) and shows a pretty huge diff. If git gurus like @koppor know a way to fix this, feel free to hijack this PR. 

----


build.gradle
```diff
@@ -132,6 +132,7 @@ dependencies {
     compile 'org.fxmisc.flowless:flowless:0.6'
     compile 'org.fxmisc.richtext:richtextfx:0.8.2'
     compile 'com.sibvisions.external.jvxfx:dndtabpane:0.1'
+    compile 'javax.inject:javax.inject:1'
```
Siedlerchr: Is that an official java dep?

tobiasdiez: Yes, it's a popular JavaEE dependency, see https://mvnrepository.com/artifact/javax.inject/javax.inject.

tobiasdiez:
By the way, here is a very simple example of a dialog that accepts input and provides a result: https://github.com/JabRef/afterburner.fx/blob/master/demo-app/src/main/java/com/airhacks/afterburner/demo/dialog/SimpleDialog.java

Siedlerchr:
Cool! Overall looks good. I would suggest we release our patched afterburner also on gradle/maven.
Could you outline the steps to create /convert the old dialogs to new ones? I think updating the wiki pages would be appropriate

stefan-kolb:
I'm not sure what's behind this but a general remark.
I'm not so happy with changes that introduce additional abstraction layers.
1. Additional knowledge is needed to use them
2. If the logic behind the library is not 100% tested we get hard to trace errors again
Probably makes you the only one that can ever change the code 😄

koppor:
`@Inject` is the key here? Then, we should definitively take it. 👍  Angular also uses that and we should train people in that.

I cannot do anything regarding the huge diffs. Git does it via heuristics. More reading:h ttps://git.wiki.kernel.org/index.php/GitFaq#Why_does_Git_not_.22track.22_renames.3F

Siedlerchr:
Injection has already been used for the whole time to inject deps

tobiasdiez:
@stefan-kolb You raise good points. In fact, one of the problems we have right now is that the afterburner.fx library has bugs that affect us but is no longer maintained. On the other hand, loading FXML files by hand using built-in classes is tedious and leads to repetitive code that would be needed for every dialog. In the end, the extracted library is very small and essentially only one class [ViewLoader](https://github.com/JabRef/afterburner.fx/blob/master/src/main/java/com/airhacks/afterburner/views/ViewLoader.java). Against my natural tendency, I tried to write as much documentation as possible so that it should be easy to understand and maintain it.

Since the PR only got positive feedback, I'll merge it now. The documentation in the wiki is adopted. The `AboutDialog` is easy enough to see which steps are required to setup a dialog (and have a look at the cleanup dialog or wiki page for dialog with return values).

PR #3881, opened 2018-03-22
Concurrent fetchers

stefan-kolb:
Trying to improve the speed of the fulltext fetcher:

- First the authoritative publisher is queried for the PDF (if user has access)
- Afterwards we query all remaining sources and take the first result

@JabRef/developers  WDYT?

### DIFF
--- a/src/main/java/org/jabref/gui/externalfiles/DownloadExternalFile.java
+++ b/src/main/java/org/jabref/gui/externalfiles/DownloadExternalFile.java
@@ -155,16 +155,9 @@ public class DownloadExternalFile {
 
         URLDownload udl = new URLDownload(url);
 
-        try {
-            // TODO: what if this takes long time?
-            // TODO: stop editor dialog if this results in an error:
-            mimeType = udl.getMimeType(); // Read MIME type
-        } catch (IOException ex) {
-            JOptionPane.showMessageDialog(frame, Localization.lang("Invalid URL") + ": " + ex.getMessage(),
-                    Localization.lang("Download file"), JOptionPane.ERROR_MESSAGE);
-            LOGGER.info("Error while downloading " + "'" + res + "'", ex);
-            return;
-        }
+        // TODO: what if this takes long time?
+        // TODO: stop editor dialog if this results in an error:
+        mimeType = udl.getMimeType(); // Read MIME type
         final URL urlF = url;
         final URLDownload udlF = udl;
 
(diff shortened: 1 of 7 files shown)


stefan-kolb:
@tobiasdiez If that really matters, we could invokeAll and assign priorities to the fetchers or something like that. I'm not sure if the preprint (in reality) really differs (or how often) from the published paper.

tobiasdiez:
Priorities or clustering in Authority, Journals and preprints would be a good solution in my opinion. 

I know a few instances where authors didn't update their arxiv preprint with the revised and published version. Since even the slightest changes could shift the equation or theorem numbering, having the published PDF is in general desirable.

stefan-kolb:
I thought about this for a moment. 
The problem that still persists is that we
- have a time complexity of max(v_1,..., v_n) then instead of min(v_1,..., v_n)
- decision complexity goes up (clustering, invokingall, checking which highest rated authority has a url, downloading this URL (hopefully it succeeds then if not probably the download is broken all the time as it will always be the highest priority....)

What really annoys me is that the download takes so much time now.
Your priority might be to get the _right_ document.

Note sure which way to go here.

stefan-kolb:
At the moment we try the original publisher for 10 seconds via the DOIResolution.
Afterwards there might be _better_ alternatives like IEEE than GoogleScholar, but it will not be the original publisher site then anyhow!

Siedlerchr:
Maybe we can offer a switch? E.g. Prefer Official papers over Preprints?

stefan-kolb:
I implemented a possible solution in #3882 
Not 100% sure if it is correct but it could be a step into the right direction.
Fetching should be a little faster now as all fetchers are queried in parallel.
I'm still not sure if I like it that way.

tobiasdiez:
Ok, these are good points. What do think about combining both approaches: we cluster the fetcher by trust level and run all fetchers in a cluster in parallel. Thus the performance is still `min` in each cluster.

Something like:
```java
for (TrustLevel trustLevel : TrustLevel.values()) {
    var callers = fetchers.stream()
            .filter(fetcher::getTrustLevel == trustLevel)
            .map(fetcher -> () -> fetcher.fetch(entry))
            .collect(list());

     try {
         return executer.invokeAny(callers);
     } catch( ExecutionException) {
          // No fetcher successful, continue in next trust level bracket
     } 
}
```
(`Fetcher.fetch` should throw an exception if no url could be found, otherwise the above code does not work).

stefan-kolb:
It's probably easier an more clear to just run all fetchers as it is now and then select the best authority. Don't see too much benefits running them after another except for multiple code loops then.

stefan-kolb:
Closed in favor of #3882

PR #3882, opened 2018-03-23
Priority fetcher

stefan-kolb:
- all fetchers are run in parallel now
- the result with the highest priority wins
- InterruptedException | ExecutionException | CancellationException ignore is more critical; not sure if the code behaves as we want it everytime

Trust Levels:
- SOURCE (highest)
- PUBLISHER
- PREPRINT
- META_SEARCH
- UNKNOWN

Current trust levels:
- DOI: SOURCE
- ScienceDircect: Publisher
- Springer: Publisher
- ACS: Publisher
- IEEE: Publisher
- Google Scholar: META_SEARCH
- Arxiv: PREPRINT
- OpenAccessDOI: META_SEARCH


src/main/java/org/jabref/gui/externalfiles/DownloadExternalFile.java
```diff
@@ -139,63 +139,66 @@ public void download(final DownloadCallback callback) throws IOException {
...
-        String mimeType;
+    public void download(URL url, String mimeType, final DownloadCallback callback) throws IOException {
+        Optional<ExternalFileType> fileType = getExternalFileType(mimeType);
 
         // First of all, start the download itself in the background to a temporary file:
-        final File tmp = File.createTempFile("jabref_download", "tmp");
-        tmp.deleteOnExit();
-
-        URLDownload udl = new URLDownload(url);
+        final File tempFile = File.createTempFile("jabref_download", "tmp");
```
Siedlerchr: Use Files.createTempFile

src/main/java/org/jabref/logic/importer/FulltextFetchers.java
```diff
@@ -20,6 +30,8 @@
 public class FulltextFetchers {
     private static final Logger LOGGER = LoggerFactory.getLogger(FulltextFetchers.class);
 
+    private static final int FETCHER_TIMEOUT = 10;
```
Siedlerchr: Maybe add the unit of time we wait. Eg. 10 seconds/minutes or hours?

src/main/java/org/jabref/logic/importer/fetcher/GoogleScholar.java
```diff
@@ -64,36 +66,51 @@ public GoogleScholar(ImportFormatPreferences importFormatPreferences) {
...
+        Document doc = Jsoup.connect(url).userAgent(URLDownload.USER_AGENT).get();
+        // Check results for PDF link
+        // TODO: link always on first result or none?
+        for (int i = 0; i < NUM_RESULTS; i++) {
+            Elements link = doc.select(String.format("div[data-rp=%S] div.gs_or_ggsm a", i));
+
+            if (link.first() != null) {
+                String target = link.first().attr("href");
+                // link present?
+                if (!"".equals(target) && new URLDownload(target).isPdf()) {
```
Siedlerchr: isEmpty?

src/main/java/org/jabref/logic/importer/fetcher/GoogleScholar.java
```diff
@@ -64,36 +66,51 @@ public GoogleScholar(ImportFormatPreferences importFormatPreferences) {
...
+                    // TODO: check title inside pdf + length?
+                    // TODO: report error function needed?! query -> result
+                    LOGGER.info("Fulltext PDF found @ Google: " + target);
```
lenhard: Is this log statement really needed? If you do a lot of downloads it kind of clutters up the log imho...

stefan-kolb:
@lenhard The SORT messages should not be there anymore. Do you have an old state of the PR?! Thanks for your review, will work that in.

src/main/java/org/jabref/logic/importer/FulltextFetchers.java
```diff
@@ -36,26 +48,93 @@ public FulltextFetchers(ImportFormatPreferences importFormatPreferences) {
...
-                        .ifPresent(e -> clonedEntry.setField(FieldName.DOI, e.getDOI()));
-            } catch (FetcherException e) {
-                LOGGER.debug("Failed to find DOI", e);
+                result = executor.invokeAll(getCallables(clonedEntry, finders), FETCHER_TIMEOUT, TimeUnit.SECONDS);
+            } catch (InterruptedException ignore) {
+
             }
+
+            return result.stream()
+                    .map(FulltextFetchers::waitForResult)
```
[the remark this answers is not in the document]
stefan-kolb: That's right, good point 👍

lenhard:
@stefan-kolb Yes, sorry, my bad. Pulling from master doesn't update the branches you had checked out before...

Tested again, the message is gone. It's also very nice how you can switch between entries during the download and the pdf is linked to the entry without killing the entry editor :)

src/main/java/org/jabref/logic/importer/FulltextFetchers.java
```diff
@@ -36,26 +48,93 @@ public FulltextFetchers(ImportFormatPreferences importFormatPreferences) {
...
+
+    private List<Callable<Optional<FetcherResult>>> getCallables(BibEntry entry, List<FulltextFetcher> fetchers) {
+        return fetchers.stream()
+                .map(f -> getCallable(entry, f))
+                .collect(Collectors.toList());
+    }
 
-                if (result.isPresent() && new URLDownload(result.get().toString()).isPdf()) {
-                    return result;
+    private void shutdownAndAwaitTermination(ExecutorService pool) {
```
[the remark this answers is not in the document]
stefan-kolb: Why isn't this the usual way? It is directl yfrom the Java docs 😄

src/main/java/org/jabref/logic/importer/FulltextFetchers.java
```diff
@@ -36,26 +48,93 @@ public FulltextFetchers(ImportFormatPreferences importFormatPreferences) {
         Optional<DOI> doi = clonedEntry.getField(FieldName.DOI).flatMap(DOI::parse);
 
         if (!doi.isPresent()) {
+            findDoiForEntry(clonedEntry);
+        }
+
+        ExecutorService executor = Executors.newCachedThreadPool();
```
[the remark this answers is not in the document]
lenhard: The only difference between shutdown and shutdownNow is that in the first case tasks are still allowed to complete and in the second they are cancelled. Tasks that never terminate will never terminate in both cases.

src/main/java/org/jabref/logic/importer/FetcherResult.java
```diff
@@ -0,0 +1,15 @@
+package org.jabref.logic.importer;
+
+import java.net.URL;
+
+import org.jabref.logic.importer.fetcher.TrustLevel;
+
+public final class FetcherResult {
+    public final TrustLevel trust;
```
tobiasdiez: Hide fields and make them accessible via getter?

src/main/java/org/jabref/logic/importer/FulltextFetchers.java
```diff
@@ -36,26 +48,93 @@ public FulltextFetchers(ImportFormatPreferences importFormatPreferences) {
...
+                result = executor.invokeAll(getCallables(clonedEntry, finders), FETCHER_TIMEOUT, TimeUnit.SECONDS);
+            } catch (InterruptedException ignore) {
+
             }
+
+            return result.stream()
+                    .map(FulltextFetchers::waitForResult)
+                    .filter(Optional::isPresent)
+                    .map(Optional::get)
+                    .filter(res -> Objects.nonNull(res.source))
```
tobiasdiez: In my opinion, the fetcher should make sure that the URL returned is always non-null. Otherwise the usage of `Optional` is superfluous.

src/main/java/org/jabref/logic/importer/FulltextFetchers.java
```diff
@@ -36,26 +48,93 @@ public FulltextFetchers(ImportFormatPreferences importFormatPreferences) {
...
+            }
+        }
+    }
+
+    private Callable<Optional<FetcherResult>> getCallable(BibEntry entry, FulltextFetcher fetcher) {
+            return () -> {
+                try {
+                    Optional<URL> result = fetcher.findFullText(entry);
+
+                    if (result.isPresent() && new URLDownload(result.get().toString()).isPdf()) {
```
tobiasdiez: Do we really need to check that the URL returned is a PDF? This is an additional request. I think, we should just trust the fetcher that it found an appropriate document.

src/main/java/org/jabref/logic/importer/FulltextFetchers.java
```diff
@@ -36,26 +48,93 @@ public FulltextFetchers(ImportFormatPreferences importFormatPreferences) {
...
+                        return Optional.of(new FetcherResult(fetcher.getTrustLevel(), result.get()));
```
tobiasdiez: Using `Optional.map` may be a bit easier to understand.

src/main/java/org/jabref/logic/importer/FulltextFetchers.java
tobiasdiez: We have the class `JabRefExecutorService` that manages executors. Reuse it here?

src/test/java/org/jabref/logic/importer/FulltextFetchersTest.java
```diff
@@ -43,4 +45,44 @@ public void rejectNonPdfUrls() throws MalformedURLException {
...
+        FulltextFetchers fetcher = new FulltextFetchers(Arrays.asList(finder));
+
+        assertEquals(Optional.of(pdfUrl), fetcher.findFullTextPDF(entry));
+    }
+
+    @Test
+    public void higherTrustLevelWins() throws MalformedURLException {
+        final URL lowUrl = new URL("http://docs.oasis-open.org/opencsa/sca-bpel/sca-bpel-1.1-spec-cd-01.pdf");
+        final URL highUrl = new URL("http://docs.oasis-open.org/wsbpel/2.0/OS/wsbpel-v2.0-OS.pdf");
+        FulltextFetcher finderHigh = new FulltextFetcher() {
```
tobiasdiez: I find mocking using mockito is slightly better / easier to read / less code.

src/test/java/org/jabref/logic/importer/WebFetchersTest.java
```diff
@@ -62,7 +71,9 @@ public void getSearchBasedFetchersReturnsAllFetcherDerivingFromSearchBasedFetche
     public void getFullTextFetchersReturnsAllFetcherDerivingFromFullTextFetcher() throws Exception {
         List<FulltextFetcher> fullTextFetchers = WebFetchers.getFullTextFetchers(importFormatPreferences);
 
-        Set<Class<? extends FulltextFetcher>> expected = reflections.getSubTypesOf(FulltextFetcher.class);
+        Set<Class<? extends FulltextFetcher>> expected = Stream.of(
```
tobiasdiez: The test was written to remind when somebody implements a new fetcher but forgets to add it in `getFullTextFetchers`. This purpose is defeated if the list of fetchers is specified explicitly here. What was the reason for the change?

src/main/java/org/jabref/logic/importer/FetcherResult.java
stefan-kolb: Ok, will do so.

src/main/java/org/jabref/logic/importer/FulltextFetchers.java
stefan-kolb: This is the single point where this is checked. Otherwise every fetcher has to do this logic. It is vital so we don't get wrong fulltexts.

src/test/java/org/jabref/logic/importer/WebFetchersTest.java
stefan-kolb: The problem was that the reflections find the classes from the comment before as sub classes then. Maybe mocking them does change this behavior!

src/main/java/org/jabref/logic/importer/FulltextFetchers.java
stefan-kolb: What kind of code should replace this?

src/main/java/org/jabref/logic/importer/FulltextFetchers.java
```diff
@@ -36,26 +50,82 @@ public FulltextFetchers(ImportFormatPreferences importFormatPreferences) {
...
+            };
+    }
+
+    private List<Callable<Optional<FetcherResult>>> getCallables(BibEntry entry, List<FulltextFetcher> fetchers) {
+        return fetchers.stream()
+                .map(f -> getCallable(entry, f))
+                .collect(Collectors.toList());
+    }
+
+    private void shutdownAndAwaitTermination(ExecutorService pool) {
```
lenhard: Please remove this now unused private method.

src/main/java/org/jabref/logic/importer/FulltextFetchers.java
tobiasdiez: Yes, I was wondering why this check is necessary at all. If the fetcher determines that a given URL contains the right full-text (say because the API tell it so), then why do we need to check that it is a PDF?

src/main/java/org/jabref/logic/importer/FulltextFetchers.java
tobiasdiez: I was thinking about something like
````java
return fetcher.findFullText(entry)
          .filter(isPdf())
          .map(url -> new FetcherResult(fetcher.getTrust(), url));
````

src/main/java/org/jabref/logic/importer/FulltextFetchers.java
stefan-kolb: Because sometimes it is an HTML site or the access is blocked for the users, resulting in something else than the fulltext 😄

stefan-kolb:
Ok guys, I think all of your suggestions are there now 😄

PR #3883, opened 2018-03-23
Reenable closing of entry preview by pressing Esc

tobiasdiez:
Problem was that we had a few key-bindings that were bound to `Esc`, but only the first binding was returned (e.g. user pressed `Esc`, was recognized as `DialogClose` but preview only reacted on `EntryPreviewClose`). 

----

### DIFF
--- a/src/main/java/org/jabref/gui/search/GlobalSearchBar.java
+++ b/src/main/java/org/jabref/gui/search/GlobalSearchBar.java
@@ -143,7 +143,7 @@ public class GlobalSearchBar extends HBox {
                     updateOpenCurrentResultsTooltip(globalSearch.isSelected());
                     focus();
                     event.consume();
-                } else if (keyBinding.get().equals(KeyBinding.CLEAR_SEARCH)) {
+                } else if (keyBinding.get().equals(KeyBinding.CLOSE)) {
                     // Clear search and select first entry, if available
                     clearSearch();
                     frame.getCurrentBasePanel().getMainTable().getSelectionModel().selectFirst();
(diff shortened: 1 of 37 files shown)


Siedlerchr:
Still leaves the problem with clear search and close both using esc

tobiasdiez:
@Siedlerchr This is also fixed with https://github.com/JabRef/jabref/pull/3883/files#diff-62258e748b1a855079c917bfbc5009d2R146

Siedlerchr:
But I think it will now have the effect that if you clear the search the entry editor will be closed  as well.

tobiasdiez:
No, it does not (just tried it out to make sure). `event.consume()` prevents the event from from bubbling up the tree and thus the control with focus gets closed and the other ones stay open.

src/main/java/org/jabref/gui/search/GlobalSearchBar.java
```diff
@@ -143,7 +143,7 @@ public void actionPerformed(ActionEvent e) {
                     updateOpenCurrentResultsTooltip(globalSearch.isSelected());
                     focus();
                     event.consume();
-                } else if (keyBinding.get().equals(KeyBinding.CLEAR_SEARCH)) {
+                } else if (keyBinding.get().equals(KeyBinding.CLOSE)) {
```
Siedlerchr: I would prefer a switch case here like in the other dialogs to have it consistent

Siedlerchr:
If you fix the checkstyle issuie it's a go for me

src/main/java/org/jabref/gui/search/GlobalSearchBar.java
tobiasdiez: Me too, but I'm too lazy right now to change it 😄

PR #3893, opened 2018-03-28
Executor service improvements

stefan-kolb:


src/main/java/org/jabref/JabRefExecutorService.java
```diff
@@ -135,18 +134,36 @@ public void submit(TimerTask timerTask, long millisecondsDelay) {
...
 
-    private class NamedRunnable implements Runnable {
+    private void gracefulShutdown(ExecutorService pool) {
+        // Disable new tasks from being submitted
+        pool.shutdown();
+        try {
+            // Wait a while for existing tasks to terminate
+            if (!pool.awaitTermination(60, TimeUnit.SECONDS)) {
+                // Cancel currently executing tasks
+                pool.shutdownNow();
```
lenhard: I don't think we should cancel regular tasks during the shutdown but let them complete. There may be cases where we definitely want to wait until a task has completed no matter how long it takes. Consider the save operation. Maybe a save of a database of 20 000 entries on a network drive takes very long. If we cancel that, JabRef would stop during the save and corrupt the data.

I think we should not use `shutdownNow` here (and below), but we could print some log method to show that JabRef is trying to shutdown, but some tasks are not complete yet.

stefan-kolb:
Ok so we keep it the way it is. I merged it into the other branch but reverted the commits!

Issue #3861, opened 2018-03-17
Convert/rework rest of the Swing Dialogs to javafx 


Siedlerchr:
This issue serves as point to see which dialogs need to be migrated and which need to be converted:
@JabRef/developers  please have a look at that list and see if I maybe missed some
Some of the dialogs are very easy to convert and could be nice beginner tasks 

- [x] Global Search Results (includes converting SearchWorker to JavaFX Tasks) 
- [x] SidePanel:  OpenOffice #4341 
- [x] SidePanel: WebSearch
- [x] Import->Append  Library Dialog
- [x] Import Inspection (assigned to @tobiasdiez #4692)
- [x] Connect to shared DB #4040 
- [x] Edit-> Manage Keywords
- [x] Edit-> Replace String #4227 (Thx to  @marisuki )
- [x] Edit-> Set/Clear/Append Rename Fields
- [x] Groups -> Add new Group #2630 
- [x] Groups -> Edit Group
- [x] Library-> New Entry Dialog #4312 
- [x] Library-> Edit Strings  #3735 
- [x] Library-> Library  properties #4617
- [x] Quality -> Find Duplicates (broken Threading due to PreviewPanel) #4601 
- [x] Quality -> Duplicate bibtex keys (Removed as part of #4696 )
- [x] Quality -> Merge Entries #4410 
- [x] Quality -> Cleanup Entries
- [x] Quality -> Find unlinked files
- [x] Tools ->New sublibrary from Aux file
- [x] Tools -> Writing XMP Data to Entry (Just a simple status? dialog)
- [x] Options -> Setup general fields #4346 (Thx to @NorwayMaple)
- [x] Options -> Manage custom imports
- [x] Options -> Manage custom exports
- [x] Options -> Manage external filetype 
- [x] Options -> Manage external filetype -> Edit file types  #4885
- [x] Options -> Manage protected Terms (add file is broken due to Threading)
- [x] Options -> Manage content Selectors  #4658 (Thx to @ricantech )
- [x] Options -> Customize entry types
- [x] Options -> Bibtex Key Patterns 
- [x]  *Push Entries to Application. Related to Options -> Preferences -> External Programs*
- [x] Review Changes dialog  #4693 
- [x] Import from plaintext
- [x] Merge shared entry dialog
- [x] Wait for save dialog #4697 

Still need to be converted to MVVM (not a beginner task):
- [ ] Options -> Preferences #4253

LinusDietz:
[...]

PR #3895, opened 2018-03-28
Add XMP Exporter

johannes-manner:
Refs [koppor#319](https://github.com/koppor/jabref/issues/319).

Adds the first functionality to export the bib entries in a single .xmp file. 
@koppor For a first review: Please check the output format of the new .xmp file.
There is also a <?xpacket begin="﻿" id="W5M0MpCehiHzreSzNTczkc9d"?> tag included. Should I remove this tag?

The cli support is planned for the next days :)
----

- [x] Check documentation status (Issue created for outdated help page at [help.jabref.org](https://github.com/JabRef/help.jabref.org/issues)?) Updated the help page within the already opened pull request.

### DIFF
--- a/src/main/java/org/jabref/logic/exporter/XmpExporter.java
+++ b/src/main/java/org/jabref/logic/exporter/XmpExporter.java
@@ -8,8 +8,8 @@ import java.nio.file.Path;
 import java.util.List;
 import java.util.Objects;
 
-import org.jabref.Globals;
 import org.jabref.logic.util.FileType;
+import org.jabref.logic.xmp.XmpPreferences;
 import org.jabref.logic.xmp.XmpUtilWriter;
 import org.jabref.model.database.BibDatabaseContext;
 import org.jabref.model.entry.BibEntry;
@@ -21,8 +21,11 @@ import org.jabref.model.entry.BibEntry;
  */
 public class XmpExporter extends Exporter {
 
-    public XmpExporter() {
-        super("XmpBib", FileType.ENDNOTE_XMP.getDescription(), FileType.ENDNOTE_XMP);
+    private final XmpPreferences xmpPreferences;
+
+    public XmpExporter(XmpPreferences xmpPreferences) {
+        super("xmp", FileType.PLAIN_XMP.getDescription(), FileType.PLAIN_XMP);
+        this.xmpPreferences = xmpPreferences;
     }
 
     @Override
@@ -35,7 +38,7 @@ public class XmpExporter extends Exporter {
         }
 
         try (BufferedWriter writer = Files.newBufferedWriter(file, StandardCharsets.UTF_8)) {
-            writer.write(XmpUtilWriter.generateXmpString(entries, Globals.prefs.getXMPPreferences()));
+            writer.write(XmpUtilWriter.generateXmpString(entries, this.xmpPreferences));
             writer.flush();
         }
     }
--- a/src/main/java/org/jabref/logic/exporter/XmpExporter.java
+++ b/src/main/java/org/jabref/logic/exporter/XmpExporter.java
@@ -1,10 +1,13 @@
 package org.jabref.logic.exporter;
 
 import java.io.BufferedWriter;
+import java.io.IOException;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
 
@@ -21,6 +24,8 @@ import org.jabref.model.entry.BibEntry;
  */
 public class XmpExporter extends Exporter {
 
+    private static final String XMP_SPLIT_PATTERN = "split";
+
     private final XmpPreferences xmpPreferences;
 
     public XmpExporter(XmpPreferences xmpPreferences) {
@@ -31,16 +36,31 @@ public class XmpExporter extends Exporter {
     @Override
     public void export(BibDatabaseContext databaseContext, Path file, Charset encoding, List<BibEntry> entries) throws Exception {
         Objects.requireNonNull(databaseContext);
+        Objects.requireNonNull(file);
         Objects.requireNonNull(entries);
 
         if (entries.isEmpty()) {
             return;
         }
 
+        // This is a distinction between writing all entries from the supplied list to a single .xmp file,
+        // or write every entry to a separate file.
+        if (file.getFileName().toString().trim().equals(XMP_SPLIT_PATTERN)) {
+            for (BibEntry entry : entries) {
+                // Avoid situations, where two cite keys are null
+                Path entryFile = Paths.get(entry.getId() + "_" + entry.getCiteKey() + ".xmp");
+
+                this.writeBibToXmp(entryFile, Arrays.asList(entry));
+            }
+        } else {
+            this.writeBibToXmp(file, entries);
+        }
+    }
+
+    private void writeBibToXmp(Path file, List<BibEntry> entries) throws IOException {
         try (BufferedWriter writer = Files.newBufferedWriter(file, StandardCharsets.UTF_8)) {
             writer.write(XmpUtilWriter.generateXmpString(entries, this.xmpPreferences));
             writer.flush();
         }
     }
-
 }
--- a/src/main/java/org/jabref/logic/exporter/XmpExporter.java
+++ b/src/main/java/org/jabref/logic/exporter/XmpExporter.java
@@ -3,7 +3,6 @@ package org.jabref.logic.exporter;
 import java.io.BufferedWriter;
 import java.io.IOException;
 import java.nio.charset.Charset;
-import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -46,19 +45,26 @@ public class XmpExporter extends Exporter {
         // This is a distinction between writing all entries from the supplied list to a single .xmp file,
         // or write every entry to a separate file.
         if (file.getFileName().toString().trim().equals(XMP_SPLIT_PATTERN)) {
+
             for (BibEntry entry : entries) {
                 // Avoid situations, where two cite keys are null
-                Path entryFile = Paths.get(entry.getId() + "_" + entry.getCiteKey() + ".xmp");
+                Path entryFile;
+                String suffix = entry.getId() + "_" + entry.getCiteKey() + ".xmp";
+                if (file.getParent() == null) {
+                    entryFile = Paths.get(suffix);
+                } else {
+                    entryFile = Paths.get(file.getParent().toString() + "/" + suffix);
+                }
 
-                this.writeBibToXmp(entryFile, Arrays.asList(entry));
+                this.writeBibToXmp(entryFile, Arrays.asList(entry), encoding);
             }
         } else {
-            this.writeBibToXmp(file, entries);
+            this.writeBibToXmp(file, entries, encoding);
         }
     }
 
-    private void writeBibToXmp(Path file, List<BibEntry> entries) throws IOException {
-        try (BufferedWriter writer = Files.newBufferedWriter(file, StandardCharsets.UTF_8)) {
+    private void writeBibToXmp(Path file, List<BibEntry> entries, Charset encoding) throws IOException {
+        try (BufferedWriter writer = Files.newBufferedWriter(file, encoding)) {
             writer.write(XmpUtilWriter.generateXmpString(entries, this.xmpPreferences));
             writer.flush();
         }
--- a/src/main/java/org/jabref/logic/exporter/XmpExporter.java
+++ b/src/main/java/org/jabref/logic/exporter/XmpExporter.java
@@ -1,0 +1,40 @@
+package org.jabref.logic.exporter;
+
+import java.io.BufferedWriter;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+import java.util.Objects;
+
+import org.jabref.logic.util.FileType;
+import org.jabref.logic.xmp.XmpUtilWriter;
+import org.jabref.model.database.BibDatabaseContext;
+import org.jabref.model.entry.BibEntry;
+
+/**
+ *
+ */
+public class XmpExporter extends Exporter {
+
+    public XmpExporter() {
+        super("XmpBib", FileType.ENDNOTE_XMP.getDescription(), FileType.ENDNOTE_XMP);
+    }
+
+    @Override
+    public void export(BibDatabaseContext databaseContext, Path file, Charset encoding, List<BibEntry> entries) throws Exception {
+        Objects.requireNonNull(databaseContext);
+        Objects.requireNonNull(entries);
+
+        if (entries.isEmpty()) {
+            return;
+        }
+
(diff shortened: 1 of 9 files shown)


src/main/java/org/jabref/logic/xmp/XmpUtilWriter.java
```diff
@@ -159,6 +171,26 @@ private static void writeDublinCore(PDDocument document,
         catalog.setMetadata(metadataStream);
     }
 
+    public static String generateXmpString(List<BibEntry> entries) {
+        XMPMetadata meta = XMPMetadata.createXMPMetadata();
+        for (BibEntry entry : entries) {
+            DublinCoreSchema dcSchema = meta.createAndAddDublinCoreSchema();
+            XmpUtilWriter.writeToDCSchema(dcSchema, entry, null, Globals.prefs.getXMPPreferences());
+        }
+        ByteArrayOutputStream os = new ByteArrayOutputStream();
```
Siedlerchr: You should always wrap those Input and Output stream in the try with resources construct, that ensures that the streams are autoclosed:
https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

src/main/java/org/jabref/logic/exporter/XmpExporter.java
```diff
@@ -0,0 +1,40 @@
...
+import java.nio.file.Path;
+import java.util.List;
+import java.util.Objects;
+
+import org.jabref.logic.util.FileType;
+import org.jabref.logic.xmp.XmpUtilWriter;
+import org.jabref.model.database.BibDatabaseContext;
+import org.jabref.model.entry.BibEntry;
+
+/**
```
koppor: No empty comments please

src/main/java/org/jabref/logic/xmp/XmpUtilWriter.java
```diff
@@ -82,6 +85,15 @@ public static void writeXmp(Path file, BibEntry entry,
         XmpUtilWriter.writeXmp(file, bibEntryList, database, xmpPreferences);
     }
 
+    /**
+     * Writes the information of the bib entry to the dublin core schema using
+     * a custom extractor.
+     *
+     * @param dcSchema
```
koppor: Remove non-commented parameters - This should automatically be done in IntelliJ if you use autoformat (ctrl+alt+l)

src/main/java/org/jabref/logic/xmp/XmpUtilWriter.java
```diff
@@ -159,6 +171,26 @@ private static void writeDublinCore(PDDocument document,
...
+            DublinCoreSchema dcSchema = meta.createAndAddDublinCoreSchema();
+            XmpUtilWriter.writeToDCSchema(dcSchema, entry, null, Globals.prefs.getXMPPreferences());
+        }
+        ByteArrayOutputStream os = new ByteArrayOutputStream();
+        XmpSerializer serializer = new XmpSerializer();
+        try {
+            serializer.serialize(meta, os, true);
+            return os.toString(StandardCharsets.UTF_8.name());
+        } catch (TransformerException e) {
+            LOGGER.warn("Tranformation into xmp not possible: " + e.getMessage());
```
koppor: Why not logging the exception itself, too?

src/main/java/org/jabref/logic/xmp/XmpUtilWriter.java
```diff
@@ -159,6 +171,26 @@ private static void writeDublinCore(PDDocument document,
         catalog.setMetadata(metadataStream);
     }
 
+    public static String generateXmpString(List<BibEntry> entries) {
```
koppor: Please comment that it returns an empty string if something goes wrong.

src/main/java/org/jabref/logic/util/FileType.java
```diff
@@ -26,6 +26,7 @@
     DIN_1505(Localization.lang("%0 file", "DIN 1505"), "rtf"),
     ENDNOTE(Localization.lang("%0 file", "EndNote/Refer"), "ref", "enw"),
     ENDNOTE_XML(Localization.lang("%0 file", "EndNote XML"), "xml"),
+    ENDNOTE_XMP(Localization.lang("%0 file", "EndNote Xmp"), "xmp"),
```
koppor: Why "Endnote XMP"? What has "Endnote" todo with XMP? I would just use "XMP", because it is an [ISO Standard](https://en.wikipedia.org/wiki/Extensible_Metadata_Platform), not something from Endnote.

src/main/java/org/jabref/logic/xmp/XmpUtilWriter.java
```diff
@@ -159,6 +171,26 @@ private static void writeDublinCore(PDDocument document,
...
+        XMPMetadata meta = XMPMetadata.createXMPMetadata();
+        for (BibEntry entry : entries) {
+            DublinCoreSchema dcSchema = meta.createAndAddDublinCoreSchema();
+            XmpUtilWriter.writeToDCSchema(dcSchema, entry, null, Globals.prefs.getXMPPreferences());
```
koppor: Globals should never be asked. The XMPPreferences should be passed to the method, not taken from globals!

src/main/java/org/jabref/logic/xmp/XmpUtilWriter.java
johannes-manner: Do you mean the stack trace of the exception?
Is there a smart solution in JabRef to print the stack trace with the logger (maybe an utility method)?

Otherwise, I would use something like this:
Arrays.asList(exception.getStackTrace())
                .stream()
                .map(Objects::toString)
                .collect(Collectors.joining("\n"))

Siedlerchr: The logger methods have  a second parameter which accepts a throwable /exception.
The stack trace will then automatically be printed (handled by the logger itself)

tobiasdiez: The code usually looks something like `LOGGER.warn("Transformation in xmp not possible", e)`.

src/main/java/org/jabref/logic/xmp/XmpUtilWriter.java
```diff
@@ -159,6 +170,38 @@ private static void writeDublinCore(PDDocument document,
...
+     * @param xmpPreferences    The user's xmp preferences.
+     *
+     * @return  If something goes wrong (e.g. an exception is thrown), the method returns an empty string,
+     *          otherwise it returns the xmp metadata as a string in dublin core format.
+     */
+    public static String generateXmpString(List<BibEntry> entries, XmpPreferences xmpPreferences) {
+        XMPMetadata meta = XMPMetadata.createXMPMetadata();
+        for (BibEntry entry : entries) {
+            DublinCoreSchema dcSchema = meta.createAndAddDublinCoreSchema();
+            XmpUtilWriter.writeToDCSchema(dcSchema, entry, null, xmpPreferences);
```
koppor: Is it possible to add an additional method `writeToDCSchema` with 3 parameters. We should avoid `null` parameters.

src/main/java/org/jabref/logic/exporter/XmpExporter.java
```diff
@@ -0,0 +1,43 @@
...
+    public void export(BibDatabaseContext databaseContext, Path file, Charset encoding, List<BibEntry> entries) throws Exception {
+        Objects.requireNonNull(databaseContext);
+        Objects.requireNonNull(entries);
+
+        if (entries.isEmpty()) {
+            return;
+        }
+
+        try (BufferedWriter writer = Files.newBufferedWriter(file, StandardCharsets.UTF_8)) {
+            writer.write(XmpUtilWriter.generateXmpString(entries, Globals.prefs.getXMPPreferences()));
```
tobiasdiez: Can you please pass the `Globals.prefs.getXMPPreferences()` dependency as a constructor argument. We are currently trying to reduce the number of calls to `Globals`.

tobiasdiez: Can you please pass the `Globals.prefs.getXMPPreferences()` dependency as a constructor argument. We are currently trying to reduce the number of calls to `Globals`.

src/main/java/org/jabref/logic/xmp/XmpUtilWriter.java
```diff
@@ -159,6 +170,38 @@ private static void writeDublinCore(PDDocument document,
...
+        }
+        try (ByteArrayOutputStream os = new ByteArrayOutputStream()) {
+            XmpSerializer serializer = new XmpSerializer();
+            serializer.serialize(meta, os, true);
+            return os.toString(StandardCharsets.UTF_8.name());
+        } catch (TransformerException e) {
+            LOGGER.warn("Tranformation into xmp not possible: " + e.getMessage());
```
tobiasdiez: Personally, I would wrap these Exceptions in an IOException and then rethrow.

johannes-manner:
I have a few question to the cli exporter.
If I read the code right, all exporters read a file through an importer and write a bib file.
So the syntax, that worked for me was: 
java -jar JabRef.jar -o export.bib import.pdf,xmp -n

So there is a xmp importer for the import of the metadata in pdf files. 
To import plain xmp and write xmp to bib, there is the need for another importer for plain xmp. Is this requested?

Another question is, if the discussed cli option is possible under the -o option (or if it is only possible under the -xmp tag) to generate a plain xmp file from a bib file? Or to generate multiple xmp files for every single entry in the bib file.
I don't know and see no option to plug-in such features in the current exporter. Maybe some of you have suggestions for me? :)

koppor:
Doesn't `-o filename,xmp` (as documented at http://help.jabref.org/en/CommandLine#export-file--o-filenameexport-format) work? :)

koppor:
Doesn't `-o filename,xmp` (as documented at http://help.jabref.org/en/CommandLine#export-file--o-filenameexport-format) work? :)

johannes-manner:
"If only filename is specified, **it will be exported in BibTeX format**. If the filename is followed by a comma and an export format, the given export filter will be used." (JabRef Help)

The functionality is an exporter, but only to .bib files and importes the exported entries from different sources (in my case reads the xmp metadata from a .pdf file and writes it to the .bib file)

-o export.bib import.pdf, xmp

I think, the requested option is not possible with the current architectuer under the -o option, if I didn't overlooked something. 
The easiest way for me is, to write a custom functionality for -xmp. Other opinions?

johannes-manner:
Is it possible, that the help page is not up-to-date?

johannes-manner:
Now, it worked for:
-o  output.xmp,xmp  test.bib  -n

There were a few strings and options, I configured the wrong way. 
An enrichment of the help page with examples would be beneficial :)

Siedlerchr:
@johannes-manner  Feel free to update the help page accordingly 
https://github.com/JabRef/help.jabref.org/blob/gh-pages/en/CommandLine.md

Siedlerchr:
@johannes-manner 
you have an architecture violation. Either you add an exception (not that good, or you find a way to pass the XMP Prefs from a gui class down to the exporter factory

>  => org.opentest4j.AssertionFailedError: The following classes are not allowed to depend on org.jabref.Globals ==> expected: <[]> but was: <[src/main/java/org/jabref/logic/exporter/ExporterFactory.java]>

src/main/java/org/jabref/logic/exporter/XmpExporter.java
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * A custom exporter to write bib entries to a .xmp file for further processing
+ * in other scenarios and applications. The xmp metadata are written in dublin
+ * core format.
+ */
+public class XmpExporter extends Exporter {
+
+    private final XmpPreferences xmpPreferences;
+
+    public XmpExporter(XmpPreferences xmpPreferences) {
```
koppor: How do the other exporters get their preferences? Shouldn't they have the same issue with the Globals dependency?

Siedlerchr: They get their prefs in the Exporter Factory. In this case it must only be expanded to pass the xmp prefs from the existing pref object

johannes-manner:
Ready for final review.

There is now also an option to generate single .xmp files for every entry in a .bib file. Commented in the help page - command line options :)

koppor:
Shouldn't XMP be written with capital M and P?

grafik

koppor:
- [ ] Why does the order of XML elements change at a second export? This IMHO should not happen.
  grafik
- [ ] There should a be a cleanup made. latex2unicode at each field.
       "Proceedings of the 10\textsuperscript{th} Central European Workshop on Services and their Composition ({ZEUS} 2018" should become "Proceedings of the 10th Central European Workshop on Services and their Composition ({ZEUS} 2018)"
- [ ] The latex cleanup should be executed. - Maybe this is done by the latex2unicode thing:  "{CEUR} Workshop Proceeding" <-- Please no `{}` in the XMP.
- [ ] I am not sure whether `<rdf:li>bibtex/url/https://adr.github.io/madr/</rdf:li>` is the correct RDF searlization for an URL.

These issues are not mandatory. We can merge and fix that in a follow-up if it is too complicated.

Test entry:

```
@InProceedings{KoppAZ-MADR-ZEUS-2018,
  author    = {Kopp, Oliver and Armbruster, Anita and Zimmermann, Olaf},
  title     = {Markdown Architectural Decision Records: Format and Tool Support},
  booktitle = {Proceedings of the 10\textsuperscript{th} Central European Workshop on Services and their Composition ({ZEUS} 2018)},
  year      = {2018},
  volume    = {2072},
  series    = {{CEUR} Workshop Proceedings},
  pages     = {55--62},
  publisher = {CEUR-WS.org},
  keywords  = {ADR; MADR; architecture decision records; architectural decision records; Nygard},
  url       = {https://adr.github.io/madr/},
}
```

Siedlerchr:
@johannes-manner  Class Prefixes Xmp, but I think @koppor  is refering to the extension description

koppor:
Yeah, the class is fine with "Xmp", because of Google Style Guide. The interface to the user should use the offical term. Same with URI, ...

johannes-manner:
I will update the description with the fix for the mentioned issues :)

Siedlerchr:
Just one more question, what exactly does the dialog  Tools -> Writing XMP Data to Entry display?
You could take this as a change to rework that simple one to javafx, see #3861

johannes-manner:
@Siedlerchr 
Is there no general dialog for informing the user about an operation? 
Because the dialog only displays, if the operation suceeded or failed...

You mean Tools->Write XMP metadata to PDFs?

koppor:
I'll merge. The JavaFX thing and my XMP wishes can be done in a separate PR. 😇 . Thank you for the continuous work on that. Reading http://dublincore.org/documents/dc-xml-guidelines/ and https://github.com/CrossRef/pdfmark/blob/master/test-data/random-xmp-example.xmp and https://www.pdflib.com/de/knowledge-base/xmp-metadaten/kostenloser-xmp-validator/, this topic is not easy.

<blockquote class="twitter-tweet" data-lang="de"><p lang="en" dir="ltr">1. Release to customers twice as often.<br>2. Fix the problems.<br>3. Repeat.<br><br>(Maybe release one third as much stuff at a time, just to avoid, you know, going out of business. It&#39;s okay, because much of the second two thirds your customer didn&#39;t want anyway and now you&#39;ll know early.)</p>&mdash; Kent Beck (@KentBeck) <a href="https://twitter.com/KentBeck/status/984053793762373633?ref_src=twsrc%5Etfw">11. April 2018</a></blockquote>

Siedlerchr:
Yes, I wondered what there is displayed when I checked the dialogs

PR #3900, opened 2018-03-29
Monthfiled in patent should be treated same as month (IEEETRAN)

### LINKED ISSUE #3896: monthfiled in patent should be treated same as month

JabRef version 4.1 and master latest on Fedora 26

Steps to reproduce:

1. Add an entry with type IEEEtran patent
2. Go to "Optional fields" tab, and checkout "Monthfiled"
3. Input `#may#`

Expected: recognized as a bibtex month macro. Saved to bib file as `monthfiled = may`.

Actual: Saved to bib file as `monthfiled = {#may#}`, which is an error.

stefan-kolb:
Closes #3896


Siedlerchr:
Interestingly Travis was green, but now it seems like there is some checkstyle

PR #3903, opened 2018-03-30
Fix pureauth

koppor:
Fixes https://github.com/koppor/jabref/issues/74.

`pureauth` was not implemented correctly. This PR fixes it and does some refactoring

- Remove `Optional` as parameter in two cases
- Split tests.


Siedlerchr:
Please also check http://help.jabref.org/en/BibtexKeyPatterns

tobiasdiez:
Merge or squash? I leave the decision in your hand @koppor .

PR #3904, opened 2018-03-30
[WIP] Add Union/Intersection Toogle to Groups Menu

### LINKED ISSUE #3269: 4.0 Groups: fast access to View options

Siedlerchr:
Fixes #3269 

Followup from #3739 

----


Siedlerchr:
Yeah, I already thought about the same. Especially a way to indicate the current status would be nice. 
 So maybe an icon status somewhere. I also like these circle indicators from set theory. e.g.:
https://materialdesignicons.com/icon/set-center

@halirutan  As you are working on the design, any idea where to put such an icon for the groups?

halirutan:
@Siedlerchr I think the header of the groups view is the best place. In IDEA this is similar. The view settings for the project tree is in the top right corner of this panel. So *for me* this would be the first place to look. If there are several of such view options, we probably need to make a button that opens a context menu. If there are only two buttons, they might fit into the existing header.

Siedlerchr:
I will create a new PR based on maintable beta.

Issue #3897, opened 2018-03-29
license of external libraries not exact enough


alick:
Steps to reproduce:

1. Check out https://github.com/JabRef/jabref/blob/master/external-libraries.txt
2. Note that it particularly says "BSD" is not exact enough
3. However, quite a few libraries has "BSD" as the license ID.

lenhard:
I have the impression that this file is not maintained as rigorously as other things. Some libraries (jgoodies, etc.) in there have already been removed from JabRef.

In anything license-related @koppor is our expert.

koppor:
Yes, we shall use the [SPDX License Identifiers](https://spdx.org/licenses/).

I wonder whether you read [line 11 onwars](https://github.com/JabRef/jabref/blob/40a007aabaf96ad963cc9cb72f94a9f7371bfedd/external-libraries.txt#L11) in the file:

> We follow the [SPDX license identifiers](https://spdx.org/licenses/).
> [...]
> For instance, "BSD" is not exact enough, there are numerous variants out there: BSD-2-Clause, BSD-3-Clause-No-Nuclear-Warranty, ...

Feel free to open a pull request fixing the license statements for each listed library.

stefan-kolb:
@lenhard We still have jgoodies stuff in there?!

alick:
OK I'll do a license audit/check when I get the time.

Please remove any libraries that are no longer in use if any.

koppor:
@alick Can you do a cross-check with `build.gralde` when you are on it? We do not know by heart which libraries we included an dropped. The build.gradle is the single source of truth.

PR #3906, opened 2018-03-30
Dependencies

stefan-kolb:
#3897

I sorted some of the dependencies in build.gradle.
Some dependencies are grouped by context, e.g., JavaFX.
I think however that we need to sort all dependencies alphabetically otherwise synching the external-libs.txt with the dependencies is impossible in my opinion. @JabRef/developers WDYT?
We might add some comments for deps that do not reveal their usage via their name.


Siedlerchr:
I thought it would be a nice idea to auto-generate this file and fortunately someone created a  gradle plugin:
https://github.com/jk1/Gradle-License-Report

lenhard:
Alphabetical sorting will more or less result in a context-sorting anyway, won't it? After all, the libraries have a url-style naming pattern and we can just add blank lines here and there to make the context stand out.

stefan-kolb:
In many cases yes. For a lot of JavaFX libraries unfortunately, no.

Siedlerchr:
Why not using that plug in I suggested?
Iz seems to auto generate such a list of dependencies to a file. We then need only invoke it then when deps changes 
https://github.com/jk1/Gradle-License-Report

stefan-kolb:
Have not tried it out yet.

koppor:
I added it. Run following command

    ./gradlew generateLicenseReport

Afterwards, find following files:

- build/reports/dependency-license/index.html
- build/reports/dependency-license/THIRD-PARTY-NOTICES.txt

Excerpt:

```

Dependency License Report for JabRef

Dependency License Report for JabRef 4.2-dev

1. Group: com.airhacks  Name: afterburner.fx  Version: 1.7.0

POM Project URL: http://afterburner.adam-bien.com

POM License: The Apache Software License, Version 2.0 - http://www.apache.org/licenses/LICENSE-2.0.txt

--------------------------------------------------------------------------------

2. Group: com.github.bkromhout  Name: java-diff-utils  Version: 2.1.1

--------------------------------------------------------------------------------

3. Group: com.github.tomtung  Name: latex2unicode_2.12  Version: 0.2.2

POM Project URL: https://github.com/tomtung/latex2unicode

POM License: Apache 2 - http://www.apache.org/licenses/LICENSE-2.0.txt
```

grafik

When it comes to embedded licenses, the text renderer is not that good:

```
--------------------------------------------------------------------------------

15. Group: com.microsoft.azure  Name: applicationinsights-core  Version: 2.0.2

POM License: The Apache Software License, Version 2.0 - http://www.apache.org/licenses/LICENSE-2.0.txt

Embedded license: 

                    ****************************************                    


                                 Apache License
                           Version 2.0, January 2004
                        http://www.apache.org/licenses/

   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
```

The HTML one is better in this case:

grafik

Since, the tool does not use [SPDX License Identifiers](https://spdx.org/licenses/), we need to make a PR to the tool proposing to use the accepted standard for identifying licenses.

We also need to provide an [importer](https://github.com/jk1/Gradle-License-Report#importers) for unknown licenses - or we need to contact these authors and ask for a proper license in the POM.

grafik

Comparison

| Field | external-libraries.txt | Gradle-License-Report |
| -- | -- | -- |
| Id | Y | Y |
| Project | Y | -  |
| Version | Y | Y |
| Project | Y | -  |
| URL| Y | only-if-embedded-in-the-POM  |
| License| Y | y  |
| Note | Y | -  |

That means, we skip the human readable project name (which is OK) and the project-URL (in most cases).

Sometimes, we have different project URLs:

```
Id:      de.undercouch.citeproc-java
Project: Citeproc-Java
URL:     http://michel-kraemer.github.io/citeproc-java/
Licence: Apache-2.0
```

vs.

```
25. Group: de.undercouch  Name: citeproc-java  Version: 1.0.1

POM Project URL: http://www.michel-kraemer.com

POM License: The Apache Software License, Version 2.0 - http://www.apache.org/licenses/LICENSE-2.0.txt
```

So, in summary, I think, it is OK to include the dependency, but we should use it only as cross-check for external-libraries.txt.

koppor:
I updated `external-libraries.txt` and an initial ADR-0003 reflecting this discussion.

koppor:
I grouped `external-libraries.txt` according to the groups of `build.gradle` (and added the group "Citation Style Language"). I removed all testing dependencies from `external-libraries.txt` as testing dependencies are not required to be re-packaged by linux distributions.

I also added following sections to `external-libraries.txt`:

    ## Transitive dependencies

    ## Dependencies included as JARs in the source tree

build.gradle
```diff
@@ -85,91 +86,76 @@ configurations {
...
     compile fileTree(dir: 'lib', includes: ['*.jar'])
 
-    compile 'com.jgoodies:jgoodies-common:1.8.1'
-    compile 'com.jgoodies:jgoodies-forms:1.9.0'
-
-    compile 'org.apache.pdfbox:pdfbox:2.0.9'
-    compile 'org.apache.pdfbox:fontbox:2.0.9'
-    compile 'org.apache.pdfbox:xmpbox:2.0.9'
-
-    // required for reading write-protected PDFs - see https://github.com/JabRef/jabref/pull/942#issuecomment-209252635
```
koppor: Should I convert the comment to an [ADR](https://adr.github.io/madr/) and link it (in a comment `// see ADR-0004`)? I added that comment, because someone might wonder why bouncycastle is integrated.

koppor: Should I convert the comment to an [ADR](https://adr.github.io/madr/) and link it (in a comment `// see ADR-0004`)? I added that comment, because someone might wonder why bouncycastle is integrated.

stefan-kolb:
Hm, so it doesn't look like we can make any use of the plugin right now? So I propose to remove it and sort all dependencies alphabetically. And keep the manual workflow for now.

koppor:
I like the grouping very much, because it eases packaging. Especially in the context of Debian (refs https://github.com/koppor/jabref/issues/135).

It is also helpful to have the plugin included in `build.gradle` so that maintainers working on `external-libraries.txt` don't have to modify `build.gradle` when working on that.

stefan-kolb:
Probably need to base this on the maintable-beta branch if we want to merge it. Otherwise, we end up with a lot of conflicts there.

koppor:
Decision: cherry-pick adr-003 into a new PR. Remainder: Leave as is.

PR #3907, opened 2018-03-31
Integrate CodeStyleTest into StringUtilTest

koppor:
During https://github.com/JabRef/jabref/pull/3704, I observed that having many classes causes difficulties in migrating to a multi module build. Since 1) we have [ArchUnit](https://www.archunit.org/) in place and 2) the current test in "CodeStyleTest" just covers `StringUtils`. Therefore, I propose to include the test included there into `StringUtilTest`.


stefan-kolb:
Can be removed after refactoring Stringutils anyway. Also, as we have a Guava dependency now, maybe we can replace a lot of our helper functions?

koppor:
We checked this approx half a year ago and none of these functions were present in Guava. They are IMHO mostly available in [Apache Commons Lang](https://commons.apache.org/proper/commons-lang/), but our "spirit" wants to remove this well-established library to the more modern Guava library (https://github.com/google/guava/wiki/StringsExplained).

stefan-kolb:
Well then you can move it anyway.

PR #3909, opened 2018-03-31
Add required mock to LinkedfileViewModelTest

koppor:
While working on #3704, test were failing in other settings. This is a quick fix to get the tests running.


Siedlerchr:
I think for this build it's okay, in maintable-beta we already have it refactored

PR #3910, opened 2018-03-31
Remove unused classes

koppor:
This deletes two classes, where IntelliJ said that they are not used any more.


koppor:
OK, I only removed "VerticalLabelUI"

PR #3913, opened 2018-03-31
Prepare multi module build

koppor:
This is a follow-up to https://github.com/JabRef/jabref/pull/3704. This PR separates logic and UI at more places. It cannot easily be based on `maintable-beta` (https://github.com/JabRef/jabref/pull/3621). I propose following procedure:

1. Get this ready to be merged into master
2. Merge master into `maintable-beta` (https://github.com/JabRef/jabref/pull/3621)

For the first step, just the following is open:

- [x] The OO preferences should be fixed (reading/writing not fully supported yet). All other features should work.


src/main/java/org/jabref/logic/exporter/TemplateExporter.java
```diff
@@ -172,7 +171,7 @@ private Reader getReader(String filename) throws IOException {
         String name = dir + filename;
         Reader reader;
         // Try loading as a resource first. This works for files inside the JAR:
-        URL reso = JabRefMain.class.getResource(name);
+        URL reso = TemplateExporter.class.getResource(name);
```
tobiasdiez: Now `name` needs to be relative to `org/jabref/logic/exporter/` and not `org/jabref/` right?

src/main/java/org/jabref/preferences/ExporterFactoryFactory.java
```diff
@@ -0,0 +1,20 @@
...
+
+import org.jabref.logic.exporter.ExporterFactory;
+import org.jabref.logic.exporter.SavePreferences;
+import org.jabref.logic.exporter.TemplateExporter;
+import org.jabref.logic.journals.JournalAbbreviationLoader;
+import org.jabref.logic.layout.LayoutFormatterPreferences;
+
+public class ExporterFactoryFactory {
+
+    public static ExporterFactory create(JabRefPreferences preferences, JournalAbbreviationLoader abbreviationLoader) {
```
tobiasdiez: What about moving this method to `JabRefPreferences` as an instance method `createExportFactory(abbrLoader)`?

src/main/java/org/jabref/preferences/JabRefPreferences.java
```diff
@@ -1421,6 +1425,41 @@ public ImportFormatPreferences getImportFormatPreferences() {
                 isKeywordSyncEnabled());
     }
 
+    public static SavePreferences loadForExportFromPreferences(JabRefPreferences preferences) {
```
tobiasdiez: Convert these two methods to instance methods and rename them to something like `loadSavePreferencesForExport`

src/test/java/org/jabref/model/groups/TexGroupTest.java
```diff
@@ -18,7 +17,7 @@
 
     @Test
     public void containsReturnsTrueForEntryInAux() throws Exception {
-        Path auxFile = Paths.get(AuxParserTest.class.getResource("paper.aux").toURI());
+        Path auxFile = Paths.get(TexGroupTest.class.getResource("paper.aux").toURI());
```
tobiasdiez: Whats the reason for this change? The only "advantage" is that we now have a copy of the `paper.aux` file...

src/main/java/org/jabref/logic/exporter/TemplateExporter.java
koppor: I checked the code and tried out the UI:

grafik

The UI works and the code IMHO never goes that path. With that change, the code is prepared for testing the "TemplateExporter".

src/test/java/org/jabref/model/groups/TexGroupTest.java
koppor: I had the fear that maintainers of the AuxParserTest do not know that the TexGroupTest also relies on the aux file. What if another test is created? I would separate tests of different classes as much as possible even if files (`.bib`, `.aux`, ...) are duplicated.

src/main/java/org/jabref/preferences/JabRefPreferences.java
koppor: Seems, I was too tired at JabCon to see the obvious.

src/main/java/org/jabref/logic/exporter/TemplateExporter.java
tobiasdiez: Sorry, but I don't really get what you want to say. You tested it and it worked although the changed code was never invoked?

https://github.com/JabRef/jabref/blob/7844fcbbea65bcbbe6eaf624352ebbe76d897946/src/main/java/org/jabref/logic/exporter/TemplateExporter.java#L161-L171
Gives a name of the form `/resource/layout/<exporter name>.layout`.  With the above change, I suspect that all layout files in `src/main/resources/resource/layout` are no longer found and thus the corresponding exporter stop to work.

koppor: TL;DR: The given name is **absolute**, so it does not matter which class is used.

Longer reasoning (with debugging JabRef):

JabRefs main class is in `org.jabref`, so the path changes from

    file:/C:/git-repositories/jabref/jabref/out/production/resources/resource/layout/html.begin.layout

to

    file:/C:/git-repositories/jabref/jabref/out/production/resources/resource/layout/html.begin.layout

So, no change when debugging JabRef.

I think, it is because the parameter `directory` is given and thus the path is an **absolute** one. (Source: https://docs.oracle.com/javase/7/docs/api/java/lang/Class.html#getResource(java.lang.String))

I would like to change the path to enable the test files residing in the test `resources` folder of "TemplateExporter".

tobiasdiez: Ok, if it works then the change is of course fine with me. I was only concerned that the exporter may break.

koppor:
OO fixed in 5c8e3e9. Tested in the UI.

So, ready for review 😇

koppor:
Since @Siedlerchr gave his "OK" and @tobiasdiez reviewed the intermediate steps, I'll fix tghe tests and merge to keep things moving.

PR #3919, opened 2018-04-03
Formatter refactoring

tobiasdiez:
Refactor the regular expression-based formatter. Previously, the regex was sent through a static `setRegex` method, now it gets passed as a normal constructor argument.

----

### DIFF
--- a/src/main/java/org/jabref/logic/formatter/Formatters.java
+++ b/src/main/java/org/jabref/logic/formatter/Formatters.java
@@ -80,22 +80,18 @@ public class Formatters {
     public static Optional<Formatter> getFormatterForModifier(String modifier) {
         Objects.requireNonNull(modifier);
 
-        if (modifier.startsWith(RegexFormatter.KEY)) {
-            String regex = modifier.substring(RegexFormatter.KEY.length());
-            return Optional.of(new RegexFormatter(regex));
-        } else {
-            Optional<Formatter> formatter = getAll().stream().filter(f -> f.getKey().equals(modifier)).findAny();
-            if (formatter.isPresent()) {
-                return formatter;
-            }
-        }
         switch (modifier) {
             case "lower":
                 return Optional.of(new LowerCaseFormatter());
             case "upper":
                 return Optional.of(new UpperCaseFormatter());
-            default:
-                return Optional.empty();
+        }
+
+        if (modifier.startsWith(RegexFormatter.KEY)) {
+            String regex = modifier.substring(RegexFormatter.KEY.length());
+            return Optional.of(new RegexFormatter(regex));
+        } else {
+            return getAll().stream().filter(f -> f.getKey().equals(modifier)).findAny();
         }
     }
 
--- a/src/main/java/org/jabref/logic/formatter/Formatters.java
+++ b/src/main/java/org/jabref/logic/formatter/Formatters.java
@@ -31,83 +31,63 @@ import org.jabref.model.cleanup.Formatter;
 
 public class Formatters {
 
-    private static final List<Formatter> CONVERTERS = Arrays.asList(
-            new HtmlToLatexFormatter(),
-            new HtmlToUnicodeFormatter(),
-            new LatexToUnicodeFormatter(),
-            new UnicodeToLatexFormatter()
-    );
-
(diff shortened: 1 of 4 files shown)


src/main/java/org/jabref/logic/formatter/Formatters.java
```diff
@@ -31,83 +31,63 @@
...
+        if (modifier.startsWith(RegexFormatter.KEY)) {
+            String regex = modifier.substring(RegexFormatter.KEY.length());
+            return Optional.of(new RegexFormatter(regex));
         } else {
-            formatter = all.stream().filter(f -> f.getKey().equals(modifier)).findAny();
-        }
-        if (formatter.isPresent()) {
-            return formatter;
+            Optional<Formatter> formatter = getAll().stream().filter(f -> f.getKey().equals(modifier)).findAny();
+            if (formatter.isPresent()) {
```
Siedlerchr: That does not make any sense does it?
The method returns an optional. So you can directly return formatter.

src/test/java/org/jabref/logic/formatter/FormatterTest.java
```diff
@@ -38,102 +38,101 @@
...
+                new LatexToUnicodeFormatter(),
+                new LowerCaseFormatter(),
+                new MinifyNameListFormatter(),
+                new NormalizeDateFormatter(),
+                new NormalizeMonthFormatter(),
+                new NormalizeNamesFormatter(),
+                new NormalizePagesFormatter(),
+                new OrdinalsToSuperscriptFormatter(),
+                new ProtectTermsFormatter(protectedTermsLoader),
+               new RegexFormatter("(\" \",\"-\")"),
```
Siedlerchr: I don't want to do nitpkicking but that Regex Formatter indentation is clear off

src/test/java/org/jabref/logic/formatter/FormatterTest.java
```diff
@@ -38,102 +38,101 @@
...
+                new UnitsToLatexFormatter(),
+                new UpperCaseFormatter());
 
+        // @formatter:on
     }
 
     @ParameterizedTest
     @MethodSource("getFormatters")
-    public void getNameReturnsNotNull(Formatter formatter) {
+    void getNameReturnsNotNull(Formatter formatter) {
```
Siedlerchr: Why do you remove the public?
Codacy or whatever will complain that you should always use explicit scoping modifiers. I see no harm in using public here.  I don't know if this even has to be public to be accessible from gradle

src/test/java/org/jabref/logic/formatter/bibtexfields/RegexFormatterTest.java
```diff
@@ -1,64 +1,55 @@
...
-import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
 /**
  * Tests in addition to the general tests from {@link org.jabref.logic.formatter.FormatterTest}
  */
-public class RegexFormatterTest {
+class RegexFormatterTest {
```
Siedlerchr: Same as above

src/main/java/org/jabref/logic/formatter/Formatters.java
tobiasdiez: If no formatter is found, then we still want to try the switch at the end. I now reordered the code a bit and it should be clearer now. Good point.

src/test/java/org/jabref/logic/formatter/FormatterTest.java
tobiasdiez: Yeah, no idea why there were these `@formatter` comments that prevented an automatic code formatting...

src/test/java/org/jabref/logic/formatter/FormatterTest.java
tobiasdiez: It's the new convention with JUnit 5 that test classes and their methods are package private. IntellJ even complains if they are marked with public.

Siedlerchr: Ah okay, that was new to me

PR #3922, opened 2018-04-04
[WIP] implement journal abbrev formatter

Siedlerchr:
I tried to implement a formatter for Journal Abbrev which can be used in the bibtexkeypattern as requested.

http://discourse.jabref.org/t/support-of-import-pattern-for-abbreviated-journal-names/1097
However, I now have a problem I am unsure how to solve.
How do I pass the JournalPrefs and the AbbrevLoader to that logic class? 
Sure, I could call 
`Globals.journalAbbreviationLoader `but that should only be acessed from the gui
@tobiasdiez @lenhard  Any idea?

----

### DIFF
--- a/src/main/java/org/jabref/logic/formatter/bibtexfields/JournalMedlineAbbreviator.java
+++ b/src/main/java/org/jabref/logic/formatter/bibtexfields/JournalMedlineAbbreviator.java
@@ -1,0 +1,54 @@
+package org.jabref.logic.formatter.bibtexfields;
+
+import java.util.Objects;
+
+import org.jabref.logic.journals.JournalAbbreviationLoader;
+import org.jabref.logic.journals.JournalAbbreviationPreferences;
+import org.jabref.model.cleanup.Formatter;
+
+public class JournalMedlineAbbreviator implements Formatter {
+
+    //TODO: How do I pass the prefs at best?
+    public JournalMedlineAbbreviator(JournalAbbreviationLoader repostioryLoader,
+            JournalAbbreviationPreferences journalAbbreviationPreferences) {
+        this.repostioryLoader = Objects.requireNonNull(repostioryLoader);
+        this.journalAbbreviationPreferences = Objects.requireNonNull(journalAbbreviationPreferences);
+    }
+
+    @Override
+    public String getName() {
+        // TODO Auto-generated method stub
+        return null;
+    }
+
+    @Override
+    public String getKey() {
+        // TODO Auto-generated method stub
+        return null;
+    }
+
+    private final JournalAbbreviationLoader repostioryLoader;
+    private final JournalAbbreviationPreferences journalAbbreviationPreferences;
+
+
+
+    @Override
+    public String format(String fieldText) {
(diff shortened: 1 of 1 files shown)


tobiasdiez:
The only place where formatter are instantiated are in `Formatters`. Thus you need to pass the dependencies there (I would propose as a constructor argument and making all static methods in this class instance methods). Probably you need to propagate the dependencies even a bit further until you reach a point where you can use `Globals`.

Siedlerchr:
Damn. Always when I think this will be quick thing to do, it emerges to a more complicated thing 👎 .
Seems like I have to introduce a GUI class for all that formatter as a wrapper/starting point

src/main/java/org/jabref/logic/formatter/bibtexfields/JournalMedlineAbbreviator.java
```diff
@@ -0,0 +1,52 @@
...
+import org.jabref.logic.journals.JournalAbbreviationPreferences;
+import org.jabref.model.cleanup.Formatter;
+
+public class JournalMedlineAbbreviator implements Formatter {
+
+    private final JournalAbbreviationLoader repostioryLoader;
+    private final JournalAbbreviationPreferences journalAbbreviationPreferences;
+
+    //TODO: How do I pass the prefs at best?
+    public JournalMedlineAbbreviator(JournalAbbreviationLoader repostioryLoader,
```
tobiasdiez: Its probably simpler to only pass the repository as the dependency.

tobiasdiez:
What's the status of this PR?

src/main/java/org/jabref/logic/formatter/bibtexfields/JournalMedlineAbbreviator.java
Siedlerchr: Still the problem is that nearly the whole codebase of JabRef is affected as I need to pass the arg from somewhere to the formatter interface

Sethur:
Is there any progress on this? This feature would be extremely helpful for us.

tobiasdiez:
I had a second look and the dependency problem and think it is not such a big issue. I would propose to
- Introduce a class FormatterProperties that hold the necessary dependencies
- Pass it to the Formatters class as a constructor argument https://github.com/JabRef/jabref/blob/master/src/main/java/org/jabref/logic/formatter/Formatters.java#L35
- The only logic classes that use the formatters are in Cleanup  and BracketedPattern. For the cleanup, I would use a similar constructor injection, for the bracketedpattern I am not sure if it might be easier to pass it as an method argument. 

It is indeed a bit of code changes but nothing hard or complicated. But I think it is worth the work as it makes the formatter ready for further enhancements in the future.

Siedlerchr:
I digged a bit through the code again and I currently see this as nearly impossible to implement without changing the whole codebase of JabRef.
Despite BrackedPattern the main culprit is the static `Cleanup.parse `method which is called over several hierarchy levels from the MetadataParser and from the BibTexParser and the SharedDBCode.
So much for easy dependency injection...:(

PR #3931, opened 2018-04-06
Resolve #3929 Correct fetcher messages

### LINKED ISSUE #3929: Fetcher messages misleading
`Full text document download failed` does not indicate a failed download but that no accessible URL was found!

stefan-kolb:
Resolve #3929


stefan-kolb:
@Siedlerchr We do?!

Issue #674, opened 2016-01-15
Add a proper way to set external application for pushing


ziggystar:
I searched for 10 minutes how to change the external application for pushing from emacs to vim, when I finally found the drop-down next to the tool bar button.

Please add a proper way of changing this setting through a menu action. I think it's not good style to have an action only available through the tool bar.

oscargus:
What would you suggest to be a proper way?

ziggystar:
One possible way would be adding a menu entry (with a sub menu) right beside `Tools -> Push Entries`.

And you're sure there is no other way to set this? Maybe I've just missed it.

oscargus:
You can edit the preferences file. :-)

I think the idea is that since one (typcially) press the button to push, having the setting exactly there is quite convenient. On the other hand, since most users do not change it continuously one can ask if it is needed. I think an additional dropdown in the preferences tab may make sense. Maybe a submenu below the "Push entries...". Let's see who implements it. :-)

lenhard:
related  #382

ziggystar:
How should this function work, suppose one were to implement it using Actions?

## One action for pushing to each application

### Pro
- ability to configure hotkeys for each application

### Con 
- you'd have more than one button and menu entry (one for push to emacs, one for push to vim...), requires some tricks to reduce burden on user (see below)

I think the current implementation partially belongs here, since you have this button with a selectable target application. But as I noted, the menu-way of accessing these actions is broken, since you cannot change the target of the menu action. The canonical way to transfer this to the menu would be a sub menu to select the target application to push to, but this is cumbersome. 

## One push action, select target application in preferences

### Pro
 - obvious how to implement button, menu, hotkey
 - clear design

### Con
 - cumbersome if you are working with multiple applications (actually I used both Vim and Emacs some weeks ago and wanted to insert references in both, but I suppose this is a rarely used feature)

## Configurable push actions

Another approach would be to not have all available applications as possible targets, but to configure explicit targets. This could include Vim twice, but with different parameters.

This would enable using two different instances of Vim, and selectively pushing to each instance from Jabref, while never having to deal with other targets in the UI (only within the preferences.

Since this would reduce the number of targets (to one in most cases), one could have a dedicated button, configurable hotkey, or menu entry, for each target.

### Pro
- reduces number of targets (usually one, more for power users)
- targeting different instances of the same application

### Con
- requires most changes to code base
- unclear how default looks like; if no application is configured initially, the feature would become invisible in the UI

[...]

PR #3942, opened 2018-04-10
[WIP] Add External Application Selection to Preferences

slyr91:
WIP - Add External Application Selection to Preferences

This pull request is in regards to #674. 
This initial commit adds a button to the settings panel for each external application which allows the user to select an application to make default. There is currently a problem with this build where the application will not update the pushToApplicationButton. I believe the issue has something to do with the interaction between javafx and swing components. Also I plan on adding keybinding options as well in the future, it's just taking me a bit of time to familiarize myself with the program between life events. Let me know what you think so far and on the direction I am taking.

### DIFF
--- a/src/main/java/org/jabref/gui/JabRefFrame.java
+++ b/src/main/java/org/jabref/gui/JabRefFrame.java
@@ -206,6 +206,7 @@ public class JabRefFrame extends BorderPane implements OutputPrinter {
     private final CountingUndoManager undoManager = new CountingUndoManager();
     private final DialogService dialogService;
     private SidePane sidePane;
+    private PushToApplicationButton pushToExternal;
 
     public JabRefFrame(Stage mainStage) {
         this.mainStage = mainStage;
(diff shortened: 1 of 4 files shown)


tobiasdiez:
This looks like a good start. Nice work!

The button is not updated because such an update mechanism is not yet implemented. The icon is generated from an action (which itself is created in `PushToApplicationButton.getMenuAction`). However, the [Action class](https://github.com/JabRef/jabref/blob/maintable-beta/src/main/java/org/jabref/gui/actions/Action.java) only has static return values. I would propose to follow the example of [controlsfx Action](https://controlsfx.bitbucket.io/org/controlsfx/control/action/Action.html) and replace the getters in the action interface by javafx properties. Then you can use bindings in the adapter class [JabRefAction](https://github.com/JabRef/jabref/blob/maintable-beta/src/main/java/org/jabref/gui/actions/JabRefAction.java) and in this way update the UI if the underlying property is changed. 

@slyr91 do you want to implement these changes or prefer if somebody from the team does it? (I would volunteer but it may take some time).

slyr91:
I will definitely give this a solid try. @tobiasdiez I am still learning about the proper procedures on GitHub but if I can implement those changes should I add them to this pull request or to a separate one?

tobiasdiez:
Good question. In general, it is better to create a few smaller PRs because that makes reviewing and understanding the code changes easier. Having said this, in this particular case, I think however that the changes are small and localized enough so that you can simply push them to this PR.

stefan-kolb:
@slyr91 Will you work on this in the future?

slyr91:
@stefan-kolb I definitely will give it a shot soon. I have just been too busy to dive into this. I had finals to deal with and now I am trying to study for a certification so all my free time goes to that. As soon as I am done with that I will be returning to this.

slyr91:
I'm getting back into this now and I noticed that the branch this was originally built on, maintainable-beta, is closed. Which branch should I apply these changes and proposed improvements to?

stefan-kolb:
Hi, on the `master` branch please 😄

Siedlerchr:
You probably have to rework your code a bit as recently the preferences dialog was transformed to javaFX as well.

tobiasdiez:
Surpassed by #5024.

PR #3947, opened 2018-04-12
Pdf exporter - delete xmp actions in the menu bar and the cli option

johannes-manner:
The idea is, to use a new exporter for writting bib entries to pdf metadata.
It enables users to select entries and export these entries via File->Export Selected Entries to a pdf file.
This exporter can also be used within the -o cli infrastructure.

java -jar JabRef.jar -o   path/article.pdf,pdf   path/entries.bib

Therefore, the xmp cli and the xmp actions in the menu bar can be deleted.

Opinions?

----


tobiasdiez:
I think having a `PdfExporter` (maybe rename to `XmpPdfExporter`?) is a good idea, especially for the cli support we get in this way. However, I would keep the "Write XMP" action in the menu. Personally, I would never look under "Export" when I want to to add metadata to my pdfs, although you are of course right that technically it is an export of some kind.

johannes-manner:
I deleted the cli option -xmp (integrated with https://github.com/JabRef/jabref/pull/3756), because the new XmpPdfExporter now enables the cli interaction via the -o exporter mechanism.

src/main/java/org/jabref/logic/exporter/XmpPdfExporter.java
```diff
@@ -0,0 +1,29 @@
...
+    private final XmpPreferences xmpPreferences;
+
+    public XmpPdfExporter(XmpPreferences xmpPreferences) {
+        super("pdf", FileType.PDF_XMP.getDescription(), FileType.PDF_XMP);
+        this.xmpPreferences = xmpPreferences;
+    }
+
+    @Override
+    public void export(BibDatabaseContext databaseContext, Path pdfFile, Charset encoding, List<BibEntry> entries) throws Exception {
+        if (pdfFile.toString().endsWith(".pdf")) {
```
Siedlerchr: Is this necessary? I think for an export it should not play a role if my flie is named  `test.asdf` for export, as long as the format is correct?

Siedlerchr:
Please have a look at the faling exporter test. Then we can merge

PR #3950, opened 2018-04-12
Improve search performance

Siedlerchr:
Before executing the search there is now a short delay.
I tested this with the Huge file bib and the performance is way better.

Still the entry editor has a high load, but that's out of scope.

Thx to @halirutan  who pointed me to the code in  #3553

----

### DIFF
--- a/src/main/java/org/jabref/gui/search/GlobalSearchBar.java
+++ b/src/main/java/org/jabref/gui/search/GlobalSearchBar.java
@@ -64,9 +64,11 @@ import org.slf4j.LoggerFactory;
 @SuppressWarnings("Duplicates")
 public class GlobalSearchBar extends JPanel {
 
+    private static final Logger LOGGER = LoggerFactory.getLogger(GlobalSearchBar.class);
+
+    private static final int SEARCH_DELAY = 400;
     private static final PseudoClass CLASS_NO_RESULTS = PseudoClass.getPseudoClass("emptyResult");
     private static final PseudoClass CLASS_RESULTS_FOUND = PseudoClass.getPseudoClass("emptyResult");
-    private static final Logger LOGGER = LoggerFactory.getLogger(GlobalSearchBar.class);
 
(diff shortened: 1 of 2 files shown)


stefan-kolb:
Does this swallow event before the delay or does it just execute the search at a later time one by one?

Siedlerchr:
@stefan-kolb  The timer is resetted when you type new chars

src/main/java/org/jabref/gui/search/GlobalSearchBar.java
```diff
@@ -186,7 +191,13 @@ public void actionPerformed(ActionEvent e) {
         updateSearchModeButtonText();
         searchModeButton.addActionListener(event -> toggleSearchModeAndSearch());
 
-        EasyBind.subscribe(searchField.textProperty(), searchText -> performSearch());
+        //Add a delay of 400 milliseconds before starting search
+        Timer searchTask = FxTimer.create(Duration.ofMillis(400), () -> {
```
stefan-kolb: Maybe extract the search delay as a constant

src/main/java/org/jabref/gui/search/GlobalSearchBar.java
```diff
@@ -55,13 +56,17 @@
...
+import org.reactfx.util.Timer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 @SuppressWarnings("Duplicates")
 public class GlobalSearchBar extends JPanel {
 
     private static final PseudoClass CLASS_NO_RESULTS = PseudoClass.getPseudoClass("emptyResult");
     private static final PseudoClass CLASS_RESULTS_FOUND = PseudoClass.getPseudoClass("emptyResult");
+    private static final Logger LOGGER = LoggerFactory.getLogger(GlobalSearchBar.class);
```
stefan-kolb: Logger should be first attribute in the class

PR #3954, opened 2018-04-14
Add button-icon for union/intersection in the groups side panel

### LINKED ISSUE #3269: 4.0 Groups: fast access to View options

Siedlerchr:
Fixes #3269
Fix display of state for union/intersection in preferences

As we currently unfortunately don't have Icons for Set Intersection/Union 
`set-center` and `set-all`  https://materialdesignicons.com/icon/set-center 
 in the Material Design Font, I used the WWW and the twitter icon to demonstrate the purpose. @halirutan  will then try to integrate the correoct  icons in JabRef 

grafik

----


src/main/java/org/jabref/gui/SidePaneComponent.java
```diff
@@ -88,7 +90,7 @@ public final Node getContentPane() {
     /**
      * @return the header pane for this component
      */
-    public final Node getHeader() {
+    public Node getHeader() {
```
tobiasdiez: Instead of overriding this method, I would propose to create a new (overridable) method `getHeaderButtons` that results additional header button to be displayed. In this way you ensure consistency as well as reduce code duplication.

src/main/java/org/jabref/gui/SidePaneComponent.java
```diff
@@ -98,13 +100,13 @@ public final Node getHeader() {
...
         down.setOnAction(event -> moveDown());
 
-        HBox buttonContainer = new HBox();
+        final HBox buttonContainer = new HBox();
         buttonContainer.getChildren().addAll(up, down, close);
         BorderPane graphic = new BorderPane();
         graphic.setCenter(icon.getGraphicNode());
-        BorderPane container = new BorderPane();
-//        container.setLeft(graphic);
+        //        container.setLeft(graphic);
```
tobiasdiez: Remove?

src/main/java/org/jabref/gui/groups/GroupSidePane.java
```diff
@@ -46,6 +89,21 @@ public Action getToggleAction() {
         return StandardActions.TOGGLE_GROUPS;
     }
 
+    private void toggleUnionIntersection() {
+        boolean intersectionToggled = preferences.getBoolean(JabRefPreferences.GROUP_INTERSECT_SELECTIONS);
+        boolean newState = !intersectionToggled;
+        preferences.putBoolean(JabRefPreferences.GROUP_INTERSECT_SELECTIONS, newState);
+
+        BasePanel basePanel = frame.getCurrentBasePanel();
```
tobiasdiez: Please remove the dependency on `frame` and replace it by dialog service (which has a `notify` method).

src/main/java/org/jabref/gui/preftabs/GroupsPrefsTab.java
```diff
@@ -114,7 +110,14 @@ public void setValues() {
         groupingField.setText(prefs.get(JabRefPreferences.GROUPS_DEFAULT_FIELD));
         keywordSeparator.setText(prefs.get(JabRefPreferences.KEYWORD_SEPARATOR));
         autoAssignGroup.setSelected(prefs.getBoolean(JabRefPreferences.AUTO_ASSIGN_GROUP));
-        multiSelectionModeIntersection.setSelected(prefs.getBoolean(JabRefPreferences.GROUP_INTERSECT_SELECTIONS));
+
+        boolean intersectionSelected = prefs.getBoolean(JabRefPreferences.GROUP_INTERSECT_SELECTIONS);
+        //Looks odd, but the prefs store true for intersection and false for union
+        if (intersectionSelected) {
```
tobiasdiez: Since this is hard to understand, I would propose to add a better named method to `JabrefPreferences`. Moreover, the code is probably easier to understand if you add an enum for the different modes.

Siedlerchr: Now introduced an enum with methods for setting/get the pref value

src/main/java/org/jabref/gui/SidePaneComponent.java
Siedlerchr: Added a protected method for gettingAddtionalHeader buttons

src/main/java/org/jabref/gui/preftabs/GroupsPrefsTab.java
tobiasdiez: Personally, I would separate the group mode enum from the icon. The mode is something that can (in principle) be defined and used in the model/logic package thus I wouldn't mix it with something gui related.

Siedlerchr: Well, however, this is purely a gui thing.  My immediate thoughts were also about storing that in model, but it would provide any additional value. And this makes it easier to directly use the icon,

Siedlerchr:
@tobiasdiez  Is this now okay? Then you could merge it with CO-Authors

src/main/java/org/jabref/gui/groups/GroupSidePane.java
```diff
@@ -20,15 +25,43 @@
 public class GroupSidePane extends SidePaneComponent {
 
     private final JabRefPreferences preferences;
+    private final DialogService dialogService;
+    private final Button intersectionUnionToggle = IconTheme.JabRefIcons.WWW.asButton();
+    private final Tooltip toggleUnion = new Tooltip(Localization.lang("Toogle union"));
```
tobiasdiez: I think there is no need to have these tooltips as instance variables. I would simply inline them.

src/main/java/org/jabref/gui/groups/GroupSidePane.java
```diff
@@ -20,15 +25,43 @@
...
+    }
+
+    @Override
+    protected Optional<Node> getAddtionalHeaderButtons() {
+        intersectionUnionToggle.setOnAction(event -> toggleUnionIntersection());
+        return Optional.of(intersectionUnionToggle);
+    }
+
+    private Node getUnionIntersectionGraphic() {
+        return preferences.getGroupViewMode().getIcon().getGraphicNode();
```
tobiasdiez: Personally, I would extract these methods to a new class `GroupViewModeViewModel` which gives you the graphic and the tooltip (constructor accepts the mode). In this way, you clearly separate everything ui-related from the actual model class.

src/main/java/org/jabref/preferences/JabRefPreferences.java
```diff
@@ -193,7 +194,8 @@
     public static final String EDITOR_EMACS_KEYBINDINGS_REBIND_CA = "editorEMACSkeyBindingsRebindCA";
     public static final String EDITOR_EMACS_KEYBINDINGS_REBIND_CF = "editorEMACSkeyBindingsRebindCF";
     public static final String GROUPS_DEFAULT_FIELD = "groupsDefaultField";
-    public static final String GROUP_INTERSECT_SELECTIONS = "groupIntersectSelections";
+    public static final String GROUP_INTERSECT_UNION_VIEW_MODE = "groupIntersectUnionViewModes";
```
tobiasdiez: this can be private now, right?

Siedlerchr:
@tobiasdiez  Fixed the issues you m entionend.

src/main/java/org/jabref/gui/groups/GroupModeViewModel.java
```diff
@@ -0,0 +1,24 @@
...
+import javafx.scene.Node;
+import javafx.scene.control.Tooltip;
+
+import org.jabref.logic.l10n.Localization;
+
+public class GroupModeViewModel {
+
+    public static Tooltip getUnionIntersectionTooltip(GroupViewMode mode) {
+        if (mode == GroupViewMode.UNION) {
+            return new Tooltip(Localization.lang("Toogle intersection"));
```
tobiasdiez: The texts seem to be interchanged by mistake.

src/main/java/org/jabref/gui/groups/GroupModeViewModel.java
```diff
@@ -0,0 +1,24 @@
...
+        }
+        if (mode == GroupViewMode.INTERSECTION) {
+            return new Tooltip(Localization.lang("Toogle union"));
+        }
+        return new Tooltip();
+
+    }
+
+    public static Node getUnionIntersectionGraphic(GroupViewMode mode) {
+        return mode.getIcon().getGraphicNode();
```
tobiasdiez: Don't you agree that a code similar to the above one for the tooltip is better than storing the icon directly in the `groupviewmode`?

src/main/java/org/jabref/preferences/JabRefPreferences.java
tobiasdiez: ?

Siedlerchr:
I fixed the remaining issues and I am going to merge this one now

PR #3964, opened 2018-04-18
Fix xmp exporter

johannes-manner:
Solves [koppor#338](https://github.com/koppor/jabref/issues/338).

Changed the behavior of the xmp exporter. Remove the ?xpacket tags with xmp metadata in dublin core format to enable users to use the .xmp files, exported by JabRef, with the xmpincl LaTeX package. 

Also adressed the month parsing problem, implemented a workaround and created an issue for the date section within xmpbox (https://issues.apache.org/jira/browse/PDFBOX-4196).


src/main/java/org/jabref/logic/exporter/XmpExporter.java
```diff
@@ -64,8 +67,16 @@ public void export(BibDatabaseContext databaseContext, Path file, Charset encodi
     }
 
     private void writeBibToXmp(Path file, List<BibEntry> entries, Charset encoding) throws IOException {
+        String xmpContent = XmpUtilWriter.generateXmpString(entries, this.xmpPreferences);
```
koppor: Can we add a second method there? `generateXmpStringWithoutXmpDeclaration`? Maybe rename to current functio to `generateXmpStringWithXmpDeclaration`. The term "XmpDeclaration" stems from "XmlDeclaration". See https://wiki.selfhtml.org/wiki/XML/Regeln/XML-Deklaration.

src/main/java/org/jabref/logic/xmp/DublinCoreExtractor.java
```diff
@@ -61,12 +63,12 @@ private void extractAuthor() {
     }
 
     /**
-     * Year + Month in BibTex - Date in DublinCore is a combination of year and month information
+     * Year in BibTex - Date in DublinCore is only the year information, because dc interprets empty months as January.
```
koppor: I don't get this change? The old code from line 81 on seems to extract the month perfectly. In case no month is there, the fallback was and still seems to be January.

src/main/java/org/jabref/logic/xmp/DublinCoreExtractor.java
```diff
@@ -134,17 +132,33 @@ private void extractPublisher() {
...
                 }
             }
+
+        };
+        List<String> relationships = dcSchema.getRelations();
+        if (relationships != null) {
+            relationships.stream()
+                    .filter(isBibTeXElement)
+                    .forEach(splitBibTeXElement);
+
```
koppor: Remove empty line 😇

src/main/java/org/jabref/logic/xmp/DublinCoreExtractor.java
```diff
@@ -134,17 +132,33 @@ private void extractPublisher() {
...
     private void extractBibTexFields() {
-        List<String> relationships = dcSchema.getRelations();
-        if (relationships != null) {
-            for (String r : relationships) {
-                if (r.startsWith("bibtex/")) {
-                    r = r.substring("bibtex/".length());
-                    int i = r.indexOf('/');
-                    if (i != -1) {
-                        bibEntry.setField(r.substring(0, i), r.substring(i + 1));
+        Predicate<String> isBibTeXElement = s -> s.startsWith("bibtex/");
```
koppor: Move this down just after `List<String relationships =`.

src/main/java/org/jabref/logic/xmp/DublinCoreExtractor.java
```diff
@@ -134,17 +132,33 @@ private void extractPublisher() {
...
+        Consumer<String> splitBibTeXElement = s -> {
+            // split solution is complicated, because some fields contains url etc.
+            String temp = s.substring("bibtex/".length());
+            int i = temp.indexOf('/');
```
koppor: Please add an example why you need this substring. It is clear from the code that it is bibtex/key/value, but it should be documented somewhere (and referenced from here).

src/main/java/org/jabref/logic/xmp/DublinCoreExtractor.java
```diff
@@ -134,17 +132,33 @@ private void extractPublisher() {
...
+            if (i != -1) {
+                String key = temp.substring(0, i);
+                String value = temp.substring(i + 1);
+                bibEntry.setField(key, value);
+
+                // only for month field - override value
+                // workaround, because the date value of the xmp component of pdf box is corrupted
+                if ("month".equals(key)) {
```
koppor: We are in the special bibtex mode. So, we hit `bibtex/month/#mar#` in this case, didn't we? Did we hit `bibtex/month/mar`? In the former case, I would assume, the export is corrupted. In the latter case, I don't know, what's happening. In all cases, it strange that the date value of PdfBox is corrupt; we are in our custom bibtex mode 😇

src/main/java/org/jabref/logic/xmp/DublinCoreExtractor.java
johannes-manner: You are right with the schema bibtex/key/value. 
In a former approach I used String#split, but in some cases the value is an url or doi etc. 

So the value in this case can contain "/" and therefore the substring is the easiest way to handle this. Documenting the url problem would be beneficial for future changes.

koppor:
Thank you for the work! I got a minor comment on the "January, 1st" hack: There should be a comment on that. Or a [MADR](https://github.com/adr/madr) linked via `// ADR(number)` 😀 . Can be done in leisure time 😃

PR #3965, opened 2018-04-18
Fixed Url column

### LINKED ISSUE #3860: The `URL` column missed in the maintable beta version
Look at the screenshot:

image

florian-beetz:
This fixes #3860 

URL Column now shows URL or DOI correctly and opens on click.

----

### DIFF
--- a/src/main/java/org/jabref/gui/maintable/CellFactory.java
+++ b/src/main/java/org/jabref/gui/maintable/CellFactory.java
@@ -14,6 +14,7 @@ import org.jabref.gui.icon.JabRefIcon;
 import org.jabref.gui.specialfields.SpecialFieldViewModel;
 import org.jabref.model.entry.FieldName;
 import org.jabref.model.entry.specialfields.SpecialField;
+import org.jabref.model.strings.StringUtil;
 
 public class CellFactory {
 
@@ -95,6 +96,10 @@ public class CellFactory {
     }
 
     public Node getTableIcon(String fieldType) {
+        if (StringUtil.isBlank(fieldType)) {
+            return null;
+        }
+
         JabRefIcon icon = TABLE_ICONS.get(fieldType);
         if (icon == null) {
             //LOGGER.info("Error: no table icon defined for type '" + fieldType + "'.");
--- a/src/main/java/org/jabref/gui/maintable/CellFactory.java
+++ b/src/main/java/org/jabref/gui/maintable/CellFactory.java
@@ -10,96 +10,98 @@ import javafx.scene.Node;
 import org.jabref.gui.externalfiletype.ExternalFileType;
 import org.jabref.gui.externalfiletype.ExternalFileTypes;
 import org.jabref.gui.icon.IconTheme;
+import org.jabref.gui.icon.JabRefIcon;
 import org.jabref.gui.specialfields.SpecialFieldViewModel;
 import org.jabref.model.entry.FieldName;
 import org.jabref.model.entry.specialfields.SpecialField;
 
 public class CellFactory {
 
-    private final Map<String, Node> TABLE_ICONS = new HashMap<>();
+    private final Map<String, JabRefIcon> TABLE_ICONS = new HashMap<>();
 
     public CellFactory(ExternalFileTypes externalFileTypes, UndoManager undoManager) {
-        Node label;
-        label = IconTheme.JabRefIcons.PDF_FILE.getGraphicNode();
-        //label.setToo(Localization.lang("Open") + " PDF");
-        TABLE_ICONS.put(FieldName.PDF, label);
+        JabRefIcon icon;
+        icon = IconTheme.JabRefIcons.PDF_FILE;
+        //icon.setToo(Localization.lang("Open") + " PDF");
+        TABLE_ICONS.put(FieldName.PDF, icon);
 
-        label = IconTheme.JabRefIcons.WWW.getGraphicNode();
-        //label.setToolTipText(Localization.lang("Open") + " URL");
-        TABLE_ICONS.put(FieldName.URL, label);
+        icon = IconTheme.JabRefIcons.WWW;
+        //icon.setToolTipText(Localization.lang("Open") + " URL");
+        TABLE_ICONS.put(FieldName.URL, icon);
 
-        label = IconTheme.JabRefIcons.WWW.getGraphicNode();
-        //label.setToolTipText(Localization.lang("Open") + " CiteSeer URL");
-        TABLE_ICONS.put("citeseerurl", label);
+        icon = IconTheme.JabRefIcons.WWW;
+        //icon.setToolTipText(Localization.lang("Open") + " CiteSeer URL");
+        TABLE_ICONS.put("citeseerurl", icon);
 
-        label = IconTheme.JabRefIcons.WWW.getGraphicNode();
-        //label.setToolTipText(Localization.lang("Open") + " ArXiv URL");
-        TABLE_ICONS.put(FieldName.EPRINT, label);
+        icon = IconTheme.JabRefIcons.WWW;
+        //icon.setToolTipText(Localization.lang("Open") + " ArXiv URL");
+        TABLE_ICONS.put(FieldName.EPRINT, icon);
 
-        label = IconTheme.JabRefIcons.DOI.getGraphicNode();
-        //label.setToolTipText(Localization.lang("Open") + " DOI " + Localization.lang("web link"));
-        TABLE_ICONS.put(FieldName.DOI, label);
+        icon = IconTheme.JabRefIcons.DOI;
+        //icon.setToolTipText(Localization.lang("Open") + " DOI " + Localization.lang("web link"));
+        TABLE_ICONS.put(FieldName.DOI, icon);
 
-        label = IconTheme.JabRefIcons.FILE.getGraphicNode();
-        //label.setToolTipText(Localization.lang("Open") + " PS");
-        TABLE_ICONS.put(FieldName.PS, label);
+        icon = IconTheme.JabRefIcons.FILE;
+        //icon.setToolTipText(Localization.lang("Open") + " PS");
+        TABLE_ICONS.put(FieldName.PS, icon);
 
-        label = IconTheme.JabRefIcons.FOLDER.getGraphicNode();
-        //label.setToolTipText(Localization.lang("Open folder"));
-        TABLE_ICONS.put(FieldName.FOLDER, label);
+        icon = IconTheme.JabRefIcons.FOLDER;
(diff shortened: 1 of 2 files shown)


src/main/java/org/jabref/gui/maintable/MainTableColumnFactory.java
```diff
@@ -269,20 +277,50 @@ private ContextMenu createFileMenu(BibEntryTableViewModel entry, List<LinkedFile
...
+     * The {@code firstField} is preferred to be shown over {@code secondField}.
      */
-    private TableColumn<BibEntryTableViewModel, String> createIconColumn(JabRefIcon icon, String firstField, String secondField) {
+    private TableColumn<BibEntryTableViewModel, String> createUrlOrDoiColumn(JabRefIcon icon, String firstField, String secondField) {
         TableColumn<BibEntryTableViewModel, String> column = new TableColumn<>();
         column.setGraphic(icon.getGraphicNode());
         column.getStyleClass().add(ICON_COLUMN);
         setExactWidth(column, GUIGlobals.WIDTH_ICON_COL);
-        column.setCellValueFactory(cellData -> EasyBind.monadic(cellData.getValue().getField(firstField)).orElse(cellData.getValue().getField(secondField)));
+        column.setCellValueFactory(cellData -> EasyBind.monadic(cellData.getValue().getField(firstField)).map(x -> firstField).orElse(EasyBind.monadic(cellData.getValue().getField(secondField)).map(x -> secondField)));
```
tobiasdiez: whats the purpose of the added `map` calls? They seem to always replace the cell value by `firstField` and `secondField`, respectively.

src/main/java/org/jabref/gui/maintable/MainTableColumnFactory.java
```diff
@@ -269,20 +277,50 @@ private ContextMenu createFileMenu(BibEntryTableViewModel entry, List<LinkedFile
...
+                field = firstField;
+            } else if (entry.getEntry().hasField(secondField)) {
+                field = secondField;
+            } else {
+                // nothing to open
+                LOGGER.debug("Requested opening DOI or URL, but none present.");
+                return;
+            }
+
+            entry.getEntry().getField(field).ifPresent(url -> {
```
tobiasdiez: I think this needs to be
````java
entry.getEntry()
   .getField(field)
   .map(value -> IdentifierParser.parse(field, value))
   .flatMap(Identifier::getExternalURI)
   .ifPresent(
                url -> {
                    try {
                        JabRefDesktop.openBrowser(url);
                    } catch (IOException ex) {
                        dialogService.showErrorDialogAndWait(Localization.lang("Unable to open link."), ex);
                    }
                }
    );
````
because the doi may not be present as a url but only in short form. Or this automatically handled by `openExternalViewer`? @JabRef/developers whats the preferred way to open dois/url ?

File: src/main/java/org/jabref/gui/maintable/MainTableColumnFactory.java (hunk 1 above)
Comment (florian-beetz) [reply src/main/java/org/jabref/gui/maintable/MainTableColumnFactory.java]: Yes, that's exactly what's happening. The reason here is that `CellFactory` is expecting the field name to choose the icon.
I thought about trying to identify if the column contains a DOI or an URL, but I think this is the better approach, as here this information is already there. 
I'm open for different approaches though.

File: src/main/java/org/jabref/gui/maintable/MainTableColumnFactory.java (hunk 2 above)
Comment (florian-beetz) [reply src/main/java/org/jabref/gui/maintable/MainTableColumnFactory.java]: `openExternalViewer` can handle DOIs directly. I did not know about `IdentifierParser`, but I can change it, if that's better.

**Comment by LinusDietz** [comment]
Hi @florian-beetz, since you have been involved in the maintable migrations do you want to join tomorrow's Devcall at 11 am? Drop me a note by mail (linus@lynyus.de). Cheers, Linus

**Comment by stefan-kolb** [comment]
It looks like `MainTableSelectionListener` uses `JabRefDesktop.openExternalViewer(panel.getBibDatabaseContext(), link, fieldName);` Line 259.

File: src/main/java/org/jabref/gui/maintable/MainTableColumnFactory.java (hunk 2 above)
Comment (stefan-kolb) [reply src/main/java/org/jabref/gui/maintable/MainTableColumnFactory.java]: It looks like MainTableSelectionListener uses JabRefDesktop.openExternalViewer(panel.getBibDatabaseContext(), link, fieldName); Line 259.

File: src/main/java/org/jabref/gui/maintable/MainTableColumnFactory.java (hunk 1 above)
Comment (tobiasdiez) [reply src/main/java/org/jabref/gui/maintable/MainTableColumnFactory.java]: ok, makes sense. The code looks a bit strange but I'm not able to find a more readable solution. Could you please add a short comment explaining the construction. Thanks!

File: src/main/java/org/jabref/gui/maintable/CellFactory.java [hunk 3]
```diff
@@ -10,96 +10,98 @@
...
-        TABLE_ICONS.put(SpecialField.PRINTED.getFieldName(), label);
+        icon = printedViewModel.getIcon();
+        //icon.setToolTipText(printedViewModel.getLocalization());
+        TABLE_ICONS.put(SpecialField.PRINTED.getFieldName(), icon);
     }
 
     public Node getTableIcon(String fieldType) {
-        Node label = TABLE_ICONS.get(fieldType);
-        if (label == null) {
+        JabRefIcon icon = TABLE_ICONS.get(fieldType);
```
tobiasdiez: If I understand the code below correctly, then this method is hit quite often with an empty (or null) string. In these cases (checked using `StringUtil.isBlank`), we can simply return null before looking in the `Map` for a suitable icon.

src/main/java/org/jabref/gui/maintable/MainTableColumnFactory.java
```diff
@@ -269,20 +277,50 @@ private ContextMenu createFileMenu(BibEntryTableViewModel entry, List<LinkedFile
...
+    private TableColumn<BibEntryTableViewModel, String> createUrlOrDoiColumn(JabRefIcon icon, String firstField, String secondField) {
         TableColumn<BibEntryTableViewModel, String> column = new TableColumn<>();
         column.setGraphic(icon.getGraphicNode());
         column.getStyleClass().add(ICON_COLUMN);
         setExactWidth(column, GUIGlobals.WIDTH_ICON_COL);
-        column.setCellValueFactory(cellData -> EasyBind.monadic(cellData.getValue().getField(firstField)).orElse(cellData.getValue().getField(secondField)));
+        column.setCellValueFactory(cellData -> EasyBind.monadic(cellData.getValue().getField(firstField)).map(x -> firstField).orElse(EasyBind.monadic(cellData.getValue().getField(secondField)).map(x -> secondField)));
                 new ValueTableCellFactory<BibEntryTableViewModel, String>()
                         .withGraphic(cellFactory::getTableIcon)
+                        .withOnMouseClickedEvent((entry, content) -> createOpenUrlOrDoiHandler(entry, firstField, secondField))
```
tobiasdiez: I hope a construction like `(entry, content) -> event -> openurlOrDoi(event, entry, content)` with method `openUrlOrDoi(MouseEvent event, BibentryTableViewmodel entry, String field)` should work and is a bit easier to parse.

src/main/java/org/jabref/gui/maintable/MainTableColumnFactory.java
```diff
@@ -269,20 +277,50 @@ private ContextMenu createFileMenu(BibEntryTableViewModel entry, List<LinkedFile
...
+                    JabRefDesktop.openExternalViewer(database, url, field);
+                } catch (IOException e) {
+                    LOGGER.error("Failed to open external viewer for '{}'", url, e);
+                }
+            });
+            event.consume();
+        };
+    }
+
     private TableColumn<BibEntryTableViewModel, String> createIconColumn(JabRefIcon icon, String field) {
```
tobiasdiez: Could you please fix this method for the `eprint` column as well. Thanks!

src/main/java/org/jabref/gui/maintable/MainTableColumnFactory.java
tobiasdiez: ok, then simply ignore my remark and keep the `openExternalViewer` call.

src/main/java/org/jabref/gui/maintable/MainTableColumnFactory.java
florian-beetz: Done!

src/main/java/org/jabref/gui/maintable/MainTableColumnFactory.java
florian-beetz: Done!

src/main/java/org/jabref/gui/maintable/CellFactory.java
florian-beetz: Done!

src/main/java/org/jabref/gui/maintable/MainTableColumnFactory.java
florian-beetz: I used `EasyBind.map()` directly, this makes this line somewhat more readable. Additionally to the comment you suggested.

src/main/java/org/jabref/gui/maintable/CellFactory.java
```diff
@@ -10,96 +10,103 @@
...
+    private final Map<String, JabRefIcon> TABLE_ICONS = new HashMap<>();
 
     public CellFactory(ExternalFileTypes externalFileTypes, UndoManager undoManager) {
-        Node label;
-        label = IconTheme.JabRefIcons.PDF_FILE.getGraphicNode();
-        //label.setToo(Localization.lang("Open") + " PDF");
-        TABLE_ICONS.put(FieldName.PDF, label);
+        JabRefIcon icon;
+        icon = IconTheme.JabRefIcons.PDF_FILE;
+        //icon.setToo(Localization.lang("Open") + " PDF");
```
Siedlerchr: What about the tooltips @tobiasdiez ? I think we should have them

tobiasdiez: Tooltips should indeed be reimplemented. However, I don't think that these old generic ones are really helpful. Instead of "Open PDF" I would prefer a tooltip that really explains what will happen if you click on the icon, i.e "Open filename.pdf".

Not sure if this should be part of this PR or a follow-up.

stefan-kolb: Lets do it as a follow up?

LinusDietz:
Okay, merging this in order to move on. I have created another issue to track the tooltips. https://github.com/JabRef/jabref/issues/3973

Issue #3854, opened 2018-03-15
Fetcher broken without anyone noticing


stefan-kolb:
As we have deactivated the tests in Travis we do not notice that they fail.
Right now, a lot of them are actually broken again. 
I think this is not a good situation if we are not aware of this.
This is a major functionality for us, even if we hide such errors from the user the results will not be as good as expected when the fetchers are broken.
Need to discuss possibilities in the dev call.

- [x] ScienceDirect
- [x] OpenAccessDOI
- [x] Isbnfetcher just returns wrong result because ebbok.de is broken
- [x] viaEbook inputstream seems to be empty via code call. browser result is fine. #3967
- [x] MathSciNet no subscription
- [ ] zbMath only available with subscrption (https://zbmath.org/subscription/)

stefan-kolb:
@tobiasdiez Did you have a chance to test the remaining two fetchers?

tobiasdiez:
MathSciNet works, but I've no longer a subscription to zbMath...

PR #3967, opened 2018-04-20
Fix ebook.de

stefan-kolb:
Refs #3854

Unfortunately not a very clean fix, but it is fine again.
Using the normal InputStream does not return anything. Probably as it is detected as automated crawler?
The interface with InputStream is just not really fitting here.

### DIFF
--- a/src/main/java/org/jabref/logic/importer/IdBasedParserFetcher.java
+++ b/src/main/java/org/jabref/logic/importer/IdBasedParserFetcher.java
@@ -1,20 +1,16 @@
 package org.jabref.logic.importer;
 
-import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.URISyntaxException;
 import java.net.URL;
-import java.nio.charset.StandardCharsets;
 import java.util.List;
 import java.util.Optional;
 
+import org.jabref.logic.net.URLDownload;
 import org.jabref.model.cleanup.Formatter;
 import org.jabref.model.entry.BibEntry;
 
-import com.mashape.unirest.http.HttpResponse;
-import com.mashape.unirest.http.Unirest;
-import com.mashape.unirest.http.exceptions.UnirestException;
 import org.jsoup.helper.StringUtil;
 import org.slf4j.Logger;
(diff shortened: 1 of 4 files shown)


src/main/java/org/jabref/logic/importer/IdBasedParserFetcher.java
```diff
@@ -61,8 +63,9 @@ default void doPostCleanup(BibEntry entry) {
             return Optional.empty();
         }
 
-        try (InputStream stream = new BufferedInputStream(getURLForID(identifier).openStream())) {
-            List<BibEntry> fetchedEntries = getParser().parseEntries(stream);
+        try {
+            HttpResponse<String> response = Unirest.get(getURLForID(identifier).toString()).asString();
```
Siedlerchr: Well, this now has the disdvantage that the Stream does not get closed. That's why try(....) is preferable.

Siedlerchr:
The ISBN fetcher tries ebook.de first and if no result is found, it switches to the chimbori/amazon fetcher. 
Is this still valid?

stefan-kolb:
Yes.

src/main/java/org/jabref/logic/importer/IdBasedParserFetcher.java
Siedlerchr: Ah sorry, got confused by the diff.
Why is this ByteArrayInputStream needed? Is there no way to directly get the response with unirest?

stefan-kolb: Caused by the interface

tobiasdiez: Not sure whats better but we should stick to one: unirest vs our UrlDownloader

Siedlerchr: We could directly use jsoup. We use that in dfiferent cases

stefan-kolb: Jsoup is for a DOM that needs to be parsed. I don't need that here. I just need the body contents.

PR #3976, opened 2018-04-24
Add GenerateKey Button to Entry Editor Toolbar

LinusDietz:
When re-designing the Entry Editor, the Generate BibTeX Button moved to the left on the Required Fields tab. Now I'm re-adding it to the toolbar on the left, because I think it should be accessible from all tabs. See screenshot

screen shot 2018-04-24 at 22 23 44

### DIFF
--- a/src/main/java/org/jabref/gui/entryeditor/EntryEditor.fxml
+++ b/src/main/java/org/jabref/gui/entryeditor/EntryEditor.fxml
@@ -39,6 +39,14 @@
                     <Tooltip text="%Change entry type"/>
                 </tooltip>
             </Button>
+            <Button fx:id="generateKeyButton" styleClass="flatButton,narrow">
+                <graphic>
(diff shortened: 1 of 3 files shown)


src/main/java/org/jabref/gui/entryeditor/EntryEditor.fxml
```diff
@@ -39,6 +39,14 @@
                     <Tooltip text="%Change entry type"/>
                 </tooltip>
             </Button>
+            <Button fx:id="generateKeyButton" styleClass="flatButton,narrow">
```
tobiasdiez: Please set the on-button-click-action via fxml as below (`onAction="#navigateToPreviousEntry"`).

LinusDietz:
I have done the modification as wished by @tobiasdiez.

@halirutan: I guess in the end we still have to develop something driven by our assumptions. If at some point in time we feel that the design is ready for evaluation, we can try to reach out to our users. 
But that would require something that the users can work with.

Issue #3341, opened 2017-10-24
Feature suggestion: Text file export for "find unlinked files"


notuntoward:
`Quality --> Find unlinked files` is handy --  I just used it to discover that an old .bib file has 51 unlinked pdfs.

However, the unlinked files are mostly due to bibkey/pdfname misspellings, and therefore don't need to have a new bib entry made for them -- the only option currently in the interface.

Instead, I'd like to simply export the list of unlinked pdfs to a text file, and then deal with this programmatically, or maybe with some Emacs cut and paste.

I did try `Quality --> Synchronize file links` but that ran for 45 minutes with no result before I gave up on it (there are about 5K entries in this bib file).

jssander:
I'd like to try this as my first issue. Is that ok?

Siedlerchr:
That sounds like a good idea. Just make sure that you name the button so that it is clear what the functionality does

jssander:
Another question: What is the preferred method of writing to a file for JabRef?

stefan-kolb:
When looking at this issue, I ask myself if writing to a file is really the best way to solve this problem or if we need to rework the functionality....

tobiasdiez:
This is now implemented thanks to @jssander.

PR #3979, opened 2018-04-25
[WIP] Add Text File Export for "Find Unlinked Files"

jssander:
This is to add the feature which allows exporting of the list of files found in "Find Unlinked Files" to a text file.

Closes issue #3341

I'm not sure if it is exactly what you are looking for, but I thought I would give it a shot. Hopefully you will be kind though I welcome constructive feedback of course.


src/main/java/org/jabref/gui/FindUnlinkedFilesDialog.java
```diff
@@ -527,6 +534,100 @@ public void stateChanged(ChangeEvent e) {
...
+    private void startExport() {
+
+        if (treeModel == null) {
+            return;
+        }
+        setDefaultCloseOperation(WindowConstants.DO_NOTHING_ON_CLOSE);
+
+        CheckableTreeNode root = (CheckableTreeNode) treeModel.getRoot();
+
+        final List<File> fileList = getFileListFromNode(root);
```
Siedlerchr: It would be preferable if you change the return type from File to Path of the method, as the Path-Class is part of the java.nio and therefore the more modern way.

src/main/java/org/jabref/gui/FindUnlinkedFilesDialog.java
```diff
@@ -527,6 +534,100 @@ public void stateChanged(ChangeEvent e) {
...
+        FileDialogConfiguration fileDialogConfiguration = new FileDialogConfiguration.Builder()
+                .withInitialDirectory(Globals.prefs.get(JabRefPreferences.WORKING_DIRECTORY)).build();
+        DialogService ds = new FXDialogService();
+
+        Optional<Path> exportPath = DefaultTaskExecutor
+                .runInJavaFXThread(() -> ds.showFileSaveDialog(fileDialogConfiguration));
+
+        VerifyingWriter verifyingWriter;
+        try {
+            verifyingWriter = new VerifyingWriter(Files.newOutputStream(exportPath.get()), Charset.defaultCharset());
```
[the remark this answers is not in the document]
Siedlerchr: I think you don't need a Veryfying Writer here, a BufferedWriter should be enough:
And you should use the try with resources , which has the advantage that any stream will be closed automatically, also on error. 
Example: 
```
 try (BufferedWriter writer = 
                 Files.newBufferedWriter(logFile, StandardCharsets.UTF_8, 
                     StandardOpenOption.WRITE)) {
```

src/main/java/org/jabref/gui/FindUnlinkedFilesDialog.java
```diff
@@ -527,6 +534,100 @@ public void stateChanged(ChangeEvent e) {
...
+        } catch (Exception e) {
+            LOGGER.warn("Error while opening file.", e);
+            exportFinishedHandler();
+            return;
+        }
+
+        threadState.set(true);
```
Siedlerchr: What is the purpose of this variable/method?

src/main/java/org/jabref/gui/FindUnlinkedFilesDialog.java
```diff
@@ -527,6 +534,100 @@ public void stateChanged(ChangeEvent e) {
...
+        JabRefExecutorService.INSTANCE.execute(() -> {
+            List<String> errors = new LinkedList<>();
```
Siedlerchr: According to the codacy output, the variable errors is never used. And in 90% the cases a normal ArrayList is sufficient, a LinkedList costs more memory. The benefits of a LinkedList seldom outweigh the memory usage in comparsion to ArrayList

jssander: I'll remove it.

src/main/java/org/jabref/gui/FindUnlinkedFilesDialog.java
jssander: This method is used for the importing which I basically copied and then modified to do the exporting.

src/main/java/org/jabref/gui/FindUnlinkedFilesDialog.java
```diff
@@ -527,6 +535,88 @@ public void stateChanged(ChangeEvent e) {
         });
     }
 
+    /**
+     * This will start the export of all files of all selected nodes in this
```
koppor: "This starts the export"

src/main/java/org/jabref/gui/FindUnlinkedFilesDialog.java
```diff
@@ -527,6 +535,88 @@ public void stateChanged(ChangeEvent e) {
...
+     * dialogs tree view. <br>
+     * <br>
+     * The export itself will run in a seperate thread, whilst this dialog will
+     * be showing a progress bar, until the thread has finished its work. <br>
+     * <br>
+     * When the export has finished, the {@link #exportFinishedHandler(java.util.List)} is
+     * invoked.
+     */
+    private void startExport() {
+
```
koppor: Remove empty line.

src/main/java/org/jabref/gui/FindUnlinkedFilesDialog.java
```diff
@@ -527,6 +535,88 @@ public void stateChanged(ChangeEvent e) {
...
+        if (treeModel == null) {
+            return;
+        }
+        setDefaultCloseOperation(WindowConstants.DO_NOTHING_ON_CLOSE);
+
+        CheckableTreeNode root = (CheckableTreeNode) treeModel.getRoot();
+
+        final List<File> fileList = getFileListFromNode(root);
+
```
koppor: Please remove empty line

src/main/java/org/jabref/gui/FindUnlinkedFilesDialog.java
```diff
@@ -527,6 +535,88 @@ public void stateChanged(ChangeEvent e) {
...
+            }
+        });
+
+        exportFinishedHandler();
+    }
+
+    /**
+     */
+    private void exportFinishedHandler() {
+
```
koppor: Please remove empty line

src/main/java/org/jabref/gui/FindUnlinkedFilesDialog.java
```diff
@@ -527,6 +535,88 @@ public void stateChanged(ChangeEvent e) {
...
+            } catch (IOException e) {
+                LOGGER.warn("IO Error.", e);
+            }
+        });
+
+        exportFinishedHandler();
+    }
+
+    /**
```
koppor: Please remove empty comment.

src/main/java/org/jabref/gui/FindUnlinkedFilesDialog.java
```diff
@@ -527,6 +535,88 @@ public void stateChanged(ChangeEvent e) {
...
+        progressBarImporting.setVisible(false);
+        labelExportingInfo.setVisible(false);
+        buttonExport.setVisible(true);
+        buttonApply.setVisible(true);
+        buttonClose.setVisible(true);
+        disOrEnableDialog(true);
+        setDefaultCloseOperation(WindowConstants.DISPOSE_ON_CLOSE);
+        frame.getCurrentBasePanel().markBaseChanged();
+    }
+
```
koppor: If you are on it - please remove the method comment 😇  - or add description. The functionality is not exactly clear by the method name.

koppor:
@jssander  Please pull before you continue to work on it. I merged master to fix conflicts due to our release today.

src/main/java/org/jabref/gui/FindUnlinkedFilesDialog.java
```diff
@@ -527,6 +535,88 @@ public void stateChanged(ChangeEvent e) {
...
+        progressBarImporting.setValue(0);
+        progressBarImporting.setString("");
+
+        FileDialogConfiguration fileDialogConfiguration = new FileDialogConfiguration.Builder()
+                .withInitialDirectory(Globals.prefs.get(JabRefPreferences.WORKING_DIRECTORY)).build();
+        DialogService ds = new FXDialogService();
+
+        Optional<Path> exportPath = DefaultTaskExecutor
+                .runInJavaFXThread(() -> ds.showFileSaveDialog(fileDialogConfiguration));
+
```
tobiasdiez: Please add a check if `exportPath` is present. If not (i.e. the user clicked cancel), then we should return early.

src/main/java/org/jabref/gui/FindUnlinkedFilesDialog.java
```diff
@@ -527,6 +535,88 @@ public void stateChanged(ChangeEvent e) {
...
+        threadState.set(true);
+        JabRefExecutorService.INSTANCE.execute(() -> {
+            try (BufferedWriter writer =
+                         Files.newBufferedWriter(exportPath.get(), StandardCharsets.UTF_8,
+                                 StandardOpenOption.CREATE)) {
+                int counter = 0;
+                for (File file : fileList) {
+                    writer.write(file.toString() + "\n");
+                    counter++;
+                    progressBarImporting.setValue(counter);
```
tobiasdiez: As you say, writing to the external file should be very quick (its below <1k lines in almost every case and usual not more then say 50). Thus the progressbar would only flicker a bit and I would hence propose to perform the export silently without it.

src/main/java/org/jabref/gui/FindUnlinkedFilesDialog.java
```diff
@@ -609,6 +701,8 @@ private void setupActions() {
          * selected nodes in this dialogs tree view. <br>
          */
         ActionListener actionListenerImportEntrys = e -> startImport();
+        ActionListener actionListenerExportEntrys = e -> startExport();
+        buttonExport.addActionListener(actionListenerExportEntrys);
```
tobiasdiez: I think you can simply inline `actionListenerExportEntrys` (and the other listener before it too, if its not used somewhere else).

jssander:
I made some changes, but there is a conflict now in the changelog. I'm not sure how to resolve it.

koppor:
@jssander Just keep both lines. I did it for you using the GitHub web ui. Just pull and you are fine at your side, too.

jssander:
Great! Thank you so much.

Issue #3239, opened 2017-09-28
IEEE preview does not display month


fylimas:
Currently, the reference are previewed in Jabref without the field month, as shown in the following:
![image](https://user-images.githubusercontent.com/17979090/30964175-d44f2376-a482-11e7-95d0-cece57920f5c.png)

However, the bib should be and is compiled correctly as follows (in which the month is displayed):
![image](https://user-images.githubusercontent.com/17979090/30964158-c508ab1c-a482-11e7-9fbb-d6b7ef6efaf0.png)

This is consistent as the ieee.csl fromcitationstyles.org
![image](https://user-images.githubusercontent.com/17979090/30964145-b6348e1c-a482-11e7-87dd-7951113d274b.png)

That is to say, jabref handles the field month correctly, but the preview with ieee.csl lack the display of field month.
I think this is an error and should be fixed.

Thanks guys for the great jobs.

vatdaell:
I can look into this and try to fix it

tokkot:
I looked into this some, but got stuck. If the month is a literal string "{january}", the preview prints correctly. So, I think there is a problem with how the month field is getting passed to citeproc-java. Which in turn, is maybe a problem with how the BibTeXEntry object is getting built by JabRef. But I'm really not sure if that is a bug in citeproc-java or in JabRef.

Siedlerchr:
@tokkot  Thanks for your analysis! Maybe you could debug through it and see where the problem lies?
I took a quick look and it seems that the conversion is done outside of JabRef in the citeproc java:
Maybe you can have a look here and an at the DateParser in detail
https://github.com/michel-kraemer/citeproc-java/blob/b60592f4e02876fc06dde233e17b8411a6703408/citeproc-java/src/main/java/de/undercouch/citeproc/bibtex/BibTeXConverter.java#L223

koppor:
Coming from the linked issue:

Possibly, our data model needs to be improved. Currently had a "hack" to be able to store

    month = mar

When entering `mar` in the entry editor, JabRef saves it to

    month = {mar}

How to produce the other result? JabRef offers to enter `#mar#` in the entry editor. Then, the field content is saved without `#`.  See also http://help.jabref.org/en/EntryEditor#field-consistency-checking

Siedlerchr:
Is probably a similar issue as #3539, modifying the StringConverter should do the job

DevSiroukane:
Guys,
I just fix both issues Month preview on IEEE style.
[...]

PR #3983, opened 2018-04-26
Fix IEEE preview does not display month 

DevSiroukane:
Fix issue [#3239](https://github.com/JabRef/jabref/issues/3239) on [maintable-beta](https://github.com/JabRef/jabref/tree/maintable-beta) branch

On month change from comboBox on OptionalFieldsTab, Month was not appear on IEEE style.
The issue get fixed.

----

### DIFF
--- a/src/main/java/org/jabref/logic/bibtex/LatexFieldFormatter.java
+++ b/src/main/java/org/jabref/logic/bibtex/LatexFieldFormatter.java
@@ -270,8 +270,8 @@ public class LatexFieldFormatter {
 
     private void writeStringLabel(String text, int startPos, int endPos,
                                   boolean first, boolean last) {
-        putIn((first ? "{" : " # ") + text.substring(startPos, endPos)
-                + (last ? "}" : " # "));
+        putIn((first ? "" : " # ") + text.substring(startPos, endPos)
+                + (last ? "" : " # "));
     }
 
     private void putIn(String s) {
--- a/src/main/java/org/jabref/logic/bibtex/LatexFieldFormatter.java
+++ b/src/main/java/org/jabref/logic/bibtex/LatexFieldFormatter.java
@@ -270,8 +270,8 @@ public class LatexFieldFormatter {
 
     private void writeStringLabel(String text, int startPos, int endPos,
                                   boolean first, boolean last) {
-        putIn((first ? "" : " # ") + text.substring(startPos, endPos)
-                + (last ? "" : " # "));
+        putIn((first ? "{" : " # ") + text.substring(startPos, endPos)
+                + (last ? "}" : " # "));
     }
 
     private void putIn(String s) {
--- a/src/main/java/org/jabref/logic/citationstyle/CSLAdapter.java
+++ b/src/main/java/org/jabref/logic/citationstyle/CSLAdapter.java
@@ -16,7 +16,6 @@ import de.undercouch.citeproc.ItemDataProvider;
 import de.undercouch.citeproc.bibtex.BibTeXConverter;
 import de.undercouch.citeproc.csl.CSLItemData;
 import de.undercouch.citeproc.output.Bibliography;
-
 import org.jbibtex.BibTeXEntry;
 import org.jbibtex.DigitStringValue;
 import org.jbibtex.Key;
--- a/src/main/java/org/jabref/logic/citationstyle/CSLAdapter.java
+++ b/src/main/java/org/jabref/logic/citationstyle/CSLAdapter.java
@@ -84,14 +84,10 @@ public class CSLAdapter {
         /**
          * Converts the {@link BibEntry} into {@link CSLItemData}.
          *
-         * Clone bibEntry to make a save changes,
          * Change month field from JabRefFormat <code>#mon#</> to ShortName <code>mon</code>
          * because CSL does not support JabRefFormat.
          */
         private static CSLItemData bibEntryToCSLItemData(BibEntry bibEntry) {
-
-            bibEntry = (BibEntry) bibEntry.clone();
-
             String citeKey = bibEntry.getCiteKeyOptional().orElse("");
             BibTeXEntry bibTeXEntry = new BibTeXEntry(new Key(bibEntry.getType()), new Key(citeKey));
 
@@ -99,14 +95,15 @@ public class CSLAdapter {
             HTMLChars latexToHtmlConverter = new HTMLChars();
             RemoveNewlinesFormatter removeNewlinesFormatter = new RemoveNewlinesFormatter();
             for (String key : bibEntry.getFieldMap().keySet()) {
-                if (FieldName.MONTH.equals(key)) {
-                    bibEntry.getFieldMap().put(FieldName.MONTH, bibEntry.getMonth().get().getShortName());
-                }
-
                 bibEntry.getField(key)
                         .map(removeNewlinesFormatter::format)
                         .map(latexToHtmlConverter::format)
-                        .ifPresent(value -> bibTeXEntry.addField(new Key(key), new DigitStringValue(value)));
+                        .ifPresent(value -> {
+                            if((FieldName.MONTH.equals(key))){
+                                value = bibEntry.getMonth().get().getShortName();
+                            }
+                            bibTeXEntry.addField(new Key(key), new DigitStringValue(value));
+                        });
             }
             return BIBTEX_CONVERTER.toItemData(bibTeXEntry);
         }
--- a/src/main/java/org/jabref/logic/citationstyle/CSLAdapter.java
+++ b/src/main/java/org/jabref/logic/citationstyle/CSLAdapter.java
@@ -9,13 +9,14 @@ import java.util.Objects;
 import org.jabref.logic.formatter.bibtexfields.RemoveNewlinesFormatter;
 import org.jabref.logic.layout.format.HTMLChars;
 import org.jabref.model.entry.BibEntry;
+import org.jabref.model.entry.FieldName;
 
 import de.undercouch.citeproc.CSL;
 import de.undercouch.citeproc.ItemDataProvider;
 import de.undercouch.citeproc.bibtex.BibTeXConverter;
 import de.undercouch.citeproc.csl.CSLItemData;
 import de.undercouch.citeproc.output.Bibliography;
-import org.jabref.model.entry.FieldName;
+
 import org.jbibtex.BibTeXEntry;
 import org.jbibtex.DigitStringValue;
 import org.jbibtex.Key;
--- a/src/main/java/org/jabref/logic/citationstyle/CSLAdapter.java
+++ b/src/main/java/org/jabref/logic/citationstyle/CSLAdapter.java
@@ -88,10 +88,6 @@ public class CSLAdapter {
 
             // create a copy of bibEntry
             bibEntry = (BibEntry) bibEntry.clone();
-            // change month field from "#mon#" to "mon"
-            if((bibEntry.getFieldMap().get(FieldName.MONTH) != null) && !bibEntry.getFieldMap().get(FieldName.MONTH).isEmpty()){
-                bibEntry.getFieldMap().put(FieldName.MONTH, bibEntry.getMonth().get().getShortName());
-            }
 
             String citeKey = bibEntry.getCiteKeyOptional().orElse("");
             BibTeXEntry bibTeXEntry = new BibTeXEntry(new Key(bibEntry.getType()), new Key(citeKey));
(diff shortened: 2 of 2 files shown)


src/main/java/org/jabref/logic/citationstyle/CSLAdapter.java
```diff
@@ -84,6 +85,14 @@ private void initialize(String newStyle, CitationStyleOutputFormat newFormat) th
          * Converts the {@link BibEntry} into {@link CSLItemData}.
          */
         private static CSLItemData bibEntryToCSLItemData(BibEntry bibEntry) {
+
+            // create a copy of bibEntry
+            bibEntry = (BibEntry) bibEntry.clone();
+            // change month field from "#mon#" to "mon"
+            if((bibEntry.getFieldMap().get(FieldName.MONTH) != null) && !bibEntry.getFieldMap().get(FieldName.MONTH).isEmpty()){
```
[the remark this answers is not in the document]
Siedlerchr: Can you test if this works?

src/main/java/org/jabref/logic/bibtex/LatexFieldFormatter.java
```diff
@@ -270,8 +270,8 @@ private void writeText(String text, int startPos, int endPos) {
 
     private void writeStringLabel(String text, int startPos, int endPos,
                                   boolean first, boolean last) {
-        putIn((first ? "" : " # ") + text.substring(startPos, endPos)
-                + (last ? "" : " # "));
+        putIn((first ? "{" : " # ") + text.substring(startPos, endPos)
```
tobiasdiez: Can you please explain the reason for this change. Thanks!

src/main/java/org/jabref/logic/citationstyle/CSLAdapter.java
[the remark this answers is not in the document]
tobiasdiez: Or add `LatexFieldFormatter fieldFormatter = new LatexFieldFormatter(...)` in front of the for loop and then add `.map(fieldFormatter ::format)`. In this way, we also support custom bibtex strings (and not only the predefined ones for the month field).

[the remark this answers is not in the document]
Siedlerchr: Does not work, as LatexFieldFormatter a) needs preferences object b) has only a two argument format method and does not implment the interface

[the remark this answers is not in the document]
tobiasdiez: ok, but it works "with obvious modifications" 😀

src/main/java/org/jabref/logic/bibtex/LatexFieldFormatter.java
DevSiroukane: This method were create only for "#%s#" forma I guess. The output should be "{%s}" to have this result for month
```
month = {mon}
```
not this
```
month = mon
``` 
[check this on #3239](https://github.com/JabRef/jabref/issues/3239#issuecomment-353082567)

tobiasdiez: Mhh, I'm pretty sure that `month = jan` is the correct form and not `month = {jan}`. Maybe @koppor can clarify this.

src/main/java/org/jabref/logic/citationstyle/CSLAdapter.java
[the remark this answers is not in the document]
DevSiroukane: @Siedlerchr I try ur suggestion but it didnt work.
I can also prefer your solution but instead of making this instruction inside if block 
```
bibEntry.getMonth().map(Month::getShortName).ifPresent(value -> bibTeXEntry.addField(new Key(key), new DigitStringValue(value)));
```
I will keep the instruction that I made.
```
bibEntry.getFieldMap().put(FieldName.MONTH, bibEntry.getMonth().get().getShortName());
```
What do you think?

src/main/java/org/jabref/logic/bibtex/LatexFieldFormatter.java
DevSiroukane: As we know that bibTeX forma should be 
```
key = {value}
```
u can add a debug instruction here [LatexFieldFormatter.java#L166](https://github.com/DevSiroukane/jabref/blob/67387113513ce9529e08b62ffc430d13e24f0b4b/src/main/java/org/jabref/logic/bibtex/LatexFieldFormatter.java#L166)
```
System.out.println(stringBuilder);
```
and see the out put after each change in any field. You will find that only month output is wrong if we miss the {} that I add on 
```
LatexFieldFormatter.java#writeStringLabel(String text, int startPos, int endPos, boolean first, boolean last)
```

Siedlerchr: @DevSiroukane You are confusing it here. If it's  `#month#` it should be displayed without braces in the source. So the entry editor part was already correct. 
Only when it's  `month` -> `{month}`

DevSiroukane: @Siedlerchr are you sure that month should not be between {} on source tab?

because I try to use generate entry using DOI (ex: [10.1109/TSG.2014.2346511](https://ieeexplore.ieee.org/document/6884808/)) and the out put is this on source tab:
```
@Article{Rahbari-Asr2014,
  author    = {Navid Rahbari-Asr and Unnati Ojha and Ziang Zhang and Mo-Yuen Chow},
  title     = {Incremental Welfare Consensus Algorithm for Cooperative Distributed Generation/Demand Response in Smart Grid},
  journal   = {{IEEE} Transactions on Smart Grid},
  year      = {2014},
  volume    = {5},
  number    = {6},
  pages     = {2836--2845},
  month     = {nov},
  doi       = {10.1109/tsg.2014.2346511},
  publisher = {Institute of Electrical and Electronics Engineers ({IEEE})},
}
```

DevSiroukane: @Siedlerchr I think I should let source tab month appear as before
```
month = mon
```
I will remove the change as you said @tobiasdiez

Siedlerchr:
Please have a look at the Travis output, regarding the checkstyle, there is still some error

src/main/java/org/jabref/logic/citationstyle/CSLAdapter.java
```diff
@@ -84,13 +85,22 @@ private void initialize(String newStyle, CitationStyleOutputFormat newFormat) th
          * Converts the {@link BibEntry} into {@link CSLItemData}.
          */
         private static CSLItemData bibEntryToCSLItemData(BibEntry bibEntry) {
+
+            // create a copy of bibEntry
```
koppor: This comment states the same as it is done in the code. So please "Remove Superfluous Comments" [Java by Comparison](http://java.by-comparison.com/).

You can also describe, *why* a clone is generated!

src/main/java/org/jabref/logic/citationstyle/CSLAdapter.java
```diff
@@ -84,13 +85,22 @@ private void initialize(String newStyle, CitationStyleOutputFormat newFormat) th
...
+
             String citeKey = bibEntry.getCiteKeyOptional().orElse("");
             BibTeXEntry bibTeXEntry = new BibTeXEntry(new Key(bibEntry.getType()), new Key(citeKey));
 
             // Not every field is already generated into latex free fields
             HTMLChars latexToHtmlConverter = new HTMLChars();
             RemoveNewlinesFormatter removeNewlinesFormatter = new RemoveNewlinesFormatter();
             for (String key : bibEntry.getFieldMap().keySet()) {
+                if (FieldName.MONTH.equals(key)) {
+                    // change month field value from "#mon#" to "mon"
```
koppor: I don't understand why this hack is necessary at this place.

In case a field value is surrounded by `#`, JabRef stores tha value as follows:

    month = mar

and not

    month = #mar#

See http://help.jabref.org/en/EntryEditor#field-consistency-checking for an explanation. So, I think, either the wrong method is used to get the month or there is something more wrong on JabRef: All other exports should have difficulties.

Siedlerchr: This is actually necessary. The problem is that csl can not handle the JabRef special Syntax with the #. Therefore the value of the month field without the # is needed.

For export usually a latex formatter is applied which takes care of this

src/main/java/org/jabref/logic/citationstyle/CSLAdapter.java
DevSiroukane: Is this a good comment?
```
/**
         * Converts the {@link BibEntry} into {@link CSLItemData}.
         *
         * Clone bibEntry to make a save changes,
         * change month field from JabRefFormat <code>#mon#</> to ShortName <code>mon</code>
         * because CSL does not support JabRefFormat.
         */
```

src/main/java/org/jabref/logic/citationstyle/CSLAdapter.java
DevSiroukane: @koppor As CSL is not part of our code we cannot inject a instructions that handle #mon# forma.
The issue on CSL plugin is still open [Bibtex month like #mar# not parsed #41](https://github.com/michel-kraemer/citeproc-java/issues/41)

src/main/java/org/jabref/logic/citationstyle/CSLAdapter.java
```diff
@@ -82,15 +83,26 @@ private void initialize(String newStyle, CitationStyleOutputFormat newFormat) th
...
             HTMLChars latexToHtmlConverter = new HTMLChars();
             RemoveNewlinesFormatter removeNewlinesFormatter = new RemoveNewlinesFormatter();
             for (String key : bibEntry.getFieldMap().keySet()) {
+                if (FieldName.MONTH.equals(key)) {
+                    bibEntry.getFieldMap().put(FieldName.MONTH, bibEntry.getMonth().get().getShortName());
+                }
+
                 bibEntry.getField(key)
                         .map(removeNewlinesFormatter::format)
                         .map(latexToHtmlConverter::format)
```
tobiasdiez: The code changes above work but are more complicated than they need to. I think an easier solution is to extract the lambda expression in this line (`ifPresent(value -> ...)`) to a new method `setFieldValue(bibTeXEntry, key, value)`. In this new method, you can then simply check `if (FieldName.MONTH.equals(key)) {` and use `Month.parse` to get the correct month value.

DevSiroukane: I did not notice that bibEntry and BibTeXEntry are different object. I think we do not need to clone bibEntry anymore. Instead of create a new method I thought of making test direct on lamda expression. What do you think?

I tested this code and it work.

```
private static CSLItemData bibEntryToCSLItemData(BibEntry bibEntry) {
    String citeKey = bibEntry.getCiteKeyOptional().orElse("");
    BibTeXEntry bibTeXEntry = new BibTeXEntry(new Key(bibEntry.getType()), new Key(citeKey));

    // Not every field is already generated into latex free fields
    HTMLChars latexToHtmlConverter = new HTMLChars();
    RemoveNewlinesFormatter removeNewlinesFormatter = new RemoveNewlinesFormatter();
    for (String key : bibEntry.getFieldMap().keySet()) {
        bibEntry.getField(key)
                .map(removeNewlinesFormatter::format)
                .map(latexToHtmlConverter::format)
                .ifPresent(value -> {
                    if((FieldName.MONTH.equals(key)) && (value.matches("#[a-z]{3}#"))){
                        value = value.substring(1,4);
                    }
                    bibTeXEntry.addField(new Key(key), new DigitStringValue(value));
                });
    }
    return BIBTEX_CONVERTER.toItemData(bibTeXEntry);
}
```

About https://github.com/JabRef/jabref/blob/master/src/test/java/org/jabref/logic/citationstyle/CitationStyleTest.java I have no idea what kind of test I will make because I never make tests for my programs.

Siedlerchr: That is also an idea, but I would not make the regex. If you have the field month, I would really get the shortname from the BibEntry then, so value = key.getMonth().getShortName

DrSiroukane:
```for (String key : bibEntry.getFieldMap().keySet()) {
    bibEntry.getField(key)
            .map(removeNewlinesFormatter::format)
            .map(latexToHtmlConverter::format)
            .ifPresent(value -> {
                if((FieldName.MONTH.equals(key))){
                    value = bibEntry.getMonth().get().getShortName();
                }
                bibTeXEntry.addField(new Key(key), new DigitStringValue(value));
            });
}
```

DevSiroukane:
```
for (String key : bibEntry.getFieldMap().keySet()) {
    bibEntry.getField(key)
            .map(removeNewlinesFormatter::format)
            .map(latexToHtmlConverter::format)
            .ifPresent(value -> {
                if(FieldName.MONTH.equals(key)){
                    value = bibEntry.getMonth().get().getShortName();
                }
                bibTeXEntry.addField(new Key(key), new DigitStringValue(value));
            });
}
```

DevSiroukane: I just post my last request

DrSiroukane:
Thank you, It's nice to work with all of you. I learn a lot.

PR #3985, opened 2018-04-26
Do not run checkstyle as part of the gradle check task

tobiasdiez:
With these changes, the `checkstyle` task is no longer invoked as part of the `build` or `check` task. The `checkstyleMain` task and its variations are kept. 

Reason: The build fails as soon as the checkstyle task fails, possibly hiding more important build or test failures (recent [example](https://travis-ci.org/JabRef/jabref/builds/371621951?utm_source=github_status&utm_medium=notification)). Moreover, the checkstyle task is invoked as a separate build command on travis anyway. 

----

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,8 +17,8 @@ env:
   global:
     - GRADLE_OPTS=-Dorg.gradle.daemon=false
   matrix:
-    - TEST_SUITE=checkstyle
     - TEST_SUITE=check OPTIONS=modernizer
+    - TEST_SUITE=checkstyle
     - TEST_SUITE=fetcherTest
     - TEST_SUITE=databaseTest
     - TEST_SUITE=guiTest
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -17,8 +17,8 @@ env:
   global:
     - GRADLE_OPTS=-Dorg.gradle.daemon=false
   matrix:
-    - TEST_SUITE=checkstyle
     - TEST_SUITE=check OPTIONS=modernizer
+    - TEST_SUITE=checkstyle
```
Siedlerchr: I would definitely execute checkstyle first. As it will always fail asap. So you already know that you have to fix this again

tobiasdiez: They are not guaranteed to run in the specified order (but are started at almost the same time anyway in most cases). Thus it is more a display thing, and the main build is definitely the more important one.

PR #4001, opened 2018-05-01
[WIP] Try to remove bundled jre after uninstall

Siedlerchr:
@halirutan  I added an option to remove the dir "jre" of the bundled installer. Can you please try this out?

----


Siedlerchr:
@halirutan  It would be really nice if you could test this as I don't have a Mac OS on hand

halirutan:
@Siedlerchr I tested in on my Mac and it doesn't seem to work. I created two test files inside the `JabRef.app` folder and after a re-installation of JabRef, where I clicked on *update existing installation*, they are still there

```
patrick@heisenberg:/Applications/JabRef.app$ find . -name "*.txt" -type f
./Contents/java/app/jre/text.txt
./Contents/vmoptions.txt
./Contents/test.txt
```

What now?

LinusDietz:
image
I have normally installed jabref with install4j from the .dmg, but do not have this jre folder.
(I also do not have any problems with freezes.)

halirutan:
@LinusDietz My comment is misleading. The major problem is that some users have old installations, where the JRE folder is indeed inside the app. This JRE seems to have priority and makes that they cannot start JabRef because the Java version is wrong. Therefore, we decided to remove the content completely during a re-installation.

I have created test-folders inside the JabRef.app folder to see if they get removed.

Siedlerchr:
I looked a bit through the help files, and it seems although update bundled jre, the action is not executed when you choose Update application.

LinusDietz:
So I was playing around with Install4J and the Mac installer and could not manage to delete any files or directories via these install4j uninstall directives.. Probably, we missed something... @Siedlerchr's approach in the PR seems to be the right way to go, but somehow it doesn't work.

stefan-kolb:
Is this https://github.com/JabRef/jabref/issues/4103?

stefan-kolb:
Without knowing the fodler where the old JRE resides @LinusDietz cannot know which folder to delete!

halirutan:
I suggest contacting Install4J's support and ask them for help. I don't know either if my test earlier does even resemble the situation that some users have. Our goal is to completely delete the contents of the old directory before filling it with the files of the new version, right? I'm certain the support can help @Siedlerchr and I wouldn't know, what else to do.

Siedlerchr:
I think I found the issue @halirutan Would be nice if you could test this with the new build again.
I changed the installatio dir to conten dir variable which seems to the right one according to the doc for macos bundled stuff

LinusDietz:
This still doesn't work. Since we all tried to make it work I'd say we close this now.

PR #4002, opened 2018-05-01
Remove gtk

Siedlerchr:
----

### DIFF
--- a/src/main/java/org/jabref/gui/preftabs/AppearancePrefsTab.java
+++ b/src/main/java/org/jabref/gui/preftabs/AppearancePrefsTab.java
@@ -20,7 +20,6 @@ import javax.swing.UIManager.LookAndFeelInfo;
 
 import org.jabref.gui.GUIGlobals;
 import org.jabref.logic.l10n.Localization;
-import org.jabref.logic.util.OS;
 import org.jabref.preferences.JabRefPreferences;
 
 import com.jgoodies.forms.builder.DefaultFormBuilder;
@@ -32,6 +31,7 @@ import org.slf4j.LoggerFactory;
 class AppearancePrefsTab extends JPanel implements PrefsTab {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(AppearancePrefsTab.class);
+    private static final String GTK_LF_CLASSNAME = "com.sun.java.swing.plaf.gtk.GTKLookAndFeel";
 
     private final JabRefPreferences prefs;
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/jabref/JabRefGUI.java
```diff
@@ -230,25 +237,20 @@ private void setLookAndFeel() {
...
             } else {
                 lookFeel = Globals.prefs.get(JabRefPreferences.WIN_LOOK_AND_FEEL);
             }
 
-            // FIXME: Open JDK problem
-            if (UIManager.getCrossPlatformLookAndFeelClassName().equals(lookFeel)
-                    && !System.getProperty("java.runtime.name").contains("OpenJDK")) {
-                // try to avoid ending up with the ugly Metal L&F
-                UIManager.setLookAndFeel("javax.swing.plaf.nimbus.NimbusLookAndFeel");
+            if (UIManager.getCrossPlatformLookAndFeelClassName().equals(lookFeel) && !GTK_LF_CLASSNAME.equals(lookFeel)) {
```
tobiasdiez: I don't understand the second test. I thought we want to prevent the `GTK L&F` by all means....

tobiasdiez: ok now I don't understand the complete if statement (and think we can simply delete it).

cases:
- system: gtk -> `lookFeel = Nimbus`, `if = false`
- system: win -> `lookFeel = win`, `if = true` (why do we set the l&f to nimbus?!)
- system: win, WIN_LOOK_AND_FEEL: win -> `lookFeel = win`, `if = true` (again wrong)
- system: gtk, WIN_LOOK_AND_FEEL: gtk -> `lookFeel = gtk`, `if = false` (but we actually wanted to prevent setting it to gtk?!)
- system: win, WIN_LOOK_AND_FEEL: gtk -> `lookFeel = gtk`, `if = false` (again, we don't want gtk)

If I understand you correctly, WIN_LOOK_AND_FEEL = gtk can never happen because of the migration, right? In this case we can ignore the last two cases, but even in the first three cases the behavior is odd.

Siedlerchr: I simplified the condition. Now the cases should be more clear.

src/main/java/org/jabref/JabRefGUI.java
```diff
@@ -239,17 +239,17 @@ private void setLookAndFeel() {
...
+                    LOGGER.warn("There seems to be problems with GTK Look&Feel. Using Nimbus L&F instead. Change to another L&F with caution.");
                 } else {
                     lookFeel = systemLookFeel;
                 }
             } else {
                 lookFeel = Globals.prefs.get(JabRefPreferences.WIN_LOOK_AND_FEEL);
             }
 
-            if (UIManager.getCrossPlatformLookAndFeelClassName().equals(lookFeel) && !GTK_LF_CLASSNAME.equals(lookFeel)) {
+            //Prevent metal l&f
```
Siedlerchr: use Default L&F:
1. System L&F = gtk => Nimbus
2. System L&F = win => win
3. System L&F = metal look and feel 

Check cross platform 
1. Metal look and feel => Nimbus

PR #4015, opened 2018-05-06
Fix custom exporter not displayed in file chooser

### LINKED ISSUE #4013: custom export filter cannot be used

JabRef version 4.2+ on Fedora 27

Steps to reproduce:

1. Add a custom export filter https://github.com/alick/jabref-export/tree/master/listrefs with the name 'test'
2. Export a selected bib entry
3. In the save file dialog, look for the custom filter entry in the popdown filter list.
4. There are two 'HTML list file (*.html)' entries in the popdown list. Neither gives the custom result. No entry for 'test'.

screenshot from 2018-05-05 21-51-15
screenshot from 2018-05-05 22-00-05

Notes:

- I think I might have filename clash with shipped filter. But after I changed my custom one to a new name, the result is still the same.
- With JabRef 4.1, the 'test' filter appears in the popdown list, but with a wrong suffix (DEFAULT) although it is .html in the 'Manage custom export' dialog.

Siedlerchr:
Fixes  #4013

Fixed several issues with the Custom Exporter.
- Filename check for layout files was wrong 
- Commandline name for custom exports was the filename and not the name, opposed to what our help says 
- Custom exports could not be added due to our hardcoded enum
- Expanded the enum with [Mixins](http://blog.pengyifan.com/how-to-extend-enum-in-java/)

----


src/main/java/org/jabref/logic/util/BasicFileType.java
```diff
@@ -0,0 +1,161 @@
...
+        allfileTypes.add(fileType);
+        return fileType;
+
+    }
+
+    public static void addnewFileType(FileType fileType) {
+        allfileTypes.add(fileType);
+    }
+
+    public static FileType parse(String description) {
```
tobiasdiez: It appears to me that the `parse` method is no longer needed and I suspect that hence the `allFileTypes` list can be removed. Right?

Siedlerchr: Yes missed that.

tobiasdiez:
ok, thanks for the explanation. I thought a bit more about this problem and came to the conclusion that the cleanest solution is probably to split the description from the "file type". If the later is only a list of extensions (xml, bib, txt, ...) and the importer/exporter itself contains the description, then you can easily create a `ExtensionFilter` for an importer/exporter (and don't need to pay special attention to custom exporters). What do you think?

Siedlerchr:
This would only work if the custom export follows the standard file types. If I want my exporter to have filetype xyz then this would not be possible.
What I did is to provide a set of standard file types and their descriptions which can be extended to support custom one. The standard file types are also used for the importer 
This is the same problem the jdk devs

tobiasdiez:
ok, good point. Do you want to improve the code around `ExternalFileType` as part of this PR or is this a follow-up?

I would still separate the file extension from the description text.

Siedlerchr:
I would rather do this in a follow up on the maintable directly

CHANGELOG.md
```diff
@@ -16,6 +16,10 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - We changed the implementation of the `[shorttitle]` key pattern. It now removes small words like `a`, `an`, `on`, `the` etc. Refer to the help page for a complete overview. [Feature request in the forum](http://discourse.jabref.org/t/jabref-differences-in-shorttitle-between-versions-3-8-1-and-4-not-discounting-the-a-an-of-in-titles/1147)
 
 ### Fixed
+We fixed an issue where the export to clipboard functionality could not be invoked [#3994](https://github.com/JabRef/jabref/issues/3994)
+We fixed an issue with the migration of invalid Look and Feels [#3995, comment](https://github.com/JabRef/jabref/issues/3995#issuecomment-385649448)
```
tobiasdiez: This looks like a duplication (probably due to a problem while merging).

src/main/java/org/jabref/gui/util/FileDialogConfiguration.java
```diff
@@ -101,7 +102,7 @@ public Builder withInitialDirectory(String directory) {
             return this;
         }
 
-        public Builder withDefaultExtension(FileType fileType) {
+        public Builder withDefaultExtension(BasicFileType fileType) {
```
tobiasdiez: I think this method should still accept `FileType`

src/main/java/org/jabref/gui/util/FileFilterConverter.java
```diff
@@ -24,11 +25,11 @@ private FileFilterConverter() {
     }
 
     private static FileChooser.ExtensionFilter convertImporter(String description, Collection<Importer> formats) {
-        List<FileType> fileTypes = formats.stream().map(Importer::getFileType).collect(Collectors.toList());
+        List<BasicFileType> fileTypes = formats.stream().map(Importer::getFileType).collect(Collectors.toList());
         return toExtensionFilter(description, fileTypes);
     }
 
-    private static FileChooser.ExtensionFilter toExtensionFilter(String description, List<FileType> fileTypes) {
+    private static FileChooser.ExtensionFilter toExtensionFilter(String description, List<BasicFileType> fileTypes) {
```
tobiasdiez: Here also `FileType`

src/main/java/org/jabref/logic/exporter/TemplateExporter.java
```diff
@@ -117,7 +118,7 @@ public TemplateExporter(String consoleName, String lfFileName, String directory,
      * @param savePreferences Preferences for saving
      * @param deleteBlankLines If blank lines should be remove (default: false)
      */
-    public TemplateExporter(String consoleName, String lfFileName, String directory, FileType extension, LayoutFormatterPreferences layoutPreferences, SavePreferences savePreferences, boolean deleteBlankLines) {
+    public TemplateExporter(String consoleName, String lfFileName, String directory, BasicFileType extension, LayoutFormatterPreferences layoutPreferences, SavePreferences savePreferences, boolean deleteBlankLines) {
```
tobiasdiez: here too

src/main/java/org/jabref/logic/importer/Importer.java
```diff
@@ -134,9 +134,9 @@ public static BufferedReader getReader(Path filePath, Charset encoding)
 
     /**
      * Returns the type of files that this importer can read
-     * @return {@link FileType} corresponding to the importer
+     * @return {@link BasicFileType} corresponding to the importer
      */
-    public abstract FileType getFileType();
+    public abstract BasicFileType getFileType();
```
tobiasdiez: here as well...

src/main/java/org/jabref/logic/util/BasicFileType.java
```diff
@@ -0,0 +1,161 @@
...
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import java.util.StringJoiner;
+import java.util.stream.Collectors;
+
+import org.jabref.logic.l10n.Localization;
+
+/**
+ * This enum contains a list of file types used in open and save dialogs.
```
tobiasdiez: We shouldn't reference dialogs here (as we are in the logic package).

src/main/java/org/jabref/logic/util/BasicFileType.java
```diff
@@ -0,0 +1,161 @@
...
+    ODS(Localization.lang("%0 file", Localization.lang("OpenDocument spreadsheet")), "ods"),
+    SXC(Localization.lang("%0 file", "OpenOffice/LibreOffice Calc"), "sxc"),
+    HTML(Localization.lang("%0 file", "HTML"), "html"),
+    RTF(Localization.lang("%0 file", "RTF"), "rtf"),
+    RDF(Localization.lang("%0 file", "RDF"), "rdf"),
+    CSV(Localization.lang("%0 file", "CSV"), "csv"),
+    DEFAULT(Localization.lang("%0 file", "DEFAULT"), "default");
+
+    private static Set<FileType> allfileTypes = new HashSet<>(EnumSet.allOf(BasicFileType.class));
+    private final String[] extensions;
```
tobiasdiez: It probably makes the code below a bit easier to read if you simply store the extensions as a List. Moreover, the comment below can probably be removed.

src/main/java/org/jabref/logic/util/FileType.java
```diff
@@ -1,105 +1,18 @@
...
-            sj.add("*." + ext);
-        }
-        return sj.toString();
-    }
+    String getDescription();
 
-    public String getFirstExtensionWithDot() {
-        return "." + extensions[0].trim();
-    }
+    String getFirstExtensionWithDot();
```
tobiasdiez: For these two method you can provide a default implementation.

src/main/java/org/jabref/logic/util/BasicFileType.java
```diff
@@ -0,0 +1,161 @@
...
+        return getFirstExtensionWithDot(extensions);
+    }
+
+    @Override
+    public List<String> getExtensionsWithDot() {
+        return getExtensionsWithDot(extensions);
+    }
+
+    public static FileType addnewFileType(String description, String... extensions) {
+        FileType fileType = new FileType() {
```
tobiasdiez: It is probably better to extract this anonymous class to a separate class since it looks to be the standard implementation that can be reused in custom importers

src/main/java/org/jabref/logic/util/FileType.java
```diff
@@ -1,105 +1,18 @@
...
+public interface FileType {
 
-    public String getDescription() {
-        StringJoiner sj = new StringJoiner(", ", description + " (", ")");
-        for (String ext : extensions) {
-            sj.add("*." + ext);
-        }
-        return sj.toString();
-    }
+    String getDescription();
```
tobiasdiez: Also on a second look, I'm strongly in favor of removing the `description` from the file type. It is actually the description of the exporter/importer that we should show in the dialog...if you want to reuse the strings you can extract the descriptions in `BasicFileType` to a new class, but its also ok if you just repeat the description in the importer and exporter.

src/main/java/org/jabref/logic/util/BasicFileType.java
Siedlerchr: I have created a factory method for this. So the caller does not need to create it's own file type objects 
 public static FileType addnewFileType(String description, String... extensions) {

src/main/java/org/jabref/logic/util/FileType.java
Siedlerchr: This would not solve the underlying problem and in fact would result in the same problem I intend to fix here. 
As we have multiple file types with the same extension e.g. html. We already have Html Table, Html LIst etc. Only with the added description, the file type is unique

tobiasdiez: I think your example is a good illustration why the current mix of file extensions and description is misleading. The file type is `html` in both examples, but the exporter used is different (tables vs list). You don't get these kind of problems if the `FileType` is only an abstraction of the extension and the description comes from the exporter...

Siedlerchr: At least we need the description in the extension filter. Otherwise we can't map it. I will think about this. I will look and see how this would affect the importer

Siedlerchr:
@tobiasdiez  I now removed the exporter stuff from the FileTypes. So currently it's only used for the importers. The key problem with the importers is, that we have an "All available import formats" plus all the extensions. In this case I see no way to filter the importer without having a description in the extension format.

tobiasdiez:
A shot in the blue: Instead of 
https://github.com/JabRef/jabref/blob/e32bdc8ef5731d41252eb53a50cefdf58ea2657d/src/main/java/org/jabref/gui/util/FileFilterConverter.java#L26-L29
try
````java
 private static FileChooser.ExtensionFilter convertImporter(Importer importer) { 
     return toExtensionFilter(importer.getDisplayName(), Collections.singletonList(importer.getFileType)); 
}
````
and at the same time change
https://github.com/JabRef/jabref/blob/e32bdc8ef5731d41252eb53a50cefdf58ea2657d/src/main/java/org/jabref/gui/util/FileFilterConverter.java#L39-L41
to
````java
 public static Optional<Importer> getImporter(FileChooser.ExtensionFilter extensionFilter, Collection<Importer> importers) { 
     return importers.stream().filter(importer -> importer.getDisplayName().equals(extensionFilter.getDescription())).findFirst(); 
} 
````

If `getImporter` returns an empty optional, you know that there is no specific importer, i.e. the user selected "any file" or "any import format".

**Comment by Siedlerchr** [comment]
@tobiasdiez  I noticed that it was already possible. I just needed to create the Map with Description/Name  and FileType like I did for export. It is now analogous to export. I removed the not needed FileTypes.
However, there are still some import formats left which have more than one extension. However, I think it's not worth changing them, too.  Would make things a lot more complicated. 
We already have a good base for future expansions and for using this in combination with ExternalFileTypes.

File: src/main/java/org/jabref/gui/exporter/ExportAction.java [hunk 12]
```diff
@@ -152,7 +155,7 @@ public void update() {
     }
 
     private static FileDialogConfiguration createExportFileChooser(ExporterFactory exportFactory, String currentDir) {
-        List<FileType> fileTypes = exportFactory.getExporters().stream().map(Exporter::getFileType).collect(Collectors.toList());
+        Map<String, FileType> fileTypes = exportFactory.getExporters().stream().collect(Collectors.toMap(Exporter::getDisplayName, Exporter::getFileType));
```
tobiasdiez: The logic to convert an exporter (and importer below) to a FileExtensionFilter should reside in `FileFilterConverter`.

Siedlerchr: No, in this case it would not have any additional benefit. It would still return a map. The map is internally handled in the file dialog configuration and there handled.

Siedlerchr:
Yes definitely.

tobiasdiez:
I've refactored a bit more (it was easier to hijack this PR than describe the refactorings in detail -sorry). If you are happy with these changes, feel free to merge; otherwise just revert my commits...

Siedlerchr:
Looks good so far.  If someone can merge this with co authoring ( i got this wrong last time)

PR #3697, opened 2018-02-06
Allow spaces in DOIs

johannes-manner:
This fixes https://github.com/koppor/jabref/issues/283

----

- [ ] Change in CHANGELOG.md described
- [ ] Tests created for changes
- [ ] Screenshots added (for bigger UI changes)
- [x] Manually tested changed features in running JabRef
- [ ] Check documentation status (Issue created for outdated help page at [help.jabref.org](https://github.com/JabRef/help.jabref.org/issues)?)

PR #4016, opened 2018-05-06
Fix titlefetcherwhitespace

Siedlerchr:
----

### DIFF
--- a/src/main/java/org/jabref/gui/EntryTypeDialog.java
+++ b/src/main/java/org/jabref/gui/EntryTypeDialog.java
@@ -34,6 +34,7 @@ import org.jabref.logic.bibtexkeypattern.BibtexKeyGenerator;
 import org.jabref.logic.importer.FetcherException;
 import org.jabref.logic.importer.IdBasedFetcher;
 import org.jabref.logic.importer.WebFetchers;
+import org.jabref.logic.importer.fetcher.TitleFetcher;
 import org.jabref.logic.l10n.Localization;
 import org.jabref.model.EntryTypes;
 import org.jabref.model.database.BibDatabaseMode;
@@ -73,6 +74,7 @@ public class EntryTypeDialog extends JabRefDialog implements ActionListener {
         setTitle(Localization.lang("Select entry type"));
 
         addWindowListener(new WindowAdapter() {
+
             @Override
             public void windowClosing(WindowEvent e) {
                 cancelAction.actionPerformed(null);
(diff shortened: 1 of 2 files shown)


src/main/java/org/jabref/gui/EntryTypeDialog.java
```diff
@@ -288,10 +291,16 @@ public void actionPerformed(ActionEvent e) {
...
 
-            Globals.prefs.put(JabRefPreferences.ID_ENTRY_GENERATOR,String.valueOf(comboBox.getSelectedItem()));
-            searchID = idTextField.getText().trim();
-            searchID = searchID.replaceAll(" ", "");
+            Globals.prefs.put(JabRefPreferences.ID_ENTRY_GENERATOR, String.valueOf(comboBox.getSelectedItem()));
+
             fetcher = WebFetchers.getIdBasedFetchers(Globals.prefs.getImportFormatPreferences()).get(comboBox.getSelectedIndex());
+
+            searchID = idTextField.getText();
+            if (!(fetcher instanceof TitleFetcher)) {
```
tobiasdiez: Why do we remove whitespace for the other fetcher? Shouldn't each fetcher decide how to proceed the input? I would just always trim it at this point...

Siedlerchr: All other fetchers have a single alphanumeric id or an url so, trimming in general makes sense.

koppor: Copy and paste DOIs from PDFs or homepages (e.g., Springer) leads to identifiers. With the fix https://github.com/JabRef/jabref/pull/3697, there is a speedup in the user workflow. One can just copy something from a PDF/homepage and paste it into JabRef and click "Generate".

From https://github.com/koppor/jabref/issues/283:

`http s://doi.org/10.1007/s13222-018-0273-1` and `DOI: 10.1109/MS.2013.97`.

Siedlerchr: I think it really makes sense to add this argument parsing/trimming to the fetcher itself as @tobiasdiez  proposes. Maybe a new method parse input?

Siedlerchr:
I now moved trimming down to the concrete fetcher classes and added tests

src/main/java/org/jabref/logic/importer/fetcher/DoiFetcher.java
```diff
@@ -44,7 +44,9 @@ public HelpFile getHelpPage() {
 
     @Override
     public Optional<BibEntry> performSearchById(String identifier) throws FetcherException {
-        Optional<DOI> doi = DOI.parse(identifier);
+        String cleanedIdentifier = identifier.trim();
```
tobiasdiez: Can you please move this clean-up to the `DOI.parse` method.

Issue #3419, opened 2017-11-09
Feature Request: OpenSSL for SQL


mwessley:
It would be really nice to include SSL support for the shared database function.
Up to now for SSL encrypted communication, tunneling is necessary.

stefan-kolb:
Would be useful, as many remote DBs are only accessible via SSL.

Siedlerchr:
Refs #4040

Siedlerchr:
Hi,
starting with the current development version, https://builds.jabref.org/master/
 JabRef now supports the usage of SSL for both MySQL and PostgreSQL. 
In a nutshell, you have to convert your certificates for the usage with Java and add them to a (custom) keystore. This file is then specified in JabRef.


I already create a PR for the documentation update for the upcoming 5.0.
Therefore you have to use this link to see the changes.
 Feel free to ask any questions.
https://github.com/JabRef/help.jabref.org/blob/74c3bf244017ba17f7533cd85044a56c9361d1c8/en/SQLDatabase.md

PR #4040, opened 2018-05-20
Convert SharedDatabaseConnect Dialog to Javafx

Siedlerchr:
Connection works so far.  No more freezes regarding addTabs. in basePanel
Somehow the dialog looks odd, @tobiasdiez  Any idea what is missing? The checkboxes are not visible correct. I just see that it looks really odd in he maintable-beta branch, too. e..g. int he Event Log

Edit// The conection and the synchronization seems to be working.  

Fixes: #3419   PostgresSQL database connection now supports SSL

TODO for documentaition:
Create and configure Server certs
https://www.postgresql.org/docs/current/static/ssl-tcp.html

Convert to keystore format
https://jdbc.postgresql.org/documentation/head/ssl-client.html

grafik

Currently I only tested it with PostgresSQL, but I see that mySQL should also work
https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-reference-using-ssl.html

----

TODO: 
- check conflict merge dialog


tobiasdiez:
The display problem probably results from us playing around with different styles and changing defaults. Since there is no checkbox in the main view, it was not noticed that some of the input controls look strange. I'll have a look at it later...

Siedlerchr:
The database connection works so far. Tested with PostgresSQL. I reworked the dialog and fixed an error in the listener which led to freezes. 
The MergeEntryDialog is not yet converted, this will be done in a future PR.

src/main/java/org/jabref/gui/shared/SharedDatabaseLoginDialog.fxml
```diff
@@ -0,0 +1,74 @@
...
+                      <RowConstraints minHeight="10.0" prefHeight="30.0" vgrow="SOMETIMES" />
+                        <RowConstraints minHeight="10.0" prefHeight="30.0" vgrow="SOMETIMES" />
+                        <RowConstraints minHeight="10.0" prefHeight="30.0" vgrow="SOMETIMES" />
+                        <RowConstraints minHeight="10.0" prefHeight="30.0" vgrow="SOMETIMES" />
+                    </rowConstraints>
+                     <children>
+                        <Label text="%Database Type:" />
+                        <ComboBox fx:id="cmbDbType" prefHeight="25.0" prefWidth="606.0" GridPane.columnIndex="1" GridPane.columnSpan="2" />
+                        <Label text="%Host/Port:" GridPane.rowIndex="1" />
+                        <TextField fx:id="tbHost" GridPane.columnIndex="1" GridPane.rowIndex="1" />
```
tobiasdiez: I don't have a strong opinion on this, but this Hungarian notation is against the usual naming convention. Probably, `host` or `hostTextField` is preferable. @koppor what's your opinion about this?

src/main/java/org/jabref/gui/shared/SharedDatabaseLoginDialogView.java
```diff
@@ -0,0 +1,320 @@
...
+    private final JabRefFrame frame;
+
+    private DBMSConnectionProperties connectionProperties;
+    private SharedDatabaseLoginDialogViewModel viewModel;
+
+    public SharedDatabaseLoginDialogView(JabRefFrame frame) {
+        this.frame = frame;
+        this.setTitle(Localization.lang("Connect to shared database"));
+
+        this.getDialogPane().getButtonTypes().addAll(connect, ButtonType.CANCEL);
```
tobiasdiez: It is preferable to set the buttons in the fxml
https://github.com/JabRef/jabref/blob/390c38740997201c6ab6c5e66bf53c2991507a48/src/main/java/org/jabref/gui/help/AboutDialog.fxml#L99-L100
and then add the action in the code-behind
https://github.com/JabRef/jabref/blob/390c38740997201c6ab6c5e66bf53c2991507a48/src/main/java/org/jabref/gui/help/AboutDialogView.java#L36

src/main/java/org/jabref/gui/shared/SharedDatabaseLoginDialogView.java
```diff
@@ -0,0 +1,320 @@
...
+        this.btnConnect = (Button) this.getDialogPane().lookupButton(connect);
+        btnConnect.setOnAction(this::openDatabase);
+
+        ViewLoader.view(this)
+                  .load()
+                  .setAsContent(this.getDialogPane());
+
+    }
+
+    private void openDatabase(ActionEvent e) {
```
tobiasdiez: This method and essentially everything below should be moved to the `ViewModel`. The view should contain no logic whatsoever and just binds the fxml file to the properties in the view model.

src/main/java/org/jabref/gui/shared/SharedDatabaseLoginDialogView.java
```diff
@@ -0,0 +1,320 @@
...
+
+    @FXML
+    private void initialize() {
+
+        viewModel = new SharedDatabaseLoginDialogViewModel();
+        cmbDbType.itemsProperty().bind(viewModel.dbmstypeProperty());
+        cmbDbType.getSelectionModel().select(0);
+        viewModel.selectedDbmstypeProperty().bind(cmbDbType.getSelectionModel().selectedItemProperty());
+
+        viewModel.selectedDbmstypeProperty().addListener((observable, oldValue, newValue) -> {
```
tobiasdiez: `EasyBind.subscribe` is usually easier to read. Moreover, this logic should also be moved to the view model.

src/main/java/org/jabref/gui/shared/SharedDatabaseLoginDialogView.java
```diff
@@ -0,0 +1,320 @@
...
+
+        applyPreferences();
+    }
+
+    @FXML
+    void openFileDialog(ActionEvent event) {
+        FileDialogConfiguration fileDialogConfiguration = new FileDialogConfiguration.Builder()
+                                                                                               .addExtensionFilter(FileType.BIBTEX_DB)
+                                                                                               .withDefaultExtension(FileType.BIBTEX_DB)
+                                                                                               .withInitialDirectory(Globals.prefs.get(JabRefPreferences.WORKING_DIRECTORY))
```
tobiasdiez: Please extract the `Globals.prefs` dependency. The new JavaFX stuff should no longer depend on `Globals`.

src/main/java/org/jabref/gui/shared/SharedDatabaseLoginDialogView.java
```diff
@@ -0,0 +1,320 @@
...
+
+            if (!tbFolder.getText().isEmpty()) {
+                try {
+                    new SaveDatabaseAction(panel, Paths.get(tbFolder.getText())).runCommand();
+                } catch (Throwable e) {
+                    LOGGER.error("Error while saving the database", e);
+                }
+            }
+
+            return; // setLoadingConnectButtonText(false) should not be reached regularly.
```
tobiasdiez: what is the comment trying to tell us?

src/main/java/org/jabref/gui/shared/SharedDatabaseLoginDialogView.java
```diff
@@ -0,0 +1,320 @@
...
+        } catch (SQLException | InvalidDBMSConnectionPropertiesException exception) {
+
+            frame.getDialogService().showErrorDialogAndWait(Localization.lang("Connection error"), exception);
+
+        } catch (DatabaseNotSupportedException exception) {
+            ButtonType openHelp = new ButtonType("Open Help", ButtonData.OTHER);
+
+            Optional<ButtonType> result = dialogService.showCustomButtonDialogAndWait(AlertType.INFORMATION,
```
tobiasdiez: Information or Warning?

src/main/java/org/jabref/gui/shared/SharedDatabaseLoginDialogView.java
```diff
@@ -0,0 +1,320 @@
...
+            result.filter(btn -> btn.equals(openHelp)).ifPresent(btn -> HelpAction.openHelpPage(HelpFile.SQL_DATABASE_MIGRATION));
+            result.filter(btn -> btn == ButtonType.OK).ifPresent(btn -> openSharedDatabase());
+
+        }
+
+        setLoadingConnectButtonText(false);
+    }
+
+    private void setLoadingConnectButtonText(boolean isLoading) {
+        btnConnect.setDisable(isLoading);
```
tobiasdiez: The `viewModel` should have a boolean property `connectionInProgress` so that the functionality of this method can be realized by binding.

src/main/java/org/jabref/gui/shared/SharedDatabaseLoginDialogView.java
```diff
@@ -0,0 +1,320 @@
...
+    private final SharedDatabasePreferences prefs = new SharedDatabasePreferences();
+
+    private final ButtonType connect = new ButtonType("Connect", ButtonData.YES);
+    private final Button btnConnect;
+    private final JabRefFrame frame;
+
+    private DBMSConnectionProperties connectionProperties;
+    private SharedDatabaseLoginDialogViewModel viewModel;
+
+    public SharedDatabaseLoginDialogView(JabRefFrame frame) {
```
tobiasdiez: I really don't like that we need a frame here. Can you please investigate if it can be replaced by the `StateManager` (maybe after extending the latter).

src/main/java/org/jabref/gui/shared/SharedDatabaseLoginDialogView.java
```diff
@@ -0,0 +1,320 @@
...
+            }
+        } else {
+            prefs.clearPassword(); // for the case that the password is already set
+        }
+
+        prefs.setRememberPassword(chkRememberPassword.isSelected());
+    }
+
+    private void checkFields() throws JabRefException {
+        if (isEmptyField(tbHost)) {
```
tobiasdiez: Validation should be done using the validation framework so that the user get immediate feedback (have a look at the usage of ControlsFxVisualizer).

src/main/java/org/jabref/gui/shared/SharedDatabaseUIManager.java
```diff
@@ -37,32 +38,40 @@
...
+                                                                                  reconnect,
+                                                                                  workOffline,
+                                                                                  closeLibrary);
+
+        if (answer.isPresent()) {
+            if (answer.get().equals(reconnect)) {
+                jabRefFrame.closeCurrentTab();
+                new SharedDatabaseLoginDialogView(jabRefFrame).showAndWait();
+
+            }
```
tobiasdiez: `else if` (and remove empty line above)

src/main/java/org/jabref/gui/shared/SharedDatabaseUIManager.java
```diff
@@ -86,12 +95,12 @@ public void listen(SharedEntryNotPresentEvent event) {
...
 
         if (Objects.nonNull(entryEditor) && (entryEditor.getEntry() == event.getBibEntry())) {
-            JOptionPane.showMessageDialog(null,
-                    Localization.lang("The BibEntry you currently work on has been deleted on the shared side.")
-                            + "\n" + Localization.lang("You can restore the entry using the \"Undo\" operation."),
-                    Localization.lang("Shared entry is no longer present"), JOptionPane.INFORMATION_MESSAGE);
 
-            SwingUtilities.invokeLater(() -> panel.closeBottomPane());
+            dialogService.showInformationDialogAndWait(Localization.lang("Shared entry is no longer present"),
+                                                       Localization.lang("The BibEntry you currently work on has been deleted on the shared side.")
```
tobiasdiez: `BibEntry` is for our internal usage and `entry` should be used for dialogs.

src/main/java/org/jabref/gui/shared/SharedDatabaseLoginDialogView.java
Siedlerchr: Unfortunately this results in an NPE which I don't understand.  The lookUpButton returns null.

src/main/java/org/jabref/gui/shared/SharedDatabaseLoginDialog.fxml
```diff
@@ -0,0 +1,81 @@
...
+               </bottom>
+            </BorderPane>
+         </center>
+      </BorderPane>
+   </content>
+   <header>
+   </header>
+   <buttonTypes>
+      <ButtonType fx:constant="CLOSE" />
+      <ButtonType fx:id="connectButton" text="Connect" />
```
Siedlerchr: Here I add the button

src/main/java/org/jabref/gui/shared/SharedDatabaseLoginDialogView.java
```diff
@@ -0,0 +1,320 @@
...
+    @FXML private TextField tbHost;
+    @FXML private TextField tbDb;
+    @FXML private TextField tbPort;
+    @FXML private TextField tbUser;
+    @FXML private PasswordField tbPwd;
+    @FXML private CheckBox chkRememberPassword;
+    @FXML private TextField tbFolder;
+    @FXML private Button btnBrowse;
+    @FXML private CheckBox chkAutosave;
+    @FXML private ButtonType connectButton;
```
Siedlerchr: The buttonType is initalized here

src/main/java/org/jabref/gui/shared/SharedDatabaseLoginDialogView.java
```diff
@@ -0,0 +1,320 @@
...
+
+    public SharedDatabaseLoginDialogView(JabRefFrame frame) {
+        this.frame = frame;
+        this.setTitle(Localization.lang("Connect to shared database"));
+
+        ViewLoader.view(this)
+                  .load()
+                  .setAsContent(this.getDialogPane());
+
+        ControlHelper.setAction(connectButton, this.getDialogPane(), event -> openDatabase());
```
Siedlerchr: In this method the lookup method from the dialogPane returns null. I don't understand why.

Siedlerchr: Somehow the buttons are in the content property of the dialog pane. Does not make any sense to me

src/main/java/org/jabref/gui/shared/SharedDatabaseLoginDialogView.java
tobiasdiez: Did you also tried it without the surrounding `buttontypes` tag?

Siedlerchr:
Yes I tried both with and without the tags. Same behavior. I don't
understand this.

Tobias Diez <notifications@github.com> schrieb am Fr., 25. Mai 2018, 15:39:

src/main/java/org/jabref/gui/shared/SharedDatabaseLoginDialogView.java
```diff
@@ -0,0 +1,320 @@
...
+    private DBMSConnectionProperties connectionProperties;
+    private SharedDatabaseLoginDialogViewModel viewModel;
+
+    public SharedDatabaseLoginDialogView(JabRefFrame frame) {
+        this.frame = frame;
+        this.setTitle(Localization.lang("Connect to shared database"));
+
+        ViewLoader.view(this)
+                  .load()
+                  .setAsContent(this.getDialogPane());
```
tobiasdiez: Should be `.setAsDialogPane(this);`

Siedlerchr:
Now moved all stuff and implemented the validationto the viewmodel and will tackle the SharedUiManager stuff next

src/main/java/org/jabref/gui/shared/SharedDatabaseLoginDialogView.java
```diff
@@ -0,0 +1,120 @@
...
+        tbPwd.textProperty().bindBidirectional(viewModel.passwordProperty());
+        tbPort.textProperty().bindBidirectional(viewModel.portProperty());
+        cmbDbType.valueProperty().bindBidirectional(viewModel.selectedDbmstypeProperty());
+
+        tbFolder.textProperty().bindBidirectional(viewModel.folderProperty());
+        btnBrowse.disableProperty().bind(chkAutosave.selectedProperty().not());
+        tbFolder.disableProperty().bind(chkAutosave.selectedProperty().not());
+        chkAutosave.selectedProperty().bindBidirectional(viewModel.autosaveProperty());
+
+        Platform.runLater(() -> {
```
tobiasdiez: Please add a comment that this is a workaround and add a link to the issue in controlsfx

src/main/java/org/jabref/gui/shared/SharedDatabaseLoginDialogView.java
```diff
@@ -0,0 +1,120 @@
...
+        this.setTitle(Localization.lang("Connect to shared database"));
+
+        ViewLoader.view(this)
+                  .load()
+                  .setAsDialogPane(this);
+
+        ControlHelper.setAction(connectButton, this.getDialogPane(), event -> openDatabase());
+        btnConnect = (Button) this.getDialogPane().lookupButton(connectButton);
+        setLoadingConnectButtonText();
+        btnConnect.disableProperty().bind(viewModel.formValidation().validProperty().not());
```
tobiasdiez: move the last line to `initialize`

src/main/java/org/jabref/gui/shared/SharedDatabaseLoginDialogView.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    @FXML
+    void openFileDialog(ActionEvent event) {
+        viewModel.openFileDialog();
+    }
+
+    private void setLoadingConnectButtonText() {
+
+        if (viewModel.loadingProperty().get()) {
```
tobiasdiez: This should be done via binding, because otherwise setting `loading = true` does not change the text to "Connecting...". You can use `EasyBind.map` to convert the boolean to the appropriate text and then bind this text to `btnConnect.textProperty`

src/main/java/org/jabref/gui/shared/SharedDatabaseLoginDialogViewModel.java
```diff
@@ -0,0 +1,324 @@
...
+import org.slf4j.LoggerFactory;
+
+public class SharedDatabaseLoginDialogViewModel extends AbstractViewModel {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(SharedDatabaseLoginDialogViewModel.class);
+
+    private final ListProperty<DBMSType> allDBMSTypes = new SimpleListProperty<>(FXCollections.observableArrayList(DBMSType.values()));
+    private final ObjectProperty<DBMSType> selectedDBMSType = new SimpleObjectProperty<>(allDBMSTypes.get(0));
+
+    private final StringProperty db = new SimpleStringProperty("");
```
tobiasdiez: no abbreviation here and above (selectedDatabaseType).

src/main/java/org/jabref/gui/shared/SharedDatabaseLoginDialogViewModel.java
```diff
@@ -0,0 +1,324 @@
...
+import de.saxsys.mvvmfx.utils.validation.Validator;
+import org.fxmisc.easybind.EasyBind;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class SharedDatabaseLoginDialogViewModel extends AbstractViewModel {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(SharedDatabaseLoginDialogViewModel.class);
+
+    private final ListProperty<DBMSType> allDBMSTypes = new SimpleListProperty<>(FXCollections.observableArrayList(DBMSType.values()));
```
tobiasdiez: since this is an static enum, you probably don't need to wrap it in a list property.

src/main/java/org/jabref/gui/shared/SharedDatabaseLoginDialogViewModel.java
```diff
@@ -0,0 +1,324 @@
...
+    private final StringProperty folder = new SimpleStringProperty("");
+    private final BooleanProperty autosave = new SimpleBooleanProperty();
+    private final BooleanProperty rememberPassword = new SimpleBooleanProperty();
+    private final BooleanProperty loading = new SimpleBooleanProperty();
+
+    private final JabRefFrame frame;
+    private final DialogService dialogService;
+    private final SharedDatabasePreferences prefs = new SharedDatabasePreferences();
+
+    private final Validator dbvalidator;
```
tobiasdiez: no abbreviation

src/main/java/org/jabref/gui/shared/SharedDatabaseLoginDialogViewModel.java
```diff
@@ -0,0 +1,324 @@
...
+
+    public SharedDatabaseLoginDialogViewModel(JabRefFrame frame, DialogService dialogService) {
+        this.frame = frame;
+        this.dialogService = dialogService;
+
+        EasyBind.subscribe(selectedDBMSType, selected -> {
+            port.setValue(Integer.toString(selected.getDefaultPort()));
+        });
+
+        Predicate<String> predicate = input -> (input != null) && !input.trim().isEmpty();
```
tobiasdiez: StringUtil has a `isNotBlank` method that does exactly this.

src/main/java/org/jabref/gui/shared/SharedDatabaseLoginDialogViewModel.java
```diff
@@ -0,0 +1,324 @@
...
+
+    private final ListProperty<DBMSType> allDBMSTypes = new SimpleListProperty<>(FXCollections.observableArrayList(DBMSType.values()));
+    private final ObjectProperty<DBMSType> selectedDBMSType = new SimpleObjectProperty<>(allDBMSTypes.get(0));
+
+    private final StringProperty db = new SimpleStringProperty("");
+    private final StringProperty host = new SimpleStringProperty("");
+    private final StringProperty port = new SimpleStringProperty("");
+    private final StringProperty user = new SimpleStringProperty("");
+    private final StringProperty password = new SimpleStringProperty("");
+    private final StringProperty folder = new SimpleStringProperty("");
```
tobiasdiez: is it a folder or actually a file?

src/main/java/org/jabref/gui/shared/SharedDatabaseLoginDialogViewModel.java
```diff
@@ -0,0 +1,324 @@
...
+            port.setValue(Integer.toString(selected.getDefaultPort()));
+        });
+
+        Predicate<String> predicate = input -> (input != null) && !input.trim().isEmpty();
+
+        dbvalidator = new FunctionBasedValidator<>(db, predicate, ValidationMessage.error(Localization.lang("Required field \"%0\" is empty.", Localization.lang("Library"))));
+        hostvalidator = new FunctionBasedValidator<>(host, predicate, ValidationMessage.error(Localization.lang("Required field \"%0\" is empty.", Localization.lang("Port"))));
+        portValidator = new FunctionBasedValidator<>(port, predicate, ValidationMessage.error(Localization.lang("Required field \"%0\" is empty.", Localization.lang("Host"))));
+        uservalidator = new FunctionBasedValidator<>(user, predicate, ValidationMessage.error(Localization.lang("Required field \"%0\" is empty.", Localization.lang("User"))));
+        folderValidator = new FunctionBasedValidator<>(folder, predicate, ValidationMessage.error(Localization.lang("Please enter a valid file path.")));
```
tobiasdiez: I think we should use the nio-path methods to check if the provided path is actually valid.

src/main/java/org/jabref/gui/shared/SharedDatabaseLoginDialogViewModel.java
```diff
@@ -0,0 +1,324 @@
...
+
+        formValidator = new CompositeValidator();
+        formValidator.addValidators(dbvalidator, hostvalidator, portValidator, uservalidator);
+    }
+
+    public void openDatabase() {
+
+        connectionProperties = new DBMSConnectionProperties();
```
tobiasdiez: I would convert this to a local variable and pass it to `openSharedDatabase` as a argument (because it may be null and a caller of `openSharedDatabase` has to know that `connectionProperties` need to be init first...)

src/main/java/org/jabref/gui/shared/SharedDatabaseLoginDialogViewModel.java
```diff
@@ -0,0 +1,324 @@
...
+            Path localFilePath = Paths.get(folder.getValue());
+
+            if (Files.exists(localFilePath) && !Files.isDirectory(localFilePath)) {
+
+                boolean overwriteFilePressed = dialogService.showConfirmationDialogAndWait(Localization.lang("Existing file"),
+                                                                                           Localization.lang("'%0' exists. Overwrite file?", localFilePath.getFileName().toString()),
+                                                                                           Localization.lang("Overwrite file"),
+                                                                                           Localization.lang("Cancel"));
+                if (!overwriteFilePressed) {
+                    // tbFolder.requestFocus();
```
tobiasdiez: remove comment

src/main/java/org/jabref/gui/shared/SharedDatabaseLoginDialogViewModel.java
```diff
@@ -0,0 +1,324 @@
...
+            prefs.clearPassword(); // for the case that the password is already set
+        }
+
+        prefs.setRememberPassword(rememberPassword.get());
+    }
+
+    /**
+     * Fetches possibly saved data and configures the control elements respectively.
+     */
+    public void applyPreferences() {
```
tobiasdiez: Make this private and call from constructor?

src/main/java/org/jabref/gui/shared/SharedDatabaseLoginDialogViewModel.java
```diff
@@ -0,0 +1,324 @@
...
+                LOGGER.error("Could not read the password due to decryption problems.", e);
+            }
+        }
+
+        rememberPassword.set(sharedDatabaseRememberPassword);
+    }
+
+    private boolean isSharedDatabaseAlreadyPresent() {
+        List<BasePanel> panels = frame.getBasePanelList();
+        return panels.parallelStream().anyMatch(panel -> {
```
tobiasdiez: Can you try to move this check using `frame` to `StateManager` (it also gives access to the list of open databases).

Siedlerchr: It currently only support the active database, needs probably be extended in JabRefFrame, there is this line:   EasyBind.subscribe(tabbedPane.getSelectionModel().selectedItemProperty(), e -> {

src/main/java/org/jabref/gui/shared/SharedDatabaseLoginDialogViewModel.java
```diff
@@ -0,0 +1,359 @@
...
+
+            result.filter(btn -> btn.equals(openHelp)).ifPresent(btn -> HelpAction.openHelpPage(HelpFile.SQL_DATABASE_MIGRATION));
+            result.filter(btn -> btn == ButtonType.OK).ifPresent(btn -> openSharedDatabase());
+
+        }
+        loading.set(false);
+    }
+
+    private void setPreferences() {
+        prefs.setType(selectedDBMSType.getValue().toString());
```
tobiasdiez: Return the preferences instead of setting a local variable? Makes the flow more logical in my opinion.

src/main/java/org/jabref/logic/shared/DBMSConnection.java
```diff
@@ -21,23 +22,30 @@
...
         try {
             DriverManager.setLoginTimeout(3);
             // ensure that all SQL drivers are loaded - source: http://stackoverflow.com/a/22384826/873282
             // we use the side effect of getAvailableDBMSTypes() - it loads all available drivers
             DBMSConnection.getAvailableDBMSTypes();
 
-            this.connection = DriverManager.getConnection(
-                    properties.getType().getUrl(properties.getHost(), properties.getPort(), properties.getDatabase()),
-                    properties.getUser(), properties.getPassword());
+            Properties props = new Properties();
```
tobiasdiez: Can you please move these additions to the `DBMSConnectionProperties ` class as two methods `Properties  asJavaProperties()` and `String getUrl`.

src/main/java/org/jabref/logic/shared/DBMSConnectionProperties.java
```diff
@@ -100,15 +101,23 @@ public void setPassword(String password) {
...
     public boolean equals(DBMSConnectionProperties properties) {
         return this.type.equals(properties.getType())
-                && this.host.equalsIgnoreCase(properties.getHost())
-                && (this.port == properties.getPort())
-                && this.database.equals(properties.getDatabase())
-                && this.user.equals(properties.getUser());
+               && this.host.equalsIgnoreCase(properties.getHost())
+               && (this.port == properties.getPort())
+               && this.database.equals(properties.getDatabase())
+               && this.user.equals(properties.getUser());
```
tobiasdiez: add `useSSL` in equals and hash

src/main/java/org/jabref/logic/shared/PostgreSQLProcessor.java
```diff
@@ -89,7 +89,8 @@ public void startNotificationListener(DBMSSynchronizer dbmsSynchronizer) {
             // Otherwise the listener is going to be deleted by GC.
             PGConnection pgConnection = connection.unwrap(PGConnection.class);
             listener = new PostgresSQLNotificationListener(dbmsSynchronizer, pgConnection);
-            listener.start();
+            new Thread(listener, "PostgresSQLNotificationListener").start();
```
tobiasdiez: Better reuse our threading infrastructure so that we are sure that the thread is shutdown on exit etc.

src/main/java/org/jabref/logic/shared/listener/PostgresSQLNotificationListener.java
```diff
@@ -13,23 +13,29 @@
...
+    private volatile boolean stop;
 
     public PostgresSQLNotificationListener(DBMSSynchronizer dbmsSynchronizer, PGConnection pgConnection) {
         this.dbmsSynchronizer = dbmsSynchronizer;
         this.pgConnection = pgConnection;
     }
 
+    public void start() {
+        stop = false;
+        new Thread(this).start();
```
tobiasdiez: same

src/main/java/org/jabref/logic/shared/prefs/SharedDatabasePreferences.java
```diff
@@ -71,6 +72,11 @@ public boolean getRememberPassword() {
         return internalPrefs.getBoolean(SHARED_DATABASE_REMEMBER_PASSWORD, false);
     }
 
+    public boolean isUseSSL()
+    {
```
tobiasdiez: code style

PR #4043, opened 2018-05-21
Add "Add Braces" Formatter

koppor:
In the context using Springer's default bst file, the "title casing" feature of BibTeX is not desired. When working with persons quickly creating their bibliography, the solution is to put `{}` around the `title` field. Up to now, JabRef supports the inverse feature by the `RemoveBracesFormatter`. This PR adds the `AddBracesFormatter`. I am very aware that the other formatters such as our `org.jabref.logic.formatter.casechanger.ProtectTermsFormatter`. However, that formatter relies on a fixed list of terms, whereas this `AddBracesFormater` "just works" (under some assumptions).

----


Siedlerchr:
You can directly use right click option -> Add braces around selected word

Siedlerchr:
What happens to latex encoded titles? .e.g If i have some Unicode latex stuff in there

koppor:
I agree. This PR helps in updating a complete BibTeX database with more than 100 entries. I could go manually through it and tell others to do the same. However, I like automating things. This option offers a functionality requested to me multiple times by different persons. I could also explain the protected terms thing, but I do not want to force people to use JabRef in a certain way if there is an easy other way.

koppor:
In the default configuration, the formatter runs as last. Thus, the protection functionality comes at last. I tried it out:

grafik

    title = {Än interesting Historüüü}

becomes

    title = {{{{\"{A}}}n interesting Histor{\"{u}}{\"{u}}{\"{u}}}

Siedlerchr:
Okay good! From my point of view good

tobiasdiez:
It is [usually recommended](https://tex.stackexchange.com/questions/10772/bibtex-loses-capitals-when-creating-bbl-file) to only "protect" acronyms or proper names and leave the rest to the citation style. In the bibtex world, the citation style decides the general casing of content words. Thus, I'm not sure if we should add a formatter that encourages the non-standard approach.

koppor:
I do not want to train 90% of the people to do the right thing we believe in. Sometimes, I just want to get a paper finished to the deadline and not spending time educating people in BibTeX. 😇 We could add a warning in the documentation to encourage other behavior.

Other stream of argumentation: None of us found time to work on the ProtectedTermsFormatter to include more words. This PR is a work around. 😇

PR #4044, opened 2018-05-21
Streamline formatter implementation

koppor:
This builds on https://github.com/JabRef/jabref/pull/4043 and streamlines the code of the "formatter" framework a bit.

- Introduce AbstractFormatter to ensure that `hashCode` and `equals` are implemented the same way
- Introduce a test to ensure that `getKey()` returns a unique key


Siedlerchr:
Why not adding the methods implemented to the interface?

koppor:
grafik

tobiasdiez:
Why not make `Formatter` an abstract class?

koppor:
If this get merged, please with normal merge so that #4045 can be merged correctly. 😇

koppor:
@stefan-kolb introduced the interface paradigma for formatters in 2015. The interface resides in our model package, whereas the concrete implementations reside in the logic package. In c09e4720, I changed the interface to an abstract class.

src/test/java/org/jabref/logic/formatter/FormatterTest.java
```diff
@@ -29,6 +31,18 @@ public static void setUp() {
...
+    /**
+     * When a new formatter is added by copy and pasting another formatter, it may happen that the <code>getKey()</code> method is not adapted. This results in duplicate keys, which this test tests for.
+     */
+    @Test
+    public void allFormatterKeysAreUnique() {
+        // idea for uniqueness checking by https://stackoverflow.com/a/44032568/873282
+       assertFalse(getFormatters().collect(Collectors.groupingBy(
+               formatter -> formatter.getKey(),
+               Collectors.counting()
+       )).entrySet().stream().anyMatch(e -> e.getValue() > 1));
```
tobiasdiez: In case somebody adds a new formatter with the same key, this test fails but the error message does not say why. I would change `anyMatch` to filter and compare (assertEquals) the resulting map to an empty map.

PR #4047, opened 2018-05-22
A first solution for the cli problem #4038

johannes-manner:
ExportFactory is not initialized at this execution point...
The printUsage method is used twice. The JabRefCLI prints the usage, if an error occurs during argument parsing and the ArgumentProcessor prints the usage, when --help option is chosen. 

I made two solutions for this problem:
 - Commit 1: Some sort of a hack to only get the names of the exporters without altering the processing in the main method. 
image 

 - Commit 2: Changed the order of execution in the main method. The argument processor needs the export factory. I don't know, if the processing of the preferences will influence the cli functionality?

Fixes #4038 

- [ ] Change in CHANGELOG.md described
- [ ] Tests created for changes
- [X] Manually tested changed features in running JabRef

PR #4054, opened 2018-05-24
Restructure startup

stefan-kolb:
Considered as necessary when reviewing #4047

### DIFF
--- a/src/main/java/org/jabref/JabRefMain.java
+++ b/src/main/java/org/jabref/JabRefMain.java
@@ -106,7 +106,7 @@ public class JabRefMain extends Application {
         final JabRefPreferences preferences = JabRefPreferences.getInstance();
         Globals.prefs = preferences;
         // Perform migrations
-        migratePreferences();
+        PreferencesMigrations.runMigrations();
 
         configureProxy(preferences.getProxyPreferences());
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/jabref/JabRefMain.java
```diff
@@ -93,58 +98,43 @@ private static void ensureCorrectJavaVersion() {
...
-        Globals.protectedTermsLoader = new ProtectedTermsLoader(Globals.prefs.getProtectedTermsPreferences());
-
         // Check for running JabRef
+        if (!allowMultipleAppInstances(args)) {
+            return;
+        }
+
+        // See if we should shut down now
+        if (argumentProcessor.shouldShutDown()) {
+            Globals.shutdownThreadPools();
```
koppor: This code is appearing twice, too. Can it be moved to `shutdownCurrentInstance()`?

Alternatively, change line this line to 

   if (!allowMultipleAppInstances(args) || argumentProcessor.shouldShutDown()) {

and remove the shutdown code from `allowMultipleAppInstances`.

src/main/java/org/jabref/JabRefMain.java
```diff
@@ -93,58 +98,43 @@ private static void ensureCorrectJavaVersion() {
...
+            Platform.exit();
+            return;
+        }
+
+        // If not, start GUI
+        SwingUtilities
+                .invokeLater(() -> new JabRefGUI(argumentProcessor.getParserResults(), argumentProcessor.isBlank()));
+    }
+
+    private static boolean allowMultipleAppInstances(String[] args) {
```
koppor: Rename that to `handleMultipleInstances`, because some "black magic" is happening here. Nut just a check whether multiple instances are allowed.

tobiasdiez:
@stefan-kolb may I ask you to merge the master now also into the maintable-beta branch. You are more familiar with the changes to the startup code than me.

PR #4056, opened 2018-05-25
Extract old Bibtex key generation action

### LINKED ISSUE #4000: 'Do not overwrite existing keys' does not work
Under:... Options > Preferences > BibTex key generator

I checked the 'Do not overwrite...' box but bib entry keys are still overwritten when I click on [generate] key icon. I have tried restarting JabRef but same result. Version 4.2
- - -
Follow up: 2018/05/07

Whether 'Do not overwrite existing keys' flag is set or not, if I double click the generate key icon, it locks the new bibtexkey as one might want but there is no visual indication that the key is actually locked. If I click once every two (2) seconds or so, the key continues to be changed. So there is a locking mechanism but not as documented - perhaps not as designed. If key is locked with double click, I can still delete the key and generate a new one.

stefan-kolb:
Resolves #4000 

This is one for the JavaFX guys.
We need a new JavaFX action here or duplicate parts of the code.

I already marked the respective code locations with TODOs.
- We could use the FXDialogService but there is an additional checkbox for "Do not ask this question again" in the old dialog.
- Also, I don't know how to properly extract such Worker code for JavaFX, otherwise I would have to duplicate the code on all occurences


Siedlerchr:
I would suggest you directly implement this on the maintable beta branch. Otherwise we will get the hell of merge conflicts and we would need to rework the AbstractWorker again.
Have a look at the functionality of  BackgroundTask.wrap(()

stefan-kolb:
Is there any possibility to extract and reuse actions in JavaFX, except from the Action in controlsfx?

Siedlerchr:
Sure,  we do this all the time. In the main table.beta branch we use the `OpenXYZAction extends SimpleCommand ` class for the things in `org.gui.jabref.actions `

stefan-kolb:
This is also functionality from a beta framework....

Siedlerchr:
> We could use the FXDialogService but there is an additional checkbox for "Do not ask this question again" in the old dialog.

The DialogService has an option for this, where you can add a consumer? which then should store the property

tobiasdiez:
I've now extracted the code of the cite key generation of a single entry and extended it by a warning dialog that is shown if a previous key is about to be overwritten. 

With these changes, the buttons in the entry editor show now a button text instead of a single icon. Fixing this requires https://bitbucket.org/controlsfx/controlsfx/pull-requests/706 to be merged. Thus, in the current state this PR is not mergable but ready-for-review.

stefan-kolb:
Thanks for your efforts! 
> With these changes, the buttons in the entry editor show now a button text instead of a single icon.
What does this mean? Why text?

tobiasdiez:
Instead of the icon, the text "Generate key" appears. This is due to a limitation of the controlsfx library that we use (hopefully, my PR there gets merged in the next few days).

stefan-kolb:
Can we merge this now? Is it released by controlsfx?

tobiasdiez:
Yes, my PR is finally integrated into the snapshot build of controlsfx and thus I'll merge it now.

Issue #3943, opened 2018-04-10
Entry editor produces constant load


stefan-kolb:
When the entry editor is opened I experience constant high load from JabRef.
This needs to be evaluated and fixed!

![image](https://user-images.githubusercontent.com/2141507/38553320-3173fbec-3cbf-11e8-9509-23de34484798.png)

When I close the entry editor load immediately goes down to idle.

stefan-kolb:
It looks like JR is producing events? in such cases.
The classes that are created over and over are:
`AccessControlContext, EventQueueItem, WeakReference`
After some time the garbage collector kills all of them and the climb to the peak starts again.
@JabRef/developers What could cause this? We really need to find out...

stefan-kolb:
Maybe relevant: https://stackoverflow.com/questions/14558266/clean-javafx-property-listeners-and-bindings-memory-leaks?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa

stefan-kolb:
I just need to click tab titles and after a while the leak starts!

stefan-kolb:
I couldn't find anything yet, tho I'm not so much into these code parts.

stefan-kolb:
I couldn't reproduce the constant load on the maintabl branch with a few tests yet, although I am experiencing this phenomen when cycling through the entries (memory spike/leak again):

![image](https://user-images.githubusercontent.com/2141507/40537148-8571e50c-600e-11e8-8eac-e75aa4b33b79.png)

stefan-kolb:
Looks like AWT is creating InvocationEvents all the time:

```
[...]

Issue #3989, opened 2018-04-27
Search crashes completely on mid-sized database


Sethur:
**Versions:**
Issue persists in dev snapshot: https://builds.jabref.org/master/JabRef-4.3-dev--snapshot--2018-04-26--master--7629b3171.jar

Tested on Windows 7 with Java x64 Build 1.8.0_162-b12

**Issue:**

On a database with about 2000 entries, most of them with associated PDFs, the search box behaves even worse in 4.2 than it did in 4.1.

Problems are:

1. Random duplication of entered search strings (see issue #3921 )
2. Random lockups of the complete search box (no new text can be entered, old text cannot be deleted, no text can be marked)
3. Random vertical shifts of the already entered characters as if a new blank line was added.

Problems 1) and 2) were already present in 4.1, Problem 3) seems to be a new addition of 4.2. I also have the feeling that the crashes will come faster now, on average it takes only 3-4 searches until something bad happens and you have to restart JabRef entirely.

Siedlerchr:
I really can not reproduce this with Win10 and a bib database with '>6.000 entries.
It could be a memory problem. How much RAM do you installed in your machine?
 Do you have the entry open when searching`? Probably at the General Tab?
When you close the entry editor, or switch to another tab the search should be way faster.

stefan-kolb:
For me it looks like this is still some double generation or reception of events.
It feels like it is occuring more often when load is on the computer and if you type fast.
Strangely enough afterwards it does not recover from this duplictaed events, which I would expect if I delete the search term.
I also experience that after deleting the term te cursor stays on the left forever. 
I think we had this bug before and "fixed" it.
It is here again.

Siedlerchr:
The search timer waits 400 ms until it executes the search. When you start typing the current search gets cancelled and the timer resetted.
I think it could maybe due to the richtextfx thing which was also used for the source tab.

juejung:
I'm experiencing a similar issue with JabRef 4.2 running on Ubuntu 16.04.
While searching a database of > 2,000 entries is fine, as soon as I double click on an entry after a search and the input window opens, JabRef freezes.

Actually, JabRef freezes as soon as I open an entry a second time (without any prior search or filtering). This did not happen with JabRef 4.1.

stefan-kolb:
Everyone that has this problems: could you please update your Java to the latest version (Version 8 Update 172). I did experience way less problems with this version. The old Java 162 has a CPU leak that causes the events to duplicate etc.

[...]

PR #4058, opened 2018-05-25
CPU leak

stefan-kolb:
See #3943
- Fix Java to later version
- Pretty restrictive for the user unfortunately!


stefan-kolb:
@JabRef/developers I worked a little bit with Jabref 4.2 and the new JDK and I feel like also the problems with duplicated entries in the search box and freezes are less frequent now.
Maybe the constant load provokes these issues even more.
Therefore, I vote to release a new JR version where we force the users to upgrade to the newest Java JRE/JDK!

Also refs #3989.
Fixes #3943.

stefan-kolb:
As this really improves the situation (at least for me), I think we should definitly include this in our release.

koppor:
@stefan-kolb Which tool did you use for the CPU monitoring?

I tried `java version "1.8.0_171"` and it doesn't seem to cause high CPU load.

Focused entry editor:

grafik

Unfocused entry editor:

grafik

stefan-kolb:
I used the visualvm. It is actually not caused by the entry editor but by switching between programs.

Issue #4025, opened 2018-05-09
Groups are focused on right-click


andre8s:
- [X] I have tested the latest master version from http://builds.jabref.org/master/ and the problem persists

JabRef version 4.2 on Windows 7 6.1 amd64, Java 1.8.0_161

I am using a database with 12000+ entries and a large number of keyword groups, which are arranged in a hierarchical tree (via using sub groups). In the group dialog, the sub branches of the group tree can be folded or unfolded via the small arrow symbols at the right side. Further, it is possible to right-click on a group or sub-group and get a context menu which allows to "add selected entries to this group". 

The problem is that already clicking on the fold/unfold symbol immediately affects the filtering status of the main entry list. Also right-clicking on a group or subgroup has the same effect. In jabref-3.x it was possible to fold/unfold a sub branch by clicking on a "+" symbol without changing the filtering state in the entry viewer. Also, getting the context menu via right-clicking did not change the filtering state. 

Currently, when I add a new literature entry and want to connect it with some keyword groups, it is not possible to do this via the right-clicking context menu, because at the moment of right-clicking, my entry disappears from the main list (because at the moment of righ-clicking it is not yet in that group, but filtering becomes immediately effective). 

An alternative way is to drag-drop the literature entry onto a group element. However in case of nested subgroups, the subbranch needs to be unfolded first, and by unfolding the subgroup the literature entry again disappears in the main entry list. Thus the way to do this is

- Find the right place in the group list and unfold the required subbranches of subgroups 
- Scroll to the top of the group list and click on "All entries"
- Scroll back to the place, where the desired subbranch is unfolded
- In some cases it is again required to find and mark the literature entry in the entry list (which temporarily disappeared)
- Finally drag-drop the literature entry onto the subgroup

A second workaround is right-clicking on the literature entry and select "add to group" in the context menu, however this opens a separate dialog box with the same list of groups, which again needs scrolling and unfolding of subgroups and allows for selecting only one group per time. In case of multiple group selections this dialog box has to be opened for each group individually and always starts at the top of the group list with all sub branches folded. 

Both workarounds are rather cumbersome. A "simple" (at least conceptually simple, I'm not aware of the coding details...) solution would be preventing the group filtering in the main entry window upon right-clicking on a group or upon clicking on the small arrow for folding/unfolding. The filtering should become effective only upon clicking with the left mouse button onto a group name. At least in jabref-3.x it always worked this way.

tobiasdiez:
[...]

PR #4061, opened 2018-05-26
Groups right click

tobiasdiez:
(Partly) fixes https://github.com/JabRef/jabref/issues/4025. Right-click on a group no longer focuses that group. It's not perfect since there is no longer feedback on which group the user clicked and the context menu belongs to.  

Clicking the disclosure node (the small arrow) still leads to an immediate selection of the group. The selection happens before any events are triggered on the row and thus I'm not able to intercept the selection.

----


tobiasdiez:
Merge this now for the sake of moving forward...

PR #4068, opened 2018-05-31
Fix JabRef leaks open threads after exit

Siedlerchr:
I noticed that whenver I run JabRef in Eclipse that after the exit of JabRef still some threads were open and not stopping. I have no idea why and where they are coming from. 

----

### DIFF
--- a/src/main/java/org/jabref/JabRefMain.java
+++ b/src/main/java/org/jabref/JabRefMain.java
@@ -150,8 +150,10 @@ public class JabRefMain extends Application {
 
     private static void shutdownCurrentInstance() {
         Globals.shutdownThreadPools();
+        Globals.REMOTE_LISTENER.stop();
         // needed to tell JavaFx to stop
         Platform.exit();
+        System.exit(0);
     }
 
     private static void applyPreferences(JabRefPreferences preferences) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/jabref/JabRefMain.java
```diff
@@ -150,8 +150,10 @@ private static boolean handleMultipleAppInstances(String[] args) {
 
     private static void shutdownCurrentInstance() {
```
tobiasdiez: Is still method still needed, i.e. cannot we just call `stop` ?

LinusDietz:
Discussed this in the devcall. Consensus was the we don't like this quickfix, but would recommend to look at the underlying reason why the threads in questions don't terminate.

tobiasdiez:
Can you please try if cherry-picking https://github.com/JabRef/jabref/pull/4109/commits/39e30cca85b79a4a1aa23a4742d06650203834a1 in 4.x solves the issue for you. The telemetry library had a bug that prevented some threads to stop https://github.com/Microsoft/ApplicationInsights-Java/issues/662.

Siedlerchr:
@tobiasdiez  Works fine! All threads are exited. I will directly push that to the 4.x branch so that we have it as back port. I will test if this is suffiient in master, too.

PR #4069, opened 2018-06-01
Additional "Cancel" and "VSCode" icon

halirutan:
In addition to the new icons, I have removed the `lcd` antialiasing for icons. We discussed several times that this setting makes icons crispy on Linux while it doesn't seem to have a major effect on other operating systems.

----


src/main/java/org/jabref/gui/icon/JabRefMaterialDesignIcon.java
```diff
@@ -26,7 +26,9 @@
     COPY("\ue909"),
     PASTE("\ue90a"),
     SET_CENTER("\ue90b"),
-    SET_ALL("\ue90c");
+    SET_ALL("\ue90c"),
+    VSCODE("\ue90d"),
+    CANCEL("\ue90e");
```
koppor: I think, we could have reused the original material design font without copy and pasting and icon? Let's see, how I can include the new icon, because constructor of `JabRefIcon` needs a `MaterialDesignIcon`.

tobiasdiez: @koppor This is on the maintable-beta branch...there JabRefIcon is an interface.

halirutan: @koppor That was a design-flaw I fixed for the maintable branch. It basically only needs to implement the `GlyphIcons` interface. In essence it was a more or less simple refactoring and now the `JabRefIcons` constructor can take both our Icons and MD icons. Look at the `org.jabref.gui.icon` package on the maintable branch. Maybe you can cherry pick it somehow.

PR #4074, opened 2018-06-02
Mac graphics

### LINKED ISSUE #4070: discrete versus integrated graphics card on MacOS
When starting jabref on MacOS the system automatically changes from the integrated to the discrete graphics card. This is not necessary and is due to the following incorrect key in the Info.plist in jabref 4.2:

```
<key>NSSupportsAutomaticGraphicsSwitching</key>
<string>YES</string>

```

which should in fact be:
```
<key>NSSupportsAutomaticGraphicsSwitching</key>
<true/>

```
This is also documented here:
[https://developer.apple.com/library/content/qa/qa1734/_index.html](https://developer.apple.com/library/content/qa/qa1734/_index.html)

stefan-kolb:
Resolves #4070 

@LinusDietz @halirutan  Can you also take a look as you own a mac, please 😄


stefan-kolb:
Should also be backported to v4.x

halirutan:
@stefan-kolb I certainly can test something, but what exactly should I look out for? With JabRef open, should my Macbook use the integrated graphics-card or the fast one? Or should I only test if JabRef works?

jcrbloch:
FYI: I just upgraded from Jabref 4.2 to 4.3.1. I had to correct the plist again to have jabref use the integrated graphics card, see [https://github.com/JabRef/jabref/issues/4070](https://github.com/JabRef/jabref/issues/4070). So I guess that we are just waiting for your corrections to be merged.

halirutan:
It seems to work. My *Activity Monitor* says that JabRef does not require high-performance GPU. Is there anything else I should look out for?

stefan-kolb:
I guess that's it. @LinusDietz

PR #4075, opened 2018-06-02
Arχiv supports TLS too.

KOLANICH:


koppor:
Thank you for your pull requests. I know that it is hard work in addition to adding an `s` at one or two places, but could you add an entry to `CHANGELOG.md`, too?.

I see that you used GitHub's feature instead of using an IDE. You could use Eclipse Che (https://www.eclipse.org/che/) in your browser to bundle the updates of multiple files into one pull reuqest.

KOLANICH:
>I know that it is hard work in addition to adding an s at one or two places, but could you add an entry to CHANGELOG.md, too?.

Please, do that yourself.

koppor:
Fetcher tests run through, so I merge. Thank you for the PR.

PR #2157, opened 2016-10-12
[WIP] Integrate BibSonomy Plugin

zesaro:
**We try to offer the latest build at http://builds.jabref.org/bibsonomy/**. Feel free to comment on the functionality in this PR.

[BibSonomy](https://www.bibsonomy.org/) is a tool for maintaining references. [PUMA](https://puma.ub.uni-stuttgart.de/) is the instance of the University of Stuttgart. For more information, check the related work of [CloudRef](https://dblp.org/rec/html/conf/zeus/KoppBM18).

Example usage of BibSonomy:

grafik

> Aim: Get first working version into JabRef again, do some marketing, and attract a maintainer for that functionality.

The aim of this pull request is to integrate the BibSonomy Plugin from JabRef 2.11 into the actual JabRef master branch. You can run BibSonomy from the menubar (between `Tools` and `Options`), search for entries and fetch them. There are still parts in the code which has to be updated. Since 2.11 much has changed. For example the MetaData which I have to investigate.

I am looking forward to hear your feedback.
- [ ] Change in CHANGELOG.md described
- [ ] Tests created for changes
- [ ] Screenshots added (for bigger UI changes)
- [ ] Manually tested changed features in running JabRef
- [x] If you changed the localization: Did you run `gradle localizationUpdate`?

## Notes

- [ ] BibSonomy seems to use following fields: `interhash;intrahash;keywords;groups;privnote`. Special handling for them is removed in this PR as it is unclear how they are used.

PR #4076, opened 2018-06-02
TLS for bibsonomy

KOLANICH:


koppor:
I didn't know that we have a bibsonomy scraper. Refs #2157.

Fetcher tests run through, so I merge. Thank you for the PR.

PR #4090, opened 2018-06-03
New Crowdin translations

Siedlerchr:
Enjoy!

PR #4094, opened 2018-06-04
Remove menu* from crowdin.yml

koppor:
Our JavaFX port does not need separate files anymore.

Refs #3873 #3877


Siedlerchr:
Do we need to configure or resync sth at crowdin?

koppor:
I paused the synchronization. At a resume, crowdin should obey the new setting. The latest crowdin PR https://github.com/JabRef/jabref/pull/4090 looks good from the side of `Menu_*.properties`.

PR #4095, opened 2018-06-04
Set minimum java to 171

### LINKED ISSUE #4093: Minimum required java version for JabRef 4.3 is higher than the most recent stable java release provided for consumers on Windows 10
JabRef version 4.3 on Windows 10 x64

This issue seems like a small increment error which was introduced in JabRef 4.3 (see [the first line here](https://github.com/JabRef/jabref/blob/v4.3/CHANGELOG.md#changed)) 

Steps to reproduce:

1. Install the most recent java version to date from [this page](https://java.com/en/download/win10.jsp) (which is 1.8.0_171 to date)
2. Install JabRef 4.3 with the installer provided
3. Start JabRef and see the error message that 1.8.0_172 is required, although 1.8.0_171 is the most recent available version

The problem is, 1.8.0_171 and 1.8.0_172 were released simultaneously, shown [here](http://www.oracle.com/technetwork/java/javase/8u172-relnotes-4308893.html) and [here](http://www.oracle.com/technetwork/java/javase/8u171-relnotes-4308888.html). 
Version 1.8.0_171 does not update to 1.8.0_172 automatically, even after checking for new updates in the scheduler, so my question is: Does the CPU leak described in [3943](https://github.com/JabRef/jabref/issues/3943) still persist in 1.8.0_171 or should JabRef support both versions?

tobiasdiez:
Fixes #4093. Since Java does not automatically update to 171, users have to manually install the 172 update. This is very convenient. 

I actually propose to cherry-pick this PR to 4.x and immediately release 4.3.1.

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 ## [Unreleased]
 
 ### Changed
+- We changed the minimum required version of Java to 1.8.0_171, as this is the latest release for which the automatic Java update works.  [4093](https://github.com/JabRef/jabref/issues/4093)
 - The special fields like `Printed` and `Read status` now show gray icons when the row is hovered.
 - We added a button in the tab header which allows you to close the database with one click. https://github.com/JabRef/jabref/issues/494
 - Sorting in the main table now takes information from cross-referenced entries into account. https://github.com/JabRef/jabref/issues/2808
--- a/build.gradle
+++ b/build.gradle
@@ -52,7 +52,7 @@ targetCompatibility = 1.8
 mainClassName = "org.jabref.JabRefMain"
 
 // These are the Java version requirements we will check on each start of JabRef
-ext.minRequiredJavaVersion = "1.8.0_172"
+ext.minRequiredJavaVersion = "1.8.0_171"
 ext.allowJava9 = false
 
 sourceSets {


stefan-kolb:
Let me check if the fix is already included in _171

stefan-kolb:
https://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-8195830

Unfortunately this is only fixed in The bug has been fixed in 8u172 b01

koppor:
For provenance: This downgrades the minimum Java version introduced at https://github.com/JabRef/jabref/pull/4058.

Siedlerchr:
Have you looked at the oracle changelog page for 171? Where you download the jdk there is a changelog linked. It could be that 171 only includes security fixes

koppor:
For a quick try out: `choco install jre8`. This installs Java 8 171. See <https://chocolatey.org/packages/jre8>.

Deep link to the Changelog:

- 171: http://www.oracle.com/technetwork/java/javase/8u171-relnotes-4308888.html
- 172: http://www.oracle.com/technetwork/java/javase/8u172-relnotes-4308893.html

The Changelog of 172 shows:

> The changes made under JDK-8033530 introduced an inconsistency between the implementation for and the documentation of the following methods:

However, JDK-8033530 is not listed in the Changelog for 171. I would assume nevertheless that this issue is fixed in 171. See https://github.com/JabRef/jabref/pull/4058#issuecomment-394593243.

stefan-kolb:
Ok, I have also tried to reproduce the bug with v171 and the code from https://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-8195830. Cannot see any spikes right now. We cannot be 100% sure as it is not in the changelog, but maybe it is worth risking it until v172 is rolled out everywhere. i guess we should still force v172 later on.

koppor:
Backported to v4.x at https://github.com/JabRef/jabref/commit/3f73e9ecc05842ef1b2f2a31b51e8c168d7dfcc5.

PR #4100, opened 2018-06-05
Migrate Review field in entry preview to comment

### LINKED ISSUE #4098: Comments are not shown in entry preview
- [X ] I have tested the latest master version from http://builds.jabref.org/master/ and the problem persists

JabRef version 4.3 on Ubuntu 18.04

Steps to reproduce:

1. Open entry preview on a entry with comments
2. Comments are not shown (older versions have shown the comments field)

How to fix it:
Change following line in the prefix preferences:
```
\begin{review}<BR><BR><b>Review: </b> \format[HTMLChars]{\review} \end{review}</dd>
```
to
```
\begin{comment}<BR><BR><b>Comment: </b> \format[HTMLChars]{\comment} \end{comment}</dd>
```

I am reporting this, because I think that all users are affected.

Thanks for your great work!

Siedlerchr:
Migrate Review field in entry preview to comment
Replace review with comment in every layout file
Fixes #4098 

----


Siedlerchr:
Added a test case

src/main/java/org/jabref/preferences/JabRefPreferences.java
```diff
@@ -366,6 +366,15 @@
     // Id Entry Generator Preferences
     public static final String ID_ENTRY_GENERATOR = "idEntryGenerator";
 
+    // Preview
+    //public because needed for pref migration
+    public static final String PREVIEW_STYLE = "previewStyle";
```
tobiasdiez: You can simply add a get/setPreviewStyle method which I would prefer.

src/main/java/org/jabref/migrations/PreferencesMigrations.java
```diff
@@ -297,20 +299,27 @@ private static void upgradeObsoleteLookAndFeels() {
...
+                  } else {
+                      String nimbusLandF = "javax.swing.plaf.nimbus.NimbusLookAndFeel";
+                      prefs.put(JabRefPreferences.WIN_LOOK_AND_FEEL, nimbusLandF);
+                      LOGGER.info("Switched from obsolete look and feel " + currentLandF + " to " + nimbusLandF);
+                  }
+              });
+    }
+
+    static void upgradePreviewStyleFromReviewToComment() {
+        JabRefPreferences prefs = Globals.prefs;
```
tobiasdiez: You can pass the preferences class as a parameter and then use `mock` in combination with `verify` in the tests below (instead of working on the actual preference object).

src/test/java/org/jabref/migrations/PreferencesMigrationsTest.java
```diff
@@ -16,9 +16,49 @@
     private JabRefPreferences prefs;
 
     private final String[] oldStylePatterns = new String[] {"\\bibtexkey",
-            "\\bibtexkey\\begin{title} - \\format[RemoveBrackets]{\\title}\\end{title}"};
+                                                            "\\bibtexkey\\begin{title} - \\format[RemoveBrackets]{\\title}\\end{title}"};
     private final String[] newStylePatterns = new String[] {"[bibtexkey]",
-            "[bibtexkey] - [fulltitle]"};
+                                                            "[bibtexkey] - [fulltitle]"};
+
+    private final String oldPreviewStyle = "<font face=\"sans-serif\">"
```
tobiasdiez: Can you please define these as local variables in the test below and try to simplify it a bit by removing unimportant lines.

src/main/java/org/jabref/migrations/PreferencesMigrations.java
Siedlerchr: I tried around with this but I currently see no way to do this. When I mock the object, calling the migration method on the object will have no effect as it operates directly on the preferences object. I don't like that operating on the real prefs object but 
Sure I can mock returns, but that's not the goal of this test. 
However, I could remove the dependency on the actual Globals.prefs object by passing a Preferences object as parametre

Siedlerchr:
I would vote for merging this, I refactored the tests and the migrations so that they no longer call  Globals.prefs. _ More mocking is not really possible.

src/test/java/org/jabref/migrations/PreferencesMigrationsTest.java
```diff
@@ -86,4 +84,57 @@ public void testArbitraryBibtexkeyPattern() {
...
+                                 + "\\begin{institution} <em>\\format[HTMLChars]{\\institution}, </em>\\end{institution}__NEWLINE__"
+                                 + "\\begin{publisher} <em>\\format[HTMLChars]{\\publisher}, </em>\\end{publisher}__NEWLINE__"
+                                 + "\\begin{year}<b>\\year</b>\\end{year}\\begin{volume}<i>, \\volume</i>\\end{volume}"
+                                 + "\\begin{pages}, \\format[FormatPagesForHTML]{\\pages} \\end{pages}__NEWLINE__"
+                                 + "\\begin{abstract}<BR><BR><b>Abstract: </b> \\format[HTMLChars]{\\abstract} \\end{abstract}__NEWLINE__"
+                                 + "\\begin{comment}<BR><BR><b>Comment: </b> \\format[HTMLChars]{\\comment} \\end{comment}"
+                                 + "</dd>__NEWLINE__<p></p></font>";
+
+        String previousStyle = prefs.getPreviewStyle();
+        prefs.setPreviewStyle(oldPreviewStyle);
```
tobiasdiez: I don't see a reason why you shouldn't be able to mock the preferences:
````java
preferences = mock(JabRefPreferences);
when(preferences.getPreviewStyle).thenReturn(oldPreviewStyle);;
verify(preferences).setPreviewStyle(newPreviewStyle);
```

Siedlerchr: Got it now, I didn't know that verify is interpreted as "assert" and makes the test fail/succeed

PR #4107, opened 2018-06-05
Fix a few things in entry editor

### LINKED ISSUE #4104: Moving to the next entry while in entry editor no longer supported
- [+] I have tested the latest master version from http://builds.jabref.org/master/ and the problem persists
JabRef 4.4-dev--snapshot--2018-06-04--master--ee669a542
Windows 7 6.1 amd64 
Java 1.8.0_172

In 4.1 I could move in the entry editor to the next entry using Ctrl-Shift ↓ (the latter from the numeric key pad. That it must be the numeric key pad is not mentioned in the help.
However with the latest snapshot this changing entries is no langer supported. Within the field Ctrl Shift ↓ or Ctrl Shift ↑ move to the end or beginning of that field. Moving to the next or previous entry seems no longer supported.

tobiasdiez:
Fixes two things:
- Reactivate next/previous entry shortcuts in entry editor (Fixes #4104)
- Preserve focus of text fields in entry editor when the user switches between entries

Please merge and don't squash.

----


CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - In the main table, the context menu appears now when you press the "context menu" button on the keyboard. [feature request in the forum](http://discourse.jabref.org/t/how-to-enable-keyboard-context-key-windows)
 - We added icons to the group side panel to quickly switch between `union` and `intersection` group view mode https://github.com/JabRef/jabref/issues/3269.
 - We use `https` for [fetching from most online bibliographic database](https://help.jabref.org/en/#-using-online-bibliographic-database).
+- We changed the default keyboard shortcuts for moving between entries when the entry editor is active to ̀ alt + up/down`.
```
Siedlerchr: Use kbd tags

src/main/java/org/jabref/gui/entryeditor/EntryEditorTab.java
```diff
@@ -39,7 +38,7 @@ public void notifyAboutFocus(BibEntry entry) {
         if (!entry.equals(currentEntry) || !currentEntryType.equals(entry.getType())) {
             currentEntry = entry;
             currentEntryType = entry.getType();
-            DefaultTaskExecutor.runInJavaFXThread(() -> bindToEntry(entry));
```
Siedlerchr: I am not sure if you can already do this. As far as I remember the import inspection or other swing dialogs will raise an fx thread error

tobiasdiez: This is necessary so that the entry editor is always bound to correct entry before other events hit it. If there are problems with certain dialogs, then those need to be fixed (call runInJavaFXThread there). I've not yet noticed any issues.

CHANGELOG.md
tobiasdiez: done

Issue #3838, opened 2018-03-12
docbook export skips species names that are in braces


bernhard-kleine:
- [ ] I have tested the latest master version from http://builds.jabref.org/master/ and the problem persists

JabRef version 4.1

Steps to reproduce:

1. 
this is the citation:

> Article (AW06) Audsley, N. & Weaver, R. J. Analysis of peptides in the brain and corpora cardiaca-corpora allata of the honey bee, {A}pis mellifera using MALDI-TOF mass spectrometry. Peptides, 2006, 27, 512-20.

2. Using a line like this
`java -jar C:\Programme\JabRe4.1\JabRef-4.1.jar -o cha13addBiB.xml,docbook -n true addkap13.bib`

3. I get

> <biblioentry xreflabel="Audsley und Weaver (2006)" xml:id="AW06">
>    <authorgroup>
>        <author><personname><firstname>N.</firstname><surname>Audsley</surname></personname></author>
>        <author><personname><firstname>R. J.</firstname><surname>Weaver</surname></personname></author> 
> 
>    </authorgroup>
>    <citetitle pubwork="article">Analysis of peptides in the brain and corpora cardiaca-corpora allata of the honey bee, mellifera using MALDI-TOF mass spectrometry.</citetitle>
>    <citetitle pubwork="journal">Peptides</citetitle>
> 
>    <volumenum>27</volumenum> 
> 
>    <artpagenums>512–20</artpagenums> 
>    <pubdate>2006</pubdate>  
> 
> 
>    </abstract>
> </biblioentry>

Note that the {A}pis is left out. This is valid not only for this example, but about 50% of my citations are errorneous.

stefan-kolb:
@johannes-manner Maybe you can help here, as you just worked on some exporters.

johannes-manner:
I tried to reproduce the bug, but I get a different xml-result.
Please share the .bib and .xml files with me for a further investigation of this bug.

bernhard-kleine:
I have tested this again and still some instances of braced e.g.
Drosophila are removed.

I used the following line to start the program

java -jar C:\Programme\JabRef4.2Test\JabRef-4.2Test.jar -o
testbraces.xml,docbook -n k -n <d h~hB g> k -n true testbraces.bib

JabRef4.2Test contains the files in the image
In that period I downloaded the following jabref install files:


.bib file and .xml file are attached.
Thanks for looking into this.

Bernhard
[...]

PR #4108, opened 2018-06-06
Docbook exporter

johannes-manner:
References #3838.

The problem are LaTeX commands, which are not resolved before executing special formatters for all fields. Therefore the presented solution is quite easy, but I cannot assess, if this solution has any side effects or if there are cases, where the LaTeX command should be also present after formatting. 

----


src/main/java/org/jabref/JabRefMain.java
```diff
@@ -186,4 +186,4 @@ private static void configureProxy(ProxyPreferences proxyPreferences) {
             Authenticator.setDefault(new ProxyAuthenticator());
         }
     }
-}
+}
```
koppor: Please keep the newline at the end of the file. Just unstage this file from the commit.

tobiasdiez:
Since there are formatters that need the latex commands to be present (e.g. because the translate latex to unicode), this solution sadly does not work. As this PR is stale for quite some time now, I'll close it now. Feel free to update the code and reopen the PR.

PR #4109, opened 2018-06-06
Smooth shutdown

tobiasdiez:
I tried to remove the `System.exit` call so that JabRef shutdown smoothly. However, currently an AWT-thread does not want to exit. In order to solve this problem, I removed most calls that would init the AWT toolkit. This includes:
- Use JavaFX clipboard instead of AWT
- Remove L&F and font customization

Moreover, the Azure telemetry SDK has a bug https://github.com/Microsoft/ApplicationInsights-Java/issues/662 that prevents certain threads from terminating. I added the workaround mentioned in this issue.

----


src/main/java/org/jabref/JabRefGUI.java
```diff
@@ -242,46 +233,11 @@ private boolean isLoaded(File fileToOpen) {
...
-            for (Object key : Collections.list(keys)) {
-                if ((key instanceof String) && ((String) key).endsWith(".font")) {
-                    Font font = (Font) UIManager.get(key);
-                    font = new FontUIResource(font.getName(), font.getStyle(), fontSize);
-                    defaults.put(key, font);
-                }
-            }
+        // On Linux, Java FX fonts look blurry per default. This can be improved by using a non-default rendering
+        // setting. See https://github.com/woky/javafx-hates-linux
+        if (Globals.prefs.getBoolean(JabRefPreferences.FX_FONT_RENDERING_TWEAK)) {
```
stefan-kolb: Is there a Linux switch anywhere?

src/main/java/org/jabref/gui/ClipBoardManager.java
```diff
@@ -22,24 +16,28 @@
...
+import org.jabref.model.util.OptionalUtil;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class ClipBoardManager implements ClipboardOwner {
+public class ClipBoardManager {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(ClipBoardManager.class);
+    public static final DataFormat XML = new DataFormat("application/xml");
```
stefan-kolb: Is this used anywhere?

src/main/java/org/jabref/gui/GUIGlobals.java
```diff
@@ -20,12 +20,10 @@
     public static Color activeBackgroundColor;
     public static Color invalidFieldBackgroundColor;
     public static Font currentFont;
-    public static final Color NULL_FIELD_COLOR = new Color(75, 130, 95); // Valid field, green.
 
-    public static final Color ACTIVE_EDITOR_COLOR = new Color(230, 230, 255);
-    public static final int WIDTH_ICON_COL = JabRefPreferences.getInstance().getInt(JabRefPreferences.ICON_SIZE_SMALL) + 12; // add some additional space to improve appearance
```
stefan-kolb: JabRefPreferences.ICON_SIZE_SMALL not used anymore?

src/main/java/org/jabref/gui/actions/CopyVersionToClipboardAction.java
```diff
@@ -1,23 +0,0 @@
...
-
-import javax.swing.AbstractAction;
-
-import org.jabref.Globals;
-import org.jabref.JabRefGUI;
-import org.jabref.gui.ClipBoardManager;
-import org.jabref.logic.l10n.Localization;
-import org.jabref.logic.util.BuildInfo;
-
-public class CopyVersionToClipboardAction extends AbstractAction {
```
stefan-kolb: Functionality totally removed?

src/test/java/org/jabref/gui/ClipBoardManagerTest.java
```diff
@@ -1,70 +0,0 @@
...
-import org.junit.jupiter.api.Test;
-import org.mockito.ArgumentMatchers;
-
-import static org.junit.Assert.assertEquals;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.when;
-
-public class ClipBoardManagerTest {
-
-    private ClipBoardManager clipBoardManager;
```
stefan-kolb: What about the tests?

src/main/java/org/jabref/JabRefGUI.java
tobiasdiez: Yes, it is turned on by default only on Linux: `defaults.put(FX_FONT_RENDERING_TWEAK, OS.LINUX);`

src/main/java/org/jabref/gui/ClipBoardManager.java
tobiasdiez: Yes, when we put xml in the clipboard (which may happen for copied citation styles)

tobiasdiez: Yes, when we put xml in the clipboard (which may happen for copied citation styles)

src/main/java/org/jabref/gui/actions/CopyVersionToClipboardAction.java
tobiasdiez: There is still a "copy version" button in the about dialog but no separate menu entry. (I didn't change anything as part of this PR; this class was completely unused).

src/test/java/org/jabref/gui/ClipBoardManagerTest.java
tobiasdiez: The clipboard manager needs to run on the JavaFX thread to work, which is not easily achieved in a test. Thus, deleting these tests was easier than trying to rewrite them ;-)

Siedlerchr:
I created a branch from master, cherry picked 39e30cc, and deleted all system.exit() calls. 
It works fine for me. For reference, I have create a PR to show the diff:  #4111

src/main/java/org/jabref/gui/ClipBoardManager.java
Siedlerchr: We have already a class DragAndDropDataFormats, maybe you could rename that and add it to the collection?

tobiasdiez: I don't want to interfere with your PR (especially, since I think it might completely remove the `DragAndDropFormats` class or make it internal to the `DragAndDropManager`).

src/main/java/org/jabref/Globals.java
```diff
@@ -90,10 +92,13 @@ public static void startBackgroundTasks() {
...
     private static void stopTelemetryClient() {
-        if (Globals.prefs.shouldCollectTelemetry()) {
-            getTelemetryClient().ifPresent(client -> client.trackSessionState(SessionState.End));
-            getTelemetryClient().ifPresent(client -> client.flush());
-        }
+        getTelemetryClient().ifPresent(client -> {
+            client.trackSessionState(SessionState.End);
+            client.flush();
+
+            // Workaround for bug https://github.com/Microsoft/ApplicationInsights-Java/issues/662
```
stefan-kolb: Add FIXME so we can revert this after they merged the bug fix.

Siedlerchr: I really would use my PR as a base. As this done does more than the initial necessary solution.

tobiasdiez: I don't understand the complication (including merge conflicts) but well it doesn't really matter. @Siedlerchr @stefan-kolb can I merge this PR?

stefan-kolb: Guess so.

Siedlerchr:
And removing setting the font size without providing an alternative was not the good idea!

stefan-kolb:
Just revert If necessary.

Christoph <notifications@github.com> schrieb am Do., 14. Juni 2018, 12:28:

tobiasdiez:
Sorry, I was so used to failing tests against the maintable-branch that I ignored the red travis icon. Tests are now fixed in master.

PR #4111, opened 2018-06-06
Fix open thread prevents shutdown

Siedlerchr:
@tobiasdiez  As this is just contains the relevant fix and works for me, I would be in favor for using this PR. You could still have your PR as refactoring PR then

----

### DIFF
--- a/src/main/java/org/jabref/Globals.java
+++ b/src/main/java/org/jabref/Globals.java
@@ -95,7 +95,7 @@ public class Globals {
             client.trackSessionState(SessionState.End);
             client.flush();
 
-            // Workaround for bug https://github.com/Microsoft/ApplicationInsights-Java/issues/662
+            //FIXME: Workaround for bug https://github.com/Microsoft/ApplicationInsights-Java/issues/662
             SDKShutdownActivity.INSTANCE.stopAll();
         });
     }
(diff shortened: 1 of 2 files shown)


tobiasdiez:
Yes, I also noted that this is enough to fix the shutdown issue. I nonetheless think that getting rid of most of the AWT stuff is desirable. It's good that you highlight the essential part of #4109 that fixes the shutdown problem but I don't think we need two PR's for this (especially since this leads to merge problems...).

src/main/java/org/jabref/Globals.java
```diff
@@ -90,10 +91,13 @@ public static void startBackgroundTasks() {
...
     private static void stopTelemetryClient() {
-        if (Globals.prefs.shouldCollectTelemetry()) {
-            getTelemetryClient().ifPresent(client -> client.trackSessionState(SessionState.End));
-            getTelemetryClient().ifPresent(client -> client.flush());
-        }
+        getTelemetryClient().ifPresent(client -> {
+            client.trackSessionState(SessionState.End);
+            client.flush();
+
+            // Workaround for bug https://github.com/Microsoft/ApplicationInsights-Java/issues/662
```
stefan-kolb: Maybe add a FIXME here, so we can revert this when the bug fix is merged

Siedlerchr: done

Siedlerchr:
I am going to merge this now.

PR #4116, opened 2018-06-07
Don't report journal name as abbreviated when full name = abbreviated name

### LINKED ISSUE #4115: Single word journal titles show up with an alert
Not sure when this problem was introduced, but in JabRef 4.3.1, single-word journal titles show up with an orange alert triangle as if they were incorrectly formatted:

jabref-crop

Perhaps this is an issue with checking for correct case?  The problem goes away if you change the entry to "Bioinformatics X" or something.

Not a huge deal, but kind of distracting to see "false positive" alerts like this.

tobiasdiez:
Fixes #4115. 

----


tobiasdiez:
Thanks for the feedback. Your solution was my initial idea but then went on to modify the `JournalRepository` because other (potential) callers of the repository also profit from the change. I wouldn't consider "Bioinformatics" abbreviated when it's also the full journal name (which is of course a question of definition).

In the end it's almost a question of taste and I'm also fine changing it to your solution (if you think the additional work is worth it ;-) ).

PR #4119, opened 2018-06-08
[WIP] Improve start-up time

tobiasdiez:
I played around a bit with the aim of improving the startup time. Ignore the code for the moment.
One of the biggest contribution comes from `ClassLoader.loadClass`:
image
@JabRef/developers Any idea why class loading is so slow? Do we mess with the classpath at some point or do other customization in this regard?

----

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -137,9 +137,9 @@ dependencies {
 
     // >1.8.0-beta is required for java 9 compatibility
     compile 'org.slf4j:slf4j-api:1.8.0-beta2'
-    compile 'org.apache.logging.log4j:log4j-slf4j-impl:2.11.0'
-    compile 'org.apache.logging.log4j:log4j-jcl:2.11.0'
-    compile 'org.apache.logging.log4j:log4j-api:2.11.0'
+    //compile 'org.apache.logging.log4j:log4j-slf4j-impl:2.11.0'
+    //compile 'org.apache.logging.log4j:log4j-jcl:2.11.0'
+    //compile 'org.apache.logging.log4j:log4j-api:2.11.0'
     compile 'org.apache.logging.log4j:log4j-core:2.11.0'
 
     // need to use snapshots as the stable version is from 2013 and doesn't support v1.0.1 CitationStyles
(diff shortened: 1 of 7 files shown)


Siedlerchr:
From the final call it seems the logger searches all class files for a given name.  But i don't know if that is the real bottleneck here. 
https://github.com/apache/logging-log4j2/blob/master/log4j-core/src/main/java/org/apache/logging/log4j/core/config/plugins/util/ResolverUtil.java

Siedlerchr:
I also checked the performance with the profiler in the branch for #4068 and noticed a few things (ensured to kill any previous instance int task mgr)
- Telemetry client throttles startup 
- Version Worker  -> Old Swing Task. Could be simplified as Task, only display a dialog.  Furthermore in the version checker, there seems to be still a HTTP connection open, at least a Keep Alive Cache thread is running the background 

- Latex2Unicode: The first initialization does either happen in the Maintable or in the Groups Pane and boils down to the fast parse -> Maybe we can init Latex2Unicode in a background task?
 
- Somehow it makes a difference if I have the entry open before I exit JabRef and restart it, or not.  Don't know why, because the entry editor is not shown when I restart JabRef

koppor:
Refs https://github.com/JabRef/jabref/issues/2966 and https://github.com/JabRef/jabref/pull/3015

tobiasdiez:
@Siedlerchr Nice observations. I already removed some of the performance bottlenecks (like the one related to telemetry). But in the end, these are minor things improving the startup time only be a few ms whereas the ClassLoader currently consumes a few seconds...

@koppor I think the high numbers for the logger are more a symptom than the reason for the bad performance. For some reason, every call to `ClassLoader.loadClass` takes at least 60 ms (sometimes even half a second). I'm aware that this method might not be the fastest, but I shouldn't take that long. Thus, also other operations like `Stage.show()` are slower than usual. However, I've no experience with these kind of issues (i.e. I'm not sure how `loadClass` is supposed to work vs how it actually works right now, etc). Most problems discussed online in relation to `ClassLoader` seem to be related to memory leaks...

Maybe our Java gurus have an idea? @stefan-kolb @lenhard @LinusDietz @simonharrer ?

tobiasdiez:
This is now obsolete. In the year since this PR I also learned a thing: it's not the class loader that takes so long but the actual initialization of the class (which is however not shown that way in the profiler output).

PR #4127, opened 2018-06-13
 Fix #4033: order of fields in customized entry types is saved correctly 

### LINKED ISSUE #4033: Customize Entry Types
In jabref 4.2, and the beta 4.3 customized entry types do no show entry fields in the correct order in the entry editor. If I, for example customize the entry type @article with other/additional fields in "optional fields" the changes are take effect BUT fields appear in different order. This is also true for the customize entry type window after closing and restarting jabref.

Steps to reproduce:

1. customize entry type article (optional fields)
2. restart jabref
3. open any @article entry in the entry editor -> optional fields
----------
In the screenshots you can see the (random?) order established after closing/restarting jabref. Origially, I customized the following order: bibtexkey, author, gender, title, shorttitle, subtitle, journalname, journal, ....

---------
2018-05-15 15_23_27-jabref - c__users_dell_documents_bsv_bsv_bibliographie bib biblatex mode
2018-05-15 15_24_18-customize entry types

tobiasdiez:
----

### DIFF
--- a/src/main/java/org/jabref/gui/customentrytypes/EntryTypeCustomizationDialog.java
+++ b/src/main/java/org/jabref/gui/customentrytypes/EntryTypeCustomizationDialog.java
@@ -8,6 +8,7 @@ import java.awt.GridLayout;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -276,11 +277,13 @@ public class EntryTypeCustomizationDialog extends JabRefDialog implements ListSe
                 if (biblatexMode) {
                     Set<String> oldPrimaryOptionalFieldsLists = oldType.get().getPrimaryOptionalFields();
                     Set<String> oldSecondaryOptionalFieldsList = oldType.get().getSecondaryOptionalFields();
-                    if (oldRequiredFieldsList.equals(requiredFieldsList) && oldPrimaryOptionalFieldsLists.equals(optionalFieldsList) &&
-                            oldSecondaryOptionalFieldsList.equals(secondaryOptionalFieldsLists)) {
+                    if (Arrays.equals(oldRequiredFieldsList.toArray(), requiredFieldsList.toArray())
+                            && Arrays.equals(oldPrimaryOptionalFieldsLists.toArray(), optionalFieldsList.toArray())
+                            && Arrays.equals(oldSecondaryOptionalFieldsList.toArray(), secondaryOptionalFieldsLists.toArray())) {
                         changesMade = false;
                     }
-                } else if (oldRequiredFieldsList.equals(requiredFieldsList) && oldOptionalFieldsList.equals(optionalFieldsList)) {
+                } else if (Arrays.equals(oldRequiredFieldsList.toArray(), requiredFieldsList.toArray())
+                        && Arrays.equals(oldOptionalFieldsList.toArray(), optionalFieldsList.toArray())) {
                     changesMade = false;
                 }
             }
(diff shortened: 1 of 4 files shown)


src/main/java/org/jabref/gui/customentrytypes/EntryTypeCustomizationDialog.java
```diff
@@ -276,11 +277,13 @@ private void applyChanges() {
                 if (biblatexMode) {
                     Set<String> oldPrimaryOptionalFieldsLists = oldType.get().getPrimaryOptionalFields();
                     Set<String> oldSecondaryOptionalFieldsList = oldType.get().getSecondaryOptionalFields();
-                    if (oldRequiredFieldsList.equals(requiredFieldsList) && oldPrimaryOptionalFieldsLists.equals(optionalFieldsList) &&
-                            oldSecondaryOptionalFieldsList.equals(secondaryOptionalFieldsLists)) {
+                    if (Arrays.equals(oldRequiredFieldsList.toArray(), requiredFieldsList.toArray())
```
stefan-kolb: Is the reason for the bug that array.equals != list.equals?

Siedlerchr: This looks really odd, normally the equals should work here:
https://docs.oracle.com/javase/8/docs/api/java/util/List.html#equals-java.lang.Object-

tobiasdiez: The contract for sets is that equals does not take the order into account (even if it is a linkedhashset). This is in contrast to equality of lists (which however allow duplication). On the other hand, arrays equals takes the order into account. The cleanest solution would be to use a data structure that is ordered by design and prevents duplication; but I'm not aware of such an implementation (as said above LinkedHashSet does not behave as an ordered set, it just provides an ordered *iterator*).

Siedlerchr: You could use a Tree Set, I used that in tests in another program to ensure that the output is correct every time https://docs.oracle.com/javase/7/docs/api/java/util/TreeSet.html

tobiasdiez: TreeSet still inherits equals from set (as does LinkedHashSet which is usually the better choice than TreeSet).

Siedlerchr: No, it does not, A TreeSet implements Comparable and uses the method of the compareTo 

>Note that the ordering maintained by a set (whether or not an explicit comparator is provided) must be consistent with equals if it is to correctly implement the Set interface. (See Comparable or Comparator for a precise definition of consistent with equals.) This is so because the Set interface is defined in terms of the equals operation, but a TreeSet instance performs all element comparisons using its compareTo (or compare) method, so two elements that are deemed equal by this method are, from the standpoint of the set, equal. The behavior of a set is well-defined even if its ordering is inconsistent with equals; it just fails to obey the general contract of the Set interface. 

https://dzone.com/articles/the-hidden-contract-between-equals-and-comparable

tobiasdiez: As you see here http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/8ed8e2b4b90e/src/share/classes/java/util/TreeSet.java the class does not override the default implementation of `equals` of http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/8ed8e2b4b90e/src/share/classes/java/util/AbstractSet.java. The comment you cite concerns equality of items not of the set itself.

Siedlerchr: This really seems to be only alternative.

Siedlerchr:
If you fix the failing tests, it's good!

PR #4133, opened 2018-06-17
Fix SNAPSHOT releases marked as outdated

LinusDietz:
- Fixes that all the SNAPSHOT releases are shown in the dependencyUpdates check https://github.com/JabRef/jabref/pull/4131
- Updates two dependencies

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -16,7 +16,7 @@ buildscript {
 
 plugins {
     id 'com.gradle.build-scan' version '1.14'
-    id 'com.install4j.gradle' version '7.0.4'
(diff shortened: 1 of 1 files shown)


Siedlerchr:
Merged too early. https://travis-ci.org/JabRef/jabref/jobs/393346860 
I think we would need your switch for all libraries except junit-pioneer. Maybe one can combine the rules?  e.g.  snapshot && not in (junit pioneer, javax inject?)

 Failed to determine the latest version for the following dependencies (use --info for details):
 - javax.inject:javax.inject
 - org.junit-pioneer:junit-pioneer

PR #4136, opened 2018-06-17
add exception for junit-pioneer in version rejection of snapshots

Siedlerchr:
I now added an exception for junit-pioneer  version checker. All other snapshots are rejected.

----

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -188,13 +188,18 @@ dependencyUpdates.revision = 'integration'
 dependencyUpdates.resolutionStrategy = {
     componentSelection { rules ->
         rules.all { ComponentSelection selection ->
-            if ( selection.candidate.module != "junit-pioneer" && selection.candidate.version ==~ /[0-9].*SNAPSHOT/ ) {
+            if ( selection.candidate.module != "junit-pioneer" && selection.candidate.module!="javax.inject"  && selection.candidate.version ==~ /[0-9].*SNAPSHOT/ ) {
                 selection.reject("Ignore SNAPSHOT releases")
             }
         }
         rules.withModule("org.controlsfx:controlsfx") { ComponentSelection selection ->
             if (selection.candidate.version ==~ /9.*/) { // Reject version 9 or higher
                  selection.reject("Cannot be updated to 9.*.* until Jabref works with Java 9")
+            }
+        }
+		rules.withModule("com.github.tomtung:latex2unicode_2.12") { ComponentSelection selection ->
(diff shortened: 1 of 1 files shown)


Siedlerchr:
@LinusDietz with exception I mean that junit-pioneer does not get rejected, because it's only available as snapshot. Same applies to javax:inject

if you run ./gradlew dU it works correclty

build.gradle
```diff
@@ -188,14 +188,19 @@ dependencyUpdates.revision = 'integration'
...
                 selection.reject("Ignore SNAPSHOT releases")
             }
         }
         rules.withModule("org.controlsfx:controlsfx") { ComponentSelection selection ->
             if (selection.candidate.version ==~ /9.*/) { // Reject version 9 or higher
                  selection.reject("Cannot be updated to 9.*.* until Jabref works with Java 9")
             }
+        }
+		rules.withModule("com.github.tomtung:latex2unicode_2.12") { ComponentSelection selection ->
+            if (selection.candidate.version ==~ /0.2.2/) { // Reject version higher than 2.0.2
```
LinusDietz: `selection.candidate.version == 0.2.4` would be correct. (you have to put in the version to be rejected, not the current.

build.gradle
```diff
@@ -188,14 +188,19 @@ dependencyUpdates.revision = 'integration'
...
+                 selection.reject("Cannot be updated to 0.2.4 until JabRef is prepared for it")
```
LinusDietz: 0.2.4 -> 2.2.4

build.gradle
```diff
@@ -188,14 +188,19 @@ dependencyUpdates.revision = 'integration'
 dependencyUpdates.resolutionStrategy = {
     componentSelection { rules ->
         rules.all { ComponentSelection selection ->
-            if (selection.candidate.version ==~ /[0-9].*SNAPSHOT/) {
+            if ( selection.candidate.module != "junit-pioneer" && selection.candidate.module!="javax.inject"  && selection.candidate.version ==~ /[0-9].*SNAPSHOT/ ) {
```
LinusDietz: javax.inject -> javax.inject:javax.inject ?

Siedlerchr: I initially had the same idea but my tests with junit pioneer showed that it didn't work. I then added a print statement and saw that the module is just the real module name
```
 componentSelection { rules ->
        rules.all { ComponentSelection selection ->
print selection.candidate.module 
```

Siedlerchr:
As this works as expected I merge this now,

PR #4137, opened 2018-06-17
Fix #3805: dash correctly terminates commands in layout files

### LINKED ISSUE #3805: Can't use a dash after a formatter
- I have not tested the latest master version from http://builds.jabref.org/master/ as the gui doesn't start.

JabRef version 4.1 on fedora 27

Steps to reproduce:

1. Use a style with a formatter \edition like
``` \begin{edition} \edition-th ed.- \end{edition}```
(I actually need a russian variant, the -th variant is here just for example)
2. The result is that only 'ed.' is printed
3. ...

<details>
  <summary>Log File</summary>
  
  ```
  Paste an excerpt of your log file here
  ```
</details>

### PR DESCRIPTION
**Pull request by tobiasdiez**
Fixes #3805. Now commands terminated by a hyphen (like `\edition-`) work correctly.

Way back in 2011, Morten added the extra check for hyphens (commit 392bc7a6dd215eb23766378aa90e3995d0847bb7). However, I'm not aware of any field that has a dash in it, so I'm not sure what was the reason for the change (I couldn't find an old bug report).

Also refactored a bit the layout tests.

----

### CODE REVIEW

File: CHANGELOG.md [hunk 1]
```diff
@@ -31,6 +31,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - We fixed an issue where in rare cases entries where overlayed in the main table. https://github.com/JabRef/jabref/issues/3281
 - We fixed an issue where selecting a group messed up the focus of the main table / entry editor. https://github.com/JabRef/jabref/issues/3367
 - We fixed an issue where composite author names were sorted incorrectly. https://github.com/JabRef/jabref/issues/2828
+- We fixed an issue where commands followed by `-` didn't worked. [#3805](https://github.com/JabRef/jabref/issues/3805)
```
koppor: Please change "worked" into "work".

src/main/java/org/jabref/logic/layout/Layout.java
```diff
@@ -33,7 +33,7 @@ public Layout(List<StringInt> parsedEntries, LayoutFormatterPreferences prefs) {
         for (StringInt parsedEntry : parsedEntries) {
             switch (parsedEntry.i) {
             case LayoutHelper.IS_LAYOUT_TEXT:
-            case LayoutHelper.IS_SIMPLE_FIELD:
+                case LayoutHelper.IS_SIMPLE_COMMAND:
```
koppor: Please double check tabs vs. spaces there and the line before/after

src/main/java/org/jabref/logic/layout/LayoutEntry.java
```diff
@@ -109,7 +109,7 @@ public LayoutEntry(StringInt si, LayoutFormatterPreferences prefs) {
         case LayoutHelper.IS_LAYOUT_TEXT:
             text = si.s;
             break;
-        case LayoutHelper.IS_SIMPLE_FIELD:
+            case LayoutHelper.IS_SIMPLE_COMMAND:
```
koppor: Why the extra indent? (applies for all case statements in this file)

Issue #4126, opened 2018-06-13
Title should be single line field


stefan-kolb:
And wrap multiline paste data in a single line

`
SeaClouds
Project
D5.1.2
–
Integrated
Platform
`

-->

`SeaClouds Project D5.1.2 – Integrated Platform`

Naitoreivun:
I want to take a look on that :)

PR #4138, opened 2018-06-18
single line text fields

Naitoreivun:
Solution for [#4126](https://github.com/JabRef/jabref/issues/4126). I've used text formatter to force single line behavior. Hope it's enough :) Particular fields, that should contain only one line, are stored in immutable set.

### DIFF
--- a/src/main/java/org/jabref/gui/fieldeditors/FieldEditors.java
+++ b/src/main/java/org/jabref/gui/fieldeditors/FieldEditors.java
@@ -22,6 +22,7 @@ import org.jabref.model.entry.InternalBibtexFields;
 import org.jabref.model.metadata.MetaData;
 import org.jabref.preferences.JabRefPreferences;
 
+import com.google.common.collect.ImmutableSet;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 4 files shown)


stefan-kolb:
Can't we just change the underlying GUI component from TextArea to TextField?
Also please use plain Java for making fields immutable or final,  not com.google.common.collect.ImmutableSet if possible.

Naitoreivun:
@stefan-kolb how about now?

Siedlerchr:
That looks good so far for me! Haven't tested it yet, but codewise it looks good, (Although the final in front of method parameter is suerpfllous

Naitoreivun:
I use finals to avoid stupid bugs, like reassigning variable. With them my IDE automatically underline such stuff. Another thing is that I've decided to make code more readable, so I've put new lines between SimpleEditor constructor parameters (because there was one long line), and now Codacy yells about that. Should I extract those parameters to another class, like SimpleEditorParameters or sth?

src/main/java/org/jabref/gui/fieldeditors/FieldEditors.java
```diff
@@ -25,16 +28,40 @@
...
 
-    public static FieldEditorFX getForField(String fieldName, TaskExecutor taskExecutor, DialogService dialogService, JournalAbbreviationLoader journalAbbreviationLoader, JournalAbbreviationPreferences journalAbbreviationPreferences, JabRefPreferences preferences, BibDatabaseContext databaseContext, String entryType, SuggestionProviders suggestionProviders, UndoManager undoManager) {
+    private static final Set<String> SINGLE_LINE_FIELDS = Collections.unmodifiableSet(new HashSet<>(
+            Arrays.asList(TITLE, AUTHOR, YEAR, INSTITUTION)
+    ));
+
+    public static FieldEditorFX getForField(final String fieldName,
+                                            final TaskExecutor taskExecutor,
+                                            final DialogService dialogService,
+                                            final JournalAbbreviationLoader journalAbbreviationLoader,
```
Siedlerchr: You could extract the loader and the prefs parameter and directly pass the  JounralAbbrev Repo stuff which is just passed onto the field checkers later 
`journalAbbreviationLoader.getRepository(journalAbbreviationPreferences)`

src/main/java/org/jabref/gui/fieldeditors/FieldEditors.java
```diff
@@ -25,16 +28,40 @@
...
+                                            final UndoManager undoManager) {
         final Set<FieldProperty> fieldExtras = InternalBibtexFields.getFieldProperties(fieldName);
 
-        AutoCompleteSuggestionProvider<?> suggestionProvider = getSuggestionProvider(fieldName, suggestionProviders, databaseContext.getMetaData());
+        final AutoCompleteSuggestionProvider<?> suggestionProvider = getSuggestionProvider(fieldName, suggestionProviders, databaseContext.getMetaData());
+
+        final FieldCheckers fieldCheckers = new FieldCheckers(
+                databaseContext,
+                preferences.getFileDirectoryPreferences(),
+                journalAbbreviationLoader.getRepository(journalAbbreviationPreferences),
```
Siedlerchr: Here the abbrev repo stuff is passed

src/main/java/org/jabref/gui/fieldeditors/FieldEditors.java
Naitoreivun: done

src/main/java/org/jabref/gui/fieldeditors/FieldEditors.java
```diff
@@ -25,16 +27,39 @@
...
+import static org.jabref.model.entry.FieldName.INSTITUTION;
+import static org.jabref.model.entry.FieldName.TITLE;
+import static org.jabref.model.entry.FieldName.YEAR;
+
 public class FieldEditors {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(FieldEditors.class);
 
-    public static FieldEditorFX getForField(String fieldName, TaskExecutor taskExecutor, DialogService dialogService, JournalAbbreviationLoader journalAbbreviationLoader, JournalAbbreviationPreferences journalAbbreviationPreferences, JabRefPreferences preferences, BibDatabaseContext databaseContext, String entryType, SuggestionProviders suggestionProviders, UndoManager undoManager) {
+    private static final Set<String> SINGLE_LINE_FIELDS = Collections.unmodifiableSet(new HashSet<>(
```
tobiasdiez: Can you please add this list two `InternalBibtexFields` (this class needs to be refactored at some point, but for now it is good to have anything at one place).

src/main/java/org/jabref/gui/fieldeditors/FieldEditors.java
tobiasdiez: In principle, the user might change the journal abbreviations at any time. However, it should be fine to use the repository since the field editors are recreated when switching to the new entry.

tobiasdiez: In principle, the user might change the journal abbreviations at any time. However, it should be fine to use the repository since the field editors are recreated when switching to the new entry.

stefan-kolb:
Ping 😄

src/main/java/org/jabref/gui/fieldeditors/FieldEditors.java
Naitoreivun: Yep. Sorry for delay, I will do that on Saturday.

PR #4139, opened 2018-06-18
Convert DOAJ fetcher to new interface

tobiasdiez:
Previously, the DOAJ fetcher was able to get all results by pagination. This version is restricted to the first 20 entries; I think, this is enough for the moment.

----


stefan-kolb:
Should we really limit all fetchers to 20 results, even if they would support more?

Siedlerchr:
same architecture violation as in your Springer fetcher

tobiasdiez:
We have to specify some number, the default used by the API is 10. I've now increased the limit to 30. Do you think that suffices?

tobiasdiez:
ok, I completely agree with this: we should support pagination at some point. However, it was not easy to integrate it in the current infrastructure and thus I left it out for the moment...

PR #4140, opened 2018-06-18
Remove OAI2Fetcher

tobiasdiez:
The OAI2 fetcher is nowhere used and thus can be removed. It actually would be nice to have a fetcher that supports tho OAI2 interface (so that we may implement further fetchers more easily in the future). However, the current implementation still uses the old fetcher infrastructure and thus reenabling it would almost amount to a complete rewrite anyway.

----


Siedlerchr:
You have some obsolete language keys now, which results in failing test. apart from that lgtm

PR #4141, opened 2018-06-18
Convert Springer fetcher to new interface

tobiasdiez:
The Springer fetcher is now implemented according to the new fetcher infrastructure. Instead of pagination support, the fetcher now only returns the first 20 items matched by the search query. 

I used the opportunity and fixed a few small things: url to pdf is now added as online file link (instead of in the url field) and the ending page is correctly parsed.

----

### DIFF
--- a/src/main/java/org/jabref/gui/importer/fetcher/SpringerFetcher.java
+++ b/src/main/java/org/jabref/gui/importer/fetcher/SpringerFetcher.java
@@ -1,111 +1,38 @@
 package org.jabref.gui.importer.fetcher;
 
-import java.io.IOException;
-import java.net.URLEncoder;
+import java.io.BufferedReader;
+import java.io.InputStreamReader;
+import java.net.MalformedURLException;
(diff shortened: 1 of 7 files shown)


Siedlerchr:
Why do you have a dependeny to gui in the logic?

>     org.opentest4j.AssertionFailedError: The following classes are not allowed to depend on org.jabref.gui ==> expected: <[]> but was: <[src/main/java/org/jabref/logic/importer/WebFetchers.java]>

src/main/java/org/jabref/gui/importer/fetcher/SpringerFetcher.java
```diff
@@ -1,122 +1,73 @@
 package org.jabref.gui.importer.fetcher;
```
Siedlerchr: Why does it reside in gui?

tobiasdiez:
Forgot to move the fetcher to logic, good point.

PR #4144, opened 2018-06-19
Convert ACM fetcher to new infrastructure

tobiasdiez:
Converts the ACM fetcher to the new infrastructure and deletes a lot of gui code that is now no longer needed. Notable changes:
- Fetcher no longer visits each citation page separately but uses the batch export provided by ACM
- Fetcher always queries the full ACM Guide to Computing Literature (previously there was also the option to use the ACM Full-Text collection, which is however completely contained in the former).
- Abstracts are no longer returned (they are simply not exported by ACM)
- All fetcher now use the class `URLDownload` to make the request, and thereby pretend to be Firefox (otherwise we were blocked by ACM). I remember I had a discussion with @stefan-kolb recently about this, but don't remember the outcome. 

----


Siedlerchr:
I always liked abstracts. Is there no other way to get them?

tobiasdiez:
No sadly the only way to get the abstracts is to visit the page of every citation and extract it from the HTML. Maybe contact ACM and ask them to include the abstract in the bibtex export?

PR #4146, opened 2018-06-20
Convert INSPIRE fetcher to new infrastructure

tobiasdiez:
----


src/main/java/org/jabref/logic/importer/fetcher/INSPIREFetcher.java
```diff
@@ -0,0 +1,92 @@
...
+    @Override
+    public Parser getParser() {
+        // Inspire returns the BibTeX result embedded in HTML
+        // So we extract the BibTeX string from the <pre>bibtex</pre> tags and pass the content to the BibTeX parser
+        return inputStream -> {
+            String response = new BufferedReader(new InputStreamReader(inputStream)).lines().collect(Collectors.joining(OS.NEWLINE));
+
+            List<BibEntry> entries = new ArrayList<>();
+            BibtexParser bibtexParser = new BibtexParser(preferences, new DummyFileUpdateMonitor());
+            Pattern pattern = Pattern.compile("<pre>(?s)(.*)</pre>");
```
Siedlerchr: extract pattern.compile to a private final or static variable

tobiasdiez: As this is an anonymous class, it is effectively a constant. I don't see any advantage of making it a final variable of the outer class, but maybe I miss something...

src/main/java/org/jabref/logic/importer/fetcher/INSPIREFetcher.java
```diff
@@ -0,0 +1,92 @@
...
+    public HelpFile getHelpPage() {
+        return HelpFile.FETCHER_INSPIRE;
+    }
+
+    @Override
+    public URL getURLForQuery(String query) throws URISyntaxException, MalformedURLException, FetcherException {
+        URIBuilder uriBuilder = new URIBuilder(INSPIRE_HOST);
+        uriBuilder.addParameter("p", query); // Query
+        //uriBuilder.addParameter("jrec", "1"); // Start index (not needed at the moment)
+        uriBuilder.addParameter("rg", "100"); // Should return up to 100 items (instead of default 25)
```
koppor: I heared that we support 30 entries only, why is there 100 here?

tobiasdiez: It was previously set to 1000, but I wanted to make sure that we are not banned...

PR #4157, opened 2018-06-26
Remove Spin

florian-beetz:
Prepares #3421

This removes `AbstractWorker` completely (where Spin was mainly used). All `AbstractWorker`s were
migrated to `BaseAction` using `BackgroundTask`s. This also simplifies `BasePanel` at some points.

Most workers could easily be wrapped in a `BackgroundTask`, as all they do on the UI Thread is call `JabRefFrame.output("..")` which is already executed on the UI Thread internally.

Some workers were joined with their corresponding actions as all the actions did, was calling the workers.

----


src/main/java/org/jabref/gui/BasePanel.java
```diff
@@ -1834,4 +1736,82 @@ public void action() throws SaveException {
...
         }
     }
+
+    private class GenerateKeysAction implements BaseAction {
+
+        List<BibEntry> entries;
+        int numSelected;
+        boolean canceled;
+
+        // Run first, in EDT:
```
Siedlerchr: That comment is obsolete now, right?>
Could this be extracted to a new file?

florian-beetz: Done. You are right, the comment was a left-over.

src/main/java/org/jabref/gui/openoffice/OpenOfficePanel.java
```diff
@@ -383,27 +386,38 @@ private void connect(boolean autoDetect) {
...
+            BackgroundTask
+                    .wrap(this::createBibBase)
+                    .onSuccess(ooBase -> {
+                        this.ooBase = ooBase;
+                        SwingUtilities.invokeLater(() -> {
+                            finalProgressDialog.dispose();
+                            diag.dispose();
+                        });
+                        if (ooBase == null) {
+                            //throw connectException; // TODO
```
koppor: What is the TODO at this place?

stefan-kolb:
Nice work! @LinusDietz will review this and merge soon!

src/main/java/org/jabref/gui/actions/CleanupAction.java
```diff
@@ -39,6 +43,26 @@ public CleanupAction(BasePanel panel, JabRefPreferences preferences) {
     }
 
     @Override
+    public void action() {
+        init();
+        Platform.runLater(() -> {
```
tobiasdiez: Please use the `runInJavaFXThread` method of `DefaultTaskExectuor` (we want to separate calls that are needed because we still have interaction with legacy Swing and "real" async operations in JavaFX)

On a second thought, I'm actually wondering if this call to `runLater` is even necessary since the action should be invoked through the menu which is already on the JavaFX thread.

src/main/java/org/jabref/gui/actions/CleanupAction.java
```diff
@@ -39,6 +43,26 @@ public CleanupAction(BasePanel panel, JabRefPreferences preferences) {
...
+                return;
+            }
+            CleanupDialog cleanupDialog = new CleanupDialog(panel.getBibDatabaseContext(), preferences.getCleanupPreset());
+
+            Optional<CleanupPreset> chosenPreset = cleanupDialog.showAndWait();
+            chosenPreset.ifPresent(cleanupPreset ->
+                    BackgroundTask
+                            .wrap(() -> {
+                                cleanup(cleanupPreset);
+                                return null;
```
tobiasdiez: Please copy the changes from https://github.com/JabRef/jabref/pull/4119/files#diff-851966b1f29f8a595c4b262cf1f895bb (it may take a while until this PR is merged). Then these `return null` statements are no longer needed (in several places).

src/main/java/org/jabref/gui/actions/CleanupAction.java
```diff
@@ -39,6 +43,26 @@ public CleanupAction(BasePanel panel, JabRefPreferences preferences) {
...
+                            })
+                            .onSuccess(x -> SwingUtilities.invokeLater(this::showResults))
```
tobiasdiez: the `showResults` method uses the JavaFX dialog service, so it should probably not run on the swing thread.

src/main/java/org/jabref/gui/exporter/ExportCommand.java
```diff
@@ -91,51 +92,21 @@ private void export(Path file, FileChooser.ExtensionFilter selectedExtensionFilt
...
-        };
-
-        // Run the export action in a background thread:
-        exportWorker.getWorker().run();
-        // Run the update method:
-        exportWorker.update();
+                    return null;
+                })
+                .onSuccess(x -> frame.output(Localization.lang("%0 export successful", format.getDisplayName())))
+                .onFailure(exception -> Platform.runLater(() -> handleError(exception)))
```
tobiasdiez: this `runLater` shouldn't be necessary as `onSuccess` and `onFailure` are always run on the JavaFX thread (it would be nice if you could add this as a comment in BackgroundTask)

src/main/java/org/jabref/gui/exporter/ExportCommand.java
```diff
@@ -147,4 +118,17 @@ private static FileDialogConfiguration createExportFileChooser(ExporterFactory e
...
+        final String errorMessage;
+        if (ex.getMessage() == null) {
+            errorMessage = ex.toString();
+        } else {
+            errorMessage = ex.getMessage();
+        }
+
+        frame.output(Localization.lang("Could not save file.") + " - " + errorMessage);
+        // Need to warn the user that saving failed!
+        frame.getDialogService().showErrorDialogAndWait(Localization.lang("Save library"), Localization.lang("Could not save file.") + "\n" + errorMessage);
```
tobiasdiez: the `DialogService` has a method that directly accepts an exception so you don't need to compute it by hand above (and the errorMessage can be removed in the call to `output`).

src/main/java/org/jabref/gui/openoffice/DetectOpenOfficeInstallation.java
```diff
@@ -44,28 +51,27 @@ public DetectOpenOfficeInstallation(JDialog parent, OpenOfficePreferences prefer
...
-    }
-
-    @Override
-    public void run() {
-        foundPaths = autoDetectPaths();
+        try {
+            Object result = BackgroundTask.wrap(this::autoDetectPaths)
+                                     .onSuccess(x -> SwingUtilities.invokeLater(progressDialog::dispose))
+                                     .executeWith(Globals.TASK_EXECUTOR)
+                                     .get();
```
tobiasdiez: This looks a bit strange to me. The `get` makes the thread wait until the task is completely, so the operation is essentially synchronous. Its probably better to return the `Task<Boolean>` and let the caller decide if he wants to wait.

src/main/java/org/jabref/gui/openoffice/OpenOfficePanel.java
florian-beetz: Good catch, this was a left-over from the error handling before. The error case is now handled better.

src/main/java/org/jabref/gui/DuplicateSearch.java
```diff
@@ -2,225 +2,219 @@
...
-
-    public void run() {
         BasePanel panel = frame.getCurrentBasePanel();
-
         panel.output(Localization.lang("Searching for duplicates..."));
 
-        bes = panel.getDatabase().getEntries();
-        if (bes.size() < 2) {
+        List<BibEntry> entries = panel.getDatabase().getEntries();
+        dups = new LinkedBlockingQueue<>();
```
LinusDietz: How about initializing the list at object creation already, i.e., l 35?

src/main/java/org/jabref/gui/DuplicateSearch.java
```diff
@@ -2,225 +2,219 @@
...
         }
 
-        // Thread cancel option
-        // no synchronized used because no "really" critical situations expected
-        public void setFinished() {
-            finished = true;
-        }
+        SwingUtilities.invokeLater(() -> {
+            BasePanel panel = frame.getCurrentBasePanel();
+            final NamedCompound ce = new NamedCompound(Localization.lang("duplicate removal"));
```
LinusDietz: again avoid abbreviations

src/main/java/org/jabref/gui/DuplicateSearch.java
```diff
@@ -2,225 +2,219 @@
...
         }
 
-        @Override
-        public void update() {
-            DuplicateResolverDialog diag = new DuplicateResolverDialog(frame, one, two, dialogType);
-            diag.setVisible(true);
-            diag.dispose();
-            reply = diag.getSelected();
-            merged = diag.getMergedEntry();
+        public synchronized void replace(BibEntry replace1, BibEntry replace2, BibEntry replacement) {
```
LinusDietz: I would go with first and second here

src/main/java/org/jabref/gui/actions/WriteXMPAction.java
```diff
@@ -1,20 +1,266 @@
...
 public class WriteXMPAction extends SimpleCommand {
 
     private final BasePanel basePanel;
+    private OptionsDialog optDiag;
+
+    private Collection<BibEntry> entries;
+
+    private BibDatabase database;
+
+    private boolean goOn = true;
```
LinusDietz: how about `shouldCcontinue`?

src/main/java/org/jabref/gui/actions/WriteXMPAction.java
```diff
@@ -1,20 +1,266 @@
...
+import org.jabref.logic.xmp.XmpUtilWriter;
+import org.jabref.model.database.BibDatabase;
+import org.jabref.model.entry.BibEntry;
+
+import com.jgoodies.forms.builder.ButtonBarBuilder;
 
 public class WriteXMPAction extends SimpleCommand {
 
     private final BasePanel basePanel;
+    private OptionsDialog optDiag;
```
LinusDietz: Please spell it out.

src/main/java/org/jabref/gui/actions/CleanupAction.java
```diff
@@ -39,6 +39,20 @@ public CleanupAction(BasePanel panel, JabRefPreferences preferences) {
     }
 
     @Override
+    public void action() {
+        init();
+        if (canceled) {
```
LinusDietz: I know it's not from you, but it would be good to rename it to `isCancelled`.

src/main/java/org/jabref/gui/actions/GenerateKeysAction.java
```diff
@@ -0,0 +1,96 @@
...
+import org.jabref.model.entry.BibEntry;
+import org.jabref.preferences.JabRefPreferences;
+
+public class GenerateKeysAction implements BaseAction {
+
+    private final BasePanel panel;
+    private final DialogService dialogService;
+    private List<BibEntry> entries;
+    private int numSelected;
+    private boolean canceled;
```
LinusDietz: Here again: please start the names of booleans with is/has etc.

src/main/java/org/jabref/gui/actions/WriteXMPAction.java
```diff
@@ -1,20 +1,266 @@
...
+        basePanel.output(Localization.lang("Finished writing XMP for %0 file (%1 skipped, %2 errors).",
+                String.valueOf(entriesChanged), String.valueOf(skipped), String.valueOf(errors)));
     }
 
+    class OptionsDialog extends JabRefDialog {
+
+        private final JButton okButton = new JButton(Localization.lang("OK"));
+        private final JButton cancelButton = new JButton(Localization.lang("Cancel"));
+
+        private boolean canceled;
```
LinusDietz: Also here the `canceled` could be named more precisely: what should be canceled?

src/main/java/org/jabref/gui/exporter/ExportToClipboardAction.java
```diff
@@ -103,11 +100,11 @@ public void run() {
             ClipboardContent clipboardContent = new ClipboardContent();
             clipboardContent.putRtf(sb.toString());
             Globals.clipboardManager.setContent(clipboardContent);
-            message = Localization.lang("Entries exported to clipboard") + ": " + entries.size();
+            return Localization.lang("Entries exported to clipboard") + ": " + entries.size();
 
         } catch (Exception e) {
             LOGGER.error("Error exporting to clipboard", e); //To change body of catch statement use File | Settings | File Templates.
-            message = Localization.lang("Error exporting to clipboard");
+            return Localization.lang("Error exporting to clipboard");
```
LinusDietz: Very good that you return that directly now!

src/main/java/org/jabref/gui/exporter/SaveDatabaseAction.java
```diff
@@ -454,4 +450,12 @@ private boolean checkExternalModification() {
         // Return false as either no external database file modifications have been found or overwrite is requested any way
         return false;
     }
+
+    @Override
+    public void action() throws Exception {
```
LinusDietz: Can you please start the name of a method with a verb? Like `performAction`?

florian-beetz: This method is inherited from `BaseAction`.
I can rename the base method though if you wish.

PR #4162, opened 2018-07-03
Convert IEEE fetcher to new interface

tobiasdiez:
The IEEE fetcher now also fits in the new infrastructure and is converted to use the new IEEE API. The original fetcher performed a quite "sophisticated" cleanup of some fields. I'm not sure if this was needed because of scraping the webpage / a buggy BibTeX export or because the data is inherently faulty. After looking at the JSON response of some queries it appears that these cleanups are no longer necessary. 

----


src/main/java/org/jabref/logic/importer/util/JSONEntryParser.java
```diff
@@ -229,4 +231,67 @@ public static BibEntry parseSpringerJSONtoBibtex(JSONObject springerJsonEntry) {
 
         return entry;
     }
+
+    /**
+     * @implNote <a href="https://developer.ieee.org/docs/read/Metadata_API_responses">documentation</a>
+     */
+    public static BibEntry parseIEEEJSONtoBibtex(JSONObject jsonEntry, Character keywordSeparator) {
```
stefan-kolb: For me this belongs to the IEEE class.

tobiasdiez: Good point. Moved it to the fetcher (as well as the other parsing code in that class)

PR #4163, opened 2018-07-03
[WIP] Cleanup build script

florian-beetz:
This prepares for splitting JabRef into several modules.

The dependencies were extracted to a separate file. This makes it much easier in the future to define shared dependencies across multiple modules, without duplicating them in the build scripts and especially keeping them all updated to the same version.

According to Gradle's guide on [writing maintainable build scripts](https://docs.gradle.org/current/userguide/authoring_maintainable_build_scripts.html), build scripts should not contain any imperative logic but just declarative language elements. Imperative logic should live as plugins in the `buildSrc` directory. So I extracted the ANTLR tasks to this directory. This also makes it much easier to move the tasks to another module in the future.

What do you think? I would like to extract the XJC-, release-, and localization-tasks to separate plugins if you approve these changes.

----


buildSrc/build.gradle
```diff
@@ -0,0 +1,13 @@
+plugins {
+    id 'java'
+    id 'groovy'
+}
+
+sourceCompatibility = 1.8
+
+repositories {
+    mavenCentral()
+}
```
Siedlerchr: Isn't this now duplicated?

florian-beetz: This specifies the repositories used for resolving dependencies for the buildSrc project.
As far as I know, this does not overwrite or duplicate the `buildscript.repositories` property in the `build.gradle`.

florian-beetz:
@Siedlerchr The `buildSrc` code is put on the classpath of the build scripts by gradle, including sub-modules. So in those sub-modules, you can just `import static Dependencies.libraries` and use them as in the main script. Same goes for the tasks I defined.

The main advantage of this is that the versions of dependencies don't need to be managed in every module, but you can just change the version in one central file and all modules will use that version.

build.gradle
```diff
@@ -75,102 +69,81 @@ repositories {
...
-    compile 'javax.inject:javax.inject:1'
+    compile libraries.fontawesome
+    compile libraries.mvvmfx
+    compile libraries.easybind
+    compile libraries.flowless
+    compile libraries.richtextfx
+    compile libraries.dndtabpane
+    compile libraries.javaInject
 
     // Cannot be updated to 9.*.* until Jabref works with Java 9
```
stefan-kolb: Move comment to library file?

build.gradle
```diff
@@ -75,102 +69,81 @@ repositories {
...
     // >1.8.0-beta is required for java 9 compatibility
-    compile 'org.slf4j:slf4j-api:1.8.0-beta2'
-    compile 'org.apache.logging.log4j:log4j-slf4j-impl:2.11.0'
-    compile 'org.apache.logging.log4j:log4j-jcl:2.11.0'
-    compile 'org.apache.logging.log4j:log4j-api:2.11.0'
-    compile 'org.apache.logging.log4j:log4j-core:2.11.0'
+    compile libraries.loggingApi
+    compile libraries.logging
 
     // need to use snapshots as the stable version is from 2013 and doesn't support v1.0.1 CitationStyles
```
stefan-kolb: Comment to library file

buildSrc/build.gradle
```diff
@@ -0,0 +1,13 @@
...
+
+dependencies {
+}
```
stefan-kolb: Do we need this if it's empty?

florian-beetz:
@stefan-kolb I agree that the dependency extraction is not optimal. I think something like mavens `dependencyManagement` would be much better, but afaik Gradle has no such thing.

As far as I can tell, shared dependencies are mainly testing, logging, guava, and antlr.
I guess when JDK 11 is released without JavaFX, it will become a shared dependency as well because the collections are used basically everywhere.

stefan-kolb:
The question is, if we should just include the plugin idea for now and duplicate the dependencies.
After the module split, we could evaluate if it is reasonable to extract the dependencies (of course we should keep the code you wrote anyway).
I'm not sure which way to go.

florian-beetz:
Sure, we can do that.

Just for completeness sake: I've read about two other possibilities to manage dependency versions in multi-module gradle builds. 
One managed the versions in a [`gradle.properties`](https://docs.gradle.org/current/userguide/build_environment.html#sec:gradle_configuration_properties) like `versionGuava=25.1-jre`, and kept the dependencies in the build scripts but making the versions variable like `compile "com.google.guava:guava:${versionGuava}"`.
The other is using [this gradle plugin](https://spring.io/blog/2015/02/23/better-dependency-management-for-gradle) together with a Maven BOM, but IMO using the Maven XML with Gradle is pretty ugly.

stefan-kolb:
Using a properties file looks promising to me.
This also would separate code and version declarations pretty well.

tobiasdiez:
The extraction of plugins is definitely a good idea.
In principle, I also like the refactored dependencies except that it is now relatively complex to add a new dependency: you have to add an entry under version, under libraries and in the main gradle file. What do you think about combining the version and libraries lists, i.e. define the libraries in the usual full gradle format like `org.apache.logging.log4j:log4j-core:2.11.0`. Then adding a new dependency is still straightforward but you also have the advantage to reuse it in other modules.

tobiasdiez:
We discussed this in the dev call, and the decision was to include the changes of the plugins (done in https://github.com/JabRef/jabref/pull/4934) and to wait with the extraction of the dependencies until we have a bit more experience with the modular build.

Thanks for your PR.

PR #4185, opened 2018-07-08
Convert CiteSeerX fetcher to new infrastructure

tobiasdiez:
The old implementation first extracted the detail pages for each article matched by the query and then went to every detail page to extract the bibliographic information. Instead, we now parse the COinS information that is already contained in the main result page and thereby reduce the number of requests to 1 per query.

With this PR, now all fetcher are converted to the new infrastructure and the old one can now be removed (to be done in a new PR).

----


src/test/java/org/jabref/logic/importer/fetcher/CiteSeerTest.java
```diff
@@ -0,0 +1,36 @@
...
+        fetcher = new CiteSeer();
+    }
+
+    @Test
+    void searchByQueryFindsEntry() throws Exception {
+        BibEntry expected = new BibEntry();
+        expected.setType(BibtexEntryTypes.MISC);
+        expected.setField("author", "Wang Wei and Zhang Pingwen and Zhang Zhifei");
+        expected.setField("title", "Rigorous Derivation from Landau-de Gennes Theory to Eericksen-leslie Theory");
+        expected.setField("doi", "10.1.1.744.5780");
```
stefan-kolb: We need test for all other parsed fields as well!
- JOURNAL
- YEAR
- VOLUME
- PAGES
- ISSUE

tobiasdiez: I was not able to find an entry that had pages/volume or issue information. However, I added a test covering journal and year.

Issue #4173, opened 2018-07-05
Silently fails when saving library to unavailable network drive (wrongfully claims "Status: Saved library '\\SERVER\\...' without indication to user that something went wrong)


Ede123:
- [ ] I have tested the latest master version from http://builds.jabref.org/master/ and the problem persists

JabRef 4.3.1
Windows 10 10.0 x86 
Java 1.8.0_172

Steps to reproduce:

1. Save a library onto a network drive (UNC path like `\\SERVER\folder\library.bib`)
2. Make sure the network drive is not available anymore
3. Save the (already opened) library in JabRef

Result:

JabRef fails silently. It claims "Status: Saved library '\\SERVER\folder\library.bib'." in the status bar when it actually did *not* save the library leading to potential data loss.

Errors are logged but not reported to the user.

<details>

```
Error when creating lock file.
java.nio.file.FileSystemException: \\SERVER\folder\library.bib.lock: Der Netzwerkpfad wurde nicht gefunden.

	at sun.nio.fs.WindowsException.translateToIOException(Unknown Source)
	at sun.nio.fs.WindowsException.rethrowAsIOException(Unknown Source)
	at sun.nio.fs.WindowsException.rethrowAsIOException(Unknown Source)
	at sun.nio.fs.WindowsFileSystemProvider.newByteChannel(Unknown Source)
	at java.nio.file.spi.FileSystemProvider.newOutputStream(Unknown Source)
	at java.nio.file.Files.newOutputStream(Unknown Source)
	at java.nio.file.Files.write(Unknown Source)
	at org.jabref.logic.util.io.FileBasedLock.createLockFile(FileBasedLock.java:115)
	at org.jabref.logic.exporter.FileSaveSession.commit(FileSaveSession.java:78)
	at org.jabref.gui.exporter.SaveDatabaseAction.saveDatabase(SaveDatabaseAction.java:256)
	at org.jabref.gui.exporter.SaveDatabaseAction.run(SaveDatabaseAction.java:142)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
	at java.lang.reflect.Method.invoke(Unknown Source)
	at spin.Invocation.evaluate(Invocation.java:175)
[...]

PR #4193, opened 2018-07-12
[WIP] Save logic #4173

stefan-kolb:
#4173
Save logic is fundamentally flawed imho 😢 

Without a status bar, we have no more possibility to signal problems to the user (it is debatable if the status bar was the right place for errors anyway).
Or at least a lot of the problems are swallowed right now.


stefan-kolb:
The entire save logic looks wierd. I just scratched the surface.

tobiasdiez:
I guess this is now surpassed by #4309.

PR #4200, opened 2018-07-14
Make it easier to rename and move files

tobiasdiez:
Removes confirmation dialogs when renaming or moving files. It is hard to accidentally invoke this action (you have to right click on the file > move/rename) and it is not irreversible. Thus, in my opinion, no confirmation dialog is needed.

----

### DIFF
--- a/src/main/java/org/jabref/gui/fieldeditors/LinkedFileViewModel.java
+++ b/src/main/java/org/jabref/gui/fieldeditors/LinkedFileViewModel.java
@@ -201,34 +201,19 @@ public class LinkedFileViewModel extends AbstractViewModel {
             // Cannot rename remote links
             return;
         }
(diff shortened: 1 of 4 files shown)


src/main/java/org/jabref/gui/fieldeditors/LinkedFileViewModel.java
```diff
@@ -201,49 +201,35 @@ public void rename() {
             // Cannot rename remote links
             return;
         }
-        Optional<Path> fileDir = databaseContext.getFirstExistingFileDir(fileDirectoryPreferences);
```
Siedlerchr: Why remove this?

tobiasdiez: This checks whether a certain file directory exists (the one where we usually put files). This is unnecessary if we just want to rename the file...

Siedlerchr:
If you fix the checkstyle issue it's good

Issue #4190, opened 2018-07-12
UI issues


stefan-kolb:
- [ ] set general fields in background/no modal dialog
- [ ] manage custom imports no modal dialog
- [ ] manage custom exports no modal dialog
- [ ] manage content selectors no modal dialog
- [ ] customize entry types in background/no modal dialog
- [x] manage journal abbreviations cannot be closed by x button (fixed with #4263)
- [x] appearance tab inside preferences is empty (fixed with #4178)
![image](https://user-images.githubusercontent.com/2141507/42633619-5fc7479a-85e1-11e8-832b-f7edaa04aef0.png)
- [x] checkboxes are not recognizable for user (appearance tab) (fixed with #4261)
![image](https://user-images.githubusercontent.com/2141507/42727853-243d3ff8-87ae-11e8-83a7-320b5bcaa30a.png)

mlep:
I believe that "appearance tab inside preferences is empty" is still not solved yet: while, indeed, the appearance tab is not empty anymore, only the font size can be selected (thanks to #4178). But the Look&Feel cannot be chosen yet. And once the Look&Feel could be set int the appearance tab, issue #4197 will be solved.

stefan-kolb:
There is no look and feel anymore.

mlep:
Thank you for the clarification. It was not clear to me from the discussion on issue #4197.

Siedlerchr:
As we are migrating the rest of the dialogs to javafx, the old look and feel settings become obsolete as they only affect the old swing dialogs See also  #4130 were we try to provide a dark theme for javafx.
As it's solely based on css new themes can easily be developed

mlep:
So, the "Freezes when "Add to Group"" bug (issue #4197) will "naturally" be solved thanks to this migration. Great!

tobiasdiez:
There is nothing we can do about the modality of the old Swing dialogs as that would need a main Swing window. Since the other issues are now fixed, I'll this now.

PR #4205, opened 2018-07-15
Fix that swing dialogs are hidden behind main window

tobiasdiez:
This fixes some of the issues mentioned in https://github.com/JabRef/jabref/issues/4190, namely that swing dialogs are hidden behind the main window. They are still not modal and not centered at the screen since for this a Swing frame is needed as the owner or parent...but I couldn't find a way to convert our JavaFX main window to a Swing frame or window. 

----

### DIFF
--- a/src/main/java/org/jabref/gui/JabRefDialog.java
+++ b/src/main/java/org/jabref/gui/JabRefDialog.java
@@ -1,7 +1,6 @@
 package org.jabref.gui;
 
 import java.awt.Frame;
-import java.awt.Window;
 
 import javax.swing.JDialog;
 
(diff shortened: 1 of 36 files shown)


src/main/java/org/jabref/gui/JabRefDialog.java
```diff
@@ -47,13 +36,19 @@
...
     private <T extends JabRefDialog> void trackDialogOpening(Class<T> clazz) {
         Globals.getTelemetryClient().ifPresent(client -> client.trackPageView(clazz.getName()));
     }
+
+    @Override
+    public void setVisible(boolean visible) {
+        super.setVisible(visible);
+
+        if (visible) {
+            // Ugly hack to ensure that new dialogs are not hidden behind the main window
```
stefan-kolb: Add a FIXME otherwise this will be there forever, I guess.

PR #4212, opened 2018-07-17
Make attached files relative to the file directory

### LINKED ISSUE #4201: Jabref 5.0dev does not respect the standard directory
JabRef 5.0-dev--snapshot--2018-07-12--master--dd9b22911
Windows 7 6.1 amd64 
Java 1.8.0_172
- [ ] I have tested the latest master version from http://builds.jabref.org/master/ and the problem persists

Steps to reproduce:

1. load jabref 5.0dev and any library
2. try to save a pdf entry for an entry
3. you will notice that the path which is saved is the full path, the directory in the standard directory entry included

This is a Verschlimmbesserung, sorry for the nice German word. it means changing not for the better, but for the worth.

For first entry is manually edited, the second not
grafik
the standard directory is:
grafik

Siedlerchr:
Fixes #4201

----


src/main/java/org/jabref/gui/filelist/AttachFileAction.java
```diff
@@ -32,11 +34,14 @@ public void execute() {
...
         FileDialogConfiguration fileDialogConfiguration = new FileDialogConfiguration.Builder()
-                .withInitialDirectory(Globals.prefs.get(JabRefPreferences.WORKING_DIRECTORY))
-                .build();
+                                                                                               .withInitialDirectory(Globals.prefs.get(JabRefPreferences.WORKING_DIRECTORY))
+                                                                                               .build();
 
         dialogService.showFileOpenDialog(fileDialogConfiguration).ifPresent(newFile -> {
-            LinkedFile newLinkedFile = new LinkedFile("", newFile.toString(), "");
+
+            Path relativePath = FileUtil.shortenFileName(newFile, panel.getBibDatabaseContext().getFileDirectoriesAsPaths(Globals.prefs.getFileDirectoryPreferences()));
```
tobiasdiez: In my opinion, this code should be refactored to use the same code as the same action in the entry editor. Otherwise, we will again run into a similar problem where the two actions are different in behavior.

src/main/java/org/jabref/gui/filelist/LinkedFilesEditDialogViewModel.java
```diff
@@ -119,7 +119,7 @@ public StringProperty descriptionProperty() {
     }
 
     public LinkedFile getNewLinkedFile() {
-        return new LinkedFile(description.getValue(), link.getValue(), selectedExternalFileType.getValue().toString());
+        return new LinkedFile(description.getValue(), link.getValue(), selectedExternalFileType.getValue() == null ? null : selectedExternalFileType.getValue().toString());
```
tobiasdiez: As we wanted to get ride of null's, I don't like this change. Maybe we should use a `MonadicBinding` (optional binding) and use an empty string as file type?

src/main/java/org/jabref/gui/desktop/JabRefDesktop.java
```diff
@@ -160,6 +155,8 @@ private static void openExternalFilePlatformIndependent(Optional<ExternalFileTyp
             } else {
                 NATIVE_DESKTOP.openFileWithApplication(filePath, application);
             }
+        } else {
+            NATIVE_DESKTOP.openFile(filePath, null);
```
tobiasdiez: Please add a short comment explaining what is done here.

src/main/java/org/jabref/gui/filelist/AttachFileAction.java
Siedlerchr: Changed it

PR #4216, opened 2018-07-17
Fix fetcher tests

Siedlerchr:
Fix fetcher test and fix l10n logging for test

@tobiasdiez  could you please check if the zbMath and the MathSciNet tests are still correct? 
I am getting 401 as I have no access.

----

### DIFF
--- a/src/main/java/org/jabref/logic/importer/fetcher/ArXiv.java
+++ b/src/main/java/org/jabref/logic/importer/fetcher/ArXiv.java
@@ -367,8 +367,8 @@ public class ArXiv implements FulltextFetcher, SearchBasedFetcher, IdBasedFetche
          * Returns the arXiv identifier
          */
         public Optional<String> getIdString() {
-            // remove leading https://arxiv.org/abs/ from abstract url to get arXiv ID
-            String prefix = "https://arxiv.org/abs/";
+            // remove leading http://arxiv.org/abs/ from abstract url to get arXiv ID
+            String prefix = "http://arxiv.org/abs/";
             return urlAbstractPage.map(abstractUrl -> {
                 if (abstractUrl.startsWith(prefix)) {
                     return abstractUrl.substring(prefix.length());
--- a/src/main/java/org/jabref/logic/importer/fetcher/ArXiv.java
+++ b/src/main/java/org/jabref/logic/importer/fetcher/ArXiv.java
@@ -59,7 +59,6 @@ public class ArXiv implements FulltextFetcher, SearchBasedFetcher, IdBasedFetche
     private static final Logger LOGGER = LoggerFactory.getLogger(ArXiv.class);
 
     private static final String API_URL = "https://export.arxiv.org/api/query";
-    // remove leading http(s)://arxiv.org/abs/ from abstract url to get arXiv ID
     private static final String ARXIV_URL_PREFIX_FOR_ID = "(https?://arxiv.org/abs/)";
     private static final Pattern URL_PATTERN = Pattern.compile(ARXIV_URL_PREFIX_FOR_ID);
 
@@ -377,6 +376,7 @@ public class ArXiv implements FulltextFetcher, SearchBasedFetcher, IdBasedFetche
             return urlAbstractPage.map(abstractUrl -> {
                 Matcher matcher = URL_PATTERN.matcher(abstractUrl);
                 if (matcher.find()) {
+                    // remove leading http(s)://arxiv.org/abs/ from abstract url to get arXiv ID
                     return abstractUrl.substring(matcher.group(1).length());
                 } else {
                     return abstractUrl;
--- a/src/main/java/org/jabref/logic/importer/fetcher/ArXiv.java
+++ b/src/main/java/org/jabref/logic/importer/fetcher/ArXiv.java
@@ -10,6 +10,8 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import javax.xml.parsers.DocumentBuilder;
@@ -53,9 +55,13 @@ import org.xml.sax.SAXException;
  * <a herf="https://gitlab.c3sl.ufpr.br/portalmec/dspace-portalmec/blob/aa209d15082a9870f9daac42c78a35490ce77b52/dspace-api/src/main/java/org/dspace/submit/lookup/ArXivService.java">dspace-portalmec</a>
  */
 public class ArXiv implements FulltextFetcher, SearchBasedFetcher, IdBasedFetcher, IdFetcher<ArXivIdentifier> {
+
     private static final Logger LOGGER = LoggerFactory.getLogger(ArXiv.class);
 
     private static final String API_URL = "https://export.arxiv.org/api/query";
+    // remove leading http(s)://arxiv.org/abs/ from abstract url to get arXiv ID
+    private static final String ARXIV_URL_PREFIX_FOR_ID = "(https?://arxiv.org/abs/)";
+    private static final Pattern URL_PATTERN = Pattern.compile(ARXIV_URL_PREFIX_FOR_ID);
 
     private final ImportFormatPreferences importFormatPreferences;
 
@@ -69,10 +75,10 @@ public class ArXiv implements FulltextFetcher, SearchBasedFetcher, IdBasedFetche
 
         try {
             Optional<URL> pdfUrl = searchForEntries(entry).stream()
-                    .map(ArXivEntry::getPdfUrl)
-                    .filter(Optional::isPresent)
-                    .map(Optional::get)
-                    .findFirst();
+                                                          .map(ArXivEntry::getPdfUrl)
+                                                          .filter(Optional::isPresent)
+                                                          .map(Optional::get)
+                                                          .findFirst();
 
             if (pdfUrl.isPresent()) {
                 LOGGER.info("Fulltext PDF found @ arXiv.");
@@ -159,7 +165,7 @@ public class ArXiv implements FulltextFetcher, SearchBasedFetcher, IdBasedFetche
     }
 
     private List<ArXivEntry> queryApi(String searchQuery, List<ArXivIdentifier> ids, int start, int maxResults)
-            throws FetcherException {
+        throws FetcherException {
         Document result = callApi(searchQuery, ids, start, maxResults);
         List<Node> entries = XMLUtil.asList(result.getElementsByTagName("entry"));
 
@@ -195,7 +201,7 @@ public class ArXiv implements FulltextFetcher, SearchBasedFetcher, IdBasedFetche
             }
             if (!ids.isEmpty()) {
                 uriBuilder.addParameter("id_list",
-                        ids.stream().map(ArXivIdentifier::getNormalized).collect(Collectors.joining(",")));
+                                        ids.stream().map(ArXivIdentifier::getNormalized).collect(Collectors.joining(",")));
             }
             uriBuilder.addParameter("start", String.valueOf(start));
             uriBuilder.addParameter("max_results", String.valueOf(maxResults));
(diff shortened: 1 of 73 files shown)


src/main/java/org/jabref/logic/l10n/Localization.java
```diff
@@ -51,8 +51,7 @@ private Localization() {
      */
     public static String lang(String key, String... params) {
         if (localizedMessages == null) {
-            // I'm logging this because it should never happen
-            LOGGER.error("Messages are not initialized before accessing key: " + key);
+            // This will happen when running junit tests
```
tobiasdiez: Why is this now a problem? While changing some localization-based stuff, I hit this error message sometimes during development and it was helpful to locate the problem. Hence, I would prefer if it could stay.

src/test/java/org/jabref/logic/importer/fetcher/ArXivTest.java
```diff
@@ -41,7 +41,7 @@ public void setUp() {
...
         sliceTheoremPaper.setField("abstract", "A general slice theorem for the action of a Fr\\'echet Lie group on a Fr\\'echet manifolds is established. The Nash-Moser theorem provides the fundamental tool to generalize the result of Palais to this infinite-dimensional setting. The presented slice theorem is illustrated by its application to gauge theories: the action of the gauge transformation group admits smooth slices at every point and thus the gauge orbit space is stratified by Fr\\'echet manifolds. Furthermore, a covariant and symplectic formulation of classical field theory is proposed and extensively discussed. At the root of this novel framework is the incorporation of field degrees of freedom F and spacetime M into the product manifold F * M. The induced bigrading of differential forms is used in order to carry over the usual symplectic theory to this new setting. The examples of the Klein-Gordon field and general Yang-Mills theory illustrate that the presented approach conveniently handles the occurring symmetries.");
-        sliceTheoremPaper.setField("eprint", "1405.2249v1");
+        sliceTheoremPaper.setField("eprint", "http://arxiv.org/abs/1405.2249v1");
```
tobiasdiez: This is wrong... it should just insert the number (i.e. the fetcher should be fixed not the test)

src/test/java/org/jabref/logic/importer/fetcher/ArXivTest.java
```diff
@@ -142,7 +142,7 @@ public void searchEntryByOldId() throws Exception {
         expected.setField("title", "Multi-Electron Production at High Transverse Momenta in ep Collisions at HERA");
         expected.setField("date", "2003-07-07");
         expected.setField("abstract", "Multi-electron production is studied at high electron transverse momentum in positron- and electron-proton collisions using the H1 detector at HERA. The data correspond to an integrated luminosity of 115 pb-1. Di-electron and tri-electron event yields are measured. Cross sections are derived in a restricted phase space region dominated by photon-photon collisions. In general good agreement is found with the Standard Model predictions. However, for electron pair invariant masses above 100 GeV, three di-electron events and three tri-electron events are observed, compared to Standard Model expectations of 0.30 \\pm 0.04 and 0.23 \\pm 0.04, respectively.");
-        expected.setField("eprint", "hep-ex/0307015v1");
+        expected.setField("eprint", "http://arxiv.org/abs/hep-ex/0307015v1");
```
tobiasdiez: dito

src/test/java/org/jabref/logic/importer/fetcher/ArXivTest.java
```diff
@@ -189,7 +189,7 @@ public void searchWithMalformedIdThrowsException() throws Exception {
     public void searchIdentifierForSlicePaper() throws Exception {
         sliceTheoremPaper.clearField(FieldName.EPRINT);
 
-        assertEquals(ArXivIdentifier.parse("1405.2249v1"), finder.findIdentifier(sliceTheoremPaper));
+        assertEquals(ArXivIdentifier.parse("http://arxiv.org/abs/1405.2249v1"), finder.findIdentifier(sliceTheoremPaper));
```
tobiasdiez: should work just with the id...

src/test/java/org/jabref/logic/importer/fetcher/LibraryOfCongressTest.java
```diff
@@ -6,16 +6,24 @@
...
 
-    private final LibraryOfCongress fetcher = new LibraryOfCongress(mock(ImportFormatPreferences.class, Answers.RETURNS_DEEP_STUBS));
+    ImportFormatPreferences prefs = mock(ImportFormatPreferences.class);
+
+    @BeforeEach
+    public void setUp() {
+        when(prefs.getKeywordSeparator()).thenReturn(',');
+    }
+
+    private final LibraryOfCongress fetcher = new LibraryOfCongress(prefs);
```
tobiasdiez: initialize prefs and fetcher in `setup`

src/test/java/org/jabref/logic/importer/fetcher/AstrophysicsDataSystemTest.java
```diff
@@ -176,8 +178,7 @@ public void testPerformSearchByIdEmptyDOI() throws Exception {
 
     @Test
     public void testPerformSearchByIdInvalidDoi() throws Exception {
-        Optional<BibEntry> fetchedEntry = fetcher.performSearchById("this.doi.will.fail");
-        assertEquals(Optional.empty(), fetchedEntry);
+        assertThrows(FetcherException.class, () -> fetcher.performSearchById("this.doi.will.fail"));
```
tobiasdiez: I'm not sure about our conventions, but whether we throw an exception or return an empty result should be consistent across id fetchers (and it should be added as a general test for all id fetchers (there should be a test class for all id fetchers))

Siedlerchr: The problem is that our URLDownload throws a FileNotFound Exception on an 404 error, because you are trying to access an input stream of a non existent resource) and thus resulting in a fetcher exception which does not return anything.

src/test/java/org/jabref/logic/importer/fetcher/ArXivTest.java
Siedlerchr: Same reason as above, the bibtex code returns the url + id and will not be equal if not.

src/main/java/org/jabref/logic/l10n/Localization.java
Siedlerchr: Yes, but the problem now is that you will hit a NotInitlaizedException in junit. Execute the isbn fetcher test for example and you will see it. It has somehow to do with the fact that the log4j tries to call another appender (GUI based in the end, though it's executed from the test).

src/test/java/org/jabref/logic/importer/fetcher/ArXivTest.java
tobiasdiez: Well, then the implementation needs be changed to strip the url part ;-)

src/main/java/org/jabref/logic/l10n/Localization.java
tobiasdiez: The appender are/can be configured separately for the tests, so removing the GUI-appender is worth a try.

src/test/java/org/jabref/logic/importer/fetcher/AstrophysicsDataSystemTest.java
stefan-kolb: I think the default is an empty result if the ID cannot be found or used.

Siedlerchr: I changed the method in UrlDownload to return an empty stream when a 404 or 400 error (the latter one is thrown by medline)  is encountered. I simply log the status response code + url .

src/test/java/org/jabref/logic/importer/fetcher/LibraryOfCongressTest.java
Siedlerchr: done

src/test/java/org/jabref/logic/importer/fetcher/ArXivTest.java
Siedlerchr: The strip was already implemented, but it checked for https and not http.

src/main/java/org/jabref/logic/l10n/Localization.java
Siedlerchr: Fixed the loging configuration, the xml file had to be renamed

src/test/java/org/jabref/logic/importer/fetcher/AstrophysicsDataSystemTest.java
stefan-kolb: Can't you catch the Exception inside the fetcher? I don't like our URLDownload implementation anyway.

src/main/java/org/jabref/logic/importer/fetcher/ArXiv.java
```diff
@@ -367,8 +367,8 @@ public static String correctLineBreaks(String s) {
          * Returns the arXiv identifier
          */
         public Optional<String> getIdString() {
-            // remove leading https://arxiv.org/abs/ from abstract url to get arXiv ID
-            String prefix = "https://arxiv.org/abs/";
+            // remove leading http://arxiv.org/abs/ from abstract url to get arXiv ID
+            String prefix = "http://arxiv.org/abs/";
```
stefan-kolb: Why no HTTPS anymore? We just changed this?!

src/test/java/org/jabref/logic/importer/fetcher/DBLPFetcherTest.java
```diff
@@ -64,4 +64,10 @@ public void findSingleEntryUsingComplexOperators() throws FetcherException {
         assertEquals(Collections.singletonList(entry), result);
     }
 
+    @Test
+    public void findNothing() throws Exception
+    {
```
stefan-kolb: Brackets alignment

src/main/java/org/jabref/logic/importer/fetcher/ArXiv.java
tobiasdiez: I am also in favor of removing both http and https prefixes

Siedlerchr: The fetcher itself still uses https, but the prefix for the eprint field does not start with https but with http

stefan-kolb: But we only remove http prefixes not https (if there exists one at a point in time)?!

src/test/java/org/jabref/logic/importer/fetcher/AstrophysicsDataSystemTest.java
Siedlerchr: Well that was done before but the fetcher exception just encapsulated a file not found exception with no way to getting the underlying http status code.

src/main/java/org/jabref/logic/importer/fetcher/ArXiv.java
Siedlerchr: Added check for both

Siedlerchr:
I fixed all remainig issues. Error response 400 or 404 are logged in the error log.

Siedlerchr:
Ebook.de and Google Scholar work again. Chimbori fetcher is broken, but author is noticed and responded that it will be fixed soon.

src/main/java/org/jabref/logic/importer/fetcher/ArXiv.java
```diff
@@ -53,9 +55,13 @@
  * <a herf="https://gitlab.c3sl.ufpr.br/portalmec/dspace-portalmec/blob/aa209d15082a9870f9daac42c78a35490ce77b52/dspace-api/src/main/java/org/dspace/submit/lookup/ArXivService.java">dspace-portalmec</a>
  */
 public class ArXiv implements FulltextFetcher, SearchBasedFetcher, IdBasedFetcher, IdFetcher<ArXivIdentifier> {
+
     private static final Logger LOGGER = LoggerFactory.getLogger(ArXiv.class);
 
     private static final String API_URL = "https://export.arxiv.org/api/query";
+    // remove leading http(s)://arxiv.org/abs/ from abstract url to get arXiv ID
```
tobiasdiez: This comment should not be here but in the method below

Issue #4092, opened 2018-06-04
Names are not in the standard BibTex format


stefan-kolb:
`L.C. Briand and S. Morasca and V.R. Basily`

Only because space or a dot is missing after firstnames.

Also remove dot at the end of the warning message: `Names are not in the standard BibTex format.`

bernhard-kleine:
This is an issue I would like to see been solved. Any entry I add using Medline has this problem with missing dots. Why don't you have the program add them? If it can recognize the missing dots, I guess it would be simple to add them, too.

tobiasdiez:
Right-click > Normalize to BibTeX name format adds the missing space (and reorders the name parts).

I'm actually not sure if we should change this behavior. As far as I'm aware, the space should be there by convention...

bernhard-kleine:
I did the rightclick > Normalize but the error (?) remains.
JabRef 5.0-dev--snapshot--2018-07-12--master--dd9b22911
Windows 7 6.1 amd64 
Java 1.8.0_172
JabRef 5.0-dev--snapshot--2018-07-12--master--dd9b22911
Windows 7 6.1 amd64 
Java 1.8.0_172
![grafik](https://user-images.githubusercontent.com/9216487/42750526-d09ca83a-88e7-11e8-9885-39fde0c20f43.png)

tobiasdiez:
The BibTeX standard seems to be that there should be a space after the dot. Hence, the check is ok like it is right now.

z3li3us:
Try to use curly brackets maybe?

koppor:
@z3li3us Longer background of the issue: JabRef has build-in checkers for fields. One is for the content in the BibTeX field. In the issue report at hand, the names are correct, but the checker complained. The tool JabRef should not force the user to change the data, but it should adapt to other formats.

PR #4225, opened 2018-07-22
Fixed: #4092 Names are not in the standard BibTex format.

1160300509:
Fixed: #4092

----

### DIFF
--- a/src/main/java/org/jabref/logic/integrity/PersonNamesChecker.java
+++ b/src/main/java/org/jabref/logic/integrity/PersonNamesChecker.java
@@ -33,13 +33,16 @@ public class PersonNamesChecker implements ValueChecker {
 
         // Remove all brackets to handle corporate names correctly, e.g., {JabRef}
         value = new RemoveBrackets().format(value);
+
+        // put ' ' after every name, e.g.,L.C. Brand -> L. C. Brand
+        String checkFormat = value.replaceAll("\\.", ". ").replaceAll("\\  ", " ");
         // Check that the value is in one of the two standard BibTeX formats:
         //  Last, First and ...
         //  First Last and ...
         AuthorList authorList = AuthorList.parse(value);
-        if (!authorList.getAsLastFirstNamesWithAnd(false).equals(value)
-                && !authorList.getAsFirstLastNamesWithAnd().equals(value)) {
-            return Optional.of(Localization.lang("Names are not in the standard %0 format.", bibMode.getFormattedName()));
+        if (!authorList.getAsLastFirstNamesWithAnd(false).equals(checkFormat)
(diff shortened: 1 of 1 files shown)


tobiasdiez:
@stefan-kolb @Siedlerchr Do you really think we should be forgiving with respect to the space? I'm fine with both but we should think a second about it...

Code-wise tests are missing, otherwise it looks good.

src/main/java/org/jabref/logic/integrity/PersonNamesChecker.java
```diff
@@ -33,13 +33,16 @@ public PersonNamesChecker(BibDatabaseContext databaseContext) {
 
         // Remove all brackets to handle corporate names correctly, e.g., {JabRef}
         value = new RemoveBrackets().format(value);
+
+        // put ' ' after every name, e.g.,L.C. Brand -> L. C. Brand
+        String checkFormat = value.replaceAll("\\.", ". ").replaceAll("\\  ", " ");
```
stefan-kolb: Why do we need `replaceAll("\\ ", " ")`

src/main/java/org/jabref/logic/integrity/PersonNamesChecker.java
```diff
@@ -33,13 +33,16 @@ public PersonNamesChecker(BibDatabaseContext databaseContext) {
 
         // Remove all brackets to handle corporate names correctly, e.g., {JabRef}
         value = new RemoveBrackets().format(value);
+
+        // put ' ' after every name, e.g.,L.C. Brand -> L. C. Brand
```
stefan-kolb: put space after...

stefan-kolb:
@tobiasdiez what do you mean?

src/main/java/org/jabref/logic/integrity/PersonNamesChecker.java
1160300509: If we don't need `replaceAll("\\ ", " ")` , for example, `L.C. Brand` ,`checkFormat` will be   `L. C. Briand` ( there are two spaces between `L.` and `C.` , `C.` and `Brand`, not one space). So there's also a warning message. We use `checkFormat` to help check the name format.

tobiasdiez:
@stefan-kolb I'm not sure if we really should change the behavior of the check. As far as I'm aware it is indeed the usual convention to add a space after the initial (e.g. `L. C. Brand` instead of `L.C. Brand`).

stefan-kolb:
Hm yeah so this should not go into the checker but to the normalize names functionality?!

tobiasdiez:
I thnik we should just leave the check as it was.

PR #4226, opened 2018-07-22
FIx chinese author bibtex key generation

### LINKED ISSUE #4169: can't generate bibtex for Chinese author

JabRef version  on 

Steps to reproduce:

1. The rule for generate key is [auth][year] and  include Chinese authors  in author field such as 王, 阳 and 张, 家发 and 胡, 喜军 and 李, 保明
2. Ctrl+G to generate key
3. only year appear in the key.

<details>
  <summary>Log File</summary>
  
  ```
  Paste an excerpt of your log file here
  ```
</details>

1160300420:
-We fixed an issue where the default bibtexkey's generation is incorrect for Chinese authors.
For example:
suppose that the author field such as 王, 阳 and 张, 家发 and 胡, 喜军 and 李, 保明, the year is 1000 and the default bibtexkey's generation is [author][year]
the original output is:
1000 (it is obviously incorrect)
we add some code to address it
now the output is:
王1000 (which is specified as user guide)
well, we also add some test code and passed these tests successfully

Fixes #4169

### DIFF
--- a/src/main/java/org/jabref/logic/bibtexkeypattern/BracketedPattern.java
+++ b/src/main/java/org/jabref/logic/bibtexkeypattern/BracketedPattern.java
@@ -570,7 +570,7 @@ public class BracketedPattern {
         if (authorList.isEmpty()) {
             return "";
         }
-        if (authorList.getAuthor(0).getVon().toString().equals("Optional.empty")) {
+        if (!authorList.getAuthor(0).getVon().isPresent()) {
             Character.UnicodeBlock ub = Character.UnicodeBlock.of(authorList.getAuthor(0).getLast().orElse("").charAt(0));
             if ((ub == Character.UnicodeBlock.CJK_UNIFIED_IDEOGRAPHS)
                 || (ub == Character.UnicodeBlock.CJK_COMPATIBILITY_IDEOGRAPHS)
--- a/src/main/java/org/jabref/logic/bibtexkeypattern/BracketedPattern.java
+++ b/src/main/java/org/jabref/logic/bibtexkeypattern/BracketedPattern.java
@@ -570,6 +570,32 @@ public class BracketedPattern {
         if (authorList.isEmpty()) {
             return "";
         }
+        if (authorList.getAuthor(0).getVon().toString().equals("Optional.empty")) {
+            Character.UnicodeBlock ub = Character.UnicodeBlock.of(authorList.getAuthor(0).getLast().orElse("").charAt(0));
+            if ((ub == Character.UnicodeBlock.CJK_UNIFIED_IDEOGRAPHS)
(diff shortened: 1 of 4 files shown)


src/main/java/org/jabref/logic/bibtexkeypattern/BracketedPattern.java
```diff
@@ -570,6 +570,32 @@ public static String firstAuthor(String authorField) {
         if (authorList.isEmpty()) {
             return "";
         }
+        if (authorList.getAuthor(0).getVon().toString().equals("Optional.empty")) {
```
Siedlerchr: Much easier, Optionals have a method names isPresent https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html

1160300420: Thanks for your suggestion! We will change it better.

src/main/java/org/jabref/logic/bibtexkeypattern/BracketedPattern.java
```diff
@@ -570,6 +570,32 @@ public static String firstAuthor(String authorField) {
...
+        } else {
+            Character.UnicodeBlock ub = Character.UnicodeBlock.of(authorList.getAuthor(0).getVon().orElse("").charAt(0));
+            if ((ub == Character.UnicodeBlock.CJK_UNIFIED_IDEOGRAPHS)
+                || (ub == Character.UnicodeBlock.CJK_COMPATIBILITY_IDEOGRAPHS)
+                || (ub == Character.UnicodeBlock.CJK_UNIFIED_IDEOGRAPHS_EXTENSION_A)
+                || (ub == Character.UnicodeBlock.GENERAL_PUNCTUATION)
+                || (ub == Character.UnicodeBlock.CJK_SYMBOLS_AND_PUNCTUATION)
+                || (ub == Character.UnicodeBlock.HALFWIDTH_AND_FULLWIDTH_FORMS)) {
+                return authorList.getAuthor(0).getVon().orElse("");
+            }
```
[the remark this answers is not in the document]
1160300420: Well, I understand your meaning. At first, I did  want to change the code in the way you suggested. But then I found that the key of author filed containing Chinese name is totally different from the situation of English name. In AuthorListParser.class, there are four situations for the split of English name(two based on no commas, one based on one commas,one based on two or more commas), while there are only two fixed situations and other instances don't apply to Chinese name. Also it's easier to fix the bug in bibtexkeypattern.class while it's complex and not neccessary to  move this fix to the AuthorListParser.class.

Siedlerchr:
If I remeber correctly this issue was now resolved with the other PR  https://github.com/JabRef/jabref/pull/4248

PR #4227, opened 2018-07-22
Fix Part of Issue#3861: Edit->Replace String

marisuki:
Hi, we fixed part of the Issue[#3861](https://github.com/JabRef/jabref/issues/3861), change the Swing Dialog to the JavaFx Dialog: the Edit -> Replace String Dialog, we add 3 files and modified the general control to link the Replace String controller of JavaFX Dialog, and we did not delete the old dialog logic (ReplaceStringDialog.java). The outlook of Dialog hasn't change a lot, and the previous logic is inherited.

The Screenshot is pasted below, just like the design before. 

githubjabrefreplacestring

----


src/main/java/org/jabref/gui/ReplaceStringAction.java
```diff
@@ -0,0 +1,20 @@
...
+
+import org.jabref.gui.actions.SimpleCommand;
+import org.jabref.model.database.BibDatabase;
+
+public class ReplaceStringAction extends SimpleCommand
+{
+    private BasePanel basePanel;
+
+    public ReplaceStringAction(BasePanel bPanel) {
+        this.basePanel = bPanel;
```
Siedlerchr: Typically you name the parameter also basePanel as the field and use this to distinguish the field from the parameter.
e.g. this.baesPanel = basePanel

src/main/java/org/jabref/gui/ReplaceString.fxml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<?import javafx.scene.control.*?>
+<?import javafx.scene.layout.*?>
+
+<DialogPane  maxHeight="-Infinity" maxWidth="-Infinity" minHeight="-Infinity" minWidth="-Infinity" prefHeight="339.0" prefWidth="476.0" xmlns:fx="http://javafx.com/fxml/1" xmlns="http://javafx.com/javafx/8" fx:controller="org.jabref.gui.ReplaceStringView" fx:id="pane">
+  <content>
+    <AnchorPane minHeight="0.0" minWidth="0.0" prefHeight="376.0" prefWidth="552.0">
```
Siedlerchr: You better should use a GridPane here for positioning the label and the text fields, that allows them to be properly aligned. Can be easily done using the Scence Builder.
You can also take a look at the FXML here from my other PR https://github.com/JabRef/jabref/pull/4040 for the db login dialog:
https://github.com/JabRef/jabref/pull/4040/files#diff-200f60f39aab2550ae174ea88d604981

src/main/java/org/jabref/gui/ReplaceStringView.java
```diff
@@ -0,0 +1,184 @@
...
+                  .setAsDialogPane(this);
+
+        st = (Stage) this.pane.getScene().getWindow();
+        st.setOnCloseRequest(event -> st.close());
+    }
+
+    @FXML
+    public void initialize() {
+        visualizer.setDecoration(new IconValidationDecorator());
+        LimitFieldInput.setEditable(true);
```
Siedlerchr: They should already be editable by default.

src/main/java/org/jabref/gui/ReplaceStringView.java
```diff
@@ -0,0 +1,184 @@
...
+     * Copied and Adapted from org.jabref.gui.ReplaceStringDialog.java
+     */
+    public int replaceItem(BibEntry be, NamedCompound ce) {
+        int counter = 0;
+        if (this.AllFieldReplace) {
+            for (String s : be.getFieldNames()) {
+                counter += replaceField(be, s, ce);
+            }
+        } else {
+            for (String fld : fieldStrings) {
```
Siedlerchr: String field : fieldStrings

src/main/java/org/jabref/gui/ReplaceStringView.java
```diff
@@ -0,0 +1,184 @@
...
+    /**
+     * Does the actual operation on a Bibtex entry based on the
+     * settings specified in this same dialog. Returns the number of
+     * occurences replaced.
+     * Copied and Adapted from org.jabref.gui.ReplaceStringDialog.java
+     */
+    public int replaceItem(BibEntry be, NamedCompound ce) {
+        int counter = 0;
+        if (this.AllFieldReplace) {
+            for (String s : be.getFieldNames()) {
```
Siedlerchr: Named Compound  ce -> NamedCompound compound

Siedlerchr: And String string : be.getFieldNames...

src/main/java/org/jabref/gui/ReplaceStringView.java
```diff
@@ -0,0 +1,184 @@
...
+    private int replaceField(BibEntry be, String fieldname, NamedCompound ce) {
+        if (!be.hasField(fieldname)) {
+            return 0;
+        }
+        String txt = be.getField(fieldname).get();
+        StringBuilder sb = new StringBuilder();
+        int ind;
+        int piv = 0;
+        int counter = 0;
+        int len1 = this.findString.length();
```
Siedlerchr: maybe better findStringLength to directly make clear what the variable represent

Siedlerchr:
@marisuki  you don't have to be sorry or take it personally ;) We really much appreciate your contribution
The reviews are meant to improve and maintain a clean code base, although they sometimes may seem a bit nitpicking.  Variable naming is one important part, ideally variables should be self-documenting, e.g. they should be meaningful. Here are some tips. https://dzone.com/articles/best-practices-variable-and

marisuki:
Thank you for all the advice above :), and I changed the dialog based on GridPane also the "ViewModel", and modified the miss using of field names, finally deleted the ReplaceStringDialog.java, which is about the old style logic: Swing. I hope it would be better. :)

src/main/java/org/jabref/gui/ReplaceStringView.java
```diff
@@ -0,0 +1,83 @@
...
+    public void buttonReplace() {
+        String findString = findField.getText();
+        String replaceString = replaceField.getText();
+        String[] fieldStrings = limitFieldInput.getText().toLowerCase().split(";");
+        if ("".equals(findString))
+        {
+            stage.close();
+            return;
+        }
+        ReplaceStringViewModel viewModel = new ReplaceStringViewModel(panel, fieldStrings, findString, replaceString, selOnly, allFieldReplace);
```
tobiasdiez: This works because the dialog is very simple, but goes against the usual strategy of JavaFX / MVVM. You should add properties in the ViewModel class and bind these properties to the corresponding properties of the controls. Please have a look at the other JavaFX dialogs to see how this is done. 

Since this approach is a bit different from what you are used to in the Swing world, let me quickly explain the big overall picture and its advantages. You are probably used to program against the UI controls (as you do here right now): for example, if you want to know the input from the user, you use `textbox.getText()` to get it. Similarly, for other input controls. This works all nicely until the user reports a bug and you have to write a test for it. The problem you then face is the following: how do I automate the user interaction in my test? There are different approaches to this issue but the general answer is: there is no good solution ;-) UI automation is usually ugly and breaks as soon as you change the design. The idea behind the MVVM approach is that you completely separate the UI from the logic that performs the action. So you create a class (the ViewModel) that holds the state of the UI (i.e. you have `string` field that corresponds to the user input, but at this point you don't care how or where this string is manipulated). This state in the ViewModel is then bound to the actual userinterface in the `View` class using JavaFX bindings. In this way, you can test the ViewModel, i.e. the complete logic, without ever knowing how the user interface looks like.

src/main/java/org/jabref/gui/ReplaceStringView.java
```diff
@@ -0,0 +1,83 @@
...
+        viewModel.replace();
+        stage.close();
```
tobiasdiez: As `BaseDialog` derives from `Dialog` you can simply use `close()` (without the reference to the stage).

marisuki:
Thank you for your comment !  :) @tobiasdiez , this time I changed the initial part of "view" and add the property bindings to make it a little more MVVM pattern alike ;-). I hope it would be better.

src/main/java/org/jabref/gui/ReplaceStringViewModel.java
```diff
@@ -0,0 +1,117 @@
...
+
+    public BooleanProperty allFieldReplaceProperty() {
+        return allFieldReplaceProperty;
+    }
+
+    public BooleanProperty selectOnlyProperty() {
+        return selectOnlyProperty;
+    }
+
+    public StringProperty getFieldStringProperty() {
```
Siedlerchr: Properties in javafx don't have the traditional get/set prefix. So just rename them as the ones abvove,
e.g. fieldStringProperty  and findStringProperty etc

src/main/java/org/jabref/gui/ReplaceStringViewModel.java
```diff
@@ -0,0 +1,117 @@
...
+    public int replace() {
+        findString = findStringProperty.getValue();
+        replaceString = replaceStringProperty.getValue();
+        fieldStrings = fieldStringProperty.getValue().split(";");
+        boolean selOnly = selectOnlyProperty.getValue();
+        allFieldReplace = allFieldReplaceProperty.getValue();
+
+        final NamedCompound compound = new NamedCompound(Localization.lang("Replace string"));
+        int counter = 0;
+        if (this.panel == null)
```
Siedlerchr: Is this check necessary? You always pass the basePanel as constructor argument.
I would suggest you add an Objects.requireNonNull(basePanel) in the constructor to ensure that the argument is never null. If it is null, it will directly throw and exception which makes it easier to see the real reason,

Siedlerchr:
From my point of view this looks good. Just some minor code improvements. Let's see what @tobiasdiez   thinks about it

src/main/java/org/jabref/gui/ReplaceStringViewModel.java
marisuki: Yes, you are right :), I forgot to check my logic after converting to a viewModel ;-). Thank You!

marisuki:
Yeah, I tried every means to avoid the conflict ;-), but it does not work (caused by delete a file) . Does this conflict be inevitable ;-)?  Or, do you have any idea to solve it? Thank you.

Siedlerchr:
Okay, I tried around and found a solution:
just do git rm src/main/java/org/jabref/gui/ReplaceStringDialog.java

tobiasdiez:
Something like the following should work:
````bash
git pull upstream master (replace "upstream" by whatever you named the JabRef rep)
> git warns about merge conflict
git mergetool
> accept deletion of file
git commit
git push
````

**Comment by marisuki** [comment]
Thank you, I will give it a try! :)

**Comment by marisuki** [comment]
I think it works! :) this time I use "git pull git@github.com:jabref/jabref.git master" to merge and then remove and push again, there becomes no conflicts :). Thank you! But there shows a failure on build process, need I do something?  (Since I did not change the related files.)

File: src/main/java/org/jabref/gui/ReplaceString.fxml [hunk 11]
```diff
@@ -0,0 +1,58 @@
...
+                                <RowConstraints minHeight="10.0" prefHeight="30.0" vgrow="SOMETIMES"/>
+                                <RowConstraints minHeight="10.0" prefHeight="30.0" vgrow="SOMETIMES"/>
+                                <RowConstraints minHeight="10.0" prefHeight="30.0" vgrow="SOMETIMES"/>
+                                <RowConstraints minHeight="10.0" prefHeight="30.0" vgrow="SOMETIMES"/>
+                                <RowConstraints minHeight="10.0" prefHeight="30.0" vgrow="SOMETIMES"/>
+                                <RowConstraints minHeight="10.0" prefHeight="30.0" vgrow="SOMETIMES"/>
+                                <RowConstraints minHeight="10.0" prefHeight="30.0" vgrow="SOMETIMES"/>
+                                <RowConstraints minHeight="10.0" prefHeight="30.0" vgrow="SOMETIMES"/>
+                            </rowConstraints>
+                            <Label text="Find and Replace"/>
```
Siedlerchr: You need to add a % sign in front of the label text, so that it will be recognized as translatable string.
That is probably one of the reasons why the l10n test fails

marisuki:
Finally! I updated the properties in resources and make the "text" value linked with the file: Jabref_en.properties, also, I deleted the unused key-values (may generated from the previous replace string dialog) to make no obsolete item in the file. And thank you again for your help! :)

src/main/java/org/jabref/gui/ReplaceString.fxml
Siedlerchr: I think you missed this one ;) But all others look good

Siedlerchr:
Thank you very much for your contribution! I think you just forgot one little string, if you add this again, then we can finally merge!

src/main/java/org/jabref/gui/ReplaceString.fxml
marisuki: Yeah ;-) I miss it. Thank you for mention! :)

Siedlerchr:
As the faling test on Travis CI is not related to your changes I am going to merge it now.

marisuki:
Thank you! :-)

PR #4230, opened 2018-07-23
[WIP] Modularize JabRef with JPMS

florian-beetz:
This splits JabRef into multiple JPMS modules. (see #3704)

JabRef runs with these changes, but there are some issues:

* [ ] modernizer and checkstyle Gradle plugins don't seem to play nice with multi-module builds
* [ ] Something seems to screw up logging for tests (`ClassCastException: org.slf4j/org.slf4j.helpers.NOPLoggerFactory cannot be cast to org.gradle.internal.logging.slf4j.OutputEventListenerBackedLoggerContext`)

----


Siedlerchr:
@florian-beetz  I just fixed some issue with the test xml config part of my fetcher fixes (it was named wrong): https://github.com/JabRef/jabref/commit/d5d67d70247b95af937c3e818dc7f1ae6dd484ee#diff-91cb0a5b893efefa2c0dbcedbe1a0f85

florian-beetz:
I just found [this](https://github.com/gradle/gradle/issues/2657) and [this](https://github.com/gradle/gradle/issues/4974). Seems to be a Gradle issue actually :(

florian-beetz:
I just tried also extracting a logic module, but I encountered a problem:
The class `org.jabref.logic.importer.OpenDatabase` depends on `org.jabref.migrations`. I can't move `org.jabref.migrations` to the logic module, because it depends on the GUI, causing a circular dependency. But I also cannot move `org.jabref.logic.importer` to the main module, because this package (transitively) depends on half of the logic module, which would also need to be moved to the main module to prevent circular dependencies, but that makes modularizing basically pointless.

Do you have any ideas how this could be refactored to be able to extract the logic module?

tobiasdiez:
Not every class in `migrations` has a dependency on `gui`. In particular, the classes used in `OpenDatabase` only depend on `logic` and `model`. Thus, I would propose to split the migration package and move parts of it to `logic.migrations` and the other parts in `gui.migrations`. In the future, it would be nice refactor/recode everything so that it fits in logic, but for the moment splitting should be fine.

florian-beetz:
@tobiasdiez Nice, that worked great 👍

koppor:
@florian-beetz @tobiasdiez - Since @lenhard pinged us because of architecture. May I ask about the state of this PR? I would not like to put work into the trash. I would really like to keep this going.

florian-beetz:
@koppor This PR is still pretty rough around the edges, and I could not work on this recently.
The biggest problem here right now is that the logging does not work for tests, which means they are currently not runnable in this version.

I also don't know if it makes much sense to continue working on this before #3421 is finished.

tobiasdiez:
> I also don't know if it makes much sense to continue working on this before #3421 is finished.

In the devcall, we agreed with this sentiment. The important thing right now is to have a workable Java 9 build, and then worry about the moduliarization of JabRef itself. Thus, I'll put this PR into sleep mode for now. Thanks nonetheless for your work!

PR #4232, opened 2018-07-24
NameChange and update to JabRef4

JoKalliauer:
1) The name has to be changed because otherwise on Ubuntu it will appear twice in activities: once the icon added to favourites and once the opened instance, see https://unix.stackexchange.com/a/59654/241592 for details. The name has to be identical with the Application.

2) Since Ubuntu 18.04 uses JabRef3, which ruins the grouping of JabRef4, just by starting JabRef! (automatic opening last files and autosave?!?), you should use JabRef4 using java -jar ~/${[[:Folder To Your Jabref:]]}/JabRef-4.3.1.jar
For more details see (Only for the germans developers): https://wiki.ubuntuusers.de/JabRef/ (which should be updated (I think that's the first place to search for Germanspeakings), it says the newest 2.9 version can be installed using sourceforge, maybe you should refer to http://help.jabref.org/en/Installation, which might also should be updated to Ubuntu 18.04.)

3) The icon does not work
You might should specify that it should be downloaded from:
https://github.com/JabRef/jabref/blob/master/src/main/resources/images/external/JabRef-icon-128.png
and should be stored f.e. in ~/.local/share/icons/

----


tobiasdiez:
@koppor can you please have a look at this PR. Thanks!

koppor:
I am not sure how this PR relates to the snap build. The place holder `SNAP` is still in there, but there seems to be a manual place holder (`{[[:Folder To Your Jabref:]]}`). So, this doesn't seem to fully work.

I need more explanation, what this thing does. Maybe a PR on our installation instructions at https://github.com/JabRef/help.jabref.org/blob/gh-pages/en/Installation.md would help.

- [ ] Add an [ADR](https://adr.github.io/madr/ on the `name`. Otherwise, it's a WTF when looking at the name field.
- [ ] It seems that `{[[:Folder To Your Jabref:]]}` raises some input field. I would like to see some automatic installation here. At least, some improvements in our installation description.
- [ ] The icon is still an open issue.
- [ ] Ubuntu Wiki page needs to be updated. We do not maintain it for ourselves, but the Ubuntu user community does it.

@JoKalliauer could you please take care about the installation instructions? You can assume your PR here is merged in that case.

LinusDietz:
Hey @JoKalliauer 👋 can you have a look at the comments and adapt the code so we can move forward with this PR?

JoKalliauer:
@LinusDietz : Thanks for reminding!

If I install jabref according to
https://github.com/JabRef/help.jabref.org/blob/gh-pages/en/Installation.md
I only install java, but I only execute a portable jabref (without creating any jabref.desktop-icon).
Is that correct?
I don't see the installation instruction for JabRef.
Maybe I missed/forget something.

Siedlerchr:
> 
> 
> @LinusDietz : Thanks for reminding!
> 
> If I install jabref according to
> https://github.com/JabRef/help.jabref.org/blob/gh-pages/en/Installation.md
> I only install java, but I only execute a portable jabref (without creating any jabref.desktop-icon).
> Is that correct?
> I don't see the installation instruction for JabRef.
> Maybe I missed/forget something.

You are correct. We only provide the standalone jar file for linux. No *deb package and therenore no icon. and the snap package was a promising idea, but it turns out that the build process is still complex and there need to be some issues resolved on the Linux/Ubuntu side which we can't do anyhting about atm. Either you adapt your code and maybe you can solve that or we decide to close the PR.

JoKalliauer:
> Either you adapt your code and maybe you can solve that or we decide to close the PR.

Maybe we should close this pull request.

Issue #4229, opened 2018-07-23
XXE in MsBibImporter


ProDigySML:
# The Issue
An XML External Entity attack is a type of attack against an application that parses XML input. This attack occurs when XML input containing a reference to an external entity is processed by a weakly configured XML parser. This attack may lead to the disclosure of confidential data, denial of service, server side request forgery, port scanning from the perspective of the machine where the parser is located, and other system impacts.

# Where the Issue Occurred
The following code snippet displays the usage of `DocumentBuilderFactory` without disabling entities:
https://github.com/JabRef/jabref/blob/2f5f2e5bbe2b42319736bd735791b0b3039096f6/src/main/java/org/jabref/logic/importer/fileformat/MsBibImporter.java#L40

The following code snippet displays the parsing of the XML:
https://github.com/JabRef/jabref/blob/2f5f2e5bbe2b42319736bd735791b0b3039096f6/src/main/java/org/jabref/logic/importer/fileformat/MsBibImporter.java#L58

Siedlerchr:
Method for fixing this:
https://www.owasp.org/index.php/XML_External_Entity_(XXE)_Prevention_Cheat_Sheet#JAXP_DocumentBuilderFactory.2C_SAXParserFactory_and_DOM4J

nicksw:
Hi, Do you all mind if I try to give this one a shot? I take it the class to look at is the one linked by ProDigySML?

ProDigySML:
Hey @nicksw! Yes it should be a simple fix. Adding in the above-mentioned prevention measures within the MsBibImporter class should fix the issue.

Siedlerchr:
@nicksw  Feel free to start over! Check out the contribution guideline and go ahead: https://github.com/JabRef/jabref/blob/master/CONTRIBUTING.md

nicksw:
Thanks for the free range @Siedlerchr though I just opted to create a private method with the fixes and also correct a spelling error. My one issue is that I am failing 5 tests when I run `gradlew check` The weird thing is that I fail the same 5 tests with the original MsBibImporter file, see attached . Did I incorrectly setup my workspace? 
[reports.zip](https://github.com/JabRef/jabref/files/2236851/reports.zip)

Siedlerchr:
[...]

PR #4240, opened 2018-07-29
Fix importer vulnerability

nicksw:
Fixed issue where importer was vulnerable to XXE attacks by
disabling DTDs. If feature is unavailable, warning is sent to logger and returns builderFactory at current state. Used [#4229](https://github.com/JabRef/jabref/issues/4229)

----

- [X] Ensured that [the git commit message is a good one](https://github.com/joelparkerhenderson/git_commit_message)


Siedlerchr:
Thanks for your contribution
I just resolved the conflict in the changelog, so that the build runs on Travis. Check the Continous Interation Travis output (just click on it, in case it fails)

src/main/java/org/jabref/logic/importer/fileformat/MsBibImporter.java
```diff
@@ -18,26 +19,31 @@
 import org.xml.sax.SAXException;
 import org.xml.sax.SAXParseException;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 /**
  * Importer for the MS Office 2007 XML bibliography format
- * By S. M. Mahbub Murshed
+ * By S. M. Mahbub Murshed & Nicholas S. Weatherley
```
Siedlerchr: Pleas remove this author comment completely as we handle author credits via our [authors-file](https://github.com/JabRef/jabref/blob/master/AUTHORS)

src/main/java/org/jabref/logic/importer/fileformat/MsBibImporter.java
```diff
@@ -85,4 +92,29 @@ public String getDescription() {
...
+    /**
+     * DocumentBuilderFactory makes a XXE safe Builder factory from dBuild. If not supported by current
+     * XML then returns original builder given and logs error.
+     * @param dBuild | DocumentBuilderFactory to be made XXE safe.
+     * @return If supported, XXE safe DocumentBuilderFactory. Else, returns original builder given
+     */
+    private DocumentBuilderFactory makeSafeDocBuilderFactory(DocumentBuilderFactory dBuild) {
+        String FEATURE = null;
+        try {
+            FEATURE = "http://apache.org/xml/features/disallow-doctype-decl";
```
Siedlerchr: Just some minor improvemtns: I would suggest extracting both urls to two separate variables and make them private static final as constants.

src/main/java/org/jabref/logic/importer/fileformat/MsBibImporter.java
```diff
@@ -85,4 +92,29 @@ public String getDescription() {
...
+            dBuild.setFeature(FEATURE, true);
+
+            FEATURE = "http://apache.org/xml/features/nonvalidating/load-external-dtd";
+            dBuild.setFeature(FEATURE, false);
+
+            dBuild.setXIncludeAware(false);
+            dBuild.setExpandEntityReferences(false);
+
+        } catch (ParserConfigurationException e) {
+            LOGGER.warn("Builder not fully configured. ParserConfigurationException was thrown. Feature:'" +
```
Siedlerchr: slf4j has the nice advantage that it supports variables via curly braces:
https://www.slf4j.org/faq.html#logging_performance

CHANGELOG.md
```diff
@@ -60,7 +60,8 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - We fixed an issue where the "Convert to BibTeX-Cleanup" moved the content of the `file` field to the `pdf` field [#4120](https://github.com/JabRef/jabref/issues/4120)
 - We fixed an issue where the preview pane in entry preview in preferences wasn't showing the citation style selected [#3849](https://github.com/JabRef/jabref/issues/3849)
 - We fixed an issue where the default entry preview style still contained the field `review`. The field `review` in the style is now replaced with comment to be consistent with the entry editor [#4098](https://github.com/JabRef/jabref/issues/4098)
-- We fixed an issue where filles added via the "Attach file" contextmenu of an entry were not made relative. [#4201](https://github.com/JabRef/jabref/issues/4201)
+- We an issue where users were vulnerable to XXE attacks during parsing [#4229](https://github.com/JabRef/jabref/issues/4229)
```
tobiasdiez: We *fixed* an...

src/main/java/org/jabref/logic/importer/fileformat/MsBibImporter.java
```diff
@@ -85,4 +92,29 @@ public String getDescription() {
...
     }
 
+    /**
+     * DocumentBuilderFactory makes a XXE safe Builder factory from dBuild. If not supported by current
+     * XML then returns original builder given and logs error.
+     * @param dBuild | DocumentBuilderFactory to be made XXE safe.
+     * @return If supported, XXE safe DocumentBuilderFactory. Else, returns original builder given
+     */
+    private DocumentBuilderFactory makeSafeDocBuilderFactory(DocumentBuilderFactory dBuild) {
+        String FEATURE = null;
```
tobiasdiez: We have the convention that only static (constant) values are all capitals. Thus, please change the name to `feature`.

src/main/java/org/jabref/logic/importer/fileformat/MsBibImporter.java
tobiasdiez: and please add the exception as a parameter so that the root cause is also logged. (then probably the part "ParserConfigurationException was thrown." is no longer necessary).

nicksw:
Thanks for the feedback. Yes, I agree with all of the fixes. Sorry about the two commits back to back. The first one stated it did not go through on my end as I needed to update HEAD. Also, I am not sure why I am failing both tests. the CI test seems to be related to GUI errors while the Codacy fail is from the extra space before importing items from slf4j.

nicksw:
No I agree. The project should be uniform in import and code style as it would look a mess without. Pushed the changes.

Issue #3823, opened 2018-03-09
File annotations: remove line breaks, increase width, and reduce font size


anassal:
JabRef 4.2-dev--snapshot--2018-03-09--master--988fb54a9
Windows 8.1 6.3 x86 
Java 1.8.0_161

1. Multi-line PDF annotations like these:

![1](https://user-images.githubusercontent.com/3094352/37212695-a907ab08-23a7-11e8-91e1-22c0ece31939.png)

are shown with "line breaks" like this.

![2](https://user-images.githubusercontent.com/3094352/37212696-a9371f6e-23a7-11e8-8a15-455fb09c3b2f.png)

2. There is currently a wide white space between the annotation text and the date/page number. Text should be warped with the maximum available/appropriate width.

3. Then, the height of each annotation item should be "auto".

4. I suggest the font size of the extracted annotations in the left-hand panel be reduced (e.g., x0.6). The right hand panel font size looks OK.

Siedlerchr:
@miderleta You are welcome! Please take a look at our [Contribution guideline](https://github.com/JabRef/jabref/blob/master/CONTRIBUTING.md) for getting started and setting up a workspace etc and you can start. 

For this specific issue it's probably sufficient to play around with the css properties and the fxml.
If you have any questions please ask here or in our [gitter chat](https://gitter.im/JabRef/jabref)

miderleta:
It seems like I cannot generate Javadoc for the project. Is that an issue or am I doing something wrong?

Siedlerchr:
@miderleta  Where do you need it for? Any modern IDE will show you the javadoc automatically for each method - if it's annotated.

Siedlerchr:
Maybe you can differentiate between a new paragraph starting vs a normal line break code wise.

anassal:
@upupming This is a valid point. But can't we reasonably assume that most annotations are in single paragraphs, since paragraphs usually contain different ideas?

For annotations that actually span over multiple paragraphs, we can see if there are statistical differences between "end-of-line" and "end-of-paragraph" linebreak characters, as @Siedlerchr suggested, in PDFs annotated in multiple readers (e.g., Adobe, Foxit, Okular). It is also worth checking if this has been solved in other libraries/tools that extract text from PDF.

Regarding the "auto" height of an annotation item (left-hand panel), I just meant auto height that would result from text wrapping and fixed/limited width. Along with text wrapping and other improvements, the summary could be limited to, say, 10-15 words.

[...]

PR #4246, opened 2018-08-02
Solved #3823 File annotation

upupming:
This pull request solved #3823, and improved the new line remover.

# Removed hyphens and newlines of highlighted text

I removed all new lines which **don't have preceded `.` or `:` becaues I think `.` or `:` is often used to start a new paragraph.

Test text:

Test result:

There is another problem with the regular expression `(\r?\n|\r)` when I try to use `(?<![.|:])(\r?\n|\r)` for excluding the new line end with `.` or `:`. 

If it's on Windows, the `\r\n` will be replaced with a whitespace, and it is OK. However if we encounter `.\r\n`, the `.\r` will be reserved and the `\n` will be replaced by a whitespace. In this way, the `.\r` will not be shown as new line int the field, and we don't need this whitespace. So instead we should change `.\r\n` to `\r\n` directly.

So I use java's `%n` to distinguish new line wisely.

# Font size of summary on the left changed to 10px

# Summary text wrapped

# The Final example:

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,9 +25,9 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - We changed the default keyboard shortcuts for moving between entries when the entry editor is active to ̀<kbd>alt</kbd> + <kbd>up/down</kbd>.
 - Opening a new file now prompts the directory of the currently selected file, instead of the directory of the last opened file.
 - Window state is saved on close and restored on start.
-- Files without a defined external file type are now directly opened with the default aplication of the operating system
+- Files without a defined external file type are now directly opened with the default application of the operating system
 - We streamlined the process to rename and move files by removing the confirmation dialogs.
-- We removed the redundant new lines of marking and wrapped the summary in File annotation tab.
+- We removed the redundant new lines of markings and wrapped the summary in the File annotation tab. [#3823](https://github.com/JabRef/jabref/issues/3823)
 
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - Window state is saved on close and restored on start.
 - Files without a defined external file type are now directly opened with the default aplication of the operating system
 - We streamlined the process to rename and move files by removing the confirmation dialogs.
+- We removed the redundant new lines of marking and wrapped the summary in File annotation tab.
 
 
 
--- a/src/main/java/org/jabref/gui/entryeditor/fileannotationtab/FileAnnotationTabView.java
+++ b/src/main/java/org/jabref/gui/entryeditor/fileannotationtab/FileAnnotationTabView.java
@@ -15,6 +15,7 @@ import javafx.scene.control.Label;
 import javafx.scene.control.ListView;
 import javafx.scene.control.SelectionMode;
 import javafx.scene.control.TextArea;
+import javafx.scene.control.Tooltip;
 import javafx.scene.layout.ColumnConstraints;
 import javafx.scene.layout.GridPane;
 import javafx.scene.layout.VBox;
@@ -63,7 +64,6 @@ public class FileAnnotationTabView {
         annotationList.getSelectionModel().setSelectionMode(SelectionMode.SINGLE);
         annotationList.getSelectionModel().selectedItemProperty().addListener((observable, oldValue, newValue) -> viewModel.notifyNewSelectedAnnotation(newValue));
         ViewModelListCellFactory<FileAnnotationViewModel> cellFactory = new ViewModelListCellFactory<FileAnnotationViewModel>()
-                .withTooltip(FileAnnotationViewModel::getMarking)
                 .withGraphic(this::createFileAnnotationNode);
         annotationList.setCellFactory(cellFactory);
         annotationList.setPlaceholder(new Label(Localization.lang("File has no attached annotations")));
@@ -94,8 +94,13 @@ public class FileAnnotationTabView {
         Label date = new Label(annotation.getDate());
         Label page = new Label(Localization.lang("Page") + ": " + annotation.getPage());
 
-        marking.setStyle("-fx-font-size: 10px; -fx-font-weight: bold");
+        marking.setStyle("-fx-font-size: 0.75em; -fx-font-weight: bold");
         marking.setMaxHeight(30);
+
+        Tooltip markingTooltip = new Tooltip(annotation.getMarking());
+        markingTooltip.setMaxWidth(800);
+        markingTooltip.setWrapText(true);
+        marking.setTooltip(markingTooltip);
 
         // add alignment for text in the list
         marking.setTextAlignment(TextAlignment.LEFT);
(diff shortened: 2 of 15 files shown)


src/main/java/org/jabref/logic/formatter/bibtexfields/RemoveHyphenatedNewlinesFormatter.java
```diff
@@ -10,7 +10,8 @@
  * Removes all hyphenated line breaks in the string.
  */
 public class RemoveHyphenatedNewlinesFormatter extends Formatter {
-    private static final Pattern HYPHENATED_WORDS = Pattern.compile("(-\r\n|-\n|-\r)");
+    private static final String newLine = String.format("%n");
+    private static final Pattern HYPHENATED_WORDS = Pattern.compile("(-" + newLine + ")");
```
tobiasdiez: You can also use `\R` to match any line breaks, see https://docs.oracle.com/javase/9/docs/api/java/util/regex/Pattern.html.

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ We refer to [GitHub issues](https://github.com/JabRef/jabref/issues) by using `#
 - Window state is saved on close and restored on start.
 - Files without a defined external file type are now directly opened with the default aplication of the operating system
 - We streamlined the process to rename and move files by removing the confirmation dialogs.
+- We removed the redundant new lines of marking and wrapped the summary in File annotation tab.
```
tobiasdiez: Slightly better: We removed the redundant new lines of marking*s* and wrapped the summary in *the* File annotation tab.

Please also add a reference to the issue #3823 (cf. e.g. some of the changelog entries under "fixed").

src/main/java/org/jabref/gui/entryeditor/fileannotationtab/FileAnnotationTabView.java
```diff
@@ -94,12 +94,14 @@ private Node createFileAnnotationNode(FileAnnotationViewModel annotation) {
         Label date = new Label(annotation.getDate());
         Label page = new Label(Localization.lang("Page") + ": " + annotation.getPage());
 
-        marking.setStyle("-fx-font-weight: bold");
+        marking.setStyle("-fx-font-size: 10px; -fx-font-weight: bold");
```
tobiasdiez: Please specify the font-size in `em` so that it scales properly if the base font size is changed (e.g. on high res displays).

src/main/java/org/jabref/gui/entryeditor/fileannotationtab/FileAnnotationViewModel.java
```diff
@@ -31,7 +31,11 @@ private void setupContentProperties(FileAnnotation annotation) {
             this.content.set(annotation.getLinkedFileAnnotation().getContent());
             String annotationContent = annotation.getContent();
             String illegibleTextMessage = Localization.lang("The marked area does not contain any legible text!");
-            this.marking.set(annotationContent.isEmpty() ? illegibleTextMessage : annotationContent);
+            String markingContent = (annotationContent.isEmpty() ? illegibleTextMessage : annotationContent);
+            // remove newlines && hyphens before linebreaks
+            markingContent = new RemoveHyphenatedNewlinesFormatter().format(markingContent);
```
tobiasdiez: It is good that you try to reuse existing logic, i.e. the existing formatters. However, these are also used to generate the bibtex key (and file name). For these use cases, it is still desired to remove all line breaks. Thus, I would prefer if you define the modified regex pattern to remove the linebreaks here in this class and leave the RemoveNewlinesFormatter untouched (except changing the pattern to use `\R`).

src/main/java/org/jabref/gui/entryeditor/fileannotationtab/FileAnnotationViewModel.java
```diff
@@ -31,7 +31,11 @@ private void setupContentProperties(FileAnnotation annotation) {
...
+            markingContent = new RemoveNewlinesFormatter().format(markingContent);
+            this.marking.set(markingContent);
```
tobiasdiez: It would be nice if you could add some tests verifying the new behavior: create a `FileAnnotion` object by hand (not by parsing a pdf) which contains such problematic hyphens/newlines and check that they are removed, indeed.

upupming:
@tobiasdiez Thanks for your review! I will fix them soon.

src/main/java/org/jabref/logic/formatter/bibtexfields/RemoveHyphenatedNewlinesFormatter.java
upupming: Thanks, it is what I didn't know before. But it seems `\R` will match any of `\r`, `\n` and `\r\n`.
image
In this example, you can see `\r` is matched on Windows. But we need only match `\r\n` on Windows, so I think `%n` is the right choice.

upupming:
Hi, I reverted the change for `RemoveHyphenatedNewlinesFormatter` and `RemoveNewlinesFormatter`. I have also fixed the tooltip for marking summary.

**Before**

**After**


Now I'm going to add some unit tests!

CHANGELOG.md
upupming: Fixed.

src/main/java/org/jabref/gui/entryeditor/fileannotationtab/FileAnnotationTabView.java
upupming: Changed to `0.75em`

src/main/java/org/jabref/gui/entryeditor/fileannotationtab/FileAnnotationViewModel.java
upupming: Now I leave the two formatter untouched and only use the pattern in this scope.

src/main/java/org/jabref/gui/entryeditor/fileannotationtab/FileAnnotationViewModel.java
upupming: Thank you for your advice, I have added a unit test!

upupming:
All done~ 😃

src/test/java/org/jabref/logic/formatter/bibtexfields/RemoveHyphenatedNewlinesFormatterTest.java
```diff
@@ -15,8 +15,7 @@ public void setUp() {
 
     @Test
     public void removeHyphensBeforeNewlines() {
-        assertEquals("water", formatter.format("wa-\nter"));
-        assertEquals("water", formatter.format("wa-\r\nter"));
-        assertEquals("water", formatter.format("wa-\rter"));
+        String newLine = String.format("%n");
```
tobiasdiez: Please keep these tests (here and below). The text that we want to format might contain linux line endings on Windows and vice versa. But you can of course also add a tests for ̀ %n`. 
Please change also all regex patterns to use `\R` instead of `%n` for newlines.

upupming: Finally leaving them untouched. 

In summary, Pattern.compile("...") changed from `(\r?\n|\r)` to `\\R`

tobiasdiez:
Thanks for the quick follow-up! Looking forward to your next PR ;)

PR #4252, opened 2018-08-08
Fix attach file does not relativize file path

### LINKED ISSUE #4241: added file link has absolute instead of relative path

JabRef version 4.3.1 on Windows 10

Jabref generates an absolute path to file if the relative path would "climb" the directory tree

Steps to reproduce:
0. assume you have a file "C:\foo\file.pdf"
1. set file directory (whether main, general, or user-specific) to "C:\bar"
2. right click on any JabRef entry -> select "Attach file"
3. navigate to and select "C:\foo\file.pdf"
4. The path to file will be stored in absolute format "C:\foo\file.pdf" instead of relative "..\foo\file"

Note that JabRef (at least on Windows) still handles correctly relative paths which include "..\" (or start with "\\") as long as you type them yourselves.

Siedlerchr:
Fixes #4241 

Attached files were only made relative before if you hit the browse button

----

### DIFF
--- a/src/main/java/org/jabref/gui/filelist/LinkedFilesEditDialogViewModel.java
+++ b/src/main/java/org/jabref/gui/filelist/LinkedFilesEditDialogViewModel.java
@@ -84,11 +84,8 @@ public class LinkedFilesEditDialogViewModel extends AbstractViewModel {
             // Store the directory for next time:
             preferences.setWorkingDir(path);
 
-            // If the file is below the file directory, make the path relative:
-            List<Path> fileDirectories = database.getFileDirectoriesAsPaths(preferences.getFileDirectoryPreferences());
-            path = FileUtil.shortenFileName(path, fileDirectories);
+            relativizeFile();
 
-            link.set(path.toString());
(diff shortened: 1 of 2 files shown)


src/main/java/org/jabref/gui/filelist/LinkedFilesEditDialogViewModel.java
```diff
@@ -128,4 +128,12 @@ public LinkedFile getNewLinkedFile() {
         return new LinkedFile(description.getValue(), link.getValue(), monadicSelectedExternalFileType.map(ExternalFileType::toString).getOrElse(""));
     }
 
+    private void relativizeFile() {
```
tobiasdiez: Right now, it is hard to tell from the outside on which path this methods operates. I would thus prefer if `relativizeFile` would accept a path and return the relativized path (probably also rename to `relativize[Path]`). Then in would look like `link.set(relativize(path))`.

Siedlerchr: I already had considered this option as well, but seem to forget about it ;)

PR #4253, opened 2018-08-08
Fix for issue 3861 : Preferences Dialog to javafx

Super-Tang:
----


1160300311:
Below are what we have done to the dialog Options->Preference. The left is the original.

fc614eb383bd210d1852503555065eb
c75201e95ecb9927765952a9bcc653e
cae239f92de3b1c952596cbf1a7bdce
02c337e56200c1496d4477377dedd76
3c1f226f65035bda9d3482c106b2393
f34b59245e53e45fda0eb1f36131d75

Siedlerchr:
I think it would be easier to rebase your changes in the branch on the current (upstream, the JabRef project master) master branch, so that would hopefully avoid that conflitcts.
https://git-scm.com/book/de/v2/Git-Branching-Rebasing

1160300311:
Thanks for the advice, I am going  to fix it. @Siedlerchr

src/main/java/org/jabref/gui/bibtexkeypattern/BibtexKeyPatternPanel.java
```diff
@@ -68,10 +68,11 @@ private void buildGUI() {
         sp.setBorder(BorderFactory.createEmptyBorder());
         pan.setLayout(gbl);
         setLayout(gbl);
-
+        javafx.scene.text.Font font1 = new javafx.scene.text.Font(10);
         // The header - can be removed
         JLabel lblEntryType = new JLabel(Localization.lang("Entry type"));
```
Siedlerchr: missed one ;)

src/main/java/org/jabref/gui/bibtexkeypattern/BibtexKeyPatternPanel.java
```diff
@@ -87,7 +88,8 @@ private void buildGUI() {
 
         JLabel lblKeyPattern = new JLabel(Localization.lang("Key pattern"));
```
Siedlerchr: here, too

1160300311: I see that, I will do that soon :)

CHANGELOG.md
```diff
@@ -130,3 +395,5 @@ The changelog of JabRef 2.11 and all previous versions is available as [text fil
 [4.0-beta]: https://github.com/JabRef/jabref/compare/v3.8.2...v4.0-beta
 [2.11.1]: https://github.com/JabRef/jabref/compare/v2.11...v2.11.1
 [JavaFX]: https://en.wikipedia.org/wiki/JavaFX
+
+>>>>>>> f7f5841bdb1648947cbd257f1128fa46c0c7b524
```
Siedlerchr: You still have this conflict markers in the file. The easiest solution to fix this would be to just copy the current changelog.md from master and paste it in your file locally and then commit the change.
So then git should be able to recognize that it didn't change

Siedlerchr:
Are you using intellij or Eclipse? Many diffs see only related to the code style.  You should better import the correct code style.
https://github.com/JabRef/jabref/tree/master/config

1160300311:
Thanks for the advice. I have imported the correct  code style, but I always forgot to check the code style after I made some changes @Siedlerchr

Siedlerchr:
@1160300311  Regarding the code style: There are keybord shortcuts in Intellij /Eclipse for code formatting and + organizing imports. So before commit just hit the shortcut and no more build errors due to code style.
https://stackoverflow.com/a/17880058
In Eclipse it's ctrl + shift + f and ctrl + shift +o (the latter one for organiziung imports)

1160300311:
Thanks for the advice. I still have a problem: The Travis CI build failed due to some test failures, but it goes ok when I run the test on my computer.  @Siedlerchr

Siedlerchr:
It looks good now on Travis. It can sometimes happen that some tests relying on external data sources/websites fail, because they are executed too often and or are related to tempoary outages.
If it's not related to your changes it seems to work.

PS: Now you've screwed up the changelog format again, I think...

1160300311:
Is BibtexKeyPatternPanel.java only used for generating a swing panel that used in the Preferences Dialog ? I am afraid other dialogs will go wrong if I removed the Swing controls. 
In addition, there are still some Swing controls in ExternalTab.java and PreviewPrefsTab.java, because there are some Swing controls I don't know how to convert.  @Siedlerchr

Siedlerchr:
@1160300311 you can easily find this out by selecting the class name and by showing the Call Hierarchy of the class/constructor. That will allow you to find out to find where the class /method is actually used.

Before I forget, it would be nice if you could change the method  to Double.parseDouble instead of integer.parseInt in org.jabref.gui.preftabs.TableColumnsTab.setValues(TableColumnsTab.java:415) 
Because when I start an older JabRef version and then the latest dev version,I will get an exception as the column sizes are not integers. So it would be nice if you could include that fix


In the case of the BibtexKeyPatternPanel it is used in two positions, in the Preferences and it's used in a dialog (Options -> BibTexKeyPattern).
If you convert the panel to a javafx Node in the method, you can simplify the dialog by doing sth like this:

```
 DialogService ds = new FXDialogService();
     DialogPane pane = new DialogPane();
                    pane.setContent(bibtexPane) //The converted javafx pane;
       ds.showCustomDialogAndWait(title, pane, buttonTypes)
```
There is an example for this somewhere in the PreviewPrefsTab.

Your second questoin: The main problem in PreviewPrefsTab seems to be the SwingWorker. The SwingWorker can be easily converted to our BackgroundTask: 

```
 BackgroundTask
                    .wrap(() ->  taskto execute ))
                    .onRunning(() -> do sth )
                    .onFinished(() ->  do sth))
                    .onSuccess( result-> {
                    do sth with result
                    })
                    .onFailure(dialogService::showErrorDialogAndWait)
                    .executeWith(taskExecutor);
```

And example for the method can be found in IdentifierEditorViewModel.

1160300311:
Thanks for the advice,  it may take me a few days to learn how to convert the SwingWorker to the BackgroundTask :(. I will change that ASAP after I finish this learning journey :). @Siedlerchr

Siedlerchr:
No hassle! Take your time. In short the workers are there to execute (long running) tasks in the background thread so that the gui does not freeze /block. 

Feel free to ask if you need any help.

Siedlerchr:
I found and fixed a bug regarding the storing of settings when you change the language.
I created a patch. You need to download the patch, and remove the .txt extension (github doesnt allow .patch), then apply the patch with 
`git am -3 < 0001-fix-storing-of-language-settings-bibtex.patch`

[0001-fix-storing-of-language-settings-bibtex.patch.txt](https://github.com/JabRef/jabref/files/2280871/0001-fix-storing-of-language-settings-bibtex.patch.txt)

1160300311:
I have solved fix the NullPointerException in setting for External Application. Then I run the JabRefMain.java, it goes ok sometimes, like this  
1cdff6dfb27e671db2cd6d05e045900
but mostly, the program doesn't respond.
d91d7626f487c9559c934a9357d3e47
Is that OK or I have messed sth up ? @Siedlerchr

Siedlerchr:
It looks right so far. The Settings Dialog is still an old Swing dialog `PushToApplicationSettingsDialog` together with `PushToApplicationSettings`
 which will be displayed as modal and blocking. You have to first close the swing dialog to make the rest of the application work again. If the dialog is not visible, use Alt+Tab to switch to the window),

For the moment I would let this dialog stay as is and concentrate to convert the rest of the Panes as it is a bit out of scope of this PR:
I added this to the list in #3861 
You can embed Swing in JavaFX by using a [SwingNode](https://docs.oracle.com/javase/8/javafx/api/javafx/embed/swing/SwingNode.html)

src/main/java/org/jabref/JabRefGUI.java
```diff
@@ -54,11 +54,10 @@ public JabRefGUI(Stage mainStage, List<ParserResult> argsDatabases, boolean isBl
 
         // passed file (we take the first one) should be focused
         focusedFile = argsDatabases.stream()
-                                   .findFirst()
-                                   .flatMap(ParserResult::getFile)
-                                   .map(File::getAbsolutePath)
-                                   .orElse(Globals.prefs.get(JabRefPreferences.LAST_FOCUSED));
-
+                .findFirst()
```
tobiasdiez: The code formatting was correct before.

build.gradle
```diff
@@ -194,12 +194,12 @@ dependencyUpdates.resolutionStrategy = {
         }
         rules.withModule("org.controlsfx:controlsfx") { ComponentSelection selection ->
             if (selection.candidate.version ==~ /9.*/) { // Reject version 9 or higher
-                 selection.reject("Cannot be updated to 9.*.* until Jabref works with Java 9")
+                selection.reject("Cannot be updated to 9.*.* until Jabref works with Java 9")
             }
         }
-		rules.withModule("com.github.tomtung:latex2unicode_2.12") { ComponentSelection selection ->
+        rules.withModule("com.github.tomtung:latex2unicode_2.12") { ComponentSelection selection ->
```
tobiasdiez: I guess these formatting changes are copied from a different PR. Did you merged the latest changes from master?

src/main/java/org/jabref/gui/SaveOrderConfigDisplay.java
```diff
@@ -38,56 +46,87 @@ private void init() {
         fieldNames.add(BibEntry.KEY_FIELD);
         Collections.sort(fieldNames);
         String[] allPlusKey = fieldNames.toArray(new String[fieldNames.size()]);
-        savePriSort = new JComboBox<>(allPlusKey);
+        savePriSort = new ComboBox<>(FXCollections.observableArrayList(allPlusKey));
```
tobiasdiez: You no longer need to use an array. Instead, directly use the `fieldNames` collection.

src/main/java/org/jabref/gui/SaveOrderConfigDisplay.java
```diff
@@ -38,56 +46,87 @@ private void init() {
...
-        saveSecDesc = new JCheckBox(Localization.lang("Descending"));
-        saveTerDesc = new JCheckBox(Localization.lang("Descending"));
-
+        savePriDesc = new CheckBox(Localization.lang("Descending"));
+        saveSecDesc = new CheckBox(Localization.lang("Descending"));
+        saveTerDesc = new CheckBox(Localization.lang("Descending"));
+
+        Font font = new Font(10);
+        GridPane builder = new GridPane();
+        Label label = new Label(Localization.lang("Primary sort criterion"));
```
tobiasdiez: Please use descriptive names for the controls, here eg `primarySortLabel`. This essentially applies to all controls on all tabs.

src/main/java/org/jabref/gui/SaveOrderConfigDisplay.java
```diff
@@ -38,56 +46,87 @@ private void init() {
...
+        label.setFont(font);
+        builder.add(label,1,1);
+        builder.add(savePriSort,2,1);
```
tobiasdiez: Our code formatting convention is that there should be a space after the comma.

src/main/java/org/jabref/gui/SaveOrderConfigDisplay.java
```diff
@@ -38,56 +46,87 @@ private void init() {
...
+        Label label2 = new Label(Localization.lang("Tertiary sort criterion"));
+        label2.setFont(font);
+        builder.add(label2,1, 3);
+        builder.add(saveTerSort,2,3);
+        builder.add(saveTerDesc,3,3);
+        panel = builder;
+
+        JComboBox<String> savePriSort1 = new JComboBox<>(allPlusKey);
+        savePriSort1.setEditable(true);
+        JComboBox<String> saveSecSort1 = new JComboBox<>(allPlusKey);
```
tobiasdiez: This part here looks like a unnecessary code duplication.

src/main/java/org/jabref/gui/help/AboutDialogView.java
```diff
@@ -31,8 +31,8 @@ public AboutDialogView() {
         this.setResizable(true);
 
         ViewLoader.view(this)
-                  .load()
-                  .setAsDialogPane(this);
+                .load()
```
tobiasdiez: The formatting was ok before (applies also to some other code style changes).

src/main/java/org/jabref/gui/preftabs/AdvancedTab.java
```diff
@@ -41,56 +43,79 @@ public AdvancedTab(DialogService dialogService, JabRefPreferences prefs) {
...
-        builder.append(useCaseKeeperOnSearch);
-        builder.nextLine();
-        builder.append(pan);
-        builder.append(useUnitFormatterOnSearch);
-
-        pan = builder.getPanel();
-        pan.setBorder(BorderFactory.createEmptyBorder(5, 5, 5, 5));
-        setLayout(new BorderLayout());
-        add(pan, BorderLayout.CENTER);
+        Font font = new Font(10);
```
tobiasdiez: Instead of specifying the font in the code for every tab again, it is better to create one css file for the whole preference dialog. Then you can add a style class `heading` with a bigger font size (and maybe higher bottom padding). In the code, you then simply use `useRemoteServer.setStyleClass("heading")`.

src/main/java/org/jabref/gui/bibtexkeypattern/BibtexKeyPatternPanel.java
```diff
@@ -153,14 +183,61 @@ private void buildGUI() {
...
+    private void addExtraText() {
+        Label []label = new Label[19];
+        Button []button = new Button[19];
+        for (int i = 0; i <= 18; i++) {
+            textFieldArray[i] = new TextField();
+            button[i] = new Button("Default");
+            button[i].setFont(new javafx.scene.text.Font(10));
+            button[i].setOnAction(e-> defaultPat.setText((String) Globals.prefs.defaults.get(JabRefPreferences.DEFAULT_BIBTEX_KEY_PATTERN)));
+        }
+        label[0] = new Label("Article");
```
tobiasdiez: Instead of hard-coding these fields, you can use a foreach loop like `for (EntryType type : EntryTypes.getAllValues(mode)) `

src/main/java/org/jabref/gui/SaveOrderConfigDisplay.java
tobiasdiez: Ok, in this case you can use a `JFXPanel` in `DatabasePropertiesDialog` to display the JavaFX controls.

src/main/java/org/jabref/gui/preftabs/AppearancePrefsTab.java
```diff
@@ -54,12 +40,12 @@ public AppearancePrefsTab(DialogService dialogService, JabRefPreferences prefs)
...
 
-        VBox container = new VBox();
         container.getChildren().addAll(overrideFonts, fontSizeContainer, fontTweaksLAF);
 
-        JFXPanel panel = CustomJFXPanel.wrap(new Scene(container));
-        panel.setBorder(BorderFactory.createEmptyBorder(5, 5, 5, 5));
-        add(panel, BorderLayout.CENTER);
+    }
+
+    public VBox getContainer() {
```
Siedlerchr: Better set the return type to Node, it's always better to return the Interface Type (https://www.quora.com/What-is-the-use-of-having-an-interface-as-a-return-type-in-Java)

src/main/java/org/jabref/gui/preftabs/ExportSortingPrefsTab.java
```diff
@@ -1,81 +1,74 @@
...
+        builder.add(exportSortOrder, 1, 1);
+        builder.add(new Separator(), 2, 1);
+        builder.add(exportInOriginalOrder,  1, 2);
+        builder.add(new Line(), 2, 3);
+        builder.add(exportInTableOrder,  1, 4);
+        builder.add(new Line(), 2, 5);
+        builder.add(exportInSpecifiedOrder, 1, 6);
+        builder.add(new Line(), 2, 7);
 
         exportOrderPanel = new SaveOrderConfigDisplay();
```
Siedlerchr: If you directly initialize this SaveOrderConfigDisplay Variable at declaration above, you can change your event listener to this neat lambda: 

```

      EventHandler<ActionEvent> listener = (event) -> {
            boolean selected = event.getSource() == exportInSpecifiedOrder;
            exportOrderPanel.setEnabled(selected);
        };
```

Super-Tang: Thanks for the advice :). Done it

src/main/java/org/jabref/gui/bibtexkeypattern/BibtexKeyPatternPanel.java
Siedlerchr: Still here

src/main/java/org/jabref/gui/bibtexkeypattern/BibtexKeyPatternPanel.java
```diff
@@ -88,6 +102,10 @@ private void buildGUI() {
         con.gridy = 1;
         con.gridx = 0;
         JLabel lab = new JLabel(Localization.lang("Default pattern"));
```
Siedlerchr: still and old jlabel

src/main/java/org/jabref/gui/bibtexkeypattern/BibtexKeyPatternPanel.java
```diff
@@ -97,6 +115,11 @@ private void buildGUI() {
         JButton btnDefault = new JButton(Localization.lang("Default"));
```
Siedlerchr: old jbutton. Please remove all rest of the old swing stuff and it would be good to merge

Super-Tang:
I am sorry for forgetting this， done it. :) @Siedlerchr

PR #4254, opened 2018-08-08
[WIP] Try to update jaxb modules

Siedlerchr:
2.4.x is finally java 9 compatible. This is a kind of POC to check if it is generally working,

Currently it  fails because an external schema cannot be accesed due to security restrictions. There is a switch, a system property-
The problem is that the system property jvm arg is not passed to the ant task. And I don't have any deeper gradle experience. So if someone might take a look at this. 

Fixes https://github.com/koppor/jabref/issues/276

----

### DIFF
--- a/xjc.gradle
+++ b/xjc.gradle
@@ -1,13 +1,25 @@
 configurations {
     xjc
 }
-
-dependencies {
-    // Cannot be updated.
-    xjc 'com.sun.xml.bind:jaxb-xjc:2.2.4-1'
+repositories {
+        mavenLocal()
+        jcenter()
+        maven {
+            url 'https://oss.sonatype.org/content/groups/public'
+        }
+          maven {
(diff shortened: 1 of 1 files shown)


Siedlerchr:
@florian-beetz  Maybe you can find a way to get this working? These v2.4.x would be finally jdk 9 compatible

tobiasdiez:
Any update on this one? Maybe it is easier to directly fix the security exception (by including the schema?)?

Siedlerchr:
I think the integration of the jaxb stuff is a more java9 related task as it is splitted into new dependencies. 
So mabye @florian-beetz can integrate that in the jdk9 branch

Siedlerchr:
The dependencies are now part of Eclipse java ee4j https://github.com/eclipse-ee4j/jaxb-ri

tobiasdiez:
Included in https://github.com/JabRef/jabref/pull/3421.

PR #3394, opened 2017-11-02
[WIP] Auto clean up url links(https://github.com/koppor/jabref/issues/254)

sudoHackIn:
* Created unit test class for Formatter

* Add Localized value key

* Add formatter to formatter list with category OTHERS

* Add URLCleanup menu item for URLEditor context menu

* Add generalized interface to EditorTextArea for uder-defined paste action handeling

* Add menu item with url cleanup functional to EditorMenus class

* Set url cleanup on paste action for TextArea in URLEditor class

* Add disable URLCleanupMenu if selected text is empty

* Update CHANGELOG.md

- [+] Change in CHANGELOG.md described
- [+] Tests created for changes
- [+] Manually tested changed features in running JabRef
- [+] If you changed the localization: Did you run `gradle localizationUpdate`?

I am currently a new person in this project and this is my first pull request ever. Could you please give me a full feedback. I have questions about logging correctness, working with new menu item add disabling menu item, if selected text is empty. Is my commit corresponds to rules? Is it correct to rebase few commits into one before pull request?
Hope you understand my questions and my tediousness)

PR #4255, opened 2018-08-09
Fix for issue koppor 254 (Auto cleanup url)

1161140207:
This issue is [koppor#254](https://github.com/koppor/jabref/issues/254).
Set auto cleanup url links on paste action and add cleanup url menu item to the url field
Most codes are modified from a previous contributor in https://github.com/JabRef/jabref/pull/3394.

Currently the menu item list changed as follows(with defalut editor menu, change case, convert, etc):

But I am not sure whether it's necessary to add defalut editor menu. If not the menu item list would be:

I have run `./gradlew check` but failed. But even if I switch to the master branch which is the same as JabRef/master, the build still failed with the same error message like:

But it seems that I did not modify related classes. I don't know what went wrong. It confuses me a lot.

I am looking forward to your feedback.

----


src/main/java/org/jabref/gui/fieldeditors/EditorTextArea.java
```diff
@@ -65,4 +72,42 @@ public void populateContextMenu(ContextMenu contextMenu) {
...
+    @FunctionalInterface
+    public interface PasteActionHandler {
+
+        void handle();
+    }
+  
+    /**
+     * Empty interface implementation to do nothing external on paste method
+     */
+    private static class EmptyPasteHandler implements PasteActionHandler {
```
Siedlerchr: You could add this simply as default method in the interface, so you won't need that emptyPastHandller I think
https://dzone.com/articles/interface-default-methods-java

1161140207: Thanks for your feedback. I will try to fix this.

src/main/java/org/jabref/gui/fieldeditors/contextmenu/EditorMenus.java
```diff
@@ -84,4 +85,24 @@
...
+    /**
+     * The default context menu with a specific menu item to cleanup  URL.
+     *
+     * @param textArea text-area that this menu will be connected to
+     * @return menu containing items of the default menu and an item to cleanup a URL
+     */
+    public static Supplier<List<MenuItem>> getCleanupURLMenu(TextArea textArea) {
+        return () -> {
+            CustomMenuItem cleanupURL = new CustomMenuItem(new Label(Localization.lang("Cleanup URL Link")));
+            //cleanupURL.setDisable(true);
```
tobiasdiez: remove comment or better bind the disable property to `textArea.textProperty().isEmpty()`

src/main/java/org/jabref/gui/fieldeditors/contextmenu/EditorMenus.java
```diff
@@ -84,4 +85,24 @@
...
+            cleanupURL.setOnAction(event -> textArea.setText(new CleanupURLFormatter().format(textArea.getText())));
+
+            List<MenuItem> menuItems = new ArrayList<>();
+            menuItems.add(cleanupURL);
+            menuItems.add(new SeparatorMenuItem());
+            menuItems.addAll(getDefaultMenu(textArea).get());
```
tobiasdiez: It should be the second option in your screenshots since "Protect terms" and the other options do not make sense for the URL field.

src/main/resources/l10n/JabRef_en.properties
```diff
@@ -2227,6 +2227,8 @@ Abbreviate\ journal\ names\ (ISO)=Abbreviate journal names (ISO)
 Abbreviate\ journal\ names\ (MEDLINE)=Abbreviate journal names (MEDLINE)
 Blog=Blog
 Check\ integrity=Check integrity
+Cleanup\ URL\ Link=Cleanup URL Link
```
tobiasdiez: it should be lower case `link` in both cases

tobiasdiez: it should be lower case `link` in both cases

Issue #3959, opened 2018-04-16
Migrate all tests to JUnit 5


stefan-kolb:

Siedlerchr:
Our Importer/Exporter and a lot of our other tests rely on this feature:

Refs https://github.com/junit-team/junit5/issues/1247
Currently planned for RC1

LinusDietz:
Do we really need to migrate? The Junit5 vintage engine should run the old junit4 tests. I already removed the old dependencies at https://github.com/JabRef/jabref/pull/3580

stefan-kolb:
@LinusDietz Good question. If JUnit 4 support does not introduce additional dependencies, but is included in JUnit 5 per default, it is really just nice to have migration and not necessary. Could also be a nice beginners task then.

Siedlerchr:
The TempoaryFolder extension will be published in the junit-pioneer project: https://github.com/junit-pioneer/junit-pioneer/issues/39

tobiasdiez:
The TempDir extension is now part of https://github.com/junit-pioneer/junit-pioneer and a test conversion to this extension can be found in https://github.com/JabRef/jabref/commit/3c6bbd2fc599ad4d24be4bdd62993cd5df4862e1 (sorry for pushing it directly onto the master).

Siedlerchr:
Very good. So then we can go ahead converting the rest. I think this was the only issue left.

koppor:
Adressed by https://github.com/JabRef/jabref/pull/5394.

PR #4258, opened 2018-08-11
migrate  to Junit5

1160300522:
[#3959](https://github.com/JabRef/jabref/issues/3959)
We migrate some packages imported from JUnit4 to JUnit5, such as org.junit.Test --> org.junit.jupiter.api.Test,  org.junit.experimental.categories.Category --> org.junit.jupiter.api.Tag, and so on. 
At the same time, we slightly modify some codes which rely on those packages and doesn't apply to JUnit5 but applys to JUnit4. For example,  we change @Test(excepted = XXX.class) into using assertThrows and lambda, @Category(XXX.class) into @Tag(“XXX”)  , @Before into @BeforeEach , and so on.

----

### DIFF
--- a/src/test/java/org/jabref/gui/AbstractUITest.java
+++ b/src/test/java/org/jabref/gui/AbstractUITest.java
@@ -18,11 +18,11 @@ import org.assertj.swing.image.ScreenshotTaker;
 import org.assertj.swing.junit.testcase.AssertJSwingJUnitTestCase;
 import org.assertj.swing.timing.Pause;
 import org.junit.Assert;
-import org.junit.experimental.categories.Category;
+import org.junit.jupiter.api.Tag;
 
 import static org.assertj.swing.launcher.ApplicationLauncher.application;
 
-@Category(GUITest.class)
+@Tag("GUITest")
 public abstract class AbstractUITest extends AssertJSwingJUnitTestCase {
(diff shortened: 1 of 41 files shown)


Siedlerchr:
First of all thanks for your contribution! 👍  I just took a quick look at the code and it seems you're on a good way.
It would be really nice, if you convert the Temporary Folder Rule from junit4 to junit5 as well.
@tobiasdiez already implemented a sample conversion in a test:
https://github.com/JabRef/jabref/commit/3c6bbd2fc599ad4d24be4bdd62993cd5df4862e1

And in addition, it would ne also nice if you could migrate the reaming Parametrized tests (The old ones rely on the RunWith).  Examples for the parametrized test can be found in the exporters I think

I think we can get rid of the swing gui tests as they are merely obsolete.

tobiasdiez:
Now that another PR (#4260) was opened witch converts the tests from Junit 4 to 5, I would propose the following:
- You make this PR compile (i.e fix all the failing tests) in a minimal intrusive way (i.e. no further conversion of other tests)
- We merge this PR and then afterwards #4260 
- Everything that is still Junit 4 at this stage, can be converted in further PRs.

Siedlerchr:
Obsolete due to #4260

PR #4260, opened 2018-08-12
Fix for issue 3959: migrate all tests to JUnit 5

Metatronwings:
Hi, I'm trying to fix the issue #3959 , and I have some problems here. Can someone help me?

**1. Some tests don't seem to have a parameter source**
Such as  `BibTeXMLExporterTest.java`, `MSBibExportFormatTestFiles.java`, `ModsExportFormatTestFiles.java`, and so on. Since I can't get feedback from both Travis CI and my IDE, I couldn't do anything at the moment.
image

**2. Something about the external libraries**
For `MainArchitectureTestsWithArchUnit.java`, 
`testCompile 'com.tngtech.archunit:archunit-junit:0.8.3'` **isn't enough for migrate this test to JUnit 5.**
Just a few days ago, `ArchUnit` just supported tests with only JUnit 5 package. But this feature needs a `testCompile 'com.tngtech.archunit:archunit-junit:0.9.0` and as far as I know, this version has not been released yet.
For `CiteKeyBasedFileFinderTest.java`, I haven't found an extension in JUnit 5 for the `MockitoJUnitRunner.class` in JUnit 4. And it seems that their (mock) official version hasn't been release yet, too.

**3. The Database Test**
As I can see, those tests have already been migrated partly to JUnit 5. But the rest of it is the PROBLEM.
I've tried to migrate `DBMSProcessorTest.java` , but the code is no longer concise after that.
For example, 
image
this is the old version, and
image
that is the new version. And that is only part of it.
The reason is that I couldn't pass the parameters to the `@BeforeEach` method in JUnit 5, but JUnit 4 can use those local variable to pass the parameters. I think this is where the difference occurred.

**4. The GUI Test**
I just change the `@Category` annotation to `@Tag` annotation, as I have known that those tests are mostly obsolete.

**5. About CleanupWorkerTest**
This test is not hard to migrate from JUnit 4 to 5. But when I finish, and run the tests on my IDE, two tests failed. And the error message is quite weird. It seems that the worker doesn't work in JUnit 5 tests.
image

And if there are any errors or something can be improved in other tests that I didn't mention, please let me know! Thanks!

----


tobiasdiez:
Thanks for your work! Most of the conversion looks good. Concerning your questions:
1. The exporter and importer tests use a filename-based convention to find and check the tests. For example, 
https://github.com/JabRef/jabref/blob/57126659b4f8ff58c9403af0f4d6aabccf77058c/src/test/java/org/jabref/logic/exporter/ModsExportFormatTestFiles.java#L59-L66
finds all files that start with `Mods`. There are usually two files with the same name: one with `bib` ending and one with `xml`. The tests then should test that exporting the content of the bib file using the Mods-exporter give the contents of the `xml` file.
2. Yes, there is not yet a release of the new archunit. So, for now, the architecture tests can stay using Junit 4.
The mockito runner is not really needed. Its only purpose is to support the `@Mock` annotation, but you can always use `variable = mock(...)`. 
3. Currently, there is no neat way to handle this problem (see the open https://github.com/junit-team/junit5/issues/878).
4. Looks good.
5. I think, the problem is the setup in https://github.com/JabRef/jabref/pull/4260/files#diff-2e1fadfb1af8461d9d830981c338f225R64. The bib file should reside on the same level as the pdf folder (and not within).

Metatronwings:
@tobiasdiez Hi! I've done some improments. But I still have some questions.

1. I understand what you say. But when I try to enable this line 
`stream.forEach(n -> System.out.println(n));` with no other code changed, I still get nothing. Looks like there're no such files on my laptop. I suspect that it is the cause of this problem.

2. Done.

3. I'll keep those tests for the time being, as that kind of code is just frustrating. And of course, if you have any requirements, I'll use the "complicated" way to handle it.
 
4. Done.

5. I have a try, the `cleanupRelativePathsConvertAbsoluteToRelativePath()` test passed, but the `cleanupRenamePdfRenamesRelativeFile()` test still failed. So I was wondering if you can give me a further guide about this? It can be a big help.
image

BTW, the tests in `FileUtilTest.java` use `create()` method, which is totally not recommended in the API guide. So when migrating it to JUnit 5, I name a folder manually to imitate the original tests. And you can see this in the new version of `FileUtilTest.java`. If I do something wrong, tell me and I'll do my best to fix them soon.

src/test/java/org/jabref/logic/cleanup/CleanupWorkerTest.java
```diff
@@ -229,117 +238,121 @@ public void cleanupMoveFilesMovesFileFromSubfolder() throws IOException {
...
+        assertEquals(Optional.of(FileFieldWriter.getStringRepresentation(newFileField)), entry.getField("file"));
     }
 
     @Test
-    public void cleanupRenamePdfRenamesRelativeFile() throws IOException {
+    void cleanupRenamePdfRenamesRelativeFile(@TempDirectory.TempDir Path bibFolder) throws IOException {
         CleanupPreset preset = new CleanupPreset(CleanupPreset.CleanupStep.RENAME_PDF);
 
-        File tempFile = bibFolder.newFile();
+        Path path = bibFolder.resolve("AnotherRandomlyNamedFile");
```
tobiasdiez: Use `AnotherRandomlyNamedFile.pdf` here

src/test/java/org/jabref/logic/cleanup/CleanupWorkerTest.java
```diff
@@ -229,117 +238,121 @@ public void cleanupMoveFilesMovesFileFromSubfolder() throws IOException {
...
+        Files.createFile(path);
+        File tempFile = path.toFile();
         BibEntry entry = new BibEntry();
         entry.setCiteKey("Toot");
         LinkedFile fileField = new LinkedFile("", tempFile.getAbsolutePath(), "");
         entry.setField("file", FileFieldWriter.getStringRepresentation(fileField));
 
         worker.cleanup(preset, entry);
         LinkedFile newFileField = new LinkedFile("", "Toot.tmp", "");
```
tobiasdiez: If you use pdf as an extension above, then this should be also `Toot.pdf`. (The temp file rule always creates `tmp` files, thus the behavior before was correct).

.travis.yml
```diff
@@ -59,8 +59,7 @@ after_failure:
 
 branches:
   only:
-    - master
-    - maintable-beta
+    - fix-for-issue-3959
```
Siedlerchr: Is there a reason why you disabled travis for the master branch?
This prevents executing travis for all PR that target master branch

Metatronwings: Oh, sorry. I forgot to restore it as it is.
I was using Travis CI to check some tests before, as those tests couldn't be run on my laptop. I'll fix it immediately.

Siedlerchr:
The failing architecture test is one thing I already experienced a  while ago. It's because it checks for public class.  I was so free to create a  simple patch which fixes this.
simply remove the .txt extension after download and execute 
` git am < 0001-fix-architecture-test-by-removing-check-for-public.patch`

[0001-fix-architecture-test-by-removing-check-for-public.patch.txt](https://github.com/JabRef/jabref/files/2283533/0001-fix-architecture-test-by-removing-check-for-public.patch.txt)
`

tobiasdiez:
The files are in `src\test\resources\org\jabref\logic\exporter` but for some reason the tests don't find them (even on travis) and thus fail.

Metatronwings:
@tobiasdiez Thanks for the previous comment! I've got a little bit about the "cannot find test files" problem. And I've fixed them in 2 commits. The first uses the origin JUnit 4, the second uses JUnit 5.
But some tests in `ModsExportFormatTestFiles.java` and `MSBibExportFormatTestFiles.java` failed. They failed both in the origin JUnit 4 and the version I write JUnit 5. Here is an example for `MSBibExportFormatTestFiles.java` .
image
image

So is that OK just for the migrate task? Is there something else I can do to improve that?

And also a big thank for @Siedlerchr 's patch. ;)

Siedlerchr:
I remember that we had problems with this before, that some resources were not int he right folder. I will take a look at that.

tobiasdiez:
That now some of the exporter tests fail, were my fear. This probably needs further investigations and fixes for the exporter; thus it goes beyond the current PR. Thus, I propose to mark the tests as ignored/disabled for now and we will come back later to fix them.

Siedlerchr:
I am just investigating this and I think I found the issue with the paths

Siedlerchr:
It's a classpath issue, for exmaple:
```
        Path path = Paths.get(MSBibExportFormatTestFiles.class.getResource("").toURI());
 => Returns E:\workspace\JabRef\jabref\bin\main\org\jabref\logic\exporter
``` 
although it should be bin/test

Siedlerchr:
Okay, I found the issue with the paths. I really wondered if it ever worked... But well, now it does. The key is to point to one existing resource. The rest can be found dynamically. 

And this leads us actually to some failing tests. But I would ignore that for the moment. That is a difference. 

[0001-fix-resolving-of-paths-in-msbibexporterpath.patch.txt](https://github.com/JabRef/jabref/files/2287509/0001-fix-resolving-of-paths-in-msbibexporterpath.patch.txt)

src/test/java/org/jabref/logic/integrity/IntegrityCheckTest.java
```diff
@@ -381,7 +382,7 @@ private void assertWrong(BibDatabaseContext context) {
                 createBibtexKeyPatternPreferences(),
                 new JournalAbbreviationRepository(new Abbreviation("IEEE Software", "IEEE SW")), true)
                 .checkBibtexDatabase();
-        assertFalse(messages.toString(), messages.isEmpty());
+        assertFalse(messages.isEmpty(), messages.toString());
```
tobiasdiez: Its always better to compare values to empty collections or strings then to use `assertFalse(...isEmpty())` because you get better error message in case the test fails. Of course, you didn't write this code but please change it to `assertEquals("", messages)`.

src/test/java/org/jabref/logic/util/io/CiteKeyBasedFileFinderTest.java
```diff
@@ -106,4 +103,34 @@ public void findAssociatedFilesInNonExistingDirectoryFindsNothing() throws Excep
         assertEquals(Collections.emptyList(), results);
     }
 
+    @AfterEach
+    void deleteTempFiles() throws IOException{
```
tobiasdiez: Is this really necessary? Usually the `TempDirectory` extension takes care of file deletion.

src/test/java/org/jabref/logic/util/io/FileUtilTest.java
```diff
@@ -355,9 +356,29 @@ public void testGetLinkedDirNameDefaultFullTitle() {
                 FileUtil.createDirNameFromPattern(null, entry, fileNamePattern));
     }
 
-    private Path createTemporaryTestFile(String name) throws IOException {
-        File testFile = temporaryFolder.newFile(name);
-        Files.write(testFile.toPath(), name.getBytes(StandardCharsets.UTF_8), StandardOpenOption.APPEND);
-        return testFile.toPath();
+    @AfterEach void delete() throws IOException{
```
tobiasdiez: here also, is this really necessary?

src/test/java/org/jabref/logic/util/io/CiteKeyBasedFileFinderTest.java
Metatronwings: I think `rootDir = temporaryFolder.getParent();` is the reason why those files didn't disappear automatically. 
In fact, I can just use `deleteIfExists(rootDir);` to solve this problem, but the temp folder on my machine has some other stuff that can't be deleted (they share one temp file folder).  So I have to use this to guarantee the tests run in a correct environment.

src/test/java/org/jabref/logic/util/io/FileUtilTest.java
Metatronwings: And yes, those folders didn't delete themselves automatically. Every time I run this test, I'll get something like tens of temp folders.

src/test/java/org/jabref/logic/integrity/IntegrityCheckTest.java
tobiasdiez: Damn...ok leave it like it was before, but this is something we also have to fix in the future...

src/test/java/org/jabref/logic/util/io/CiteKeyBasedFileFinderTest.java
tobiasdiez: Good observation. In fact, you should never use the parent of the temporary folder, since a user might not even have write access to it. Can you please change the tests (here and below) so that they don't use the parent. Thanks.

src/test/java/org/jabref/logic/cleanup/MoveFilesCleanupTest.java
```diff
@@ -117,15 +125,17 @@ public void movesFileFromSubfolderWithFileDirPattern() throws IOException {
...
     @Test
-    public void movesFileFromSubfolderWithSubdirPattern() throws IOException {
+    public void movesFileFromSubfolderWithSubdirPattern(@TempDirectory.TempDir Path bibFolder) throws IOException {
         BibEntry local_entry = (BibEntry) entry.clone();
         local_entry.setField("year", "1989");
-        File subfolder = bibFolder.newFolder();
+        Path path = bibFolder.resolve("AnotherRandomlyNamedFolder");
+        Files.createDirectory(path);
+        File subfolder = path.toFile();
         File fileBefore = new File(subfolder, "test.pdf");
```
Siedlerchr: Just one more minor thing I have before we can merge.
Please use the nio methods here and in the other things as well, e.g. instead of the old File = new File use  
Files.createFile(Path path)

src/test/java/org/jabref/logic/cleanup/RenamePdfCleanupTest.java
```diff
@@ -51,31 +54,36 @@ public void setUp() throws Exception {
...
     @Test
-    public void cleanupRenamePdfRenamesFileEvenIfOnlyDifferenceIsCase() throws IOException {
+    public void cleanupRenamePdfRenamesFileEvenIfOnlyDifferenceIsCase(@TempDirectory.TempDir Path testFolder) throws IOException {
         String fileNamePattern = "[bibtexkey]";
-        File tempFile = testFolder.newFile("toot.tmp");
+        Path path = testFolder.resolve("toot.tmp");
+        Files.createFile(path);
+        File tempFile = path.toFile();
+
         LinkedFile fileField = new LinkedFile("", tempFile.getAbsolutePath(), "");
```
Siedlerchr: The path class also has a method toAbsolutePath which you can use here., e.g. just toAbsolutePath().toString()

src/test/java/org/jabref/logic/exporter/BibTeXMLExporterTestFiles.java
```diff
@@ -46,33 +42,31 @@
...
     }
 
-    @Test
-    public final void testPerformExport() throws IOException, SaveException {
+    @ParameterizedTest
+    @MethodSource("fileNames")
+    public final void testPerformExport(String filename) throws IOException, SaveException {
         String xmlFileName = filename.replace(".bib", ".xml");
         Path importFile = resourceDir.resolve(filename);
         String tempFilename = tempFile.getCanonicalPath();
```
Siedlerchr: Path has a method getFileName which you should use here

src/test/java/org/jabref/logic/cleanup/MoveFilesCleanupTest.java
Metatronwings: Are there other tests that have this issue? Or just change this test?

src/test/java/org/jabref/logic/exporter/BibTeXMLExporterTestFiles.java
Metatronwings: I'm afraid that is a naming error. The `tempFilename` doesn't represent "filename" but "an absolute path". And you can see that in the pic below. I print `tempFilename` to the console.
image
So I guess if it is better to rename this variable to something like `tempFilePath` ? And I've done that in the commit.

Metatronwings:
I'm very glad to see this issue will be resolved by us soon. Only a few more small questions: 

**1. Some tests have not been migrated yet.**
The `DBMSSynchronizerTest.java` and `SynchronizationTestSimulator.java` have not been migrated yet as there is no good way to handle chaos code after migration. (See example in `DBMSProcessorTest.java`)
Do I need to migrate them?
The `MainArchitectureTestsWithArchUnit.java` have not, too. As there is not yet a release of the new archunit.


**2. Since I didn't migrate them, how should I write the** `CHANGELOG.md` **?**

And thanks for all of you! 👍

Siedlerchr:
I just converted the two other classes with the legacy file methods to use the nio stuff.
[0001-include-main-java-in-resources-as-well.patch.txt](https://github.com/JabRef/jabref/files/2290800/0001-include-main-java-in-resources-as-well.patch.txt)

Metatronwings:
Sorry, but I must do that tomorrow morning because it's 12:00p.m. in my timezone. Thanks for the patch!

PR #4263, opened 2018-08-13
Fix "Manage journal abbreviations" dialog

tobiasdiez:
Make it possible to close the "Manage journal abbreviations" dialog using the x-button. Fixes part of https://github.com/JabRef/jabref/issues/4190.

### DIFF
--- a/src/main/java/org/jabref/gui/journals/ManageJournalAbbreviationsView.java
+++ b/src/main/java/org/jabref/gui/journals/ManageJournalAbbreviationsView.java
@@ -2,7 +2,6 @@ package org.jabref.gui.journals;
 
 import javax.inject.Inject;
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/jabref/gui/journals/ManageJournalAbbreviationsView.java
```diff
@@ -202,7 +195,7 @@ protected Void call() {
             }
         };
         new Thread(task).start();
```
Siedlerchr: DO we really need it to execute on a new thread or can't we reuse one of ours from the thread pools?

tobiasdiez: Good observation. Fixed.

PR #4266, opened 2018-08-15
Convert Part of the Issue#3861: Library-> New Entry Dialog to Javafx

1160300608:
Hello, we convert part of the Issue #3861 ,  Library-> New Entry Dialog, to Javafx. We add 2 files, fxml file for the layout of the dialog, and EntryTypeView.java for the controller of the dialog. We also delete the old file (EntryTypeDialog.java) and changed the related files to fit the new dialog. There is no big change in the logic parts.

Here is our screenshot of the dialog.

image

----


Siedlerchr:
Thanks for your work. It would be better if you move your changes to a branch to avoid conflicts with the master branch and then create a PR based on this.

src/main/java/org/jabref/gui/BasePanel.java
```diff
@@ -798,10 +798,13 @@ public BibEntry newEntry(EntryType type) {
         EntryType actualType = type;
         if (actualType == null) {
             // Find out what type is wanted.
-            final EntryTypeDialog etd = new EntryTypeDialog(frame);
+            //final EntryTypeDialog etd = new EntryTypeDialog(frame);
```
tobiasdiez: Please clean-up the code and remove all the uncommented old code.

src/main/java/org/jabref/gui/EntryType.fxml
```diff
@@ -0,0 +1,70 @@
...
+<?import java.lang.*?>
+<?import javafx.scene.control.*?>
+<?import javafx.scene.layout.*?>
+
+<DialogPane xmlns="http://javafx.com/javafx/8" xmlns:fx="http://javafx.com/fxml/1" fx:controller="org.jabref.gui.EntryTypeView">
+   <content>
+       
+      <BorderPane>             
+         <bottom>
+            <Button fx:id="cancelButton" text="Cancel" BorderPane.alignment="CENTER" />
```
tobiasdiez: Buttons in JavaFX dialogs should be added as ButtonTypes and not as normal buttons (because otherwise you cannot close the dialog using the x button). Have a look at the other FXML-files on how to do this.

src/main/java/org/jabref/gui/EntryType.fxml
```diff
@@ -0,0 +1,70 @@
...
+         </bottom>
+         <center>
+            <GridPane alignment="CENTER" BorderPane.alignment="CENTER">
+               <children>
+                  <Label text="ID type" />
```
tobiasdiez: Every text that is localized needs to be prepanded by `%`, e.g. here `text="%ID type"`

src/main/java/org/jabref/gui/EntryTypeView.java
```diff
@@ -0,0 +1,291 @@
...
+        });
+
+        idTextField.onActionProperty().addListener(event -> fetcherWorker.run());
+
+        if (basePanel.getBibDatabaseContext().isBiblatexMode()) {
+            biblatexPane.setContent(createPane(BiblatexEntryTypes.ALL));
+            vBox.getChildren().remove(bibTexPane);
+            vBox.getChildren().remove(ieeetranPane);
+            List<EntryType> customTypes = EntryTypes.getAllCustomTypes(BibDatabaseMode.BIBLATEX);
+            if (EntryTypes.getAllCustomTypes(BibDatabaseMode.BIBLATEX).isEmpty()) {
```
tobiasdiez: reuse declared variable: `if(customTypes.isEmpty())`...

src/main/java/org/jabref/gui/EntryTypeView.java
```diff
@@ -0,0 +1,291 @@
...
+            //vBox.getChildren().remove(customPane);
+            //bibTexPane.setVisible(false);
+            //ieeetranPane.setVisible(false);
+        } else {
+            bibTexPane.setContent(createPane(BibtexEntryTypes.ALL));
+            ieeetranPane.setContent(createPane(IEEETranEntryTypes.ALL));
+            vBox.getChildren().remove(biblatexPane);
+
+            List<EntryType> customTypes = EntryTypes.getAllCustomTypes(BibDatabaseMode.BIBTEX);
+            if (EntryTypes.getAllCustomTypes(BibDatabaseMode.BIBTEX).isEmpty()) {
```
tobiasdiez: here too

src/main/java/org/jabref/gui/EntryTypeView.java
```diff
@@ -0,0 +1,291 @@
...
+                customPane.setContent(createPane(customTypes));
+            }
+            //vBox.getChildren().remove(customPane);
+            //biblatexPane.setVisible(false);
+        }
+
+    }
+
+    private GridPane createPane(Collection<? extends EntryType> entries) {
+        GridPane gridpane = new GridPane();
```
tobiasdiez: It is probably better to use a [FlowPane](https://docs.oracle.com/javase/8/javafx/api/javafx/scene/layout/FlowPane.html) which automatically takes care of organizing its children in columns and rows depending on the preferred width.

src/main/java/org/jabref/gui/EntryTypeView.java
```diff
@@ -0,0 +1,291 @@
...
+        gridpane.setPadding(new Insets(4));
+        gridpane.setVgap(4);
+        // row count
+        int row = 0;
+        // col count
+        int col = 0;
+        for (EntryType entryType : entries) {
+            TypeButton entryButton = new TypeButton(entryType.getName(), entryType);
+            entryButton.setOnAction(event -> cancelHandle(event));
```
tobiasdiez: I think, the usage of a custom button class is a bit too complicated. An action like the following should work:
````
entryButton.setOnAction(event -> {
   type.set(entryType);
   close();
});
````
and change `type` to `ObjectProperty`.

Siedlerchr:
It would be really nice if you could finish your PR and fix the remaining parts. If you need any help, just ask.

Siedlerchr:
Superseeded by #4312

Issue #4148, opened 2018-06-21
Unicode to Latex normalization should not happen for the bibtex key


johann-petrak:
There is an option that allows to specified which normalisations should be carried out when saving the library, and these include translating unicode characters to the corresponding Latex command.

Apparently, jabref does this for bibtex keys as well which means a library where this happens will not be able to load back the changed entry, and the error:

     WARN  org.jabref.logic.importer.OpenDatabase - Error occurred when parsing entry: 'Error in line 26831:Character '{' is not allowed in bibtex keys.'. Skipped entry.

will be shown. It would maybe be better to not do this conversion for bibtex entries in the first place and instead just warn that there is a character in the key which should not be there in the first place?

NorwayMaple:
I created a fix for this on a new branch in my forked repo, and then realized I could not create the original problem on the master branch.  How did you find the option to normalize the library?

tobiasdiez:
It should be under Quality > Cleanup entries and/or Library > Library properties.

NorwayMaple:
Thanks!  I managed to test it and it worked.  The fix is basically a check in org.jabref.model.cleanup.FieldFormatterCleanup.java that does nothing (returns an empty ArrayList) when asked to clean up the value for KEY_FIELD. 

This would have the effect of ignoring attempts to clean up KEY_FIELD.  Would it be better to simply make KEY_FIELD impossible to modify by any means (including cleanup) after it is set, perhaps by making it final?  In other words, should we let the user change the BibTeX ID after it he/she sets it?

tobiasdiez:
That sounds like a good fix to me. At least I cannot think of a use case where the users wants to run a formatter on the key field; however, experience tells us that users always manage to surprise us and find obscure uses cases, I would say that formatters that applies to "all" fields should ignore the bibtexkey field; however, if the users really targets the key field, the cleanup should still run.

The user may edit the key field and thus marking it as final does not work.

tobiasdiez:
Thanks to @NorwayMaple this should be fixed in the latest development version. Could you please check the build from http://builds.jabref.org/master/. Thanks! Please remember to make a backup of your library before trying-out this version.

PR #4275, opened 2018-08-18
Changed the behavior of the field formatting dialog

NorwayMaple:
I made the loops for formatting all fields in the FieldFormatterCleanup class test for not being the KEY_FIELD.  I originally had this in the cleanupSingleString method to prevent KEY_FIELD from being cleaned up (reformatted) entirely, but decided to let the user specifically format that field.  When the user chooses to format all fields at once, however, the KEY_FIELD will be not be reformatted. This solves issue [#4148](https://github.com/JabRef/jabref/issues/4148)


NorwayMaple:
I created a few unit tests.  For the last one, I hard coded the LaTeX version of the "ç" character.

NorwayMaple:
My pleasure!

PR #4277, opened 2018-08-20
convert bibtexkeypatterndlg to javafx as well

Siedlerchr:
Followup from #4253  in the Bibtex key pattern panel there was an exception thrown if you had defined custom entry types or used biblatex mode.

@Super-Tang In your PR you used a fixed size for the number of labels and buttons. However, as it is possible in JabRef to customize the entry types, e.g. add new ones, the size is no longer fixed. Therefore if the number of entry types was > 19 the array threw an exception.
I now used the size of the entry type list to setup the array size and wrapped it in a Scroll Pane

Edit// I set colspan to remaining, so it uses all space
grafik

----

### DIFF
--- a/src/main/java/org/jabref/gui/actions/BibtexKeyPatternAction.java
+++ b/src/main/java/org/jabref/gui/actions/BibtexKeyPatternAction.java
@@ -13,7 +13,6 @@ public class BibtexKeyPatternAction extends SimpleCommand {
 
     @Override
     public void execute() {
-        BibtexKeyPatternDialog dlg = new BibtexKeyPatternDialog(frame.getCurrentBasePanel());
-        dlg.showAndWait();
+       new BibtexKeyPatternDialog(frame.getCurrentBasePanel()).showAndWait();
     }
 }
--- a/src/main/java/org/jabref/gui/actions/BibtexKeyPatternAction.java
+++ b/src/main/java/org/jabref/gui/actions/BibtexKeyPatternAction.java
@@ -5,7 +5,7 @@ import org.jabref.gui.bibtexkeypattern.BibtexKeyPatternDialog;
 
 public class BibtexKeyPatternAction extends SimpleCommand {
 
-    private JabRefFrame frame;
+    private final JabRefFrame frame;
 
     public BibtexKeyPatternAction(JabRefFrame frame) {
         this.frame = frame;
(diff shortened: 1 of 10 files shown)


Siedlerchr:
I decided to quickly convert the Bibtex Key Pattern Dialog as well to javafx., it is using the same panel from the pref tabs

Super-Tang:
Sorry for not thinking so much. :( @Siedlerchr

Siedlerchr:
@Super-Tang  No problem! Now you know it for the future and I bet you won't come in such a situation again ;)

tobiasdiez:
I implemented a similar fix in #4280, which goes even a step further and completely removes the textbox array. Sorry for not being quicker with my PR and the double effort. So that we don't run in a merge hell, I would propose you remove your changes to this key pattern pane class. The code looks good otherwise.

src/main/java/org/jabref/gui/actions/BibtexKeyPatternAction.java
```diff
@@ -5,16 +5,15 @@
...
     public BibtexKeyPatternAction(JabRefFrame frame) {
         this.frame = frame;
     }
 
     @Override
     public void execute() {
-        BibtexKeyPatternDialog bibtexKeyPatternDialog = new BibtexKeyPatternDialog(frame.getCurrentBasePanel());
-        bibtexKeyPatternDialog.setLocationRelativeTo(null);
-        bibtexKeyPatternDialog.setVisible(true);
+        BibtexKeyPatternDialog dlg = new BibtexKeyPatternDialog(frame.getCurrentBasePanel());
```
tobiasdiez: Please no abbreviations

src/main/java/org/jabref/gui/bibtexkeypattern/BibtexKeyPatternDialog.java
```diff
@@ -44,53 +24,20 @@ public BibtexKeyPatternDialog(BasePanel panel) {
...
-        Action cancelAction = new AbstractAction() {
-
-            @Override
-            public void actionPerformed(ActionEvent e) {
-                dialog.dispatchEvent(new WindowEvent(dialog, WindowEvent.WINDOW_CLOSING));
+        this.setResultConverter(button -> {
+            if (button == ButtonType.APPLY) {
+                metaData.setCiteKeyPattern(bibtexKeyPatternPanel.getKeyPatternAsDatabaseBibtexKeyPattern());
+                panel.markNonUndoableBaseChanged();
+            } else {
```
tobiasdiez: else branch is not really needed

src/main/java/org/jabref/gui/util/BaseDialog.java
```diff
@@ -20,7 +20,7 @@ protected BaseDialog() {
         });
 
         setDialogIcon(IconTheme.getJabRefImageFX());
-
+        setResizable(true);
```
tobiasdiez: It's a good idea to set this globally. Can you please remove the other explicit `setResizable` calls that are now no longer needed.

Siedlerchr: I removed it everywhere except in the prefs dialog to not create a conflict with your other PR

PR #4286, opened 2018-08-23
Fix Export to clipbaord action

Siedlerchr:
Fixes #4286  

The content is correctly put, but I can't paste it in Notepad++ or wherever 

----

### DIFF
--- a/src/main/java/org/jabref/gui/exporter/ExportToClipboardAction.java
+++ b/src/main/java/org/jabref/gui/exporter/ExportToClipboardAction.java
@@ -4,6 +4,7 @@ import java.io.BufferedReader;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Optional;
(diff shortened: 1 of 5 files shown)


src/main/java/org/jabref/gui/exporter/ExportToClipboardAction.java
```diff
@@ -77,29 +84,21 @@ private String exportToClipboard(Exporter exporter) {
...
-            ClipboardOwner owner = (clipboard, content) -> {
-                // Do nothing
-            };
-            ClipboardContent clipboardContent = new ClipboardContent();
-            clipboardContent.putRtf(sb.toString());
-            Globals.clipboardManager.setContent(clipboardContent);
+
+            String content = readFileToString(tmp);
+
+            DefaultTaskExecutor.runInJavaFXThread(() -> {
```
tobiasdiez: It is slightly better if you return the content here at this point and put it in the clipboard in the `onSuccess` method of the BackgroundTask (which is always executed on the JavaFX thread).

src/main/java/org/jabref/gui/exporter/ExportToClipboardAction.java
```diff
@@ -77,29 +84,21 @@ private String exportToClipboard(Exporter exporter) {
 
             // Write to file:
             exporter.export(panel.getBibDatabaseContext(), tmp,
-                    panel.getBibDatabaseContext()
-                         .getMetaData()
-                         .getEncoding()
-                         .orElse(Globals.prefs.getDefaultEncoding()),
-                    entries);
+                            panel.getBibDatabaseContext()
```
tobiasdiez: Do we really need to export to a tmp file? The exporter should be able to work on a output stream and then we can simply pass a StringOutputBuffer here. I know this may be a bigger task, so take it as a suggestion and not requirement for merge ;-)

Siedlerchr: It's out of scope and I think we have some other more important bugs to fix atm.

PR #4291, opened 2018-08-23
fix loading and storing of XMP Field Exclusions

### LINKED ISSUE #4072: Cannot add exclusions into `XMP-metadata`'s `XMP export privacy settings`
I guess we should
1 fix the exclusions mechanism
2 insert 'file' into default exclusions and enable it by default
3 stop storing files paths in bibtex

Siedlerchr:
Fixes #4072 

@koppor  I am not that into the XMP stuff, are those "fields" I can add there bibtexfields which I can exclude from the export? 
Then it would make sense to have a combobox analogous to the Cleanup panel with all fields

Edit// Decided to add the combobox with all fields

grafik

----

### DIFF
--- a/src/main/java/org/jabref/gui/preferences/XmpPrefsTab.java
+++ b/src/main/java/org/jabref/gui/preferences/XmpPrefsTab.java
@@ -12,12 +12,12 @@ import javafx.collections.FXCollections;
 import javafx.scene.Node;
 import javafx.scene.control.Button;
 import javafx.scene.control.CheckBox;
+import javafx.scene.control.ComboBox;
 import javafx.scene.control.Label;
 import javafx.scene.control.ScrollPane;
 import javafx.scene.control.TableColumn;
 import javafx.scene.control.TableColumn.CellEditEvent;
 import javafx.scene.control.TableView;
-import javafx.scene.control.TextField;
 import javafx.scene.layout.BorderPane;
 import javafx.scene.layout.GridPane;
 import javafx.scene.layout.HBox;
@@ -25,6 +25,7 @@ import javafx.scene.layout.Pane;
 
 import org.jabref.gui.util.ValueTableCellFactory;
 import org.jabref.logic.l10n.Localization;
+import org.jabref.model.entry.InternalBibtexFields;
 import org.jabref.model.strings.StringUtil;
 import org.jabref.preferences.JabRefPreferences;
 
@@ -60,10 +61,8 @@ class XmpPrefsTab extends Pane implements PrefsTab {
 
         column.setPrefWidth(350);
         tableView.getColumns().add(column);
+        ComboBox<String> bibtexFields = new ComboBox<>(FXCollections.observableArrayList(InternalBibtexFields.getAllPublicAndInternalFieldNames()));
 
-        TextField addName = new TextField();
-        addName.setPromptText("name");
-        addName.setPrefSize(200, 30);
         BorderPane tablePanel = new BorderPane();
         ScrollPane scrollPane = new ScrollPane();
         scrollPane.setMaxHeight(400);
@@ -74,9 +73,8 @@ class XmpPrefsTab extends Pane implements PrefsTab {
         Button add = new Button("Add");
         add.setPrefSize(80, 20);
         add.setOnAction(e -> {
-            if (!addName.getText().isEmpty()) {
-                XMPPrivacyFilter tableRow = new XMPPrivacyFilter(addName.getText());
-                addName.clear();
+            if (!StringUtil.isNullOrEmpty(bibtexFields.getSelectionModel().getSelectedItem())) {
+                XMPPrivacyFilter tableRow = new XMPPrivacyFilter(bibtexFields.getSelectionModel().getSelectedItem());
                 fields.add(tableRow);
             }
         });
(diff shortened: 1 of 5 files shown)


Siedlerchr:
I am going to merge this now if no further comments are coming ;)

src/main/java/org/jabref/gui/preferences/XmpPrefsTab.java
```diff
@@ -68,34 +72,22 @@ public XmpPrefsTab(JabRefPreferences prefs) {
...
+                fields.add(tableRow);
             }
         });
         Button delete = new Button("Delete");
-        delete.setPrefSize(80, 20);
-        delete.setOnAction(e-> {
-            if (tableView.getFocusModel() != null && tableView.getFocusModel().getFocusedIndex() != -1) {
-                tableChanged = true;
+        delete.setOnAction(e -> {
+            if ((tableView.getFocusModel() != null) && (tableView.getFocusModel().getFocusedIndex() != -1)) {
```
tobiasdiez: Why do you use the focus and not selected item ala `tableView.getSelectionModel().getSelectedItem()` ?

src/main/java/org/jabref/gui/preferences/XmpPrefsTab.java
```diff
@@ -107,35 +99,20 @@ public XmpPrefsTab(JabRefPreferences prefs) {
...
      * Load settings from the preferences and initialize the table.
      */
     @Override
     public void setValues() {
-        tableRows.clear();
-        tableRows.addAll(data);
-        privacyFilterCheckBox.setSelected(JabRefPreferences.getInstance().getBoolean(
-                JabRefPreferences.USE_XMP_PRIVACY_FILTER));
+        List<XMPPrivacyFilter> xmpExclusions = prefs.getStringList(JabRefPreferences.XMP_PRIVACY_FILTERS).stream().map(XMPPrivacyFilter::new).collect(Collectors.toList());
+        fields.clear();
```
tobiasdiez: I'm not sure, but I thought there is a `setAll` method (instead of clear + addAll)

src/main/java/org/jabref/gui/preferences/XmpPrefsTab.java
```diff
@@ -174,4 +138,30 @@ public boolean validateSettings() {
...
+
+        public void setField(String field) {
+            this.field.set(field);
+        }
+
+        public String getField() {
+            return field.get();
+        }
+
+        public StringProperty fieldName() {
```
tobiasdiez: should be `field()`

src/test/java/org/jabref/logic/exporter/XmpExporterTest.java
```diff
@@ -55,10 +57,23 @@ public void exportSingleEntry(@TempDirectory.TempDir Path testFolder) throws Exc
         entry.setField("author", "Alan Turing");
 
         exporter.export(databaseContext, file, encoding, Collections.singletonList(entry));
-
-        List<String> lines = Files.readAllLines(file);
-        assertEquals(15, lines.size());
-        assertEquals("<rdf:li>Alan Turing</rdf:li>", lines.get(4).trim());
+        String actual = Files.readAllLines(file).stream().collect(Collectors.joining("\n")); //we are using \n to join, so we need it in the expected string as well, \n would fail
```
tobiasdiez: I don't understand the comment, especially the `\n would fail` part. You can also simply compare the file contents as arrays, then there is no ambiguity concerning line endings.

src/main/java/org/jabref/gui/preferences/XmpPrefsTab.java
```diff
@@ -68,34 +72,22 @@ public XmpPrefsTab(JabRefPreferences prefs) {
...
+                XMPPrivacyFilter tableRow = new XMPPrivacyFilter(bibtexFields.getSelectionModel().getSelectedItem());
+                fields.add(tableRow);
             }
         });
         Button delete = new Button("Delete");
-        delete.setPrefSize(80, 20);
-        delete.setOnAction(e-> {
-            if (tableView.getFocusModel() != null && tableView.getFocusModel().getFocusedIndex() != -1) {
-                tableChanged = true;
+        delete.setOnAction(e -> {
```
tobiasdiez: The delete button should be shown in the table per row on hover (e.g. similar to the buttons in the maintable and I guess manage journal abbreviations and/or key bindings)

src/main/java/org/jabref/gui/preferences/XmpPrefsTab.java
```diff
@@ -174,4 +138,30 @@ public boolean validateSettings() {
     public String getTabName() {
         return Localization.lang("XMP-metadata");
     }
+
+    private class XMPPrivacyFilter {
+
+        private final SimpleStringProperty field;
```
tobiasdiez: is it really necessary to make this a property instead of a simple get/set java prop? Moreover, I like the idea of creating a new class for this but would go one step further: use it everywhere (i.e. the properties class should set/get a list of privacy filter).

src/test/java/org/jabref/logic/exporter/XmpExporterTest.java
Siedlerchr: The comment should have been `\r\n` would fail.
I prefer this equals variant because at least eclipse shows the differences highlighted

src/main/java/org/jabref/gui/preferences/XmpPrefsTab.java
Siedlerchr: I understand your idea but it is necessary as the cellValueFactory expects an Observable.
Another idea would be to have a converter/PropertyValueFactory. 
 lastNameCol.setCellValueFactory(new PropertyValueFactory<Person,String>("lastName"));

Siedlerchr:
grafik

PR #4302, opened 2018-08-30
Convert to circle CI 2.0

Siedlerchr:
I just found that @koppor  already had a  branch and configuration for circle CI 2.0

----


.circleci/config.yml
```diff
@@ -0,0 +1,79 @@
...
+  downloadinstall4j:
+    docker:
+      - image: circleci/openjdk:8-jdk
+    steps:
+      - restore_cache:
+          keys:
+            - install4j
+      - checkout
+      - run: scripts/download-install4j-and-jres.sh
+      - save_cache:
```
tobiasdiez: add `- run: scripts/extract-install4j.sh` already here? Then you also don't need to cache the `downloads` folder below.

Siedlerchr: No idea about it and I don't have time to figure that out. And as of today we need it working I won't make any experiments with it. Maybe you should create an issue  at koppors repo

PR #4309, opened 2018-09-01
Completely rework save operation

tobiasdiez:
This is a complete rework of the file writing logic. Since this is the most essential part of JabRef, I would kindly ask to every @JabRef/developers to have a look at the code (at least on the new `AtomicOutputStream` which does the hard work).

The new strategy for saving the bib file is as follows:
- Create a backup (with .bak suffix) of the original file in the same directory as the destination file (if set in preferences).
- Write to a temporary file (with .tmp suffix) in the same directory.
- Create a save-backup (with .sav suffix) of the original file (if it exists) in the same directory.
- Move the temporary file to the correct place, overwriting any file that already exists at that location.
- Delete the save-backup file.

What is not clear to me is how one should handle (almost) concurrent saving processes of different users to the same file. Is it enough to lock a file (`nio.FileLock`) and if yes, which (the tmp?)?

----


src/main/java/org/jabref/logic/exporter/AtomicFileOutputStream.java
```diff
@@ -0,0 +1,197 @@
...
+        } catch (IOException exception) {
+            LOGGER.debug("Unable to abort writing to file " + temporaryFile, exception);
+        }
+    }
+
+    private void cleanup() {
+        try {
+            Files.deleteIfExists(temporaryFile);
+        } catch (IOException exception) {
+            LOGGER.debug("Unable to delete file " + temporaryFile, exception);
```
Siedlerchr: For arguments in the logger you can simply use a pair of curly braces {} instead of concat with plus. 
https://www.slf4j.org/faq.html#logging_performance
Here and in above

src/main/java/org/jabref/logic/exporter/AtomicFileOutputStream.java
```diff
@@ -0,0 +1,197 @@
...
+                    try {
+                        oldFilePermissions = Files.getPosixFilePermissions(targetFile);
+                    } catch (IOException exception) {
+                        LOGGER.warn("Error getting file permissions for file {}.", targetFile, exception);
+                    }
+                }
+            }
+
+            // Move temporary file (replace original if it exists)
+            Files.move(temporaryFile, targetFile, StandardCopyOption.ATOMIC_MOVE, StandardCopyOption.REPLACE_EXISTING);
```
Siedlerchr: We have a move in FileUtil. And you better check if the move was successufl

Siedlerchr:
Yeah, I generally like that you tackle this complex stuff! 👍  We should really test this carefully and also check how it interacts with the FileUpdateMonitor.   
Some TemplateExporter tests fail, seems to be related to the resource closing

koppor:
Tobias Diez <notifications@github.com> schrieb am So., 2. Sep. 2018, 10:58:

> This is a complete rework of the file writing logic. Since this is the
> most essential part of JabRef, I would kindly ask to every
> @JabRef/developers <https://github.com/orgs/JabRef/teams/developers> to
> have a look at the code (at least on the new AtomicOutputStream which
> does the hard work).
>

Thank you for stepping in!

The new strategy for saving the bib file is as follows:
>
>    - Create a backup (with .bak suffix) of the original file in the same
>    directory as the destination file (if set in preferences).
>
> This is from the early days of JabRef. Why not creating it in %APPDATA% or
/tmp? It confused much of our users.



>    - Write to a temporary file (with .tmp suffix) in the same directory.
>
>
Why not writing to the same file? The only risk is out-of-space, isn't?
When copying things around, we'll have issues with Unix access rights and
ownerships on NFS shares.


>    - Create a save-backup (with .sav suffix) of the original file (if it
>    exists) in the same directory.
>
>
What is the diff to .bak?


>    - Move the temporary file to the correct place, overwriting any file
>    that already exists at that location.
>    - Delete the save-backup file.
>
> What is not clear to me is how one should handle (almost) concurrent
> saving processes of different users to the same file. Is it enough to lock
> a file (nio.FileLock) and if yes, which (the tmp?)?
>

Proposal: lock the .bib and write to it.

tobiasdiez:
Thanks @koppor for the valuable input (especially for pointing out that the `bak` and `sav` file are the same). The idea of writing to a temporary file is to make sure that any point in time the `bib` file is valid and un-damaged, even in cases of powerloss, drive disconnect etc.

Siedlerchr:
As I experienced this first hand in another project, a locking of the file is practically useless on Unix systems. Even if you grab a write lock, it doesn't stop other processes from accessing the file or writing to it. That's because file locks on Unix are only advisory, e.g.just a signal to other process.
https://docs.oracle.com/javase/7/docs/api/java/nio/channels/FileLock.html

src/main/java/org/jabref/logic/exporter/AtomicFileOutputStream.java
tobiasdiez: This is not possible here since the there is no appropriate debug method that also accepts an exception.

src/main/java/org/jabref/logic/exporter/AtomicFileOutputStream.java
tobiasdiez: Yes, good point - I don't understand why we have a own move method that actually looks more like a "read and rewrite content"-method. Moreover, the move is successful if there is no thrown exception (I assume).

tobiasdiez:
I included all the feedback from @koppor and @Siedlerchr. Thanks for the good suggestions!
Since this PR has been in "ready-for-review" for two weeks now, I'll merge now.

PR #4312, opened 2018-09-02
Convert New entry dialog to javafx

Siedlerchr:
Followup from  #4266

bibtex +custom 
grafik

biblatex + custom
grafik

----


src/main/java/org/jabref/gui/EntryTypeView.java
```diff
@@ -0,0 +1,256 @@
...
+                        new BibtexKeyGenerator(basePanel.getBibDatabaseContext(), prefs.getBibtexKeyPatternPreferences()).generateAndSetKey(bibEntry);
+                        // Update Timestamps
+                        if (prefs.getTimestampPreferences().includeCreatedTimestamp()) {
+                            bibEntry.setField(prefs.getTimestampPreferences().getTimestampField(), prefs.getTimestampPreferences().now());
+                        }
+                        basePanel.insertEntry(bibEntry);
+                    }
+
+                    close();
+                } else if (searchID.trim().isEmpty()) {
```
tobiasdiez: better disable the generate button if the search field is empty (and use StringUtil.isBlank)

src/main/java/org/jabref/gui/EntryTypeView.java
```diff
@@ -0,0 +1,256 @@
...
+                    dialogService.showWarningDialogAndWait(Localization.lang("Empty search ID"), Localization.lang("The given search ID was empty."));
+                } else if (!fetcherException) {
+                    dialogService.showErrorDialogAndWait(Localization.lang("No files found.", Localization.lang("Fetcher '%0' did not find an entry for id '%1'.", fetcher.getName(), searchID) + "\n" + fetcherExceptionMessage));
```
tobiasdiez: "No files found" is a strange header.

src/main/java/org/jabref/gui/EntryTypeView.java
```diff
@@ -0,0 +1,256 @@
...
+            Optional<BibEntry> bibEntry = Optional.empty();
+
+            viewModel.searchingProperty().setValue(true);
+            viewModel.storeSelectedFetcher();
+            fetcher = viewModel.selectedItemProperty().getValue();
+            searchID = idTextField.getText();
+            if (!searchID.isEmpty()) {
+                try {
+                    bibEntry = fetcher.performSearchById(searchID);
+                } catch (FetcherException e) {
```
tobiasdiez: There is no need to catch exceptions here. The Task class has already ways built-in to handle exceptions and signal errors (you need to overwrite a different method to handle exceptions). It might be easier to just use the `BackgroundTask` wrapper.

src/main/java/org/jabref/gui/BasePanel.java
```diff
@@ -772,10 +772,13 @@ public BibEntry newEntry(EntryType type) {
         EntryType actualType = type;
         if (actualType == null) {
             // Find out what type is wanted.
-            final EntryTypeDialog etd = new EntryTypeDialog(frame);
+            //final EntryTypeDialog etd = new EntryTypeDialog(frame);
```
tobiasdiez: Is this method even used with null argument? At least the code below suggests that the JabRefFrame uses the NewEntryAction and does not calls this method.

src/main/java/org/jabref/gui/actions/NewEntryFromPlainTextAction.java
```diff
@@ -31,13 +37,15 @@ public void execute() {
             return;
         }
 
-        EntryTypeDialog typeChoiceDialog = new EntryTypeDialog(jabRefFrame);
-        typeChoiceDialog.setVisible(true);
+        //EntryTypeDialog typeChoiceDialog = new EntryTypeDialog(jabRefFrame);
```
tobiasdiez: remove old code

src/main/java/org/jabref/pdfimport/PdfImporter.java
```diff
@@ -224,8 +224,10 @@ private void doContentImport(String fileName, List<BibEntry> res) {
 
     private Optional<BibEntry> createNewEntry() {
         // Find out what type is desired
-        EntryTypeDialog etd = new EntryTypeDialog(frame);
-        etd.setVisible(true);
+        //EntryTypeDialog etd = new EntryTypeDialog(frame);
```
tobiasdiez: here too

src/main/java/org/jabref/pdfimport/PdfImporter.java
```diff
@@ -224,8 +224,10 @@ private void doContentImport(String fileName, List<BibEntry> res) {
...
+        EntryTypeView etd = new EntryTypeView(frame.getCurrentBasePanel(), frame.getDialogService(), Globals.prefs);
+        //etd.setVisible(true);
+        etd.showAndWait();
```
tobiasdiez: The code below looks like yet another place with code duplication (which is however slightly different to the one above, eg. timestamp generation).

src/main/java/org/jabref/preferences/JabRefPreferences.java
```diff
@@ -1472,13 +1472,13 @@ public ExporterFactory getExporterFactory(JournalAbbreviationLoader abbreviation
...
     public BibtexKeyPatternPreferences getBibtexKeyPatternPreferences() {
         return new BibtexKeyPatternPreferences(
-                get(KEY_PATTERN_REGEX),
-                get(KEY_PATTERN_REPLACEMENT),
-                getBoolean(KEY_GEN_ALWAYS_ADD_LETTER),
-                getBoolean(KEY_GEN_FIRST_LETTER_A),
-                getBoolean(ENFORCE_LEGAL_BIBTEX_KEY),
-                getKeyPattern(),
-                getKeywordDelimiter());
+                                               get(KEY_PATTERN_REGEX),
```
tobiasdiez: the code formatting was ok before...this indention looks ugly to be honest.

src/main/java/org/jabref/gui/EntryTypeView.java
```diff
@@ -0,0 +1,256 @@
...
+    private final BasePanel basePanel;
+    private final DialogService dialogService;
+    private final JabRefPreferences prefs;
+
+    private EntryType type;
+    private Task<Optional<BibEntry>> fetcherWorker = new FetcherWorker();
+    private EntryTypeViewModel viewModel;
+
+    public EntryTypeView(BasePanel basePanel, DialogService dialogService, JabRefPreferences preferences) {
+        this.basePanel = basePanel;
```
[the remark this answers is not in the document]
Siedlerchr: Nope, the basePanel.insertEntry is called which does a few things, e.g. mark the database changed, create undoable entry etc

src/main/java/org/jabref/gui/EntryTypeView.java
```diff
@@ -0,0 +1,256 @@
...
+        btnGenerate.textProperty().bind(EasyBind.map(viewModel.searchingProperty(), searching -> (searching) ? Localization.lang("Searching...") : Localization.lang("Generate")));
+        btnGenerate.disableProperty().bind(viewModel.searchingProperty());
+
+    }
+
+    private void addEntriesToPane(FlowPane pane, Collection<? extends EntryType> entries) {
+
+        for (EntryType entryType : entries) {
+            Button entryButton = new Button(entryType.getName());
+            entryButton.setUserData(entryType);
```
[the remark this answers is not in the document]
Siedlerchr: Nope, the fxml controller expects exactly one argument for the method

src/main/java/org/jabref/gui/BasePanel.java
Siedlerchr: I just found out that this method and the insertEntry method are doing roughly the same, I think it makes sense to combine their functionaltiy

src/main/java/org/jabref/gui/EntryTypeView.java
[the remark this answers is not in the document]
tobiasdiez: I think, this method is not even called from the fxml.

Siedlerchr:
Adressed all comments, The Duplicate check is actually different handled and is slightly coupled to the import inspection dialog. This is stuff for a new PR.

Siedlerchr:
I would like to merge this now, if no further comments are added.,

PR #4313, opened 2018-09-02
[WIP] Make dark theme selectable in preferences

ngrant123:
----

darkthemeprefbox
darkthemeoption

This pull request allows the user to pick the dark theme option in the preferences box. After the user restarts the application the UI would have the dark theme in it. I also added the dark theme in the preferences box. An critique on my coding will be greatly appreciated.

### DIFF
--- a/src/main/java/org/jabref/gui/FXDialogService.java
+++ b/src/main/java/org/jabref/gui/FXDialogService.java
@@ -64,6 +64,7 @@ public class FXDialogService implements DialogService {
         alert.setHeaderText(null);
         alert.setContentText(content);
         alert.getDialogPane().setMinHeight(Region.USE_PREF_SIZE);
+        alert.showAndWait();
         return alert;
     }
 
--- a/src/main/java/org/jabref/gui/preferences/AppearancePrefsTab.java
+++ b/src/main/java/org/jabref/gui/preferences/AppearancePrefsTab.java
@@ -2,6 +2,8 @@ package org.jabref.gui.preferences;
 
 import javafx.geometry.Insets;
 import javafx.scene.Node;
+import javafx.scene.control.Alert;
+import javafx.scene.control.Alert.AlertType;
 import javafx.scene.control.CheckBox;
 import javafx.scene.control.Label;
 import javafx.scene.control.TextField;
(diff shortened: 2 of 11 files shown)


src/main/java/org/jabref/gui/preferences/AppearancePrefsTab.java
```diff
@@ -10,58 +12,151 @@
...
+
+            } catch (IOException e) {
+                e.printStackTrace();
+            }
+            accessDarkTheme.setSelected(true);
+        }
+
+    }
+
+    private void copy(File file1,File file2) throws IOException {
```
Siedlerchr: You can use the copy method from FileUtil

src/main/java/org/jabref/gui/FXDialogService.java
```diff
@@ -64,6 +64,7 @@ private static FXDialog createDialog(AlertType type, String title, String conten
         alert.setHeaderText(null);
         alert.setContentText(content);
         alert.getDialogPane().setMinHeight(Region.USE_PREF_SIZE);
+        alert.showAndWait();
```
Siedlerchr: This is not a good idea to add it here as is the internal method for creating a dialog. Normally you call one of the public method from the Dialog Service, get a dialog object back and then call show and wait on that object. Just look in other places how it's done

src/main/java/org/jabref/gui/preferences/AppearancePrefsTab.java
```diff
@@ -10,58 +12,151 @@
...
         return container;
     }
 
+    //Basically replace each of the files with each other
+    //Ultimately you have to read the contents of the file and replace it
+    private void addLogic(CheckBox accessDarkTheme) throws URISyntaxException {
+
+        if(accessDarkTheme.isSelected()==true){
+
+            File basecaseFile= new File(DEFAULT_PATH_MAIN_CSS);
```
Siedlerchr: Please use the nio Path methods here. Have a look at the exporter tests to see how to access a resource file

src/main/java/org/jabref/gui/preferences/AppearancePrefsTab.java
```diff
@@ -10,58 +12,151 @@
...
+            File darkthemecssFile= new File(DARK_THEME_PATH_CSS);
```
Siedlerchr: I don't understand why the copy and replace is necessary. I thought it's just pointing the JavaFX loader to the new path 🤔

src/main/java/org/jabref/gui/util/ThemeLoader.java
```diff
@@ -51,6 +55,9 @@ public ThemeLoader(FileUpdateMonitor fileUpdateMonitor) {
      */
     public void installBaseCss(Scene scene, JabRefPreferences preferences) {
         addAndWatchForChanges(scene, DEFAULT_PATH_MAIN_CSS, 0);
+        addAndWatchForChanges(scene,DARK_THEME_PATH_CSS,0);
```
tobiasdiez: It should work without copying (and modifying) the css files if you add here something like:
````
if (preferences.useDarkTheme) {
   addAndWatchForChanges(scene,DARK_THEME_PATH_CSS, 1);
}
````

stefan-kolb:
I would rather habe a dropdown to select the two different styles than a checkbox for the dark theme.

ngrant123:
Cool I'll look at everyone's advice and suggestions and try to fix it.

src/main/java/org/jabref/gui/util/ThemeLoader.java
Siedlerchr: Do we really need to watch for changes? Maybe just have a second method which just adds/loads the css and doesn't  watch for changes.

tobiasdiez: It only watches for changes if the corresponding command line flag is passed, i.e. only for us developers.

tobiasdiez:
https://github.com/JabRef/jabref/pull/4372 was a bit quicker to implement this. Sorry!

Issue #4223, opened 2018-07-21
Readd notifications for important actions (like save status)


bernhard-kleine:
- [ ] I have tested the latest master version from http://builds.jabref.org/master/ and the problem persists

JabRef 5.0-dev--snapshot--2018-07-15--master--a44d31668
Windows 7 6.1 amd64 
Java 1.8.0_181

I have got to rely on the status line to know that the library has been saved. In the snapshot 5.0 there is no status line left. Please let me know to re-enable it or make it available again.

tobiasdiez:
I'm not a big fan of the old status line, especially when it is used to communicate errors to the user. In most cases, displaying a message right in the context of the action is preferable (e.g. a small flyout of the button that triggered the action). Only for actions that are triggered trough the menu, something like a global status line makes sense (because the menu vanishes after the action is triggered and hence there is no context to show the result). But then again, if the information is important enough to show it to the user, a status line has usual not the visual weight to ensure that the user really reads the message. I would use something like the [NotificationPane](http://fxexperience.com/controlsfx/features/#notifications) in such cases (and remove most of the other status line messages).

Siedlerchr:
I am very much in favor of the Notficiation pane or either as the other option, a notification "window" at the right of the desktop (jdownloader 2 does it that way, with a notifaction window which can be moved.

tobiasdiez:
This is now implemented in the latest development version. Could you please check the build from http://builds.jabref.org/master/. Thanks! Please remember to make a backup of your library before trying-out this version.

PR #4314, opened 2018-09-03
Readd status line in form of a snack bar

tobiasdiez:
Bring back the status line in the following form:
image
Fixes https://github.com/JabRef/jabref/issues/4223.
Since these notifications have a rather great visual weight, we may revisit some of the existing notifications that are trivial/not useful and remove them. 

As I've learned these things are called snack bars and the notifications toasts. In this sense: 🍻 

----

### DIFF
--- a/src/main/java/org/jabref/gui/JabRefFrame.java
+++ b/src/main/java/org/jabref/gui/JabRefFrame.java
@@ -32,7 +32,6 @@ import javafx.scene.control.Alert;
 import javafx.scene.control.Button;
 import javafx.scene.control.ButtonBar;
 import javafx.scene.control.ButtonType;
-import javafx.scene.control.Label;
 import javafx.scene.control.Menu;
 import javafx.scene.control.MenuBar;
 import javafx.scene.control.ProgressBar;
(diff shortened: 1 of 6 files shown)


stefan-kolb:
Looks interesting. Can we switch back to a normal bar easily if we decide otherwise?

tobiasdiez:
Yes, going back to the old status line consists essentially of replacing the new `Snackbar` control by a good-old Label.

src/main/java/org/jabref/gui/JabRefFrame.java
```diff
@@ -832,27 +831,26 @@ private MenuBar createMenu() {
         }
 
         edit.getItems().addAll(
-                               factory.createMenuItem(StandardActions.MANAGE_KEYWORDS, new ManageKeywordsAction(this)),
-                               factory.createMenuItem(StandardActions.REPLACE_ALL, new OldDatabaseCommandWrapper(Actions.REPLACE_ALL, this, Globals.stateManager)),
-                               factory.createMenuItem(StandardActions.MASS_SET_FIELDS, new MassSetFieldAction(this))
+                factory.createMenuItem(StandardActions.MANAGE_KEYWORDS, new ManageKeywordsAction(this)),
```
Siedlerchr: This will create again merge conflicts with the indentation

tobiasdiez: Yes, I think, we need to revisist the eclipse formatting rules. Having such a huge indention for method arguments is really ugly.

